PR #880, opened 2017-06-15
Adding second scenario where IPv6 is not supported

missingcharacter:
When trying to create a security group in AWS CN I ran into the following error:

```
* module.security_groups.aws_security_group.dev-https-in: 1 error(s) occurred:

* aws_security_group.dev-https-in: Error revoking default IPv6 egress rule for Security Group (sg-56489bc5): InvalidParameterValue: remote-ipv6-range
	status code: 400, request id: beeb90dd-7b32-4905-9b94-59e3e00f77f5
```

Added this check as another way to determine IPv6 is not supported therefore ignore the error and continue.
I already tested with the same case that gave me these errors and I no longer get them and my security groups are created.

I'm not sure the code is exactly how it should be, I'll be more than happy to modify it if needed.

### DIFF
--- a/aws/resource_aws_security_group.go
+++ b/aws/resource_aws_security_group.go
@@ -316,7 +316,7 @@ func resourceAwsSecurityGroupCreate(d *schema.ResourceData, meta interface{}) er
 		if err != nil {
 			//If we have a NotFound or InvalidParameterValue, then we are trying to remove the default IPv6 egress of a non-IPv6
 			//enabled SG
-			if ec2err, ok := err.(awserr.Error); ok && (ec2err.Code() != "InvalidPermission.NotFound" || ec2err.Code() != "InvalidParameterValue: remote-ipv6-range") {
+			if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() != "InvalidPermission.NotFound" && ec2err.Code() != "InvalidParameterValue" {
 				return fmt.Errorf(
 					"Error revoking default IPv6 egress rule for Security Group (%s): %s",
 					d.Id(), err)
--- a/aws/resource_aws_security_group.go
+++ b/aws/resource_aws_security_group.go
@@ -314,9 +314,9 @@ func resourceAwsSecurityGroupCreate(d *schema.ResourceData, meta interface{}) er
 
 		_, err = conn.RevokeSecurityGroupEgress(req)
 		if err != nil {
-			//If we have a NotFound, then we are trying to remove the default IPv6 egress of a non-IPv6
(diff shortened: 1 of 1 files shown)


missingcharacter:
Looks like I didn't fix my problem, with this, I have to run terraform twice with this code change, I'll be posting an update once I'm able to make this work.

missingcharacter:
ok, I'm now confident this works, let me know if I should change anything.

radeksimko:
Hi @missingcharacter 
thanks for the PR.

While we do have some AWS CN specific code, it's always scoped explicitly to that region, so it's safe to assume it will only affect users of that region.

We (HashiCorp) do not have access to the Chinese AWS region, unfortunately, so we can't reliably test this - which is the main reason this is hanging here. Is this something you could potentially help us with?

An acceptable alternative is pasting full HTTP response (available from [debug log](https://www.terraform.io/docs/internals/debugging.html)) and decide how to tighten the condition to something like `isAWSErr("InvalidParameterValue", "specific IPv6 err message")`. The error code seems to be a bit too generic and there's risk of ignoring errors we don't intend to ignore.

Admittedly `InvalidPermission.NotFound` code match is already quite generic and brings some risk.

timonwong:
FYI: The HTTP 400 bad request returned from AWS CN is:
```xml
<Response>
<Errors>
<Error><Code>InvalidParameterValue</Code><Message>remote-ipv6-range</Message></Error></Errors>
<RequestID>xxxxxxxxxxxxxxxxxxxxxxxxxxx</RequestID>
</Response>
```

Ninir:
Hi @timonwong 

Is `remote-ipv6-range` the exact message that is returned by AWS or actually a placeholder?
If it's a placeholder, we should check that this message is actually in the form of `2620:0:2d0:200::7/32`, etc.

timonwong:
Hi @Ninir, it's the exact message returned from AWS CN.

Ninir:
Then as Radek exposed it, can you update this work using:

```go
isAWSErr("InvalidParameterValue", "remote-ipv6-range")
```

to handle this case?

Thanks!

missingcharacter:
Sorry for the long delay on these changes, made the change and tested, this time I kept `TF_LOG=DEBUG` and saved the logs in case you need them.

missingcharacter:
@Ninir yes, this fixes my issue too

Ninir:
Thank you @missingcharacter ! :)

hoffmannliu-ayla:
Hi,
When will new AWS provider including this PR available ? I ran into this issue too....:'(

Issue #856, opened 2017-06-14
Caching on API Gateway does not work for path parameters


betabandido:
When enabling caching on an API that uses path parameters (or query strings) it is necessary to create a cache key for the parameter (or query string). But, Terraform currently does not seem to allow to set cache keys. If caching is enabled for a given method (something currently possible to do with Terraform), the same cached value will be returned no matter whether the path parameter value is the same or not.

In `resourceAwsApiGatewayIntegrationCreate` method the method integration is created but both `CacheKeyParameters` and `CacheNamespace` are set to `nil`:

```
	_, err := conn.PutIntegration(&apigateway.PutIntegrationInput{
		HttpMethod: aws.String(d.Get("http_method").(string)),
		ResourceId: aws.String(d.Get("resource_id").(string)),
		RestApiId:  aws.String(d.Get("rest_api_id").(string)),
		Type:       aws.String(d.Get("type").(string)),
		IntegrationHttpMethod: integrationHttpMethod,
		Uri:                 uri,
		RequestParameters:   aws.StringMap(parameters),
		RequestTemplates:    aws.StringMap(templates),
		Credentials:         credentials,
		CacheNamespace:      nil,
		CacheKeyParameters:  nil,
		PassthroughBehavior: passthroughBehavior,
		ContentHandling:     contentHandling,
	})
```
(see code at https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_api_gateway_integration.go#L153)

Additionally, when trying to enable caching for a path parameter in the AWS console (after an API has been created by Terraform) the operation fails. This might be related to using `nil` in the `PutIntegration` call.

We are working on a pull request to provide a working solution. Any help would be appreciated.

### Terraform Version

Terraform v0.9.8

The issue seems to be present in the master branch too.

### Affected Resource(s)
aws_api_gateway_integration

betabandido:
Enabling caching does **not** properly work even for very simple use cases (such as when having path parameters or query strings). So, IMHO I would not say this is just an "enhancement."

We have created a pull request to fix this issue. Hopefully it can make it into the next release :)

Ninir:
Hi @betabandido 

As your PR was merged, and mine about [the documentation](https://github.com/terraform-providers/terraform-provider-aws/pull/1602/files) will be too soon I hope, could you tell me if you are still having this issue or is it all resolved?

Thanks!

betabandido:
@Ninir 

[...]

PR #893, opened 2017-06-16
Add support for specifying cache key parameters in API Gateway.

betabandido:
Pull request to fix issue #856 

This pull request provides a way to specify cache key parameters, so that caching can take into account varying path parameters or query strings.

I have tested the changes, and I was able to effectively create, read or update cache key parameters. The following (adapted) example shows how to do so for path parameters:

```
resource "aws_api_gateway_resource" "path1" {
  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
  parent_id   = "${aws_api_gateway_rest_api.api.root_resource_id}"
  path_part   = "{param1}"
}

resource "aws_api_gateway_resource" "path2" {
  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
  parent_id   = "${aws_api_gateway_resource.path1.id}"
  path_part   = "{param2}"
}

resource "aws_api_gateway_method" "method" {
  rest_api_id   = "${aws_api_gateway_rest_api.api.id}"
  resource_id   = "${aws_api_gateway_resource.path2.id}"
  http_method   = "GET"
  authorization = "NONE"
  request_parameters = {
      "method.request.path.param1" = true
      "method.request.path.param2" = true
  }
}

resource "aws_api_gateway_integration" "integration" {
  rest_api_id   = "${aws_api_gateway_rest_api.api.id}"
  resource_id   = "${aws_api_gateway_resource.path2.id}"
  http_method   = "${aws_api_gateway_method.method.http_method}"
  type          = "AWS"
  uri           = "arn:aws:apigateway:${var.region}:lambda:path/2015-03-31/functions/${var.lambda_arn}/invocations"
  integration_http_method = "POST"
  request_parameters = {
      "integration.request.path.param1" = "method.request.path.param1"
      "integration.request.path.param2" = "method.request.path.param2"
  }
  cache_key_parameters = ["method.request.path.param1", "method.request.path.param2"]
}
```

The example shows how to set the new field (`cache_key_parameters`) as well as `request_parameters` (both in the method and integration resources).

### DIFF
--- a/aws/resource_aws_api_gateway_integration.go
+++ b/aws/resource_aws_api_gateway_integration.go
@@ -101,6 +101,13 @@ func resourceAwsApiGatewayIntegration() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validateApiGatewayIntegrationPassthroughBehavior,
 			},
+
+			"cache_key_parameters": {
+				Type:     schema.TypeSet,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+				Optional: true,
+			},
 		},
 	}
 }
@@ -150,6 +157,11 @@ func resourceAwsApiGatewayIntegrationCreate(d *schema.ResourceData, meta interfa
 		contentHandling = aws.String(val.(string))
 	}
 
+	var cacheKeyParameters []*string
+	if v, ok := d.GetOk("cache_key_parameters"); ok {
+		cacheKeyParameters = expandStringList(v.(*schema.Set).List())
+	}
+
 	_, err := conn.PutIntegration(&apigateway.PutIntegrationInput{
 		HttpMethod: aws.String(d.Get("http_method").(string)),
 		ResourceId: aws.String(d.Get("resource_id").(string)),
@@ -160,8 +172,8 @@ func resourceAwsApiGatewayIntegrationCreate(d *schema.ResourceData, meta interfa
 		RequestParameters:   aws.StringMap(parameters),
 		RequestTemplates:    aws.StringMap(templates),
 		Credentials:         credentials,
-		CacheNamespace:      nil,
-		CacheKeyParameters:  nil,
+		CacheNamespace:      aws.String(d.Get("resource_id").(string)),
+		CacheKeyParameters:  cacheKeyParameters,
 		PassthroughBehavior: passthroughBehavior,
 		ContentHandling:     contentHandling,
 	})
@@ -215,6 +227,8 @@ func resourceAwsApiGatewayIntegrationRead(d *schema.ResourceData, meta interface
 	if integration.ContentHandling != nil {
 		d.Set("content_handling", integration.ContentHandling)
 	}
+
+	d.Set("cache_key_parameters", flattenStringList(integration.CacheKeyParameters))
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


aws/resource_aws_api_gateway_integration.go
```diff
@@ -160,8 +172,8 @@ func resourceAwsApiGatewayIntegrationCreate(d *schema.ResourceData, meta interfa
 		RequestParameters:   aws.StringMap(parameters),
 		RequestTemplates:    aws.StringMap(templates),
 		Credentials:         credentials,
-		CacheNamespace:      nil,
-		CacheKeyParameters:  nil,
+		CacheNamespace:      aws.String(d.Get("resource_id").(string)),
```
radeksimko: Is there any reason the cache namespace should always be resource ID?
I had a peek 👀  into the official API docs and couldn't find any suggestions around that: https://docs.aws.amazon.com/apigateway/api-reference/link-relation/integration-put/#cacheNamespace

If there isn't a reason I think we should let the user specify their own namespace.

Also I'm not sure what kind of data it contains and how these are structured, so it's hard to tell (for me at least) whether `resource_id` is the best namespace key - e.g. does it automatically assign the cache distribution to a given API ID so that we don't need to put API ID to the namespace? Same questions about other fields. If it's not clear from the docs it would be worth asking the AWS support - I'm happy to ask for you.

radeksimko: Also I think we should leave this as `nil` in case the user didn't specify any caching.

aws/resource_aws_api_gateway_integration_test.go
```diff
@@ -89,6 +89,28 @@ func TestAccAWSAPIGatewayIntegration_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("aws_api_gateway_integration.test", "request_templates.application/xml", "#set($inputRoot = $input.path('$'))\n{ }"),
 				),
 			},
+
+			{
+				Config: testAccAWSAPIGatewayIntegrationConfigCacheKeyParameters,
```
radeksimko: Instead of expanding this `_basic` test would you mind creating a separate one? That way we can also run tests much faster as we generally run whole tests in parallel, we cannot do that with test _steps_.

aws/resource_aws_api_gateway_integration.go
betabandido: I observed that when caching was manually enabled the namespace set by AWS matched the resource id. But, as you well say it might not be a requirement. I will conduct a test using a custom namespace. If it works, users should indeed be able to specify it.

In the meanwhile, if you can please ask AWS support that would be great. Hopefully we can then have a clear understanding of the usage (and the potential restrictions) of a cache namespace.

Setting the default to `nil` sounds good to me. I will do that.

aws/resource_aws_api_gateway_integration_test.go
betabandido: Do you mean adding a new function named `func TestAccAWSAPIGatewayIntegration_cache_key_parameters(t *testing.T)` (or similar) to the same file `resource_aws_api_gateway_integration_test.go`?

radeksimko: `TestAccAWSAPIGatewayIntegration_cache_key_parameters` - that's pretty much what I mean. There is no need for a new file.

aws/resource_aws_api_gateway_integration.go
betabandido: I just updated the pull request with your feedback. I did some testing and using a user-defined cache namespace works too. So, I added a field named `cache_namespace`.

radeksimko:
Thank you for making those changes.

I ran the relevant acceptance tests and it looks like AWS is setting the namespace by default to the `resource_id` (even if not specified in the create request), which is causing spurious diff:

```
=== RUN   TestAccAWSAPIGatewayIntegration_basic
--- FAIL: TestAccAWSAPIGatewayIntegration_basic (51.39s)
	testing.go:428: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		UPDATE: aws_api_gateway_integration.test
		  cache_namespace: "or4rhf" => ""

```

The good news is that this behaviour answers my earlier question about how the data are structured in the cache. If AWS chooses to use this key by default then it should be good (unique) enough.

Do you mind setting `cache_namespace` as `Computed`, so that Terraform can accept such default namespace instead of trying to change it to empty?

It would be obviously best if we could just set a `Default` to `resource_id`, but we don't have a good mechanism in the schema to do this - there's no way to get access to values of other fields in the schema, so `Computed` is the only option for now.

betabandido:
Done :) I ran the two acceptance tests for the integration resource, and both of them passed.

Issue #571, opened 2017-06-13
Feature request: AWS Organizations support


hashibot:
_This issue was originally opened by @bootswithdefer as hashicorp/terraform#12337. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

AWS Organizations has gone GA and it would be very helpful to manage it with Terraform.  Among other things, Organizations provides management of multiple accounts and lets you apply Service Control Policies (similar to IAM policies) to those accounts from outside of their administrative namespace.

[https://aws.amazon.com/documentation/organizations/](url)

barundel:
any movement on this?

DevBOFH:
Would love to see this work merged. Big help for those of us managing a lot of AWS accounts through Organizations.

beginbot:
This would be great for my current companies needs. so If there is any testing that could be done to help move this along, I offer my services!

AzCii:
We are also really hoping for this to happen soon, as this would be very useful.

woohgit:
:+1:

DaveBlooman:
Any updates on https://github.com/terraform-providers/terraform-provider-aws/pull/903, would like to see this go in

asedge:
I haven't heard a thing.  I also have a branch that can create an `aws_organization_account` but was hoping to get #903 merged first.

rwdaigle:
Do we even know what HashiCorp’s stance on this functionality is? Are there strategic treasons they wouldn’t support this? I’ve just seen very little guidance or indication for this addition but also not sure what their MO usually is for such things?

tamsky:
@rwdaigle asks:
> Do we even know what HashiCorp’s stance on this functionality is? 

Yes, their stance is in https://github.com/terraform-providers/terraform-provider-aws/pull/903#issuecomment-318053756 on Jul 26, 2017:


> Cases requiring a validation are always tough to handle. We try to avoid these complex stuff by either not allowing, either finding a way to do so.
> Here, storing the handshake could bring a few drawbacks, since organisations are pretty new.
We started a talk with @catsby on that, will get back to you soon!
> 
> Thanks for being patient on this topic :)

bflad:
[...]

PR #903, opened 2017-06-19
New Resource: aws_organizations_organization

asedge:
A resource to help implement `AWS Organizations`: #571.

Originally attempted a PR at: https://github.com/hashicorp/terraform/pull/13831.  Noticed the providers had been split so I'm trying again to get this merged.


website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+resource "aws_organization" "org" {
+  feature_set = "ALL"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `feature_set` - (Optional) Specify "ALL" (default) or "CONSOLIDATED_BILLING.
```
tamsky: missing trailing-`"`

aws/resource_aws_organization_test.go
```diff
@@ -0,0 +1,89 @@
...
+			return nil
+		}
+
+		if resp.Organization != nil {
+			return fmt.Errorf("Bad: Organization still exists: %q", rs.Primary.ID)
+		}
+	}
+
+	return nil
+
```
tamsky: rm extra NL

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,104 @@
...
+
+func resourceAwsOrganizationDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).orgsconn
+	_, err := conn.DeleteOrganization(&organizations.DeleteOrganizationInput{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+
```
tamsky: rm extra NL

asedge:
@tamsky Thanks for the review.  Submitted a fix for the issues you pointed out.

asedge:
Thanks again @MrGossett!

tpoindessous:
Hi

any idea this PR will be merged ? 

Thanks !

Ninir:
Hey @asedge, everyone,

This looks good to me. I am going to add some fixes and the missing upgrade from CONSOLIDATED_BILLING to ALL (http://docs.aws.amazon.com/organizations/latest/APIReference/API_EnableAllFeatures.html).

Will keep you informed when everything sounds ok.

Thanks!

asedge:
@Ninir, fine by me. Thanks!

asedge:
@Ninir, do you think anything else is needed? I'm willing to make changes if necessary.

Ninir:
Hey @asedge 

Sorry for keeping you not up-to-date on this.
I'm struggling with something a bit tricky at the moment, regarding the feature_set.

As you know, an AWS Organization can be created in 2 modes: CONSOLIDATED_BILLING or ALL.

There are 2 cases right now:
1.       If the option is set to ALL or not set, you cannot switch back to CONSOLIDATED BILLING (I added a validation error in my local code)
2.       If the option is CONSOLIDATED_BILLING and you want to switch to ALL, the [EnableAllFeatures](http://docs.aws.amazon.com/organizations/latest/APIReference/API_EnableAllFeatures.html) call is made.
 
EnableAllFeatures for AWS Organizations needs unanimity from its members, thus everyone receives an email to either accept or decline the change.
As soon as everyone accepted the enabling of the ALL mode, a new email is received by the Organization Owner, and the owner needs to go to the console, validating the change manually.
The action can take at min 1min (you’re alone in the org), at max 15 days (expiration).
 
There are 2 ways to proceed there:
1. Let the user change the value from one to the other and throw an error or make the EnableCall
2. Remove the ability to change this specific attribute by adding the ForceNew flag
 
In case 1, the template configuration could then be updated, the API Call (EnableAllFeatures) would be made, but Terraform would ask still for the change until you finalize the process manually in the console
In case 2, if you already had members, you would need to remove them from it.

We are currently discussing the UX perspective with @catsby on this.
What are your thougths there? :)

asedge:
@Ninir Is it possible to store some information about the `Handshake` in Terraform's state?  The presence of a `Handshake` would indicate there's no need to call EnableAllFeatures again plus the `Handshake` includes the expiration date: [Handshake structure](http://docs.aws.amazon.com/organizations/latest/APIReference/API_EnableAllFeatures.html#API_EnableAllFeatures_ResponseSyntax) - which we could use to time out the presence of the `Handshake`.  Perhaps a bit too over complicated.

There's a complication with case 2 as well.  If any of the Organization member accounts have been created from Organization master account they cannot be removed from the Organization without completing some manual steps: [Removing a Member Account From Your Organization](http://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_accounts_remove.html#leave-without-all-info).  I have another PR for creating `aws_organization_account` and some of the discussion around this happened here: https://github.com/hashicorp/terraform/issues/12337

There don't seem to be many good options.

Ninir:
Cases requiring a validation are always tough to handle. We try to avoid these complex stuff by either not allowing, either finding a way to do so.
Here, storing the handshake could bring a few drawbacks, since organisations are pretty new. 
We started a talk with @catsby on that, will get back to you soon!

Thanks for being patient on this topic :)

barundel:
any update? would really like to see this functionality ;)

sionsmith:
Can someone resolve this conflict so we can get this into GA.

ellerbrock:
just started to play for the first time with aws and terraform and would love to start the whole account setup with terraform. so when i understand it right its not yet merged and currently not possible to do the overall:

- create organisiation
- create subaccounts
- add organisation policies to subaccounts ...

yet, right?

cheers maik

asedge:
@ellerbrock You are correct, it's not yet possible. I have `aws_organization_account` ready to file a PR but was hoping to get this merged first.  Both resources depend on adding some new libs from `aws-sdk-go`.  I had both PRs filed when providers were still tightly coupled with Terraform but ended up just constantly fixing merge conflicts in `vendor/vendor.json`.

ellerbrock:
any news if stuff get merged?
i have to setup a huge enterprise account and would love to do all with one awesome tool called terraform :)

JoshiiSinfield:
+1.

DevBOFH:
+1 Seems like there's a lot of us that would love to have this merged.

radeksimko:
Hi folks,
we do appreciate the interest and +1's if these don't generate notifications. 😅 

We'll get to this PR 🔜 .
I'll lock this thread temporarily, until we do so, to avoid cluttering everyone's notification feed.

Thanks.

bflad:
Hi, everyone. 👋  Sorry for the extended delay in getting this new resource into the provider. I have just unlocked the commenting here because I'll be picking up this PR to take it across the finish line shortly. 😄 

@asedge quick question -- will you have time to address new PR feedback or would you prefer if the maintainers took it over? No wrong answer, just trying to gauge current status here. Either way, we'll keep your commits and make sure to acknowledge your work. Thanks!

asedge:
@bflad I would like to see it through despite not needing it anymore personally. I guess it depends on how much needs to be changed.

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+	resp, err := conn.CreateOrganization(createOpts)
+	if err != nil {
+		return fmt.Errorf("Error creating organization: %s", err)
+	}
+
+	org := resp.Organization
+	d.SetId(*org.Id)
+	log.Printf("[INFO] Organization ID: %s", d.Id())
+
+	return resourceAwsOrganizationUpdate(d, meta)
```
bflad: This should go right to calling `resourceAwsOrganizationRead(d, meta)`

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"log"
+)
+
+func resourceAwsOrganization() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsOrganizationCreate,
+		Read:   resourceAwsOrganizationRead,
+		Update: resourceAwsOrganizationUpdate,
```
bflad: As written, this resource does not support updates. We should remove this line from the resource and its associated function.

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+			},
+			"master_account_email": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"master_account_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"feature_set": {
```
bflad: Please set `ForceNew: true` on this attribute as updates are not implemented

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "ALL",
```
bflad: Minor nitpick: we should use the SDK constant here too: `organizations.OrganizationFeatureSetAll`

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+	log.Printf("[DEBUG] Creating Organization: %#v", createOpts)
+
+	resp, err := conn.CreateOrganization(createOpts)
+	if err != nil {
+		return fmt.Errorf("Error creating organization: %s", err)
+	}
+
+	org := resp.Organization
+	d.SetId(*org.Id)
+	log.Printf("[INFO] Organization ID: %s", d.Id())
```
bflad: Can you please remove this line or set it to `DEBUG`? 👍

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+	return resourceAwsOrganizationUpdate(d, meta)
+}
+
+func resourceAwsOrganizationRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).organizationsconn
+
+	log.Printf("[INFO] Reading Organization: %s", d.Id())
+	org, err := conn.DescribeOrganization(&organizations.DescribeOrganizationInput{})
+	if err != nil {
+		if orgerr, ok := err.(awserr.Error); ok && orgerr.Code() == "AWSOrganizationsNotInUseException" {
```
bflad: Minor nitpick: We can simplify this with `if isAWSErr(err, organizations.ErrCodeAWSOrganizationsNotInUseException, "") {`

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
...
+	conn := meta.(*AWSClient).organizationsconn
+
+	log.Printf("[INFO] Deleting Organization: %s", d.Id())
+
+	_, err := conn.DeleteOrganization(&organizations.DeleteOrganizationInput{})
+	if err != nil {
+		return fmt.Errorf("Error deleting Organization: %s", err)
+	}
+
+	d.SetId("")
```
bflad: Nitpick: While its still around in a bunch of the codebase, this is actually not necessary on deletion 👍

aws/resource_aws_organization.go
```diff
@@ -0,0 +1,109 @@
+package aws
+
+import (
```
bflad: Nitpick: can you please structure the imports as:
```go
import (
  // stdlib ones

  // others
)
```

FYI many folks here are using [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) to automatically add, remove, and format imports 😄

aws/import_aws_organization_test.go
```diff
@@ -0,0 +1,28 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSOrganizationImportBasic(t *testing.T) {
```
bflad: Can you rename this to `_importBasic` please? Thanks!

website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,33 @@
+---
+layout: "aws"
+page_title: "AWS: aws_organization
+sidebar_current: "docs-aws-resource-organization|"
+description: |-
+  Provides a resource to create an organization.
+---
+
+# aws\_organization
```
bflad: Nitpick: Backslashes are no longer required for documentation titles

website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+  feature_set = "ALL"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `feature_set` - (Optional) Specify "ALL" (default) or "CONSOLIDATED_BILLING".
+
```
bflad: Can you please add all the other attributes here? e.g.
```
## Attributes Reference

The following additional attributes are exported:

* `arn` - ARN of the organization
* `id` - Identifier of the organization
* `master_account_arn` - ...
...
```

website/docs/r/organization.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+## Import
+
+The AWS organization can be imported by using the `account_id`, e.g.
+
+```
+$ terraform import aws_organization.my_org 111111111111
```
bflad: While the read function will currently work with anything here at the moment, we should be encouraging importing by the organization ID (`o-XXXXXXX`) for consistent usage with the documentation above and potentially to future proof this resource.

aws/resource_aws_organization_test.go
```diff
@@ -0,0 +1,122 @@
...
+			return fmt.Errorf("Organization %q does not exist", rs.Primary.ID)
+		}
+
+		a = resp.Organization
+
+		return nil
+	}
+}
+
+func testAccAWSOrganizationConfig() string {
```
bflad: Nitpick: this can just be a `const`

aws/resource_aws_organization_test.go
```diff
@@ -0,0 +1,122 @@
...
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("Organization ID not set")
+		}
+
+		conn := testAccProvider.Meta().(*AWSClient).organizationsconn
+		params := &organizations.DescribeOrganizationInput{}
+
+		resp, err := conn.DescribeOrganization(params)
+
+		if err != nil || resp == nil {
```
bflad: This should be returning errors:

```go
if err != nil {
  return err
}

if resp == nil || resp.Organization == nil {
  return fmt.Errorf("Organization %q does not exist", rs.Primary.ID)
}
```

aws/resource_aws_organization_test.go
```diff
@@ -0,0 +1,122 @@
...
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_organization" {
+			continue
+		}
+
+		params := &organizations.DescribeOrganizationInput{}
+
+		resp, err := conn.DescribeOrganization(params)
+
+		if err != nil || resp == nil {
```
bflad: This should be checking and returning errors:

```go
if err != nil {
  if isAWSErr(err, organizations.ErrCodeAWSOrganizationsNotInUseException, "") {
    return nil
  }
  return err
}

if resp != nil && resp.Organization != nil {
  return fmt.Errorf("Organization still exists: %q", rs.Primary.ID)
}
```

Ninir:
@bflad we certainly need to address https://github.com/terraform-providers/terraform-provider-aws/pull/903#issuecomment-315400548 in some way as it is sure that people will encounter the issue a day or the other :)

aws/provider.go
```diff
@@ -454,6 +454,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_opsworks_user_profile":                    resourceAwsOpsworksUserProfile(),
 			"aws_opsworks_permission":                      resourceAwsOpsworksPermission(),
 			"aws_opsworks_rds_db_instance":                 resourceAwsOpsworksRdsDbInstance(),
+			"aws_organization":                             resourceAwsOrganization(),
```
bflad: Its going to look ugly at first and Naming is Hard, but we should probably also rename this `aws_organizations_organization` for a few reasons:
* The AWS Organizations service supports multiple resources we'll want to manage, e.g. `aws_organizations_account`, `aws_organizations_policy`, this will keep everything bundled together nicely naming-wise
* To prevent confusion with other AWS services that might use the terminology "organization"
* If AWS ever creates a separate account organization management service/method

Ideally the resource function names will be updated to long form as well: `resourceAwsOrganizationsOrganization`, `resourceAwsOrganizationsOrganizationRead`, etc

asedge: I think I handled all the other feedback you provided.  I'm not against doing this but it seemed like more work than I was willing to put in tonight.

asedge:
@bflad I believe I've taken care of everything you've given me feedback on.

bflad:
@Ninir I think we are okay for now with https://github.com/terraform-providers/terraform-provider-aws/pull/903#issuecomment-315400548 since we are no longer allowing updates in the resource (`ForceNew: true` on `feature_set`). We can and certainly should address those two scenarios if or when updates are added to the resource. Please correct me if I'm wrong and I'll address ASAP. Thanks!

bflad:

asedge:
@bflad It looks like I screwed up the web page: https://www.terraform.io/docs/providers/aws/r/organizations_organization.html

PR #910, opened 2017-06-19
Fixing malformed ARN attribute for aws_security_group data source

tbell83:
As per the docs:
http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arn-syntax-ec2

`arn:aws:ec2:region:account-id:security-group/security-group-id`


stack72:
Hi @tbell83 

Thanks for the work here - this LGTM!

Paul

PR #923, opened 2017-06-20
Fix the sample policy for redshift_service_account

lewchuk:
Missing `:` causes invalid principle error. Policy is now aligned with: https://docs.aws.amazon.com/redshift/latest/mgmt/db-auditing.html.


stack72:
Hi @lewchuk 

Thanks for this - apologies this was missed the first time around

Paul

PR #937, opened 2017-06-21
Update docs to include old AWS env vars

otakup0pe:
Looks like the old AWS env vars are still used. This is relevant as in order to make use of `AWS_SHARED_CREDENTIALS_FILE` / `AWS_PROFILE` you must unset _all_ other AWS vars.


radeksimko:
Hi @otakup0pe 
I believe this is intentionally undocumented. AWS encourages everyone to use the new env variables (also by removing old variables from most of their docs) and if we're supposed to be compliant with their suggestions we should do the same thing.

We want users to stop using these variables and documenting it would have the exact opposite effect as it would make them think it's ok to use it. One day we might remove the support for these deprecated variable altogether and this is just the first step in our way there. It's part of the deprecation process.

Thanks.

Issue #685, opened 2017-06-13
Decryption of aws_kms_secret is not correct when using it in other resources


hashibot:
_This issue was originally opened by @MattiasGees as hashicorp/terraform#13533. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.9.2

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_kms_secret

### Terraform Configuration Files
```hcl
provider "aws" {
  region = "eu-west-1"
}

data "aws_kms_secret" "mysql_password" {
  secret {
    name    = "db_user"
    payload = "<encrypted_key>"

    context {
      mysql = "password"
    }
  }
}

provider "mysql" {
  endpoint = "127.0.0.1:3306"
  username = "root"
  password = "myrootpassword"
}

resource "mysql_user" "user" {
  user     = "db_user"
  host     = "%"
  password = "${data.aws_kms_secret.mysql_password.db_user}"
}

output "db_password"{
  value = "${data.aws_kms_secret.mysql_password.db_user}"
}
```

### Expected Behavior
The user gets created and we are able to login with the password we encrypted with KMS

### Actual Behavior
The user gets created correctly but we are not able to login with the mysql password we generated and that was encrypted with KMS. The output of Terraform matches with the original key and gets decrypted correctly. There is probably something wrong with the pass from the data source to the `mysql_user` user.

When we give the password in plain text to the `mysql_user` resource it works without any problems and we are able to login to mysql with that user.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Create KMS key
2. Encrypt password 
```
echo 'password' > /tmp/plaintext-password
aws kms encrypt --key-id keyid --plaintext fileb:///tmp/plaintext-password --encryption-context mysql=password --output text --query CiphertextBlob --region eu-west-1
rm /tmp/plaintext-password
```
3. Fill the encrypted value in the terraform example
4. terraform plan
5. terraform apply
6. mysql -u db_user -p

tanhaa:
Is this bug the reason why I get the following error: 
```
Error modifying DB Instance novalibrary-dev: InvalidParameterValue: The parameter MasterUserPassword is not a valid password. Only printable ASCII characters besides '/', '@', '"', ' ' may be used.
```
when using this in an db resource:
```
[...]

Issue #616, opened 2017-06-13
Need clarification on documentation for kms_secret - data.


hashibot:
_This issue was originally opened by @derFunk as hashicorp/terraform#12800. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

this issue is about a clarification of the documentation on https://www.terraform.io/docs/providers/aws/d/kms_secret.html, yet with a `aws_db_instance` instead of a `aws_rds_cluster `.

What I assume as a _expected behaviour_ is derived from this document.

### Terraform Version
v0.9.0

### Affected Data Source
- aws_kms_secret
### Affected Resource
- aws_db_instance

### Preconditions
```sh
$ echo 'master-password' > plaintext-password
$ aws kms encrypt \
> --key-id ab123456-c012-4567-890a-deadbeef123 \
> --plaintext fileb://plaintext-example \
> --encryption-context foo=bar \
> --output text --query CiphertextBlob
AQECAHgaPa0J8 ... MmDBdqP8dPp28OoAQ==
``` 

### Terraform Configuration Files
```hcl
data "aws_kms_secret" "db" {
  secret {
    name    = "master_password"
    payload = "AQECAHgaPa0J8 ... MmDBdqP8dPp28OoAQ=="

    context {
      foo = "bar"
    }
  }
}

resource "aws_db_instance" "db" {

   # ...
  identifier = "test"
  storage_encrypted  = "${data.aws_kms_secret.db.master_password}"
  kms_key_id         = "ab123456-c012-4567-890a-deadbeef123"
  storage_encrypted  = true
  # ...

}
```

### Debug Output
```
* aws_db_instance.db: Error creating DB Instance: InvalidParameterValue:
The parameter MasterUserPassword is not a valid password.
Only printable ASCII characters besides '/', '@', '"', ' ' may be used.
```

### Expected Behavior
The RDS instance should be initialized with the decrypted password `master-password`.

### Actual Behavior
The database is initialized with the encrypted password, which fails.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
- The key `ab123456-c012-4567-890a-deadbeef123` is a valid KMS key in my test case (of course mine has another ID).

### References
- Docs: https://www.terraform.io/docs/providers/aws/d/kms_secret.html
- https://github.com/hashicorp/terraform/pull/7886
[...]

PR #940, opened 2017-06-22
Update the aws_kms_secret docs

SamClinckspoor:
The documentation led users to create secrets with unexpected
characters, leading to issues when using the datasource. This
patch warns the user and sets `echo` to use no newline.

### DIFF
--- a/website/docs/d/kms_secret.html.markdown
+++ b/website/docs/d/kms_secret.html.markdown
@@ -22,13 +22,17 @@ Please take care to secure your secret data outside of resource definitions.
 First, let's encrypt a password with KMS using the [AWS CLI
 tools](http://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).  This
 requires you to have your AWS CLI setup correctly, and you would replace the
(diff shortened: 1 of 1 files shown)


SamClinckspoor:
Should also fix #685 and #616

website/docs/d/kms_secret.html.markdown
```diff
@@ -22,13 +22,14 @@ Please take care to secure your secret data outside of resource definitions.
 First, let's encrypt a password with KMS using the [AWS CLI
 tools](http://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).  This
 requires you to have your AWS CLI setup correctly, and you would replace the
-key-id with your own.
+key-id with your own. Be sure you don't have an unexpected newline character when using a file as input.
```
Ninir: Could you add a small note about why you shouldn't have a newline there (as it may not be obvious for everyone)?

SamClinckspoor: Rephrased and added some clarification.

PR #942, opened 2017-06-22
Update refs to TLS + template providers

radeksimko:
This is a second step after https://github.com/terraform-providers/terraform-provider-aws/pull/941

```
$ grep -R 'builtin/' ./* | wc -l
       0
```


aws/resource_aws_lb_ssl_negotiation_policy_test.go
```diff
@@ -191,8 +191,6 @@ func policyAttributesToMap(attributes *[]*elb.PolicyAttributeDescription) map[st
 }
 
 // Sets the SSL Negotiation policy with attributes.
-// The IAM Server Cert config is lifted from
-// builtin/providers/aws/resource_aws_iam_server_certificate_test.go
```
tombuildsstuff: is it worth updating this comment to the new location?

radeksimko: a) I plan to use generated certs from TLS and replace all static certs
b) IMO tests shouldn't be coupled anyways, unless absolutely necessary :)

PR #944, opened 2017-06-22
data.aws_ecr_repository: Add new ECR Repository data source

RyanFrench:
Adds a new data source, for retrieving information about an ECR repository using the name.

### DIFF
--- a/aws/data_source_aws_ecr_repository.go
+++ b/aws/data_source_aws_ecr_repository.go
@@ -1,0 +1,65 @@
+package aws
+
+import (
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsEcrRepository() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsEcrRepositoryRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"registry_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"repository_url": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsEcrRepositoryRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ecrconn
+
+	repositoryName := d.Get("name").(string)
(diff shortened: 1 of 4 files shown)


aws/data_source_aws_ecr_repository.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsEcrRepository() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsEcrRepositoryRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
```
Ninir: These redundant composite literals can be safely removed (doable since Go 1.7), so this could be:

```
"name": {
```

Could you reflect the same modification everywhere where needed? Thanks!

aws/data_source_aws_ecr_repository.go
```diff
@@ -0,0 +1,65 @@
...
+			},
+		},
+	}
+}
+
+func dataSourceAwsEcrRepositoryRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ecrconn
+
+	repositoryName := d.Get("name").(string)
+	log.Printf("[DEBUG] Reading repository %s", repositoryName)
```
Ninir: To better debug things, could you update this sentence to: `Reading ECR repository named %s`?

aws/data_source_aws_ecr_repository.go
```diff
@@ -0,0 +1,65 @@
...
+		if ecrerr, ok := err.(awserr.Error); ok && ecrerr.Code() == "RepositoryNotFoundException" {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	repository := out.Repositories[0]
+
+	log.Printf("[DEBUG] Received repository %s", out)
```
Ninir: Could you also update this log for something like: `Received ECR repository %s`?

Also, if you are exposing the `out` structure, I would suggest something like `Received ECR repository %#v`, what do you think?

website/docs/d/ecr_repository.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+data "aws_ecr_repository" "service" {
+  name = "ecr-repository"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the ECR Repository
```
Ninir: As for attributes, could you also add a dot at the end?

website/docs/d/ecr_repository.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `arn` - Full ARN of the repository.
+* `registry_id` - The registry ID where the repository was created.
+* `repository_url` - The URL of the repository (in the form `aws_account_id.dkr.ecr.region.amazonaws.com/repositoryName`
```
Ninir: Small typo there: missing bracket at the end

RyanFrench:
Hi @Ninir 

Thank you for the review. I believe I have now addressed your comments.

PR #945, opened 2017-06-22
Retry timeout parameter for aws_internet_gateway

jasminSPC:
Running acceptance tests sometimes times out. There is a need to override existing timeouts.


stack72:
Hi @jasminSPC 

have you seen the new timeout blocks that were added? If not, checkout aws_db_instance.go. If we are going to add timeout support, we should potentially move to that format

Thanks

P.

jasminSPC:
Are you talking about this?
```
			Create: schema.DefaultTimeout(40 * time.Minute),
			Update: schema.DefaultTimeout(80 * time.Minute),
			Delete: schema.DefaultTimeout(40 * time.Minute),
		},
```
So this will do the same I've implemented here?

radeksimko:
@jasminSPC do you mind attaching the acceptance test output? I think the general rule we applied in the past was to set the **default** timeout high enough that all acceptance tests (which are testing common use cases) pass consistently.

In other words the user should _not_ be forced to set custom timeout for regular usage of the resource. That is why I think the best solution in the context of IGW is just bumping the timeout, rather than introducing new timeout fields.

jasminSPC:
@radeksimko when I said acceptance tests I mean some tests that I'm running not acceptance tests of TF. I have an infrastructure which after some time errors out due short timeouts.
Anyhow here is the acceptance test output for the stuff I changed:
```
 make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSInternetGateway_basic'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSInternetGateway_basic -timeout 120m
=== RUN   TestAccAWSInternetGateway_basic
--- PASS: TestAccAWSInternetGateway_basic (101.94s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       101.980s
```

radeksimko:
I see, that's 👌  then.

Would you mind sharing any more details about the infrastructure? It would be useful to know what kind of aspect cause the IGW to take more time to delete (or wait for deletion).

radeksimko:
And yes - @stack72 was talking about those timeouts.

jasminSPC:
@radeksimko About the infrastructure essentially it is an AWS environment under a VPC with 3 servers. We run tests daily basically we do apply/destroy. Every now and then it happens that we get such message: 
`* aws_internet_gateway.igw: Error waiting for internet gateway (igw-7957c91f) to detach: couldn't find resource (31 retries)` we believe that this is due to defaul timeout duration.

jasminSPC:
I also think it would be a good idea to add this feature through out other AWS resources.

edevenport:
We are experiencing various timeouts with smoke tests, particularly with removing the internet gateway on clean up. The behavior is not consistent and seems to be load related. The approach already taken with `aws_db_instance.go` seems like a good approach and allows for additional flexibility for automated tests. We would love to expand this to other resources.

I agree a user should not be forced to set a custom timeout, but there is a case for allowing users to override reasonable defaults. Simply increasing the hardcoded timeout would add pain when testing or troubleshooting issues while being forced to wait for the predefined timeout.

s-urbaniak:
@jasminSPC short question, does the above also override the hardcoded timeout in https://github.com/terraform-providers/terraform-provider-aws/blob/757055f/aws/resource_aws_internet_gateway.go#L242-L244? The referenced timeout was the actual source of flakes when destroying clusters, i.e. 30 retries * 10 seconds = 300 seconds = 5 minutes timeout. This 5 minute timeout is too short in ~70%ile of destroy cases.

s-urbaniak:
@radeksimko The investigation details can be found here: https://github.com/coreos/tectonic-installer/issues/1017#issuecomment-307780549. The TLDR is that we believe that there is an implicit dependency inside AWS between the internet gateway and AWS EIPs and/or AWS route tables not reflected in the internal Terraform dependency graph. The API does not expose it as an explicit dependency, hence Terraform destroys those resources concurrently, eventually giving up on the internet gateway if the start of destroying it was too early. We tried to mitigate this by setting explicit dependencies, i.e. in https://github.com/coreos/tectonic-installer/pull/1092, but that is still flaky :-/

jasminSPC:
@s-urbaniak I think that extending the resource with `not_found_check_count` which would be used to replace default value of 30 retries could be the best option to solve this problem. 

What do you think?

s-urbaniak:
@jasminSPC that would help in our use case, yes. But having said that, accounting detaching the internet gateway in the total destroy timeout would be the ideal solution from the users perspective.

jasminSPC:
@s-urbaniak detach is called in from two places one is delete and the other is update. I've implemented resource level timeouts for create, delete and update because timeouts for all three operations are different. 
So in my opinion I would add `not_found_check_count` parameter to the resource to allow for plus we could replace default timeout on [detach](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_internet_gateway.go#L242) with delete timeout with default value of 10.

jasminSPC:
I also believe that the similar change should be added to other resources, but it is worth mentioning that some of them are not using `resource.Retry` but `resource.StateChangeConf` which has more parameters than `resource.Retry`.

s-urbaniak:
@jasminSPC adding `not_found_check_count` would certainly help, agreed.

jasminSPC:
@s-urbaniak what about other timeouts? We add `not_found_check_count` and on top of that we parameterise timeouts. If we decide to leave timeouts what about `resourceAwsInternetGatewayAttach` and `resourceAwsInternetGatewayDetach` to which group of timeouts they belong `update` or `delete`?

radeksimko:
I'd still like to understand the complexity or specifics of your TF configuration, ideally from a config + debug log, before introducing `not_found_check_count`. It is possible that this doesn't affect just you, but some other users that may be using Terraform in a reasonable way. If that is the case we certainly don't expect such users to set a new special field to make things work. For common cases it should work out of the box, even if that means raising the timeout or toleration of failures generally for everyone.

TL;DR It would be better to have some real data to confirm our thoughts here before discussing and implementing solution. 😉 

Thanks.

s-urbaniak:
@radeksimko @jasminSPC I emailed you the destroy log which leads to the error reported here. Note that in order to provoke it to a maximum I did the following nasty change to the internet gateway provider ;-)

```
diff --git a/builtin/providers/aws/resource_aws_internet_gateway.go b/builtin/providers/aws/resource_aws_internet_gateway.go
index 3834aa58e..ba7f2cba3 100644
--- a/builtin/providers/aws/resource_aws_internet_gateway.go
+++ b/builtin/providers/aws/resource_aws_internet_gateway.go
@@ -241,7 +241,7 @@ func resourceAwsInternetGatewayDetach(d *schema.ResourceData, meta interface{})
                Refresh:        detachIGStateRefreshFunc(conn, d.Id(), vpcID.(string)),
                Timeout:        15 * time.Minute,
                Delay:          10 * time.Second,
-               NotFoundChecks: 30,
+               NotFoundChecks: 1,
        }
        if _, err := stateConf.WaitForState(); err != nil {
                return fmt.Errorf(
```

jasminSPC:
@s-urbaniak Thank you

radeksimko:
Thank you for providing the debug log and extra details. That's very helpful.

I noticed in your log (and during my own debugging) that this is the error we retry on:

`<Response><Errors><Error><Code>DependencyViolation</Code><Message>Network vpc-a**b**c* has some mapped public address(es). Please unmap those public address(es) before detaching the gateway.</Message></Error></Errors><RequestID>REDACTED</RequestID></Response>`

This error code is represented as "not found" by the retrier which is the cause of the problem IMO.

https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_internet_gateway.go#L274
```go
case "DependencyViolation":
	return nil, "detaching", nil
```
^ that first `nil` is the cause.

We should retry normally in such cases and be bound by the timeout, not by `NotFoundChecks`.

Treating `DependecyViolation` as a standard retry error instead of "not found" should give us more retries overall. 30 "not found" checks with the current retrier settings gives us `4m22.7s`, whereas the overall timeout for standard retries is `15mins`. Those extra `10mins` might actually be sufficient.

This is very simplified version of [our retrier](https://github.com/hashicorp/terraform/blob/master/helper/resource/state.go) https://play.golang.org/p/XHm-k0MUrS in case you wonder how did I come up with the numbers.

I will send a separate PR to address this as I think the linked code deserves some care.

radeksimko:
btw. the public IPs which are lying around come from 2 of your public ELBs in the configuration. I'll take a look if we could do some cleanup in the ELB resource too, to speed things up a bit.

jasminSPC:
Regardless of @radeksimko elegant solution I think that customisable resource timeouts are a good idea. Also @radeksimko's solution renders `not_found_check_count` unnecessary.  

How should we proceed?

radeksimko:
My suggestion on going forward is to hear some feedback from @s-urbaniak based on stress testing and if the feedback is positive and someone on the team gives the PR 👍  then I'd merge it and then close this one.

Customisable timeouts are useful in some contexts - as explained earlier - specifically when it's clear the time-to-provision is going to vary and it's reasonable for the time to vary - e.g. EC2 instance size (the bigger, the long it takes to provision), Autoscaling Group (the more instances in the group the longer it may take to reach min. size), RDS, CloudFormation Stack (because it may be managing all of the mentioned resources and much more).

Here's where we support standard customisable timeouts today:

```
aws_alb
aws_autoscaling_group
aws_cloudformation_stack
aws_db_instance
aws_db_option_group
aws_db_snapshot
aws_dms_replication_instance
aws_instance
aws_opsworks_instance
aws_rds_cluster
aws_rds_cluster_instance
aws_spot_instance_request
```
I'm not sure why we have timeouts in `db_option_group` and in`alb` to be honest, but those two resources are the only two that look suspicious to me.

radeksimko:
I'm going to close this PR with the explanation above. I think the two PRs ([one increasing timeout](https://github.com/terraform-providers/terraform-provider-aws/issues/1021), [other one cleaning up ELB's ENIs](https://github.com/terraform-providers/terraform-provider-aws/pull/1036)) should be sufficient to fix it, but please do let me know if you keep seeing this error or similar - we can revisit.

Thanks you both @s-urbaniak and @jasminSPC for bringing this to our attention and suggesting solutions.

PR #946, opened 2017-06-22
[WIP/Request Review] New Resource: aws_ses_identity_notification_topic

### LINKED ISSUE #931: SES Noification Settings
At present `aws_ses_domain_identity` only allows for me to configure the domain.
I would like to configure the Topic Configuration i.e. bounce/complaint/delivery SNS topics.

benoj:
Attempts to resolve #931 

Would like some feedback on general style and testing?

### DIFF
--- a/aws/resource_aws_ses_identity_notification.go
+++ b/aws/resource_aws_ses_identity_notification.go
@@ -1,0 +1,97 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ses"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsSesNotification() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesNotificationSet,
+		Read:   resourceAwsSesNotificationRead,
+		Update: resourceAwsSesNotificationSet,
+		Delete: resourceAwsSesNotificationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"topic_arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: false,
+			},
+
+			"notification_type": &schema.Schema{
+				Type:     schema.TypeSet,
+				Required: true,
+			},
+
+			"identity": &schema.Schema{
+				Type:     schema.TypeSet,
+				Required: true,
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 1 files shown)


aws/resource_aws_ses_identity_notification.go
```diff
@@ -0,0 +1,95 @@
...
+		if err := d.Set("topic", notificationAttributes.DeliveryTopic); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func resourceAwsSesNotificationDelete(d *schema.ResourceData, meta interface{}) error {
+	d.Set("topic_arn", nil)
```
Ninir: To delete it, I think we should call the `SetIdentityNotificationTopic` method, passing nil / empty for the SNS Topic. Thoughts? :)

aws/resource_aws_ses_identity_notification.go
```diff
@@ -0,0 +1,95 @@
...
+	response, err := conn.GetIdentityNotificationAttributes(getOpts)
+
+	if err != nil {
+		return fmt.Errorf("Error reading SES Identity Notification: %s", err)
+	}
+
+	notificationAttributes := response.NotificationAttributes[identity]
+	switch notification {
+	case "Bounce":
+		if err := d.Set("topic", notificationAttributes.BounceTopic); err != nil {
```
Ninir: It seems that this `topic` field does not exist in the schema. Did you think of something or is it a typo?

Also, it could be really interesting to add a `attributes` field, which would contain everything retrieved/sent back by the AWS API when reading the field. Thoughts? :)

aws/resource_aws_ses_identity_notification.go
```diff
@@ -0,0 +1,95 @@
...
+	return &schema.Resource{
+		Create: resourceAwsSesNotificationSet,
+		Read:   resourceAwsSesNotificationRead,
+		Update: resourceAwsSesNotificationSet,
+		Delete: resourceAwsSesNotificationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"topic_arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: false,
```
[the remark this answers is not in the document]
Ninir: Also, we need to add a `ValidateFunc: validateArn,` attribute, so that it uses a function to validate the value passed in the configuration. `validateArn` is an existing function that validates the format using a given RegExp, so you don't nee much more here :)

For other attributes, it can be nice to also validate the value passed, using custom functions!

bflad:
This PR has been superseded by #2640 and merged into master! The new `aws_ses_identity_notification_topic` resource will release with v1.14.0 of the AWS provider, likely by the end of the week. Thanks!

PR #947, opened 2017-06-22
Fix ssm_parameter doc line break

OliverCole:
Renders fine on GitHub without the additional linebreak, but the eventual deployed page is incorrect: https://www.terraform.io/docs/providers/aws/r/ssm_parameter.html


Ninir:
Hi @OliverCole 

This seems good to me. Thanks!

PR #949, opened 2017-06-22
Make the ASG delete timeout configurable

catsby:
Originally https://github.com/hashicorp/terraform/pull/14396

Note that this won't actually respect the delete timeout until https://github.com/hashicorp/terraform/pull/15374 is merged and we bump our internal dependency on `helper/schema`

TODO:

- [x] get https://github.com/hashicorp/terraform/pull/15374 merged
- [x] bump our dep of terraform/helper/schema
- [ ] verify tests

------

Tests passed:

screen shot 2017-06-23 at 10 55 37 am

### DIFF
--- a/aws/resource_aws_autoscaling_group.go
+++ b/aws/resource_aws_autoscaling_group.go
@@ -28,7 +28,7 @@ func resourceAwsAutoscalingGroup() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
-			Delete: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(10 * time.Minute),
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


catsby:
@jringuette here's the other part of that ASG PR you opened 😄

aws/resource_aws_autoscaling_group.go
```diff
@@ -834,7 +838,7 @@ func resourceAwsAutoscalingGroupDrain(d *schema.ResourceData, meta interface{})
 
 	// Next, wait for the autoscale group to drain
 	log.Printf("[DEBUG] Waiting for group to have zero instances")
-	return resource.Retry(10*time.Minute, func() *resource.RetryError {
+	return resource.Retry(d.Timeout(schema.TimeoutDelete), func() *resource.RetryError {
```
radeksimko: Is there any reason to decrease the default timeout from 10 to 5 mins here?
We could either introduce custom draining timeout or set the default to 10mins, I think.

catsby: Ah no, that's a strange typo or copy/paste error, I'll bump that back up to 10, thanks!

PR #950, opened 2017-06-22
AWS_METADATA_TIMEOUT configuration environment variable

handlerbot:
We have a custom Terraform authentication server which impersonates the EC2 metadata API endpoint, but our use and implementation would benefit from being able to boost the 100ms API endpoint timeout in our environment.

### DIFF
--- a/aws/auth_helpers.go
+++ b/aws/auth_helpers.go
@@ -125,8 +125,25 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 	// Build isolated HTTP client to avoid issues with globally-shared settings
 	client := cleanhttp.DefaultClient()
 
-	// Keep the timeout low as we don't want to wait in non-EC2 environments
+	// Keep the default timeout (100ms) low as we don't want to wait in non-EC2 environments
 	client.Timeout = 100 * time.Millisecond
+
+	const userTimeoutEnvVar = "AWS_METADATA_TIMEOUT"
+	userTimeout := os.Getenv(userTimeoutEnvVar)
+	if userTimeout != "" {
+		newTimeout, err := time.ParseDuration(userTimeout)
+		if err == nil {
+			if newTimeout.Nanoseconds() > 0 {
+				client.Timeout = newTimeout
+			} else {
+				log.Printf("[WARN] Non-positive value of %s (%s) is meaningless, ignoring", userTimeoutEnvVar, newTimeout.String())
+			}
+		} else {
+			log.Printf("[WARN] Error converting %s to time.Duration: %s", userTimeoutEnvVar, err)
+		}
+	}
+
+	log.Printf("[INFO] Setting AWS metadata API timeout to %s", client.Timeout.String())
 	cfg := &aws.Config{
 		HTTPClient: client,
 	}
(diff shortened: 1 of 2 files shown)


aws/auth_helpers.go
```diff
@@ -125,8 +126,19 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 	// Build isolated HTTP client to avoid issues with globally-shared settings
 	client := cleanhttp.DefaultClient()
 
-	// Keep the timeout low as we don't want to wait in non-EC2 environments
-	client.Timeout = 100 * time.Millisecond
+	// Keep the default timeout low as we don't want to wait in non-EC2 environments
+	timeoutMillis := 100
```
Ninir: What do you think of calling this `defaultTimeout` and rename `timeoutString` as `userTimeout`?

aws/auth_helpers.go
```diff
@@ -125,8 +126,19 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
...
+	timeoutString := os.Getenv("AWS_METADATA_TIMEOUT_MILLISECONDS")
```
Ninir: I would set the env variable name as a constant for reusability here and in the log

website/docs/index.html.markdown
```diff
@@ -113,6 +113,9 @@ which reduces the chance of leakage.
 You can provide the custom metadata API endpoint via the `AWS_METADATA_ENDPOINT` variable
 which expects the endpoint URL, including the version, and defaults to `http://169.254.169.254:80/latest`.
 
+The default deadline for the EC2 metadata API endpoint is 100 milliseconds, but you can
+override the default by setting the `AWS_METADATA_TIMEOUT_MILLISECONDS` variable.
```
Ninir: What do you think of updating this to:

```
The default deadline for the EC2 metadata API endpoint is 100 milliseconds, which can be overidden by setting the `AWS_METADATA_TIMEOUT_MILLISECONDS` environment variable.
```
?

aws/auth_helpers.go
handlerbot: I like `userTimeout`, but I think I'll go with `timeoutMillis` -> `timeout` rather than `defaultTimeout`. Once a value from the user is assigned to the variable, it's no longer the the default, no? :-)

radeksimko:
I'm 👌 with the idea overall and leave the final review on @Ninir 

Just one question from me here - is it worth accepting the standard duration format, e.g. `1m`, `20sec`, `100ms` etc? I'm aware it would probably change a bunch of things in the code, e.g. `AWS_METADATA_TIMEOUT_MILLISECONDS` to `AWS_METADATA_TIMEOUT` and the parsing logic to [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) and I don't want to cause much pain, but I also feel it's worth considering. 😉 

btw. by convention we always leave the Changelog outside of PRs as it's responsibility of the person merging the PR. We can avoid unnecessary conflicts that way.

Ninir:
Hey @handlerbot 

Could you estimate the work to make regarding @radeksimko 's proposal?
If it's too much for this work, this could be done in another PR.

Thanks!

PR #963, opened 2017-06-24
Mention AWS environment variable precedence

otakup0pe:
A more appropriate take on #937

Unrelated; is there a trick to running tests locally?

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -60,9 +60,11 @@ provider "aws" {
 
 You can provide your credentials via the `AWS_ACCESS_KEY_ID` and
 `AWS_SECRET_ACCESS_KEY`, environment variables, representing your AWS
-Access Key and AWS Secret Key, respectively.  The `AWS_DEFAULT_REGION`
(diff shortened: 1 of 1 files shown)


radeksimko:
Thanks, this is helpful 👍 

re tests:

you can run all unit tests (the fast ones, which shouldn't talk to any external API) like this:

```
make test
```
^ these take less than a minute (which includes compilation) on `3.3 GHz Intel Core i7`.

and acceptance tests (the slower ones, which actually create real resources inside the chosen AWS account - i.e. may cost money 💰) e.g. like this:

```
AWS_PROFILE=myprofile make testacc TESTARGS='-run=TestAccAWSDefaultVpc_basic'
```
I'd suggest to always run only a small scope of tests, e.g. VPC only - if you're changing VPC resource/datasource, because all tests would take _a lot of time_. We run all AWS tests daily with parallelism set to 20 (20 tests in parallel) and it takes 2-3hours to finish.

otakup0pe:
I'm going to assume it is something funky with my local environment causing `make test` to not work... I'll revisit this if I end up making non-doc changes.

Issue #169, opened 2017-06-13
Network Interface private_ips in different order


hashibot:
_This issue was originally opened by @Shaiou as hashicorp/terraform#6750. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

- aws_network_interface

Hi there,
 I'm running into some issue when creating a network interface, I specified a list of private_ips and made sure the one I wanted as primary was listed first, however it does respect that order and the second one appears as primary: 

Here's my ressource: 

```
resource "aws_network_interface" "waf_1_pub" {
    subnet_id = "${aws_subnet.public.1.id}"
    security_groups = ["${aws_security_group.waf_prv.id}"]
    # We put the basic ip for sync and all and one IP per VS
    private_ips = ["${cidrhost(aws_subnet.public.1.cidr_block, var.waf_ip_offset)}", "${join(",",template_file.waf_1_vs_ips.*.rendered)}"]
    security_groups = [ "${aws_security_group.waf_pub.id}" ]
    attachment {
        instance = "${aws_instance.waf.1.id}"
        device_index = 1
    }
}
```

And the output of the terraform show after the apply: 

```
aws_network_interface.waf_1_pub:
  id = eni-8ba85cf3
  attachment.# = 1
  attachment.3009111673.attachment_id = eni-attach-2baa9ae8
  attachment.3009111673.device_index = 1
  attachment.3009111673.instance = i-18be9790
  description = 
  private_ips.# = 2
  private_ips.3297441504 = 10.220.1.210
  private_ips.3717318561 = 10.220.1.200 => that was the first one in my list and
  security_groups.# = 2
  security_groups.1563906976 = sg-9a10b6fd
  security_groups.973784294 = sg-9910b6fe
  source_dest_check = true
  subnet_id = subnet-e2c50e86
  tags.# = 0
```

Can you help ?

certnew:
Need this please

melbit-anthonyw:
Yep I just hit this while attempting to build a SQL availability group. All this stemmed from the lack of support to define primary and secondary IP addresses for network interfaces.

ArturZurawski:
Yes, same for me. It gets a bit weird: 

```
resource "aws_network_interface" "interface" {
...
  private_ips = [
     "10.113.0.4",
     "10.113.0.5"
  ]
}
```

produces:

```
[...]

PR #967, opened 2017-06-26
Fix private ips returning in wrong order

aschepp:
This is a pull request to the new repository of the old pull request by @freimer that fixes issue #169
hashicorp/terraform#6986


justinclayton:
Why was this closed?

PR #970, opened 2017-06-26
IAM Server Cert test updates

catsby:
Update our IAM Server cert tests to use the TLS provider so we don't have to update the certs every year. 

I also generated new certs for the test-fixtures because we want to keep a test for windows line endings. 

I created a `.pem.winfile` to denote to `gitignore` that this file was to keep its line endings.


aws/provider_test.go
```diff
@@ -8,9 +8,11 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/terraform-providers/terraform-provider-template/template"
+	"github.com/terraform-providers/terraform-provider-tls/tls"
 )
 
 var testAccProviders map[string]terraform.ResourceProvider
+var testAccProvidersWithTLS map[string]terraform.ResourceProvider
```
catsby: Here we add the `TLS` provider to just the new `testAccProvidersWithTLS` map, so tests that need it can explicitly use `testAccProvidersWithTLS` instead of `testAccProviders`. This way only those tests have the additional provider and the rest remain untainted

aws/resource_aws_iam_server_certificate_test.go
```diff
@@ -87,28 +93,35 @@ func TestAccAWSIAMServerCertificate_disappears(t *testing.T) {
 }
 
 func TestAccAWSIAMServerCertificate_file(t *testing.T) {
-	var cert iam.ServerCertificate
+	var _cert iam.ServerCertificate
```
catsby: I use `_` here to denote that I don't actually care about `_cert` in this test; loading the file (with and without windows line endings) and applying should be sufficient, because follow up `plan` as part of the test will tell us if there's a diff due to endings.

aws/resource_aws_iam_server_certificate_test.go
```diff
@@ -141,14 +145,27 @@ func testAccCheckCertExists(n string, cert *iam.ServerCertificate) resource.Test
...
 			return fmt.Errorf("Bad Server Cert Name: %s", *cert.ServerCertificateMetadata.ServerCertificateName)
 		}
 
-		if *cert.CertificateBody != strings.TrimSpace(certBody) {
-			return fmt.Errorf("Bad Server Cert body\n\t expected: %s\n\tgot: %s\n", certBody, *cert.CertificateBody)
+		if *cert.CertificateBody != strings.TrimSpace(*certBody) {
+			return fmt.Errorf("Bad Server Cert body\n\t expected: %s\n\tgot: %s\n", *certBody, *cert.CertificateBody)
 		}
 		return nil
 	}
```
radeksimko: The old certificate in `var certBody` on line 200 (apparently I cannot comment on lines outside of context) can go away too I think?

radeksimko: I made that change and pushed to your branch, so we can get more tests green asap. All `TestAccAWSIAMServerCertificate_` tests are passing after the removal, so I reckon it's 👌

radeksimko:
Merging to make tests green, butt feel free to respond to my comments when you're back on Monday.

catsby:
Thanks @radeksimko !

PR #974, opened 2017-06-27
aws/data/subnet: Add filter example to website

tamsky:
I found the lack of an example to be extremely unfriendly... 

After consulting the API documentation, I believed the correct syntax was `tag:name=Name` vs the actually correct `tag:Name`.

It was this example which saved me: https://github.com/hashicorp/terraform/issues/10074#issuecomment-260439012


stack72:
Hi @tamsky 

Thanks so much for this - sorry this was unfriendly for you! 

Paul

tamsky:
>  sorry this was unfriendly for you!

Thanks for the friendly fix!

stack72:
Nothing to apologise for at all :)

PR #976, opened 2017-06-27
enlarge origin_read_timeout validation

cmartinez85:
We asked to AWS to increase the limitation of the parameter origin_read_timeout of cloudfront distribution to 150 seconds (default is 60) and they granted it us.
When we tried to apply these 150 seconds in this parameter with terraform the bellow message appear:

`* module.cloudfront_es.aws_cloudfront_distribution.web: expected origin.1.custom_origin_config.0.origin_read_timeout to be in the range (4 - 60), got 150. `

So we need to increase (or remove) this limitation in terraform too.

### DIFF
--- a/aws/resource_aws_cloudfront_distribution.go
+++ b/aws/resource_aws_cloudfront_distribution.go
@@ -367,7 +367,7 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 										Type:         schema.TypeInt,
 										Optional:     true,
 										Default:      30,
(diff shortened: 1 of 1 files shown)


stack72:
Hi @cmartinez85 

Thanks for the PR here - I am going to handle this in a slightly different manner. I am going to remove the validation completely and add some documentation to suggest that there is a default limit of 60 but that you can ask AWS to increase

thanks for this

Paul

cmartinez85:
thanks!

PR #979, opened 2017-06-27
Add aws_ecr_policy_statement

DennyLoko:
This PR was originally opened by @greenboxal at hashicorp/terraform#12078.

I'm using this and has been very useful. I believe it can be fully merged on this provider.

PR #985, opened 2017-06-27
New Resource: aws_ecr_repository_policy_statement

greenboxal:
This resource allows to manage ECR policy statements as single resources, allowing to spread them across several modules. This is useful as a ECR repository can have only one policy document.

It was implemented in the same way that aws_security_group_rule works: It locks the resource mutex and applies the diff (add, remove or change the statement) over the policy document and applies it on the repository.

### DIFF
--- a/aws/resource_aws_ecr_repository_policy_statement.go
+++ b/aws/resource_aws_ecr_repository_policy_statement.go
@@ -1,0 +1,294 @@
+package aws
+
+import (
+	"encoding/json"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+type ecrPolicy struct {
+	Version    string                `json:",omitempty"`
+	ID         string                `json:",omitempty"`
+	Statements []*ecrPolicyStatement `json:"Statement"`
+}
+
+type ecrPolicyStatement struct {
+	Sid           string
+	Effect        string      `json:",omitempty"`
+	Actions       interface{} `json:"Action,omitempty"`
+	NotActions    interface{} `json:"NotAction,omitempty"`
+	Resources     interface{} `json:"Resource,omitempty"`
+	NotResources  interface{} `json:"NotResource,omitempty"`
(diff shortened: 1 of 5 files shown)


greenboxal:
@stack72 as discussed in #979, just improved the documentation.

Ninir:
Hi @greenboxal

Sorry for taking this long to get back to you.
I have read the discussion in the other PR, and I am just wondering why not using the `aws_ecr_repository_policy` policy along with the `template_file` (for reusability)?

Hope I understood the use-case well 😄 

THanks!

greenboxal:
You can only apply one policy per repository. A lot of times you want to deploy the same application in different infrastructures, using the same repository. Maybe for creating a staging -> production pipeline.

Let's say you have a module called `my-cool-app`, and that I want to provision it twice, one for production, one for staging. Still, I want to use the same ECR repository.

If I want to use the same repository, I can't create the repo inside the module, otherwise I would end up with two different repositories. I need to create a policy for each `my-cool-app` module instance, so its EC2 instances can download the images from the repo.

The first approach that come to my head is: create one policy for each module instance and attach them to the repo. But then you have a issue here: ECR repos can only have one policy.

This resources solves this, in the same way that `aws_security_group_rule` resource solves a related issue with security groups. You can do a kind of "dependency injection".

aws/resource_aws_ecr_repository_policy_statement.go
```diff
@@ -0,0 +1,294 @@
...
+		if err := json.Unmarshal([]byte(*out.PolicyText), policy); err != nil {
+			return nil, nil, err
+		}
+	} else {
+		policy.Version = "2008-10-17"
+	}
+
+	if policy.Statements == nil {
+		policy.Statements = make([]*ecrPolicyStatement, 0)
+	}
```
Ninir: @greenboxal Do you think you could take some time to create a data source called `aws_iam_policy_document` which would have a source_policy document, that would allow to merge your policy into the source one?

Your use case seems totally valid and it would be a great addition to have it! 👍 
I am just wondering that maybe we could make this part reusable so that we can handle this case in several situations. What do you think? :)

bflad:
I believe this will be helped/solved by #2890

aws/resource_aws_ecr_repository_policy_statement.go
devonbleak: @Ninir this sounds almost exactly like what I did with #2890

bflad:
As noted above, this was solved more generically with the `aws_iam_policy_document` data source. 👍

Issue #143, opened 2017-06-13
AWS IoT support


hashibot:
_This issue was originally opened by @vrcsix as hashicorp/terraform#6138. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Edit by Ninir: TODOs
- [ ] Thing
- [x] Thing Type
- [ ] Thing Group
- [x] Certificate
- [x] Policy
- [ ] Policy Attachment
- [x] Topic

<hr>
Opening this to spec out Terraform support for [AWS IoT](https://aws.amazon.com/iot/).

**Thing**
Represents a thing in the thing registry.

``` hcl
resource "aws_iot_thing" "device3" {
  name = "MyDevice3"
  principals = ["${aws_iot_certificate.cert.arn}"]

  attributes {
    Manufacturer = "Amazon"
    Type = "IoT Device A"
    SerialNumber = "10293847562912"
  }
}
```

``` go
func Create(t Thing) {
  iot.CreateThing(t)
  for p := range t.principals {
    iot.AttachThingPrincipal(t, p)
  }
}

func Read(t Thing) {
  update(t, iot.DescribeThing(t))
}

func Update(t Thing) {
  toBeDetached, toBeAttached := principalAttachmentChanges(t)

  for p := range toBeDetached {
    iot.DetachThingPrincipal(t, p)
  }

  for p := range toBeAttached {
    iot.AttachThingPrincipal(t, p)
  }

  iot.UpdateThing(t)
}

func Delete(t Thing) {
  for p := range t.principals {
    iot.DetachThingPrincipal(t, p)
  }

  iot.DeleteThing(t)  
}

```

**Certificate**
Represents a X.509 certificate for use with IoT.

``` hcl
resource "aws_iot_certificate" "cert" {
  csr = "${file("/my/csr.pem")}"
  active = true 
}
```

``` go
func Create(c Certificate) {
  iot.CreateCertificateFromCsr(c)
}

func Read(c Certificate) {
  update(c, iot.DescribeCertificate(c))
}

func Update(c Certificate) {
  if hasChanged(c.csr) {
    forceNewResource(c)
  } else {
    iot.UpdateCertificate(c)
  }
}

func Delete(c Certificate) {
  iot.UpdateCertificate(c, Input{NewStatus: "INACTIVE"})
  iot.DeleteCertificate(c)
}
```

**Policy**
[...]

PR #986, opened 2017-06-28
Issue #143 - Add aws_iot_policy resource based on work from @jhedev

AlexMabry:
My goal is to add support for the AWS IoT service.  This is my first time committing, so I just wanted to add a simple resource to get my feet wet.  I believe I am followed the guidelines, but please point out anything that needs to be added or change.


aws/resource_aws_iot_policy_test.go
```diff
@@ -0,0 +1,109 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSIoTPolicyDestroy_basic,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSIoTPolicy_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSIoTPolicyExists_basic("aws_iot_policy.pubsub"),
+				),
```
radeksimko: Because we run all acceptance tests nightly in parallel (and because tests sometimes do fail and leave named resources behind, blocking the same test from running again) we need to avoid static names.

Have a look at what we do in other resources, e.g.
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_dynamodb_table_test.go#L19

Also do you mind adding a few more checks here? e.g.

```go
resource.TestCheckResourceAttr("aws_iot_policy.pubsub", "name", rName),
resource.TestCheckResourceAttrSet("aws_iot_policy.pubsub", "arn"),
resource.TestCheckResourceAttrSet("aws_iot_policy.pubsub", "default_version_id"),
```

aws/resource_aws_iot_policy_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func testAccCheckAWSIoTPolicyDestroy_basic(s *terraform.State) error {
+	conn := testAccProvider.Meta().(*AWSClient).iotconn
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_iot_policy" {
+			continue
+		}
+
+		out, err := conn.ListPolicies(&iot.ListPoliciesInput{})
```
radeksimko: Wouldn't it be easier to call `GetPolicy` here as we only care about the single policy we created as part of the test in this context?

aws/resource_aws_iot_policy.go
```diff
@@ -0,0 +1,131 @@
...
+		PolicyName: aws.String(d.Id()),
+	})
+
+	if err != nil {
+		log.Printf("[ERROR] %s", err)
+		return err
+	}
+
+	d.Set("arn", out.PolicyArn)
+	d.Set("defaultVersionId", out.DefaultVersionId)
```
radeksimko: 👀  I think the field name should be `default_version_id`.

website/docs/r/iot_policy.html.markdown
```diff
@@ -0,0 +1,49 @@
+---
+layout: "aws"
+page_title: "AWS: aws_iot_policy"
+sidebar_current: "docs-aws-resource-iot-policy"
+description: |-
+  Provides an IoT policy.
+---
```
radeksimko: We also need to add a link to the sidebar to make this page discoverable, see https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/aws.erb

aws/resource_aws_iot_policy_test.go
```diff
@@ -0,0 +1,109 @@
...
+				return fmt.Errorf("IoT policy still exists:\n%s", t)
+			}
+		}
+
+	}
+
+	return nil
+}
+
+func testAccCheckAWSIoTPolicyExists_basic(name string) resource.TestCheckFunc {
```
radeksimko: We usually leverage helpers like `resource.TestCheckResourceAttr` and `resource.TestCheckResourceAttrSet` to check whether the resource was recorded in the state.

The point of having a function like this is to allow more thorough check of the API/SDK struct, take a look how we do it in a VPC test: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_vpc_test.go#L25-L26

It's not mandatory to perform such thorough checking, but if you decide not to do it there's no point of having this function when the standard helpers can do the job.

PR #1018, opened 2017-06-30
Fix typo in aws_security_group data source example

aitorpazos:
Variable used in the website documentation aws_security_group data source should be security_group_id as it is declared.


stack72:
LGTM! thanks for this :)

PR #1021, opened 2017-06-30
r/internet_gateway: Retry properly on DependencyViolation

radeksimko:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/181

Root cause of the bug is described in https://github.com/terraform-providers/terraform-provider-aws/pull/945#issuecomment-312249655

In addition to retrying I also added some extra debugging info which should help us in case this problem reoccurs - at least we'll know what ENIs are lying around and what/who created them.

I can confirm through my repro case + debug logs provided by @s-urbaniak that ELB (public one) is one service which can leave IPs around for a while and cause this error, but @s-urbaniak also mentioned that NAT Gateway can be blamed in https://github.com/coreos/tectonic-installer/issues/1017#issuecomment-307780549 so this is possibly wide-spread problem throughout AWS services which allocate public IPs.

btw. We previously ignored unknown errors before, so I'm fixing that too in this PR.


s-urbaniak:
@radeksimko Thanks a lot for this PR, that's great! I will include it in our CI pipeline and recheck if we get rid of that flake.

Issue #1008, opened 2017-06-29
Changing the name option for a aws_ssm_parameter creates a new parameter in the Parameter Store


FransUrbo:
If I only change the `name` option to a `aws_ssm_parameter` resource, TF will create a NEW parameter, not rename the old one (or possibly delete the old, create the new).

### Terraform Version
0.9.8

### Affected Resource(s)
- aws_ssm_parameter

### Terraform Configuration Files
```hcl
provider "aws" {}
resource "aws_ssm_parameter" "test" {
  name                   = "test1"
  type                   = "SecureString"
  value                  = "secret"
}
```

### Expected Behavior
The old parameter should be either renamed to the new name, or possibly (if easier) simply been deleted and the new resource created.

### Actual Behavior
I ended up with TWO parameters, one with the old name and one with the new name.

### Steps to Reproduce
1. `terraform apply`
2. Change the `name` option to `test2`
3. `terraform apply`

FransUrbo:
PS. Please back port to 0.9 if possible!

FransUrbo:
Additionally, if I after running `apply` the second time then run `destroy` only the last parameter will be deleted. The initial one, `test1` will remain and have to be deleted manually.

FransUrbo:
No. I shouldn't have to enable overwrite to be able to rename things. That's not how anything else in TF works!

It doesn't complain at all if I change the `name` option, it will simply/silently create a new parameter.

stack72:
Can you confirm what the behaviour is using the cli?

FransUrbo:
This will of course create two resources:
```
aws ssm put-parameter --name test1 --type SecureString --value "test"
aws ssm put-parameter --name test2 --type SecureString --value "test"
```

This works of course, so this is what TF then needs to do, and I'm fine with that:
```
aws ssm put-parameter --name test1 --type SecureString --value "test1"
aws ssm delete-parameter --name test1
aws ssm put-parameter --name test2 --type SecureString --value "test"
```
That is, TF should know that the `name` option have changed, and would/should force a recreate of the resource (TF does that elsewhere for similar things). This is "bad", but because you don't get a "handle" or anything back in this case (that I could see), there's no other choice...
But on the other hand, a recreate in this case isn't that big of a deal - it's the same value and it's quick to do...

Although there's throttling somewhere (not sure how much and when - it only happened once and i wasn't paying attention to that at the moment).

stack72:
[...]

PR #1022, opened 2017-06-30
resource/aws_ssm_parameter: ForceNew on ssm_parameter rename

stack72:
Fixes: #1008

There doesn't seem to be a way to change the name of an AWS SSM
Parameter. Therefore, it should be marked as ForceNew.

```
% make testacc TEST=./aws TESTARGS='-run=TestAccAWSSSMParameter_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSSMParameter_ -timeout 120m
--- PASS: TestAccAWSSSMParameter_basic (25.06s)
--- PASS: TestAccAWSSSMParameter_update (43.21s)
--- PASS: TestAccAWSSSMParameter_changeNameForcesNew (40.84s)
--- PASS: TestAccAWSSSMParameter_secure (37.90s)
--- PASS: TestAccAWSSSMParameter_secure_with_key (73.26s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	220.297s
```

### DIFF
--- a/aws/resource_aws_ssm_parameter_test.go
+++ b/aws/resource_aws_ssm_parameter_test.go
@@ -50,6 +50,27 @@ func TestAccAWSSSMParameter_update(t *testing.T) {
 	})
 }
 
+func TestAccAWSSSMParameter_changeNameForcesNew(t *testing.T) {
+	before := acctest.RandString(10)
+	after := acctest.RandString(10)
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSSMParameterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSSSMParameterBasicConfig(before, "bar"),
+			},
+			{
+				Config: testAccAWSSSMParameterBasicConfig(after, "bar"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSSSMParameterDestroyed(before),
+				),
+			},
+		},
+	})
+}
+
 func TestAccAWSSSMParameter_secure(t *testing.T) {
 	name := acctest.RandString(10)
 	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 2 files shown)


aws/resource_aws_ssm_parameter_test.go
```diff
@@ -183,6 +202,16 @@ resource "aws_ssm_parameter" "foo" {
 `, rName, value)
 }
 
+func testAccAWSSSMParameterBasicConfigChangeName(rName string, value string) string {
```
radeksimko: What is the reason for having a copy of `testAccAWSSSMParameterBasicConfig` with a different name instead of just passing a different name as argument? 🙂

aws/resource_aws_ssm_parameter_test.go
```diff
@@ -50,6 +50,25 @@ func TestAccAWSSSMParameter_update(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSSMParameterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSSSMParameterBasicConfig(name, "bar"),
+			},
+			{
+				Config:             testAccAWSSSMParameterBasicConfigChangeName(name, "baz1"),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
```
radeksimko: I think this test isn't testing the new behaviour. 🤔  (a) It is changing the value, not the name. (b) even if it was changing the name, we'd still need to check existence of the old SSM parameter via API as it gets wiped from state, so `testAccCheckAWSSSMParameterDestroy` wouldn't catch it.

stack72:
@radeksimko you are correct - the test wasn't correct. This has been updated now and it checks the API to make sure that the old Parameter has been removed :)

aws/resource_aws_ssm_parameter_test.go
```diff
@@ -147,6 +168,29 @@ func testAccCheckAWSSSMParameterType(n string, v string) resource.TestCheckFunc
...
+		conn := testAccProvider.Meta().(*AWSClient).ssmconn
+
+		paramInput := &ssm.GetParametersInput{
+			Names: []*string{
+				aws.String(name),
+			},
+		}
+
+		resp, err := conn.GetParameters(paramInput)
+		if err != nil {
```
tombuildsstuff: as mentioned on Slack - I'm surprised this doesn't return a 404 but ¯\_(ツ)_/¯

stack72: It will return an empty response object :) But a response none the less

PR #1028, opened 2017-06-30
resource/aws_ecs_service: Prevent destroying tasks

nhuray:
Hi everyone,

This is a Pull request to fix an annoying issue when using `aws_ecs_service` described by @BrunoBonacci here [#9690](https://github.com/hashicorp/terraform/issues/9690). 

I already submit this PR in the past in the Terraform project (see [#11892](https://github.com/hashicorp/terraform/pull/11892))

The idea is to used `desired_count` to create or update the number of ECS service tasks, but the `desired_count` is not updated in the state when reading from the real-infrastructure (during `refresh`).

This prevents to destroy tasks  (running containers) if the number of tasks has been changed manually in AWS Console or by auto-scaling rules.

### DIFF
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -47,6 +47,7 @@ func resourceAwsEcsService() *schema.Resource {
 			"desired_count": {
 				Type:     schema.TypeInt,
 				Optional: true,
+				Computed: true,
 			},
 
 			"iam_role": {


radeksimko:
Hi @nhuray 
thanks for the PR.

I think I understand the issue here - we have experienced very similar one in `aws_autoscaling_group` which can be also scaled manually or autoscaled.

This PR would cause diff to be ignored even if users do actually scale manually, which is certainly not desired. How about we instead add `Computed: true` to the affected field (`desired_count`)?

That way we can either show diff to the user if they have `desired_count` specified in the config or just respect anything that comes out of the API if it's not defined in the config.

nhuray:
Hi @radeksimko 

Your proposal makes completely sense. 

After reading the documentation it's more clear:
```
	// If Computed is true, then the result of this value is computed
	// (unless specified by config) on creation.
```

I thought specifying `Compute: true` means the value is computed when the resource is created but I didn't realized that it's possible to define `Computed: true` and use the parameter also in the config.

I can modify my Pull Request. My concern is more about testing the change. Do you think we need tests for this change ? How to test the behaviour ?

Nicolas

catsby:
@radeksimko / @nhuray what is the workflow that supports this become? In order to create the resource, we "require" `desired_count` (`""` because it's actually optional in the schema, but the API requires it). 

With this new behavior though, you don't get the intended effect (not showing diffs in the event the count changes due to autoscaling) unless the user then removes the `desired_count` from the config? That need for two steps seems strange.. am I imagining it correctly?

Would it make more sense to introduce a new attribute (`initial_count` or something) that is only used during creation? That would allow users to omit `desired_count` completely, but still satisfy the creation. 

Thoughts?

radeksimko:
@catsby You're right, the field [does](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html#ECS-CreateService-request-desiredCount) seem to be required, so new attribute is probably better solution. 👍  I didn't notice that.

radeksimko:
> Do you think we need tests for this change ? How to test the behaviour ?

It's always helpful to have tests, but I can't think of any easy/reasonable way to test this, other than somehow triggering the scaling event. 🤔

nhuray:
@catsby @radeksimko 

Personally I will prefer to keep the `desired_count` parameter to define the number of ECS tasks desired and having a computed parameter `running_count` which effectively reflect the number of tasks running. This solution does not introduce breaking changes and match the [AWS API](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeServices.html) contract.

We might also to add a computed parameter `pending_count` defining the tasks scheduled but not started.

About tests, I will create an acceptance test doing:  
1. - Create the service with `desired_count=1`. 
     - Refresh the Terraform state
     - Ensure `desired_count=1` and  `running_count=1` 
2. - Scale with the AWS CLI (`desired_count=2`). 
     - Refresh the Terraform state
     - Ensure  `desired_count=1` and `running_count=2`

nhuray:
@catsby @radeksimko  @apparentlymart Any thoughts about my last comment ?

radeksimko:
> Personally I will prefer to keep the desired_count parameter to define the number of ECS tasks desired and having a computed parameter running_count which effectively reflect the number of tasks running.

While it may not be obvious in this context detecting drifts is a desired and expected behaviour of Terraform for all fields, unless documented otherwise. One-way fields (where we ignore changes) are intentionally very rare and even in these rare cases we tend to make sure the difference is clear to the user - i.e. they should not expect drifts to be detected for this field.

If a field is called `desired_count` and `desired_count = 5` as a user I'd expect Terraform to fix any drifts - i.e. if someone or something scales down/up the service Terraform will change it back to _desired_ state.

If we had a field called e.g. `initial_count` then it's more obvious that it will only be respected for the _initial_ creation, but ignored for all other purposes and the user would be still given a chance to bypass the autoscaling should they need/wish via `desired_count`.

TL;DR It's about satisfying different personas with different needs and contexts.

I hope the explanation makes sense - let me know.

bflad:
The `aws_ecs_service` resource documentation has been updated to include the `ignore_changes` configuration that allows this problem to be solved (works generically with all Terraform resources/attributes): https://www.terraform.io/docs/providers/aws/r/ecs_service.html#ignoring-changes-to-desired-count

```hcl
  # ... other configurations ...

  # Example: Create service with 2 instances to start
  desired_count = 2

  # Optional: Allow external changes without Terraform plan difference
  lifecycle {
    ignore_changes = ["desired_count"]
  }
}
```

PR #1033, opened 2017-07-02
Support for ECS credentials provider

fabienrenaud:
This addresses issue https://github.com/terraform-providers/terraform-provider-aws/issues/259

Adds the ECS credentials provider to the credentials chain before the EC2RoleProvider when AWS_CREDENTIALS_RELATIVE_UR environment variable is defined.

The code introduced merely mimics aws' defaults: https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L114

Added 1 new unit test.
I did not find acceptance tests for functions in auth_helpers.go so I believe I don't need to write any for this change...

Instructions to build a fully working terraform binary with this plugin would be appreciated. 

I do not know how to point my local terraform checkout to use my locally built terraform-provider-aws. Everything is in my ~/go dir and my GOPATH is configured to point ~/go/bin as instructed. govendor faq didn't help. So far, I've tested the new terraform binary by manually copying auth_helpers.go to the terraform project. I ran the new binary in a container on ECS and it did work:

    2017/07/01 21:13:11 [INFO] AWS EC2 instance detected via default metadata API endpoint, EC2RoleProvider added to the auth chain
    2017/07/01 21:13:11 [INFO] AWS Auth provider used: "CredentialsEndpointProvider"

and got me through making calls to dynamodb and run most of terraform plan (which failed before this change because the instance-profile hasn't enough permissions). But later in the execution, terraform seems to using an older version of the plugin and therefore falls back to the instance-profile role: 

    2017/07/02 18:12:03 [DEBUG] plugin: terraform-provider-aws_v0.1.2_x4: 2017/07/02 18:12:03 [INFO] No assume_role block read from configuration
    2017/07/02 18:12:03 [DEBUG] plugin: terraform-provider-aws_v0.1.2_x4: 2017/07/02 18:12:03 [INFO] Building AWS region structure
    2017/07/02 18:12:03 [DEBUG] plugin: terraform-provider-aws_v0.1.2_x4: 2017/07/02 18:12:03 [INFO] Building AWS auth structure
    2017/07/02 18:12:03 [DEBUG] plugin: terraform-provider-aws_v0.1.2_x4: 2017/07/02 18:12:03 [INFO] Setting AWS metadata API timeout to 100ms
    2017/07/02 18:12:03 [DEBUG] root.ecs_service- [...]
    2017/07/02 18:12:03 [DEBUG] plugin: terraform-provider-aws_v0.1.2_x4: 2017/07/02 18:12:03 [INFO] AWS EC2 instance detected via default metadata API endpoint, EC2RoleProvider added to the auth chain

Feel free to take over this change.

### DIFF
--- a/aws/auth_helpers.go
+++ b/aws/auth_helpers.go
@@ -12,7 +12,9 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	awsCredentials "github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/credentials/endpointcreds"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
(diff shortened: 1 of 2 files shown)


radeksimko:
Hi @fabienrenaud 
firstly I really apologise I missed your PR before reviewing #1425 

The good news is that the next AWS provider release is going to include the functionality.

I appreciate the time you invested into this, especially into the attached tests. I'm willing to review & merge a patch adding tests, but I think that better approach would be to use [`AWS_CONTAINER_CREDENTIALS_FULL_URI`](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L102) instead of custom ENV variable.

Also, it's usually better practice to just rely on the SDK to do its job in regards to defaults and avoid duplicating any logic and/or constants, where possible - in this case `169.254.170.2`.

Let me know what you think.

PR #1038, opened 2017-07-02
Fix OpsWorks 'set permissions' failing to set ssh access.

charles-at-geospock:
When the permissions being changed in OpsWorks were only the flags
for the sudo and ssh permissions of the current user, terraform
would send a 'level' parameter containing an empty string. This is
not valid, and AWS will respond to the request saying you cannot
change your own privilege level. Clearly an empty string ("") is
not taken as being 'unchanged', but nil is.

This change detects that an empty string would be sent to AWS and
if so, sends 'nil' in its place (which makes the parameter unset,
and allows the operation). Consequently, this allows the OpsWorks
SetPermissions operation to complete, changing the Sudo or SSH
permissions.

Fixes: https://github.com/terraform-providers/terraform-provider-aws/issues/458

### DIFF
--- a/aws/resource_aws_opsworks_permission.go
+++ b/aws/resource_aws_opsworks_permission.go
@@ -125,19 +125,15 @@ func resourceAwsOpsworksPermissionRead(d *schema.ResourceData, meta interface{})
 func resourceAwsOpsworksSetPermission(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*AWSClient).opsworksconn
 
-	/* Level must be supplied as 'nil' if we are not changing it */
-	level_string := d.Get("level").(string)
-	level_aws := aws.String(level_string)
-	if level_string == "" {
-		level_aws = nil
-	}
-
 	req := &opsworks.SetPermissionInput{
 		AllowSudo:  aws.Bool(d.Get("allow_sudo").(bool)),
 		AllowSsh:   aws.Bool(d.Get("allow_ssh").(bool)),
-		Level:      level_aws,
 		IamUserArn: aws.String(d.Get("user_arn").(string)),
 		StackId:    aws.String(d.Get("stack_id").(string)),
+	}
+
+	if v, ok := d.GetOk("level"); ok {
+		req.Level = aws.String(v.(string))
 	}
 
 	err := resource.Retry(2*time.Minute, func() *resource.RetryError {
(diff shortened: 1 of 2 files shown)


aws/resource_aws_opsworks_permission.go
```diff
@@ -125,10 +125,17 @@ func resourceAwsOpsworksPermissionRead(d *schema.ResourceData, meta interface{})
 func resourceAwsOpsworksSetPermission(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*AWSClient).opsworksconn
 
+	/* Level must be supplied as 'nil' if we are not changing it */
+	level_string := d.Get("level").(string)
+	level_aws := aws.String(level_string)
+	if level_string == "" {
+		level_aws = nil
+	}
```
radeksimko: We have a more idiomatic way of dealing with this exact problem, take a look at `d.GetOk()`

aws/resource_aws_opsworks_permission.go
```diff
@@ -125,10 +125,17 @@ func resourceAwsOpsworksPermissionRead(d *schema.ResourceData, meta interface{})
...
+
 	req := &opsworks.SetPermissionInput{
 		AllowSudo:  aws.Bool(d.Get("allow_sudo").(bool)),
 		AllowSsh:   aws.Bool(d.Get("allow_ssh").(bool)),
-		Level:      aws.String(d.Get("level").(string)),
+		Level:      level_aws,
 		IamUserArn: aws.String(d.Get("user_arn").(string)),
 		StackId:    aws.String(d.Get("stack_id").(string)),
 	}
```
radeksimko: The common pattern you can spot elsewhere in the codebase would 1st build the basic request with all required params and then set any optional ones if specified, e.g.

```go
if v, ok := d.GetOk("level"); ok {
    req.Level = aws.String(v.(string))
}
```

aws/resource_aws_opsworks_permission.go
charles-at-geospock: Ta; will fix.

aws/resource_aws_opsworks_permission.go
charles-at-geospock: That's a lot cleaner, thanks!

charles-at-geospock:
Changes to the code made as requested; makes things a bit nicer and now I know a little more. I'll look more at other examples of the code in the future to make sure that I can make it more similar.

Did you want the ChangeLog change removing as well? I can strip that out as well, if that would help.

charles-at-geospock:
Awesome, thanks very much. I expect I will be submitting more changes in the future, so I'll try to get those things right in the future and make it smoother all around.

PR #1050, opened 2017-07-04
add primary private ip to the documentation

AmineChikhaoui:
Fixing the documentation of the eni resource as I noticed that `private_ip` for primary private ip is not specified in the documentation eventhough I can see it's in the schema of the eni resource implementation.


AmineChikhaoui:
hm after looking further into the code, looks like this is not intended to be set from the user. Terraform doesn't allow assigning a primary private ip address apparently eventhough the api allows that ?

PR #1051, opened 2017-07-04
r/resource_aws_subnet: increase deletion timeout

s-urbaniak:
This adds a bump in the timeout for removing subnets as we experienced a high flake count upon cluster destruction. Introducing this timeout reduces flakes significantly

This is a stop-gap solution until resource providers get timeout override support.

Fixes https://github.com/coreos/tectonic-installer/issues/1245

/cc @radeksimko @jasminSPC

### DIFF
--- a/aws/resource_aws_subnet.go
+++ b/aws/resource_aws_subnet.go
@@ -294,7 +294,7 @@ func resourceAwsSubnetDelete(d *schema.ResourceData, meta interface{}) error {
 	wait := resource.StateChangeConf{
 		Pending:    []string{"pending"},
 		Target:     []string{"destroyed"},
-		Timeout:    10 * time.Minute,
+		Timeout:    30 * time.Minute,
 		MinTimeout: 1 * time.Second,
 		Refresh: func() (interface{}, string, error) {
 			_, err := conn.DeleteSubnet(req)


radeksimko:
See https://github.com/terraform-providers/terraform-provider-aws/pull/1052#issuecomment-312884633

radeksimko:
Hi @s-urbaniak 
as discussed in other places and conversations - raising the timeout should always be the last option unless the API explicitly exposes state (e.g. `pending`), so we know for sure it's worth waiting. 

This isn't subnet's case, so I believe it's avoidable. We want to prevent users with genuine dependency violations from waiting too long until they see the error.

By looking at the resources you use in `tectonic-installer` I think there's a chance it was already addressed in either https://github.com/terraform-providers/terraform-provider-aws/pull/1427 or https://github.com/terraform-providers/terraform-provider-aws/pull/1036

Generally speaking one of the most common reasons I bumped into this error is when AWS service managing its own ENIs left some interfaces around after deleting the resource (e.g. VPC-enabled Lambda, ELB, ALB, ...). Unfortunately this is also the reason why it's so hard to find the root cause.

With that I'm going to close this issue, but I'd advise you to run the destruction part of your smoke tests like this:

```sh
aws ec2 describe-network-interfaces --filters=Name=vpc-id,Values=$VPC_ID | tee network-interfaces-before.json
terraform destroy ...
aws ec2 describe-network-interfaces --filters=Name=vpc-id,Values=$VPC_ID | tee network-interfaces-after.json
```
so that we can inspect ENIs present before and after deletion if you do bump into this error again.

s-urbaniak:
@radeksimko thanks a lot for the heads-up! I am very glad and thankful to see this going forward, hence I am fine with closing this then. We are already using stock terraform without custom patches like this one in our pipeline now too.

/cc @cpanato @mxinden can we add Radek's hint to use `aws ec2 describe-network-interfaces ...` to our CI pipeline and add trace points in destroy failure loops to enable ENI inspection?

mxinden:
@s-urbaniak https://github.com/coreos/tectonic-installer/pull/2157 addresses this.

@radeksimko Thanks a lot for your help and advice!

PR #1052, opened 2017-07-04
r/resource_aws_security_group: increase deletion timeout

s-urbaniak:
This adds a bump in the timeout for removing security groups as we experienced a high flake count upon cluster destruction. Introducing this timeout reduces flakes significantly

This is a stop-gap solution until resource providers get timeout override support.

Fixes https://github.com/coreos/tectonic-installer/issues/1242

/cc @radeksimko @jasminSPC

### DIFF
--- a/aws/resource_aws_security_group.go
+++ b/aws/resource_aws_security_group.go
@@ -415,7 +415,7 @@ func resourceAwsSecurityGroupDelete(d *schema.ResourceData, meta interface{}) er
 		return fmt.Errorf("Failed to delete Lambda ENIs: %s", err)
 	}
 
-	return resource.Retry(5*time.Minute, func() *resource.RetryError {
+	return resource.Retry(30*time.Minute, func() *resource.RetryError {
 		_, err := conn.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
 			GroupId: aws.String(d.Id()),
 		})


radeksimko:
Hi @s-urbaniak 
I'd love to solve this problem, but I think 30mins is a bit too high for such resource and consequence. Keep in mind this timeout also applies to users that have a genuine `DependencyViolation` - i.e. users that have a dependency lying around will have to wait 30mins until they see the error. That's not great user experience.

The good news is that if you can consistently reproduce this problem, we can dig in and find out what's causing this, like we did https://github.com/terraform-providers/terraform-provider-aws/pull/1021 - do you mind following the same process? i.e. sending me the debug logs + relevant tf configs?

Thanks.

s-urbaniak:
@radeksimko I'll try to provoke the timeout as I did in the other precedence, sure.

jasminSPC:
Maybe It wouldn't hurt to maybe parameterise Delete timeout?

radeksimko:
@jasminSPC I'd prefer not to do it in this context for this resource. See my longer explanation in https://github.com/terraform-providers/terraform-provider-aws/pull/945#issuecomment-312916643

There's no good reason for an SG removal to take so long. Yes, APIs are eventually consistent and sometimes laggy, but that's matter of minutes, not half an hour. If the user has specified references between resources correctly (= we aren't deleting all things at the same time) then terraform will go resource-by-resource and the only reason for being stuck for so long in this situation is just resource not cleaning up after itself.

That is likely what's happening here, so when we understand what's holding the SG from removal we can do the cleanup ourselves without having to crank up timeouts.

Ninir:
Hey @s-urbaniak ,

As @radeksimko exposes it, this issue seems to me like a specific one since there are not so many people complaining about it.

Don't want to urge anyone on this one (😅) but find a solution for that if there is a real issue, or either close it if it's not reproductible.

Do you think you would have time to investigate?

Thanks!

s-urbaniak:
closing, as we will try to find more datapoints having `TF_TRACE` enabled and come up with a more sensible solution.

Ninir:
Thank you for that @s-urbaniak :)

PR #1064, opened 2017-07-05
Add support for Api Gateway method request validators.

betabandido:
This PR should fix terraform-providers#772

### DIFF
--- a/aws/resource_aws_api_gateway_method.go
+++ b/aws/resource_aws_api_gateway_method.go
@@ -76,6 +76,11 @@ func resourceAwsApiGatewayMethod() *schema.Resource {
 				ConflictsWith: []string{"request_parameters"},
 				Deprecated:    "Use field request_parameters instead",
 			},
+
+			"request_validator_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


aws/resource_aws_api_gateway_method.go
```diff
@@ -226,6 +236,22 @@ func resourceAwsApiGatewayMethodUpdate(d *schema.ResourceData, meta interface{})
...
 
+	if d.HasChange("request_validator_id") {
+		var request_validator_id *string
+		if v, ok := d.GetOk("request_validator_id"); ok {
+			// requestValidatorId cannot be an empty string; it must either be nil
+			// or it must have some value. Otherwise, updating fails.
+			if s := v.(string); len(s) > 0 {
+				request_validator_id = &s
+			}
+		}
```
catsby: I think we need to compare the old value here with the new one. If the old one is an ID, and the new one is `""`, then we're deleting a request valuator, and we should issue a `remove` update here on the `method`, correct? It doesn't look like we can otherwise remove anything as-is

catsby: I checked this out and tried manually, it seems we're sending `nil` in the operations below which is effectively doing what we want and removing it, so I guess this is find 👍

bflad:
The missing documentation has been submitted: https://github.com/terraform-providers/terraform-provider-aws/pull/5110

Issue #1076, opened 2017-07-07
Request Support for VPC ClassicLink DNS Support


peimanja:
### Terraform Version
v0.9.11

### Affected Resource(s)
 - aws_vpc

### Terraform Configuration Files
```
resource "aws_vpc" "main" {
  cidr_block  = "10.0.0.0/16"
  enable_classiclink  = true
  enable_dns_classiclink_support = true
}
```
### Expected Behavior
Create a VPC with ClassicLink DNS Support enabled

### Actual Behavior
Nothing, because this syntax isn't supported yet.

PR #1079, opened 2017-07-07
resource/aws_vpc: Add support for classiclink_dns_support

stack72:
Fixes: #1076

```
% make testacc TEST=./aws TESTARGS='-run=TestAccAWSVpc'
TF_ACC=1 go test ./aws -v -run=TestAccAWSVpc -timeout 120m
=== RUN   TestAccAWSVpcEndpoint_importBasic
--- PASS: TestAccAWSVpcEndpoint_importBasic (68.54s)
=== RUN   TestAccAWSVpc_importBasic
--- PASS: TestAccAWSVpc_importBasic (53.29s)
=== RUN   TestAccAWSVpcEndpointRouteTableAssociation_basic
--- PASS: TestAccAWSVpcEndpointRouteTableAssociation_basic (59.57s)
=== RUN   TestAccAWSVpcEndpoint_basic
--- PASS: TestAccAWSVpcEndpoint_basic (58.48s)
=== RUN   TestAccAWSVpcEndpoint_withRouteTableAndPolicy
--- PASS: TestAccAWSVpcEndpoint_withRouteTableAndPolicy (117.51s)
=== RUN   TestAccAWSVpcEndpoint_WithoutRouteTableOrPolicyConfig
--- PASS: TestAccAWSVpcEndpoint_WithoutRouteTableOrPolicyConfig (53.33s)
=== RUN   TestAccAWSVpcEndpoint_removed
--- PASS: TestAccAWSVpcEndpoint_removed (49.14s)
=== RUN   TestAccAWSVpc_basic
--- PASS: TestAccAWSVpc_basic (45.19s)
=== RUN   TestAccAWSVpc_enableIpv6
--- PASS: TestAccAWSVpc_enableIpv6 (132.93s)
=== RUN   TestAccAWSVpc_dedicatedTenancy
--- PASS: TestAccAWSVpc_dedicatedTenancy (46.52s)
=== RUN   TestAccAWSVpc_tags
--- PASS: TestAccAWSVpc_tags (87.11s)
=== RUN   TestAccAWSVpc_update
--- PASS: TestAccAWSVpc_update (85.61s)
=== RUN   TestAccAWSVpc_DisabledDnsSupport
--- PASS: TestAccAWSVpc_DisabledDnsSupport (43.34s)
=== RUN   TestAccAWSVpc_classiclinkOptionSet
--- PASS: TestAccAWSVpc_classiclinkOptionSet (45.99s)
=== RUN   TestAccAWSVpc_classiclinkDnsSupportOptionSet
--- PASS: TestAccAWSVpc_classiclinkDnsSupportOptionSet (52.46s)
PASS
ok	github.com/terraform-providers/terraform-provider-aws/aws	1009.082s
```

### DIFF
--- a/aws/resource_aws_vpc.go
+++ b/aws/resource_aws_vpc.go
@@ -58,6 +58,12 @@ func resourceAwsVpc() *schema.Resource {
 				Computed: true,
 			},
 
+			"enable_classiclink_dns_support": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
+			},
+
 			"assign_generated_ipv6_cidr_block": {
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


aws/resource_aws_vpc.go
```diff
@@ -235,6 +241,30 @@ func resourceAwsVpcRead(d *schema.ResourceData, meta interface{}) error {
...
+		VpcIds: []*string{&vpcid},
+	}
+
+	respClassiclinkDnsSupport, err := conn.DescribeVpcClassicLinkDnsSupport(DescribeClassiclinkDnsOpts)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "UnsupportedOperation" {
+			log.Printf("[WARN] VPC Classic Link DNS Support is not supported in this region")
+		} else {
+			return err
+		}
```
catsby: Should we default this value to `false`? It looks like here in the error case the value will not get set at all (`d.Set` only called in the `else` clause here). I'm imagining a situation where a user has an account/region that doesn't support this, they try to enable it with `true` here, it fails, but we never set it ourselves to false and so they end up with `true` in the state file, which doesn't reflect reality. Is that possible or am I missing something?

stack72: This follows *exactly* the same pattern as `enable_classiclink` - that seems to work well so therefore, we don't set it to default - thoughts?

catsby: 👍

PR #1080, opened 2017-07-07
resource/aws_lambda_function: Add support for lambda_function vpc_config update

stack72:
Fixes: #1073

Changing to not ForceNew on any chances to vpc_config

Also, updates to `dead_letter_config` were supported in the code, but
the schema said that it would ForceNew on changes. So removed this
restriction

screen shot 2017-07-07 at 17 49 40

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -62,7 +62,6 @@ func resourceAwsLambdaFunction() *schema.Resource {
 			"dead_letter_config": {
 				Type:     schema.TypeList,
 				Optional: true,
-				ForceNew: true,
 				MinItems: 0,
 				MaxItems: 1,
 				Elem: &schema.Resource{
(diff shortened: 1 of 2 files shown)


aws/resource_aws_lambda_function.go
```diff
@@ -600,6 +600,32 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
...
 		}
 	}
+	if d.HasChange("vpc_config") {
+		vpcConfigRaw := d.Get("vpc_config").([]interface{})
+		vpcConfig, ok := vpcConfigRaw[0].(map[string]interface{})
+		if !ok {
+			return errors.New("vpc_config is <nil>")
+		}
+
+		if vpcConfig != nil {
```
Ninir: Without looking too much into the Terraform core, could this be something other than nil, like empty or so?
(Just wondering if the check is needed, since the check above guard us from nil structures, but may be missing something)

stack72: This is just an extra panic guard :)

PR #1083, opened 2017-07-07
Stored a sha256 hashed password for dms endpoint

iceycake:
https://github.com/hashicorp/terraform/pull/12554 was not brought back into this new project.  We need to continue discuss what is the better way to implement this.

### DIFF
--- a/aws/resource_aws_dms_endpoint.go
+++ b/aws/resource_aws_dms_endpoint.go
@@ -94,6 +94,7 @@ func resourceAwsDmsEndpoint() *schema.Resource {
 				Type:      schema.TypeString,
 				Optional:  true,
 				Sensitive: true,
+				StateFunc: hashSum,
 			},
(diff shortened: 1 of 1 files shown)


grubernaut:
Going to echo the same comment I left here, https://github.com/terraform-providers/terraform-provider-aws/pull/1084#issuecomment-314092829, for this change. But, as before, will defer to @catsby

bflad:
Hi @iceycake 👋 Thanks for submitting this. Unfortunately this is a breaking change and usage of `StateFunc` is discouraged in the Terraform Provider SDK, similar to #1084, so we cannot accept this as-is. A solution similar to #1740 might be preferable, although the maintainers would prefer to have this discussed in a design issue that can be applied more holistically across all potential sensitive schema attributes before working on targeted resource implementations. Thank you again.

PR #1090, opened 2017-07-10
Append the info about IAM Server Certificate import

cazorla19:
I noticed that we missed an information about IAM certificates import. From the user side it looks like a Terraform cannot do this. So, I tried import by `id` and by `name` both. Only the last one succeeded.
So, please, update the docs.

### DIFF
--- a/website/docs/r/iam_server_certificate.html.markdown
+++ b/website/docs/r/iam_server_certificate.html.markdown
@@ -121,7 +121,7 @@ The following arguments are supported:
 IAM Server Certificates can be imported using the `name`, e.g.
 
 ```
-$ terraform import aws_iam_server_certificate.sertificate example.com-certificate-until-2018
+$ terraform import aws_iam_server_certificate.certificate example.com-certificate-until-2018
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/r/iam_server_certificate.html.markdown
```diff
@@ -116,6 +116,13 @@ The following arguments are supported:
 * `name` - The name of the Server Certificate
 * `arn` - The Amazon Resource Name (ARN) specifying the server certificate.
 
+## Import
+
+IAM Server Certificates can be imported using the `name`, e.g.
+
+```
+$ terraform import aws_iam_server_certificate.sertificate example.com-certificate-until-2018
```
Ninir: is the `sertificate` wanted there, with a `s`?

stack72: @Ninir I am going to make the change on this branch and then merge this - I believe it was a typo :)

stack72:
Hi @cazorla19 

Thanks for this work :) 

Paul

website/docs/r/iam_server_certificate.html.markdown
cazorla19: Surely, not. Thanks for fixing my typo!

PR #1099, opened 2017-07-10
Ensure bootstrap actions are ordered the way they are defined

apetresc:
Currently, the bootstrap actions are a TypeSet, so their relative ordering is random. This was fixed by an earlier PR for the args list for each action, but not for the actions themselves. This fixes that.


apetresc:
Excellent points. I'll address these ASAP.

Ninir:
Hi @apetresc 

Did you find time to work on that? :)
If you need help, please tell us!

apetresc:
@Ninir: I believe I have just about finished it – writing the test now :)

Ninir:
Hey @apetresc 

Awesome! Tell us if we can be of any help! 👍

apetresc:
Hey guys! (@Ninir and/or @radeksimko)

So, I was in the process of writing a test for this migration, and decided as a sanity check to also reproduce the crash caused by parsing an old state file with this patch applied. Much to my surprise, and for reasons I'm unsure of, there is no crash - even without any migration at all, Terraform is doing exactly the right thing. That is, it's reading the old state, and re-applying it as a list.

In painful detail, here's the steps I'm taking:

1. With latest Terraform v0.10.4 and default aws provider (checksum `067ed7bf24c0f111bef09f3f3b0d113a8d6a8422ec90be4847efbe2f45f3e0b7`), I am `terraform apply`-ing the minimal EMR cluster example on [the `aws_emr_cluster` resource documentation page](https://www.terraform.io/docs/providers/aws/r/emr_cluster.html), but with a few extra `bootstrap_action`s thrown in so it's easy to track their ordering.
2. The generated state file shows that the `boostrap_action` field is a Set:
    ```json
                    "bootstrap_action.#": "4",
                    "bootstrap_action.1293383300.args.#": "2",
                    "bootstrap_action.1293383300.args.0": "instance.isMaster=true",
                    "bootstrap_action.1293383300.args.1": "echo bootstrap action 1",
                    "bootstrap_action.1293383300.name": "runif",
                    ... (etc)
    ```
    Furthermore, the AWS console clearly shows that the bootstrap actions are in the wrong order (1, 3, 4, 2)
3. At this point, I switch my branch of `terraform-provider-aws` over to commit e2f4267b (which does NOT have any sort of migration in place, but does contain this patch), build it, and put it in the `.terraform/plugins` folder, updating `lock.json`.
4. I run `terraform apply` and expect the crash that @radeksimko predicted. Instead, terraform simply decides it needs to recreate the `aws_emr_cluster` resource, with `bootstrap_action` now a list:
    ```json
                    bootstrap_action.#:                                 "" => "4"
                    bootstrap_action.0.args.#:                          "" => "2"
                    bootstrap_action.0.args.0:                          "" => "instance.isMaster=true"
                    bootstrap_action.0.args.1:                          "" => "echo bootstrap action 1"
                    bootstrap_action.0.name:                            "" => "runif"
                    bootstrap_action.0.path:                            "" => "s3://elasticmapreduce/bootstrap-actions/run-if"
                    bootstrap_action.1.args.#:                          "" => "2"
                    bootstrap_action.1.args.0:                          "" => "instance.isMaster=true"
                    ... (etc)
    ```
5. The apply finishes without any problems, and indeed the AWS console now shows the bootstrap actions in the right order (1, 2, 3, 4). The state file shows the same thing.

So at this point, I don't see any reason for the migration. Judging from this, as well as the fact that the only thing my migration even _did_ was read the state and write it back out again, leads me to think that Terraform already has some sort of casting logic in place between Sets and Lists that just handles this transparently. I can submit the no-op migration if you really insist, but I can't really submit a _test_ for that migration because I can't figure out how to create a failing test :)

Hopefully this is helpful!

Ninir:
Hi @apetresc ,

Thanks for your feedback on this.

> So at this point, I don't see any reason for the migration. Judging from this, as well as the fact that the only thing my migration even did was read the state and write it back out again, leads me to think that Terraform already has some sort of casting logic in place between Sets and Lists that just handles this transparently. 

From a User Experience point-of-view, you don't expect a change to your infrastructure when upgrading from a version to an other.
A migration would allow to make all of this transparent for the user: we migrate it internally without the user knowing it.

Even if no errors is thrown, having to recreate the EMR infra because of such change seems like a no-go: if it was about EMR API and the way it is managed, then the recreation wouldn't matter. 
In our case we are switching types, which is all about Terraform.

Does it makes sense to you?

 > I can submit the no-op migration if you really insist, but I can't really submit a test for that migration because I can't figure out how to create a failing test :)

Why do you want a failing test? (it really is a question to better clarify it in my mind)

I think we could use this [kind of test example](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance_migrate_test.go), which checks the TF state in the migration. Thougths?

apetresc:
Oh, I see - I thought you just wanted the migration in order to prevent a crash when loading a statefile created with a previous version of Terraform (Radek had said "If we didn't do the migration then Terraform would simply crash on line 285 for those users"), which turned out not to be the case.

Using a migration to do a one-time sort of the TypeSet into a List in order to prevent resource re-creation doesn't really make sense to me. The _actual_ EMR cluster running on AWS will still have the bootstrap actions in the wrong order, so the next time Terraform actually uses the API to check the state of the cluster, it will notice that the state is wrong, whether it is stored as a List or a Set. 

Fundamentally, the current version of `aws_emr_cluster` resource *is wrong* in that it creates the EMR cluster with the bootstrap actions in a random order. I don't see how a migration can trick Terraform into thinking the order is correct when it isn't, nor do I see why we would want to do that. Current `aws_emr_cluster` resources _should_ be recreated (even though the "mistake" is relatively minor, since bootstrap action ordering doesn't matter in most cases... but when it does matter, it *really* matters :smile:)

As I was typing this, I realized that a migration _would_ save a recreate if the `bootstrap_action` list happened to have length 0 or 1, since then the sort from the AWS API would always match the migrated state file. Is that what you meant? Because in that case, you're right, I can fix this much :smile:

apetresc:
Okay, I just did some manual testing on a blank configuration and I can confirm:
  - For old configurations with 0 or 1 `bootstrap_action`s, this change does *not* require a recreate (since the sorting of the automatically-casted List will match the sorting of the old Set)
  - For old configurations with 2+ `bootstrap_action`s, a recreate is required unless (by coincidence) the Set sorting happened to be the correct sorting (1/2 chance for 2 actions, 1/6 chance for 3 actions, ..., 1/(n!) chance for n actions). For the reasons I stated above, I don't think this could/should be "corrected", since clusters with a wrong sorting on bootstrap actions *are* being meaningfully semantically affected by this change.

apetresc:
Pinging @Ninir @radeksimko 

Any updates on this? Just to reiterate, according to my testing, setups with 0-1 bootstrap actions *are* fully backwards compatible without a recreate, and setups with 2+ bootstrap actions *may* require a recreate since the ordering of bootstrap actions _is_ actually changing, and it would be neither desirable nor easily-achievable to change this.

Ninir:
Hi @apetresc 

Even if it work in your case, we need to handle every possible case using a state migration here.

Do you think you would be able to achieve this? here is an example about how to query the AWS API in the migration and write the state: https://github.com/terraform-providers/terraform-provider-aws/pull/1195/files

Let me know if you need anything more :)

apetresc:
@Ninir: I still seem to be being misunderstood. I am **_not_** simply making a statement about how it "works for me". I'm claiming that what you are asking for is fundamentally impossible.

The sequence of events is:
1. The user's .tf file (before this patch) specified bootstrap actions in the order `[A, B, C]`.
2. The user applied this file, but Terraform messed up the order to `[B, C, A]`, and so the EMR cluster was created with bootstrap actions in the order `[B, C, A]`
3. The user now upgrades Terraform and has this patch.
4. The user re-applies the exact same .tf file, but because Terraform now understands that the correct order is `[A, B, C]`, it realizes that the EMR cluster was created incorrectly with `[B, C, A]`. Since the EMR cluster **_is actually semantically wrong_**, and bootstrap actions are not editable, it recreates it. It **should** recreate it, it never matched what was in the .tf file, this was just undetected before.

I don't see how a state migration would help here. The only conceivable way to work around it with a state migration is to somehow store the incorrect old ordering and tell Terraform to ignore that it's wrong just this once, until something else forces a recreate anyway. This seems like a terrible hack for what is essentially an ephemeral resource anyway.

Am I misunderstanding some part of this process? In which step is my mistake?

Ninir:
@apetresc Ok it is now crystal clear, thanks for taking the time to explain with other words 🙂 .

You are indeed totally right saying this can not be fixed by a migration, especially if it is ordered as `[B, C, A]` on the AWS side.

> The only conceivable way to work around it with a state migration is to somehow store the incorrect old ordering and tell Terraform to ignore that it's wrong just this once, until something else forces a recreate anyway.

I think this is the way to go: migrating the state BUT warning people that they should taint this resource since it avoids breaking things and delaying the merge too much.

Do you think you could get this done? Let me know if I can be of any help here.

phani308:
> @apetresc Ok it is now crystal clear, thanks for taking the time to explain with other words 🙂 .
> 
> You are indeed totally right saying this can not be fixed by a migration, especially if it is ordered as `[B, C, A]` on the AWS side.
> 
> > The only conceivable way to work around it with a state migration is to somehow store the incorrect old ordering and tell Terraform to ignore that it's wrong just this once, until something else forces a recreate anyway.
> 
> I think this is the way to go: migrating the state BUT warning people that they should taint this resource since it avoids breaking things and delaying the merge too much.
> 
> Do you think you could get this done? Let me know if I can be of any help here.

@Ninir  @apetresc I am still facing EMR bootstrap actions executing in random order even with 0.11.8t terraform version. Is this resolved, I am passing a list of 3 actions in required order but it always takes (b, a,c) ORDER

ALFNeT:
@phani308 I'm also facing the same issues, did you find a PR or Issue where to follow this?

PR #1121, opened 2017-07-11
docs/aws_redshift_cluster: Fix default value of skip_final_snapshot

minamijoyo:
The default value of `skip_final_snapshot` was changed to `false`
with the following change, but the document was not updated.
https://github.com/hashicorp/terraform/pull/11695


stack72:
Thanks so much for this @minamijoyo 

Apologies for the mission

PR #1127, opened 2017-07-12
New Resource: aws_kms_external_key

MCyprien:
Hi,

this PR adds origin to aws_kms_key and manage EXTERNAL keys.

Thanks for review :)

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -359,6 +359,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_kinesis_stream":                           resourceAwsKinesisStream(),
 			"aws_kms_alias":                                resourceAwsKmsAlias(),
 			"aws_kms_key":                                  resourceAwsKmsKey(),
+			"aws_kms_external_key":                         resourceAwsKmsExternalKey(),
 			"aws_lambda_function":                          resourceAwsLambdaFunction(),
 			"aws_lambda_event_source_mapping":              resourceAwsLambdaEventSourceMapping(),
 			"aws_lambda_alias":                             resourceAwsLambdaAlias(),
--- a/aws/resource_aws_kms_external_key.go
+++ b/aws/resource_aws_kms_external_key.go
@@ -1,0 +1,399 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsKmsExternalKey() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsKmsExternalKeyCreate,
+		Read:   resourceAwsKmsExternalKeyRead,
+		Update: resourceAwsKmsExternalKeyUpdate,
+		Delete: resourceAwsKmsExternalKeyDelete,
+		Exists: resourceAwsKmsExternalKeyExists,
+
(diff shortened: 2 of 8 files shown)


radeksimko:
Hi @MCyprien 
thank you for bringing this KMS feature to our attention. It looks like something that's worth supporting.

Given the nature of external keys, do you think it's worth implementing this as a separate resource, e.g. `aws_kms_external_key` or `aws_kms_key_external`? It seems there are API/SDK methods exposed to do the import and removal (`kms/ImportKeyMaterial` + `kms/DeleteImportedKeyMaterial`), both accept KMS Key ID, so the key still needs to be created, but I'm thinking this approach of a separate resource would allow the user to manage the whole lifecycle instead of creating a resource which is half-way finished (in `PendingImport` state).

MCyprien:
Hi @radeksimko,
First, thanks for the review :)

I hesitated when doing this feature if another resource would be better, now it just solved the problem.
I will change the code and come back for another review.

I will be back ;)

MCyprien:
Hi @radeksimko ,

It looks like travis is stuck on my PR, can you help me fix that?
Else my code is ready and tests are OK to me.

Can someone review the code and tell me if I need to change anything ?

Thanks

radeksimko:
I'm not sure we understood each other in the initial conversation about the split-up. Maybe I wasn't clear enough - sorry.

The main motivation and reason I was suggesting it is because we can leverage these two API calls to actually manage the key material, not just key "placeholder":

 - `kms/ImportKeyMaterial`
 - `kms/DeleteImportedKeyMaterial`

I don't see either of these two calls in the CRUD or related fields in the schema. Do you mind adding that functionality to the PR? 🙂 

I'd expect to see a few fields added to the schema, probably 3 (`encrypted_key_material`, `expiration_model`, `valid_to`)?

Thanks.

MCyprien:
Hi,

I don't really understand what you want :/
When you create an EXTERNAL key in KMS, it's like an empty shell for the key. To upload the key material, AWS give you a public key, to encrypt your key material and a token, to upload your key. But the two of them are only available for 24h, so the `encrypted_key_material` is not valid anymore after this time. 
`expiration_model` and `valid_to` are set when uploading the key material, I can add this parameters to the resource but it will always be set by kms and not by terraform.
So I think `kms/ImportKeyMaterial` cannot be added because of the above reasons.
I can add `kms/DeleteImportedKeyMaterial` if needed to delete the key material already existing.

What are your thought on that?

MCyprien:
Hi @radeksimko,
Let me know if anything is blocking this PR.

Best regards,

johntdyer:
Would love to see this soon !

LucaLanziani:
I would like to see this in ASAP too, we are currently having problem on managing an EXTERNAL key our self

MCyprien:
Hi @radeksimko ,
Let me know if anything is blocking this PR.

Best regards,

opsj:
@MCyprien any update on this?

MCyprien:
@opsj I'm sorry, I don't have much time right now to come back on this PR.

Also I don't want to work on something that is pending for too long so if someone tells me what is wrong with this PR and what to do for this PR to be mergeable, I'll try to find time to work on it, else this PR will stay like that :)

petems:
@MCyprien I'm working with a team that is looking for this feature! I'm going to see if I can cherry-pick your commit and pull it into a new pull-request to merge in 👍

bflad:
Hi @MCyprien 👋 Your work will be included as part of #8066, which will hopefully be released this week. Thanks for your contribution here.

bflad:
This has been released in [version 2.4.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#240-march-29-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #1136, opened 2017-07-12
Updating documentation

thomasbiddle:
**Contribution Guidelines**
This is an update to the documentation. Reasoning is explained in the below paragraph. This is for the current version of Terraform and should have no issues being deployed immediately AFAIK.

---

In the `aws_vpc_peering_connection` resource the `peer_owner_id` is actually optional.

This can be seen here: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_vpc_peering_connection.go#L30

The documentation after commit already explains what it defaults to (The current account ID).

### DIFF
--- a/website/docs/r/vpc_peering.html.markdown
+++ b/website/docs/r/vpc_peering.html.markdown
@@ -74,7 +74,7 @@ more information.
 
 The following arguments are supported:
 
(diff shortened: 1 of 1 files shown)


stack72:
Hi @thomasbiddle 

Thanks for that - you are indeed correct :) So this LGTM!

Paul

Issue #1081, opened 2017-07-07
AWS RDS missing dbi-resource-id reference for IAM auth


opetrovskyi:
Hi there,

This is not a bug but a feature request. 
AWS Aurora cluster have a feature to use IAM role for authorization. To create a IAM policy you have to use "Resource ID" of a cluster instance in resource arn line. aws_rds_cluster or aws_rds_cluster_instance do not provide with "Resource ID" reference though it has 2 references called "identifier"  and "id" but they both refer to one entity - The Instance identifier. The right reference is "dbi-resource-id". It'll be very nice to have this. 
Here is the AWS doc of using IAM roles for auth in RDS database.
http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html

the example policy has such lines:
"Resource": [
             "arn:aws:rds-db:us-west-2:12345678:dbuser:**db-12ABC34DEFG5HIJ6KLMNOP78QR**/jane_doe",
the bold string is the dbi-resource-id missing in aws_rds_ reference.

stack72:
Hi @opetrovskyi 

I believe we already have this value available as an `rds_cluster` exported attribute

You should be able to use the value `cluster_resource_id` - we set this value as follows:

```
d.Set("cluster_resource_id", dbc.DbClusterResourceId)
```

Please can you tell me if this works for you?

Thanks

Paul

opetrovskyi:
Hi @stack72,

I've checked this reference and it is not the one needed.
reference you mentioned is a cluster reference, like
rds_cluster_resource_id = cluster-OL1111111122222233333Y
but i was talking about the database(instance) reference. It'll be something like this:
db-12ABC34DEFG5HIJ6KLMNOP78QR

stack72:
Hi @opetrovskyi 

Got it - will add it today for you :)

Paul

PR #1142, opened 2017-07-13
resource/rds_cluster_instance: Export the RDI Resource ID for the instance

stack72:
Fixes: #1081

### DIFF
--- a/aws/resource_aws_rds_cluster_instance.go
+++ b/aws/resource_aws_rds_cluster_instance.go
@@ -109,6 +109,11 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"dbi_resource_id": {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_rds_cluster_instance.go
```diff
@@ -109,6 +109,11 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"dbi_resource_id": {
+				Type:     schema.TypeBool,
```
grubernaut: Is this a bool, seems to be a string(?)

stack72:
Crap! Good catch - will fix and merge on green :)

stack72:
Hi @grubernaut 

This is now fixed and the tests are passing as expected:

screen shot 2017-07-14 at 13 53 20

P

PR #1162, opened 2017-07-15
CodeDeploy: Blue/Green deployments and In-Place deployments with traffic control.

niclic:
Issue: https://github.com/terraform-providers/terraform-provider-aws/issues/504

This PR adds Blue/Green Deployment support to CodeDeploy Deployment Groups, as well as supporting In-Place deployments with Traffic Control (i.e. Load Balancer).

It was originally opened on February 4, 2017, as https://github.com/hashicorp/terraform/pull/11700.

This PR adds the following resources to a Deployment Group:

* DeploymentStyle
* LoadBalancerInfo
* BlueGreenDeploymentConfiguration

The original work is contained in a single new commit. The original PR retains all original commits and discussion.

These enhancements can also be used for in-place deployments with traffic control.

```hcl
deployment_style {
  deployment_option = "WITH_TRAFFIC_CONTROL"
  deployment_type = "IN_PLACE"
}
```

Since this scenario was not supported when the original PR was created, I created a new commit that includes a test for this specific configuration, as well as updating the documentation to reflect this recent change. No code changes were required to support this scenario.

### DIFF
--- a/aws/resource_aws_codedeploy_deployment_group.go
+++ b/aws/resource_aws_codedeploy_deployment_group.go
@@ -716,12 +716,13 @@ func buildAlarmConfig(configured []interface{}) *codedeploy.AlarmConfiguration {
 // expandDeploymentStyle converts a raw schema list containing a map[string]interface{}
 // into a single codedeploy.DeploymentStyle object
 func expandDeploymentStyle(list []interface{}) *codedeploy.DeploymentStyle {
-	result := &codedeploy.DeploymentStyle{}
 	if len(list) == 0 || list[0] == nil {
-		return result
+		return nil
 	}
 
 	style := list[0].(map[string]interface{})
+	result := &codedeploy.DeploymentStyle{}
+
 	if v, ok := style["deployment_option"]; ok {
 		result.DeploymentOption = aws.String(v.(string))
 	}
@@ -739,9 +740,8 @@ func expandLoadBalancerInfo(list []interface{}) *codedeploy.LoadBalancerInfo {
 		return nil
 	}
 
-	loadBalancerInfo := &codedeploy.LoadBalancerInfo{}
-
 	lbInfo := list[0].(map[string]interface{})
+	loadBalancerInfo := &codedeploy.LoadBalancerInfo{}
 
 	if attr, ok := lbInfo["elb_info"]; ok {
 		elbs := attr.(*schema.Set).List()
@@ -773,12 +773,12 @@ func expandLoadBalancerInfo(list []interface{}) *codedeploy.LoadBalancerInfo {
 // expandBlueGreenDeploymentConfig converts a raw schema list containing a map[string]interface{}
 // into a single codedeploy.BlueGreenDeploymentConfiguration object
 func expandBlueGreenDeploymentConfig(list []interface{}) *codedeploy.BlueGreenDeploymentConfiguration {
-	blueGreenDeploymentConfig := &codedeploy.BlueGreenDeploymentConfiguration{}
 	if len(list) == 0 || list[0] == nil {
-		return blueGreenDeploymentConfig
+		return nil
 	}
 
 	config := list[0].(map[string]interface{})
+	blueGreenDeploymentConfig := &codedeploy.BlueGreenDeploymentConfiguration{}
 
 	if attr, ok := config["deployment_ready_option"]; ok {
 		a := attr.([]interface{})
@@ -926,10 +926,8 @@ func alarmConfigToMap(config *codedeploy.AlarmConfiguration) []map[string]interf
 // flattenDeploymentStyle converts a codedeploy.DeploymentStyle object
 // into a []map[string]interface{} list containing a single item
 func flattenDeploymentStyle(style *codedeploy.DeploymentStyle) []map[string]interface{} {
-	result := make([]map[string]interface{}, 0, 1)
-
 	if style == nil {
-		return result
+		return nil
 	}
 
 	item := make(map[string]interface{})
@@ -939,6 +937,8 @@ func flattenDeploymentStyle(style *codedeploy.DeploymentStyle) []map[string]inte
 	if style.DeploymentType != nil {
 		item["deployment_type"] = *style.DeploymentType
 	}
+
+	result := make([]map[string]interface{}, 0, 1)
 	result = append(result, item)
 	return result
 }
@@ -946,10 +946,8 @@ func flattenDeploymentStyle(style *codedeploy.DeploymentStyle) []map[string]inte
 // flattenLoadBalancerInfo converts a codedeploy.LoadBalancerInfo object
 // into a []map[string]interface{} list containing a single item
 func flattenLoadBalancerInfo(loadBalancerInfo *codedeploy.LoadBalancerInfo) []map[string]interface{} {
-	result := make([]map[string]interface{}, 0, 1)
-
 	if loadBalancerInfo == nil {
-		return result
+		return nil
 	}
 
 	elbs := make([]interface{}, 0, len(loadBalancerInfo.ElbInfoList))
@@ -969,18 +967,18 @@ func flattenLoadBalancerInfo(loadBalancerInfo *codedeploy.LoadBalancerInfo) []ma
 	lbInfo := make(map[string]interface{})
 	lbInfo["elb_info"] = schema.NewSet(loadBalancerInfoHash, elbs)
 	lbInfo["target_group_info"] = schema.NewSet(loadBalancerInfoHash, targetGroups)
-	result = append(result, lbInfo)
 
+	result := make([]map[string]interface{}, 0, 1)
+	result = append(result, lbInfo)
 	return result
 }
 
 // flattenBlueGreenDeploymentConfig converts a codedeploy.BlueGreenDeploymentConfiguration object
 // into a []map[string]interface{} list containing a single item
 func flattenBlueGreenDeploymentConfig(config *codedeploy.BlueGreenDeploymentConfiguration) []map[string]interface{} {
-	list := make([]map[string]interface{}, 0)
 
 	if config == nil {
-		return list
+		return nil
 	}
 
 	m := make(map[string]interface{})
@@ -1024,6 +1022,7 @@ func flattenBlueGreenDeploymentConfig(config *codedeploy.BlueGreenDeploymentConf
 		m["terminate_blue_instances_on_deployment_success"] = append(c, blueInstanceTerminationOption)
 	}
 
+	list := make([]map[string]interface{}, 0)
 	list = append(list, m)
 	return list
 }
--- a/aws/resource_aws_codedeploy_deployment_group.go
+++ b/aws/resource_aws_codedeploy_deployment_group.go
@@ -749,8 +749,13 @@ func expandLoadBalancerInfo(list []interface{}) *codedeploy.LoadBalancerInfo {
 
 		for _, v := range elbs {
 			elb := v.(map[string]interface{})
+			name, ok := elb["name"].(string)
+			if !ok {
+				continue
+			}
+
 			loadBalancerInfo.ElbInfoList = append(loadBalancerInfo.ElbInfoList, &codedeploy.ELBInfo{
-				Name: aws.String(elb["name"].(string)),
+				Name: aws.String(name),
 			})
 		}
 	}
@@ -761,8 +766,13 @@ func expandLoadBalancerInfo(list []interface{}) *codedeploy.LoadBalancerInfo {
 
 		for _, v := range targetGroups {
 			targetGroup := v.(map[string]interface{})
+			name, ok := targetGroup["name"].(string)
+			if !ok {
+				continue
+			}
+
 			loadBalancerInfo.TargetGroupInfoList = append(loadBalancerInfo.TargetGroupInfoList, &codedeploy.TargetGroupInfo{
-				Name: aws.String(targetGroup["name"].(string)),
+				Name: aws.String(name),
 			})
 		}
 	}
@@ -952,6 +962,9 @@ func flattenLoadBalancerInfo(loadBalancerInfo *codedeploy.LoadBalancerInfo) []ma
 
 	elbs := make([]interface{}, 0, len(loadBalancerInfo.ElbInfoList))
 	for _, elb := range loadBalancerInfo.ElbInfoList {
+		if elb.Name == nil {
+			continue
+		}
 		item := make(map[string]interface{})
 		item["name"] = *elb.Name
 		elbs = append(elbs, item)
@@ -959,6 +972,9 @@ func flattenLoadBalancerInfo(loadBalancerInfo *codedeploy.LoadBalancerInfo) []ma
 
 	targetGroups := make([]interface{}, 0, len(loadBalancerInfo.TargetGroupInfoList))
 	for _, targetGroup := range loadBalancerInfo.TargetGroupInfoList {
+		if targetGroup.Name == nil {
+			continue
+		}
 		item := make(map[string]interface{})
 		item["name"] = *targetGroup.Name
 		targetGroups = append(targetGroups, item)
--- a/aws/resource_aws_codedeploy_deployment_group.go
+++ b/aws/resource_aws_codedeploy_deployment_group.go
@@ -52,6 +52,91 @@ func resourceAwsCodeDeployDeploymentGroup() *schema.Resource {
 				},
 			},
 
+			"deployment_style": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"deployment_option": &schema.Schema{
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validateDeploymentOption,
+						},
+
+						"deployment_type": &schema.Schema{
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validateDeploymentType,
+						},
+					},
+				},
+			},
+
+			"blue_green_deployment_config": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"deployment_ready_option": &schema.Schema{
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"action_on_timeout": &schema.Schema{
+										Type:         schema.TypeString,
+										Optional:     true,
+										ValidateFunc: validateDeploymentReadyOption,
+									},
+									"wait_time_in_minutes": &schema.Schema{
+										Type:     schema.TypeInt,
+										Optional: true,
+									},
+								},
+							},
+						},
+
+						"green_fleet_provisioning_option": &schema.Schema{
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"action": &schema.Schema{
+										Type:         schema.TypeString,
+										Optional:     true,
+										ValidateFunc: validateGreenFleetProvisioningOption,
+									},
+								},
+							},
+						},
+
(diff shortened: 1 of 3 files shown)


website/docs/r/codedeploy_deployment_group.html.markdown
```diff
@@ -108,35 +152,108 @@ The following arguments are supported:
...
+_Only one `alarm_configuration` block is allowed_.
+
+### Deployment Style
+You can configure the type of deployment, either standard or blue/green, you want to run and whether to route deployment traffic behind a load balancer. `deployment_style` supports the following:
+
+* `deployment_option` - (Optional) Indicates whether to route deployment traffic behind a load balancer. Valid Values are `WITH_TRAFFIC_CONTROL` or `WITHOUT_TRAFFIC_CONTROL`.
+
+* `deployment_type` - (Optional) Indicates whether to run a standard deployment or a blue/green deployment. Valid Values are `IN_PLACE` or `BLUE_GREEN`.
+
+  * `IN_PLACE` deployment type is not supported with the `WITH_TRAFFIC_CONTROL` deployment option.
```
LegNeato: This is no longer true, right?

niclic: You are correct. I will fix this.

niclic:
Documentation updated to reflect support of traffic control with in-place deployments.

zdobrenen:
Will this PR be merged in the next version release? If so, when can that be expected to happen? This functionality will be much appreciated.

aredridel:
Ooh, this just bit us, too!

LegNeato:
Not sure if this changes the PR:

https://aws.amazon.com/about-aws/whats-new/2017/08/aws-codedeploy-now-supports-using-application-load-balancers-for-blue-green-and-in-place-deployments/

niclic:
Thanks for the heads up. @LegNeato 

There is now an optional `targetGroupInfoList` array on `loadBalancerInfo` to support using  Application Load Balancer as part of a deployment.

I can add support for this feature, although I'd prefer if the PR was merged _as is_. It has been six months after all. ;)

PauloMigAlmeida:
It just hit us too.

Is there any plan for when this will be merged to master?

niclic:
Added support for using an [Application Load Balancer](https://aws.amazon.com/about-aws/whats-new/2017/08/aws-codedeploy-now-supports-using-application-load-balancers-for-blue-green-and-in-place-deployments/) in a deployment.

A note on `load_balancer_info` exception tests.

* two `elb_info`?

    * `aws_codedeploy_deployment_group.foo_group: InvalidLoadBalancerInfoException: The specification for load balancing in the deployment group is invalid. The elbInfoList string contains more than one load balancer name, but only one load balancer is supported per deployment group.`

* two `target_group_info`?

    * `aws_codedeploy_deployment_group.foo_group: InvalidLoadBalancerInfoException: The specification for load balancing in the deployment group is invalid. The targetGroupInfoList string contains more than one target group name, but only one target group is supported per deployment group.`

* one each?

    * `aws_codedeploy_deployment_group.foo_group: InvalidLoadBalancerInfoException: The specification for load balancing in the deployment group is invalid. Both a load balancer and a target group have been specified in loadBalancerInfo, but only one can be used in a single deployment.`

These exceptions are mentioned in the documentation for this resource.

PauloMigAlmeida:
@niclic Good work! It is undoubtedly a long-awaited feature for both CodeDeploy and Terraform. 🥇 
ping @radeksimko: is there anything else that needs to be done in order to get this merged?

aws/resource_aws_codedeploy_deployment_group.go
```diff
@@ -538,6 +712,98 @@ func buildAlarmConfig(configured []interface{}) *codedeploy.AlarmConfiguration {
...
+
+	blueGreenDeploymentConfig := &codedeploy.BlueGreenDeploymentConfiguration{}
+
+	config := list[0].(map[string]interface{})
+
+	if attr, ok := config["deployment_ready_option"]; ok {
+		m := attr.([]interface{})[0].(map[string]interface{})
+		deploymentReadyOption := &codedeploy.DeploymentReadyOption{
+			ActionOnTimeout:   aws.String(m["action_on_timeout"].(string)),
+			WaitTimeInMinutes: aws.Int64(int64(m["wait_time_in_minutes"].(int))),
```
radeksimko: As mentioned above - these are also optional fields, so we should only be passing them to the API if they're actually defined in the config.

aws/resource_aws_codedeploy_deployment_group.go
```diff
@@ -538,6 +712,98 @@ func buildAlarmConfig(configured []interface{}) *codedeploy.AlarmConfiguration {
...
+			Action: aws.String(m["action"].(string)),
+		}
+		blueGreenDeploymentConfig.GreenFleetProvisioningOption = greenFleetProvisioningOption
+	}
+
+	if attr, ok := config["terminate_blue_instances_on_deployment_success"]; ok {
+		m := attr.([]interface{})[0].(map[string]interface{})
+		blueInstanceTerminationOption := &codedeploy.BlueInstanceTerminationOption{
+			Action: aws.String(m["action"].(string)),
+			TerminationWaitTimeInMinutes: aws.Int64(int64(m["termination_wait_time_in_minutes"].(int))),
```
radeksimko: As mentioned above - these are also optional fields, so we should only be passing them to the API if they're actually defined in the config.

aws/resource_aws_codedeploy_deployment_group.go
```diff
@@ -630,6 +896,85 @@ func alarmConfigToMap(config *codedeploy.AlarmConfiguration) []map[string]interf
...
+func deploymentStyleToMap(style *codedeploy.DeploymentStyle) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, 1)
+
+	if style == nil {
+		return result
+	}
+
+	item := make(map[string]interface{})
+	item["deployment_option"] = *style.DeploymentOption
+	item["deployment_type"] = *style.DeploymentType
```
radeksimko: Are you sure that none of these options will ever be `nil`? Otherwise this would cause a crash here.

aws/resource_aws_codedeploy_deployment_group.go
```diff
@@ -630,6 +896,85 @@ func alarmConfigToMap(config *codedeploy.AlarmConfiguration) []map[string]interf
...
+	targetGroups := make([]interface{}, 0, len(loadBalancerInfo.TargetGroupInfoList))
+	for _, targetGroup := range loadBalancerInfo.TargetGroupInfoList {
+		item := make(map[string]interface{})
+		item["name"] = *targetGroup.Name
+		targetGroups = append(targetGroups, item)
+	}
+
+	lbInfo := make(map[string]interface{})
+	lbInfo["elb_info"] = schema.NewSet(loadBalancerInfoHash, elbs)
+	lbInfo["target_group_info"] = schema.NewSet(loadBalancerInfoHash, targetGroups)
```
radeksimko: It may look unintuitive I think we should be able to just use `[]interface{}` here, i.e. `elbs` and `targetGroups` directly. We can avoid the custom hashing function that way.

aws/resource_aws_codedeploy_deployment_group.go
```diff
@@ -630,6 +896,85 @@ func alarmConfigToMap(config *codedeploy.AlarmConfiguration) []map[string]interf
...
+	greenFleetProvisioningOption := make(map[string]interface{})
+	greenFleetProvisioningOption["action"] = *config.GreenFleetProvisioningOption.Action
+	m["green_fleet_provisioning_option"] = append(b, greenFleetProvisioningOption)
+
+	c := make([]map[string]interface{}, 0)
+	blueInstanceTerminationOption := make(map[string]interface{})
+	blueInstanceTerminationOption["action"] = *config.TerminateBlueInstancesOnDeploymentSuccess.Action
+	blueInstanceTerminationOption["termination_wait_time_in_minutes"] = *config.TerminateBlueInstancesOnDeploymentSuccess.TerminationWaitTimeInMinutes
+	m["terminate_blue_instances_on_deployment_success"] = append(c, blueInstanceTerminationOption)
+
```
radeksimko: Are you sure that none of these options will ever be `nil`? Otherwise this would cause a crash here.

aws/resource_aws_codedeploy_deployment_group.go
```diff
@@ -670,6 +1015,21 @@ func resourceAwsCodeDeployTriggerConfigHash(v interface{}) int {
 	return hashcode.String(buf.String())
 }
 
+func loadBalancerInfoHash(v interface{}) int {
```
radeksimko: This custom hash function shouldn't be necessary as the nested schema seems quite simple, so the default one (which is used when `Set` is not defined) should suffice.

niclic:
@radeksimko 

Thank you so much for the review! I really appreciate getting great feedback.

I wil review your comments in detail and address them with new changes as soon as possible.

**Naming**
I chose to be consistent with the naming convention of similar functions in this file, but I was aware that they were inconsistent with those used elsewhere. _I am happy to change them to be consistent with the desired convention._

Question: Do you think I should change the names of only the functions in this PR, or for all similar functions in this resource file?

**Optional v Required**
The optional or required nature of things has bothered me right from the very start of this PR. I [commented](https://github.com/hashicorp/terraform/pull/11700#issuecomment-280874601) on this in the original PR.

All types (and their sub types) are indicated as being _optional_ in the [SDK](http://docs.aws.amazon.com/codedeploy/latest/APIReference/API_BlueGreenDeploymentConfiguration.html), which obviously is not the case (your simple test proves that). However, this was the reason why I hesitated marking them as required in terraform - to avoid a mismatch between the official documentation and terraform, which might be confusing. But if they are going to be indicated as being optional to users of terraform, then more robust null checking is mandatory.

Question: For these cases, would it be better to mark them with `Required: true` using schema attributes, rather than adding null checks in the code? The former option seems simpler and requires less code, but the latter has the added advantage of remaining consistent with the SDK.

radeksimko:
> Question: Do you think I should change the names of only the functions in this PR, or for all similar functions in this resource file?

I'd only change the ones you have created as part of this PR and deal with the rest in a separate PR as this one is already getting quite big in terms of LOC. 😅 

Related to this - we also tend to store all flatteners & expanders in here: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/structure.go to keep the LOC in `resource_*` files with the actual CRUD logic as low as possible.

Again - no big deal, but it's a convention we developed over time and it's nice to stick to it, if we can.

> Question: For these cases, would it be better to mark them with Required: true using schema attributes, rather than adding null checks in the code? The former option seems simpler and requires less code, but the latter has the added advantage of remaining consistent with the SDK.

I'm honestly not sure, but I remember I marked some fields Required in the past in a different API, despite them being documented as optional to later discover they _actually_ were optional, just under certain context - e.g. in combination with some other fields and/or values of those fields. I don't know if that's the case here, but I'd double check this with AWS support before drifting away from documentation.

Keeping it optional and adding `nil` checks seems safer at this point, but if you can get a confirmation of docs being wrong by AWS, then making fields required is obviously much cleaner solution.

niclic:
### TODO

* [ ] Don't pass "optional" fields to the api:
   * [ ] `buildDeploymentStyle`
   * [ ] `buildBlueGreenDeploymentConfig`

* [ ] `nil` checks on "optional" fields:
   * [ ] `deploymentStyleToMap`
   * [ ] `blueGreenDeploymentConfigToMap`

 * [ ] Rename _flatteners_ and _expanders_ to be consistent with established convention.

* [ ] Move _flatteners_ and _expanders_ to `structure.go`?

* [ ] Remove `loadBalancerInfoHash` function.

* [ ] Does `deployment_style` need to be `Computed`?

niclic:
@radeksimko 

RE: `loadBalancerInfo`
Can you clarify which option you are suggesting, since two options are presented:
* use `schema.HashString` in place of `loadBalancerInfoHash`, or
* model `elb_info` and `target_group_info` as `TypeList` (with `MaxItems = 1`), instead of as `schema.TypeSet`

On reflection, I'm not sure _why_ `elb_info` is `TypeSet` and not a `TypeList`, so I think I will change that. I guess I got tired of dealing with lists that could only have one item. Is there a better way of modelling such simple structs?

Also, I did not add any `nil` checks for the `name` field on `elb` or `targetGroup`, since you didn't explicitly call those out. This field is supposed to be required, even though it appears as optional in the docs. When making the above changes, I can add these additional checks to `expandLoadBalancerInfo` and `flattenLoadBalancerInfo`.

radeksimko:
@niclic What I meant originally is to keep both fields as `TypeSet` (because ordering doesn't/shouldn't matter) and set value as `[]interface{}` so we can remove the custom hash function and let `helper/schema` use the default one.

```go
lbInfo["elb_info"] = elbs
lbInfo["target_group_info"] = targetGroups
```
See how we do the slice -> set conversion here: https://github.com/hashicorp/terraform/blob/master/helper/schema/field_writer_map.go#L271-L309 
and how we use the default hash function here:
https://github.com/hashicorp/terraform/blob/004f6cc9e29ccecb25676f162584db113813e91c/helper/schema/schema.go#L275-L281 which is the best we can do in this particular case anyway. 

`schema.HashString` would be useful in case we had something like this in the schema:

```go
"elb_info": &schema.Schema{
	Type:     schema.TypeSet,
	Optional: true,
	Set:      schema.HashString,
},
```

The nested structure in our case here isn't a simple string though, it's a `Resource` with `name` field.

Let me know if that makes sense and/or whether you need any further help.

niclic:
@radeksimko 


Okay. When I remove the custom hash function, I get this error in any test that includes `load_balancer_info`:

```
* Invalid address to set: []string{"load_balancer_info", "0", "elb_info"}
```

I may be missing something, but if I change `elb_Info` and `target_group_info` to `TypeList` all tests pass. Kinda like in this prior [issue](https://github.com/hashicorp/terraform/pull/7105/files).

Let me know if you want me to push these commits, so you see for yourself. Or if I should just proceed with changing `TypeSet` to `TypeList`. Note that currently, only one `elb_info` or `target_group_info` is allowed by the SDK.

niclic:
Also, I'm considering removing the flatten/expand function tests. Here are the tests I added for this PR (there are many others form earlier contributions, never mind all the brittle validation functions as well, but that's for another time).

```
TestAWSCodeDeployDeploymentGroup_expandDeploymentStyle
TestAWSCodeDeployDeploymentGroup_flattenDeploymentStyle
TestAWSCodeDeployDeploymentGroup_expandLoadBalancerInfo
TestAWSCodeDeployDeploymentGroup_flattenLoadBalancerInfo
TestAWSCodeDeployDeploymentGroup_expandBlueGreenDeploymentConfig
TestAWSCodeDeployDeploymentGroup_flattenBlueGreenDeploymentConfig
```
I'm not sure if these sorts of tests add much value. I wrote them to guide the writing of flatten and expand functions. There are plenty of acceptance tests covering these resources now and besides, despite their lines of code, the tests are very basic, and do not cover every possible code path.

Thoughts?

niclic:
@radeksimko 

RE: `Computed` on `DeploymentStyle`
I can't get the `TestAccAWSCodeDeployDeploymentGroup_deploymentStyle_delete` test to pass without using `Computed: true,`. If I send `nil`, I get the error below:

```
--- FAIL: TestAccAWSCodeDeployDeploymentGroup_deploymentStyle_delete (34.05s)
  testing.go:427: Step 1 error: After applying this step, the plan was not empty:

    DIFF:

    UPDATE: aws_codedeploy_deployment_group.foo_group
      deployment_style.#:                   "1" => "0"
      deployment_style.0.deployment_option: "WITH_TRAFFIC_CONTROL" => ""
      deployment_style.0.deployment_type:   "BLUE_GREEN" => ""

    STATE:

    aws_codedeploy_app.foo_app:
      ID = 44aaa267-7e15-4642-a5a4-3552f34ff7a5:foo-app-18afv
      name = foo-app-18afv
    aws_codedeploy_deployment_group.foo_group:
      ID = a75207ec-a2b7-4acd-98d6-9a292ab2c5b9
      alarm_configuration.# = 0
      app_name = foo-app-18afv
      auto_rollback_configuration.# = 0
      autoscaling_groups.# = 0
      blue_green_deployment_config.# = 0
      deployment_config_name = CodeDeployDefault.OneAtATime
      deployment_group_name = foo-group-18afv
      deployment_style.# = 1
      deployment_style.0.deployment_option = WITH_TRAFFIC_CONTROL
      deployment_style.0.deployment_type = BLUE_GREEN
```

If I send an empty `&codedeploy.DeploymentStyle{}`, I get an exception:

```
* aws_codedeploy_deployment_group.foo_group:
InvalidDeploymentStyleException: Deployment type or deployment option cannot be null or empty
```

I'm probably missing something here. But it seems that removing `deployment_style` from my configuration does not trigger a reversion to the default configuration, but instead retains the previous state.

I have sent a question to AWS SDK support regarding the using of `Required: No` in the SDK documentation. Have yet to hear back from them. I should probably try a more direct line of communication.

radeksimko:
> When I remove the custom hash function, I get this error in any test that includes `load_balancer_info`:

You're right and I apologise for providing misleading info and letting you spend the extra time on this! 😞 
As I discovered myself + after chatting with colleague from the core team we only support this behaviour for Sets in the 1st level, not nested sets, like here. I frankly wasn't aware of this limitation.

> if I change elb_Info and target_group_info to TypeList all tests pass.

One of the reasons we should use TypeSet instead of TypeList here is because the names of ELBs or Target Groups may come from the API in an unpredictable order (AFAIK) and the user would see spurious diff because ordering in the config doesn't match the one in the API response.

To sum this up - do you mind removing the last two commits? Thanks and sorry again!

-------

> Also, I'm considering removing the flatten/expand function tests.

I think they're useful, I'd keep them there.

> There are plenty of acceptance tests covering these resources now and besides, despite their lines of code, the tests are very basic, and do not cover every possible code path.

I think the intention of a test isn't to cover every possible codepath, that's almost impossible in reality. It's however useful to have tests which test different parts/functions of the program, so that if that small part breaks we can have a small, isolated test making it obvious which part is broken.

If you had an acceptance test failure you'd need to dig deeper to understand that it's caused by flattener/expander.

-------

> I can't get the TestAccAWSCodeDeployDeploymentGroup_deploymentStyle_delete test to pass without using Computed: true,

I see, technically Terraform does what it's told - `Computed` means "anything is ok", including the previous value user had in config prior to update (because that's what comes from the API). Whether that's what the user expects is questionable. Either way I have an open AWS support ticket about inability to remove/reset blue-green configuration.

Here's what I received on 17th September:

> I am engaging with our internal team to optimize the output of deployment group for you. Right now there seems no way to remove B/G configuration from a deployment group once added. I will update this ticket when I have the feedback from the internal team.

I don't expect quick response or solution there, so I'd keep it `Computed` for the time being.

-------

> I have sent a question to AWS SDK support regarding the using of Required: No in the SDK documentation.

How do you feel about adding the extra nil checks and keeping it `Optional`, so that we can merge this PR and eventually follow up later, once AWS comes back with an answer?

niclic:
> To sum this up - do you mind removing the last two commits? Thanks and sorry again!

No problem!

> How do you feel about adding the extra nil checks and keeping it Optional, so that we can merge this PR and eventually follow up later, once AWS comes back with an answer?

Sounds good to me.

I will remove the last couple of commits and revisit the overall PR before signing off.

Thanks again for all your help. @radeksimko

radeksimko:
@niclic Hey,
I can take this across the finish line if you want. I know you already spent quite some time on this and part of that was because of my mistake, so I feel I should pay it back 😅 

I'd like to cut next release of the provider this week and have this PR in it, so let me know if you're ok with me making the proposed changes and pushing them to your branch.

niclic:
@radeksimko 

I had planned on getting this completed this past weekend, but best laid plans, etc.

I am totally fine with you completing this and finally seeing it merged. Thanks for all your help!

JimtotheB:
@radeksimko or @niclic Since you guys seem to be behind this functionality, what is the preferred way to keep track of the ASG created with

```
green_fleet_provisioning_option {
      action = "COPY_AUTO_SCALING_GROUP"
}
```

Currently I am bringing in a ASG from another module, which is destroyed when I deploy the app. I have tried to `terraform import` the new ASG in, but of course running `terraform apply` attempts to recreate the original named ASG from the other module. Is there a correct way that either of you guys had in mind on how to keep track of this?

SvanBoxel:
Hey @JimtotheB! Did you find out how to fix this? Running into the same problem.

niclic:
Please see [this comment](https://github.com/terraform-providers/terraform-provider-aws/issues/4678#issuecomment-419982879), which explains this issue in detail, plus some options for working around it.

PR #1168, opened 2017-07-17
Added API Gateway Gateway response resource

Ninir:
## Description 
This adds the API Gateway Gateway response, which allows to configure.
This is a required step to complete the basic-auth protected HTTPs endpoint test to end https://github.com/terraform-providers/terraform-provider-aws/pull/861.

Using this resource, we are now able to specify headers & response templates on a https status code basis, so per 401, 403, 500 errors.

Documentation: https://docs.aws.amazon.com/apigateway/api-reference/link-relation/gatewayresponse-update/

## Tests
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayGatewayResponse_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayGatewayResponse_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayGatewayResponse_basic
--- PASS: TestAccAWSAPIGatewayGatewayResponse_basic (39.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	39.649s
```

## TODOs
- [x] Add the resource + acceptance tests
- [x] Handle timeouts in the resource CRUD
- [x] Add the documentation

### DIFF
--- a/aws/resource_aws_api_gateway_gateway_response.go
+++ b/aws/resource_aws_api_gateway_gateway_response.go
@@ -1,0 +1,145 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/apigateway"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsApiGatewayGatewayResponse() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsApiGatewayGatewayResponsePut,
+		Read:   resourceAwsApiGatewayGatewayResponseRead,
+		Update: resourceAwsApiGatewayGatewayResponsePut,
+		Delete: resourceAwsApiGatewayGatewayResponseDelete,
+
+		Schema: map[string]*schema.Schema{
+			"rest_api_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"response_type": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"status_code": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"response_templates": {
+				Type:     schema.TypeMap,
+				Elem:     schema.TypeString,
+				Optional: true,
+			},
+
+			"response_parameters": {
+				Type:     schema.TypeMap,
+				Elem:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceAwsApiGatewayGatewayResponsePut(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	templates := make(map[string]string)
+	if kv, ok := d.GetOk("response_templates"); ok {
+		for k, v := range kv.(map[string]interface{}) {
+			templates[k] = v.(string)
+		}
+	}
+
+	parameters := make(map[string]string)
+	if kv, ok := d.GetOk("response_parameters"); ok {
+		for k, v := range kv.(map[string]interface{}) {
+			parameters[k] = v.(string)
+		}
+	}
+
+	input := apigateway.PutGatewayResponseInput{
+		RestApiId:          aws.String(d.Get("rest_api_id").(string)),
+		ResponseType:       aws.String(d.Get("response_type").(string)),
+		ResponseTemplates:  aws.StringMap(templates),
+		ResponseParameters: aws.StringMap(parameters),
+	}
+
+	if v, ok := d.GetOk("status_code"); ok {
+		input.StatusCode = aws.String(v.(string))
+	}
+
+	log.Printf("[DEBUG] Putting API Gateway Gateway Response: %s", input)
+
+	_, err := conn.PutGatewayResponse(&input)
+	if err != nil {
+		return fmt.Errorf("Error putting API Gateway Gateway Response: %s", err)
+	}
+
+	d.SetId(fmt.Sprintf("aggr-%s-%s", d.Get("rest_api_id").(string), d.Get("response_type").(string)))
+	log.Printf("[DEBUG] API Gateway Gateway Response put (%q)", d.Id())
+
+	return resourceAwsApiGatewayGatewayResponseRead(d, meta)
+}
+
+func resourceAwsApiGatewayGatewayResponseRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	log.Printf("[DEBUG] Reading API Gateway Gateway Response %s", d.Id())
+	gatewayResponse, err := conn.GetGatewayResponse(&apigateway.GetGatewayResponseInput{
+		RestApiId:    aws.String(d.Get("rest_api_id").(string)),
+		ResponseType: aws.String(d.Get("response_type").(string)),
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFoundException" {
+			log.Printf("[WARN] API Gateway Gateway Response (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	log.Printf("[DEBUG] Received API Gateway Gateway Response: %s", gatewayResponse)
+
+	d.Set("response_type", gatewayResponse.ResponseType)
+	d.Set("status_code", gatewayResponse.StatusCode)
+	d.Set("response_templates", aws.StringValueMap(gatewayResponse.ResponseTemplates))
+	d.Set("response_parameters", aws.StringValueMap(gatewayResponse.ResponseParameters))
+
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,152 @@
...
+func resourceAwsApiGatewayGatewayResponse() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsApiGatewayGatewayResponseCreate,
+		Read:   resourceAwsApiGatewayGatewayResponseRead,
+		Update: resourceAwsApiGatewayGatewayResponseCreate,
+		Delete: resourceAwsApiGatewayGatewayResponseDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
```
radeksimko: Is there any particular reason to have the customizable timeout for this resource?
This feature was IMO mainly aimed at resources which are going to take varying time to create/update/delete depending on fields, e.g. instance size, cluster size, ...

I described the full reasoning here https://github.com/terraform-providers/terraform-provider-aws/pull/945#issuecomment-312916643 (beginning from the 2nd paragraph).

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,152 @@
...
+				Elem:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceAwsApiGatewayGatewayResponseCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+	log.Printf("[DEBUG] Putting API Gateway Gateway Response")
```
radeksimko: Nitpick: I think we can make this log message a lot more useful if we move it down and log the `input` before sending it to the API.

```go
log.Printf("[DEBUG] Putting API Gateway Gateway Response: %s", input)
```

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,152 @@
...
+			return nil
+		}
+
+		apigatewayErr, ok := err.(awserr.Error)
+
+		if !ok {
+			return resource.NonRetryableError(err)
+		}
+
+		if apigatewayErr.Code() == "TooManyRequestsException" {
```
radeksimko: This code is handled (and retried on) already in the AWS SDK: https://github.com/aws/aws-sdk-go/blob/master/aws/request/retryer.go#L41

If you think we don't retry enough times we should pass that feedback back to AWS SDK and deal with it there.

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,152 @@
...
+
+		if err == nil {
+			return nil
+		}
+
+		apigatewayErr, ok := err.(awserr.Error)
+
+		if !ok {
+			return resource.NonRetryableError(err)
+		}
```
radeksimko: You can omit these three lines if you add `ok &&` to the condition handling `NotFoundException` - it will just fall through to the last `return`. 😉

aws/resource_aws_api_gateway_gateway_response_test.go
```diff
@@ -0,0 +1,148 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAPIGatewayGatewayResponseExists("aws_api_gateway_gateway_response.test", &conf),
+					resource.TestCheckResourceAttr("aws_api_gateway_gateway_response.test", "status_code", "401"),
+					resource.TestCheckResourceAttr("aws_api_gateway_gateway_response.test", "response_parameters.gatewayresponse.header.Authorization", "'Basic'"),
+					resource.TestCheckResourceAttr("aws_api_gateway_gateway_response.test", "response_templates.application/xml", "#set($inputRoot = $input.path('$'))\n{ }"),
+					resource.TestCheckNoResourceAttr("aws_api_gateway_gateway_response.test", "response_templates.application/json"),
+				),
+			},
+
+			resource.TestStep{
```
radeksimko: As in any slice/array of structs in Go this "annotation" is optional and we tend to omit it (less characters to read 😃 ) - like you did in the schema map.

https://play.golang.org/p/JcV3cTcxvn

aws/resource_aws_api_gateway_gateway_response_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+func testAccCheckAWSAPIGatewayGatewayResponseExists(n string, res *apigateway.UpdateGatewayResponseOutput) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No API Gateway Method ID is set")
```
radeksimko: 👀

aws/resource_aws_api_gateway_gateway_response_test.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		req := &apigateway.GetGatewayResponseInput{
+			RestApiId:    aws.String(s.RootModule().Resources["aws_api_gateway_rest_api.main"].Primary.ID),
+			ResponseType: aws.String(rs.Primary.Attributes["response_type"]),
+		}
+		_, err := conn.GetGatewayResponse(req)
+
+		if err == nil {
+			return fmt.Errorf("API Gateway Method still exists")
```
radeksimko: 👀

aws/resource_aws_api_gateway_gateway_response.go
Ninir: Didn't even know that :O Big win on that, thanks!

aws/resource_aws_api_gateway_gateway_response.go
Ninir: Just wanted to make it all flexible. Your point seems legit, will remove the related stuff!

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,152 @@
...
+	return resourceAwsApiGatewayGatewayResponseRead(d, meta)
+}
+
+func resourceAwsApiGatewayGatewayResponseRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	log.Printf("[DEBUG] Reading API Gateway Gateway Response %s", d.Id())
+	gatewayResponse, err := conn.GetGatewayResponse(&apigateway.GetGatewayResponseInput{
+		RestApiId:    aws.String(d.Get("rest_api_id").(string)),
+		ResponseType: aws.String(d.Get("response_type").(string)),
```
Ninir: Yup, don't see any counter argument :)

aws/resource_aws_api_gateway_gateway_response.go
Ninir: Good catch!

Ninir:
@radeksimko Seems ok on my side!

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,143 @@
...
+			},
+		},
+	}
+}
+
+func resourceAwsApiGatewayGatewayResponsePut(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	templates := make(map[string]string)
+	for k, v := range d.Get("response_templates").(map[string]interface{}) {
```
radeksimko: Shouldn't this be also wrapped in `GetOk()` like the other optional field, as it's optional? Otherwise I think we might crash here when casting `nil` to a map.

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,152 @@
...
+		for k, v := range kv.(map[string]interface{}) {
+			parameters[k] = v.(string)
+		}
+	}
+
+	input := apigateway.PutGatewayResponseInput{
+		RestApiId:          aws.String(d.Get("rest_api_id").(string)),
+		ResponseType:       aws.String(d.Get("response_type").(string)),
+		ResponseTemplates:  aws.StringMap(templates),
+		ResponseParameters: aws.StringMap(parameters),
```
radeksimko: As discussed via Slack, this doesn't matter and we need it for updates, so 👍

aws/resource_aws_api_gateway_gateway_response.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil
+		}
+		return err
+	}
+
+	log.Printf("[DEBUG] Received API Gateway Gateway Response: %s", gatewayResponse)
+
+	d.Set("response_type", gatewayResponse.ResponseType)
+	d.Set("status_code", gatewayResponse.StatusCode)
+	d.Set("response_templates", gatewayResponse.ResponseTemplates)
```
radeksimko: This comes back as `map[string]*string`, so shouldn't we also convert it to flat map via `aws.StringValueMap` like the other field below?

aws/resource_aws_api_gateway_gateway_response.go
Ninir: Guhuhu, indeed it should have been set...!

aws/resource_aws_api_gateway_gateway_response.go
Ninir: Good catch >_<"

PR #1172, opened 2017-07-18
Add support for aws_cloudwatch_dashboard resource

ttacon:
Hi there! I noticed that there wasn't any support for CloudWatch dashboards yet, so I though I'd throw together a little something to at least see if I'm approaching this from the correct standpoint! If this looks mostly good, I'll go ahead with writing more acceptance tests and writing up the documentation for this section.

Thanks!

### DIFF
--- a/aws/resource_aws_cloudwatch_dashboard.go
+++ b/aws/resource_aws_cloudwatch_dashboard.go
@@ -1,0 +1,139 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCloudWatchDashboard() *schema.Resource {
+	return &schema.Resource{
+		Create:        resourceAwsCloudWatchDashboardCreate,
+		Read:          resourceAwsCloudWatchDashboardRead,
+		Update:        resourceAwsCloudWatchDashboardUpdate,
+		Delete:        resourceAwsCloudWatchDashboardDelete,
+		SchemaVersion: 1,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"dashboard_arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"dashboard_body": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateJsonString,
+				StateFunc: func(v interface{}) string {
+					json, _ := normalizeJsonString(v)
+					return json
+				},
+			},
+			"dashboard_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceAwsCloudWatchDashboardCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchconn
+
+	params := getAwsCloudWatchPutDashboardInput(d)
+
+	log.Printf("[DEBUG] Creating CloudWatch Dashboard: %#v", params)
+	_, err := conn.PutDashboard(&params)
+	if err != nil {
+		return fmt.Errorf("Creating dashboard failed: %s", err)
+	}
+	d.SetId(d.Get("dashboard_name").(string))
+	log.Println("[INFO] CloudWatch Dashboard created")
+
+	return resourceAwsCloudWatchDashboardRead(d, meta)
+}
+
+func resourceAwsCloudWatchDashboardRead(d *schema.ResourceData, meta interface{}) error {
+	a, err := getAwsCloudWatchDashboard(d, meta)
+	if err != nil {
+		return err
+	}
+	if a == nil {
+		d.SetId("")
+		return nil
+	}
+
+	log.Printf("[DEBUG] Reading CloudWatch Dashboard: %s", d.Get("dashboard_name"))
+
+	d.Set("dashboard_name", a.DashboardName)
+	d.Set("dashboard_body", a.DashboardBody)
+	return nil
+}
+
+func resourceAwsCloudWatchDashboardUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchconn
+	params := getAwsCloudWatchPutDashboardInput(d)
+
+	log.Printf("[DEBUG] Updating CloudWatch Dashboard: %#v", params)
+	_, err := conn.PutDashboard(&params)
+	if err != nil {
+		return fmt.Errorf("Updating dashboard failed: %s", err)
+	}
+	log.Println("[INFO] CloudWatch Dashboard updated")
+
+	return resourceAwsCloudWatchDashboardRead(d, meta)
+}
+
+func resourceAwsCloudWatchDashboardDelete(d *schema.ResourceData, meta interface{}) error {
+	p, err := getAwsCloudWatchDashboard(d, meta)
+	if err != nil {
+		return err
+	}
+	if p == nil {
+		log.Printf("[DEBUG] CloudWatch Dashboard %s is already gone", d.Id())
+		return nil
+	}
+
+	log.Printf("[INFO] Deleting CloudWatch Dashboard: %s", d.Id())
+
+	conn := meta.(*AWSClient).cloudwatchconn
+	params := cloudwatch.DeleteDashboardsInput{
+		DashboardNames: []*string{aws.String(d.Id())},
+	}
+
+	if _, err := conn.DeleteDashboards(&params); err != nil {
+		return fmt.Errorf("Error deleting CloudWatch Dashboard: %s", err)
+	}
+	log.Println("[INFO] CloudWatch Dashboard deleted")
+
+	d.SetId("")
+	return nil
+}
+
+func getAwsCloudWatchPutDashboardInput(d *schema.ResourceData) cloudwatch.PutDashboardInput {
+	return cloudwatch.PutDashboardInput{
+		DashboardBody: aws.String(d.Get("dashboard_body").(string)),
+		DashboardName: aws.String(d.Get("dashboard_name").(string)),
+	}
+}
+
+func getAwsCloudWatchDashboard(d *schema.ResourceData, meta interface{}) (*cloudwatch.GetDashboardOutput, error) {
+	conn := meta.(*AWSClient).cloudwatchconn
+
+	params := cloudwatch.GetDashboardInput{
+		DashboardName: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetDashboard(&params)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
--- a/aws/resource_aws_cloudwatch_dashboard.go
+++ b/aws/resource_aws_cloudwatch_dashboard.go
@@ -5,22 +5,26 @@ import (
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
 func resourceAwsCloudWatchDashboard() *schema.Resource {
 	return &schema.Resource{
-		Create:        resourceAwsCloudWatchDashboardCreate,
-		Read:          resourceAwsCloudWatchDashboardRead,
-		Update:        resourceAwsCloudWatchDashboardUpdate,
-		Delete:        resourceAwsCloudWatchDashboardDelete,
-		SchemaVersion: 1,
+		Create: resourceAwsCloudWatchDashboardCreate,
+		Read:   resourceAwsCloudWatchDashboardRead,
+		Update: resourceAwsCloudWatchDashboardUpdate,
+		Delete: resourceAwsCloudWatchDashboardDelete,
 
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
 
+		// Note that we specify both the `dashboard_body` and
+		// the `dashboard_name` as being required, even though
+		// according to the REST API documentation both are
+		// optional: http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutDashboard.html#API_PutDashboard_RequestParameters
 		Schema: map[string]*schema.Schema{
 			"dashboard_arn": {
 				Type:     schema.TypeString,
@@ -28,7 +32,7 @@ func resourceAwsCloudWatchDashboard() *schema.Resource {
 			},
 			"dashboard_body": {
 				Type:         schema.TypeString,
-				Optional:     true,
+				Required:     true,
 				ValidateFunc: validateJsonString,
 				StateFunc: func(v interface{}) string {
 					json, _ := normalizeJsonString(v)
@@ -37,7 +41,7 @@ func resourceAwsCloudWatchDashboard() *schema.Resource {
 			},
 			"dashboard_name": {
 				Type:     schema.TypeString,
-				Optional: true,
+				Required: true,
 			},
 		},
 	}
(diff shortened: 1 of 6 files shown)


aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCloudWatchDashboard() *schema.Resource {
+	return &schema.Resource{
+		Create:        resourceAwsCloudWatchDashboardCreate,
+		Read:          resourceAwsCloudWatchDashboardRead,
+		Update:        resourceAwsCloudWatchDashboardUpdate,
+		Delete:        resourceAwsCloudWatchDashboardDelete,
+		SchemaVersion: 1,
```
Ninir: This is not needed at the moment

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
Ninir: Doing so, we should add an import acc test

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+	a, err := getAwsCloudWatchDashboard(d, meta)
+	if err != nil {
+		return err
+	}
+	if a == nil {
+		d.SetId("")
+		return nil
+	}
+
+	log.Printf("[DEBUG] Reading CloudWatch Dashboard: %s", d.Get("dashboard_name"))
```
Ninir: I would place this at the beginning of the function, so that the log is more logic:
- log to explain that we will read
- make the actual call
- etc

What do you think?

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+func getAwsCloudWatchDashboard(d *schema.ResourceData, meta interface{}) (*cloudwatch.GetDashboardOutput, error) {
+	conn := meta.(*AWSClient).cloudwatchconn
+
+	params := cloudwatch.GetDashboardInput{
+		DashboardName: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetDashboard(&params)
+	if err != nil {
+		return nil, err
```
Ninir: As per the other functions, I would also add a log prefix:
```go
return nil, fmt.Errorf("Reading dashboard failed: %s", err)
```

Also, we should check for the awserr `ResourceNotFound` here, and warn if the dashboard is not found using this kind of log:

```go
log.Printf("[WARN] CloudWatch Dashboard (%s) not found, removing from state", d.Id())
```

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+	d.Set("dashboard_name", a.DashboardName)
+	d.Set("dashboard_body", a.DashboardBody)
+	return nil
+}
+
+func resourceAwsCloudWatchDashboardUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchconn
+	params := getAwsCloudWatchPutDashboardInput(d)
+
+	log.Printf("[DEBUG] Updating CloudWatch Dashboard: %#v", params)
```
Ninir: Same here, could you swap line 81 & 83, just for structuration?

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return fmt.Errorf("Updating dashboard failed: %s", err)
+	}
+	log.Println("[INFO] CloudWatch Dashboard updated")
+
+	return resourceAwsCloudWatchDashboardRead(d, meta)
+}
+
+func resourceAwsCloudWatchDashboardDelete(d *schema.ResourceData, meta interface{}) error {
+	p, err := getAwsCloudWatchDashboard(d, meta)
```
Ninir: It would be better to avoid this call, and check the DeleteDashboards function return codes for the `ResourceNotFound` awserr.
Thus, we could move the code from `getAwsCloudWatchDashboard` to the read function only

aws/resource_aws_cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,100 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSCloudWatchDashboardDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSCloudWatchDashboardConfig(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckCloudWatchDashboardExists("aws_cloudwatch_dashboard.foobar", &dashboard),
+				),
+			},
```
Ninir: It would be nice to have a second step, so that we can check the update of the resource

aws/resource_aws_cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,100 @@
...
+		conn := testAccProvider.Meta().(*AWSClient).cloudwatchconn
+		params := cloudwatch.GetDashboardInput{
+			DashboardName: aws.String(rs.Primary.ID),
+		}
+
+		resp, err := conn.GetDashboard(&params)
+		if err != nil {
+			return err
+		}
+		if len(*resp.DashboardName) == 0 {
```
Ninir: It would be better to check for the return code `ResourceNotFound`, and same line 71

aws/resource_aws_cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,100 @@
...
+  dashboard_body = <<EOF
+  {
+    "widgets": [{
+      "type": "text",
+      "x": 0,
+      "y": 0,
+      "width": 6,
+      "height": 6,
+      "properties": {
+        "markdown": "Hi there from CloudFormation"
```
Ninir: Hi there from Terraform / Cloudwatch, no? 😄

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+				Optional:     true,
+				ValidateFunc: validateJsonString,
+				StateFunc: func(v interface{}) string {
+					json, _ := normalizeJsonString(v)
+					return json
+				},
+			},
+			"dashboard_name": {
+				Type:     schema.TypeString,
+				Optional: true,
```
Ninir: This should be replaced by `Required: true` as name & body are required by the API

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,139 @@
...
+	return resourceAwsCloudWatchDashboardRead(d, meta)
+}
+
+func resourceAwsCloudWatchDashboardDelete(d *schema.ResourceData, meta interface{}) error {
+	p, err := getAwsCloudWatchDashboard(d, meta)
+	if err != nil {
+		return err
+	}
+	if p == nil {
+		log.Printf("[DEBUG] CloudWatch Dashboard %s is already gone", d.Id())
```
Ninir: I would WARN and rephrase for something like:

```go
log.Printf("[WARN] CloudWatch Dashboard (%s) not found, removing from state", d.Id())
```

aws/resource_aws_cloudwatch_dashboard.go
ttacon: Interestingly enough, these are [not specified as being required](http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutDashboard.html#API_PutDashboard_RequestParameters) to create a CloudWatch Dashboard.

aws/resource_aws_cloudwatch_dashboard.go
ttacon: Sounds totally good to me! (Was just copy-pastying logic from the CloudWatch Metric Alarm resource.)

aws/resource_aws_cloudwatch_dashboard_test.go
ttacon: Totally, agree, I hadn't written the rest of the acc tests at the moment.

aws/resource_aws_cloudwatch_dashboard_test.go
ttacon: Haha, true, I copied this from the blog post re: announcing the CloudWatch Dashboard API methods.

aws/resource_aws_cloudwatch_dashboard.go
ttacon: Could you point me to a good example test for testing import? Thanks!

Ninir: Sure! 😄 
here is an example: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/import_aws_cloudwatch_event_rule_test.go

aws/resource_aws_cloudwatch_dashboard.go
Ninir: I agree :( Whether it's the console or the CLI, it appears the name is required.
Body is required on the CLI only, so I would go with both required there

ttacon:
@Ninir let me know what you think of the changes!

aws/resource_aws_cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,198 @@
...
+    "x": 0,
+    "y": 0,
+    "width": 6,
+    "height": 6,
+    "properties": {
+      "markdown": "Hi there from Terraform: CloudWatch - updated"
+    }
+  }]
+}`
+)
```
Ninir: Why not putting these widgets in `testAccAWSCloudWatchDashboardConfig` & `testAccAWSCloudWatchDashboardConfig_updateBody`?

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,136 @@
...
+		if isResourceNotFoundErr(err) {
+			log.Printf("[WARN] CloudWatch Dashboard %q not found, removing", dashboardName)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Reading dashboard failed: %s", err)
+	}
+
+	d.Set("dashboard_name", resp.DashboardName)
+	d.Set("dashboard_body", resp.DashboardBody)
```
Ninir: We are never setting the `dashboard_arn` attribute, is this wanted or just an oversight?

aws/resource_aws_cloudwatch_dashboard_test.go
```diff
@@ -0,0 +1,198 @@
...
+			continue
+		}
+
+		params := cloudwatch.GetDashboardInput{
+			DashboardName: aws.String(rs.Primary.ID),
+		}
+
+		if _, err := conn.GetDashboard(&params); err == nil {
+			return fmt.Errorf("Dashboard still exists: %s", rs.Primary.ID)
+		} else if !isResourceNotFoundErr(err) {
```
Ninir: This would be better as another if, since the first one returns! 👍

aws/resource_aws_cloudwatch_dashboard_test.go
ttacon: I left it out so that it could be defined once and used multiple times instead of duplicating it multiple times. This also makes updating the test much easier.

ttacon:
@Ninir did some minor cleanup and replied to one of the questions [here](https://github.com/terraform-providers/terraform-provider-aws/pull/1172#discussion_r128983381), thanks for the feedback on this!

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,123 @@
...
+				StateFunc: func(v interface{}) string {
+					json, _ := normalizeJsonString(v)
+					return json
+				},
+				DiffSuppressFunc: suppressEquivalentJsonDiffs,
+			},
+			"dashboard_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
Ninir: Forgot to say that `dashboard_body` & `dashboard_name` should also be set as `ForceNew`, as those are required!

Also, according to the [documentation](http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutDashboard.html) the name should be validated: ` The maximum length is 255, and valid characters are A-Z, a-z, 0-9, "-", and "_".`

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,123 @@
...
+
+	d.Set("dashboard_arn", resp.DashboardArn)
+	d.Set("dashboard_name", resp.DashboardName)
+	d.Set("dashboard_body", resp.DashboardBody)
+	return nil
+}
+
+func resourceAwsCloudWatchDashboardPut(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cloudwatchconn
+	params := getAwsCloudWatchPutDashboardInput(d)
```
Ninir: It would be better to put the params structure here directly instead of using the function, since we are using it once

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("Putting dashboard failed: %s", err)
+	}
+	d.SetId(d.Get("dashboard_name").(string))
+	log.Println("[INFO] CloudWatch Dashboard put finished")
+
+	return resourceAwsCloudWatchDashboardRead(d, meta)
+}
+
+func resourceAwsCloudWatchDashboardDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Println("[INFO] Deleting CloudWatch Dashboard")
```
Ninir: We should expose the id in the log, for better debug purposes!

aws/resource_aws_cloudwatch_dashboard.go
```diff
@@ -0,0 +1,123 @@
...
+	conn := meta.(*AWSClient).cloudwatchconn
+	params := cloudwatch.DeleteDashboardsInput{
+		DashboardNames: []*string{aws.String(d.Id())},
+	}
+
+	if _, err := conn.DeleteDashboards(&params); err != nil {
+		if isCloudWatchDashboardNotFoundErr(err) {
+			log.Printf("[WARN] CloudWatch Dashboard %s is already gone", d.Id())
```
Ninir: We should also add `d.SetId("")` right after this

aws/resource_aws_cloudwatch_dashboard_test.go
Ninir: Got it!

aws/resource_aws_cloudwatch_dashboard.go
ttacon: I'm not sure that `dashboard_body` and `dashboard_name` should be set to `ForceNew`. According to the terraform documentation, that implies changes to those resources should mean that a new resource should be created? However, they should be update-able (even though this code treats them as required, they're not required as per the AWS docs, meaning we shouldn't force the creation of new resources due to changes in these fields).

ttacon:
Thanks for the feedback @Ninir - added the validation logic and commented on the discussion re: `ForceNew`.

ttacon:
I should also mention that I have been running the acceptance tests:

```
$ make testacc TESTARGS="-run=TestAccAWSCloudWatchDashboard_*"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAWSCloudWatchDashboard_* -timeout 120m
=== RUN   TestAccAWSCloudWatchDashboard_importBasic
--- PASS: TestAccAWSCloudWatchDashboard_importBasic (17.25s)
=== RUN   TestAccAWSCloudWatchDashboard_basic
--- PASS: TestAccAWSCloudWatchDashboard_basic (15.56s)
=== RUN   TestAccAWSCloudWatchDashboard_update
--- PASS: TestAccAWSCloudWatchDashboard_update (29.64s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       62.478s
```

Ninir:
Hey @ttacon 

Just updated a bit the validation function.

Ran tests and it seems all ok:
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSCloudWatchDashboard_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSCloudWatchDashboard_ -timeout 120m
=== RUN   TestAccAWSCloudWatchDashboard_importBasic
--- PASS: TestAccAWSCloudWatchDashboard_importBasic (16.30s)
=== RUN   TestAccAWSCloudWatchDashboard_basic
--- PASS: TestAccAWSCloudWatchDashboard_basic (14.92s)
=== RUN   TestAccAWSCloudWatchDashboard_update
--- PASS: TestAccAWSCloudWatchDashboard_update (26.81s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	58.134s
```

I'm not just playing a bit with this specific version locally, and will merge afterwards :)

Ninir:
Hey @ttacon 

All good to me, works really nicely!

Thanks a lot for the whole work you gave here, and for fixing it quickly when required!
That's a really good first contribution you made 👍

Issue #227, opened 2017-06-13
aws_instance associate_public_ip_address


hashibot:
_This issue was originally opened by @mengesb as hashicorp/terraform#8244. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Related to hashicorp/terraform#5764 I'd like to re-raise the issue of `associate_public_ip_address` in the `aws_instance` resource.

If the subnet being used has a `map_public_ip = true` setting, it will always map a public IP. I rather this be interpreted as, "If not specified, use `map_public_ip` setting for the subnet". It seems this is not the case, and that the subnet settings can override the `aws_instance` settings.

Is this an API limitation? Even if I create a subnet with public IP mapping, there are circumstances where I want the VMs created to not have public IPs. I know that if I have a subnet that DOES NOT map public IPs, and I set `associate_public_ip_address = true` on the `aws_instance` resource, I get a public IP. When I manually create an instance via the AWS console, I get the option to disable the 'auto-assign public IP', accept the subnet default, or enable. Seems pretty clear-cut here that if I specify `false` for `associate_public_ip_address` I should NOT get a public IP no matter what the subnet settings are.
### Terraform Version

0.7.0
### Affected Resource(s)
- aws_instance
### Terraform Configuration Files

``` hcl
resource "aws_instance" "instance" {
  ami = ..
  ebs_optimized = ..
  instance_type = ..
  associate_public_ip_address = false
  subnet_id = ..
 ...
}
```
### Debug Output

N/A
### Panic Output

N/A
### Expected Behavior

if `associate_public_ip_address` is set to false, I get an instance without a public IP. If `associate_public_ip_address is set to true, I get an instance with a public IP. This doesn't matter whether or not the subnet is set to automatically map public IPs or not.
### Actual Behavior

if `associate_public_ip_address` is set to false, I get an instance with a public IP if the subnet is configured to `map_public_ip = true` (done in terraform or not done in terraform).
if `associate_public_ip_address` is set to false, I get an instance without a public IP if the subnet is configured to `map_public_ip = false` (done in terraform or not done in terraform).
### Steps to Reproduce

---
1. Create terraform plan with aws instance resource. set `associate_public_ip_address` to `false`
2. Use a subnet id that maps public IPs to instances (`map_public_ip = true`)
3. `terraform apply`
4. VM has public IP associated

---
[...]

PR #1185, opened 2017-07-19
AssociatePublicIPs: value when false should work.

alejandroEsc:
PR is an attempt to resolve the issue: https://github.com/terraform-providers/terraform-provider-aws/issues/227

To summarize:
Currently the key associate_public_ip_addresses can be in three states: true, false, nil (not defined). The behavior of the provider, because of how GetOk behaves sets the value to nil when false was the value in the template. This is not the expected behavior, and the default type should NOT be boolean in the schema.

Why we need this:
When launching kubernetes nodes using a LC it will become important to disable public IPs in a production environment and not necessarily adopt the default of the vpc/subnet. 

Note to reviewer:
Please feel free to help me make this cleaner/clearer and consistent with the expectations of this repo, which I am new to. Thank you for your time!

### DIFF
--- a/aws/resource_aws_launch_configuration.go
+++ b/aws/resource_aws_launch_configuration.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"log"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


alejandroEsc:
@radeksimko could I get an eta on when this would get reviewed? Thanks!

grubernaut:
Hi @alejandroEsc, thank you so much for the contribution!

For now, going to close this PR with the response given in #227, but happy to re-open and discuss further if needed. 
Thanks again!

alejandroEsc:
@grubernaut please warn and contact me before closing PR and allow for a discussion. I would like to strongly ask you to revert and open this PR again. I have, as you said, left a comment on #227 as a response.

PR #1193, opened 2017-07-20
Remove `network_interface_id`

grubernaut:
Finally remove `network_interface_id` in favor of
`primary_network_interface_id`.


catsby:
We discussed that we would delay merging this PR a few days, so please hold of

PR #1195, opened 2017-07-20
Print whole container_definitions in aws_ecs_task_definition instead of checksum in the plan

s-maj:
Currently, when aws_ecs_task_definition is updated, people can see only container_definitions checksum in the plan. There is no visibility what exactly is being pushed to ECS and I think this quite confusing. In other resources, like aws_iam_policy, old version of policy and new version policy is visible in the plan. Also, this change would allow to utilize terraform-landspace to produce a pretty plan for ECS tasks ([screenshot](http://imgur.com/a/KVXKu)).

Tests:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSEcsTaskDefinition'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEcsTaskDefinition -timeout 120m
=== RUN   TestAccAWSEcsTaskDefinition_basic
--- PASS: TestAccAWSEcsTaskDefinition_basic (19.93s)
=== RUN   TestAccAWSEcsTaskDefinition_withScratchVolume
--- PASS: TestAccAWSEcsTaskDefinition_withScratchVolume (12.85s)
=== RUN   TestAccAWSEcsTaskDefinition_withEcsService
--- PASS: TestAccAWSEcsTaskDefinition_withEcsService (138.14s)
=== RUN   TestAccAWSEcsTaskDefinition_withTaskRoleArn
--- PASS: TestAccAWSEcsTaskDefinition_withTaskRoleArn (19.50s)
=== RUN   TestAccAWSEcsTaskDefinition_withNetworkMode
--- PASS: TestAccAWSEcsTaskDefinition_withNetworkMode (21.53s)
=== RUN   TestAccAWSEcsTaskDefinition_constraint
--- PASS: TestAccAWSEcsTaskDefinition_constraint (10.78s)
=== RUN   TestAccAWSEcsTaskDefinition_changeVolumesForcesNewResource
--- PASS: TestAccAWSEcsTaskDefinition_changeVolumesForcesNewResource (19.44s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	242.187s
```

### DIFF
--- a/aws/ecs_task_definition_equivalency.go
+++ b/aws/ecs_task_definition_equivalency.go
@@ -1,0 +1,80 @@
+package aws
+
+import (
+	"bytes"
+	"encoding/json"
+	"reflect"
+
+	"github.com/aws/aws-sdk-go/private/protocol/json/jsonutil"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/mitchellh/copystructure"
+)
+
+func ecsContainerDefinitionsAreEquivalent(def1, def2 string) (bool, error) {
+	var obj1 containerDefinitions
+	err := json.Unmarshal([]byte(def1), &obj1)
+	if err != nil {
+		return false, err
+	}
+	err = obj1.Reduce()
+	if err != nil {
+		return false, err
+	}
+	canonicalJson1, err := jsonutil.BuildJSON(obj1)
+	if err != nil {
+		return false, err
+	}
+
+	var obj2 containerDefinitions
+	err = json.Unmarshal([]byte(def2), &obj2)
+	if err != nil {
+		return false, err
+	}
+	err = obj2.Reduce()
+	if err != nil {
+		return false, err
+	}
+	canonicalJson2, err := jsonutil.BuildJSON(obj2)
+	if err != nil {
+		return false, err
+	}
+
+	return bytes.Compare(canonicalJson1, canonicalJson2) == 0, nil
+}
+
+type containerDefinitions []*ecs.ContainerDefinition
+
+func (cd containerDefinitions) Reduce() error {
+	for i, def := range cd {
+		// Deal with special fields which have defaults
+		for j, pm := range def.PortMappings {
+			if pm.Protocol != nil && *pm.Protocol == "tcp" {
+				cd[i].PortMappings[j].Protocol = nil
+			}
+			if pm.HostPort != nil && *pm.HostPort == 0 {
+				cd[i].PortMappings[j].HostPort = nil
+			}
+		}
+
+		// Create a mutable copy
+		defCopy, err := copystructure.Copy(def)
+		if err != nil {
+			return err
+		}
+
+		definition := reflect.ValueOf(defCopy).Elem()
+		for i := 0; i < definition.NumField(); i++ {
+			sf := definition.Field(i)
+
+			// Set all empty slices to nil
+			if sf.Kind() == reflect.Slice {
+				if sf.IsValid() && !sf.IsNil() && sf.Len() == 0 {
+					sf.Set(reflect.Zero(sf.Type()))
+				}
+			}
+		}
+		iface := definition.Interface().(ecs.ContainerDefinition)
+		cd[i] = &iface
+	}
+	return nil
+}
--- a/aws/ecs_task_definition_equivalency_test.go
+++ b/aws/ecs_task_definition_equivalency_test.go
@@ -1,0 +1,173 @@
+package aws
+
+import (
+	"testing"
+)
+
+func TestAwsEcsContainerDefinitionsAreEquivalent_basic(t *testing.T) {
+	cfgRepresention := `
+[
+    {
+      "name": "wordpress",
+      "links": [
+        "mysql"
+      ],
+      "image": "wordpress",
+      "essential": true,
+      "portMappings": [
+        {
+          "containerPort": 80,
+          "hostPort": 80
+        }
+      ],
+      "memory": 500,
+      "cpu": 10
+    },
+    {
+      "environment": [
+        {
+          "name": "MYSQL_ROOT_PASSWORD",
+          "value": "password"
+        }
+      ],
+      "name": "mysql",
+      "image": "mysql",
+      "cpu": 10,
+      "memory": 500,
+      "essential": true
+    }
+]`
+
+	apiRepresentation := `
+[
+    {
+        "name": "wordpress",
+        "image": "wordpress",
+        "cpu": 10,
+        "memory": 500,
+        "links": [
+            "mysql"
+        ],
+        "portMappings": [
+            {
+                "containerPort": 80,
+                "hostPort": 80,
+                "protocol": "tcp"
+            }
+        ],
+        "essential": true,
+        "environment": [],
+        "mountPoints": [],
+        "volumesFrom": []
+    },
+    {
+        "name": "mysql",
+        "image": "mysql",
+        "cpu": 10,
+        "memory": 500,
+        "portMappings": [],
+        "essential": true,
+        "environment": [
+            {
+                "name": "MYSQL_ROOT_PASSWORD",
+                "value": "password"
+            }
+        ],
+        "mountPoints": [],
+        "volumesFrom": []
+    }
+]`
+
+	equal, err := ecsContainerDefinitionsAreEquivalent(cfgRepresention, apiRepresentation)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !equal {
+		t.Fatal("Expected definitions to be equal.")
+	}
+}
+
+func TestAwsEcsContainerDefinitionsAreEquivalent_portMappings(t *testing.T) {
+	cfgRepresention := `
+[
+    {
+      "name": "wordpress",
+      "image": "wordpress",
+      "essential": true,
+      "portMappings": [
+        {
+          "containerPort": 80
+        }
+      ],
+      "memory": 500,
+      "cpu": 10
+    }
+]`
+
+	apiRepresentation := `
+[
+    {
+        "name": "wordpress",
+        "image": "wordpress",
+        "cpu": 10,
+        "memory": 500,
+        "portMappings": [
+            {
+                "containerPort": 80,
+                "hostPort": 0,
+                "protocol": "tcp"
+            }
+        ],
+        "essential": true,
+        "environment": [],
+        "mountPoints": [],
+        "volumesFrom": []
+    }
+]`
+
+	equal, err := ecsContainerDefinitionsAreEquivalent(cfgRepresention, apiRepresentation)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !equal {
+		t.Fatal("Expected definitions to be equal.")
+	}
+}
+
+func TestAwsEcsContainerDefinitionsAreEquivalent_negative(t *testing.T) {
+	cfgRepresention := `
+[
+    {
+      "name": "wordpress",
+      "image": "wordpress",
+      "essential": true,
+      "memory": 500,
+      "cpu": 10,
+      "environment": [
+        {"name": "EXAMPLE_NAME", "value": "foobar"}
+      ]
+    }
+]`
+
+	apiRepresentation := `
+[
+    {
+        "name": "wordpress",
+        "image": "wordpress",
+        "cpu": 10,
+        "memory": 500,
+        "essential": true,
+        "environment": [],
+        "mountPoints": [],
+        "volumesFrom": []
+    }
+]`
+
+	equal, err := ecsContainerDefinitionsAreEquivalent(cfgRepresention, apiRepresentation)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if equal {
+		t.Fatal("Expected definitions to differ.")
+	}
+}
--- a/aws/resource_aws_ecs_task_definition.go
+++ b/aws/resource_aws_ecs_task_definition.go
@@ -2,8 +2,6 @@ package aws
 
 import (
 	"bytes"
-	"crypto/sha1"
-	"encoding/hex"
 	"fmt"
 	"log"
 	"strings"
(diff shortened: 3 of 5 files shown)


astawiarski:
This would be an incredible improvement - it would solve our current visibility issues during the plan stage.

s-maj:
Hi @radeksimko! Is there a chance to merge this soon? I took a look at the latest Hashicorp blog post (https://www.hashicorp.com/blog/continuous-deployment-with-nomad-and-terraform/) and even Nomad provider uses full jobspec instead of hash :)

Ninir:
Hey folks,

In my opinion, this is a great addition. The thing that bothers me is that people will see a diff when migrating from the hash-ed version to this plain-text one.
I would like to keep it silent, using the DiffSuppressFunc, thus being able to detect whether the value is encrypted or not...

What do you think @radeksimko, is it worth it?

radeksimko:
Hi @s-maj
thanks for the contribution.
The idea looks reasonable, I can't remember why did I use the hash back then in the initial implementation 🤔 

> The thing that bothers me is that people will see a diff when migrating from the hash-ed version to this plain-text one.

Yep, that's exactly my concern, too.

We can't merge this as is otherwise we'd upset many users that have existing task definitions in their state and schema change like this would cause recreation of the whole TD because hash != content from diff perspective. Just try creating TD with current stable version of Terraform/AWS provider, then run `plan` with compiled provider from your branch and you'll see what I mean.

The best (and idiomatic) way to address this is to add a state migration.

Take a look at [existing state migrations](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_key_pair_migrate.go) and [tests](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_key_pair_migrate_test.go) if you need any inspiration and/or let me know if anything's unclear in regards to migrations.

Here's full diff (hopefully) explaining what it takes to do the whole migration: https://github.com/terraform-providers/terraform-provider-aws/commit/e253f88df45c45abfdee99c37c20a049a40d1bad

s-maj:
Hi @radeksimko

Yep, totally. 

About schema migration, from what I understand, task definition "blob" is being reduced to sha1 hash and this is stored in state file as an attribute of the resource. By doing migration I can only operate on the hash which might quite hard to reverse :) 

I guess I need to replace old value in this field with new value from the input, but I don't if this is possible or if it's a good approach.

radeksimko:
> I guess I need to replace old value in this field with new value from the input, but I don't if this is possible or if it's a good approach.

There isn't any way to get access to the config, which may is probably by design.

I think the best way forward is to pull down the current body from the ECS API. The migration function interface has `meta` which contains the ECS connection, so you should be able to do that.

It may not always be accurate, because state may be drifted from the reality, but Terraform should refresh the state automatically as part of `plan` or `apply`, unless the user passed `-refresh=false`.

s-maj:
Alright, I guess it should look like this:
```
conn := meta.(*AWSClient).ecsconn

out, err := conn.DescribeTaskDefinition(&ecs.DescribeTaskDefinitionInput{
	TaskDefinition: aws.String(is.Attributes["arn"]),
})

if err != nil {
	return is, nil
}

taskDefinition := out.TaskDefinition

strB, _ := json.Marshal(taskDefinition.ContainerDefinitions)
is.Attributes["container_definitions"] = string(strB)
```
but I am getting tons of nulls or empty lists (fields not present in Terraform) in the output:
```
[{\"Command\":null,\"Cpu\":10,\"DisableNetworking\":null,\"DnsSearchDomains\":null,\"DnsServers\":null,\"DockerLabels\":null,\"DockerSecurityOptions\":null,\"EntryPoint\":null,\"Environment\":[],\"Essential\":true,\"ExtraHosts\":null,\"Hostname\":null,\"Image\":\"service-first\",\"Links\":null,\"LogConfiguration\":null,\"Memory\":512,\"MemoryReservation\":null,\"MountPoints\":[],\"Name\":\"first\",\"PortMappings\":[{\"ContainerPort\":80,\"HostPort\":80,\"Protocol\":\"tcp\"}],\"Privileged\":null,\"ReadonlyRootFilesystem\":null,\"Ulimits\":null,\"User\":null,\"VolumesFrom\":[],\"WorkingDirectory\":null},{\"Command\":null,\"Cpu\":10,\"DisableNetworking\":null,\"DnsSearchDomains\":null,\"DnsServers\":null,\"DockerLabels\":null,\"DockerSecurityOptions\":null,\"EntryPoint\":null,\"Environment\":[],\"Essential\":true,\"ExtraHosts\":null,\"Hostname\":null,\"Image\":\"service-second\",\"Links\":null,\"LogConfiguration\":null,\"Memory\":256,\"MemoryReservation\":null,\"MountPoints\":[],\"Name\":\"second\",\"PortMappings\":[{\"ContainerPort\":443,\"HostPort\":443,\"Protocol\":\"tcp\"}],\"Privileged\":null,\"ReadonlyRootFilesystem\":null,\"Ulimits\":null,\"User\":null,\"VolumesFrom\":[],\"WorkingDirectory\":null}]" => "[\n  {\n    \"name\": \"first\",\n    \"image\": \"service-first\",\n    \"cpu\": 10,\n    \"memory\": 512,\n    \"essential\": true,\n    \"portMappings\": [\n      {\n        \"containerPort\": 80,\n        \"hostPort\": 80\n      }\n    ]\n  },\n  {\n    \"name\": \"second\",\n    \"image\": \"service-second\",\n    \"cpu\": 10,\n    \"memory\": 256,\n    \"essential\": true,\n    \"portMappings\": [\n      {\n        \"containerPort\": 443,\n        \"hostPort\": 443\n      }\n    ]\n  }\n]\n" (forces new resource)
```
which is more or less exptected since `json:",omitempty"` is not set in ContainerDefinition. There is a ticket about this topic (https://github.com/aws/aws-sdk-go/issues/271) but it's not going to happen any time soon. 

I tried to work on string output from SDK but even AWS API responds with empty fields:
```
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws: 2017/08/27 13:48:25 [DEBUG] Received task definition {
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:   TaskDefinition: {
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     ContainerDefinitions: [{
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Cpu: 10,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Environment: [],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Essential: true,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Image: "service-first",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Memory: 512,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         MountPoints: [],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Name: "first",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         PortMappings: [{
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:             ContainerPort: 80,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:             HostPort: 80,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:             Protocol: "tcp"
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:           }],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         VolumesFrom: []
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:       },{
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Cpu: 10,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Environment: [],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Essential: true,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Image: "service-second",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Memory: 256,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         MountPoints: [],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         Name: "second",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         PortMappings: [{
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:             ContainerPort: 443,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:             HostPort: 443,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:             Protocol: "tcp"
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:           }],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:         VolumesFrom: []
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:       }],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     Family: "service2",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     PlacementConstraints: [],
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     Revision: 1,
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     Status: "ACTIVE",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     TaskDefinitionArn: "arn:aws:ecs:eu-west-2:XXXXXXXXXXXX:task-definition/service2:1",
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:     Volumes: []
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws:   }
2017/08/27 13:48:25 [DEBUG] plugin: terraform-provider-aws: }
```
where orginal task definion is:
```
[
  {
    "name": "first",
    "image": "service-first",
    "cpu": 10,
    "memory": 512,
    "essential": true,
    "portMappings": [
      {
        "containerPort": 80,
        "hostPort": 80
      }
    ]
  },
  {
    "name": "second",
    "image": "service-second",
    "cpu": 10,
    "memory": 256,
    "essential": true,
    "portMappings": [
      {
        "containerPort": 443,
        "hostPort": 443
      }
    ]
  }
]
```

@radeksimko could you please advise how to approach this?

radeksimko:
I see - that's annoying - it might have been the reason why I originally decided to use hash instead 😄 

However we now have some ways to deal with equivalent JSON structures which aren't exactly same in string-based match. Take a look at https://github.com/terraform-providers/terraform-provider-aws/blob/fd6ad1d8c976ee779ff9a543d1ad128166959d01/aws/resource_aws_cloudwatch_dashboard.go#L40

which should prevent spurious diffs from appearing. I hope the problem is generic enough we can use the existing generic JSON helper. Eventually we'd need to create a custom one, similar to the one we did for IAM policies https://github.com/terraform-providers/terraform-provider-aws/blob/fd6ad1d8c976ee779ff9a543d1ad128166959d01/aws/resource_aws_iam_policy.go#L42

which is work worth its own [package/repository](https://github.com/jen20/awspolicyequivalence).

s-maj:
Existing diff suppression functions won't help in this case and fixing this is a bit beyond my skill set :) But it seems NLB will be part of 1.0.0 which will break backward compatibility for ECS services (at least web based containers) so it might good moment to add this PR to this mix (maybe along with https://github.com/terraform-providers/terraform-provider-aws/pull/1476)

radeksimko:
@s-maj I just pushed a few enhancements to your branch:
 - state migration, so it's not a breaking change anymore
 - diff suppression function, to avoid showing no-op differences between API representation and config, like empty array and no array
 - fixed flattening of definitions, so that we can correctly show genuine differences during `plan`

Can you take a look at those changes and try them out to see if it's working for you as expected in your context?

btw. the reason you were getting the odd diff with `null` values is because the SDK struct does not use standard `json` tags which means that standard (un)marshaler doesn't (un)marshal the data correctly. I had to use the custom marshaler from SDK's package.

radeksimko:
FYI: I will merge this on Monday, 13th if there's no further feedback.

s-maj:
I'll take a look over weekend. Thanks!

s-maj:
@radeksimko All grand, thanks a million!

achille-roussel:
Hey guys, we're seeing terraform recreate the task definitions on every plan and all that changed is the version of the AWS provider on our end.

Is anybody else experiencing this issue?

bturbes:
We are also seeing this same behavior.

ziggythehamster:
Also seeing it here.

This code needs to be amended so that empty arrays coming back from AWS are treated the same as the JSON missing the empty array. Also, the `environment` array can come back in any order from AWS even if your task definition has them in a different order.

nathanielks:
likewise!

ckyoog:
Hi @radeksimko This change is good, but I prefer the chksum way. In my case, my container definition contains some credential information, I don't want them to be printed out on screen when doing plan. Could you please at least offer an option to make it still print chksum number instead of whole JSON text?

tomelliff:
I wouldn't recommend putting secrets in the task definition. I presume you're injecting them as environment variables? You might want to consider fetching them from SSM parameter store or something like Vault at runtime in an entrypoint script.

ckyoog:
@tomelliff, yes, you guess right. I know it is not a good idea to use environment variables for such thing. But anyway, I think it still might be useful to print and store chksum in some situations (?). I will appreciate it very much if you can think about it. :)

achille-roussel:
This change also has a sad side effect of making the plan output so large that its truncated in the terraform enterprise web view, making the plan impossible to review. 

It’s kind of a TFE issue but making it somehow an opt-in feature would be useful in that case.

PR #1197, opened 2017-07-20
Add a body property to API Gateway RestAPI for Swagger import support

lukehoban:
## Description

The `body` property allows supplying a Swagger/OpenAPI spec to a RestAPI to create the API definition using the routes and integrations defined in the spec.

Related to https://github.com/hashicorp/terraform/issues/11034 and the work done in https://github.com/hashicorp/terraform/pull/6175.

Provides similar behaviour to what's available in CloudFormation with: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-body.

## Tests
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayRestApi_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayRestApi_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayRestApi_basic
--- PASS: TestAccAWSAPIGatewayRestApi_basic (50.66s)
=== RUN   TestAccAWSAPIGatewayRestApi_openapi
--- PASS: TestAccAWSAPIGatewayRestApi_openapi (28.66s)
PASS
ok  	github.com/pulumi/terraform-provider-aws/aws	79.354s
```

### DIFF
--- a/aws/resource_aws_api_gateway_rest_api.go
+++ b/aws/resource_aws_api_gateway_rest_api.go
@@ -37,6 +37,11 @@ func resourceAwsApiGatewayRestApi() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 
+			"body": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
 			"root_resource_id": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -75,6 +80,17 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
 	}
 
 	d.SetId(*gateway.Id)
+
+	if body, ok := d.GetOk("body"); ok {
+		_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
+			RestApiId: gateway.Id,
+			Mode:      aws.String("overwrite"),
+			Body:      []byte(body.(string)),
+		})
+		if err != nil {
+			return fmt.Errorf("Error putting API Gateway specification: %s", err)
+		}
+	}
 
 	if err = resourceAwsApiGatewayRestApiRefreshResources(d, meta); err != nil {
 		return err
@@ -154,6 +170,19 @@ func resourceAwsApiGatewayRestApiUpdateOperations(d *schema.ResourceData) []*api
 func resourceAwsApiGatewayRestApiUpdate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).apigateway
 	log.Printf("[DEBUG] Updating API Gateway %s", d.Id())
+
+	if d.HasChange("body") {
+		if body, ok := d.GetOk("body"); ok {
+			_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
+				RestApiId: aws.String(d.Id()),
+				Mode:      aws.String("overwrite"),
+				Body:      []byte(body.(string)),
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
 
 	_, err := conn.UpdateRestApi(&apigateway.UpdateRestApiInput{
 		RestApiId:       aws.String(d.Id()),
--- a/aws/resource_aws_api_gateway_rest_api.go
+++ b/aws/resource_aws_api_gateway_rest_api.go
@@ -8,6 +8,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/apigateway"
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
@@ -82,13 +83,14 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
 	d.SetId(*gateway.Id)
 
 	if body, ok := d.GetOk("body"); ok {
+		log.Printf("[DEBUG] Initializing API Gateway from OpenAPI spec %s", d.Id())
 		_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
 			RestApiId: gateway.Id,
-			Mode:      aws.String("overwrite"),
+			Mode:      aws.String(apigateway.PutModeOverwrite),
 			Body:      []byte(body.(string)),
 		})
 		if err != nil {
-			return fmt.Errorf("Error putting API Gateway specification: %s", err)
+			return errwrap.Wrapf("Error creating API Gateway specification: {{err}}", err)
 		}
 	}
 
@@ -173,13 +175,14 @@ func resourceAwsApiGatewayRestApiUpdate(d *schema.ResourceData, meta interface{}
 
 	if d.HasChange("body") {
 		if body, ok := d.GetOk("body"); ok {
+			log.Printf("[DEBUG] Updating API Gateway from OpenAPI spec: %s", d.Id())
 			_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
 				RestApiId: aws.String(d.Id()),
-				Mode:      aws.String("overwrite"),
+				Mode:      aws.String(apigateway.PutModeOverwrite),
 				Body:      []byte(body.(string)),
 			})
 			if err != nil {
-				return err
+				return errwrap.Wrapf("Error updating API Gateway specification: {{err}}", err)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


lukehoban:
@radeksimko Please let me know if there's any questions on concerns on this that I can help address.

lukehoban:
Updated tests to create valid REST API definitions and add test results to summary.

cc @radeksimko @Ninir @betabandido @apparentlymart for feedback.

aws/resource_aws_api_gateway_rest_api.go
```diff
@@ -76,6 +81,17 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
...
 	d.SetId(*gateway.Id)
 
+	if body, ok := d.GetOk("body"); ok {
+		_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
+			RestApiId: gateway.Id,
+			Mode:      aws.String("overwrite"),
+			Body:      []byte(body.(string)),
+		})
+		if err != nil {
+			return fmt.Errorf("Error putting API Gateway specification: %s", err)
```
Ninir: Could you use `errwrap.Wrapf("Error creating API Gateway specification: {{err}}", err)` there please? (notice the creating word, to make it clear we are putting by creating)

aws/resource_aws_api_gateway_rest_api.go
```diff
@@ -76,6 +81,17 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(*gateway.Id)
 
+	if body, ok := d.GetOk("body"); ok {
+		_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
+			RestApiId: gateway.Id,
+			Mode:      aws.String("overwrite"),
```
Ninir: Can you use the `apigateway.PutModeOverwrite` constant here and line 178?

aws/resource_aws_api_gateway_rest_api_test.go
```diff
@@ -144,3 +216,81 @@ resource "aws_api_gateway_rest_api" "test" {
...
+              "statusCode": 200
+            }
+          }
+        }
+      }
+    }
+  }
+}
+EOF
+}
```
Ninir: Can you fix the space indentation here please? 😄 (mostly at the beginning of the JSON structure)

aws/resource_aws_api_gateway_rest_api.go
```diff
@@ -155,6 +171,19 @@ func resourceAwsApiGatewayRestApiUpdate(d *schema.ResourceData, meta interface{}
...
 
+	if d.HasChange("body") {
+		if body, ok := d.GetOk("body"); ok {
+			_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
+				RestApiId: aws.String(d.Id()),
+				Mode:      aws.String("overwrite"),
+				Body:      []byte(body.(string)),
+			})
+			if err != nil {
+				return err
```
Ninir: I would go for the same as line 91 here: `errwrap.Wrapf("Error updating API Gateway specification: {{err}}", err)`. Thoughts?

aws/resource_aws_api_gateway_rest_api.go
```diff
@@ -76,6 +81,17 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(*gateway.Id)
 
+	if body, ok := d.GetOk("body"); ok {
+		_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
```
Ninir: Could you log in the logs that we are setting the specification?
Just for verbosity & debugging :D

Same for the update would be highly appreciated!

aws/resource_aws_api_gateway_rest_api.go
radeksimko: It does no harm, but I think the wrapper here isn't necessary as the only place the error will escalate to is the UI, where all that matters is string-based error message user will read, not the original error type.

aws/resource_aws_api_gateway_rest_api.go
radeksimko: It does no harm, but I think the wrapper here isn't necessary as the only place the error will escalate to is the UI, where all that matters is string-based error message user will read, not the original error type.

radeksimko:
Thanks @Ninir - other than what you mentioned we should also document clearly which resources collide with this field/approach, I believe these are:

 - `aws_api_gateway_gateway_response`
 - `aws_api_gateway_integration`
 - `aws_api_gateway_integration_response`
 - `aws_api_gateway_method`
 - `aws_api_gateway_method_response`
 - `aws_api_gateway_method_settings`
 - `aws_api_gateway_model`
 - `aws_api_gateway_resource`

i.e. it should be clear to the user that they cannot use both Swagger and these resources together for the same API.

lukehoban:
Thanks @ninir and @radeksimko!  

Updated PR to address feedback.  Please take a look at the doc updates in particular.

website/docs/r/api_gateway_rest_api.html.markdown
```diff
@@ -26,6 +26,18 @@ The following arguments are supported:
 * `name` - (Required) The name of the REST API
 * `description` - (Optional) The description of the REST API
 * `binary_media_types` - (Optional) The list of binary media types supported by the RestApi. By default, the RestApi supports only UTF-8-encoded text payloads.
+* `body` - (Optional) An OpenAPI specification that defines the set of routes and integrations to create as part of the REST API.
+
+__Note__: If the `body` argument is provided, the OpenAPI specification will be used to configure the resources, methods and integrations for the Rest API. If this argument is provided, the following resources should not be configured manually for the same REST API, as updates may cause manual resource updates to be overwritten:
```
radeksimko: > the following resources should not be configured manually

I'd rather say _should not be managed as separate ones_ as that's more in-line with the language used in the rest of our docs.

Ninir:
hey @lukehoban !

Just pushed the documentation update following @radeksimko 's recommandation.
Thanks for all the work here, looks very good! :)

Cheers!

kirangangineni:
Hi All - When will this feature be available. I have v0.10.5 on my mac. 

resource "aws_api_gateway_rest_api" "MyDemoAPI" {
  name        = "MyDemoAPI"
  description = "This is my API for demonstration purposes"
  body = "./api/swagger.yml"
}

when I run terraform plan , I get 
1 error(s) occurred:

* module.apigateway.aws_api_gateway_rest_api.MyDemoAPI: : invalid or unknown key: body

Am I missing some thing here. Please let me know

bvingerh:
Same with v0.10.6: body argument not recognized.
That's the fourth release of Terraform without a release of the AWS provider (which is stuck at v0.1.4 of August 8)...

kirangangineni:
I built the terraform-provider-aws using the instructions from README. Now when I run terraform plan, the body tag is recognized. But when I apply, I get the following error. I used the same swagger with AWS SDK/CLI and it worked fine. Can you please help

$ terraform apply
module.apigateway.aws_api_gateway_rest_api.MyDemoAPI: Creating...
  body:             "" => "../api/swagger.json"
  created_date:     "" => "<computed>"
  description:      "" => "This is my API for demonstration purposes"
  name:             "" => "MyDemoAPI"
  root_resource_id: "" => "<computed>"
Error applying plan:

1 error(s) occurred:

* module.apigateway.aws_api_gateway_rest_api.MyDemoAPI: 1 error(s) occurred:

* aws_api_gateway_rest_api.MyDemoAPI: Error creating API Gateway specification: BadRequestException: Invalid Swagger 2.0 input.
	status code: 400, request id: 5ec75ec4-a235-11e7-b1f5-7b2e853b338e

woody3549:
Hi all,

Thanks for this great work.
Can you please let me know when it's going to be released ?

Thanks

ebarault:
i have the same error as @kirangangineni, with the same difference that i'm using a yaml swagger file. Are we supposed to provide the file in a special manner?

Ninir:
Hi folks,

the AWS provider (this repository) is release independently from the core one (https://github.com/hashicorp/terraform).
This means you can use Terraform 0.10.8 with the AWS Provider version 1.0.0, 1.1.0, etc..

the body property was added in the 1.0.0 release: to ensure you get it right, remove the `.terraform/plugins` directory and run `terraform init` again: this will get the latest version and you should be all set.

Thus being said, @ebarault is right: the body property expects the body of the file, not the path to the file. As in:

# using data template
```hcl
resource "aws_api_gateway_rest_api" "api" {
  ...
  body = "${data.template_file.swagger_api.rendered}"
}
```

# using a file
```hcl
resource "aws_api_gateway_rest_api" "api" {
  ...
  body = "${file("myfile.json"}"
}
```

# using inline strings
```hcl
resource "aws_api_gateway_rest_api" "api" {
  ...
  body = <<YAML
myfile
YAML
}
```

Tell us how it goes! :)

Willis0826:
@Ninir 
Thanks for the clearly demonstration : ) But here is a typo in your post :
> body = "${file("myfile.json"}"
I believe it should be ${file("...")}

PR #1200, opened 2017-07-21
resource_aws_instance: allow private instance in public subnet.  

alejandroEsc:
We allow the proper use of the associate_public_ip_addresses flag which means the ability to create a private instance when defining a public subnet id.

To summarize:
Currently the key associate_public_ip_addresses can be in three states: true, false, nil (not defined). The behavior of the provider, because of how GetOk behaves sets the value to nil when false was the value in the template. This is not the expected behavior, and the default type should NOT be boolean in the schema.

Please note this is also related to the PR: https://github.com/terraform-providers/terraform-provider-aws/pull/1185

Why we need this:
When launching kubernetes nodes, it may become important to deploy instances (such as in the case of etcd) and ensure private ips only.

Note to reviewer:
Please feel free to help me make this cleaner/clearer and consistent with the expectations of this repo, which I am new to. Thank you for your time!

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strconv"
 	"strings"
 	"time"
 
@@ -46,7 +47,7 @@ func resourceAwsInstance() *schema.Resource {
 			},
 
 			"associate_public_ip_address": {
-				Type:     schema.TypeBool,
+				Type:     schema.TypeString,
 				ForceNew: true,
 				Computed: true,
 				Optional: true,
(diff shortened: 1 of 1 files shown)


alejandroEsc:
@radeksimko this resolves a bug tightly related to PR: #1185 mentioned above. Could I get an eta when this would be reviewed?

grubernaut:
Hi @alejandroEsc, thank you so much for the contribution!

For now, going to close this PR with the response given in #227, but happy to re-open and discuss further if needed. 
Thanks again!

alejandroEsc:
@grubernaut please warn and contact me before closing PR and allow for a discussion. I would like to strongly ask you to revert and open this PR again. I have, as you said, left a comment  on https://github.com/terraform-providers/terraform-provider-aws/issues/227 as a response.

PR #1201, opened 2017-07-21
Append the info about IAM role policy import

cazorla19:


website/docs/r/iam_role_policy.html.markdown
```diff
@@ -72,3 +72,11 @@ assign a random, unique name.
 * `name` - The name of the policy.
 * `policy` - The policy document attached to the role.
 * `role` - The role to which this policy applies.
+
+## Import
+
+IAM Role Policies can be imported using the `name`, e.g.
+
+```
+$ terraform import aws_iam_role_policy.mypolicy mypolicy_name
```
Ninir: It looks like we are missing a bit more info there, as you can see in the [code](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_iam_role_policy.go#L80): the ID for this resource is constructed as `<role name>:<policy name>`.

cazorla19: Thanks for the remark, Ninir. Yes, actually it's a `role_name:policy_name` syntax. I tried an import too far and because of this I did a mistake. Sorry for this. Please, review my last commit

PR #1203, opened 2017-07-21
fix empty key_name on aws_spot_fleet_request

svenwltr:
This fixes one part of https://github.com/terraform-providers/terraform-provider-aws/issues/492.

Basically Terraform sends an empty string for the `key_name` values to the AWS API, if the value isn't set in the template. This causes the instance creation to fail:

```
Instance fail to start with failed: Invalid value '' for keyPairNames. It should not be blank.
```

I am not sure, if it is worth doing an acceptance test here. It seems a bit complicated, because it would require to wait for the spot fleet instance creation. I am not sure how to handle this in a acceptance test (maybe with timeouts?). Please tell me, if I should add an test here.

### DIFF
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -350,7 +350,7 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
 		opts.UserData = aws.String(base64Encode([]byte(v.(string))))
 	}
 
-	if v, ok := d["key_name"]; ok {
+	if v, ok := d["key_name"]; ok && v != "" {
 		opts.KeyName = aws.String(v.(string))
 	}
 
--- a/aws/resource_aws_spot_fleet_request_test.go
+++ b/aws/resource_aws_spot_fleet_request_test.go
@@ -1364,9 +1364,10 @@ resource "aws_spot_fleet_request" "foo" {
     target_capacity = 1
     valid_until = "2019-11-04T20:44:20Z"
     terminate_instances_with_expiration = true
+    wait_for_fulfillment = true
     launch_specification {
         instance_type = "m1.small"
-        ami = "ami-d06a90b0"
+        ami = "ami-516b9131"
 
 	ebs_block_device {
             device_name = "/dev/xvda"


radeksimko:
Hi @svenwltr 
thanks for this patch!

> I am not sure, if it is worth doing an acceptance test here. It seems a bit complicated, because it would require to wait for the spot fleet instance creation. I am not sure how to handle this in a acceptance test (maybe with timeouts?). Please tell me, if I should add an test here.

It would be better to add a test covering this bugfix. We [already wait for fulfilment](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L610-L624), so there should be no need to add any extra logic in that sense. Just make sure the spot price & instance size is aligned with other requests from our testing suite. Feel free to take the "copy-paste-modify" approach if you like.
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request_test.go

svenwltr:
@radeksimko AFAIS `active` only means, that the spot fleet request is created and enabled. It might still fail, if there is an error in the `launch_specification`.

I created [a test](https://github.com/rebuy-de/terraform-provider-aws/pull/1/commits/d9d85a4017e5639d6b5ad3eaa6b487595344c713) and removed my fix for testing. It succeeds, but it succeeded which shouldn't happen. This ist the History of the Spot Fleet Request:

screenshot 2017-07-24 14 30 39 02 00

svenwltr:
Okay, this is what I found out after digging a bit deeper:

* A Spot Fleet Request has an `ActivityStatus` and a `SpotFleetRequestState` ([source](https://godoc.org/github.com/aws/aws-sdk-go/service/ec2#SpotFleetRequestConfig)). The "State" doesn't respect the progress of the "child instances".
* The [fulfillment check](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L610-L624) uses `SpotFleetRequestState`, which means that the Terraform run might succeed, despite the actual instances don't start.
* If I use the `ActivityStatus` in the [fulfillment check](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L610-L624) other tests also fail (or at least `TestAccAWSSpotFleetRequest_placementTenancy` with: `Non-Windows instances with a virtualization type of 'hvm' are currently not supported for this instance type.`)

It looks easy to change the behavior of the [fulfillment check](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L610-L624), so it waits for the instances to actually be ready, but I cannot estimate the impact for such a change.

radeksimko:
Ah, I see, apologies for misleading info then!

It looks like we should add an optional fulfilment check, like we have in https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_instance_request.go#L54-L58

Are you willing to do this in a separate PR?

> If I use the ActivityStatus in the fulfillment check other tests also fail (or at least TestAccAWSSpotFleetRequest_placementTenancy with: Non-Windows instances with a virtualization type of 'hvm' are currently not supported for this instance type.)

Yeah, that means those tests are _actually_ broken. 😞

svenwltr:
@radeksimko I guess it makes sense to add the fulfilment check to all existing test cases, right?

radeksimko:
@svenwltr pretty much. I first thought we could default to wait, but it seems the `aws_spot_instance` [**doesn't** wait by default](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_instance_request.go#L57), so better to stay consistent, which does mean we'll need to modify all existing test cases.

svenwltr:
Here is the implementation so far: terraform-providers/terraform-provider-aws#1241

I still need some time to fix the tests.

svenwltr:
@radeksimko I now rebased to the current master and fixed the `TestAccAWSSpotFleetRequest_withEBSDisk` test. Since this one already doesn't use a `key_name`, I didn't create an extra test. Is this sufficient?

PR #1225, opened 2017-07-24
Add aws_iot_certificate resource 

AlexMabry:
Add IoT Certificate resource.  Code is based on work from @jhedev.  This one is super simple.


grubernaut:
Needs `go fmt` ran on source files to fix failing tests

aws/resource_aws_iot_certificate.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	log.Printf("[DEBUG] Created certificate from csr")
+
+	d.SetId(*out.CertificateId)
+
+	return resourceAwsIotCertificateRead(d, meta)
+}
+
+func resourceAwsIotCertificateRead(d *schema.ResourceData, meta interface{}) error {
+
```
grubernaut: Extra line after function header

aws/resource_aws_iot_certificate.go
```diff
@@ -0,0 +1,147 @@
...
+	out, err := conn.DescribeCertificate(&iot.DescribeCertificateInput{
+		CertificateId: aws.String(d.Id()),
+	})
+
+	if err != nil {
+		log.Printf("[ERROR] %s", err)
+		return err
+	}
+
+	d.Set("arn", out.CertificateDescription.CertificateArn)
```
grubernaut: is `active` also returned from the `DescribeCertificate` API call? If so, we need to set that here as well, to be able to catch any diffs between infra+state.

aws/resource_aws_iot_certificate.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			log.Printf("[ERROR] %s", err)
+			return nil
+		}
+
+		// If everything worked, make the old one inactive
+		_, err = conn.UpdateCertificate(&iot.UpdateCertificateInput{
+			CertificateId: aws.String(d.Id()),
+			NewStatus:     aws.String("INACTIVE"),
+		})
```
grubernaut: If I'm reading this correctly, if a user updates the `csr`, Terraform will create a new certificate resource, and leak the existing one? Should we delete the original certificate here as well?

aws/resource_aws_iot_certificate.go
AlexMabry: ok

aws/resource_aws_iot_certificate.go
AlexMabry: ok

aws/resource_aws_iot_certificate.go
AlexMabry: There is really no way to update the csr in API.  Would it be more straightforward to delete the old one or just not allow changing it in the first place?

AlexMabry:
I changed it to prevent changing the CSR, since it requires creating a new cert

Issue #1213, opened 2017-07-22
Empty assume_role causes a crash


charles-at-geospock:
### Terraform Version
Terraform v0.9.4

### Affected Resource(s)
assume_role {}

### Terraform Configuration Files
```hcl
provider "aws" {
  assume_role {
  }
}

data "aws_caller_identity" "current" {}
```
( not a fragment - that's all you need to provoke it )

### Panic Output
https://gist.github.com/charles-at-geospock/94c9b4ff7a1b406ce8cbcc490398cf59

### Expected Behavior
Nothing at all, in this example.

### Actual Behavior
Bang. It crashes because we try to use 'nil' to create a map.

### Steps to Reproduce
1. `terraform apply` on the example terraform script supplied above.

### Important Factoids
I tried to patch and came up with this:

```diff
diff --git a/aws/provider.go b/aws/provider.go
index fe99eead..5228994c 100644
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -688,6 +688,9 @@ func assumeRoleSchema() *schema.Schema {
 
 func assumeRoleToHash(v interface{}) int {
        var buf bytes.Buffer
+       if v == nil {
+               return 0
+       }
        m := v.(map[string]interface{})
        buf.WriteString(fmt.Sprintf("%s-", m["role_arn"].(string)))
        buf.WriteString(fmt.Sprintf("%s-", m["session_name"].(string)))
```

But I'm sure this is just a single case of a more widespread issue - I saw a couple of places where the ToHash functions were operating on values that were not checked. It would be nicer if a warning about an empty structure were generated, but I cannot see how to do that.

charles-at-geospock:
Also observed in 0.9.11, and this was actually tested with the separate provider in this repository against the 0.9.4 version.

charles-at-geospock:
ooh. deleting code. always a good thing :-)

PR #1227, opened 2017-07-24
provider: Remove assumeRoleHash

grubernaut:
The default hash function is good enough to use for the `assume_role` schema,
previously the custom hash function was causing a panic in the provider if the
`assume_role` configuration block was present but empty.

Fixes: #1213


charles-at-geospock:
Awesome; thanks guys :)

PR #1240, opened 2017-07-25
[WIP] resource/aws_dynamodb_table: Use just the name to determine hashcode of GSI

mcwqy9:
I'm trying to make changes to DynamoDB GSI capacity ignorable. This is likely the wrong approach. #CunninghamsLaw. I picked up the idea from something that was removed in https://github.com/hashicorp/terraform/pull/13256.

The catch-22 with this approach is if something is not included in calculating the return value of the Set function then when you change said thing it's not updated. I just wanted it to not change the resource address so that I can use `ignore_changes`. For people that _do_ want to control their GSI's capacity in terraform, a change should still be applicable. 

Even if I wanted to hardcode something nasty like `ignore_changes = ["global_secondary_index.3291674533.read_capacity"]`, I can't because "3291674533" is calculated based, in part, on the read_capacity. 

Other ideas:

1) Modify ignore_changes to accept the right kind of wildcard to support this case
2) Are GSIs a bad fit for the type TypeSet, given their mutability? Would it be better to split GSIs out into their own resource?

Closes https://github.com/terraform-providers/terraform-provider-aws/issues/671

### DIFF
--- a/aws/resource_aws_dynamodb_table.go
+++ b/aws/resource_aws_dynamodb_table.go
@@ -177,6 +177,12 @@ func resourceAwsDynamoDbTable() *schema.Resource {
 						},
 					},
 				},
+				Set: func(v interface{}) int {
+					var buf bytes.Buffer
+					m := v.(map[string]interface{})
+					buf.WriteString(fmt.Sprintf("%s-", m["name"].(string)))
+					return hashcode.String(buf.String())
+				},
 			},
 			"stream_enabled": {
 				Type:     schema.TypeBool,


bflad:
Maybe a new dynamodb index resource might be more composable for your situation? This would be similar to portions of S3 configuration which work better with separate resources rather than trying to make everything work in one.

mcwqy9:
I think I need an education on what sorts of things ought to be their own resource, and which ought to be an Set attribute of another resource. 
Is there a way to have only a few fields in that set's schema be used to identify uniqueness, while still allowing changes to all fields trigger an update? 

To ask an even broader question: Is it even supported to have arguments of type schema.TypeSet that are partially mutable?

mfinelli:
@mcwqy9 thanks for your work on this! any update on getting this merged in?

bflad:
Hi folks 👋 Unfortunately, due to the way that `TypeSet` attributes operate in the current Terraform Provider SDK, the `Set` hash must include all elements of the underlying arguments, otherwise Terraform cannot see differences in them to perform updates or force resource recreation. Rather than leaving an unactionable pull request open, I'm going to go ahead and close this. Please feel free to continue/track design discussions in the originating issue.

PR #1245, opened 2017-07-26
Added missing attribute in the API Gateway Resource documentation

Ninir:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/1229

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayResource_'                               
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayResource_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayResource_basic
--- PASS: TestAccAWSAPIGatewayResource_basic (15.40s)
=== RUN   TestAccAWSAPIGatewayResource_update
--- PASS: TestAccAWSAPIGatewayResource_update (46.89s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	62.323s
```

### DIFF
--- a/aws/resource_aws_api_gateway_resource.go
+++ b/aws/resource_aws_api_gateway_resource.go
@@ -60,9 +60,8 @@ func resourceAwsApiGatewayResourceCreate(d *schema.ResourceData, meta interface{
 	}
 
 	d.SetId(*resource.Id)
-	d.Set("path", resource.Path)
 
-	return nil
+	return resourceAwsApiGatewayResourceRead(d, meta)
 }
 
 func resourceAwsApiGatewayResourceRead(d *schema.ResourceData, meta interface{}) error {
@@ -76,6 +75,7 @@ func resourceAwsApiGatewayResourceRead(d *schema.ResourceData, meta interface{})
 
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFoundException" {
+			log.Printf("[WARN] API Gateway Resource (%s) not found, removing from state", d.Id())
 			d.SetId("")
 			return nil
 		}
--- a/website/docs/r/api_gateway_resource.html.markdown
+++ b/website/docs/r/api_gateway_resource.html.markdown
@@ -37,5 +37,7 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-* `id` - The ID of the API resource
-* `path` - The complete path for this API resource, including all parent paths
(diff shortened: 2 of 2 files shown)


aws/resource_aws_api_gateway_resource.go
```diff
@@ -76,6 +75,7 @@ func resourceAwsApiGatewayResourceRead(d *schema.ResourceData, meta interface{})
 
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFoundException" {
+			log.Printf("[WARN] API Gateway Resource (%s) not found, removing from state", d.Id())
```
radeksimko: 👍

website/docs/r/api_gateway_resource.html.markdown
```diff
@@ -37,5 +37,7 @@ The following arguments are supported:
 
 The following attributes are exported:
 
-* `id` - The ID of the API resource
-* `path` - The complete path for this API resource, including all parent paths
+* `id` - The resource's identifier.
+* `parent_id` - The parent resource's identifier.
+* `path` - The complete path for this API resource, including all parent paths.
+* `path_parth` - The last path segment for this resource.
```
radeksimko: This does no harm, but we tend to avoid documenting attributes which are exposed because they're also arguments. i.e. we don't double-document the same fields as arguments and attributes.

Instead we could just use different wording for the sentence above:

> In addition to all arguments above, the following attributes are exported:

aws/resource_aws_api_gateway_resource.go
```diff
@@ -60,9 +60,8 @@ func resourceAwsApiGatewayResourceCreate(d *schema.ResourceData, meta interface{
 	}
 
 	d.SetId(*resource.Id)
-	d.Set("path", resource.Path)
 
-	return nil
+	return resourceAwsApiGatewayResourceRead(d, meta)
```
radeksimko: Was there any specific reason for this change - any bug or misbehaviour you experienced or is this just to stick to conventions?

Ninir: Conventions mainly :)

website/docs/r/api_gateway_resource.html.markdown
Ninir: Agreed, will change, thanks!

Ninir: Changed!

Issue #188, opened 2017-06-13
Allow configuring CloudFront cache_behavior precedence


hashibot:
_This issue was originally opened by @randomeizer as hashicorp/terraform#7253. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

0.6.16
### Affected Resource(s)

Please list the resources as a list, for example:
- cloudfront_distribution > cache_behavior
### Terraform Configuration Files

`````` hcl
# Distro pointing at the specified domain.
resource "aws_cloudfront_distribution" "cloudfront" {
  origin {
    domain_name = "${var.lde_vm_domain_name}"
    origin_id = "lde_vm"
    custom_origin_config {
      http_port = 80
      https_port = 443
      origin_protocol_policy = "match-viewer"
      origin_ssl_protocols = [
        "SSLv3",
        "TLSv1"]
    }
    custom_header = {
      name = "${var.lde_custom_header_name}"
      value = "${var.lde_custom_header_value}"
    }
  }
  aliases = ["${var.public_domain_name}"]
  enabled = true
  comment = "CloudFront Server"
  default_cache_behavior {
    target_origin_id = "lde_vm"
    allowed_methods = [
      "DELETE",
      "GET",
      "HEAD",
      "OPTIONS",
      "PATCH",
      "POST",
      "PUT"]
    cached_methods = [
      "GET",
      "HEAD"]
    compress = true
    smooth_streaming = false
    forwarded_values {
      cookies {
        forward = "all"
      }
      headers = ["Host"]
      query_string = true
    }
    viewer_protocol_policy = "allow-all"
    min_ttl = 0
    default_ttl = 3600
    max_ttl = 31536000
  }

# private digital content cache
  cache_behavior {
    path_pattern = "user/digital_content/download/*"
    target_origin_id = "lde_vm"
    allowed_methods = [
      "GET",
      "HEAD",
      "OPTIONS"]
    cached_methods = [
      "GET",
      "HEAD"]
    compress = true
    smooth_streaming = false
    forwarded_values {
      cookies {
        forward = "none"
      }
      headers = ["Host"]
      query_string = true
    }
    trusted_signers = ["self"]
    viewer_protocol_policy = "allow-all"
    min_ttl = 0
    default_ttl = 3600
    max_ttl = 31536000
  }

  # JS files
  cache_behavior {
[...]

PR #1269, opened 2017-07-27
resource/aws_cloudfront_distribution: Add precendence for cache behaviors

acejam:
The goal of this PR is to address #188.

This PR changes `cache_behavior` of the `aws_cloudfront_distribution` resource from a `TypeSet` to a `TypeList`. By making this a list, we can set the precedence of cache behaviors in CloudFront.

Tested against AWS using `Terraform v0.10.0-dev` local build.

### DIFF
--- a/aws/resource_aws_cloudfront_distribution.go
+++ b/aws/resource_aws_cloudfront_distribution.go
@@ -35,18 +35,17 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 				Set:      aliasesHash,
 			},
 			"cache_behavior": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
-				Set:      cacheBehaviorHash,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"allowed_methods": {
-							Type:     schema.TypeList,
+							Type:     schema.TypeSet,
 							Required: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
 						"cached_methods": {
-							Type:     schema.TypeList,
+							Type:     schema.TypeSet,
 							Required: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
@@ -189,12 +188,12 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"allowed_methods": {
-							Type:     schema.TypeList,
+							Type:     schema.TypeSet,
 							Required: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
 						"cached_methods": {
-							Type:     schema.TypeList,
+							Type:     schema.TypeSet,
 							Required: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
--- a/aws/cloudfront_distribution_configuration_structure.go
+++ b/aws/cloudfront_distribution_configuration_structure.go
@@ -42,7 +42,7 @@ func (p StringPtrSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
 // Used by the aws_cloudfront_distribution Create and Update functions.
 func expandDistributionConfig(d *schema.ResourceData) *cloudfront.DistributionConfig {
 	distributionConfig := &cloudfront.DistributionConfig{
-		CacheBehaviors:       expandCacheBehaviors(d.Get("cache_behavior").(*schema.Set)),
+		CacheBehaviors:       expandCacheBehaviors(d.Get("cache_behavior").([]interface{})),
 		CustomErrorResponses: expandCustomErrorResponses(d.Get("custom_error_response").(*schema.Set)),
 		DefaultCacheBehavior: expandDefaultCacheBehavior(d.Get("default_cache_behavior").(*schema.Set).List()[0].(map[string]interface{})),
 		Enabled:              aws.Bool(d.Get("enabled").(bool)),
@@ -219,12 +219,12 @@ func defaultCacheBehaviorHash(v interface{}) int {
 		buf.WriteString(fmt.Sprintf("%d-", d.(int)))
 	}
 	if d, ok := m["allowed_methods"]; ok {
-		for _, e := range sortInterfaceSlice(d.([]interface{})) {
+		for _, e := range sortInterfaceSlice(d.(*schema.Set).List()) {
 			buf.WriteString(fmt.Sprintf("%s-", e.(string)))
 		}
 	}
 	if d, ok := m["cached_methods"]; ok {
-		for _, e := range sortInterfaceSlice(d.([]interface{})) {
+		for _, e := range sortInterfaceSlice(d.(*schema.Set).List()) {
 			buf.WriteString(fmt.Sprintf("%s-", e.(string)))
 		}
 	}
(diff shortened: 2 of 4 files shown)


aws/resource_aws_cloudfront_distribution.go
```diff
@@ -35,18 +35,17 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 				Set:      aliasesHash,
 			},
 			"cache_behavior": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
```
Ninir: There is also a Schema migration to create here, as for security groups:
- [Resource](https://github.com/terraform-providers/terraform-provider-aws/blob/b8d4e1570fc43a2acee6b6e47f63c9db6b067fa2/aws/resource_aws_security_group_rule.go#L26)
- [Related migration](https://github.com/terraform-providers/terraform-provider-aws/blob/b8d4e1570fc43a2acee6b6e47f63c9db6b067fa2/aws/resource_aws_security_group_rule_migrate.go)

This would allow to not break things for users, and make the change invisible :)

gazoakley:
Hi @acejam 

I've just run into this issue myself from having a CloudFormation configuration where behavior precedence matters. The changes seem to work fine for me (and I think the choice to use TypeList instead of TypeSet makes a lot of sense) - but I think it's an unavoidable breaking change:

- Previously Terraform would create a hash for each `cache_behavior` and write the distribution XML in the A-Z order of the hash
- Now Terraform takes the same configuration and uses the item order instead. When Terraform refreshes state from AWS, it's an unavoidably different order (but now it's determinable for the user by looking at the config)

I don't think a migration can fix this - instead this is a behavior (no pun intended!) change. There's a risk of breakages:

- Users that do rely on precedence that have got lucky so far and don't have `cache_behavior` resources in the correct order already may experience an issue if they don't notice a change in the plan
- Users that otherwise would rely on precedence are already impacted - they can't currently write a configuration that works
- Users that don't rely on precedence (i.e. they don't have overlapping `path_pattern`s) will see a change in the plan after upgrading, but it shouldn't cause them an issue even if they ignore the changes

I don't know if a warning can be shown specifically during a migration (haven't seen it in other migrations) but that might be doable. 

There have been breaking changes with this resource before: https://github.com/hashicorp/terraform/blob/master/CHANGELOG.md#096-may-25-2017

acejam:
@Ninir, could you review @gazoakley's comment above and confirm if this is indeed the case?

Ninir:
Hey folks, 

Sorry for the silence here, been really busy elsewhere! 😅 

> I've just run into this issue myself from having a CloudFormation configuration where behavior precedence matters. The changes seem to work fine for me (and I think the choice to use TypeList instead of TypeSet makes a lot of sense) - but I think it's an unavoidable breaking change:

That is why we could use a migration for that: avoid the Breaking Change. However, I'm wondering whether we can call the AWS API from a migration... it would help getting the correct order, update from hash to list, and then fix the order. It would then be invisible for users (if I don't omit anything).

So, if the call is not doable, then we would need to mark this work as breaking-change, and delay the merge for a major version :/ 

What do you folks think?

danielfm:
@Ninir  what about introducing a new required `priority` field to `cache_behavior`, instead of relying on the order of `cache_behavior` blocks?

This would be a breaking change in the way it would force the operator to fill those values manually, but it would be a safe change in the way it would not make assumptions regarding the order of such blocks.

Ninir:
@danielfm this is an excellent idea and a safe way to manage it in my opinion: just wondering how to properly do that from a dev point of view/ Not saying it's not doable, just want to have the best way to do that, and error / warning when needed :)

skyzyx:
I also have an issue where precedence matters. The sooner we can get support for this into Terraform, the less I'll need to deal with my engineering manager hassling me about it.

acejam:
I haven't forgotten about this - I will try and see if I can change this to match @danielfm's idea.

skyzyx:
Pinging again. As we're getting ready to roll out to production, we're hitting these errors multiple deploys per day in non-prod. I don't have any proficiency in Golang, otherwise I would absolutely jump in.

Is this something that @Ninir or someone else from Hashicorp can escalate? This patch is intended to fix a genuine bug in Terraform.

SmoshySmosh:
Echoing what @skyzyx is saying.

@Ninir any luck moving this forward?

Ninir:
Hi @SmoshySmosh, @skyzyx 

One could still handle this new `precedence` parameter and use a schema migration: In my opinion, I feel like this parameter should be flagged as required.

This is a breaking change, and thus should be delayed to 2.0.0. 

If one feels able to do it, please take over it, we would be happy to review.

danielfm:
I hacked together a change in which I added support for a `precedence` field for sorting cache behaviors, like this:

```tf
# Resource 1
resource "aws_cloudfront_distribution" "cloudfront-1" {
  # ...

  cache_behavior {
    precedence = 0
    # ...
  }

  cache_behavior {
    precedence = 1
    # ...
  }

  cache_behavior {
    precedence = 2
    # ...
  }
}
```

After playing around with this for a while, it turns out it has one major drawback which makes me believe the solution proposed in this PR is better.

The AWS SDK stores the cache behaviors in a list, thus the precedence is implicitly defined by how that list is ordered. This means that the precedence we specify in the Terraform configuration "gets lost" depending on the values we choose as the precedence for each cache behavior.

An example:

```tf
# Resource 2
resource "aws_cloudfront_distribution" "cloudfront-2" {
  # ...

  cache_behavior {
    precedence = 1
    # ...
  }

  cache_behavior {
    precedence = 2
    # ...
  }

  cache_behavior {
    precedence = 3
    # ...
  }
}
```

If you look closely, both resources `cloudfront-1` and `cloudfront-2` should be considered the same with respect to the precedence even though we have different precedence values (all shifted by one), and indeed, in my "solution", both resources are created respecting the precedence.

However, when running a `terraform plan` for the `cloudfront-2` resource, we are unable to "reconstruct" the expected precedence values from the data we get from the AWS SDK; the best we can do is to set the precedence after the index of each element (0, 1, and 2), but this confuses Terraform into thinking the actual distribution is different from the one we are expecting (when in fact it's not).

In case anyone wants to see, the code I've been playing around (that has the issue I'm describing) is available here:
https://github.com/danielfm/terraform-provider-aws/commit/36e4767f927d62f7f309b2f7887341af70a9f83b

leecookson:
I'm not sure if there's a policy for how terraform maps features to AWS API's, but in the API, the precedence is just a side-effect of the index within the "cache_behaviors" array, right?  I see (esp for very large distributions) validating and fixing overlapping precedence attributes as burdensome, when the natural order within the tf file already expresses the expected outcome.

FYI, I've tested the PR from Jul 28 and it seems to behave as expected.

A different breaking change that may imply the order is controlled in the source may be a `cache_behaviors` list that contains the individual cache_behavior objects, which would strongly imply that order is maintained.

```
cache_behaviors {
  cache_bahavior {
    ...
  }
  cache_bahavior {
    ...
  }
}
```

I prefer the July 28 solution since it doesn't require another field to manage (ugly, scattered commits when a low-numbered precedence is added) and validate.

acejam:
@Ninir Bump to this. This bug really needs to be fixed.

SmoshySmosh:
@ninir Giving this another bump. Need to get this merged.

Ninir:
Hi folks,

As exposed earlier, this work is a breaking change that cannot be merged before 2.0.0.
However, the team is thinking/working on an alternative solution that would allow us get this top-to-bottom listing without breaking things.

We will keep you up-to-date as soon as possible. Hope it makes sense.

Ninir:
Hey folks,

Just to let you know that we worked on an alternative solution that would help getting a fix earlier without breaking things, still deprecating the old and broken behavior.

#4117 implements a new attribute called `ordered_behavior_cache`, which implies that you replace your old and broken `cache_behavior` in your stacks to this new one.

It is the only solution we came up with. Thanks for the initial work @acejam! 🤖 🎆 🚀 

Tell me If we can be of any help on that. Closing!

PR #1285, opened 2017-07-29
Add Waf RateBasedRule

DDRBoxman:
https://docs.aws.amazon.com/waf/latest/APIReference/API_regional_RateBasedRule.html#WAF-Type-regional_RateBasedRule-MatchPredicates

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -478,6 +478,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_waf_byte_match_set":                       resourceAwsWafByteMatchSet(),
 			"aws_waf_ipset":                                resourceAwsWafIPSet(),
 			"aws_waf_rule":                                 resourceAwsWafRule(),
+			"aws_waf_rate_based_rule":                      resourceAwsWafRateBasedRule(),
 			"aws_waf_size_constraint_set":                  resourceAwsWafSizeConstraintSet(),
 			"aws_waf_web_acl":                              resourceAwsWafWebAcl(),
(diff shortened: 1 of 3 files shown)


bekbulatov:
Hi @DDRBoxman, have your changes been tested? The acceptance tests fail as you don't pass `RateKey` and `RateLimit` when you create a `RateBasedRule` object.

bekbulatov:
The next issue is that `aws_waf_web_acl` takes `rule_id` and doesn't know the type of rule. Here is the example from docs.

```
resource "aws_waf_web_acl" "waf_acl" {
  depends_on  = ["aws_waf_ipset.ipset", "aws_waf_rule.wafrule"]
  name        = "tfWebACL"
  metric_name = "tfWebACL"

  default_action {
    type = "ALLOW"
  }

  rules {
    action {
      type = "BLOCK"
    }

    priority = 1
    rule_id  = "${aws_waf_rule.wafrule.id}"
  }
}
```

Current implementation of the aws provider doesn't set the type and it always requests a REGULAR (as default) rule for WebACL and fails if it sets as a RATE_BASED rule.

I propose to add type for rules for WebACL like this:

```
rules {
  action {
    type = "BLOCK"
  }

  priority = 1
  rule_id  = "${aws_waf_rule.wafrule.id}"
  type = "RATE_BASED"
}
```

@radeksimko @grubernaut Are you happy with these changes?

bekbulatov:
I have updated this PR, tested and raised https://github.com/terraform-providers/terraform-provider-aws/pull/1606

PR #1295, opened 2017-08-01
New ssm parameters features

jocgir:
Hi,

I made some changes on the aws_ssm_parameter resource.

I added support for the new features (description, tags and allowed pattern).
I also fix some problems with the current implementation. If a parameter was manually deleted from the parameter store, that causes problems with terraform plan, apply and destroy. A resource that no longer exists must not be considered as an error.

I also changed the default value for overwrite. Forcing the value to be false by default is not very useful in a terraform project. If we wish to preserve the value, we should use the default lifecycle rules to protect manually changed value. Setting this new parameter to false by default causes breaking change in the already written code since this parameter did not exists before and the values were overwritten by default.


jocgir:
Sorry, I was on vacation. I mixup internal changes to make our fork buildable on the submitted branche and I am not able to cleanly reset the unwanted commit. I will close the branch and resubmit a new clean one. Sorry for the inconvenience.

PR #1303, opened 2017-08-02
Populate listener_arn for aws_alb_listener_rule

### LINKED ISSUE #779: aws_alb_listener_rule doesn't import listener arn
_This issue was originally opened by @dannyhw as hashicorp/terraform#14568. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

After successfully importing a listener rule I check the state and there is no listener arn even though this is required in the listener_rule config.

```
$ terraform state show aws_alb_listener_rule.CH-EU-PROD-MEDIAPOOL-LB-443-1
id                        = arn:aws:elasticloadbalancing:eu-central-...
action.#                  = 1
action.0.target_group_arn = arn:aws:elasticloadbalancing:eu-central-...
action.0.type             = forward
arn                       = arn:aws:elasticloadbalancing:eu-central-...
condition.#               = 1
condition.0.field         = path-pattern
condition.0.values.#      = 1
condition.0.values.0      = /mediapool/*
priority                  = 1
```
_I removed some of the information from the output however I kept all the fields that were returned._

I'm running Terraform v0.9.5 on ubuntu 14.04.

To reproduce just import an existing listener rule using 

` terraform import aws_alb_listener_rule.some_name some_arn`

sj26:
Importing a listener rule currently doesn't populate the listener_arn so the next plan attempts to recreate the rule. The AWS API doesn't return the listener arn when describing listener rules. But the listener arn can be inferred from the listener rule arn.

Closes #779

### DIFF
--- a/aws/resource_aws_alb_listener_rule.go
+++ b/aws/resource_aws_alb_listener_rule.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"regexp"
 	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -147,6 +148,10 @@ func resourceAwsAlbListenerRuleRead(d *schema.ResourceData, meta interface{}) er
 	rule := resp.Rules[0]
 
 	d.Set("arn", rule.RuleArn)
+
+	// The listener arn isn't in the response but can be derived from the rule arn
+	d.Set("listener_arn", albListenerARNFromRuleARN(*rule.RuleArn))
+
 	// Rules are evaluated in priority order, from the lowest value to the highest value. The default rule has the lowest priority.
 	if *rule.Priority == "default" {
 		d.Set("priority", 99999)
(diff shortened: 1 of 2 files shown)


aws/resource_aws_alb_listener_rule.go
```diff
@@ -147,6 +148,15 @@ func resourceAwsAlbListenerRuleRead(d *schema.ResourceData, meta interface{}) er
...
 
 	d.Set("arn", rule.RuleArn)
+
+	// The listener arn isn't in the response but can be derived from the rule arn:
+	// arn:aws:elasticloadbalancing:us-east-1:012345678912:listener-rule/app/name/0123456789abcdef/abcdef0123456789/456789abcedf1234
+	listenerArn := strings.Replace(*rule.RuleArn, ":listener-rule/", ":listener/", 1)
+	// arn:aws:elasticloadbalancing:us-east-1:012345678912:listener/app/name/0123456789abcdef/abcdef0123456789/456789abcedf1234
+	listenerArn = listenerArn[0:strings.LastIndex(listenerArn, "/")]
+	// arn:aws:elasticloadbalancing:us-east-1:012345678912:listener/app/name/0123456789abcdef/abcdef0123456789
+	d.Set("listener_arn", listenerArn)
```
radeksimko: This seems functionally ok, but do you mind decoupling this new logic into a function and adding a test for it?

sj26:
@radeksimko thanks for the feedback! Sorry, I was away for a few days. I've extracted that logic into a little function and added some testing modelled after the [alb target group cloudwatch suffix tests](https://github.com/terraform-providers/terraform-provider-aws/blob/4f9196eaf93734078a986decf859f8f995c25383/aws/resource_aws_alb_target_group_test.go#L17-L46). I also switched to using a regexp because it wasn't handling the empty case gracefully.

PR #1312, opened 2017-08-02
r/cloudtrail: Retry if IAM role isn't propagated yet

### LINKED ISSUE #693: CloudTrail resource creation failure when creating aws_iam_role and aws_iam_role_policy in single run
_This issue was originally opened by @tmclaugh as hashicorp/terraform#13631. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

There appears to be an issue with creating cloudtrails when the iam role for cloudwatch logs 
is created in the same run.  Terraform says the resource creation has completed but when creating the CloudTrail an InvalidCloudWatchLogsRoleArnException error occurs.  A subsequent run completes just fine.  This leads me to believe that the issue is with the IAM role policy change not being propagated across AWS fast enough for Terraform.

### Terraform Version
Terraform v0.9.3

### Affected Resource(s)
- aws_cloudtrail
- aws_iam_role_policy
- aws_iam_role

https://gist.github.com/0dd20ed6bc2a9dcd5c316c9d5193cdc2

### Debug Output
https://gist.github.com/690acefc73b6509d89e53f389981cdd0

### Expected Behavior
aws_cloudtrail is successful created

### Actual Behavior
It appears that the role policy is not properly propagated by AWS before attempting to create the cloudtrail resource.  A second run successfully creates wht aws_cloudtrail resource.
```
* aws_cloudtrail.ct: InvalidCloudWatchLogsRoleArnException: Access denied. Check the trust relationships for your role.
        status code: 400, request id: 4a209c1f-206b-11e7-bb0a-0d6cd06b0119
```

### Steps to Reproduce
1. `terraform apply`

radeksimko:
Closes #693

The first error code (`InvalidCloudWatchLogsRoleArnException`) comes from the bug report, second one (`InvalidCloudWatchLogsLogGroupArnException`) comes from testing of https://github.com/terraform-providers/terraform-provider-aws/pull/1357.

### DIFF
--- a/aws/resource_aws_cloudtrail.go
+++ b/aws/resource_aws_cloudtrail.go
@@ -3,9 +3,11 @@ package aws
 import (
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudtrail"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -116,7 +118,21 @@ func resourceAwsCloudTrailCreate(d *schema.ResourceData, meta interface{}) error
 		input.SnsTopicName = aws.String(v.(string))
 	}
 
-	t, err := conn.CreateTrail(&input)
+	var t *cloudtrail.CreateTrailOutput
+	err := resource.Retry(15*time.Second, func() *resource.RetryError {
+		var err error
+		t, err = conn.CreateTrail(&input)
+		if err != nil {
+			if isAWSErr(err, cloudtrail.ErrCodeInvalidCloudWatchLogsRoleArnException, "Access denied.") {
+				return resource.RetryableError(err)
+			}
+			if isAWSErr(err, cloudtrail.ErrCodeInvalidCloudWatchLogsLogGroupArnException, "Access denied.") {
+				return resource.RetryableError(err)
+			}
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


aws/resource_aws_cloudtrail.go
```diff
@@ -116,7 +118,21 @@ func resourceAwsCloudTrailCreate(d *schema.ResourceData, meta interface{}) error
...
-	t, err := conn.CreateTrail(&input)
+	var t *cloudtrail.CreateTrailOutput
+	err := resource.Retry(15*time.Second, func() *resource.RetryError {
+		var err error
+		t, err = conn.CreateTrail(&input)
+		if err != nil {
+			if isAWSErr(err, "InvalidCloudWatchLogsRoleArnException", "Access denied.") {
+				return resource.RetryableError(err)
+			}
+			if isAWSErr(err, "InvalidCloudWatchLogsLogGroupArnException", "Access denied.") {
```
Ninir: What do you think of having a single if, checking for both conditions?

aws/resource_aws_cloudtrail.go
```diff
@@ -116,7 +118,21 @@ func resourceAwsCloudTrailCreate(d *schema.ResourceData, meta interface{}) error
 		input.SnsTopicName = aws.String(v.(string))
 	}
 
-	t, err := conn.CreateTrail(&input)
+	var t *cloudtrail.CreateTrailOutput
+	err := resource.Retry(15*time.Second, func() *resource.RetryError {
+		var err error
+		t, err = conn.CreateTrail(&input)
+		if err != nil {
+			if isAWSErr(err, "InvalidCloudWatchLogsRoleArnException", "Access denied.") {
```
Ninir: I would use the `cloudtail.ErrCodeInvalidCloudWatchLogsRoleArnException` constant, just for sanity and typo-less issues. Thoughts?

aws/resource_aws_cloudtrail.go
Ninir: Also, would also use `ErrCodeInvalidCloudWatchLogsLogGroupArnException` here.

aws/resource_aws_cloudtrail.go
```diff
@@ -250,7 +266,21 @@ func resourceAwsCloudTrailUpdate(d *schema.ResourceData, meta interface{}) error
...
-	t, err := conn.UpdateTrail(&input)
+	var t *cloudtrail.UpdateTrailOutput
+	err := resource.Retry(15*time.Second, func() *resource.RetryError {
+		var err error
+		t, err = conn.UpdateTrail(&input)
+		if err != nil {
+			if isAWSErr(err, "InvalidCloudWatchLogsRoleArnException", "Access denied.") {
+				return resource.RetryableError(err)
+			}
+			if isAWSErr(err, "InvalidCloudWatchLogsLogGroupArnException", "Access denied.") {
```
Ninir: Same here for constants :)

aws/resource_aws_cloudtrail.go
radeksimko: It's 17 extra characters! :trollface: but trolling aside - yes, constant makes sense in this context and it may help avoiding typos. I'll change it.

aws/resource_aws_cloudtrail.go
radeksimko: I think the two if-s look tiny little bit cleaner (easier to read) compared to 1 if with 2 conditionals which would need to be line-wrapped (b/c long lines aren't easy to read either).

radeksimko: Agreed about constant though. I'll change that.

radeksimko:
@Ninir addressed comments regarding constants. 😉

Ninir:
👍

PR #1339, opened 2017-08-03
Kinesis concurrent CreateStream backoff

joshmyers:
See: http://docs.aws.amazon.com/kinesis/latest/APIReference/API_CreateStream.html

Similar to DynamoDB, only up to 5 kinesis streams are able to be in the 
CREATING state simultaneously otherwise a LimitExceededException is
thrown as can be seen below:

```
	* aws_kinesis_stream.test_stream.6: [WARN] Error creating Kinesis Stream: "This request would have exceed the limit on the number  of streams that can simultaneously be in CREATING or DELETING for the account XXXX. Limit: 5.", code: "LimitExceededException"
		* aws_kinesis_stream.test_stream[1]: 1 error(s) occurred
```

We create about 20 streams per environment (read: single Terraform run) 
and had been seeing these errors frequently, having to do several 
`terraform apply` to get a clean run. After speaking with AWS support, 
this is a hard limit (5) and they have no way of tuning per account.

### DIFF
--- a/aws/resource_aws_kinesis_stream.go
+++ b/aws/resource_aws_kinesis_stream.go
@@ -111,6 +111,11 @@ func resourceAwsKinesisStreamCreate(d *schema.ResourceData, meta interface{}) er
 		if isAWSErr(err, "LimitExceededException", "simultaneously be in CREATING or DELETING") {
 			return resource.RetryableError(err)
 		}
+		// AWS (un)helpfully raises LimitExceededException
+		// rather than ThrottlingException here
+		if isAWSErr(err, "LimitExceededException", "Rate exceeded for stream") {
+			return resource.RetryableError(err)
+		}
 		return resource.NonRetryableError(err)
 	})
 
--- a/aws/resource_aws_kinesis_stream.go
+++ b/aws/resource_aws_kinesis_stream.go
@@ -106,14 +106,19 @@ func resourceAwsKinesisStreamCreate(d *schema.ResourceData, meta interface{}) er
 		StreamName: aws.String(sn),
 	}
 
-	_, err := conn.CreateStream(createOpts)
-	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
-			return fmt.Errorf("[WARN] Error creating Kinesis Stream: \"%s\", code: \"%s\"", awsErr.Message(), awsErr.Code())
+	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
+		_, err := conn.CreateStream(createOpts)
+		if isAWSErr(err, "LimitExceededException", "simultaneously be in CREATING or DELETING") {
+			return resource.RetryableError(err)
 		}
-		return err
+		return resource.NonRetryableError(err)
+	})
+
+	if err != nil {
+		return fmt.Errorf("Unable to create stream: %s", err)
 	}
 
+	// No error, wait for ACTIVE state
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"CREATING"},
 		Target:     []string{"ACTIVE"},
--- a/aws/resource_aws_kinesis_stream_test.go
+++ b/aws/resource_aws_kinesis_stream_test.go
@@ -36,6 +36,45 @@ func TestAccAWSKinesisStream_basic(t *testing.T) {
 	})
 }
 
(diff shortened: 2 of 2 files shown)


joshmyers:
Rebased.

joshmyers:
@radeksimko any updates on getting this merged?

joshmyers:
Bump.

aws/resource_aws_kinesis_stream.go
```diff
@@ -12,6 +13,12 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
+// Number of attempts to create the Kinesis streams
+const KINESIS_MAX_ATTEMPTS = 10
+
+// How long to sleep if a limit-exceeded event happens
+const KINESIS_LIMIT_EXCEEDED_SLEEP = 10 * time.Second
+
```
radeksimko: Probably a nitpick, but these constants are only ever going to be used in this file in a single place, so we may as well just hard code those values and avoid the extra 5 LOC.

aws/resource_aws_kinesis_stream.go
```diff
@@ -85,36 +92,54 @@ func resourceAwsKinesisStreamCreate(d *schema.ResourceData, meta interface{}) er
...
 
-	_, err := conn.CreateStream(createOpts)
-	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok {
-			return fmt.Errorf("[WARN] Error creating Kinesis Stream: \"%s\", code: \"%s\"", awsErr.Message(), awsErr.Code())
-		}
-		return err
-	}
+	attemptCount := 1
+	for attemptCount <= KINESIS_MAX_ATTEMPTS {
```
radeksimko: Instead of (re)building a complex retry logic, counting attempts and time elapsed I think we should use the helper from [`helper/resource`](https://github.com/hashicorp/terraform/blob/master/helper/resource/wait.go#L10)

i.e. 
```go
err := resource.Retry(30 * time.Second, func() *resource.RetryableError {
    _, err := conn.CreateStream(createOpts)
    if isAWSErr(err, "LimitExceededException", "simultaneously be in CREATING or DELETING") {
    }
})
```

so we could reduce ~20 LOC to (IMO) more readable ~7 lines.

aws/resource_aws_kinesis_stream.go
```diff
@@ -85,36 +92,54 @@ func resourceAwsKinesisStreamCreate(d *schema.ResourceData, meta interface{}) er
...
+						attemptCount += 1
+					}
+				default:
+					return fmt.Errorf("[WARN] Error creating Kinesis Stream: \"%s\", code: \"%s\"", awsErr.Message(), awsErr.Code())
+				}
+			} else {
+				// Non-AWS exception occurred, give up
+				return fmt.Errorf("Error creating Kinesis stream: %s", err)
+			}
+		} else {
```
radeksimko: Since we `return` on errors above I don't think the else block is necessary in this context. It can be just below any retry logic, just assuming we'll only ever get to it if creation was successful -> less nesting => easier to read 🙂

joshmyers:
Hey @radeksimko no worries, I know you are busy. Comments addressed, hopefully good to go. Thanks!

aws/resource_aws_kinesis_stream.go
```diff
@@ -85,14 +85,15 @@ func resourceAwsKinesisStreamCreate(d *schema.ResourceData, meta interface{}) er
...
-			return fmt.Errorf("[WARN] Error creating Kinesis Stream: \"%s\", code: \"%s\"", awsErr.Message(), awsErr.Code())
+	err := resource.Retry(30*time.Second, func() *resource.RetryError {
+		_, err := conn.CreateStream(createOpts)
+		if isAWSErr(err, "LimitExceededException", "simultaneously be in CREATING or DELETING") {
+			return resource.RetryableError(err)
 		}
-		return err
-	}
+		return resource.NonRetryableError(err)
+	})
```
radeksimko: Actually do you mind adding an error check here?

radeksimko: Then I'm happy to click the merge button 😃

joshmyers: @radeksimko done :)

joshmyers:
@radeksimko Hopefully sorted

Issue #1232, opened 2017-07-25
add support tags for aws_spot_fleet_request


hashibot:
_This issue was originally opened by @jefimm as hashicorp/terraform#15632. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

https://aws.amazon.com/about-aws/whats-new/2017/07/tag-your-spot-fleet-ec2-instances/

s-maj:
I was just looking how to extend Terraform to support this (and it should be fairly easy imho) but first, aws-sdk has to be bumped to [1.10.15](https://github.com/aws/aws-sdk-go/pull/1416).

jefimm:
Please be advised that there is ec2:CreateTags permission that is required to implement tags
Fleet role needs to include it and if default role is used it needs to be aws-ec2-spot-fleet-tagging-role and aws-ec2-spot-fleet-role

stack72:
Depends on https://github.com/terraform-providers/terraform-provider-aws/pull/1235

pdolan:
I don't believe there should be any further blockers on this? Would like to add my name to the hat for people who are actively looking forward to this feature!

jekh:
I just submitted a PR for this: #2042

PR #1366, opened 2017-08-08
add support for tags on spot fleet requests

taylorsmithgg:
Solves #1232


aws/resource_aws_spot_fleet_request.go
```diff
@@ -256,6 +256,30 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
 							Computed: true,
 							ForceNew: true,
 						},
+						"tag": {
```
catsby: I'm curious if you were able to examine other resource's support of EC2 tags before coming to this implementation? Most resources use our helper file aws/tags.go:

https://github.com/terraform-providers/terraform-provider-aws/blob/494221fd522c1d41c47790e1a04ce78fcd1c3ab6/aws/tags.go#L20-L25

Example usage from `aws_instance`:

https://github.com/terraform-providers/terraform-provider-aws/blob/494221fd522c1d41c47790e1a04ce78fcd1c3ab6/aws/resource_aws_instance.go#L262-L264

aws/resource_aws_spot_fleet_request.go
```diff
@@ -256,6 +256,30 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
...
+									"value": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+								},
+							},
+							Set: spotFleetTagToHash,
+						},
+
+						"tags": {
```
catsby: How does this attribute work, and how does it differ from `tag` set above?  For instance, in the `aws_instance` resource, you see this for configuration:

```
tags {
  Name = "HelloWorld"
}
```

Can you elaborate on how this `tags` attribute is used?

aws/resource_aws_spot_fleet_request.go
```diff
@@ -362,6 +455,28 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
...
+
+	if v, ok := d["tags"]; ok && v != "" {
+		tags, _ := spotFleetTagsFromList(v.([]interface{}))
+		opts.TagSpecifications = []*ec2.SpotFleetTagSpecification{
+			{
+				ResourceType: aws.String(ec2.ResourceTypeInstance),
+				Tags:         tags,
+			},
+		}
+	}
```
catsby: It seems by the implementation above that a user cannot specify both `tags` and `tag`. By the looks of it, `tags` would take predecease if both are supplied , is the the intent? If so, they should use the helper/schema `ConflictsWith` attribute. An example of this usage can be found in `aws_instance`:

https://github.com/terraform-providers/terraform-provider-aws/blob/494221fd522c1d41c47790e1a04ce78fcd1c3ab6/aws/resource_aws_instance.go#L106-L110

taylorsmithgg: This was originally in the reference implementation, but caused some minor issues that I didn't have capacity to work on at the time. I intend to add it back in.

aws/resource_aws_spot_fleet_request.go
taylorsmithgg: The class for aws-sdk is `SpotFleetTagSpecification` with uses `key` and `value`. I just attempted to follow along with the aws model without making too many modifications.

```
tags = [
  {
    key = "key"
    value = "${var.name}"
  }
]
```

aws/resource_aws_spot_fleet_request.go
taylorsmithgg: I previewed `aws_instance` but with different specs, opted to go with the model used here: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_autoscaling_group.go

moee:
First of all thanks for this PR, I'm looking very forward to be able to use this.

The PR now proposes this syntax:

```
tags = [
  {
    key = "key"
    value = "${var.name}"
  }
]
```

Is there a specific reason why it doesn't follow the [SpotFleetLaunchSpecification](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotFleetLaunchSpecification.html) naming from the AWS API  and use the [SpotFleetTagSpecification](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotFleetTagSpecification.html) data structure? In this case, the tags would we wrapped around `tag_specification` 

```
tag_specifications = [
  {
    resource_type = "instance"
    tags = [
      {
        key = "key"
        value = "${var.name}"
      }
    ]
  }
]
```

It adds some verbosity (considering that at the moment the only resource type that is supported is `instance`), but reading the API documentation, my first guess would have been to use `tag_specifications` in terraform. Also, if at some point more resource types are supported, you can provide this functionality without breaking compatibility.

aaroncline:
Please revive this feature!

nullck:
Please revive this feature[2]

taylorsmithgg:
I'd be happy to, I'll make the requested changes in the next week. Sorry for the delays!

davidfic:
Any update on this PR getting merged?

Ninir:
Hy @taylorsmithgg 

Did you find time to continu the work? :)
We would love to this merged! 🚀

galindro:
@taylorsmithgg any return??

sbv-siteimprove:
Is  this the same as https://github.com/terraform-providers/terraform-provider-aws/pull/2042 ?

taylorsmithgg:
Apologies, I've been traveling a ton and haven't had time. Looks like #2042 wasn't working for some folks. I'll take a look now and see if I can get it in for you guys!

seandtaber:
@taylorsmithgg Any update here? I am running terraform-provider-aws version 1.8.0 and see an error when trying to add tags to the spot fleet saying it is an invalid or unknown key.

bflad:
This pull request predates my maintainership by quite a bit so I'm really not sure the history here, but it appears this support has already been handled by #2042 and the related issue #1232 is closed. I'm going to close this pull request based on that to clean up the open PRs in the repository, but if I'm missing something, please let me know.

PR #1368, opened 2017-08-08
Removed unused deps from vendor

radeksimko:
Basically I just reverted https://github.com/terraform-providers/terraform-provider-aws/pull/1362, added `github.com/hashicorp/terraform/backend` to the vendor ignore list and then reran

```
govendor fetch github.com/hashicorp/terraform/...@v0.10.0
```


catsby:
Needs rebase I think, other than that, 👍

PR #1371, opened 2017-08-09
Change documented bucket name to use hyphens instead of underscore

### LINKED ISSUE #1370: Docs - aws_s3_bucket name example uses underscores , hence fails to create
### Terraform Version
Terraform v0.10.0

### Affected Resource(s)
- aws_s3_bucket

From the existing docs:

```hcl
resource "aws_s3_bucket" "b" {
  bucket = "my_tf_test_bucket"
  acl    = "private"

  tags {
    Name        = "My bucket"
    Environment = "Dev"
  }
}
```

### Expected Behavior
The documented example should be syntactically correct so that further changes based on it should result in working resources except when it fails due to non-syntax related failures (example failure due to resource conflict which will happen if anyone creates the s3 bucket as mentioned in the example, no one else will be able to create/repeat the same).

### Actual Behavior
Error applying plan:

1 error(s) occurred:

* aws_s3_bucket.b: 1 error(s) occurred:

* aws_s3_bucket.b: Error validating S3 bucket name: only lowercase alphanumeric characters and hyphens allowed in "my_tf_test_bucket"

### Steps to Reproduce

 `terraform apply`

anshprat:
Change the bucket name to use hyphens instead of underscore.

	modified:   website/docs/r/s3_bucket.html.markdown

Fixes #1370

### DIFF
--- a/website/docs/r/s3_bucket.html.markdown
+++ b/website/docs/r/s3_bucket.html.markdown
@@ -16,7 +16,7 @@ Provides a S3 bucket resource.
 
 ```hcl
 resource "aws_s3_bucket" "b" {
-  bucket = "my_tf_test_bucket"
+  bucket = "my-tf-test-bucket"
   acl    = "private"
 
   tags {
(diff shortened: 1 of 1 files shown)


Ninir:
Hey @anshprat,

This looks good to me.

Thanks!

PR #1375, opened 2017-08-09
r/emr_cluster: enable configuration of ebs root volume size

rolandjohann:
EBS root volume size configuration implementation


Ninir:
Hey @rmmjohann 

This seems very good for a 1st-time contribution! 👍 

Could you also add an acceptance test that set the newly-added attribute, update it & check for the attribute existence?

If you need any help on that, feel free to ask!

rolandjohann:
I'll do this within the next days

rolandjohann:
@Ninir done

rolandjohann:
Sure, here are the docs: https://docs.aws.amazon.com/sdk-for-go/api/service/emr/#RunJobFlowInput.
Just saw that root volume size is supported by EMR 4.x upwards only. Will check that tomorrow.

Your test output doesn't contain the test for the root volume size. Is this the full output?

Yes, docs are crutial - didn't had them in mind. I'll fix them tomorrow as well.

Ninir:
> Your test output doesn't contain the test for the root volume size. Is this the full output?

It seems the output has been cut, but the test do passed :)

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSEMRCluster_root_volume_size'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEMRCluster_root_volume_size -timeout 120m
=== RUN   TestAccAWSEMRCluster_root_volume_size
--- PASS: TestAccAWSEMRCluster_root_volume_size (1530.05s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1530.085s
```

> Sure, here are the docs: https://docs.aws.amazon.com/sdk-for-go/api/service/emr/#RunJobFlowInput.
Just saw that root volume size is supported by EMR 4.x upwards only. Will check that tomorrow.

Yes, this was my concern when checking the docs: can't find anything about this particular attribute in the API / CF but in the CLI & SDK. 

> Yes, docs are crutial - didn't had them in mind. I'll fix them tomorrow as well.

Thank your for that! :)

rolandjohann:
Strange, that there is no reference at the API and CloudFormation docs.

The terraform docs are now updated.

Setting `ebs_root_volume_size` when using EMR version below 4.x is a misconfiguration, and shouldn't be handled by terraform, IMO. Are there any conventions at terraform how to handle those errors?

rolandjohann:
@Ninir is there anything left to do?

Issue #517, opened 2017-06-13
DynamoDB item creation using a aws_dynamodb_table_item resource


hashibot:
_This issue was originally opened by @drieselliott as hashicorp/terraform#11773. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,

It would be nice to have a 'aws_dynamodb_table_item' resource that allows you to write configuration entries to DynamoDB tables when provisioning an environment. It should support both the DynamoDB JSON and the normal JSON format.

```hcl
resource "aws_dynamodb_table_item" "configuration_item" {
    table = "${aws_dynamodb_table.configuration_table.name}"
    format = "json"
    item = <<EOF
       //Some JSON content
EOF
}
```

table: The table name
format: Either 'json' or 'dynamodb' for 'normal json' or 'dynamodb json'
item: The item to insert into the table (either as inline content or template)

This is very useful in serverless applications that use DynamoDb as their configuration store and Terraform to provision everything. Otherwise you need to rely on a separate non-terraform script to upload your configuration entries.

Thanks for considering this.

Dries

galadriel2143:
Started a PR into implement this feature. Mostly works but there are a few questions I would have before it gets merged back in.

Behavior on update: Should it force attributes back to the ones specified in the resource definition of they don't match? I think this would be fine based on how terraform handles attributes on other resources.
Behavior on externally removed item: Recreate, leave, alone, or hard fail? Recreating when it's missing seems to fall in line with other resources.

Output values: Currently I'm returning the range attribute value, hash attribute value, and the "query key", which contains both and can be used in a DDB query to get the object if necessary. Is there anything else that should be returned, or maybe we shouldn't return these at all...

Query setup: I have the range_key and hash_key names specified on the resource. These probably could be determined by querying the table metadata, but that would add more HTTP calls, and specifying them on the resource theoretically offers some protection from the key schema changing in an unexpected way.

jantman:
Is there any movement on this? This would be an extremely useful feature for us...

ccgagnon:
+1 It would be very useful feature for us too.

galadriel2143:
@jantman @ccagnon Are either of you familiar with Terraform's test suite, and/or want to take a crack at it? Do you see anything that needs to be improved with my code? Please let me know.

galadriel2143:
@ccgagnon

[...]

PR #1378, opened 2017-08-09
New Resource: aws_dynamodb_table_item

galadriel2143:
#517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ BUG FIXES:
 * resource/aws_db_parameter_group: fix permanent diff when specifying parameters with database-default values [GH-3182]
 * resource/aws_route53_record: Fix dualstack alias name regression trimming too many characters [GH-3187]
 * resource/aws_ses_template: Send only specified attributes for update [GH-3214]
-* resource/aws_dynamodb_table: Allow disabling stream with empty `stream_view_type` [GH-3197]
+* resource/aws_dynamodb_table: Allow disabling stream with empty `stream_view_type` [GH-3197] [GH-3224]
 * resource/aws_dx_connection_association: Retry disassociation [GH-3212]
 
 ## 1.8.0 (January 29, 2018)


jniesen:
Is there any timeframe for when this might get reviewed? My team would love to use this feature.

lorengordon:
Very interested in this feature also!

adamdiy:
Would be useful here as well...

joeharrison714:
This would be helpful for me

OliverCole:
I don't know what the rules are, but I suspect Hashicorp won't want to merge this without docs. I had a go at doing it myself, but actually I don't fully understand it, or have a golang development environment.

https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/dynamodb_table.html.markdown would be a good starting point.

radeksimko:
Hi @galadriel2143

firstly sorry for the long delay in reviewing your PR.

I took some time to think about this resource and also discuss internally
the challenges and benefits which may arise from supporting it.

I think it's :ok_hand: to support the use case of storing
some simple configuration data as we already allow folks to do via
[`aws_s3_bucket_object`](https://www.terraform.io/docs/providers/aws/r/s3_bucket_object.html)

However there are much better ways to **manage application configuration** (e.g. feature flags, connection strings etc.).
Tools like Consul, etcd, Vault, Chef/Puppet/Ansible/Salt or service like [AWS SSM](https://www.terraform.io/docs/providers/aws/r/ssm_parameter.html)
are all _designed_ to handle this problem.

I'm fine with enabling people who know the limitations, risks, trade-offs,
and reasons for using DynamoDB in this context. I also feel
the responsibility to prompt others to question the suitability though.
For that reason we should have a clear explanation of the intended
use case in the docs.

I'd also discourage people from using this resource for regular data
**backup**, basically for managing anything beyond 1-5 items, because
(again) there are much better ways to do this. e.g. scheduling
regular backups via CloudWatch Events & Lambda & [DynamoDB API](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateBackup.html).

**Restoration** is something Terraform could theoretically support
in a similar way [we support restoration in RDS](https://www.terraform.io/docs/providers/aws/r/db_instance.html#snapshot_identifier). i.e. restoring
from a given backup at the time of creating `aws_dynamodb_table`.

I hope I didn't miss any valid or invalid use case here.
Do let me know if I did.

----

So I'm happy to pull this resource in, but the PR will require some significant changes.
It's been sitting here for some time and my reservations are related mostly to code
which was inspired by the existing `aws_dynamodb_table` resource
which I refactored recently in https://github.com/terraform-providers/terraform-provider-aws/pull/3136.

For that reason I will breach the etiquette a bit (usually I'd review & let you fix)
and make those changes for you and push back to your branch.

I hope it's understandable and you don't mind me making those changes.

Thanks.

radeksimko:
@galadriel2143 I tried pushing back to your branch, but somehow ended up closing the PR. I assume it has something to do with the fact that the PR was opened from `master`. It's generally advisable to always open PRs from feature branches for the future. 😉

It not only makes it easier for reviewers, but also the author as they can keep rebasing the branch and work keep multiple PRs opened without having to force-push into `master`.

However #3238 was opened with your original commit and my changes.

bflad:

PR #1386, opened 2017-08-10
Log error when failing to delete log subscription filter.

paddycarver:
We were swallowing the error being returned when a
cloudwatch_log_subscription_filter resource's Delete method failed. This
just includes the error in the log message.


radeksimko:
btw. vast majority of AWS error types have meaningful implementations of string-based versions of errors, so `%s` may be just sufficient here, unless you have some other intentions.

paddycarver:
> Good point - just wondering how did you find this? 😃

Debugging an issue where the destroy failed, and discovered the error text wasn't incredibly helpful :)

> btw. vast majority of AWS error types have meaningful implementations of string-based versions of errors, so %s may be just sufficient here, unless you have some other intentions.

Oh, good to know, thanks! I just use `%+v` by habit now, as a lot of Go errors are.... not meaningful implementations.

PR #1400, opened 2017-08-13
Create default VPC for region if it has been deleted

ewbankkit:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/1289.
After some thinking about this I have gone with just creating the default VPC for a region if it doesn't exist.
This seems a more natural fit that having a `create_if_missing` attribute.
Documentation updated accordingly.
Acceptance tests:
```
make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSDefaultVpc_'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSDefaultVpc_ -timeout 120m
=== RUN   TestAccAWSDefaultVpc_basic
--- PASS: TestAccAWSDefaultVpc_basic (22.85s)
=== RUN   TestAccAWSDefaultVpc_createNew
--- PASS: TestAccAWSDefaultVpc_createNew (39.87s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	62.736s
```


aws/resource_aws_default_vpc.go
```diff
@@ -45,16 +45,27 @@ func resourceAwsDefaultVpcCreate(d *schema.ResourceData, meta interface{}) error
...
 	}
+	if describeResp.Vpcs == nil || len(describeResp.Vpcs) == 0 {
+		log.Printf("[INFO] No default VPC found in this region")
+	} else {
+		vpcId = aws.StringValue(describeResp.Vpcs[0].VpcId)
+	}
 
-	if resp.Vpcs == nil || len(resp.Vpcs) == 0 {
-		return fmt.Errorf("No default VPC found in this region.")
+	if vpcId == "" {
```
stack72: Why do this here rather than inside the if statement in line 53?

website/docs/r/default_vpc.html.markdown
```diff
@@ -15,10 +15,6 @@ For AWS accounts created after 2013-12-04, each region comes with a Default VPC.
 **This is an advanced resource**, and has special caveats to be aware of when
 using it. Please read this document in its entirety before using this resource.
 
-The `aws_default_vpc` behaves differently from normal resources, in that
```
stack72: I believe it's worth adding a note to say that it now creates it if it didn't exist prior

radeksimko:
> now sure this really is a breaking change - thoughts on why it is?

I'm ambivalent about this, but thought someone might be using the existing resource as a data source and not expecting new VPC to appear - but you could say the opposite too, I guess. 🤷‍♂️ 

I was expecting to see deletion of the VPC also being implemented (which is IMO right thing to do), in which case it would become "more dangerous" in this context.

Maybe it's not a BC but it's definitely worth mentioning under "BREAKING CHANGES / NOTES" section of changelog.

stack72:
totally agree on the BC/NOTES section - i was curious rather than saying it was incorrect :)

website/docs/r/default_vpc.html.markdown
ewbankkit: Will do.

ewbankkit:
@radeksimko I tend to agree about the default VPC deletion now that we have creation - That definitely would be a breaking change and/or a new attribute. Thoughts on implementing that also?

radeksimko:
> Thoughts on implementing that also?

I believe it should be implemented - the rule of thumb is that if Terraform creates something it should also be able to delete it (where API allows it) - i.e. own the whole lifecycle.

tburt11:
Whoa..   You are forcing a default VPC on Terraform users?  We specifically do not want a Default VPC in any of our 6 AWS Sub Accounts.  Please tell me that I am wrong that I will now have a Default VPC where I have previously intentionally deleted them.

stack72:
Hi @tburt11 

This isn't the case - this is only if people actually use the `aws_default_vpc` resource

Is this not what you expect?

Paul

tburt11:
@stack72   Thank you for the clarification.  As we expand to more regions, it would be nice to be able to delete that troublesome default VPC.  Troublesome because a former engineer made our prod subnets 172.31.0.0 and our other sub-accounts must be able to route to it.  A default VPC becomes a problem in this case.  Please give us the ability to delete the Default VPC.

radeksimko:
@tburt11 As Paul already mentioned unless you're using `aws_default_vpc` **resource**  the behaviour won't change.

I understand the need for a way to delete default VPC(s) as I previously worked for a company where we had to do this as part of initial provisioning of a each new AWS account. The CIDR reason is a good one, but the other reason for us was security as default VPC might not have the desired "secure" default settings (e.g. has IGW & subnets assign public IPs by default).

So I can feel your pain - with all that said I'm not sure Terraform is the right tool for this job. If you wish to discuss more concrete proposals on how this could be implemented with the respect to how resources work (i.e. w/out confusing users by creating a resource which only _deletes_ stuff), feel free to [open a fresh issue](https://github.com/terraform-providers/terraform-provider-aws/issues/new) as relevant ideas and conversation will be lost here as it's off-topic.

Thanks.

aws/resource_aws_default_vpc.go
ewbankkit: Changed as suggested.

website/docs/r/default_vpc.html.markdown
ewbankkit: Changed as suggested.

ewbankkit:
Made changes for the 2 review comments and rebased.

ewbankkit:
WRT deletion, we don't allow Terraform to delete some of the other "default" resources that are deletable via the API/Console if certain conditions are met (e.g. default route table not associated with any subnets):
* [aws_default_route_table](https://www.terraform.io/docs/providers/aws/r/default_route_table.html)
* [aws_default_subnet](https://www.terraform.io/docs/providers/aws/r/default_subnet.html)
* [aws_default_vpc_dhcp_options](https://www.terraform.io/docs/providers/aws/r/default_vpc_dhcp_options.html)

so at this point it seems more consistent to continue the non-deletion of default VPCs.
Thoughts?

ewbankkit:
**N.B.** This adds a requirement on the IAM `ec2:CreateDefaultVpc` action.

PR #1411, opened 2017-08-14
Db instance html markdown list re-order and cleanup

coingraham:
My first pull request.  Would like to update the order of the items in the list on this page to be alphabetical.  I've cleaned up the extraneous spacing and line feeds.  Added some language so it's more consistent and added punctuation to lines that were missing it.

### DIFF
--- a/website/docs/r/db_instance.html.markdown
+++ b/website/docs/r/db_instance.html.markdown
@@ -88,8 +88,6 @@ The following arguments are supported:
 
 The following attributes are exported:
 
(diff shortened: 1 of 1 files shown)


catsby:
https://github.com/terraform-providers/terraform-provider-aws/commit/2172a930, if anyone is interested

coingraham:
@catsby I believe there's an error in line 202 now.  "Timeouts" is not displaying correctly as a heading on the page.

PR #1420, opened 2017-08-15
resource/aws_efs_file_system: Add support for encryption

pjfbashton:
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSEFSFileSystem_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEFSFileSystem_ -timeout 120m
=== RUN   TestAccAWSEFSFileSystem_importBasic
--- PASS: TestAccAWSEFSFileSystem_importBasic (50.69s)
=== RUN   TestAccAWSEFSFileSystem_basic
--- PASS: TestAccAWSEFSFileSystem_basic (113.40s)
=== RUN   TestAccAWSEFSFileSystem_pagedTags
--- PASS: TestAccAWSEFSFileSystem_pagedTags (48.26s)
=== RUN   TestAccAWSEFSFileSystem_KmsKey
--- PASS: TestAccAWSEFSFileSystem_KmsKey (87.40s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	299.772s
```


aws/resource_aws_efs_file_system.go
```diff
@@ -76,6 +91,14 @@ func resourceAwsEfsFileSystemCreate(d *schema.ResourceData, meta interface{}) er
 		createOpts.PerformanceMode = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("encrypted"); ok {
+		createOpts.Encrypted = aws.Bool(v.(bool))
+	}
+
+	if v, ok := d.GetOk("kms_key_id"); ok {
+		createOpts.KmsKeyId = aws.String(v.(string))
+	}
```
Ninir: We should check that `Encrypted` is true if `KmsKeyId` is set, and catch the related Exception if the API call is made (BadRequest perhaps?).

aws/resource_aws_efs_file_system_test.go
```diff
@@ -170,6 +171,28 @@ func TestAccAWSEFSFileSystem_pagedTags(t *testing.T) {
...
+		CheckDestroy: testAccCheckEfsFileSystemDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEFSFileSystemConfigWithKmsKey(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestMatchResourceAttr(
+						"aws_efs_file_system.foo-with-kms",
+						"kms_key_id",
+						keyRegex,
+					),
```
Ninir: We should also check that `encrypted` is well set here :)

aws/resource_aws_efs_file_system_test.go
```diff
@@ -351,3 +374,33 @@ resource "aws_efs_file_system" "foo-with-performance-mode" {
...
+	      "Action": "kms:*",
+	      "Resource": "*"
+	    }
+	  ]
+	}
+	POLICY
+	}
+
+	resource "aws_efs_file_system" "foo-with-kms" {
+		encrypted = true
```
Ninir: `encrypted` should be aligned with `kms_key_id`

aws/resource_aws_efs_file_system_test.go
pjfbashton: Thanks, good call. Check added

aws/resource_aws_efs_file_system_test.go
```diff
@@ -351,3 +374,33 @@ resource "aws_efs_file_system" "foo-with-performance-mode" {
 	performance_mode = "maxIO"
 }
 `
+
+func testAccAWSEFSFileSystemConfigWithKmsKey(rInt int) string {
+	return fmt.Sprintf(`
+	resource "aws_kms_key" "foo" {
```
pjfbashton: I actually dont think we need that policy at all, test runs fine without it so I removed it

aws/resource_aws_efs_file_system.go
pjfbashton: Looking into this now, thanks ;)

Ninir:
Ping me when you're all ok on that @pjfbashton, will make the review. Thanks for your efforts :)

pjfbashton:
@Ninir seems that `KmsKeyId` is ignored by sdk when `Encrypted` is false so I added a check that will throw an error if `encrypted` is not `true` when  `kms_key_id` is specified. would that work?

pjfbashton:
That's great @Ninir ! Thanks for your help :+1: ;)

shezaan:
Stupid question, which version of TF will this be added to? I am already on 0.10.4.
Documentation still doesn't reflect this if its already there.
Thanks in advance!

Ninir:
Hi @shezaan !

Since the [Terraform split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/), providers (AWS, Google, Azure, Kubernetes, etc...) provide their own releases.

The one including EFS encryption will then be in the next AWS release. I can't provide any ETA of it yet, sadly.
I will try to get back to you on that as soon as possible! 👍

Issue #259, opened 2017-06-13
Terraform does not use IAM Role for ECS Task as credential provider


hashibot:
_This issue was originally opened by @iwat as hashicorp/terraform#8746. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

Terraform v0.7.3
### Affected Resource(s)
- aws_alb_target_group
- aws_security_group
- a lot

This affects all AWS related command.
### Terraform Configuration Files

``` ccl
resource "aws_security_group_rule" "demo_pri_ingress_vpn_service" {
    security_group_id = "${aws_security_group.demo_pri.id}"
    type = "ingress"

    from_port   = 80
    to_port     = 80
    protocol    = "tcp"
    cidr_blocks = ["${data.terraform_remote_state.infra.vpn-cidr_block}"]
}
```
### Debug Output

https://gist.github.com/iwat/df0b0ebfe2f8db62adfd5953bfd6b92c
### Panic Output

None
### Expected Behavior

It should work by using IAM Role for ECS Task.
awscli works
### Actual Behavior

It was using EC2 Instance Role which does not allow this action.

Error retrieving Target Group: AccessDenied: User: arn:aws:sts::872767853649:assumed-role/myrole/i-0223aeb98c19f2d0d
### Steps to Reproduce
- Setup an EC2, do not provide any critical IAM action.
- Setup ECS task, provide required IAM action for testing.
- Try AWSCLI inside the running ECS task, it should work fine.
- Run terraform on AWS ECS Task.
### Important Factoids

None
### References
- http://docs.aws.amazon.com/de_de/AmazonECS/latest/developerguide/task-iam-roles.html
- https://github.com/hashicorp/terraform/blob/v0.7.3/builtin/providers/aws/auth_helpers.go#L97
- https://github.com/hashicorp/terraform/blob/v0.7.3/builtin/providers/aws/auth_helpers.go#L128
- https://github.com/aws/aws-sdk-go/blob/v1.4.8/aws/defaults/defaults.go#L100

jch254:
Pls pls

jch254:
I hope Terraform will be able to get these credentials for use in CodeBuild. For now I am doing the following:

```
[...]

PR #1425, opened 2017-08-15
Add support for Task Roles when running on ECS or CodeBuild

jekh:
This is a port of my original PR [#14199](https://github.com/hashicorp/terraform/pull/14199) from the terraform repo, before the provider spin-off in 0.10+. It addresses #259.

Here's the comment from the original PR:
> [This PR] adds the RemoteCredProvider from the AWS SDK if the AWS_CONTAINER_CREDENTIALS_RELATIVE_URI environment variable is set. The RemoteCredProvider uses the value of that environment variable and appends it to the hard-coded URL http://169.254.170.2, and uses the resulting URL to obtain Task Role credentials.

The benefit of this approach is that it falls back to the default AWS SDK to obtain ECS credentials. It may be valuable to always fall back to the default SDK unless specific terraform overrides are configured, rather than providing a custom credentials bootstrap or mimicking the SDK behavior, so that terraform is able to automatically take advantage of future updates to the SDK's credential-finding behavior.

### DIFF
--- a/aws/auth_helpers.go
+++ b/aws/auth_helpers.go
@@ -13,6 +13,7 @@ import (
 	awsCredentials "github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
+	"github.com/aws/aws-sdk-go/aws/defaults"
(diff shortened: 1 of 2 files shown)


kshcherban:
When is this going to be merged?

jch254:
Is this solving https://github.com/hashicorp/terraform/issues/16278? I'm still facing the issue with TF  0.11.1 and AWS provider 1.6 - Error configuring the backend "s3": No valid credential sources found for AWS Provider.

PR #1430, opened 2017-08-16
Add retries for AppScaling policies throttling exceptions

s-maj:
Fix for https://github.com/terraform-providers/terraform-provider-aws/issues/1428, allows retry on DescribeScalingPolices throttling error. 

```
TF_ACC=1 go test ./aws -v -run=TestAccAWSAppautoScalingPolicy -timeout 120m
=== RUN   TestAccAWSAppautoScalingPolicy_basic
--- PASS: TestAccAWSAppautoScalingPolicy_basic (128.68s)
=== RUN   TestAccAWSAppautoScalingPolicy_spotFleetRequest
--- PASS: TestAccAWSAppautoScalingPolicy_spotFleetRequest (84.73s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	213.435s
```

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -399,6 +399,20 @@ func (c *Config) Client() (interface{}, error) {
 		}
 	})
 
+	// Workaround for https://github.com/aws/aws-sdk-go/issues/1472
+	client.appautoscalingconn.Handlers.Retry.PushBack(func(r *request.Request) {
+		if !strings.HasPrefix(r.Operation.Name, "Describe") && !strings.HasPrefix(r.Operation.Name, "List") {
+			return
+		}
+		err, ok := r.Error.(awserr.Error)
+		if !ok || err == nil {
+			return
+		}
+		if err.Code() == applicationautoscaling.ErrCodeFailedResourceAccessException {
+			r.Retryable = aws.Bool(true)
+		}
+	})
+
 	return &client, nil
 }
 


Ninir:
@s-maj Got it :)

Do you think you could provide a TF configuration (acceptance test) including a lot of alarms, etc, so that we can reproduce it and validate please?

Would help a lot for the review :)

Thanks!

s-maj:
Sure thing, but first I would like to add some extra background. Initially, a very similar issue was reported https://github.com/aws/aws-sdk-go/issues/1376 but since AWS API is quirky, this is working as designed (client is responsible for handling this, not SDK). Then https://github.com/terraform-providers/terraform-provider-aws/pull/1085 was submitted and merged. 
I asked a question https://github.com/aws/aws-sdk-go/issues/1472 and looking at the answer this is also working as expected. I had a chat with AWS support they confirmed that this also should be implemented on the client-side. 
Now, testing can be quirky too:
If you have a really slow Internet connection/lack of CPU, this error will never occur to you.
If you have increased cloudwatch DescribeAlarms limit (yep, it's per account setting, see: http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html) it will affect how this issue is triggered.

s-maj:
You can test it using something like this: https://gist.github.com/s-maj/0bf81fef71274d31605d4caf43631a2c

Issue #1410, opened 2017-08-14
AWS S3 Bucket Policy Documentation missing


flacito:
https://www.terraform.io/docs/providers/aws/r/s3_bucket_policy.html

Not much there. Gonna have to go read the code to get it right.

coingraham:
@flacito are you expecting something like https://www.terraform.io/docs/providers/aws/r/iam_group_policy_attachment.html ?

flacito:
I guess it would be nice to have a functional policy instead of something like "#omitted".  Most of the resources are well-documented and create functional AWS artifact(s).  That's all.  I did figure out how to get it going from the the main [s3_bucket](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html) documentation. So I'm good.

Go ahead and close if you think the page is OK.

coingraham:
I'll look at doing a PR in reference to this Issue sometime this week and see what the devs think.

flacito:
Wow thanks.  Tremendously responsive support!

coingraham:
@flacito Take a look at my PR and see what you think.

flacito:
Sorry couldn't get to it until now. Took a look. Great change.  

You guys are super responsive.   THANKS!

PR #1433, opened 2017-08-16
S3 bucket policy html markdown update

coingraham:
Fixes [#1410](https://github.com/terraform-providers/terraform-provider-aws/issues/1410#issuecomment-322566852).  Provides a working example for the aws_s3_bucket_policy resource and removes confusing reference to "Versioning" that was likely a copy paste error from somewhere else.


coingraham:
Some reason my previous commits were included.

PR #1434, opened 2017-08-16
S3 bucket policy html markdown update

coingraham:
Second try.

Fixes [#1410](https://github.com/terraform-providers/terraform-provider-aws/issues/1410#issuecomment-322566852). Provides a working example for the aws_s3_bucket_policy resource and removes confusing reference to "Versioning" that was likely a copy paste error from somewhere else.

### DIFF
--- a/website/docs/r/s3_bucket_policy.html.markdown
+++ b/website/docs/r/s3_bucket_policy.html.markdown
@@ -12,20 +12,37 @@ Attaches a policy to an S3 bucket resource.
 
 ## Example Usage
 
-### Using versioning
+### Creating a custom policy and adding to the bucket
 
 ```hcl
 resource "aws_s3_bucket" "b" {
-  # Arguments
+  bucket = "my_tf_test_bucket"
 }
 
-data "aws_iam_policy_document" "b" {
-  # Policy statements
+variable "my_bucket_policy" {
+  default = <<EOF
+{
+  "Version": "2012-10-17",
+  "Id": "MYBUCKETPOLICY",
+  "Statement": [
+    {
+      "Sid": "IPAllow",
+      "Effect": "Deny",
+      "Principal": "*",
+      "Action": "s3:*",
+      "Resource": "arn:aws:s3:::my_tf_test_bucket/*",
+      "Condition": {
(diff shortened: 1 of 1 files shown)


website/docs/r/s3_bucket_policy.html.markdown
```diff
@@ -12,20 +12,37 @@ Attaches a policy to an S3 bucket resource.
 
 ## Example Usage
 
-### Using versioning
+### Creating a custom policy and adding to the bucket
```
Ninir: Since this is basic, I would either remove this line, or write basic. Thoughts 🤷‍♂️  😄 ?

website/docs/r/s3_bucket_policy.html.markdown
```diff
@@ -12,20 +12,37 @@ Attaches a policy to an S3 bucket resource.
...
+    } 
+  ]
+}
+EOF
 }
 
 resource "aws_s3_bucket_policy" "b" {
   bucket = "${aws_s3_bucket.b.id}"
-  policy = "${data.aws_iam_policy_document.b.json}"
+  policy = "${var.my_bucket_policy}"
```
Ninir: I would use the policy itself there, rather than depending on the variable, to make things simpler.
What do you think? :)

website/docs/r/s3_bucket_policy.html.markdown
coingraham: Sure.  I did a spot check on a couple of other pages and guessed that there needed to be some language here.  If it's not required, drop it.  I was more concerned with removing the "Using versioning" part.

Ninir: Let's just put "Basic" / "Basic Usage" if you don't mind then :)

website/docs/r/s3_bucket_policy.html.markdown
coingraham: Oh, sure.  I was keeping the pattern with the data source, but had issues getting that to work because of the "Principal" requirement for S3 bucket policies. I'll test and update soon.

website/docs/r/s3_bucket_policy.html.markdown
coingraham: Okie dokie

PR #1438, opened 2017-08-16
Add ability to specify EMR instance group bid price

chussenot:
Spot Instances in Amazon EMR provide an option for you to purchase Amazon EC2 instance capacity at a reduced cost as compared to On-Demand purchasing.

The `bid_price` for spot instances. This must be a string. To define Spot Instances in the instance group, you have to specify a single Spot Instance type and a bid price. 
Changing this forces a new resource to be created.

http://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-instance-purchasing-options.html

```hcl
resource "aws_emr_instance_group" "task" {
    cluster_id     = "${aws_emr_cluster.tf-test-cluster.id}"
    instance_count = 1
    instance_type  = "m1.small"
    bid_price = "100"
  }
```

Partially fixes https://github.com/terraform-providers/terraform-provider-aws/issues/1147

### DIFF
--- a/aws/resource_aws_emr_instance_group.go
+++ b/aws/resource_aws_emr_instance_group.go
@@ -55,6 +55,11 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+			"bid_price": {
+				Type: schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
 			"ebs_config": {
 				Type:     schema.TypeSet,
 				Optional: true,
@@ -116,6 +121,7 @@ func readEmrEBSConfig(d *schema.ResourceData) *emr.EbsConfiguration {
 	return result
 }
 
+// Please also see https://docs.aws.amazon.com/goto/WebAPI/elasticmapreduce-2009-03-31/AddInstanceGroups
 func resourceAwsEMRInstanceGroupCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).emrconn
 
@@ -123,6 +129,7 @@ func resourceAwsEMRInstanceGroupCreate(d *schema.ResourceData, meta interface{})
 	instanceType := d.Get("instance_type").(string)
 	instanceCount := d.Get("instance_count").(int)
 	groupName := d.Get("name").(string)
+	bidPrice := d.Get("bid_price").(string)
 
 	ebsConfig := readEmrEBSConfig(d)
 
@@ -133,6 +140,7 @@ func resourceAwsEMRInstanceGroupCreate(d *schema.ResourceData, meta interface{})
 				InstanceCount:    aws.Int64(int64(instanceCount)),
 				InstanceType:     aws.String(instanceType),
 				Name:             aws.String(groupName),
+				BidPrice:         aws.String(bidPrice),
 				EbsConfiguration: ebsConfig,
 			},
 		},
--- a/aws/resource_aws_emr_instance_group.go
+++ b/aws/resource_aws_emr_instance_group.go
@@ -59,6 +59,7 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
 				Type: schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				ValidateFunc: validateAwsEmrBidPrice,
 			},
 			"ebs_config": {
 				Type:     schema.TypeSet,
--- a/aws/resource_aws_emr_instance_group_test.go
+++ b/aws/resource_aws_emr_instance_group_test.go
@@ -29,6 +29,22 @@ func TestAccAWSEMRInstanceGroup_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSEMRInstanceGroup_spot(t *testing.T) {
+	var ig emr.InstanceGroup
+	rInt := acctest.RandInt()
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEmrInstanceGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEmrInstanceGroupConfig_bid(rInt),
+				Check:  testAccCheckAWSEmrInstanceGroupExists("aws_emr_instance_group.task", &ig),
+			},
+		},
+	})
+}
+
 // Confirm we can scale down the instance count. Regression test for https://github.com/terraform-providers/terraform-provider-aws/issues/1264
 func TestAccAWSEMRInstanceGroup_zero_count(t *testing.T) {
 	var ig emr.InstanceGroup
@@ -402,6 +418,17 @@ func testAccAWSEmrInstanceGroupConfig(r int) string {
     cluster_id     = "${aws_emr_cluster.tf-test-cluster.id}"
     instance_count = 1
     instance_type  = "m1.small"
+  }
+	`, r, r, r, r, r, r)
+}
+
+func testAccAWSEmrInstanceGroupConfig_bid(r int) string {
+	return fmt.Sprintf(testAccAWSEmrInstanceGroupBase+`
+	resource "aws_emr_instance_group" "task" {
+    cluster_id     = "${aws_emr_cluster.tf-test-cluster.id}"
+    instance_count = 1
+    instance_type  = "m1.small"
+    bid_price = "0.10"
   }
 	`, r, r, r, r, r, r)
 }
--- a/aws/validators.go
+++ b/aws/validators.go
@@ -861,6 +861,16 @@ func validateAwsEcsPlacementStrategy(stratType, stratField string) error {
 	return nil
 }
 
+func validateAwsEmrBidPrice(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+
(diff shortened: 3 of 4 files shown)


aws/resource_aws_emr_instance_group.go
```diff
@@ -116,23 +121,26 @@ func readEmrEBSConfig(d *schema.ResourceData) *emr.EbsConfiguration {
...
 	groupName := d.Get("name").(string)
+	bidPrice := d.Get("bid_price").(string)
 
 	ebsConfig := readEmrEBSConfig(d)
 
 	params := &emr.AddInstanceGroupsInput{
 		InstanceGroups: []*emr.InstanceGroupConfig{
 			{
-				InstanceRole:     aws.String("TASK"),
+				InstanceRole:     aws.String("TASK"), // MASTER|CORE|TASK
```
Ninir: This should not be part of this PR I think

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -406,6 +406,17 @@ func testAccAWSEmrInstanceGroupConfig(r int) string {
 	`, r, r, r, r, r, r)
 }
 
+func testAccAWSEmrInstanceGroupConfig_bid(r int) string {
```
Ninir: This is only a config that is not ran as an acceptance test.

To use this config, you should write an acceptance test, as the ones that are named `Test*` (with the capital T), generally at the top of `_test.go` files.

aws/resource_aws_emr_instance_group.go
```diff
@@ -55,6 +55,11 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+			"bid_price": {
```
Ninir: We should also check the value returned by AWS when reading the resource, i.e. in the `resourceAwsEMRInstanceGroupRead` function, around lines 190. 
This helps Terraform knows if the value changed, for example from a manual edit in the console.

aws/resource_aws_emr_instance_group.go
chussenot: Sure...

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -406,6 +406,17 @@ func testAccAWSEmrInstanceGroupConfig(r int) string {
...
+	return fmt.Sprintf(testAccAWSEmrInstanceGroupBase+`
+	resource "aws_emr_instance_group" "task" {
```
chussenot: Done

aws/resource_aws_emr_instance_group_test.go
chussenot: My bad

aws/resource_aws_emr_instance_group.go
chussenot: A little help for this point will be highly appreciated 🥇  @Ninir

Ninir: [Here](https://github.com/terraform-providers/terraform-provider-aws/pull/1438/files#diff-c081c3ba2c4eb51067d67a5e2dd80a85R191), you can see `d.Set("instance_type", group.InstanceType)`, which writes the value returned by the AWS API call, to the state. 
We should use the same approach for the newly added argument, doing some sort of `d.Set("bid_price", group.BidPrice)`

aws/validators.go
```diff
@@ -861,6 +861,16 @@ func validateAwsEcsPlacementStrategy(stratType, stratField string) error {
 	return nil
 }
 
+func validateAwsEmrBidPrice(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+
+	if len(value) > 256 {
+		errors = append(errors, fmt.Errorf("%q cannot be longer than 256 characters", k))
+	}
```
Ninir: You should also check for the `[\u0020-\uD7FF\uE000-\uFFFD\uD800\uDC00-\uDBFF\uDFFF\r\n\t]*` pattern validation.

Also, this will need to have a unit test attached. You can find many examples in the `validators_test.go` file

website/docs/r/emr_instance_group.html.md
```diff
@@ -41,6 +41,8 @@ The following arguments are supported:
 * `size` - (Optional) The volume size, in gibibytes (GiB). This can be a number from 1 - 1024. If the volume type is EBS-optimized, the minimum value is 10.
 * `type` - (Optional) The volume type. Valid options are 'gp2', 'io1' and 'standard'.
 * `volumes_per_instance` - (Optional) The number of EBS Volumes to attach per instance.
+* `bid_price` - (Optional) Bid price for each EC2 instance in the instance group when launching nodes as Spot Instances, expressed in USD.
+This must be a string, for exemple '0.10'. Changing this forces a new resource to be created.
```
Ninir: This line should be part of the item bullet right above

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -406,6 +422,17 @@ func testAccAWSEmrInstanceGroupConfig(r int) string {
...
+    cluster_id     = "${aws_emr_cluster.tf-test-cluster.id}"
+    instance_count = 1
+    instance_type  = "m1.small"
+    bid_price = "0.10"
+  }
+	`, r, r, r, r, r, r)
+}
```
Ninir: I would go for the following indentation here:

```go
	return fmt.Sprintf(testAccAWSEmrInstanceGroupBase+`
resource "aws_emr_instance_group" "task" {
  cluster_id     = "${aws_emr_cluster.tf-test-cluster.id}"
  instance_count = 1
  instance_type  = "m1.small"
  bid_price      = "0.10"
}
`, r, r, r, r, r, r)
}
```

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -29,6 +29,22 @@ func TestAccAWSEMRInstanceGroup_basic(t *testing.T) {
...
+	var ig emr.InstanceGroup
+	rInt := acctest.RandInt()
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEmrInstanceGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEmrInstanceGroupConfig_bid(rInt),
+				Check:  testAccCheckAWSEmrInstanceGroupExists("aws_emr_instance_group.task", &ig),
```
Ninir: To add multiple checks for this configuration, we should definitely use this approach:

```go
        Check: resource.ComposeAggregateTestCheckFunc(
            testAccCheckAWSEmrInstanceGroupExists("aws_emr_instance_group.task", &ig),
            resource.TestCheckResourceAttr("aws_emr_instance_group.task", "bid_price", "0.10"),
        ),
```

This would allow to check that the EmrInstanceGroupExists and that the bid_price is well set. This depends on the `d.Set("bid_price", group.BidPrice)` instruction I talked about :)

aws/resource_aws_emr_instance_group.go
```diff
@@ -116,13 +122,15 @@ func readEmrEBSConfig(d *schema.ResourceData) *emr.EbsConfiguration {
 	return result
 }
 
+// Please also see https://docs.aws.amazon.com/goto/WebAPI/elasticmapreduce-2009-03-31/AddInstanceGroups
```
Ninir: Could you put this instruction above line 137?
A create call *might* do multiple calls, like creating the instance group, attach options... it hardly depends on the API-side, but hey... 🤷‍♂️ 😄

aws/resource_aws_emr_instance_group.go
chussenot: Perfect!

aws/resource_aws_emr_instance_group.go
```diff
@@ -55,6 +55,11 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+			"bid_price": {
+				Type: schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
chussenot: Done

damdr:
I think we can close this issue.
The bid price feature is now realized.
https://www.terraform.io/docs/providers/aws/r/emr_cluster.html#bid_price

Ninir:
@damdr The `bid_price` is implemented at the `resource_aws_emr_cluster` level, whereas this work is about the `resource_aws_emr_instance_group` 😄 

While those two were different in the past, we do think the second one should be deprecated in favor of the emr_cluster one.

Still, tell us if you need any help resolving this :)

damdr:
My bad!
Aws offers to add or modify an instance_group in sdk and cli.
I think it can still be useful to have a resource to only handle the instance_group.

http://docs.aws.amazon.com/cli/latest/reference/emr/add-instance-groups.html
http://docs.aws.amazon.com/cli/latest/reference/emr/modify-instance-groups.html

chussenot:
:+1:

bflad:
@chussenot are you able to finish the feedback from this PR?

bcornils:
looks like we have a check failing amongst other things. Has anyone had time to assess or review criticality of the failures?

chussenot:
@bflad Nope I'm sorry but don't plan to continue this work. I think we should close this PR @Ninir

bflad:
Closing by request 👍

PR #1439, opened 2017-08-16
Set availability_zone attribute on RDS cluster instance

kl4w:
First PR so let me know if I'm missing anything.

Current docs state that `aws_rds_cluster_instance` returns the `availability_zones` attribute but is not the case https://www.terraform.io/docs/providers/aws/r/rds_cluster_instance.html#availability_zones 

Enhanced resource to set the attribute and remove plural from docs.

### DIFF
--- a/aws/resource_aws_rds_cluster_instance.go
+++ b/aws/resource_aws_rds_cluster_instance.go
@@ -159,6 +159,11 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
 				Default:  0,
 			},
 
+			"availability_zone": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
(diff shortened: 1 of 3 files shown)


aws/resource_aws_rds_cluster_instance.go
```diff
@@ -296,6 +301,7 @@ func resourceAwsRDSClusterInstanceRead(d *schema.ResourceData, meta interface{})
 	d.Set("promotion_tier", db.PromotionTier)
 	d.Set("preferred_backup_window", db.PreferredBackupWindow)
 	d.Set("preferred_maintenance_window", db.PreferredMaintenanceWindow)
+	d.Set("availability_zone", db.AvailabilityZone)
```
Ninir: Could you augment an existing acceptance test to check for this to be set? perhaps the basic one?

kl4w: Hi @Ninir 

I've added to the basic test now

PR #1444, opened 2017-08-17
Add `stopped` as a valid state for volume attachment

### LINKED ISSUE #1311: aws_volume_attachment shouldnt have to wait for instance to be running to attach
Terraform 0.9.8

I have an aws_instance terraformed, and in the 'stopped' state (by design). I should easily be able to attach an ebs volume to it in this state with an aws_volume_attachment. However terraforming an aws_volume_attachment to the instance leads to this error:

```
1 error(s) occurred:

* aws_volume_attachment.foo: 1 error(s) occurred:

* aws_volume_attachment.foo: Error waiting for instance (i-11112222333344445) to become ready: unexpected state 'stopped', wanted target 'running'. last error: %!s(<nil>)

```

This seems to be bad terraform design behaviour - the volume attachment should work just fine - indeed it is often safer to attach volumes to a stopped instance than to a started one.

bodgit:
Should fix #1311

### DIFF
--- a/aws/resource_aws_volume_attachment.go
+++ b/aws/resource_aws_volume_attachment.go
@@ -81,8 +81,8 @@ func resourceAwsVolumeAttachmentCreate(d *schema.ResourceData, meta interface{})
 		// a spot request and whilst the request has been fulfilled the
 		// instance is not running yet
 		stateConf := &resource.StateChangeConf{
-			Pending:    []string{"pending"},
-			Target:     []string{"running"},
+			Pending:    []string{"pending", "stopping"},
+			Target:     []string{"running", "stopped"},
 			Refresh:    InstanceStateRefreshFunc(conn, iID, "terminated"),
 			Timeout:    10 * time.Minute,
 			Delay:      10 * time.Second,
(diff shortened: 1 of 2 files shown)


Ninir:
Hi @bodgit 

Do you think you would have time to finish this work and add the acceptance test asked by @radeksimko ?
Would be much appreciated! :)

bodgit:
I've added a test that passes:
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSVolumeAttachment_attachStopped'
TF_ACC=1 go test ./aws -v -run=TestAccAWSVolumeAttachment_attachStopped -timeout 120m
=== RUN   TestAccAWSVolumeAttachment_attachStopped
--- PASS: TestAccAWSVolumeAttachment_attachStopped (199.85s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	199.880s
```
I also reverted my change and checked that the test also fails correctly.

jen20:
Thanks for adding the test here - this looks good to me now!

PR #1449, opened 2017-08-18
Updated DMS docs for creating a dynamodb endpoint.

chrisjharding:
This refers to https://github.com/terraform-providers/terraform-provider-aws/pull/1002. The docs are missing an argument for this endpoint. As part of this PR, the `service_access_role` param was added.
It should probably be deployed soon but it's not critical.


stack72:
Nice work on this addition!! Sorry i missed this when bringing it over

PR #1463, opened 2017-08-21
Support for standalone acceptance testing

mubeta06:
In this PR I have simply exposed the various AWS connections outside of the aws package to support the development of standalone acceptance testing. We have a number of remote modules that we have put together that we would like to integrate into CI / CD processes. Rather than roll our own acceptance testing framework we thought we would attempt to leverage the existing terraform acceptance testing framework.

If there is an alternate path or better suggestion I'd be happy to hear about it.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -178,12 +178,220 @@ type AWSClient struct {
 	iotconn               *iot.IoT
 }
 
-func (c *AWSClient) S3() *s3.S3 {
-	return c.s3conn
+func (c *AWSClient) CloudFormation() *cloudformation.CloudFormation {
+	return c.cfconn
+}
+
+func (c *AWSClient) CloudFront() *cloudfront.CloudFront {
+	return c.cloudfrontconn
+}
+
+func (c *AWSClient) CloudTrail() *cloudtrail.CloudTrail {
+	return c.cloudtrailconn
+}
+
+func (c *AWSClient) CloudWatch() *cloudwatch.CloudWatch {
+	return c.cloudwatchconn
+}
+
+func (c *AWSClient) CloudWatchLogs() *cloudwatchlogs.CloudWatchLogs {
+	return c.cloudwatchlogsconn
+}
+
+func (c *AWSClient) CloudWatchEvents() *cloudwatchevents.CloudWatchEvents {
+	return c.cloudwatcheventsconn
+}
+
+func (c *AWSClient) CognitoIdentity() *cognitoidentity.CognitoIdentity {
+	return c.cognitoconn
+}
+
+func (c *AWSClient) ConfigService() *configservice.ConfigService {
+	return c.configconn
+}
+
+func (c *AWSClient) DeviceFarm() *devicefarm.DeviceFarm {
+	return c.devicefarmconn
+}
+
+func (c *AWSClient) DatabaseMigrationService() *databasemigrationservice.DatabaseMigrationService {
+	return c.dmsconn
+}
+
+func (c *AWSClient) DirectoryService() *directoryservice.DirectoryService {
+	return c.dsconn
 }
 
 func (c *AWSClient) DynamoDB() *dynamodb.DynamoDB {
 	return c.dynamodbconn
+}
+
+func (c *AWSClient) EC2() *ec2.EC2 {
+	return c.ec2conn
+}
+
+func (c *AWSClient) ECR() *ecr.ECR {
+	return c.ecrconn
+}
+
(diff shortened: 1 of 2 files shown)


catsby:
Hey @mubeta06 – can you elaborate (with pseudo code even?) how this is used? Thanks!

mubeta06:
Sure thing @catsby. So something similar to this
```
package arrakis

import (
	...

	// eventually the cozero fork should die when
	// https://github.com/terraform-providers/terraform-provider-aws/pull/1463
	provider "github.com/cozero/terraform-provider-aws/aws"
)

func TestAccArrakisIAMStandardRoles(t *testing.T) {
	var role iam.GetRoleOutput
	dev := 123456789100
	elmo := 123456789100

	resource.Test(t, resource.TestCase{
		CheckDestroy: testAccCheckAWSRoleDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccStandardRoles(dev, module, elmo),
					testAccCheckAWSRoleExists(module, "aws_iam_role.cloudwatch_readonly", &role),
					testAccCheckAWSRoleExists(module, "aws_iam_role.deploy", &role),
				),
			},
		},
	})
}

func testAccCheckAWSRoleExists(module, n string, res *iam.GetRoleOutput) resource.TestCheckFunc {
		module := s.ModuleByPath([]string{"root", module})
		if module == nil {
			return fmt.Errorf("module is nil")
		}

		rs, ok := module.Resources[n]
		if !ok {
			return fmt.Errorf("Not found: %s", n)
		}

		if rs.Primary.ID == "" {
			return fmt.Errorf("No Role name is set")
		}

		iamconn := testAccProvider.Meta().(*provider.AWSClient).IAM()

		resp, err := iamconn.GetRole(&iam.GetRoleInput{
			RoleName: aws.String(rs.Primary.ID),
		})
		if err != nil {
			return err
		}

		*res = *resp

		return nil
	}
}

func testAccCheckAWSRoleDestroy(s *terraform.State) error {
	iamconn := testAccProvider.Meta().(*provider.AWSClient).IAM()

	...

	return nil
}

func testAccStandardRoles(remote int, module string, account int) string {
	return fmt.Sprintf(`
module "%s" {
  source = "git::ssh://git@myrepo/arrakis-iam.git?ref=v0.1.1//standard-roles"

  master_account_id = "%d"
}
`, remote, module, account)
}
```

And we run our acceptance tests in the usual way.

mubeta06:
Nudge!

mubeta06:
@Ninir what gives?????

Ninir:
Hi @mubeta06 

I would have prefer @catsby to reply here since it was him who originally started handling this.

In my opinion, I'm just wondering that one needs to update this work on each new available service. I would probably prefer to return the AWSClient directly rather than returning single connections. Thoughts?

mubeta06:
Hi @Ninir,

Thanks for the feedback. I am not sure how what you are proposing would work. I have my acceptance tests that live in a different package to terraform-provider-aws and, therefore, the client's connections to the various services (with the exception of DynamoDB and S3) are not at my disposal outside of the terraform-provider-aws package. Is there some way I can use the AWSClient directly as you say that I am overlooking?

Ninir:
I'm thinking about something like exposing:

```go
func (c *AWSClient) AWSClient() *cloudformation.CloudFormation {
	return c
}
```

And being able tu use `c := client.AWSClient()` and `c.3conn` for instance.

Just an idea I'm thinking about... :)

mubeta06:
I'm still confused. I believe what you are proposing would not compile.

Ninir:
@mubeta06 Why that? (there was a typo I fixed, which was the return type)
The provided code above compiles! thoughts?

mubeta06:
@Ninir access to the client itself is not the problem but rather access to each of the connections which are private members of the `*AWSClient` structure.

Ninir:
@mubeta06 Sorry my answer was not complete.

I'm not for or against having public methods: it's just that we would need to add one each time a new service is added to Terraform. 
Using reflection, you wouldn't have to deal with that, nor with the newly services that could be added afterwards.

On your side, it is "just" a matter of doing that, nothing more elsewhere.

What do you think? 😄

bflad:
For anyone following this pull request, I am going to close it just due to very long inactivity and staleness with its support.

This is not to say we would not allow support for using the provider client/connections with outside libraries in the future, but merely to help clean up the open repository pull requests with actionable items. If folks are still interested in something like this functionality, we should probably open a new GitHub issue to discuss the implementation. If this pull request does in fact look fine after further discussion, we can always reopen this pull request. 👍

Issue #1469, opened 2017-08-22
What arn should one use in aws_cloudwatch_event_target if using ecs_target?


scream314:
Hi there,

I am trying to create an ECS Scheduled Task with Terraform.

### Terraform Version
```
terraform -v
Terraform v0.10.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_cloudwatch_event_target

### Terraform Configuration Files
```hcl
resource "aws_cloudwatch_log_group" "scheduled-job" {
  name = "scheduled-job"
}

data "template_file" "scheduled-job_task_definition" {
  template = "${file("${path.module}/scheduled-job.json")}"
  vars {
    log_group_name = "${aws_cloudwatch_log_group.scheduled-job.name}"
    log_group_region = "${var.region}"
    log_stream_prefix = "scheduled-job_"
  }
}

resource "aws_ecs_task_definition" "scheduled-job" {
  family = "scheduled-job"
  container_definitions = "${data.template_file.scheduled-job_task_definition.rendered}"
}

resource "aws_cloudwatch_event_target" "scheduled-job" {
  target_id = "scheduled-job"
  rule      = "${aws_cloudwatch_event_rule.scheduled-job.name}"
  arn = "${aws_ecs_task_definition.scheduled-job.arn}"
  ecs_target {
    task_count = 1
    task_definition_arn = "${aws_ecs_task_definition.scheduled-job.arn}"
  }
}

resource "aws_cloudwatch_event_rule" "scheduled-job" {
  name        =  "scheduled-job"
  description = "Run a scheduled job every 1 hours"
  schedule_expression = "rate(1 hour)"
}
```

### Expected Behavior
I would expect my scheduled task to be created.

### Actual Behavior
```
* aws_cloudwatch_event_target.scheduled_job: Creating CloudWatch Event Target failed: ValidationException: Parameter arn:aws:ecs:[region]:[accountid]:task-definition/scheduled-job:1 is not valid. Reason: Provided Arn is not in correct format.
```
I think the error is correct, it is not a correct format (note the last colon between task definition name and version).
I am also not sure what ARN should be specified here, all examples I found used Lambda or Kinesis, where the ARN to be used is obvious.
The ARN also cannot be omitted, as it is required.
The desired Scheduled Task can be manually configured under the ECS Service (on the Console).

### Steps to Reproduce
1. Create a scheduled task like in the snipped above.
1. `terraform apply`

### Important Factoids
N/A

[...]

PR #1470, opened 2017-08-22
GH-1469 - More documentation about using ecs_target.

scream314:
https://github.com/terraform-providers/terraform-provider-aws/issues/1469

### DIFF
--- a/website/docs/r/cloudwatch_event_target.html.markdown
+++ b/website/docs/r/cloudwatch_event_target.html.markdown
@@ -68,11 +68,11 @@ The following arguments are supported:
 
 * `rule` - (Required) The name of the rule you want to add targets to.
 * `target_id` - (Optional) The unique target assignment ID.  If missing, will generate a random, unique id.
-* `arn` - (Required) The Amazon Resource Name (ARN) associated of the target.
+* `arn` - (Required) The Amazon Resource Name (ARN) associated of the target. (It should be the ID of the ECS Cluster if `ecs_target` is used.)
 * `input` - (Optional) Valid JSON text passed to the target.
 * `input_path` - (Optional) The value of the [JSONPath](http://goessner.net/articles/JsonPath/)
 	that is used for extracting part of the matched event when passing it to the target.
-* `role_arn` - (Optional) The Amazon Resource Name (ARN) of the IAM role to be used for this target when the rule is triggered.
+* `role_arn` - (Optional) The Amazon Resource Name (ARN) of the IAM role to be used for this target when the rule is triggered. (Required if `ecs_target` is used.)
 * `run_command_targets` - (Optional) Parameters used when you are using the rule to invoke Amazon EC2 Run Command. Documented below. A maximum of 5 are allowed.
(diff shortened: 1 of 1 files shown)


website/docs/r/cloudwatch_event_target.html.markdown
```diff
@@ -68,11 +68,11 @@ The following arguments are supported:
 
 * `rule` - (Required) The name of the rule you want to add targets to.
 * `target_id` - (Optional) The unique target assignment ID.  If missing, will generate a random, unique id.
-* `arn` - (Required) The Amazon Resource Name (ARN) associated of the target.
+* `arn` - (Required) The Amazon Resource Name (ARN) associated of the target. (It should be the ID of the ECS Cluster if `ecs_target` is used.)
 * `input` - (Optional) Valid JSON text passed to the target.
 * `input_path` - (Optional) The value of the [JSONPath](http://goessner.net/articles/JsonPath/)
 	that is used for extracting part of the matched event when passing it to the target.
-* `role_arn` - (Optional) The Amazon Resource Name (ARN) of the IAM role to be used for this target when the rule is triggered.
+* `role_arn` - (Optional) The Amazon Resource Name (ARN) of the IAM role to be used for this target when the rule is triggered. (Required if `ecs_target` is used.)
```
Ninir: Can you remove brackets?

website/docs/r/cloudwatch_event_target.html.markdown
```diff
@@ -68,11 +68,11 @@ The following arguments are supported:
 
 * `rule` - (Required) The name of the rule you want to add targets to.
 * `target_id` - (Optional) The unique target assignment ID.  If missing, will generate a random, unique id.
-* `arn` - (Required) The Amazon Resource Name (ARN) associated of the target.
+* `arn` - (Required) The Amazon Resource Name (ARN) associated of the target. (It should be the ID of the ECS Cluster if `ecs_target` is used.)
```
Ninir: As you exposed it in your issue, the `aws_ecs_cluster.*.id` attribute is the ARN. Since the related documentation exposes it, we should not redo the work here, especially since I'm agrai it will confuse people :/
The other reason is that there are so many edge cases in so many resources, that adding such indications would result in enormous documentation, which we will try to avoid 😄 

However, one thing that could be done is the normalization of the data source & resource, so that both uses the `arn` too. Does it seem reasonnable to you?

scream314: It does.

website/docs/r/cloudwatch_event_target.html.markdown
scream314: bccb22d

Ninir: Sorry, loved aussi those brackets, was meant to remove parenthesis, or `(`. Dunno bracket was the same for ` & `(` or `)`
So would go for `Required if `ecs_target` is used.`, simply :D

Sorry for the extra work ahah!

scream314: No worries.
Cheers.

PR #1489, opened 2017-08-23
resource/aws_instance: Add computed field for volume_id of block device

### LINKED ISSUE #1485: Feature request: please expose EBS volume ids in the aws_instance object
Hi there,

### Terraform Version
0.10.2

### Affected Resource(s)
aws_instance

### Expected Behavior
I would like to be able to get the volume ids of EBS volumes attached to an aws instance, including the root volume.

My specific use case is that I want to create a cloudwatch event to make snapshots of the volumes on a regular schedule, and this requires the volume ids.

### Actual Behavior
This information is not available.

rowleyaj:
Should resolve #1485.

Thought I would take a stab at adding this feature. I'm not sure how to add tests if any are required for this kind of change, pointers in the right direction there would be appreciated.

Manually tested using the below config:

```
# Create a new instance of the latest Ubuntu 14.04 on an
# t2.micro node with an AWS Tag naming it "HelloWorld"
provider "aws" {
  region = "us-east-1"
}

variable "subnet_id" {}

data "aws_ami" "ubuntu" {
  most_recent = true

  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
  }

  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }

  owners = ["099720109477"] # Canonical
}

  ami           = "${data.aws_ami.ubuntu.id}"
  instance_type = "t2.micro"
  subnet_id = "${var.subnet_id}"

  root_block_device {
    volume_type = "standard"
    volume_size = "8"
  }

  tags {
    Name = "HelloWorld"
  }
}

output "root_volume_id" {
  value = "${aws_instance.web.root_block_device.0.volume_id}"
}
```

cc @fields is this what you were looking for?

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -323,6 +323,11 @@ func resourceAwsInstance() *schema.Resource {
 							Computed: true,
 							ForceNew: true,
 						},
+
+						"volume_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
 					},
 				},
 				Set: func(v interface{}) int {
@@ -405,6 +410,11 @@ func resourceAwsInstance() *schema.Resource {
 							Optional: true,
 							Computed: true,
 							ForceNew: true,
+						},
+
+						"volume_id": {
+							Type:     schema.TypeString,
+							Computed: true,
 						},
 					},
 				},
@@ -1137,6 +1147,8 @@ func readBlockDevicesFromInstance(instance *ec2.Instance, conn *ec2.EC2) (map[st
 	for _, vol := range volResp.Volumes {
 		instanceBd := instanceBlockDevices[*vol.VolumeId]
 		bd := make(map[string]interface{})
+
+		bd["volume_id"] = *vol.VolumeId
 
 		if instanceBd.Ebs != nil && instanceBd.Ebs.DeleteOnTermination != nil {
 			bd["delete_on_termination"] = *instanceBd.Ebs.DeleteOnTermination
--- a/aws/resource_aws_instance_test.go
+++ b/aws/resource_aws_instance_test.go
@@ -254,12 +254,16 @@ func TestAccAWSInstance_blockDevices(t *testing.T) {
 						"aws_instance.foo", "ebs_block_device.#", "3"),
 					resource.TestCheckResourceAttr(
 						"aws_instance.foo", "ebs_block_device.2576023345.device_name", "/dev/sdb"),
+					resource.TestMatchResourceAttr(
+						"aws_instance.foo", "ebs_block_device.2576023345.volume_id", regexp.MustCompile("vol-[a-z0-9]+")),
 					resource.TestCheckResourceAttr(
 						"aws_instance.foo", "ebs_block_device.2576023345.volume_size", "9"),
 					resource.TestCheckResourceAttr(
 						"aws_instance.foo", "ebs_block_device.2576023345.volume_type", "standard"),
 					resource.TestCheckResourceAttr(
 						"aws_instance.foo", "ebs_block_device.2554893574.device_name", "/dev/sdc"),
+					resource.TestMatchResourceAttr(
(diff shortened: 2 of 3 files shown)


fields:
Assuming it does what it seems to do, yes - that would be exactly what I'm looking for.

fields:
There can only ever be one root block device, so making that an array seems a little weird to me, but that's not a huge problem.

rowleyaj:
@fields I believe it's always been an array/list, set to a maximum of 1 item though. Thats in the code here: https://github.com/rowleyaj/terraform-provider-aws/blob/master/aws/resource_aws_instance.go#L372-L376

The other ebs volumes also have the ids exported in the same way.

kwilczynski:
@rowleyaj looks good to me! But we need to wait for someone like @radeksimko to make the final call.

rowleyaj:
Thanks @grubernaut. Thanks for taking the time to review my attempt here :)

I took a look at https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance_test.go and I think I worked out how to add an acceptance test for a static value and check the resource attributes (by changing the config and using `resource.TestCheckResourceAttr`, but how would I go about adding an acceptance test for this where I need to know the id of the volume beforehand to confirm it is set correctly?

Do I need to create a volume in a similar way to this: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance_test.go#L60 and then use `resource.TestCheckResourceAttr` in a similar way to this: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance_test.go#L78 but with vol.VolumeId passed in to check?

I can try and take a stab at this tomorrow. One important question, how do I run the acceptance tests, ideally just the one I write/edit so I don't have to spin up everything. I'm guessing that https://github.com/hashicorp/terraform/blob/master/.github/CONTRIBUTING.md#running-an-acceptance-test still applies even though the providers are now split out, and it will just require standard AWS credentials to with permission to create an instance + volume?

Sorry for all the questions.

radeksimko:
> I think I worked out how to add an acceptance test for a static value and check the resource attributes (by changing the config and using resource.TestCheckResourceAttr, but how would I go about adding an acceptance test for this where I need to know the id of the volume beforehand to confirm it is set correctly?

You have basically two options:

`resource.TestCheckResourceAttrSet` can check whether the field was set or not (simple empty/non-empty check). See example at https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_alb_test.go#L71-L74

`resource.TestMatchResourceAttr` provides checking via regexp, so you can check whether the field matches something like `vol-[a-z0-9]+`. See example at https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_alb_test.go#L135-L136

I don't personally have a preference - so pick the one you like.

> Do I need to create a volume in a similar way to this

Unless I'm mistaken you shouldn't need to. You just need to pick an AMI which has a volume baked into it. See example at https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance_test.go#L196-L284

> how do I run the acceptance tests, ideally just the one I write/edit

We don't have a provider-specific guide (yet) for running acceptance tests, but it indeed works pretty much the same way it did prior to provider split, e.g.

```
AWS_PROFILE=xyz make testacc TEST=./aws TESTARGS='-run= TestAccAWSInstance_blockDevices'
```

Ninir:
Hi @rowleyaj

Do you feel you could have time to finish this? let us know if you need any help here! :)

rowleyaj:
Sorry @Nini - I completely dropped the ball on this one. I've added acceptance tests now, for both my change and the ebs volume_id as I didn't see any for that either. Thanks for the pointers how to do this @radeksimko.

rowleyaj:
Sorry @Ninir - I completely dropped the ball on this one. I've added acceptance tests now, for both my change and the ebs volume_id as I didn't see any for that either. Thanks for the pointers how to do this @radeksimko.

bflad:

PR #1492, opened 2017-08-24
Add importer to aws_iam_user_login_profile

mcanevet:


Ninir:
Hey @mcanevet 

Thanks for this :)
Could you also add an acceptance test, so that we can better follow the addition in the future?

Thanks!

mcanevet:
@Ninir I just added an acceptance test using copy/paste/replace driven development, but to be honest I didn't test it because I don't known how to do it.

Ninir:
Hey @mcanevet !

After looking into details, I'm not sure this will allow us to get everything we wan. When you create a login profile, you can only get the following attributes:
* CreateDate
* PasswordResetRequired
* UserName

Since the initial password is not retrievable from the API (nor the encrypted password), we can't set the `key_fingerprint` attribute which is based on the (initial) password, nor the `encrypted_password`, etc.

Also, we can't set the `gpg_Key` attribute since we don't know how it is attached to your user: perhaps your are using your own PGP key, or maybe it is a keybase-based one...

At this point, it seems not possible to allow such import. Thoughts?

mcanevet:
@Ninir yes I'm aware of that.
I'm using 
```hcl
resource "aws_iam_user_login_profile" "foo" {
  user = "foo"
  pgp_key = "keybase:foo"

  lifecycle {
    ignore_changes = ["password_length", "password_reset_required", "pgp_key", "user"]
  }
}
```
To prevent this.
This is not ideal though...

Ninir:
Indeed, not ideal 😄 

This resource is neither readable nor updatable, so one very specific use-case, sadly :/
In fact, what would be the purpose for such import, since you would need to fill all the details on your own?

mcanevet:
@Ninir I just want to import what exists declaring this resource in a generic module so that future declaration of this module will create the user login profiles.

Ninir:
@mcanevet I don't really see the point for import in your use-case. Could you elaborate a bit more please?

To be honnest, I think that this is one of the resource that is hard to import: you will need to set every details on your own (as exposed earlier), so this does not bring any value to me here...

Having an import just for the sake of having imports here seems unreasonable to me, but... just my opinion 🤷‍♂️

mcanevet:
@Ninir what would be the impact of applying a `aws_iam_user_login_profile` on an already exisiting account? Would it reinitialize the password?

Ninir:
@mcanevet You would get a `EntityAlreadyExists` exception, so no changes on the current object.
It would not reinitialize the password though! :)

The general point of importing is fine, however in our case we would need to fill so much details on our own it would not be useful I think, especially if yu don't have those details... 

Let me ask @radeksimko's idea on that :)

radeksimko:
Hi folks,
I'd be usually inclined to support importing resources even when we can't get all fields from the API because the user can fill in the gaps as presumably they are the ones who created those resources and use them, so they are likely going to know the details.

That said there are exceptions to that, and I'm inclined to what @Ninir said - i.e. that value of such import is questionable.

Both password reset and initial (encrypted) password are set during initial creation and then become un-recoverable (un-importable), so the only use case left is "you can import the login profile to either `taint` it or `destroy` it".

@mcanevet Do you believe this scenario is common enough that it's worth supporting?

The downside of supporting this is that some users may be confused and think they can easily import the login profile which clearly isn't true. Documentation may help, but sometimes it's better to just avoid confusing behaviours in the first place.

> what would be the impact of applying a aws_iam_user_login_profile on an already exisiting account? Would it reinitialize the password?

I _think_ as long as you set `gpg_key` to any non-empty string it _should_ be noop - that's what I can tell by looking at the schema, but someone would need to check for real.

mcanevet:
@radeksimko as long as I can create the resources for existing accounts I can live without import.
I think we can close this.

radeksimko:
@mcanevet I think there are two situations you can be in which may cause the confusion and drift between mine & @Ninir 's response - it _shouldn't_ matter where/how you manage the `aws_iam_user` (manually, within the same TF config or elsewhere) - as long as it exists at the time of creation `aws_iam_user_login_profile` then you're 🆗 . All that login profile resource cares about is username.

I'm not sure what happens when the user already has a login profile though - it may error out as @Ninir mentioned.

Issue #1262, opened 2017-07-27
data source for aws_s3_bucket please


gtmtech:
Terraform 0.9.8

Strange that there is a data source for aws_s3_bucket_object but not one for aws_s3_bucket yet...

Thanks!

radeksimko:
Hi @gtmtech 
do you mind sharing a use-case for this particular potential data source?

Thanks.

gtmtech:
Sure... 

In a security-segregated terraform setup, one team should be responsible for all aws policies. This includes IAM policies, KMS policies, and Bucket policies. This team should therefore either be in charge of buckets + bucket policies, or just the policies. Because policy is an attribute on bucket, they need to be in charge of buckets & bucket policies, and therefore they manage the aws_s3_bucket resource. 

Another team is a user of the bucket. They wish to manage a DNS entry for a static website to that bucket. They wish to create an aws_route53_record , type=A, with alias{} section referencing name, zone_id of the bucket. These are available through parameters website_domain, and hosted_zone_id on the bucket resource, however they cannot create a datasource to the aws_s3_bucket in order to work out what these parameters are. They end up having to hardcode values as a workaround which is bad. 

Modules does not solve this BTW, because in a security-segregated environment, the second team does not have the credentials to refresh the resources in the modules/submodules for good security practice- i.e. the second team is not entitled to know what the policies are on the buckets, kms keys, iam roles, they just have to work under them.

gtmtech:
Or you can just ignore all the security stuff above, and just say - from the teams perspective there is a bucket not managed by terraform, for which they wish to create a corresponding aws_route53_record resource object pointing to that bucket - in order to do so , they want a datasource to be able to read attributes of that bucket

radeksimko:
Doesn't shared remote state solve this problem also?

https://www.terraform.io/docs/providers/terraform/d/remote_state.html

gtmtech:
No, it requires access to the remote state which would not be available due to security restrictions. E.g. team A has given team B permission to s3:Get* and s3:List* on bucket X, so a data resource would be able to retrieve the bucket attributes that might be useful in setting up a route53_record. However, team A has obviously not given team B permission to see team A's statefile which is full of passwords and sensitive information. 

But in any case, if thats the answer, why have any data sources at all? There are loads of useful ones, instances, volumes, odd that bucket is not in the list.

radeksimko:
[...]

PR #1505, opened 2017-08-25
Add s3 bucket data source

kl4w:
Started creating s3 bucket data source. Solves for #1262.

Please let me know if I'm missing anything.


kl4w:
@radeksimko 
Ahh cheers, that's perfect.
I'll keep an eye out for that in the future.

Thank you very much

PR #1520, opened 2017-08-28
New ssm parameters features

jocgir:
Resubmit of the PR #1295.

I made some changes on the aws_ssm_parameter resource.

I added support for the new features (description, tags and allowed pattern).
I also fix some problems with the current implementation. If a parameter was manually deleted from the parameter store, that causes problems with terraform plan, apply and destroy. A resource that no longer exists must not be considered as an error.

I also changed the default value for overwrite. Forcing the value to be false by default is not very useful in a terraform project. If we wish to preserve the value, we should use the default lifecycle rules to protect manually changed value. Setting this new parameter to false by default causes breaking change in the already written code since this parameter did not exists before and the values were overwritten by default.


pmorton:
@jocgir +1 on reverting the overwrite behavior of overwrite. I was perplexed to find the original implementation reverted. You reasoning is how it was originally implanted. The lifecycle rules you introduced make sense to me. 

Out of curiosity, on the allowed pattern, what is the use case for this? Pattern validation seems to be a concern that superceeds the responsibility of the aws provider. My spidy sense is that this feature may hold up many of the other great improvements in his PR.

 IHMO if you need to validate variable input, we should have a generic provider that takes a pattern and an input and emits the validated value. That way various strategies can be implemented in a provider agnostic way.

jocgir:
@pmorton, The allowed pattern has been added by aws to enable custom validation when the parameter is edited manually through the console. I just added support to be able to define that parameter when defining parameter store resource.

jocgir:
I noticed a problem when using lifecycle rule to ignore change on the value. The resource is destroyed and replaced. I must investigate this behaviour to check what is the issue.

Please, do not merge until confirmation.

pmorton:
@jocgir BTW I created a pull request that makes the update/overwrite lifecycle more robust. #1556

jocgir:
OK, I made several changes to ensure proper lifecycle rules management.

I integrated the idea from @pmorton to set the overwrite parameter to false by default on new resource, but I changed the behaviour a bit to let terraform handle the overwrite on already existing resource through it's normal lifecycle rules.

I did not see any reason why changing the encryption key or the type should replace the resource. Parameters values are importants and I think we must try to protect them as much as possible.

jocgir:
Hi @Ninir,

Is it possible to check if that PR could be integrated it the the next release. It has been submitted several week ago. We use it for a while and we did not face any issue.

I just merged in some changes that have been made by my colleague @julienduchesne to avoid conflit since its PR has been merged very quickly by @radeksimko.

We also have some other PR also waiting in the pipeline for a while #1945, #1745.

Regards,

Jo
Coveo Team

pccowboy:
@Ninir @radeksimko FWIW, we have had this PR integrated into our custom AWS provider for a little over a month now, and it is operating as expected.  Any idea when it might make it to a review?

bflad:
Hi @jocgir it looks like `aws/resource_aws_ssm_parameter_test.go` is conflicting at the moment, can you please resolve that so we can start reviewing this? Thanks and sorry for the delay in getting you a review. 😅

jocgir:
I fixed the conflict. Thank you for the review.

bashims:
Hi,

Any idea when this PR will be merged? We would love to take advantage of the new features, especially tags!

Thanks,

-ben

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,45 +52,88 @@ func resourceAwsSsmParameter() *schema.Resource {
...
-	paramInput := &ssm.GetParametersInput{
-		Names: []*string{
-			aws.String(d.Id()),
-		},
+	resp, err := ssmconn.GetParameters(&ssm.GetParametersInput{
+		Names:          []*string{aws.String(d.Id())},
 		WithDecryption: aws.Bool(true),
+	})
+	if err != nil {
+		return errwrap.Wrapf("[ERROR] Error getting SSM parameter: {{err}}", err)
```
bflad: Minor nitpick: We don't need to use `errwrap.Wrapf` normally if we are just returning an error message, can simply be `return fmt.Errorf("error getting SSM parameter: %s", err)`

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,45 +52,88 @@ func resourceAwsSsmParameter() *schema.Resource {
...
+				Key:    aws.String("Name"),
+				Values: []*string{aws.String(d.Get("name").(string))},
+			},
+		},
+	})
 	if err != nil {
 		return errwrap.Wrapf("[ERROR] Error describing SSM parameter: {{err}}", err)
 	}
 
-	if len(resp.Parameters) == 0 {
```
bflad: Is there a reason why this logic was removed? We should force recreation of the parameter if its missing or at least prevent a Terraform crash by ensuring we won't get an index out of bounds error if `resp.Parameters` has no elements.

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,45 +52,88 @@ func resourceAwsSsmParameter() *schema.Resource {
...
-		log.Printf("[WARN] SSM Param %q not found, removing from state", d.Id())
-		d.SetId("")
-		return nil
+	detail := respDetailed.Parameters[0]
+	if detail.Description != nil {
+		// Trailing spaces are not considered as a difference
+		*detail.Description = strings.TrimSpace(*detail.Description)
+	}
+	if _, ok := d.GetOk("key_id"); !ok && detail.KeyId != nil && *detail.KeyId == "alias/aws/ssm" {
```
bflad: We should be able to avoid this logic by adding `Computed: true` to the attribute.

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,45 +52,88 @@ func resourceAwsSsmParameter() *schema.Resource {
...
+	})
 	if err != nil {
 		return errwrap.Wrapf("[ERROR] Error describing SSM parameter: {{err}}", err)
 	}
 
-	if len(resp.Parameters) == 0 {
-		log.Printf("[WARN] SSM Param %q not found, removing from state", d.Id())
-		d.SetId("")
-		return nil
+	detail := respDetailed.Parameters[0]
```
bflad: We should ensure there is a check for `len(respDetailed.Parameters) == 0` that at least returns an error otherwise it can cause a Terraform crash.

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,45 +52,88 @@ func resourceAwsSsmParameter() *schema.Resource {
...
 			},
 			"overwrite": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
 			},
+			"allowed_pattern": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "",
```
bflad: Minor nitpick: you can remove this, `""` is the default for strings.

aws/resource_aws_ssm_parameter.go
```diff
@@ -124,24 +169,47 @@ func resourceAwsSsmParameterPut(d *schema.ResourceData, meta interface{}) error
...
+		Type:           aws.String(d.Get("type").(string)),
+		Value:          aws.String(d.Get("value").(string)),
+		Overwrite:      aws.Bool(shouldUpdateSsmParameter(d)),
+		AllowedPattern: aws.String(d.Get("allowed_pattern").(string)),
 	}
+
+	if description, ok := d.GetOk("description"); ok {
+		paramInput.SetDescription(description.(string))
+	} else if d.HasChange("description") {
+		// There is a "bug" in the AWS API and is it not possible to unset a description once
```
bflad: Is this still the case? It sounds like it will cause a perpetual diff if this logic is left as-is (e.g. always showing `description "" => " "` when its removed until a configuration is re-updated to `description = " "` instead). What happens if you set `description = ""` in your configuration?

In either case if the AWS API bug still exists, we should instead force a new resource and document the behavior in the resource documentation.

aws/resource_aws_ssm_parameter_test.go
```diff
@@ -273,20 +273,33 @@ resource "aws_ssm_parameter" "foo" {
 func testAccAWSSSMParameterBasicConfigOverwrite(rName, pType, value string) string {
 	return fmt.Sprintf(`
 resource "aws_ssm_parameter" "foo" {
-  name  = "%s"
-  type  = "%s"
-  value = "%s"
+  name  = "test_parameter-%[1]s"
+  description  = "description for parameter %[1]s"
```
bflad: Since `description` is an optional field, we should probably give it its own acceptance test for adding it, updating it, and removing it in configuration. Especially given the API bug mentioned in the code.

julienduchesne:
Updated the PR @bflad

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,35 +51,47 @@ func resourceAwsSsmParameter() *schema.Resource {
 			"key_id": {
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
```
bflad: Is `key_id` modifiable now? Can you please update `TestAccAWSSSMParameter_secure_with_key` to test this?

aws/resource_aws_ssm_parameter.go
```diff
@@ -47,35 +51,47 @@ func resourceAwsSsmParameter() *schema.Resource {
...
 
+	log.Printf("[DEBUG] Reading SSM Parameter: %s", d.Id())
+
+	resp, err := ssmconn.GetParameters(&ssm.GetParametersInput{
+		Names:          []*string{aws.String(d.Id())},
+		WithDecryption: aws.Bool(true),
+	})
 	if err != nil {
-		return errwrap.Wrapf("[ERROR] Error describing SSM parameter: {{err}}", err)
+		return errwrap.Wrapf("[ERROR] Error getting SSM parameter: {{err}}", err)
```
bflad: Minor nitpick: It's currently in a lot of the codebase 😢 , but we do not need `errwrap` for returning simple error strings. `fmt.Errorf("error getting SSM parameter: %s", err)` is fine here

aws/resource_aws_ssm_parameter.go
```diff
@@ -87,6 +103,48 @@ func resourceAwsSsmParameterRead(d *schema.ResourceData, meta interface{}) error
...
+		},
+	}
+	detailedParameters := []*ssm.ParameterMetadata{}
+	err = ssmconn.DescribeParametersPages(describeParamsInput,
+		func(page *ssm.DescribeParametersOutput, lastPage bool) bool {
+			detailedParameters = append(detailedParameters, page.Parameters...)
+			return !lastPage
+		})
+	if err != nil {
+		return errwrap.Wrapf("[ERROR] Error describing SSM parameter: {{err}}", err)
```
bflad: Minor nitpick: same here with `fmt.Errorf()`

aws/resource_aws_ssm_parameter.go
```diff
@@ -87,6 +103,48 @@ func resourceAwsSsmParameterRead(d *schema.ResourceData, meta interface{}) error
...
+	}
+	if len(detailedParameters) == 0 {
+		log.Printf("[WARN] SSM Param %q not found, removing from state", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	detail := detailedParameters[0]
+	if detail.Description != nil {
+		// Trailing spaces are not considered as a difference
```
bflad: Does this still apply? If so, looks like we should move this to a `DiffSuppressFunc` on the attribute to make the behavior more clear

aws/resource_aws_ssm_parameter.go
```diff
@@ -124,24 +179,45 @@ func resourceAwsSsmParameterPut(d *schema.ResourceData, meta interface{}) error
...
-		Overwrite: aws.Bool(d.Get("overwrite").(bool)),
+		Name:           aws.String(d.Get("name").(string)),
+		Type:           aws.String(d.Get("type").(string)),
+		Value:          aws.String(d.Get("value").(string)),
+		Overwrite:      aws.Bool(shouldUpdateSsmParameter(d)),
+		AllowedPattern: aws.String(d.Get("allowed_pattern").(string)),
+	}
+
+	if description, ok := d.GetOk("description"); ok {
+		paramInput.SetDescription(description.(string))
```
bflad: We should probably avoid using the `Set...()` methods of the AWS SDK as they are likely going away in v2 of the SDK: `paramInput.Description = description.(string)`

That said, this is always writing a value to the description parameter and should be able to match similar code of other resources:

```go
if d.HasChange("description") {
  _, n := d.GetChange("description")
  paramInput.Description = aws.String(n.(string))
}
```

In the case where its removed from the configuration, it should still write "". Just something to consider.

aws/resource_aws_ssm_parameter.go
```diff
@@ -124,24 +179,45 @@ func resourceAwsSsmParameterPut(d *schema.ResourceData, meta interface{}) error
...
-	_, err := ssmconn.PutParameter(paramInput)
-
-	if err != nil {
+	log.Printf("[DEBUG] Waiting for SSM Parameter %v to be updated", d.Get("name"))
+	if _, err := ssmconn.PutParameter(paramInput); err != nil {
 		return errwrap.Wrapf("[ERROR] Error creating SSM parameter: {{err}}", err)
 	}
 
+	if err := setTagsSSM(ssmconn, d, d.Get("name").(string), "Parameter"); err != nil {
+		return errwrap.Wrapf("[ERROR] Error creating SSM parameter tags: {{err}}", err)
```
bflad: Minor nitpick: same here with `fmt.Errorf()`

aws/resource_aws_ssm_parameter.go
julienduchesne: Yes it works well, I tested it manually but I'll be adding tests of course.

aws/resource_aws_ssm_parameter.go
julienduchesne: This actually was a hack due to the API bug which is apparently resolved. Removing.

julienduchesne:
Here you go @bflad. Thanks for the review. Let me know if there's anything else that I should change!

aws/resource_aws_ssm_parameter_test.go
```diff
@@ -273,26 +331,88 @@ resource "aws_ssm_parameter" "foo" {
...
+}
+
+func testAccAWSSSMParameterSecureConfigWithKey(rName string, value string, keyAlias string) string {
+	return fmt.Sprintf(`
+resource "aws_ssm_parameter" "secret_foo" {
+  name  = "test_secure_parameter-%[1]s"
+  description  = "description for parameter %[1]s"
+  type  = "SecureString"
+  value = "%[2]s"
+  key_id = "alias/%[3]s"
```
bflad: Minor nitpick: `${aws_kms_alias.test_alias.name}` is a little easier to understand here instead of depends_on 😉

bflad:

PR #1540, opened 2017-08-30
Fixed recreation of EIP when it does not exist anymore

### LINKED ISSUE #1529: EIP deletion is not detected, unable to recover via taint
### Terraform Version

Terraform v0.10.2
AWS provider v0.1.4

### Affected Resource(s)
- aws_eip

```hcl
resource "aws_eip" "prod-otacontrol" {
  vpc = true
}
```

### Expected and Actual Behavior

When the resource is deleted externally, `terraform refresh` should pick this up and recreate the resource. However the `InvalidAllocationID.NotFound` for the EIP is silently ignored and Terraform assumes the EIP still exists.

Trying to remedy the situation manually via `terraform taint` also does not work; When attempting to delete the (non-existant) EIP, Terraform keeps retrying to perform a `ReleaseAddress` indefinitely.

### Debug Output

https://gist.github.com/ksperling/f5aeec3b59dd2f423904df58a8801ff0

### Steps to Reproduce

1. Create an EIP via Terraform (plan / apply)
2. Delete the EIP via AWS console
3. Terraform refresh / plan. Observe EIP deletion is not detected.
4. Attempt to recover via taint / plan / apply also does not work.

### References

The bug seems to have been introduced in this commit e4244f8b481f2796f06e443a413510468fa765ec as part of #1053 

When the EIP is not found in the "existing resource" code path, `d.SetId("")` needs to be called.

Ninir:
Fixes #1529

This missing bit was not marking the resource as needing recreation, when it should have been.
In case the EIP was created and removed from the console, this read was simply warning in the logs, but never doing anything more.

### DIFF
--- a/aws/resource_aws_eip.go
+++ b/aws/resource_aws_eip.go
@@ -159,6 +159,7 @@ func resourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
 			if ok && (awsErr.Code() == "InvalidAllocationID.NotFound" ||
 				awsErr.Code() == "InvalidAddress.NotFound") {
 				log.Printf("[WARN] EIP not found, removing from state: %s", req)
+				d.SetId("")
 				return nil
 			}
 			return err


Ninir:
Closing since #1551 has a nice way to handle the acceptance test.

PR #1546, opened 2017-08-30
Fix insecure aws_sns_topic_policy example

### LINKED ISSUE #1545: aws_sns_topic_policy example allows any AWS account to publish?
I spotted that someone on a project of ours had set up an SNS policy that allowed any account to access the SNS topic, and wondered why they had configured it that way... Turns out it was a copy from the example for aws_sns_topic_policy in the Terraform documentation! It should probably be changed to something a little less dangerous security-wise...

edmundcraske:
Fixes #1545

Also add note to clarify AWS's behaviour when setting a principal of just an account ID, causing terraform to always see a diff.

### DIFF
--- a/website/docs/r/sns_topic_policy.html.markdown
+++ b/website/docs/r/sns_topic_policy.html.markdown
@@ -10,6 +10,8 @@ description: |-
 
 Provides an SNS topic policy resource
 
+~> **NOTE:** If a Principal is specified as just an AWS account ID rather than an ARN, AWS silently converts it to the ARN for the root user, causing future terraform plans to differ. To avoid this problem, just specify the full ARN, e.g. `arn:aws:iam::123456789012:root`
+
 ## Example Usage
 
 ```hcl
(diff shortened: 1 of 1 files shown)


edmundcraske:
Thought I'd also move from inline json to using the aws_iam_policy_document data source, hope that makes sense :)

website/docs/r/sns_topic_policy.html.markdown
```diff
@@ -10,36 +10,59 @@ description: |-
 
 Provides an SNS topic policy resource
 
+~> **NOTE:** If a Principal is specified as just an AWS account ID rather than an ARN, AWS silently converts it to the ARN for the root user, causing future terraform plans to differ. To avoid this problem, just specify the full ARN, e.g. `arn:aws:iam::123456789012:root`
```
radeksimko: It's good to call it out, but unfortunately this is a problem that appears in all kinds of policies of all AWS resources - so SNS Topic Policy isn't special in this case. 😑

PR #1547, opened 2017-08-30
r/aws_api_gateway_authorizer: URI must contain region name

elliotwms:
Was having an issue with gateway authorizer permissions and it took a while to track this down but it looks like the authorizer URI needs to contain the region it's being created in. The resource will be created but it won't work properly until you go in and save it, at which point the authorizer will be able to be called properly.

### DIFF
--- a/website/docs/r/api_gateway_authorizer.html.markdown
+++ b/website/docs/r/api_gateway_authorizer.html.markdown
@@ -16,7 +16,7 @@ Provides an API Gateway Authorizer.
 resource "aws_api_gateway_authorizer" "demo" {
   name                   = "demo"
   rest_api_id            = "${aws_api_gateway_rest_api.demo.id}"
-  authorizer_uri         = "arn:aws:apigateway:region:lambda:path/2015-03-31/functions/${aws_lambda_function.authorizer.arn}/invocations"
+  authorizer_uri         = "arn:aws:apigateway:${var.myregion}:lambda:path/2015-03-31/functions/${aws_lambda_function.authorizer.arn}/invocations"
   authorizer_credentials = "${aws_iam_role.invocation_role.arn}"
 }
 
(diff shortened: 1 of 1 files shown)


website/docs/r/api_gateway_authorizer.html.markdown
```diff
@@ -16,7 +16,7 @@ Provides an API Gateway Authorizer.
 resource "aws_api_gateway_authorizer" "demo" {
   name                   = "demo"
   rest_api_id            = "${aws_api_gateway_rest_api.demo.id}"
-  authorizer_uri         = "arn:aws:apigateway:region:lambda:path/2015-03-31/functions/${aws_lambda_function.authorizer.arn}/invocations"
+  authorizer_uri         = "arn:aws:apigateway:${var.myregion}:lambda:path/2015-03-31/functions/${aws_lambda_function.authorizer.arn}/invocations"
```
Ninir: I think it is worth providing the `aws_lambda_function.authorizer.invoke_arn` attribute here.
This is an attribute exposed by Terraform, based on a helper we created in order to avoid constructing the ARN on your own, and falling into issues like the one you provided! :)

Here is the code behind:
```go
func buildLambdaInvokeArn(lambdaArn, region string) string {
	apiVersion := "2015-03-31"
	return fmt.Sprintf("arn:aws:apigateway:%s:lambda:path/%s/functions/%s/invocations",
		region, apiVersion, lambdaArn)
}
```

As you can see, it uses the Lambda ARN plus the region we retrieved from the provider (either local, defined in env var, etc).

Tell me what you think!

elliotwms: Yep this makes a lot more sense and seems to do exactly what it's supposed to! I'll update the PR to include this

PR #1548, opened 2017-08-30
r/aws_sns_topic: Provider now errors with fake iam role

mbfrahry:
This PR is a migration from https://github.com/hashicorp/terraform/pull/14226 and fixes https://github.com/terraform-providers/terraform-provider-aws/issues/738. 

```
make testacc TEST=./aws TESTARGS="-run=TestAccAWSSNSTopic_withFakeIAMRole"
TF_ACC=1 go test ./aws -v -run=TestAccAWSSNSTopic_withFakeIAMRole -timeout 120m
=== RUN   TestAccAWSSNSTopic_withFakeIAMRole
--- PASS: TestAccAWSSNSTopic_withFakeIAMRole (8.34s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	8.375s
```

### DIFF
--- a/aws/resource_aws_sns_topic.go
+++ b/aws/resource_aws_sns_topic.go
@@ -147,7 +147,7 @@ func resourceAwsSNSUpdateRefreshFunc(
 				// if the error contains the PrincipalNotFound message, we can retry
 				if strings.Contains(awsErr.Message(), "PrincipalNotFound") {
 					log.Printf("[DEBUG] Retrying AWS SNS Topic Update: %s", params)
-					return nil, "retrying", nil
+					return nil, "retrying", err
 				}
 			}
 			return nil, "failed", err
--- a/aws/resource_aws_sns_topic.go
+++ b/aws/resource_aws_sns_topic.go
@@ -9,7 +9,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/sns"
 	"github.com/hashicorp/errwrap"
-	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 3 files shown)


aws/resource_aws_sns_topic.go
```diff
@@ -147,7 +147,7 @@ func resourceAwsSNSUpdateRefreshFunc(
 				// if the error contains the PrincipalNotFound message, we can retry
 				if strings.Contains(awsErr.Message(), "PrincipalNotFound") {
 					log.Printf("[DEBUG] Retrying AWS SNS Topic Update: %s", params)
-					return nil, "retrying", nil
+					return nil, "retrying", err
```
catsby: Returning `err` here instead of `nil` will cause this WaitForState method to quit immediately. We have to return `nil` here because the _intent_ is to retry right here:

- https://github.com/terraform-providers/terraform-provider-aws/blob/068ff1beeb935dc6034b52a6ee38cef9f6fd60bd/vendor/github.com/hashicorp/terraform/helper/resource/state.go#L41

I agree with Radek, reworking this into `resource.Retry` may be better. I realize this isn't your code or change so let me know if you want to hand it off. Until `StateChangeConf` can surface "error before last", we're in this spot.

mbfrahry:
Thanks for pointing me towards that retry function!

radeksimko:
Thanks for making those changes, `retryOnAwsCode` is probably even better.

Do you mind replacing this in the two other operations (`resourceAwsSnsTopicPolicyUpsert` & `resourceAwsSnsTopicPolicyDelete`) too & removing `resourceAwsSNSUpdateRefreshFunc` altogether?

mbfrahry:
Alright! It's all gone and tests are passing

Issue #1004, opened 2017-06-29
aws_ssm_parameter resource does not have a overwrite option


FransUrbo:
I'm trying to update/replace a *Parameter Store* value, but TF gives `ParameterAlreadyExists`.

### Terraform Version
0.9.8

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_ssm_parameter

FransUrbo:
I notice after I've deleted the param, run `terraform apply` and then ran `terraform apply` again some time later, that it can't detect that the parameter already exists!

So a better option (because I'm encrypting the parameter with a KMS key and I'm not sure TF can get the param, decrypt it with the key - how would it know WHICH key?? - and then check if it differs from what is in the state or the `value` option) might be to simply try see if the param already exists, and then ignore it (as in, not to create one).

FransUrbo:
Another issue:

If the `name` value is changed (not sure if it's related to the param also been deleted), I get:
```
[ERROR] SSM Parameter XXX is invalid
```

stack72:
@FransUrbo if possible, can you try and recreate that issue and then open an alternative about that?

FransUrbo:
Ok, no worries!

Regarding the "can't detect already exists" above. This MIGHT be because I had the same resource name (but different `name` value) in several modules.

As in:
```
resource "aws_ssm_parameter" "hiera-private" {
    [...]
}
```
in all my modules. Prefixing the resource name, seems to have solved the "recreation issue"...

FransUrbo:
PS. Please back port to 0.9 if possible!

stack72:
Hi @FransUrbo 

there will be no further releases of the 0.9 services I'm afraid

Paul

FransUrbo:
Ok, bummer :(. So 0.10 is starting to be ok? Last I saw it was just a beta release... ?

stack72:
@FransUrbo it was the first release in the major version series - it will take a few betas before we GA it

FransUrbo:
If that's so, I'd very much like you to continue supporting 0.9. I'm not ready (i.e., brave/stupid enough), to hand my whole production environment to beta code!

stack72:
Sorry, we are only making critical big fixes to 0.9.x series. Ann
development effort is focused on 0.10 series now

It will be ready for GA in a few weeks

Paul

FransUrbo:
[...]

Issue #1481, opened 2017-08-23
Not possible to allow SSM Parameter overwrites only for updates (and not creates)


smphhh:
### Terraform Version
Terraform v0.10.2

### Affected Resource(s)
- aws_ssm_parameter

### Expected Behavior
Should be possible to update the values of existing parameters while making sure that Terraform won't change the values of parameters created outside Terraform.

### Actual Behavior
With `overwrite = false` Terraform won't allow changing the values of any parameter. With `overwrite = true` Terraform allows changing the values of existing Terraform parameters but also allows adding a parameter to the configuration with a name of a parameter that was created outside Terraform (or even another parameter created by Terraform elsewhere).

### Important Factoids
I think the default behavior of the SSM parameter resource should be to have `overwrite = false` for API calls for resource creation and `overwrite = true` for resource update API calls. This would make the behavior consistent with pretty much every other AWS Terraform resource. With that behavior I'm not sure we really even need the overwrite parameter exposed in the Terraform API at all.

In fact it seems that we had exactly the above behavior before [this](https://github.com/terraform-providers/terraform-provider-aws/commit/f446bc6860ddbcb86098e2cd0e124c2db2f7720f). I don't really understand the issue referred to in that commit, you shouldn't get `ParameterAlreadyExists` errors with the above behavior unless your state gets corrupted somehow.

smphhh:
Fixed in #1520.

PR #1556, opened 2017-08-31
Allow parameters to follow standard lifecycle rules

pmorton:
This pull request refines the commits that were merged in response to #1004 and #1006. When 1006 was merged it introduced a new flag which allows the user to overwrite an existing parameter. I understand that this can be helpful if the parameter already exists in your system, however the introduction of this flags has broken the update lifecycle. Specifically the way that updates interacts with overwrite is non-sensical. Under the introduced in 1006 you must have overwrite set to true if the value of that parameter needs to be updated. This creates an unsafe scenario in that a new resource will overwrite an existing value that it did not create _and_ this should throw an error unless the user decides that it should always be overwritten. 

I am proposing the following the following rules to govern update.
1. If the user has specified the overwrite flag, overwrite the parameters value.
2. If the resource is new, do not overwrite the parameters value. This will result in an error if the parameter already exists and is not managed by terraform. By default we do not want to overwrite parameters which we don't not own.
3. All other cases, allow the parameter to be overwritten.

### DIFF
--- a/aws/resource_aws_ssm_parameter.go
+++ b/aws/resource_aws_ssm_parameter.go
@@ -42,7 +42,6 @@ func resourceAwsSsmParameter() *schema.Resource {
 			"overwrite": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
 			},
 		},
 	}
@@ -101,12 +100,11 @@ func resourceAwsSsmParameterPut(d *schema.ResourceData, meta interface{}) error 
 	ssmconn := meta.(*AWSClient).ssmconn
 
 	log.Printf("[INFO] Creating SSM Parameter: %s", d.Get("name").(string))
-
 	paramInput := &ssm.PutParameterInput{
 		Name:      aws.String(d.Get("name").(string)),
 		Type:      aws.String(d.Get("type").(string)),
 		Value:     aws.String(d.Get("value").(string)),
-		Overwrite: aws.Bool(d.Get("overwrite").(bool)),
+		Overwrite: aws.Bool(shouldUpdateSsmParameter(d)),
 	}
 	if keyID, ok := d.GetOk("key_id"); ok {
 		log.Printf("[DEBUG] Setting key_id for SSM Parameter %s: %s", d.Get("name").(string), keyID.(string))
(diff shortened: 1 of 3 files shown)


smphhh:
I created an issue (#1481) earlier with the same proposed behavior so :+1:.

pmorton:
@Ninir I was wondering if this PR would be within your domain to review? The PR is pretty straight forward and passes unit and acceptance tests.

Ninir:
Hi @pmorton 

I have to admit I love the idea and your implementation :) 👍 

Here is what I did:
1. Create a parameter in the store with name `test_parameter` and value `foo` in the console
2. Create the `aws_ssm_parameter` resource with the same name but value `bar`, without the overwrite argument
3. Ran the apply: resource is created BUT the initial parameter is overridden, even though it was new on Terraform.
4. Expected and had the ParameterAlreadyExists exception
5. Updated the overwrite configuration with overwrite = false, and still expected and had the ParameterAlreadyExists exception
6. Updated the overwrite configuration with overwrite = true, and still expected not to have the ParameterAlreadyExists exception, and didn't have it

This seems like the proper way to do it, just want to clarify everything :)

Thanks!

aws/resource_aws_ssm_parameter.go
```diff
@@ -124,3 +122,20 @@ func resourceAwsSsmParameterPut(d *schema.ResourceData, meta interface{}) error
 
 	return resourceAwsSsmParameterRead(d, meta)
 }
+
+func shouldUpdateSsmParameter(d *schema.ResourceData) bool {
+	// If the user has specifed a preference, return their preference
+	if value, ok := d.GetOkExists("overwrite"); ok == true {
```
Ninir: the `ok == true` isn't needed here

website/docs/r/ssm_parameter.html.markdown
```diff
@@ -56,7 +56,7 @@ The following arguments are supported:
 * `type` - (Required) The type of the parameter. Valid types are `String`, `StringList` and `SecureString`.
 * `value` - (Required) The value of the parameter.
 * `key_id` - (Optional) The KMS key id or arn for encrypting a SecureString.
-* `overwrite` - (Optional) Overwrite an existing parameter. If not specified, will default to `false`.
+* `overwrite` - (Optional) Overwrite an existing parameter even if it was created outside of terraform.
```
Ninir: I think we should really detail this argument a bit more here, to provide better inputs about what is going on.
Just for the case "if specified and set to true,...". Thoughts? :)

schmod:
any chance of this getting merged soon?  the current behavior is moderately unusable (and, frankly, having `overwrite` as a special-case with SSM parameters is kind of odd, given that Terraform generally doesn't understand "upsert" operations on other kinds of resources)

bflad:
It looks like this functionality made it in via #1520, which was previously released in [version 1.11.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1110-march-09-2018) and has been available in all releases since. Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #1560, opened 2017-09-01
Proposal: r/codepipeline: Add note with link on how to find more information regarding arguments

hussfelt:
⚡️ Proposal ⚡️ 

Since digging through the documentation the past weeks I find it more and more frustrating that the details on each argument is lacking.

Take this resource for instance:
https://www.terraform.io/docs/providers/aws/r/codepipeline.html#configuration

*(Optional) A Map of the action declaration's configuration.*

There is no further details on what the map should look like. I need to circle back to the AWS CLI documentation at this point (because I know to look there, which took me a while to figure out), and then to the best of my abilities translate that into terraform.

A link like this would lower the threshold for on-boarding terraform.
It would be beneficial to add this type of link block on all resources, not only codepipeline - possibly this could be done in programatically...

Feedback please! :-)

### DIFF
--- a/website/docs/r/codepipeline.markdown
+++ b/website/docs/r/codepipeline.markdown
@@ -120,6 +120,8 @@ resource "aws_codepipeline" "foo" {
 
 ## Argument Reference
 
+~> **Note:** For more in depth details regarding arguments below, refer to the [AWS CLI documentation](http://docs.aws.amazon.com/cli/latest/reference/codepipeline/index.html) for codepipeline.
+
 The following arguments are supported:
 
 * `name` - (Required) The name of the pipeline.
--- a/website/docs/r/codepipeline.markdown
+++ b/website/docs/r/codepipeline.markdown
@@ -120,8 +120,6 @@ resource "aws_codepipeline" "foo" {
 
 ## Argument Reference
 
-~> **Note:** For more in depth details regarding arguments below, refer to the [AWS CLI documentation](http://docs.aws.amazon.com/cli/latest/reference/codepipeline/index.html) for codepipeline.
-
 The following arguments are supported:
 
 * `name` - (Required) The name of the pipeline.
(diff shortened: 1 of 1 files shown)


hussfelt:
@Ninir I'd definitely pull it off on this specific page - but finding all those resources lacking these kinds of pointers in the whole documentation is most probably a couple of weeks work... :)

Is there any other thoughts around how to accomplish a better and deeper documentation?
Even pointing to the resource above or the direct ones leaves quite a lot to grasp, I still have not been able to terraform my pipeline with the properer configuration.

Ninir:
> @Ninir I'd definitely pull it off on this specific page - but finding all those resources lacking these kinds of pointers in the whole documentation is most probably a couple of weeks work... :)

Was writing about this resource only :) it can be a step-by-step work: and we do prefer that! :)
So for the map of configuration, a documentation link on this specific line would be awesome, like:

*(Optional) A Map of the action declaration's configuration. Find out more about configuring action configurations in the [Reference Pipeline Structure documentation](http://docs.aws.amazon.com/codepipeline/latest/userguide/reference-pipeline-structure.html#action-requirements).*

> Is there any other thoughts around how to accomplish a better and deeper documentation?
> Even pointing to the resource above or the direct ones leaves quite a lot to grasp, I still have not been able to terraform my pipeline with the properer configuration.

If you have any issue, please let us know! :)

hussfelt:
@ninir Awesome, I'll fix this asap and push an update. Your request makes total sense and is in line with the feedback I was hoping for. Understandable that you'd like to do it step-by-step, maybe a pointer somewhere though when creating new resources to do these deep-links, or something...

Cheers!

hussfelt:
@Ninir Fixed this occurrence. Feel free to merge, I'll go ahead and use the above convention when pushing suggestions!

Thanks for the input! 🍻

PR #1572, opened 2017-09-01
aws_ami: Add root_snapshot_id attribute

phinze:
Current limitations in HCL prevent it from being possible to reference
an AMI's root snapshot ID via its ID. This is because the snapshot ID is
deeply nested in the `block_device_mappings` set. HCL is not
yet flexible enough to allow a lookup expression to extract the snapshot
ID embedded within that structure.

This makes it impossible to use resources like
`aws_snapshot_create_volume_permission` without manually doing the AMI
ID -> Snapshot ID translation outside of Terraform.

Here, we promote the Snapshot ID of the root volume to a top-level
`root_snapshot_id` attribute for AMI data sources and resources.

This field will be blank for instance store AMIs or for weird conditions
where there's no snapshot ID recorded (See commit message here:
bed5f624f8f914080ecad627f855d1e9224b6ad3), but it enables usage of the
resource referenced above for the majority of scenarios.


phinze:
```
❯ go test -run 'TestAccAWSAmiDataSource_*' -v .
=== RUN   TestAccAWSAmiDataSource_natInstance
--- PASS: TestAccAWSAmiDataSource_natInstance (17.32s)
=== RUN   TestAccAWSAmiDataSource_windowsInstance
--- PASS: TestAccAWSAmiDataSource_windowsInstance (14.51s)
=== RUN   TestAccAWSAmiDataSource_instanceStore
--- PASS: TestAccAWSAmiDataSource_instanceStore (11.44s)
=== RUN   TestAccAWSAmiDataSource_owners
--- PASS: TestAccAWSAmiDataSource_owners (214.44s)
=== RUN   TestAccAWSAmiDataSource_ownersEmpty
--- PASS: TestAccAWSAmiDataSource_ownersEmpty (105.74s)
=== RUN   TestAccAWSAmiDataSource_localNameFilter
--- PASS: TestAccAWSAmiDataSource_localNameFilter (140.15s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       503.864s
```

```
=== RUN   TestAccAWSAMI_basic
--- PASS: TestAccAWSAMI_basic (52.48s)
=== RUN   TestAccAWSAMI_snapshotSize
--- PASS: TestAccAWSAMI_snapshotSize (51.43s)
```

/cc @apparentlymart as we discussed this a bit

aws/resource_aws_ami_test.go
```diff
@@ -202,12 +205,14 @@ func testAccCheckAmiEbsBlockDevice(bd *ec2.BlockDeviceMapping, ed *ec2.EbsBlockD
...
- 	availability_zone = "us-west-2a"
- 	size = 8
- 	tags {
- 	  Name = "testAccAmiConfig_basic"
- 	}
+  availability_zone = "${data.aws_availability_zones.available.names[0]}"
+  size = 8
+  tags {
+    Name = "testAccAmiConfig_basic"
+  }
```
radeksimko: Nitpick - I think the convention is **2** spaces per indentation.

aws/resource_aws_ami_test.go
```diff
@@ -232,12 +237,14 @@ resource "aws_ami" "foo" {
...
- 	availability_zone = "us-west-2a"
- 	size = 20
- 	tags {
- 	  Name = "testAccAmiConfig_snapshotSize"
- 	}
+  availability_zone = "${data.aws_availability_zones.available.names[0]}"
+  size = 20
+  tags {
+    Name = "testAccAmiConfig_snapshotSize"
+  }
```
radeksimko: ^

radeksimko:
I'm not sure why did I point out the correct indentation as incorrect 🤔 
Anyways - I just resolved conflicts... 🚢 

Thanks for the PR. 👍

Issue #1588, opened 2017-09-05
Feature: aws_alb_target_group_attachment - Support for new "IP" target type


hashibot:
_This issue was originally opened by @simonclausen as hashicorp/terraform#16018. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

AWS recently added the possibility to use a "IP address" target type in their target group config for ALB, see: https://aws.amazon.com/blogs/aws/new-application-load-balancing-via-ip-address-to-aws-on-premises-resources/

It would of course be great to be able to make use of that with terraform, so here's a note on the subject.

Thank you very much.

PR #1589, opened 2017-09-05
resource/aws_alb_target_group: Add support for `target_type`

stack72:
Fixes: #1588

```
% make testacc TEST=./aws TESTARGS='-run=TestAccAWSALBTargetGroup_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSALBTargetGroup_ -timeout 120m
=== RUN   TestAccAWSALBTargetGroup_basic
--- PASS: TestAccAWSALBTargetGroup_basic (73.55s)
=== RUN   TestAccAWSALBTargetGroup_namePrefix
--- PASS: TestAccAWSALBTargetGroup_namePrefix (72.26s)
=== RUN   TestAccAWSALBTargetGroup_generatedName
--- PASS: TestAccAWSALBTargetGroup_generatedName (74.94s)
=== RUN   TestAccAWSALBTargetGroup_changeNameForceNew
--- PASS: TestAccAWSALBTargetGroup_changeNameForceNew (129.78s)
=== RUN   TestAccAWSALBTargetGroup_changeProtocolForceNew
--- PASS: TestAccAWSALBTargetGroup_changeProtocolForceNew (151.66s)
=== RUN   TestAccAWSALBTargetGroup_changePortForceNew
--- PASS: TestAccAWSALBTargetGroup_changePortForceNew (129.73s)
=== RUN   TestAccAWSALBTargetGroup_changeVpcForceNew
--- PASS: TestAccAWSALBTargetGroup_changeVpcForceNew (112.48s)
=== RUN   TestAccAWSALBTargetGroup_tags
--- PASS: TestAccAWSALBTargetGroup_tags (124.72s)
=== RUN   TestAccAWSALBTargetGroup_updateHealthCheck
--- PASS: TestAccAWSALBTargetGroup_updateHealthCheck (122.61s)
=== RUN   TestAccAWSALBTargetGroup_updateSticknessEnabled
--- PASS: TestAccAWSALBTargetGroup_updateSticknessEnabled (170.23s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1161.982s
```

### DIFF
--- a/aws/resource_aws_alb_target_group_test.go
+++ b/aws/resource_aws_alb_target_group_test.go
@@ -38,7 +38,7 @@ func TestALBTargetGroupCloudwatchSuffixFromARN(t *testing.T) {
 	}
 
 	for _, tc := range cases {
-		actual := albTargetGroupSuffixFromARN(tc.arn)
+		actual := lbTargetGroupSuffixFromARN(tc.arn)
 		if actual != tc.suffix {
 			t.Fatalf("bad suffix: %q\nExpected: %s\n     Got: %s", tc.name, tc.suffix, actual)
 		}
--- a/aws/resource_aws_alb_target_group_test.go
+++ b/aws/resource_aws_alb_target_group_test.go
@@ -1,0 +1,789 @@
+package aws
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestALBTargetGroupCloudwatchSuffixFromARN(t *testing.T) {
+	cases := []struct {
+		name   string
+		arn    *string
+		suffix string
+	}{
+		{
+			name:   "valid suffix",
+			arn:    aws.String(`arn:aws:elasticloadbalancing:us-east-1:123456:targetgroup/my-targets/73e2d6bc24d8a067`),
+			suffix: `targetgroup/my-targets/73e2d6bc24d8a067`,
+		},
+		{
+			name:   "no suffix",
+			arn:    aws.String(`arn:aws:elasticloadbalancing:us-east-1:123456:targetgroup`),
+			suffix: ``,
+		},
+		{
+			name:   "nil ARN",
+			arn:    nil,
+			suffix: ``,
+		},
+	}
+
+	for _, tc := range cases {
+		actual := albTargetGroupSuffixFromARN(tc.arn)
+		if actual != tc.suffix {
+			t.Fatalf("bad suffix: %q\nExpected: %s\n     Got: %s", tc.name, tc.suffix, actual)
+		}
+	}
+}
+
+func TestAccAWSALBTargetGroup_basic(t *testing.T) {
+	var conf elbv2.TargetGroup
+	targetGroupName := fmt.Sprintf("test-target-group-%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_alb_target_group.test",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckAWSALBTargetGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSALBTargetGroupConfig_basic(targetGroupName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckAWSALBTargetGroupExists("aws_alb_target_group.test", &conf),
+					resource.TestCheckResourceAttrSet("aws_alb_target_group.test", "arn"),
+					resource.TestCheckResourceAttr("aws_alb_target_group.test", "name", targetGroupName),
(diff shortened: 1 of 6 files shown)


aws/resource_aws_alb_target_group.go
```diff
@@ -79,6 +80,14 @@ func resourceAwsAlbTargetGroup() *schema.Resource {
 				ValidateFunc: validateAwsAlbTargetGroupDeregistrationDelay,
 			},
 
+			"target_type": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "instance",
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{"ip", "instance"}, true),
```
Ninir: Could you prefer the use of constants (TargetTypeEnumInstance & TargetTypeEnumIp) here?

website/docs/r/alb_target_group.html.markdown
```diff
@@ -39,6 +39,12 @@ The following arguments are supported:
 * `deregistration_delay` - (Optional) The amount time for Elastic Load Balancing to wait before changing the state of a deregistering target from draining to unused. The range is 0-3600 seconds. The default value is 300 seconds.
 * `stickiness` - (Optional) A Stickiness block. Stickiness blocks are documented below.
 * `health_check` - (Optional) A Health Check block. Health Check blocks are documented below.
+* `target_type` - (Optional) The type of target that you must specify when registering targets with this target group.
+The possible values are i`nstance` (targets are specified by instance ID) or `ip` (targets are specified by IP address).
```
Ninir: Typo in *i`nstance`*

website/docs/r/alb_target_group.html.markdown
```diff
@@ -39,6 +39,12 @@ The following arguments are supported:
...
+The default is `instance`. Note that you can't specify targets for a target group using both instance IDs and IP addresses.
+If the target type is `ip` , specify IP addresses from the subnets of the virtual private cloud (VPC) for the target group,
```
Ninir: Additional space before the comma

aws/resource_aws_alb_target_group.go
stack72: I don't see what extra that would give us? It's for validation of what the user inputs - i'm interested in seeing what you think

stack72:
Thanks for the review @Ninir  - made the docs changes - I don't agree with the constants change but i'm willing to listen to why

P.

aws/resource_aws_alb_target_group.go
Ninir: By using SDK constants everywhere, you rely on the same exact value everywhere. 
Relying on a user-typed string can brings issues if any typo is made. Thus, that's why we try to stick to them as much as possible, whenever they exist.

stack72: I will update the PR to remove the validation - I don't see any reason why we need to use constants when the values are `ip` or `instance` - if they were longer strings then I would understand

Ninir: Paul,

Perhaps I was misleading in my comment, but the validation was really fine to me.

Was just wondering if we could do that:
```go
ValidateFunc: validation.StringInSlice([]string{elbv2.TargetTypeEnumInstance, elbv2. TargetTypeEnumIp}, true),
```

Sorry if was not clear at all 😅

stack72:
Hi @Ninir 

I made the docs changes and just removed the validation check - I believe this is ready to go now

P.

Ninir:
This looks very good to me!

Just a small question: shouldn't we add a validation for that? do you think adding it will restrict us as soon as there is a new potential target?

Will make the last review & run tests as soon as I get your answer on that :)

Thanks a lot! 👍

stack72:
@Ninir I think validation of that type will restrict us going forward - I had a huge conversation with @tombuildsstuff about validation the other day and the types we do feel wrong. Think of how long it has been since an AWS provider release

IF a new target was added, then it stops people using the provider

f3lang:
@stack72 At first, great change :) I just did a test in my staging environment and stumbled upon an issue, that currently exists, when you try to attach ip targets. 

When you attach an instance target, aws knows the availability zone. But when you attach an ip target to a target group with the target_type "ip", you need to tell AWS the Availability Zone of the target. If the target is not within a VPC, but on-premise, you need to set the Availability Zone of the target to "all".

The aws_alb_target_group_attachment doesn't have the availability_zone option right now.

stack72:
Hi @f3lang 

Thanks for the feedback here - you are indeed correct, I just updated the PR to include `availability_zone` in the `alb_target_group_attachment`

```
terraform-provider-aws [f-aws-alb-target-group-type●] % acctests aws TestAccAWSALBTargetGroupAttachment_
=== RUN   TestAccAWSALBTargetGroupAttachment_basic
--- PASS: TestAccAWSALBTargetGroupAttachment_basic (143.64s)
=== RUN   TestAccAWSALBTargetGroupAttachment_withoutPort
--- PASS: TestAccAWSALBTargetGroupAttachment_withoutPort (150.73s)
=== RUN   TestAccAWSALBTargetGroupAttachment_ipAddress
--- PASS: TestAccAWSALBTargetGroupAttachment_ipAddress (150.28s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	444.678s
```

Thanks for the nudge on this :)

P.

stack72:
@Ninir I don't believe we should add validation here so I believe this can be finalized

Paul

Ninir:
@stack72 hey Paul,

Could you fix the Go issue (aws/resource_aws_alb_target_group_test.go:41:13: undefined: albTargetGroupSuffixFromARN)? Can't test it right now :/

Will run tests & merge as soon as it's all ok. Thanks!

alex1x:
Would be great to have this finalized, this is an important missing feature.

PR #1590, opened 2017-09-05
[WIP] Fix assuming role via ".aws/config"

### LINKED ISSUE #1184: Using terraform with assumed roles defined in shared configuration are ineffective
### Terraform Version
Tested with 0.9.4 through to 0.9.11 and cannot get the role assumption to work.

### Affected Resource(s)
Authentication to access environment using assumed role.

_none_

Note that the configuration does not include a 'assume_role' block - this is defined at the command line.

### Debug Output
Full debug dot provided; an output can be provided if this is to be investigated.

The relevant section of authentication within the log is (this is short enough to include without a gist):
```
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] No assume_role block read from configuration
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] Building AWS region structure
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] Building AWS auth structure
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] AssumeRoleARN:
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] Endpoint Config: &{CredentialsChainVerboseErrors:<nil> Credentials:<nil> Endpoint:<nil> EndpointResolver:<nil> EnforceShouldRetryCheck:<nil> Region:<nil> DisableSSL:<nil> HTTPClient:0xc420ab0f60 LogLevel:<nil> Logger:<nil> MaxRetries:<nil> Retryer:<nil> DisableParamValidation:<nil> DisableComputeChecksums:<nil> S3ForcePathStyle:<nil> S3Disable100Continue:<nil> S3UseAccelerate:<nil> EC2MetadataDisableTimeoutOverride:<nil> UseDualStack:<nil> SleepDelay:<nil> DisableRestProtocolURICleaning:<nil>}
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] Our Config: &{AccessKey: SecretKey: CredsFilename: Profile: Token: Region:us-east-1 MaxRetries:25 AssumeRoleARN: AssumeRoleExternalID: AssumeRoleSessionName: AssumeRolePolicy: AllowedAccountIds:[] ForbiddenAccountIds:[] CloudFormationEndpoint: CloudWatchEndpoint: CloudWatchEventsEndpoint: CloudWatchLogsEndpoint: DynamoDBEndpoint: DeviceFarmEndpoint: Ec2Endpoint: ElbEndpoint: IamEndpoint: KinesisEndpoint: KmsEndpoint: RdsEndpoint: S3Endpoint: SnsEndpoint: SqsEndpoint: Insecure:false SkipCredsValidation:false SkipGetEC2Platforms:false SkipRegionValidation:false SkipRequestingAccountId:false SkipMetadataApiCheck:false S3ForcePathStyle:false}
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] Credential chain: &{creds:{AccessKeyID: SecretAccessKey: SessionToken: ProviderName:} forceRefresh:true m:{state:0 sema:0} provider:0xc420ab0fc0}
2017/07/18 11:58:06 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:06 [INFO] Attempt to get credentials: {AccessKeyID:ALITTLELAMB SecretAccessKey:ANOTHERLAMB SessionToken: ProviderName:SharedCredentialsProvider}
2017/07/18 11:58:07 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:07 [INFO] Ignoring AWS metadata API endpoint at default location as it doesn't return any instance-id
2017/07/18 11:58:07 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:07 [INFO] AWS Auth provider used: "SharedCredentialsProvider"
2017/07/18 11:58:07 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:07 [INFO] Initializing DeviceFarm SDK connection
2017/07/18 11:58:07 [DEBUG] plugin: terraform: aws-provider (internal) 2017/07/18 11:58:07 [DEBUG] [aws-sdk-go] DEBUG: Request sts/GetCallerIdentity Details:
```

Additional debug, to print the contents of parameters exists within this output in order to try to determine the parameters being processed, has been added here (otherwise there's not a lot of extra information.

### Intended usage
I work in DevOps, and my accounts on AWS are part of our 'DevOps / Build and Test account'. The teams I work with have their own accounts - the backend team use their 'Backend' account, and the R&D account have an 'Experimental account'. The usual way of working is that we have users within our own accounts, but we can assume roles within other accounts depending on our access needs. This way of working means that we can restrict the permissions in different areas across accounts, and not proliferate users within accounts that are not needed.

It is not appropriate for us to put the assume_role into the terraform files, because they should be generic - applicable to any team's account. Thus, it makes more sense for us to define the profile using the 'AWS_PROFILE'/'AWS_DEFAULT_PROFILE' settings.

### Expected Behavior
I expect that I can create a shared credentials file (~/.aws/credentials) containing (with actual credentials obscured):

```
[default]
aws_access_key_id=ALITTLELAMB
aws_secret_access_key=ANOTHERLAMB
```

And a shared configuration file containing:

```
[profile backend]
assume_role_arn = arn:aws:iam::01234567890123:role/Maintenance
default_profile = default
```

And the environment variable `AWS_PROFILE` set to '`backend`'.
Also the `AWS_DEFAULT_PROFILE` is set to '`backend`' (as it seems unclear how these are used within the SDK).

It is possible that the `AWS_SDK_LOAD_CONFIG` variable needs to be set to a 'truthy' value ('`1`'), so I have tried this with the value set and not set, and neither was effective.

### Actual Behavior
The Terraform behaviour appears to use the default account, not the assumed role.

### Steps to Reproduce
See above.

### Important Factoids
See 'Intended usage' above.

During trying to testing there were a few configurations that resulted in crashes, which will be reported independantly.

We would like to use this behaviour, but after a day or so of testing with different combinations and adding further debugging to the go code, and reading the SDK code and documentation, I have not managed to make the assumed roles work in this way (assuming roles using the access/secret/session keys manually generated by `aws sts assume-role` works, but is a much more awkward mechanism). The implication on the `aws-go-sdk` documentation are that the use of NewSession would take care of these configuration settings; but I get lost in the use of the chained credentials that is called by the auth_helpers. Go code, and the internals of terraform, are not speciality, but although I'm happy to look at it, I've reached the end of what I can investigate and so... I ask for help on this - even if it's not a 'bug', but just that I'm Doing It All Wrong.

svenwltr:
> Fixes #1184

This PR allows assuming roles via shared configuration `~/.aws/config`. It still supports assuming other roles for different provider configurations afterwards.

I am somehow not fully satisfied with the code yet, so feel free to nitpick everything.

**Open Questions**

* Should I try to reduce code duplication due to assuming roles twice?
* I am not entirely sure, if MFA would work correctly, but AFAIS when using MFA one has to do the `aws sts assume-role` manually anyway, so it shouldn't be affected by this change.

### DIFF
--- a/aws/auth_helpers.go
+++ b/aws/auth_helpers.go
@@ -22,7 +22,6 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/go-multierror"
-	homedir "github.com/mitchellh/go-homedir"
 )
 
 func GetAccountInfo(iamconn *iam.IAM, stsconn *sts.STS, authProviderName string) (string, string, error) {
@@ -172,9 +171,25 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 		}
 	}
 
+	// Try to read AssumeRole data from ~/.aws/config
 	assumeRoleConfig := readAssumeRoleConfig(c)
 	if assumeRoleConfig != nil {
-		assumeRoleConfig.assumeRole()
+		creds, err := assumeRoleConfig.assumeRole(c)
+		if err != nil {
+			return nil, err
+		}
+
+		value, err := creds.Get()
+		if err != nil {
+			return nil, err
+		}
+
+		// One might still want to assume a different role in the provider
+		// config, so we just update the providers instead of directly
+		// returning the credentials.
+		providers = []awsCredentials.Provider{
+			&awsCredentials.StaticProvider{Value: value},
+		}
 	}
 
 	// This is the "normal" flow (i.e. not assuming a role)
@@ -253,27 +268,29 @@ type assumeRoleConfig struct {
 }
 
 func readAssumeRoleConfig(c *Config) *assumeRoleConfig {
-	configFilename, err := homedir.Expand(c.ConfigFilename)
-	if err != nil {
-		log.Printf("[WARN] Unable to expand config file path: %v", err)
-		return nil
+	profile := os.Getenv("AWS_PROFILE")
+	if profile == "" {
+		profile = c.Profile
+	}
+	if profile == "" {
+		profile = "default"
 	}
 
-	configContent, err := ioutil.ReadFile(configFilename)
+	configContent, err := ioutil.ReadFile(c.ConfigFilename)
 	if err != nil {
-		log.Printf("[WARN] Unable read config file (%s): %v", configFilename, err)
+		log.Printf("[WARN] Unable read config file (%s): %v", c.ConfigFilename, err)
 		return nil
 	}
 
 	iniData, err := ini.Load(configContent)
 	if err != nil {
-		log.Printf("[WARN] Unable parse config file (%s): %v", configFilename, err)
+		log.Printf("[WARN] Unable parse config file (%s): %v", c.ConfigFilename, err)
 		return nil
 	}
 
-	section, err := iniData.GetSection(fmt.Sprintf("profile %s", c.Profile))
+	section, err := iniData.GetSection(fmt.Sprintf("profile %s", profile))
 	if err != nil {
-		log.Printf("[DEBUG] Profile (%s) doesn't exist in config file (%s): %v", c.Profile, configFilename, err)
+		log.Printf("[DEBUG] Profile (%s) doesn't exist in config file (%s): %v", profile, c.ConfigFilename, err)
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


radeksimko:
Hey @svenwltr 
thanks for the contribution.

It makes sense for Terraform to support assuming roles via `~/.aws/config`. That said I'd rather see the INI parsing logic and all the format-specific logic, or generally as much of that code as possible in the upstream SDK, e.g. here https://github.com/aws/aws-sdk-go/tree/master/aws/credentials
The SDK is IMO much better place for it and we can make sure it's aligned with AWS' conventions/expectations and kept up to date as AWS engineers maintain the SDK.

Have you looked into the SDK and considered raising a PR there?

svenwltr:
Hey @radeksimko,

the SDK is already able to do this, as I wrote in the actual [issue](https://github.com/terraform-providers/terraform-provider-aws/issues/1184#issuecomment-326281749):

```golang
sess, err := session.NewSessionWithOptions(session.Options{
    Config: aws.Config{
        Region: "my-region",
    },
    SharedConfigState: session.SharedConfigEnable,
    Profile:           "my-profile",
})
```

I did it the manual way, because Terraform already does more manually than necessary, which confused me a bit. I also strongly prefer to use the SDKs ability.

Also I am very fine, if we go with the PR #1608, which solves the same problem in another way.

radeksimko:
Ah, I see - sorry I didn't read the whole conversation there. 🙈 

Have you seen this PR? https://github.com/terraform-providers/terraform-provider-aws/pull/1608

It will probably need some tests & more thorough review, but it's more in-line with what we're discussing here. What do you think?

radeksimko:
I should read more before posting comments, clearly...

> Also I am very fine, if we go with the PR #1608, which solves the same problem in another way.

Cool. With that are you happy for me to close this PR?

svenwltr:
Yeah, the issues and PRs got a bit messed up :-)

Ok, it looks like we're on the same line here. Therefore I close this in favor of #1608.

PR #1606, opened 2017-09-07
Add Waf RateBasedRule support

bekbulatov:
Updated version of https://github.com/terraform-providers/terraform-provider-aws/pull/1285. 
I had to add the required `type` field to the rules block under `aws_waf_rule`. The value could be either `REGULAR` or `RATE_BASED` regarding which type of rule you are using.
The docs are updated as well.

### DIFF
--- a/aws/resource_aws_waf_rate_based_rule.go
+++ b/aws/resource_aws_waf_rate_based_rule.go
@@ -68,10 +68,24 @@ func resourceAwsWafRateBasedRule() *schema.Resource {
 			"rate_key": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					if value != waf.RateKeyIp {
+						errors = append(errors, fmt.Errorf("%q must be only %s", k, waf.RateKeyIp))
+					}
+					return
+				},
 			},
 			"rate_limit": &schema.Schema{
 				Type:     schema.TypeInt,
 				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(int)
+					if value < 2000 {
+						errors = append(errors, fmt.Errorf("%q cannot be less than 2000", k))
+					}
+					return
+				},
 			},
 		},
 	}
@@ -86,6 +100,8 @@ func resourceAwsWafRateBasedRuleCreate(d *schema.ResourceData, meta interface{})
 			ChangeToken: token,
 			MetricName:  aws.String(d.Get("metric_name").(string)),
 			Name:        aws.String(d.Get("name").(string)),
+			RateKey:     aws.String(d.Get("rate_key").(string)),
+			RateLimit:   aws.Int64(int64(d.Get("rate_limit").(int))),
 		}
 
 		return conn.CreateRateBasedRule(params)
@@ -142,8 +158,9 @@ func resourceAwsWafRateBasedRuleUpdate(d *schema.ResourceData, meta interface{})
 	if d.HasChange("predicates") {
 		o, n := d.GetChange("predicates")
 		oldP, newP := o.(*schema.Set).List(), n.(*schema.Set).List()
+		rateLimit := aws.Int64(int64(d.Get("rate_limit").(int)))
 
-		err := updateWafRateBasedRuleResource(d.Id(), oldP, newP, conn)
+		err := updateWafRateBasedRuleResource(d.Id(), oldP, newP, rateLimit, conn)
 		if err != nil {
 			return fmt.Errorf("Error Updating WAF Rule: %s", err)
 		}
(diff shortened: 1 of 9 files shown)


aws/resource_aws_waf_rate_based_rule.go
```diff
@@ -0,0 +1,219 @@
...
+								}
+								return
+							},
+						},
+					},
+				},
+			},
+			"rate_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
```
radeksimko: It seems that this field isn't updatable (based on API reference), shouldn't it be marked as `ForceNew` then?

aws/resource_aws_waf_rate_based_rule.go
```diff
@@ -0,0 +1,219 @@
...
+		return conn.DeleteRateBasedRule(req)
+	})
+	if err != nil {
+		return fmt.Errorf("Error deleting WAF Rate Based Rule: %s", err)
+	}
+
+	return nil
+}
+
+func updateWafRateBasedRuleResource(id string, oldP, newP []interface{}, rateLimit *int64, conn *waf.WAF) error {
```
radeksimko: It might be more convenient to accept `rateLimit int` here instead and do all the pointer magic inside the function to reduce the repetition, but no biggie.

aws/resource_aws_waf_web_acl.go
```diff
@@ -64,6 +64,18 @@ func resourceAwsWafWebAcl() *schema.Resource {
...
+							Required: true,
+							ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+								value := v.(string)
+								if value != waf.WafRuleTypeRegular && value != waf.WafRuleTypeRateBased {
+									errors = append(errors, fmt.Errorf(
+										"%q must be one of %s | %s", k, waf.WafRuleTypeRegular, waf.WafRuleTypeRateBased))
+								}
+								return
+							},
+						},
```
radeksimko: Is there any particular reason why this change should be included in this PR? How is it related to WAF Rate Based Rule?

aws/resource_aws_waf_rate_based_rule.go
```diff
@@ -0,0 +1,219 @@
...
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"negated": &schema.Schema{
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+						"data_id": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
```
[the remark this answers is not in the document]
bekbulatov: Sure, I will update the schema for `Rule` as well

aws/resource_aws_waf_rate_based_rule.go
bekbulatov: `Rate_key` could be only `IP` now. Why do you think it should be recreated if changed?

aws/resource_aws_waf_rate_based_rule.go
```diff
@@ -0,0 +1,219 @@
...
+			"rate_key": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					if value != waf.RateKeyIp {
+						errors = append(errors, fmt.Errorf("%q must be only %s", k, waf.RateKeyIp))
+					}
+					return
+				},
```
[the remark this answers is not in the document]
bekbulatov: Agreed, anyway the value will be validated by AWS.

aws/resource_aws_waf_web_acl.go
bekbulatov: I mentioned this in https://github.com/terraform-providers/terraform-provider-aws/pull/1285. By default, WebACL works with `REGULAR` rules and you will get an error if you pass `rule_id` of a `RATE_BASED` rule. This is documented in https://github.com/aws/aws-sdk-go/blob/master/service/waf/api.go#L5708-L5713

aws/resource_aws_waf_rate_based_rule.go
radeksimko: well, you're right that there's no way to change it, but if the user is unaware they might try and at the moment we'd just ignore the field (in `Update`) and support user's assumption that the field was successfully changed.

If we instead delegate the validation work to `Create` then it will fail and user will also realize during `plan` that this can't be changed.

aws/resource_aws_waf_rate_based_rule.go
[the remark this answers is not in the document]
radeksimko: Yep, I think the AWS API validation is the best choice here.

aws/resource_aws_waf_web_acl.go
radeksimko: I see, but I think we can't just introduce a `Required` field - that would break it for users with existing configs which don't have this field there.
This applies to your other change in `aws/resource_aws_waf_rule.go` too.

radeksimko: Do you mind changing it to Optional?

radeksimko: btw. you might be able to avoid changing the `updateWebAclResource` if you also add the current default value as `Default` - that way it should just work the same way as before, yet allow users to specify `type = "RATE_BASED"` if they need to.

bekbulatov: @radeksimko Thank you for the review. I believe the `Required` change for `aws/resource_aws_waf_rule.go` is fine, as we just duplicate AWS validation, it's not a breaking change.
I made `type` optional and default to `REGULAR` as recommended.

radeksimko: You're right - just checked the [API docs](http://docs.aws.amazon.com/waf/latest/APIReference/API_Predicate.html) - it was probably just overlooked.

Thanks.

PR #1614, opened 2017-09-07
updated with latest redshift service account ID's

david-wells-1:
AWS Redshift service accounts IDs added to included newest regions where Redshift is now supported 

Documentation: http://docs.aws.amazon.com/fr_fr/redshift/latest/mgmt/db-auditing.html

### DIFF
--- a/aws/data_source_aws_redshift_service_account_test.go
+++ b/aws/data_source_aws_redshift_service_account_test.go
@@ -20,7 +20,7 @@ func TestAccAWSRedshiftServiceAccount_basic(t *testing.T) {
 			resource.TestStep{
 				Config: testAccCheckAwsRedshiftServiceAccountExplicitRegionConfig,
 				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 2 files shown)


aws/data_source_aws_redshift_service_account_test.go
```diff
@@ -33,6 +33,6 @@ data "aws_redshift_service_account" "main" { }
 
 const testAccCheckAwsRedshiftServiceAccountExplicitRegionConfig = `
 data "aws_redshift_service_account" "regional" {
-	region = "eu-west-1"
+	region = "eu-west-2"
```
Ninir: Is there a reason to change this test?

david-wells-1: Possibly not - I'm new to updating the terraform source code

I thought it would help validate the new Redshift service account (eu-west-2) being added - remove if not required

Ninir: No worries, was just wondering :)

PR #1621, opened 2017-09-08
Instance - Instance profile, uses ID as input not name of the profile

willfarley:
Instance Profile parameter takes ID not the name of the Instance profile.


Ninir:
Hi @willfarley 

Thank you for this change.

The documentation was correct and still is with your change: the profile iam_instance_profile ID is also the name, as [exposed in the code](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_iam_instance_profile.go#L294).

I would prefer to keep the `name` in the documentation, since it has more meaning than just the ID.

What do you think?

Ninir:
Hi @willfarley 

For the reason above, I'm going to close this PR.

Thanks!

PR #1626, opened 2017-09-08
remove "id" fields from schema definitions

jbardin:
The "id" field is built-in and not required in the schema. Future
updates to helper/schema will cause these to fail validation.

### DIFF
--- a/aws/data_source_aws_availability_zone.go
+++ b/aws/data_source_aws_availability_zone.go
@@ -79,7 +79,6 @@ func dataSourceAwsAvailabilityZoneRead(d *schema.ResourceData, meta interface{})
 	nameSuffix := (*az.ZoneName)[len(*az.RegionName):]
 
 	d.SetId(*az.ZoneName)
-	d.Set("id", az.ZoneName)
 	d.Set("name", az.ZoneName)
 	d.Set("name_suffix", nameSuffix)
 	d.Set("region", az.RegionName)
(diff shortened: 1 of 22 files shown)


jbardin:
Good idea @radeksimko. I updated the calls throughout, and found a few more top-level "id" fields that weren't caught by the unit tests.

I also removed some `SetPartial` calls to "id", which shouldn't be required since `SetId` is being called outside of `Partial`.

radeksimko:
Thanks for making those changes - I just ran a few experiments to verify @paddycarver 's thoughts about `Optional` fields he mentioned in https://github.com/terraform-providers/terraform-provider-google/pull/399#pullrequestreview-61627974 and I have to admit he's right.

We can't _just_ remove them 😢 , because in some cases they may be working (if the CRUD is correctly using `d.Id()` to get it and `d.SetId()` to set it). At this stage we'd break those data sources, e.g. `aws_eip`:

```hcl
data "aws_eip" "e" {
  id = "eipalloc-0128413b" # This is currently valid and works just fine
}
```

I'm thinking my validation may be more aggressive than I thought. I'm pondering with an idea of excluding data sources from the validation or allowing `Optional` `id` fields to exist if they're also `Deprecated` or `Removed`.

I'm more inclined to the latter (slowly allow phasing out) because it can be confusing to work with this field in the CRUD. It's not immediately obvious `d.Id()` and `d.SetId()` are dedicated getters and setters.

What do you think?

azman0101:
It seems the documentation doesn't mention built-in id attribute https://www.terraform.io/docs/providers/aws/d/iam_server_certificate.html#attributes-reference

PR #1637, opened 2017-09-11
Update ssm_association.html.markdown

Ninir:
Added missing required argument in the ssm_association documentation, as [per the code](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_ssm_document.go#L44)


stack72:
LGTM! Thanks for picking this up @Ninir

Issue #996, opened 2017-06-28
aws_network_interface.private_ips and aws_network_interface.private_ips_count act in an unexpected way


aschepp:
### Terraform Version
Terraform v0.9.8

### Affected Resource(s)
- aws_network_interface

### Terraform Configuration Files
```hcl
resource "aws_network_interface" "interface1" {
	private_ips = [ "${var.ip1}", "${var.ip2}" ]
}

resource "aws_network_interface" "interface2" {
	private_ips_count = 1
}

output "primaryIP" {
  value = "${aws_network_interface.interface2.private_ips[0]}"
}

output "secondaryIP" {
  value = "${aws_network_interface.interface2.private_ips[1]}"
}
```

### Expected Behavior
- The interface 1 should have two IPs, with ip1 being the primary ip and ip2 being the secondary ips.
- The interface 2 should have only one IP total assigned to it.
- The output primaryIP should have the primary IP of the interface2.
- The output secondaryIP should have the secondary IP of the interface2.

### Actual Behavior
- Interface 1 will have ip1 and ip2 assigned in a undetermined order. This is because private_ips is a set with string hash as sorting.
- Interface 2 will have 2 IP assigned to it, one primary and one secondary IP. The parameter should be renamed to secondary_private_ips_count.
- The outputs primaryIP and secondaryIP return the IPs in a undetermined order, because again, private_ips is a set with string hash as sorting.

### Workaround
As there is no way to reliable get the secondary IP assigned to the interface, you have to use something like this:

```hcl
resource "aws_network_interface" "interface2" {
	private_ips_count = 1
}

output "primaryIP" {
  value = "${aws_network_interface.interface2.private_ip}"
}

output "secondaryIP" {
  value = "${replace(join(",",aws_network_interface.interface2.private_ips),"/(^|,)${aws_network_interface.interface2.private_ip}(,|$)/", "")}"
}
```
### Possible Fixes
- Use a List instead of a Set in aws_network_interface.go to store private_ips.
- Or do not include the primary ip in private_ips, as it's available as private_ip and rename private_ips to secondary_private_ips.
- Rename private_ips_count to secondary_private_ips_count to make it less confusing.

### References
This issues has already been mentioned in #836, but I wanted to extend it.

buec011:
> * Rename private_ips_count to secondary_private_ips_count to make it less confusing.

At least that has been documented now:
[...]

Issue #836, opened 2017-06-13
Sort private ips for an AWS instance in Terraform


hashibot:
_This issue was originally opened by @Joe6547654 as hashicorp/terraform#15173. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi, I am using terraform to create an instance with multiple ips in AWS. I specified a list of private_ips in the network interface block and I made sure the one I wanted as primary ip was listed first on the list, however it does not respect that order when I create the instance in AWS. Order doesn't change even when when I change the ips around. I also tried setting with private_ip for the primary ip and private_ips for secondary ips, but it doesn't work. 

### Affected Resource(s)
- aws_instance
- aws_network_interface

### Terraform Configuration Files
```
resource "aws_network_interface" "test" {
	subnet_id = "subnet-xxxxxxx"
	private_ips = ["172.24.3.249", "172.24.3.250", "172.24.3.5", "172.24.3.9", "172.24.3.13"]
}
resource "aws_instance" "test" {
	ami = "ami-xxxxxxxx" 	
	instance_type = "m4.2xlarge"
	key_name = "name"
	network_interface {
	network_interface_id = "${aws_network_interface.test.id}"
	device_index = 0
	}
}
```

### Expected Behavior
The first ip address on the list will appear as the primary ip, and the rest as the secondary ips.

### Actual Behavior
"172.24.3.5" becomes the primary instead of the first ip on the list. It does not change even if we switch the order of the ips. 

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- GH-6986

nbaztec:
#1749 fixes this behavior but hasn't received any attention since more than a year.

Fazered:
How is such a fundamental use case still an issue? 

This problem has cost us thousands as we have to use larger instance types that support more nics/ips as we can only assign a single ips on the primary nic when it needs to be known. 

We expected it would only be temporary but have now been doing it for over a year. It is frustrating.

buec011:
> How is such a fundamental use case still an issue?
> 
> This problem has cost us thousands as we have to use larger instance types that support more nics/ips as we can only assign a single ips on the primary nic when it needs to be known.
> 
> We expected it would only be temporary but have now been doing it for over a year. It is frustrating.

[...]

PR #1672, opened 2017-09-15
resource/aws_network_interface: Use first of private_ips as primary

matt-deboer:
This changes `private_ips` to List (instead of Set) for more intuitive/predictable behavior, which helps to solve issues #836 and #996.

Before this change, there was no reliable way to configure the primary_ip --> you simply got the ip address whose string value's hash had the lowest lexical sort order. Deterministic, yes, but not determined by the order the ips were specified in the list :)

### DIFF
--- a/aws/resource_aws_network_interface.go
+++ b/aws/resource_aws_network_interface.go
@@ -46,11 +46,10 @@ func resourceAwsNetworkInterface() *schema.Resource {
 			},
 
 			"private_ips": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
-				Set:      schema.HashString,
 			},
 
 			"private_ips_count": &schema.Schema{
@@ -119,7 +118,7 @@ func resourceAwsNetworkInterfaceCreate(d *schema.ResourceData, meta interface{})
 		request.Groups = expandStringList(security_groups)
 	}
 
-	private_ips := d.Get("private_ips").(*schema.Set).List()
+	private_ips := d.Get("private_ips").([]interface{})
 	if len(private_ips) != 0 {
 		request.PrivateIpAddresses = expandPrivateIPAddresses(private_ips)
 	}
--- a/aws/resource_aws_network_interface.go
+++ b/aws/resource_aws_network_interface.go
@@ -273,21 +273,21 @@ func resourceAwsNetworkInterfaceUpdate(d *schema.ResourceData, meta interface{})
 	if d.HasChange("private_ips") {
 		o, n := d.GetChange("private_ips")
 		if o == nil {
-			o = new(schema.Set)
+			o = []interface{}{}
 		}
 		if n == nil {
-			n = new(schema.Set)
+			n = []interface{}{}
 		}
 
-		os := o.(*schema.Set)
-		ns := n.(*schema.Set)
+		os := o.([]interface{})
+		ns := n.([]interface{})
 
 		// Unassign old IP addresses
-		unassignIps := os.Difference(ns)
-		if unassignIps.Len() != 0 {
+		unassignIps := difference(os, ns)
+		if len(unassignIps) != 0 {
 			input := &ec2.UnassignPrivateIpAddressesInput{
 				NetworkInterfaceId: aws.String(d.Id()),
-				PrivateIpAddresses: expandStringList(unassignIps.List()),
+				PrivateIpAddresses: expandStringList(unassignIps),
 			}
 			_, err := conn.UnassignPrivateIpAddresses(input)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


justinclayton:
This is pretty important to anyone who needs to attach more than one interface to an instance in a particular order, such as anything using a `keepalived`-style pattern.

curator:
Agreed ... this keeps us from building infrastructures where HA is dependent on deterministic addressing order.

matt-deboer:
@radeksimko any chance of taking another look at this one?

thallam08:
This issue might be resolved by updating the documentation.  The option "private_ip" is not documented at https://www.terraform.io/docs/providers/aws/r/network_interface.html

nbaztec:
I tested the above configuration however the privateIP seemed to be ignored:

```
resource "aws_vpc" "main" {
  cidr_block = "10.0.0.0/16"
}

resource "aws_subnet" "public_a" {
  vpc_id = "${aws_vpc.main.id}"
  cidr_block = "10.0.0.0/24"
  availability_zone = "eu-central-1a"
}

resource "aws_network_interface" "test" {
  subnet_id   = "${aws_subnet.public_a.id}"
  private_ip  = "10.0.0.60"
  private_ips = ["10.0.0.51", "10.0.0.40", "10.0.0.30"]
}
```

```
$ terraform apply
aws_vpc.main: Creating...
  assign_generated_ipv6_cidr_block: "" => "false"
  cidr_block:                       "" => "10.0.0.0/16"
  default_network_acl_id:           "" => "<computed>"
  default_route_table_id:           "" => "<computed>"
  default_security_group_id:        "" => "<computed>"
  dhcp_options_id:                  "" => "<computed>"
  enable_classiclink:               "" => "<computed>"
  enable_classiclink_dns_support:   "" => "<computed>"
  enable_dns_hostnames:             "" => "<computed>"
  enable_dns_support:               "" => "true"
  instance_tenancy:                 "" => "<computed>"
  ipv6_association_id:              "" => "<computed>"
  ipv6_cidr_block:                  "" => "<computed>"
  main_route_table_id:              "" => "<computed>"
aws_vpc.main: Creation complete after 2s (ID: vpc-75ec0b1e)
aws_subnet.public_a: Creating...
  assign_ipv6_address_on_creation: "" => "false"
  availability_zone:               "" => "eu-central-1a"
  cidr_block:                      "" => "10.0.0.0/24"
  ipv6_cidr_block:                 "" => "<computed>"
  ipv6_cidr_block_association_id:  "" => "<computed>"
  map_public_ip_on_launch:         "" => "false"
  vpc_id:                          "" => "vpc-75ec0b1e"
aws_subnet.public_a: Creation complete after 1s (ID: subnet-2844ef43)
aws_network_interface.test: Creating...
  attachment.#:           "" => "<computed>"
  private_dns_name:       "" => "<computed>"
  private_ip:             "" => "10.0.0.60"
  private_ips.#:          "" => "3"
  private_ips.2391807628: "" => "10.0.0.51"
  private_ips.2949480860: "" => "10.0.0.30"
  private_ips.3767277403: "" => "10.0.0.40"
  private_ips_count:      "" => "<computed>"
  security_groups.#:      "" => "<computed>"
  source_dest_check:      "" => "true"
  subnet_id:              "" => "subnet-2844ef43"
```

However the primary was not set correctly as per the console.

```
Network interface ID: eni-1cb07249 | Subnet ID: subnet-2844ef43
VPC ID: vpc-75ec0b1e
MAC address: 02:87:95:cf:fe:3c 
Primary private IPv4 IP: 10.0.0.51
Secondary private IPv4 IPs: 10.0.0.30, 10.0.0.40
```

https://imgur.com/a/UwYMT

thallam08:
I can confirm that the template above is not working.

Maybe the logic needs to be:
- If "private_ip" is set, then use that and prepend it to the "private_ips" if its not already there
- Or if "private_ip" is not set, then set it to the first IP in the "private_ips" list
- Always set the primary IP for an interface to the "private_ip"

Alternatively drop the "private_ip" as it's not documented and just use the first item in "private_ips"

thallam08:
The AWS API works by specifying that one of the IP addresses is primary. http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_PrivateIpAddressSpecification.html

Order is not the determining factor. 

Maybe the cleanest fix is to make sure that "private_ip" is honoured as the primary IP address, and this option is documented.

thallam08:
From the CLI documentation:
`
--private-ip-address (string)

    The primary private IPv4 address of the network interface. If you don't specify an IPv4 address, Amazon EC2 selects one for you from the subnet's IPv4 CIDR range. If you specify an IP address, you cannot indicate any IP addresses specified in privateIpAddresses as primary (only one IP address can be designated as primary).

--private-ip-addresses (list)

    One or more private IPv4 addresses.
`

nbaztec:
I agree with @thallam08 

After reviewing the AWS API it seems the plugin wasn't handling the private IP as it was supposed to be.
I've updated #1749 to behave in a reliable way and updated the tests accordingly. It's highly important to us that we switch away from our custom plugin to an officially supported/working one.

thallam08:
There's a related issue that may not have been logged.  If you change the list of IPs then the primary IP may change.  The AWS CLI allows for adding and removing secondary IPs.  As implemented this module does not allow us to do this reliably as it may try to change the primary IP and either have to recreate the resource or fail (eg primary interface).

yn-academia:
If you include both the "private_ip" and "private_ips", and make sure that "private_ip" is included in the list of "private_ips", it does what you want.

yn-academia:
The real issue here is that "private_ip" is ignored completely, even when set. It's set in the state, but it doesn't seem to propagate to AWS as the primary

thallam08:
That does not work for a non-primary interface.
Not sure it always works for primary interface.

YakDriver:
This is closed by #17846. Let us know if that does not solve for your situation.

PR #1694, opened 2017-09-18
Adding servicecatalog_portfolio resource

bw-intuit:
@ninir here is a pr for the first resource.  Once we get this and the vendor merged I'll iterate in a similar fashion on the remaining resources.

Dependent on: https://github.com/terraform-providers/terraform-provider-aws/pull/1693

More info: https://github.com/terraform-providers/terraform-provider-aws/pull/1663


aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,132 @@
...
+	d.Set("provider_name", portfolioDetail.ProviderName)
+	return nil
+}
+
+func resourceAwsServiceCatalogPortfolioUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).scconn
+	input := servicecatalog.UpdatePortfolioInput{}
+	input.Id = aws.String(d.Id())
+
+	if v, ok := d.GetOk("name"); ok {
```
Ninir: `d.GetOk` will always returns the value plus the existence check if the field is set.
We better need `d.HasChange("name")` here and below, so that Terraform is well-aware of changes and then use GetOk to get the value.

This way, we only update fields that we need to update, and reduce the amount of edge cases. Thoughts?

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if v, ok := d.GetOk("description"); ok {
+		input.Description = aws.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("provider_name"); ok {
+		input.ProviderName = aws.String(v.(string))
+	}
+
+	_, err := conn.UpdatePortfolio(&input)
```
Ninir: Here also, would be nice to have a log instruction with parameters, as for the create method

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,132 @@
...
+	return resourceAwsServiceCatalogPortfolioRead(d, meta)
+}
+
+func resourceAwsServiceCatalogPortfolioRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).scconn
+	input := servicecatalog.DescribePortfolioInput{}
+	input.Id = aws.String(d.Id())
+
+	resp, err := conn.DescribePortfolio(&input)
+	if err != nil {
```
Ninir: We should really check for `ResourceNotFoundException` and if this is the case (removed from the console for instance), we should add a log instruction like:

    log.Printf("[WARN] Service Catalog %q not found, removing from state", d.Id())

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return resourceAwsServiceCatalogPortfolioRead(d, meta)
+}
+
+func resourceAwsServiceCatalogPortfolioDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).scconn
+	input := servicecatalog.DeletePortfolioInput{}
+	input.Id = aws.String(d.Id())
+
+	_, err := conn.DeletePortfolio(&input)
```
Ninir: Here also, a log instruction saying that we are deleting would be nice :)

aws/provider.go
```diff
@@ -438,6 +438,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_network_interface_sg_attachment":          resourceAwsNetworkInterfaceSGAttachment(),
 			"aws_default_security_group":                   resourceAwsDefaultSecurityGroup(),
 			"aws_security_group_rule":                      resourceAwsSecurityGroupRule(),
+			"aws_servicecatalog_portfolio":                 resourceAwsServiceCatalogPortfolio(),
```
Ninir: Can you also add the documentation for this resource?

bw-intuit: @Ninir will do

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,132 @@
+package aws
```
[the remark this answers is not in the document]
bw-intuit: @Ninir will do

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+
+func resourceAwsServiceCatalogPortfolioRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).scconn
+	input := servicecatalog.DescribePortfolioInput{}
+	input.Id = aws.String(d.Id())
+
+	log.Printf("[DEBUG] Reading Service Catalog Portfolio: %#v", input)
+	resp, err := conn.DescribePortfolio(&input)
+	if err != nil {
+		if scErr, ok := err.(awserr.Error); ok && scErr.Code() == "ResourceNotFoundException" {
```
bw-intuit: @Ninir can you suggest, or point me to a similar example, how to idiomatically write a test for this type of error in the tf test framework?

Ninir: This kind of test is not simple to implement, due to the way to test: we need to create the resource, add a custom step to remove the object, and expect a non-empty plan.

In [this example](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_customer_gateway_test.go#L50-L93), you first have a test that expects an error, and then a test that expects a non empty plan.
Depending on our implementation, we will go for one or the other (I think it will be the non-empty plan)

The best way to proceed is perhaps to `make build` the provider and check it manually.

bw-intuit:
@Ninir added a first pass at documentation.  Is there a way I can compile the site to test myself?

aws/resource_aws_servicecatalog_portfolio.go
bw-intuit: @ninir Ran it manually, removed the resource and it works.

```
2017/09/19 06:17:46 [WARN] Service Catalog "port-123456781234" not found, removing from state
```

Still a bit hazy but will try to construct a non-empty plan test using the above.

bw-intuit: @ninir added a first pass at a test.  Can you provide feedback?

https://github.com/terraform-providers/terraform-provider-aws/pull/1694/commits/037d11a42117a7e6c242a94872a241a6ce24417f

Thanks

bw-intuit:
@Ninir added docs and tested. Believe I have addressed all your current requests and updates to the tests.  Can you provide some feedback?  I have time this week and would like to hammer out the rest of them if this is moving in the right direction.

screen shot 2017-09-21 at 8 32 41 am

bw-intuit:
@Ninir any feedback on the updates?  Thanks

trung:
I would like to give hands on this. How should we proceed?

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: true,
+			},
+			"name": {
```
Ninir: What about adding the `AcceptLanguage` and `Tags` attributes?

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
```
Ninir: ForceNew being false is the default value, so no need to specify it :)

Ninir: We should also validate each atttribute, as in: http://docs.aws.amazon.com/servicecatalog/latest/dg/API_CreatePortfolio.html#servicecatalog-CreatePortfolio-request-ProviderName.
For this one, name should be required and have a max length of 20 for instance

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+			},
+		},
+	}
+}
+
+func resourceAwsServiceCatalogPortfolioCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).scconn
+	input := servicecatalog.CreatePortfolioInput{}
+	if v, ok := d.GetOk("name"); ok {
```
Ninir: When an attribute is required, GetOk will return false, so we can directly fill the input knowing that

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+	d.Set("description", portfolioDetail.Description)
+	d.Set("display_name", portfolioDetail.DisplayName)
+	d.Set("provider_name", portfolioDetail.ProviderName)
+	return nil
+}
+
+func resourceAwsServiceCatalogPortfolioUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).scconn
+	input := servicecatalog.UpdatePortfolioInput{}
+	input.Id = aws.String(d.Id())
```
Ninir: This could be filled directly with:

```go
input := servicecatalog.UpdatePortfolioInput{
    Id: aws.String(d.Id()),
}
```

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
+			Delete: schema.DefaultTimeout(30 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: true,
```
Ninir: This can't be set as ForceNew, this it's computed and won't be mutable from an AWS point-of-view

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: false,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
```
Ninir: This should be Computed: true, as it may be updated on the console, and we would like to set the value on that :)

(Almost every retrieved attributes can be set as Computed)

aws/resource_aws_servicecatalog_portfolio.go
```diff
@@ -0,0 +1,145 @@
...
+		if scErr, ok := err.(awserr.Error); ok && scErr.Code() == "ResourceNotFoundException" {
+			log.Printf("[WARN] Service Catalog Portfolio %q not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Reading ServiceCatalog Portfolio '%s' failed: %s", *input.Id, err.Error())
+	}
+	portfolioDetail := resp.PortfolioDetail
+
+	d.Set("description", portfolioDetail.Description)
```
Ninir: We should also have those attributes: `arn` (computed), `created_time` (computed)

Created time should be set as in:

```go
	if err := d.Set("created_time", portfolioDetail.CreatedTime.Format(time.RFC3339)); err != nil {
		log.Printf("[DEBUG] Error setting created_time: %s", err)
	}
```

aws/resource_aws_servicecatalog_portfolio_test.go
```diff
@@ -0,0 +1,150 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckServiceCatlaogPortfolioDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccCheckAwsServiceCatalogPortfolioResourceConfig_basic1,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("aws_servicecatalog_portfolio.test", "name", "test-1"),
+					resource.TestCheckResourceAttr("aws_servicecatalog_portfolio.test", "description", "test-2"),
+					resource.TestCheckResourceAttr("aws_servicecatalog_portfolio.test", "provider_name", "test-3"),
```
Ninir: When we'll have more attributes, we should also check them here :)

aws/resource_aws_servicecatalog_portfolio_test.go
```diff
@@ -0,0 +1,150 @@
...
+			return fmt.Errorf("Portfolio still exists")
+		}
+	}
+
+	return nil
+}
+
+const testAccCheckAwsServiceCatalogPortfolioResourceConfig_basic1 = `
+resource "aws_servicecatalog_portfolio" "test" {
+  name = "test-1"
```
Ninir: We should definitely randomize test values, so that we don't have conflicts when running nightly tests.
FOr instance, we can [pass](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lambda_function_test.go#L55) a [given value](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lambda_function_test.go#L46), and then [use it](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lambda_function_test.go#L1064)

website/docs/r/servicecatalog_portfolio.html.markdown
```diff
@@ -0,0 +1,43 @@
+---
+layout: "aws"
+page_title: "AWS: aws_servicecatalog_portfolio"
+sidebar_current: "docs-aws-resource-servicecatalog-portfolio"
+description: |-
+  Provides a resource to create a Service Catalog portfolio
+---
+
+# aws\_servicecatalog\_portfolio
```
Ninir: Those antislashes aren't needed anymore due to a recent work by Seth Vargo :)

website/docs/r/servicecatalog_portfolio.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+The following attributes are exported:
+
+* `id` - The ID of the Service Catalog Portfolio.
+
+## Import
+
+Service Catalog Portfolios can be imported using the `service catalog portfolio id`, e.g.
+
+```
+$ terraform import aws_servicecatalog_portfolio.testfolio port-12344321
```
Ninir: Can you also add a test for imports? as in: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/import_aws_vpc_test.go

bw-intuit:
@Ninir updated based on code review.  AcceptLanguage is a request parameter which is not persisted.  I've hard set it to "en" for now.  How you would recommend parameters such as this are processed?

bw-intuit:
@Ninir with the exception of my question around AcceptLanguage, all other comments should be addressed.  Can you review and provide feedback?  Thanks

bw-intuit:
@Ninir thanks for the merge!  I will submit PRs for the remaining resources for service catalog using the same approach.

Ninir:
Awesome! ⭐️ 
Looking forward to this :)

PR #1702, opened 2017-09-19
[WIP][NeedHelp] Bugfix for scaling elasticache cluster with memcached

### LINKED ISSUE #272: aws_elasticache_cluster should force destroy when changing node_type with memcached engine
_This issue was originally opened by @mblakele as hashicorp/terraform#8865. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
$ terraform -v
Terraform v0.7.3
```
### Affected Resource(s)
- aws_elasticache_cluster (with `engine=memcached`)

Original:

``` hcl
resource "aws_elasticache_cluster" "fubar" {
  cluster_id = "fubar"
  engine = "memcached"
  node_type = "cache.t2.micro"
}
```

Updated to:

``` hcl
resource "aws_elasticache_cluster" "fubar" {
  cluster_id = "fubar"
  engine = "memcached"
  node_type = "cache.t2.medium"
}
```
### Expected Behavior

Because AWS doesn't support changing the memcached node type, the terraform plan should destroy and recreate the resource.
### Actual Behavior

Running `terraform plan` shows resource modification:

```
~ aws_elasticache_cluster.fubar
    node_type:       "cache.t2.micro" => "cache.t2.medium"
```

Then `terraform apply` fails with:

```
InvalidParameterCombination: Scaling is not supported for engine memcached
```
### Important Factoids

I can't find clear AWS documentation for this limitation, but it seems to be real.

shadycuz:
This PR aims to fix #272. When an elasticache cluster is scaled up or down terraform apply fails when the engine is memcached. This can be avoided by destroying the cluster and recreating it with the new node size.

I'm new to terraform and golang, I have one other PR but it has not been approved yet.

I did not change the docs as I believe this should work as intended and the docs don't mention scaling doesn't work with memcached.

I didn't write a unit test but should.

### DIFF
--- a/aws/resource_aws_elasticache_cluster.go
+++ b/aws/resource_aws_elasticache_cluster.go
@@ -523,8 +523,8 @@ func resourceAwsElasticacheClusterUpdate(d *schema.ResourceData, meta interface{
 
 	if requestRecreate {
 		log.Printf("[DEBUG] Recreating ElastiCache Cluster (%s), opts:\n%s", d.Id(), req)
-		resourceAwsElasticacheClusterDelete(d,meta)
-		resourceAwsElasticacheClusterCreate(d,meta)
+		resourceAwsElasticacheClusterDelete(d, meta)
+		resourceAwsElasticacheClusterCreate(d, meta)
 	}
 
 	return resourceAwsElasticacheClusterRead(d, meta)
--- a/aws/resource_aws_elasticache_cluster.go
+++ b/aws/resource_aws_elasticache_cluster.go
@@ -426,6 +426,7 @@ func resourceAwsElasticacheClusterUpdate(d *schema.ResourceData, meta interface{
 	}
 
 	requestUpdate := false
+	requestRecreate := false
 	if d.HasChange("security_group_ids") {
 		if attr := d.Get("security_group_ids").(*schema.Set); attr.Len() > 0 {
 			req.SecurityGroupIds = expandStringList(attr.List())
@@ -464,8 +465,13 @@ func resourceAwsElasticacheClusterUpdate(d *schema.ResourceData, meta interface{
 	}
 
 	if d.HasChange("node_type") {
-		req.CacheNodeType = aws.String(d.Get("node_type").(string))
-		requestUpdate = true
+		if d.Get("engine").(string) == "memcached" {
+			req.CacheNodeType = aws.String(d.Get("node_type").(string))
+			requestRecreate = true
+		} else {
+			req.CacheNodeType = aws.String(d.Get("node_type").(string))
+			requestUpdate = true
+		}
 	}
 
 	if d.HasChange("snapshot_retention_limit") {
@@ -513,6 +519,12 @@ func resourceAwsElasticacheClusterUpdate(d *schema.ResourceData, meta interface{
 		if sterr != nil {
 			return fmt.Errorf("Error waiting for elasticache (%s) to update: %s", d.Id(), sterr)
 		}
+	}
+
+	if requestRecreate {
+		log.Printf("[DEBUG] Recreating ElastiCache Cluster (%s), opts:\n%s", d.Id(), req)
+		resourceAwsElasticacheClusterDelete(d,meta)
+		resourceAwsElasticacheClusterCreate(d,meta)
 	}
 
 	return resourceAwsElasticacheClusterRead(d, meta)


aws/resource_aws_elasticache_cluster.go
```diff
@@ -515,6 +521,12 @@ func resourceAwsElasticacheClusterUpdate(d *schema.ResourceData, meta interface{
 		}
 	}
 
+	if requestRecreate {
+		log.Printf("[DEBUG] Recreating ElastiCache Cluster (%s), opts:\n%s", d.Id(), req)
+		resourceAwsElasticacheClusterDelete(d, meta)
+		resourceAwsElasticacheClusterCreate(d, meta)
+	}
```
shadycuz: Pretty crude, not sure if that will cut it, maybe I should add the return here, I think create returns the read function so maybe `return resourceAwsElasticacheClusterCreate(d, meta)`

Ninir:
Hi @shadycuz ,

Thanks for the work you started here! 👍 

The only concern I have is that we should not delete the cluster when it was not planned. 
It goes against the purpose of the plan, as it is an operation made in the backgroun. You could also lose data that are crucial for you, and that you didn't want to lose. 

This is something that should be handled by the user and exposed at plan time.

We could add a `force_upgrade` attribute that defaults to false. If set to true, it would create a new resource.

What do you think?

shadycuz:
@Ninir 

You mean an argument? https://www.terraform.io/docs/providers/aws/r/elasticache_cluster.html

as in `d.Get("force_upgrade").(string) == "true" then RecreateCluster` That way we can put it in the docs that if use use force_upgrade it will always recreate a new resource? 

I like that but I still don't understand how we change the output of plan? I know some resources will state when you run plan that a resource will be created or modified, and sometimes if its modified it will say next to it that the resource will, in fact, be destroyed and recreated. 

How do make `terraform plan` out put that the cluster resource will be destroyed and recreated when we set `force_upgrade` to `true`?

shadycuz:
We could add`force_new` to `true` for `node_type` but that is not accurate for engine type of `redis`. I kinda feel for something like this we should have 2 resources. One for elasticache redis and one for memcached.

shadycuz:
@Ninir No I don't think that PR allows scaling up and down of a cluster. The AWS API does not allow scaling for memcache cluster. That is why it must be deleted and recreated. If you could give me info or a link to documentation for `force_upgrade` attribute I can finish this PR.

shadycuz:
@Ninir @radeksimko If I could get a bump in the right direction I would love to finish this out.

radeksimko:
Hey @shadycuz 
sorry for any confusion caused when scaling the memcache cluster.
It comes down to some improvements we can make in the schema to support this kind of behaviour, specifically `ForceNewWhen` as described in https://github.com/hashicorp/terraform/issues/5895 & WIP implementation in https://github.com/hashicorp/terraform/pull/14887

I believe that destruction of a storage cluster is not an operation that we should _ever_ perform silently (without exposing it to the user via `plan`) - It's way too dangerous, I agree with @Ninir here.

I'm not a huge fan of the special attribute either because it's just not obvious enough from the plan that we're going to destroy the cluster. I'd rather keep it as difficult as it is as it's rarely a good thing to make it simple and easy to _destroy_ things (and potentially loose data and/or cause outage).

We can however advise the user to `taint` the resource in such case, until we have a proper support in the schema, e.g. by catching the error:

```go
if isAWSErr(err, "InvalidParameterCombination", "Scaling is not supported for engine") {
  // TODO: Revisit after https://github.com/hashicorp/terraform/issues/5895
  return fmt.Errorf("%s\nHint: You can taint the resource to mark it for recreation.", err)
}
```

Let me know what you think.

shadycuz:
@radeksimko Yeah you guys are right about not destroying the cluster. [ForceNew](https://github.com/vancluever/terraform/blob/0c0ae3ca7c73bda71db828d0efdeab02b18083a4/helper/schema/resource_diff.go#L292-L321) is now available.

 I still think we would need a new attribute like @Ninir suggested but at least now we can show a new diff which would remind them or another person that the "force_upgrade_memcached" is set. Let me rebase my fork on the latest and give that a try this weekend.

Thanks

radeksimko:
@shadycuz I don't think we need a special attribute anymore.
The linked PR was merged and we now have a mechanism to deal with this "natively" in the schema, see https://github.com/hashicorp/terraform/pull/14887/files#diff-456d1e45f8a8d1585e8228597b9a2176R2845 for an example.

You could have a logic inside `CustomizeDiff` which decides whether to mark `node_type` as ForceNew based on `engine`.

Do you think you could implement that as part of this PR?

bflad:
This PR was re-implemented with `CustomizeDiff` for setting plan-time `ForceNew` in #3857 and merged into master. It will be released in v1.12.0 of the AWS provider, likely at the end of this week.

PR #1710, opened 2017-09-20
Add support for batch_job_definition and batch_job_queue

ashinohara:
## Description
Adding resources `aws_batch_job_definition` and `aws_batch_job_queue`.

Additionally, I found an issue with some of the tests for `aws_batch_compute_environment` which seems to be a condition where the `aws_iam_role_policy_attachment` gets deleted before the compute environment and causes the `aws_batch_compute_environment ` to be a dangling resource.  I have added a note in the documentation about it, and have also fixed the tests by adding a `depends_on` link in the `aws_batch_compute_environment ` to the `aws_iam_role_policy_attachment`.

## Tests
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSBatchJobDefinition'
TF_ACC=1 go test ./aws -v -run=TestAccAWSBatchJobDefinition -timeout 120m
=== RUN   TestAccAWSBatchJobDefinition
--- PASS: TestAccAWSBatchJobDefinition (14.26s)
=== RUN   TestAccAWSBatchJobDefinitionUpdateForcesNewResource
--- PASS: TestAccAWSBatchJobDefinitionUpdateForcesNewResource (22.15s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	36.442s
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSBatchJobQueue'
TF_ACC=1 go test ./aws -v -run=TestAccAWSBatchJobQueue -timeout 120m
=== RUN   TestAccAWSBatchJobQueue
--- PASS: TestAccAWSBatchJobQueue (89.57s)
=== RUN   TestAccAWSBatchJobQueueUpdate
--- PASS: TestAccAWSBatchJobQueueUpdate (104.32s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	193.919s
```

### DIFF
--- a/aws/resource_aws_batch_job_definition.go
+++ b/aws/resource_aws_batch_job_definition.go
@@ -63,7 +63,6 @@ func resourceAwsBatchJobDefinition() *schema.Resource {
 			},
 			"revision": {
 				Type:     schema.TypeInt,
-				Optional: true,
 				Computed: true,
 			},
 			"arn": {
--- a/aws/resource_aws_batch_job_definition.go
+++ b/aws/resource_aws_batch_job_definition.go
@@ -2,10 +2,7 @@ package aws
 
 import (
 	"fmt"
-	"log"
 
-	"crypto/sha1"
-	"encoding/hex"
 	"encoding/json"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -32,10 +29,11 @@ func resourceAwsBatchJobDefinition() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 				StateFunc: func(v interface{}) string {
-					hash := sha1.Sum([]byte(v.(string)))
-					return hex.EncodeToString(hash[:])
+					json, _ := normalizeJsonString(v)
+					return json
 				},
-				ValidateFunc: validateAwsBatchJobContainerProperties,
+				DiffSuppressFunc: suppressEquivalentJsonDiffs,
+				ValidateFunc:     validateAwsBatchJobContainerProperties,
 			},
 			"parameters": {
 				Type:     schema.TypeMap,
@@ -44,7 +42,7 @@ func resourceAwsBatchJobDefinition() *schema.Resource {
 				Elem:     schema.TypeString,
 			},
 			"retry_strategy": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: true,
 				MaxItems: 1,
@@ -70,7 +68,6 @@ func resourceAwsBatchJobDefinition() *schema.Resource {
 			},
 			"arn": {
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 		},
@@ -99,7 +96,7 @@ func resourceAwsBatchJobDefinitionCreate(d *schema.ResourceData, meta interface{
 	}
 
 	if v, ok := d.GetOk("retry_strategy"); ok {
-		input.RetryStrategy = expandJobDefinitionRetryStrategy(v.(*schema.Set))
+		input.RetryStrategy = expandJobDefinitionRetryStrategy(v.([]interface{}))
 	}
 
 	out, err := conn.RegisterJobDefinition(input)
@@ -118,11 +115,15 @@ func resourceAwsBatchJobDefinitionRead(d *schema.ResourceData, meta interface{})
 	if err != nil {
 		return fmt.Errorf("%s %q", err, arn)
 	}
-	d.Set("arn", *job.JobDefinitionArn)
-	d.Set("container_properties", *job.ContainerProperties)
-	d.Set("parameters", job.Parameters)
+	if job == nil {
+		d.SetId("")
+		return nil
+	}
+	d.Set("arn", job.JobDefinitionArn)
+	d.Set("container_properties", job.ContainerProperties)
+	d.Set("parameters", aws.StringValueMap(job.Parameters))
 	d.Set("retry_strategy", flattenRetryStrategy(job.RetryStrategy))
-	d.Set("revision", *job.Revision)
+	d.Set("revision", job.Revision)
 	d.Set("type", job.Type)
 	return nil
 }
@@ -152,10 +153,12 @@ func getJobDefinition(conn *batch.Batch, arn string) (*batch.JobDefinition, erro
 	numJobDefinitions := len(resp.JobDefinitions)
 	switch {
 	case numJobDefinitions == 0:
-		log.Printf("[DEBUG] Job Definition %q is already gone", arn)
 		return nil, nil
 	case numJobDefinitions == 1:
-		return resp.JobDefinitions[0], nil
+		if *resp.JobDefinitions[0].Status == "ACTIVE" {
+			return resp.JobDefinitions[0], nil
+		}
+		return nil, nil
 	case numJobDefinitions > 1:
 		return nil, fmt.Errorf("Multiple Job Definitions with name %s", arn)
 	}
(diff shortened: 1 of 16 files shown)


aws/resource_aws_batch_job_definition.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil, err
+	}
+
+	numJobDefinitions := len(resp.JobDefinitions)
+	switch {
+	case numJobDefinitions == 0:
+		log.Printf("[DEBUG] Job Definition %q is already gone", arn)
+		return nil, nil
+	case numJobDefinitions == 1:
+		return resp.JobDefinitions[0], nil
```
radeksimko: Shouldn't we also check the status of the definition here, since we treat `INACTIVE` definition as "deleted" I think it should be reflected here too, unless there's a strong reason not to do it.

aws/resource_aws_batch_job_definition_test.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+		}
+		return nil
+	}
+}
+
+func testAccCheckJobDefinitionRecreated(t *testing.T,
+	before, after *batch.JobDefinition) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		if *before.Revision == *after.Revision {
```
radeksimko: That's quite clever! 😄  👍

aws/resource_aws_batch_job_definition_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+func testAccCheckBatchJobDefinitionDestroy(s *terraform.State) error {
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_batch_job_definition" {
+			continue
+		}
+		conn := testAccProvider.Meta().(*AWSClient).batchconn
+		js, err := getJobDefinition(conn, rs.Primary.Attributes["arn"])
+		if err == nil {
+			if *js.Status != "INACTIVE" {
```
radeksimko: Really a nitpick, but positive conditionals tend to be easier to read, i.e. `*js.Status == "ACTIVE"`.

aws/resource_aws_batch_job_queue.go
```diff
@@ -0,0 +1,206 @@
...
+				Required: true,
+			},
+			"state": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{batch.JQStateDisabled, batch.JQStateEnabled}, true),
+			},
+			"arn": {
+				Type:     schema.TypeString,
+				Optional: true,
```
radeksimko: Is there any reason why `arn` should be Optional? I don't see it being used anywhere in the CRUD other than in `Set()` which makes me think it should be Computed-only.

aws/resource_aws_batch_job_queue.go
```diff
@@ -0,0 +1,206 @@
...
+func resourceAwsBatchJobQueueUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).batchconn
+
+	updateInput := &batch.UpdateJobQueueInput{
+		ComputeEnvironmentOrder: createComputeEnvironmentOrder(d),
+		JobQueue:                aws.String(d.Get("name").(string)),
+		Priority:                aws.Int64(int64(d.Get("priority").(int))),
+		State:                   aws.String(d.Get("state").(string)),
+	}
+	_, err := conn.UpdateJobQueue(updateInput)
```
radeksimko: Shouldn't we also wait here for the job until it goes from `batch.JQStatusUpdating` to `batch.JQStatusValid`?

aws/resource_aws_batch_job_queue.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return fmt.Errorf(
+			"Error waiting for Job Queue (%s) to be destroyed: %s",
+			sn, err)
+	}
+	d.SetId("")
+	return nil
+}
+
+func createComputeEnvironmentOrder(d *schema.ResourceData) (envs []*batch.ComputeEnvironmentOrder) {
```
radeksimko: Maybe a nitpick, but the interface here seems to be quite greedy - any reason we can't just pass an `[]interface{}` and call `d.Get()` outside of this function since it only needs access to a single field, not all ResourceData?

aws/resource_aws_batch_job_queue.go
```diff
@@ -0,0 +1,206 @@
...
+	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
+		log.Printf("[DEBUG] Trying to delete Job Queue %s", sn)
+		_, err = conn.DeleteJobQueue(&batch.DeleteJobQueueInput{
+			JobQueue: aws.String(sn),
+		})
+		if err == nil {
+			return nil
+		}
+		return resource.RetryableError(err)
+	})
```
radeksimko: Is there any particular reason why this should fail and/or why we should retry? Can we retry on a specific error code, instead of _any_ error?

radeksimko: Also if we really need the retry logic, then we should also catch the `err` coming from it - right now it seems to be shadowed on line 152.

aws/validators.go
```diff
@@ -1420,10 +1420,10 @@ func validateSsmParameterType(v interface{}, k string) (ws []string, errors []er
 	return
 }
 
-func validateBatchComputeEnvironmentName(v interface{}, k string) (ws []string, errors []error) {
+func validateBatchName(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
 	if !(regexp.MustCompile(`^[A-Za-z0-9_]*$`).MatchString(value) && len(value) <= 128) {
-		errors = append(errors, fmt.Errorf("computeEnvironmentName must be up to 128 letters (uppercase and lowercase), numbers, and underscores."))
+		errors = append(errors, fmt.Errorf("Name must be up to 128 letters (uppercase and lowercase), numbers, and underscores."))
```
radeksimko: I have raised https://github.com/terraform-providers/terraform-provider-aws/pull/1760/files to address the generic-ness of this function - I agree it can be still renamed, but either way - do you mind rebasing?

website/docs/r/batch_job_definition.html.markdown
```diff
@@ -0,0 +1,78 @@
...
+
+# aws_batch_job_definition
+
+Provides a Batch Job Definition resource.
+
+## Example Usage
+
+```hcl
+resource "aws_batch_job_definition" "test" {
+	name = "tf-test-batch-job-definition"
```
radeksimko: I think we should replace hyphens with underscores here, per validation function.

aws/resource_aws_batch_job_definition.go
```diff
@@ -0,0 +1,209 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateBatchName,
+			},
+			"container_properties": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				StateFunc: func(v interface{}) string {
```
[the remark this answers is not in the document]
ashinohara: Yes, that was the intent.  Thanks for pointing me to that function.

aws/resource_aws_batch_job_queue.go
ashinohara: I was using it to wait until the Job Queue was disabled, but I think a more elegant solution is to use `WaitForState`.  I am working on that change.

ashinohara:
@radeksimko Thank you very much for your insightful review.  I believe I have addressed all of your comments and made sure that tests are still passing.  Let me know if there is anything else you would like me to change.

aws/resource_aws_batch_job_definition.go
```diff
@@ -0,0 +1,212 @@
...
+			},
+			"type": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice([]string{batch.JobDefinitionTypeContainer}, true),
+			},
+			"revision": {
+				Type:     schema.TypeInt,
+				Optional: true,
```
radeksimko: The `revision` here is still `Optional`, but I will take care of that prior to merging.

pat-nomalab:
Hi,
thanks for this feature
I've managed to create batch compute env with job queue link to it, but I have an issue when i'm trying to upgrade compute env definition "Cannot delete, found existing JobQueue relationship"

PR #1515, opened 2017-08-27
AWS DynamoDB Accelerator Support

gazoakley:
This adds support for managing AWS DynamoDB Accelerator (DAX). It adds two new resources:

- `aws_dax_cluster`
- `aws_dax_subnet_group`

There should probably also be an `aws_dax_parameter_group`, but I think this PR is probably big enough already (I've put the vendoring change/each resource in a separate commit for easier reviewing)

PR #1722, opened 2017-09-21
New Resource: aws_dax_cluster

gazoakley:
This adds support for managing AWS DynamoDB Accelerator (DAX). It adds a new `aws_dax_cluster` resource. There is code ready for a `aws_dax_subnet_group` but it is not present in this PR to simplify review.


gazoakley:
@radeksimko Any chance of a review? This was previously opened as #1515 but split out to only have 1 resource for the first review

website/docs/r/dax_cluster.html.markdown
```diff
@@ -0,0 +1,90 @@
...
+# aws\_dax\_cluster
+
+Provides a DAX Cluster resource.
+
+## Example Usage
+
+```hcl
+resource "aws_dax_cluster" "bar" {
+  cluster_id         = "cluster-example"
+  iam_role_arn       = "arn:aws:iams:us-east-1:012345678999"
```
complexsplit: I think an IAM role ARN looks like `arn:aws:iam:us-east-1:123456789012:role/foo`.

Ninir:
Hi @gazoakley 

Do you think you could move the vendoring part into its own PR, so that it's easier to review this work? :)
Please note that we need to lock the service version so that we always match it exactly, and keep it in-sync with other AWS service versions.

Thanks!

Issue #1724, opened 2017-09-21
aws_iam_role documentation outdated


oddsocksio:
example code in documentaion 
```
data "aws_iam_role" "example" {
  role_name = "an_example_role_name"
}
```

When running this example code the below error is generated
```
Warnings:

  * data.aws_iam_role.example: "role_name": [DEPRECATED] Use `name` instead

1 error(s) occurred:
```

PR #1725, opened 2017-09-21
Update iam_role data source example to provide working example

oddsocksio:
Fix  broken example code for #1724


Ninir:
Hey @oddsocksio 

Thanks for the fix, looks good to me :) 👍

PR #1729, opened 2017-09-22
Bring documented Lightsail Instance Blueprint IDs up to current

slancio:
AWS updated these in June and there are many new options and versions available.


slancio:
Added the updated regions as well

Issue #1691, opened 2017-09-18
aws_vpc_endpoint data source does not expose prefix_list_id, but the equivalent resource does


mikemoate:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
v0.10.4

### Affected Resource(s)
aws_vpc_endpoint data source (https://www.terraform.io/docs/providers/aws/d/vpc_endpoint.html)

### Expected Behavior
aws_vpc_endpoint data source exposes the prefix list id of the endpoint's associated service as an attribute named prefix_list_id. The resource for creating a VPC endpoint behaves like this (https://www.terraform.io/docs/providers/aws/r/vpc_endpoint.html)

### Actual Behavior
There is no prefix_list_id for the aws_vpc_endpoint data source.

PR #1733, opened 2017-09-23
data_source/aws_vpc_endpoint: Exposing prefix_list_id in datasource

stack72:
Fixes: #1691

```
terraform-provider-aws [b-aws-vpce-ds-1691●] % acctests aws TestAccDataSourceAwsVpcEndpoint_
=== RUN   TestAccDataSourceAwsVpcEndpoint_basic
--- PASS: TestAccDataSourceAwsVpcEndpoint_basic (60.06s)
=== RUN   TestAccDataSourceAwsVpcEndpoint_withRouteTable
--- PASS: TestAccDataSourceAwsVpcEndpoint_withRouteTable (64.76s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	124.860s
```

### DIFF
--- a/aws/data_source_aws_vpc_endpoint.go
+++ b/aws/data_source_aws_vpc_endpoint.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/hashicorp/errwrap"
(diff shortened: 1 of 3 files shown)


radeksimko:
Hey @stack72 
is there any particular reason why we should expose the first prefix ID instead of all? i.e. why should this field be different to `route_table_ids`?

stack72:
@radeksimko I followed the same code as `aws_vpc_endpoint` which only exposes the first. As this is the data source of the resource, I thought it should work the same

stack72:
@radeksimko ping on this one - thoughts?

Issue #1697, opened 2017-09-19
aws_ses_event_destination: SNS destination?


hashibot:
_This issue was originally opened by @cmalek as hashicorp/terraform#16113. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I want to set up my SES Configuration Set to send its bounce and complain messages to an SNS queue, as described here: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/event-publishing-add-event-destination-sns.html.

But looking at the docs for the `aws_ses_event_destination` resource (https://www.terraform.io/docs/providers/aws/r/ses_event_destination.html), it seems that only cloudwatch and kinesis firehose are supported. 

Please can we have SNS as an event destination?  Pretty please?

bflad:
This has been released in [terraform-provider-aws version 1.7.0](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#170-january-12-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

inge4pres:
Hi 

was just about to open a new issue and found this: I can't seem to get working this resource.
Whenever I use it in such form

```
resource "aws_ses_domain_identity" "identity" {
  domain = "${var.tld}"
}

resource "aws_ses_event_destination" "sns" {
  name                   = "ses-sns"
  configuration_set_name = "${aws_ses_configuration_set.email.name}"
  enabled                = true
  matching_types         = ["bounce", "complaint", "reject"]

  sns_destination {
    topic_arn = "${aws_sns_topic.email-notifications.arn}"
  }
}

resource "aws_sns_topic" "email-notifications" {
  name = "mail-notifications"
}

resource "aws_ses_configuration_set" "email" {
  name = "mail-set"
}```

I also tried to add an explicit dependency link
```
resource "aws_ses_domain_identity" "identity" {
  domain = "${var.tld}"
  depends_on = ["aws_sns_topic.email-notifications", "aws_ses_event_destination.sns"]
}```

but the result is the SNS notification is not set in the domain, but only in the config set...
Any idea?
Thank you

inge4pres:
Nevermind, I found out why... I needed to use the `aws_ses_identity_notification_topic` resource...

PR #1737, opened 2017-09-23
resource/aws_ses_event_destination: Add support for SNS destinations

stack72:
Fixes: #1697

```
terraform-provider-aws [master●] % acctests aws TestAccAWSSESEventDestination_basic
=== RUN   TestAccAWSSESEventDestination_basic
--- PASS: TestAccAWSSESEventDestination_basic (159.86s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	159.898s
```

### DIFF
--- a/aws/resource_aws_ses_event_destination.go
+++ b/aws/resource_aws_ses_event_destination.go
@@ -19,26 +19,26 @@ func resourceAwsSesEventDestination() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"configuration_set_name": &schema.Schema{
+			"configuration_set_name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
 
-			"enabled": &schema.Schema{
+			"enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
 				ForceNew: true,
 			},
 
-			"matching_types": &schema.Schema{
+			"matching_types": {
 				Type:     schema.TypeSet,
 				Required: true,
 				ForceNew: true,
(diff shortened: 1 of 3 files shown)


aws/resource_aws_ses_event_destination_test.go
```diff
@@ -156,6 +158,10 @@ data "aws_iam_policy_document" "fh_felivery_document" {
     }
 }
 
+resource "aws_sns_topic" "ses_destination" {
+  name = "ses-destination-test"
```
radeksimko: Do you mind randomizing this name here?

aws/resource_aws_ses_event_destination_test.go
```diff
@@ -178,10 +184,22 @@ resource "aws_ses_event_destination" "cloudwatch" {
...
-  cloudwatch_destination = {
+  cloudwatch_destination {
     default_value = "default"
 	dimension_name = "dimension"
 	value_source = "emailHeader"
   }
 }
+
+resource "aws_ses_event_destination" "sns" {
+  name = "event-destination-sns",
```
radeksimko: Likewise ^ do you mind randomizing it?

aws/resource_aws_ses_event_destination_test.go
```diff
@@ -26,6 +26,8 @@ func TestAccAWSSESEventDestination_basic(t *testing.T) {
 						"aws_ses_event_destination.kinesis", "name", "event-destination-kinesis"),
 					resource.TestCheckResourceAttr(
 						"aws_ses_event_destination.cloudwatch", "name", "event-destination-cloudwatch"),
+					resource.TestCheckResourceAttr(
+						"aws_ses_event_destination.sns", "name", "event-destination-sns"),
```
radeksimko: Nitpick - but do you mind adding check for `sns_destination.0.topic_arn` as that is the field being added and that's the one that should also be tested?

stack72:
@radeksimko so I looked at this and the reason the original tests are not actually randomized is that there is nothing happening in the Read func - therefore, when we check that a destroy happens, we are actually checking the hardcoded SES destination to make sure it no longer exists

Thoughts?

P.

caJaeger:
@radeksimko Is this dead in the water? I would love to see this in the next release...

sefaYavuz:
+1 need this!

pccowboy:
@stack72 I have what might be valid tests for what @radeksimko has asked for - how best to get them into this PR?

radeksimko:
There are other things to improve on that resource - e.g. turning all the `MaxItems: 1` `TypeSet`s into `TypeList`s to make nested fields easier to reference in tests and elsewhere, but that's certainly outside of scope of this PR.

Merging now...

bflad:

aws/resource_aws_ses_event_destination.go
```diff
@@ -53,20 +53,21 @@ func resourceAwsSesEventDestination() *schema.Resource {
 				Type:          schema.TypeSet,
 				Optional:      true,
 				ForceNew:      true,
-				ConflictsWith: []string{"kinesis_destination"},
+				MaxItems:      1,
```
bflad: Interestingly enough, this resource actually did appear to support multiple `cloudwatch_destination` previously, even if it wasn't appropriately acceptance tested. This change will likely be reverted in #6690

PR #1740, opened 2017-09-24
Optional password hashing before writing to state - RDS DB

gspencerfabian:
In reference to existing PR: #1084

**Summary**
Gives users the option to hash their RDS master password before it's written clear text in the state file. 

**Comments**
Unable to utilize SchemeStateFunc since ResourceData is not available from within. This is why all the code is in the Create/Update functions. This could all be fixed if we had access to ResourceData/context inside SchemaStateFunc.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -581,9 +581,10 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			CopyTagsToSnapshot:      aws.Bool(d.Get("copy_tags_to_snapshot").(bool)),
 		}
 
-		// If hash_password config is true, the create switch will hash the password.
-		// Otherwise no work is done.
-		managePasswordHash(d, "create")
+		// If hash_password config is true, hash the password
+		if d.Get("hash_password").(bool) {
+			d.Set("password", hashPassword(d.Get("password")))
+		}
 
 		attr := d.Get("backup_retention_period")
 		opts.BackupRetentionPeriod = aws.Int64(int64(attr.(int)))
@@ -954,7 +955,7 @@ func resourceAwsDbInstanceUpdate(d *schema.ResourceData, meta interface{}) error
 	}
 	if d.HasChange("password") || d.HasChange("hash_password") {
 		d.SetPartial("password")
-		requestUpdate, req.MasterUserPassword = managePasswordHash(d, "update")
+		requestUpdate, req.MasterUserPassword = managePasswordHashUpdate(d)
 	}
 	if d.HasChange("multi_az") {
 		d.SetPartial("multi_az")
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -57,6 +57,12 @@ func resourceAwsDbInstance() *schema.Resource {
 				Sensitive: true,
 			},
 
+			"hash_password": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
 			"engine": {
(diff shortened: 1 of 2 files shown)


iceycake:
@Ninir can you check on this PR to see if matching with what we had discussed in https://github.com/terraform-providers/terraform-provider-aws/pull/1084 ?

gspencerfabian:
@Ninir have you had a chance to review this?

gspencerfabian:
@Ninir @radeksimko @bflad this PR has been open for a while, it would be nice to get some closure. I'm at a new company and there is still a need for this. Storing clear text passwords in the state file is a serious security concern.

bflad:
Hi @gspencerfabian 👋 Thank you for submitting this so long ago and apologies for the radio silence on the matter. Terraform as a whole has had a relatively unsatisfying security story when it comes to unavoidable sensitive values being in the Terraform state, such as this case. The silence on the matter more relates to some internal discussions that we have had around the subject where sometimes there seems to be path forward, but priorities change. I apologize for that. It would be more fair to you as the contributor and those following this to be more transparent.

The maintainers of the Terraform AWS Provider, other Terraform Providers, and the Terraform CLI/Core team all recognize that _something_ needs to be done in this space, but we have concerns about attempting to implement conditional logic like this as the practice cannot scale in a maintainable way across all providers, resources, and attributes where sensitive values may live. Different people will have differing opinions about what it potentially sensitive, as we have seen even when it comes to marking attributes as `Sensitive: true` (some are more obvious like here; however some are not such as #8820 for marking `aws_s3_bucket_object` content as sensitive). This type of value hashing also can introduce potential confusion for operators not familiar with the practice being implemented in certain places, but not others, if they try to reference the value to connect to the database. Terraform resources will also happily pass along the hashed value instead of the real value on updates (since that is all they see), which is also problematic as seen in the few other resources that implement this practice currently.

The general recommendation, barring better solutions in this space, is to encrypt your Terraform state and protect access to it. Out of the remote state backends, the Terraform Cloud and S3 backends support encryption at rest (in addition to encryption in transit), while other major ones like AzureRM and GCP I believe have open pull requests to implement similar encryption at rest functionality. Another specific recommendation in this case would be separating access to this specific value in state by using the [`aws_db_instance` data source](https://www.terraform.io/docs/providers/aws/d/db_instance.html) in Terraform configurations with less privileged access, which cannot retrieve/save the master password in the state.

I personally think that, ultimately, a solution for this problem needs to allow operators to decide which values are sensitive themselves for their own security/compliance needs, while also enabling passthrough encryption, protected by role-based access control of the encryption system itself, so Terraform can operate as expected while protecting the content. This is a large undertaking though, affecting a large amount of the Terraform Core and Terraform Plugin SDK handling of refreshing values and performing drift detection for differences.

My opinions aside, the current references and further tracking of this problem can be found here:

- [Terraform CLI State documentation on sensitive data](https://www.terraform.io/docs/state/sensitive-data.html)
- [Extending Terraform documentation on sensitive data](https://www.terraform.io/docs/extend/best-practices/sensitive-state.html)
- [Version 3.0.0 Issue to Remove Hashing from State Storage](https://github.com/terraform-providers/terraform-provider-aws/issues/13406)
- [Upstream Terraform Core Issue for State Encryption](https://github.com/hashicorp/terraform/issues/516)

I apologize if this continues to be an unsatisfactory response to this situation and that in light of the above, we are going to close this pull request for now. If you would like to discuss this more, my best suggestion is adding constructive feedback or suggestions in https://github.com/hashicorp/terraform/issues/516 -- the Product Manager and Engineering Manager for the Terraform Core team (which would likely drive any potential generic solution) actively monitor highly upvoted issues there.

PR #1745, opened 2017-09-25
Force s3 object content type to retrieve the body of known object type

lpbedard:
Force interpretation of the resource to be of a different content type (useful to get the body of non-binary resource that are not considered as text even if they are, i.e. certificates)

Use case:
We want to be able to read the content of a certificate from an s3 object. However, the mime type being application/x-x509-ca-cert, even though it is in text format, the regular expression blocks the retrieval of the body. This parameter tells terraform to process this content as what you are expecting.


julienduchesne:
This would be ready to merge. Any update on this?

julienduchesne:
This may also fix #1323

PR #1753, opened 2017-09-26
New Resource: aws_wafregional_rule

knehring:
This PR is a port of the originally opened by @yusukegoto at hashicorp/terraform#13710, with some minor tweaks based off of conversations in other PRs for wafregional

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -486,6 +486,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_waf_sql_injection_match_set":              resourceAwsWafSqlInjectionMatchSet(),
 			"aws_wafregional_byte_match_set":               resourceAwsWafRegionalByteMatchSet(),
 			"aws_wafregional_ipset":                        resourceAwsWafRegionalIPSet(),
+			"aws_wafregional_rule":                         resourceAwsWafRegionalRule(),
 			"aws_batch_compute_environment":                resourceAwsBatchComputeEnvironment(),
 		},
 		ConfigureFunc: providerConfigure,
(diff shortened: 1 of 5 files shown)


radeksimko:
Hi @knehring 
this PR was already re-raised in https://github.com/terraform-providers/terraform-provider-aws/pull/1045 and there are other WAF regional PRs pending some work based on feedback I provided in https://github.com/terraform-providers/terraform-provider-aws/pull/1013 or https://github.com/terraform-providers/terraform-provider-aws/pull/1014

Do you have any plans to implement the feedback?

If not, that's 👌  - just FYI given that this PR was re-opened couple of times and always went stale I will go ahead and implement the changes on top of @yusukegoto's PRs (because he we the first one raising PRs for this).

giuliocalzolari:
Hi do you plan to merge it?

knehring:
All of the changes that you linked in the above PR's have been implemented in this PR, its been awhile since I looked at this, but from looking at the feedback on those PR some of them do not apply in this case.

1. No plural tuple/tuples (not used in this)
2. This is not missing any fields afaik
3. The Update code uses existing code in the global waf rule for diffing the updates and should be good here
4. the tests reflect what is currently there for the global WAF Rule, if I am missing any tests here please let me know

I don't think there need to be any more changes to this bit of code. The only other thing I think would be beneficial is to have the other resources implemented as well. That shouldn't stop this PR from being implemented because this would allow the use of the byte match set and ipsets that are already implemented. @radeksimko

radeksimko:
https://github.com/terraform-providers/terraform-provider-aws/pull/3756 which added this resource was just merged. Thanks to everyone involved 🎉

PR #1786, opened 2017-10-01
Add AWS SES DKIM resource

### LINKED ISSUE #1347: DKIM for SES
### Terraform Version
Terraform v0.10.0

### Affected Resource(s)
- aws_ses_domain_identity

### Expected Behavior
It doesn't look like the AWS provider allows you to generate DKIM settings for a domain.

### References
- https://docs.aws.amazon.com/ses/latest/DeveloperGuide/easy-dkim.html

arr-dev:
Added aws_ses_dkim resource, according to http://docs.aws.amazon.com/ses/latest/DeveloperGuide/easy-dkim-dns-records.html

Transfered the PR from hashicorp/terraform#14831

Closes #1347


aws/resource_aws_ses_domain_dkim.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func resourceAwsSesDomainDkim() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesDomainDkimCreate,
+		Read:   resourceAwsSesDomainDkimRead,
+		Delete: resourceAwsSesDomainDkimDelete,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
```
handlerbot: I am not sure what the `arn` field is used for in this resource? I didn't find any references to constructing an ARN for `dkim/` as you do below, and I'm not sure where this is used. I was able to get my local use of this resource working fine even after removing the all `arn`-related code.

aws/resource_aws_ses_domain_dkim.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/aws/aws-sdk-go/service/ses"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsSesDomainDkim() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesDomainDkimCreate,
+		Read:   resourceAwsSesDomainDkimRead,
+		Delete: resourceAwsSesDomainDkimDelete,
+
```
handlerbot: Can we add an `Importer` here so importing works?

```
},
```

The above worked for me.

aws/resource_aws_ses_domain_dkim.go
```diff
@@ -0,0 +1,89 @@
...
+	verificationAttrs, ok := response.DkimAttributes[domainName]
+	if !ok {
+		log.Printf("[WARN] Domain not listed in response when fetching verification attributes for %s", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("arn", fmt.Sprintf("arn:%s:ses:%s:%s:dkim/%s", meta.(*AWSClient).partition, meta.(*AWSClient).region, meta.(*AWSClient).accountid, d.Id()))
+	d.Set("dkim_tokens", verificationAttrs.DkimTokens)
```
handlerbot: Most important: **this resource did not work with this line as written!**

Without changing this, the tokens weren't actually written to the state file under `dkim_tokens`, and the next plan attempted to update the resource, which isn't possible, so Terraform errored out.

I needed to wrap the `DkimTokens` in a `aws.StringValueSlice` to get the `[]*string` converted to `[]string`, as (sort of) documented here: https://godoc.org/github.com/aws/aws-sdk-go/aws

```
d.Set("dkim_tokens", aws.StringValueSlice(verificationAttrs.DkimTokens))
```

aws/resource_aws_ses_domain_dkim.go
arr-dev: I've based this resource on `aws_ses_domain_identity` which computes `arn` the same way.
I understand that it's not used, but I wasn't sure about leaving it out.

I'll remove it.

arr-dev:
@handlerbot Yes, you are right and I'd be glad to complete this PR.
I'll do follow-up commits in the next couple of days.

arr-dev:
I've added a commit with doc fixes, 
but am having issues running acceptance tests, even without any changes of my own.
I've tried running it for single resource only, but that doesn't seem to work.

`TEST="aws/resource_aws_ses_domain_identity_test.go" make testacc`

Not sure if I'm running it correctly.

arr-dev:
Managed to run Acceptance tests, so I've added tests which verify tokens state.

If anyone needs it, command to run specific acceptance test:
`make testacc TESTARGS='-run=TestAccAwsSESDomainDkim'`

PR #1805, opened 2017-10-03
data/aws_iam_user: Add a data source for IAM user

slaunay:
# New feature
This data source can be used to fetch information about a specific IAM user.
By using this data source, you can reference IAM user properties without having to hard code ARNs or unique IDs as input (e.g. referencing a user `aws:userId` when creating an IAM policy).

# Code change
- adding `aws_iam_user` data source
- adding acceptance test cases (existing user and non existing user)
- adding documentation for the new data source

# Testing done
```
TF_ACC=1 go test -run=TestAccAWSDataSourceIAMUser_basic -v
=== RUN   TestAccAWSDataSourceIAMUser_basic
--- PASS: TestAccAWSDataSourceIAMUser_basic (11.09s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	11.118s
```

# Misc
The code of this contribution was based a lot on #1140, thanks @stack72.

### DIFF
--- a/aws/data_source_aws_iam_user.go
+++ b/aws/data_source_aws_iam_user.go
@@ -1,0 +1,54 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsIAMUser() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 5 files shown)


slaunay:
👍 @radeksimko Thanks for making those changes and merging them.

The failure test helped me find out that the `resp` variable is never `nil` (after cloning the `d/iam_group` feature) when looking at the code coverage:
https://github.com/terraform-providers/terraform-provider-aws/blob/5815c0d72834326f2315c6e9ceaab460f93d1b77/aws/data_source_aws_iam_group.go#L46-L52

But I agree that the regexp makes the test brittle.

Issue #1618, opened 2017-09-07
Support for network load balancer


robinbowes:
AWS recently announced a new load balancer: the network load balancer

https://aws.amazon.com/blogs/aws/new-network-load-balancer-effortless-scaling-to-millions-of-requests-per-second/

Would be good to get support for this into this provider

karnauskas:
👍

aakash-garg-olx:
+1

bentterp:
The ability to build a load-balancer from a number of EIP's is particularly interesting for services where moving to another IP-address causes huge operational headaches.

et304383:
I had a feeling this was going to be requested right away.

NLB is amazing for supporting a static IP load balanced layer.  Definitely looking forward to terraform support.

enderv:
Reading over the docs looks like pretty much all the ALB stuff can be copied over and just renamed. Only need to add support for EIP and change the protocol options to TCP only on the target group. I will try to take a stab at this over the weekend but I have no idea how far I will make it.

robinbowes:
I wondered whether this might be the case.

I also wondered whether it might be an idea to support both types with one code base, ie. have a `type` field to choose between `alb` and `nlb`?

enderv:
I thought about that as well, but I think if we go that route would have to rename all the alb resources to something more generic and I'm not sure on the policy of breaking changes. I am in favor of just creating new resource type for nlb just in case they diverge more in the future but thats just my thoughts.

robinbowes:
BTW @stack72 recently tweeted that he was looking at this.

bubunyo:
+1

kcormier:
Excited to see this merged!  Do we have an estimate of when a new release will be cut?

ben-ms:
+++1 Looking to leverage nlb for multiple service offerings and this issue is blocking the adoption of terraform.

bentterp:
@ben-ms But it's not blocking anymore, it's merged and released.

robinbowes:
@ben-ms It was released in v1.1.0 of the aws provider: https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#110-october-16-2017

bacoboy:
While merged, I believe there are still issues in #1838.
Somebody seems to be working on a ton of fixes in #1948 so I am patiently watching those 2 items.  Cheers

robomon1:
[...]

PR #1629, opened 2017-09-09
provider/aws: Add support for Network Loadbalancers

stack72:
Fixes: #1618

In terraform, we had the idea of an alb. In AWS this doesn't exist. ALBs
are actually Load balancers of type `application`

Therefore, the first part of this PR adds a new parameter to ALBs called
`load_balancer_type`. We default this to `application` to follow the
same idea as the current behaviour

The next part of the PR will then change the idea of an alb -> lb

In order to preserve backwards compatibility, we have added another
resource name to the same schema type. This means we effectively have an
alias of aws_alb and aws_lb. This includes updating *all* of the tests
to make sure and remove the idea of ALB and rename to LB and then we
will add a check to make sure we can still check that an ALB can be
created in the old resource

PR #1806, opened 2017-10-04
provider/aws: Add support for Network Loadbalancers

stack72:
Fixes: #1618

In terraform, we had the idea of an alb. In AWS this doesn't exist. ALBs
are actually Load balancers of type `application`

Therefore, the first part of this PR adds a new parameter to ALBs called
`load_balancer_type`. We default this to `application` to follow the
same idea as the current behaviour

The next part of the PR will then change the idea of an alb -> lb

In order to preserve backwards compatibility, we have added another
resource name to the same schema type. This means we effectively have an
alias of aws_alb and aws_lb. This includes updating *all* of the tests
to make sure and remove the idea of ALB and rename to LB and then we
will add a check to make sure we can still check that an ALB can be
created in the old resource


stack72:
Not sure what happened in #1629 - my remote branch got deleted so i pushed a new version and it closed the PR

radeksimko:
> When you are happy it's the same, I will make all of the changes you asked for

I'm pretty confident it's the same, based on patch I was able to pull from my local branch when reviewing the old PR: https://gist.github.com/radeksimko/f0e953ba5bae73c05732648d5bafc032

So feel free to make the changes.

stack72:
Hi @radeksimko 

Ok, I made the docs changes. The issue with this line:

```
if d.Get("load_balancer_type").(string) != "network" && d.HasChange("idle_timeout") {
```

We set a default value for `idle_timeout` and the way that Create calls Update func, we don't want to set that value if the type of loadbalancer is `network`

Does this make more sense? Do you think there is a better way to handle this?

P.

aws/resource_aws_lb_listener.go
```diff
@@ -252,15 +252,15 @@ func resourceAwsAlbListenerDelete(d *schema.ResourceData, meta interface{}) erro
...
 	if port < 1 || port > 65536 {
 		errors = append(errors, fmt.Errorf("%q must be a valid port number (1-65536)", k))
 	}
 	return
 }
 
-func validateAwsAlbListenerProtocol(v interface{}, k string) (ws []string, errors []error) {
+func validateAwsLbListenerProtocol(v interface{}, k string) (ws []string, errors []error) {
 	value := strings.ToLower(v.(string))
 	if value == "http" || value == "https" {
```
jwinter: Network Load Balancers only support TCP listeners, so I think line 265 needs to be updated to allow for a value of "tcp".

https://github.com/aws/aws-sdk-go/blob/master/service/elbv2/api.go#L3294-L3296

stack72: @jwinter you are indeed correct - just pushed this change!

stack72:
@radeksimko I agree - let's certainly chat about this and see if we can solve this in another fashion as a followup

radeksimko:
@handlerbot I think we first need to implement mechanism for deprecating resources. Then we're _actually_ going to mark those old names as deprecated and they'll be reported as such when used in configs. This is work to be done in core schema package. Until that's done, aliases will remain in place as we want to avoid surprises.

There is no migration tool/mechanism available at this point for any deprecations, although we did discuss the idea of having one internally within the team. If we don't have one at the time of real deprecation then yes, it will require `terraform state mv` surgery + config changes. It is highly likely that we'll present a semi-automated way leveraging unix tools like `find`/`sed` etc. but we're talking future here. I'm aware of how inconvenient would such migration be.

I don't have any timelines to share around any of the above, but I'm pretty sure that from introducing resource deprecation (in the schema package) it would take at least a couple of months, or rather half a year until the aliases are removed.

aws/resource_aws_lb_target_group.go
```diff
@@ -410,15 +410,15 @@ func validateAwsAlbTargetGroupHealthCheckProtocol(v interface{}, k string) (ws [
...
 	if port < 1 || port > 65536 {
 		errors = append(errors, fmt.Errorf("%q must be a valid port number (1-65536)", k))
 	}
 	return
 }
 
-func validateAwsAlbTargetGroupProtocol(v interface{}, k string) (ws []string, errors []error) {
+func validateAwsLbTargetGroupProtocol(v interface{}, k string) (ws []string, errors []error) {
 	protocol := strings.ToLower(v.(string))
 	if protocol == "http" || protocol == "https" {
```
cbarbara-okta: this needs to allow "tcp", just like your change to `validateAwsLbListenerProtocol`. "tcp" is a valid network load balancer target group protocol.

PR #1809, opened 2017-10-04
Fix ECS cluster ARN attribute name/value for aws_batch_compute_environment resource

rxacevedo:
## Description
Small change here, the ECS Cluster ARN attribute was named `ecc_cluster_arn` and was actually being assigned the ARN of the compute environment instead of the cluster. I have updated the code to use the ECS cluster ARN instead, as well updating the attribute name in the code/documentation to `ecs_cluster_arn`.

## Test
```
AWS_ACCESS_KEY_ID=$(aws configure get aws_access_key_id) AWS_SECRET_ACCESS_KEY=$(aws configure get aws_secret_access_key) AWS_REGION=us-east-1 make testacc TEST=./aws TESTARGS='-run=TestAccAWSBatchComputeEnvironment_createUnmanaged'
TF_ACC=1 go test ./aws -v -run=TestAccAWSBatchComputeEnvironment_createUnmanaged -timeout 120m
=== RUN   TestAccAWSBatchComputeEnvironment_createUnmanaged
--- PASS: TestAccAWSBatchComputeEnvironment_createUnmanaged (52.43s)
=== RUN   TestAccAWSBatchComputeEnvironment_createUnmanagedWithComputeResources
--- PASS: TestAccAWSBatchComputeEnvironment_createUnmanagedWithComputeResources (41.30s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	93.764s
```

### DIFF
--- a/aws/resource_aws_batch_compute_environment.go
+++ b/aws/resource_aws_batch_compute_environment.go
@@ -121,6 +121,11 @@ func resourceAwsBatchComputeEnvironment() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"ecc_cluster_arn": {
+				Type:       schema.TypeString,
+				Computed:   true,
+				Deprecated: "Use ecs_cluster_arn instead",
+			},
 			"ecs_cluster_arn": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -266,6 +271,7 @@ func resourceAwsBatchComputeEnvironmentRead(d *schema.ResourceData, meta interfa
 	}
 
 	d.Set("arn", computeEnvironment.ComputeEnvironmentArn)
+	d.Set("ecc_cluster_arn", computeEnvironment.EcsClusterArn)
 	d.Set("ecs_cluster_arn", computeEnvironment.EcsClusterArn)
 	d.Set("status", computeEnvironment.Status)
 	d.Set("status_reason", computeEnvironment.StatusReason)
--- a/aws/resource_aws_batch_compute_environment.go
+++ b/aws/resource_aws_batch_compute_environment.go
@@ -121,7 +121,7 @@ func resourceAwsBatchComputeEnvironment() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"ecc_cluster_arn": {
+			"ecs_cluster_arn": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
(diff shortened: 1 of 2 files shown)


radeksimko:
Hi @rxacevedo 
thanks for spotting this and for raising the PR.

In principle the change makes sense, but there may be people who are (for any weird) reason using the wrong field and this PR contains breaking change in that sense.

Would you therefore mind keeping the existing field there and marking it as `Deprecated: "Use ecs_cluster_arn or arn"`? We can remove it in the next major version.

FYI: We made a promise to stick to semantic versioning of providers from 1.0.0, see https://www.hashicorp.com/blog/hashicorp-terraform-provider-versioning/

Thanks.

rxacevedo:
Hi @radeksimko - thanks for the prompt response. No problem - I've gone ahead and added `ecc_cluster_arn` back as an attribute on the `Schema` and set it to the same value as `ecs_cluster_arn` and added a `Deprecated` attribute as I've seen done in other classes. In my local usage of this newly compiled version of the provider, I am not seeing the warnings being printed to the console when I use the `Deprecated` attribute. Is there extra logic that I need to add in order to have the warning print out, or should adding the attribute be enough? Example:

```
There are warnings related to your configuration. If no errors occurred,
Terraform will continue despite these warnings. It is a good idea to resolve
these warnings in the near future.

Warnings:

  * aws_appautoscaling_policy.in: "adjustment_type": [DEPRECATED] Use step_scaling_policy_configuration -> adjustment_type instead
  * aws_appautoscaling_policy.in: "cooldown": [DEPRECATED] Use step_scaling_policy_configuration -> cooldown instead
  * aws_appautoscaling_policy.in: "metric_aggregation_type": [DEPRECATED] Use step_scaling_policy_configuration -> metric_aggregation_type instead
  * aws_appautoscaling_policy.in: "step_adjustment": [DEPRECATED] Use step_scaling_policy_configuration -> step_adjustment instead
  * aws_appautoscaling_policy.out: "adjustment_type": [DEPRECATED] Use step_scaling_policy_configuration -> adjustment_type instead
  * aws_appautoscaling_policy.out: "cooldown": [DEPRECATED] Use step_scaling_policy_configuration -> cooldown instead
  * aws_appautoscaling_policy.out: "metric_aggregation_type": [DEPRECATED] Use step_scaling_policy_configuration -> metric_aggregation_type instead
  * aws_appautoscaling_policy.out: "step_adjustment": [DEPRECATED] Use step_scaling_policy_configuration -> step_adjustment instead
```

I used `resource_aws_appautoscaling_policy.go` as a reference but am not finding any logic that explicitly prints these warnings, so was wondering if you could point me in the direction. Thanks!

rxacevedo:
@radeksimko Actually - another observation, this is an exported attribute and not an input on the resource itself, so perhaps there is no warning to print since there is no user input for this? Let me know if allowing both attributes to exist/be exported and having one marked as `Deprecated` via that attribute is sufficient. Thanks!

PR #1827, opened 2017-10-06
aws_route53_record alias name documentation (must be all in lowercase letters)

ecbrodie:
This PR is a response to an issue I faced when working with the`aws_route53_record` resource, similar to what was described in https://github.com/hashicorp/terraform/issues/10004.

The problem for me was that my ELB had a name containing some uppercase letters. This meant that this route53 record's alias name contained uppercase letters and thus semantically violated the constraints of a domain name (must be in lowercase letters).

My hope is that this small additional bit of documentation can help a Terraform developer who faces this problem to more quickly find a possible solution (ie, use the `lower` function) instead of having to spend additional time Google-ing around for a solution.

I hope this helps. Happy Hacktoberfest ;)

### DIFF
--- a/website/docs/r/route53_record.html.markdown
+++ b/website/docs/r/route53_record.html.markdown
@@ -113,7 +113,7 @@ Exactly one of `records` or `alias` must be specified: this determines whether i
 
 Alias records support the following:
 
-* `name` - (Required) DNS domain name for a CloudFront distribution, S3 bucket, ELB, or another resource record set in this hosted zone.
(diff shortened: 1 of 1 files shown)


ecbrodie:
@radeksimko is a valid bugfix to just have Terraform automatically run the `lower` function in the background?

radeksimko:
@ecbrodie Pretty much, I think the best way to fix this would be to add `DiffSuppressFunc` to the field. I can't find an example of lowercase/uppercase in the AWS provider, but take a look at how do it in AzureRM:

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_container_group.go#L39

radeksimko:
With that do you mind me closing this PR?

ecbrodie:
Go ahead

PR #1830, opened 2017-10-06
Add s3_backup_mode option in Firehose Redshift destination

ApsOps:
http://docs.aws.amazon.com/cli/latest/reference/firehose/create-delivery-stream.html

### DIFF
--- a/aws/resource_aws_kinesis_firehose_delivery_stream.go
+++ b/aws/resource_aws_kinesis_firehose_delivery_stream.go
@@ -41,6 +41,59 @@ func cloudWatchLoggingOptionsSchema() *schema.Schema {
 	}
 }
 
+func S3BackupConfigurationSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
+		MaxItems: 1,
+		Optional: true,
+		Computed: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"bucket_arn": {
+					Type:     schema.TypeString,
+					Required: true,
+				},
+
+				"buffer_size": {
+					Type:     schema.TypeInt,
+					Optional: true,
+					Default:  5,
+				},
+
+				"buffer_interval": {
+					Type:     schema.TypeInt,
+					Optional: true,
+					Default:  300,
+				},
+
+				"compression_format": {
+					Type:     schema.TypeString,
+					Optional: true,
+					Default:  "UNCOMPRESSED",
+				},
+
+				"kms_key_arn": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validateArn,
+				},
+
+				"role_arn": {
+					Type:     schema.TypeString,
+					Required: true,
+				},
+
+				"prefix": {
+					Type:     schema.TypeString,
+					Optional: true,
+				},
+
+				"cloudwatch_logging_options": cloudWatchLoggingOptionsSchema(),
+			},
+		},
+	}
+}
+
 func processingConfigurationSchema() *schema.Schema {
 	return &schema.Schema{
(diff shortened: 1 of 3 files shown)


ApsOps:
@radeksimko Hi. I had tried adding tests, but unable to check them completely since even `make test` is failing for me locally, which is passing in the CI.
I didn't find any specific steps to do except running `make test`.

`go version go1.9 darwin/amd64`
`GOPATH` is set and `$GOPATH/bin` is added in `$PATH`.

ApsOps:
@radeksimko Here's the output of failed tests. - https://gist.github.com/ApsOps/123386f091924aafe80d07f628f3198f

Do I need to setup something to mock the EC2 metadata API?

ApsOps:
Vendor packages for aws-sdk-go (for firehose) needs to be updated for this to work.

ApsOps:
@radeksimko Added the acceptance tests. These are passing for me.
Please let me know if I missed something.

Thanks a lot for helping out with this!

aws/resource_aws_kinesis_firehose_delivery_stream.go
```diff
@@ -41,6 +41,59 @@ func cloudWatchLoggingOptionsSchema() *schema.Schema {
 	}
 }
 
+func S3BackupConfigurationSchema() *schema.Schema {
+	return &schema.Schema{
+		Type:     schema.TypeSet,
```
ApsOps: @radeksimko I have a doubt here. When do we use `TypeSet` vs `TypeList`?
I see `S3ConfigurationSchema` has `TypeList` with `MaxItems: 1`. Is this significant only during type assertion?

radeksimko: In other words it would be better to use `TypeList` here.

Can't we call this `s3ConfigurationSchema` and use it in `"s3_configuration"` too to reduce the repetition a bit?

ApsOps: >Can't we call this s3ConfigurationSchema and use it in "s3_configuration" too to reduce the repetition a bit?

That's what I was going for and noticed that the schemaTypes are different.

ApsOps: > Set is used when order doesn't matter

@radeksimko Just curious. Why do we use `TypeList` here then? The order doesn't matter here, right?

aws/resource_aws_kinesis_firehose_delivery_stream.go
```diff
@@ -467,6 +536,34 @@ func createS3Config(d *schema.ResourceData) *firehose.S3DestinationConfiguration
 	return configuration
 }
 
+func createS3BackupConfig(d *schema.ResourceData) *firehose.S3DestinationConfiguration {
```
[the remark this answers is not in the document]
ApsOps: @radeksimko We're not using "expanders" for `createS3Config` and `createExtendedS3Config` - few lines above and below this snippet.

aws/resource_aws_kinesis_firehose_delivery_stream.go
radeksimko: Correct, it doesn't, but also there's really nothing to order 😃  as we only ever have a single item in the list and `TypeList` makes it more convenient for referencing purposes, i.e. `field_name.0.nested_field` instead of `field_name.<computed-index>.nested_field`.

aws/resource_aws_kinesis_firehose_delivery_stream.go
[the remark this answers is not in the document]
radeksimko: I didn't notice that previously - good catch - we should certainly fix that, but I'd prefer to do it in a separate PR which may come afterwards to keep the context & LOC low here. 😉

PR #1858, opened 2017-10-11
New Resource: aws_iot_topic_rule

abingham:
This PR is fundamentally just the "IoT rules" subset of the [large PR created by @jhedev](https://github.com/hashicorp/terraform/pull/6961/) a while back. My hope is to help push this work forward by providing it in digestible chunks. 

This seems to work insofar as I'm able to manage "aws_iot_topic_rule" resources: I can create them and destroy them, and they seem to be what I expect on the AWS side. The test suite provided in the original PR also seems to pass (though I don't know how substantial it is).

NB: I'm a complete noob at go and the terraform code base. Let me know if there are problems that need to be fixed, but don't be surprised if I have questions :)

TODO:
- [x] Flatten actions
- [x] Add the import
- [x] Update acceptance tests to check destination attributes
- [x] Add import test


abingham:
This is apropos to #143

rob-smallshire:
I notice that this PR #1858 seems equivalent to #1723. If nothing else this demonstrates a certain level of demand for this feature.

cwardcode:
@rob-smallshire The lack of expediency to support features by terraform is one of the major reasons we ended up going with CloudFormation in the end.

rob-smallshire:
Is there anything the community can do to expedite either this PR #1858 or PR #1723 ?

Ninir:
Hi @abingham 

Thanks for the work here. Can you expose in why this seems easier? it's just me trying to make sure I understand correctly what's going on here! :)

The number of additions is the same, so before looking at a given work, it would probably be better for us to get the full explanation.

Thanks, will review & merge IoT Rules this week! 🚀

abingham:
> Can you expose in why this seems easier?

I'm not sure I understand the question, but I think you're asking why a smaller pull request (i.e. this one) is better than a large one (i.e. the original one I copied from). The answer is that the original one seemed to be stuck because it was too big for this project to digest. It really was a monster, and nobody seemed to have the time/energy/patience to deal with it, including the original author. The perception - both mine and that of some others following this line of development - is that a smaller set of focused PRs will get the changes in faster.

Ninir:
Hi @abingham 

What I meant is that #1723 has 865 additions, this work has 858 ones and looking briefly, it seems to be the same kind of work: schema structure, documentation, etc.

It's just me trying to understand why this work is easier to review, as I may be overlooking something! 😄

abingham:
@Ninir They two PRs are probably almost identical. They both attempt to extract the "iot topic" elements from [a much larger PR that has foundered](https://github.com/hashicorp/terraform/pull/6961/). If you're considering how to decide between the two, at first approximation you should probably just chose the first. 

It might be useful to get both PRs in to local branches and do a `git diff` between them. Maybe that'll shed some light on the differences (or lack thereof) between them.

abingham:
@Ninir Actually, I take that back. The diff between the two is pretty monumental, largely I think because my PR is based on a more recent master version. That may be a good reason to review *this* PR rather than the other, but of course you could take the route of rebasing each PR on the same master to find the real differences.

abingham:
@Ninir Ok, I've done the "rebase and compare" step, and it looks like the changes are almost identical. The main difference is that the #1723 seems to have leading "+"s on the lines in `website/docs/r/iot_topic_rule.html.markdown`; my guess is that these are diff markers that somehow snuck in. The remaining differences seem to be inconsequential whitespace issues.

So, if my analysis is right, you should review this PR because it's a) a bit fresher and b) doesn't include the leading plus signs. In the end, though, either should be fine.

Ninir:
@abingham By convention, we will first have a look at #1723 and see how it goes. If however the author does not have time to update, explicitely says someone can take over it or there is a lack of answer, we will check & review this one.

This is just us trying to be fair with authors, respecting ordering & so. Hope it makes sense!

abingham:
Certainly, I'll try to do that tomorrow.

abingham:
@Ninir I got some time to start working on this today, but it'll take me some time to work through everything you listed in the review. I'll push changes up as I have them, and it might be useful if you can look them over as I do, both to make sure I'm doing what you've asked and also to make sure I'm doing things idiomatically (since I don't really know go).

Since your comments are on a separate PR, this might get a bit awkward. I could do something like add notes to the other PR when I've pushed up relevant changes. How does that sound to you?

abingham:
@Ninir I got some time to start working on this today, but it'll take me some time to work through everything you listed in the review. I'll push changes up as I have them, and it might be useful if you can look them over as I do, both to make sure I'm doing what you've asked and also to make sure I'm doing things idiomatically (since I don't really know go).

Since your comments are on a separate PR, this might get a bit awkward. I could do something like add notes to the other PR when I've pushed up relevant changes. How does that sound to you?

Ninir:
Hey @abingham 

I'm currently fixing last issues before merging, hope you don't mind.
This should be ready this day.

Still need to flatten actions when reading & complete the import.

Cheers!

bflad:

PR #1879, opened 2017-10-13
New Resource: aws_budget

xchapter7x:
building out budget resource functionality in order to provide the ability to create/configure budgets from the terraform aws provider

[https://github.com/terraform-providers/terraform-provider-aws/issues/753]


radeksimko:
Hi @xchapter7x ,
thanks for the contribution.

Do you mind submitting all vendor changes in a separate PR (we can merge such PR fairly quickly) to make the review of the real code easier?

xchapter7x:
sure thing @radeksimko . 

Here is the PR for the govendor of the budgets package https://github.com/terraform-providers/terraform-provider-aws/pull/1898

xchapter7x:
hi @radeksimko , 
this PR is ready for review. Whenever you have time to provide your thoughts and feedback, that would be much appreciated.

Side question, the docs say for new resources that terraform docs should be updated to reflect the new feature. Since the docs seem to be in a separate repo, I'm not sure what the desired flow is.

Let me know. Thanks.

-John

xchapter7x:
Hey @radeksimko, just wanted to check in and see if you have some free time to review this PR? Feedback would be greatly appreciated. Let me know thanks.

xchapter7x:
Hey @radeksimko ,

Any word on this PR? Any feedback you can provide would be much appreciated. 
Thanks :)

nicksantamaria:
I had a few thoughts while reviewing this PR and [writing some docs](https://github.com/xchapter7x/terraform-provider-aws/pull/1).

* According to the [AWS docs](http://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/data-type-budget.html#time-period), `time_period_end` defaults to June 15, 2087 when created in the console. It would be good to mirror that behavior, and make the `time_period_end` argument optional. 
* There are 3 `include_` arguments provided by this resource, but the [AWS CostTypes documentation](http://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/data-type-cost.html) shows there are 10 possible options. I would like to see support added for the following options:
  * `include_credit`
  * `include_discount`
  * `include_other_subscription`
  * `include_recurring`
  * `include_refund`
  * `include_credit`
  * `include_support`
  * `include_upfront`
* Add support for `name_prefix` as an alternative argument to `name`.

xchapter7x:
hi @nicksantamaria ,

thanks for the PR feedback. Let me set aside some time to get some commits together for the suggested enhancements. Stay tuned. Ill put this back to [wip] in the meantime.

xchapter7x:
hi @nicksantamaria ,

latest changes should reflect your suggested enhancements for `name_prefix`, `cost_types` & `time_period_end`.

any other suggestions or feedback would be most appreciated. Thanks.

- John

xchapter7x:
hi @radeksimko & @nicksantamaria , 

Happy New Year!

Just wanted to check in and ping for the preferred path forward for this PR.
Any further guidance, thoughts or feedback on the implementation and functionality would be greatly appreciated. 

Let me know. Thanks :)

John

eperdeme:
Been testing this locally looks to work really well. 
Is there anyway to also do notifications for budges like you can in the CLI/UI ? Ie, aws budgets create-budget --account-id X --budget X --notifications-with-subscribers X

https://docs.aws.amazon.com/cli/latest/reference/budgets/create-notification.html

xchapter7x:
@eperdeme I agree the ability to configure notifications would be a valuable addition.

It was intentionally left out of this PR to control the scope. Reason being, I feel adding notifications support is more appropriate to be done as a feature enhancement, after the core value add of the budget resource is merged in.

bflad:
Hey @xchapter7x! Thanks for this contribution! Sorry for the long maintainer silence on this PR as we have been trying to work through the PR backlog. I'll be able to check this out in the coming week or two. In the meantime, can you fix the conflicts or rebase with master please? That'll be a huge help here. 😄

xchapter7x:
@bflad 
I've just rebased this PR to master. Should be all good for review when you get a chance. Thanks in advance.

eperdeme:
Happy to see this progressing. For now for notifications I'll fudge it with some local-exec's.

eg;
resource "aws_budget" "cost" {
XXXXX
  provisioner "local-exec" {
    command = "aws budgets ..... ${aws_organization_account.mainaccount.id} etc"
  }
}

bflad:
I'll be taking a look at this later this week 👍

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+		},
+		"limit_amount": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"limit_unit": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"include_credit": {
```
bflad: While its more complex right now, it feels like these should all be nested under a `cost_types` attribute to match the API. We have run into issues with other AWS services where new API attributes are introduced that are similar or the same name as existing attributes. e.g.

```go
"cost_types": {
  Type:     schema.TypeList,
  Optional: true,
  MaxItems: 1,
  Elem:     &schema.Resource{
      "include_credit": {
        // ...
      },
      // ...
      "use_blended": {
        // ...
      },
  },
},
```

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+For more detailed documentation about each argument, refer to the [AWS official
+documentation](http://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/data-type-budget.html).
+
+The following arguments are supported:
```
bflad: Each of these arguments should denote `(Required)` or `(Optional)` - someday we'll have automated documentation 🙁

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+* `use_blended` - A boolean value whether to use blended costs in the cost budget. Defaults to `false`
+* `limit_amount` - The amount of cost or usage being measured for a budget.
+* `limit_unit` - The unit of measurement used for the budget forecast, actual spend, or budget threshold, such as dollars or GB. See [Spend ](http://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/data-type-spend.html) documentation.
+* `time_period_end` - The end of the time period covered by the budget. There are no restrictions on the end date. Format: `2017-01-01_12:00`.
+* `time_period_start` - The start of the time period covered by the budget. The start date must come before the end date. Format: `2017-01-01_12:00`.
+* `time_unit` - The length of time until a budget resets the actual and forecasted spend. Valid values: `MONTHLY`, `QUARTERLY`, `ANNUALLY`.
+
+## Attributes Reference
+
+The following attributes are exported:
```
bflad: `In addition to the arguments, the following additional attributes are exported:`

And then only include those additional attributes like `id` and `name` (always returned)

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+  * `AZ`
+  * `LinkedAccount`
+  * `Operation`
+  * `PurchaseType`
+  * `UsageType:<service name>`
+  * `TagKeyValue`
+
+Refer to [AWS CostFilter documentation](http://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/data-type-filter.html) for further detail.
+
+## Examples
```
bflad: These examples should be included in the examples section above

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+```hcl
+resource "aws_budget" "s3" {
+  ...
+  budget_type  = "usage"
+  limit_amount = "3"
+  limit_unit   = "GB"
+}
+```
+
+## Import
```
bflad: This should not be included unless the ID is updated as mentioned before, then should be updated to say:

```
Budgets can be imported using `AccountID:BudgetName`, e.g.

`$ terraform import aws_budgets_budget.myBudget 123456789012:myBudget`
```

aws/resource_aws_budget_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/service/budgets"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAwsBudget_basic(t *testing.T) {
```
bflad: All acceptance tests should be prefixed with `TestAccAWS...`

aws/resource_aws_budget_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+			{
+				Config: testBudgetHCLPrefix(configBasicUpdate),
+				Check:  newComposedBudgetTestCheck(configBasicUpdate, testAccProvider, "name_prefix"),
+			},
+		},
+	})
+}
+
+func newComposedBudgetTestCheck(config budgetTestConfig, provider *schema.Provider, nameField string) resource.TestCheckFunc {
```
bflad: We do not need this extra function redirection (simpler is better when it comes to the testing).

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+func budgetExists(budgetName string, meta interface{}) bool {
+	_, err := describeBudget(budgetName, meta)
+	if isBudgetNotFoundException(err) {
+		return false
+	}
+
+	return true
+}
+
+func isBudgetNotFoundException(err error) bool {
```
bflad: We do not need this extra function, any usage can be handled with:

```go
isAWSErr(err, budgetsErrCodeNotFoundException, "")
```

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+	deleteBudgetInput.SetAccountId(accountID)
+	_, err := client.DeleteBudget(deleteBudgetInput)
+	if err != nil {
+		return fmt.Errorf("delete budget failed: %v", err)
+	}
+
+	return nil
+}
+
+func newBudget(d *schema.ResourceData) (*budgets.Budget, error) {
```
bflad: Function names are global across the provider currently. We generally name functions like these `expandBudgetsBudget` to denote we're "expanding" the HCL configuration into a certain SDK data type.

Preferably, this would be split into its constituent parts as well (depending on resource schema nesting):
* `func expandBudgetsBudget()`
* `func expandBudgetsCostTypes(m map[string]interface{}) (*budgets.CostTypes, error)`

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	if _, ok := d.GetOk("name"); ok {
+		d.Set("name", describeBudgetOutput.Budget.BudgetName)
+	}
+
+	d.Set("budget_type", describeBudgetOutput.Budget.BudgetType)
+	d.Set("limit_amount", describeBudgetOutput.Budget.BudgetLimit.Amount)
+	d.Set("limit_unit", describeBudgetOutput.Budget.BudgetLimit.Unit)
+	d.Set("include_credit", describeBudgetOutput.Budget.CostTypes.IncludeCredit)
```
bflad: When this gets converted to a nested structure, we'll wind up setting this like:

```go
if err := d.Set("cost_types", flattenBudgetsCostTypes(output.Budget.CostTypes)); err != nil {
  return err
}
```

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+		"time_period_end": {
+			Type:     schema.TypeString,
+			Optional: true,
+			Default:  "2087-06-15_12:00",
+		},
+		"time_unit": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"cost_filters": &schema.Schema{
```
bflad: Minor nitpick: `&schema.Schema` is extraneous here

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+* `include_tax` - whether to include tax in the cost budget
+* `include_upfront` - whether to include upfront costs in the cost budget
+* `use_blended` - whether to use blended costs in the cost budget
+* `limit_amount` - The amount of cost or usage being measured
+* `limit_unit` - The unit of measurement used, such as dollars or GB.
+* `time_period_end` - The end of the time period covered by the budget
+* `time_period_start` - The start of the time period covered by the budget
+* `time_unit` - The length of time until the budget resets the actual and forecasted spend.
+
+### CostFilters
```
bflad: Arguments should be above Attributes

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return fmt.Errorf("describe budget failed: %v", err)
+	}
+
+	if _, ok := d.GetOk("name"); ok {
+		d.Set("name", describeBudgetOutput.Budget.BudgetName)
+	}
+
+	d.Set("budget_type", describeBudgetOutput.Budget.BudgetType)
+	d.Set("limit_amount", describeBudgetOutput.Budget.BudgetLimit.Amount)
```
bflad: These nested attributes could be very panic prone. They should be checked, e.g.

```go
budgetLimit := describeBudgetOutput.Budget.BudgetLimit
if budgetLimit != nil {
  d.Set("limit_amount", budgetLimit.Amount)
}
```

Although ideally, this is another case where we should introduce nested attributes and the state set becomes:

```go
if err := d.Set("limit", flattenBudgetsBudgetLimit(describeBudgetOutput.Budget.BudgetLimit)); err != nil {
  return err
}
```

aws/resource_aws_budget.go
```diff
@@ -0,0 +1,303 @@
...
+
+func isBudgetNotFoundException(err error) bool {
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == budgets.ErrCodeNotFoundException {
+		return true
+	}
+
+	return false
+}
+
+func describeBudget(budgetName string, meta interface{}) (*budgets.DescribeBudgetOutput, error) {
```
bflad: This function seems extraneous as well. Its not really saving any complexity and is just introducing additional redirection.

xchapter7x:
@bflad 

great feedback and suggestions, thanks!

I should have some time and be able to get something together in the next week or so. 

-john

xchapter7x:
hey @bflad,

We should be in a better spot now. Let me know your thoughts/feedback on the latest, when you get a chance. Thanks.

-john

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/budgets"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsBudgetsBudget() *schema.Resource {
+	return &schema.Resource{
+		Schema: resourceAwsBudgetsBudgetSchema(),
```
bflad: Is there a reason for the indirection here? (e.g. reusing the schema for some other resource) Generally we prefer to inline it unless its reused multiple times.

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+		accountID = meta.(*AWSClient).accountid
+	}
+
+	budget, err := expandBudgetsBudgetUnmarshal(d)
+	if err != nil {
+		return fmt.Errorf("failed creating budget: %v", err)
+	}
+
+	_, err = client.CreateBudget(&budgets.CreateBudgetInput{
+		AccountId: &accountID,
```
bflad: Minor nitpick: we generally use the aws-sdk-go `aws` package helper function for these: `AccountId: aws.String(accountID),`

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+		Budget:    budget,
+	})
+	if err != nil {
+		return fmt.Errorf("create budget failed: %v", err)
+	}
+
+	d.SetId(fmt.Sprintf("%s:%s", accountID, *budget.BudgetName))
+	return resourceAwsBudgetsBudgetUpdate(d, meta)
```
bflad: Do we need to call update here? We generally prefer calling read after create.

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func resourceAwsBudgetsBudgetUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*AWSClient).budgetconn
+	var accountID string
+	if v, ok := d.GetOk("account_id"); ok {
```
bflad: If we don't need to call update after create (its not typical), this logic gets replaced by the `decodeBudgetsBudgetId()` call 👍

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func resourceAwsBudgetsBudgetDelete(d *schema.ResourceData, meta interface{}) error {
+	accountID, budgetName, err := decodeBudgetsBudgetID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	client := meta.(*AWSClient).budgetconn
+	_, err = client.DescribeBudget(&budgets.DescribeBudgetInput{
```
bflad: We can remove the API call here by moving the `isAWSErr()` check under the `if err != nil` check under the `DeleteBudget` call:

```go
if err != nil {
  if isAWSErr(err, budgets.ErrCodeNotFoundException, "") {
    return nil
  }
  return fmt.Errorf("delete budget failed: %s", err)
}
```

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+		}
+
+		convertedCostFilters[k] = strings.Join(filterValues, ",")
+	}
+
+	return convertedCostFilters
+}
+
+func expandBudgetsBudgetUnmarshal(d *schema.ResourceData) (*budgets.Budget, error) {
+	var budgetName string
```
bflad: This budgetName logic should only live in the create function. It won't be necessary anywhere else when we call `d.Set("name", ...)` during read 👍

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+		return fmt.Errorf("describe budget failed: %v", err)
+	}
+
+	flattenedBudget, err := expandBudgetsBudgetFlatten(describeBudgetOutput.Budget)
+	if err != nil {
+		return fmt.Errorf("failed flattening budget output: %v", err)
+	}
+
+	if _, ok := d.GetOk("name"); ok {
+		d.Set("name", flattenedBudget.name)
```
bflad: We should always call `d.Set("name", flattenedBudget.name)` even for generated names/name prefixes. We just need to mark the `name` attribute as `Computed: true`

aws/resource_aws_budgets_budget.go
```diff
@@ -0,0 +1,431 @@
...
+		return err
+	}
+
+	client := meta.(*AWSClient).budgetconn
+	describeBudgetOutput, err := client.DescribeBudget(&budgets.DescribeBudgetInput{
+		BudgetName: &budgetName,
+		AccountId:  &accountID,
+	})
+	if isAWSErr(err, budgets.ErrCodeNotFoundException, "") {
+		d.SetId("")
```
bflad: Can you please include a warning message before here? `log.Printf("[WARN] Budget %s not found, removing from state", d.Id())`

aws/resource_aws_budgets_budget.go
xchapter7x: yeah, I'm beginning to realize that with the mono-package approach, minimizing the number of functions really reduces the overall package noise and potential for namespace conflict.

Do you feel there are benefits with one approach over the other (package per resource vs. mono-package), ie. are there reasons a single package approach was chosen originally?

aws/resource_aws_budgets_budget.go
xchapter7x: 👍

aws/resource_aws_budgets_budget.go
xchapter7x: i was trying to align to convention, seems i copied a non-standard resources flow :).. lol `s3_bucket`

makes more sense without the create->update

aws/resource_aws_budgets_budget.go
xchapter7x: 👍

aws/resource_aws_budgets_budget.go
xchapter7x: 👍

xchapter7x:
@bflad thanks for the feedback. I should be able to turn these around in the next few days.

-john

aws/resource_aws_budgets_budget.go
bflad: I can't speak for the original implementation (way before my time with the codebase and HashiCorp), but generally we lean towards mono-package nowadays since it tends to reduce friction for new contributors and less debates about code boundaries.

website/aws.erb
```diff
@@ -403,6 +403,15 @@
                     </ul>
                 </li>
 
+                <li<%= sidebar_current("docs-aws-resource-budget") %>>
+                    <a href="#">Budget Resources</a>
+                    <ul class="nav nav-visible">
+                        <li<%= sidebar_current("docs-aws-resource-budget") %>>
```
bflad: This should be updated to reflect the new nesting `sidebar_current("docs-aws-resource-budgets-budget")`

website/aws.erb
```diff
@@ -403,6 +403,15 @@
...
+                            <a href="/docs/providers/aws/r/budget.html">aws_budget</a>
```
bflad: The file path and resource name need to be updated for the nesting `<a href="/docs/providers/aws/r/budgets_budget.html">aws_budgets_budget</a>`

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,130 @@
+---
+layout: "aws"
+page_title: "AWS: aws_budgets_budget"
+sidebar_current: "docs-aws-resource-budget"
```
bflad: Similar update here: `"docs-aws-resource-budgets-budget"`

website/aws.erb
```diff
@@ -403,6 +403,15 @@
                     </ul>
                 </li>
 
+                <li<%= sidebar_current("docs-aws-resource-budget") %>>
```
bflad: Needs `s` appended so its `budgets` to match nesting

aws/resource_aws_budgets_budget_test.go
```diff
@@ -0,0 +1,408 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/service/budgets"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSBudget_basic(t *testing.T) {
```
bflad: These function names need to be updated as well, e.g. `TestAccAWSBudgetsBudget_basic`

aws/resource_aws_budgets_budget_test.go
```diff
@@ -0,0 +1,408 @@
...
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "time_period_end", configBasicUpdate.TimePeriodEnd),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "time_unit", configBasicUpdate.TimeUnit),
+					testBudgetExists(configBasicUpdate, testAccProvider),
+				),
+			},
+		},
+	})
+}
+
+func testBudgetExists(config budgetTestConfig, provider *schema.Provider) resource.TestCheckFunc {
```
bflad: Generally we separate out existence tests and attribute tests.

The signature for this should be `testAccCheckAWSBudgetsBudgetExists(resourceName string, budget *budget.Budget) resource.TestCheckFunc`

Where the provider is already available via an already initialized `testAccProvider` variable, so `testAccProvider.Meta().(*AWSClient).budgetconn` works automatically.

And the budget is made available for other attribute check functions via `*budget = *b.Budget` before returning.

aws/resource_aws_budgets_budget_test.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	if strconv.FormatBool(*costTypes.UseBlended) != config.UseBlended {
+		return fmt.Errorf("UseBlended not set properly '%v' should be '%v'", *costTypes.UseBlended, config.UseBlended)
+	}
+
+	return nil
+}
+
+func testCheckBudgetDestroy(budgetName string, provider *schema.Provider) error {
```
bflad: This should be checking all possible resources in the acceptance testing Terraform state:

```go
func testAccCheckAWSBudgetsBudgetDestroy(s *terraform.State) error {
  conn := testAccProvider.Meta().(*AWSClient).budgetconn
  for _, rs := range s.RootModule().Resources {
    if rs.Type != "aws_budgets_budget" {
      continue
    }
    // rest of logic using rs.Primary.ID instead of budgetName
  }
  
  return nil
}
```

aws/resource_aws_budgets_budget_test.go
```diff
@@ -0,0 +1,408 @@
...
+		AccountId:  &accountID,
+	})
+	if !isAWSErr(err, budgets.ErrCodeNotFoundException, "") {
+		return fmt.Errorf("Budget '%s' was not deleted properly", budgetName)
+	}
+
+	return nil
+}
+
+type budgetTestConfig struct {
```
bflad: If we could remove this custom configuration indirection here, that would be wonderful. Being more verbose with the configurations and the attribute testing (looking inside the `*budget.Budget` from the API directly or even creating a `&budget.Budget{}` for reflect.DeepEqual) makes the testing easier to understand.

aws/resource_aws_budgets_budget_test.go
```diff
@@ -0,0 +1,408 @@
...
+		{{.FilterKey}} = "{{.FilterValue}}"
+	}
+}
+`))
+	var doc bytes.Buffer
+	t.Execute(&doc, budgetConfig)
+	return doc.String()
+}
+
+func testBudgetHCLPrefixUseDefaults(budgetConfig budgetTestConfig) string {
```
bflad: General naming convention: `testAccAWSBudgetsBudgetConfig_XXX`

aws/resource_aws_budgets_budget_test.go
```diff
@@ -0,0 +1,408 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckNoResourceAttr("aws_budgets_budget.foo", "account_id"),
+					resource.TestMatchResourceAttr("aws_budgets_budget.foo", "name", regexp.MustCompile(configBasicDefaults.BudgetName)),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "budget_type", configBasicDefaults.BudgetType),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "limit_amount", configBasicDefaults.LimitAmount),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "limit_unit", configBasicDefaults.LimitUnit),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "time_period_start", configBasicDefaults.TimePeriodStart),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "time_period_end", configBasicDefaults.TimePeriodEnd),
+					resource.TestCheckResourceAttr("aws_budgets_budget.foo", "time_unit", configBasicDefaults.TimeUnit),
+					testBudgetExists(configBasicDefaults, testAccProvider),
```
bflad: The existence testing should be first so the acceptance testing error is more clear if it cannot create the resource (and also so we can later reference the `&budget`). 😄

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,130 @@
+---
```
bflad: This file needs to be renamed to match the nesting: `website/docs/r/budgets_budget.html.markdown`

website/docs/r/budget.html.markdown
```diff
@@ -0,0 +1,130 @@
...
+```hcl
+resource "aws_budgets_budget" "s3" {
+  ...
+  budget_type  = "usage"
+  limit_amount = "3"
+  limit_unit   = "GB"
+}
+```
+
+## Attributes Reference
```
bflad: We generally list attributes after arguments and should only list attributes that aren't already arguments like `id`. 👍

aws/resource_aws_budgets_budget_test.go
xchapter7x: agreed, this struct always kind of bothered me. intent was to flatten the config so it was more manageable, but to your point it likely just ends up adding complexity.

aws/resource_aws_budgets_budget_test.go
xchapter7x: my thought in putting the provider as an argument in these functions was to avoid dependencies on those global vars. I guess since they are there to be used and are standard, it makes sense to conform.

with regard to your last comment of setting `*budget = *b.Budget` im not sure i follow what you're suggesting. Could you expand that thought a bit?

xchapter7x:
@bflad , 

just pushed up changes in line with the last refactor suggestion. 
If you could give it a look when you have a moment, and let me know if there are any further refactors, suggestions or thoughts.

thanks for your continued feedback and support in driving this to completion.

-John

website/docs/r/budgets_budget.html.markdown
```diff
@@ -0,0 +1,120 @@
...
+# aws_budgets_budget
+
+Provides a budgets budget resource. Budgets use the cost visualisation provided by Cost Explorer to show you the status of your budgets, to provide forecasts of your estimated costs, and to track your AWS usage, including your free tier usage.
+
+## Example Usage
+
+```hcl
+resource "aws_budgets_budget" "ec2" {
+  name                  = "budget-ec2-monthly"
+  budget_type           = "spend"
```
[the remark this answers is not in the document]
xchapter7x: 👍 
thanks for pointing these out @fmasuhr . ill try to push up a change this evening.

bflad:

PR #1882, opened 2017-10-13
Check VPC array size

andreaTP:
This missing check causes a terraform crash when used with Eucalyptus Cloud.
I know that supporting AWS clones is not a target for this project but the fix here just increase robustness of the implementation IMO.


radeksimko:
Thank you for the contribution.
To speed things up I just added a one-line comment in ea5fb39 to explain why we perform this check (for context) as the official AWS API doesn't behave this way. I hope you don't mind.

Your contribution is still included (and still credited to you), with the appropriate modifications.

andreaTP:
sounds perfect to me @radeksimko thanks a lot, sorry first contribution to this codebase ...

andreaTP:
if you prefer I can squash commits to have a cleaner tree

andreaTP:
ok wow, thanks! :-)

Issue #1486, opened 2017-08-23
Support AWS Athena


countergram:
AWS Athena supports the creation of tables and named queries both which are persistent, configurable objects. It would be valuable to support them in Terraform so that a set of known tables and queries can be configured in an AWS environment without being manually added.

https://aws.amazon.com/documentation/athena/

atsushi-ishibashi:
It might be a good idea to wait until AWS provides api to create database for Athena.

betabandido:
@atsushi-ishibashi It is actually already possible to create a database by running a `CREATE DATABASE` query. Therefore, I wonder if Amazon is going to add soon any direct support in the SDK for creating a database, given there is already a way to do it.

atsushi-ishibashi:
Done #1893

radeksimko:
As mentioned we now support the following two resources:

 - [`athena_database`](https://www.terraform.io/docs/providers/aws/r/athena_database.html)
 - [`athena_named_query`](https://www.terraform.io/docs/providers/aws/r/athena_named_query.html)

I have explained some reasoning behind my inclination to _not_ supporting `athena_table` in https://github.com/terraform-providers/terraform-provider-aws/pull/1893#issuecomment-351300973

I'm going to close this issue now. If someone feels that `athena_table` is totally must have I'd recommend opening a new issue focused on that resource only and attach some use cases (which presumably aren't "data backup/recovery"). We can also monitor upvotes (interest) that way and understand how relevant/important it is for users.

Thanks all and especially @atsushi-ishibashi for getting it done! 🎉

PR #1893, opened 2017-10-14
New Resource: Athena Named Query

atsushi-ishibashi:
#1486 

Adds the feature to create athena's named query. Aws cli doesn't provide api to create database but named query api requires database name.
So you have to create database in advance.
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSAthena'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAthena -timeout 120m
=== RUN   TestAccAWSAthenaNamedQuery
--- PASS: TestAccAWSAthenaNamedQuery (39.28s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	39.318s
```

### DIFF
--- a/aws/resource_aws_athena_named_query.go
+++ b/aws/resource_aws_athena_named_query.go
@@ -1,8 +1,6 @@
 package aws
 
 import (
-	"fmt"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/athena"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -68,7 +66,7 @@ func resourceAwsAthenaNamedQueryRead(d *schema.ResourceData, meta interface{}) e
 
 	_, err := conn.GetNamedQuery(input)
 	if err != nil {
-		if isAWSErr(err, athena.ErrCodeInvalidRequestException, fmt.Sprintf("Athena Named Query (%s) does not exist", d.Id())) {
+		if isAWSErr(err, athena.ErrCodeInvalidRequestException, d.Id()) {
 			d.SetId("")
 			return nil
 		}
--- a/aws/resource_aws_athena_named_query.go
+++ b/aws/resource_aws_athena_named_query.go
@@ -1,0 +1,43 @@
+package aws
+
+func resourceAwsAthenaNamedQuery() *schema.Resource {
+ 	return &schema.Resource{
+ 		Create: resourceAwsAthenaNamedQueryCreate,
+ 		Read:   resourceAwsAthenaNamedQueryRead,
+ 		Delete: resourceAwsAthenaNamedQueryDelete,
+
+ 		Schema: map[string]*schema.Schema{
+      "name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+      "query": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+      "database": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+      "description": &schema.Schema{
+				Type:     schema.TypeString,
+        Optional:     true,
+			},
+    }
+  }
+}
+
+func resourceAwsAthenaNamedQueryCreate(d *schema.ResourceData, meta interface{}) error {
+  return nil
+}
+
+func resourceAwsAthenaNamedQueryRead(d *schema.ResourceData, meta interface{}) error {
+  return nil
+}
+
+func resourceAwsAthenaNamedQueryDelete(d *schema.ResourceData, meta interface{}) error {
+  return nil
+}
--- a/aws/resource_aws_athena_named_query.go
+++ b/aws/resource_aws_athena_named_query.go
@@ -82,5 +82,6 @@ func resourceAwsAthenaNamedQueryDelete(d *schema.ResourceData, meta interface{})
 	if err != nil {
 		return err
 	}
+	d.SetId("")
 	return nil
 }
--- a/aws/resource_aws_athena_named_query.go
+++ b/aws/resource_aws_athena_named_query.go
@@ -1,43 +1,86 @@
 package aws
 
-func resourceAwsAthenaNamedQuery() *schema.Resource {
- 	return &schema.Resource{
- 		Create: resourceAwsAthenaNamedQueryCreate,
- 		Read:   resourceAwsAthenaNamedQueryRead,
- 		Delete: resourceAwsAthenaNamedQueryDelete,
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/athena"
+	"github.com/hashicorp/terraform/helper/schema"
+)
 
- 		Schema: map[string]*schema.Schema{
-      "name": &schema.Schema{
+func resourceAwsAthenaNamedQuery() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsAthenaNamedQueryCreate,
+		Read:   resourceAwsAthenaNamedQueryRead,
+		Delete: resourceAwsAthenaNamedQueryDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-      "query": &schema.Schema{
+			"query": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-      "database": &schema.Schema{
+			"database": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-      "description": &schema.Schema{
+			"description": &schema.Schema{
 				Type:     schema.TypeString,
-        Optional:     true,
+				Optional: true,
+				ForceNew: true,
 			},
-    }
-  }
+		},
+	}
 }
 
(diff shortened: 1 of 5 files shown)


radeksimko:
Hi @atsushi-ishibashi 
thanks for the PR. We'd love to have the support for Athena in Terraform, but I'd rather not have tests relying on a static database name created manually.

I reached to the AWS support to ask about management of DBs and they sent me this helpful answer:

> So while there is not a specific call to accomplish this, there is a perfectly viable method to accomplish this with the existing API. Something as simple as:
> ```
> aws athena start-query-execution --query-string "create database <database name>;" --result-configuration "OutputLocation=s3://<output location>/"
> ```
> ...will do the job.

How do you feel about submitting a _separate_ PR for `aws_athena_database` which would basically map the CRUD to such simple Athena queries?

It has some gotchas admittedly...

I assume we'll need to figure out how to deal with the S3 relationship. I'd probably pick a unique random S3 bucket in the form of `tf-athena-db-<db-name>-<random-id>` which would be created as part of the execution and we should also make sure it's deleted right after creation and/or error.

This behaviour should also be clearly stated in the docs, so that the user is aware.

Let me know what you think.

atsushi-ishibashi:
@radeksimko Thank you for your review and research. That's a good idea to create database.
Ok, I'll submit PR for aws_athena_database in advance👍

radeksimko:
Do you mind rebasing this PR and leveraging `aws_athena_database` here? Thanks 😉

atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSAthenaNamedQuery_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAthenaNamedQuery_basic -timeout 120m
=== RUN   TestAccAWSAthenaNamedQuery_basic
--- PASS: TestAccAWSAthenaNamedQuery_basic (90.47s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	90.519s
```

atsushi-ishibashi:
@radeksimko You're welcome to review👍

aws/resource_aws_athena_named_query_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+		input := &athena.GetNamedQueryInput{
+			NamedQueryId: aws.String(rs.Primary.ID),
+		}
+
+		resp, err := conn.GetNamedQuery(input)
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case athena.ErrCodeInvalidRequestException:
```
radeksimko: The error code here seems quite generic - is this supposed to be an error returned in case the query is gone? Can we make it a bit more specific by any chance? e.g. `isAWSErr(err, athena.ErrCodeInvalidRequestException, "something something")`

aws/resource_aws_athena_named_query.go
```diff
@@ -0,0 +1,87 @@
...
+func resourceAwsAthenaNamedQueryRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).athenaconn
+
+	input := &athena.GetNamedQueryInput{
+		NamedQueryId: aws.String(d.Id()),
+	}
+
+	_, err := conn.GetNamedQuery(input)
+	if err != nil {
+		return err
```
radeksimko: Do you mind adding a check here for cases when the query is gone and remove it from the state via `d.SetId("")`, please? I don't know if it returns an error or just `nil` in such case (that may need some research), but we should have a check in place.

aws/resource_aws_athena_named_query.go
```diff
@@ -0,0 +1,93 @@
...
+func resourceAwsAthenaNamedQueryRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).athenaconn
+
+	input := &athena.GetNamedQueryInput{
+		NamedQueryId: aws.String(d.Id()),
+	}
+
+	_, err := conn.GetNamedQuery(input)
+	if err != nil {
+		if isAWSErr(err, athena.ErrCodeInvalidRequestException, fmt.Sprintf("Athena Named Query (%s) does not exist", d.Id())) {
```
radeksimko: I'm afraid this won't match with the (admittedly odd) error message:

```
2017/11/10 10:12:32 [DEBUG] [aws-sdk-go] DEBUG: Request athena/GetNamedQuery Details:
---[ REQUEST POST-SIGN ]-----------------------------
POST / HTTP/1.1
Host: athena.us-west-2.amazonaws.com
User-Agent: aws-sdk-go/1.12.25 (go1.9.2; darwin; amd64) APN/1.0 HashiCorp/1.0 Terraform/0.11.0-beta1
Content-Length: 55
Authorization: AWS4-HMAC-SHA256 Credential=*REDACTED*/20171110/us-west-2/athena/aws4_request, SignedHeaders=content-length;content-type;host;x-amz-date;x-amz-target, Signature=*REDACTED*
Content-Type: application/x-amz-json-1.1
X-Amz-Date: 20171110T101232Z
X-Amz-Target: AmazonAthena.GetNamedQuery
Accept-Encoding: gzip

{"NamedQueryId":"fb403822-2e67-4344-b530-fc488aa79c49"}
-----------------------------------------------------
2017/11/10 10:12:32 [DEBUG] [aws-sdk-go] DEBUG: Response athena/GetNamedQuery Details:
---[ RESPONSE ]--------------------------------------
HTTP/1.1 400 Bad Request
Connection: close
Content-Length: 181
Content-Type: application/x-amz-json-1.1
Date: Fri, 10 Nov 2017 10:12:32 GMT
X-Amzn-Requestid: aa8710fc-c5ff-11e7-97cf-5fde2fdbf5a4


-----------------------------------------------------
2017/11/10 10:12:32 [DEBUG] [aws-sdk-go] {"__type":"InvalidRequestException","AthenaErrorCode":"NamedQuery: %s does not exists","ErrorCode":"NamedQuery: %s does not exists","Message":"fb403822-2e67-4344-b530-fc488aa79c49"}
```

do you mind loosening up the matching? e.g. just to `"does not exist"`

aws/resource_aws_athena_named_query_test.go
```diff
@@ -0,0 +1,85 @@
...
+			continue
+		}
+
+		input := &athena.GetNamedQueryInput{
+			NamedQueryId: aws.String(rs.Primary.ID),
+		}
+
+		resp, err := conn.GetNamedQuery(input)
+		if err != nil {
+			if isAWSErr(err, athena.ErrCodeInvalidRequestException, "does not exist") {
```
radeksimko: I'm not sure what's going on here, but the test is failing for me:

```
=== RUN   TestAccAWSAthenaNamedQuery_basic
--- FAIL: TestAccAWSAthenaNamedQuery_basic (104.99s)
	testing.go:563: Error destroying resource! WARNING: Dangling resources

		Error: Check failed: InvalidRequestException: fb403822-2e67-4344-b530-fc488aa79c49
			status code: 400, request id: aa8710fc-c5ff-11e7-97cf-5fde2fdbf5a4

		State: <no state>
```
Is it passing for you? 🤔

aws/resource_aws_athena_named_query.go
atsushi-ishibashi: I misunderstood the behaviour of `isAWSErr`. I'll modify. And the same reason why the test didn't pass.

atsushi-ishibashi:
@radeksimko I confirmed that what `err.Message()` is just named query id which we use for input.
Sorry for confusing you because of skipping `testacc` after the previous commit 🙇 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSAthenaNamedQuery_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAthenaNamedQuery_basic -timeout 120m
=== RUN   TestAccAWSAthenaNamedQuery_basic
--- PASS: TestAccAWSAthenaNamedQuery_basic (93.92s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	93.977s
```

dewang:
Hi,

Great to see support for database and named queries. Is there any chance we might also see support for creating tables?

Thanks!

atsushi-ishibashi:
@dewang We already have supported athena database👍
https://www.terraform.io/docs/providers/aws/r/athena_database.html

radeksimko:
We support database as @atsushi-ishibashi mentioned. I'm not sure it's a good idea to support _tables_ though. We've been traditionally hesitant to adding resources which manage data in one way or another and tables are in that grey area.

There's a lot of use cases where people are much better just using some standard Athena/SQL-based tools (whatever these are) which scale better with growing data/structures. Terraform is not designed to handle long-running data backup/recovery/migration which may stop half-way through for any reason. It offers no effective ways to recover from such state and no way to report progress.

atsushi-ishibashi:
@dewang Sorry, I missed you talked about table(^^;;

betabandido:
I had a look into this a while ago, and I also believe it is not a good idea to support tables. At least for now. One reason is that obtaining information on the existing tables would require parsing the output of a command such as `DESCRIBE TABLE`. The result of that command is a formatted text that I suppose Amazon may change without giving any notice. Therefore, attempting to parse that text would be very fragile in my opinion.

dewang:
@radeksimko Athena tables are _external_ tables, so only describing data schema, not really managing data. They would never need to handle backup/recovery in any way.

@betabandido fair shout, if the `describe table` response is bare text that needs to be parsed it would indeed be very fragile. It's just that we manage all of the rest of our setup using terraform, so would have been cool to add this to the same deployment process as well :-)

Guess we'll have to go with bash scripts to call aws cli with the right params for each environment.

Issue #1259, opened 2017-07-27
FR: Support AWS SSM Resource Data Sync


sfrode:
I would kindly request an [AWS SSM Resource Data sync](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_CreateResourceDataSync.html) resource to support the pipeline as described [here](https://aws.amazon.com/blogs/aws/new-s3-sync-capability-for-ec2-systems-manager-query-visualize-instance-software-inventory/).

atsushi-ishibashi:
Close

PR #1895, opened 2017-10-14
New Resource: aws_ssm_resource_data_sync

atsushi-ishibashi:
#1259
In my local environment, I got error like below. But all checks have passed in travis-ci...
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSSsmResourceDataSync'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSsmResourceDataSync -timeout 120m
=== RUN   TestAccAWSSsmResourceDataSync
--- FAIL: TestAccAWSSsmResourceDataSync (44.23s)
	testing.go:434: Step 0 error: Check failed: Check 1/1 error: Not found: aws_ssm_resource_data_sync.foo
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	44.270s
make: *** [testacc] Error 1
```

### DIFF
--- a/aws/resource_aws_ssm_resource_data_sync.go
+++ b/aws/resource_aws_ssm_resource_data_sync.go
@@ -1,8 +1,11 @@
 package aws
 
 import (
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ssm"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -56,15 +59,25 @@ func resourceAwsSsmResourceDataSync() *schema.Resource {
 func resourceAwsSsmResourceDataSyncCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ssmconn
 
-	input := &ssm.CreateResourceDataSyncInput{
-		S3Destination: expandSsmResourceDataSyncS3Destination(d),
-		SyncName:      aws.String(d.Get("name").(string)),
-	}
+	err := resource.Retry(1*time.Minute, func() *resource.RetryError {
+		input := &ssm.CreateResourceDataSyncInput{
+			S3Destination: expandSsmResourceDataSyncS3Destination(d),
+			SyncName:      aws.String(d.Get("name").(string)),
+		}
+		_, err := conn.CreateResourceDataSync(input)
+		if err != nil {
+			if isAWSErr(err, ssm.ErrCodeResourceDataSyncInvalidConfigurationException, "S3 write failed for bucket") {
+				return resource.RetryableError(err)
+			}
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
 
-	_, err := conn.CreateResourceDataSync(input)
 	if err != nil {
 		return err
 	}
+
 	d.SetId(d.Get("name").(string))
 	return resourceAwsSsmResourceDataSyncRead(d, meta)
 }
--- a/aws/resource_aws_ssm_resource_data_sync.go
+++ b/aws/resource_aws_ssm_resource_data_sync.go
@@ -1,15 +1,11 @@
 package aws
 
 import (
-	"bytes"
 	"fmt"
-	"log"
-	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ssm"
-	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -25,18 +21,18 @@ func resourceAwsSsmResourceDataSync() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-			"destination": {
-				Type:     schema.TypeSet,
+			"s3_destination": {
+				Type:     schema.TypeList,
 				Required: true,
 				ForceNew: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"key": {
+						"kms_key_arn": {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
-						"bucket": {
+						"bucket_name": {
 							Type:     schema.TypeString,
 							Required: true,
 						},
@@ -48,20 +44,12 @@ func resourceAwsSsmResourceDataSync() *schema.Resource {
 							Type:     schema.TypeString,
 							Required: true,
 						},
+						"sync_format": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  ssm.ResourceDataSyncS3FormatJsonSerDe,
+						},
 					},
-				},
-				Set: func(v interface{}) int {
-					var buf bytes.Buffer
-					m := v.(map[string]interface{})
-					buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(m["bucket"].(string))))
-					buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(m["region"].(string))))
-					if val, ok := m["key"]; ok {
-						buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(val.(string))))
-					}
-					if val, ok := m["prefix"]; ok {
-						buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(val.(string))))
-					}
-					return hashcode.String(buf.String())
 				},
 			},
 		},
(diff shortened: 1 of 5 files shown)


atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSSsmResourceDataSync'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSsmResourceDataSync -timeout 120m
=== RUN   TestAccAWSSsmResourceDataSync
--- PASS: TestAccAWSSsmResourceDataSync (60.41s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	60.457s
```

aws/resource_aws_ssm_resource_data_sync.go
```diff
@@ -0,0 +1,180 @@
...
+		for _, v := range resp.ResourceDataSyncItems {
+			if *v.SyncName == d.Get("name").(string) {
+				found = true
+			}
+		}
+		if found || *resp.NextToken == "" {
+			break
+		}
+		nextToken = *resp.NextToken
+	}
```
radeksimko: Do you mind moving this for loop into a separate function, e.g. `findResourceDataSync(conn *ssm.SSM, name string) *ssm.ResourceDataSyncItem`?

aws/resource_aws_ssm_resource_data_sync.go
```diff
@@ -0,0 +1,180 @@
...
+	if !found {
+		log.Printf("[INFO] No Resource Data Sync found for SyncName: %s", d.Get("name").(string))
+	}
```
radeksimko: There's a few things missing here - we need to re-set all the fields here based on what came from the API, e.g.

```go
d.Set("s3_destination", flattenS3Destination(ds.S3Destination))
```
where the flattener will turn the SDK object into `[]interface{}`

and secondly if there's no sync found we should remove it from state by calling `d.SetId("")`.

aws/resource_aws_ssm_resource_data_sync.go
```diff
@@ -0,0 +1,180 @@
...
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case ssm.ErrCodeResourceDataSyncNotFoundException:
+				return nil
+			default:
+				return err
+			}
+		}
+		return err
+	}
```
radeksimko: All of the above can be IMO simplified to something like this:

```go
if isAWSErr(err, ssm.ErrCodeResourceDataSyncNotFoundException, "") {
    return nil
}
return err
```

aws/resource_aws_ssm_resource_data_sync_test.go
```diff
@@ -0,0 +1,130 @@
...
+					found = true
+				}
+			}
+			if resp.NextToken != nil {
+				nextToken = *resp.NextToken
+			}
+			if found || nextToken == "" {
+				break
+			}
+		}
```
radeksimko: Once you have decoupled the logic for finding the right data sync, it can also be reused here 😉

aws/resource_aws_ssm_resource_data_sync_test.go
```diff
@@ -0,0 +1,130 @@
...
+		if found {
+			return fmt.Errorf("[DELETE ERROR] Resource Data Sync found for SyncName: %s", rs.Primary.Attributes["name"])
```
radeksimko: Nitpick, but there's no reason to prefix the message with `[DELETE ERROR]` - it's already clear that it's error.

aws/resource_aws_ssm_resource_data_sync_test.go
```diff
@@ -0,0 +1,130 @@
...
+                    }
+                }
+            }
+          ]
+      }
+      EOF
+    }
+
+    resource "aws_ssm_resource_data_sync" "foo" {
+      name = "foo"
```
radeksimko: Do you mind randomizing the name here?

aws/resource_aws_ssm_resource_data_sync.go
```diff
@@ -0,0 +1,180 @@
...
+					m := v.(map[string]interface{})
+					buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(m["bucket"].(string))))
+					buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(m["region"].(string))))
+					if val, ok := m["key"]; ok {
+						buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(val.(string))))
+					}
+					if val, ok := m["prefix"]; ok {
+						buf.WriteString(fmt.Sprintf("%s-", strings.ToLower(val.(string))))
+					}
+					return hashcode.String(buf.String())
```
radeksimko: Can you please explain why do we need to lowercase all the values here? It seems unnecessary to me. 🤔

aws/resource_aws_ssm_resource_data_sync.go
```diff
@@ -0,0 +1,180 @@
...
+		Delete: resourceAwsSsmResourceDataSyncDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"destination": {
+				Type:     schema.TypeSet,
```
radeksimko: Since this is a singleton I think we can just make it `TypeList` as there's no value in computing the hash and it would also complicate referencing. TypeList can be referenced as `s3_destination.0.key` etc., whereas TypeSet is non-trivial to reference because the address is `s3_destination.<computed-idx>.key`.

atsushi-ishibashi:
I absolutely agree with your feedback. I'm sorry for bothering you with trivial mistakes because I couldn't understand the behaivour when I submitted this PR 🙇 
I added `sync_format` with default value.
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSSsmResourceDataSync_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSsmResourceDataSync_basic -timeout 120m
=== RUN   TestAccAWSSsmResourceDataSync_basic
--- PASS: TestAccAWSSsmResourceDataSync_basic (63.07s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	63.119s
```

aws/resource_aws_ssm_resource_data_sync.go
```diff
@@ -0,0 +1,160 @@
...
+	conn := meta.(*AWSClient).ssmconn
+
+	syncItem, err := findResourceDataSyncItem(conn, d.Get("name").(string))
+	if err != nil {
+		if _, ok := err.(awserr.Error); ok {
+			return err
+		} else {
+			d.SetId("")
+			return nil
+		}
```
radeksimko: I'm slightly worried about this conditional here, because we may add more errors to `findResourceDataSyncItem` which aren't `awserr` and those may have nothing to do with "resource not found".
How do you feel about tweaking `findResourceDataSyncItem` to return `nil, nil` if no resource was found and then checking the `nil` here, instead of checking the error _isn't_ awserr.Error?

aws/resource_aws_ssm_resource_data_sync_test.go
```diff
@@ -0,0 +1,112 @@
...
+		if rs.Type != "aws_ssm_resource_data_sync" {
+			continue
+		}
+		_, err := findResourceDataSyncItem(conn, rs.Primary.Attributes["name"])
+		if err != nil {
+			if _, ok := err.(awserr.Error); ok {
+				return err
+			} else {
+				return nil
+			}
```
radeksimko: ^ I share the same concerns here as above

aws/resource_aws_ssm_resource_data_sync_test.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		return nil
+	}
+}
+
+func testAccSsmResourceDataSyncConfig(rInt int, rName string) string {
+	return fmt.Sprintf(`
+    resource "aws_s3_bucket" "hoge" {
+      bucket = "tf-test-bucket-%d"
+      region = "us-east-1"
```
radeksimko: Any particular reason why this bucket should be in `us-east-1`? It's certainly no big deal, just that our primary test region is `us-west-2`.

aws/resource_aws_ssm_resource_data_sync.go
atsushi-ishibashi: Sure👍 I'll apply that idea!

radeksimko:
@atsushi-ishibashi I just added retry logic to work around the eventually consistent nature of S3 which emerged in the acceptance test:

```
=== RUN   TestAccAWSSsmResourceDataSync_basic
--- FAIL: TestAccAWSSsmResourceDataSync_basic (61.77s)

		* aws_ssm_resource_data_sync.foo: 1 error(s) occurred:

		* aws_ssm_resource_data_sync.foo: ResourceDataSyncInvalidConfigurationException: S3 write failed for bucket [tf-test-bucket-5869476133795886706] with prefix [null] with sync name [tf-test-ssm-2hzj4] due to [Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: 31B58B3D98C61C90; S3 Extended Request ID: ra9YFypLVbeh6yzILpZ8S39ukPnmkCw6UBFWwSKdYeTLU7uKBgI5a0Mwut4Ds/7J)]. Please refer: http://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-inventory-configuring.html#sysman-inventory-datasync
			status code: 400, request id: ccedb0c2-c5f8-11e7-9213-61c5dcb61d04
```

Let me know if you have any questions in regards to that patch, otherwise I'm happy to merge this.

atsushi-ishibashi:
@radeksimko I missed that point... Of course, I appreciate the patch👍

Issue #434, opened 2017-06-13
Support for AWS Workspaces


hashibot:
_This issue was originally opened by @mootpt as hashicorp/terraform#10794. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Support for AWS Workspaces would be great, especially since there is existing support for AWS directory service directories in Terraform. Something like:

```
resource "aws_workspaces_workspace" {
  bundle_id              = "Some Bundle Id Here"
  directory_id           = "${aws_directory_service_directory.bar.id}"
  user_name              = "foo"
  root_volume_encryption = true,
  user_volume_encryption = false,
  volume_encryption_key  = "${aws_kms_key.foo.key_id}"
}
```

Ideally it would be nice to also have a `aws_workspaces_bundle` resource, but I'd settle for the former.

zachgatesak:
Checking in on this ticket any updates or has it made it onto the roadmap?

apparentlymart:
Based on the outcome of hashicorp/terraform#7806 it seems like this is not solvable at this time, though it may have been resolved in the mean time. Nobody on the team at HashiCorp is currently working on this, but we'd be happy to review a new PR for this (possibly based on the existing one) if the necessary API actions are now in place.

foxx:
Any plans to add support for this soon?

pjohnstonv:
This would be nice.

apobbati:
Any updates?

kmishra9:
Will also indicate my interest in this feature here (+1)

eidast:
+1

ryancbutler:
AWS still hasn't provided any API functionality to register a directory for workspaces making this impossible.  Been following this https://forums.aws.amazon.com/thread.jspa?threadID=237801 which AWS has stated there is no ETA.

bharat109puri:
@apparentlymart Do we have support for this feature or any ETA ?

porthunt:
@bharat109puri AWS has no ETA:

> At this moment, we don't have an ETA on when this feature will be released. However, as soon as it gets released, it should be publicly announced in either the 'AWS News Blog' page [1] or on our 'What's New with AWS' page [2]. I would also recommend you to subscribe to our 'AWS Announcements' page [3].

https://forums.aws.amazon.com/thread.jspa?threadID=237801

Last message from January 15th.

[...]

PR #1917, opened 2017-10-16
[Pending] New Resource: aws_workspaces_workspace

atsushi-ishibashi:
#434 
I have to crawl a website which only suits IE so I'd like to add workspaces resource.
AWS API provides `CreateWorkspaces` which can create some workspaces at a time, but it's complicated to write multi workspaces in one resource like below.
```
resource "aws_workspaces_workspaces" "hoge" {
  workspace = {
    a = a1
    b = b1
    ....
  }

  workspace = {
    a = a2
    b = b2
    ....
  }
}
```
So I'm going to add `aws_workspaces_workspace` and for each `CreateWorkspaces` works with one element.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -80,6 +80,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/waf"
 	"github.com/aws/aws-sdk-go/service/wafregional"
+	"github.com/aws/aws-sdk-go/service/workspaces"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
@@ -206,6 +207,7 @@ type AWSClient struct {
 	athenaconn            *athena.Athena
 	dxconn                *directconnect.DirectConnect
 	mediastoreconn        *mediastore.MediaStore
+	wsconn                *workspaces.WorkSpaces
 }
 
 func (c *AWSClient) S3() *s3.S3 {
@@ -451,6 +453,7 @@ func (c *Config) Client() (interface{}, error) {
 	client.athenaconn = athena.New(sess)
 	client.dxconn = directconnect.New(sess)
 	client.mediastoreconn = mediastore.New(sess)
+	client.wsconn = workspaces.New(sess)
 
 	// Workaround for https://github.com/aws/aws-sdk-go/issues/1376
 	client.kinesisconn.Handlers.Retry.PushBack(func(r *request.Request) {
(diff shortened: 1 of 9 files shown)


atsushi-ishibashi:
I got the below error and I couldn't api to register the directory to workspaces.
```
The specified directory could not be found in the specified region.
```
I asked aws support but they said that they could't support api to register so you had to do it from web console.

So this PR is pending until aws supports it.

liamjbennett:
Reached the same issue here: https://github.com/hashicorp/terraform/pull/7806

bcornils:
looks like we have a check failing and the branch has conflicts. Has anyone had time to assess or review criticality of the failure?

sstarcher:
@atsushi-ishibashi are you going to be able to pick this back up or do you need someone to take it over.

bflad:
Closing this PR due to lack of response. I should note that the `aws_workspaces_bundle` data source is separately being merged in #3243 and that will be released in version 1.40.0 of the AWS provider, likely middle of this week. 👍

cmaxwellau:
@atsushi-ishibashi @liamjbennett

I work for AWS, and have a customer that would like to use terraform for workspaces. 

Would one of you be able to concisely describe the issue for me so that I can pass it on to the service teams?
xCam

tamsky:
@cmaxwellau
> Would one of you be able to concisely describe the issue for me so that I can pass it on to the service teams?

tl;dr:
The AWS service team must create a new API endpoint, which currently doesn't exist (endpoint name is up to the service team, but this is my guess as to what it would be called:) `CreateWorkspaceDirectories`.

----

Given that there hasn't been a reply from anyone else in the last 90 days,
I'll try my best to give you my (admittedly non-expert) understanding of the issue.

To anyone reading this, it's highly likely that you know this issue better than I.
Please correct me if you see anything wrong here:

AIUI, in order to create a workspace, using the API call `CreateWorkspaces`[1] one must first have _already_ registered a workspace directory and provide a valid `DirectoryId` in the API call payload.

The act of "registering a workspace directory" is currently described in the adminguide at [2].
This appears to be an interactive process that demands a human use a web browser to click and type on the AWS web console.

Terraform automated tests require api-based environment creation/destruction, so we can't automate tests unless we have an API endpoint for creating a Workspace Directory.

- [1] https://docs.aws.amazon.com/workspaces/latest/api/API_CreateWorkspaces.html
- [2] https://docs.aws.amazon.com/workspaces/latest/adminguide/register-deregister-directory.html

tamsky:
I see now that @liamjbennett replied over in
https://github.com/hashicorp/terraform/issues/10794#issuecomment-435563758

There was a post from an AWS rep on the forum thread [1]:
> Our internal team has confirmed that there is already an existing feature request placed on our system to address this requirement.
> At this moment, we don't have an ETA on when this feature will be released. 

- [1] https://forums.aws.amazon.com/thread.jspa?messageID=749994#749994

PR #1923, opened 2017-10-17
Deprecate `region` option in aws_s3_bucket

tmshn:
## Background

AWS S3 API has bit tricky characteristics on how it handles region:

1. Each region has its own dedicated endpoint, and
1. Each endpoint only accepts requests to buckets which region matches the endpoint's; i.e: "cross-region request" is not supported.

In the current implementation of terraform-provider-aws, endpoint is defined in the `provider` (by `region` option or `endpoints.s3` option).

Therefore, the only way to create S3 bucket on a desired region is to use a provider of that region.

Although `region` option in the `aws_s3_bucket` can be used to define LocationConstaint, it works iff the region is same as the one in the provider; otherwise an error will occur. Yes, it's useless option.

## What this PR do

In this pr, I deprecated the region option (though it should remain as a readable attribute).

I suggest to remove this option in a future version, but leave core team to decide when.

#### AWS Documents

> Endpoints currently do not support cross-region requests—ensure that you create your endpoint in the same region as your bucket

[Endpoints for Amazon S3 - Amazon Virtual Private Cloud](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-endpoints-s3.html)

[AWS Regions and Endpoints - Amazon Web Services](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region)

#### Relevant issues

- https://github.com/terraform-providers/terraform-provider-aws/issues/1656
- https://github.com/terraform-providers/terraform-provider-aws/issues/592
- https://github.com/terraform-providers/terraform-provider-aws/issues/867
- https://github.com/terraform-providers/terraform-provider-aws/issues/762


tmshn:
Ping

tmshn:
@Ninir Do you have any plan for reviewing or merging this?

bflad:
Hi @tmshn 👋 Thank you for submitting this and sorry it has been pending review for a long time.

One of the reasons this may have been sitting here is that Terraform Providers currently do not have a way to signal to Terraform that an attribute deprecation is only for the ability to configure it. As such, I believe Terraform configurations referencing the attribute will still get a deprecation warning, which is not behavior we wish to introduce.

I have added https://github.com/terraform-providers/terraform-provider-aws/issues/592 to our upcoming 3.0.0 major release milestone and will denote there that we should make the breaking change of removing `Optional: true` from the attribute.

PR #1943, opened 2017-10-17
data_source_aws_security_group: add description

### LINKED ISSUE #1940: Provide support for security group rules description.
_This issue was originally opened by @ssala7 as hashicorp/terraform#16372. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

#https://www.terraform.io/docs/providers/aws/d/security_group.html 

Add a description/comment for ingress and egress rules.

Example:

resource "aws_security_group" "my_sg" {
  name        = "my_sg"
  description = "my_sg"
  vpc_id      = "vpc-xxxxxxx"

  ingress {
    from_port   = 443
    to_port     = 443
    protocol    = "tcp"
    cidr_blocks = ["0.0.0.0/0"]
    description = "Open to internet"
  }

darkowlzz:
From #1940 I came across https://www.terraform.io/docs/providers/aws/d/security_group.html#description where `description` is an attribute reference. But in real, `description` isn't available. This change adds `description` to `data_source_aws_security_group`.

This doesn't f-ixes #1940, which talks about description for the SG rules, which I see already exist 🤷‍♀️.

### DIFF
--- a/aws/data_source_aws_security_group.go
+++ b/aws/data_source_aws_security_group.go
@@ -38,6 +38,11 @@ func dataSourceAwsSecurityGroup() *schema.Resource {
 			},
 
 			"tags": tagsSchemaComputed(),
+
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
--- a/aws/data_source_aws_security_group_test.go
+++ b/aws/data_source_aws_security_group_test.go
@@ -21,6 +21,7 @@ func TestAccDataSourceAwsSecurityGroup(t *testing.T) {
 				Config: testAccDataSourceAwsSecurityGroupConfig(rInt),
 				Check: resource.ComposeTestCheckFunc(
 					testAccDataSourceAwsSecurityGroupCheck("data.aws_security_group.by_id"),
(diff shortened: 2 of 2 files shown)


aws/data_source_aws_security_group.go
```diff
@@ -38,6 +38,12 @@ func dataSourceAwsSecurityGroup() *schema.Resource {
 			},
 
 			"tags": tagsSchemaComputed(),
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
```
Ninir: Optional is used when the user can pass a value for this attribute. Since we are only retrieving the description (and we don't need it to retrieve the Security Group info), this is not needed here and can be removed

darkowlzz:
@Ninir thanks for the review.

Removed `Optional`.

About adding the attribute to documentation, isn't it already there in the docs? https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/d/security_group.html.markdown#attributes-reference

Should I add it to some other docs too?

Issue #1838, opened 2017-10-09
Impossible to create LBs of type `network`


else:
I just gave #1806 a spin (on 5a7d0ccdc15f122eb6aab4a3dd1b493c7e8d0cf7). With the current state it looks like I am unable to create LBs of type `network`:

* Health check path and matcher only exist for LBs of type `http[s]`, resulting in a crash
* The target group protocol must support protocols of type `tcp`
* Setting subnets is not supported for LBs of type `network`, instead availability zones have to be provided

First two things are easy to fix, however I don't know how to properly address the last point, because AZs cannot be passed as part of the `CreateLoadBalancerInput` and instead have to be set via `lb.SetAvailabilityZones`.

bubunyo:
NLBs will come with v1.0.1 which is yet to be released. Check the change logs.

else:
Thanks, according to #1618 I was under the impression that this should already work in latest master.

jwinter:
I was under the same impression, i.e. that after #1806 was merged Network Load Balancers would work. @bubunyo do you know of any additional work in progress for NLB support? I don't see any other Issues or PRs.

bubunyo:
I think NLB support has been implemented and merged into master which is yet to be released in v1.0.1. I don't exactly know what exactly the hold up is for the release for the said version. I also don't know what the eta for that is. Perhaps @radeksimko can help with that.

radeksimko:
I will take a look this week, hopefully, as I wanted to cut a release.

Lacking features can certainly be added in future releases but if there's a known **crash** or significantly buggy behaviour introduced by #1806 then that would make me reconsider and fix the bug before cutting the release.

Do you mind providing an example config that causes a crash @else ?

else:
Sorry for being unclear, the crash only occured during my experiments when I allowed target group protocol `tcp` which is necessary for NLBs.

This is the diff that I had to apply in order to get `terraform` actually do something and not complain about erroneously configured provisioners:

```diff
diff --git a/aws/resource_aws_lb_target_group.go b/aws/resource_aws_lb_target_group.go
index 9cab9601..132fab88 100644
--- a/aws/resource_aws_lb_target_group.go
+++ b/aws/resource_aws_lb_target_group.go
@@ -420,11 +420,11 @@ func validateAwsLbTargetGroupPort(v interface{}, k string) (ws []string, errors
 
[...]

Issue #494, opened 2017-06-13
Wait for ALB Target Group health threshold


hashibot:
_This issue was originally opened by @apparentlymart as hashicorp/terraform#11451. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

There's a common pattern with autoscaling groups in Terraform of having the `aws_autoscaling_group` resource wait until a certain number of instances are healthy in an associated ELB before considering the autoscaling group as created.

This pattern doesn't work for the new *Application Load Balancer* because there's an extra indirection in the form of a *target group*.

A naive attempt to use the usual pattern with ALB might look like this:

```hcl
resource "aws_alb" "example" {
  # ...
}

resource "aws_alb_target_group" "example" {
  name = "example-${var.version_id}"
  # ...

  lifecycle {
    create_before_destroy = true
  }
}

resource "aws_autoscaling_group" "example" {
  name = "example-${var.version_id}"
  target_group_arns = ["${aws_alb_target_group.example.arn}"]
  min_elb_capacity = 2
  # ...

  lifecycle {
    create_before_destroy = true
  }
}

resource "aws_alb_listener" "example" {
  load_balancer_arn = "${aws_alb.example.arn}"
  # ...

  default_action {
    target_group_arn = "${aws_alb_target_group.example.arn}"
    type = "forward"
  }

  # Don't update target_group_arn to new target group until
  # the autoscaling group has healthy instances.
  depends_on = ["aws_autoscaling_group.example"]
}
```

The problem here is the non-obvious dependency cycle that causes a deadlock: the `min_elb_capacity` attribute on `aws_autoscaling_group` causes that resource to block completion until it sees instances in the ELB, but the instances can't get into the ELB until the `aws_alb_listener` resource is updated to include the new target group.

---

[The `DescribeTargetHealth` operation](http://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_DescribeTargetHealth.html) allows us to ask for the healthcheck status of the members of a target group without first attaching the target group to an ALB.

Terraform could detect that `target_group_arns` is set on the `aws_autoscaling_group` resource and prefer to use the ALB `DescribeTargetHealth` operation instead of the ELB status operation, which would then allow the subsequent `aws_alb_listener` update to complete only once the target group has become healthy enough to satisfy the constraint.

[...]

PR #1948, opened 2017-10-17
Rolling Blue-Green Deployments for AWS-LB (ALB/*NLB)

matt-deboer:
This PR adds the capability of rolling (zero-downtime) deployments to the newer `aws_lb` resources, similar to the rolling deployment pattern used for `aws_elb` resources. 
Even better, in this case, instances from the new ASG are available for routing until all of them are ready, and the change-over is seamless so that a given user will not received mixed-version answers from the app.

This solves #494, although the desired goal of no schema changes was not possible, due to the inherent differences in the new load-balancer resources and their decoupling from the autoscaling groups to which they route--however, if the defaults are acceptable, the 2 new optional attributes can be omitted, and the experience will be the same from the terraform user's perspective.

**Changes**: ( on the `aws_lb_listener` resource )

- Adds a `min_target_group_capacity` attribute (optional); during creation and updates, the listener will wait for this number of instances to report healthy status before updating the default action. When not specified, this value is determined from the (sum of) `DesiredCapacity` value(s) on the autoscaling group(s) targeted by the target group. Set to `0` to disable waiting.

- Adds a `wait_for_capacity_timeout` attribute (optional); this configures how long the listener should wait for the desired target group capacity before failing in error. Defaults to `10m`.

The waiting occurs on the `aws_lb_listener` resource since Target Groups only become eligible for health checks upon attachment to a Load Balancer. This approach uses a "dummy rule" (essentially non-routable) which serves to enable health checks on the target group before it is routed any requests from the load balancer.

**Edit (2017-10-23):**

**Note**: NLB is not currently supported due to the lack of that capability in the provider overall (follow #1838 for details)


bacoboy:
I can't tell if there is anything left to be done on this?

website/docs/r/lb_listener.html.markdown
```diff
@@ -61,6 +72,60 @@ The following attributes are exported in addition to the arguments listed above:
 * `id` - The ARN of the listener (matches `arn`)
 * `arn` - The ARN of the listener (matches `id`)
 
+## Waiting for Capacity
+
+A newly-created Target Group is initially empty, with it's eventual members usually
+composed of ASG instances that initialize as the ASG comes up to it's desired
```
seh: - [ ] s/it's/its/

aws/resource_aws_lb_listener_rolling_update_test.go
```diff
@@ -0,0 +1,402 @@
...
+		url:            url,
+		maxConcurrency: maxConcurrency,
+		errors:         make(chan error, maxConcurrency),
+		successes:      make(chan string, maxConcurrency),
+		killSwitch:     make(chan struct{}, maxConcurrency),
+	}
+}
+
+// Poll causes the poller to continuously test the output of
+// it's configured url with the specified number of concurrent
```
seh: - [ ] s/it's/its/
- [ ] s/url/URL/

aws/resource_aws_lb_listener_rolling_update_test.go
```diff
@@ -0,0 +1,402 @@
...
+// clients.
+func (p *rollingVersionPoller) Poll(concurrency int) error {
+
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	errCount, err := p.waitForURLReady()
+	if err != nil {
+		return fmt.Errorf("Timed out waiting for url to become initially ready; %v", err)
```
seh: - [ ] s/url/URL/

aws/resource_aws_lb_listener_rolling_update_test.go
```diff
@@ -0,0 +1,402 @@
...
+			go p.runPoller()
+			p.pollers++
+		} else {
+			return fmt.Errorf("Already reached a maximum of %d pollers; no more will be created", p.pollers)
+		}
+	}
+	return nil
+}
+
+// waitForURLReady continuously pings the configured url until
```
seh: - [ ] s/url/URL/

aws/resource_aws_lb_listener_rolling_update_test.go
```diff
@@ -0,0 +1,402 @@
...
+					startVersion, endVersion, version)
+				return
+			}
+		}
+	}
+}
+
+// Complete stops the poller's clients and returns
+// an aggregate error describing any of the errors
+// which occurred while polling the configured url
```
seh: - [ ] s/url/URL/
- [ ] Punctuate this sentence.

aws/resource_aws_lb_listener_rolling_update_test.go
```diff
@@ -0,0 +1,402 @@
...
+						return err
+					},
+				),
+			},
+		},
+	})
+}
+
+// rollingVersionPoller exists to continuously ping a particular
+// url, testing the string result returned, and treating it the 'version' for
```
seh: - [ ] s/url/URL/

aws/resource_aws_lb_listener_rolling_update_test.go
```diff
@@ -0,0 +1,402 @@
...
+// that endpoint.
+//
+// Any errors in fetching a string value from the endpoint are reflected as errors
+// when the Complete method is called.
+//
+// The first string value returned from the endpoint is considered the 'blue' version;
+// after that, the endpoint is allowed to return one other value, considered the
+// 'green' version--once the endpoint has returned a value for the 'green' version,
+// it will be considered an error if the endpoint returns any other value.
+//
```
seh: - [ ] Remove blank comment line.

markwellis:
Is this PR going to be merged? Is there another way to do this without it being merged? Thanks

markwellis:
@aeschright @seh any chance this can be looked at? I really will make my life so much better.

Thanks

kaplanben:
any chance this will get merged any time soon? It's causing production errors on apply..

PR #1951, opened 2017-10-18
Add "snapshot_identifier" to "aws_db_instance".

javierbeaumont:
Add "snapshot_identifier" to "aws_db_instance" to link with "aws_db_snapshot" id.

### DIFF
--- a/website/docs/d/db_snapshot.html.markdown
+++ b/website/docs/d/db_snapshot.html.markdown
@@ -25,6 +25,7 @@ resource "aws_db_instance" "default" {
   password             = "bar"
   db_subnet_group_name = "my_database_subnet_group"
   parameter_group_name = "default.mysql5.6"
+  snapshot_identifier  = "${data.aws_db_snapshot.db_snapshot.id}"
 }
 
 data "aws_db_snapshot" "db_snapshot" {


website/docs/d/db_snapshot.html.markdown
```diff
@@ -25,6 +25,7 @@ resource "aws_db_instance" "default" {
   password             = "bar"
   db_subnet_group_name = "my_database_subnet_group"
   parameter_group_name = "default.mysql5.6"
+  snapshot_identifier  = "${data.aws_db_snapshot.db_snapshot.id}"
```
Ninir: Hi @javierbeaumont 

This will certainly bring a circular dependency issue where `aws_db_instance` has a dependency to `data.aws_db_snapshot.db_snapshot` and the latter also has a dependency on `aws_db_instance.default.identifier` regarding the identifier.

Were you having an issue with the documentation?

javierbeaumont: Ok, sorry I see my mistake. My user case was the creation of a database with a snapshot of another database and I assumed the documentation error.

In any case I see a problem with this implementation. I can have two scenarios:
- Create two databases and both use first database's snapshot.
- Create two databases and only the second one use first database's snapshot.

Is it possible to resolve both cases?

Ninir: @javierbeaumont I think so! using one or 2 data sources, but it may be possible. Could you try it and tell us how it goes?

I will also close this issue if you don't mind: if you encounter any issue with a given use case, could you open a new issue?

Would help :)

Thanks!

edmorley:
I think #3623 should make how to use the data source clearer, but without adding a circular dependency.

PR #1958, opened 2017-10-19
Support VPC configuration of aws_elasticsearch_domain resources.

handlerbot:
For https://github.com/terraform-providers/terraform-provider-aws/issues/1949

### DIFF
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -336,7 +336,7 @@ func waitForElasticSearchDomainCreation(conn *elasticsearch.ElasticsearchService
 			return resource.NonRetryableError(err)
 		}
 
-		if !*out.DomainStatus.Processing && out.DomainStatus.Endpoint != nil {
+		if !*out.DomainStatus.Processing && (out.DomainStatus.Endpoint != nil || out.DomainStatus.Endpoints != nil) {
 			return nil
 		}
 
@@ -379,9 +379,6 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 	d.Set("domain_id", ds.DomainId)
 	d.Set("domain_name", ds.DomainName)
 	d.Set("elasticsearch_version", ds.ElasticsearchVersion)
-	if ds.Endpoint != nil {
-		d.Set("endpoint", *ds.Endpoint)
-	}
 
 	err = d.Set("ebs_options", flattenESEBSOptions(ds.EBSOptions))
 	if err != nil {
@@ -408,6 +405,21 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 		err = d.Set("vpc_id", *ds.VPCOptions.VPCId)
 		if err != nil {
 			return err
+		}
+		endpoints := pointersMapToStringList(ds.Endpoints)
+		err = d.Set("endpoint", endpoints["vpc"])
+		if err != nil {
+			return err
+		}
+		if ds.Endpoint != nil {
+			return fmt.Errorf("%q: Elasticsearch domain in VPC expected to have null Endpoint value", d.Id())
+		}
+	} else {
+		if ds.Endpoint != nil {
+			d.Set("endpoint", *ds.Endpoint)
+		}
+		if ds.Endpoints != nil {
+			return fmt.Errorf("%q: Elasticsearch domain not in VPC expected to have null Endpoints value", d.Id())
 		}
 	}
 
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -143,6 +143,12 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"availability_zones": {
+							Type:     schema.TypeSet,
+							Computed: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
+						},
 						"security_group_ids": {
 							Type:     schema.TypeSet,
 							Optional: true,
@@ -155,20 +161,12 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 							Set:      schema.HashString,
 						},
+						"vpc_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
 					},
 				},
-			},
-			"vpc_availability_zones": {
-				Type:     schema.TypeSet,
-				Optional: true,
-				Computed: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
-				Set:      schema.HashString,
-			},
-			"vpc_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
 			},
 			"elasticsearch_version": {
 				Type:     schema.TypeString,
@@ -391,14 +389,6 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 	}
 	if ds.VPCOptions != nil {
 		err = d.Set("vpc_options", flattenESVPCDerivedInfo(ds.VPCOptions))
-		if err != nil {
-			return err
-		}
-		err = d.Set("vpc_availability_zones", schema.NewSet(schema.HashString, flattenStringList(ds.VPCOptions.AvailabilityZones)))
-		if err != nil {
-			return err
-		}
-		err = d.Set("vpc_id", *ds.VPCOptions.VPCId)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


aws/resource_aws_elasticsearch_domain.go
```diff
@@ -137,6 +137,38 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
...
+				},
+			},
+			"vpc_availability_zones": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
+			"vpc_id": {
```
radeksimko: Is there any particular reason for keeping those two fields outside of `vpc_options` and drift away from the API?

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -431,6 +504,17 @@ func resourceAwsElasticSearchDomainUpdate(d *schema.ResourceData, meta interface
 		}
 	}
 
+	if d.HasChange("vpc_options") {
+		options := d.Get("vpc_options").([]interface{})
+
+		if len(options) > 1 {
+			return fmt.Errorf("Only a single vpc_options block is expected")
```
radeksimko: We can do the same validation just by adding `MaxItems: 1` to the field in the schema which is a bit more idiomatic and more importantly the user will get the validation error during plan, rather than apply.

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -448,3 +467,48 @@ resource "aws_elasticsearch_domain" "example" {
...
+  availability_zone = "${data.aws_availability_zones.available.names[0]}"
+}
+
+resource "aws_default_subnet" "second" {
+  availability_zone = "${data.aws_availability_zones.available.names[1]}"
+}
+
+resource "aws_security_group" "first" {
+  vpc_id = "${aws_default_vpc.default.id}"
+}
```
radeksimko: To avoid sharing any resources and allow running multiple tests in parallel do you mind building custom VPC & subnets here, instead of creating default ones?

aws/resource_aws_elasticsearch_domain.go
handlerbot: Done. I was cribbing/copying/keeping style with the other optional config sections in this provider, I didn't know that `MaxItems` was even an option! :-/

aws/resource_aws_elasticsearch_domain_test.go
handlerbot: Done & re-running this test as I write this. It hasn't completed yet, but it brought up the VPC and subnets and SGs, and the ES domain is creating, so I have very high confidence the tear-down will work. :-)

aws/resource_aws_elasticsearch_domain.go
handlerbot: The ES-in-VPC API is kind of weird compared to the other AWS APIs, from what I've seen. You don't actually pass in the `vpc_id` or `availability_zones` as a parameter to the API, you pass in a list of subnets that you want the ES domain to have endpoints in, and AWS enforces that all of those subnets are in the same VPC, and calculates `vpc_id` and `availability_zones` from the subnets, and returns them to you in a slightly different datatype.

Structure `VPCOptions` has `SubnetIds` and `SecurityGroupIds`, and you use that for create and update requests for the domain.

Structure `VPCDerivedInfo` has `VPCId`, `SubnetIds`, `AvailabilityZones`, and `SecurityGroupIds`, and you get that back as part of a `ElasticsearchDomainStatus` struct during create or delete requests.

And then there's a structure `VPCDerivedInfoStatus`, which contains both a `VPCDerivedInfo` and an `OptionStatus` structure (unrelated to VPC), which you get back as part of `ElasticsearchDomainConfig` if you describe the domain.

And then to make your life even more interesting, the API uses the name `VPCOptions` to refer to all three of the structures, without even trying to do anything Go-idiomatic like defining an interface that just has the two common fields (which would work for `VPCOptions` and `VPCDerivedInfo`, but not for `VPCDerivedInfoStatus`, because `VPCDerivedInfo` is a named member of it).

```
$ grep 'VPCOptions.*structure' vendor/github.com/aws/aws-sdk-go/service/elasticsearchservice/api.go
	VPCOptions *VPCOptions `type:"structure"`
	VPCOptions *VPCDerivedInfoStatus `type:"structure"`
	VPCOptions *VPCDerivedInfo `type:"structure"`
	VPCOptions *VPCOptions `type:"structure"`
```

So, a) the whole thing is kind of a mess, but after puzzling all of that out, I figured the least complicated thing was, for the Terraform code, to keep `vpc_options` equivalent to `VPCOptions`.

But the end of it is: You can't specify the VPC id or the availability zones via the API, you have to let AWS compute them for you and return them, so they're never options that an operator will specify. And, once they are returned, I wanted to export them as top-level attributes of the resource, as I see other resources do, and I don't see a way to export schema sub-members as top-level attributes, other than defining them twice and exporting them twice?

Feedback encouraged if I am missing a more Terraform-idiomatic way to implement this!

agarstang: SubnetIds are globally unique. This behavior of submitting just subnets is seen elsewhere (ie [ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_CreateLoadBalancer.html)) and should be familiar to anyone used to using VPC. 

Filtering by VPC is often just client-side to narrow down subnet list.

There are exceptions of course, for resources such as security groups that go into a VPC but not into a subnet.

radeksimko: @handlerbot understood, but I don't think that gives us a reason to drift away from the API. They can both live nested under `vpc_options` if I understand correctly?

Additionally we should remove `Optional` from both mentioned fields as the user cannot specify them in the config.

radeksimko: > I don't see a way to export schema sub-members as top-level attributes, other than defining them twice and exporting them twice?

I'm not sure I follow - the user can access nested attributes of any `TypeList` like this:

```
aws_elasticsearch_domain.vpc_options.0.vpc_id
```
the index may feel a bit weird, but that's how most other resources do it, so I'd just stick with it. It's something we plan to address in core/HCL eventually.

handlerbot: I only found one example of the docs specifying a usage like that, and only in passing (`cache_nodes` in https://www.terraform.io/docs/providers/aws/r/elasticache_cluster.html). I am not personally convinced, but you folks are the bosses, so. :-)

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -448,3 +466,54 @@ resource "aws_elasticsearch_domain" "example" {
...
+resource "aws_security_group" "first" {
+  vpc_id = "${aws_vpc.elasticsearch_in_vpc.id}"
+}
+
+resource "aws_security_group" "second" {
+  vpc_id = "${aws_vpc.elasticsearch_in_vpc.id}"
+}
+
+resource "aws_elasticsearch_domain" "example" {
+  domain_name = "tf-test-in-vpc"
```
Ninir: Would it be possible to add randomization to this name? 'would help avoid conflicts when running tests :)

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -145,6 +145,24 @@ func TestAccAWSElasticSearchDomain_complex(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckESDomainDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccESDomainConfig_vpc(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckESDomainExists("aws_elasticsearch_domain.example", &domain),
+				),
+			},
```
Ninir: Can you add another step, where we would update the VPC configuration, so that we ensure the update works as expected?

radeksimko:
Thanks, there is one last thing before we can merge this. The attached test is failing:

```
=== RUN   TestAccAWSElasticSearchDomain_vpc
--- FAIL: TestAccAWSElasticSearchDomain_vpc (61.01s)

		* aws_elasticsearch_domain.example: 1 error(s) occurred:

		* aws_elasticsearch_domain.example: ValidationException: Before you can proceed, you must enable a service-linked role to give Amazon ES permissions to access your VPC.
			status code: 400, request id: 9830a6e4-b965-11e7-a4b8-8330fd83b650
FAIL
```

I assume the config in the test will need an IAM role with the right policy.
Do you mind taking a look into it?

aws/resource_aws_elasticsearch_domain_test.go
handlerbot: @Ninir Sure, I actually had that (name randomization) earlier and then removed it because I dumbly assumed that you could have duplicate ES domain names within your account as long as they were unique within the VPC, but that turned out to be incorrect, so I've got the restoration of the randomness in my pending change.

aws/resource_aws_elasticsearch_domain_test.go
handlerbot: @Ninir Yeah, I can do that, once I sort the larger IAM problem for the test (update re: that coming next).

handlerbot: Disregard "larger IAM problem", I figured it out. 🎉

Issue #1762, opened 2017-09-27
[FEATURE REQUEST] Import for ecs resources


hashibot:
_This issue was originally opened by @jensskott as hashicorp/terraform#16180. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I would like to be able to import the ecs resources such as cluster with container instances aswell as services and taskdefinitions.

bflad:
`aws_ecs_cluster` import has been released in [terraform-provider-aws version 1.7.0](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#170-january-12-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

dchait-simplr:
Where do we stand with task importing?

luisamador:
Task definition importing is not supported yet :(

luisamador:
I've just found myself int he same situation, It would be really nice to be able to import task definition resources soon. Thank you.

bflad:
Support for importing `aws_ecs_task_definition` resources has been merged in and will release with version 1.51.0 of the AWS provider, likely later today. Since this rounds out support for all the existing ECS resources, closing this issue!

PR #1962, opened 2017-10-19
aws_ecs_cluster: add importer

darkowlzz:
Related to #1762


TylerBrock:
Would love to see this get merged if possible @radeksimko.

kwent:
I also need this feature

loivis:
@bflad With your suggestion, are we forcing to use arn to import the resource? I think it's easier to import using cluster name only.

bflad:

Issue #50, opened 2017-06-13
what's the proper way of dealing with an instance marked for retirement?


hashibot:
_This issue was originally opened by @rgs1 as hashicorp/terraform#3915. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

In AWS (surely there's something similar in other environments) an instance can be marked for retirement:

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-retirement.html

From a quick git grep in the repo and search in the issues tracker, I couldn't see any logic to handle these instances. Has any thought been put into what Terraform should provide to deal with these instances (or maybe there is a way of doing this already)?

hashibot:
_This comment was originally opened by @apparentlymart as https://github.com/hashicorp/terraform/issues/3915#issuecomment-156586810. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

This is a very good question.

So far when I've encountered this problem I've just manually run `terraform taint` on the resource in question and had Terraform rebuild it, but I've only done this so far for instances thwt were ephemeral anyway, and so they were designed to be replaced often.

One way to deal with this issue, at least for EBS-based instances, is to restart the instance so that it migrates to new hardware. That could potentially be done outside of Terraform, but for Terraform to do it would look like what was disussed in terraform-providers/terraform-provider-aws#22.

We could potentially have an option on the `ec2_instance` resource that tells Terraform to explicitly restart an instance when it's retired, which could then just be done as part of "updating" the instance, but it's not great to do that sort of multi-step workflow inside a single resource action because Terraform can't easily represent the intermediate steps in the state and thus won't be able to recover well if any part of the process fails.

So yeah... definitely an important use-case but one I think Terraform's current model doesn't handle very well. Thanks for starting this discussion! I'm curious to see what other folks come up with.

hashibot:
_This comment was originally opened by @k3ypad as https://github.com/hashicorp/terraform/issues/3915#issuecomment-159927185. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original comment is below._

<hr>

[...]

PR #1980, opened 2017-10-20
start/stop instances - should close #22, #1998 and maybe #50

### LINKED ISSUE #22: Stop instances
_This issue was originally opened by @johnjelinek as hashicorp/terraform#1579. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

How can I specify the state of my instances to be `stopped`?

artburkart:
I've added a couple acceptance tests for this feature. Since I'd already taken a stab at this about a year ago and never thought of a good way to deal with the transition from `terminated` to `stopped` and don't really think it's necessary to get the brunt of the value here, I've made it so you can only set `instance_state` to `stopped` or `running`. You can even set an instance to `stopped` at creation time, which might be useful for something, ¯\_(ツ)_/¯. Also, while writing my tests, I realized you can't stop instances when their root block device is of type `instance-store`. I had no idea.

I haven't had a ton of luck getting my code merged into this code base, so please don't hesitate to tell me if there's something I can do to improve this PR.

Cheers!

UPDATE: Also going to reference https://github.com/hashicorp/terraform/issues/1579 here so people know I did it.

### DIFF
--- a/aws/data_source_aws_instance_test.go
+++ b/aws/data_source_aws_instance_test.go
@@ -1,9 +1,9 @@
 package aws
 
 import (
-	"testing"
-
 	"fmt"
+	"regexp"
+	"testing"
 
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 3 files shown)


artburkart:
🤔 🤔 🤔

artburkart:
What does "size/L" mean?

ericofusco:
@artburkart @radeksimko 

I believe this implementation should be improved in order to use with other attributes that don't require the resource to be recreated, just state to be temporarily changed to make the changes.

For example we also have the attributes **user_data** and **ebs_optimized** that require the instance to be at stopped state and the instance is recreated for them which is not necessary.

One approach that I suggest is adding another setting to the attribute like ForceNew that tells which state the resource should be at before making that change. Once the changes are verified terraform checks if resource needs to be at one particular state and changes the resource first and then make the changes and returns to the original value if it's different from the required one. Ignore that if resource needs to be recreated. Maybe something that could be added to terraform core instead of aws provider.

Also terraform should tell that the resource will be changed from "running" to "stopped" on this case, users should know that, when instance_type is changed now terraform shows the change like it changing while the instance is running, it doesn't show that a particular state is required and it might impact user's production environment by stopping the instance.

Please let me know youf thoughts.

artburkart:
@ericofusco - All good thoughts. I'm not interested in supporting this PR. If someone would like to take the code and improve it, I'll leave my branch around for them to use. In the meantime, I'll close this PR in favor of someone else taking ownership. Cheers!

PR #1986, opened 2017-10-20
resource/aws_spot_instance_request: Fix being able to stop requests

tomelliff:
As mentioned in https://github.com/terraform-providers/terraform-provider-aws/pull/1735#issuecomment-338249736 this was failing locally for me.

Fixes an issue when setting the instance interruption behaviour to anything other than terminate (hibernate has since been added but not included in this PR) and adds a test to cover it.

### DIFF
--- a/aws/resource_aws_spot_instance_request.go
+++ b/aws/resource_aws_spot_instance_request.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsSpotInstanceRequest() *schema.Resource {
@@ -81,8 +82,12 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
 			s["instance_interruption_behaviour"] = &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "terminate",
+				Default:  ec2.InstanceInterruptionBehaviorTerminate,
 				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					ec2.InstanceInterruptionBehaviorTerminate,
+					ec2.InstanceInterruptionBehaviorStop,
+				}, false),
 			}
 			return s
 		}(),
@@ -115,7 +120,6 @@ func resourceAwsSpotInstanceRequestCreate(d *schema.ResourceData, meta interface
 			ImageId:             instanceOpts.ImageID,
 			InstanceType:        instanceOpts.InstanceType,
 			KeyName:             instanceOpts.KeyName,
-			Placement:           instanceOpts.SpotPlacement,
 			SecurityGroupIds:    instanceOpts.SecurityGroupIDs,
 			SecurityGroups:      instanceOpts.SecurityGroups,
 			SubnetId:            instanceOpts.SubnetID,
@@ -130,6 +134,11 @@ func resourceAwsSpotInstanceRequestCreate(d *schema.ResourceData, meta interface
 
 	if v, ok := d.GetOk("launch_group"); ok {
 		spotOpts.LaunchGroup = aws.String(v.(string))
+	}
+
+	// Placement GroupName can only be specified when instanceInterruptionBehavior is not set to 'stop'
+	if v, exists := d.GetOkExists("instance_interruption_behaviour"); v.(string) == ec2.InstanceInterruptionBehaviorTerminate || !exists {
+		spotOpts.LaunchSpecification.Placement = instanceOpts.SpotPlacement
 	}
 
 	// Make the spot instance request
--- a/aws/resource_aws_spot_instance_request.go
+++ b/aws/resource_aws_spot_instance_request.go
@@ -87,6 +87,7 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{
 					ec2.InstanceInterruptionBehaviorTerminate,
 					ec2.InstanceInterruptionBehaviorStop,
+					ec2.InstanceInterruptionBehaviorHibernate,
 				}, false),
 			}
 			return s
@@ -136,7 +137,7 @@ func resourceAwsSpotInstanceRequestCreate(d *schema.ResourceData, meta interface
 		spotOpts.LaunchGroup = aws.String(v.(string))
 	}
 
-	// Placement GroupName can only be specified when instanceInterruptionBehavior is not set to 'stop'
+	// Placement GroupName can only be specified when instanceInterruptionBehavior is not set or set to 'terminate'
 	if v, exists := d.GetOkExists("instance_interruption_behaviour"); v.(string) == ec2.InstanceInterruptionBehaviorTerminate || !exists {
 		spotOpts.LaunchSpecification.Placement = instanceOpts.SpotPlacement
 	}
--- a/aws/resource_aws_spot_instance_request_test.go
+++ b/aws/resource_aws_spot_instance_request_test.go
@@ -389,7 +389,7 @@ func testAccCheckAWSSpotInstanceRequestAttributesVPC(
 	}
 }
 
-func TestAccAWSSpotInstanceRequestStopInterrupt(t *testing.T) {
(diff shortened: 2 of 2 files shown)


tomelliff:
Well... that makes no sense to me. Running that test locally fails (although I've just noticed it's failing with a different issue in the test):
```
		
		* aws_spot_instance_request.foo: 1 error(s) occurred:
		
		* aws_spot_instance_request.foo: Error requesting spot instances: InvalidParameterCombination: An AMI with a root device type of 'instance-store' is not supported when instanceInterruptionBehavior is set to 'stop'.
			status code: 400, request id: 89e4fe73-fdde-4261-ac6e-5f263b8548d0

Process finished with exit code 1
```

tomelliff:
I need to head off now and the Travis build is still pending but locally I now get
```
		
		* aws_spot_instance_request.foo: 1 error(s) occurred:
		
		* aws_spot_instance_request.foo: Error requesting spot instances: InvalidParameterCombination: The parameter GroupName within placement information cannot be specified when instanceInterruptionBehavior is set to 'stop'.
			status code: 400, request id: 189ebe25-fb67-402d-8f20-7d54b5cfef9f

Process finished with exit code 1
```

tomelliff:
I was considering adding a simple validate function on the instance interruption behaviours as well (because plan time failures are always nicer than apply time failures when you typo something) and then spotted that we have `instance_interruption_behavior` in the docs but it's `instance_interruption_behaviour` in the actual code.

The British part of me wants to change the docs to `instance_interruption_behaviour` but it doesn't align with the AWS API and most things want US spelling instead. Unfortunately, changing the key in the schema will then technically be a breaking change. However, it doesn't currently work so I'm not sure if people want to consider it a non breaking change and I can just change it all to `behavior`? @radeksimko  any thoughts?

tomelliff:
Also it's a shame that that Travis build on Friday got canned rather than fail because it doesn't show the failing build but if you want to verify it just checkout the commit with the test from my fork and run the provided test.

tomelliff:
I decided to go ahead with the "breaking" change considering it didn't work anyway this seems like the best solution instead of leaving it as is and updating docs or adding the US spelling and deprecating `instance_interruption_behaviour`.

tomelliff:
It might be worth deprecating the `aws_spot_instance_request` resource in light of https://aws.amazon.com/blogs/aws/amazon-ec2-update-streamlined-access-to-spot-capacity-smooth-price-changes-instance-hibernation/ and as mentioned in https://github.com/terraform-providers/terraform-provider-aws/issues/1243

Also if the validator is kept from here we need to add whatever the enum value is for `hibernate` (probably that) now that's been added: https://aws.amazon.com/about-aws/whats-new/2017/11/amazon-ec2-spot-lets-you-pause-and-resume-your-workloads/

tomelliff:
I've migrated away from spot instance requests directly to only using spot instances in ASGs so have no real need for this anymore but it might still be useful for others if anyone is happy to review this pull request?

I still feel like the `aws_spot_instance_request` might want deprecating in favour of refactoring the `aws_instance` resource to use launch templates so that resource can more directly create spot instance requests for a much simpler API but that's going to be a fair bit more work and I'm not sure how many people are using this resource as is.

billyfoss:
I would like to see this fixed.  Between the spelling mismatch in documentation and being consistent with the AWS API, I agree with removing the 'u'.  
I agree that long term, the focus should be on directly using the spot support in aws_instance, but if a little work will get this working, I'd prefer to deprecate the usage with working code.  

I've merged this PR into my fork and have added a little to support for the hibernate option.  I tested the 'stop' support and it is working for my usage.  I have not tested the 'hibernate' support.

aws/resource_aws_spot_instance_request.go
```diff
@@ -78,11 +78,12 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			}
-			s["instance_interruption_behaviour"] = &schema.Schema{
```
bflad: Can we please save this breaking change for another PR? It'll need to be handled in the 2.0.0 upgrade when that occurs. In the meantime, we should deprecate the existing attribute (but that change can be made separately).

aws/resource_aws_spot_instance_request.go
```diff
@@ -78,11 +78,12 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
...
-				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "terminate",
-				ForceNew: true,
+			s["instance_interruption_behavior"] = &schema.Schema{
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "terminate",
```
bflad: Nitpick: SDK provided constant available: `ec2.InstanceInterruptionBehaviorTerminate`

aws/resource_aws_spot_instance_request.go
```diff
@@ -78,11 +78,12 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
...
+				ForceNew:     true,
+				ValidateFunc: validateInstanceInterruptionBehavior,
```
bflad: Nitpick: Custom function can be removed for:

```go
  ec2.InstanceInterruptionBehaviorStop,
  ec2.InstanceInterruptionBehaviorTerminate,
}, false),
```

bflad: There is also `ec2.InstanceInterruptionBehaviorHibernate,` available.

tomelliff:
Still want the change separating when it's currently broken anyway? I'd definitely agree if it was working but it has never worked so I don't think we have to worry about breaking changes here and it's then simpler to fix both things in one go. Or separate them and merge the "breaking" change first without incrementing the major version and then merge the fix.

aws/resource_aws_spot_instance_request.go
```diff
@@ -132,6 +132,11 @@ func resourceAwsSpotInstanceRequestCreate(d *schema.ResourceData, meta interface
 		spotOpts.LaunchGroup = aws.String(v.(string))
 	}
 
+	// Placement GroupName can only be specified when instanceInterruptionBehavior is not set to 'stop'
+	if v, exists := d.GetOkExists("instance_interruption_behavior"); v == "terminate" || !exists {
```
bflad: I think this needs to be `v.(string) == ec2.InstanceInterruptionBehaviorTerminate`

bflad:
Sorry if I'm being dense, but I don't understand where this is coming from:

>  but it has never worked

It looks like the attribute was passed in just fine before with the old (incorrectly documented) name. Plenty of people could have changed their configurations to use the different spelling and it should have worked just fine for them, unless I'm missing something.

bflad:
For reference, people could have found this issue and adjusted their configurations: https://github.com/terraform-providers/terraform-provider-aws/issues/2670

bflad:
PR to fix the documentation: #4195

tomelliff:
I guess technically someone might be using it to directly specify they want it to `terminate` instead of leaving it as the default which I hadn't considered until now. In that case then yeah I guess this would be a breaking change.

tomelliff:
@bflad I've removed the breaking change, rebased on master and squashed my commits. Thanks for the review and comments earlier, and sorry it's taken a while to find time to address the feedback.

I've also added a test for hibernating instances and expanded the validation for that to work. If you'd rather I can split that commit out to a different PR to make the PR as small as possible.

Did you also want me to raise a PR with the breaking change to move things over to `instance_interruption_behavior` that can be merged when 2.0 is ready to be released?

piontec:
Hi!
We're hitting this issue as well - we can't start spot instance with `interruption_behavior = stop` at all. Can you please review this PR @bflad ?

tomelliff:
@bflad not sure I follow about the validation as there is validation for all 3 behaviours right now (`terminate`, `stop`, `hibernate`) and refactored out of a stand alone validation function as per your feedback. Am I missing something there? I guess there's not an explicit `terminate` acceptance test if that's what you meant?

And did you want me to raise the breaking change PR to move things to use `instance_interruption_behavior`?

bflad:
> I guess there's not an explicit terminate acceptance test if that's what you meant?

We can provide an error (and acceptance test that error) if there is an invalid configuration (defining `placement_group` and a bad `instance_interruption_behaviour`) rather than silently skipping the invalid part of the configuration.

> And did you want me to raise the breaking change PR to move things to use instance_interruption_behavior?

If you so desire. The first PR should support both the old and new attribute names (using `ConflictsWith` between them).

Issue #1877, opened 2017-10-12
Possible typo in documentation for data aws_route53_zone


abinashmeher999:
### Affected Resource(s)
- data "aws_route53_zone"

### Reason
There seems to be a typo in the docs for `data "aws_route53_zone"` where in [Argument Reference](https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/d/route53_zone.html.markdown#argument-reference) > `tags`, the sentence '... must exactly match a pair on the desired security group.' seems to have mistakenly referred 'Hosted Zone' as 'security group'. The same reflects in the [latest docs](https://www.terraform.io/docs/providers/aws/d/route53_zone.html) on the website. Kindly confirm if that's the case. I can send a PR to fix it.

### Relevant Terraform Version
There's no immediate need to deploy it to the site.

abinashmeher999:
Closed in favor of #1997

PR #1997, opened 2017-10-21
Fix typo in d/route53_zone.html

abinashmeher999:
### Reasoning for docs update
There was a typo in the docs for `data "aws_route53_zone"` as noted in #1877.
> There seems to be a typo in the docs for `data "aws_route53_zone"` where in [Argument Reference](https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/d/route53_zone.html.markdown#argument-reference) > `tags`, the sentence '... must exactly match a pair on the desired security group.' seems to have mistakenly referred 'Hosted Zone' as 'security group'. The same reflects in the [latest docs](https://www.terraform.io/docs/providers/aws/d/route53_zone.html) on the website

### Relevant Terraform version
This update doesn't refer to a specific Terraform version but is relevant for the latest docs on the website.  There is no immediate need to deploy it to the site.


abinashmeher999:
Thanks for merging this! 🙂

PR #1999, opened 2017-10-21
New Data Source: aws_iam_policy

### LINKED ISSUE #1346: aws_iam_policy as a data source
_This issue was originally opened by @colout as hashicorp/terraform#15734. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hello,

Is it currently on your roadmap to create a data source for `aws_iam_policy`?

I built some `aws_iam_policy`s in a single state, and I consume them in higher states (I plan to have dozens of states apply these shared policies).  

The specific use case is to generate some shared polices that every instance's `aws_iam_instance_profile` would apply a-la-carte (describing ec2 tags, access an s3 bucket, etc).  

I identified two ways of accomplishing this via Terraform, and neither are ideal:  
1. We can define some `aws_iam_policy_document`s on the lower state and query for them using  `"${data.aws_iam_policy_document.policy_name.json}"`.   This is far from ideal since we plan on spinning up hundreds of ASGs, and this creates a giant mess in the AWS console.  On top of this adding a single new single new policy to all instances could exponentially increase our policy count (we'll be forced to request less-than-sane aws account limits).

2. The other option is to piece together the policy ARN using our namespacing prefix and our account number (which we'll have to input to the state as a variable now).  This doesn't feel like an ideal pattern for terraform, but it works: `arn:aws:iam::${var.aws_account_id}:policy/${var.environment_namespace}.ec2_describetags`

Thanks, and keep up the great work!

Edit:  I typo'd `aws_iam_role` in the first sentence when i meant `aws_iam_policy`.

darkowlzz:
fixes #1346

### DIFF
--- a/aws/data_source_aws_iam_policy.go
+++ b/aws/data_source_aws_iam_policy.go
@@ -1,0 +1,40 @@
+package aws
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsIAMPolicy() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsIAMPolicyRead,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"policy": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"path": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsIAMPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	d.SetId(d.Get("arn").(string))
+	return resourceAwsIamPolicyRead(d, meta)
+}
--- a/aws/data_source_aws_iam_policy_test.go
+++ b/aws/data_source_aws_iam_policy_test.go
@@ -1,0 +1,61 @@
+package aws
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSDataSourceIAMPolicy_basic(t *testing.T) {
+	policyName := fmt.Sprintf("test-policy-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAwsDataSourceIamPolicyConfig(policyName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.aws_iam_policy.test", "name", policyName),
+					resource.TestCheckResourceAttr("data.aws_iam_policy.test", "description", "My test policy"),
+					resource.TestCheckResourceAttr("data.aws_iam_policy.test", "path", "/"),
+					resource.TestCheckResourceAttrSet("data.aws_iam_policy.test", "policy"),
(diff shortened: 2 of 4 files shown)


aws/data_source_aws_iam_policy.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func dataSourceAwsIAMPolicy() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsIAMPolicyRead,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Optional: true,
```
trung: I would define as `Required: true` instead

bflad:

chris-weck:
Hi @bflad,
I´ve upgraded the tf provider to 1.10.0.
When I try this:

> _data "aws_iam_policy" "AdministratorAccess" {
>  arn = "arn:aws:iam::aws:policy/AdministratorAccess"
> }
> output "policy" {
>  value = "${aws_iam_policy.AdministratorAccess.arn}"
> }_
I will receive this error:

**Error: output 'policy': unknown resource 'aws_iam_policy.AdministratorAccess' referenced in variable aws_iam_policy.AdministratorAccess.arn**

bflad:
@chris-weck for Terraform data sources you need to prepend `data.` to the front of your interpolation:

```hcl
output "policy" {
  value = "${data.aws_iam_policy.AdministratorAccess.arn}"
}
```

[Documentation](https://www.terraform.io/docs/configuration/data-sources.html):
> Each data instance will export one or more attributes, which can be interpolated into other resources using variables of the form data.TYPE.NAME.ATTR.

chris-weck:
Oh, many thanks. I´m new with terraform. It´s working now :)

TBeijen:
Quoting @bflad:
>  I'm hoping you or others will enhance it in the future so it can lookup AWS provided policies by name to fetch their ARN.

Ran into this and was surprised it requires an ``arn`` as a lot of similar data sources (e.g. [aws_iam_user](https://www.terraform.io/docs/providers/aws/d/iam_user.html), [aws_iam_role](https://www.terraform.io/docs/providers/aws/d/iam_role.html)) operate based on ``name`` instead of ``arn``.

This looks like a nice thing to pick up and get my feet wet on contributing to Terraform.

What would be a good pattern to follow? As ``arn`` is already an argument, the situation of a user (inadvertedly) providing both ``arn`` and ``name`` should be handled. This could be:

1. Raise error, only one should be specified.
2. Use ``arn``, ignore ``name``
3. Use ``arn``, raise error if resulting resource name doesn't match given ``name``.

First seems best imo. Opinions?

bflad:
@TBeijen its probably fine to raise an error for now. You can cross-add `ConflictsWith: []string{"arn"},` to the new `name` attribute and `ConflictsWith: []string{"name"},` to the existing `arn` attribute to have Terraform automatically error if both are specified.

tomelliff:
@TBeijen are you still wanting to pick this up? I was just looking at this data source as I currently build the ARN of our managed policies and it would be nicer if we could just refer to it by name.

Also I would probably go with your third option where any combination of things can be specified but they still have to return exactly one result. So something like this should be fine:

```hcl
data "aws_iam_policy" "AdministratorAccess" {
  arn = "arn:aws:iam::aws:policy/AdministratorAccess"
  name = "AdministratorAccess"
  path = "policy"
}
```
But this would fail:
```hcl
data "aws_iam_policy" "AdministratorAccess" {
  arn = "arn:aws:iam::aws:policy/AdministratorAccess"
  name = "ReadOnlyAccess"
  path = "policy"
}
```
It would also be good to export/consume the account ID (also allowing for the magic `aws` account ID for AWS managed policies).

TBeijen:
Hi @tomelliff,
Was ambitiously planning to but so far life/work got in the way so no progress whatsoever.

Feel free to pick this up, no risk of efforts crossing.

saravanan30erd:
@TBeijen @tomelliff any update in this?

TBeijen:
> Was ambitiously planning to but so far life/work got in the way so no progress whatsoever.
> Feel free to pick this up, no risk of efforts crossing.

@saravanan30erd Above is still 100% accurate, so no progress from my part.

Issue #384, opened 2017-06-13
Feature Request: Add support for CreateVPCAssociationAuthorization AWS API


hashibot:
_This issue was originally opened by @rginev as hashicorp/terraform#10208. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hey there,

Terraform does have [route53_zone_association](https://www.terraform.io/docs/providers/aws/r/route53_zone_association.html) resource, but it works for private zones and VPCs in the same AWS account. 
If you want to associate a VPC that you created with one AWS account with a private hosted zone that you created with a different account you had to ask AWS support to create authorisation manually. 
Now they added an [API](http://docs.aws.amazon.com/Route53/latest/APIReference/API_CreateVPCAssociationAuthorization.html) for this. It would be great to have this implemented in terraform e.g. new resource.

Thanks in advance!

### References
- [Associating an Amazon VPC and a Private Hosted Zone That You Created with Different AWS Accounts](http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-associate-vpcs-different-accounts.html)

josephholsten:
@RyanJarv I don't know if you've looked here since the provider split. Let me know if you want to help get a fix together for this.

Zordrak:
Should only require the creation of an authorization resource against a specific zone, with parameter inputs for source vpc id and source vpc region.

Would love to see this implemented.

RyanJarv:
@josephholsten Hey sorry, yeah that would be awesome I've just been pretty busy and not sure when I'll have time to get back to this.

Worth mentioning this actually does work currently for creating authorizations but like @Zordrak pointed out the association resource can't describe route53 zone and fails currently. This is talked about a bit more at https://github.com/hashicorp/terraform/pull/12553. If we don't mind the association and authorization resource working together currently I think this could probably be merged as is with some added docs and a note on the issue.

robh007:
Is anyone still interested in trying to get this feature out?

RyanJarv:
Went ahead and cleaned this up a bit and added documentation in https://github.com/terraform-providers/terraform-provider-aws/pull/2005. Last PR was before the repo split.

julianalves:
Awesome! this will be very useful! Thanks for the hard work!

luisamador:
[...]

PR #2005, opened 2017-10-22
New Resource: aws_route53_vpc_association_authorization

RyanJarv:
I've been sitting on this for a bit hoping the cross account aws_route53_zone_association issue would sort it self out (#384). A few people have expressed interest in this even though the association will need to be done outside of Terraform so I think it's fine to just have a note in the docs about this issue.

Much of the code is copied from aws_route53_zone_association since the logic is pretty similar.

## Related Issues/PRs
https://github.com/hashicorp/terraform/issues/10208
https://github.com/hashicorp/terraform/pull/12553 (PR before the repo split)
https://github.com/terraform-providers/terraform-provider-aws/issues/384

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -418,6 +418,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_redshift_subnet_group":                    resourceAwsRedshiftSubnetGroup(),
 			"aws_route53_delegation_set":                   resourceAwsRoute53DelegationSet(),
 			"aws_route53_record":                           resourceAwsRoute53Record(),
+			"aws_route53_vpc_association_authorization":    resourceAwsRoute53VPCAssociationAuthorization(),
 			"aws_route53_zone_association":                 resourceAwsRoute53ZoneAssociation(),
 			"aws_route53_zone":                             resourceAwsRoute53Zone(),
 			"aws_route53_health_check":                     resourceAwsRoute53HealthCheck(),
--- a/website/docs/r/route53_zone_authorization_association.html.markdown
+++ b/website/docs/r/route53_zone_authorization_association.html.markdown
@@ -1,0 +1,62 @@
+---
+layout: "aws"
+page_title: "AWS: aws_route53_vpc_association_authorization"
+sidebar_current: "aws_route53_vpc_association_authorization"
+description: |-
+  Authorizes a VPC in a peer account to associated with a local Route53 Hosted Zone
+---
+
+# aws\_route53\_vpc\_association\_authorization
+
+Authorizes a VPC in a peer account to associated with a local Route53 Hosted Zone
+
+~> **NOTE:** Currently the route53\_zone\_assocation resource does not work across accounts. If you use this resource you will need to do the association the VPC to the Zone outside of Terraform.
+
+## Example Usage
+
+```hcl
+provider "aws" {
+    region = "us-west-1"
+    // Requester's credentials.
+}
+provider "aws" {
+    alias = "peer"
+    region = "us-east-2"
+}
+resource "aws_vpc" "foo" {
+	cidr_block = "10.6.0.0/16"
+	enable_dns_hostnames = true
+	enable_dns_support = true
+}
+resource "aws_vpc" "peer" {
+    provider = "aws.peer"
+	cidr_block = "10.7.0.0/16"
+	enable_dns_hostnames = true
+	enable_dns_support = true
+}
+resource "aws_route53_zone" "foo" {
+	name = "foo.com"
+	vpc_id = "${aws_vpc.foo.id}"
+}
+resource "aws_route53_vpc_association_authorization" "peer" {
+    zone_id = "${aws_route53_zone.foo.id}"
+    vpc_id  = "${aws_vpc.peer.id}"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `zone_id` - (Required) The private hosted zone to associate.
+* `vpc_id` - (Required) The VPC to associate with the private hosted zone.
+* `vpc_region` - (Optional) The VPC's region. Defaults to the region of the AWS provider.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The calculated unique identifier for the association.
+* `zone_id` - The ID of the hosted zone for the association.
+* `vpc_id` - The ID of the VPC for the association.
+* `vpc_region` - The region in which the VPC identified by `vpc_id` was created.
--- a/website/docs/r/route53_zone_authorization_association.html.markdown
+++ b/website/docs/r/route53_zone_authorization_association.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 Authorizes a VPC in a peer account to associated with a local Route53 Hosted Zone
 
-~> **NOTE:** Currently the route53\_zone\_assocation resource does not work across accounts. If you use this resource you will need to do the association the VPC to the Zone outside of Terraform.
+~> **NOTE:** Currently the route53\_zone\_assocation resource does not work across accounts. If you use this resource you will need to do the association between the VPC and Route53 Zone using another tool.
 
 ## Example Usage
 
--- a/aws/resource_aws_route53_vpc_association_authorization.go
+++ b/aws/resource_aws_route53_vpc_association_authorization.go
@@ -1,0 +1,127 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+func resourceAwsRoute53VPCAssociationAuthorization() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsRoute53VPCAssociationAuthorizationCreate,
+		Read:   resourceAwsRoute53VPCAssociationAuthorizationRead,
+		Delete: resourceAwsRoute53VPCAssociationAuthorizationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"zone_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"vpc_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"vpc_region": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceAwsRoute53VPCAssociationAuthorizationCreate(d *schema.ResourceData, meta interface{}) error {
+	r53 := meta.(*AWSClient).r53conn
+
+	req := &route53.CreateVPCAssociationAuthorizationInput{
+		HostedZoneId: aws.String(d.Get("zone_id").(string)),
+		VPC: &route53.VPC{
+			VPCId:     aws.String(d.Get("vpc_id").(string)),
+			VPCRegion: aws.String(meta.(*AWSClient).region),
+		},
+	}
+	if w := d.Get("vpc_region"); w != "" {
+		req.VPC.VPCRegion = aws.String(w.(string))
+	}
+
+	log.Printf("[DEBUG] Creating Route53 VPC Association Authorization for hosted zone %s with VPC %s and region %s", *req.HostedZoneId, *req.VPC.VPCId, *req.VPC.VPCRegion)
+	var err error
+	_, err = r53.CreateVPCAssociationAuthorization(req)
+	if err != nil {
+		return err
+	}
+
+	// Store association id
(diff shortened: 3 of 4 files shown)


tobyclemson:
We're running into exactly this problem and it would be great to get this merged in.

julianalves:
AWSome!

aggallim:
Is this on someone's todo list? Would love to have this functionality before I start off down the route of automating it using the aws cli!

bevanbennett:
Thumbs up for this. About half our terraform apply runs now always report an error "NotAuthorizedException: The VPC: vpc-XXXXXXX has not authorized to associate with your hosted zone." even when we authorize it explicitly or have already associated manually.

ewbankkit:
Any ideas on what is required to get this merged?

ewbankkit:
@RyanJarv I took the liberty of playing a bit with your code, see [here](https://github.com/ewbankkit/terraform-provider-aws/tree/issue-384) - no substantive changes, mainly just a rename of `aws_route53_vpc_association_authorization` to `aws_route53_zone_association_authorization` and some code cleanup to match other recent changes.
In testing I hit the issue I describe [here](https://github.com/terraform-providers/terraform-provider-aws/issues/384#issuecomment-357049374) which is an artifact of the way the Route 53 API works - No zone association resource with an ID is exposed so we have to query the zone resource and don't have permissions to do that in the account that didn't create the zone.

RyanJarv:
@ewbankkit Awesome, thanks! I will try to take a look at your changes soon and get them added to this PR.

Since the error you mentioned is expected in that use case, and it's likely more people will run into it I'm wondering if we can/should add a message about the issue there.

I would also be interested if you have any ideas on how to approach that issue. It's been a while since I looked at this but it may be worth seeing if any other resources have dealt with something similar since.

bflad:
@ewbankkit / @RyanJarv sorry for the long maintainer silence on this PR. Can you reach out to me if/when its ready to be looked at or if you have specific questions? Thanks!

smastrorocco:
Any chance this is getting merged soon?  Running into scenario where I need this.  Any workaround in meantime?

RyanJarv:
@smastrorocco I would +1 the AWS forms thread at https://forums.aws.amazon.com/thread.jspa?threadID=271250 if you need this.

This is really something that needs to be fixed by AWS.

RyanJarv:
Hey @ewbankkit, wondering if you ever here anything back from your AWS account team on this.

ewbankkit:
No, I never heard anything back.

lorengordon:
Any further progress on this? The approach mentioned by @JoshiiSinfield in the associated issue would work for me.

https://github.com/terraform-providers/terraform-provider-aws/issues/384#issuecomment-372276327

wobondar:
Gents @ewbankkit @RyanJarv , is there any chance to get it merged? Thanks!

aws/provider.go
```diff
@@ -418,6 +418,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_redshift_subnet_group":                    resourceAwsRedshiftSubnetGroup(),
 			"aws_route53_delegation_set":                   resourceAwsRoute53DelegationSet(),
 			"aws_route53_record":                           resourceAwsRoute53Record(),
+			"aws_route53_vpc_association_authorization":    resourceAwsRoute53VPCAssociationAuthorization(),
```
FransUrbo: As mentioned elsewhere, this should probably be `aws_route53_zone_association_authorization`. Even better, would probably be `aws_route53_zone_association_accepter` (depending on what it actually do - I'm not a fan of Go! :).

This because we already have:

```
aws_vpc_peering_connection
aws_vpc_peering_connection_accepter
```

So having the r53 resources being named
```
aws_route53_zone_association
aws_route53_zone_association_accepter
```

makes sense. Pretty if nothing else :D

website/docs/r/route53_zone_authorization_association.html.markdown
```diff
@@ -0,0 +1,62 @@
...
+* `vpc_region` - (Optional) The VPC's region. Defaults to the region of the AWS provider.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The calculated unique identifier for the association.
+* `zone_id` - The ID of the hosted zone for the association.
+* `vpc_id` - The ID of the VPC for the association.
+* `vpc_region` - The region in which the VPC identified by `vpc_id` was created.
```
FransUrbo: Should this possibly be `vpc_id` and something like `peer_zone_id` (much like `aws_vpc_peering_connection` does it)?

zdoherty: In the case of `aws_vpc_peering_connection`, the `peer` prefix is needed to disambiguate between the two VPC IDs in the arguments. Since there's only one `zone_id` in these arguments, I feel like just `zone_id` would be sufficient.

aws/provider.go
zdoherty: `aws_route53_vpc_association_authorization` would match the name of the API call that creates it, though:

https://docs.aws.amazon.com/Route53/latest/APIReference/API_CreateVPCAssociationAuthorization.html

nckbnv:
I've just come across this one while trying to do cross-account authorization myself and after I tried every possible workaround that I can think of (and I was able to find across other github issues).

It's strange that this one is sitting for more than a year now, while as it looks like there is a working PR.

jkroepke:
LGTM or not?

lucas044:
@FransUrbo Is this far off being merged?

freimer:
The AWS docs basically have you do a three-step process:

1. create-vpc-association-authorization
2. associate-vpc-with-hosted-zone
3. delete-vpc-association-authorization

Reference [AWS Docs](https://aws.amazon.com/premiumsupport/knowledge-center/private-hosted-zone-different-account/)
While overall the association is not atomic for cross-account associations, as it has three steps, from a TF perspective I'm not sure why we wouldn't just have a single resource that makes the association.  The fact that the provide might make three AWS API calls is besides the point.  Just because other resources expose the authorization doesn't mean that was the best choice or that we have to here.  We can embed the logic within the single resource to handle linking within a single account and across accounts, by providing the alternate provider name for the other account.

I recommend not merging this, and updating [aws_route53_zone_association](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_route53_zone_association.go) to accept an optional vpc_provider attribute to specify the TF provider name for the VPC if it differs from the provider for the hosted zone.  It would then do the authorization before the association, and delete the authorization after the association as per AWS recommended best practices.  Basically, I think aws_route53_zone_association is broken, and should handle the use-case of linking between accounts if required.  This resource should not be needed.

bevanbennett:
I agree with freimer, especially because as currently implemented you need a different provider to make the change from the one that can check whether it needs changing.

Adding the second provider optionally but explicitly would allow this resource to do its job properly.

rfarro:
Is there any movement on this? We have been needing to create the authorizations with aws cli commands. Here is an example on how we accomplish this with our transit account. If we could get this piece in terraform it would solve a world of problems with our Route53 module
```
resource "null_resource" "route53_association" {

  provisioner "local-exec" {
    command = "aws route53 create-vpc-association-authorization --profile ${var.account_alias} --hosted-zone-id ${element(aws_route53_zone.private_hosted_zone.*.id, count.index)} --vpc VPCRegion=${data.terraform_remote_state.transit_peer_east.region},VPCId=${data.terraform_remote_state.transit_peer_east.vpc_id}"
  }

  provisioner "local-exec" {
    command = "aws route53 associate-vpc-with-hosted-zone --profile naic-transit --hosted-zone-id ${element(aws_route53_zone.private_hosted_zone.*.id, count.index)} --vpc VPCRegion=${data.terraform_remote_state.transit_peer_east.region},VPCId=${data.terraform_remote_state.transit_peer_east.vpc_id}"
  }
}
```

feraudet:
Any chance to see aws_route53_vpc_association_authorization merged soon ?

nunofernandes:
I'm also blocked in this.. Any ideas/timeline on when it will be merged?

hdpoe:
Just throwing out there that we my organization is also facing a block because of this. I also feel that the Terraform docs should be updated to indicated that cross account zone association is not currently possible as the current wording led to me spending no small amount of time trying various configurations of provider, vpc_id and hosted zone_id.

bluemalkin:
It's pretty obvious that it's not possible to do cross aws account resources without manual approval. Vpc peering requires manual approval between accounts.

bevanbennett:
> It's pretty obvious that it's not possible to do cross aws account resources without manual approval. Vpc peering requires manual approval between accounts.

This is provably untrue. The only issue is that the request and approval need to be executed with different credentials. There is zero requirement for "manual" anything and this can be automated today with the CLI. The solution here would be to either split this resource into two pieces (association_authorization and association) or to allow us to specify a separate provider to perform the authorization than the one being used for the association. Splitting it into two resources is probably cleanest.

maciejdrozdzowski:
> > It's pretty obvious that it's not possible to do cross aws account resources without manual approval. Vpc peering requires manual approval between accounts.
> 
> This is provably untrue. The only issue is that the request and approval need to be executed with different credentials. There is zero requirement for "manual" anything and this can be automated today with the CLI. The solution here would be to either split this resource into two pieces (association_authorization and association) or to allow us to specify a separate provider to perform the authorization than the one being used for the association. Splitting it into two resources is probably cleanest.

That is correct, peering can be done with Terraform without any manual intervention.

bluemalkin:
@maciejdrozdzowski peering can be done with Terraform within the same AWS account yes, not with a different account, you'd still need to accept that separately with another terraform state or manually.

maciejdrozdzowski:
@bluemalkin that's how it was in my case - it was a cross-account peering, with different state files. The fact that it needed no manual steps was a pleasant surprise.

lorengordon:
@bluemalkin You can indeed accept the VPC peering request in the same terraform state, even between accounts. You just need to use multiple providers, one with a credential to the requester, and one with a credential to the accepter. 

```
provider "aws" {}

provider "aws" {
  alias = "peer"
}

resource "aws_vpc_peering_connection" "this" {
  peer_owner_id = "${var.peer_owner_id}"
  peer_vpc_id   = "${var.peer_vpc_id}"
  vpc_id        = "${var.vpc_id}"
}

resource "aws_vpc_peering_connection_accepter" "this" {
  provider = "aws.peer"

  vpc_peering_connection_id = "${aws_vpc_peering_connection.this.id}"
  auto_accept               = true
}
```

hdpoe:
I apologize that my comment was unclear. There is currently not a way to use native terraform aws resources to associate a vpc in one account with the private hosted zone of a vpc in another account. I'll make an update to clarify what was intended to be communicated.

rfarro:
I just want to throw out that we are testing getting away from hosting our own DHCP options set and using unbound as our DNS traffic cop. AWS has recently released route 53 resolver https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resolver-getting-started.html which allows you to create outbound/inbound endpoint so that you can get answers from both onprim server and aws servers.

ewbankkit:
* https://aws.amazon.com/premiumsupport/knowledge-center/private-hosted-zone-different-account/

aeschright:
Hi @RyanJarv ! Do you have a minute to resolve the merge conflict? Unfortunately the PR has been sitting for a while. We also need the output from your test run, which you can add to the PR description like so:

```

...
```

RyanJarv:
Want to note there is a lot of comments here that are distractions from the real issue. This is likely because discussion about it got split up between several issue making it fairly confusing to follow.

About to look into this again to see if anything has changed. But, from my memory (which granted is probably poor since this issue is fairly old now), the issue is mainly that the API around these resources doesn't map well to how terraform manages resources...

*It is certainly possible to hack something together, I simply chose not to do that here especially considering this seems to be an upstream issue IMO*

Here's a bit more info on the API issue (posted this in another issue prior): https://blog.ryanjarv.sh/2019/05/24/backdooring-route53-with-cross-account-dns.html

Granted I'm not sure if this is still the case anymore. Will update this comment shortly.

aeschright:
@RyanJarv We've been trying to sort out whether the API will map to how Terraform works too, so I'd be interested in any thoughts you might have on that given the [cli workflow](https://aws.amazon.com/premiumsupport/knowledge-center/private-hosted-zone-different-account/) linked above.

RyanJarv:
@aeschright This seems fine to me, with the exception of using —profile instead of connecting to nodes in different accounts this is how I’ve done it in the past.

Also saw this today which may be helpful https://medium.com/@dalethestirling/managing-route53-cross-account-zone-associations-with-terraform-e1e45de8f3ea.

aeschright:
After internal discussion we've decided not to go forward with the PR, for the reasons described above. There doesn't seem to be a way to do this through the API that will be effective within Terraform's design -- we're especially thinking about the need to delete the authorization afterward. I'm going to make sure the cli info is visible on the issue ticket. Thanks so much for your work on this!

RyanJarv:
Thanks @aeschright, that makes sense to me.

I cleaned up my post around the API issue a bit (https://blog.ryanjarv.sh/2019/05/24/backdooring-route53-with-cross-account-dns.html), so if anyone is concerned about this I recommend mentioning it to AWS support. There is an internal ticket on it somewhere, but it's been about 2 years without any action as far as I know.

Issue #2010, opened 2017-10-23
Support VPC Option for Elasticsearch


atsushi-ishibashi:
https://aws.amazon.com/jp/blogs/aws/amazon-elasticsearch-service-now-supports-vpc/

So far, aws-sdk-go's api reference didn't correctly update but v1.12.12 adds support for VPC access to Amazon Elasticsearch Service.

atsushi-ishibashi:
duplicate #1949

PR #2013, opened 2017-10-23
Support VPC Options for Elasticsearch

atsushi-ishibashi:
#2010

### DIFF
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"regexp"
@@ -147,6 +148,7 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 			"vpc_options": {
 				Type:     schema.TypeList,
 				Optional: true,
+				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"subnet_ids": &schema.Schema{
@@ -253,16 +255,11 @@ func resourceAwsElasticSearchDomainCreate(d *schema.ResourceData, meta interface
 
 	if v, ok := d.GetOk("vpc_options"); ok {
 		options := v.([]interface{})
-
-		if len(options) > 1 {
-			return fmt.Errorf("Only a single vpc_options block is expected")
-		} else if len(options) == 1 {
-			if options[0] == nil {
-				return fmt.Errorf("At least one field is expected inside vpc_options")
-			}
-			m := options[0].(map[string]interface{})
-			input.VPCOptions = expandESVPCOptions(m)
+		option, ok := options[0].(map[string]interface{})
+		if !ok {
+			return errors.New("vpc_options is <nil>")
 		}
+		input.VPCOptions = expandESVPCOptions(option)
 	}
 
 	log.Printf("[DEBUG] Creating ElasticSearch domain: %s", input)
@@ -472,16 +469,11 @@ func resourceAwsElasticSearchDomainUpdate(d *schema.ResourceData, meta interface
 
 	if d.HasChange("vpc_options") {
 		options := d.Get("vpc_options").([]interface{})
-
-		if len(options) > 1 {
-			return fmt.Errorf("Only a single vpc_options block is expected")
-		} else if len(options) == 1 {
-			if options[0] == nil {
-				return fmt.Errorf("At least one field is expected inside vpc_options")
-			}
-			m := options[0].(map[string]interface{})
-			input.VPCOptions = expandESVPCOptions(m)
+		option, ok := options[0].(map[string]interface{})
+		if !ok {
+			return errors.New("vpc_options is <nil>")
 		}
+		input.VPCOptions = expandESVPCOptions(option)
 	}
 
 	_, err := conn.UpdateElasticsearchDomainConfig(&input)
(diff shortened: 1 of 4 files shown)


atsushi-ishibashi:
In my environment, testacc finished with timeout error...
Acctualy, it took a lot but according to GUI console the domain was completely created and error occured while it was deleting...
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticSearchDomain_VPCOptions'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticSearchDomain_VPCOptions -timeout 120m
=== RUN   TestAccAWSElasticSearchDomain_VPCOptions
--- FAIL: TestAccAWSElasticSearchDomain_VPCOptions (3954.14s)
		
		* aws_elasticsearch_domain.example: 1 error(s) occurred:
		
		* aws_elasticsearch_domain.example: "arn:aws:es:us-east-1:hogehoge:domain/tf-test-4577410669025848591": Timeout while waiting for the domain to be created
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	3954.203s
make: *** [testacc] Error 1
```

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -143,6 +144,26 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 				Default:  "1.5",
 				ForceNew: true,
 			},
+			"vpc_options": {
+				Type:     schema.TypeList,
+				Optional: true,
```
Ninir: This implementation implies that we can have more than a single vpc_options section.
What do you think of adding a `MaxItems` option set to 1, so that we can only define a single VPC configuration?

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -230,6 +251,20 @@ func resourceAwsElasticSearchDomainCreate(d *schema.ResourceData, meta interface
 		}
 	}
 
+	if v, ok := d.GetOk("vpc_options"); ok {
+		options := v.([]interface{})
+
+		if len(options) > 1 {
+			return fmt.Errorf("Only a single vpc_options block is expected")
+		} else if len(options) == 1 {
```
Ninir: The else is not needed since the first if returns, meaning we can simplify this to:

```go
if len(options) > 1 {
    return fmt.Errorf("Only a single vpc_options block is expected")
}

if len(options) == 1 {
...
}
```

However, I do think we can simplify this a lot. 
Since only a single `vpc_options` section should be configurable as far as I understand the API, checking more than 2 is unnecessary, hence the `MaxItems` option.

Then, we can also add this line 256:

```go
option, ok := options[0].(map[string]interface{})

if !ok {
    return errors.New("vpc_options is <nil>")
}

if option != nil {
    //...
}
```

The same kind of logic applied to the update section would then be required. What do you think? 😄

website/docs/r/elasticsearch_domain.html.markdown
```diff
@@ -85,6 +86,10 @@ The following arguments are supported:
 * `automated_snapshot_start_hour` - (Required) Hour during which the service takes an automated daily
 	snapshot of the indices in the domain.
 
+**vpc_options** supports the following attribute:
```
Ninir: Nitpick: missing s to attribute

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -349,6 +395,52 @@ resource "aws_elasticsearch_domain" "example" {
...
+	}
+}
+
+resource "aws_elasticsearch_domain" "example" {
+  domain_name = "tf-test-%d"
+  ebs_options {
+    ebs_enabled = true
+    volume_size = 10
+  }
+	vpc_options {
```
Ninir: Could you fix the indentation here so that it matches the rest of the test configuration?

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -349,6 +395,52 @@ resource "aws_elasticsearch_domain" "example" {
...
+	cidr_block = "10.10.0.0/16"
+}
+
+resource "aws_subnet" "foo" {
+	cidr_block = "10.10.1.0/24"
+	vpc_id = "${aws_vpc.foo.id}"
+}
+
+resource "aws_security_group" "foo" {
+  name = "terraform_acceptance_test_example"
```
Ninir: Can you add randomization to this name also? it will help avoiding conflicts when running acceptance tests on our side 😄

aws/resource_aws_elasticsearch_domain.go
atsushi-ishibashi: I agree with all reviews:bow:
And after increasing timeout, I'll run testacc and report the result. Of course, I won't add the 
 timeout change.

Ninir:
@atsushi-ishibashi I'm seing that there was another work for VPC options opened earlier: https://github.com/terraform-providers/terraform-provider-aws/pull/1958

Perhaps you could help & contribute on that? :)

atsushi-ishibashi:
@Ninir Oh, I missed it... Sure👍
Btw, when creating ES domain with vpc options, `DescribeElasticsearchDomainOutput. DomainStatus.Endpoint` seems to be nil so `waitForElasticSearchDomainCreation` couldn't return nil..

radeksimko:
#1958 was just merged and will be part of the next release.

Thanks for the spending the time & investing the effort into this PR though.

darrenhaken:
Any idea when this will go live?

s7anley:
@darrenhaken `v1.2.0` was released about 2 hours ago.

PR #2021, opened 2017-10-23
Force new iam user if name changes

kushmansingh:


Ninir:
Hi @kushmansingh 

Thank you for opening this issue.

From an API point of view, it is totally valid to update the name, and the code already does that. Here is the documentation excerpt:

> You should understand the implications of changing an IAM user's path or
> name. For more information, see Renaming an IAM User (http://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_manage.html#id_users_renaming)
> and Renaming an IAM Group (http://docs.aws.amazon.com/IAM/latest/UserGuide/id_groups_manage_rename.html)
> in the IAM User Guide.
> 
> To change a user name the requester must have appropriate permissions on
> both the source object and the target object. For example, to change Bob
> to Robert, the entity making the request must have permission on Bob and
> Robert, or must have permission on all (*). For more information about permissions,
> see Permissions and Policies (http://docs.aws.amazon.com/IAM/latest/UserGuide/PermissionsAndPolicies.html).

Could you provide more details about why this should be a new resource?

kushmansingh:
You're right, I didn't realize users were not uniquely identified by the name. This is the use case I have https://github.com/terraform-providers/terraform-provider-aws/issues/1318 where I am initializing iam_users from a list but deletions and additions in the same step don't work since it tries to rename instead of deleting and adding. I think the work around for this is to split the operations into steps; apply all the deletes first then apply all the additions.
I would like to get some feedback on how this scenario could be handled though.

calebcase:
It seems like a distinction needs to be drawn between the terraform `name` which should be used to uniquely identify a logic user and the AWS `username` which could be modified. Otherwise I'm not sure how one could have a list of users in terraform that worked properly.

kushmansingh:
Looks like this is a more generalized problem where any list based resources are hard to manage due to the uniqueness by count rather than element names.
https://github.com/hashicorp/terraform/issues/14275

radeksimko:
Uniqueness of resources in general can be addressed using the `lifecycle` block which will reverse the ordering of operations (originally destroy -> create), e.g.:

```hcl
lifecycle {
  create_before_destroy = true
}
```
https://www.terraform.io/docs/configuration/resources.html#create_before_destroy

I believe Terraform is working as intended here and we shouldn't be forcing the recreation when the API allows updates. For that reason I'm going to close this PR.

Thanks for contribution.

PR #2029, opened 2017-10-24
Add support for SES MAIL FROM

mintuhouse:
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/mail-from-set.html

Would be the last missing piece after https://github.com/terraform-providers/terraform-provider-aws/issues/1347

### DIFF
--- a/aws/resource_aws_ses_domain_mail_from.go
+++ b/aws/resource_aws_ses_domain_mail_from.go
@@ -11,8 +11,9 @@ import (
 
 func resourceAwsSesDomainMailFrom() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceAwsSesDomainMailFromCreate,
+		Create: resourceAwsSesDomainMailFromSet,
 		Read:   resourceAwsSesDomainMailFromRead,
+		Update: resourceAwsSesDomainMailFromSet,
 		Delete: resourceAwsSesDomainMailFromDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
@@ -27,25 +28,30 @@ func resourceAwsSesDomainMailFrom() *schema.Resource {
 			"mail_from_domain": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
+			},
+			"behavior_on_mx_failure": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  ses.BehaviorOnMXFailureUseDefaultValue,
 			},
 		},
 	}
 }
 
-func resourceAwsSesDomainMailFromCreate(d *schema.ResourceData, meta interface{}) error {
+func resourceAwsSesDomainMailFromSet(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).sesConn
 
+	behaviorOnMxFailure := d.Get("behavior_on_mx_failure").(string)
 	domainName := d.Get("domain").(string)
 	mailFromDomain := d.Get("mail_from_domain").(string)
 
-	createOpts := &ses.SetIdentityMailFromDomainInput{
-		BehaviorOnMXFailure: aws.String("UseDefaultValue"),
+	input := &ses.SetIdentityMailFromDomainInput{
+		BehaviorOnMXFailure: aws.String(behaviorOnMxFailure),
 		Identity:            aws.String(domainName),
 		MailFromDomain:      aws.String(mailFromDomain),
 	}
 
-	_, err := conn.SetIdentityMailFromDomain(createOpts)
+	_, err := conn.SetIdentityMailFromDomain(input)
 	if err != nil {
 		return fmt.Errorf("Error setting MAIL FROM domain: %s", err)
 	}
@@ -59,7 +65,6 @@ func resourceAwsSesDomainMailFromRead(d *schema.ResourceData, meta interface{}) 
 	conn := meta.(*AWSClient).sesConn
 
 	domainName := d.Id()
-	d.Set("domain", domainName)
 
 	readOpts := &ses.GetIdentityMailFromDomainAttributesInput{
 		Identities: []*string{
@@ -67,10 +72,20 @@ func resourceAwsSesDomainMailFromRead(d *schema.ResourceData, meta interface{}) 
 		},
 	}
 
-	_, err := conn.GetIdentityMailFromDomainAttributes(readOpts)
+	out, err := conn.GetIdentityMailFromDomainAttributes(readOpts)
 	if err != nil {
-		log.Printf("[WARN] Error fetching MAIL FROM domain attributes for %s: %s", d.Id(), err)
+		log.Printf("error fetching MAIL FROM domain attributes for %s: %s", domainName, err)
 		return err
+	}
+
+	d.Set("domain", domainName)
+
+	if v, ok := out.MailFromDomainAttributes[domainName]; ok {
+		d.Set("behavior_on_mx_failure", v.BehaviorOnMXFailure)
+		d.Set("mail_from_domain", v.MailFromDomain)
+	} else {
+		d.Set("behavior_on_mx_failure", v.BehaviorOnMXFailure)
+		d.Set("mail_from_domain", "")
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


petergoldstein:
@Ninir @radeksimko Is there anything that needs to be done to push this across the finish line?

madmod:
If this needs anything else done I'm willing to take it on.

bflad:
I'll be taking a look at this later today.

aws/resource_aws_ses_domain_mail_from.go
```diff
@@ -0,0 +1,96 @@
...
+		Schema: map[string]*schema.Schema{
+			"domain": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"mail_from_domain": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
bflad: We should not need `ForceNew: true` here since be able to support updates by calling `SetIdentityMailFromDomain` just like create. 😄

aws/resource_aws_ses_domain_mail_from.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func resourceAwsSesDomainMailFromCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	domainName := d.Get("domain").(string)
+	mailFromDomain := d.Get("mail_from_domain").(string)
+
+	createOpts := &ses.SetIdentityMailFromDomainInput{
+		BehaviorOnMXFailure: aws.String("UseDefaultValue"),
```
bflad: Two things here:
* Nitpick: SDK has a constant available `ses.BehaviorOnMXFailureUseDefaultValue`
* Seems easy enough to support this as an optional attribute, that defaults to this

aws/resource_aws_ses_domain_mail_from.go
```diff
@@ -0,0 +1,96 @@
...
+			aws.String(domainName),
+		},
+	}
+
+	_, err := conn.GetIdentityMailFromDomainAttributes(readOpts)
+	if err != nil {
+		log.Printf("[WARN] Error fetching MAIL FROM domain attributes for %s: %s", d.Id(), err)
+		return err
+	}
+
```
bflad: We should really be resetting the `mail_from_domain` attribute in the Terraform state each read to catch drift from the Terraform configuration, e.g.

```go
out, err := conn.GetIdentityMailFromDomainAttributes(readOpts)
if err != nil {
	log.Printf("[WARN] Error fetching MAIL FROM domain attributes for %s: %s", d.Id(), err)
	return err
}

if v, ok := out.MailFromDomainAttributes[domainName]; ok {
	d.Set("mail_from_domain", v.MailFromDomain)
} else {
	d.Set("mail_from_domain", "")
}
```

aws/resource_aws_ses_domain_mail_from.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+func resourceAwsSesDomainMailFromDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	domainName := d.Get("domain").(string)
+
+	deleteOpts := &ses.SetIdentityMailFromDomainInput{
+		BehaviorOnMXFailure: aws.String("UseDefaultValue"),
```
bflad: Nitpick: this parameter is not required here

aws/resource_aws_ses_domain_mail_from.go
```diff
@@ -0,0 +1,96 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func resourceAwsSesDomainMailFromDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	domainName := d.Get("domain").(string)
```
bflad: Nitpick: We should use `d.Id()`

aws/resource_aws_ses_domain_mail_from_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAwsSESDomainMailFrom_basic(t *testing.T) {
+	domain := fmt.Sprintf(
+		"%s.terraformtesting.com",
+		acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+
+	resource.Test(t, resource.TestCase{
```
bflad: Can you please implement a `CheckDestroy:` function for the `TestCase`? This is one final check in our acceptance tests to verify that the delete/destroy actually did what it was supposed to in AWS. e.g.

```go
func testAccCheckSESDomainMailFromDestroy(s *terraform.State) error {
	conn := testAccProvider.Meta().(*AWSClient).sesConn

	for _, rs := range s.RootModule().Resources {
		if rs.Type != "aws_ses_domain_mail_from" {
			continue
		}

		input := &ses.GetIdentityMailFromDomainAttributesInput{
			Identities: []*string{aws.String(rs.Primary.ID)},
		}

		out, err := conn.GetIdentityMailFromDomainAttributes(input)
		if err != nil {
			return fmt.Errorf("error fetching MAIL FROM domain attributes: %s", err)
		}
		if v, ok := out.MailFromDomainAttributes[domainName]; ok && len(v.MailFromDomain) > 0 {
			return fmt.Errorf("MAIL FROM domain was not removed, found: %s", v.MailFromDomain)
		}
	}

	return nil
}
```

website/docs/r/ses_domain_mail_from.html.markdown
```diff
@@ -0,0 +1,67 @@
+---
+layout: "aws"
+page_title: "AWS: ses_domain_mail_from"
+sidebar_current: "docs-aws-resource-ses-domain-mail-from"
+description: |-
+  Provides an SES domain MAIL FROM resource
+---
+
+# aws\_ses\_domain\_dkim
```
bflad: Copypasta 🍝 : should be `aws_ses_domain_mail_from` (backslashes are no longer required in the documentation)

website/docs/r/ses_domain_mail_from.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+
+## Followup Steps
+
+Find out more about setting MAIL FROM domain with Amazon SES in [docs](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/mail-from-set.html).
+
+* Create an MX record used to verify SES MAIL FROM setup
+
+* (Optionally) If you want your emails to pass SPF checks, you must publish an SPF record to the DNS server of the custom MAIL FROM domain.
+
+## Example Usage
```
bflad: Nitpick: We generally keep the example usage above arguments.

website/docs/r/ses_domain_mail_from.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+  domain = "example.com"
+}
+
+resource "aws_ses_domain_mail_from" "example" {
+  domain           = "example.com"
+  mail_from_domain = "bounce.example.com"
+}
+
+resource "aws_route53_record" "example_amazonses_mail_from_mx_record" {
+  zone_id = "ABCDEFGHIJ123" # Change to appropriate Route53 Zone ID
```
bflad: Feel free to point these at a "fake" resource to encourage proper references, e.g. `${aws_route53_zone.example.id}`

website/docs/r/ses_domain_mail_from.html.markdown
```diff
@@ -0,0 +1,67 @@
...
+  name    = "bounce.example.com"
```
bflad: Nitpick: I think we should cross-reference either the `${aws_ses_domain_mail_from.example.mail_from_domain}` here or `${aws_route53_record. example_amazonses_mail_from_mx_record.name}` in `aws_ses_domain_mail_from` to better show the linkage between the two

aws/resource_aws_ses_domain_mail_from_test.go
```diff
@@ -0,0 +1,76 @@
...
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(testAccAwsSESDomainMailFromConfig, domain),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsSESDomainMailFromExists("aws_ses_domain_mail_from.test"),
```
bflad: Sorry I forgot to mention in the initial review -- we should verify the attributes are set correctly in the Terraform state as well 👍  e.g.
```go
resource.TestCheckResourceAttr("aws_ses_domain_mail_from.test", "domain", domain),
resource.TestCheckResourceAttr("aws_ses_domain_mail_from.test", "mail_from_domain", fmt.Sprintf("bounce.%s", domain)),
```

mintuhouse:
@bflad The project where I originally used this code is no longer active and changed PC since then. 
I will have to refresh my go and terraform (which was super easy to pick up by the way :+1:) for the changes. Will someone else (@madmod, @bflad) be able to take this up?

bflad:
Not a problem, I can add a review commit to yours and get this in for you. Thanks so much for your initial work here, very appreciated!

bflad:

PR #2033, opened 2017-10-24
resource/vpc_dhcp_options: Updated documentation about how to maintain default AWS DNS servers

### LINKED ISSUE #2031: Documentation: dhcp_options_set should explain how to maintain default AWS NS
The documentation here doesn't mention anything about how to keep the default AWS NS: https://www.terraform.io/docs/providers/aws/r/vpc_dhcp_options.html

More specifically, when you set DHCP options, if you're not specifying external/own DNS NS, then you would (probably) want to default to the standard AWS NS. This should probably be the default behavior, but it isn't. What actually happens when you don't specify `domain_name_servers` is that resolution breaks (or rather you have to manually configure it to make it work by editing `/etc/resolv.conf` to point to default AWS NS or such). That's fine, but it should at least be mentioned in the docs.

For more on this: http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_DHCP_Options.html#AmazonDNS

I'll make a PR to include a note and warning about this behavior in the docs.

alex1x:
…ult AWS DNS

Fixes #2031

### DIFF
--- a/website/docs/r/vpc_dhcp_options.html.markdown
+++ b/website/docs/r/vpc_dhcp_options.html.markdown
@@ -41,7 +41,7 @@ resource "aws_vpc_dhcp_options" "foo" {
 The following arguments are supported:
 
 * `domain_name` - (Optional) the suffix domain name to use by default when resolving non Fully Qualified Domain Names. In other words, this is what ends up being the `search` value in the `/etc/resolv.conf` file.
-* `domain_name_servers` - (Optional) List of name servers to configure in `/etc/resolv.conf`.
+* `domain_name_servers` - (Optional) List of name servers to configure in `/etc/resolv.conf`. If you want to use the default AWS nameservers you should set this to `AmazonProvidedDNS`.
 * `ntp_servers` - (Optional) List of NTP servers to configure.
 * `netbios_name_servers` - (Optional) List of NETBIOS name servers.
 * `netbios_node_type` - (Optional) The NetBIOS node type (1, 2, 4, or 8). AWS recommends to specify 2 since broadcast and multicast are not supported in their network. For more information about these node types, see [RFC 2132](http://www.ietf.org/rfc/rfc2132.txt).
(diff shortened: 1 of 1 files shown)


website/docs/r/vpc_dhcp_options.html.markdown
```diff
@@ -41,7 +41,7 @@ resource "aws_vpc_dhcp_options" "foo" {
 The following arguments are supported:
 
 * `domain_name` - (Optional) the suffix domain name to use by default when resolving non Fully Qualified Domain Names. In other words, this is what ends up being the `search` value in the `/etc/resolv.conf` file.
-* `domain_name_servers` - (Optional) List of name servers to configure in `/etc/resolv.conf`.
+* `domain_name_servers` - (Optional) List of name servers to configure in `/etc/resolv.conf`. Note that if you leave this empty, DNS resolution will break. If you want to use the default AWS nameservers you should set this to `AmazonProvidedDNS`.
```
Ninir: Can we change this to a more positive message, just keeping the second sentence?

Also, if this is empty, is the DNS resolution really broken or just non-functional?

alex1x: Sure- made the requested change. As mentioned in #2031 it breaks in the sense that your instances are provisioned without nameservers so it doesn't work. If you manually set them to use nameservers after the fact then they will work, but by default your DNS won't work.

Ninir: Understood: reachable by API but not by DNS name, that's right?

alex1x: More specifically: if you attempt to do something like `ping my-instance.example.com` you will get "Unable to resolve". Then you would look at `/etc/resolv.conf` and you would need to add the default AWS NS to it, for example: `nameserver 10.0.150.2` (if you are on 10.0.150.0/24). Now DNS works.

By setting `AmazonProvidedDNS` in your terraform, you will not get this "Unable to resolve" error and you will not need to manually add the nameserver IP to your `resolv.conf`, it will already be there (AWS will figure out what the value needs to be depending on your VPC and pre-populate it for you, thus DNS works by default for each new instance).

Ninir: OK so it's all ok then :) thank you @alex1x ! 🚀

PR #2042, opened 2017-10-25
Add 'tags' support to aws_spot_fleet_request

### LINKED ISSUE #1232: add support tags for aws_spot_fleet_request

jekh:
This adds support for [tags on spot fleet requests](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotFleetTagSpecification.html). I tried to follow the pattern of the nearly-identical `tags` block in `resource_aws_instance.go` (the difference being Spot Fleet only supports `instance`, not `volume`, as a resource type).

Since the documentation for `aws_spot_fleet_request` simply references the `aws_instance` documentation, I didn't add any additional info about `tags`, but I did include a `tags` block in one of the examples.

This should resolve feature request #1232.

Closes #1232

### DIFF
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -382,10 +382,10 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
 		}
 	}
 
-	if v, ok := d["tags"]; ok {
+	if m, ok := d["tags"].(map[string]interface{}); ok && len(m) > 0 {
 		tagsSpec := make([]*ec2.SpotFleetTagSpecification, 0)
 
-		tags := tagsFromMap(v.(map[string]interface{}))
+		tags := tagsFromMap(m)
 
 		spec := &ec2.SpotFleetTagSpecification{
 			ResourceType: aws.String("instance"),
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -256,6 +256,11 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
 							Computed: true,
 							ForceNew: true,
 						},
+						"tags": {
+							Type:     schema.TypeMap,
+							Optional: true,
+							ForceNew: true,
+						},
 					},
 				},
 				Set: hashLaunchSpecification,
@@ -375,6 +380,21 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
 				securityGroupIds = append(securityGroupIds, aws.String(v.(string)))
 			}
 		}
+	}
+
+	if v, ok := d["tags"]; ok {
+		tagsSpec := make([]*ec2.SpotFleetTagSpecification, 0)
+
+		tags := tagsFromMap(v.(map[string]interface{}))
+
+		spec := &ec2.SpotFleetTagSpecification{
+			ResourceType: aws.String("instance"),
+			Tags:         tags,
+		}
+
+		tagsSpec = append(tagsSpec, spec)
+
+		opts.TagSpecifications = tagsSpec
 	}
 
 	subnetId, hasSubnetId := d["subnet_id"]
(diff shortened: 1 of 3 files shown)


aws/resource_aws_spot_fleet_request.go
```diff
@@ -377,6 +382,21 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
 		}
 	}
 
+	if v, ok := d["tags"]; ok {
```
radeksimko: Do you mind adding an extra check here? i.e.

```go
if m, ok := d["tags"].(map[string]interface{}); ok && len(m) > 0 {
```
otherwise this would break all configs without tags, see output from relevant acceptance tests:

```
=== RUN   TestAccAWSSpotFleetRequest_associatePublicIpAddress
--- FAIL: TestAccAWSSpotFleetRequest_associatePublicIpAddress (682.03s)

		* aws_spot_fleet_request.foo: 1 error(s) occurred:

		* aws_spot_fleet_request.foo: Error requesting spot fleet: [WARN] Error creating Spot fleet request, retrying: InvalidSpotFleetRequestConfig: Tag value cannot be null. Use empty string instead.
			status code: 400, request id: 42b4061d-3214-4bae-9fba-0751e13458ab
```

jekh:
Done! Thanks, for some reason this is one of the test cases I didn't test. That's what the acceptance tests are for, I suppose. Please let me know if there are any other issues you see with the change.

aaroncline:
This was supposed to be in 1.2 according to the changelog, but I am unable to successfully apply tags to a spot fleet request.  I do not get an error and the spot fleet is created, but I do not get any tags.   I have placed the tag block in the lauch_specification block.

jedi4ever:
@radeksimko in which TF version will this be available?

bflad:
This was previously released in [terraform-provider-aws version 1.2.0](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#120-october-31-2017) and has been available in all releases since. Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

cemo:
You could also search the issue number 2042 in [CHANGELOG].

[CHANGELOG]: https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md

yogeek:
Sorry if I misuderstood but it seems to me that "tags" are possible only in the `launch_specification{...}` part of the `aws_spot_fleet_request ` resource and not in the resource itself. 

Indeed, if I apply the example in the documentation by using a tag like this :
```
resource "aws_spot_fleet_request" "cheap_compute" {
  iam_fleet_role          = "${var.arn}"
  allocation_strategy     = "lowestPrice"
  target_capacity         = 1

  launch_specification {
    ami                       = "${var.ami}"
    instance_type             = "${var.instance_type}"
    key_name                  = "${var.key_name}"
    vpc_security_group_ids    = "${var.vpc_security_group_ids}"
    subnet_id                 = "${var.subnet_id}"

    tags = {
        Name = "${var.name}"
    }
  }
}
```
the tag is in fact only on the spot instance created by the request but not on the request itself (whereas its possible to add tags in AWS conole for this resource)

And if I try to add a `tag` section directly under the `aws_spot_fleet_request ` resource :
```
resource "aws_spot_fleet_request" "cheap_compute" {
  iam_fleet_role          = "${var.arn}"
  allocation_strategy     = "lowestPrice"
  target_capacity         = 1

  tags = {
        Name = "${var.name}"
    }

  launch_specification {
    ami                       = "${var.ami}"
    instance_type             = "${var.instance_type}"
    key_name                  = "${var.key_name}"
    vpc_security_group_ids    = "${var.vpc_security_group_ids}"
    subnet_id                 = "${var.subnet_id}"

    tags = {
        Name = "${var.name}"
    }
  }
}
``` 
I have the following error :
```
Error: aws_spot_fleet_request.cheap_compute: : invalid or unknown key: tags
```

I thought this PR would be to add tag on the resource directly...Did I misunderstood the PR here ?
Thank you

bgaillard:
Hi, we encounter the same problem as @yogeek.

```
Error: module.jenkins.aws_spot_fleet_request.jenkins-spot-fleet-v2: : invalid or unknown key: tags
```

I upgraded the AWS provider to version 1.42 and this still not work. 

Do you know what's the problem ?

PR #2055, opened 2017-10-25
r/kinesis_firehose_delivery_stream: Use data source instead of acc ID in tests

radeksimko:
There's no real reason for a static ID. I'm lowering the barrier this way so that anyone can `make testacc` without having find & specify the ID.


radeksimko:
Re eventual consistency - interesting, we have retry logic in place, we retry for 1 minute. We can (probably should) raise that - to say 90 seconds? or 2 minutes? We don't want to go too high as there may be genuine policy issues. Do you mind raising a PR for that?

PR #2080, opened 2017-10-27
Fix Elastic Beanstalk dependency issue causing delete failure.

gerph:
Commonly we were seeing that Elastic Beanstalk failed to destroy
before it timed out. The reason that it failed to destroy was that
the destroy operation was silently ignoring an error on the Elastic
Beanstalk Application - reporting that it could not be destroyed
because there was still an Elastic Beanstalk Application Version
attached to it.

This change fixes the Terraform code so that this error is reported
when if the delete happens.

### DIFF
--- a/aws/resource_aws_elastic_beanstalk_application.go
+++ b/aws/resource_aws_elastic_beanstalk_application.go
@@ -112,11 +112,14 @@ func resourceAwsElasticBeanstalkApplicationDelete(d *schema.ResourceData, meta i
 	_, err = beanstalkConn.DeleteApplication(&elasticbeanstalk.DeleteApplicationInput{
 		ApplicationName: aws.String(d.Id()),
 	})
+	if err != nil {
+		return err
+	}
 
 	return resource.Retry(10*time.Second, func() *resource.RetryError {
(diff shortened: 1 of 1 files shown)


charles-at-geospock:
That's correct - what was happening previously would be that it would...

* Try to destroy the elastic beanstalk application
* Get an error back to say that there was an application version still present.
* Ignore the error (this is thing fixed in this change)
* Repeatedly call 'Is the application still there?'.
* Timeout because the application is still there - the request to destroy it failed, so of course it's still there.

Now, at the 'ignore the error' we'll report the error that the application version is present and will stop without the timeout being triggered :-)

The reason we were seeing the problem at all was that we had a configuration issue in our .tf modules that meant that the application version was being referred to with a application name of "${var.application_name}", rather than "${aws_elasticsearch_application.our_app.name}" - and that meant that the two were not dependant on one another, and thus the application could be destroyed before the application version. If we'd got our dependencies right, we would not have seen this problem. It probably won't affect most people, although if they were to manually create more application versions they would get a timeout every time they tried to destroy it.

Thank you for the merge; it will make our lives a lot easier - despite the fact that we've now fixed our dependencies :-)

PR #2086, opened 2017-10-27
resource/aws_cloudwatch_log_metric_filter: fixed flattenCloudWachLogMetricTransformations return value

Ninir:
This updates the `flattenCloudWachLogMetricTransformations`  method.
Until now, the `d.Set("metric_transformation")` for this particular structure was not working as expected: the value passed was just never set, meaning you could not update one of the structure value in the console, making Terraform be aware of it at plan-time. The value stored in the state only depended on the value set by the user in the configuration file.

Required for https://github.com/terraform-providers/terraform-provider-aws/pull/1578


Ninir:
@radeksimko Indeed!

will merge then, thanks :)

PR #2090, opened 2017-10-27
Add Redis AUTH, in-transit and at-rest encryption

nathanielks:
AWS [announced](https://aws.amazon.com/blogs/security/amazon-elasticache-now-supports-encryption-for-elasticache-for-redis/) support for `AUTH`, in-transit and at-rest encryption. This aims to add that to terraform!

### DIFF
--- a/aws/resource_aws_elasticache_replication_group.go
+++ b/aws/resource_aws_elasticache_replication_group.go
@@ -87,6 +87,28 @@ func resourceAwsElasticacheReplicationGroup() *schema.Resource {
 	resourceSchema["engine"].Default = "redis"
 	resourceSchema["engine"].ValidateFunc = validateAwsElastiCacheReplicationGroupEngine
 
+	resourceSchema["at_rest_encryption"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+		ForceNew: true,
+	}
+
+	resourceSchema["transit_encryption"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+		ForceNew: true,
+	}
+
+	resourceSchema["auth_token"] = &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		Sensitive:    true,
+		ForceNew:     true,
+		ValidateFunc: validateAwsElastiCacheReplicationGroupAuthToken,
+	}
+
 	return &schema.Resource{
 		Create: resourceAwsElasticacheReplicationGroupCreate,
 		Read:   resourceAwsElasticacheReplicationGroupRead,
@@ -166,6 +188,18 @@ func resourceAwsElasticacheReplicationGroupCreate(d *schema.ResourceData, meta i
 
 	if v, ok := d.GetOk("snapshot_name"); ok {
 		params.SnapshotName = aws.String(v.(string))
+	}
+
+	if _, ok := d.GetOk("transit_encryption"); ok {
+		params.TransitEncryptionEnabled = aws.Bool(d.Get("transit_encryption").(bool))
+	}
+
+	if _, ok := d.GetOk("at_rest_encryption"); ok {
+		params.AtRestEncryptionEnabled = aws.Bool(d.Get("at_rest_encryption").(bool))
+	}
+
+	if v, ok := d.GetOk("auth_token"); ok {
+		params.AuthToken = aws.String(v.(string))
 	}
 
 	clusterMode, clusterModeOk := d.GetOk("cluster_mode")
@@ -313,6 +347,9 @@ func resourceAwsElasticacheReplicationGroupRead(d *schema.ResourceData, meta int
 		}
 
 		d.Set("auto_minor_version_upgrade", c.AutoMinorVersionUpgrade)
+		d.Set("at_rest_encryption", c.AtRestEncryptionEnabled)
+		d.Set("transit_encryption", c.TransitEncryptionEnabled)
+		d.Set("auth_token", c.TransitEncryptionEnabled)
 	}
 
 	return nil
@@ -532,6 +569,19 @@ func validateAwsElastiCacheReplicationGroupId(v interface{}, k string) (ws []str
 	if regexp.MustCompile(`-$`).MatchString(value) {
 		errors = append(errors, fmt.Errorf(
 			"%q cannot end with a hyphen", k))
+	}
+	return
+}
+
+func validateAwsElastiCacheReplicationGroupAuthToken(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	if (len(value) < 16) || (len(value) > 128) {
+		errors = append(errors, fmt.Errorf(
+			"%q must contain from 16 to 128 alphanumeric characters or symbols (excluding @, \", and /)", k))
+	}
+	if !regexp.MustCompile(`^[^@"\/]+$`).MatchString(value) {
+		errors = append(errors, fmt.Errorf(
+			"only alphanumeric characters or symbols (excluding @, \", and /) allowed in %q", k))
 	}
 	return
 }
--- a/aws/resource_aws_elasticache_replication_group.go
+++ b/aws/resource_aws_elasticache_replication_group.go
@@ -87,14 +87,14 @@ func resourceAwsElasticacheReplicationGroup() *schema.Resource {
 	resourceSchema["engine"].Default = "redis"
 	resourceSchema["engine"].ValidateFunc = validateAwsElastiCacheReplicationGroupEngine
 
-	resourceSchema["at_rest_encryption"] = &schema.Schema{
+	resourceSchema["at_rest_encryption_enabled"] = &schema.Schema{
 		Type:     schema.TypeBool,
 		Optional: true,
 		Default:  false,
 		ForceNew: true,
 	}
 
-	resourceSchema["transit_encryption"] = &schema.Schema{
+	resourceSchema["transit_encryption_enabled"] = &schema.Schema{
 		Type:     schema.TypeBool,
 		Optional: true,
 		Default:  false,
@@ -190,12 +190,12 @@ func resourceAwsElasticacheReplicationGroupCreate(d *schema.ResourceData, meta i
 		params.SnapshotName = aws.String(v.(string))
 	}
 
-	if _, ok := d.GetOk("transit_encryption"); ok {
-		params.TransitEncryptionEnabled = aws.Bool(d.Get("transit_encryption").(bool))
+	if _, ok := d.GetOk("transit_encryption_enabled"); ok {
+		params.TransitEncryptionEnabled = aws.Bool(d.Get("transit_encryption_enabled").(bool))
 	}
 
-	if _, ok := d.GetOk("at_rest_encryption"); ok {
-		params.AtRestEncryptionEnabled = aws.Bool(d.Get("at_rest_encryption").(bool))
+	if _, ok := d.GetOk("at_rest_encryption_enabled"); ok {
+		params.AtRestEncryptionEnabled = aws.Bool(d.Get("at_rest_encryption_enabled").(bool))
 	}
 
 	if v, ok := d.GetOk("auth_token"); ok {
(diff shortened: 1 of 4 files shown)


nathanielks:
This is ready for review! Questions I have for reviewers are:

- AWS has additional requirements for in-transit encryption like redis version, node types, etc. Do we need to add validation before attempting to create the resource, or just defer to the API rejecting the request? I opted for that to keep this simple.
- When should I put together a PR for docs? Now or after this has been merged?
---

This is my first code contribution and it's been a _great_ experience. Thank you for writing such good software! The testing helpers and clear patterns made contributing a breeze. I hadn't ever really looked at source till last night and I've got a PR put together today!! Thanks again!

ewbankkit:
@nathanielks Thanks for tackling this and glad you found contributing so straightforward.

You should also update the corresponding [documentation page](https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/elasticache_replication_group.html.markdown) and since https://github.com/terraform-providers/terraform-provider-aws/pull/2088 was merged, you can undo the changes you made in the `vendor` tree and `git rebase` from the upstream master branch.

For completeness I suggest updating the `aws_elasticache_cluster` resource and data source for those not using replication groups.

nathanielks:
@ewbankkit hitting a bit of a blocker and not sure how to resolve! The AWS Go SDK has a few different structs it's using for Elasticache. Terraform uses the `CreateCacheClusterInput` and `CreateReplicationGroupInput` structs. Transit Encryption and At Rest Encryption are only available on the `CacheCluster`, `CreateReplicationGroupInput`, and `ReplicationGroup` structs. AuthToken is available on `CreateCacheClusterInput`, however. If the encryption options aren't available on the `Input` struct, how do we get around that?

nathanielks:
I've tested by adding just the auth token to the CreateCacheClusterInput and the API errors with the following:

```
* aws_elasticache_cluster.bar: Error creating Elasticache: InvalidParameterValue: This AWS account is not whitelisted for Redis Auth but AuthToken is provided.
	status code: 400, request id: 50ef2190-bdc7-11e7-a938-c3d192d48d
```

tavisma:
Eagerly awaiting this!

nathanielks:
Maybe @radeksimko or another core committer can recommend how to resolve this issue?

https://github.com/terraform-providers/terraform-provider-aws/pull/2090#issuecomment-340601280

nathanielks:
@ewbankkit Can I get your eyes on https://github.com/terraform-providers/terraform-provider-aws/pull/2090#issuecomment-340601280?

nathanielks:
@handlerbot 
> What do you think about splitting the AUTH work to a separate PR

Unfortunately they're tied together. AUTH depends on in-transit encryption (which practically is a good thing, but annoying at the moment!)

> you may want to engage with AWS support on the error you're getting

That's a great call! They may also be able to explain why the Go sdk is structured like it is! I'll reach out on their repo as well 😄

nathanielks:
(also, the auth-only tests were to illustrate they'll need in-transit encryption)

nathanielks:
@handlerbot Ah! Thanks for clarifying! So the problem isn't AUTH, it's 100% with the structs that the Go SDK provides for us. AUTH in this PR actually works fine, but it's only for Replication Groups as they are the only resource that exposes any encryption parameters. I'm now trying to figure out how to get Encryption set up on basic Clusters per @ewbankkit's request. You can actually get the same functionality with a replication group as you can with a cluster. Enabling encryption forces the creation of a new resource, so we're forced to create a new _something_. Whether that's a cache cluster or a replication group, I'm not sure it entirely matters. To get a replication group that behaves like a cache cluster, go ahead and use this config:

```
resource "aws_elasticache_replication_group" "bar" {
  replication_group_id          = "${var.name}"
  replication_group_description = "replication-group-example"
  node_type                     = "cache.t2.micro"
  number_cache_clusters         = "1"
  port                          = 6379
  parameter_group_name          = "default.redis3.2"
  engine_version                = "3.2.4"
}
```

To that end, @ewbankkit, is it 100% necessary to add encryption support to `aws_elasticache_cluster`, consider it's a redis-only feature and it's already supported on the replication group resource? Users will need to a create a new _something_ as explained above, so why not force them to use the replication group resource instead of the cache cluster?

nathanielks:
And more fuel in support of using replication groups, the in-transit option is only supported by redis 3.2.4 and above, which is what replication groups require...

nathanielks:
@ewbankkit looking into this further, the encryption options are only supported by Replication Groups in the API as well: https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html, so **we can't create a cache cluster with encryption enabled. It's only replication groups.**

ewbankkit:
@nathanielks Yes, my bad. The [User Guide](http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/in-transit-encryption.html#in-transit-encryption-enable) seems to indicate that `CreateCacheCluster` supports those options but the [API documentation](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateCacheCluster.html) makes no mention.

myoung34:
Would it be easier to split this up into 2 pr's, one for auth one for encryption?
I need encryption ASAP and it seems to be more straight forward vs the auth changes

nathanielks:
@ewbankkit do you know how to bump this to get this reviewed? I think I've followed the contribution guidelines and haven't seen any core committers come around.

nathanielks:
Acceptance tests are passing:
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption -timeout 120m
=== RUN   TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption
--- PASS: TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption (875.50s)
PASS
ok  	github.com/nathanielks/terraform-provider-aws/aws	875.530s

$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption -timeout 120m
=== RUN   TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption
--- PASS: TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption (1059.05s)
PASS
ok  	github.com/nathanielks/terraform-provider-aws/aws	1059.087s
```

nathanielks:
@myoung34 No need to split into two PR's. Encryption can only be added to Replication Groups and that's what this PR does.

ewbankkit:
@nathanielks I suggest patience 😄.

nathanielks:
@radeksimko this is ready for you! Acceptance tests written and passing!

sstarcher:
@radeksimko does any additional work need done to get this merged in?

ChrisMcKee:
image
image

It's on their radar

nathanielks:
Good eyes, @ChrisMcKee

aws/resource_aws_elasticache_replication_group.go
```diff
@@ -165,6 +187,18 @@ func resourceAwsElasticacheReplicationGroupCreate(d *schema.ResourceData, meta i
 		params.SnapshotName = aws.String(v.(string))
 	}
 
+	if _, ok := d.GetOk("transit_encryption"); ok {
```
radeksimko: It will sound like a nitpick, but would you mind changing the field name to `transit_encryption_enabled`. It will better communicate the fact that this is a boolean field and make space for potential future `transit_encryption` block which may define some further settings of the encryption.

aws/resource_aws_elasticache_replication_group.go
```diff
@@ -165,6 +187,18 @@ func resourceAwsElasticacheReplicationGroupCreate(d *schema.ResourceData, meta i
...
+		params.TransitEncryptionEnabled = aws.Bool(d.Get("transit_encryption").(bool))
+	}
+
+	if _, ok := d.GetOk("at_rest_encryption"); ok {
```
radeksimko: Likewise - would you mind changing it to `at_rest_encryption_enabled` for the same reasons as the above field?

aws/resource_aws_elasticache_replication_group.go
```diff
@@ -310,6 +344,9 @@ func resourceAwsElasticacheReplicationGroupRead(d *schema.ResourceData, meta int
 		}
 
 		d.Set("auto_minor_version_upgrade", c.AutoMinorVersionUpgrade)
+		d.Set("at_rest_encryption", c.AtRestEncryptionEnabled)
+		d.Set("transit_encryption", c.TransitEncryptionEnabled)
+		d.Set("auth_token", c.TransitEncryptionEnabled)
```
radeksimko: Typo ☝️ 

Because we cannot get the token back from the API (for obvious reasons) I think we should just make it `nil` if `c.AuthTokenEnabled != nil && !*c.AuthTokenEnabled` otherwise leave it alone during `Read()`.

aws/resource_aws_elasticache_replication_group.go
```diff
@@ -532,3 +569,16 @@ func validateAwsElastiCacheReplicationGroupId(v interface{}, k string) (ws []str
 	}
 	return
 }
+
+func validateAwsElastiCacheReplicationGroupAuthToken(v interface{}, k string) (ws []string, errors []error) {
```
radeksimko: Nitpick: We tend to keep our validation functions in `aws/validators.go` + tests in `aws/validators_test.go`. It would be nice to move it there too, it's not a blocker for this PR though.

aws/resource_aws_elasticache_replication_group_test.go
```diff
@@ -970,3 +1048,108 @@ resource "aws_elasticache_replication_group" "bar" {
     }
 }`, rInt, rInt, rInt, rInt, rName)
 }
+
+var testAccAWSElasticacheReplicationGroupEnableAtRestEncryptionConfig = fmt.Sprintf(`
```
radeksimko: Nitpick: It's a common convention to turn these configs into functions and pass the random strings from the body of the test as argument. It makes it easier to build & compare fields which include these random strings and also makes the config reusable (as it doesn't contain a random string generated once during initiation of the variable).

I know we don't follow that convention everywhere yet, but it's been a common theme of many recent test changes. 😉

aws/resource_aws_elasticache_replication_group.go
nathanielks: Sure thing!

aws/resource_aws_elasticache_replication_group.go
nathanielks: 👍

aws/resource_aws_elasticache_replication_group.go
nathanielks: ah, great catch!

aws/resource_aws_elasticache_replication_group.go
nathanielks: I'll move them!

nathanielks:
@radeksimko requested changes have been made, tests still passing!

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption -timeout 120m
=== RUN   TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption
--- PASS: TestAccAWSElasticacheReplicationGroup_enableAuthTokenTransitEncryption (674.39s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       674.431s

$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption -timeout 120m
=== RUN   TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption
 --- PASS: TestAccAWSElasticacheReplicationGroup_enableAtRestEncryption (744.07s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       744.118s
```

nathanielks:
Hooray!!

PR #2091, opened 2017-10-27
Fix TestAccAWSEMRCluster_terminationProtected

catsby:
Fix the `TestAccAWSEMRCluster_terminationProtected` test by using a consistent function to generate the configuration between test steps. `TestAccAWSEMRCluster_terminationProtected` was failing and leaking EMR clusters (and vpcs, et. al). 

It's a 3 step test:

1) create the repo with config A, with `termination_protection` set to `false`. It also has `ebs_root_volume_size` set to `21`. Verify `termination_protection` is `false` in state.
2)  run config B, a copy of config A but with `termination_protection` set to `true`. It also has `ebs_root_volume_size` omitted. Verify `termination_protection` is `true` in state.
3) run config A again, where `termination_protection` is `false` and  `ebs_root_volume_size` is again specified.

The test is failing going from step 2 to 3, because `ebs_root_volume_size` is an `Optional` attribute; it's omission is ok, however, it's also a `ForceNew` attribute, so changing it's value (set to not set) is a `ForceNew` operation. 

Step 1 -> Step 2 is fine, but it does not happen by an in-place update as intended, but a force-new recreation. In this transition, `termination_protection` is turned on. Step 2->Step 3 is where `termination_protection` is turned off, but because `ebs_root_volume_size` is also changed, it's by recreation and not in-place update, so Terraform doesn't try to update the `termination_protection` attribute, it tries to destroy the old and create the new. This fails because `termination_protection` is still on. 

In the PR I use the same method to generate config for A and B, with `ebs_root_volume_size` consistent in both, and `termination_protection` injected in each step. 

I also fix some configuration issues, like a proper `Name` tag, and no longer use the deprecated `roles` attribute of `aws_iam_instance_profile`

### DIFF
--- a/aws/resource_aws_emr_cluster_test.go
+++ b/aws/resource_aws_emr_cluster_test.go
@@ -112,7 +112,7 @@ func TestAccAWSEMRCluster_terminationProtected(t *testing.T) {
 		CheckDestroy: testAccCheckAWSEmrDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccAWSEmrClusterConfig(r),
+				Config: testAccAWSEmrClusterConfigTerminationPolicy(r, "false"),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAWSEmrClusterExists("aws_emr_cluster.tf-test-cluster", &cluster),
 					resource.TestCheckResourceAttr(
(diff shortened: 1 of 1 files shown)


mbfrahry:
```
=== RUN   TestAccAWSEMRCluster_terminationProtected
--- PASS: TestAccAWSEMRCluster_terminationProtected (458.41s)
```
LGTM!

PR #2094, opened 2017-10-28
Update lb_target_group.html.markdown

jkodroff:


radeksimko:
@jkodroff Hey, just thing I noticed - this PR was opened against `stable-website`. Please always open PRs against `master`, we can cherry-pick to `stable-website` from there.

If I didn't notice this then those commits would've been force-pushed over and lost as part of the next release as `master` is the ultimate point of truth.

Thanks for understanding.

PR #2095, opened 2017-10-28
Ignore the 'ResourceNotFoundException' when deleting ElasticSearch Po…

charles-at-geospock:
…licies.

If the Elastic Search Domain Policy has already been deleted, and then you
try to destroy it with Terraform, the detroy will fail, reporting that the
resource is not found. This goes against the rest of the implementation of
the Terraform destroy process that already gone resources are silently
ignored.

The problem was caused by one instance of the code using the wrong name for
the exception, and another instance failing to ignore the lack of the
resource as an acceptable error.

Both of these are addressed here. In passing through, more debugging was
added, in line with other functions, and the common domainName value
held in a local variable.

### DIFF
--- a/aws/resource_aws_elasticsearch_domain_policy.go
+++ b/aws/resource_aws_elasticsearch_domain_policy.go
@@ -40,7 +40,7 @@ func resourceAwsElasticSearchDomainPolicyRead(d *schema.ResourceData, meta inter
 		DomainName: aws.String(name),
 	})
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFound" {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
 			log.Printf("[WARN] ElasticSearch Domain %q not found, removing", name)
 			d.SetId("")
 			return nil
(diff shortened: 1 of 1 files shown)


charles-at-geospock:
This fix only helps in the case where the resources have been destroyed out of band of the Terraform environment (or at least outside the knowledge of the Terraform state). All the other resource types seem to be very happy to ignore resources that have already gone away.

In my case, the resources were destroyed by the Terraform tool, BUT the assumed role that was in use was expired before the completion of the `terraform destroy`. This meant that the S3 upload of the state information never happened, and therefore the next `terraform destroy` was run over the resources and failed for the ESDP. After adding my fix, the destroy completed successfully :-)

aws/resource_aws_elasticsearch_domain_policy.go
```diff
@@ -93,19 +93,27 @@ func resourceAwsElasticSearchDomainPolicyUpsert(d *schema.ResourceData, meta int
...
 		AccessPolicies: aws.String(""),
 	})
 	if err != nil {
+		log.Printf("[WARN] ElasticSearch Domain Policy for %s delete failed: %s", domainName, err)
+
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
+			log.Printf("[WARN] ElasticSearch Domain %q not found, removing", domainName)
+			d.SetId("")
+			return nil
+		}
```
radeksimko: The condition here in `Delete()` isn't necessary, because by default `apply` always refreshes the state and runs `Read()`, so the above condition should be sufficient and more aligned with behaviour of majority of existing resources in the codebase.

aws/resource_aws_elasticsearch_domain_policy.go
```diff
@@ -40,7 +40,7 @@ func resourceAwsElasticSearchDomainPolicyRead(d *schema.ResourceData, meta inter
 		DomainName: aws.String(name),
 	})
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFound" {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
```
radeksimko: This was already addressed in https://github.com/terraform-providers/terraform-provider-aws/pull/2249

radeksimko:
Closing for the reasons mentioned.

Thanks.

PR #2110, opened 2017-10-31
Find LB listener by the LB ARN

tomelliff:
Right now the aws_lb_listener data source isn't particularly useful unless you somehow know the ARN of the listener.
Being able to search by the load balancer ARN means you can chain the aws_lb data source to find the LB listener from an LB name.

As an example we can now do something like this:

```hcl
variable "lb_name" {
  default = "foo"
}

variable "vpc" {
  default = "my_vpc"
}

variable "listener_rule_priority" {
  default = 99
}

data "aws_vpc" "vpc" {
  tags {
    Name = "${var.vpc}"
  }
}

data "aws_lb" "load_balancer" {
  name = "${var.lb_name}"
}

data "aws_lb_listener" "listener" {
  load_balancer_arn = "${data.aws_lb.load_balancer.arn}"
}

resource "aws_lb_listener_rule" "host_based_routing" {
  listener_arn = "${data.aws_lb_listener.listener.arn}"
  priority     = "${var.listener_rule_priority}"

  action {
    type             = "forward"
    target_group_arn = "${aws_lb_target_group.service.arn}"
  }

  condition {
    field  = "host-header"
    values = ["my-service.*.terraform.io"]
  }
}

resource "aws_lb_target_group" "service" {
  name     = "dynamic"
  port     = 80
  protocol = "HTTP"
  vpc_id   = "${data.aws_vpc.vpc.id}"
}
```

In particular this will allow me to go from creating an ALB with every single ECS service I deploy to a single one per ECS cluster (separated by environments) because now I can easily create a listener rule to forward to the service's target group.

In reality I don't really want to have to provide the priority statically with each service so I intend to add some more functionality to the AWS provider at a later date that fetches the highest priority rule for a load balancer listener and then just increment from there. There is an issue for this already at https://github.com/terraform-providers/terraform-provider-aws/issues/1574

Note that if a load balancer has multiple listeners this will error when searching by load_balancer_arn.

### DIFF
--- a/aws/data_source_aws_lb_listener.go
+++ b/aws/data_source_aws_lb_listener.go
@@ -1,6 +1,13 @@
 package aws
 
-import "github.com/hashicorp/terraform/helper/schema"
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+	"log"
+)
 
 func dataSourceAwsLbListener() *schema.Resource {
 	return &schema.Resource{
@@ -9,11 +16,13 @@ func dataSourceAwsLbListener() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"arn": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Computed: true,
 			},
 
 			"load_balancer_arn": {
 				Type:     schema.TypeString,
+				Optional: true,
 				Computed: true,
 			},
 			"port": {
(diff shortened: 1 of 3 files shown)


tomelliff:
For now I've left it so that if you search for a listener by the load balancer's ARN and the load balancer has multiple listeners then the data source should error.

I was thinking about expanding it so you could filter by port but I don't have a use case for that right now and am happy with it working just for a single listener (in my case 443/HTTPS). If it's thought that this would be useful to have before this is merged then I can take a look at it but I'd say this is worth merging as is for now and the other work can be done later if there is a need for it.

tomelliff:
I was hoping to pick up the second part of this to close https://github.com/terraform-providers/terraform-provider-aws/issues/1574 some time over Christmas if I get some free time but would like to know if there's any useful feedback on this part first and ideally get it merged? Is there anything more needed here?

It would cut a fair chunk out of our AWS bill by allowing us to massively reduce the amount of load balancers we have due to having lots of services and then lots of dynamic review environments plus reduce the time to create them (vast majority of the creation time is in the load balancer creation time) which would help a bit.

bflad:
Hi @tomelliff, looks like this functionality (with port filtering as well) was just merged into master in #2886 and will be released in v1.9.0 of the AWS provider. I'm very sorry that your PR here got lost in the shuffle to get a proper review when it should have sooner, along with getting you proper credit. 😢 We just completed labelling the backlog of issues and PRs to hopefully prevent this from happening in the future.

tomelliff:
@bflad No worries, just glad the functionality is in. Combined with https://github.com/terraform-providers/terraform-provider-aws/pull/3219 (or adding the max listener to the ALB data source) means I can reduce a lot of ALBs in my setup and make it much quicker to provide dynamic environments.

Issue #630, opened 2017-06-13
RDS Aurora Cross-Region replication for encrypted cluster failing


hashibot:
_This issue was originally opened by @asadhu as hashicorp/terraform#12985. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
v0.9.1_1

### Affected Resource(s)
RDS Cross-Region Cluster replication

### Terraform Configuration Files
```hcl
resource "aws_rds_cluster" "replica-west" {
  cluster_identifier      = "aurora-replica-cluster-demo"
  replication_source_identifier = "arn:aws:rds:us-east-1:*:cluster:aurora-cluster"
  #"${aws_rds_cluster.default-east.id}"
  availability_zones      = ["us-west-2a"]
  backup_retention_period = 5
  preferred_backup_window = "07:00-09:00"
  storage_encrypted       = "true"
  kms_key_id              = "arn:aws:kms:us-east-1:*:key/[key_name]"
  db_cluster_parameter_group_name = "default.aurora5.6"
  #source-region           = "us-east-1"
}

```

### Debug Output
* aws_rds_cluster.replica-west: InvalidParameterCombination: Source cluster arn:aws:rds:us-east-1:*:cluster:aurora-cluster is encrypted; pre-signed URL has to be specified
	status code: 400

### Expected Behavior
What should have happened?

I have an AWS RDS Aurora Cluster in US East and am trying to create a Cross-Region Cluster on the west. The Replica Cluster should have been created.

### Actual Behavior
What actually happened?

The East Cluster is encrypted and requires kms_key_id and source_region/presigned_url. 'source-region/presigned_url' is not supported in Terraform 0.9.1_1.
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Aurora.Replication.CrossRegion.html

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

asadhu:
@twofivetechnology @grubernaut Any work around that can help to remediate the problem??

asadhu:
Is the SourceRegion being specified

[...]

PR #2111, opened 2017-10-31
RDS Aurora Cross Region replication for encrypted clusters

mitchelldavis44:
This fixes issue RDS Aurora Cross-Region replication for encrypted cluster failing #630.  Issue was it was missing source_region within the module which was making it fail when trying to create a cross region replication cluster that was encrypted with the errror "aws_rds_cluster.replica: InvalidParameterCombination: Source cluster arn:aws:rds:us-east-1:<id>:cluster:<cluster_name> is encrypted; pre-signed URL has to be specified status code: 400, request id: b59ab0b3-812b-11e7-8467-c192637e53bf"

### DIFF
--- a/aws/resource_aws_rds_cluster.go
+++ b/aws/resource_aws_rds_cluster.go
@@ -232,6 +232,11 @@ func resourceAwsRDSCluster() *schema.Resource {
 				Optional: true,
 			},
 
+			"source_region": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
 			"iam_roles": {
 				Type:     schema.TypeSet,
 				Optional: true,
@@ -365,7 +370,8 @@ func resourceAwsRDSClusterCreate(d *schema.ResourceData, meta interface{}) error
 			Engine:                      aws.String(d.Get("engine").(string)),
 			StorageEncrypted:            aws.Bool(d.Get("storage_encrypted").(bool)),
 			ReplicationSourceIdentifier: aws.String(d.Get("replication_source_identifier").(string)),
-			Tags: tags,
+			SourceRegion:                aws.String(d.Get("source_region").(string)),
+			Tags:                        tags,
 		}
 
 		if attr, ok := d.GetOk("port"); ok {
--- a/aws/resource_aws_rds_cluster.go
+++ b/aws/resource_aws_rds_cluster.go
@@ -233,8 +233,8 @@ func resourceAwsRDSCluster() *schema.Resource {
 			},
 
 			"source_region": {
-				Type:     	   schema.TypeString,
-				Optional: 		 true,
+				Type:          schema.TypeString,
+				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"snapshot_identifier"},
 			},
--- a/aws/resource_aws_rds_cluster.go
+++ b/aws/resource_aws_rds_cluster.go
@@ -233,10 +233,10 @@ func resourceAwsRDSCluster() *schema.Resource {
 			},
 
 			"source_region": {
-				Type:     	  	schema.TypeString,
-				Optional: 			true,
-				ForceNew: 			true,
-				ConflictsWith: 	[]string{"snapshot_identifier"},
+				Type:     	   schema.TypeString,
+				Optional: 		 true,
+				ForceNew:      true,
+				ConflictsWith: []string{"snapshot_identifier"},
 			},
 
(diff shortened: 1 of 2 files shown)


mitchelldavis44:
Any objections to merging this?  This is really holding my work back as far as being able to deploy encrypted cross region RDS clusters.

mitchelldavis44:
@radeksimko I believe there should also be a bug label on this since it was missing source_region causing it to fail when trying to deploy a cross region replication cluster that's encrypted.

aws/resource_aws_rds_cluster.go
```diff
@@ -232,6 +232,11 @@ func resourceAwsRDSCluster() *schema.Resource {
 				Optional: true,
 			},
 
+			"source_region": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
radeksimko: I believe we should mark this as `ConflictsWith: []string{"snapshot_identifier"}` because this field will be ignored if the DB is restored, instead of created from scratch.

radeksimko: Also it seems it's not updatable, so we should mark it as `ForceNew: true`.

mitchelldavis44:
Hi @radeksimko 

I have made the changes you've requested, I've never added an acceptance test before but I tried my hand at adding one.  Let me know if you need anything else, thanks!

mitchelldavis44:
@radeksimko is this ready to merge, or are there more changes needed?

radeksimko:
@mitchelldavis44 thanks for modifying the schema, I think we'll need a bit more testing there - i.e. a separate test which spins up an Aurora cluster with both encryption and `source_region`. The current test is just testing whether the field is present, on a config which doesn't do cross-region replication.

mitchelldavis44:
@radeksimko I added a completely new test for cross region replicas that checks that `source_region` & `kms_key_id` are defined as well as sets `storage_encrypted` to true.  Let me know if this will work.

aws/resource_aws_rds_cluster_test.go
```diff
@@ -46,6 +46,40 @@ func TestAccAWSRDSCluster_basic(t *testing.T) {
...
+func TestAccAWSRDSCluster_crossRegion(t *testing.T) {
+	var v rds.DBCluster
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSClusterConfig(acctest.RandInt()),
```
radeksimko: I think we'll also need to create a config which sets up the cross-region replication, `testAccAWSClusterConfig` doesn't do it. Any minimalistic configuration will do in this context, but it should _actually_ setup the replication and test this new behaviour/field - ideally we'd be checking the value of `source_region`, not just whether it's set.

mitchelldavis44: This is a little out of my comfort zone.  Unfortunately I don't really have the experience to do what you're asking, I'm not very experienced in writing go.

mitchelldavis44: @radeksimko

mitchelldavis44: @radeksimko is there anyway we can have this merged?  I'm not the guy to make more complex tests, but this is definitely a needed flag in terraform.

jamisonhyatt:
It looks like there are 2 conflicting PR's solving the same issue - This one, and another https://github.com/terraform-providers/terraform-provider-aws/pull/2808 by @dselvan. 

Both are requesting help with the test - although @dselvan has an almost working Terraform test setup, it doesn't work with both providers (which is necessary to deploy to both east and west for cross-region replication).  

Is there a different way this test can be composed - is it somehow related to `aws_caller_identity` while having multiple `aws` providers aliased?

I have tried searching the tests to find an example of doing something similar (2 providers while calling `aws_caller_identity` and haven't been successful.  Anyone willing to step in the muck and get us unstuck?

mitchelldavis44:
This is getting kinda ridiculous, is there someone that can help with the tests or at the very least a timetable on when you guys will release this?  The code I have in this pull request works, for both east and west and we've been using our "own" provider since this hasn't been merged.  Again, cross region replication works completely with this pull request, any assistance on the test would be greatly appreciated.

bflad:

PR #2126, opened 2017-11-01
validateBatchName should allow hyphens

yzach:
Links to AWS docs 
http://docs.aws.amazon.com/batch/latest/APIReference/API_CreateComputeEnvironment.html#Batch-CreateComputeEnvironment-request-computeEnvironmentName
http://docs.aws.amazon.com/batch/latest/APIReference/API_RegisterJobDefinition.html#Batch-RegisterJobDefinition-request-jobDefinitionName

### DIFF
--- a/aws/validators.go
+++ b/aws/validators.go
@@ -1490,8 +1490,8 @@ func validateSsmParameterType(v interface{}, k string) (ws []string, errors []er
 
 func validateBatchName(v interface{}, k string) (ws []string, errors []error) {
 	value := v.(string)
-	if !regexp.MustCompile(`^[A-Za-z0-9_]{1,128}$`).MatchString(value) {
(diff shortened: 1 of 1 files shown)


toddZilla0130:
We also noticed this and I did a little more research into the matter, testing various edges when naming the 3 Batch resources. In addition to the above, I discovered that:
- names must begin with an alphanumeric
- job queues have the same naming rules as compute environments and job definitions.

The regex I came up with is this `^[0-9a-zA-Z]{1}[0-9a-zA-Z_\-]{0,127}$` 

and I added 
```
ValidateFunc: validateBatchName,
```
to the `name` block of `resourceAwsBatchJobQueue` in `resource_aws_batch_job_queue.go`.

I had opened (and will close) this: https://github.com/terraform-providers/terraform-provider-aws/issues/2108

but if you'd like me to follow up on the above with a new PR I will be happy to do so.

PR #2130, opened 2017-11-01
s3 bucket object usage example typo fix

mannion007:
Corrected error with aws s3 bucket object usage example


mannion007:
More than happy to give something back! Keep up the awesome work!

PR #2132, opened 2017-11-01
resource/aws_launch_configuration: Added support for no_device

maxfortun:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-mapping.html

Launch configs need ability to pass NoDevice for AMIs that come preconfigured with partitions.

### DIFF
--- a/aws/resource_aws_launch_configuration.go
+++ b/aws/resource_aws_launch_configuration.go
@@ -175,6 +175,12 @@ func resourceAwsLaunchConfiguration() *schema.Resource {
 							ForceNew: true,
 						},
 
+						"no_device": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							ForceNew: true,
+						},
+
 						"iops": {
 							Type:     schema.TypeInt,
 							Optional: true,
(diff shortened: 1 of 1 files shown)


maxfortun:
Any estimate on a pull? Should I contact someone? Could really use this in the next release.

aws/resource_aws_launch_configuration.go
```diff
@@ -355,8 +361,10 @@ func resourceAwsLaunchConfigurationCreate(d *schema.ResourceData, meta interface
 		vL := v.(*schema.Set).List()
 		for _, v := range vL {
 			bd := v.(map[string]interface{})
-			ebs := &autoscaling.Ebs{
-				DeleteOnTermination: aws.Bool(bd["delete_on_termination"].(bool)),
+			ebs := &autoscaling.Ebs{}
+
+			if v, ok := bd["no_device"].(bool); !ok && v {
+				ebs.DeleteOnTermination = aws.Bool(bd["delete_on_termination"].(bool))
```
radeksimko: I'm not sure I follow what's happening here. Was this supposed to be a fix of existing code? i.e.

```go
if v, ok := bd["delete_on_termination"].(bool); ok {
    ebs.DeleteOnTermination = aws.Bool(v)
```
if not then can you please explain why can't we just leave the validation of potentially conflicting fields to the API?

radeksimko:
Hi @maxfortun 
do you think you'll find time to finish this?

It's totally 👌 if not - in which case would you be happy for someone else to pick it up?

Thanks.

maxfortun:
Sorry, don't have the time to finish this right now. And no, I do not mind at all if someone takes over and finishes this enhancement. Found a bug in this too. The state does not retain the change to no_device. This needs to be fixed as well.

bflad:

Issue #420, opened 2017-06-13
error redeploying API gateway stage when mapped to custom domain path


hashibot:
_This issue was originally opened by @marioharper as hashicorp/terraform#10674. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

When redeploying an aws_api_gateway_deployment to the same stage and that stage is mapped to an aws_api_gateway_base_path_mapping, the deployment fails with error: 

`aws_api_gateway_deployment.deployment: BadRequestException: Active stages pointing to this deployment must be moved or deleted`

### Terraform Version
7.11

### Affected Resource(s)
- aws_api_gateway_deployment

### Terraform Configuration Files
```hcl
# INPUTS
variable "api_stage" {
  default = "v1"
}

variable "domain_name" {
  default = "example.domain.com"
}

variable "domain_base_path" {
  default = "examplepath"
}


# API
resource "aws_api_gateway_rest_api" "api" {
  name = "my api"
  description = "example api for bug"
}

# API Deployment
resource "aws_api_gateway_deployment" "deployment" {
  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
  stage_name = "${var.api_stage}"
  variables = {
    "functionAlias" = "${var.api_stage}"
    #this is to force a new API deployment on every 'terraform apply'
    "deploymentGUID" = "${uuid()}"
  }
}

# Custom domain path mapping
resource "aws_api_gateway_base_path_mapping" "path_mapping" {
  depends_on=[
    "aws_api_gateway_deployment.deployment"
  ]
  api_id      = "${aws_api_gateway_rest_api.api.id}"
  domain_name = "${var.domain_name}"
  base_path = "${var.domain_base_path}",
  stage_name = "${var.api_stage}"
}
```

### Expected Behavior
API redeployed to the stage and the base path mapping still pointed to that stage. 

### Actual Behavior
API deployment failed with error: 
```sh
Error applying plan:


1 error(s) occurred:

* aws_api_gateway_deployment.deployment: BadRequestException: Active stages pointing to this deployment must be moved or deleted

2016/12/12 16:26:58 [DEBUG] plugin: waiting for all plugin processes to complete...
	status code: 400, request id: cd69598e-c087-11e6-b7ec-1d35a1dd5635
```

### Steps to Reproduce
1. `terraform apply`

[...]

PR #2147, opened 2017-11-01
Apigateway stage import

ericfode:
importer for apigateway_stage 
this allows us to work around #420


Ninir:
Hi @ericfode 

This is getting into shape! is it still a WIP? do you need any help?

Thanks! :)

ericfode:
No, I don't think it is. I would appreciate a code review. 
Currently, my only concern with this PR is around dealing with tags on API Gateway stages. Both the stage and the deployment seem to modify the tags. The changes I added seem to make it more deterministic, but it's nothing I'm proud of.  Any suggestions are appreciated.  @Ninir

aws/import_aws_api_gateway_stage_test.go
```diff
@@ -0,0 +1,27 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSApiGatewayStage_importBasic(t *testing.T) {
+	resourceName := "tf-acc-test.dev"
```
bflad: This should be a reference to a resource in a test configuration, for `testAccAWSAPIGatewayStageConfig_basic` its resource definition is:

```hcl
resource "aws_api_gateway_stage" "test" {
  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
  stage_name = "prod"
  deployment_id = "${aws_api_gateway_deployment.dev.id}"
  cache_cluster_enabled = true
  cache_cluster_size = "0.5"
  variables {
    one = "1"
    two = "2"
  }
  tags {
    Name = "tf-test"
  }
}
```

So this should be `resourceName := "aws_api_gateway_stage.test"`.

It also looks that we require name randomization for `testAccAWSAPIGatewayStageConfig_basic()` now. You can create a variable `rName := acctest.RandomWithPrefix("tf-acc-test")` and call that function as `testAccAWSAPIGatewayStageConfig_basic(rName)`

aws/resource_aws_api_gateway_stage.go
```diff
@@ -300,6 +304,12 @@ func diffVariablesOps(prefix string, oldVars, newVars map[string]interface{}) []
 			if oldValue == newValue {
 				continue
 			}
+		} else {
```
bflad: Can you explain this change here?

aws/resource_aws_api_gateway_stage.go
```diff
@@ -18,6 +19,9 @@ func resourceAwsApiGatewayStage() *schema.Resource {
 		Read:   resourceAwsApiGatewayStageRead,
 		Update: resourceAwsApiGatewayStageUpdate,
 		Delete: resourceAwsApiGatewayStageDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceAwsApiGatewayStageImportState,
```
bflad: To support import when the read function is not using `d.Id()` directly, we just need to simply `d.Set()` the attributes required for read. I would also recommend splitting on `/` which matches the format of the execution ARN. Something like this should work:

```go
State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
	idParts := strings.Split(d.Id(), "/")
	if len(idParts) != 2 || len(d.Id()) < 3 {
		return []*schema.ResourceData{}, fmt.Errorf("Wrong format of ID: %s. Please follow 'rest-api-id/stage-name'", d.Id())
	}
	d.Set("rest_api_id", idParts[0])
	d.Set("stage_name", idParts[1])
	return []*schema.ResourceData{d}, nil
},
```

bflad:
Closing due to inactivity -- happy to review again if someone addresses the feedback.

PR #2154, opened 2017-11-02
New Resource: aws_dx_lag

atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsDxLag_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAwsDxLag_basic -timeout 120m
=== RUN   TestAccAwsDxLag_basic
--- PASS: TestAccAwsDxLag_basic (40.91s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	40.984s
```

### DIFF
--- a/aws/resource_aws_dx_lag.go
+++ b/aws/resource_aws_dx_lag.go
@@ -1,0 +1,162 @@
+package aws
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/directconnect"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsDxLag() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsDxLagCreate,
+		Read:   resourceAwsDxLagRead,
+		Update: resourceAwsDxLagUpdate,
+		Delete: resourceAwsDxLagDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"band_width": &schema.Schema{
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateDxLagBandWidth,
+			},
+			"location": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"num_connections": &schema.Schema{
+				Type:     schema.TypeInt,
+				Required: true,
+				ForceNew: true,
+			},
+			"force_destroy": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+		},
+	}
+}
+
+func resourceAwsDxLagCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dxconn
+
+	input := &directconnect.CreateLagInput{
+		ConnectionsBandwidth: aws.String(d.Get("band_width").(string)),
+		LagName:              aws.String(d.Get("name").(string)),
+		Location:             aws.String(d.Get("location").(string)),
+		NumberOfConnections:  aws.Int64(int64(d.Get("num_connections").(int))),
+	}
+	resp, err := conn.CreateLag(input)
+	if err != nil {
+		return err
+	}
+	d.SetId(*resp.LagId)
+	return resourceAwsDxLagRead(d, meta)
+}
+
+func resourceAwsDxLagRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dxconn
+
+	lagId := d.Id()
+	input := &directconnect.DescribeLagsInput{
+		LagId: aws.String(lagId),
+	}
+	resp, err := conn.DescribeLags(input)
+	if err != nil {
+		d.SetId("")
+		return err
+	}
+	if len(resp.Lags) != 1 {
+		d.SetId("")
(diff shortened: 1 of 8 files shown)


atsushi-ishibashi:
Related #876

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+		LagId: aws.String(lagId),
+	}
+	resp, err := conn.DescribeLags(input)
+	if err != nil {
+		d.SetId("")
+		return err
+	}
+	if len(resp.Lags) != 1 {
+		d.SetId("")
+		return fmt.Errorf("[ERROR] Number of DX Lag (%s) isn't one, got %d", lagId, len(resp.Lags))
```
radeksimko: I'm ok with erroring out if we get != 1 lags, but I don't think it's a reason for removing the LAG from state. We should only be removing it from state if it's not found - i.e. `len(resp.Lags) < 1`

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+func resourceAwsDxLagRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dxconn
+
+	lagId := d.Id()
+	input := &directconnect.DescribeLagsInput{
+		LagId: aws.String(lagId),
+	}
+	resp, err := conn.DescribeLags(input)
+	if err != nil {
+		d.SetId("")
```
radeksimko: I can't find any particular error that would mean LAG is gone in this context: http://docs.aws.amazon.com/directconnect/latest/APIReference/API_DescribeLags.html#c
so I don't think error here is a reason for removing the LAG from state. 🤔

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+		return err
+	}
+	if len(resp.Lags) != 1 {
+		d.SetId("")
+		return fmt.Errorf("[ERROR] Number of DX Lag (%s) isn't one, got %d", lagId, len(resp.Lags))
+	}
+	if d.Id() != *resp.Lags[0].LagId {
+		d.SetId("")
+		return fmt.Errorf("[ERROR] DX Lag (%s) not found", lagId)
```
radeksimko: It makes sense to error out if wrong ID was found (not sure how/why would that happen, but strange things do happen sometimes), but I'm not sure if it's the reason for removing LAG from state.

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+	input := &directconnect.DeleteLagInput{
+		LagId: aws.String(d.Id()),
+	}
+	_, err := conn.DeleteLag(input)
+	if err != nil {
+		return err
+	}
+	deleteStateConf := &resource.StateChangeConf{
+		Pending:    []string{directconnect.LagStateAvailable, directconnect.LagStateRequested, directconnect.LagStatePending},
+		Target:     []string{directconnect.LagStateDeleted, directconnect.LagStateDeleting},
```
radeksimko: Is there any reason we should treat `directconnect.LagStateDeleting` as `Target` state here?

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+		Read:   resourceAwsDxLagRead,
+		Update: resourceAwsDxLagUpdate,
+		Delete: resourceAwsDxLagDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"band_width": &schema.Schema{
```
radeksimko: Nitpick, but is there any reason to drift away from the name in the API? i.e. can't we just call it `connections_bandwidth`?

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateDxLagBandWidth,
+			},
+			"location": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"num_connections": &schema.Schema{
```
radeksimko: Nitpick, but is there any reason to drift away from the name in the API? i.e. can't we just call it `number_of_connections `?

aws/resource_aws_dx_lag_test.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	}
+}
+
+func testAccDxLagConfig(rName string) string {
+	return fmt.Sprintf(`
+    resource "aws_dx_lag" "hoge" {
+      name = "tf-dx-lag-%s"
+      band_width = "1Gbps"
+      location = "EqDC2"
```
radeksimko: Do you mind changing the location to a one which is available in `us-west-2` as all of our tests run in that region? e.g. `EqSe2`

aws/resource_aws_dx_lag.go
atsushi-ishibashi: I totally agree with your reviews!

aws/resource_aws_dx_lag.go
```diff
@@ -0,0 +1,162 @@
...
+	input := &directconnect.DescribeLagsInput{
+		LagId: aws.String(lagId),
+	}
+	resp, err := conn.DescribeLags(input)
+	if err != nil {
+		return err
+	}
+	if len(resp.Lags) != 1 {
+		if len(resp.Lags) < 1 {
+			d.SetId("")
```
radeksimko: I think in this case (when Lag was removed) we don't want to error out - at least it's a convention not to do that.

Also - which is rather a nitpick - we can move this condition above or below the other condition and avoid the nesting & indentation so it's easier to read.

atsushi-ishibashi: I understood the convention 🙇

PR #2155, opened 2017-11-02
Dependency: directconnect@v1.12.20

atsushi-ishibashi:
I'm developing Direct Connect Gateway.
https://aws.amazon.com/jp/blogs/aws/new-aws-direct-connect-gateway-inter-region-vpc-access/

```
govendor fetch github.com/aws/aws-sdk-go/service/directconnect@v1.12.20
```

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -519,12 +519,12 @@
 			"versionExact": "v1.12.19"
 		},
 		{
-			"checksumSHA1": "vJAA3F/3UhmdWX7s3r+5YxEUuUg=",
+			"checksumSHA1": "2139YRKSqXN9N49CFTopzwftCDw=",
 			"path": "github.com/aws/aws-sdk-go/service/directconnect",
-			"revision": "ee1f179877b2daf2aaabf71fa900773bf8842253",
-			"revisionTime": "2017-10-26T21:07:55Z",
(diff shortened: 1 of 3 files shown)


vendor/vendor.json
```diff
@@ -519,12 +519,12 @@
...
+			"checksumSHA1": "2139YRKSqXN9N49CFTopzwftCDw=",
 			"path": "github.com/aws/aws-sdk-go/service/directconnect",
-			"revision": "ee1f179877b2daf2aaabf71fa900773bf8842253",
-			"revisionTime": "2017-10-26T21:07:55Z",
-			"version": "v1.12.19",
-			"versionExact": "v1.12.19"
+			"revision": "80dcc100bd75b8742a5ffc6cfc2200efbe5a59d3",
+			"revisionTime": "2017-11-01T17:10:25Z",
+			"version": "v1.12.20",
+			"versionExact": "v1.12.20"
```
Ninir: When updating a given vendor, we should update all other ones, in order not to have side effects.
This will be handled by https://github.com/terraform-providers/terraform-provider-aws/pull/2169, where I bumped every dependency to match 1.12.21 :)

So might be better to close this one then. Thoughts? 😄

atsushi-ishibashi: I got it👍

PR #2159, opened 2017-11-02
Batch name extra checks

toddZilla0130:
- Augment the `validateBatchName` method to ensure name begins with alphanumeric (i.e., no underscore or hyphen). 
- Enforce the same name qualifications for **job queues**.

Neither of the above is officially documented by AWS; however they are real restrictions that will cause resource creation to fail if not adhered to. All have been hand-tested.

### DIFF
--- a/aws/resource_aws_batch_job_queue.go
+++ b/aws/resource_aws_batch_job_queue.go
@@ -27,8 +27,9 @@ func resourceAwsBatchJobQueue() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 			"name": {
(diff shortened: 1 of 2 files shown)


toddZilla0130:
I wrote feedback on the relevant documentation pages and will reach out to our account rep. Right now they're 💯  focused on re:Invent so not much else short of an actual fire is going to get a lot of attention.

Thanks for merging this and I will continue to make contributions as my knowledge of Go increases.

radeksimko:
@toddZilla0130 Right, I wasn't suggesting (nor expecting) they fix it straight away 😄 just making sure the feedback is passed and _eventually_ processed.

PR #2162, opened 2017-11-02
Documentation: r/ecs_task_definition note about quote escaping container_definition

### LINKED ISSUE #2142: Terraform ECS task environment variable with json
_This issue was originally opened by @sunilkumarmohanty as hashicorp/terraform#16519. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

### Terraform Version
0.10.0

```hcl
# Copy-paste your Terraform configurations here.
#

I am creating an ecs task definition with terraform. My variable looks like the below
some_json_env = "{\"foo\":\"bar\"}"

In ecs task definition I am setting the environment variable

{
          "name": "SOME_JSON",
          "value": "${var.some_json_env}"
      }

I get the 

### Expected Behavior
I want the environment variable in the container to be as below
SOME_JSON = {"foo":"bar"}

### Actual Behavior
I get an error saying error decoding JSON: invalid character 'f' after object key:value pair

bflad:
This is a targeted documentation note, but maybe it could be put somewhere more generic and referenced in various places where the Terraform/API input is raw JSON for a field and could contain quotes in values?

Closes #2142

### DIFF
--- a/website/docs/r/ecs_task_definition.html.markdown
+++ b/website/docs/r/ecs_task_definition.html.markdown
@@ -6,7 +6,7 @@ description: |-
   Provides an ECS task definition.
 ---
 
-# aws\_ecs\_task\_definition
+# aws_ecs_task_definition
 
 Provides an ECS task definition to be used in `aws_ecs_service`.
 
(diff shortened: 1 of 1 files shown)


Ninir:
Hi @bflad 

Good call! :)

I do think this would have sense in the [interpolation documentation](https://www.terraform.io/docs/configuration/interpolation.html) for instance, and then link notes in other pages to it. Thoughts? :)

catsby:
Added to the Interpolation docs in https://github.com/hashicorp/terraform/pull/16940

bflad:
💯 🎉

PR #2168, opened 2017-11-03
Docs: Remove tags from spot fleet example

DavidWittman:
As indicated later in the documentation for this resource, `tags` is not
a valid key in the launch specification. Adding tags will result in the
following error at runtime:

    invalid or unknown key: tags

### DIFF
--- a/website/docs/r/spot_fleet_request.html.markdown
+++ b/website/docs/r/spot_fleet_request.html.markdown
@@ -42,10 +42,6 @@ resource "aws_spot_fleet_request" "cheap_compute" {
       volume_size = "300"
       volume_type = "gp2"
     }
-
-    tags {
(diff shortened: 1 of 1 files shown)


Ninir:
Hi @DavidWittman 

This `tags` attribute is valid and does work since [the recent work merge](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L259), and is now available using the AWS Provider 1.2.0.

To proceed, go to your Terraform stack and run `terraform init -upgrade`. Tried it myself and everything works as expected! 😄 

For this reason, i'm going to close this.

Happy Terraforming!

DavidWittman:
Thanks @Ninir!

PR #2173, opened 2017-11-04
New Resource: aws_dx_connection

atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsDxConnection_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAwsDxConnection_basic -timeout 120m
=== RUN   TestAccAwsDxConnection_basic
--- PASS: TestAccAwsDxConnection_basic (64.31s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	64.364s
```

### DIFF
--- a/aws/resource_aws_dx_connection.go
+++ b/aws/resource_aws_dx_connection.go
@@ -22,7 +22,7 @@ func resourceAwsDxConnection() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-			"band_width": &schema.Schema{
+			"bandwidth": &schema.Schema{
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
@@ -41,7 +41,7 @@ func resourceAwsDxConnectionCreate(d *schema.ResourceData, meta interface{}) err
 	conn := meta.(*AWSClient).dxconn
 
 	input := &directconnect.CreateConnectionInput{
-		Bandwidth:      aws.String(d.Get("band_width").(string)),
+		Bandwidth:      aws.String(d.Get("bandwidth").(string)),
 		ConnectionName: aws.String(d.Get("name").(string)),
 		Location:       aws.String(d.Get("location").(string)),
 	}
@@ -62,15 +62,16 @@ func resourceAwsDxConnectionRead(d *schema.ResourceData, meta interface{}) error
 	}
 	resp, err := conn.DescribeConnections(input)
 	if err != nil {
-		d.SetId("")
 		return err
 	}
-	if len(resp.Connections) != 1 {
+	if len(resp.Connections) < 1 {
 		d.SetId("")
+		return nil
+	}
+	if len(resp.Connections) != 1 {
 		return fmt.Errorf("[ERROR] Number of DX Connection (%s) isn't one, got %d", connectionId, len(resp.Connections))
 	}
 	if d.Id() != *resp.Connections[0].ConnectionId {
-		d.SetId("")
 		return fmt.Errorf("[ERROR] DX Connection (%s) not found", connectionId)
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


atsushi-ishibashi:
After this PR and #2154 will be merged, I'll develop `aws_dx_lag_association` to associate connection with lag.

aws/resource_aws_dx_connection.go
```diff
@@ -0,0 +1,116 @@
...
+		Read:   resourceAwsDxConnectionRead,
+		Delete: resourceAwsDxConnectionDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"band_width": &schema.Schema{
```
radeksimko: nitpick, but I never saw this spelled as "band width" and the API also uses 1 word - bandwidth - mind changing that?

aws/resource_aws_dx_connection.go
```diff
@@ -0,0 +1,116 @@
...
+func resourceAwsDxConnectionRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dxconn
+
+	connectionId := d.Id()
+	input := &directconnect.DescribeConnectionsInput{
+		ConnectionId: aws.String(connectionId),
+	}
+	resp, err := conn.DescribeConnections(input)
+	if err != nil {
+		d.SetId("")
```
radeksimko: As mentioned in your other DX PR - we shouldn't remove a resource from the state on any error and there's no relevant error that would signal missing resource: http://docs.aws.amazon.com/directconnect/latest/APIReference/API_DescribeConnections.html#API_DescribeConnections_Errors

aws/resource_aws_dx_connection.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+	if len(resp.Connections) != 1 {
+		d.SetId("")
```
radeksimko: Same here - we should only ever remove the resource from state if `len(resp.Connections) < 1` (and _not_ error out in such case).

aws/resource_aws_dx_connection.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("[ERROR] Number of DX Connection (%s) isn't one, got %d", connectionId, len(resp.Connections))
+	}
+	if d.Id() != *resp.Connections[0].ConnectionId {
+		d.SetId("")
```
radeksimko: Same here ^

aws/resource_aws_dx_connection.go
```diff
@@ -0,0 +1,116 @@
...
+	input := &directconnect.DeleteConnectionInput{
+		ConnectionId: aws.String(d.Id()),
+	}
+	_, err := conn.DeleteConnection(input)
+	if err != nil {
+		return err
+	}
+	deleteStateConf := &resource.StateChangeConf{
+		Pending:    []string{directconnect.ConnectionStatePending, directconnect.ConnectionStateOrdering, directconnect.ConnectionStateAvailable, directconnect.ConnectionStateRequested},
+		Target:     []string{directconnect.ConnectionStateDeleted, directconnect.ConnectionStateDeleting},
```
radeksimko: Any particular reason to treat `directconnect.ConnectionStateDeleting` as target here?

aws/resource_aws_dx_connection_test.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil
+	}
+}
+
+func testAccDxConnectionConfig(rName string) string {
+	return fmt.Sprintf(`
+    resource "aws_dx_connection" "hoge" {
+      name = "tf-dx-%s"
+      band_width = "1Gbps"
+      location = "EqDC2"
```
radeksimko: As mentioned in your other PR - we run all acceptance tests in `us-west-2`, so we'll need to set a different location here, e.g. `EqSe2`

aws/validators.go
```diff
@@ -1651,3 +1651,21 @@ func validateCognitoRoles(v map[string]interface{}, k string) (errors []error) {
 
 	return
 }
+
+func validateDxConnectionBandWidth(v interface{}, k string) (ws []string, errors []error) {
```
radeksimko: How do you feel about renaming the other validation function from the previous PR and reusing it here? It seems to be validating the exact same thing.

aws/resource_aws_dx_connection_test.go
```diff
@@ -0,0 +1,74 @@
...
+		input := &directconnect.DescribeConnectionsInput{
+			ConnectionId: aws.String(rs.Primary.ID),
+		}
+
+		resp, err := conn.DescribeConnections(input)
+		if err != nil {
+			return err
+		}
+		for _, v := range resp.Connections {
+			if *v.ConnectionId == rs.Primary.ID && !(*v.ConnectionState == directconnect.ConnectionStateDeleted || *v.ConnectionState == directconnect.ConnectionStateDeleting) {
```
radeksimko: `directconnect.ConnectionStateDeleting` shouldn't be treated as "gone" here, IMO

aws/validators.go
atsushi-ishibashi: I think `validateDxConnectionBandWidth` is better name so I replace dx_lag validate func with it.

Issue #1416, opened 2017-08-15
Feature request: Support AWS Glue


darrenhaken:
AWS Glue has now become GA https://aws.amazon.com/glue/

It would be nice to have this running through Terraform. Do you accept PR's as I could look to add it?

Ninir:
Hey @darrenhaken 

It seems the service is available in the Go SDK and CRUDs operations are also available.
Feel free to open a PR for that, we'll be more than happy to review & help on that :)

darrenhaken:
Thanks :-) is there any docs on how to write Terraform to hook into the SDK? I'll do some digging myself of course.
		
		

		
	




On Tue, Aug 15, 2017 at 10:27 AM +0100, "Gauthier Wallet" <notifications@github.com> wrote:












Hey @darrenhaken


It seems the service is available in the Go SDK and CRUDs operations are also available.

Feel free to open a PR for that, we'll be more than happy to review & help on that :)

Ninir:
#### Vendor dependencies
Generally speaking, you first need to vendor the dependencies, as exposed here: https://github.com/terraform-providers/terraform-provider-aws/pull/1417.

#### Connection management
To be able to interact with the SDK and thus the API, you would need to hook the connection, in the aws/config.go file. Then, you would be able to get & use it in the resource.

#### Resource creation
A good start is the plugin provider [documentation page](https://www.terraform.io/docs/plugins/provider.html), along with the [core repository documentation](https://github.com/hashicorp/terraform#developing-terraform).

At first you would need to create a HCL structure that matches the API. This is about creating a schema, which is an abstraction of the resource itself. 

#### Tests
Adding acceptance tests is a good way of testing without the need of compiling yourself but reproducing a real-world use-case.
I encourage you to check what was already made by other contributors, which is a good start.

Once the test is written, run it:

    make testacc TEST=./aws TESTARGS='-run=TestAccAWSResourceGlueClassifier_'

This will run only the test with name prefix being `TestAccAWSResourceGlueClassifier_`.

#### Documentation
The documentation is a required step for people to use it. It should be crystal clear about how to use each options, what is required, optional.
Again, I encourage you to check what was already done, will help a lot 😄 

#### Global example
Check out [this example](https://github.com/terraform-providers/terraform-provider-aws/commit/106b7ea6cbed6665297697dc3389b85498314f27) about hooking into the SDK, resource, tests, documentation, etc.

[...]

PR #2175, opened 2017-11-05
New Resource: AWS Glue Catalog Database

drewsonne:
Implemented the AWS Glue Catalog database resource as described here http://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-catalog-databases.html to address #1416.

This is my first PR here so I want to keep it simple and only implement a single resource and will implement other resources for AWS glue once I've done the process once


drewsonne:
@bflad thanks for the heads up. 

Have changed this to a WIP PR, and added a new PR for the vendor changes https://github.com/terraform-providers/terraform-provider-aws/pull/2177

drewsonne:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSGlueCatalogDatabase_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSGlueCatalogDatabase_basic -timeout 120m
=== RUN   TestAccAWSGlueCatalogDatabase_basic
--- PASS: TestAccAWSGlueCatalogDatabase_basic (21.46s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	21.488s
```

drewsonne:
Is there anything else which is needed to get this PR in?

aws/resource_aws_glue_catalog_database_test.go
```diff
@@ -0,0 +1,128 @@
...
+		if rs.Type != "aws_glue_catalog_database" {
+			continue
+		}
+
+		input := &glue.GetDatabaseInput{
+			Name: aws.String(rs.Primary.ID),
+		}
+		if _, err := conn.GetDatabase(input); err != nil {
+			//Verify the error is what we want
+			if ae, ok := err.(awserr.Error); ok && ae.Code() == "EntityNotFoundException" {
```
bflad: Two notes about this:
* We have a helper function which can let you remove the `awserr` import from this file
* I would recommend using the SDK constant here

`isAWSErr(err, glue.ErrCodeEntityNotFoundException, "")`

aws/resource_aws_glue_catalog_database_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/glue"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSGlueCatalogDatabase_basic(t *testing.T) {
```
bflad: Can you please add a second `resource.TestStep` that updates the attributes and ensures the updates propagate properly to AWS/Terraform?

drewsonne:
Amazing thank you! I will have a work through those when I next get a chance. I based this off the glacier resource, so that's where the conventions came from.

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,149 @@
...
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"parameters": &schema.Schema{
+				Type:     schema.TypeMap,
+				Elem:     schema.TypeString,
+				Optional: true,
+			},
+			"create_time": &schema.Schema{
```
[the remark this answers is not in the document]
drewsonne: There is no specific use case, my motivation was simply to expose the Glue API for the [Database](http://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-catalog-databases.html#aws-glue-api-catalog-databases-Database) object with as little logic as possible. If the convention within the AWS resources in terraform is not to expose this, I will remove it. Can you just confirm that hiding the `CreateTime` attribute would be part of the convention within the provider?

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,149 @@
...
+			},
+		},
+	}
+}
+
+func resourceAwsGlueCatalogDatabaseCreate(d *schema.ResourceData, meta interface{}) error {
+	glueconn := meta.(*AWSClient).glueconn
+	name := d.Get("name").(string)
+
+	input := &glue.CreateDatabaseInput{
```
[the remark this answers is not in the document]
drewsonne: I do not think this should be supported within the create statement. I am not entirely sure why it's there, but it's not listed in the [`DatabaseInput` structure](http://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-catalog-databases.html#aws-glue-api-catalog-databases-DatabaseInput), and in my integration tests, adding it caused failures from the API.

aws/resource_aws_glue_catalog_database_test.go
drewsonne: I have attempted to do this, but I struggling to get it working.
When I add a second step with a modified `description` field, I get:

```
testing.go:503: Step 0 error: After applying this step, the plan was not empty:
	
	DIFF:
	
	DESTROY/CREATE: aws_glue_catalog_database.test
	  catalog_id:        "440474553311" => "" (forces new resource)
	  description:       "A test catalog from terraform" => "A test catalog from terraform"
	  location_uri:      "my-location" => "my-location"
	  name:              "my_test_catalog_database_6335041086110733121" => "my_test_catalog_database_6335041086110733121"
	  parameters.%:      "3" => "3"
	  parameters.param1: "value1" => "value1"
	  parameters.param2: "1" => "1"
	  parameters.param3: "50" => "50"
	
	STATE:
	
	aws_glue_catalog_database.test:
	  ID = 440474553311:my_test_catalog_database_6335041086110733121
	  catalog_id = 440474553311
	  description = A test catalog from terraform
	  location_uri = my-location
	  name = my_test_catalog_database_6335041086110733121
	  parameters.% = 3
	  parameters.param1 = value1
	  parameters.param2 = 1
	  parameters.param3 = 50
```

When I take the second step out, it works, so I am not sure why it's saying the plan was not empty at the end if it succeeded with the first step. If there's only one step, does it not test for an empty plan?

I've had a look at https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_ecs_service_test.go to get an idea of a config change test, but I can't see why mine fails and that one works. @bflad , are you able to shed any light on this? I'm keen to get this test working, as it's a problem I ran across in my own provider. That is, how to test diffs successfully.

bflad: Try adding `Computed: true` to the optional `catalog_id` attribute in the schema. This should tell Terraform that its value is allowed to be set from the provider instead of just the configuration. 😄

bflad: tl;dr `Computed: true` should be set on any attributes that AWS might return to be set in the state that aren't required by writing Terraform configuration.

website/docs/r/glue_catalog_database.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+  name = "MyCatalogDatabase"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the database.
+* `catalogID` - (Optional) ID of the Glue Catalog to create the database in. If omitted, this defaults to the AWS Account Id.
```
bflad: Minor typo: `catalog_id`

website/docs/r/glue_catalog_database.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+* `description` - (Optional) Description of the database.
+* `location_uri` - (Optional) The location of the database (for example, an HDFS path).
+* `parameters` - (Optional) A list of key-value pairs that define parameters and properties of the database.
+
+## Import
+
+Glue Catalog Databases can be imported using the `name`, e.g.
```
bflad: Thanks for adding the catalog ID! Hopefully this makes this resource a little more future-proof.  Bump to mention we also support importing via `catalog_id:name` 👍

bflad:
This is looking awesome! Just a few more little things from my perspective (most importantly the update test), we'll verify the testing, then 🚀  . Thanks for your effort on this!

aws/resource_aws_glue_catalog_database_test.go
drewsonne: Great thanks! I was under the impression `Computed: true` was *only* for attributes which are computed, and not attributes which *could* be computed. This is working on my integration tests now.
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSGlueCatalogDatabase_*'
TF_ACC=1 go test ./aws -v -run=TestAccAWSGlueCatalogDatabase_* -timeout 120m
=== RUN   TestAccAWSGlueCatalogDatabaseVault_importBasic
--- PASS: TestAccAWSGlueCatalogDatabaseVault_importBasic (22.05s)
=== RUN   TestAccAWSGlueCatalogDatabase_basic
--- PASS: TestAccAWSGlueCatalogDatabase_basic (33.43s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	55.511s
```

drewsonne:
No worries :-) I have the glue crawler patiently waiting in another branch so there will be another PR probably in the next couple of weeks :-)

aws/import_aws_glue_catalog_database_test.go
```diff
@@ -0,0 +1,29 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccAWSGlueCatalogDatabaseVault_importBasic(t *testing.T) {
```
bflad: Nitpick: Dang, sorry I didn't notice this copypasta earlier - just needs `Vault` removed

drewsonne:
@bflad ok, I think I'm done fiddling with this now, so I shan't change it unless there are requests.

bflad:
This is passing locally for me! The only change I can think of left is adding `Computed: true` to the other `Optional: true` attributes (we're not explicitly testing their lack of configuration in the `basic` test, which is why testing is passing currently - I don't think we need to go crazy and test each optional attribute separately, but maybe the other maintainers will have an opinion). Otherwise this LGTM and :shipit: 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSGlueCatalogDatabase'
TF_ACC=1 go test ./aws -v -run=TestAccAWSGlueCatalogDatabase -timeout 120m
=== RUN   TestAccAWSGlueCatalogDatabase_importBasic
--- PASS: TestAccAWSGlueCatalogDatabase_importBasic (13.54s)
=== RUN   TestAccAWSGlueCatalogDatabase_basic
--- PASS: TestAccAWSGlueCatalogDatabase_basic (19.02s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	32.602s
```

Hashicorp employees are on vacation until the new year so might take a little longer for this PR to be reviewed/merged by someone with access to the merge button unless another outside maintainer is poking around. 👍

drewsonne:
The other optional parameters would not be filled in with anything from AWS nor any defaults beyond empty values. Would we need to test that those values are still empty after a read operation from AWS?

bflad:
In general, we want to know if something occurred outside Terraform as well (e.g. a manual update in the UI). I believe the intent of `Computed: true` for the other optional attributes is to catch these scenarios. If you look around at other resource schemas, you should see that the two are often paired together.

bflad:
LGTM once the Jenkins testing is happy.

twistedgrim:
@bflad  Is there an update on release on this, as its now after the holiday season? 
Thank you!

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,180 @@
...
+				Optional: true,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsGlueCatalogDatabaseCreate(d *schema.ResourceData, meta interface{}) error {
+	glueconn := meta.(*AWSClient).glueconn
+	catalogID := createAwsGlueCatalogID(d, meta)
```
radeksimko: Nitpick, but do you mind reducing the interface of the function a bit? It IMO doesn't need to take the whole `meta`, maybe we can just pass the account ID?

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,180 @@
...
+	input := &glue.CreateDatabaseInput{
+		CatalogId: aws.String(catalogID),
+		DatabaseInput: &glue.DatabaseInput{
+			Name: aws.String(name),
+		},
+	}
+
+	_, err := glueconn.CreateDatabase(input)
+	if err != nil {
+		return fmt.Errorf("Error creating Catalogue Database: %s", err)
```
radeksimko: Nitpick: the service name is spelled `catalog` (American spelling).

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,180 @@
...
+	glueconn := meta.(*AWSClient).glueconn
+
+	catalogID, name := readAwsGlueCatalogID(d.Id())
+	input := &glue.UpdateDatabaseInput{
+		CatalogId: aws.String(catalogID),
+		DatabaseInput: &glue.DatabaseInput{
+			Name:        aws.String(name),
+			Description: aws.String(d.Get("description").(string)),
+			LocationUri: aws.String(d.Get("location_uri").(string)),
+			Parameters:  make(map[string]*string),
```
radeksimko: AFAICT the above 3 fields are optional, so we shouldn't be sending empty strings if the user has not defined them in the config. You can use `v, ok := d.GetOk("field_name")` to check if the field was defined and only append it to the input if it actually was.

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	for key, value := range d.Get("parameters").(map[string]interface{}) {
+		input.DatabaseInput.Parameters[key] = aws.String(value.(string))
+	}
+
+	doUpdate := d.HasChange("description")
+	doUpdate = doUpdate || d.HasChange("location_uri")
+	doUpdate = doUpdate || d.HasChange("parameters")
+
+	if doUpdate {
```
radeksimko: Nitpick, but this is a tiny little bit more difficult to read/understand - could be simplified to `if d.HasChange("description") || d.HasChange("location_uri") || d.HasChange("parameters") {`.

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,180 @@
...
+	catalogID, name := readAwsGlueCatalogID(d.Id())
+
+	input := &glue.GetDatabaseInput{
+		CatalogId: aws.String(catalogID),
+		Name:      aws.String(name),
+	}
+
+	out, err := glueconn.GetDatabase(input)
+	if err != nil {
+		return fmt.Errorf("Error reading Glue Cataloge Database: %s", err.Error())
```
radeksimko: nitpick: typo 👀

radeksimko: Also do you mind adding a check for database which was removed outside of Terraform? something like this should suffice:

```go
if isAWSErr(err, glue.ErrCodeEntityNotFoundException, "") {
	log.Printf("[WARN] Glue Catalog Database (%s) not found, removing from state", d.Id())
	d.SetId("")
}
```
you may also add an acceptance test verifying this behaviour, similar to e.g. https://github.com/terraform-providers/terraform-provider-aws/blob/79ed212c594fd8e848d0274cd0efc70bbf4cede8/aws/resource_aws_eip_test.go#L220-L238

but the test is not a blocker.

website/docs/r/glue_catalog_database.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+  Provides a Glue Catalog Database.
+---
+
+# aws_glue_catalog_database
+
+Provides a Glue Catalog Database Resource. You can refer to the [Glue Developer Guide](http://docs.aws.amazon.com/glue/latest/dg/populate-data-catalog.html) for a full explanation of the Glue Data Catalog functionality
+
+## Example Usage
+
+```hcl-terraform
```
radeksimko: Nitpick: We tend to use just `hcl` here for simplicity.

aws/resource_aws_glue_catalog_database.go
```diff
@@ -0,0 +1,180 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"parameters": {
+				Type:     schema.TypeMap,
+				Elem:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
```
radeksimko: What is the reason for setting the above 3 fields as `Computed`? I tried removing the flag and all acceptance tests passed which makes me think there is no reason, but maybe I'm wrong and the behaviour just isn't covered  in tests?

drewsonne: @radeksimko the motivation is because of the above discussion re computed flags with @bflad 

bflad:
>  In general, we want to know if something occurred outside Terraform as well (e.g. a manual update in the UI). I believe the intent of Computed: true for the other optional attributes is to catch these scenarios. If you look around at other resource schemas, you should see that the two are often paired together.

I don't know the functional requirements here, and testing the functionality describe i the quote here to import would require a step in between each `resource.TestStep` which goes into AWS and modifies the AWS resource.

I'm happy to do this, but I have not seen any prior art as to how one would go about this. But also, I would like to have some sort of idea about what is expected in terms of functionality so I don't end up bouncing back and forth between two different opinions of Reviewers :-)

I'm not sure who is closer to the project between yourself @radeksimko and @bflad :-)

aws/resource_aws_glue_catalog_database.go
drewsonne: Ah noo! I'm an Australian living in London! I think the world would be a bit better off with a generous sprinkling of vowels here and there ;-P

aws/resource_aws_glue_catalog_database.go
drewsonne: As I mentioned above, I wish to round out the whole AWS Glue product set for terraform. For the sake of my sanity (getting a PR merged), can I put this test in a subsequent PR?

radeksimko: Adding a test in a subsequent PR is totally fine. 😉

drewsonne:
```
$ date; make testacc TEST=./aws TESTARGS='-run=TestAccAWSGlueCatalogDatabase'
Wed 10 Jan 2018 07:43:30 GMT
TF_ACC=1 go test ./aws -v -run=TestAccAWSGlueCatalogDatabase -timeout 120m
=== RUN   TestAccAWSGlueCatalogDatabase_importBasic
--- PASS: TestAccAWSGlueCatalogDatabase_importBasic (26.66s)
=== RUN   TestAccAWSGlueCatalogDatabase_full
--- PASS: TestAccAWSGlueCatalogDatabase_full (57.37s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	84.055s
```

aws/resource_aws_glue_catalog_database.go
radeksimko: I will ask @bflad about this when he wakes up.

drewsonne:
Also... if I wanted to add a list of things I have learned in this PR about how one goes about building a resource (code style, terraform-provider-aws idiomatic process, what type of tests, when to use `computed`), which .md file in the repo would I put it in?

radeksimko:
> if I wanted to add a list of things I have learned in this PR about how one goes about building a resource (code style, terraform-provider-aws idiomatic process, what type of tests, when to use computed), which .md file in the repo would I put it in?

https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md is probably the best place for such things.

cemo:
Hope that other resources will be added too. 👍 Thanks for the great PR.

aws/resource_aws_glue_catalog_database.go
radeksimko: Just confirmed with @bflad privately that this was just a misunderstanding and we mutually agreed it should be removed from the schema here.

Sorry for any confusion caused by this!

After that I believe we're ready to merge it 🎉

drewsonne: For my own understanding, aside from obvious things (like instance ids i-12345678, etc), when should the `Computed:` tag be used?

drewsonne:
Sigh. 

Ok, the glue module in vendor has been deleted from the master branch, so I'm going to go through the vendor PR's and find out where it went.

drewsonne:
Looks like it was deleted during an update to terraform 0.11.2.  @bflad @radeksimko does this need a separate PR to pull this dependency back into master, or can I put it in this PR?

https://github.com/terraform-providers/terraform-provider-aws/commit/65b06012e6708f41dbf2e854ad307d7a4562de6c#diff-bd290170e2912d3e8694db1a151066e5L619

aws/resource_aws_glue_catalog_database.go
radeksimko: `Computed` w/ `Optional` is typically used when the API automatically sets unpredictable default value or when the value is generally not static and depends on other attributes.

It is an expectation of Terraform users to see difference in `plan` if something has changed outside of Terraform (e.g. someone manually changed description through AWS Console or AWS CLI). User has some options to ignore certain changes if they wish (via `lifecycle` block), but overall this is a desired behaviour.

The downside of marking something `Computed` is that it will ignore such changes, but sometimes it's necessary. We always prefer static `Default` where possible.

`Computed` only is used for any IDs or anything not defined in the config and returned by the API.

radeksimko:
Ahh, sorry for this. Looks like we were a bit too excited in https://github.com/terraform-providers/terraform-provider-aws/pull/2722 about removal of some "unused" packages...

I can track down the rest of missing packages - feel free to send a PR for `glue` - I can merge it straight away!

drewsonne:
@radeksimko PR in https://github.com/terraform-providers/terraform-provider-aws/pull/2933/files

radeksimko:
@drewsonne merged - feel free to rebase.

aws/resource_aws_glue_catalog_database.go
drewsonne: So, there is a case above https://github.com/terraform-providers/terraform-provider-aws/pull/2175#discussion_r157543208 where I have a value (`catalog_id`), which at the moment is *only* able to be set as an AWS AccountID (not necessarily the same AccountID as TF is running as). If the `catalog_id` is omitted, AWS will fill it in.

I assume this is because AWS will eventually have the ability for multiple catalogues in a single account, a globally unique namespace for public catalogues, or something such. This then is a case where the user could provide `catalog_id`, but if they don't AWS will compute it.

This variable is currently `Computed: true` and given what you said above it seems like it is an ID, but... it could be a user set value, so it feels like a grey area, but... if it changes, the resource needs to be re-built, so it should never turn up as a diff in an update operation.

But, I would be correct in saying that this `catalog_id` is a valid use case for `Computed: true`?

drewsonne:
@radeksimko Integration tests pass.

```
$ date; make testacc TEST=./aws TESTARGS='-run=TestAccAWSGlueCatalogDatabase'
Thu 11 Jan 2018 08:08:57 GMT
TF_ACC=1 go test ./... -v -run=TestAccAWSGlueCatalogDatabase -timeout 120m
=== RUN   TestAccAWSGlueCatalogDatabase_importBasic
--- PASS: TestAccAWSGlueCatalogDatabase_importBasic (29.81s)
=== RUN   TestAccAWSGlueCatalogDatabase_full
--- PASS: TestAccAWSGlueCatalogDatabase_full (58.42s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	88.262s
```

So, @radeksimko, we think this is good to go now? :-)
poke

aws/resource_aws_glue_catalog_database.go
radeksimko: That's right, `catalog_id` is a valid use case. It's kind of a special case I didn't mention in my previous explanation.

drewsonne:
Wee! Thanks @radeksimko! Will get another one for glue in soonish

bflad:

PR #2178, opened 2017-11-05
data/source_aws_db_snapshot: Avoid nil pointer when snapshot is still being created

atsushi-ishibashi:
#1794

### DIFF
--- a/aws/data_source_aws_db_snapshot.go
+++ b/aws/data_source_aws_db_snapshot.go
@@ -212,7 +212,11 @@ func dbSnapshotDescriptionAttributes(d *schema.ResourceData, snapshot *rds.DBSna
 	d.Set("source_region", snapshot.SourceRegion)
 	d.Set("status", snapshot.Status)
 	d.Set("vpc_id", snapshot.VpcId)
-	d.Set("snapshot_create_time", snapshot.SnapshotCreateTime.Format(time.RFC3339))
+	if snapshot.SnapshotCreateTime != nil {
+		d.Set("snapshot_create_time", snapshot.SnapshotCreateTime.Format(time.RFC3339))
+	} else {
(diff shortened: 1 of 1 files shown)


atsushi-ishibashi:
This commit set `fmt.Sprintf("creating at %s", time.Now().UTC().Format(time.RFC3339))` as `snapshot_create_time` when the snapshot is still being created. But if `snapshot_create_time` is important, it should be better to wail until created.

aws/data_source_aws_db_snapshot.go
```diff
@@ -212,7 +212,11 @@ func dbSnapshotDescriptionAttributes(d *schema.ResourceData, snapshot *rds.DBSna
 	d.Set("source_region", snapshot.SourceRegion)
 	d.Set("status", snapshot.Status)
 	d.Set("vpc_id", snapshot.VpcId)
-	d.Set("snapshot_create_time", snapshot.SnapshotCreateTime.Format(time.RFC3339))
+	if snapshot.SnapshotCreateTime != nil {
+		d.Set("snapshot_create_time", snapshot.SnapshotCreateTime.Format(time.RFC3339))
+	} else {
+		d.Set("snapshot_create_time", fmt.Sprintf("creating at %s", time.Now().UTC().Format(time.RFC3339)))
+	}
```
Ninir: In the case where SnapshotCreateTime is nil, the attribute is set as a sentence containing the formatted time: this exposes a wrong information to the end user, and is not consistent with the other case, being just a formatted time.

I think we should only check for the case where it is not nil, because we have the correct snapshot creation time. THoughts? :)

atsushi-ishibashi: I agree with you. It's inconsistent.

PR #2184, opened 2017-11-06
resource/aws_elasticache_cluster: Added endpoints

lscheidler:
Problem
-----------

For an elasticache memcached cluster it is necessary to get a list of endpoints with their corresponding ports, that we can use in provisioning. At the moment it is not possible GH-67 to get a list from cache_nodes.

Solution
-----------

Added a new attribute endpoints, which contains a list of all endpoints with endpoint:port, which can be used in provisioner:

```
> ${join(aws_elasticache_cluster.bar.endpoints, ",")}
cluster-example.abc123.0001.euc1.cache.amazonaws.com:11211,cluster-example.abc123.0002.euc1.cache.amazonaws.com:11211,cluster-example.abc123.0003.euc1.cache.amazonaws.com:11211
```

This doesn't solve the general problem with cache_nodes from GH-67, but should be sufficient in most cases.

### DIFF
--- a/aws/resource_aws_elasticache_cluster.go
+++ b/aws/resource_aws_elasticache_cluster.go
@@ -176,6 +176,12 @@ func resourceAwsElasticacheCluster() *schema.Resource {
 		Computed: true,
 	}
 
+	resourceSchema["endpoints"] = &schema.Schema{
+		Type:     schema.TypeList,
+		Computed: true,
+		Elem:     &schema.Schema{Type: schema.TypeString},
+	}
+
 	resourceSchema["replication_group_id"] = &schema.Schema{
 		Type:     schema.TypeString,
 		Computed: true,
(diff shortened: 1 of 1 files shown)


bflad:
Hi @lscheidler 👋 Thanks for submitting this. Workarounds like these are no longer necessary in Terraform 0.12, which fully supports indexed splat (`*`) notation with list attribute values.

Given this configuration:

```hcl
terraform {
  required_providers {
    aws = "2.20.0"
  }
  required_version = "0.12.5"
}

provider "aws" {
  region = "us-east-2"
}

resource "aws_elasticache_cluster" "test" {
  cluster_id      = "testing"
  engine          = "memcached"
  node_type       = "cache.t2.micro"
  num_cache_nodes = 3
}

output "test" {
  value = join(",", formatlist("%s:%s", aws_elasticache_cluster.test.cache_nodes[*].address, aws_elasticache_cluster.test.cache_nodes[*].port))
}
```

We can get the expected output:

```
testing.asxu6y.0001.use2.cache.amazonaws.com:11211,testing.asxu6y.0002.use2.cache.amazonaws.com:11211,testing.asxu6y.0003.use2.cache.amazonaws.com:11211
```

Enjoy! 🚀

PR #2185, opened 2017-11-06
vendor: github.com/hashicorp/terraform/...@v0.11.0-beta1

radeksimko:
This is mainly to take advantage of https://github.com/hashicorp/terraform/pull/16506 and fix a couple of our acceptance tests which are currently crashing.


stack72:
@radeksimko is taking a vendor on not a beta a risky idea? Do we know what the feedback is on the beta? Or is this low key?

radeksimko:
Firstly I think we'll find out soon from the acceptance tests and if not, I don't know how else.
Secondly there's a very low chance of breaking anything because providers generally only vendor `hashicorp/terraform` because of all the helper (schema etc.) packages, which should be compatible - if not, we'll see it soon in acceptance tests and we can fix it before cutting next provider release 😉

PR #2191, opened 2017-11-06
"SetSubnets is not supported for load balancers of type 'network'

sandromolchanov:
Error message "Failure Setting ALB Subnets: InvalidConfigurationRequest: SetSubnets is not supported for load balancers of type 'network'"
appears, because it is impossible to update the subnets for Network Load Balancer.


wheelybird:
This would be a bug fix, not an enhancement.

catsby:
Marking this as a duplicate of https://github.com/terraform-providers/terraform-provider-aws/pull/1941 , where I've identified 2 things to add in addition but should get resolved this week.

Thank you!

PR #2197, opened 2017-11-07
d/aws_ecs_container_definition: add port mappings to output

ValFadeev:
This adds container port mappings to data source output. One use case for this is creating lb target groups based on exposed ports:

```hcl
variable "task_definition" {}
variable "container_name" {}
variable "vpc_id" {}

data "aws_ecs_task_definition" "task" {
  task_definition = "${var.task_definition}"
}

data "aws_ecs_container_definition" "container" {
  task_definition = "${data.aws_ecs_task_definition.task.id}"
  container_name  = "${var.container_name}"
}

resource "aws_lb_target_group" "test" {
  count    = "${length(data.aws_ecs_container_definition.container.port_mappings)}"
  name     = "tf-example-lb-tg"
  port     = "${lookup(data.aws_ecs_container_definition.container.port_mappings[count.index], "host_port")}"
  protocol = "HTTP"
  vpc_id   = "${var.vpc_id}"
}
```

### DIFF
--- a/aws/data_source_aws_ecs_container_definition.go
+++ b/aws/data_source_aws_ecs_container_definition.go
@@ -59,6 +59,11 @@ func dataSourceAwsEcsContainerDefinition() *schema.Resource {
 				Computed: true,
 				Elem:     schema.TypeString,
 			},
+			"port_mappings": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeMap, Elem: schema.TypeString},
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


aws/data_source_aws_ecs_container_definition.go
```diff
@@ -93,10 +98,20 @@ func dataSourceAwsEcsContainerDefinitionRead(d *schema.ResourceData, meta interf
...
 
 		var environment = map[string]string{}
-		for _, keyValuePair := range def.Environment {
-			environment[aws.StringValue(keyValuePair.Name)] = aws.StringValue(keyValuePair.Value)
+		for _, envKeyValuePair := range def.Environment {
+			environment[aws.StringValue(envKeyValuePair.Name)] = aws.StringValue(envKeyValuePair.Value)
 		}
 		d.Set("environment", environment)
+
+		var portMappings = make([]map[string]string, 0)
```
Ninir: Can you change this so that it has a fixed and expected size, based on def.PortMappings?
As in: `var portMappings = make([]map[string]string, len(def.PortMappings), len(def.PortMappings))`

aws/data_source_aws_ecs_container_definition.go
```diff
@@ -59,6 +59,11 @@ func dataSourceAwsEcsContainerDefinition() *schema.Resource {
 				Computed: true,
 				Elem:     schema.TypeString,
 			},
+			"port_mappings": &schema.Schema{
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeMap, Elem: schema.TypeString},
```
Ninir: Since we know the structure, it would be better to set the schema rather than allowing unknown strings.
This would make us update the acc test also, to check given attributes :)

ValFadeev:
Hi @Ninir, thanks for your comments! It is definitely better to enforce the schema. Please kindly check if it's ok now.

PR #2202, opened 2017-11-07
r/redshift_cluster: Make master_username ForceNew

radeksimko:
This field is not updatable per API docs: http://docs.aws.amazon.com/redshift/latest/APIReference/API_ModifyCluster.html

## Test Results
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSRedshiftCluster_forceNewUsername -timeout 120m
=== RUN   TestAccAWSRedshiftCluster_forceNewUsername
--- PASS: TestAccAWSRedshiftCluster_forceNewUsername (1449.60s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1449.644s
```

### DIFF
--- a/aws/resource_aws_redshift_cluster_test.go
+++ b/aws/resource_aws_redshift_cluster_test.go
@@ -57,7 +57,7 @@ func TestAccAWSRedshiftCluster_basic(t *testing.T) {
 	var v redshift.Cluster
 
 	ri := rand.New(rand.NewSource(time.Now().UnixNano())).Int()
-	config := fmt.Sprintf(testAccAWSRedshiftClusterConfig_basic, ri)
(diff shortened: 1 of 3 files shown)


aws/resource_aws_redshift_cluster_test.go
```diff
@@ -320,6 +320,37 @@ func TestAccAWSRedshiftCluster_tags(t *testing.T) {
 	})
 }
 
+func TestAccAWSRedshiftCluster_forceNewUsername(t *testing.T) {
+	var v redshift.Cluster
```
tombuildsstuff: _minor_ as you're storing this you could compare the old and new clusters to ensure it's been recreated?

Issue #1853, opened 2017-10-10
ALB SNI Support


oarmstrong:
[Announced today](https://aws.amazon.com/about-aws/whats-new/2017/10/elastic-load-balancing-application-load-balancers-now-support-multiple-ssl-certificates-and-smart-certificate-selection-using-server-name-indication-sni/), is support for SNI with application load balancers. This allows for multiple certificates to be attached to the load balancer and SNI will automatically select the correct certificate.

Would it be possible to support this functionality?  I imagine we'd want to have an additional option in the `aws_alb_listener` target to take a list of certificates to attach.  There does still need to be a default certificate, it may be better to take this as the first item in the certificate list though and just have a single option.

I may take a look at this myself in a few days if nobody else gets around to it in the meantime.

ebekker:
The API does treat the *additional* certs distinctly from the *default* cert, so it may be easier to implement if you mirror this relationship.  The other thing is that it's possible to actually specify the same cert twice, once as the default and once as an additional cert -- in their UI, they actually recommend doing this if you're going to "swap out" one default for another in order to maintain continuity during the transition.

ebekker:
Similar to the way SG rules are added to security groups, it may also be useful to support defining additional certs as either *inline* or as separate resources -- for example, in the former case you know all of your certs up front, or in the latter, they may be added on as part of separate configurations where each add-on cert is more closely tied to other resources.

oarmstrong:
I can see keeping the `certificate_arn` argument then and adding an additional `additional_certificate_arns` argument which would be a list of ARNs for additional certificates then. Makes sense to keep it inline with how the API treats it and to be able to neatly follow the recommendation you found of adding the default certificate as an additional certificate too (although I wasn't able to see that in the UI anywhere).

Great idea regarding another resource for adding certificates. I'd definitely get use out of that.

ebekker:
It's in the console UI for managing the additional certs, if you click on the default cert to expand this message:

![image](https://user-images.githubusercontent.com/3308726/31408637-3b8b2686-add7-11e7-8c6e-8fef132925e5.png)

oarmstrong:
Cheers. I didn't actually notice it was expandable at first.

nubbthedestroyer:
+1 this would be useful

mindnuts:
:+1:  

[...]

PR #2203, opened 2017-11-07
r/aws_lb_listener: add support for additional SSL certificates

oarmstrong:
This PR is for #1853.

We add an additional argument (`additional_certificate_arns`) to the `aws_lb_listener` resource. This argument allows for adding additional non-default certificates to a listener.

I'm well aware that there are currently no tests or documentation and that still needs to be done. It has been manually tested with as many cases as I could think of though. The existing test suite still runs, but I have not run the acceptance tests at all.

If at all possible, some feedback on the general style of my code would be great. Variable naming and error checking are two points that I've been particularly bad at in the past. I'm also not sure about defining https://github.com/terraform-providers/terraform-provider-aws/pull/2203/files#diff-6f5ff88b797d88c79786bcdb04a23eefR314 in line like that. Do we have an existing function somewhere that does the same job or should I move this elsewhere?

Another suggestion in the linked issue is for it to be possible to define additional certificates as a separate resource - must like security group rules. I'm not too sure if this design would allow for that?

As a final question, how would you like the commits before this can be merged in? Squished? Some kind of prefix to the messages?

Appreciate your time. Hopefully there isn't anything too major wrong with my code and I can get the documentation and tests done quickly too.

### DIFF
--- a/aws/resource_aws_lb_listener.go
+++ b/aws/resource_aws_lb_listener.go
@@ -154,7 +154,39 @@ func resourceAwsLbListenerCreate(d *schema.ResourceData, meta interface{}) error
 		return errors.New("Error creating LB Listener: no listeners returned in response")
 	}
 
-	d.SetId(*resp.Listeners[0].ListenerArn)
+	larn := *resp.Listeners[0].ListenerArn
+
+	err = resource.Retry(2*time.Minute, func() *resource.RetryError {
+		if carns, ok := d.GetOk("additional_certificate_arns"); ok {
+			for _, carn := range carns.([]string) {
+				_, err := elbconn.AddListenerCertificates(&elbv2.AddListenerCertificatesInput{
+					Certificates: []*elbv2.Certificate{&elbv2.Certificate{
+						CertificateArn: aws.String(carn),
+						IsDefault:      aws.Bool(false),
+					}},
+					ListenerArn: aws.String(larn),
+				})
+				if awsErr, ok := err.(awserr.Error); ok {
+					if awsErr.Code() == "TooManyCertificates" || awsErr.Code() == "CertificateNotFound" {
+						log.Printf("[WARN] Got an error while trying to add certificate to listener with ARN: %s: %s", lbArn, err)
+						return resource.RetryableError(err)
+					}
+				}
+
+				if err != nil {
+					return resource.NonRetryableError(err)
+				}
+			}
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return errwrap.Wrapf("Error adding certificates to LB Listener: {{err}}", err)
+	}
+
+	d.SetId(larn)
 
 	return resourceAwsLbListenerRead(d, meta)
 }
--- a/aws/resource_aws_lb_listener.go
+++ b/aws/resource_aws_lb_listener.go
@@ -162,7 +162,6 @@ func resourceAwsLbListenerCreate(d *schema.ResourceData, meta interface{}) error
 				_, err := elbconn.AddListenerCertificates(&elbv2.AddListenerCertificatesInput{
 					Certificates: []*elbv2.Certificate{&elbv2.Certificate{
 						CertificateArn: aws.String(carn),
-						IsDefault:      aws.Bool(false),
 					}},
 					ListenerArn: aws.String(larn),
 				})
--- a/aws/resource_aws_lb_listener.go
+++ b/aws/resource_aws_lb_listener.go
@@ -291,6 +291,62 @@ func resourceAwsLbListenerUpdate(d *schema.ResourceData, meta interface{}) error
 		return errwrap.Wrapf("Error modifying LB Listener: {{err}}", err)
 	}
 
+	var carnList []*string
+	if carns, ok := d.GetOk("additional_certificate_arns"); ok {
+		carnList = expandStringList(carns.(*schema.Set).List())
+	}
+
+	awsCerts, err := elbconn.DescribeListenerCertificates(&elbv2.DescribeListenerCertificatesInput{
+		ListenerArn: aws.String(d.Id()),
+	})
+	if err != nil {
+		return errwrap.Wrapf("Error retrieving ListenerCertficates: {{err}}", err)
+	}
+	var awsCarnList []*string
+	if awsCerts.Certificates != nil {
+		for _, cert := range awsCerts.Certificates {
+			if !*cert.IsDefault {
+				awsCarnList = append(awsCarnList, cert.CertificateArn)
+			}
+		}
+	}
+
+	stringInSlice := func(a *string, list []*string) bool {
+		for _, b := range list {
+			if b == a {
+				return true
+			}
+		}
+		return false
+	}
+
+	for _, carn := range awsCarnList {
+		if !stringInSlice(carn, carnList) {
+			_, err := elbconn.RemoveListenerCertificates(&elbv2.RemoveListenerCertificatesInput{
+				Certificates: []*elbv2.Certificate{&elbv2.Certificate{
+					CertificateArn: carn,
+				}},
+				ListenerArn: aws.String(d.Id()),
+			})
+			if err != nil {
+				return errwrap.Wrapf("Error modifying LB Listener: {{err}}", err)
+			}
+		}
+	}
+	for _, carn := range carnList {
+		if !stringInSlice(carn, awsCarnList) {
+			_, err := elbconn.AddListenerCertificates(&elbv2.AddListenerCertificatesInput{
+				Certificates: []*elbv2.Certificate{&elbv2.Certificate{
+					CertificateArn: carn,
+				}},
+				ListenerArn: aws.String(d.Id()),
+			})
+			if err != nil {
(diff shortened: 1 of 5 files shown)


oarmstrong:
Documentation written. I can also confirm that the acceptance suit for this resource passes. Currently writing the updated tests for this argument now.

oarmstrong:
That's just added the tests.

I've also modified the corresponding data source, not sure if you'd prefer that in a separate PR though.

oarmstrong:
It may make more sense to rename the original `certificate_arn` argument to `default_certificate_arn` as the API and documentation seems to refer to it as that. Not sure how you'd feel about the breaking change though.

ozbillwang:
@oarmstrong 

Thanks to raise this PR to add this feature. When will it be merged?

Regarding argument naming, I am fine for  `additional_certificate_arns` more than to rename `certificate_arn` to `default_certificate_arn` to keep version compatible.

oarmstrong:
> When will it be merged?

It's good to go from me. Just awaiting feedback from the maintainers.

ozbillwang:
Seems we missed the release in 0.11, hope to get it merged in next release.

Puneeth-n:
@ozbillwang Once, the PR is merged, the maintainers update the CHANGELOG

tatsuo48:
Is there anyone who knows when to merge?
I'm waiting this merge 😢

oarmstrong:
I'm still not 100% with it and assume it will need some polish. Just want a bit of a review at this point.

Puneeth-n:
@radeksimko @Ninir any idea when this will be merged :)

oarmstrong:
I'm able to get on any review points pretty much immediately!

aws/resource_aws_lb_listener.go
```diff
@@ -147,7 +154,38 @@ func resourceAwsLbListenerCreate(d *schema.ResourceData, meta interface{}) error
...
+		if carns, ok := d.GetOk("additional_certificate_arns"); ok {
+			for _, carn := range expandStringList(carns.(*schema.Set).List()) {
+				_, err := elbconn.AddListenerCertificates(&elbv2.AddListenerCertificatesInput{
+					Certificates: []*elbv2.Certificate{&elbv2.Certificate{
+						CertificateArn: carn,
+					}},
+					ListenerArn: aws.String(larn),
+				})
+				if awsErr, ok := err.(awserr.Error); ok {
+					if awsErr.Code() == "TooManyCertificates" || awsErr.Code() == "CertificateNotFound" {
```
radeksimko: What is the reason for retrying on `TooManyCertificates` here? 🤔

aws/resource_aws_lb_listener.go
```diff
@@ -147,7 +154,38 @@ func resourceAwsLbListenerCreate(d *schema.ResourceData, meta interface{}) error
...
+					return resource.NonRetryableError(err)
+				}
+			}
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return errwrap.Wrapf("Error adding certificates to LB Listener: {{err}}", err)
```
radeksimko: `errwrap` is typically useful in situations where we need to keep track of the error type. In this case error will only ever be escalated to the UI where all that matters is strings ("stringifyed" error message), not the original type.

In other words `fmt.Errorf()` will do the job, IMO.

aws/resource_aws_lb_listener.go
```diff
@@ -183,6 +221,23 @@ func resourceAwsLbListenerRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("certificate_arn", listener.Certificates[0].CertificateArn)
 	}
 
+	certs, err := elbconn.DescribeListenerCertificates(&elbv2.DescribeListenerCertificatesInput{
+		ListenerArn: aws.String(d.Id()),
+	})
+	if err != nil {
+		return errwrap.Wrapf("Error retrieving ListenerCertficates: {{err}}", err)
```
radeksimko: `errwrap` is typically useful in situations where we need to keep track of the error type. In this case error will only ever be escalated to the UI where all that matters is strings ("stringifyed" error message), not the original type.

In other words `fmt.Errorf()` will do the job, IMO.

aws/resource_aws_lb_listener.go
```diff
@@ -236,6 +291,62 @@ func resourceAwsLbListenerUpdate(d *schema.ResourceData, meta interface{}) error
...
+	var carnList []*string
+	if carns, ok := d.GetOk("additional_certificate_arns"); ok {
+		carnList = expandStringList(carns.(*schema.Set).List())
+	}
+
+	awsCerts, err := elbconn.DescribeListenerCertificates(&elbv2.DescribeListenerCertificatesInput{
+		ListenerArn: aws.String(d.Id()),
+	})
+	if err != nil {
+		return errwrap.Wrapf("Error retrieving ListenerCertficates: {{err}}", err)
```
radeksimko: `errwrap` is typically useful in situations where we need to keep track of the error type. In this case error will only ever be escalated to the UI where all that matters is strings ("stringifyed" error message), not the original type.

In other words `fmt.Errorf()` will do the job, IMO.

aws/resource_aws_lb_listener.go
```diff
@@ -236,6 +291,62 @@ func resourceAwsLbListenerUpdate(d *schema.ResourceData, meta interface{}) error
...
+	for _, carn := range awsCarnList {
+		if !stringInSlice(carn, carnList) {
+			_, err := elbconn.RemoveListenerCertificates(&elbv2.RemoveListenerCertificatesInput{
+				Certificates: []*elbv2.Certificate{&elbv2.Certificate{
+					CertificateArn: carn,
+				}},
+				ListenerArn: aws.String(d.Id()),
+			})
+			if err != nil {
+				return errwrap.Wrapf("Error modifying LB Listener: {{err}}", err)
```
radeksimko: `errwrap` is typically useful in situations where we need to keep track of the error type. In this case error will only ever be escalated to the UI where all that matters is strings ("stringifyed" error message), not the original type.

In other words `fmt.Errorf()` will do the job, IMO.

aws/resource_aws_lb_listener.go
```diff
@@ -236,6 +291,62 @@ func resourceAwsLbListenerUpdate(d *schema.ResourceData, meta interface{}) error
...
+	for _, carn := range carnList {
+		if !stringInSlice(carn, awsCarnList) {
+			_, err := elbconn.AddListenerCertificates(&elbv2.AddListenerCertificatesInput{
+				Certificates: []*elbv2.Certificate{&elbv2.Certificate{
+					CertificateArn: carn,
+				}},
+				ListenerArn: aws.String(d.Id()),
+			})
+			if err != nil {
+				return errwrap.Wrapf("Error modifying LB Listener: {{err}}", err)
```
radeksimko: `errwrap` is typically useful in situations where we need to keep track of the error type. In this case error will only ever be escalated to the UI where all that matters is strings ("stringifyed" error message), not the original type.

In other words `fmt.Errorf()` will do the job, IMO.

Puneeth-n:
@radeksimko Great point. I was thinking on similar lines. However with `aws_security_group_rule` how does it behave when I am adding and removing the same rule under different "resource names" ? does terraform error out or does terraform detect that we are adding the same  rule twice under different names?

oarmstrong:
Thanks for the review. I'll get on to it a little later.

aws/resource_aws_lb_listener.go
oarmstrong: Only as I essentially copied `CertificateNotFound`. Neither makes sense to retry on IMO.

aws/resource_aws_lb_listener.go
oarmstrong: Right. I'm with you. Comment noted for all instances.

oarmstrong:
So, the reason for this resource. Basically I was attempting to get this done quickly and simply so editing an existing resource felt much easier for me. It was noted in the issue that a separate resource would be useful too and I agreed 100%. I wasn't aware of how impossible it was to have this resource as well as a separate resource (e.g. the security group case).

It's probably best to do it right from the outset then and just abandon this attempt and go straight for the new resource. If that's your opinion for the "correct" implementation (and I tend to agree with that too).

I can take a crack at the new resource but I may need guidance. Would you prefer to close this PR and for myself to start a new one when I have a first attempt?  I'll reuse a lot of the code from this attempt though as it seems to have passed your initial review.

Thanks for your time!

/cc @radeksimko

radeksimko:
> I can take a crack at the new resource but I may need guidance.

Sure, feel free to ask. Just FYI I will be mostly unavailable from the end of this week until the EOY.

> Would you prefer to close this PR and for myself to start a new one when I have a first attempt? I'll reuse a lot of the code from this attempt though as it seems to have passed your initial review.

Yep, the code looked good - you can reuse most of it and yes it's probably best to close this PR.

If you need some guidance, take a look at any merged `new-resource` PR, e.g. https://github.com/terraform-providers/terraform-provider-aws/pull/2154/files - also note the docs updates.

oarmstrong:
Cheers. I'll close this PR and open a new one with the revised method.

PR #2213, opened 2017-11-08
Dynamic DNS records count in aws_ses_domain_dkim example

pracucci:
The example for `aws_ses_domain_dkim` DNS records had a static `count = 3` value. It's correct, since AWS SES DKIM currently uses 3 tokens, but I'm wondering if it would be more correct to have it dynamic.

### DIFF
--- a/website/docs/r/ses_domain_dkim.html.markdown
+++ b/website/docs/r/ses_domain_dkim.html.markdown
@@ -41,7 +41,7 @@ resource "aws_ses_domain_dkim" "example" {
 }
 
 resource "aws_route53_record" "example_amazonses_verification_record" {
-  count   = 3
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @pracucci 
thanks for the PR.

This is an ideal config, but doesn't actually work at this point. You'll get the following error if you try applying it.

```
Error: Error running plan: 1 error(s) occurred:

* aws_route53_record.example_amazonses_verification_record: aws_route53_record.example_amazonses_verification_record: value of 'count' cannot be computed
```

We have [some plans](https://github.com/hashicorp/terraform/issues/4149) for allowing this behaviour. We don't have any timelines to share for the mentioned improvements so I'm going to close this PR for now.

I hope the reasoning makes sense - let me know if it doesn't.

PR #2219, opened 2017-11-08
f/instance d/instance: Get (Windows) password data

deftflux:
This is one half of the puzzle which was previously https://github.com/hashicorp/terraform/pull/5675.

This addresses issue https://github.com/hashicorp/terraform/issues/3148 which was moved to https://github.com/terraform-providers/terraform-provider-aws/issues/30.

Basically, when an instance is read from AWS (either an instance resource or data source), you now have the option to read the encrypted password data as well.  Since not all instances will have password data and it takes time for that data to become available, retrieving this data is optional based on the value of `get_password_data`, which defaults to `false`.  When true, a `password_data` attribute is populated with the encrypted data.

This PR is ready to merge, containing updated docs and working acceptance tests.  I chose the latest Windows Server 2016 Core image for the tests, since it boots the fastest (and thus the password data becomes available the quickest).

The other half of the puzzle will be decrypting this password data.  This will involve implementing built-in functions in Terraform proper to enable this decryption using an interpolated string.

### DIFF
--- a/aws/data_source_aws_instance_test.go
+++ b/aws/data_source_aws_instance_test.go
@@ -216,52 +216,22 @@ func TestAccAWSInstanceDataSource_VPCSecurityGroups(t *testing.T) {
 	})
 }
 
-func TestAccAWSInstanceDataSource_getPasswordData_true(t *testing.T) {
-	resource.Test(t, resource.TestCase{
-		PreCheck:  func() { testAccPreCheck(t) },
-		Providers: testAccProviders,
-		Steps: []resource.TestStep{
-			{
-				Config: testAccInstanceDataSourceConfig_getPasswordData(true),
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.aws_instance.foo", "get_password_data", "true"),
-					resource.TestCheckResourceAttrSet("data.aws_instance.foo", "password_data"),
-				),
-			},
-		},
-	})
-}
-
-func TestAccAWSInstanceDataSource_getPasswordData_false(t *testing.T) {
-	resource.Test(t, resource.TestCase{
-		PreCheck:  func() { testAccPreCheck(t) },
-		Providers: testAccProviders,
-		Steps: []resource.TestStep{
-			{
-				Config: testAccInstanceDataSourceConfig_getPasswordData(false),
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.aws_instance.foo", "get_password_data", "false"),
-					resource.TestCheckNoResourceAttr("data.aws_instance.foo", "password_data"),
-				),
-			},
-		},
-	})
-}
-
 func TestAccAWSInstanceDataSource_getPasswordData_trueToFalse(t *testing.T) {
+	rInt := acctest.RandInt()
+
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccInstanceDataSourceConfig_getPasswordData(true),
+				Config: testAccInstanceDataSourceConfig_getPasswordData(true, rInt),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("data.aws_instance.foo", "get_password_data", "true"),
 					resource.TestCheckResourceAttrSet("data.aws_instance.foo", "password_data"),
 				),
 			},
 			{
-				Config: testAccInstanceDataSourceConfig_getPasswordData(false),
+				Config: testAccInstanceDataSourceConfig_getPasswordData(false, rInt),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("data.aws_instance.foo", "get_password_data", "false"),
 					resource.TestCheckNoResourceAttr("data.aws_instance.foo", "password_data"),
@@ -272,19 +242,21 @@ func TestAccAWSInstanceDataSource_getPasswordData_trueToFalse(t *testing.T) {
 }
 
 func TestAccAWSInstanceDataSource_getPasswordData_falseToTrue(t *testing.T) {
+	rInt := acctest.RandInt()
+
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccInstanceDataSourceConfig_getPasswordData(false),
+				Config: testAccInstanceDataSourceConfig_getPasswordData(false, rInt),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("data.aws_instance.foo", "get_password_data", "false"),
 					resource.TestCheckNoResourceAttr("data.aws_instance.foo", "password_data"),
 				),
 			},
 			{
-				Config: testAccInstanceDataSourceConfig_getPasswordData(true),
+				Config: testAccInstanceDataSourceConfig_getPasswordData(true, rInt),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("data.aws_instance.foo", "get_password_data", "true"),
 					resource.TestCheckResourceAttrSet("data.aws_instance.foo", "password_data"),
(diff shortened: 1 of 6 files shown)


smastrorocco:
+1 Would really like to see this merged

deftflux:
The second half of this is ready now: https://github.com/hashicorp/terraform/pull/16647

deftflux:
Can I get a merge here?  The necessary `rsadecrypt` function has been merged.  Once this PR is merged and both are released, we can do secure Windows provisioning in AWS like this:

```hcl
resource "aws_instance" "foo" {
  get_password_data = true
  ...

  connection {
    password = "${rsadecrypt(self.password_data, file("/path/to/private_key.pem"))}"
    ...
  }
}
```

And/or:

```hcl
resource "null_resource" "foo" {
  ...

  connection {
    password = "${rsadecrypt(aws_instance.foo.password_data, file("/path/to/private_key.pem"))}"
    ...
  }
}
```

Or:

```hcl
data "aws_instance" "foo" {
  get_password_data = true
  ...
}

resource "null_resource" "foo" {
  ...

  connection {
    password = "${rsadecrypt(data.aws_instance.foo.password_data, file("/path/to/private_key.pem"))}"
    ...
  }
}
```

jakemig:
I'd also love to see this functionality in the AWS provider.

itsamenathan:
I would love to see this as well.  Looks like the other half of this merge has landed in core as of yesterday.

grig-tar:
When you planning to merge this PR?

jeremygaither:
I also need this...

aws/resource_aws_instance.go
```diff
@@ -1528,6 +1543,46 @@ func readSecurityGroups(d *schema.ResourceData, instance *ec2.Instance, conn *ec
...
+		if err := d.Set("password_data", nil); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	log.Printf("[INFO] Reading password data for instance %s", d.Id())
+
+	timeout := time.Now().Add(15 * time.Minute)
```
bflad: Nitpick: we have the `resource.Retry` function we regularly use for these type of retryable operations:

```go
var passwordData string

input := &ec2.GetPasswordDataInput{
  InstanceId: aws.String(instanceID),
}

err := resource.Retry(15 * time.Minute, func() *resource.RetryError {
  out, err := conn.GetPasswordData(input)

  if err != nil {
  }

  if resp.PasswordData == nil || *resp.PasswordData == "" {
    return resource.RetryableError(fmt.Errorf("Password data is blank for instance ID: %s", instanceID))
  }

  passwordData = strings.TrimSpace(*resp.PasswordData)
  return nil
})

if err != nil {
  return "", err
}

return passwordData, nil
```

aws/resource_aws_instance.go
```diff
@@ -1528,6 +1543,46 @@ func readSecurityGroups(d *schema.ResourceData, instance *ec2.Instance, conn *ec
 	return nil
 }
 
+func readPasswordData(d *schema.ResourceData, instance *ec2.Instance, conn *ec2.EC2) error {
```
bflad: We should try to reduce this function's signature to one action and leave the state management outside of this function, e.g. 

```go
func getAwsEc2InstancePasswordData(instanceID string, conn *ec2.EC2) (string, error)
```

aws/resource_aws_instance.go
```diff
@@ -1528,6 +1543,46 @@ func readSecurityGroups(d *schema.ResourceData, instance *ec2.Instance, conn *ec
...
+	if !d.Get("get_password_data").(bool) {
+		if err := d.Set("password_data", nil); err != nil {
```
bflad: When using `d.Set()` with `*string`, we do not need the additional `err` checking. 👍

aws/resource_aws_instance_test.go
```diff
@@ -2719,3 +2821,21 @@ resource "aws_instance" "foo" {
   }
 }`, rInt, rInt)
 }
+
+func testAccInstanceConfig_getPasswordData(val bool) string {
+	return fmt.Sprintf(`
+	resource "aws_key_pair" "foo" {
+		key_name = "tf-acc-winpasswordtest"
```
bflad: The key name needs to be randomized for parallel or broken test runs. `key_name = "tf-acc-%d"` (where `%d` can come from the top level test function, e.g. `rInt := acctest.RandInt()`

aws/data_source_aws_instance_test.go
```diff
@@ -502,3 +580,25 @@ data "aws_instance" "foo" {
   instance_id = "${aws_instance.foo_instance.id}"
 }
 `
+
+func testAccInstanceDataSourceConfig_getPasswordData(val bool) string {
+	return fmt.Sprintf(`
+	resource "aws_key_pair" "foo" {
+		key_name = "tf-acc-winpasswordtest"
```
bflad: The key name needs to be randomized for parallel or broken test runs. `key_name = "tf-acc-%d"` (where `%d` can come from the top level test function, e.g. `rInt := acctest.RandInt()`

aws/data_source_aws_instance_test.go
```diff
@@ -216,6 +216,84 @@ func TestAccAWSInstanceDataSource_VPCSecurityGroups(t *testing.T) {
 	})
 }
 
+func TestAccAWSInstanceDataSource_getPasswordData_true(t *testing.T) {
```
bflad: Can we reduce these down to two acceptance tests please? true -> false and false -> true should be sufficient

aws/resource_aws_instance_test.go
```diff
@@ -1288,6 +1288,108 @@ func TestAccAWSInstance_associatePublic_overridePrivate(t *testing.T) {
 	})
 }
 
+func TestAccAWSInstance_getPasswordData_true(t *testing.T) {
```
bflad: Can we reduce this down to two acceptance tests please? false -> true and true -> false should be sufficient

website/docs/r/instance.html.markdown
```diff
@@ -64,6 +64,7 @@ instance. Amazon defaults this to `stop` for EBS-backed instances and
 instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#Using_ChangingInstanceInitiatedShutdownBehavior) for more information.
 * `instance_type` - (Required) The type of instance to start
 * `key_name` - (Optional) The key name to use for the instance.
+* `get_password_data` - (Optional) If true, wait for password data to become available and retrieve it.  The password data is exported to the `password_data` attribute.
```
bflad: For all mentions, we should specifically add `Microsoft Windows` in the documentation and potentially link to the API documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetPasswordData.html

joeabbey:
👍 Would love to see this merged

deftflux:
Thanks for the feedback, @bflad.  It took a while to get around to making the changes and testing, but it's done now.  Please review.

discentem:
This might be a little off-topic, but does terraform support winrm over https?

bflad:
@discentem: There is a `https` flag for WinRM connections: https://www.terraform.io/docs/provisioners/connection.html#https

deftflux:
@discentem a little off-topic, but yes.  However, it does require some magic sauce in the user data to set up WinRM over HTTPS on the instance, unless you're using an AMI that already has this set up.  The built-in Windows AMI's provided by Amazon do not have WinRM set up at all.  Setting up HTTPS also requires generating a certificate within that user data script.

Here's some magic sauce:

```
<powershell>

echo "Generate certificate for WinRM"
$cert = New-SelfSignedCertificate -CertStoreLocation cert:\LocalMachine\My -DnsName $env:computername
$thumb = $cert.Thumbprint

echo "Enable WinRM"
winrm quickconfig -q
winrm set winrm/config '@{MaxTimeoutms="1800000"}'
winrm set winrm/config/winrs '@{MaxMemoryPerShellMB="300"}'
winrm set winrm/config/service/auth '@{Basic="true"}'
winrm create winrm/config/listener?Address=*+Transport=HTTPS "@{CertificateThumbprint=`"$thumb`";Port=`"5986`"}"

echo "Enable WinRM firewall rules"
if ((get-service -DisplayName "Windows Firewall").StartType -ne "Disabled") {
	Remove-NetFirewallRule -DisplayName "WinRM" -ErrorAction SilentlyContinue
	New-NetFirewallRule -DisplayName "WinRM" -Direction Inbound -Protocol TCP -LocalPort 5986 | Out-Null
}

echo "Restart WinRM service"
Stop-Service winrm
Set-Service winrm -StartupType Automatic
Start-Service winrm

echo "Finished PowerShell script"

</powershell>
```

Same thing is required when using Packer to build an image.  Also, since this uses a self-signed cert, the `insecure` flag is required as well.

deftflux:
@bflad Not sure how to bounce this back to you, but I made the requested change.  The tests should pass now.

deftflux:
👍 🎉

jeremygaither:
@deftflux THANK YOU!

🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

discentem:
Anyone willing to post instructions on overriding the released aws provider with this? I built it, dropped it into `~/.terraform.d/plugins/`, ran terraform init (no changes), but terraform still doesn't recognize `get_password_data`.

discentem:
I pieced it together. A bit complicated though. It seems terraform doesn't respect a higher z value when parsing plugin_v**x.y.z**. Had to bump y by 1.

bflad:

discentem:
Thanks @bflad. I found the docs for specifying the version in the provider code 👍

bodgit:
I don't think this was implemented for spot requests so it's breaks if you try and spin up a Windows instance as a spot request and try and access the password data. I'll see about creating a PR to fix this.

Midacts:
Does anyone know if this is documented anywhere? Maybe in the `aws_instance` resource docs?
The solution to this problem was not easy to find.

PR #2224, opened 2017-11-09
New Data Source: aws_kms_key

### LINKED ISSUE #2009: Support ARNs for existing KMS keys
There is currently no way to directly retrieve an ARN for a pre-existing KMS Key. This makes it difficult / unnecessarily complicated to assign existing KMS keys to newly created resources.

### Terraform Version
0.10.7

### Affected Resource(s)
- `aws_kms_alias`
- `aws_lambda_function`
- Any resource which accepts a `kms_key_arn` argument

```hcl
data "aws_kms_alias" "existing_cmk" {                                                                                                                                                                         
  name = "alias/MyPreExistingKey"                                                                                                                                                                        
}

resource "aws_lambda_function" "new_function" {
  /* other arguments ommitted */
  kms_key_arn = "${data.aws_kms_alias.existing_cmk.arn}"
}  
```
### Expected Behavior
The newly created Lambda function would have the `MyPreExisting` KMS key set.

### Actual Behavior
The Lambda function gets assigned either no KMS key, or the default `aws/lambda` key.

### Important Factoids
After failing to have the expected KMS key assigned, I looked at the plan output and saw that the value of `${data.aws_kms_alias.existing_cmk.arn}` was in the format of 

```arn:aws:kms:eu-west-1:123456789012:alias/MyExistingKey```

This is the ARN of the alias, not the key itself, which are seemingly not analogous. Logically this makes sense, but as there is no `aws_kms_key` data provider, there is no way to directly retrieve an ARN for an existing key.

You *can* indirectly solve this by using some substitution, but it would be nice not to have to do this:

```hcl
locals {                                                                                                                                                                                                    
  kms_arn = "${replace(data.aws_kms_alias.existing_cmk.arn, data.aws_kms_alias.existing_cmk.name, "key/${data.aws_kms_alias.existing_cmk.target_key_id}")}"                                                       
}  
```

trung:
To fix #2009


trung:
```
❯ make testacc TESTARGS="-run=TestAccDataSourceAwsKmsKey"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDataSourceAwsKmsKey -timeout 120m
=== RUN   TestAccDataSourceAwsKmsKey
--- PASS: TestAccDataSourceAwsKmsKey (48.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	48.937s
```

bflad:
While I certainly see the value here, it seems it would also be fairly simple to provide a `target_key_arn` on the existing `aws_kms_alias` data source by substituting the target key ID into the alias ARN. We can get away with this since KMS aliases and keys are bound to the same partition, account, and region.

See also [CreateAlias documentation](http://docs.aws.amazon.com/sdk-for-go/api/service/kms/#KMS.CreateAlias):
> The alias and the CMK it is mapped to must be in the same AWS account and the same region. You cannot perform this operation on an alias in a different AWS account.

aws/validators.go
```diff
@@ -1972,3 +1972,20 @@ func validateDxConnectionBandWidth(v interface{}, k string) (ws []string, errors
 	errors = append(errors, fmt.Errorf("expected %s to be one of %v, got %s", k, validBandWidth, val))
 	return
 }
+
+func validateKmsKey(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	arnPrefixPattern := `arn:[\w-]+:([a-zA-Z0-9\-])+:([a-z]{2}-(gov-)?[a-z]+-\d{1})?:(\d{12})?:`
```
bflad: Can we switch this to not assume anything about the format of AWS partitions, regions, and account IDs in the future please? `arn:[^:]+:kms:[^:]+:[^:]+:`

website/docs/d/kms_key.html.markdown
```diff
@@ -0,0 +1,59 @@
+---
+layout: "aws"
+page_title: "AWS: aws_kms_key"
+sidebar_current: "docs-aws-datasource-kms-key"
+description: |-
+  Get information on a AWS Key Management Service (KMS) Key
+---
+
+# aws\_kms\_key
```
bflad: Nitpick: Backslashes are no longer required in the documentation headers

trung:
@bflad thanks for the review
I'm on Lunar New Year celebration now so i won't be able to make changes in the next couple of days. If this is not hurry, i can certainly work on it sometime next week

bflad:
@trung No worries at all, please enjoy your holiday! 🎉

trung:
@bflad I've made some changes per your review

```
❯ make testacc TESTARGS="-run=TestAccDataSourceAwsKmsKey_basic"
TF_ACC=1 go test ./... -v -run=TestAccDataSourceAwsKmsKey_basic -timeout 120m
=== RUN   TestAccDataSourceAwsKmsKey_basic
--- PASS: TestAccDataSourceAwsKmsKey_basic (41.88s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	41.922s
```

bflad:

PR #2230, opened 2017-11-09
r/redshift_cluster: Nest all logging fields

radeksimko:
This is to make the schema aligned more with the API and also to leave some name space for future new fields.

## Test Results

screen shot 2017-11-09 at 12 00 53

### DIFF
--- a/aws/resource_aws_redshift_cluster.go
+++ b/aws/resource_aws_redshift_cluster.go
@@ -219,22 +219,54 @@ func resourceAwsRedshiftCluster() *schema.Resource {
 				Set:      schema.HashString,
 			},
 
-			"enable_logging": {
-				Type:     schema.TypeBool,
+			"logging": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
 				Optional: true,
-				Default:  false,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enable": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+
+						"bucket_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+
+						"s3_key_prefix": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+						},
+					},
+				},
+			},
+
+			"enable_logging": {
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Computed:      true,
+				Deprecated:    "Use enable in the 'logging' block instead",
+				ConflictsWith: []string{"logging"},
 			},
 
 			"bucket_name": {
-				Type:     schema.TypeString,
-				Optional: true,
-				Computed: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				Deprecated:    "Use bucket_name in the 'logging' block instead",
+				ConflictsWith: []string{"logging"},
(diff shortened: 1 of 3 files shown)


aws/resource_aws_redshift_cluster.go
```diff
@@ -219,22 +219,54 @@ func resourceAwsRedshiftCluster() *schema.Resource {
...
+					},
+				},
+			},
+
+			"enable_logging": {
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Computed:      true,
+				Deprecated:    "Use enable in the 'logging' block instead",
+				ConflictsWith: []string{"logging"},
```
tombuildsstuff: would it make more sense to make this `logging.0.enabled`?

aws/resource_aws_redshift_cluster.go
```diff
@@ -711,17 +747,20 @@ func resourceAwsRedshiftClusterUpdate(d *schema.ResourceData, meta interface{})
...
-	if d.HasChange("enable_logging") || d.HasChange("bucket_name") || d.HasChange("s3_key_prefix") {
+	deprecatedHasChange := (d.HasChange("enable_logging") || d.HasChange("bucket_name") || d.HasChange("s3_key_prefix"))
+	if d.HasChange("logging") || deprecatedHasChange {
 		var loggingErr error
-		if _, ok := d.GetOk("enable_logging"); ok {
 
+		logging, ok := d.GetOk("logging.0.enable")
+		_, deprecatedOk := d.GetOk("enable_logging")
+
+		if (ok && logging.(bool)) || deprecatedOk {
```
tombuildsstuff: I think you want to check the value of `enable_logging`, rather than that it's set to a value?

aws/resource_aws_redshift_cluster.go
```diff
@@ -219,22 +219,54 @@ func resourceAwsRedshiftCluster() *schema.Resource {
...
+			"logging": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
 				Optional: true,
-				Default:  false,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enable": {
+							Type:     schema.TypeBool,
+							Required: true,
```
tombuildsstuff: does this not need a State Migration?

aws/resource_aws_redshift_cluster.go
radeksimko: As it is a boolean field it shouldn't matter, b/c it will always be present - that is in fact why we introduced [`d.GetOkExists`](https://github.com/hashicorp/terraform/blob/4cde21501c50dab41425569da7bfc71861289937/helper/schema/resource_data.go#L109), secondly I intentionally left the implementation as is since it will be deprecated eventually.

aws/resource_aws_redshift_cluster.go
radeksimko: I think there's no harm in conflicting with the whole `logging` block. Either they use deprecated field or the whole new block. I'd rather not deal with mix of deprecated and non-deprecated fields. 😉

aws/resource_aws_redshift_cluster.go
radeksimko: I don't think so - it's only required within the context of `logging` block, i.e. you can't define an empty `logging {}`, it shouldn't affect any existing configs without this block.

Issue #2227, opened 2017-11-09
Support: Schedule Action for Application Auto Scaling


atsushi-ishibashi:
http://docs.aws.amazon.com/ApplicationAutoScaling/latest/APIReference/API_PutScheduledAction.html
Require v1.12.25

atsushi-ishibashi:
Close #2231

PR #2231, opened 2017-11-09
New Resource: aws_appautoscaling_scheduled_action

atsushi-ishibashi:
#2227


atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsAppautoscalingScheduledAction_dynamo'
TF_ACC=1 go test ./aws -v -run=TestAccAwsAppautoscalingScheduledAction_basic -timeout 120m
=== RUN   TestAccAwsAppautoscalingScheduledAction_basic
--- PASS: TestAccAwsAppautoscalingScheduledAction_basic (90.87s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	90.907s 
```
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsAppautoscalingScheduledAction_ECS'
TF_ACC=1 go test ./aws -v -run=TestAccAwsAppautoscalingScheduledAction_ECS -timeout 120m
=== RUN   TestAccAwsAppautoscalingScheduledAction_ECS
--- PASS: TestAccAwsAppautoscalingScheduledAction_ECS (122.79s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	122.848s
```
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsAppautoscalingScheduledAction_EMR'
TF_ACC=1 go test ./aws -v -run=TestAccAwsAppautoscalingScheduledAction_EMR -timeout 120m
=== RUN   TestAccAwsAppautoscalingScheduledAction_EMR
--- PASS: TestAccAwsAppautoscalingScheduledAction_EMR (701.34s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	701.407s
```
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsAppautoscalingScheduledAction_SpotFleet'
TF_ACC=1 go test ./aws -v -run=TestAccAwsAppautoscalingScheduledAction_SpotFleet -timeout 120m
=== RUN   TestAccAwsAppautoscalingScheduledAction_SpotFleet
--- PASS: TestAccAwsAppautoscalingScheduledAction_SpotFleet (330.93s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	330.970s
```

aws/resource_aws_appautoscaling_scheduled_action.go
```diff
@@ -0,0 +1,222 @@
...
+			"scalable_dimension": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					validDimensions := []string{"ecs:service:DesiredCount", "ec2:spot-fleet-request:TargetCapacity",
+						"elasticmapreduce:instancegroup:InstanceCount", "appstream:fleet:DesiredCapacity",
+						"dynamodb:table:ReadCapacityUnits", "dynamodb:table:WriteCapacityUnits",
+						"dynamodb:index:ReadCapacityUnits", "dynamodb:index:WriteCapacityUnits"}
```
radeksimko: While offline, plan-time validation is usually useful I'm afraid it could be counter-productive here as the list of valid dimensions may and will expand and it did recently. I think we'd play a constant catch-up with AWS here and rushing to cut a release just to support a new dimension just because of this strict validation.
Theoretically we could validate for something like `[a-zA-Z]+:[a-zA-Z]+:[a-zA-Z]+` but frankly I'd just rather avoid the validation altogether here.

aws/resource_aws_appautoscaling_scheduled_action.go
```diff
@@ -0,0 +1,222 @@
...
+				Required: true,
+				ForceNew: true,
+			},
+			"service_namespace": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					validServiceNamespaces := []string{"ecs", "elasticmapreduce", "ec2", "appstream", "dynamodb"}
```
radeksimko: As below - I'm afraid the list of valid service namespaces may/will grow and I'd like us to avoid the maintenance burden associated with it, if possible.

aws/resource_aws_appautoscaling_scheduled_action.go
```diff
@@ -0,0 +1,222 @@
...
+			},
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsAppautoscalingScheduledActionCreate(d *schema.ResourceData, meta interface{}) error {
```
radeksimko: Per [docs](http://docs.aws.amazon.com/cli/latest/reference/application-autoscaling/put-scheduled-action.html) I think we should support updates of this resource and basically just rename this function to something like `resourceAwsAppautoscalingScheduledActionCreateOrUpdate` or `resourceAwsAppautoscalingScheduledActionPut` and reference it in `Update` too.

aws/resource_aws_appautoscaling_scheduled_action.go
```diff
@@ -0,0 +1,222 @@
...
+	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
+		_, err := conn.PutScheduledAction(input)
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case applicationautoscaling.ErrCodeObjectNotFoundException:
+					return resource.RetryableError(err)
+				default:
+					return resource.NonRetryableError(err)
+				}
```
radeksimko: Q: Is this related to eventual consistency?

Regardless of the reason I think the above conditional starting on L153 can be simplified to something like this:

```go
if err != nil {
  if isAWSErr(err, applicationautoscaling.ErrCodeObjectNotFoundException, "") {
  }
}
```

aws/resource_aws_appautoscaling_scheduled_action.go
```diff
@@ -0,0 +1,222 @@
...
+	resp, err := conn.DescribeScheduledActions(input)
+	if err != nil {
+		return err
+	}
+	if len(resp.ScheduledActions) < 1 {
+		d.SetId("")
+		return nil
+	}
+	if len(resp.ScheduledActions) != 1 {
+		return fmt.Errorf("Number of Scheduled Action (%s) isn't one, got %d", saName, len(resp.ScheduledActions))
```
radeksimko: Grammar nitpick, do you mind changing this to `Expected 1 scheduled action under %s, found %d`?

aws/resource_aws_appautoscaling_scheduled_action.go
```diff
@@ -0,0 +1,222 @@
...
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	d.SetId(d.Get("name").(string))
```
radeksimko: Shouldn't the ID consist of name + service namespace as that's how we look it up? In other words my question is: Can there be more than 1 scheduled action of the same name? If not then ignore me.

aws/resource_aws_appautoscaling_scheduled_action_test.go
```diff
@@ -0,0 +1,679 @@
...
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ecs:DescribeServices",
+        "ecs:UpdateService",
+				"cloudwatch:DescribeAlarms"
```
radeksimko: Nitpick - extra whitespace here 👀

aws/resource_aws_appautoscaling_scheduled_action_test.go
```diff
@@ -0,0 +1,679 @@
...
+        "*"
+      ]
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_ecs_cluster" "hoge" {
+	name = "tf-ecs-cluster-%s"
```
radeksimko: extra whitespace here 👀

aws/resource_aws_appautoscaling_scheduled_action_test.go
```diff
@@ -0,0 +1,679 @@
...
+}
+
+resource "aws_ecs_task_definition" "hoge" {
+  family = "foobar"
```
radeksimko: Do you mind using a random name here too, please?

website/docs/r/appautoscaling_scheduled_action.html.markdown
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the scheduled action.
+* `service_namespace` - (Required) The namespace of the AWS service. Documentation can be found in the parameter at: [AWS Application Auto Scaling API Reference](https://docs.aws.amazon.com/ApplicationAutoScaling/latest/APIReference/API_PutScheduledAction.html#ApplicationAutoScaling-PutScheduledAction-request-ServiceNamespace) Valid Values: ecs | elasticmapreduce | ec2 | appstream | dynamodb
```
radeksimko: I would avoid documenting all valid values as that will become a maintenance burden as these expand. How about we just mention a few examples? The user can always IMO look up all valid values in the docs, right?

website/docs/r/appautoscaling_scheduled_action.html.markdown
```diff
@@ -0,0 +1,89 @@
...
+* `resource_id` - (Required) The identifier of the resource associated with the scheduled action. Documentation can be found in the parameter at: [AWS Application Auto Scaling API Reference](https://docs.aws.amazon.com/ApplicationAutoScaling/latest/APIReference/API_PutScheduledAction.html#ApplicationAutoScaling-PutScheduledAction-request-ResourceId)
+* `scalable_dimension` - (Optional) The scalable dimension. Documentation can be found in the parameter at: [AWS Application Auto Scaling API Reference](https://docs.aws.amazon.com/ApplicationAutoScaling/latest/APIReference/API_PutScheduledAction.html#ApplicationAutoScaling-PutScheduledAction-request-ScalableDimension) Valid Values: ecs:service:DesiredCount | ec2:spot-fleet-request:TargetCapacity | elasticmapreduce:instancegroup:InstanceCount | appstream:fleet:DesiredCapacity | dynamodb:table:ReadCapacityUnits | dynamodb:table:WriteCapacityUnits | dynamodb:index:ReadCapacityUnits | dynamodb:index:WriteCapacityUnits
```
radeksimko: I would avoid documenting all valid values as that will become a maintenance burden as these expand. How about we just mention a few examples? The user can always IMO look up all valid values in the docs, right?

aws/resource_aws_appautoscaling_scheduled_action.go
atsushi-ishibashi: I assumed that soon after creating dynamo table and on...

aws/resource_aws_appautoscaling_scheduled_action.go
atsushi-ishibashi: According to `DeleteScheduledActionInput` docs, `ResourceId, ScheduledActionName, ServiceNamespace` are required so we have to generate id from these three.
Should I convert to hash?

aws/resource_aws_appautoscaling_scheduled_action.go
atsushi-ishibashi: Ok, I removed

aws/resource_aws_appautoscaling_scheduled_action.go
atsushi-ishibashi: [docs](https://docs.aws.amazon.com/ApplicationAutoScaling/latest/APIReference/API_PutScheduledAction.html#ApplicationAutoScaling-PutScheduledAction-request-ScalableDimension) lists constant strings so I removed `ValidateFunc`.

aws/resource_aws_appautoscaling_scheduled_action.go
atsushi-ishibashi: OK👍

aws/resource_aws_appautoscaling_scheduled_action.go
radeksimko: I'd suggest we make it something like this:

```go
d.SetId(name+"-"+serviceNamespace+"-"+resourceId)
```
assuming the underscore has no special meaning in either of these fields.

PR #1884, opened 2017-10-13
Load balancer fixes

DaveBlooman:
The loadbalancer changes don't allow for a NLB to be created because the NLB uses TCP only, with most of the code only allowing HTTP or HTTPS.

I think there is probably some additional code coverage that can be added here, especially as NLB's are much more restrictive than ALB/ELB

Example

```sh
resource "aws_lb" "main" {
  name               = "test-nlb-tf"
  internal           = false
  load_balancer_type = "network"
  subnets            = ["subnet-123456"]

  enable_deletion_protection = true

  tags {
    Environment = "test"
  }
}

resource "aws_lb_target_group" "test" {
  name     = "tf-example"
  port     = 8085
  protocol = "TCP"
  vpc_id   = "vpc-c333ccc3"

  health_check {
    interval            = 30
    port                = "traffic-port"
    timeout             = 10
    protocol            = "TCP"
    healthy_threshold   = 3
    unhealthy_threshold = 3
  }
}

resource "aws_lb_listener" "front_end" {
  load_balancer_arn = "${aws_lb.main.id}"
  port              = "80"
  protocol          = "TCP"

  default_action {
    target_group_arn = "${aws_lb_target_group.test.id}"
    type             = "forward"
  }
}

resource "aws_lb_target_group_attachment" "test" {
  target_group_arn = "${aws_lb_target_group.test.arn}"
  target_id        = "i-0000000000000"
  port             = 80
}

```

PR #2242, opened 2017-11-10
Add support for NLB target groups.

markhibberd:
Support for network load balancers was added in https://github.com/terraform-providers/terraform-provider-aws/pull/1806, but the protocol validation prevents target groups being created for them. This change just extends the validation to allow `tcp` protocol and health checks (which is the only valid value for network load balancers).


keichan34:
This is addressed in #1884, I believe

markhibberd:
@keichan34 Yep, missed that. Thanks, will close this.

PR #2256, opened 2017-11-11
resource/aws_lb: Disable access_logs for network load balancers

### LINKED ISSUE #2145: aws_lb access_logs ignores enabled
_This issue was originally opened by @bluemalkin as hashicorp/terraform#16516. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.10.8

```resource "aws_lb" "foo" {
  name            = "foo"
  security_groups = ["${aws_security_group.foo.id}"]
  subnets         = ["${aws_subnet.foo.*.id}"]

  access_logs {
    bucket = "${aws_s3_bucket.foo.bucket}"
    enabled= "0"
  }
}
```

### Expected Behavior
Access logs should be disabled.

### Actual Behavior
Instead it's always enabled and does not detect values "false" or "0".

### Update
I checked the state files, they have the right value, I think it's the output/changing state that is not reported correctly. Can someone else please try with different enabled values ? Thanks

gmccue:
Fixes #2145


aws/resource_aws_lb.go
```diff
@@ -316,7 +316,8 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	attributes := make([]*elbv2.LoadBalancerAttribute, 0)
 
-	if d.HasChange("access_logs") {
+	// Access Logs are not supported for Network Loadbalancers
+	if d.Get("load_balancer_type").(string) != "network" && d.HasChange("access_logs") {
```
Ninir: I'm not feeling comfortable in allowing a `terraform apply` when the code will just ignore it. I do think we should add a guard when planning/applying, saying that this is not allowed rather than ignoring it, which may people think that logs are configured whereas it is not.

Thoughts? 😄

gmccue: I do agree, but preventing `apply` entirely might cause some issues as well. For my use case, see: https://github.com/terraform-aws-modules/terraform-aws-alb/blob/master/main.tf#L15. In certain modules it wouldn't be possible to create `network` load balancers this way.
But I'm a bit of a TF noob. Is there a way to do a check within the module?

Otherwise we could check against the `access_logs.enabled` parameter.

What do you think?

gmccue: We could also add a `log.Printf("[WARN]")`, but I'm not sure what the preferred method is for handling these cases.

gmccue: @Ninir Any thoughts?

Ninir: Let's go for an INFO notice here :)

gmccue: @Ninir Done!

aws/resource_aws_lb_test.go
```diff
@@ -941,6 +992,51 @@ resource "aws_subnet" "alb_test" {
...
+func testAccAWSLBConfig_networkLoadbalancerAccessLogs(lbName string) string {
+	return fmt.Sprintf(`resource "aws_lb" "lb_test" {
+  name            = "%s"
+  internal        = true
+  load_balancer_type = "network"
+
+  enable_deletion_protection = false
+
+  access_logs {
+    bucket = "${aws_s3_bucket.logs.bucket}"
```
Ninir: This line and the one after will trigger issues: 
* no such `var.bucket_prefix` variable
* no such logs bucket

Could you have a look? otherwise looks good to me!

gmccue:
@Ninir I'm a bit stuck here. The problem is that the new acceptance test to check that access logs are disabled fails:

```
=== RUN   TestAccAWSLB_networkLoadbalancerAccessLogsDisabled
--- FAIL: TestAccAWSLB_networkLoadbalancerAccessLogsDisabled (233.30s)
	testing.go:503: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		UPDATE: aws_lb.lb_test
		  access_logs.0.enabled: "" => "true"
		  access_logs.0.prefix:  "" => "test_bucket"
```

As far as I can tell, this is because the test setup does a `diff` against two different states. Because we are passing an `access_log` attribute, one of the states attempts to create the access logs, while the other does not.
I guess the simple fix for this would just be not to create a new acceptance test for this case, like is currently done for the `idle_timeout` parameter (doesn't have a unique acceptance test, but does skip creation if the load balancer type is `network`). Any thoughts?

gmccue:
@Ninir any thoughts?

gmccue:
@Ninir @radeksimko Sorry to bother again with this, but do you have any more guidance or input for this PR? Or should I just close it?

gmccue:
@bflad @radeksimko is this still relevant, or should I go ahead and close this PR?

bflad:
Hi @gmccue! 👋  Sorry that we've been slow to re-review this PR. I can't speak for the other maintainers (and would prefer they continued their review unless they don't have time) but it might be beneficial to migrate some of the logic into the `CustomizeDiff` function that was recently updated to handle plan-time validation of `stickiness` for ALB vs NLB in #2746. Without diving too much into your actual code changes in this PR, I would imagine we'll need similar handling to allow the `access_logs` configuration for NLBs as long as its disabled to handle Terraform modules easier. If you're already doing that then you're probably already on the right track. 👍 

Admittedly, the handling of ALB vs NLB in these resources has proven to be quite onerous and generated a ton of issues. To better remedy the situation, we have talked internally about potentially splitting the resources into `aws_alb*`/`aws_nlb*`, but there are currently no official plans or timelines for doing that.

PR #2266, opened 2017-11-13
New Data Source: aws_instances

### LINKED ISSUE #574: aws_instances data source
_This issue was originally opened by @mattlqx as hashicorp/terraform#12376. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I would find it useful to have an `aws_instances` data source that would be able to execute a filter and return the instance IDs of multiple instances. Presently the `aws_instance` data source says:

```
NOTE: If more or less than a single match is returned by the search, Terraform will fail. Ensure that your search is specific enough to return a single Instance ID only.
```

My use-case is that have a Terraform repo that has some common shared infrastructure (like HTTP hosts that hand-out 301s for whatever URL they're given) and I want to attach that to multiple AWS ALBs that are in separate Terraform repos (per application). I would like to do a search on those instances based on tags and be able to use the result list in an `aws_alb_target_group_attachment`. I'd really like to be able to have an `aws_alb_target_group` data source, but this is an acceptable (and useful in more ways) alternative.

Right now, the only real way I can get those instances is to hardcode the instance IDs in individual `aws_instance` data sources. I can work out some outputs where the hosts are created, and work them into being imported into the application Terraforms but this additional layer increases complexity and can become stale or out-of-sync if not done correctly. It's better just to hardcode the IDs for now.

radeksimko:
Closes #574
Replaces https://github.com/hashicorp/terraform/pull/12856

## Test Results

```
TF_ACC=1 go test ./aws -v -run=TestAccAWSInstancesDataSource_ -timeout 120m
=== RUN   TestAccAWSInstancesDataSource_basic
--- PASS: TestAccAWSInstancesDataSource_basic (163.70s)
=== RUN   TestAccAWSInstancesDataSource_tags
--- PASS: TestAccAWSInstancesDataSource_tags (155.65s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	319.375s
```

### DIFF
--- a/aws/data_source_aws_instances.go
+++ b/aws/data_source_aws_instances.go
@@ -1,0 +1,112 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsInstances() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsInstancesRead,
+
+		Schema: map[string]*schema.Schema{
+			"filter":        dataSourceFiltersSchema(),
+			"instance_tags": tagsSchemaComputed(),
+
+			"ids": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"private_ips": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"public_ips": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func dataSourceAwsInstancesRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	filters, filtersOk := d.GetOk("filter")
+	tags, tagsOk := d.GetOk("instance_tags")
+
+	if !filtersOk && !tagsOk {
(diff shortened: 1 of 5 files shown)


aws/data_source_aws_instances.go
```diff
@@ -0,0 +1,106 @@
...
+		params.Filters = append(params.Filters, buildEC2TagFilterList(
+			tagsFromMap(tags.(map[string]interface{})),
+		)...)
+	}
+
+	log.Printf("[INFO] Describing EC2 instances: %s", params)
+
+	var instanceIds, privateIps, publicIps []string
+	err := conn.DescribeInstancesPages(params, func(resp *ec2.DescribeInstancesOutput, isLast bool) bool {
+		// loop through reservations, and remove terminated instances, populate instance slice
```
catsby: Option: could always use a `filter` on `instance-state-name` and append user filters to it, to save you this looping and possibly entire API pagination calls

- http://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html

aws/data_source_aws_instances.go
```diff
@@ -0,0 +1,106 @@
...
+		return err
+	}
+
+	if len(instanceIds) < 1 {
+		return fmt.Errorf("Your query returned no results. Please change your search criteria and try again.")
+	}
+
+	log.Printf("[DEBUG] Found %d instances via given filter", len(instanceIds))
+
+	d.SetId(time.Now().String())
```
catsby: Optional/question: is there an advantage or reason to use `time.Now()` here for the `id` instead of using [resource.UniqueId()](https://github.com/hashicorp/terraform/blob/aef082d1eceafd42f303748825d753c057555e00/helper/resource/id.go#L17)?

radeksimko: UniqueID sounds _safer_, I'll change it - good point.

aws/data_source_aws_instances.go
radeksimko: Ha, very valid point, I'll change that 👍

PR #2271, opened 2017-11-14
Add arn attribute to aws_ebs_volume resource and datasource

bflad:
Not provided directly by EC2 DescribeVolumes API, so built manually. References:
* http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#Volume
* http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arn-syntax-ec2

Closes #2261 

```
make testacc TESTARGS='-run=TestAccAWSEbsVolumeDataSource_basic'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAWSEbsVolumeDataSource_basic -timeout 120m
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
=== RUN   TestAccAWSEbsVolumeDataSource_basic
--- PASS: TestAccAWSEbsVolumeDataSource_basic (25.51s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	25.547s

make testacc TESTARGS='-run=TestAccAWSEBSVolume_basic'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAWSEBSVolume_basic -timeout 120m
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
=== RUN   TestAccAWSEBSVolume_basic
--- PASS: TestAccAWSEBSVolume_basic (22.02s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	22.057s
```

PR #2272, opened 2017-11-14
Add arn attribute to aws_elb resource

### LINKED ISSUE #2123: Add ARN attribute to `aws_elb` resource
It seems that the `aws_elb` resource doesn't output an `arn` attribute (whereas the `aws_lb` resource does). The use case for this is a custom Lambda function that we have written which takes a list of domains to generate certificates for using LetsEncrypt, as well as a list of ELB ARNs to use for updating certificates.

bflad:
Not provided directly by DescribeLoadBalancers/DescribeLoadBalancerAttributes APIs, so built manually. References:
* http://docs.aws.amazon.com/sdk-for-go/api/service/elb/#LoadBalancerDescription
* http://docs.aws.amazon.com/sdk-for-go/api/service/elb/#LoadBalancerAttributes
* http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arn-syntax-elb

Closes #2123 

I cannot run `TestAccAWSELB_basic` since the default VPC has been removed from my own AWS account, but these are similar changes as #2271.

### DIFF
--- a/aws/resource_aws_elb.go
+++ b/aws/resource_aws_elb.go
@@ -47,7 +47,6 @@ func resourceAwsElb() *schema.Resource {
 
 			"arn": &schema.Schema{
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 
--- a/aws/resource_aws_elb.go
+++ b/aws/resource_aws_elb.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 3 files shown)


aws/resource_aws_elb.go
```diff
@@ -44,6 +45,12 @@ func resourceAwsElb() *schema.Resource {
 				ValidateFunc: validateElbNamePrefix,
 			},
 
+			"arn": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
```
radeksimko: Any particular reason why it should be optional? i.e. why would the user need to define it in the config?

bflad: Ah sorry, delicious copypasta. It should probably always be just Computed.

bflad:
Pushed update to remove `Optional`, which sounds like the right thing to do. 👍

PR #2276, opened 2017-11-14
r/aws_elb: Add listener ssl_certificate_id ARN validation

### LINKED ISSUE #2122: aws_elb: misleading error message when specifying aws_iam_server_certificate.id instead of aws_iam_certificate.arn
_This issue was originally opened by @kevinburkeotto as hashicorp/terraform#16513. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

The documentation for the `ssl_certificate_id` parameter is as follows: (https://www.terraform.io/docs/providers/aws/r/elb.html#ssl_certificate_id)

```
ssl_certificate_id - (Optional) The ARN of an SSL certificate you have uploaded to AWS IAM. Note ECDSA-specific restrictions below. Only valid when lb_protocol is either HTTPS or SSL
```

I missed the part about "ARN" and instead tried to specify it like this:

```hcl
    ssl_certificate_id = "${aws_iam_server_certificate.tserver_certificate.id}"
```

It seems like a reasonable assumption that a field named "ssl_certificate_id" would accept an id. When you do this, you get a misleading error message:

```
Error creating ELB Listener with SSL Cert, retrying: CertificateNotFound: Server Certificate not found for the key: ASCA...
```

I'm confused, and searching other issues turns up issues like hashicorp/terraform#2499, which indicate there was a delay issue between the certificate appearing and the ELB being able to identify it.

It seems like we can provide a better error message in this case, or a better parameter name (`ssl_certificate_arn` maybe?), or detect whether you are trying to pass a value that is not an ARN, and throw a better error in that case.

bflad:
Closes #2122

### DIFF
--- a/aws/resource_aws_elb.go
+++ b/aws/resource_aws_elb.go
@@ -178,8 +178,9 @@ func resourceAwsElb() *schema.Resource {
 						},
 
 						"ssl_certificate_id": &schema.Schema{
-							Type:     schema.TypeString,
-							Optional: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validateArn,
 						},
 					},
 				},


bflad:
Makes sense -- I wonder if core during the plan could use the validateFunc of the computed attribute to ensure its "output" aligns with the input? If that doesn't work even a simple new "outputFormat" schema option that provides a necessary validation string/type? I agree it is definitely a core enhancement.

I'll poke around the core issues and create something if I can't find anything. :smile:

PR #2277, opened 2017-11-14
r/aws_elasticache_security_group: add import support

### LINKED ISSUE #2260: Unable to import aws_elasticache_security_group
### Terraform Version
Terraform v0.10.8
AWS Provider v1.2

### Affected Resource(s)
- `aws_elasticache_security_group`

```hcl
resource aws_elasticache_security_group "redis" {
  name = "redis"
  description = "redis"

  security_group_names = ["web", "app"]
}
```

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
Resource was imported successfully

### Actual Behavior
```hcl
aws_elasticache_security_group.redis (import id: redis): import aws_elasticache_security_group.redis (id: redis): resource aws_elasticache_security_group doesn't support import
```

### Steps to Reproduce
1. Create an elasticache security group through the AWS console called `redis` with a description of `redis`.
2. `terraform import aws_elasticache_security_group.redis redis`

### Important Factoids
- EC2 Classic

bflad:
Closes #2260 

Sorry I cannot run `TestAccAWSElasticacheSecurityGroup_Import` as my AWS account has its default VPCs removed.

### DIFF
--- a/aws/resource_aws_elasticache_security_group.go
+++ b/aws/resource_aws_elasticache_security_group.go
@@ -17,6 +17,12 @@ func resourceAwsElasticacheSecurityGroup() *schema.Resource {
 		Create: resourceAwsElasticacheSecurityGroupCreate,
 		Read:   resourceAwsElasticacheSecurityGroupRead,
 		Delete: resourceAwsElasticacheSecurityGroupDelete,
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				d.Set("name", d.Id())
+				return []*schema.ResourceData{d}, nil
+			},
+		},
 
 		Schema: map[string]*schema.Schema{
 			"description": &schema.Schema{
--- a/aws/resource_aws_elasticache_security_group.go
+++ b/aws/resource_aws_elasticache_security_group.go
@@ -18,10 +18,7 @@ func resourceAwsElasticacheSecurityGroup() *schema.Resource {
 		Read:   resourceAwsElasticacheSecurityGroupRead,
 		Delete: resourceAwsElasticacheSecurityGroupDelete,
 		Importer: &schema.ResourceImporter{
-			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
-				d.Set("name", d.Id())
-				return []*schema.ResourceData{d}, nil
-			},
+			State: schema.ImportStatePassthrough,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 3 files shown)


aws/resource_aws_elasticache_security_group.go
```diff
@@ -17,6 +17,12 @@ func resourceAwsElasticacheSecurityGroup() *schema.Resource {
 		Create: resourceAwsElasticacheSecurityGroupCreate,
 		Read:   resourceAwsElasticacheSecurityGroupRead,
 		Delete: resourceAwsElasticacheSecurityGroupDelete,
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				d.Set("name", d.Id())
```
radeksimko: Instead of adding a custom importer, how do you feel about replacing `d.Get("name")` with `d.Id()`? It seems a bit cleaner as we may then use the simple passthrough importer here.

radeksimko:
> Can you verify the acceptance testing passes in your account as-is first?

It does not and I believe it's because during import Terraform won't look at the static provider block (which rightly contains `us-east-1` where we have EC2Classic enabled), it expects it to be configured from outside. Here's an example of how to do that within the test:

```go
	oldRegion := os.Getenv("AWS_DEFAULT_REGION")
	os.Setenv("AWS_DEFAULT_REGION", "us-east-1")
	defer os.Setenv("AWS_DEFAULT_REGION", oldRegion)
```

you can find more examples in other tests if you search for `AWS_DEFAULT_REGION`.

> Would you prefer bundling the refactoring of this one in this PR now or them all together?

I'd prefer 1 PR per resource, if possible. Smaller PRs are much easier to review.

Thanks.

radeksimko:
btw. when I add the region workaround I'm still getting failure:

```
=== RUN   TestAccAWSElasticacheSecurityGroup_Import
--- FAIL: TestAccAWSElasticacheSecurityGroup_Import (40.75s)
	testing.go:503: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) {
		}


		(map[string]string) (len=2) {
		 (string) (len=22) "security_group_names.#": (string) (len=1) "1",
		 (string) (len=31) "security_group_names.4153974498": (string) (len=41) "tf-test-security-group-653022196438882106"
		}
```

bflad:
Sorry for my slow response here. 😦  I'll convert this over to `d.Id()` and fix up the test right now.

bflad:
Ah, the test failure previously was due to a lack of setting the security group names in the state during read, fixed! Now we're also using `d.Id()` on read with `ImportStatePassthrough` for import. With the addition of the import test EC2-Classic workaround, this verified locally for me (just so happens I have an EC2-Classic enabled us-east-1 account I can use). 👍 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticacheSecurityGroup'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticacheSecurityGroup -timeout 120m
=== RUN   TestAccAWSElasticacheSecurityGroup_basic
--- PASS: TestAccAWSElasticacheSecurityGroup_basic (9.18s)
=== RUN   TestAccAWSElasticacheSecurityGroup_Import
--- PASS: TestAccAWSElasticacheSecurityGroup_Import (7.06s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	16.281s
```

PR #2292, opened 2017-11-14
Fix example: reflect required root_device_type

jlsjonas:


kwilczynski:
I would not count this as proper documentation update, to be honest. 👎

Ninir:
Hey @jlsjonas 

I'm going to close this one for the reason above. If you feel this is the needed fix, can you tell us bringing a configuration that is not working? we will then reopen.

Thanks!

PR #1941, opened 2017-10-17
SetSubnets not allowed on LB type network

DaveBlooman:
Covers #1925

PR #2293, opened 2017-11-14
SetSubnets not allowed on LB type network (supersedes #1941)

catsby:
Continuation of https://github.com/terraform-providers/terraform-provider-aws/pull/1941

- [x] document this in the lb documentation
- [x] mention that terraform taint is the way to do change the subnets in our error message
- [x] update the check on subnet update to only fire if it's not a new resource

### DIFF
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -389,9 +389,14 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 
 	}
 
-	if d.HasChange("subnets") {
+	if d.HasChange("subnets") && !d.IsNewResource() {
+		// subnets are assigned at Create; the 'change' here is an empty map for old
+		// and current subnets for new, so this change is redundant when the
+		// resource is just created.
 		if d.Get("load_balancer_type").(string) == "network" {
-			return fmt.Errorf("Unable to update subnets for loadbalancer %s, updating subnets is not supported for type network", d.Id())
+			// Load Balancers of type 'network' cannot update their subnets at this
+			// time.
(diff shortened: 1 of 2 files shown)


aws/resource_aws_lb.go
```diff
@@ -389,7 +389,16 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
...
 
-	if d.HasChange("subnets") {
+	if d.HasChange("subnets") && !d.IsNewResource() {
+		// subnets are assigned at Create; the 'change' here is an empty map for old
+		// and current subnets for new, so this change is redundant when the
+		// resource is just created.
+		if d.Get("load_balancer_type").(string) == "network" {
+			// Load Balancers of type 'network' cannot update their subnets at this
+			// time.
+			return fmt.Errorf("Unable to update subnets for loadbalancer %s, updating subnets is not supported for type network. Please use the 'taint' command to force a recreation of this resource", d.Id())
```
radeksimko: Nitpick (read as _I'm happy to land this as is_): Could we possibly use the new `CustomizeDiff` to mark `subnets` as `ForceNew` in this case? https://github.com/hashicorp/terraform/pull/14887/files#diff-456d1e45f8a8d1585e8228597b9a2176R2845 instead of instructing the user to taint?

catsby:
I'm going to reopen this with a fancy new `CustomizeDiff`

aws/resource_aws_lb.go
catsby: TIL about `CustomizeDiff`! I've opened https://github.com/terraform-providers/terraform-provider-aws/pull/2310 as a replacement

catsby:
Replacement at https://github.com/terraform-providers/terraform-provider-aws/pull/2310

PR #2305, opened 2017-11-15
[WIP] Dms task lifecycle

chrisjharding:
Closes https://github.com/terraform-providers/terraform-provider-aws/issues/2236

It adds lifecycle behavior to dms replication tasks. This behavior is switched under the `handle_task_lifecycle` param so it can be turned off if not desirable. When enabled, it will start the task after creating it. When a modification is needed, it will stop the task prior to modifying it, then resume the task. It also stops the task before deletion.

First attempt. To test this, it actually requires running the dms task which means setting up real dms endpoints and having real data to migrate. 

Currently, the tests are failing due to remote-exec not being available. I need the remote exec on the bastion to populate the db with data to migrate. 

```
$ TF_LOG=DEBUG TF_ACC=1 go test ./aws -v -run=TestAccAwsDmsReplicationTaskHandleLifecycle -timeout 60m

--- FAIL: TestAccAwsDmsReplicationTaskHandleLifecycle (2.83s)
	testing.go:503: Step 0 error: Configuration is invalid.

		Warnings: []string(nil)

		Errors: []string{"1 error(s) occurred:\n\n* aws_instance.bastion: provisioner remote-exec couldn't be found"}
FAIL
exit status 1
```

One last note, I have had to copy in a hardcoded rsa key as the tls provider is not available when running these tests. 

Is anyone able to advise on how I can use remote-exec provisioner with an aws instance inside AWS ACC tests please?


danielnitz:
really looking forward for this to make it into master!

Zechw:
Excited for this as well! Stuck on the same bug. Any ideas on timeframe, or what's blocking this?

jbergknoff-rival:
@bflad any chance you or another maintainer could review this PR/give feedback on test strategy? This seems like a very useful feature to have.

pancochea:
We are facing the same issue. We cannot manage DMS tasks on terraform cloud as it's impossible to apply changes to the tasks until the plugin manages the lifecycle.

ewbankkit:
Closing as stale due to inactivity.

PR #2310, opened 2017-11-15
r/aws_lb: mark subnets as ForceNew for network load balancers

### LINKED ISSUE #1925: SetSubnets not allowed on LB type network
Network load balancers cannot update subnets, so the following will return an error.

https://github.com/terraform-providers/terraform-provider-aws/blob/5eb6320f9f4628a255b871c234bee52c4cf83197/aws/resource_aws_lb.go#L400-L404

The example use case would be adding or removing a subnet from the list provided

I guess destroy then recreate could be offered, or just return an error indicating action isn't possible.

catsby:
Replaces #2293

Subnets cannot be updated on Load Balancers of type `network` at this time:

- [x] Documentation 
- [x] Conditionally mark `subnets` as `ForceNew` if the `load_balancer_type` is set to `network`

Fixes #1925

### DIFF
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -24,6 +24,8 @@ func resourceAwsLb() *schema.Resource {
 		Read:   resoureAwsLbRead,
 		Update: resourceAwsLbUpdate,
 		Delete: resourceAwsLbDelete,
+		// Subnets are ForceNew for Network Load Balancers
+		CustomizeDiff: customizeDiffNLBSubnets,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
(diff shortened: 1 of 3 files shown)


catsby:
cc @DaveBlooman , this should address #1925

aws/resource_aws_lb.go
```diff
@@ -687,3 +693,50 @@ func flattenAwsLbResource(d *schema.ResourceData, meta interface{}, lb *elbv2.Lo
...
+	// handling is the only known difference between Network Load Balancers and
+	// Application Load Balancers, so the logic below is simple individual checks.
+	// If other differences arise we'll want to refactor to check other
+	// conditions in combinations, but for now all we handle is subnets
+	lbType := diff.Get("load_balancer_type").(string)
+	if "network" != lbType {
+		return nil
+	}
+
+	if "" == diff.Id() {
```
radeksimko: nitpick, but `diff.Id() == ""` would make more sense. 😉

catsby: Old defensive habit from my Ruby days, trying to save myself from typos like 

```
if variable = "something" {
  //stuff 
}
``` 

when I meant 

```
if variable == "something" {
  //stuff
}
```

Issue #2306, opened 2017-11-15
AWS Data Source for Network Interface


hashibot:
_This issue was originally opened by @djl87 as hashicorp/terraform#16659. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Could you please add a data source for the aws_network_interface?

I have a requirement to have some static network interfaces for licencing purposes which I need to then attach to instances when launched.

Currently I have to have them as mapped variables according to the region of deployment.

atsushi-ishibashi:
Close

radeksimko:
Thanks for the ping @atsushi-ishibashi, implemented in https://github.com/terraform-providers/terraform-provider-aws/pull/2316

PR #2316, opened 2017-11-16
New Datasource: aws_network_interface

atsushi-ishibashi:
#2306 
```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsNetworkInterface_basic'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsNetworkInterface_basic -timeout 120m
=== RUN   TestAccDataSourceAwsNetworkInterface_basic
--- PASS: TestAccDataSourceAwsNetworkInterface_basic (65.61s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	66.025s
```

### DIFF
--- a/aws/data_source_aws_network_interface.go
+++ b/aws/data_source_aws_network_interface.go
@@ -17,7 +17,7 @@ func dataSourceAwsNetworkInterface() *schema.Resource {
 				Required: true,
 			},
 			"association": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
@@ -45,7 +45,7 @@ func dataSourceAwsNetworkInterface() *schema.Resource {
 				},
 			},
 			"attachment": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Computed: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
@@ -107,7 +107,7 @@ func dataSourceAwsNetworkInterface() *schema.Resource {
 				Computed: true,
 			},
 			"private_ips": &schema.Schema{
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
@@ -143,7 +143,7 @@ func dataSourceAwsNetworkInterfaceRead(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("no matching network interface found")
 	}
 	if len(resp.NetworkInterfaces) > 1 {
-		return fmt.Errorf("multiple network interfaces matched")
+		return fmt.Errorf("multiple network interfaces matched %s", d.Id())
 	}
 
 	eni := resp.NetworkInterfaces[0]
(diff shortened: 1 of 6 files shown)


website/docs/d/network_interface.html.markdown
```diff
@@ -0,0 +1,47 @@
+---
+layout: "aws"
+page_title: "AWS: aws_network_interface"
+sidebar_current: "docs-aws-datasource-network-interface"
+description: |-  
+  Get information on anNetwork Interface resource.
```
eheydrick: on a Network Interface

aws/data_source_aws_network_interface.go
```diff
@@ -0,0 +1,174 @@
...
+func dataSourceAwsNetworkInterface() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsNetworkInterfaceRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"association": &schema.Schema{
+				Type:     schema.TypeSet,
```
radeksimko: Since there's only ever going to be a single association I think this field should be TypeList so it's possible to reference it like `data.aws_network_interface.name.association.0.public_ip` etc.

aws/data_source_aws_network_interface.go
```diff
@@ -0,0 +1,174 @@
...
+						},
+						"public_ip": &schema.Schema{
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"attachment": &schema.Schema{
+				Type:     schema.TypeSet,
```
radeksimko: Since there's only ever going to be a single attachment I think this field should be TypeList so it's possible to reference it like `data.aws_network_interface.name.attachment.0.instance_id` etc.

aws/data_source_aws_network_interface.go
```diff
@@ -0,0 +1,174 @@
...
+			"private_dns_name": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"private_ip": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"private_ips": &schema.Schema{
+				Type:     schema.TypeSet,
```
radeksimko: Any reason to keep this list of IPs Set? It looks like a field people would like to reference and Set would prevent them in doing so due to computed index. That said I'm not sure if the ordering of IPs has any meaning - i.e. if they're sorted by networking interfaces (primary first, then others)?

aws/data_source_aws_network_interface.go
```diff
@@ -0,0 +1,174 @@
...
+
+	resp, err := conn.DescribeNetworkInterfaces(input)
+	if err != nil {
+		return err
+	}
+	if resp == nil || len(resp.NetworkInterfaces) == 0 {
+		return fmt.Errorf("no matching network interface found")
+	}
+	if len(resp.NetworkInterfaces) > 1 {
+		return fmt.Errorf("multiple network interfaces matched")
```
radeksimko: Do you mind improving the error message here to also include the ID we looked for? e.g. `multiple network interfaces matched %s", d.Id()`

aws/data_source_aws_network_interface.go
atsushi-ishibashi: I used `Set` because I thought the ordering has no meaning.
I agree with your suggestion👍

atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsNetworkInterface_basic'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsNetworkInterface_basic -timeout 120m
=== RUN   TestAccDataSourceAwsNetworkInterface_basic
--- PASS: TestAccDataSourceAwsNetworkInterface_basic (95.81s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	95.852s
```

PR #2209, opened 2017-11-07
d/aws_nat_gateway: Add missing address attributes to the schema

darend:
## what

Include Address attributes in NAT Gateway data source

## why

Address attributes were not listed in the schema which prevented them from being referenced

## Sample TF Project

```
provider "aws" {
  region  = "${var.region}"
  version = ">=1.2.0"
}

data "aws_nat_gateway" "default" {
    id = "${var.nat_id}"
}

output "subnet_id" {
    value = "${data.aws_nat_gateway.default.subnet_id}"
}

output "private_ip" {
    value = "${data.aws_nat_gateway.default.private_ip}"
}
```

Running terraform with DEBUG logging I see this warning:

```
2017/11/07 13:04:19 [WARN] Output interpolation "private_ip" failed: Resource 'data.aws_nat_gateway.default' does not have attribute 'private_ip' for variable 'data.aws_nat_gateway.default.private_ip'
```

See https://github.com/terraform-providers/terraform-provider-aws/pull/1294#discussion_r149505469

PR #2324, opened 2017-11-16
data-nat-gateway: added more fields to read

nicgrayson:
This read matches what is currently in the docs


nicgrayson:
oops, dup of #2209

PR #2339, opened 2017-11-17
r/ecs_task_definition: Fix equivalency comparator

### LINKED ISSUE #2336: aws_ecs_task_definition: Regression from issue #1195 causing an endless cycle of diffs
The change in #1195 is causing issues with existing ECS task definitions. The task definitions coming back from AWS are logically equivalent but are structurally different. For instance, my `environment` array is in a different order than what I have on disk. Also, there are a few properties which are empty arrays (like `volumesFrom`) which come back despite not being specified in my file. And the `protocol` for a port comes back as `tcp` even though that's optional and the default.

Anyway, all of these "changes" result in a diff. The code handling the JSON from AWS needs to transform the input JSON and AWS's JSON so that the data is homogenized the same way to avoid unnecessary diffs.

(Also, sorry for not filling in the form, if I didn't give enough detail please let me know and I'll fill this in correctly tomorrow.)

radeksimko:
Closes #2336

@ziggythehamster @nathanielks @achille-roussel @bturbes - do you folks mind testing this with your configs and/or pasting it here so we can test it to make sure the bug doesn't re-appear?

## Test results

```
TF_ACC=1 go test ./aws -v -run=TestAccAWSEcsTaskDefinition -timeout 120m
=== RUN   TestAccAWSEcsTaskDefinition_basic
--- PASS: TestAccAWSEcsTaskDefinition_basic (57.24s)
=== RUN   TestAccAWSEcsTaskDefinition_withScratchVolume
--- PASS: TestAccAWSEcsTaskDefinition_withScratchVolume (35.02s)
=== RUN   TestAccAWSEcsTaskDefinition_withEcsService
--- PASS: TestAccAWSEcsTaskDefinition_withEcsService (149.91s)
=== RUN   TestAccAWSEcsTaskDefinition_withTaskRoleArn
--- PASS: TestAccAWSEcsTaskDefinition_withTaskRoleArn (38.40s)
=== RUN   TestAccAWSEcsTaskDefinition_withNetworkMode
--- PASS: TestAccAWSEcsTaskDefinition_withNetworkMode (66.97s)
=== RUN   TestAccAWSEcsTaskDefinition_constraint
--- PASS: TestAccAWSEcsTaskDefinition_constraint (33.79s)
=== RUN   TestAccAWSEcsTaskDefinition_changeVolumesForcesNewResource
--- PASS: TestAccAWSEcsTaskDefinition_changeVolumesForcesNewResource (55.91s)
=== RUN   TestAccAWSEcsTaskDefinition_arrays
--- PASS: TestAccAWSEcsTaskDefinition_arrays (34.10s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	471.371s
```

### DIFF
--- a/aws/ecs_task_definition_equivalency.go
+++ b/aws/ecs_task_definition_equivalency.go
@@ -3,8 +3,11 @@ package aws
 import (
 	"bytes"
 	"encoding/json"
+	"log"
 	"reflect"
+	"sort"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/private/protocol/json/jsonutil"
 	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/mitchellh/copystructure"
@@ -34,11 +37,14 @@ func ecsContainerDefinitionsAreEquivalent(def1, def2 string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+
 	canonicalJson2, err := jsonutil.BuildJSON(obj2)
 	if err != nil {
 		return false, err
 	}
 
+	log.Printf("[DEBUG] Comparing canonical definitions,\nFirst: %s\nSecond: %s\n",
+		canonicalJson1, canonicalJson2)
 	return bytes.Compare(canonicalJson1, canonicalJson2) == 0, nil
 }
 
@@ -47,6 +53,12 @@ type containerDefinitions []*ecs.ContainerDefinition
 func (cd containerDefinitions) Reduce() error {
 	for i, def := range cd {
 		// Deal with special fields which have defaults
+		if def.Cpu != nil && *def.Cpu == 0 {
+			def.Cpu = nil
+		}
+		if def.Essential == nil {
+			def.Essential = aws.Bool(true)
+		}
 		for j, pm := range def.PortMappings {
 			if pm.Protocol != nil && *pm.Protocol == "tcp" {
 				cd[i].PortMappings[j].Protocol = nil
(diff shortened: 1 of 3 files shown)


ziggythehamster:
I'm happy to test - how do I go about doing that?

radeksimko:
@ziggythehamster Try cloning the repository and building the binary locally: https://github.com/terraform-providers/terraform-provider-aws#building-the-provider

then in the directory with your ECS configs just `rm -rf ./.terraform && terraform init` to load the new binary and carry on with `terraform plan` which should present no diff.

aws/ecs_task_definition_equivalency.go
```diff
@@ -34,11 +37,14 @@ func ecsContainerDefinitionsAreEquivalent(def1, def2 string) (bool, error) {
...
 		return false, err
 	}
+
 	canonicalJson2, err := jsonutil.BuildJSON(obj2)
 	if err != nil {
 		return false, err
 	}
 
+	log.Printf("[DEBUG] Comparing canonical definitions,\nFirst: %s\nSecond: %s\n",
+		canonicalJson1, canonicalJson2)
```
catsby: nit: not sure if this is debugging that you left in unintentionally, however if intended, maybe we should log instead the result of the compare; `logical match` or `difference` etc. Right here we're just logging both and not actually saying if Terraform seems them as equal, that's left up to the ~user~ debugger

radeksimko: right, good point - I'm happy to add more context (i.e. whether it's a match).

The reason I left it there is because I had hard times comparing those JSONs in a failing test, where you can only get the original (non-canonical) JSONs which is also kind of intention, because this process should remain hidden as implementation detail.

PR #2344, opened 2017-11-17
bump aws-sdk-go version to v1.12.29

Puneeth-n:
Relates to #2341

PR #2349, opened 2017-11-17
resource/aws_sfn_state_machine: Support Update State machine call

### LINKED ISSUE #2341: Step function update state machine
AWS [released](https://aws.amazon.com/about-aws/whats-new/2017/11/aws-step-functions-adds-support-for-updating-state-machines/) update API for state machines. 

Also relates to the issue I created  hashicorp/terraform#15485

Puneeth-n:
Resolves #2341 

requires #2344 to be merged first

### DIFF
--- a/aws/resource_aws_sfn_state_machine.go
+++ b/aws/resource_aws_sfn_state_machine.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"fmt"
 	"log"
 	"time"
 
@@ -16,6 +17,7 @@ func resourceAwsSfnStateMachine() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceAwsSfnStateMachineCreate,
 		Read:   resourceAwsSfnStateMachineRead,
+		Update: resourceAwsSfnStateMachineUpdate,
 		Delete: resourceAwsSfnStateMachineDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
@@ -25,7 +27,6 @@ func resourceAwsSfnStateMachine() *schema.Resource {
 			"definition": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ForceNew:     true,
 				ValidateFunc: validateSfnStateMachineDefinition,
 			},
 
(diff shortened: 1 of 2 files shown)


aws/resource_aws_sfn_state_machine.go
```diff
@@ -125,6 +124,32 @@ func resourceAwsSfnStateMachineRead(d *schema.ResourceData, meta interface{}) er
...
+	}
+
+	_, err := conn.UpdateStateMachine(params)
+
+	if err != nil {
+
+		if awserr, ok := err.(awserr.Error); ok {
+			if awserr.Code() == "NotFoundException" || awserr.Code() == "StateMachineDoesNotExist" {
+				d.SetId("")
+				return nil
```
Ninir: In place of these 2 lines, it would be preferable to return an error, as in:

```
return fmt.Errorf("Error updating Step Function State Machine: %s", err)
```

Setting an ID to empty removes it from the state, step done by the Read part, always happening before updating/deleting

Puneeth-n:
@Ninir done :)

Ninir:
Could you add an acceptance test with 2 steps (one for creation, the other one for update)? thanks :)

Puneeth-n:
@Ninir I am not sure how I can check the `sfn` definition changes in the acceptance tests.

```
root@5f4b89633452:/go/src/github.com/terraform-providers/terraform-provider-aws# make testacc TESTARGS='-run=TestAccAWSSfnStateMachine_createUpdate'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAWSSfnStateMachine_createUpdate -timeout 120m
=== RUN   TestAccAWSSfnStateMachine_createUpdate
--- PASS: TestAccAWSSfnStateMachine_createUpdate (84.16s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	84.182s
```

Ninir:
@Puneeth-n We could store the output and use it afterwards, as in https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_flow_log_test.go#L15-L29

We could then check if the ID changed, and hash the structure to ensure that the content changed too. Thoughts?

Puneeth-n:
I used regex to check the output. after the state machine was updated. what do you think about it?

Ninir:
@Puneeth-n Depending on the regex written, this could be sufficient, nice idea :)

Puneeth-n:
@Ninir I have already included it in the Acceptance tests. Let me know if you want any changes.

Puneeth-n:
@Ninir Thanks a lot :) Wanted to get this feature into mainline so that we can get rid of our fork. :)

Issue #2270, opened 2017-11-14
Updating network_interface_id in aws_route fails


lethalpaga:
### Terraform Version
Terraform v0.10.8
provider.aws 1.2.0

### Affected Resource(s)
- aws_route

### Terraform Configuration Files
Config 1:
```hcl
provider "aws" {
  version             = "~>1.2.0"
}

resource "aws_route" "route_to_eni" {
  route_table_id         = "rtb-ba9acbbb"
  destination_cidr_block = "1.2.3.0/24"

  network_interface_id = "eni-35cadddd"
}
```
Config 2 (updated network_interface_id):
```hcl
provider "aws" {
  version             = "~>1.2.0"
}

resource "aws_route" "route_to_eni" {
  route_table_id         = "rtb-ba9acbbb"
  destination_cidr_block = "1.2.3.0/24"

  network_interface_id = "eni-35cadddd"
}
```

### Debug Output
https://gist.github.com/lethalpaga/28714fe7180b2debce09489c77ba1556

### Expected Behavior
The route should have been updated with the new ENI id

### Actual Behavior
The apply fails with
```
* aws_route.route_to_eni: InvalidInstanceID: There are multiple interfaces attached to instance 'i-0ec77fe2cdde1'. Please specify an interface ID for the operation instead.
	status code: 400, request id: 060b4661-c5e8-4254-92c0-ac1d13f21df7
```

### Steps to Reproduce

1. `terraform apply` with config1 to create the route
2. `terraform apply` with config2 to attempt updating it

### Important Factoids

* The route creation works fine. Updating it fails.
* In both config1 and config2 the ENI is attached to an instance that has 2 network interfaces attached.

### References
- Possibly related to GH-1426

raylu-stripe:
this bug is caused by https://github.com/hashicorp/terraform/pull/7686

YakDriver:
I was able to successfully update a route using new `aws_route` import ability in PR #5657 .  The steps I took were as follows:

1. Create route
2. Import the route with a new config
3. Apply

Terraform properly updated the route resource and I verified that the change occurred in AWS.

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  ~ aws_route.internal-default-route
      network_interface_id: "eni-0b5793f41da64217d" => "eni-07b3859117af14835"

Plan: 0 to add, 1 to change, 0 to destroy.
...
[...]

PR #2353, opened 2017-11-17
resource/aws_route: Support changing network_interface_id

raylu:
`network_interface_id` and `instance_id` are both optional and computed. If you specify one, the other is computed by `resourceAwsRouteSetResourceData`.

https://github.com/hashicorp/terraform/pull/7686 made it so that if you specified `instance_id`, applied, changed the `instance_id`, and applied again, we would ignore the computed `network_interface_id` (an ENI on the old instance) and update the `instance_id` correctly. This broke the opposite case (specify `network_interface_id`, apply, change, apply again).

This PR checks which has changed instead of preferring one over the other, fixing #2270. It depends on https://github.com/hashicorp/terraform/pull/16693.

### DIFF
--- a/aws/resource_aws_route.go
+++ b/aws/resource_aws_route.go
@@ -299,8 +299,8 @@ func prepareRouteUpdate(d *schema.ResourceData) (*ec2.ReplaceRouteInput, error) 
 		"egress_only_gateway_id",
 		"gateway_id",
 		"nat_gateway_id",
-		"network_interface_id",
 		"instance_id",
+		"network_interface_id",
 		"vpc_peering_connection_id",
 	}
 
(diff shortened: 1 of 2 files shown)


bcornils:
looks like we have a check failing. Has anyone had time to assess or review criticality of the failure?

bflad:
Hi folks 👋 It looks like this old PR has merge conflicts, fails TravisCI testing, and hasn't seen an update in a long time since last asked so I'm going to close it. If there is interest in this still, please feel free to create a new PR and we'll review it. 👍

raylu:
sorry for missing bcornils' message from june. as mentioned in the description, this PR depends on hashicorp/terraform#16693 (which introduces `TestResourceDataStateRaw` which I use here)

raylu:
@bflad / @bcornils could I get this re-opened? alternatively, could I get eyes on https://github.com/hashicorp/terraform/pull/16693 ?

PR #2360, opened 2017-11-18
New Resource: aws_dx_connection_association

atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsDxConnectionAssociation_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAwsDxConnectionAssociation_basic -timeout 120m
=== RUN   TestAccAwsDxConnectionAssociation_basic
--- PASS: TestAccAwsDxConnectionAssociation_basic (100.67s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	100.725s
```

### DIFF
--- a/aws/resource_aws_dx_connection_association.go
+++ b/aws/resource_aws_dx_connection_association.go
@@ -63,6 +63,10 @@ func resourceAwsDxConnectionAssociationRead(d *schema.ResourceData, meta interfa
 	if len(resp.Connections) != 1 {
 		return fmt.Errorf("Found %d DX connections for %s, expected 1", len(resp.Connections), d.Id())
 	}
+	if *resp.Connections[0].LagId != d.Get("lag_id").(string) {
+		d.SetId("")
+		return nil
+	}
 
 	return nil
 }
--- a/aws/resource_aws_dx_connection_association.go
+++ b/aws/resource_aws_dx_connection_association.go
@@ -62,18 +62,9 @@ func resourceAwsDxConnectionAssociationRead(d *schema.ResourceData, meta interfa
 		return nil
 	}
 	if len(resp.Connections) != 1 {
-		return fmt.Errorf("Number of DX Connection (%s) isn't one, got %d", connectionId, len(resp.Connections))
+		return fmt.Errorf("Found %d DX connections for %s, expected 1", len(resp.Connections))
 	}
-	if d.Id() != *resp.Connections[0].ConnectionId {
-		return fmt.Errorf("DX Connection (%s) not found", connectionId)
-	}
-	if resp.Connections[0].LagId == nil {
-		d.SetId("")
-		return nil
-	}
-	if *resp.Connections[0].LagId != d.Get("lag_id").(string) {
-		return fmt.Errorf("DX Connection (%s) is associated with unexpected LAG (%s)", connectionId, *resp.Connections[0].LagId)
-	}
+
 	return nil
 }
 
@@ -88,9 +79,6 @@ func resourceAwsDxConnectionAssociationDelete(d *schema.ResourceData, meta inter
 	resp, err := conn.DisassociateConnectionFromLag(input)
 	if err != nil {
 		return err
-	}
-	if resp.LagId != nil {
-		return fmt.Errorf("DX Connection (%s) is still associated with LAG", d.Id())
 	}
 
 	d.SetId("")
--- a/aws/resource_aws_dx_connection_association.go
+++ b/aws/resource_aws_dx_connection_association.go
@@ -1,0 +1,98 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/directconnect"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsDxConnectionAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsDxConnectionAssociationCreate,
+		Read:   resourceAwsDxConnectionAssociationRead,
+		Delete: resourceAwsDxConnectionAssociationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"connection_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"lag_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+	}
+}
+
+func resourceAwsDxConnectionAssociationCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dxconn
+
+	input := &directconnect.AssociateConnectionWithLagInput{
+		ConnectionId: aws.String(d.Get("connection_id").(string)),
+		LagId:        aws.String(d.Get("lag_id").(string)),
+	}
+	resp, err := conn.AssociateConnectionWithLag(input)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(*resp.ConnectionId)
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


aws/resource_aws_dx_connection_association.go
```diff
@@ -0,0 +1,98 @@
...
+		return fmt.Errorf("Number of DX Connection (%s) isn't one, got %d", connectionId, len(resp.Connections))
+	}
+	if d.Id() != *resp.Connections[0].ConnectionId {
+		return fmt.Errorf("DX Connection (%s) not found", connectionId)
+	}
+	if resp.Connections[0].LagId == nil {
+		d.SetId("")
+		return nil
+	}
+	if *resp.Connections[0].LagId != d.Get("lag_id").(string) {
```
radeksimko: As far as I understand there's 1-to-1 relationship between a connection and a LAG which this resource should represent, so shouldn't we instead treat this as if the resource (i.e. unique pair) is gone?

aws/resource_aws_dx_connection_association.go
```diff
@@ -0,0 +1,98 @@
...
+	resp, err := conn.DescribeConnections(input)
+	if err != nil {
+		return err
+	}
+	if len(resp.Connections) < 1 {
+		d.SetId("")
+		return nil
+	}
+	if len(resp.Connections) != 1 {
+		return fmt.Errorf("Number of DX Connection (%s) isn't one, got %d", connectionId, len(resp.Connections))
```
radeksimko: Just grammar nitpick - do you mind changing this to `Found %d DX connections for %s, expected 1`?

aws/resource_aws_dx_connection_association.go
```diff
@@ -0,0 +1,98 @@
...
+	input := &directconnect.DisassociateConnectionFromLagInput{
+		ConnectionId: aws.String(d.Id()),
+		LagId:        aws.String(d.Get("lag_id").(string)),
+	}
+
+	resp, err := conn.DisassociateConnectionFromLag(input)
+	if err != nil {
+		return err
+	}
+	if resp.LagId != nil {
```
radeksimko: Have you ever experienced this in reality? Is it just a matter of time until it becomes _actually_ disassociated or is it an erroneous/unrecoverable state? If it's the former then I'd suggest we just put a retry logic in place and wait until it's disassociated. If it's the latter then I guess this is a correct behaviour - I'm just trying to ensure that we know what we're doing here.

website/docs/r/dx_connection_association.html.markdown
```diff
@@ -0,0 +1,40 @@
+---
+layout: "aws"
+page_title: "AWS: aws_dx_connection_association"
+sidebar_current: "docs-aws-resource-dx-connection-association"
+description: |-
+  Associates a Connection with LAG.
```
radeksimko: Nitpick, but do you mind changing this to `Associates a Direct Connect Connection with a LAG.`?

website/docs/r/dx_connection_association.html.markdown
```diff
@@ -0,0 +1,40 @@
...
+---
+
+# aws_dx_connection_association
+
+Associates a Connection with LAG.
```
radeksimko: Nitpick, but do you mind changing this to `Associates a Direct Connect Connection with a LAG.`?

atsushi-ishibashi:
I removed some extra error handling.

aws/resource_aws_dx_connection_association.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	resp, err := conn.DescribeConnections(input)
+	if err != nil {
+		return err
+	}
+	if len(resp.Connections) < 1 {
+		d.SetId("")
+		return nil
+	}
```
radeksimko: I think there was a bit of misunderstanding about what I meant by my question about LAG ID mismatch - I wasn't implying it should be removed. I meant we should treat it the same way as if there is no connection - i.e. `d.SetId(""); return nil`.

aws/resource_aws_dx_connection_association_test.go
```diff
@@ -0,0 +1,87 @@
...
+			ConnectionId: aws.String(rs.Primary.ID),
+		}
+
+		resp, err := conn.DescribeConnections(input)
+		if err != nil {
+			return err
+		}
+		for _, v := range resp.Connections {
+			if *v.ConnectionId == rs.Primary.ID && v.LagId != nil {
+				return fmt.Errorf("Dx Connection (%s) is not diasociated with Lag", rs.Primary.ID)
```
radeksimko: nitpick: `is not dissociated from Lag`

aws/resource_aws_dx_connection_association.go
atsushi-ishibashi: Oh, I got it👍

radeksimko:
So I took a deeper look into how Direct Connect works with LAGs & connections and I think we'll need to add a for loop to read through all connections for the given connection ID as we cannot assume that there's only 1 connection per lag.

Basically we need to make the following config work:

```hcl
resource "aws_dx_connection" "example" {
  name = "example"
  bandwidth = "1Gbps"
  location = "EqSe2"
}

resource "aws_dx_lag" "example" {
  name = "example"
  connections_bandwidth = "1Gbps"
  location = "EqSe2"
  number_of_connections = 1
}

resource "aws_dx_lag" "example2" {
  name = "example2"
  connections_bandwidth = "1Gbps"
  location = "EqSe2"
  number_of_connections = 1
}

resource "aws_dx_connection_association" "example" {
  connection_id = "${aws_dx_connection.example.id}"
  lag_id = "${aws_dx_lag.example.id}"
}

resource "aws_dx_connection_association" "example2" {
  connection_id = "${aws_dx_connection.example.id}"
  lag_id = "${aws_dx_lag.example2.id}"
}
```
can you also add it as an acceptance test and make it pass, please?

atsushi-ishibashi:
Could I understand what you meant? @radeksimko 
```
TF_ACC=1 go test ./aws -v -run=TestAccAwsDxConnectionAssociation_multiConns -timeout 120m
=== RUN   TestAccAwsDxConnectionAssociation_multiConns
--- PASS: TestAccAwsDxConnectionAssociation_multiConns (41.34s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	41.386s
```

radeksimko:
Having a test for multiple connections is certainly helpful (thanks 👍 ), but I was talking about multiple LAGs. I just committed a failing test to you branch which hopefully explains what I mean.

You should see the following failure:

```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsDxConnectionAssociation_multiLags'
TF_ACC=1 go test ./aws -v -run=TestAccAwsDxConnectionAssociation_multiLags -timeout 120m
=== RUN   TestAccAwsDxConnectionAssociation_multiLags
--- FAIL: TestAccAwsDxConnectionAssociation_multiLags (44.62s)
	testing.go:503: Step 0 error: After applying this step and refreshing, the plan was not empty:

		DIFF:

		CREATE: aws_dx_connection_association.test2
		  connection_id: "" => "dxcon-ffo66i6d" (forces new resource)
		  lag_id:        "" => "dxlag-fhdhkf6q" (forces new resource)
...
```

Let me know if you need any further help in reproducing it.

atsushi-ishibashi:
@radeksimko 
At first, I added acctest which you provided above but I had the same error. On web console, the connection which is associated with a LAG doesn't show available connection list of another LAG.

radeksimko:
@atsushi-ishibashi I see, thanks for the explanation.

It looks like the AWS API is behaving unexpectedly in the sense that it just accepts any association and overrides the existing one. 😞 

I assume there isn't much we can do in this case - other than ask AWS to change/fix the API.

I'll merge it as is then - after removing my test.

Issue #2225, opened 2017-11-09
ssm_maintenance_window_target should support multiple targets


srhaber:
The SSM Maintenance Window Target resource can accept up to 5 different key/val tags.  However, the provider limits us to 1 key/val.

https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_ssm_maintenance_window_target.go#L35

This should be a simple fix in theory (just change MaxItems to 5).

atsushi-ishibashi:
Close

radeksimko:
Addressed in #2361 - thanks @atsushi-ishibashi !

PR #2361, opened 2017-11-18
r/aws_ssm_maintenance_window_target: Change MaxItems of targets

atsushi-ishibashi:
#2225 
Not yet testacc.
According to [Docs](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_RegisterTargetWithMaintenanceWindow.html#API_RegisterTargetWithMaintenanceWindow_RequestSyntax), minimum number is 0.
But I got a validation error and I'm asking aws-sdk-go.
```
ValidationException: 1 validation error detected: Value '[]' at 'targets' failed to satisfy constraint: Member must have length greater than or equal to 1
```

### DIFF
--- a/aws/resource_aws_ssm_maintenance_window_target.go
+++ b/aws/resource_aws_ssm_maintenance_window_target.go
@@ -30,7 +30,7 @@ func resourceAwsSsmMaintenanceWindowTarget() *schema.Resource {
 
 			"targets": {
 				Type:     schema.TypeList,
-				Optional: true,
+				Required: true,
 				ForceNew: true,
 				MaxItems: 5,
 				Elem: &schema.Resource{
--- a/aws/resource_aws_ssm_maintenance_window_target.go
+++ b/aws/resource_aws_ssm_maintenance_window_target.go
@@ -30,9 +30,9 @@ func resourceAwsSsmMaintenanceWindowTarget() *schema.Resource {
 
 			"targets": {
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
-				MaxItems: 1,
+				MaxItems: 5,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"key": {
--- a/aws/structure.go
+++ b/aws/structure.go
@@ -2686,7 +2686,7 @@ func flattenAwsSsmTargets(targets []*ssm.Target) []map[string]interface{} {
 
 	result := make([]map[string]interface{}, 0, len(targets))
 	for _, target := range targets {
-		t := make(map[string]interface{})
+		t := make(map[string]interface{}, 1)
 		t["key"] = *target.Key
 		t["values"] = flattenStringList(target.Values)
 
--- a/aws/structure.go
+++ b/aws/structure.go
@@ -2661,7 +2661,7 @@ func sliceContainsMap(l []interface{}, m map[string]interface{}) (int, bool) {
 }
 
 func expandAwsSsmTargets(d *schema.ResourceData) []*ssm.Target {
-	var targets []*ssm.Target
+	targets := make([]*ssm.Target, 0)
 
 	targetConfig := d.Get("targets").([]interface{})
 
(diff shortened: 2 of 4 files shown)


aws/structure.go
```diff
@@ -2685,13 +2685,13 @@ func flattenAwsSsmTargets(targets []*ssm.Target) []map[string]interface{} {
 	}
 
 	result := make([]map[string]interface{}, 0, len(targets))
-	target := targets[0]
+	for _, target := range targets {
+		t := make(map[string]interface{})
```
Ninir: To be more memory-efficient, we could set: `make(map[string]interface{}, 1)`, thoughts? :)

atsushi-ishibashi: It seems the same
```
func hoge1() {
	results := make([]map[string]interface{}, 100)
	for i := range results {
		t := make(map[string]interface{})
		t["key"], t["values"] = "key", "value"
		results[i] = t
	}
}

func hoge2() {
	results := make([]map[string]interface{}, 100)
	for i := range results {
		t := make(map[string]interface{}, 1)
		t["key"], t["values"] = "key", "value"
		results[i] = t
	}
}

func hoge3() {
	results := make([]map[string]interface{}, 100)
	for i := range results {
		t := map[string]interface{}{}
		t["key"], t["values"] = "key", "value"
		results[i] = t
	}
}

func BenchmarkHoge1(b *testing.B) {
	for i := 0; i < b.N; i++ {
		hoge1()
	}
}

func BenchmarkHoge2(b *testing.B) {
	for i := 0; i < b.N; i++ {
		hoge2()
	}
}

func BenchmarkHoge3(b *testing.B) {
	for i := 0; i < b.N; i++ {
		hoge3()
	}
}
```

```
go test -bench BenchmarkHoge -benchmem
BenchmarkHoge1-4   	   50000	     32459 ns/op	   33600 B/op	     200 allocs/op
BenchmarkHoge2-4   	   50000	     32529 ns/op	   33600 B/op	     200 allocs/op
BenchmarkHoge3-4   	   50000	     36636 ns/op	   33600 B/op	     200 allocs/op
```

aws/resource_aws_ssm_maintenance_window_target.go
```diff
@@ -30,9 +30,9 @@ func resourceAwsSsmMaintenanceWindowTarget() *schema.Resource {
 
 			"targets": {
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
```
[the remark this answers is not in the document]
atsushi-ishibashi: Sure👍

atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSSSMMaintenanceWindowTarget_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSSMMaintenanceWindowTarget_basic -timeout 120m
=== RUN   TestAccAWSSSMMaintenanceWindowTarget_basic
--- PASS: TestAccAWSSSMMaintenanceWindowTarget_basic (52.91s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	52.970s
```

aws/structure.go
Ninir: Third option is even better, interesting 🤔 
Still, can you go for the second one, since it has a fixed size?

website/docs/r/ssm_maintenance_window_target.html.markdown
```diff
@@ -36,11 +36,11 @@ The following arguments are supported:
 
 * `window_id` - (Required) The Id of the maintenance window to register the target with.
 * `resource_type` - (Required) The type of target being registered with the Maintenance Window. Possible values `INSTANCE`.
-* `targets` - (Required) The targets (either instances or tags). Instances are specified using Key=instanceids,Values=instanceid1,instanceid2. Tags are specified using Key=tag name,Values=tag value.
+* `targets` - (Required) The targets (either instances or tags). Instances are specified using Key=instanceids,Values=instanceid1,instanceid2. Tags are specified using Key=tag name,Values=tag value. Minimum number of 0 items. Maximum number of 5 items.
```
Ninir: Can you remove the first part of the change, since it is required? thanks

aws/structure.go
atsushi-ishibashi: Ok👍

PR #2386, opened 2017-11-21
resource/aws_db_instance: Allow ARN for the replicate_source_db when in the same region

pccowboy:
I believe this addresses a possible bug we ran into that was triggered by https://github.com/terraform-providers/terraform-provider-aws/pull/865/files and mentioned at https://groups.google.com/forum/#!topic/terraform-tool/zX-bWBZgViI

If you pass both an ARN and a kms_key_id to create a replica in the same region as the master, you get an error of the form "* aws_db_instance.read_replica: Error creating DB Instance: InvalidParameterCombination: Your request does not require the preSignedUrl parameter. Please remove the preSignedUrl parameter and try your request again."

I think this is because of AWS code - in aws/aws-sdk-go/service/rds/customizations.go, a preSignedUrl will get added to the CreateDBInstanceReadReplicaInput if the SourceRegion is set for the replica.  If SourceRegion is nil, the presignedURL is skipped.
```
     58         if originParams.SourceRegion == nil || originParams.PreSignedUrl != nil || originParams.DestinationRegion != nil {
     59                 return
     60         }
```

We are still verifying that this is working for our installation, hence the WIP tag.  We are able to create the same-region replica reliably (after push c84c16d), but we are getting a destroy-add on the next plan for the replica.  We are tracking that down currently.

I've modified the test for creating a replica, and added two more, one for the replica created with a source ARN in the same region, and one for creating a replica from a source ARN in a different region.  I have not yet run the tests, however.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -858,7 +858,7 @@ func resourceAwsDbInstanceRead(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("[DEBUG] Error setting replicas attribute: %#v, error: %#v", replicas, err)
 	}
 
-	//  If an ARN was passed in, do NOT use what AWS passes back for replicaste_source_id,
+	//  If an ARN was passed in, do NOT use what AWS passes back for replicate_source_id,
 	//  as it passes back the master's ID-
 	//  see https://github.com/terraform-providers/terraform-provider-aws/issues/2399
 	_, arnErr := arn.Parse(d.Get("replicate_source_db").(string))
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -420,11 +420,11 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 		//
 		// go figure, eh?
 		//
+		replicaRegion := meta.(*AWSClient).region
+
 		arnParts, arnErr := arn.Parse(d.Get("replicate_source_db").(string))
 		if arnErr == nil {
-			var replicaRegion string
-			replicaRegion = (string)(*opts.AvailabilityZone)
-			if arnParts.Region != replicaRegion[0:len(replicaRegion)-1] {
+			if arnParts.Region != replicaRegion {
 				opts.SourceRegion = aws.String(arnParts.Region)
 			}
 		}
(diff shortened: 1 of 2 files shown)


pccowboy:
@Ninir Can you take a look at this latest, and tell me if it looks like a valid fix for #2399 ?  If it looks like a fix y'all might take, we'll test it out, since testing it costs some money.

Otherwise, any pointers on producing a fix?  This is not blocking us, but it did make us create a module specific to same-region replicas, which seems less than ideal.

pccowboy:
@Ninir ping?  We are using this fix currently, and it seems to be working.  Any feedback y'all have on the current PR would be appreciated.

pccowboy:
@radeksimko @Ninir Hello, can we get your feedback on this PR, please?

aws/resource_aws_db_instance.go
```diff
@@ -407,11 +437,11 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			opts.DBSubnetGroupName = aws.String(attr.(string))
 		}
 
+		// TODO: Only allow this param if the master is not encrypted or
```
jen20: Is this still an outstanding concern?

aws/resource_aws_db_instance.go
```diff
@@ -399,6 +401,34 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			opts.AvailabilityZone = aws.String(attr.(string))
 		}
 
+		//
+		// If we are called with a Source DB ARN, and the ARN is a different region
```
jen20: :+1: for the block comment explaining the rationale here.

aws/resource_aws_db_instance.go
pccowboy: @jen20 I wrote that based on the ambiguity in the AWS documentation I describe above.  When creating a replica, they seem to have two situations where they allow an encryption key to be specified for a replica.  When creating an encrypted replica in the same region as the master, you cannot specify the key.

I had thought that maybe we should show an error if a key was specified for a same-region replica.

However, maybe it is best to let the AWS API throw an error if this param is used when it is not allowed.

Unless you have advice on which way to go, I can take those lines out.  The PR is working as-is for us.

pccowboy:
@radeksimko Is there anything about this PR that is blocking it from being merged in?  I am happy to make whatever adjustments are needed.

pccowboy:
@radeksimko bump -anything I can do to move this through?

pccowboy:
@radeksimko @jen20   Actually, I have a cleaner branch, where I removed all the duplicate commits.  I think I am going to close this PR, and open a new one from that cleaner branch.

PR #2387, opened 2017-11-21
Fixed the default values in the lb_target_group document

jun06t:
The health check default values of ``lb_target_group`` are changed in the following commit.
https://github.com/terraform-providers/terraform-provider-aws/commit/79467dc32944ceb42acaa60bb91da86fa974ef22
So i fixed the document.

### DIFF
--- a/website/docs/r/lb_target_group.html.markdown
+++ b/website/docs/r/lb_target_group.html.markdown
@@ -60,8 +60,8 @@ Health Check Blocks (`health_check`) support the following:
 * `port` - (Optional) The port to use to connect with the target. Valid values are either ports 1-65536, or `traffic-port`. Defaults to `traffic-port`.
 * `protocol` - (Optional) The protocol to use to connect with the target. Defaults to `HTTP`.
 * `timeout` - (Optional) The amount of time, in seconds, during which no response means a failed health check. Defaults to 5 seconds.
-* `healthy_threshold` - (Optional) The number of consecutive health checks successes required before considering an unhealthy target healthy. Defaults to 5.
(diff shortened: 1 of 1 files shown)


website/docs/r/lb_target_group.html.markdown
```diff
@@ -59,9 +59,9 @@ Health Check Blocks (`health_check`) support the following:
 * `path` - (Optional) The destination for the health check request. Default `/`.
 * `port` - (Optional) The port to use to connect with the target. Valid values are either ports 1-65536, or `traffic-port`. Defaults to `traffic-port`.
 * `protocol` - (Optional) The protocol to use to connect with the target. Defaults to `HTTP`.
-* `timeout` - (Optional) The amount of time, in seconds, during which no response means a failed health check. Defaults to 5 seconds.
-* `healthy_threshold` - (Optional) The number of consecutive health checks successes required before considering an unhealthy target healthy. Defaults to 5.
-* `unhealthy_threshold` - (Optional) The number of consecutive health check failures required before considering the target unhealthy. Defaults to 2.
+* `timeout` - (Optional) The amount of time, in seconds, during which no response means a failed health check. Defaults to 10 seconds.
```
Ninir: Can you just remove this part of the fix please?

@catsby is working on that in https://github.com/terraform-providers/terraform-provider-aws/pull/2380/files, so this will bring a conflict :)

Thanks!

jun06t: Thanks for your review :-)
I've removed the fix of `timeout` part to avoid a conflict.

Issue #2279, opened 2017-11-14
aws_iam_role.force_detach_policies no longer works


hashibot:
_This issue was originally opened by @aleybovich as hashicorp/terraform#16637. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Reproduce: 
1. Create an aws role with terraform

 ```
resource "aws_iam_role" "sample-instance-role" {
    name               = "sample-instance-role"
    path               = "/"
    force_detach_policies = true
    ... other stuff ...
}
```

2. Manually add another policy to that role (e.g. via aws console)
3. run `terraform destroy`

You get an error `sample-instance-role: DeleteConflict: Cannot delete entity, must detach all policies first.`

`force_detach_policies` feature has been added a few months ago specifically for this purpose - it detaches all policies that weer added outside of terraform so that the role can be deleted. This used to work briefly in one or two versions but it no longer does in 0.10.7

Documentation: https://www.terraform.io/docs/providers/aws/r/iam_role.html#force_detach_policies

atsushi-ishibashi:
@aleybovich There is a possibility that policies is truncated because a lot of policies are attached.
Could you share how many policies are attached?
https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#ListAttachedRolePoliciesOutput

aleybovich:
There are five policies that get attached to all aws roles automatically
via Turbot, after those roles are added. That was the reason that I've
filed the initial change request a year or so ago and it was implemented.
It used to work for a bit but now no longer does.

atsushi-ishibashi:
@aleybovich 
https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListRolePolicies.html
> Lists the names of the inline policies that are embedded in the specified IAM role. 
> An IAM role can also have managed policies attached to it. To list the managed policies that are attached to a role, use ListAttachedRolePolicies. 

Probably inline policies are still attached. For `force_detach_policies`, `ListAttachedRolePolicies` is used.

aleybovich:
@atsushi-ishibashi - the only inline policy that role has is the one I add in the same terraform template; all Turbot policies are attached.

atsushi-ishibashi:
@aleybovich I submitted PR to fix this👍

aleybovich:
Thank you! Whats the usual turnaround for a pull request like this one?

[...]

PR #2388, opened 2017-11-21
r/iam_role: Delete inline policies when force_detach_policies = true

atsushi-ishibashi:
Fix: #2279

### DIFF
--- a/aws/resource_aws_iam_role.go
+++ b/aws/resource_aws_iam_role.go
@@ -285,16 +285,22 @@ func resourceAwsIamRoleDelete(d *schema.ResourceData, meta interface{}) error {
 				}
 			}
 		}
+
 		// For inline policies
-		inlinePoliciesResp, err := iamconn.ListRolePolicies(&iam.ListRolePoliciesInput{
+		rolePolicyNames := make([]*string, 0)
+		err = iamconn.ListRolePoliciesPages(&iam.ListRolePoliciesInput{
 			RoleName: aws.String(d.Id()),
+		}, func(page *iam.ListRolePoliciesOutput, lastPage bool) bool {
+			for _, v := range page.PolicyNames {
+				rolePolicyNames = append(rolePolicyNames, v)
+			}
+			return len(page.PolicyNames) > 0
 		})
 		if err != nil {
 			return fmt.Errorf("Error listing inline Policies for IAM Role (%s) when trying to delete: %s", d.Id(), err)
 		}
-		// Loop and remove the Policies from the Role
-		if len(inlinePoliciesResp.PolicyNames) > 0 {
-			for _, pname := range inlinePoliciesResp.PolicyNames {
+		if len(rolePolicyNames) > 0 {
+			for _, pname := range rolePolicyNames {
 				_, err := iamconn.DeleteRolePolicy(&iam.DeleteRolePolicyInput{
 					PolicyName: pname,
 					RoleName:   aws.String(d.Id()),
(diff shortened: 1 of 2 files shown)


atsushi-ishibashi:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSIAMRole_force_detach_policies'
TF_ACC=1 go test ./aws -v -run=TestAccAWSIAMRole_force_detach_policies -timeout 120m
=== RUN   TestAccAWSIAMRole_force_detach_policies
--- PASS: TestAccAWSIAMRole_force_detach_policies (40.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	41.050s
```

atsushi-ishibashi:
the current `force_detach_policies` use `ListAttachedRolePolicies` to find policies which will be deatached.
But [ListRolePolicies](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListRolePolicies.html) says
> Lists the names of the inline policies that are embedded in the specified IAM role.
An IAM role can also have managed policies attached to it. To list the managed policies that are attached to a role, use ListAttachedRolePolicies. 

So when inlined policies are attached to role, you got error `DeleteConflict: Cannot delete entity, must detach all policies first.` even though `force_detach_policies=true`

In this PR, delete inline policies when `force_detach_policies=true`. I wonder if `force_detach_policies` is appropriate because it will delete inline policies...

aleybovich:
@atsushi-ishibashi `force_detach_policies` is definitely appropriate - the idea of that flag is to be able to destroy a role with policies attached, no matter whether they are inline or separate.

aws/resource_aws_iam_role_test.go
```diff
@@ -375,3 +394,71 @@ POLICY
 }
 `, rName)
 }
+
+func testAccAWSIAMRoleConfig_force_detach_policies(rName string) string {
+	return fmt.Sprintf(`
+resource "aws_iam_role_policy" "test" {
```
radeksimko: Technically we're not really testing the new functionality here IMO, because the policy will get destroyed first due to its relationship and `force_destroy` won't get chance to ever take action.
I think we'd need to remove those policies from state as part of the test in order to verify it actually works.

aws/resource_aws_iam_role.go
```diff
@@ -285,6 +285,25 @@ func resourceAwsIamRoleDelete(d *schema.ResourceData, meta interface{}) error {
 				}
 			}
 		}
+		// For inline policies
+		inlinePoliciesResp, err := iamconn.ListRolePolicies(&iam.ListRolePoliciesInput{
```
radeksimko: Aren't we going to miss some policies here potentially due to pagination? How do you feel about using `ListRolePoliciesPages` instead?

aws/resource_aws_iam_role_test.go
atsushi-ishibashi: That's right👍
But I couldn't have an idea to realize it... Could you give me advice?

radeksimko: So I think the easier way is to add policies outside of the config, instead of removing them from state. We could add the following to your new test:

```
testAccAddAWSIAMRolePolicy(&conf),
```
which would just call the API to add a policy outside of any Terraform CRUD.

Testing framework will then attempt to destroy the role and exercise this new functionality.

atsushi-ishibashi:
@radeksimko Ok👍

bflad:

PR #2393, opened 2017-11-21
Fix conflict exception in API gateway method response

### LINKED ISSUE #483: aws_api_gateway_method_response Concurrency
_This issue was originally opened by @steve-gray as hashicorp/terraform#11395. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

When creating multiple method response codes that live beneath the same method, Terraform will typically fail due to concurrency/conflict errors on the AWS side. It appears that AWS (whilst not documented anywhere visible) is treating the api_gateway_method as the concurrency boundary.

### Terraform Version
0.8.2

### Affected Resource(s)
- aws_api_gateway_method_response

### Expected Behavior
Multiple response codes should be created correctly for an API response

### Actual Behavior
Fails with error such as

````
 aws_api_gateway_method_response.dummy_function_response1: Error creating API Gateway Method Response: ConflictException: Response already exists for this resource
	status code: 409, request id: 56acda62-e292-11e6-b56a-85b5b33b47bd
* aws_api_gateway_method_response.dummy_function_response3: Error creating API Gateway Method Response: ConflictException: Response already exists for this resource
	status code: 409, request id: 56ae3a9c-e292-11e6-8218-a9497181c142
* aws_api_gateway_method_response.http409: ConflictException: Unable to complete operation due to concurrent modification. Please try again later.
	status code: 409, request id: 56acb348-e292-11e6-8d8e-77f3e0ae3f54
* aws_api_gateway_method_response.http401: ConflictException: Unable to complete operation due to concurrent modification. Please try again later.
	status code: 409, request id: 56ac6527-e292-11e6-8d8e-77f3e0ae3f54
* aws_api_gateway_method_response.dummy_function_response2: Error creating API Gateway Method Response: ConflictException: Unable to complete operation due to concurrent modification. Please try again later.
	status code: 409, request id: 56ac65c5-e292-11e6-99f9-efaae93431b6
````

### Steps to Reproduce
terraform apply where you have an API gateway operation with multiple (2+) HTTP response codes defined which are new and require creating. Subsequent attempts *may* work as individual responses get created, but will require many retries before the race conditions are not encountered.

betabandido:
This hopefully fixes #483 

I have added a mutex to prevent concurrent method response creations. As this does not help with simultaneous executions of terraform (e.g., multiple developers deploying to the same AWS account), I have also added a retry block.

@radeksimko @steve-gray Does this look good to you?

### DIFF
--- a/aws/resource_aws_api_gateway_method_response.go
+++ b/aws/resource_aws_api_gateway_method_response.go
@@ -12,7 +12,10 @@ import (
 	"github.com/aws/aws-sdk-go/service/apigateway"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"sync"
 )
+
+var resourceAwsApiGatewayMethodResponseMutex = &sync.Mutex{}
 
 func resourceAwsApiGatewayMethodResponse() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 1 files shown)


steve-gray:
@betabandido - LGTM. This was the bane of my life when I'm building terraform that deploys API Gateway instances.

I'd been running a hacked version of the provider locally with essentially the same change applied - though I used really aggressive retries, rather than a lock, because I was looking to go fast where possible. Would happily switch over to this though.

PR #2398, opened 2017-11-22
r/aws_instance: Set placement_group in state on read if available

bflad:
Closes #2392.

New test:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSInstance_placementGroup'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSInstance_placementGroup -timeout 120m
=== RUN   TestAccAWSInstance_placementGroup
--- PASS: TestAccAWSInstance_placementGroup (76.91s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	76.948s
```

Hopefully passing other tests still, e.g.
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSInstance_vpc'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSInstance_vpc -timeout 120m
=== RUN   TestAccAWSInstance_vpc
--- PASS: TestAccAWSInstance_vpc (95.32s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	95.352s
```

PR #2400, opened 2017-11-22
d/aws_instance: Set placement_group if available

bflad:
This is complementary PR to #2398 since the read functions are different between the resource and data source. The [data source docs](https://www.terraform.io/docs/providers/aws/d/instance.html#placement_group) already note that `placement_group` may be set.

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSInstanceDataSource_PlacementGroup'
TF_ACC=1 go test ./aws -v -run=TestAccAWSInstanceDataSource_PlacementGroup -timeout 120m
=== RUN   TestAccAWSInstanceDataSource_PlacementGroup
--- PASS: TestAccAWSInstanceDataSource_PlacementGroup (84.20s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	84.240s
```

### DIFF
--- a/aws/data_source_aws_instance_test.go
+++ b/aws/data_source_aws_instance_test.go
@@ -179,14 +179,16 @@ func TestAccAWSInstanceDataSource_VPC(t *testing.T) {
 }
 
 func TestAccAWSInstanceDataSource_PlacementGroup(t *testing.T) {
+	rStr := acctest.RandString(5)
+
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccInstanceDataSourceConfig_PlacementGroup,
+				Config: testAccInstanceDataSourceConfig_PlacementGroup(rStr),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("data.aws_instance.foo", "placement_group", "testAccInstanceDataSourceConfig_PlacementGroup"),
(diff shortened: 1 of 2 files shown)


aws/data_source_aws_instance_test.go
```diff
@@ -429,6 +444,42 @@ data "aws_instance" "foo" {
...
+  }
+}
+
+resource "aws_subnet" "foo" {
+  cidr_block = "10.1.1.0/24"
+  vpc_id = "${aws_vpc.foo.id}"
+}
+
+resource "aws_placement_group" "foo" {
+  name = "testAccInstanceDataSourceConfig_PlacementGroup"
```
Ninir: We will need to add some name randomization here, so that we avoid conflicts when running tests

bflad:
Comment addressed! Passes testing for me still:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSInstanceDataSource_PlacementGroup'
TF_ACC=1 go test ./aws -v -run=TestAccAWSInstanceDataSource_PlacementGroup -timeout 120m
=== RUN   TestAccAWSInstanceDataSource_PlacementGroup
--- PASS: TestAccAWSInstanceDataSource_PlacementGroup (88.35s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	88.392s
```

stack72:
@bflad / @Ninir I believe the documentation also needs to be updated as well - currently, for data_aws_instance, we list the outputs AFAICT

Thoughts?

Ninir:
HI @stack72 

You're indeed right. @bflad could you also add the entry in the related documentation?

Thanks!

bflad:
Are you referring to a location different than https://www.terraform.io/docs/providers/aws/d/instance.html#placement_group ?

bflad:
@stack72 / @Ninir simple question above. Let me know if/where anything else needs adjustment. 😄

PR #2409, opened 2017-11-22
r/aws_db_instance: Move create/delete/update pending states to their own variables and add storage-optimization

### LINKED ISSUE #2408: Bug when RDS is update from storage type gp2 to io1
Hi,

I am using the latest version of terraform. 
When I am updating my database from storage_type gp2 to io1, Terraform is incorrectly stopping when the RDS status is storage-optimization. Terraform quits instead of waiting for the process to finish.

```
module.rds.aws_db_instance.default: Modifying... (ID: db-service-data)
  allow_major_version_upgrade:       "" => "true"
  apply_immediately:                 "" => "true"
  engine_version:                    "5.6.27" => "5.6.37"
  final_snapshot_identifier:         "" => "db-service-data-finalsnapshot"
  iops:                              "0" => "1000"
  skip_final_snapshot:               "true" => "false"
  storage_type:                      "gp2" => "io1"
```

Message : 

`* aws_db_instance.default: unexpected state 'storage-optimization', wanted target 'available'. last error: %!s(<nil>)`

bflad:
Closes #2408 

There are some things that should be noted though, according to the [DB Instance Status docs](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Status.html):
* For `storage-optimization`: "The storage optimization process is usually short, but can sometimes take up to and even beyond 24 hours." Likely this means this will generally hit the timeout. Should this status be moved to the `resource.StateChangeConf` `Target` instead?
* Any reason why `configuring-enhanced-monitoring` is not in delete pending states?
* I noticed we don't have statuses like `stopping` and `starting` in any of the pending states, should we add them? There might be others too.

Thoughts and opinions on the above are appreciated!

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -525,8 +525,7 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 				"[INFO] Waiting for DB Instance to be available")
 
 			stateConf := &resource.StateChangeConf{
-				Pending: []string{"creating", "backing-up", "modifying", "resetting-master-credentials",
-					"maintenance", "renaming", "rebooting", "upgrading"},
+				Pending:    resourceAwsDbInstanceCreatePendingStates,
 				Target:     []string{"available"},
 				Refresh:    resourceAwsDbInstanceStateRefreshFunc(d, meta),
 				Timeout:    d.Timeout(schema.TimeoutCreate),
@@ -687,8 +686,7 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 		"[INFO] Waiting for DB Instance to be available")
 
 	stateConf := &resource.StateChangeConf{
-		Pending: []string{"creating", "backing-up", "modifying", "resetting-master-credentials",
-			"maintenance", "renaming", "rebooting", "upgrading", "configuring-enhanced-monitoring"},
+		Pending:    resourceAwsDbInstanceCreatePendingStates,
 		Target:     []string{"available"},
 		Refresh:    resourceAwsDbInstanceStateRefreshFunc(d, meta),
 		Timeout:    d.Timeout(schema.TimeoutCreate),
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @bflad 
thanks for the PR! Replies below.

> * For `storage-optimization`: "The storage optimization process is usually short, but can sometimes take up to and even beyond 24 hours." Likely this means this will generally hit the timeout. Should this status be moved to the `resource.StateChangeConf` `Target` instead?

Moving it to `Target` (during creation and updates, _not_ deletion) sounds like a reasonable thing to do, given described circumstances.

> * Any reason why `configuring-enhanced-monitoring` is not in delete pending states?

I can't think of any reason, probably just a new state that popped up recently and nobody noticed. Do you mind adding it?

> * I noticed we don't have statuses like `stopping` and `starting` in any of the pending states, should we add them? There might be others too.

Sounds like a good idea, please add them. 👍 

----

In regards to the code, I have mixed feelings about decoupling the slices from the CRUD. I don't see much value in such decoupling and I'm slightly more inclined to keeping it as is (i.e. within functions), but I won't block the PR on this - just a thought/opinion to consider.

bflad:
> I have mixed feelings about decoupling the slices from the CRUD. I don't see much value in such decoupling and I'm slightly more inclined to keeping it as is (i.e. within functions)

I was on the fence about this, however I was thinking it was nice to:
* Have the documentation link in one place
* Have the states in one place to notice items missing/not changing in all places consistently across the various usages. I think this is easier to maintain when Amazon decides to further refine these.
* At the very least make the slices more readable -- a blob of unalphabetized states across a line or two is harder to quickly understand in my opinion. We could just as easily do the same line breaks for the slice items, but now the function is a lot more LoC when it could just be stuck somewhere else... hence separated into its own variables. 😄 

I can change it back though and add the documentation in each of the spots if you wish. I definitely don't want to hold this up. I'll address the other items in the meantime.

bflad:
PR updated! 🚀

yelvert:
So when can 1.5.1 be expected to be released? Currently I can't even create Postgres instances with gp2 storage.

PR #2413, opened 2017-11-22
Updated documentation to reflect ALB requires path and matcher

blaines:
Followup to #2251, 1.3.0 (November 16, 2017)


blaines:
Ultimately I feel like the code is a bit misleading as it's conditionally optional based on the protocol, which defaults to HTTP, which requires `matcher` and `path`.
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lb_target_group.go#L134
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lb_target_group.go#L171

Ninir:
Hi @blaines 

Could you rebase your PR against master so that we can get this merged?

Thanks!

Issue #2402, opened 2017-11-22
update aws_db_subnet_group/aws_db_parameter_group fail (with new resource required) 


hsvlz:
### Terraform Version
Terraform v0.11.0
+ provider.aws v1.2.0
+ provider.template v1.0.0

### Affected Resource(s)
aws_db_instance (postgres)
aws_db_parameter_group
aws_db_subnet_group

### Debug Output
```
-/+ module.central-servers.aws_db_parameter_group.default (new resource required)
      id:                    "rds-pg" => <computed> (forces new resource)
      arn:                   "arn:aws:rds:eu-west-1:018597388725:pg:rds-pg" => <computed>
      description:           "Managed by Terraform" => "Managed by Terraform"
      family:                "postgres9.6" => "postgres9.6"
      name:                  "rds-pg" => "alba-rds-pg" (forces new resource)
      name_prefix:           "" => <computed>

-/+ module.central-servers.aws_db_subnet_group.postgresql_deployment (new resource required)
      id:                    "postgresql_deployment" => <computed> (forces new resource)
      arn:                   "arn:aws:rds:eu-west-1:018597388725:subgrp:postgresql_deployment" => <computed>
      description:           "Managed by Terraform" => "Managed by Terraform"
      name:                  "postgresql_deployment" => "alba_postgresql_deployment" (forces new resource)
      name_prefix:           "" => <computed>
      subnet_ids.#:          "2" => "2"
      subnet_ids.1695880707: "subnet-a1ccb8e8" => "subnet-a1ccb8e8"
      subnet_ids.3025816245: "subnet-f0187497" => "subnet-f0187497"
      tags.%:                "1" => "1"
      tags.Name:             "alba" => "alba"

  ~ module.central-servers.module.postgresql_deployment.aws_db_instance.postgresql
      db_subnet_group_name:  "postgresql_deployment" => "alba_postgresql_deployment"
      parameter_group_name:  "rds-pg" => "alba-rds-pg"
```

### Error
```
2 error(s) occurred:

* module.central-servers.aws_db_subnet_group.postgresql_deployment (destroy): 1 error(s) occurred:

* aws_db_subnet_group.postgresql_deployment: InvalidDBSubnetGroupStateFault: Cannot delete the subnet group 'postgresql_deployment' because at least one database instance: alba-deployment-db is still using it.
        status code: 400, request id: x
* module.central-servers.aws_db_parameter_group.default (destroy): 1 error(s) occurred:

[...]

PR #2415, opened 2017-11-23
[WIP, Need Help]r/db_parameter_group: Add force_detach to aws_db_parameter_group

atsushi-ishibashi:
To fix: #2402 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSDBParameterGroup_withDBInstance'
TF_ACC=1 go test ./aws -v -run=TestAccAWSDBParameterGroup_withDBInstance -timeout 120m
=== RUN   TestAccAWSDBParameterGroup_withDBInstance
--- PASS: TestAccAWSDBParameterGroup_withDBInstance (724.81s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	724.857s
```

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -1153,6 +1153,21 @@ func resourceAwsDbInstanceStateRefreshFunc(
 	}
 }
 
+func dbInstanceRefreshStatusFunc(conn *rds.RDS, id string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		opts := &rds.DescribeDBInstancesInput{
+			DBInstanceIdentifier: aws.String(id),
+		}
+
+		resp, err := conn.DescribeDBInstances(opts)
+		if err != nil {
+			return nil, "failed", err
+		}
+
+		return resp, *resp.DBInstances[0].DBInstanceStatus, nil
+	}
+}
+
 func buildRDSARN(identifier, partition, accountid, region string) (string, error) {
 	if partition == "" {
 		return "", fmt.Errorf("Unable to construct RDS ARN because of missing AWS partition")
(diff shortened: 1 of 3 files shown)


atsushi-ishibashi:
If you created instance with the parameter group manually on the console after creating the parameter group from terraform, this operation will affect resource outside terraform.
So I added `force_detach` option with default false.

atsushi-ishibashi:
I have one problem. I have to set another parameter group and currently I set `default.mysql5.6`... 😢 
I had an idea that copying the parameter group which is used by db and modifying with the copy one. However, I have no idea how to delete it after `resourceAwsDbParameterGroupUpdate`.
Should I store it as global variables such like `parameterGroupNamesToDelete` and refer it in `resourceAwsDbParameterGroupUpdate`?
@Ninir @radeksimko Did you experience the similar problem?
https://github.com/terraform-providers/terraform-provider-aws/pull/2415/files#diff-497d6ad0ec0d61809a57fc6ff6ffaadbR268

radeksimko:
Hi @atsushi-ishibashi 

I don't believe this is solving #2402 for the reasons I mentioned in https://github.com/terraform-providers/terraform-provider-aws/issues/2402#issuecomment-347028624

Also I don't think that a parameter group resource should be modifying DB instance unless there's a strong reason for it. I'd try to avoid it because users would otherwise end up with two resources fighting with each other about who's right (in this case `aws_db_instance.parameter_group_name` and `aws_db_parameter_group` deletion).

Unless there's another good reason for having this functionality I'm inclined to close this PR.

Thanks for all the effort you invested into this either way!

atsushi-ishibashi:
Yeah, I absolutely agree with that `lifecycle` is better solution. I'll close this PR:bow:

PR #2422, opened 2017-11-24
Adding "Data Source" header to data source resources.

RavenAlly:
Related issue: https://github.com/hashicorp/terraform-website/issues/18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,15 @@ website/vendor
 
 # Keep windows files with windows line endings
 *.winfile eol=crlf
+/.vs
+/.vs/VSWorkspaceState.json
+/.vs/terraform-provider-aws/v15/.suo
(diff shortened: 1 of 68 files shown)


bflad:
:+1: personally. Thank you! Totally not necessary in this PR but it'd be nice to clean up the extraneous slashes in the titles too

Ninir:
Hi @RavenAlly 

I do think this is a good idea to better identify whether you're on a resource or a data source 👍 

Since it needs to be reflected among other providers, just would like to call @sethvargo here and know what he thinks of it if you don't mind :) (perhaps he even has an approach to deal with this).

catsby:
Thank you! I'm closing this in favor of #2713 that I just opened, it's this PR but with a commit where I merge master and resolve the above merge conflicts.

Issue #2423, opened 2017-11-24
email_verification_message of cognito_user_pool change every time


atsushi-ishibashi:
In `testAccAWSCognitoUserPoolConfig_withVerificationMessageTemplate`, I changed `verification_message_template.email_message` to `"Foo {####} Hoge"` and keep `email_verification_message` as `"Foo {####} Bar"`. Of course I chaged `TestCheckResourceAttr` as well.
I run `TestAccAWSCognitoUserPool_withVerificationMessageTemplate` and the below error happened.
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSCognitoUserPool_withVerificationMessageTemplate'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoUserPool_withVerificationMessageTemplate -timeout 120m
=== RUN   TestAccAWSCognitoUserPool_withVerificationMessageTemplate
--- FAIL: TestAccAWSCognitoUserPool_withVerificationMessageTemplate (22.51s)
	testing.go:503: Step 0 error: After applying this step, the plan was not empty:
		
		DIFF:
		
		UPDATE: aws_cognito_user_pool.pool
		  email_verification_message: "Foo {####} Hoge" => "Foo {####} Bar"
		
		STATE:
		
		aws_cognito_user_pool.pool:
		  ID = us-west-2_VqO35We5K
		  admin_create_user_config.# = 1
		  admin_create_user_config.0.allow_admin_create_user_only = false
		  admin_create_user_config.0.invite_message_template.# = 0
		  admin_create_user_config.0.unused_account_validity_days = 7
		  creation_date = 2017-11-24T04:10:17Z
		  device_configuration.# = 0
		  email_configuration.# = 0
		  email_verification_message = Foo {####} Hoge
		  email_verification_subject = FooBar {####}
		  lambda_config.# = 0
		  last_modified_date = 2017-11-24T04:10:17Z
		  mfa_configuration = OFF
		  name = terraform-test-pool-vg4qv
		  password_policy.# = 1
		  password_policy.0.minimum_length = 8
		  password_policy.0.require_lowercase = true
		  password_policy.0.require_numbers = true
		  password_policy.0.require_symbols = true
		  password_policy.0.require_uppercase = true
		  sms_configuration.# = 0
		  sms_verification_message = {####} Baz
		  tags.% = 0
		  verification_message_template.# = 1
[...]

PR #2425, opened 2017-11-24
r/cognito_user_pool: Add ConflictsWith to a few fields

atsushi-ishibashi:
Related: #2423

### DIFF
--- a/aws/resource_aws_cognito_user_pool.go
+++ b/aws/resource_aws_cognito_user_pool.go
@@ -137,17 +137,19 @@ func resourceAwsCognitoUserPool() *schema.Resource {
 			},
 
 			"email_verification_subject": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ValidateFunc: validateCognitoUserPoolEmailVerificationSubject,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ValidateFunc:  validateCognitoUserPoolEmailVerificationSubject,
+				ConflictsWith: []string{"verification_message_template.0.email_subject"},
 			},
 
 			"email_verification_message": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ValidateFunc: validateCognitoUserPoolEmailVerificationMessage,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ValidateFunc:  validateCognitoUserPoolEmailVerificationMessage,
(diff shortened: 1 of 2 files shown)


atsushi-ishibashi:
According to AWS support, 
> If both are specified and the values are different, either one is prioritized and updated.
> So far `verification_message_template` is prioritized, but this operation is not specified in the document, possibly at the present moment only, so it will change in the future without notice.

website/docs/r/cognito_user_pool.markdown
```diff
@@ -31,7 +31,11 @@ The following arguments are supported:
 * `email_configuration` (Optional) - The [Email Configuration](#email-configuration).
 * `name` - (Required) The name of the user pool.
 * `email_verification_subject` - (Optional) A string representing the email verification subject.
+
+~> **Note** `verification_message_template.email_subject` replaces this string when you specify these two.
```
Ninir: I'm wondering whether we should set email_verification_subject as `ConflictsWith: []string{"verification_message_template.0. email_subject"},`, since we cannot have both being different

Thus, we could avoid such notice... what do you think @atsushi-ishibashi 

Also pinging @radeksimko to get it's opinion :)

atsushi-ishibashi: I'm worrying that it becomes breaking-change...
There are users who are forced update `.tf` files.

Ninir: Yes, it would be, but since it's one or the other according to the AWS support... this is a required thing IMO.

atsushi-ishibashi: I got it👍

radeksimko: @atsushi-ishibashi Correct me if I'm wrong but the API misbehaves and causes spurious diffs if you provide both?

The only case when I see `ConflictsWith` causing problems/BC is when folks use the exact same values.

We can defer the `ConflictsWith` addition to 2.0.0 and document this as suggested in the meantime.

@Ninir what do you think?

Ninir: > @atsushi-ishibashi Correct me if I'm wrong but the API misbehaves and causes spurious diffs if you provide both?

Exactly!

> We can defer the ConflictsWith addition to 2.0.0 and document this as suggested in the meantime.

Do agree on that :) 👍

atsushi-ishibashi: @radeksimko That's right!
> We can defer the ConflictsWith addition to 2.0.0 and document this as suggested in the meantime.

I also agree!

PR #2428, opened 2017-11-26
Cognito User Pool documentation clarifications

ctranstrum:
Updating the documentation of the  Cognito User Pool resource to clarify the requirements for values:
* Verification Message Template: default_email_option can accept only one of two choices, neither of which were listed.
* Invite Message Template: email_message and sms_message both require `{username}`, which wasn’t documented.

### DIFF
--- a/website/docs/r/cognito_user_pool.markdown
+++ b/website/docs/r/cognito_user_pool.markdown
@@ -111,12 +111,12 @@ The following arguments are supported:
 
 #### Verification Message Template
 
-  * `default_email_option` (Optional) - The default email option.
-  * `email_message` (Optional) - The email message template. Must contain the `{####}` placeholder.
+  * `default_email_option` (Optional) - The default email option. Must be either `CONFIRM_WITH_CODE` or `CONFIRM_WITH_LINK`.
+  * `email_message` (Optional) - The email message template. Must contain the `{####}` and `{username}` placeholders.
   * `email_message_by_link` (Optional) - The email message template for sending a confirmation link to the user.
   * `email_subject` (Optional) - The subject line for the email message template.
   * `email_subject_by_link` (Optional) - The subject line for the email message template for sending a confirmation link to the user.
-  * `sms_message` (Optional) - The SMS message template. Must contain the `{####}` placeholder.
+  * `sms_message` (Optional) - The SMS message template. Must contain the `{####}` and `{username}` placeholders.
 
 ## Attribute Reference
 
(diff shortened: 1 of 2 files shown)


website/docs/r/cognito_user_pool.markdown
```diff
@@ -86,10 +86,10 @@ The following arguments are supported:
 
 #### Schema Attributes
 
-  * `attribute_data_type` (Optional) - The attribute data type.
+  * `attribute_data_type` (Required) - The attribute data type. Must be one of `Boolean`, `Number`, `String`, `DateTime`.
```
Ninir: Since we put that as Required (which is a good catch by the way), we should also change the related code attribute from Optional to Required

website/docs/r/cognito_user_pool.markdown
```diff
@@ -86,10 +86,10 @@ The following arguments are supported:
...
   * `developer_only_attribute` (Optional) - Specifies whether the attribute type is developer only.
   * `mutable` (Optional) - Specifies whether the attribute can be changed once it has been created.
-  * `name` (Optional) - The name of the attribute.
+  * `name` (Required) - The name of the attribute.
```
Ninir: Since we put that as Required (which is a good catch by the way), we should also change the related code attribute from Optional to Required

website/docs/r/cognito_user_pool.markdown
```diff
@@ -107,11 +107,11 @@ The following arguments are supported:
 #### SMS Configuration
 
   * `external_id` (Required) - The external ID used in IAM role trust relationships. For more information about using external IDs, see [How to Use an External ID When Granting Access to Your AWS Resources to a Third Party](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html).
-  * `source_arn` (Required) - The ARN of the Amazon SNS caller.
+  * `sns_caller_arn` (Required) - The ARN of the Amazon SNS caller. This is usually the IAM role that you've given Cognito permission to assume.
 
 #### Verification Message Template
 
-  * `default_email_option` (Optional) - The default email option.
+  * `default_email_option` (Optional) - The default email option. Must be either `CONFIRM_WITH_CODE` or `CONFIRM_WITH_LINK`.
```
Ninir: We should also add: Defaults to `CONFIRM_WITH_CODE`

website/docs/r/cognito_user_pool.markdown
ctranstrum: Ah. Thanks. I was thinking I was just helping out with the documentation, but of course you're right. Code updated as well.

website/docs/r/cognito_user_pool.markdown
ctranstrum: Code updated.

Issue #2050, opened 2017-10-25
Codebuild artifacts packaging changes is not applied


SiahaanBernard:
Hi
I am creating aws codebuild project with artifact packaging type "NONE", however i want to change this artifact packaging type to "ZIP" but when i change the packaging type to "ZIP", i run terraform plan and there is no changes detected, run terraform apply doesn't make a changes to the codebuild project.

### Terraform Version
v0.10.4

### Affected Resource(s)
- AWS CodeBuild

### Terraform Configuration Files
here is my first terraform configuration
```hcl
resource "aws_codebuild_project" "packaging_test" {
  name         = "packaging_test"
  service_role = "code_build_run"

  "source" {
    type      = "CODECOMMIT"
    location  = "${var.codecommit_url}"
    buildspec = "buildspec_build.yml"
  }

  "environment" {
    compute_type = "BUILD_GENERAL1_SMALL"
    image        = "aws/codebuild/ubuntu-base:14.04"
    type         = "LINUX_CONTAINER"
  }

  "artifacts" {
    type      = "S3"
    location  = "${var.s3_bucket_path}"
    name      = "packaging_test"
    packaging = "NONE"
  }
}
```
then change the artifacts section to
```hcl
...
  "artifacts" {
    type      = "S3"
    location  = "${var.s3_bucket_path}"
    name      = "packaging_test"
    packaging = "ZIP"
  }
...
```

### Expected Behavior
- Codebuild artifact packaging type should be change from NONE become ZIP on AWS

### Actual Behavior
- terraform does not detect any change and does nothing

barryoneill:
I'm also seeing changes in the buildspec file not being picked up (e.g. if I change one of the pre_build commands), so the apply sees nothing to do.   Could be related?  (Terraform v0.11.1)

bflad:
The fixes for having the `aws_codebuild_project` resource `artifacts` configuration block perform better drift detection and updates have been merged and will release with version 2.22.0 of the Terraform AWS Provider, tomorrow. 👍

PR #2431, opened 2017-11-26
resource/aws_codebuild_project: Detect changes in artifacts

atsushi-ishibashi:
Fixed: #2050

### DIFF
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -22,6 +22,9 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 		Update: resourceAwsCodeBuildProjectUpdate,
 		Delete: resourceAwsCodeBuildProjectDelete,
 
+		SchemaVersion: 1,
+		MigrateState:  resourceAwsCodebuildMigrateState,
+
 		Schema: map[string]*schema.Schema{
 			"artifacts": {
 				Type:     schema.TypeSet,
@@ -57,7 +60,6 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 						},
 					},
 				},
-				Set: resourceAwsCodeBuildProjectArtifactsHash,
 			},
 			"description": {
 				Type:         schema.TypeString,
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -560,9 +560,22 @@ func resourceAwsCodeBuildProjectArtifactsHash(v interface{}) int {
 	var buf bytes.Buffer
 	m := v.(map[string]interface{})
 
-	artifactType := m["type"].(string)
-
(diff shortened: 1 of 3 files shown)


radeksimko:
Hi @atsushi-ishibashi 
thanks for the PR.

I think you found the root cause - i.e. I agree this is the right place to fix it.

We might get away with removing the `Set` function altogether and let the schema use the default hash function though.

Either way I'm pretty sure this will require state migration as we'd otherwise trigger no-op diff for all users which have this resource already in state (as the index will change). You can verify this by `apply`ing the resource using stable (`master`) version of the provider and then `plan`ing with version compiled from your branch. If you see a diff then we need a migration - if not then I'll be really confused and apologetic. 😄 

Example of state migration can be found here:

https://github.com/terraform-providers/terraform-provider-aws/blob/19503d1fe0ff14c0b0ab99c209fc65d5d6f2497d/aws/resource_aws_dynamodb_table.go#L42-L43
The state/schema is versioned as `0` if there's no version.

atsushi-ishibashi:
I have a problem. Below is tfstate which was generated `terraform apply` of master version.
```
"primary": {
                        "id": "arn:aws:codebuild:us-west-2:295335637120:project/tf-packaging_test",
                        "attributes": {
                            "artifacts.#": "1",
                            "artifacts.537882814.location": "tf-codebuild-93892171401399123",
                            "artifacts.537882814.name": "packaging_test",
                            "artifacts.537882814.namespace_type": "NONE",
                            "artifacts.537882814.packaging": "ZIP",
                            "artifacts.537882814.path": "",
                            "artifacts.537882814.type": "S3",
                            "build_timeout": "60",
                            "description": "",
```

Then I ran 
```
$ make build
$ mv $GOPATH/bin/terraform-provider-aws <plugin-dir>
$ terraform init -plugin-dir=<plugin-dir>
$ terraform plan

  ~ aws_codebuild_project.packaging_test
      artifacts.2191994982.location:       "tf-codebuild-93892171401399123" => ""
      artifacts.2191994982.name:           "packaging_test" => ""
      artifacts.2191994982.namespace_type: "NONE" => ""
      artifacts.2191994982.packaging:      "ZIP" => ""
      artifacts.2191994982.path:           "" => ""
      artifacts.2191994982.type:           "S3" => ""
      artifacts.962148601.location:        "" => "tf-codebuild-93892171401399123"
      artifacts.962148601.name:            "" => "packaging_test"
      artifacts.962148601.namespace_type:  "" => ""
      artifacts.962148601.packaging:       "" => "ZIP"
      artifacts.962148601.path:            "" => ""
      artifacts.962148601.type:            "" => "S3"
```

bflad:
Hi @atsushi-ishibashi 👋 It looks like this pull request has not been visited in a very long time and has many merge conflicts with master. Rather than keeping this open in an un-mergeable state, we're going to opt to close this for now, but if you're interested in re-opening a new pull request please feel free. Thanks so much.

PR #2447, opened 2017-11-27
Bump to aws-sdk-go@v1.12.34

atsushi-ishibashi:
`govendor fetch github.com/aws/aws-sdk-go/....@v1.12.34` didn't fetch new files😕 
So additionally I ran the below commands
```
govendor fetch github.com/aws/aws-sdk-go/service/mediastore@v1.12.34
govendor fetch github.com/aws/aws-sdk-go/service/medialive@v1.12.34
govendor fetch github.com/aws/aws-sdk-go/service/mediaconvert@v1.12.34
govendor fetch github.com/aws/aws-sdk-go/service/mediapackage@v1.12.34
govendor fetch github.com/aws/aws-sdk-go/service/mediastoredata@v1.12.34
```

PR #2448, opened 2017-11-27
New Resource: aws_media_store_container

atsushi-ishibashi:
Required: #2447 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsMediaStoreContainer_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAwsMediaStoreContainer_basic -timeout 120m
=== RUN   TestAccAwsMediaStoreContainer_basic
--- PASS: TestAccAwsMediaStoreContainer_basic (135.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	135.666s
```

### DIFF
--- a/aws/resource_aws_media_store_container.go
+++ b/aws/resource_aws_media_store_container.go
@@ -1,0 +1,149 @@
+package aws
+
+import (
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/mediastore"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsMediaStoreContainer() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsMediaStoreContainerCreate,
+		Read:   resourceAwsMediaStoreContainerRead,
+		Update: resourceAwsMediaStoreContainerUpdate,
+		Delete: resourceAwsMediaStoreContainerDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					if !regexp.MustCompile("^\\w+$").MatchString(value) {
+						errors = append(errors, fmt.Errorf("%q must match \\w+", k))
+					}
+					return
+				},
+			},
+			"policy": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				ValidateFunc:     validateIAMPolicyJson,
+				DiffSuppressFunc: suppressEquivalentAwsPolicyDiffs,
+			},
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"endpoint": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsMediaStoreContainerCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).mediastoreconn
+
+	input := &mediastore.CreateContainerInput{
+		ContainerName: aws.String(d.Get("name").(string)),
+	}
+
+	_, err := conn.CreateContainer(input)
+	if err != nil {
+		return err
+	}
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{mediastore.ContainerStatusCreating},
+		Target:     []string{mediastore.ContainerStatusActive},
+		Refresh:    mediaStoreContainerRefreshStatusFunc(conn, d.Get("name").(string)),
+		Timeout:    10 * time.Minute,
+		Delay:      10 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("[WARN] Error waiting for MediaStore Container status to be \"ACTIVE\": %s", err)
+	}
+
+	d.SetId(d.Get("name").(string))
+	return resourceAwsMediaStoreContainerUpdate(d, meta)
+}
+
+func resourceAwsMediaStoreContainerRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).mediastoreconn
+
+	input := &mediastore.DescribeContainerInput{
+		ContainerName: aws.String(d.Id()),
+	}
+	resp, err := conn.DescribeContainer(input)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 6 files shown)


aws/resource_aws_media_store_container.go
```diff
@@ -0,0 +1,149 @@
...
+		Target:     []string{mediastore.ContainerStatusActive},
+		Refresh:    mediaStoreContainerRefreshStatusFunc(conn, d.Get("name").(string)),
+		Timeout:    10 * time.Minute,
+		Delay:      10 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("[WARN] Error waiting for MediaStore Container status to be \"ACTIVE\": %s", err)
```
radeksimko: Nitpick, but I think [the error](https://github.com/hashicorp/terraform/blob/master/helper/resource/error.go#L53) which comes out of `WaitForState` is sufficiently detailed so we don't need to duplicate the context here.

aws/resource_aws_media_store_container.go
```diff
@@ -0,0 +1,149 @@
...
+	d.Set("endpoint", resp.Container.Endpoint)
+	return nil
+}
+
+func resourceAwsMediaStoreContainerUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).mediastoreconn
+
+	if !d.HasChange("policy") {
+		return resourceAwsMediaStoreContainerRead(d, meta)
+	}
```
radeksimko: I believe the above 3 lines will never be executed since all fields are `ForceNew` (so the only way the function ever gets executed is when `policy` has changed), unless I'm mistaken?

aws/resource_aws_media_store_container.go
```diff
@@ -0,0 +1,149 @@
...
+	_, err := conn.DeleteContainer(input)
+	if err != nil {
+		if ecrerr, ok := err.(awserr.Error); ok {
+			switch ecrerr.Code() {
+			case mediastore.ErrCodeContainerNotFoundException:
+				d.SetId("")
+				return nil
+			}
+		}
```
radeksimko: Nitpick, but I think we could simplify the above error handling to something like 

```go
if isAWSErr(err, mediastore.ErrCodeContainerNotFoundException, "") {
    d.SetId("")
    return nil
}
```

aws/resource_aws_media_store_container_test.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		resp, err := conn.DescribeContainer(input)
+		if err != nil {
+			if ecrerr, ok := err.(awserr.Error); ok {
+				switch ecrerr.Code() {
+				case mediastore.ErrCodeContainerNotFoundException:
+					return nil
+				}
+			}
```
radeksimko: As mentioned elsewhere - I think the error handling can be simplified a bit.

aws/resource_aws_media_store_container.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
```
radeksimko: As this resource is stateful shouldn't we also wait here until it's deleted instead of leaving it in "deleting" state?

aws/resource_aws_media_store_container_test.go
```diff
@@ -0,0 +1,109 @@
...
+  name = "tf_mediastore_%s"
+	policy = <<POLICY
+{
+  "Version": "2012-10-17",
+  "Statement": [{
+    "Sid": "MediaStoreFullAccess",
+    "Action": [ "mediastore:*" ],
+    "Principal": {"AWS" : "*"},
+    "Effect": "Allow",
+    "Resource": "arn:aws:mediastore:us-west-2:${data.aws_caller_identity.test.account_id}:container/tf_mediastore_%s/*",
```
radeksimko: It looks like we'd benefit from having `aws_media_store_container_policy` so we can reference the container ARN instead of building it like this. 🤔 

It's certainly outside of scope of this PR, just thinking out loud as I'm reading the code...

website/docs/r/media_store_container.html.markdown
```diff
@@ -0,0 +1,33 @@
...
+resource "aws_media_store_container" "example" {
+  name = "example"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the container. Must atisfy regular expression pattern: `\w+`
```
radeksimko: Can we make this a little bit more human-readable for folks who don't know regular expressions or omit this sentence (as the validation error message will tell them anyways)?

aws/resource_aws_media_store_container.go
```diff
@@ -0,0 +1,149 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(string)
+					if !regexp.MustCompile("^\\w+$").MatchString(value) {
+						errors = append(errors, fmt.Errorf("%q must match \\w+", k))
```
radeksimko: Can we make this error message a bit more human friendly for folks who don't know regular expressions? e.g. `must contain alphanumeric characters or underscores`?

aws/resource_aws_media_store_container_test.go
atsushi-ishibashi: Yeah, I agree with you...
`"Resource"` in policy must be scope of `arn:aws:mediastore:us-west-2:1234567890:container/tf_mediastore_%s/` (I got error if I set `*` to Resource) so I thought the same when implementing.
If we'll develop `aws_media_store_container_policy`, should I remove `policy` from `aws_media_store_container`?

aws/resource_aws_media_store_container.go
atsushi-ishibashi: I called `resourceAwsMediaStoreContainerUpdate` from the end of `resourceAwsMediaStoreContainerCreate` so I needed the above 3 lines.
Separating policy as `aws_media_store_container_policy` will solve this complication.

aws/resource_aws_media_store_container_test.go
radeksimko: No, keep it here... as I said it's outside of scope of this PR 😃

aws/resource_aws_media_store_container.go
radeksimko: Ah, I see - that's confusing. It actually makes me reconsider what I said in my other comment - I think we should remove all the policy related logic from this PR and implement it as a separate resource in another PR.

Do you agree?

aws/resource_aws_media_store_container_test.go
radeksimko: I'm taking back what I said - see my other comment about policy.

aws/resource_aws_media_store_container.go
atsushi-ishibashi: Yes, I agree👍 I'll remove

atsushi-ishibashi:
I removed `policy`. It's ready to review👍

PR #2454, opened 2017-11-28
resource/aws_network_acl_rule: Fix DiffSuppressFunc

enm10k:
Reference https://github.com/terraform-providers/terraform-provider-aws/issues/222

~The following issue was already closed but it's only for one protocol.~
~Now the bug is not fixed in other protocols.~
~This code fix it.~
~issue: https://github.com/hashicorp/terraform/issues/13012~
~pr: https://github.com/hashicorp/terraform/pull/13049~

This is `"best effort" implementations of acceptance tests`.
I hope Terraform team check this pr.

Thanks.


enm10k:
@radeksimko 
I read #222 again and now I understood this problem is not only for aws_network_acl_rule.
If anyone reviews my changes, I'll check these other resources: aws_noetwork_acl, aws_security_group, aws_security_group_rule.
I want to fix this function for my use.

bflad:

Issue #2357, opened 2017-11-17
Add edition option for aws_directory_service_directory resource


hashibot:
_This issue was originally opened by @TechDawg270 as hashicorp/terraform#16694. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

When selecting a directory type of `Microsoft AD` two options are available which are not currently configurable. Those two options are `Standard` and `Enterprise` for edition

bflad:
This has been merged into master and will be released in v1.10.0 of the AWS provider, likely the end of this week.

bflad:
This has been released in [version 1.10.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1100-february-24-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #2458, opened 2017-11-28
Add DirectoryService size parameter for MicrosoftAD

philharle:


philharle:
Requires change in the aws-sdk-go project, see https://github.com/terraform-providers/terraform-provider-aws/issues/2192

bflad:
Hi @philharle 👋  sorry for the long delay in a maintainer review of this PR. It looks like the correct answer here was an `Edition` parameter instead of using `Size` as noted upstream in the SDK issue https://github.com/aws/aws-sdk-go/issues/1664. Since this PR is invalid according to the AWS API and SDK, I'm going to close this.

The good news is that we have a tracking issue for supporting `Edition` in #2357 and a recently opened PR to implement this in #3421! 🎉

PR #2462, opened 2017-11-29
resource/aws_db_instance: Add ForceNew to db_subnet_group_name

nathanielks:
Unfortunately, it's not possible to change an existing RDS instance's DB Subnet Group without a [workaround](https://serverfault.com/a/817598). Addresses https://github.com/terraform-providers/terraform-provider-aws/issues/512 by requiring a new DB instance to be created when changing.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -234,6 +234,7 @@ func resourceAwsDbInstance() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
+				ForceNew: true,
 			},
 
 			"parameter_group_name": {


nathanielks:
Edit: added link to https://serverfault.com/a/817598 to explain how the bug is AWS side and not terraform side.

bflad:
Thanks for this contribution and sorry for the long maintainer silence on it. 😅 

I've thought about this and looking at the modification behavior (same VPC vs different VPC), I do not think we should prevent folks from modifying an existing instance in the scenario of migrating it to a different VPC. The AWS API may someday lift the same VPC restriction. If we can come up with some way to easily determine if the new subnet group is in the same VPC, we can take a look at implementing this differently, but for now I'm going to close this PR to not keep this old PR lingering.

PR #2464, opened 2017-11-29
aws/structure: add canonicalXML helper function

catsby:


aws/structure_test.go
```diff
@@ -1322,3 +1322,188 @@ func TestNormalizeCloudFormationTemplate(t *testing.T) {
...
+</rant>
+</note>`,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.Name, func(t *testing.T) {
+			old, err := canonicalXML(tc.Config)
+			if err != nil {
+				t.Fatalf("Error getting canonical xml: %s", err)
```
radeksimko: It may help if the error message mentioned this was error from _old_/_new_ XML.

catsby: You ask too much


8e56a91c

PR #2470, opened 2017-11-29
docs: Fix incomplete example of mq_configuration

radeksimko:


website/docs/r/mq_configuration.html.markdown
```diff
@@ -16,7 +16,20 @@ For more information on Amazon MQ, see [Amazon MQ documentation](https://docs.aw
 
 ```hcl
 resource "aws_mq_configuration" "example" {
-
+  description = "Example Configuration"
```
Ninir: Nitpick: Would just align values 😄

PR #2477, opened 2017-11-30
New Resource: aws_sagemaker_endpoint_configuration

jckuester:
***Please only review the [last commit](https://github.com/terraform-providers/terraform-provider-aws/pull/2477/commits/f74eae4af2b0ec8983cb49248c2712df59921bb4). The reason is that this PR depends and is rebased on https://github.com/terraform-providers/terraform-provider-aws/pull/2478, as the tests need the resource `aws_sagemaker_endpoint_configuration` to work.***

Add the resource `aws_sagemaker_endpoint_configuration` for the newly announced service called [SageMaker](https://aws.amazon.com/blogs/aws/sagemaker/):

* Documentation/Website 
* Schema
* Implementation of CRUD operations
* Acceptance tests

### Acceptance tests

```
╰─$ make testacc TESTARGS='-run=TestAccAWSSagemakerEndpointConfig_*'                                                                     2 ↵
TF_ACC=1 go test ./... -v -run=TestAccAWSSagemakerEndpointConfig_* -timeout 120m
=== RUN   TestAccAWSSagemakerEndpointConfig_basic
--- PASS: TestAccAWSSagemakerEndpointConfig_basic (72.46s)
=== RUN   TestAccAWSSagemakerEndpointConfig_kmsKeyId
--- PASS: TestAccAWSSagemakerEndpointConfig_kmsKeyId (97.70s)
=== RUN   TestAccAWSSagemakerEndpointConfig_tags
--- PASS: TestAccAWSSagemakerEndpointConfig_tags (109.41s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	279.620s
```

### Updates

* *14th Feb 2018:* Support new parameter `kms_key_id`

### DIFF
--- a/aws/resource_aws_sagemaker_endpoint_configuration.go
+++ b/aws/resource_aws_sagemaker_endpoint_configuration.go
@@ -1,16 +1,17 @@
 package aws
 
 import (
-	"bytes"
 	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/validation"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/sagemaker"
-	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
-	"log"
-	"time"
 )
 
 func resourceAwsSagemakerEndpointConfiguration() *schema.Resource {
@@ -38,7 +39,7 @@ func resourceAwsSagemakerEndpointConfiguration() *schema.Resource {
 			},
 
 			"production_variants": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Required: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
@@ -56,9 +57,10 @@ func resourceAwsSagemakerEndpointConfiguration() *schema.Resource {
 						},
 
 						"initial_instance_count": {
-							Type:     schema.TypeInt,
-							Required: true,
-							ForceNew: true,
+							Type:         schema.TypeInt,
+							Required:     true,
+							ForceNew:     true,
+							ValidateFunc: validation.IntAtLeast(1),
 						},
 
 						"instance_type": {
@@ -68,24 +70,25 @@ func resourceAwsSagemakerEndpointConfiguration() *schema.Resource {
 						},
 
 						"initial_variant_weight": {
-							Type:     schema.TypeFloat,
-							Required: true,
+							Type:         schema.TypeFloat,
+							Optional:     true,
+							ForceNew:     true,
+							ValidateFunc: FloatAtLeast(0),
+						},
+
+						"accelerator_type": {
+							Type:     schema.TypeString,
+							Optional: true,
 							ForceNew: true,
 						},
 					},
 				},
-				Set: resourceAwsSagmakerEndpointConfigEntryHash,
 			},
 
 			"kms_key_id": {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-			},
-
-			"creation_time": {
-				Type:     schema.TypeString,
-				Computed: true,
 			},
 
 			"tags": tagsSchema(),
@@ -107,7 +110,7 @@ func resourceAwsSagemakerEndpointConfigurationCreate(d *schema.ResourceData, met
 		EndpointConfigName: aws.String(name),
 	}
 
-	prodVariants, err := expandProductionVariants(d.Get("production_variants").(*schema.Set).List())
+	prodVariants, err := expandProductionVariants(d.Get("production_variants").([]interface{}))
 	if err != nil {
 		return err
 	}
@@ -117,18 +120,18 @@ func resourceAwsSagemakerEndpointConfigurationCreate(d *schema.ResourceData, met
 		createOpts.KmsKeyId = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("tags"); ok {
+		createOpts.Tags = tagsFromMapSagemaker(v.(map[string]interface{}))
+	}
+
 	log.Printf("[DEBUG] Sagemaker endpoint configuration create config: %#v", *createOpts)
-	resp, err := conn.CreateEndpointConfig(createOpts)
+	_, err = conn.CreateEndpointConfig(createOpts)
 	if err != nil {
-		return fmt.Errorf("Error creating Sagemaker endpoint configuration: %s", err)
+		return fmt.Errorf("error creating Sagemaker endpoint configuration: %s", err)
 	}
-
 	d.SetId(name)
-	if err := d.Set("arn", resp.EndpointConfigArn); err != nil {
-		return err
-	}
 
-	return resourceAwsSagemakerEndpointConfigurationUpdate(d, meta)
+	return resourceAwsSagemakerEndpointConfigurationRead(d, meta)
 }
 
 func resourceAwsSagemakerEndpointConfigurationRead(d *schema.ResourceData, meta interface{}) error {
@@ -140,11 +143,12 @@ func resourceAwsSagemakerEndpointConfigurationRead(d *schema.ResourceData, meta 
 
 	endpointConfig, err := conn.DescribeEndpointConfig(request)
 	if err != nil {
-		if sagemakerErr, ok := err.(awserr.Error); ok && sagemakerErr.Code() == "ResourceNotFound" {
+		if sagemakerErr, ok := err.(awserr.Error); ok && sagemakerErr.Code() == "ValidationException" {
+			log.Printf("[INFO] unable to find the sagemaker endpoint configuration resource and therefore it is removed from the state: %s", d.Id())
 			d.SetId("")
 			return nil
 		}
-		return fmt.Errorf("Error reading Sagemaker endpoint configuration %s: %s", d.Id(), err)
+		return fmt.Errorf("error reading Sagemaker endpoint configuration %s: %s", d.Id(), err)
 	}
 
 	if err := d.Set("arn", endpointConfig.EndpointConfigArn); err != nil {
@@ -156,14 +160,19 @@ func resourceAwsSagemakerEndpointConfigurationRead(d *schema.ResourceData, meta 
 	if err := d.Set("production_variants", flattenProductionVariants(endpointConfig.ProductionVariants)); err != nil {
 		return err
 	}
-
 	if err := d.Set("kms_key_id", endpointConfig.KmsKeyId); err != nil {
 		return err
 	}
-	if err := d.Set("creation_time", endpointConfig.CreationTime.Format(time.RFC3339)); err != nil {
+
+	tagsOutput, err := conn.ListTags(&sagemaker.ListTagsInput{
+		ResourceArn: endpointConfig.EndpointConfigArn,
+	})
+	if err != nil {
+		return fmt.Errorf("error listing tags of Sagemaker endpoint configuration %s: %s", d.Id(), err)
+	}
+	if err := d.Set("tags", tagsToMapSagemaker(tagsOutput.Tags)); err != nil {
 		return err
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,264 @@
...
+							ForceNew: true,
+						},
+
+						"instance_type": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"initial_variant_weight": {
```
mbfrahry: We should validate that this is a minimum of 0

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,264 @@
...
+			"name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				ValidateFunc: validateSagemakerName,
+			},
+
+			"production_variants": {
+				Type:     schema.TypeSet,
```
mbfrahry: Is there a reason this is a TypeSet? If it's because this comes back from the api unordered then this is perfect to use. Otherwise we should use a TypeList as they are easier to access and use because of it's ordered nature.

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,264 @@
...
+				Set: resourceAwsSagmakerEndpointConfigEntryHash,
+			},
+
+			"kms_key_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+
+			"creation_time": {
```
mbfrahry: Is there a need for this attribute. I don't think it adds a ton of value inside of Terraform and can be removed.

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	log.Printf("[DEBUG] Sagemaker endpoint configuration create config: %#v", *createOpts)
+	resp, err := conn.CreateEndpointConfig(createOpts)
+	if err != nil {
+		return fmt.Errorf("Error creating Sagemaker endpoint configuration: %s", err)
+	}
+
+	d.SetId(name)
+	if err := d.Set("arn", resp.EndpointConfigArn); err != nil {
```
mbfrahry: Since this is set in the read, I think we can remove it from here

aws/resource_aws_sagemaker_endpoint_configuration.go
jckuester: Done.

aws/resource_aws_sagemaker_endpoint_configuration.go
jckuester: Done.

aws/resource_aws_sagemaker_endpoint_configuration.go
jckuester: Done.

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,264 @@
...
+							ForceNew: true,
+						},
+
+						"model_name": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+
+						"initial_instance_count": {
```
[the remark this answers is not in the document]
jckuester: Done.

aws/resource_aws_sagemaker_endpoint_configuration.go
jckuester: Done.

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	return containers, nil
+}
+
+func flattenProductionVariants(list []*sagemaker.ProductionVariant) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, len(list))
+	for _, i := range list {
+		l := map[string]interface{}{
+			"variant_name":           *i.VariantName,
```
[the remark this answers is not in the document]
jckuester: Done. Only testing the optional parameters against `nil`.

jckuester:
@mbfrahry I am done here, too, and ran tests.

aws/resource_aws_sagemaker_endpoint_configuration.go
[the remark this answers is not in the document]
xmj: @mbfrahry could you please review the changes pushed on Jan 9th? be great if this resource made it into a release near-term! :-)

xmj:
How can we help to get this merged?

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+	if err := d.Set("tags", tagsToMapSagemaker(tagsOutput.Tags)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func resourceAwsSagemakerEndpointConfigurationUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	d.Partial(true)
```
bflad: Nit: Since this resource does not require partial Terraform state updates, we should remove the usage of `d.Partial()` and `d.SetPartial()`

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+
+		if sagemakerErr.Code() == "ResourceNotFound" {
+			return resource.RetryableError(err)
+		}
+
+		return resource.NonRetryableError(fmt.Errorf("Error deleting Sagemaker endpoint configuration: %s", err))
+	})
+}
+
+func expandProductionVariants(configured []interface{}) ([]*sagemaker.ProductionVariant, error) {
```
bflad: Two things:
* This function never returns an `error` so that should be removed from the function signature
* Nit: Since this codebase uses a very flat Go package structure, we should include the AWS service name in the expand and flatten function names

```suggestion
func expandSagemakerProductionVariants(configured []interface{}) []*sagemaker.ProductionVariant {
```

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+		if v, ok := data["variant_name"]; ok {
+			l.VariantName = aws.String(v.(string))
+		} else {
+			l.VariantName = aws.String(resource.UniqueId())
+		}
+
+		if v, ok := data["initial_variant_weight"]; ok {
+			l.InitialVariantWeight = aws.Float64(v.(float64))
+		} else {
+			l.InitialVariantWeight = aws.Float64(1)
```
bflad: Should we add this as a `Default` for the schema attribute?

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+
+	return containers, nil
+}
+
+func flattenProductionVariants(list []*sagemaker.ProductionVariant) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, len(list))
+
+	for _, i := range list {
+		l := map[string]interface{}{
+			"instance_type":          *i.InstanceType,
```
bflad: To prevent potential panics, we should prefer to use the [AWS Go SDK provided dereferencing functions](https://docs.aws.amazon.com/sdk-for-go/api/aws/) here and below. We can likely also remove `nil` checks below as well with them. e.g.

```suggestion
			"instance_type":          aws.StringValue(i.InstanceType),
```

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+
+	createOpts := &sagemaker.CreateEndpointConfigInput{
+		EndpointConfigName: aws.String(name),
+	}
+
+	prodVariants, err := expandProductionVariants(d.Get("production_variants").([]interface{}))
+	if err != nil {
+		return err
+	}
+	createOpts.ProductionVariants = prodVariants
```
bflad: This can be simplified and moved to the struct above when the function signature is updated 👍 

```go
	createOpts := &sagemaker.CreateEndpointConfigInput{
		EndpointConfigName: aws.String(name),
		ProductionVariants: expandProductionVariants(d.Get("production_variants").([]interface{})),
	}
```

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+						"accelerator_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ForceNew: true,
+						},
+					},
+				},
+			},
+
+			"kms_key_id": {
```
bflad: Since the [Sagemaker API Reference](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateEndpointConfig.html#SageMaker-CreateEndpointConfig-request-KmsKeyId) notes this should be an Amazon Resource Name anyways, we should consider renaming this to `kms_key_arn`? We have been trying to start doing this places where we can to reduce confusion. We can also validate it partially with `ValidateFunc: validateArn`

aws/resource_aws_sagemaker_endpoint_configuration_test.go
```diff
@@ -0,0 +1,551 @@
...
+func testAccCheckSagemakerEndpointConfigDestroy(s *terraform.State) error {
+	conn := testAccProvider.Meta().(*AWSClient).sagemakerconn
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_sagemaker_endpoint_configuration" {
+			continue
+		}
+
+		resp, err := conn.ListEndpointConfigs(&sagemaker.ListEndpointConfigsInput{
+			NameContains: aws.String(rs.Primary.ID),
```
bflad: This seems like it can report false positives and should use `DescribeEndpointConfig` instead, e.g.

```go
input := &sagemaker.DescribeEndpointConfigInput{
	EndpointConfigName: aws.String(rs.Primary.ID),
}

output, err := conn.DescribeEndpointConfig(input)

if isAWSErr(err, sagemaker.ErrCodeResourceNotFound, "") {
  continue
}

if err != nil {
  return err
}

return fmt.Errorf("Sagemaker Endpoint Configuration (%s) still exists", rs.Primary.ID)
```

aws/resource_aws_sagemaker_endpoint_configuration.go
```diff
@@ -0,0 +1,278 @@
...
+func resourceAwsSagemakerEndpointConfigurationRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	request := &sagemaker.DescribeEndpointConfigInput{
+		EndpointConfigName: aws.String(d.Id()),
+	}
+
+	endpointConfig, err := conn.DescribeEndpointConfig(request)
+	if err != nil {
+		if sagemakerErr, ok := err.(awserr.Error); ok && sagemakerErr.Code() == "ValidationException" {
```
bflad: Nit: We have a helper function for this type of checking:

```suggestion
		if isAWSErr(err, "ValidationException", "") {
```

website/docs/r/sagemaker_endpoint_configuration.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Optional) The name of the endpoint configuration. If omitted, Terraform will assign a random, unique name.
+* `production_variants` - (Required) Fields are documented below.
+* `kms_key_id` - (Optional) KMS key to encrypt the model data.
```
bflad: Matching the [Sagemaker API Reference](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateEndpointConfig.html#SageMaker-CreateEndpointConfig-request-KmsKeyId) (and suggested rename above)

```suggestion
* `kms_key_arn` - (Optional) Amazon Resource Name (ARN) of a AWS Key Management Service key that Amazon SageMaker uses to encrypt data on the storage volume attached to the ML compute instance that hosts the endpoint.
```

website/docs/r/sagemaker_endpoint_configuration.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+The `production_variant` block supports:
+
+* `variant_name` - (Optional) The name of the variant. If omitted, Terraform will assign a random, unique name.
```
bflad: Nit: These attributes should be alphabetically sorted either as a whole or required first then optional

jckuester:
@bflad thanks for the feedback! I will implement the feedback in the next days (probably Monday).

jckuester:
@bflad I am done here. See the last commit.

```
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSSagemakerEndpointConfiguration* -timeout 120m
--- PASS: TestAccAWSSagemakerEndpointConfiguration_ProductionVariants_AcceleratorType (51.70s)
--- PASS: TestAccAWSSagemakerEndpointConfiguration_Basic (56.96s)
--- PASS: TestAccAWSSagemakerEndpointConfiguration_ProductionVariants_InitialVariantWeight (70.52s)
--- PASS: TestAccAWSSagemakerEndpointConfiguration_Tags (79.69s)
--- PASS: TestAccAWSSagemakerEndpointConfiguration_KmsKeyId (105.20s)
```

bflad:
This has been released in [version 2.4.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#240-march-29-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #2496, opened 2017-11-30
Docs fix: name is an attribute on lb target group

genevievelesperance:


genevievelesperance:
Ah, okay. Totally glossed over that part!

PR #2512, opened 2017-12-01
New Resource: aws_api_gateway_vpc_link

atsushi-ishibashi:
- [x] test
- [x] docs

### DIFF
--- a/aws/resource_aws_api_gateway_vpc_link.go
+++ b/aws/resource_aws_api_gateway_vpc_link.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"log"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -26,10 +27,11 @@ func resourceAwsApiGatewayVpcLink() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-			"target_arn": {
-				Type:     schema.TypeString,
+			"target_arns": {
+				Type:     schema.TypeSet,
 				Required: true,
 				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
 		},
 	}
@@ -40,7 +42,7 @@ func resourceAwsApiGatewayVpcLinkCreate(d *schema.ResourceData, meta interface{}
 
 	input := &apigateway.CreateVpcLinkInput{
 		Name:       aws.String(d.Get("name").(string)),
-		TargetArns: []*string{aws.String(d.Get("target_arn").(string))},
+		TargetArns: expandStringList(d.Get("target_arns").(*schema.Set).List()),
 	}
 	if v, ok := d.GetOk("description"); ok {
 		input.Description = aws.String(v.(string))
@@ -50,6 +52,8 @@ func resourceAwsApiGatewayVpcLinkCreate(d *schema.ResourceData, meta interface{}
 	if err != nil {
 		return err
 	}
+
+	d.SetId(*resp.Id)
 
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{apigateway.VpcLinkStatusPending},
@@ -61,10 +65,10 @@ func resourceAwsApiGatewayVpcLinkCreate(d *schema.ResourceData, meta interface{}
 
 	_, err = stateConf.WaitForState()
 	if err != nil {
+		d.SetId("")
 		return fmt.Errorf("[WARN] Error waiting for APIGateway Vpc Link status to be \"%s\": %s", apigateway.VpcLinkStatusAvailable, err)
 	}
 
-	d.SetId(*resp.Id)
 	return nil
 }
 
@@ -78,6 +82,7 @@ func resourceAwsApiGatewayVpcLinkRead(d *schema.ResourceData, meta interface{}) 
 	resp, err := conn.GetVpcLink(input)
 	if err != nil {
 		if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
+			log.Printf("[WARN] VPC Link %s not found, removing from state", d.Id())
 			d.SetId("")
 			return nil
 		}
@@ -86,7 +91,7 @@ func resourceAwsApiGatewayVpcLinkRead(d *schema.ResourceData, meta interface{}) 
 
 	d.Set("name", resp.Name)
 	d.Set("description", resp.Description)
-	d.Set("target_arn", resp.TargetArns[0])
+	d.Set("target_arn", flattenStringList(resp.TargetArns))
 	return nil
 }
 
@@ -119,6 +124,7 @@ func resourceAwsApiGatewayVpcLinkUpdate(d *schema.ResourceData, meta interface{}
 	_, err := conn.UpdateVpcLink(input)
 	if err != nil {
 		if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
+			log.Printf("[WARN] VPC Link %s not found, removing from state", d.Id())
 			d.SetId("")
 			return nil
 		}
@@ -151,7 +157,6 @@ func resourceAwsApiGatewayVpcLinkDelete(d *schema.ResourceData, meta interface{}
 	_, err := conn.DeleteVpcLink(input)
 	if err != nil {
 		if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
-			d.SetId("")
 			return nil
 		}
 		return err
@@ -161,7 +166,7 @@ func resourceAwsApiGatewayVpcLinkDelete(d *schema.ResourceData, meta interface{}
 		Pending: []string{apigateway.VpcLinkStatusPending,
 			apigateway.VpcLinkStatusAvailable,
 			apigateway.VpcLinkStatusDeleting},
-		Target:     []string{"deleted"},
+		Target:     []string{""},
 		Timeout:    5 * time.Minute,
 		MinTimeout: 1 * time.Second,
 		Refresh: func() (interface{}, string, error) {
@@ -170,7 +175,7 @@ func resourceAwsApiGatewayVpcLinkDelete(d *schema.ResourceData, meta interface{}
 			})
 			if err != nil {
 				if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
-					return 1, "deleted", nil
+					return 1, "", nil
 				}
 				return nil, "failed", err
 			}
@@ -182,7 +187,6 @@ func resourceAwsApiGatewayVpcLinkDelete(d *schema.ResourceData, meta interface{}
 		return err
 	}
 
-	d.SetId("")
 	return nil
 }
 
--- a/aws/resource_aws_api_gateway_vpc_link.go
+++ b/aws/resource_aws_api_gateway_vpc_link.go
@@ -29,6 +29,7 @@ func resourceAwsApiGatewayVpcLink() *schema.Resource {
 			},
 			"target_arns": {
 				Type:     schema.TypeSet,
+				MaxItems: 1,
 				Required: true,
 				ForceNew: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
--- a/aws/resource_aws_api_gateway_vpc_link.go
+++ b/aws/resource_aws_api_gateway_vpc_link.go
@@ -56,7 +56,6 @@ func resourceAwsApiGatewayVpcLinkCreate(d *schema.ResourceData, meta interface{}
 		Target:     []string{apigateway.VpcLinkStatusAvailable},
 		Refresh:    apigatewayVpcLinkRefreshStatusFunc(conn, *resp.Id),
 		Timeout:    5 * time.Minute,
-		Delay:      10 * time.Second,
 		MinTimeout: 3 * time.Second,
 	}
 
@@ -131,7 +130,6 @@ func resourceAwsApiGatewayVpcLinkUpdate(d *schema.ResourceData, meta interface{}
 		Target:     []string{apigateway.VpcLinkStatusAvailable},
 		Refresh:    apigatewayVpcLinkRefreshStatusFunc(conn, d.Id()),
 		Timeout:    5 * time.Minute,
-		Delay:      10 * time.Second,
 		MinTimeout: 3 * time.Second,
 	}
 
@@ -159,7 +157,30 @@ func resourceAwsApiGatewayVpcLinkDelete(d *schema.ResourceData, meta interface{}
 		return err
 	}
 
-	time.Sleep(60 * time.Second)
+	stateConf := resource.StateChangeConf{
+		Pending: []string{apigateway.VpcLinkStatusPending,
+			apigateway.VpcLinkStatusAvailable,
+			apigateway.VpcLinkStatusDeleting},
+		Target:     []string{"deleted"},
+		Timeout:    3 * time.Minute,
+		MinTimeout: 1 * time.Second,
+		Refresh: func() (interface{}, string, error) {
+			resp, err := conn.GetVpcLink(&apigateway.GetVpcLinkInput{
+				VpcLinkId: aws.String(d.Id()),
+			})
+			if err != nil {
+				if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
+					return nil, "deleted", nil
+				}
+				return nil, "failed", err
+			}
+			return resp, *resp.Status, nil
+		},
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		return err
+	}
 
 	d.SetId("")
 	return nil
--- a/aws/resource_aws_api_gateway_vpc_link.go
+++ b/aws/resource_aws_api_gateway_vpc_link.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/apigateway"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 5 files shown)


atsushi-ishibashi:
```
BadRequestException: Target arns are not allowed to be changed/added.
    status code: 400, request id: c45e681f-d6b7-11e7-b794-51b5b296ccda

BadRequestException: More than one target arn specified for vpc link.
```
So I set `target_arn` as `ForceNew` and `TypeString`. I'll ask AWS support center.

atsushi-ishibashi:
😕 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsAPIGatewayVpcLink_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAwsAPIGatewayVpcLink_basic -timeout 120m
=== RUN   TestAccAwsAPIGatewayVpcLink_basic
--- FAIL: TestAccAwsAPIGatewayVpcLink_basic (515.16s)
	testing.go:563: Error destroying resource! WARNING: Dangling resources
		
		* aws_lb.test_a (destroy): 1 error(s) occurred:
		
		* aws_lb.test_a: Error deleting ALB: OperationNotPermitted: Load balancer 'arn:aws:elasticloadbalancing:us-west-2:295335637120:loadbalancer/net/tf-lb-a-sa2n7/25a8919dc2841c43' cannot be deleted because deletion protection is enabled
			status code: 400, request id: 7fe9a326-d721-11e7-a8d8-3922343a3b38
		
		State: aws_lb.test_a:
		  ID = arn:aws:elasticloadbalancing:us-west-2:295335637120:loadbalancer/net/tf-lb-a-sa2n7/25a8919dc2841c43
		  access_logs.# = 1
		  arn = arn:aws:elasticloadbalancing:us-west-2:295335637120:loadbalancer/net/tf-lb-a-sa2n7/25a8919dc2841c43
		  arn_suffix = net/tf-lb-a-sa2n7/25a8919dc2841c43
		  dns_name = tf-lb-a-sa2n7-25a8919dc2841c43.elb.us-west-2.amazonaws.com
		  enable_deletion_protection = false
		  idle_timeout = 60
		  internal = true
		  ip_address_type = ipv4
		  load_balancer_type = network
```

aws/resource_aws_api_gateway_vpc_link_test.go
```diff
@@ -0,0 +1,144 @@
...
+  availability_zone = "us-west-2a"
+}
+
+resource "aws_api_gateway_vpc_link" "test" {
+  name = "tf-apigateway-update-%s"
+  description = "test update"
+  target_arn = "${aws_lb.test_a.arn}"
+}
+`, rName, rName)
+}
```
Ninir: We should also add a test with an actual API Gateway and a HTTP endpoint, so that we ensure it's all ok.
I'm actually wondering whether we need a hard dependency (using `depends_on`) on the HTTP integration since this resource is just creating a link between the API Gateway service & the VPC targets... 🤔 
Not sure if this could result in race dependencies! 🤷‍♂️

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,187 @@
...
+	resp, err := conn.CreateVpcLink(input)
+	if err != nil {
+		return err
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{apigateway.VpcLinkStatusPending},
+		Target:     []string{apigateway.VpcLinkStatusAvailable},
+		Refresh:    apigatewayVpcLinkRefreshStatusFunc(conn, *resp.Id),
+		Timeout:    10 * time.Minute,
```
Ninir: Does the timeout really need to be this big? (same goes for the update)

website/docs/r/api_gateway_vpc_link.html.markdown
```diff
@@ -0,0 +1,45 @@
+---
+layout: "aws"
+page_title: "AWS: aws_api_gateway_vpc_link"
+sidebar_current: "docs-aws-resource-api-gateway-vpc-link"
+description: |-
+  Provides an API Gateway VPC Link.
+---
+
+# aws_api_gateway_usage_plan
```
Ninir: Copy/pasta typo? 😄

aws/resource_aws_api_gateway_vpc_link_test.go
atsushi-ishibashi: Sorry, I couldn't understand what you meant so cloud you describe more details:bow: ? @Ninir

aws/resource_aws_api_gateway_vpc_link.go
atsushi-ishibashi: Docs says
> Creates a VPC link, under the caller's account in a selected region, in an asynchronous operation that typically takes 2-4 minutes to complete and become operational.

So I set 5 minutes:bow:

website/docs/r/api_gateway_vpc_link.html.markdown
atsushi-ishibashi: 😅

aws/resource_aws_api_gateway_vpc_link_test.go
radeksimko: I guess what @Ninir means is that we should add a test which tests the full setup including API Gateway integration leveraging the VPC Link.

It seems we'd also need to update other API Gateway resources to make that happen, namely to add few more attributes to `aws_api_gateway_integration` per http://docs.aws.amazon.com/apigateway/api-reference/resource/integration/#connectionType

I'm personally ok with shipping this resource as is and addressing the above in a separate PR to keep diffs small and easier to review - what do you think @Ninir ?

SmithKevin:
Just checking on the status of this as I am really looking forward to this feature.   Looks like it has stalled a bit over the holidays.

Ninir:
@atsushi-ishibashi Just encountering an issue with the tests:

```
$ TF_LOG=DEBUG TF_LOG_PATH=terraform.log make testacc TEST=./aws TESTARGS='-run=TestAccAwsAPIGatewayVpcLink_'

TF_ACC=1 go test ./aws -v -run=TestAccAwsAPIGatewayVpcLink_ -timeout 120m
=== RUN   TestAccAwsAPIGatewayVpcLink_basic
--- FAIL: TestAccAwsAPIGatewayVpcLink_basic (582.98s)
	testing.go:563: Error destroying resource! WARNING: Dangling resources
		
		* aws_lb.test_a (destroy): 1 error(s) occurred:
		* aws_lb.test_a: Error deleting ALB: ResourceInUse: Load balancer 'arn:aws:elasticloadbalancing:us-west-2:XXXXXXXXX:loadbalancer/net/tf-lb-a-esy6k/c86aa43c1258c6ce' cannot be deleted because it is currently associated with another service
			status code: 400, request id: b55115ef-f0bb-11e7-8cc6-6b0dd342af2b
```

Do you have time to investigate?

Thanks!

atsushi-ishibashi:
@Ninir Uhhh.. There may be possibility that vpc link was deleting when api to delete nlb was called.

atsushi-ishibashi:
After `DeleteVpcLink` was called, wait for a minute then test passed.
```
func resourceAwsApiGatewayVpcLinkDelete(d *schema.ResourceData, meta interface{}) error {
	conn := meta.(*AWSClient).apigateway

	input := &apigateway.DeleteVpcLinkInput{
		VpcLinkId: aws.String(d.Id()),
	}

	_, err := conn.DeleteVpcLink(input)
	if err != nil {
		if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
			d.SetId("")
			return nil
		}
		return err
	}

	time.Sleep(60 * time.Second)

	d.SetId("")
	return nil
}
```

What is the best way to wait until completely deleted?

atsushi-ishibashi:
`VpcLinkStatus` doesn't have `DELETED` so we cannot apply `WaitForState` 😕 
```
const (
    // VpcLinkStatusAvailable is a VpcLinkStatus enum value
    VpcLinkStatusAvailable = "AVAILABLE"

    // VpcLinkStatusPending is a VpcLinkStatus enum value
    VpcLinkStatusPending = "PENDING"

    // VpcLinkStatusDeleting is a VpcLinkStatus enum value
    VpcLinkStatusDeleting = "DELETING"

    // VpcLinkStatusFailed is a VpcLinkStatus enum value
    VpcLinkStatusFailed = "FAILED"
)
```

bflad:
I have it's possible to use `resource.StateChangeConf` in these cases where you set `Pending` to all the mentioned status constants and `Target` to `""` which you can return when you receive the `NotFoundException`

atsushi-ishibashi:
Good idea👍 Thanks!

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,200 @@
...
+	resp, err := conn.CreateVpcLink(input)
+	if err != nil {
+		return err
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{apigateway.VpcLinkStatusPending},
+		Target:     []string{apigateway.VpcLinkStatusAvailable},
+		Refresh:    apigatewayVpcLinkRefreshStatusFunc(conn, *resp.Id),
+		Timeout:    5 * time.Minute,
```
lukehoban: I was trying this PR out, and hit a case where this operation timed out after 5 minutes.  The VPC Link did end up being successfully created.  The AWS docs note that this can take "2-4 minutes", so it may be worth allowing more buffer here (and on update), to ensure these don't time out under normal conditions?

atsushi-ishibashi: Thank you for your advice!

atsushi-ishibashi:
@Ninir Ok, the test passed
```
TF_ACC=1 go test ./aws -v -run=TestAccAwsAPIGatewayVpcLink_basic -timeout 120m
=== RUN   TestAccAwsAPIGatewayVpcLink_basic
--- PASS: TestAccAwsAPIGatewayVpcLink_basic (505.03s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	505.066s
```

website/docs/r/api_gateway_vpc_link.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name used to label and identify the VPC link.
+* `description` - (Optional) The description of the VPC link.
+* `target_arn` - (Required, ForceNew) The ARN of network load balancer of the VPC targeted by the VPC link.
```
a-h: (Required, ForceNew) The ARN of a network load balancer in the VPC targeted by the VPC link.

christophwitzko:
Hey folks, this looks very promising. What is the status of this pull request? Would love to see this resource in terraform. ✌️

mattvv:
Also checking in on this.

I'm also thinking that `aws_api_gateway_integration` will need to be updated to allow resources with type **VPC**

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,200 @@
...
+		Timeout:    8 * time.Minute,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("[WARN] Error waiting for APIGateway Vpc Link status to be \"%s\": %s", apigateway.VpcLinkStatusAvailable, err)
+	}
+
+	d.SetId(*resp.Id)
```
radeksimko: We usually set the ID as soon as we can after receiving it from the API, _before_ making any further requests so that it's recorded in the state in case any future API requests fail for any reason. Do you mind moving it above the `StateChangeConf` block?

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,200 @@
...
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"target_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
radeksimko: Is there any particular reason this should be implemented as 1-1 relationship instead of reflecting the API, i.e. allowing users to assign multiple LBs to one VPC link?

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,200 @@
...
+	conn := meta.(*AWSClient).apigateway
+
+	input := &apigateway.GetVpcLinkInput{
+		VpcLinkId: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetVpcLink(input)
+	if err != nil {
+		if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
+			d.SetId("")
```
radeksimko: Nitpick: Do you mind adding our usual `WARN` log message here?

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,200 @@
...
+			return nil
+		}
+		return err
+	}
+
+	stateConf := resource.StateChangeConf{
+		Pending: []string{apigateway.VpcLinkStatusPending,
+			apigateway.VpcLinkStatusAvailable,
+			apigateway.VpcLinkStatusDeleting},
+		Target:     []string{"deleted"},
```
radeksimko: Using empty string here is equally valid and in fact preferred, over making up our own statuses 😉

aws/resource_aws_api_gateway_vpc_link.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+			return resp, *resp.Status, nil
+		},
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		return err
+	}
+
+	d.SetId("")
```
radeksimko: Nitpick: I know we still have it in many resources, but it's redundant. The ID is emptied automatically in the core.

See https://github.com/hashicorp/terraform/blob/096847c9f774bfb946de7413260b30f9f6041241/helper/schema/resource.go#L197-L206

aws/resource_aws_api_gateway_vpc_link_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+		resp, err := conn.GetVpcLink(input)
+		if err != nil {
+			if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
+				return nil
+			}
+			return err
+		}
+
+		if *resp.Status != apigateway.VpcLinkStatusDeleting {
```
radeksimko: I think we should treat any positive response here as a failure - the VPC link should be already deleted by the time test finishes and practically have no _status_ (i.e. it shouldn't be in `deleting` status).

aws/resource_aws_api_gateway_vpc_link_test.go
```diff
@@ -0,0 +1,132 @@
...
+	return nil
+}
+
+func testAccCheckAwsAPIGatewayVpcLinkExists(name string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		_, ok := s.RootModule().Resources[name]
+		if !ok {
+			return fmt.Errorf("Not found: %s", name)
+		}
+
```
radeksimko: Do you mind checking the API here also?

website/docs/r/api_gateway_vpc_link.html.markdown
atsushi-ishibashi: Thank you for correct me!

atsushi-ishibashi:
@radeksimko I remembered that the interface could accept a multiple of lb but you can apply only one lb. Refer to web console as well.
```
TF_ACC=1 go test ./aws -v -run=TestAccAwsAPIGatewayVpcLink_basic -timeout 120m
=== RUN   TestAccAwsAPIGatewayVpcLink_basic
--- FAIL: TestAccAwsAPIGatewayVpcLink_basic (589.96s)
	testing.go:503: Step 0 error: Check failed: Check 4/4 error: aws_api_gateway_vpc_link.test: Attribute 'target_arns.#' expected "2", got "1"
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	589.996s
make: *** [testacc] Error 1
```
However in the future, aws may support multiple so it's better to use `MaxItem` for now. What do you think?

aws/resource_aws_api_gateway_vpc_link_test.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+data "aws_availability_zones" "test" {}
+
+resource "aws_subnet" "test" {
+  vpc_id = "${aws_vpc.test.id}"
+  cidr_block = "10.10.0.0/21"
+  availability_zone = "${data.aws_availability_zones.test.names[0]}"
+}
+`, rName, rName)
```
radeksimko: The main reason the test is failing the way it's failing is because two resources (`aws_lb.test_a` and `aws_lb.test_b`) both have the same name and the `TypeSet` will internally merge them into a single ARN since the ARN is the same.

radeksimko: Once you fix that you get more reasonable error which I think we can just present to the user:

```
* aws_api_gateway_vpc_link.test: BadRequestException: More than one target arn specified for vpc link.
  status code: 400, request id: d4eabe6c-0ff8-11e8-bea2-d9187e5e8757
```

As you rightly mentioned Amazon currently doesn't support more than 1 LB though, so we'll have to keep this test focused on single LB only.

atsushi-ishibashi: That's right! Thank you for correct.

atsushi-ishibashi:
@radeksimko Ok👍
```
TF_ACC=1 go test ./aws -v -run=TestAccAwsAPIGatewayVpcLink_basic -timeout 120m
=== RUN   TestAccAwsAPIGatewayVpcLink_basic
--- PASS: TestAccAwsAPIGatewayVpcLink_basic (633.92s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	633.961s
```

bflad:

PR #2522, opened 2017-12-04
Fixes example

alindeman:
The example uses an `ssh-rsa` formatted key, but the `encoding` is set to `PEM`.


Ninir:
Hi @alindeman 

Since both of these protocols are returning the exact same thing (tracing the api call returns the same key body), not sure to get the change.

Can you provide more inputs there?

Thanks!

alindeman:
@Ninir All I know is that `ssh-rsa ...` is not `PEM` formatted. The example using `ssh-rsa ...` but `PEM` struck me as wrong.

jen20:
This change looks to me - the key in the example is indeed in the OpenSSH format and not PEM-encoded. Thanks @alindeman!

PR #2530, opened 2017-12-04
Add organisations data source

dblooman:
This adds the ability to get a list of account ID's from the organisations service.

### DIFF
--- a/aws/data_source_aws_organization_attributes.go
+++ b/aws/data_source_aws_organization_attributes.go
@@ -1,0 +1,91 @@
+package aws
+
+import (
+	"log"
+
+	"github.com/aws/aws-sdk-go/service/organizations"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsOrganizationAttributes() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsOrganizationsAttributesRead,
+
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"master_account_arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"master_account_email": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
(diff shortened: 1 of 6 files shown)


aws/data_source_aws_organizations.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsOrganizations() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsOrganizationsAccountsRead,
+
+		Schema: map[string]*schema.Schema{
+			"account_id": {
```
Ninir: As a Terraform User, I would expect this data source to return Organisation information, not only members.

So first I would expect [describe organisation attributes](https://docs.aws.amazon.com/organizations/latest/APIReference/API_DescribeOrganization.html) and then call the ListAccounts endpoint to get users.
It could even be a separated data source.

Thoughts?

DaveBlooman:
Hi @Ninir.  I have made some changes, split the data sources into two, this should hopefully be clearer now.

aws/data_source_aws_organization_attributes.go
```diff
@@ -0,0 +1,91 @@
+package aws
```
Ninir: Why not simply naming the file data_source_aws_organization.go? (and other ones below)
Isn't this the main object representing an Organization?

bflad:
@DaveBlooman will you have time to address PR feedback? No worries if not, we can finish this up for you. 👍

DaveBlooman:
@bflad Probably wont have time this week, so go ahead :)

brandonstevens:
@dblooman I realize it's been awhile, but are you able to pick this back up. Since you last left off, a few new PRs duplicate and build on your work. If you're not able to, I'm happy to take it across the finish line for you.

If I understand the feedback correctly (@Ninir correct me if I'm wrong), but the data sources (and file names) just need to be updated. For consistency, it seems like they should be updated to `aws_organizations_organization` and `aws_organizations_account_ids`. Once merged in, the other PRs can be reworked to build on it.

brandonstevens:
@dblooman , @Ninir do you guys mind if I take over this PR and make the required updates myself?

PR #2543, opened 2017-12-05
resource/aws_s3_bucket: Mark replication_configuration rules id attribute as required

### LINKED ISSUE #665: aws_s3_bucket: replication_configuration shows changes when there are none
_This issue was originally opened by @PeteGoo as hashicorp/terraform#13352. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
0.8.8
0.9.2

### Affected Resource(s)
- aws_s3_bucket

```hcl
provider "aws" {
  region = "us-east-1"
}

provider "aws" {
  region = "us-west-2"
  alias  = "us-west-2"
}

resource "aws_s3_bucket" "origin-bucket" {
  bucket = "715489234-origin-bucket"
  acl    = "private"

  versioning {
    enabled = true
  }

  replication_configuration {
    role = "${aws_iam_role.replication.arn}"

    rules {
      prefix = ""
      status = "Enabled"

      destination {
        bucket        = "arn:aws:s3:::${aws_s3_bucket.replicated-bucket.id}"
        storage_class = "STANDARD"
      }
    }
  }
}

resource "aws_s3_bucket" "replicated-bucket" {
  bucket   = "715489234-replicated-bucket-us-west-2"
  acl      = "private"
  region   = "us-west-2"
  provider = "aws.us-west-2"

  versioning {
    enabled = true
  }
}

resource "aws_iam_role" "replication" {
  name = "715489234-replication-role"

  assume_role_policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "sts:AssumeRole",
      "Principal": {
        "Service": "s3.amazonaws.com"
      },
      "Effect": "Allow",
      "Sid": ""
    }
  ]
}
POLICY
}

resource "aws_iam_role_policy" "replication" {
  role = "${aws_iam_role.replication.id}"
  name = "715489234-replication-policy"

  policy = <<POLICY
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "s3:GetReplicationConfiguration",
        "s3:ListBucket"
      ],
      "Effect": "Allow",
      "Resource": [
        "arn:aws:s3:::${aws_s3_bucket.origin-bucket.id}"
      ]
    },
    {
      "Action": [
        "s3:GetObjectVersion",
        "s3:GetObjectVersionAcl"
      ],
      "Effect": "Allow",
      "Resource": [
        "arn:aws:s3:::${aws_s3_bucket.origin-bucket.id}/*"
      ]
    },
    {
      "Action": [
        "s3:ReplicateObject",
        "s3:ReplicateDelete"
      ],
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::${aws_s3_bucket.replicated-bucket.id}/*"
    }
  ]
}
POLICY
}

```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
A plan after the first apply should be empty

### Actual Behavior
The plan after the first apply shows changes in the replication_configuration

```
~ aws_s3_bucket.origin-bucket
    replication_configuration.0.rules.2757951880.destination.#:                        "1" => "0"
    replication_configuration.0.rules.2757951880.destination.1666013429.bucket:        "arn:aws:s3:::715489234-replicated-bucket-us-west-2" => ""
    replication_configuration.0.rules.2757951880.destination.1666013429.storage_class: "STANDARD" => ""
    replication_configuration.0.rules.2757951880.id:                                   "NjA4ZDc2YzAtNzdkNy00NGM0LWI3MmMtZDI1MjgwYjNhZTQw" => ""
    replication_configuration.0.rules.2757951880.prefix:                               "" => ""
    replication_configuration.0.rules.2757951880.status:                               "Enabled" => ""
    replication_configuration.0.rules.3415946663.destination.#:                        "0" => "1"
    replication_configuration.0.rules.3415946663.destination.1666013429.bucket:        "" => "arn:aws:s3:::715489234-replicated-bucket-us-west-2"
    replication_configuration.0.rules.3415946663.destination.1666013429.storage_class: "" => "STANDARD"
    replication_configuration.0.rules.3415946663.id:                                   "" => ""
    replication_configuration.0.rules.3415946663.prefix:                               "" => ""
    replication_configuration.0.rules.3415946663.status:                               "" => "Enabled"
```

### Steps to Reproduce
1. `terraform apply`
2. `terraform plan`
3. repeat

### Important Factoids
The id of the replication rule seems to be the only thing that changes in the plan. Perhaps it is being inconsistently used to calculate a hash for change detection?

bflad:
Closes #665 

The `rules` under `replication_configuration` are currently TypeSet with the hash computed including the ID attribute when available. However, when creating a rule that omits the ID attribute (allowing AWS to generate the ID), the plan will always show a difference due to the hash changing. Rather than deal with a state migration to TypeList on a critical resource for seemingly little benefit, set this attribute as `Required` as the least effort option to make the resource configuration less surprising.

I do not believe this needs to be marked as a breaking change, due to the current behavior.

If there are better ways to handle this via the `Set` function or some other way, please let me know!

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -351,7 +351,7 @@ func resourceAwsS3Bucket() *schema.Resource {
 								Schema: map[string]*schema.Schema{
 									"id": {
 										Type:         schema.TypeString,
-										Optional:     true,
+										Required:     true,
 										ValidateFunc: validateS3BucketReplicationRuleId,
 									},
 									"destination": {
--- a/website/docs/r/s3_bucket.html.markdown
+++ b/website/docs/r/s3_bucket.html.markdown
@@ -381,7 +381,7 @@ The `replication_configuration` object supports the following:
 
 The `rules` object supports the following:
 
-* `id` - (Optional) Unique identifier for the rule.
+* `id` - (Required) Unique identifier for the rule.
 * `destination` - (Required) Specifies the destination for the rule (documented below).
 * `prefix` - (Required) Object keyname prefix identifying one or more objects to which the rule applies. Set as an empty string to replicate the whole bucket.
 * `status` - (Required) The status of the rule. Either `Enabled` or `Disabled`. The rule is ignored if status is not Enabled.


radeksimko:
Hi @bflad 
I'd be ok with making it `Required` if there's a way to modify the config to comply with this change. I don't believe there's an easy way at this point. I'd have to go to AWSCLI/API/Console to find the ID to paste it into config _after_ the replication was enabled, unless I'm mistaken?

---

We have historically kept all of S3 bucket-related functionality inside `aws_s3_bucket` which has pros and cons.

Pros include easier protection against effects of eventual consistency (because all retries are in a single place and happen roughly at the same time sequentially: https://github.com/terraform-providers/terraform-provider-aws/pull/891/files ).

Cons include that it's trickier to address bugs like this one. Having a separate resource like `aws_s3_bucket_replication_configuration` would IMO help here.

We discussed this together in https://github.com/terraform-providers/terraform-provider-aws/pull/916 in the past and this PR is bringing more arguments to that discussion - thank you! 👍 

The transition towards the state won't be easy because we'll possibly have to deprecate a bunch of fields and/or revisit the idea of nested resources where the CRUD is shared.

---

Discussions aside I think in short-term we could theoretically make this field `Computed` and omit it from the Set hash, so it's not raising spurious diffs.

Let me know what you think.

bflad:
I believe your assessment of how to "fix" the spurious diff when `id` is omitted is correct. Essentially we require it at the moment, either you setting it upfront or finding it after the fact and setting it in the configuration to prevent the diff.

Would changing the Set hash to omit `id` be a "breaking" change for folks who already have rules with `id` configured in their hash (e.g. plan diff to recreate the same rules with new hashes)? I was thinking that would require a state migration to not show the one time diff. 

I'm certainly not opposed to doing it the better way here with state migration, just requires more effort and testing. :)

bflad:
Closing this one out for now, I don't have time to properly adjust it.

PR #2547, opened 2017-12-05
Support CodeBuild within VPC

### LINKED ISSUE #2495: Add support for CodeBuild with VPC
Amazon [announced](https://aws.amazon.com/blogs/devops/access-resources-in-a-vpc-from-aws-codebuild-builds/) support for running CodeBuild builds within a VPC last week.

### Affected Resource(s)

* aws_codebuild_project

### Expected Behavior

Create & update of projects should optionally allow the specification of VPC, subnets & security groups, likely within a new optional block to match the Create and Update APIs. Read will also need additions for these properties.

* If VPC is supplied, subnets must also be supplied.
* If subnets is supplied, VPC must also be supplied.
* If security_groups is supplied, both VPC and subnets must be supplied.

```hcl
resource "aws_codebuild_project" "foo" {
  name = "test-project"

  artifacts {
    type = "NO_ARTIFACTS"
  }

  environment {
    compute_type = "BUILD_GENERAL1_SMALL"
    image        = "2"
    type         = "LINUX_CONTAINER"
  }

  source {
    type     = "GITHUB"
    location = "https://github.com/mitchellh/packer.git"
  }

  vpc_config {
    vpc_id = "vpc-725fca"

    subnets = [
       "subnet-ba35d2e0",
       "subnet-ab129af1",
    ]

    security_group_ids = [
       "sg-f9f27d91",
       "sg-e4f48g23",
    ]
  }
}
```

### References

[Blog post](https://aws.amazon.com/blogs/devops/access-resources-in-a-vpc-from-aws-codebuild-builds/)
[Documentation](https://docs.aws.amazon.com/codebuild/latest/userguide/vpc-support.html)

drsolaris:
Fixes #2495 

First attempt at a go contribution, happy to implement any suggestions.

- [x] Update `aws_codebuild_project`.
- [x] Acceptance test.
- [x] Documentation.

There is an issue with the acceptance test (hence WIP). Creation of the CodeBuild project requires the DescribeSecurityGroups permission. This is added to the IAM role, but there is a race condition between the policy actually being applied and the creation of the CodeBuild project, meaning that it can fail. I've added a `depends_on = ["aws_iam_policy_attachment.codebuild_policy_attachment"]` clause to the `aws_codebuild_policy` resource, but this doesn't help. I guess it is the natural lag in applying IAM policies. Any ideas? Edit: **this is fixed now**.

### DIFF
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -258,6 +258,10 @@ func resourceAwsCodeBuildProjectCreate(d *schema.ResourceData, meta interface{})
 				return resource.RetryableError(err)
 			}
 
+			if isAWSErr(err, "InvalidInputException", "Not authorized to perform DescribeSecurityGroups") {
+				return resource.RetryableError(err)
+			}
+
 			return resource.NonRetryableError(err)
 		}
 
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -179,6 +179,31 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 				ValidateFunc: validateAwsCodeBuildTimeout,
 			},
 			"tags": tagsSchema(),
+			"vpc_config": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"vpc_id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"subnets": {
+							Type:     schema.TypeList,
+							Required: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							MaxItems: 16,
+						},
+						"security_group_ids": {
+							Type:     schema.TypeList,
+							Required: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							MaxItems: 5,
+						},
+					},
+				},
+			},
 		},
 	}
 }
@@ -211,6 +236,11 @@ func resourceAwsCodeBuildProjectCreate(d *schema.ResourceData, meta interface{})
 
 	if v, ok := d.GetOk("build_timeout"); ok {
 		params.TimeoutInMinutes = aws.Int64(int64(v.(int)))
+	}
+
+	if _, ok := d.GetOk("vpc_config"); ok {
+		vpcConfig := expandVpcConfig(d)
+		params.VpcConfig = vpcConfig
 	}
 
 	if v, ok := d.GetOk("tags"); ok {
@@ -324,6 +354,29 @@ func expandProjectEnvironment(d *schema.ResourceData) *codebuild.ProjectEnvironm
 	return projectEnv
 }
 
+func expandVpcConfig(d *schema.ResourceData) *codebuild.VpcConfig {
+	configs := d.Get("vpc_config").(*schema.Set).List()
+	data := configs[0].(map[string]interface{})
+
+	vpcConfig := codebuild.VpcConfig{
+		VpcId: aws.String(data["vpc_id"].(string)),
+	}
+
+	var vpcConfigSubnets []*string
+	for _, v := range data["subnets"].([]interface{}) {
+		vpcConfigSubnets = append(vpcConfigSubnets, aws.String(v.(string)))
+	}
+	vpcConfig.Subnets = vpcConfigSubnets
+
+	var vpcSecurityGroupIds []*string
+	for _, s := range data["security_group_ids"].([]interface{}) {
+		vpcSecurityGroupIds = append(vpcSecurityGroupIds, aws.String(s.(string)))
+	}
+	vpcConfig.SecurityGroupIds = vpcSecurityGroupIds
+
+	return &vpcConfig
+}
+
 func expandProjectSource(d *schema.ResourceData) codebuild.ProjectSource {
 	configs := d.Get("source").(*schema.Set).List()
 	projectSource := codebuild.ProjectSource{}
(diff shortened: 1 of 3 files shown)


bflad:
Ah yes, the evils of IAM eventual consistency! 🙁 In many downstream resources we opt to retry creation actions if the AWS error is specific enough to indicate its some form of permissions error code/message.

I haven't looked at the codebuild resource, but here's an example from codedeploy which might help: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_codedeploy_deployment_group.go#L391-L423

drsolaris:
Aha! Thanks for the quick reply. Guess I should have looked harder. The CodeBuild code has the same retry logic, but the error message slightly doesn't match.

```go
if isAWSErr(err, "InvalidInputException", "CodeBuild is not authorized to perform") {
}
```

My error is: `InvalidInputException: Not authorized to perform DescribeSecurityGroups`

Should I change the existing line to this? (I think it might end up retrying lots if the end user's IAM credentials are not authorized to perform actions, rather than just CodeBuild)

```go
if isAWSErr(err, "InvalidInputException", "ot authorized to perform") {
}
```

The safe way is to just add another error message capture, I guess. Preferences?

drsolaris:
Went ahead with the safer route after giving it some more thought. This should be good to go now.

kilsbo:
Please merge! 💯

aws/resource_aws_codebuild_project.go
```diff
@@ -324,6 +358,29 @@ func expandProjectEnvironment(d *schema.ResourceData) *codebuild.ProjectEnvironm
 	return projectEnv
 }
 
+func expandVpcConfig(d *schema.ResourceData) *codebuild.VpcConfig {
```
bflad: We really try to limit the usage of `*schema.ResourceData` in downstream functions unless its necessary in multiple pieces of logic. It helps make the code more testable and reusable (if necessary in the future). Could you update this function's signature to `expandCodeBuildVpcConfig(m map[string]interface{}) *codebuild.VpcConfig`? Thanks!

aws/resource_aws_codebuild_project_test.go
```diff
@@ -401,6 +442,18 @@ resource "aws_codebuild_project" "foo" {
   tags {
     "Environment" = "Test"
   }
+
+  vpc_config {
```
bflad: Since this new attribute is optional and has some complexity, we should probably split it out into its own acceptance test. We need to verify the resource creation (what's done here already), the update of `vpc_config` attributes (e.g. add/remove subnet), and the removal of the `vpc_config` altogether from a configuration (if removing VPC configuration is supported). It would also be good to check that the new attributes are at least making it correctly into the state. e.g.
```
resource.TestCheckResourceSet("aws_codebuild_project.foo", "vpc_config.0.vpc_id"),
resource.TestCheckResource("aws_codebuild_project.foo", "vpc_config.0.subnets.#", "1"),
resource.TestCheckResource("aws_codebuild_project.foo", "vpc_config.0.security_group_ids.#", "1"),
```

aws/resource_aws_codebuild_project.go
```diff
@@ -324,6 +358,29 @@ func expandProjectEnvironment(d *schema.ResourceData) *codebuild.ProjectEnvironm
...
+	configs := d.Get("vpc_config").(*schema.Set).List()
+	data := configs[0].(map[string]interface{})
```
bflad: This line is generating a panic for me running the acceptance test:
```
=== RUN   TestAccAWSCodeBuildProject_basic
panic: runtime error: index out of range

goroutine 2579 [running]:
github.com/terraform-providers/terraform-provider-aws/aws.expandVpcConfig(0xc42084b110, 0x3715e47)
	/Users/bflad/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_codebuild_project.go:363 +0x790
```

It will go away with the function signature refactor and TypeList changes mentioned above.

aws/resource_aws_codebuild_project.go
```diff
@@ -179,6 +179,31 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 				ValidateFunc: validateAwsCodeBuildTimeout,
 			},
 			"tags": tagsSchema(),
+			"vpc_config": {
+				Type:     schema.TypeSet,
```
bflad: I would switch this to `Type:     schema.TypeList,`. There are subtleties between the two, but the gist being that list in this context is much easier to work with when there's only 1 item.

aws/resource_aws_codebuild_project.go
```diff
@@ -213,6 +238,11 @@ func resourceAwsCodeBuildProjectCreate(d *schema.ResourceData, meta interface{})
 		params.TimeoutInMinutes = aws.Int64(int64(v.(int)))
 	}
 
+	if _, ok := d.GetOk("vpc_config"); ok {
```
bflad: I would change this to (note about function signature update below in another comment):
```
if v, ok := d.GetOk("vpc_config"); ok {
	params.VpcConfig = expandCodeBuildVpcConfig(v.([]interface{})[0].(map[string]interface{}))
}
```

aws/resource_aws_codebuild_project.go
```diff
@@ -390,6 +447,10 @@ func resourceAwsCodeBuildProjectRead(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
+	if err := d.Set("vpc_config", schema.NewSet(resourceAwsCodeBuildVpcConfigHash, flattenAwsCodebuildVpcConfig(project.VpcConfig))); err != nil {
```
bflad: When this is TypeList this can become:
```
if err := d.Set("vpc_config", flattenAwsCodebuildVpcConfig(project.VpcConfig)); err != nil {
```

aws/resource_aws_codebuild_project.go
```diff
@@ -567,6 +653,25 @@ func resourceAwsCodeBuildProjectArtifactsHash(v interface{}) int {
 	return hashcode.String(buf.String())
 }
 
+func resourceAwsCodeBuildVpcConfigHash(v interface{}) int {
```
bflad: This is extraneous with switch to `TypeList` 👍

bflad:
This support has been merged in with your commits squashed via #3324 and will be released in v1.10.0 of the AWS provider, likely tomorrow. Thanks for you work here!

bflad:

PR #2551, opened 2017-12-05
d/aws_kms_alias: Add target_key_arn attribute

### LINKED ISSUE #2009: Support ARNs for existing KMS keys

bflad:
This is an alternate implementation to #2224. Closes #2009. 

KMS keys and aliases are bound to the same AWS partition, account, and region so it is safe to assemble the target key ARN from the alias ARN and target key ID. Reference: http://docs.aws.amazon.com/sdk-for-go/api/service/kms/#KMS.CreateAlias

> The alias and the CMK it is mapped to must be in the same AWS account and the same region. You cannot perform this operation on an alias in a different AWS account.

```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsKmsAlias'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsKmsAlias -timeout 120m
=== RUN   TestAccDataSourceAwsKmsAlias
--- PASS: TestAccDataSourceAwsKmsAlias (41.76s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	41.791s
```

### DIFF
--- a/aws/data_source_aws_kms_alias_test.go
+++ b/aws/data_source_aws_kms_alias_test.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/acctest"
@@ -44,6 +45,15 @@ func testAccDataSourceAwsKmsAliasCheck(name string) resource.TestCheckFunc {
 				"arn is %s; want %s",
 				attr["arn"],
 				kmsKeyRs.Primary.Attributes["arn"],
+			)
+		}
+
+		expectedTargetKeyArnSuffix := fmt.Sprintf("key/%s", kmsKeyRs.Primary.Attributes["target_key_id"])
+		if !strings.HasSuffix(attr["target_key_arn"], expectedTargetKeyArnSuffix) {
+			return fmt.Errorf(
+				"target_key_arn is %s; want suffix %s",
+				attr["target_key_arn"],
+				expectedTargetKeyArnSuffix,
 			)
 		}
 
--- a/aws/data_source_aws_kms_alias.go
+++ b/aws/data_source_aws_kms_alias.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -20,6 +21,10 @@ func dataSourceAwsKmsAlias() *schema.Resource {
 				ValidateFunc: validateAwsKmsName,
 			},
 			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"target_key_arn": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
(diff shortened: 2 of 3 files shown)


aws/data_source_aws_kms_alias_test.go
```diff
@@ -47,6 +48,19 @@ func testAccDataSourceAwsKmsAliasCheck(name string) resource.TestCheckFunc {
...
+			kmsKeyRs.Primary.Attributes["name"],
+			fmt.Sprintf("key/%s", kmsKeyRs.Primary.Attributes["target_key_id"]),
+			1)
+		if attr["target_key_arn"] != targetKeyArn {
+			return fmt.Errorf(
+				"target_key_arn is %s; want %s",
+				attr["target_key_arn"],
+				targetKeyArn,
+			)
+		}
```
radeksimko: How do you feel about `regexp` matching with `$` here? It seems that would be a little bit cleaner approach compared to modification of the original attribute.

starlton-eb: This would be awesome to have right now. I racked by brain all day wondering why my aws_efs_file_system instances were being recreated and discovered that it was due to a recent update to our stack that includes looking up a kms key:

bflad:
I'll update and merge this PR after we release v1.7.1 (hopefully tomorrow). 🚀

bflad:
Rebased and updated if you'd like to give another look
```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsKmsAlias'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsKmsAlias -timeout 120m
=== RUN   TestAccDataSourceAwsKmsAlias
--- PASS: TestAccDataSourceAwsKmsAlias (40.72s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	40.769s
```

ColinHebert:
Helps/fixes #3019

bflad:

Issue #85, opened 2017-06-13
Terraform unable to create ecs_service if iam_role is set


hashibot:
_This issue was originally opened by @egoldschmidt as hashicorp/terraform#5292. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I've come across an issue in which an ECS service apparently cannot be created if `iam_role` is set. As you can see from this example I've set it to the ARN of the default ECS service user, but I've tried a custom user to no avil. The only thing that appears to resolve the problem is removing the `iam_role` field altogether. Obviously this is a non-starter because we need ELBs, so any ideas would be appreciated.

```
module.vpn_service.aws_ecs_service.vpn: Creating...
  cluster:                                 "" => "arn:aws:ecs:us-west-2:xxx:cluster/vpn"
  desired_count:                           "" => "1"
  iam_role:                                "" => "arn:aws:iam::xxx:role/ecsServiceRole"
  load_balancer.#:                         "" => "2"
  load_balancer.3901468935.container_name: "" => "vpn"
  load_balancer.3901468935.container_port: "" => "943"
  load_balancer.3901468935.elb_name:       "" => "vpn-elb"
  load_balancer.450980314.container_name:  "" => "vpn"
  load_balancer.450980314.container_port:  "" => "443"
  load_balancer.450980314.elb_name:        "" => "vpn-elb"
  name:                                    "" => "vpn"
  task_definition:                         "" => "arn:aws:ecs:us-west-2:xxx:task-definition/vpn:4"
Error applying plan:

1 error(s) occurred:

* aws_ecs_service.vpn: timeout while waiting for state to become '[success]'
```

chrisLeeTW:
I also have similar problem. here is my error message.

terraform version : 0.11.2

Error: Error applying plan:

1 error(s) occurred:

* aws_ecs_service.test_service_001: 1 error(s) occurred:

* aws_ecs_service.test_service_001: InvalidParameterException: You cannot specify an IAM role for services that require a service linked role.
	status code: 400, request id: 666f86b5-fb80-11e7-bd3a-cfd17371dfbc "test-service-001"

jackric:
This is affecting me also.
Looking at the code, service creation should fail sooner:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_ecs_service.go#L294
[...]

PR #2559, opened 2017-12-05
Feature/fargate support v2

johnnorton:
The previous merge for ECS Fargate support did not include the ability to ENABLE "Assign Public IP" when creating the ECS Service.

From what I can tell, this is an option that only works for services on Fargate.  This option assigns a public IP to the ENI.  This is required to access the internet from the container if you do not have a NAT gateway setup.


radeksimko:
Hi @johnnorton 
AFAICT the networking block is updatable per API docs:
http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html#ECS-UpdateService-request-networkConfiguration

Can you tell us more about the motivation behind the `ForceNew` in all fields?

Also `Elem` is used to define elements nested under the field, which is only relevant to `TypeList` or `TypeSet`, potentially `TypeMap`, but certainly not `TypeString`.

Finally we'll need to add a test to exercise this new field and ensure it works. Let me know if you need any help with that.

Thanks.

johnnorton:
@radeksimko 
Thanks for the input.  My reasoning behind forcenew on networking was based on the Console not allowing updates to a service.  I will test with the CLI and see.

This is my first foray into GoLang.  I will will take a look at your suggestions and create a test.

Standby....

John

aws/resource_aws_ecs_service.go
```diff
@@ -481,9 +494,8 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
-	if d.HasChange("network_configration") {
-		input.NetworkConfiguration = expandEcsNetworkConfigration(d.Get("network_configuration").([]interface{}))
-	}
+	//d.HasChange("network_configration") is not working, so explicity calling method.
```
johnnorton: Not sure why the the func HadChange is not detecting a change on "network_configuration".  Maybe we need to implement equals on the structure?

aws/resource_aws_ecs_service_test.go
```diff
@@ -1261,6 +1261,7 @@ resource "aws_ecs_service" "main" {
 	network_configuration {
 		security_groups = ["${aws_security_group.allow_all_a.id}", "${aws_security_group.allow_all_b.id}"]
 		subnets = ["${aws_subnet.main.*.id}"]
+		assign_public_ip = "ENABLED"
```
johnnorton: I feel like the test needs something more but I am not sure what else to do besides this....

johnnorton:
@radeksimko.  Yes updating the network config does not need to force a new resource but the existing tasks will not get the new settings as per the docs.  

Please review my comments in the code and let me know.

Thanks!

aws/resource_aws_ecs_service.go
johnnorton: HasChange was not detecting a change here.  Maybe we need to implement the "equals" method?

aws/resource_aws_ecs_service_test.go
johnnorton: I am not sure if this is all I need to do for the test.  Seems like I am missing something

johnnorton:
@radeksimko Anything else needed here?

headlessme:
Would also like to see this merged

dkuntz2:
Is there any news on getting this merged? Not being able to set auto assign public IP is preventing me from configuring fargate-backed ECS services in terraform.

As a note on the PR itself, I think it would be nicer to be able to use `true`/`false` rather than `ENABLED`/`DISABLED` as the value, given that this is just a boolean value and that feels like it's exposing too much of the underlying API in a way that isn't particularly helpful.

johnnorton:
@radeksimko 

Made change suggested by @dkuntz2 dkuntz2

austinkelleher:
This is also preventing me from configuring Fargate with Terraform. @radeksimko Can we please get this merged soon?

aws/resource_aws_ecs_service.go
```diff
@@ -133,6 +133,11 @@ func resourceAwsEcsService() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 							Set:      schema.HashString,
 						},
+						"assign_public_ip": &schema.Schema{
```
bflad: You can remove `&schema.Schema` from this line here.

aws/resource_aws_ecs_service.go
```diff
@@ -410,9 +415,14 @@ func flattenEcsNetworkConfigration(nc *ecs.NetworkConfiguration) []interface{} {
 	if nc == nil {
 		return nil
 	}
+
 	result := make(map[string]interface{})
 	result["security_groups"] = schema.NewSet(schema.HashString, flattenStringList(nc.AwsvpcConfiguration.SecurityGroups))
 	result["subnets"] = schema.NewSet(schema.HashString, flattenStringList(nc.AwsvpcConfiguration.Subnets))
+	result["assign_public_ip"] = "true"
+	if *nc.AwsvpcConfiguration.AssignPublicIp == "DISABLED" {
```
bflad: The AWS SDK provides constants we can use here instead of directly testing our own strings (this applies for all "DISABLED"/"ENABLED" in this PR), e.g. `ecs.AssignPublicIpDisabled`/`ecs.AssignPublicIpEnabled`

website/docs/r/ecs_service.html.markdown
```diff
@@ -102,6 +102,7 @@ Guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cluster-query
 
 * `subnets` - (Required) The subnets associated with the task or service.
 * `security_groups` - (Optional) The security groups associated with the task or service. If you do not specify a security group, the default security group for the VPC is used.
+* `assign_public_ip` - (Optional) Valid values are "ENABLED" or "DISABLED".  Will assign a public IP address to the ENI.
```
bflad: Documentation needs to be updated since the switch from string values to boolean values and should also note the default of `false`.

aws/resource_aws_ecs_service_test.go
```diff
@@ -1517,6 +1517,7 @@ resource "aws_ecs_service" "main" {
 	network_configuration {
 		security_groups = ["${aws_security_group.allow_all_a.id}", "${aws_security_group.allow_all_b.id}"]
 		subnets = ["${aws_subnet.main.*.id}"]
+		assign_public_ip = true
```
bflad: Can you please add an appropriate attribute test where this is used (and one in the other Fargate test where it should be the default of false)? e.g. `resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.assign_public_ip", "true")`

johnnorton:
@bflad @radeksimko Sorry I didn't get on this sooner.  Learning Go :) . Anyway please review and see if we can get this merged.

bflad:
I'll take another look in a few hours. Thanks for the updates and contributions while learning Go! 😃

austinkelleher:
Thanks for your work on this @johnnorton. Looking forward to seeing this merged.

aws/resource_aws_ecs_service.go
```diff
@@ -410,9 +415,14 @@ func flattenEcsNetworkConfigration(nc *ecs.NetworkConfiguration) []interface{} {
 	if nc == nil {
 		return nil
 	}
+
 	result := make(map[string]interface{})
 	result["security_groups"] = schema.NewSet(schema.HashString, flattenStringList(nc.AwsvpcConfiguration.SecurityGroups))
 	result["subnets"] = schema.NewSet(schema.HashString, flattenStringList(nc.AwsvpcConfiguration.Subnets))
+	result["assign_public_ip"] = "true"
+	if *nc.AwsvpcConfiguration.AssignPublicIp == ecs.AssignPublicIpDisabled {
```
bflad: If `nc.AwsvpcConfiguration.AssignPublicIp` is `nil` (its listed as optional in the [ECS API documentation](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AwsVpcConfiguration.html)), this can cause Terraform to crash. I can fix this (and clean it up a little since it will automatically default to `"false"`) prior to merge via:
```go
if nc.AwsvpcConfiguration.AssignPublicIp != nil {
	result["assign_public_ip"] = fmt.Sprintf("%v", *nc.AwsvpcConfiguration.AssignPublicIp == ecs.AssignPublicIpEnabled)
}
```

aws/resource_aws_lb_listener.go
```diff
@@ -34,7 +34,6 @@ func resourceAwsLbListener() *schema.Resource {
 			"load_balancer_arn": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
```
bflad: Unrelated change - will fix prior to merge

bflad:
@johnnorton thank you so much for your contributions here! I will be pulling in your commits, but I am running into some issues acceptance testing this appropriately, which isn't really from your PR. I will fix those on top of this as it requires some refactoring of the resource/testing itself.

I wound up down this rabbit hole because ECS will return this error if you try to enable `assign_public_ip` on EC2 launch types:
`InvalidParameterException: Assign public IP is not supported for this launch type.`

Unfortunately, as noted in #85, the error handling in this resource is already causing trouble as it is retrying on some conditions like the above when it shouldn't be. The launch type and assign_public_ip combination is something we can validate at plan time, but the implementation is a little more complex than just few lines. I think it'll be best if I fix both of these issues together as its already hindering my acceptance testing. More soon.

johnnorton:
@bflad Excellent.  Please let me know if I can help.

cemo:
@bflad milestone seems `v1.10.0`, it should be `v1.9.0`

bflad:
@cemo good catch, thanks and updated. Also, shout out to @johnnorton for working though the original PR.

bflad:

PR #2572, opened 2017-12-06
New resource: aws_dynamodb_backup 

DaveBlooman:
Creates resource for backing up DynamoDB tables.  This will not work in all regions right now, but will soon.

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -319,6 +319,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_dx_connection":                            resourceAwsDxConnection(),
 			"aws_dx_connection_association":                resourceAwsDxConnectionAssociation(),
 			"aws_dynamodb_table":                           resourceAwsDynamoDbTable(),
+			"aws_dynamodb_backup":                          resourceAwsDynamoDbBackup(),
 			"aws_ebs_snapshot":                             resourceAwsEbsSnapshot(),
 			"aws_ebs_volume":                               resourceAwsEbsVolume(),
 			"aws_ecr_lifecycle_policy":                     resourceAwsEcrLifecyclePolicy(),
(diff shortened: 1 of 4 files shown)


DaveBlooman:
Need some guidance on testing.  It appears AWS take some time to enable the backup feature on newly created tables. Im not sure if this is because the feature is new, or all new Dynamo tables are queued for backup to be enabled.  Either way, the test will fail as the backup feature is not available to be used with the new terraform resource.  Any ideas?

radeksimko:
Hi @DaveBlooman 
thanks for raising a PR.

I'm not sure on-demand backup is a good fit for Terraform. People should perform backups on more regular basis, rather than once in a while via Terraform, IMO.

Do you mind explaining how do you expect this resource to be used?

I'm aware we currently support `aws_ebs_snapshot` and `aws_db_snapshot` which have questionable use cases for the same reason.

What I'd imagine to be a good practice is scheduling backups e.g. via Lambda & CloudWatch Event Rules and use predictable table names (e.g. `backup-original-table-name-YYYY-MM-DD`), then possibly tune the existing [`aws_dynamodb_table` data source](https://www.terraform.io/docs/providers/aws/d/dynamodb_table.html) so we can find the most recent backup and finally add new field to [`aws_dynamodb_table` resource](https://www.terraform.io/docs/providers/aws/r/dynamodb_table.html) - say `restore_from_table` which would perform a one-off restoration at the creation time.

This is how I believe most users deal with backups in case of RDS at least - i.e. find the most recent snapshot via [`aws_db_snapshot` data source](https://www.terraform.io/docs/providers/aws/d/db_snapshot.html) and use it as [`snapshot_identifier`](https://www.terraform.io/docs/providers/aws/r/db_instance.html#snapshot_identifier).

That way the lifecycle of backups is safe and separate from the rest of otherwise potentially ephemeral infrastructure managed directly by Terraform.

radeksimko:
FYI: I'm going to close this PR in 1 month from now if it receives no further reaction. This is to allow us to focus on active PRs and avoid stale ones.

Thanks for understanding.

PR #2297, opened 2017-11-15
resource/aws_ssm_association: Allow for multiple targets

tom-henderson:
AWS SSM Association supports a maximum of 5 targets - http://docs.aws.amazon.com/systems-manager/latest/APIReference/API_CreateAssociation.html

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSSSMAssociation'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSSSMAssociation -timeout 120m
=== RUN   TestAccAWSSSMAssociation_basic
--- PASS: TestAccAWSSSMAssociation_basic (141.40s)
=== RUN   TestAccAWSSSMAssociation_withTargets
--- PASS: TestAccAWSSSMAssociation_withTargets (39.53s)
=== RUN   TestAccAWSSSMAssociation_withMultipleTargets
--- PASS: TestAccAWSSSMAssociation_withMultipleTargets (39.34s)
=== RUN   TestAccAWSSSMAssociation_withParameters
--- PASS: TestAccAWSSSMAssociation_withParameters (66.54s)
=== RUN   TestAccAWSSSMAssociation_withDocumentVersion
--- PASS: TestAccAWSSSMAssociation_withDocumentVersion (40.69s)
=== RUN   TestAccAWSSSMAssociation_withOutputLocation
--- PASS: TestAccAWSSSMAssociation_withOutputLocation (149.35s)
=== RUN   TestAccAWSSSMAssociation_withScheduleExpression
--- PASS: TestAccAWSSSMAssociation_withScheduleExpression (65.68s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	542.558s
```

PR #2574, opened 2017-12-06
aws_ssm_association: Improve documentation

ziggythehamster:
I made several improvements to this documentation page because it took me several tries to get a resource structure that actually worked when I came from a CloudFormation resource that worked. It also turns out that some of the documentation was actually not correct anyways (in particular, how `targets` are supposed to be specified - the original doc describes how the API wants it, not how Terraform wants it, and completely leaves out the proper format for tags).

This will conflict with #2297, which adds support for multiple targets, but I wanted to make this change against what is currently out there in case that doesn't land anytime soon. In other words, if you approve this, it should be fine to deploy whenever you feel like it.


RyanJarv:
Ran into this for the second time today. Would be nice to get the documentation updated so it reflects the current usage.

PR #2580, opened 2017-12-07
fix lb_target_group health check inconsistencies

eguven:
* path defaults to "/"  (which is what docs say) and validate function checks for "/" prefix (from AWS: "The destination path for health checks. This path must begin with a '/' character, and can be at most 1024 characters in length."
* timeout doc reflects the actual default (10) in code. AWS default is actually 5 but I did not want to change that and possibly break users' configuration.

When `path` is not provided the following error is produced:

    Error: Error applying plan:

    1 error(s) occurred:

    aws_alb_target_group.some-tg-name: Error creating LB Target Group: InvalidParameter: 1 validation error(s) found.
    - minimum field size of 1, CreateTargetGroupInput.HealthCheckPath.


aws/resource_aws_lb_target_group.go
```diff
@@ -132,6 +132,7 @@ func resourceAwsLbTargetGroup() *schema.Resource {
 						"path": {
 							Type:         schema.TypeString,
 							Optional:     true,
+							Default:      "/",
```
radeksimko: This was intentionally removed in order to bring support for network load balancers where path is irrelevant.

See https://github.com/terraform-providers/terraform-provider-aws/commit/79467dc32944ceb42acaa60bb91da86fa974ef22#diff-375aea487c27a6ada86edfd817ba2401

eguven:
@radeksimko Thanks for the response, done. I hadn't considered network load balancers, sorry.

bflad:

PR #2590, opened 2017-12-08
resource/aws_db_option_group: Add version field

chrisminton:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/748

Certain Oracle RDS options require a `version` attribute set, e.g. [OEM_AGENT](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Oracle.Options.OEMAgent.html). This change adds that attribute to the option.

The acceptance test is wrong since I'm not sure about how to get the correct option number (obviously not `option.0.version`) from state. If anyone can give me a steer it would be appreciated.

### DIFF
--- a/aws/resource_aws_db_option_group.go
+++ b/aws/resource_aws_db_option_group.go
@@ -107,6 +107,10 @@ func resourceAwsDbOptionGroup() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 							Set:      schema.HashString,
 						},
+						"version": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
 					},
 				},
 				Set: resourceAwsDbOptionHash,
@@ -353,6 +357,11 @@ func resourceAwsDbOptionHash(v interface{}) int {
 	for _, sgRaw := range m["db_security_group_memberships"].(*schema.Set).List() {
 		buf.WriteString(fmt.Sprintf("%s-", sgRaw.(string)))
 	}
+
+	if _, ok := m["version"]; ok {
+		buf.WriteString(fmt.Sprintf("%s-", m["version"].(string)))
+	}
+
 	return hashcode.String(buf.String())
 }
 
--- a/aws/resource_aws_db_option_group_test.go
+++ b/aws/resource_aws_db_option_group_test.go
@@ -262,6 +262,44 @@ func TestAccAWSDBOptionGroup_sqlServerOptionsUpdate(t *testing.T) {
 	})
 }
 
+func TestAccAWSDBOptionGroup_OracleOptionsUpdate(t *testing.T) {
+	var v rds.OptionGroup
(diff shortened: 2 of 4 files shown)


aws/resource_aws_db_option_group.go
```diff
@@ -107,6 +107,10 @@ func resourceAwsDbOptionGroup() *schema.Resource {
 							Elem:     &schema.Schema{Type: schema.TypeString},
 							Set:      schema.HashString,
 						},
+						"version": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
```
Ninir: Can this attribute be retrieved by the AWS API without a user setting it?
If so, this should also have a `Computed: true,` option

aws/resource_aws_db_option_group_test.go
```diff
@@ -262,6 +262,44 @@ func TestAccAWSDBOptionGroup_sqlServerOptionsUpdate(t *testing.T) {
...
+			{
+				Config: testAccAWSDBOptionGroupOracleEEOptionSettings(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSDBOptionGroupExists("aws_db_option_group.bar", &v),
+					resource.TestCheckResourceAttr(
+						"aws_db_option_group.bar", "name", rName),
+					resource.TestCheckResourceAttr(
+						"aws_db_option_group.bar", "option.#", "1"),
+					resource.TestCheckResourceAttr(
+						"aws_db_option_group.bar", "option.0.version", "12.1.0.4.v1"),
```
Ninir: As `option` is of type `TypeSet` (https://github.com/terraform-providers/terraform-provider-aws/pull/2590/files#diff-e61872fad819968684cae97c19506df5R70), a given entry can be retrieved with a hash, as in: https://github.com/chrisminton/terraform-provider-aws/blob/89c8a140afce2bc728b7e3e07c526f961c02f875/aws/resource_aws_db_option_group_test.go#L226

What I usually do is get the hash from the logs, using `TF_LOG=DEBUG` with `TF_LOG_PATH=terraform.log` as in:

    TF_LOG=DEBUG make testacc TEST=./aws TESTARGS='-run= TestAccAWSDBOptionGroup_multipleOptions'

This will put multiple log lines in the `aws/terraform.log` file, making it easier for you to debug :)

Note that the hash will change if parts of the `option` change.

Ninir:
Hi @chrisminton 

Do you think you would have time to fix the comments so that we get this merged?

Thanks!

chrisminton:
Hi @Ninir - yes! I've been caught up in other matters. It looks to me that the hash changes every time in the debug logs so this may not be a useful test but i'll test again.

nmische:
Is there any update on this issue? This is blocking us from being able to manage Oracle RDS with Terraform. Thanks.

chrisminton:
Thanks @bflad and sorry I didn't have any time to get back to this!

nmische:
Thanks for getting this merged!

PR #2595, opened 2017-12-08
documentation: remove antislashes in page titles

Ninir:
This cleans up the documentation so that we ensure next Pull Requests won't need review on this specific lines.

If this is too big in a single change, can still split it in multiple ones, but not sure it's worth it! 😄


website/docs/r/elastic_beanstalk_application.html.markdown
```diff
@@ -6,7 +6,7 @@ description: |-
   Provides an Elastic Beanstalk Application Resource
 ---
 
-# aws\_elastic\_beanstalk\_<wbr>application
+# aws_elastic_beanstalk_<wbr>application
```
bflad: Nitpick: While you're touching these lines, could you remove these `<wbr>` things too?

Ninir: Sure thing! didn't catch my 👀

Ninir:
Thanks folks! :)

PR #2623, opened 2017-12-11
Add more example and missing field

lologhi:
Because I had some trouble to find how to add "URL Query String Parameters" and a request validator.


lologhi:
Thanks !

rudresh-ajgaonkar:
I need to know how i can assign Validate body to the request validators option in the API Gateway? Is that possible?

PR #2632, opened 2017-12-11
resource/aws_elasticsearch_domain: Add support for encrypt_at_rest

tomelliff:
Adds encrypt at rest on ES domains.

Will close https://github.com/terraform-providers/terraform-provider-aws/issues/2591


tomelliff:
As mentioned, I think a second operation will show it wanting to rebuild the ES domain because it hasn't read the encrypt at rest options back from the API right now but the test is passing:
```
$ make testacc TEST=./aws TESTARGS="-run=TestAccAWSElasticSearchDomain_encrypt_at_rest"
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticSearchDomain_encrypt_at_rest -timeout 120m
=== RUN   TestAccAWSElasticSearchDomain_encrypt_at_rest
--- PASS: TestAccAWSElasticSearchDomain_encrypt_at_rest (886.48s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	886.493s
```
and I can see the domain is encrypted before it's deleted by the test:
```
$ aws es describe-elasticsearch-domains --domain-names tf-test-3881985383712683714 --region us-west-2
{
    "DomainStatusList": [
        {
            "ElasticsearchClusterConfig": {
                "DedicatedMasterEnabled": false, 
                "InstanceCount": 1, 
                "ZoneAwarenessEnabled": false, 
                "InstanceType": "m4.large.elasticsearch"
            }, 
            "Endpoint": "search-tf-test-3881985383712683714-tsewb5p4ajycmoo2jkqwlag4ru.us-west-2.es.amazonaws.com", 
            "Created": true, 
            "Deleted": true, 
            "DomainName": "tf-test-3881985383712683714", 
            "EBSOptions": {
                "VolumeSize": 10, 
                "VolumeType": "gp2", 
                "EBSEnabled": true
            }, 
            "SnapshotOptions": {
                "AutomatedSnapshotStartHour": 0
            }, 
            "DomainId": "123456789101/tf-test-3881985383712683714", 
            "AccessPolicies": "", 
            "Processing": true, 
            "AdvancedOptions": {
                "rest.action.multi.allow_explicit_index": "true"
            }, 
            "EncryptionAtRestOptions": {
                "KmsKeyId": "arn:aws:kms:us-west-2:123456789101:key/d454b888-53a9-4422-93d4-11d919a9b02c", 
                "Enabled": true
            }, 
            "ElasticsearchVersion": "6.0", 
            "ARN": "arn:aws:es:us-west-2:123456789101:domain/tf-test-3881985383712683714"
        }
    ]
}
```
I'll sort the read part tomorrow and see if I can add a test for it too.

As briefly mentioned in https://github.com/terraform-providers/terraform-provider-aws/pull/2548#issuecomment-350816972 should I be checking the attributes of the ES domain are what I expect? This is what I've done elsewhere but this resource doesn't seem to do that and I'm unsure why.

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -89,6 +89,24 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 					},
 				},
 			},
+			"encrypt_at_rest": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
```
Puneeth-n: What purpose does this `bool` serve here? I think the behavior should be:
- If `kms_key_id` is given, create the ES with the `kms_key_id`
- If `kms_key_id` is not given, create the ES without the `kms_key_id`
- When things change, delete and create the es domain

tomelliff: I was mirroring the API which takes those as options. I did consider setting it automatically when kms_key_id is set but it would prevent you from having Terraform encrypt it with the service KMS key that you get by default when using the console to create an encrypted domain.

The `aws_db_instance` resource has them as separate parameters to the resource (rather than an options block) and defaults to using the service KMS key if kms_key_id isn't specified but that's largely because the API looks that way.

Puneeth-n: Yea true. Does AWS API create a default key if only `enable` is passed? If so, then having `enable` makes sense. If not, I am not sure whether kms key creation is within the scope of this resource.

For example, when adding triggers to lambda functions via aws console, aws does a lot of stuff in the background like creating IAM roles and adding policies and stuff. However in Terraform we don't do it and all the resources required to trigger a lambda function must be created.

tomelliff: Yeah looks like it does, added a test that runs without specifying the key id and that works nicely without needing to do anything extra which is neat.

Puneeth-n: Lovely! :)

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -291,6 +309,21 @@ func resourceAwsElasticSearchDomainCreate(d *schema.ResourceData, meta interface
 		}
 	}
 
+	if v, ok := d.GetOk("encrypt_at_rest"); ok {
+		options := v.([]interface{})
+
+		if len(options) > 1 {
+			return fmt.Errorf("Only a single encrypt_at_rest block is expected")
```
Puneeth-n: I think we should also validate the instance types here since, like you mentioned, not all instance types support encryption at rest. I saw a similar approach in `t2.unlimited` [PR](https://github.com/terraform-providers/terraform-provider-aws/pull/2619/files#diff-28aade55afa0cda57279a9d35d523488R1043). I thought it was 🆒  :)

tomelliff:
I'd guess it's a spurious failure in https://travis-ci.org/terraform-providers/terraform-provider-aws/builds/315313868 but it's showing:
```
The command "make vendor-status" exited with 0.
3.90s$ make vet
go vet .
aws/resource_aws_elasticsearch_domain_test.go:384: arg *conf.Enabled for printf verb %d of wrong type: bool
exit status 1
Vet found suspicious constructs. Please check the reported constructs
and fix them if necessary before submitting the code for review.
make: *** [vet] Error 1
The command "make vet" exited with 2.
```
when I haven't touched any vendored packages.

I'm not done with the PR yet so it will re-run on future pushes but is this something we should be aware of?

aws/resource_aws_elasticsearch_domain.go
tomelliff: I don't think that will run at plan time though so it's just returning a different error to what you'll get from the AWS API on apply time. You already get the following error on apply (and immediately):
```
Error: Error applying plan:

1 error(s) occurred:

* aws_elasticsearch_domain.elasticsearch: 1 error(s) occurred:

* aws_elasticsearch_domain.elasticsearch: InvalidTypeException: Encryption at rest is not supported with t2.small.elasticsearch instances
	status code: 409, request id: 085a21b5-df33-11e7-babb-3f32bfcf1d9b

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```
From memory the validatefuncs in the schema can't validate different parameters to a resource so we can't put it there unfortunately to give us a plan time failure.

tomelliff:
Updated docs.

Building the provider locally it seems to work as is although I'm wary that there's nothing to read the encryption config from the API right now and trying to work out if we actually need to cover this considering changing this forces a recreation.

Right now when you build an encrypted or unencrypted ES domain the state has the encrypt_at_rest config already so it's enough to know if you change this that it needs to recreate the domain. The only thing I can think of is someone could Terraform an unencrypted ES domain, someone could then delete the existing domain and recreate it as encrypted outside of Terraform and then update Terraform to want an encrypted domain and Terraform would rebuild the ES domain needlessly.

This seems like an odd thing to do but maybe it's worth reading the config from the API just to cover it anyway? Are there any other things that could cause the lack of reading the encrypted config to do bad things?

tomelliff:
Okay, think this is done now. Added code to read the KMS key from the AWS API. I noticed that the API takes a KMS key id in short form but reading the ES domain config on the API always returns the full API so I suppress the diff in that case.

tomelliff:
Tests passing locally:
```
$ make testacc TEST=./aws TESTARGS="-run=TestAccAWSElasticSearchDomain_encrypt_at_rest_"
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticSearchDomain_encrypt_at_rest_ -timeout 120m
=== RUN   TestAccAWSElasticSearchDomain_encrypt_at_rest_default_key
--- PASS: TestAccAWSElasticSearchDomain_encrypt_at_rest_default_key (886.10s)
=== RUN   TestAccAWSElasticSearchDomain_encrypt_at_rest_specify_key
--- PASS: TestAccAWSElasticSearchDomain_encrypt_at_rest_specify_key (1084.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1970.795s
```

aws/resource_aws_elasticsearch_domain.go
handlerbot: (Also, hardcoding what instance types support encryption in the provider means that the list has to be re-synchronized if it changes, and people need to wait for a new release of the provider, or you have to give them an override command-line flag... Too much complexity for little benefit, IMO!)

Quentin-M:
+1 Thanks a bunch.

evhen14:
Hey guys, when do you think it's going to be merged?

sonarjames:
Can this be merged now? Would love to make use of this to rebuild my ElasticSearch with encryption

vseguin:
+1 for merge - would use it asap in our terraform

billcrook:
> +1 for merge - would use it asap in our terraform

ditto

gundam-zeta:
+1

Sabotaz:
+1

semka:
Can't wait to use it

billcrook:
> Can't wait to use it

Me either. No, literally - we manually created the cluster until this is ready. Please please please let's get this in!

rerolled:
merge plz

FireballDWF:
+1 merge please

tomelliff:
Just seen the merge conflicts but I'll hang off on fixing them until there's been a review on this if that's okay?

Also, for those not aware but wanting to add their :+1: to this pull request, Github allows you to add a reaction to any specific post by clicking the +(smiley face) in the top right of the post which shows everyone that you want this feature (or disagree with a :-1: or equivalent) without spamming everyone subscribed to the pull request with emails.

MichaelYak:
+1 Can the feature be merged soon?

aws/resource_aws_elasticsearch_domain.go
radeksimko: I agree with @Puneeth-n here that we should move validation to the schema here.
@tomelliff you're also right that `ValidateFunc` can't help in this context, but we can use `MaxItems: 1` which should do the job. 😉

aws/structure.go
```diff
@@ -1045,6 +1045,32 @@ func expandESEBSOptions(m map[string]interface{}) *elasticsearch.EBSOptions {
 	return &options
 }
 
+func flattenESEncryptAtRestOptions(o *elasticsearch.EncryptionAtRestOptions) []map[string]interface{} {
+	m := map[string]interface{}{}
```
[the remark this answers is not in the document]
tomelliff: Not sure I follow that, what would cause the crash? If `o` is nil doesn't it already return an empty `[]map[string]interface{}{}` due to the nil checks on `o.Enabled` and `o.KmsKeyId`?

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"strings"
```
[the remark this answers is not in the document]
tomelliff: Aha, didn't spot that `gofmt` and/or Goland's import optimiser doesn't do this for you. Seems odd that `goimports` seems to add newlines between stdlib stuff when adding it and then `gofmt` won't rectify that either which is annoying considering `gofmt` is so good otherwise.

aws/resource_aws_elasticsearch_domain.go
tomelliff: That's nice, means I can remove some logic there. Is it worth doing the same to some of the other blocks that are using that other style? Or (if you do want it changing) would you prefer that in a separate PR?

aws/structure.go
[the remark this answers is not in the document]
radeksimko: > If o is nil doesn't it already return an empty []map[string]interface{}{} due to the nil checks on o.Enabled and o.KmsKeyId?

No, it will crash on line 1052 in such case because that condition is trying to access a field of a struct which isn't struct, but `nil`. In other words that condition is one step ahead.

See https://play.golang.org/p/R07zPiMT92M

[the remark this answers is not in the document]
tomelliff: Right, sorry, completely missed that for some reason. Fixed now, thanks for the review.

tomelliff:
Are you planning on squashing this PR or would you prefer me to rebase things to clean up the commit history and the merge?

aws/structure.go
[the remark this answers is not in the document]
radeksimko: No worries 😸  That's why we do reviews.

radeksimko:
> Are you planning on squashing this PR or would you prefer me to rebase things to clean up the commit history and the merge?

I'm going to do that as part of merging this PR, so no worries.

bflad:

tehmaspc:
danke!

rawatProgrammer:
@tomelliff above pasted AWS cloud formation is not working. Highlighted field KMS encryption is also not in AWS documentation . 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html#cfn-elasticsearch-domain-vpcoptions

2018_02_03_18_41_23_resource_aws_elasticsearch_domain_add_support_for_encrypt_at_rest_by_tomelliff_

Do we have other option to solve this cloud formation issue ?

tomelliff:
What Cloudformation issue? This added support for it to Terraform, not Cloudformation.

PR #2638, opened 2017-12-12
docs/elasticache_replication_group: Add missing fields

radeksimko:
Forgotten in https://github.com/terraform-providers/terraform-provider-aws/pull/2090

### DIFF
--- a/website/docs/r/elasticache_replication_group.html.markdown
+++ b/website/docs/r/elasticache_replication_group.html.markdown
@@ -63,6 +63,10 @@ The following arguments are supported:
 * `automatic_failover_enabled` - (Optional) Specifies whether a read-only replica will be automatically promoted to read/write primary if the existing primary fails. Defaults to `false`.
 * `auto_minor_version_upgrade` - (Optional) Specifies whether a minor engine upgrades will be applied automatically to the underlying Cache Cluster instances during the maintenance window. Defaults to `true`.
 * `availability_zones` - (Optional) A list of EC2 availability zones in which the replication group's cache clusters will be created. The order of the availability zones in the list is not important.
+* `engine` - (Optional) The name of the cache engine to be used for the clusters in this replication group. e.g. `redis`
+* `at_rest_encryption_enabled` - (Optional) Whether to enable encryption at rest.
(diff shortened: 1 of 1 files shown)


website/docs/r/elasticache_replication_group.html.markdown
```diff
@@ -63,6 +63,10 @@ The following arguments are supported:
 * `automatic_failover_enabled` - (Optional) Specifies whether a read-only replica will be automatically promoted to read/write primary if the existing primary fails. Defaults to `false`.
 * `auto_minor_version_upgrade` - (Optional) Specifies whether a minor engine upgrades will be applied automatically to the underlying Cache Cluster instances during the maintenance window. Defaults to `true`.
 * `availability_zones` - (Optional) A list of EC2 availability zones in which the replication group's cache clusters will be created. The order of the availability zones in the list is not important.
+* `engine` - (Optional) The name of the cache engine to be used for the clusters in this replication group. e.g. `redis`
+* `at_rest_encryption_enabled` - (Optional) Whether to enable encryption at rest.
+* `transit_encryption_enabled` - (Optional) Whether to enable encryption in transit.
+* `auth_token` - (Optional) The password used to access a password protected server. Can be specified only if `transit_encryption_enabled = true`.
```
tombuildsstuff: >  `auth_token` - (Optional) The password used to access a password protected server.

feels like that’d be better as “The auth token used to access an encrypted server”?

radeksimko: In the context of Redis (which is currently the only engine supporting this) it's called `password`: https://redis.io/commands/auth

AFAIK the transit encryption is unrelated to the auth token - you can encrypt it whether or not you use `auth_token`.

I assume the encryption in transit was added by AWS so that the password isn't transmitted in plaintext over the wire.

btw. the description matches description from the API reference: http://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html

tombuildsstuff: 👍 seems sensible

PR #2649, opened 2017-12-13
[WIP] New Resource: LB Listener Additional Certificates

oarmstrong:
This PR is for #1853 and is a second attempt of the PR #2203.

We create a new resource `aws_lb_listener_certificate` which is used to add additional certificates into an HTTPS load balancer listener. This does not effect the `certificate_arn` argument to `aws_lb_listener` as there must still be a default certificate assigned to the listener.

I need help with the acceptance tests currently. They currently fail in the destroy phase as I believe the listener resource is being destroyed before the listener certificate resource and therefore the listener cannot be found. This could fairly easily be worked around by allowing a `ListenerNotFound` error in the destroy test but I don't believe that'd be testing the destruction of this specific resource correctly. Maybe @radeksimko can be of assistance here (as he reviewed my previous PR).

Other than that failing test this is ready to go IMO.

### DIFF
--- a/aws/resource_aws_lb_listener_certificate.go
+++ b/aws/resource_aws_lb_listener_certificate.go
@@ -1,0 +1,119 @@
+package aws
+
+import (
+	"errors"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsLbListenerCertificate() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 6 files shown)


aws/resource_aws_lb_listener_certificate.go
```diff
@@ -0,0 +1,119 @@
...
+			if *cert.IsDefault {
+				continue
+			}
+
+			if *cert.CertificateArn == d.Get("certificate_arn").(string) {
+				found = true
+			}
+		}
+
+		if *resp.NextMarker != "" {
```
Puneeth-n: `if resp.NextMarker != nil {`

Puneeth-n: Otherwise causes segfault

oarmstrong:
Closing in favour of #2686.

PR #2652, opened 2017-12-13
Changing dependency manager from govendor to dep

stack72:
it's just a simple command that converts it
`dep init`
it walks the govendor json file and works out the specific references / tags etc


stack72:
I am happy to close this - but saying dep is slower than govendor is not correct at all IMO

PR #2676, opened 2017-12-15
r/lb_target_group: Use default(200) when matcher isn't specified

atsushi-ishibashi:
#2653


atsushi-ishibashi:
@tombuildsstuff Not enhancement, bug fix

atsushi-ishibashi:
The problem was already fixed by #2380

PR #2680, opened 2017-12-15
Data Sources cleaning

Ninir:
This:
* Removes redundant type declarations in composite literals
* Adds debug lines about what is happening, normalising texts over files

Context: Data Sources

### DIFF
--- a/aws/data_source_aws_ecs_container_definition.go
+++ b/aws/data_source_aws_ecs_container_definition.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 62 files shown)


aws/data_source_aws_ecs_container_definition.go
```diff
@@ -66,9 +67,11 @@ func dataSourceAwsEcsContainerDefinition() *schema.Resource {
 func dataSourceAwsEcsContainerDefinitionRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ecsconn
 
-	desc, err := conn.DescribeTaskDefinition(&ecs.DescribeTaskDefinitionInput{
+	params := &ecs.DescribeTaskDefinitionInput{
 		TaskDefinition: aws.String(d.Get("task_definition").(string)),
-	})
+	}
+	log.Printf("[DEBUG] Reading ECS Container Definition: %#v", params)
```
catsby: FWIW the AWS SDK struct implement `Stringer()`, so you can use `%s` here instead of `%#v`

Ninir: @catsby Made the change, thanks for letting me know! :)

PR #2683, opened 2017-12-15
Replace ARN string creation of the form 'fmt.Sprintf("arn:...")' with AWS ARN structure

ewbankkit:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/1783.
Replaces https://github.com/terraform-providers/terraform-provider-aws/pull/1897.

Ended up with a less aggressive approach of using the `arn.ARN` structure directly and not wrapping.


ewbankkit:
Rebased with changes from https://github.com/terraform-providers/terraform-provider-aws/pull/2680.

PR #2684, opened 2017-12-16
resource/aws_lambda_function: Make it possible to remove dead_letter_config

stefansundin:
Make it possible to remove `dead_letter_config` by removing the block.

## Issue

Currently there is an issue removing `dead_letter_config` from a lambda function. The current work around is to:
1. Set `target_arn = ""`
2. Apply.
3. Remove `dead_letter_config` entirely.

If you don't do step 3, then you will get the following diff on subsequent plans:
```
  ~ aws_lambda_function.func
      dead_letter_config.#: "0" => "1"
```

With this change, you can just remove the whole `dead_letter_config` block and then apply.

## Tests

I tried looking into writing tests to cover this, but I don't have time to get up to speed on how to do that correctly right now. Feel free to add commits to this branch (edits from maintainers are allowed).

Thanks!!

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -565,13 +565,14 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
 	}
 	if d.HasChange("dead_letter_config") {
 		dlcMaps := d.Get("dead_letter_config").([]interface{})
+		configReq.DeadLetterConfig = &lambda.DeadLetterConfig{
+			TargetArn: aws.String(""),
+		}
 		if len(dlcMaps) == 1 { // Schema guarantees either 0 or 1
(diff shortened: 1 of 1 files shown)


jen20:
Hi @stefansundin! Thanks for opening a pull request here - the change looks valid to me. You're right that it probably does want test coverage - this one should be fairly easy to test: we need to create a lambda function with a DLQ configured, then remove it in a second test step and ensure that there is no diff the second apply.

bflad:
@stefansundin will you have a chance to implement the acceptance testing for this or would you like one of the maintainers to take this across the finish line? Thanks!

stefansundin:
Hi @bflad. I haven't had the time to learn about the acceptance tests yet. If you'd like to, then feel free to take it across the finish line. Thanks!

stefansundin:
Thanks @bflad :)

bflad:

PR #1476, opened 2017-08-23
resource/aws_ecs_service: Change type of placement_strategy to TypeList

atsushi-ishibashi:
#1466

PR #2685, opened 2017-12-16
[WIP, NeedHelp] resource/aws_elasticache_cluster: Change availability_zones from Set to List

atsushi-ishibashi:
Fix: #2658 
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticacheCluster_multiAZInVpc -timeout 120m
=== RUN   TestAccAWSElasticacheCluster_multiAZInVpc
--- PASS: TestAccAWSElasticacheCluster_multiAZInVpc (1039.93s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1039.959s
```
・TODO
- [ ] Sort list to avoid update
- [ ] State migration because this will cause breaking-change.

### DIFF
--- a/aws/resource_aws_elasticache_cluster_test.go
+++ b/aws/resource_aws_elasticache_cluster_test.go
@@ -514,7 +514,7 @@ resource "aws_elasticache_cluster" "bar" {
     cluster_id = "tf-%s"
     engine = "memcached"
     node_type = "cache.m1.small"
-    num_cache_nodes = 2
+    num_cache_nodes = 3
     port = 11211
     subnet_group_name = "${aws_elasticache_subnet_group.bar.name}"
     security_group_ids = ["${aws_security_group.bar.id}"]
@@ -522,7 +522,8 @@ resource "aws_elasticache_cluster" "bar" {
     az_mode = "cross-az"
     availability_zones = [
         "us-west-2a",
-        "us-west-2b"
+        "us-west-2b",
+        "us-west-2a"
     ]
 }
 `, acctest.RandInt(), acctest.RandInt(), acctest.RandInt(), acctest.RandInt(), acctest.RandString(10))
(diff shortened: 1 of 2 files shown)


aws/resource_aws_elasticache_cluster_test.go
```diff
@@ -514,15 +514,16 @@ resource "aws_elasticache_cluster" "bar" {
...
+    num_cache_nodes = 3
     port = 11211
     subnet_group_name = "${aws_elasticache_subnet_group.bar.name}"
     security_group_ids = ["${aws_security_group.bar.id}"]
     parameter_group_name = "default.memcached1.4"
     az_mode = "cross-az"
     availability_zones = [
         "us-west-2a",
-        "us-west-2b"
+        "us-west-2b",
```
cep21: The order does not matter.  Does Terraform have a type where it sorts the list first before comparing for changes (for example, if this list changes to "2a 2b 2a" from "2a 2a 2b", that's not actually something that needs to be updated.

atsushi-ishibashi: @cep21 Sorting list is good idea to avoid not necessary update. Thank you!

atsushi-ishibashi:
Do you have any idea to sort list before set, like `StateFunc`?

bflad:
@atsushi-ishibashi thanks for your work here and apologies you got stuck in the same spot here as #1476! 😄  In a similar manner to #1476, I'm going to close this PR in preference of a migration path similar to #4117. For other folks, there is a much longer explanation in that PR. I will make a note to personally get an updated PR in for this resource in v1.17.0 or shortly after.

bflad:
The replacement PR here, #4741, was merged and will release with version 1.23.0 of the AWS provider. It deprecates `availability_zones` (TypeSet) in favor of a new attribute `preferred_availability_zones` (TypeList).

PR #2720, opened 2017-12-20
r/aws_volume_attachment: make sure we properly update the state

svanharmelen:
Without taking the device name into account, certain changes (were the volume ID and instance ID remain the same) will not work as expected.

In those cases you will not be able to get rid of the old resource in your state file, without manual actions.


bflad:

PR #2724, opened 2017-12-20
Fix the elasticache documentation about the endpoints

gopisaba:
The current documentation for Elasticache attributes looks wrong. The `configuration_endpoint_address` is available for Memcache and Clustered Redis and the `primary_endpoint_address` is available for Standalone Redis only

- Reference: http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Endpoints.html


jen20:
This agrees with my reading of the AWS SDK documentation. Thanks @gopisaba!

PR #2725, opened 2017-12-20
Use s3 bucket "id" attribute for "aws_s3_bucket_object" docs

charlieduong94:
Reasoning for docs update:

Looks like the `s3_bucket_object` docs use the `bucket` attribute from the `aws_s3_bucket` resource. While it does look like the `bucket` attribute is sometimes exported (if I read the code correctly), it's not documented on the `aws_s3_bucket` page. Also, other pages use the `id` attribute, so this change might be good for the sake of consistency.


jen20:
Thanks @charlieduong94! `bucket` was the original name, and is still used in some older examples and older Terraform code. `id` is probably preferable, as you point out.

PR #2728, opened 2017-12-20
Add RDS Restore Option

jasonmcintosh:
Adds support for creating a DB Instance from an S3 xtrabackup snapshot
https://github.com/terraform-providers/terraform-provider-aws/issues/2410
First attempt to do go, so certain points are not entirely ideal.  Also would love to rewrite the restore and create so many of the required options are set in a function vs. duplicated, but that can be refactored later


jasonmcintosh:
Been a month... any updates?

bflad:
Hi @jasonmcintosh sorry for the maintainer delay in getting this reviewed for you. I don't have time to give this a fully exhaustive review right now but skimming the acceptance test, is it possible to actually test and have this succeed currently? It looks like it would be trying to pull the S3 backup from a random bucket and not succeed. Does AWS have a sample S3 backup that might work with this? Otherwise, we might need to get real creative.

jasonmcintosh:
XtraBackup database - we can create and share a sample file.  When I was testing in the provider, we only did unit tests vs. integration tests.  Then, we built the provider and did manual integration tests.  If you want an automatic integration test I can look at adding that support.  It'll be trickier but doable - just have to create a random bucket, a bucket item from the sample file, and then run the integration tests and tear down.  We'd need to store the xtrabackup file in the tests but thankfully it's a tiny binary file.

jasonmcintosh:
Bringing in master changes (which look like really one file) to validate/test again.

jasonmcintosh:
@bflad Note, this COULD be a misunderstanding on our part - we're pretty new to Go (some of us are).  Were these acceptance tests and we just missed something on this?

bflad:
Sorry our terminology for "acceptance" testing is equivalent to "integration" testing. It expects to fully test a Terraform configuration against the AWS APIs and have it succeed end-to-end. In order to accomplish that, it looks like we'll need to have the test perform [quite a few steps according to the documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/MySQL.Procedural.Importing.html#MySQL.Procedural.Importing.Backup) including setting up an IAM role that RDS can assume and having a S3 location ready to go with a valid backup file in an automated fashion. 

As for handling a valid backup file in S3, I can think of a few automated options:
* Create all the infrastructure each test run to create a valid backup file in S3 such as creating an EC2 instance running MySQL that runs Percona XtraBackup and uploads to S3 each time (very brittle and time consuming)
* Find and use a publicly accessible backup file for testing (potentially brittle)
* Create a backup file for testing, provide to a maintainer here, we upload it to a known and managed location in S3 that is publicly accessible, and the acceptance test is updated to point there
* Create a backup file for testing, add this file to the repository, then update the acceptance test to create a bucket and upload a backup file each time

But similar to what you said previously, we're probably going to want to go with one of the last two options, but I think it really depends on the size of the backup file. We probably don't want to include something in the repository over a few MB.

So short story here is that at the very least, the acceptance test for this feature will need to handle the IAM role creation, then we need to figure out the S3 backup file handling.

If you don't have time or interest in doing this, that's perfectly okay too 👍 . It just means this PR will be delayed as this functionality is not currently a priority (that I know about) for any of the Hashicorp employed maintainers.

jasonmcintosh:
We're happy to add this support  :)  Looking at options 3 & 4 right now - we're thinking option #4 should be viable (creating a test file now, getting it as small as we can).  On role... looks like there's another test that creates a monitoring role.  Can see if I can't re-use some of that.  Good feedback/information - thanks!  It may be early next week before we can get to this though.  We can run the provider locally for now to get us moving while we add support.

jasonmcintosh:
@bflad  We have a 1.3MB tiny backup with a test DB.  Do you want it in git (it's a .tar.gz file) or an email so you can create a public S3 location?  Working on the tests shortly :)  Note, we'll be looking to add the same functionality to aurora - do you want those changes in the same pull request or separate?

bflad:
@jasonmcintosh for now can you email it to my Github username at hashicorp dot com? I'll discuss how we want to handle this internally.

Separate pull request would be preferred. Thanks!

jasonmcintosh:
@bflad We've got acceptance tests quasi-working.  It creates the database from an s3 restore.  However, we've got a problem where IAM role creation is required for this to work, and the IAM role doesn't seem to be fully propagated before it's used.  Do you have any thoughts on solutions around this?  e.g. how to add a sleep into the code to make sure the IAM role is there first?  We tried a depends_on block similar to the way the enhanced monitoring RDS role works, but that doesn't seem to work in the case of a restore.  As we understand it, Step blocks in acceptance tests are destroyed before the next steps, so we can't do a Step "Pre-setup resources", Step "Sleep", Step "Create S3"?

bflad:
Inside the `resource.Retry()` that wraps `conn.RestoreDBInstanceFromS3()`, add a conditional that allows it to retry based on the error coming back. You didn't provide exactly what you're seeing but something like this should work:

```go
		err = resource.Retry(5*time.Minute, func() *resource.RetryError {
			_, err = conn.RestoreDBInstanceFromS3(&opts)
			if err != nil {
				if isAWSErr(err, "InvalidParameterValue", "... IAM role message here ...") {
				}
				if isAWSErr(err, "InvalidParameterValue", "ENHANCED_MONITORING") {
				}
			}
			return nil
		})
```

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
 				},
 			},
 
+			"s3_import": {
+				Type:     schema.TypeSet,
```
bflad: For simple nested attributes with `MaxItems: 1` we prefer `schema.TypeList` 👍

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
...
+				ConflictsWith: []string{
+					"snapshot_identifier",
+					"replicate_source_db",
+				},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"bucket_name": {
+							Type:     schema.TypeString,
+							Required: true,
+							Optional: false,
```
bflad: `Required: true` is defined, so `Optional: false` is extraneous

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
...
+							ForceNew: true,
+						},
+						"bucket_prefix": {
+							Type:     schema.TypeString,
+							Required: false,
```
bflad: `Optional: true` is defined, so `Required: false` is extraneous

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
...
+							Optional: true,
+							Default:  "",
```
bflad: The zero value for `schema.TypeString` is `""` so `Default: ""` is extraneous

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
 				},
 			},
 
+			"s3_import": {
```
bflad: We should add `ForceNew: true` to all child attributes of `s3_import` as there is no "update" available for it. 😄

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
...
+						"bucket_prefix": {
+							Type:     schema.TypeString,
+							Required: false,
+							Optional: true,
+							Default:  "",
+						},
+						"ingestion_role": {
+							Type:     schema.TypeString,
+							Required: true,
+							Optional: false,
```
bflad: `Required: true` is defined, so `Optional: false` is extraneous

aws/resource_aws_db_instance.go
```diff
@@ -220,6 +220,49 @@ func resourceAwsDbInstance() *schema.Resource {
...
+						},
+						"source_engine": {
+							Type:     schema.TypeString,
+							Required: false,
+							Optional: true,
+							Default:  "mysql",
```
bflad: If/when this is supported for other database engines, this will become awkward with having engine and engine version defaults. I think we should set both these two `Required: true` (like the API), remove the `Default:`s and document the valid values for each attribute for now.

aws/resource_aws_db_instance.go
```diff
@@ -451,6 +494,175 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
...
+		if attr, ok := d.GetOk("iam_database_authentication_enabled"); ok {
+			opts.EnableIAMDatabaseAuthentication = aws.Bool(attr.(bool))
+		}
+
+		log.Printf("[DEBUG] DB Instance S3 Restore configuration: %#v", opts)
+		var err error
+		err = resource.Retry(5*time.Minute, func() *resource.RetryError {
+			_, err = conn.RestoreDBInstanceFromS3(&opts)
+			if err != nil {
+				if awsErr, ok := err.(awserr.Error); ok {
```
bflad: This can be simplified with:

```
					if isAWSErr(err, "InvalidParameterValue", "ENHANCED_MONITORING") {
						return resource.RetryableError(awsErr)
					}
					if isAWSErr(err, "InvalidParameterValue", "S3_SNAPSHOT_INGESTION") {
					}
					if isAWSErr(err, "InvalidParameterValue", "S3 bucket cannot be found") {
					}
```

aws/resource_aws_db_instance_test.go
```diff
@@ -316,6 +316,30 @@ func TestAccAWSDBInstance_snapshot(t *testing.T) {
 	})
 }
 
+func TestAccAWSDBInstance_s3(t *testing.T) {
+	var snap rds.DBInstance
+	bucket := acctest.RandString(5)
```
bflad: Can you please switch these to use one of the following? Thanks!

```
acctest.RandomWithPrefix("tf-acc-test")
fmt.Sprintf("tf-acc-test-%s", acctest.RandString(5))
```

aws/resource_aws_db_instance_test.go
```diff
@@ -1046,6 +1070,135 @@ resource "aws_db_instance" "snapshot" {
...
+	return fmt.Sprintf(`
+
+resource "aws_s3_bucket" "xtrabackup" {
+  bucket = "%s"
+}
+
+resource "aws_s3_bucket_object" "xtrabackup_db" {
+  bucket = "${aws_s3_bucket.xtrabackup.id}"
+  key    = "%s/sample.tar.gz"
+  source = "../files/2018-02-26_19-31-02.tar.gz"
```
bflad: Can we please rename the backup file to something meaningful? e.g. `mysql-5-6-xtrabackup.tar.gz`

website/docs/r/db_instance.html.markdown
```diff
@@ -175,11 +175,31 @@ for more information.
...
 associate.
+* `s3_import` - (Optional) Restore from a Percona Xtrabackup in S3.  See [Importing Data into an Amazon RDS MySQL DB Instance](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/MySQL.Procedural.Importing.html)
 
 ~> **NOTE:** Removing the `replicate_source_db` attribute from an existing RDS
 Replicate database managed by Terraform will promote the database to a fully
 standalone database.
 
+### S3 Import Options
+Full details on the core parameters and impacts are in the API Docs: [RestoreDBInstanceFromS3](http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBInstanceFromS3.html).  Sample 
+```hcl
```
bflad: I think this requires an extra newline above or the formatting will not work -- it can be verified using `make website` (available very recently in master)

aws/resource_aws_db_instance_test.go
```diff
@@ -1046,6 +1070,135 @@ resource "aws_db_instance" "snapshot" {
...
+      },
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_iam_policy" "test" {
+  name   = "tf-s3-rds-access-policy"
```
bflad: This needs to be randomized with `-%s` in case of test failures/concurrency. 👍

aws/resource_aws_db_instance_test.go
```diff
@@ -1046,6 +1070,135 @@ resource "aws_db_instance" "snapshot" {
...
+	cidr_block = "10.1.2.0/24"
+	availability_zone = "us-west-2b"
+	vpc_id = "${aws_vpc.foo.id}"
+	tags {
+		Name = "tf-acc-db-instance-with-subnet-group-2"
+	}
+}
+
+resource "aws_db_subnet_group" "foo" {
+	name = "foo"
```
bflad: This needs to be randomized with `-%s` in case of test failures/concurrency. 👍

aws/resource_aws_db_instance.go
jasonmcintosh: Added this, though I could see long term an interesting debate.  if it's use as a spring-board to create a database, should be able to remove the backup file after the database is created.  SO in theory, deleting the block shouldn't cause a rebuild/recreate.  But I guess that can be done with ignore attributes block instead...

jasonmcintosh:
Thanks a ton @bflad will have one for Aurora soon.  I'm HOPING in some spare time I can work on refactoring some of that codebase.  There's a lot of duplication there that I think could be cleaned up.

bflad:

PR #2734, opened 2017-12-21
Fix FARGATE ECS task definition doc

### LINKED ISSUE #2732: Documentation is confusing for Fargate Support
On the docs here: https://www.terraform.io/docs/providers/aws/r/ecs_task_definition.html, the value of *launch_types* is spoken about as if it were a primary key for the object itself. However, there is no documentation around what a *launch_type* is. It looks like it's actually referring to the *requires_compatibilities* field?

**Examples**:

cpu - (Optional) The number of cpu units used by the task. If the *launch_type* is FARGATE this field is required.

memory - (Optional) The amount (in MiB) of memory used by the task. If the *launch_type* is FARGATE this field is required.

requires_compatibilities - (Optional) A set of launch types required by the task. The valid values are EC2 and FARGATE.

Can we update the documentation to better point this out? Or can we rename requires_compatibilities field to better reflect it as *launch_type*?

Puneeth-n:
Fixes #2732


jen20:
:+1: This matches the AWS documentation.

jen20:
Thanks @Puneeth-n!

PR #2753, opened 2017-12-22
r/aws_sns_topic: Add support of prefixes

ColinHebert:


bflad:

PR #2754, opened 2017-12-22
r/aws_db_event_subscription: Add support of prefixes

ColinHebert:


bflad:

PR #2759, opened 2017-12-23
[WIP] Set cloudwatch log metric filter default_value to nil as default

jocelynthode:
* Improve Acceptance testing of default_value


jocelynthode:
I am not sure to understand why when the point DefaultValue in struct MetricTransformation is set to "nil" it defaults to 0 afterwards. Any idea ?

jocelynthode:
Any help on debugging this issue @Ninir ?

92kz1h:
Is there any news on getting this PR merged? This is making CloudWatch Metrics useless for anybody who tries to calculate the average of metric events.

jocelynthode:
I'm still waiting on help on this issue as I encountered the problem with nil being converted to 0.

ryanpark91:
:+1:

macmacs:
It would be a great relief to have this fixed as we intend to use about 100 metric filters at a time and don't want to correct it by hand.

sbmarcos:
ETA on this? Thanks much!

ahl:
@jocelynthode you're hitting a variant of this:
https://github.com/hashicorp/terraform/issues/16796

I'd suggest our options are to either:
1. add a new field to Schema that indicates that optional, non-present fields should be omitted from the final map

2. change the resource definition to nest the default value in an optional structure (that therefore may have the value nil):
```
  default {
    value = 17
  }
```

3. Use a sentinel value e.g. NaN


1. is probably a dead end given how long 16796 has be idle
2. is pretty ugly pushes the code ugliness to the user
3. has the downside of showing the user the NaN in the computed configuration

I've prototyped all of them; I'd suggest that 3 is the simplest both for the user and in terms of implementation; it's also an easy migration if/when 16796 is ever actually fixed.

I'll submit my own PR if I don't hear soon. I don't want to step on your toes, but you might have reasonably given up after almost 2 years...

Issue #97, opened 2017-06-13
Fail during plan if user_data on aws_instance is > 16KB


hashibot:
_This issue was originally opened by @l2ol33rt as hashicorp/terraform#5525. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

From the [AWS docs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data):

```
User data is limited to 16 KB. This limit applies to the data in raw form, not
base64-encoded form.
```

When user-data is `>16 KB` the output of `terraform apply` only notifies the user that the state was not applied due to `timeout`:

```
Error applying plan:

1 error(s) occurred:

* aws_launch_configuration.main: Error creating launch configuration: timeout while waiting for state to become '[success]'
```

Setting `TF_LOG=DEBUG` didn't seem to help in this situation. But taking the size of the file `<16 KB` fixes the issue, but it took some digging to realize what was going on initially. 

I would think an error should be triggered during `terraform plan` when user-data exceeds 16 KB. Let me know what you think!

`Terraform version: v0.6.11`

bflad:
This issue has been addressed in the now released [terraform-provider-aws version 1.7.0](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#170-january-12-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #2765, opened 2017-12-24
r/aws_instance: validate user data size during plan

loivis:
Related issue: #97

Clear error message is returned during `terraform apply`.
```
Error: Error applying plan:

1 error(s) occurred:

* aws_instance.this: 1 error(s) occurred:

* aws_instance.this: Error launching source instance: InvalidParameterValue: User data is limited to 16384 bytes
	status code: 400, request id: f87cabb3-e79b-4bba-81a0-632c95e54891
```
However, it might still help to validate user data size during `terraform plan`.

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -116,6 +116,7 @@ func resourceAwsInstance() *schema.Resource {
 						return ""
 					}
 				},
+				ValidateFunc: validateInstanceUserDataSize,
 			},
 
 			"user_data_base64": {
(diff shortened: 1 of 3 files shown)


tonglil:
Is this implementation for `plan` or `apply`?

The title says `plan` but the comment suggests it is only for `apply`.

Clarification would be appreciated!

alewando:
The description is stating that currently you only get the error message when doing an "apply" (the error message comes from the remote end), but that it would be nice to get warned in the "plan" phase, which this PR implements.

tonglil:
Thanks @alewando.

bflad:

PR #2775, opened 2017-12-26
document import for ENIs

rremer:
For consistency, let's document that ENIs can be imported and how to do that.


jen20:
Thanks @rremer!

PR #2777, opened 2017-12-27
d/aws_ssm_parameter: Support returning raw encrypted SecureString value.

modax:
It is useful to have it for triggers or other purposes. Defaults to
returning decrypted value as before.

```
$ make testacc TESTARGS="-run 'TestAccAwsSsmParameterDataSource_'"                                                                                                                                                                                                                 148 ↵
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run 'TestAccAwsSsmParameterDataSource_' -timeout 120m
=== RUN   TestAccAwsSsmParameterDataSource_basic
--- PASS: TestAccAwsSsmParameterDataSource_basic (55.34s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       55.363s
```

### DIFF
--- a/aws/data_source_aws_ssm_parameter.go
+++ b/aws/data_source_aws_ssm_parameter.go
@@ -32,6 +32,11 @@ func dataSourceAwsSsmParameter() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+			"with_decryption": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


jen20:
Hi @modax! Thanks for opening up a pull request for this. I'm going to leave this one to the core team of HashiCorp employees to look at to ensure it is compatible with whatever strategy they are using to reduce the number of potential secrets present in Terraform state. cc @radeksimko.

modax:
@jen20, @radeksimko: Hi! Well, actually current behavior of this data source is to "leak" secrets to tfstate as the value is always decrypted. My patch adds a possibility to return value encrypted. Sure, it is possible to decrypt it manually via KMS API calls but we intend to use it for detecting if value has changed. In either case, it is much better than having a secret leaked in plaintext in tfstate when you don't really need its value. Still, however, it is still far from perfect :/

Actually you got me thinking that it makes more sense to rely on SSM parameter metadata like version or last modified date for detection if the value has changed. But implementation for that would be much more complex and would be a topic of another PR I believe.

bflad:

PR #2780, opened 2017-12-27
Update the Fargate documentation 

sebastianwahn:
Update the ECS Fargate documentation by moving the "launch_type" property to the correct place in aws_ecs_service.

Currently, it doesn't match the implemented aws_ecs_service resource, since the "launch_type" is beneath the "load_balancer", according to the docs.


jen20:
Thanks @sebastianwahn!

PR #2789, opened 2017-12-28
resource/aws_lambda_function: Validate the function name

ColinHebert:
Lambda function names (not the ARN, the actual name) has a restriction of 64 characters and alphanum with dash and underscore. (https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html)

> Note that the length constraint applies only to the ARN. If you specify only the function name, it is limited to 64 characters in length.

This adds a validator for those and rename the existing validator for function names that takes ARNs.

screen shot 2017-12-28 at 9 28 23 am
screen shot 2017-12-28 at 9 28 36 am

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -78,6 +78,7 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
+				ValidateFunc: validateLambdaFunctionName,
 			},
 			"handler": {
 				Type:     schema.TypeString,
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -75,9 +75,9 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				},
 			},
 			"function_name": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
 				ValidateFunc: validateLambdaFunctionName,
 			},
 			"handler": {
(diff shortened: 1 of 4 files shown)


ewbankkit:
When creating a Lambda function:

> The function name can contain only letters, numbers, hyphens, and underscores.
> The maximum length is 64 characters.

The current validator (`validateLambdaFunctionName`) is used when adding a Lambda permission:

> FunctionName
> The name of the Lambda function, version, or alias.
> 
> Name formats
> 
> Function name - my-function (name-only), my-function:v1 (with alias).
> Function ARN - arn:aws:lambda:us-west-2:123456789012:function:my-function.
> Partial ARN - 123456789012:function:my-function.
>
>You can append a version number or alias to any of the formats. The length constraint applies only to the full ARN. If you specify only the function name, it is limited to 64 characters in length.
>
> Length Constraints: Minimum length of 1. Maximum length of 140.
>
> Pattern: (arn:(aws[a-zA-Z-]*)?:lambda:)?([a-z]{2}(-gov)?-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\$LATEST|[a-zA-Z0-9-_]+))?

bflad:
Hi @ColinHebert 👋  Thank you for submitting this awhile ago and apologies for the long delay in review.

In the Lambda service, the `FunctionName` parameter across the various APIs is annoyingly ambiguous between accepting just a function name, the full ARN, or even a partial ARN. In the `CreateFunction` request, the [`FunctionName` parameter](https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-request-FunctionName) curiously accepts an ARN. Given that the Terraform AWS Provider `aws_lambda_function` resource has accepted full ARNs or partial ARNs in the past, we would need to keep that support for backward compatibility.

Given that we do not want to introduce a potentially breaking change to this validation, we are going to close this for now. However, if you would still like to introduce validation for the `aws_lambda_function` resource `function_name` argument, we would accept one that allows for any of the API supported formats. Thanks again.

PR #2804, opened 2017-12-29
r/aws_elasticsearch_domain: export kibana endpoint

### LINKED ISSUE #2802: Kibana Attribute Not available for elasticsearch
I want to be able to reference the Kibana endpoint for DNS or anything else I would like use it for later.  Currently there is only id, access_policies and domain_name that are attributes that are exported. 

Can we add kibana endpoint? 
$ terraform -version
Terraform v0.11.1
+ provider.aws v1.6.0

```
  id = esd-policy-dev-logs
  access_policies = {"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"es:*","Resource":"arn:aws:es:us-east-********/*"}]}
  domain_name = dev-logs
```

loivis:
Fixes #2802 

@fewknow, is this what you want, or rather export with `https://` scheme? The `endpoint` for elasticsearch domain itself is just a dns record, without any scheme.

### DIFF
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -484,14 +484,14 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 		if err != nil {
 			return err
 		}
-		d.Set("kibana_endpoint", endpoints["vpc"].(string)+"/_plugin/kibana/")
+		d.Set("kibana_endpoint", getKibanaEndpoint(d))
 		if ds.Endpoint != nil {
 			return fmt.Errorf("%q: Elasticsearch domain in VPC expected to have null Endpoint value", d.Id())
 		}
 	} else {
 		if ds.Endpoint != nil {
 			d.Set("endpoint", *ds.Endpoint)
-			d.Set("kibana_endpoint", *ds.Endpoint+"/_plugin/kibana/")
+			d.Set("kibana_endpoint", getKibanaEndpoint(d))
 		}
 		if ds.Endpoints != nil {
 			return fmt.Errorf("%q: Elasticsearch domain not in VPC expected to have null Endpoints value", d.Id())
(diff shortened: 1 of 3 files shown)


fewknow:
@loivis looks like this will satisfy my use case.   I just wanted to be able to assign DNS to this in terraform so my end users can hit kibana.dev.corvestacloud.net.   

Thanks

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -480,12 +484,14 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 		if err != nil {
 			return err
 		}
+		d.Set("kibana_endpoint", endpoints["vpc"].(string)+"/_plugin/kibana/")
```
Ninir: Could we make use of a function that build this endpoint for us here and below?

loivis:
Hi @Ninir 

I added documentation for the attribute and extracted the function to populate kibana endpoint.

Regarding test, I just add simple check in acceptance test. Not sure if we need this check when setting the attribute. It is only basic string concatenation of `endpoint` attribute and `/_plugin/kibana/`

loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticSearchDomain_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSElasticSearchDomain_basic -timeout 120m
=== RUN   TestAccAWSElasticSearchDomain_basic
--- PASS: TestAccAWSElasticSearchDomain_basic (870.58s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	870.636s
```

website/docs/r/elasticsearch_domain.html.markdown
```diff
@@ -109,6 +109,7 @@ The following attributes are exported:
 * `arn` - Amazon Resource Name (ARN) of the domain.
 * `domain_id` - Unique identifier for the domain.
 * `endpoint` - Domain-specific endpoint used to submit index, search, and data upload requests.
+* `kibana_endpoint` - Domain-specific endpoint for kibana without https scheme, the same as `endpoint`.
```
Ninir: By "the same as `endpoint`", do you mean the same as in "without https"?
If so, I think this is misleading with the beginning of the sentence, what do you think? :)

loivis: Exactly. Now removed those confusing words.

bflad:

Issue #2536, opened 2017-12-05
cannot change targets on aws_ssm_association 


hashibot:
_This issue was originally opened by @af6140 as hashicorp/terraform#16837. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

When changing the aws_ssm_association resource target attributes  with new tag name and value, the change is not applied.

bflad:
This has been released in [version 1.10.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1100-february-24-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #2807, opened 2017-12-31
resource/aws_ssm_association: Allow updating targets

atsushi-ishibashi:
Fixed: #2536 
Update MaxItems of targets. [ref](https://docs.aws.amazon.com/ja_jp/systems-manager/latest/APIReference/API_CreateAssociation.html#EC2-CreateAssociation-request-Targets)

```
TF_ACC=1 go test ./aws -v -run=TestAccAWSSSMAssociation_withTargets -timeout 120m
=== RUN   TestAccAWSSSMAssociation_withTargets
--- PASS: TestAccAWSSSMAssociation_withTargets (41.74s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	41.774s
```

### DIFF
--- a/aws/resource_aws_ssm_association.go
+++ b/aws/resource_aws_ssm_association.go
@@ -73,7 +73,6 @@ func resourceAwsSsmAssociation() *schema.Resource {
 			"targets": {
 				Type:     schema.TypeList,
 				Optional: true,
-				ForceNew: true,
 				Computed: true,
 				MaxItems: 5,
 				Elem: &schema.Resource{
--- a/aws/resource_aws_ssm_association.go
+++ b/aws/resource_aws_ssm_association.go
@@ -213,6 +213,10 @@ func resourceAwsSsmAssocationUpdate(d *schema.ResourceData, meta interface{}) er
 		associationInput.OutputLocation = expandSSMAssociationOutputLocation(d.Get("output_location").([]interface{}))
 	}
 
+	if d.HasChange("targets") {
+		associationInput.Targets = expandAwsSsmTargets(d)
+	}
+
 	_, err := ssmconn.UpdateAssociation(associationInput)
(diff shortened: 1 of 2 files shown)


aws/resource_aws_ssm_association.go
```diff
@@ -75,7 +75,7 @@ func resourceAwsSsmAssociation() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
```
radeksimko: Since we're now able to update this I think we should also remove `ForceNew` from here, otherwise the new code will never be executed. 😉

aws/resource_aws_ssm_association_test.go
```diff
@@ -40,6 +40,7 @@ func TestAccAWSSSMAssociation_withTargets(t *testing.T) {
 				Config: testAccAWSSSMAssociationBasicConfigWithTargets(name),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAWSSSMAssociationExists("aws_ssm_association.foo"),
+					resource.TestCheckResourceAttr("aws_ssm_association.foo", "targets.#", "2"),
```
radeksimko: Do you mind adding a test step here which will add or remove a target and check the no of targets in each step to verify that the update actually works?

atsushi-ishibashi:
@radeksimko Ok👍
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSSSMAssociation_withTargets -timeout 120m
=== RUN   TestAccAWSSSMAssociation_withTargets
--- PASS: TestAccAWSSSMAssociation_withTargets (116.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	116.984s
```

bflad:

PR #2809, opened 2017-12-31
resource/aws_autoscaling_policy: Replace GetOk with GetOkExists

atsushi-ishibashi:
Fixed: #2528 
When you specify `0`, `GetOk` returns false as exists so replaced with `GetOkExists`.

### DIFF
--- a/aws/resource_aws_autoscaling_policy.go
+++ b/aws/resource_aws_autoscaling_policy.go
@@ -287,7 +287,7 @@ func getAwsAutoscalingPutScalingPolicyInput(d *schema.ResourceData) (autoscaling
 		params.Cooldown = aws.Int64(int64(v.(int)))
 	}
 
-	if v, ok := d.GetOk("estimated_instance_warmup"); ok {
+	if v, ok := d.GetOkExists("estimated_instance_warmup"); ok {
 		params.EstimatedInstanceWarmup = aws.Int64(int64(v.(int)))
 	}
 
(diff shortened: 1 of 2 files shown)


atsushi-ishibashi:
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSAutoscalingPolicy_zerovalue -timeout 120m
=== RUN   TestAccAWSAutoscalingPolicy_zerovalue
--- PASS: TestAccAWSAutoscalingPolicy_zerovalue (115.46s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	115.487s
```
I got the below error so added `validation.IntAtLeast(1)` as ValidateFunc.
```
aws_autoscaling_policy.foobar_step: Error putting scaling policy: ValidationError: If specified, MinAdjustmentMagnitude must be greater than zero.
```

radeksimko:
I was tempted to say that we're not really testing the new behaviour, because AWS API just doesn't return the default value (`300`), but `nil` instead, which `d.Set()` then turns back into `0`.
This made me think we could only call `d.Set()` if the field isn't `nil`, but that might break some people's configs where they're referencing the field in conditionals, so it's probably ok to leave it as is. 🤷‍♂️ 

I just rebased your branch and resolved conflicts for you, so we can 🚢 this.

Thanks for the PR, as always! 👍

bflad:

PR #2814, opened 2017-12-31
[WIP, NeedHelp]New Resource: codebuild_webhook

atsushi-ishibashi:
Support: #2571 

The below error occured, how should we do?
```
* aws_codebuild_webhook.test: ResourceNotFoundException: Could not find access token for server type github
```

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -301,6 +301,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_codecommit_repository":                    resourceAwsCodeCommitRepository(),
 			"aws_codecommit_trigger":                       resourceAwsCodeCommitTrigger(),
 			"aws_codebuild_project":                        resourceAwsCodeBuildProject(),
+			"aws_codebuild_webhook":                        resourceAwsCodeBuildWebhook(),
 			"aws_codepipeline":                             resourceAwsCodePipeline(),
 			"aws_customer_gateway":                         resourceAwsCustomerGateway(),
 			"aws_db_event_subscription":                    resourceAwsDbEventSubscription(),
--- a/aws/resource_aws_codebuild_webhook.go
+++ b/aws/resource_aws_codebuild_webhook.go
@@ -1,0 +1,84 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/codebuild"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCodeBuildWebhook() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCodeBuildWebhookCreate,
+		Read:   resourceAwsCodeBuildWebhookRead,
+		Delete: resourceAwsCodeBuildWebhookDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"url": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsCodeBuildWebhookCreate(d *schema.ResourceData, meta interface{}) error {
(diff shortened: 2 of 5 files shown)


radeksimko:
Hi @atsushi-ishibashi 
I believe the error is caused by the fact that webhooks can only be set up for repositories which Codebuild has access to (that is access to setup hooks, not just to clone it). It seems (unfortunately) that the only way to give Codebuild access is to go to the AWS Console and follow the OAuth flow per https://docs.aws.amazon.com/codebuild/latest/userguide/create-project.html#create-project-console

It seems the token remains stored somewhere in the AWS account (presumably per region) and _then_ can be used automatically when creating new projects and/or webhooks.

We _might_ be able to automate the auth part via https://www.terraform.io/docs/providers/github/r/organization_webhook.html but I didn't test it myself, so I don't know for sure. In case it's possible we'd need to vendor & import the github provider in this repo. It's possible though, that once you create the webhook on Github's side this way it will also be automatically created in AWS, so this resources kind of becomes redundant. I'm just guessing how it works though - it needs testing.

The other option (if we cannot automate the auth) would be to expose a few ENV variables and provide clear instructions, e.g. `AWS_CODEBUILD_AUTHED_LOCATION` which the user would set to a repository that is within the OAuth scope they previously authorized in the Console. Then anyone could run acceptance tests by 1st authorizing Github in the Console and 2nd providing `AWS_CODEBUILD_AUTHED_LOCATION` as part of the test run.

Do you mind exploring some options in this area?

salvianreynaldi:
I think now CodeBuild's GitHub webhook supports branch filtering, https://docs.aws.amazon.com/cli/latest/reference/codebuild/create-webhook.html
It's not announced, but it's there. can you please support this parameter as well?

joestump:
Anything I can do to help get this PR across the finish line?

joestump:
This adds branch filtering per @salvianreynaldi's request.

```diff
--- a/aws/resource_aws_codebuild_webhook.go
+++ b/aws/resource_aws_codebuild_webhook.go
@@ -18,6 +18,10 @@ func resourceAwsCodeBuildWebhook() *schema.Resource {
                                Required: true,
                                ForceNew: true,
                        },
+                       "branch_filter": {
+                               Type:     schema.TypeString,
+                               Optional: true,
+                       },
                        "url": {
                                Type:     schema.TypeString,
                                Computed: true,
@@ -30,7 +34,8 @@ func resourceAwsCodeBuildWebhookCreate(d *schema.ResourceData, meta interface{})
        conn := meta.(*AWSClient).codebuildconn
 
        resp, err := conn.CreateWebhook(&codebuild.CreateWebhookInput{
-               ProjectName: aws.String(d.Get("name").(string)),
+               ProjectName:  aws.String(d.Get("name").(string)),
+               BranchFilter: aws.String(d.Get("branch_filter").(string)),
        })
        if err != nil {
                return err
```

@radeksimko as for tying this in with GitHub, you'd use the `github_repository_webhook` resource to tie the two together:

```hcl
resource "aws_codebuild_webhook" "github" {
    name = "my-project"
}

resource "github_repository_webhook" "aws_codebuild" {

  name = "awscodedeploy"

  configuration {
    url          = "${aws_codebuild_webhook.github.url}"
    content_type = "json"
  }

  events = ["pull_request", "push"]
}
```

joestump:
We're heading towards month five on this PR. I'm happy to create my own version of it with the `branchFilter` stuff and some better docs for examples if it'll increase the chances of it getting merged. This is a highly desired resource on my team and we've got Go/Terraform experience.

@radeksimko is adding `branchFilter` and better docs on tying it to GitHub all that you can see this in need of?

joestump:
@atsushi-ishibashi @radeksimko I've opened #4473 to address the comments here as well as the original error regarding not being able to find an access token.

bflad:
Work for this was finished up in #4473 and in a commit afterwards. The new `aws_codebuild_webhook` resource will be available in version 1.21.0 of the AWS provider, likely releasing middle of next week. 🎉

bflad:

Issue #2731, opened 2017-12-21
Add AWS Lightsail DNS Entry Support


tankertuf:
Hi there,

Would it be possible to add support for adding/removing DNS records to Lightsail domains?

API references:
https://docs.aws.amazon.com/lightsail/2016-11-28/api-reference/API_CreateDomainEntry.html
https://docs.aws.amazon.com/lightsail/2016-11-28/api-reference/API_DeleteDomainEntry.html

Thank you,
John

bflad:
Hi @tankertuf, thanks for submitting the feature request! Double checking the AWS Go SDK, its certainly supported: https://docs.aws.amazon.com/sdk-for-go/api/service/lightsail/

Given the holidays and other priorities, the Terraform team at HashiCorp won't be able to work on this in the near future, but we'd be happy to review a pull request if you or someone else has the time and motivation to implement it. Alternatively, if others would also like to see this implemented I'd encourage adding a :+1: upvote reaction to the original issue comment (not to this comment), which we use as one of the inputs to prioritize work for the Terraform team.

dereknex:
Does anyone working it?

arenstar:
Yes please!!! id like to see this feature too

joakiboxxx:
I'm considering putting in some effort towards implementing better Lightsail support. Should Lightsail be a part of the AWS provider, or perhaps split into a separate provider? @bflad?

benfortuna:
@joakiboxxx  As lightsail product is competing with the likes of Digital Ocean I think splitting to separate provider could be justified.

imansilla:
i'm waiting for this :)

Rendanic:
Are there any news?

brittandeyoung:
@Rendanic 
For Awareness I have begun work on creating a separate AWS lightsail provider in order to help with Speeding up contributions and to give it dedicated attention. 

The Github repo is located here: https://github.com/DeYoungTech/terraform-provider-awslightsail

The provider and documentation is published here: https://registry.terraform.io/providers/DeYoungTech/awslightsail/latest

With the most recent version, I have added support for the `domain_entry`  resource.

If you want to start using this resource now, feel free to check it out and provide feedback.

jackivanov:
no more updates?

PR #2815, opened 2017-12-31
New Resource: aws_lightsail_domain_entry

loivis:
lightsail domain entry for #2731 

This is my first attempt to implement a new resource. Comments and feedbacks are welcome.

The resource is at least working for now. But I do have a few questions about the implementation.

1. The resource schema follows the same schema as AWS API for [CreateDomainEntryInput](https://docs.aws.amazon.com/sdk-for-go/api/service/lightsail/#CreateDomainEntryInput),
 [UpdateDomainEntryInput](https://docs.aws.amazon.com/sdk-for-go/api/service/lightsail/#UpdateDomainEntryInput), [DeleteDomainEntryInput](https://docs.aws.amazon.com/sdk-for-go/api/service/lightsail/#DeleteDomainEntryInput). I'm not sure if it's good idea to flat all into common `TypeString`. In that case, it will make the implementation a bit easier but break the identical mapping between `terraform` and AWS API.

2. Name for the domain entry in AWS API is the full FQDN with domain name. From user perspective, it would be easier to just use a string without domain name.  We can append domain name before sending request to AWS. The same concern of identical mapping applies here as well. Still I'm following convention for AWS API.

3. In [AWS API DomainEntry](https://docs.aws.amazon.com/lightsail/2016-11-28/api-reference/API_DomainEntry.html), `options` is already marked as `Deprecated`. So I didn't add this parameter at all. 

4. I haven't fully figured out how import works, so support for import is not there yet.

5. I'd like to validate `target` argument based on `is_alias`. AFAIK, `ValidateFunc` works only for the current argument. I don't have it in mind where similar logic is implemented.

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -393,6 +393,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_lambda_permission":                        resourceAwsLambdaPermission(),
 			"aws_launch_configuration":                     resourceAwsLaunchConfiguration(),
 			"aws_lightsail_domain":                         resourceAwsLightsailDomain(),
+			"aws_lightsail_domain_entry":                   resourceAwsLightsailDomainEntry(),
 			"aws_lightsail_instance":                       resourceAwsLightsailInstance(),
 			"aws_lightsail_key_pair":                       resourceAwsLightsailKeyPair(),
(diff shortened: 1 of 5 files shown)


loivis:
I won't be able to continue the work so close it.

PR #2834, opened 2018-01-03
resource/aws_api_gateway_integration: Allow update of URI attributes 

mdlavin:
Instead of always forcing a new resource creation for URI changes (like a Lambda function version update), the integration:update action allows for in place updates.

### DIFF
--- a/aws/resource_aws_api_gateway_integration.go
+++ b/aws/resource_aws_api_gateway_integration.go
@@ -51,7 +51,6 @@ func resourceAwsApiGatewayIntegration() *schema.Resource {
 			"uri": {
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: true,
 			},
 
 			"credentials": {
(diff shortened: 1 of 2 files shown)


aws/resource_aws_api_gateway_integration.go
```diff
@@ -370,6 +369,17 @@ func resourceAwsApiGatewayIntegrationUpdate(d *schema.ResourceData, meta interfa
 		})
 	}
 
+	// The documentation https://docs.aws.amazon.com/apigateway/api-reference/link-relation/integration-update/ says
+	// that uri changes are only supported for non-mock types. Because the uri value is not used in mock
+	// resources, it means that the uri can always be updated
```
Ninir: Can you add a testcase that updates the URI?

mdlavin:
I added a testcase that updated the URI, but I didn't see an obvious way to assert that an in-place update was used. I'm happy to update the testcase to assert the in-place update if you can point me at an example test that does that sort of assertion.

Ninir:
@mdlavin I'm not even sure it's doable due to the nature of the API :/
Will merge as is for this reason, after running the tests. THanks!

bflad:

PR #2843, opened 2018-01-03
New Resource: aws_gamelift_build

radeksimko:
I want to take a look into the ZIP file to see if there's any way to bring the size down from 15M, but otherwise this is ready for review.

The ZIP file was basically created by following steps in https://github.com/awslabs/aws-gamelift-sample/blob/master/deployment/deployment.md#gamelift----game-server-build--fleet-creation
and it unfortunately contains a few binaries which practically can't be compressed.

The build resource here practically doesn't require a ZIP file with fully working game, but `aws_gamelift_fleet` (for which I already have a PR in a queue) will.

## Import

The main reason I decided not to implement import is because `DescribeBuild` doesn't return `storage_location` which is `ForceNew`, so having such empty field in the state would result in recreation.

## Test Results

```
$ make testacc TESTARGS='-run=TestAccAWSGamelift'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccAWSGamelift -timeout 120m
=== RUN   TestAccAWSGameliftBuild_basic
--- PASS: TestAccAWSGameliftBuild_basic (313.15s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	313.202s
```

### DIFF
--- a/aws/resource_aws_gamelift_build.go
+++ b/aws/resource_aws_gamelift_build.go
@@ -1,0 +1,187 @@
+package aws
+
+import (
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/gamelift"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsGameliftBuild() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsGameliftBuildCreate,
+		Read:   resourceAwsGameliftBuildRead,
+		Update: resourceAwsGameliftBuildUpdate,
+		Delete: resourceAwsGameliftBuildDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 1024),
+			},
+			"operating_system": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateGameliftOperatingSystem,
+			},
+			"storage_location": {
+				Type:     schema.TypeList,
+				Required: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"bucket": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"key": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"role_arn": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     true,
+							ValidateFunc: validateArn,
+						},
+					},
+				},
+			},
+			"version": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringLenBetween(1, 1024),
+			},
+		},
+	}
+}
+
+func resourceAwsGameliftBuildCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).gameliftconn
+
+	sl := expandGameliftStorageLocation(d.Get("storage_location").([]interface{}))
+	input := gamelift.CreateBuildInput{
+		Name:            aws.String(d.Get("name").(string)),
+		OperatingSystem: aws.String(d.Get("operating_system").(string)),
+		StorageLocation: sl,
+	}
+	if v, ok := d.GetOk("version"); ok {
+		input.Version = aws.String(v.(string))
+	}
+	log.Printf("[INFO] Creating Gamelift Build: %s", input)
+	var out *gamelift.CreateBuildOutput
+	err := resource.Retry(30*time.Second, func() *resource.RetryError {
+		var err error
+		out, err = conn.CreateBuild(&input)
+		if err != nil {
+			if isAWSErr(err, gamelift.ErrCodeInvalidRequestException, "Provided build is not accessible.") {
+				return resource.RetryableError(err)
+			}
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	d.SetId(*out.Build.BuildId)
+
+	stateConf := resource.StateChangeConf{
+		Pending: []string{gamelift.BuildStatusInitialized},
+		Target:  []string{gamelift.BuildStatusReady},
+		Timeout: 1 * time.Minute,
+		Refresh: func() (interface{}, string, error) {
+			out, err := conn.DescribeBuild(&gamelift.DescribeBuildInput{
+				BuildId: aws.String(d.Id()),
+			})
+			if err != nil {
+				return 42, "", err
+			}
+
+			return out, *out.Build.Status, nil
+		},
+	}
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return err
+	}
+
+	return resourceAwsGameliftBuildRead(d, meta)
+}
+
+func resourceAwsGameliftBuildRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).gameliftconn
+
+	log.Printf("[INFO] Reading Gamelift Build: %s", d.Id())
+	out, err := conn.DescribeBuild(&gamelift.DescribeBuildInput{
+		BuildId: aws.String(d.Id()),
+	})
+	if err != nil {
+		if isAWSErr(err, gamelift.ErrCodeNotFoundException, "") {
+			log.Printf("[WARN] Gamelift Build (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+	b := out.Build
+
+	d.Set("name", b.Name)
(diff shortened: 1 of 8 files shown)


stack72:
@radeksimko can't this zip be hosted on a public S3 bucket and then downloaded as part of the test?

radeksimko:
Ideally I'd like the whole test suite to be self-sustainable, not relying on precisely named files in a precisely named S3 bucket. It makes the test a bit fragile.

radeksimko:
This is the contents of the ZIP file:

```
$ du -h ./*
100K	./GomokuServer.exe
1000K	./aws-cpp-sdk-core.dll
1.9M	./aws-cpp-sdk-gamelift-server.dll
688K	./aws-cpp-sdk-sqs.dll
4.0K	./config.ini
4.0K	./install.bat
 14M	./vc_redist.x64.exe
```
where the last binary (which was downloaded from MSFT's website per linked tutorial) in the list occupies the majority of the space. This leads me to a conclusion that we can't unfortunately make it much smaller.

To put this into context here's usage of the whole repo:

```
du -s -h ./*
 56K	./CHANGELOG.md
4.0K	./GNUmakefile
 20K	./LICENSE
4.0K	./README.md
8.5M	./aws
304K	./examples
4.0K	./main.go
 16K	./scripts
 59M	./vendor
1.8M	./website
```
which has close to 70MB. It is unfortunate, but IMO not a _huge_ deal and more importantly I can't think of a better solution (other than what was mentioned by @stack72 above).

bflad:
Starting to take a look at this now.

At first glance through the comments above, I think I would also prefer we don't add "large" binaries to the repository unless absolutely necessary, but I have not dug into options/caveats here. Incurring additional MBs of Git download for everyone developing on the repository (versus developing the specific resource that requires it) is not friendly for folks with poor internet connections. I think in the worst case we could .gitignore the file location and `t.Fatalf("Missing X file, download it from Y")` or just automatically download(/build) the file when running the specific acceptance tests and its missing.

aws/resource_aws_gamelift_build.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func resourceAwsGameliftBuildDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).gameliftconn
+
+	log.Printf("[INFO] Deleting Gamelift Build: %s", d.Id())
+	_, err := conn.DeleteBuild(&gamelift.DeleteBuildInput{
+		BuildId: aws.String(d.Id()),
+	})
+	if err != nil {
```
bflad: Nitpick: We should ignore missing build IDs:
```
if err != nil {
  if isAwsErr(err, gamelift.ErrCodeNotFoundException, "") {
    return nil
  }
  return err
}
```

aws/resource_aws_gamelift_build_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	if len(resp.Builds) == 0 {
+		log.Print("[DEBUG] No Gamelift Builds to sweep")
+		return nil
+	}
+
+	log.Printf("[INFO] Found %d Gamelift Builds", len(resp.Builds))
+
+	for _, build := range resp.Builds {
+		if !strings.HasPrefix(*build.Name, "tf_acc_build_") {
```
bflad: Nitpick: I think we should use a constant here if we're going to expect continuity between the acceptance test naming and the sweeper.

aws/resource_aws_gamelift_build_test.go
```diff
@@ -0,0 +1,302 @@
...
+func testAccCheckAWSGameliftBuildDestroy(s *terraform.State) error {
+	conn := testAccProvider.Meta().(*AWSClient).gameliftconn
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_gamelift_build" {
+			continue
+		}
+
+		req := gamelift.DescribeBuildInput{
+			BuildId: aws.String(rs.Primary.ID),
```
bflad: Do we need to check `rs.Primary.ID` beforehand? e.g.

```go
if rs.Primary.ID == "" {
  continue
}
```

aws/resource_aws_gamelift_build_test.go
```diff
@@ -0,0 +1,302 @@
...
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "name", buildName),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "operating_system", "WINDOWS_2012"),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "storage_location.#", "1"),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "storage_location.0.bucket", bucketName),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "storage_location.0.key", "tf-acc-test-gl-build.zip"),
+					resource.TestMatchResourceAttr("aws_gamelift_build.test", "storage_location.0.role_arn", roleArnRe),
+				),
+			},
+			{
+				Config: testAccAWSGameliftBuildBasicUpdateConfig(uBuildName, bucketName, zipPath, roleName, policyName),
```
bflad: Nitpick: I don't think we need the "UpdateConfig" since you've already parameterized the original configuration. 😄

aws/resource_aws_gamelift_build_test.go
```diff
@@ -0,0 +1,302 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSGameliftBuildBasicConfig(buildName, bucketName, zipPath, roleName, policyName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSGameliftBuildExists("aws_gamelift_build.test", &conf),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "name", buildName),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "operating_system", "WINDOWS_2012"),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "storage_location.#", "1"),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "storage_location.0.bucket", bucketName),
+					resource.TestCheckResourceAttr("aws_gamelift_build.test", "storage_location.0.key", "tf-acc-test-gl-build.zip"),
```
bflad: Does this need to be parameterized for parallelism/failures?

radeksimko: Not anymore, I found a better way which also eliminates the big ZIP file 😃

aws/resource_aws_gamelift_build_test.go
radeksimko: Yeah, that's a good idea.

aws/resource_aws_gamelift_build.go
```diff
@@ -0,0 +1,186 @@
...
+		Timeout: 1 * time.Minute,
+		Refresh: func() (interface{}, string, error) {
+			out, err := conn.DescribeBuild(&gamelift.DescribeBuildInput{
+				BuildId: aws.String(d.Id()),
+			})
+			if err != nil {
+				return 42, "", err
+			}
+
+			return out, *out.Build.Status, nil
```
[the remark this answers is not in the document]
radeksimko: Well, I'd normally opt in for providing more context if there was more context - e.g. any event feed we can pull the context (reasons of failure) from, but I think in this case the error message is more or less equivalent. It still doesn't tell you _why_ things have failed, unfortunately.

aws/resource_aws_gamelift_build.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/gamelift"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsGameliftBuild() *schema.Resource {
+	return &schema.Resource{
```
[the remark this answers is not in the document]
radeksimko: 🤷‍♂️

aws/resource_aws_gamelift_build.go
radeksimko: We could, but `DeleteBuild` doesn't ever return such error if you try deleting build which doesn't exist. 🤷‍♂️ 

It just returns `200 OK`.

aws/resource_aws_gamelift_build_test.go
radeksimko: I'd say hopefully not? I mean we don't do it in in any `CheckDestroy` function for any other acceptance test and I'm hoping core would've already swiped the resource away from state if it doesn't have ID, if not then we have much bigger problems IMO.

radeksimko:
It took some effort but I found a way to avoid uploading the big ZIP file to the repo. 😺 

I went through the Gamelift tutorial in AWS Console and traced down the location of the Amazon's sample game from CloudTrail. Then I did it for every region Gamelift is currently available in, just to make tests work in potentially any region.

aws/resource_aws_gamelift_build_test.go
```diff
@@ -0,0 +1,230 @@
...
+	roleArn := fmt.Sprintf("arn:aws:iam::%s:role/sample-build-upload-role-%s", accId, region)
+
+	return &gamelift.S3Location{
+		Bucket:  aws.String(bucket),
+		Key:     aws.String(key),
+		RoleArn: aws.String(roleArn),
+	}, nil
+}
+
+// Account ID found from CloudTrail event (role ARN) after finishing tutorial in given region
```
bflad: 😍

bflad:

PR #2846, opened 2018-01-03
resource/aws_security_group: Allow empty rule description

### LINKED ISSUE #2831: aws_security_group_rule description validation is incorrect
### Terraform Version
v0.11.1
aws v1.6.0

### Affected Resource(s)
- aws_security_group_rule
Could affect other places that have `description`, but was not tested

Setting the description to "" is not allowed by the validation regex and will give the error: 
"description" doesn't comply with restrictions ("^[A-Za-z0-9 \\.\\_\\-\\:\\/\\(\\)\\#\\,\\@\\[\\]\\+\\=\\;\\{\\}\\!\\$\\*]+$"): ""
But you can set it to " ", which will get converted to "". If you do this, terraform will then think there is a change every run.

Since "" is obviously allowed by AWS, the '+' in the regex should be changed to '*'
Also, should not allow a description of nothing but spaces if it is always going to be converted to an empty string

This will allow definitions to be created where the description is optional

### Expected Behavior
Setting description to "" should work

### Actual Behavior
Setting description to "" produces an error:
"description" doesn't comply with restrictions ("^[A-Za-z0-9 \\.\\_\\-\\:\\/\\(\\)\\#\\,\\@\\[\\]\\+\\=\\;\\{\\}\\!\\$\\*]+$"): ""

jen20:
```
❯ make testacc TESTARGS="-run TestAccAWSSecurityGroup_emptyRuleDescription"
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccAWSSecurityGroup_emptyRuleDescription -timeout 120m
=== RUN   TestAccAWSSecurityGroup_emptyRuleDescription
--- PASS: TestAccAWSSecurityGroup_emptyRuleDescription (195.99s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	196.033s
```

Fixes #2831.

### DIFF
--- a/aws/resource_aws_security_group_test.go
+++ b/aws/resource_aws_security_group_test.go
@@ -1420,6 +1420,28 @@ func TestAccAWSSecurityGroup_egressWithPrefixList(t *testing.T) {
 	})
 }
 
+func TestAccAWSSecurityGroup_emptyRuleDescription(t *testing.T) {
+	var group ec2.SecurityGroup
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSecurityGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSSecurityGroupConfigEmptyRuleDescription,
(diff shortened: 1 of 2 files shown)


bflad:

PR #2847, opened 2018-01-03
Wrap aws_lambda_function.environment.variables type in &schema.Schema

jimmy-btn:
Wrap element type in &schema.Schema so that it conforms to the expected Terraform Schema structure

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -170,7 +170,6 @@ func resourceAwsLambdaFunction() *schema.Resource {
 						"variables": {
 							Type:     schema.TypeMap,
 							Optional: true,
-							Elem:     schema.TypeString,
 						},
 					},
 				},
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -170,6 +170,7 @@ func resourceAwsLambdaFunction() *schema.Resource {
 						"variables": {
 							Type:     schema.TypeMap,
 							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
 						},
 					},
 				},


Ninir:
Hi @jimmy-btn 

Could you give more details about how this is happening? 
If you can provide debug logs, HCL configurations and steps to reproduce, this would be greatly appreciated. Thanks!

jimmy-btn:
Hi Ninir!

tl;dr This isn't an issue today, because Terraform doesn't validate the types of elements passed to Maps properly, but will cause issues in the future.

This one is a bit tricky - I noticed it while making some changes inside Terraform, but haven't worked out how to tickle it from the public interface because Terraform doesn't validate the types for elements passed to maps via the schema provider interface.

This is definitely an invalid schema - although I mis-understood why. Terraform is happy to have an Elem field for TypeMap, but expects Elem to be a *Schema or a *Resource. I've modified this PR to wrap the field correctly.

**Why is this a problem?**
In some cases, Terraform will traverse the full Schema, and expects Elem to be a *Schema or a *Resource.

See CoreConfigSchema() function in helper/schema/core_schema.go for an example of code which is going to panic if this is not the case.

**Reproducing**
This doesn't cause any real problems today for Maps, but we can reproduce the issue using a List.

Add a field to an existing resource which implements a list of Integers, and specify the type in Elem directly (as we have with the environment.variables map for aws_lambda_function):

```
"test": {
    Type:     schema.TypeList,
    Optional: true,
    Elem:     Type: schema.TypeInt,
},
```

Terraform will not pick up on the error when validating the schema, but if you try and create or import a resource which uses that field but supplies the incorrect type, Terraform will panic and crash. For example, the following resource will cause a panic:

```
resource "my_resource" "name" {
    test = ["string_not_integer"]
}
```

If you specify Elem as a *Schema:

```
"test": {
    Type:     schema.TypeList,
    Optional: true,
    Elem:     &schema.Schema(Type: schema.TypeInt),
},
```

Terraform will correctly identify the type mismatch and output 
`test.0: cannot parse '' as bool: strconv.ParseBool: parsing "string_not_integer": invalid syntax` instead. Much nicer!

In https://github.com/hashicorp/terraform/pull/17037 I am proposing adding an explicit test for this to InternalValidate to make sure all provider schemas are in good shape, in case we ever want to add type validation for Maps.

jimmy-btn:
I'm going to drop this PR and replace with one which covers all instances of this issue in the codebase. Thank you!

PR #2849, opened 2018-01-03
Make it possible to import Route53 records containing a wildcard

jimmy-btn:
Today, route 53 records can only be manually imported into Terraform if they do not contain a wildcard matcher ('*'). This PR consistently applies normalization of the record during import so that wildcard domains can also be imported.


jimmy-btn:
Hey folks; this one has been hanging out for a while. Let me know if there's anything I can do to move it forward!

bflad:
@jimmy-btn sorry this pull request got lost in the mix. 😅  I believe this got fixed in #4183 which was released in [version 1.15.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1150-april-18-2018).

I'm going to close this but please do ping me back if its not resolved.

PR #2850, opened 2018-01-03
resource/aws_instance: Ensure at least one security group is being attached

Jonnymcc:
VPC-EC2 instances require at least 1 security group be attached. I was able to run apply and try to remove all security groups from an instance and it fails silently. terraform apply says that it is modifying the aws_instance resource but never finishes.

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -893,6 +893,10 @@ func resourceAwsInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
 				groups = append(groups, aws.String(v.(string)))
 			}
 		}
+
+		if len(groups) < 1 {
+			return fmt.Errorf("At least one security group needs to be attached!")
+		}
 		// If a user has multiple network interface attachments on the target EC2 instance, simply modifying the
 		// instance attributes via a `ModifyInstanceAttributes()` request would fail with the following error message:
 		// "There are multiple interfaces attached to instance 'i-XX'. Please specify an interface ID for the operation instead."
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -895,7 +895,7 @@ func resourceAwsInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 
 		if len(groups) < 1 {
-			return fmt.Errorf("At least one security group needs to be attached!")
+			return fmt.Errorf("VPC-based instances require at least one security group to be attached.")
 		}
 		// If a user has multiple network interface attachments on the target EC2 instance, simply modifying the
 		// instance attributes via a `ModifyInstanceAttributes()` request would fail with the following error message:
--- a/aws/resource_aws_instance_test.go
+++ b/aws/resource_aws_instance_test.go
@@ -701,6 +701,43 @@ func TestAccAWSInstance_NetworkInstanceSecurityGroups(t *testing.T) {
 	})
 }
 
+func TestAccAWSInstance_NetworkInstanceRemovingAllSecurityGroups(t *testing.T) {
+	var v ec2.Instance
+
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
+		IDRefreshName: "aws_instance.foo_instance",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckInstanceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccInstanceNetworkInstanceVPCSecurityGroupIDs(rInt),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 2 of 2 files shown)


aws/resource_aws_instance.go
```diff
@@ -893,6 +893,10 @@ func resourceAwsInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
 				groups = append(groups, aws.String(v.(string)))
 			}
 		}
+
+		if len(groups) < 1 {
+			return fmt.Errorf("At least one security group needs to be attached!")
```
Ninir: Since it is true only for VPC-based instances, what do you think of enhancing this message with: `VPC-based instances require at least one security group to be attached`?

Source: 
> [EC2-VPC] Changes the security groups of the instance. You must specify at least one security group, even if it's just the default security group for the VPC. You must specify the security group ID, not the security group name.

Jonnymcc: Does this not apply in EC2 Classic? I haven't tested there and haven't found anything relevant in the AWS docs. If it does only apply to VPC security groups then I think we can make the message more explicit.

Ninir: According to https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyInstanceAttribute.html#API_ModifyInstanceAttribute_RequestParameters under `GroupId.N` you will found the extract I wrote above. Apart from that, can't really tell 🤷‍♂️ 

If it's also applicable to EC2-classic, then we would need to add `MaxItems: 1` to the `vpc_security_group_ids` attribute, which would handle it in all cases.

Can you check it against EC2-classic?

Jonnymcc: I'm unable to test in EC2-classic since my account does not support it. I can only create instances in a VPC. I'll update the error message though.

Jonnymcc: Actually, I was able to test in EC2-classic and after creating an instance security groups cannot be changed. Trying to set vpc_security_groups_ids = [] results in the following error...
```
* aws_instance.web: Failed to update vpc_security_group_ids on "i-0f1be7ea20d091c29",
which does not contain a primary network interface
```

Ninir: Is it possible to *create* an instance without security groups?

Jonnymcc: I do not think so, I had not specified one in the template and on creation it had attached our default VPC security group.

radeksimko:
Hi @Jonnymcc 
thanks for sending the PR.

Do you mind attaching an acceptance test which verifies this new patch?

bflad:
@Jonnymcc `make testacc TEST=./aws TESTARGS='-run=TestAccXXX'`

Jonnymcc:
Not sure I wrote the test correctly. How do I verify that the expected error was returned when trying to remove the security group in step 2?

bflad:
```
{
	Config:      testAccInstanceNetworkInstanceVPCSecurityGroupIDs(rInt),
	ExpectError: regexp.MustCompile(`VPC-based instances require at least one security group to be attached`),
},
```

Jonnymcc:
Thanks @bflad!

aws/resource_aws_instance.go
```diff
@@ -893,6 +893,10 @@ func resourceAwsInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
 				groups = append(groups, aws.String(v.(string)))
 			}
 		}
+
+		if len(groups) < 1 {
```
bflad: Minor nitpick: In a future enhancement, we should be able to move this into a `CustomizeDiff` function to perform this validation during planning 👍

aws/resource_aws_instance_test.go
```diff
@@ -701,6 +701,44 @@ func TestAccAWSInstance_NetworkInstanceSecurityGroups(t *testing.T) {
...
+						"aws_instance.foo_instance", &v),
+					resource.TestCheckResourceAttr(
+						"aws_instance.foo_instance", "security_groups.#", "0"),
+					resource.TestCheckResourceAttr(
+						"aws_instance.foo_instance", "vpc_security_group_ids.#", "1"),
+				),
+			},
+			{
+				Config: testAccInstanceNetworkInstanceVPCRemoveSecurityGroupIDs(rInt),
+				Check: resource.ComposeTestCheckFunc(
```
bflad: Minor nitpick: `Check` is not required with `ExpectError`, but I'm surprised that `TestStep` didn't complain. 🤷‍♂️ These extra checks can be removed in a followup PR or better yet it won't matter if a `CustomizeDiff` is handling this at plan time as it won't be possible to apply. 👍 

I verified that `TestStep` did the right thing as an error was returned in the acceptance test:

```
2018/03/09 14:51:28 [ERROR] root: eval: *terraform.EvalApplyPost, err: 1 error(s) occurred:

* aws_instance.foo_instance: VPC-based instances require at least one security group to be attached.
```

bflad:

PR #2851, opened 2018-01-03
data-source/aws_network_interface: Add filter attribute

Gufran:


Gufran:
@Ninir @bflad Any feedback on this change? Is it likely to go out with next release?

aws/data_source_aws_network_interface.go
```diff
@@ -132,20 +133,37 @@ func dataSourceAwsNetworkInterface() *schema.Resource {
 func dataSourceAwsNetworkInterfaceRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ec2conn
 
-	input := &ec2.DescribeNetworkInterfacesInput{
-		NetworkInterfaceIds: []*string{aws.String(d.Get("id").(string))},
+	filters, filtersOk := d.GetOk("filter")
+	id, idOk := d.GetOk("id")
+	if idOk && filtersOk {
+		return fmt.Errorf("Only one of filter or id can be assigned")
```
bflad: Is there a technical reason for this limitation?
* If both `id` and `filter` are supplied and the filter successfully matches the resource corresponding to the ID, its okay and the data source will return the one result.
* If both `id` and `filter` are supplied and the filter doesn't successfully match the resource corresponding to the ID, the data source will return the error.
* Folks might want to create a module with this data source and toggle between both attributes. HCL doesn't support a `nil` value at the moment so this would prevent that use case.

aws/data_source_aws_network_interface.go
```diff
@@ -132,20 +133,37 @@ func dataSourceAwsNetworkInterface() *schema.Resource {
...
+	}
+
+	if !idOk && !filtersOk {
+		return fmt.Errorf("One of id or filter must be assigned")
+	}
+
+	input := &ec2.DescribeNetworkInterfacesInput{}
+	if idOk {
```
bflad: We should probably check to ensure ID actually has a value `if idOk && id.(string) != "" {`

aws/data_source_aws_network_interface_test.go
```diff
@@ -55,3 +56,130 @@ data "aws_network_interface" "test" {
...
+  vpc_id = "${aws_vpc.test.id}"
+}
+
+resource "aws_network_interface" "test" {
+  subnet_id = "${aws_subnet.test.id}"
+  private_ips = ["10.0.0.60"]
+  security_groups = ["${aws_security_group.test.id}"]
+}
+
+data "aws_network_interface" "test" {
```
bflad: This needs `depends_on = ["aws_network_interface.test"]` otherwise it fails due to dependency ordering:

```
--- FAIL: TestAccDataSourceAwsNetworkInterface_filters (47.96s)
	testing.go:503: Step 0 error: Error applying: 1 error(s) occurred:

		* data.aws_network_interface.test: data.aws_network_interface.test: no matching network interface found
```

aws/data_source_aws_network_interface_test.go
```diff
@@ -55,3 +56,130 @@ data "aws_network_interface" "test" {
...
+
+  filter {
+	name = "group-id"
+	values = ["${aws_security_group.test.id}"]
+  }
+}
+    `, rName)
+}
+
+func TestAccDataSourceAwsNetworkInterface_error_exclusive(t *testing.T) {
```
bflad: This test and its configuration can be removed given above comment.

website/docs/d/network_interface.html.markdown
```diff
@@ -22,7 +22,10 @@ data "aws_network_interface" "bar" {
 
 The following arguments are supported:
 
-* `id` – (Required) The identifier for the network interface.
+* `id` – (Optional) The identifier for the network interface.
+* `filter` – (Optional) One or more name/value pairs to filter off of. There are several valid keys, for a full reference, check out [describe-network-interfaces](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-network-interfaces.html) in the AWS CLI reference.
+
+~> **NOTE:** Either `id` or `filter` must be specified. Both `id` and `filter` cannot be specified together.
```
bflad: This note can be removed given above comment.

Gufran:
@bflad I made the changes. and I'm sorry about the problem with tests, not sure if there is a way for me to run tests locally without racking up a bill on AWS.

bflad:
Since you weren't able to run them, I made some minor adjustments to make the acceptance testing happy consistently (mainly setting `id` attribute `Computed: true`). Thanks so much for your work here! Merging! 🚀 

```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsNetworkInterface'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsNetworkInterface -timeout 120m
=== RUN   TestAccDataSourceAwsNetworkInterface_basic
--- PASS: TestAccDataSourceAwsNetworkInterface_basic (33.42s)
=== RUN   TestAccDataSourceAwsNetworkInterface_filters
--- PASS: TestAccDataSourceAwsNetworkInterface_filters (31.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	65.364s
```

bflad:

PR #2857, opened 2018-01-04
r/iam_role: Use ListAttachedRolePoliciesPages

atsushi-ishibashi:
Managed policies may exceed the first response of `ListAttachedRolePolicies`.

### DIFF
--- a/aws/resource_aws_iam_role.go
+++ b/aws/resource_aws_iam_role.go
@@ -267,17 +267,23 @@ func resourceAwsIamRoleDelete(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if d.Get("force_detach_policies").(bool) {
-		policiesResp, err := iamconn.ListAttachedRolePolicies(&iam.ListAttachedRolePoliciesInput{
+		// For managed policies
+		managedPolicies := make([]*string, 0)
+		err = iamconn.ListAttachedRolePoliciesPages(&iam.ListAttachedRolePoliciesInput{
 			RoleName: aws.String(d.Id()),
+		}, func(page *iam.ListAttachedRolePoliciesOutput, lastPage bool) bool {
+			for _, v := range page.AttachedPolicies {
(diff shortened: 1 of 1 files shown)


atsushi-ishibashi:
@Ninir @bflad This is not enhanccement, but bug.

bflad:
@atsushi-ishibashi are you hitting this and missing results? If so, do you have a lot of managed IAM roles or did AWS increase your role policy attachment limit? The [IAM limits documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html#reference_iam-limits-entities) states by default:
> Managed policies attached to an IAM role: 10

I guess you could be hitting the second part of [ListAttachedRolePoliciesInput documentation](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#ListAttachedRolePoliciesInput) for `MaxItems` by default:
```
    // (Optional) Use this only when paginating results to indicate the maximum
    // number of items you want in the response. If additional items exist beyond
    // the maximum you specify, the IsTruncated response element is true.
    //
    // If you do not include this parameter, it defaults to 100. Note that IAM might
    // return fewer results, even when there are more results available. In that
    // case, the IsTruncated response element returns true and Marker contains a
    // value to include in the subsequent call that tells the service where to continue
    // from.
    MaxItems *int64 `min:"1" type:"integer"`
```

Please confirm what you're seeing and we'll prioritize accordingly. Thanks.

atsushi-ishibashi:
@bflad Thanks. Actually I didn't come across the bug but when I implemented #2388, I rememberd that I would fix pagination similar to `ListRolePoliciesPages`.

bflad:
This has been released in [terraform-provider-aws version 1.7.1](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#171-january-19-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

Issue #2842, opened 2018-01-03
Feature Request: Caching for AWS CodeBuild projects


hashibot:
_This issue was originally opened by @kaofelix as hashicorp/terraform#17019. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

AWS CodeBuild projects support caching of dependencies for a build. This allows a user to specify an S3 bucket and a path within that bucket that will be used as cache storage for paths specified in a section of the buildspec YAML file.

I've seen this introduced in this [blog post](https://aws.amazon.com/blogs/devops/how-to-enable-caching-for-aws-codebuild/), where it is shown how to configure a CodeBuild project with caching from the AWS dashboard. The example use case is about Maven dependencies, which is close to my team's use case, albeit for Gradle instead. Using caching would help speed up our build significantly and since we are using terraform to manage our pipeline, it would be great to be able to do it from there.

### Expected Behavior

It would be helpful if  the`aws_codebuild_project` resource supported cache configuration. Maybe something along the lines of:

```hcl
resource "aws_codebuild_project" "foobar" {
  [...]

  cache {
    # s3 is the only option at the moment other than no cache at all, 
    # but there's a dropdown for it, so it might expand
    type = "s3" 
    location = "some-bucket"
    path_prefix = "/some/path"
  }

  [...]
}
```

### Factoids

From the dashboard, it seems that it's only possible to add cache from a newly created project, but not adding it to an existing one when editing. However, boto3 seems to support configuring cache when calling [`update_project`](http://boto3.readthedocs.io/en/latest/reference/services/codebuild.html#CodeBuild.Client.update_project) in an existing project, so I guess it should be possible programatically somehow. I'm not familiar with how this is implemented in Terraform as I didn't have time to look at the source yet.

kaofelix:
I decided to give it a try and managed to code the feature in. I submitted a PR (#2860) for that.

bflad:
This support has been merged into master via #2860 and will release with v1.14.0 of the AWS provider, likely later today. 🎉

PR #2860, opened 2018-01-04
Add cache option to AWS CodeBuild projects

kaofelix:
This is for implementing support for the cache feature I requested in #2842.


nathanielks:
@kaofelix Thanks for putting this together! I was going to add this myself. I'm not a maintainer, but on the last PR I submitted, they asked that validators be put into https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/validators.go. I imagine they'll ask the same thing here!

kaofelix:
Hey @nathanielks, thanks for the heads up! I was not aware of the `validators.go` file. I ended up putting it in there since that's were I found the other validators for CodeBuild projects. Do you think it would make sense to also move the other validators to this file, or would a refactoring like this normally be out of scope for a PR about a more specific feature? It does feel kind of odd to just move my own new validator to that file and leave the others still in `resource_aws_codebuild_project.go`, but I also don't want to go overboard on the changes, what do you think?

nathanielks:
@kaofelix I'd say that's out of scope for this PR. They just had me add my individual validator to the file, as well as move my test to `validators_test.go`. I'm sure they wouldn't mind a separate PR moving them!

nathanielks:
For reference: https://github.com/terraform-providers/terraform-provider-aws/pull/2090#discussion_r155290664

kaofelix:
Cool, @nathanielks! I moved it to the validators file, thanks again for the heads up :)

aws/resource_aws_codebuild_project.go
```diff
@@ -59,6 +59,25 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 				},
 				Set: resourceAwsCodeBuildProjectArtifactsHash,
 			},
+			"cache": {
+				Type:     schema.TypeSet,
```
radeksimko: Since there won't ever be more than 1 `cache` stanza we should make it `TypeList` instead as there's no point in computing the hash for a set and dealing with ordering (as there's nothing to order). Also it will be ~easier~ possible to reference nested fields like `${aws_codebuild_project.cache.0.type}` etc.

aws/resource_aws_codebuild_project.go
```diff
@@ -59,6 +59,25 @@ func resourceAwsCodeBuildProject() *schema.Resource {
...
+							Required:     true,
+							ValidateFunc: validateAwsCodeBuildCacheType,
+						},
+						"location": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+				Set: resourceAwsCodeBuildProjectCacheHash,
```
radeksimko: The `Set` will become unnecessary after turning the parent field into `TypeList`.

aws/resource_aws_codebuild_project.go
```diff
@@ -507,6 +557,24 @@ func flattenAwsCodebuildProjectArtifacts(artifacts *codebuild.ProjectArtifacts)
 	return &artifactSet
 }
 
+func flattenAwsCodebuildProjectCache(cache *codebuild.ProjectCache) []interface{} {
+	values := map[string]interface{}{}
+
+	if cache.Type != nil {
+		if *cache.Type == "NO_CACHE" {
```
radeksimko: Have you considered making `NO_CACHE` the `Default` value of `type` field instead of "magically" turning it into empty string here?

I'm not sure where's the balance between hiding API's implementation details and helping the user to be honest - just food for thought.

kaofelix: I wasn't aware of `Default` so I haven't even tried to go this route. But I do think it sounds a lot cleaner and transparent than what I came up with.

I gave it a go just now, however I couldn't make it work. I can't set `Default` for only the `type`, since `type` is required when `cache` is present. My next try was to give a `Default` to the whole `cache` block, but it won't allow me since it's a list. I figured in this case I should use `DefaultFunc` instead, however I can't make it work. I came up with 

```golang
func defaultAwsCodeBuildProjectCache() (interface{}, error) {
	values := make([]map[string]interface{}, 1)
	values[0]["type"] = "NO_CACHE"
	return values, nil
}
```
and I was getting `Error: aws_codebuild_project.foo: cache.0: expected object, got invalid` when trying to generate a plan containing a codebuild project with no cache defined. So I guessed maybe I need to allocate those objects since the might have been gc'ed or something like, so I tried:

```golang
func defaultAwsCodeBuildProjectCache() (interface{}, error) {
    defaultValue := make([]map[string]interface{}, 1)
    defaultValue[0] = make(map[string]interface{}, 1)
    defaultValue[0]["type"] = "NO_CACHE"
    return defaultValue, nil
}
```
But I'm still getting the same error as above. Any ideas on what might be wrong? Please bear in mind that this is my first real contact with Go whatsoever, so I might be stuck in something very basic :-)

aws/resource_aws_codebuild_project.go
kaofelix: Ok, I've changed it into a List now, seems to work fine. Thanks for the tip!

aws/resource_aws_codebuild_project.go
kaofelix: I gave it another try yesterday and came to the conclusion that it doesn't make sense to pursue this as a default, since "NO_CACHE" is not a user facing value that is going to be used in HCL. The behaviour of Codebuild projects in the AWS side is a bit weird in the sense that if you don't specify Cache you get a project without caching enabled, while once you set a caching option, the only way to remove it is by updating the project with "NO_CACHE" set. I don't think it makes sense to ask from the user that, once they enabled cached for their Codebuild project on Terraform, they must apply a configuration with "NO_CACHE" explicitly set in order for caching to be disabled. If I remove the cache block from my aws_codebuild_project I wouldn't expect for caching to still be enabled.

If anyone is waiting on me to merge, you can go ahead since I'm done with it unless anyone has some more feedback about it :-)

nathanielks:
@kaofelix There are merge conflicts present, would you mind patching those up? Looking forward to this!

kaofelix:
@nathanielks I didn’t have time to do it yesterday, but I’ll do it later today. Thanks for the heads up!

TaiSHiNet:
@nathanielks are we go with this one?

nathanielks:
@TaiSHiNet unfortunately I'm not a core contributor 😅 just helping where I can!

byF:
Any plans to finish this @radeksimko @nathanielks ? :)

nathanielks:
I'm anxiously awaiting this myself!

kaofelix:
I'm solving the conflicts again, also hoping it gets merged soon

barryoneill:
It'd be awesome if someone would review this - I've got some big ivy builds begging for a cache :)

aws/validators.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
 	"github.com/aws/aws-sdk-go/service/configservice"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/gamelift"
```
bflad: Looks like there was some wonky behavior from merge conflicts. I think the validation functions for gamelift/guardduty are not present in this file in master.

aws/resource_aws_codebuild_project_test.go
```diff
@@ -459,6 +464,11 @@ resource "aws_codebuild_project" "foo" {
     type = "NO_ARTIFACTS"
   }
 
+  cache {
```
bflad: Rather than modifying the baseline configuration, we should probably create another acceptance test and configuration(s) that implements these steps:
* Create project without the `cache` configuration
* Adding the `cache` configuration to the project
* Updates the location
* Removes the `cache` configuration from the project

aws/resource_aws_codebuild_project.go
```diff
@@ -59,6 +59,24 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 				},
 				Set: resourceAwsCodeBuildProjectArtifactsHash,
 			},
+			"cache": {
```
bflad: This parent attribute needs to be set as `Computed: true` due to some quirks working with default values in children attributes. Otherwise this shows up as a perpetual difference for anyone who does not define the cache configuration (also failing the acceptance testing):

```
=== RUN   TestAccAWSCodeBuildProject_default_build_timeout
--- FAIL: TestAccAWSCodeBuildProject_default_build_timeout (24.72s)
    testing.go:518: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: aws_codebuild_project.foo
          cache.#: "1" => "0"

--- FAIL: TestAccAWSCodeBuildProject_sourceAuth (24.39s)
    testing.go:518: Step 1 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: aws_codebuild_project.foo
          cache.#: "1" => "0"
```

aws/resource_aws_codebuild_project.go
```diff
@@ -59,6 +59,24 @@ func resourceAwsCodeBuildProject() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateAwsCodeBuildCacheType,
```
bflad: Nitpick: This can be simplified with:
```go
  codebuild.CacheTypeNoCache,
  codebuild.CacheTypeS3,
}, false),
```

aws/resource_aws_codebuild_project.go
kaofelix: hi @bflad! Thanks for your very detailed and helpful feedback. I think I've addressed all of your comments. The new acceptance test I wrote also showed me the effect of lacking the `Computed` flag you mentioned. 

If you have some time, would you mind explaining in a little more detail what this flag is actually doing? I know if I faced this error on my own I would never be able to guess that I should use it, but also if I face something similar in the future I don't wanna feel I'm just cargo culting my way out of it.

bflad: Sorry I do not have time to give a full and well thought out response, but the short of it is that `Computed` tells Terraform to ignore changes to the attribute if it is undefined in the Terraform configuration. It has a side effect of meaning that Terraform cannot reflect drift of a configuration from a default to the operator. Its necessary in this case because we are setting a default for the `type` child attribute where we allow the operator to not define the configuration to not break backwards compatibility.

A better implementation here would be using `CustomizeDiff` to manage the nested attributes so we can properly show drift (e.g. someone enabling S3 caching outside TF, but it not being defined in their configuration). In the interest of not holding up this PR longer, I am going to merge this PR as is. We can remove the `Computed` later.

We will be releasing some public documentation about provider development shortly that will hopefully clear up the usage and caveats with this flag.

bflad:

TaiSHiNet:
Cheers! Thanks a lot!

ebarault:
Many thanks @kaofelix for your efforts ! 🎉

catsby:
Hi there! I'm afraid we need to revisit this `NO_CACHE` default idea. See https://github.com/terraform-providers/terraform-provider-aws/issues/4113 . 

> since "NO_CACHE" is not a user facing value that is going to be used in HCL

I don't agree here, because... 

> If I remove the cache block from my aws_codebuild_project I wouldn't expect for caching to still be enabled.

Unfortunately `cache` is an `Optional` + `Computed` + `MaxItems: 1` attribute, and tricky trifecta, which means if you apply with it and then subsequently remove it, Terraform cannot distinguish between "remove this thing" from "I no longer need to care about tracking it's changes". Terraform then doesn't prompt remove the attribute. Assuming I'm reading the code correctly....

bflad:
Followup PR submitted: #4134

PR #2869, opened 2018-01-05
resource/aws_db_instance: Force update when snapshot_identifier is set

### LINKED ISSUE #2859: aws_db_instance parameter_group_name Ignored
_This issue was originally opened by @dgarstang as hashicorp/terraform#17031. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Following code:

```
resource "aws_db_instance" "alpha" {
  allocated_storage    = 300
  storage_type         = "gp2"
  engine               = "mysql"
  engine_version       = "5.5.46"
  instance_class       = "db.m4.large"
  identifier           = "${var.name}"
  db_subnet_group_name = "default"
  parameter_group_name = "allow-triggers"
  snapshot_identifier  = "${var.rds_snapshot_name}"
  skip_final_snapshot  = true
  tags = {
    Name = "${var.name}"
  }
}
```

When running apply, output displays:
```
parameter_group_name:                           "allow-triggers"
```

However, when I look at the RDS instance in the AWS cli, it shows:
```
Parameter group
default.mysql5.5 (in-sync)
```

Seems that parameter_group_name is completely ignored.
Terraform v0.11.1

atsushi-ishibashi:
Closes #2859 

If there is a reason why update was executed only when `password, vpc_security_group_ids, security_group_names` were set, please correct me🙇

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -522,8 +522,7 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			return err
 		}
 
-		err = resourceAwsDbInstanceUpdate(d, meta)
-		if err != nil {
+		if err = resourceAwsDbInstanceUpdate(d, meta); err != nil {
 			return err
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


atsushi-ishibashi:
@radeksimko Do you have any ideas of an acceptance test for snapshot?
I'm thinking that I make 2 test steps, the first creating db_instance with specifying `final_snapshot_identifier` and the second restoring db_instance with the `snapshot_identifier`.
How do you think?

PR #1608, opened 2017-09-07
Add support in TF file configuration for assuming roles via profiles defined in ~/.aws/config

mikemoate:
This is simply a port of https://github.com/hashicorp/terraform/pull/11734 to the new terraform-provider-aws repo. (the original PR won't be merged since the code it relates to has been moved to this repo).

This should address #1184.

This won't support assuming roles where an MFA device has been specified for the profile (in ~/.aws/config). The original PR has some discussion of this, however, Terraform seems to utilise multiple, simultaneous sessions with AWS which complicates the situation somewhat. (I certainly saw repeated prompts for MFA when testing adding this)

I can't see much in the way of test code in config_test.go so am not immediately sure where and how to test this successfuly.

Original work by @craiglink in https://github.com/hashicorp/terraform/pull/11734.

PR #2883, opened 2018-01-05
provider: Fallback to session-derived credentials

### LINKED ISSUE #1184: Using terraform with assumed roles defined in shared configuration are ineffective

twang-rs:
#1608 allows for the session to figure out credentials by setting `opt.SharedConfigState = session.SharedConfigEnable` whenever the Profile is not empty.

This appropriately allows shared config (`~/.aws/config`) to work correctly (including using assumed roles) _iff_ the `profile` is set within the Terraform provider configuration.

This does **not** work correctly in the case of an assumed role when relying on environment variables `AWS_PROFILE` and `AWS_SDK_LOAD_CONFIG`.

This patch alters the behavior to the following:

- We still use the existing method of manually creating a credential chain and attempting to load credentials via a mix of explicit configuration and environment variables found in auth_helpers.go `GetCredentials`
- If this fails, we create a default session object and attempt to extract credentials from AWS credential code found in the session.  If this is successful, we use that credential.
- Otherwise, we bomb out with an error message.

I am a little worried that assumed role (or any form of temporary credential) does not seem to have the capability to auto-renew.  I'm not sure how much of a problem this is, as terraform plans are (generally) not long-lived processes.

This should effectively resolve #1184.
  
NOTE: we'll need to bump terraform proper (as in hashicorp/terraform#16661) again, if this PR is accepted/merged.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -265,16 +265,21 @@ func (c *Config) Client() (interface{}, error) {
 	cp, err := creds.Get()
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NoCredentialProviders" {
-			// If a profile wasn't specified then error out
-			if c.Profile == "" {
+			// The session may still be able to resolve credentials from shared config.
+			sess, err := session.NewSession()
+			if err != nil {
 				return nil, errors.New(`No valid credential sources found for AWS Provider.
   Please see https://terraform.io/docs/providers/aws/index.html for more information on
   providing credentials for the AWS Provider`)
 			}
-			// add the profile and enable share config file usage
-			log.Printf("[INFO] AWS Auth using Profile: %q", c.Profile)
-			opt.Profile = c.Profile
-			opt.SharedConfigState = session.SharedConfigEnable
+			_, err = sess.Config.Credentials.Get()
+			if err != nil {
+				return nil, errors.New(`No valid credential sources found for AWS Provider.
+  Please see https://terraform.io/docs/providers/aws/index.html for more information on
+  providing credentials for the AWS Provider`)
+			}
+			log.Printf("[INFO] Using session-derived AWS Auth")
+			opt.Config.Credentials = sess.Config.Credentials
 		} else {
 			return nil, fmt.Errorf("Error loading credentials for AWS Provider: %s", err)
 		}
--- a/aws/config.go
+++ b/aws/config.go
@@ -265,21 +265,27 @@ func (c *Config) Client() (interface{}, error) {
 	cp, err := creds.Get()
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NoCredentialProviders" {
-			// The session may still be able to resolve credentials from shared config.
-			sess, err := session.NewSession()
-			if err != nil {
-				return nil, errors.New(`No valid credential sources found for AWS Provider.
-  Please see https://terraform.io/docs/providers/aws/index.html for more information on
-  providing credentials for the AWS Provider`)
+			// If a profile wasn't specified, the session may still be able to resolve credentials from shared config.
+			if c.Profile == "" {
+				sess, err := session.NewSession()
+				if err != nil {
+					return nil, errors.New(`No valid credential sources found for AWS Provider.
+	Please see https://terraform.io/docs/providers/aws/index.html for more information on
+	providing credentials for the AWS Provider`)
+				}
+				_, err = sess.Config.Credentials.Get()
+				if err != nil {
+					return nil, errors.New(`No valid credential sources found for AWS Provider.
+	Please see https://terraform.io/docs/providers/aws/index.html for more information on
+	providing credentials for the AWS Provider`)
+				}
+				log.Printf("[INFO] Using session-derived AWS Auth")
+				opt.Config.Credentials = sess.Config.Credentials
+			} else {
+				log.Printf("[INFO] AWS Auth using Profile: %q", c.Profile)
+				opt.Profile = c.Profile
+				opt.SharedConfigState = session.SharedConfigEnable
 			}
-			_, err = sess.Config.Credentials.Get()
-			if err != nil {
-				return nil, errors.New(`No valid credential sources found for AWS Provider.
-  Please see https://terraform.io/docs/providers/aws/index.html for more information on
-  providing credentials for the AWS Provider`)
-			}
-			log.Printf("[INFO] Using session-derived AWS Auth")
-			opt.Config.Credentials = sess.Config.Credentials
 		} else {
 			return nil, fmt.Errorf("Error loading credentials for AWS Provider: %s", err)
 		}


jen20:
Hi @tamsky! Thanks for opening a pull request. I'm going to refer this one to @catsby since it needs more than a quick merge, but I'll offer my :+1: here - I believe it will also resolve the issue with ECS task roles that have been reported in a few different places.

mikemoate:
I've built terraform (which includes the terraform-provider-aws source directly) and the provider with this change on macOS using https://gist.github.com/mikemoate/99776bee1c0b161f5bc2db207c771727

Testing locally shows that this change breaks the behaviour I fixed in #1608, specifically the following configuration no longer works:

```
// main.tf

terraform {
  backend "s3" {
    bucket  = "some-bucket-name"
    key     = "some-bucket-key/some-state.tfstate"
    profile = "some_profile"
    region  = "eu-west-2"
    acl     = "some-bucket-acl"
  }
}

provider "aws" {
  profile = "some_profile"
  region  = "eu-west-2"
}
```

```
// ~/.aws/config

[default]
output = json
region = eu-west-2

[profile some_profile]
role_arn = arn:aws:iam::123456789012:role/some_role
source_profile = default
```

```
// ~/.aws/credentials

[default]
aws_access_key_id=SOMEACCESSKEY
aws_secret_access_key=SomeSecretAccessKey
```
`terraform init` gives the following error:
```
$ terraform init
Initializing modules...
- module.app_db

Initializing the backend...
Error loading state: AccessDenied: Access Denied
	status code: 403, request id: 53B359144E5BCDA6, host id: BWnGEXTGRp4+htxEcrhvDsRqvJNXXPT+N1fNqlAxt03apSQUHLWfCIl2Ox7yBVdNwPacKzm0RA4=
```

I suspect this is related to the fact that `opt.SharedConfigState = session.SharedConfigEnable` is no longer set with this change, though have not had chance to debug further.

As it stands I (unsurprisingly) would not be in favour of merging this PR.

twang-rs:
@mikemoate: Thanks for the report. I'll take a look this afternoon and update the PR.

twang-rs:
@mikemoate: pushed up a change the restores functionality in #1608.  Did some preliminary testing and it seemed good, let me know if satisfies your use case.

twang-rs:
Any more feedback on the PR?

mikemoate:
@twang-rs I've re-tested our use case (that was originally fixed in #1608) with this latest change, and it now works for us, so we're in favour of the PR. Apologies for the delay in getting to this.

mikemoate:
@catsby or @Ninir any chance we could get this PR reviewed and merged ASAP?

It fixes a valid problem pointed out in https://github.com/terraform-providers/terraform-provider-aws/issues/1184 (which is becoming a highly trafficked thread) and improves on my original changes under https://github.com/terraform-providers/terraform-provider-aws/pull/1608.

bflad:
> I am a little worried that assumed role (or any form of temporary credential) does not seem to have the capability to auto-renew. I'm not sure how much of a problem this is, as terraform plans are (generally) not long-lived processes.

You're correct and we've run into this before, especially in large scale environments. Hopefully the addition of max session duration into IAM roles can help mitigate this until we can come up with a proper provider-wide fix. At first blush this pull request seems good so I'll probably pull it in.

bflad:

PR #2889, opened 2018-01-08
Switch aws_appautoscaling_target's role_arn to optional

### LINKED ISSUE #2750: aws_appautoscaling_target.RoleArn should be optional
### Terraform Version
0.11.0

### Affected Resource(s)
- aws_appautoscaling_target 

```hcl
resource "aws_appautoscaling_target" "dynamodb_table_read_target" {
  max_capacity       = 100
  min_capacity       = 5
  resource_id        = "table/tableName"
  role_arn           = "${data.aws_iam_role.DynamoDBAutoscaleRole.arn}"
  scalable_dimension = "dynamodb:table:ReadCapacityUnits"
  service_namespace  = "dynamodb"
}
```

### Expected Behavior
role_arn is set to appriopriate role for that autoscaling_target

### Actual Behavior
role_arn is set to 'AWSServiceRoleForApplicationAutoScaling_DynamoDBTable' - this causes this future deployments to *always* try and update this autoscaling target

### Steps to Reproduce
1. ensure us-east-1
1. `terraform apply`

### Important Factoids
I've been talking AWS service team about this issue as it also happened through the AWS Console.  Turns out this is a new "feature" currently only active on "us-east-1".  Basically it creates it own role regardless of what you specify.  See http://docs.aws.amazon.com/ApplicationAutoScaling/latest/APIReference/application-autoscaling-service-linked-roles.html

I'm fine either way but RoleArn is 'required' by the terraform resource.  I've worked around it by hardcoding the RoleArn to when deploying to 'us-east-1'.  Probably should makes this property optional (it's optional in AWS CLI btw)

armanshan12:
fixes #2750

### DIFF
--- a/aws/resource_aws_appautoscaling_target_test.go
+++ b/aws/resource_aws_appautoscaling_target_test.go
@@ -217,58 +217,6 @@ func testAccCheckAWSAppautoscalingTargetExists(n string, target *applicationauto
 func testAccAWSAppautoscalingTargetConfig(
 	randClusterName string) string {
 	return fmt.Sprintf(`
-resource "aws_iam_role" "autoscale_role" {
-	name = "autoscalerole%s"
-	path = "/"
-
-	assume_role_policy = <<EOF
-{
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Effect": "Allow",
-      "Principal": {
-        "Service": "application-autoscaling.amazonaws.com"
-      },
-      "Action": "sts:AssumeRole"
-    }
-  ]
-}
-EOF
-}
-
-resource "aws_iam_role_policy" "autoscale_role_policy" {
-	name = "autoscalepolicy%s"
-	role = "${aws_iam_role.autoscale_role.id}"
-
-	policy = <<EOF
-{
-    "Version": "2012-10-17",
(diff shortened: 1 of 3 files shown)


aws/resource_aws_appautoscaling_target_test.go
```diff
@@ -192,58 +217,6 @@ func testAccCheckAWSAppautoscalingTargetExists(n string, target *applicationauto
 func testAccAWSAppautoscalingTargetConfig(
 	randClusterName string) string {
 	return fmt.Sprintf(`
-resource "aws_iam_role" "autoscale_role" {
```
radeksimko: It's great to have a test which tests the new behaviour (i.e. default role assigned by AWS), but we should keep existing tests as is - setting custom role should still be working and be tested the same way as before.

armanshan12:
Hi @radeksimko,

Regarding your concern about testing custom `role_arn`, `TestAccAWSAppautoScalingTarget_spotFleetRequest` continues to use its own `role_arn` and it passes successfully. AWS might break this one as well in the future, but for now, this test verifies the custom `role_arn` functionality.

bflad:

PR #2246, opened 2017-11-10
Fix missing negation when validating Cognito RoleMapping of type token

makapuu:
Adds the missing negation as described in issue #2232 , because otherwise `ok` always leads to an error in case of `RoleMappingTypeToken `.

PR #1803, opened 2017-10-03
Add 'aws_cognito_user_pool_client' resource

psyvision:
## Description

The following pull request builds upon the work of @Ninir implementing [Cognito User Pools](https://github.com/terraform-providers/terraform-provider-aws/pull/1419) (not yet merged) by adding support for Cognito User Pool Clients.

I've added some documentation on using the new resource.

Note: This doesn't provide full API support at current but does allow a client to be created and the `ClientId` returned.

## API Documentation 

- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPoolClient.html

## Test

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSCognitoUserPoolClient_'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoUserPoolClient_ -timeout 120m
=== RUN   TestAccAWSCognitoUserPoolClient_basic
--- PASS: TestAccAWSCognitoUserPoolClient_basic (19.11s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       19.126s
```

## Resources

```
resource "aws_cognito_user_pool_client" "client" {
}
```

PR #2894, opened 2018-01-08
Fix Cognito RoleMapping validation for Token types

tomelliff:
If the mapping_rule isn't set then this should be the equivalent of it not having any mapping rules defined.
Looks like the negation was missed on this.

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/2232

### DIFF
--- a/aws/validators.go
+++ b/aws/validators.go
@@ -1891,7 +1891,7 @@ func validateCognitoRoleMappingsRulesConfiguration(v map[string]interface{}) (er
 		errors = append(errors, fmt.Errorf("mapping_rule is required for Rules"))
 	}
 
-	if (ok || valLength > 0) && t == cognitoidentity.RoleMappingTypeToken {
+	if (!ok || valLength > 0) && t == cognitoidentity.RoleMappingTypeToken {
 		errors = append(errors, fmt.Errorf("mapping_rule must not be set for Token based role mapping"))
 	}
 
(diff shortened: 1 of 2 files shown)


tomelliff:
Just seen another pull request for this after checking for the issue but it was missing tests: https://github.com/terraform-providers/terraform-provider-aws/pull/2246.

Happy to rebase this on the previous PR so that @makapuu's work can be accepted too.

Also, I wanted to add a wider acceptance test for `aws_cognito_identity_pool_roles_attachment` so that we created a token based role mapping there (there's an acceptance test for the error https://github.com/terraform-providers/terraform-provider-aws/blob/79ed212c594fd8e848d0274cd0efc70bbf4cede8/aws/resource_aws_cognito_identity_pool_roles_attachment_test.go#L126-L140) but token based role mappings can only be done with Cognito identity providers and that requires an app client so we'd need https://github.com/terraform-providers/terraform-provider-aws/pull/2874 to be merged first.

bflad:
@tomelliff thanks for your work here. Given https://github.com/terraform-providers/terraform-provider-aws/pull/2246#issuecomment-355774533 and #1803 being merged could you rebase against #2246 and help finish this up? We'd greatly appreciate it. 😄  🚀  :shipit:

tomelliff:
@bflad I've rebased against that commit so it's now included along with my further changes plus the test.

Let me know if you need anything else.

tomelliff:
@bflad Did you need anything else from me on this? Would be great if we can get this fix in for the next release because then I think I can tear out my Cloudformation workarounds for what was the missing Terraform Cognito resources plus this bug.

bflad:
@tomelliff thanks for the gentle ping - I missed coming back to this for you. Sorry about that. 😢  Taking a look now.

bflad:

Issue #2384, opened 2017-11-20
aws_cloudwatch_log_metric_filter: default_value is automatically set to 0


iamam34:
When I apply a terraform configuration that does NOT include default_value, terraform creates a log filter in AWS that has the default_value set to 0.

### Terraform Version
Terraform v0.10.3 with aws provider v1.3.1

### Affected Resource(s)
- aws_cloudwatch_log_metric_filter

### Terraform Configuration Files
```hcl
resource "aws_cloudwatch_log_metric_filter" "my-function-memory-used" {
  name           = "MyFunction_MemoryUsed"
  pattern        = "[type=REPORT, ... , memory_used, unit]"
  log_group_name = "/aws/lambda/MyFunction"

  metric_transformation {
    name      = "MemoryUsed"
    namespace = "MyFunction.CustomMetrics"
    value     = "$memory_used"
  }
}
```

### Expected Behavior
Creates a log filter with default_value **not set**, so that log lines that don't match the filter don't output any value for the metric. http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/MonitoringLogData.html

### Actual Behavior
Creates a log filter with default_value **set to 0**, so that every log line outputs a value for the metric (non-matching lines output 0). 

### Steps to Reproduce
1. Write terraform for a log_filter WITHOUT a default_value
1. `terraform apply` (note that the console output does not mention default_value at all)
    ```
    aws_cloudwatch_log_metric_filter.my-function-memory-used: Creating...
    [11:12:27][Step 2/2]   log_group_name:                    "" => "/aws/lambda/MyFunction"
    [11:12:27][Step 2/2]   metric_transformation.#:           "" => "1"
    [11:12:27][Step 2/2]   metric_transformation.0.name:      "" => "MemoryUsed"
    [11:12:27][Step 2/2]   metric_transformation.0.namespace: "" => "MyFunction.CustomMetrics"
    [11:12:27][Step 2/2]   metric_transformation.0.value:     "" => "$memory_used"
    [11:12:27][Step 2/2]   name:                              "" => "MyFunction_MemoryUsed"
    [11:12:27][Step 2/2]   pattern:                           "" => "[type=REPORT, ... , memory_used, unit]"
    ```
1. Visit the AWS Console CloudWatch and see that your new metric has a minimum value of 0 and an unexpectedly low average. 

### References
- #1578 - this PR from a month ago added support for the optional default_value

iamam34:
[...]

PR #2902, opened 2018-01-09
[WIP] resource/aws_cloudwatch_log_metric_filter: Unable to keep default value unset

modax:
Hello,

there is a bug in r/cloudwatch_log_metric_filter where `metric_transformation/default_value` gets set to 0 via API even if it is missing in the tf code like:

```hcl
resource "aws_cloudwatch_log_metric_filter" "foobar" {
  name = "MyAppAccessCount-%d"
  pattern = ""
  log_group_name = "${aws_cloudwatch_log_group.dada.name}"

  metric_transformation {
  	name = "EventCount"
  	namespace = "YourNamespace"
  	value = "1"
  }
}
```

`$ aws logs describe-metric-filters --region us-west-2`
```
{
    "metricFilters": [
        {
            "filterName": "MyAppAccessCount-11111111",
            "metricTransformations": [
                {
                    "metricName": "EventCount",
                    "metricNamespace": "YourNamespace",
                    "metricValue": "1",
                    "defaultValue": 0.0
                }
            ],
            "creationTime": 1515482147102,
            "logGroupName": "MyApp/access-11111.log"
        }
    ]
}
```

In this WIP PR so far I have improved acc tests to fail this case:

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSCloudWatchLogMetricFilter_' 
TF_ACC=1 go test ./aws -v -run=TestAccAWSCloudWatchLogMetricFilter_ -timeout 120m
=== RUN   TestAccAWSCloudWatchLogMetricFilter_basic
--- FAIL: TestAccAWSCloudWatchLogMetricFilter_basic (27.05s)
        testing.go:503: Step 0 error: Check failed: Check 10/10 error: Expected default value to be present: false, received: true
FAIL
exit status 1
FAIL    github.com/terraform-providers/terraform-provider-aws/aws       27.079s
GNUmakefile:20: recipe for target 'testacc' failed
make: *** [testacc] Error 1
```

However, I have failed to resolve this bug since it appears there is some complicated magic involved here which always sets golangs Zero value on `TypeFloat` parameter which is `Optional` and non-present, e.g. here: 

https://github.com/hashicorp/terraform/blame/master/helper/schema/field_reader.go#L242

There is similar magic which converts `nil` or entirely missing parameter in data map to `0` during either during [d.Set()](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_cloudwatch_log_metric_filter.go#L132) or during serialization to tfstate, just not sure.

Anyway, I'm lost and I seeking your advice here how to do solve this properly. To be honest, this implicit zero value trend is one of the most annoying "features" of golang in my opinion since it can be so unpredictable sometimes like this case  :/

### DIFF
--- a/aws/resource_aws_cloudwatch_log_metric_filter_test.go
+++ b/aws/resource_aws_cloudwatch_log_metric_filter_test.go
@@ -56,6 +56,7 @@ func TestAccAWSCloudWatchLogMetricFilter_basic(t *testing.T) {
 						MetricName:      aws.String("AccessDeniedCount"),
 						MetricNamespace: aws.String("MyNamespace"),
 						MetricValue:     aws.String("2"),
+						DefaultValue:    aws.Float64(1),
 					}),
 				),
 			},
@@ -107,6 +108,14 @@ func testAccCheckCloudWatchLogMetricFilterTransformation(mf *cloudwatchlogs.Metr
 		if *given.MetricValue != *expected.MetricValue {
 			return fmt.Errorf("Expected metric value: %q, received: %q",
 				*expected.MetricValue, *given.MetricValue)
+		}
+
+		if (given.DefaultValue != nil) != (expected.DefaultValue != nil) {
+			return fmt.Errorf("Expected default value to be present: %t, received: %t",
+				expected.DefaultValue != nil, given.DefaultValue != nil)
+		} else if (given.DefaultValue != nil) && *given.DefaultValue != *expected.DefaultValue {
+			return fmt.Errorf("Expected metric value: %g, received: %g",
+				*expected.DefaultValue, *given.DefaultValue)
 		}
 
 		return nil
--- a/aws/structure.go
+++ b/aws/structure.go
@@ -1541,7 +1541,7 @@ func expandCloudWachLogMetricTransformations(m map[string]interface{}) ([]*cloud
 		MetricValue:     aws.String(m["value"].(string)),
 	}
 
-	if m["default_value"] != "" {
+	if m["default_value"] != nil {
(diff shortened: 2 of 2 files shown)


92kz1h:
This appears to be a duplicate of #2384 (from #3131).

I'm not sure which PR is preferred.

jocelynthode:
Yes this pr is a duplicate and apparently you're stuck exactly on the same problem as I am. Unfortunately it's been more than one month and I have failed to get the attention of any maintainer on this issue.

Please feel free to take a look at my PR ans submit a commit to my branch if you feel that my tests are not thorough enough

92kz1h:
Sorry, as much as I would like to help on the issue. I'm not familiar with Go either and this `nil`/`0` casting issue is a bit beyond me.

sbmarcos:
ETA on this? Thanks much!

bflad:
Hi @modax 👋 Very sorry that this pull request got missed previously and that we weren't able to provide a good answer to help move this forward when it was submitted (or shortly after). I have been actively working with our product team to come up with a good solution to reduce the open issue and pull request count, since its such a moving target and often means older issues can get lost in the shuffle.

The good news here is that we have just merged #5933 which does contains the fix (a fairly odd workaround we have used in a few other places to convert the attribute types to `schema.TypeString` when we want to support `""` as a different empty value).

There are some changes in the upcoming Terraform 0.12 release relating to a richer type system that might be able to help in these cases. It will take a little bit for the Terraform provider SDK to be updated in this provider (effectively removing support for 0.11 and earlier in some places), but hopefully problems like these can be made a thing of the past.

Thanks again for your submission and apologies again for the miss on our part here.

Issue #2489, opened 2017-11-30
Add GuardDuty Support


bflad:
AWS has announced support for managing a new intelligent threat detection service called GuardDuty: https://aws.amazon.com/about-aws/whats-new/2017/11/announcing-amazon-guardduty-intelligent-threat-detection/

Prerequisite: aws-sdk-go v1.12.36 (#2474)

### Terraform Version

```
terraform 0.10+
terraform-provider-aws 1.5.0
```

### Affected Resource(s)

These are new resources.

- `aws_guardduty_detector`
- `aws_guardduty_invite`
- `aws_guardduty_invite_accepter` (similar to `aws_vpc_peering_connection_accepter`)
- `aws_guardduty_ipset`
- `aws_guardduty_member`
- `aws_guardduty_threatintelset`

### Expected Behavior

Create, update, delete, and import GuardDuty resources. e.g.

```hcl
# Enable GuardDuty

resource "aws_guardduty_detector" "master" {
  enable = true
}

# Manage IPSets/ThreatIntelSets

resource "aws_s3_bucket" "bucket" {
  acl = "private"
}

resource "aws_s3_bucket_object" "MyIPSet" {
  acl     = "public-read"
  content = "10.0.0.0/8\n"
  bucket  = "${aws_s3_bucket.bucket.id}"
  key     = "MyIPSet"
}

resource "aws_s3_bucket_object" "MyThreatIntelSet" {
  acl     = "public-read"
  content = "192.168.1.1/32\n"
  bucket  = "${aws_s3_bucket.bucket.id}"
  key     = "MyThreatIntelSet"
}

resource "aws_guardduty_ipset" "MyIPSet" {
  activate    = true
  detector_id = "${aws_guardduty_detector.master.id}"
  format      = "TXT"
  location    = "https://s3.amazonaws.com/${aws_s3_bucket_object.MyIPSet.bucket}/${aws_s3_bucket_object.MyIPSet.key}"
  name        = "MyIPSet"
}

resource "aws_guardduty_threatintelset" "MyThreatIntelSet" {
  activate    = true
  detector_id = "${aws_guardduty_detector.master.id}"
  format      = "TXT"
  location    = "https://s3.amazonaws.com/${aws_s3_bucket_object.MyThreatIntelSet.bucket}/${aws_s3_bucket_object.MyThreatIntelSet.key}"
  name        = "MyThreatIntelSet"
}

# Monitor GuardDuty in another account

[...]

PR #2911, opened 2018-01-09
New Resource: aws_guardduty_member

bflad:
References:
* Reference specification: #2489 
* Go docs: https://docs.aws.amazon.com/sdk-for-go/api/service/guardduty/
* API docs: https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_api_ref.html

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSGuardDuty'
TF_ACC=1 go test ./aws -v -run=TestAccAWSGuardDuty -timeout 120m
=== RUN   TestAccAWSGuardDuty
=== RUN   TestAccAWSGuardDuty/Detector
=== RUN   TestAccAWSGuardDuty/Detector/basic
=== RUN   TestAccAWSGuardDuty/Detector/import
=== RUN   TestAccAWSGuardDuty/Member
=== RUN   TestAccAWSGuardDuty/Member/basic
=== RUN   TestAccAWSGuardDuty/Member/import
--- PASS: TestAccAWSGuardDuty (77.94s)
    --- PASS: TestAccAWSGuardDuty/Detector (47.27s)
        --- PASS: TestAccAWSGuardDuty/Detector/basic (33.38s)
        --- PASS: TestAccAWSGuardDuty/Detector/import (13.89s)
    --- PASS: TestAccAWSGuardDuty/Member (30.66s)
        --- PASS: TestAccAWSGuardDuty/Member/basic (15.09s)
        --- PASS: TestAccAWSGuardDuty/Member/import (15.58s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	77.977s
```

### DIFF
--- a/aws/resource_aws_guardduty_member.go
+++ b/aws/resource_aws_guardduty_member.go
@@ -67,9 +67,10 @@ func resourceAwsGuardDutyMemberCreate(d *schema.ResourceData, meta interface{}) 
 func resourceAwsGuardDutyMemberRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).guarddutyconn
 
-	idParts := strings.Split(d.Id(), ":")
-	accountID := idParts[1]
-	detectorID := idParts[0]
+	accountID, detectorID, err := decodeGuardDutyMemberID(d.Id())
+	if err != nil {
+		return err
+	}
 
 	input := guardduty.GetMembersInput{
 		AccountIds: []*string{aws.String(accountID)},
@@ -93,9 +94,9 @@ func resourceAwsGuardDutyMemberRead(d *schema.ResourceData, meta interface{}) er
 		return nil
 	}
 	member := gmo.Members[0]
-	d.Set("account_id", *member.AccountId)
+	d.Set("account_id", member.AccountId)
 	d.Set("detector_id", detectorID)
-	d.Set("email", *member.Email)
+	d.Set("email", member.Email)
 
 	return nil
 }
@@ -103,9 +104,10 @@ func resourceAwsGuardDutyMemberRead(d *schema.ResourceData, meta interface{}) er
 func resourceAwsGuardDutyMemberDelete(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).guarddutyconn
 
-	idParts := strings.Split(d.Id(), ":")
-	accountID := idParts[1]
-	detectorID := idParts[0]
+	accountID, detectorID, err := decodeGuardDutyMemberID(d.Id())
+	if err != nil {
+		return err
+	}
 
 	input := guardduty.DeleteMembersInput{
 		AccountIds: []*string{aws.String(accountID)},
@@ -113,9 +115,20 @@ func resourceAwsGuardDutyMemberDelete(d *schema.ResourceData, meta interface{}) 
 	}
 
 	log.Printf("[DEBUG] Delete GuardDuty Member: %s", input)
-	_, err := conn.DeleteMembers(&input)
+	_, err = conn.DeleteMembers(&input)
 	if err != nil {
 		return fmt.Errorf("Deleting GuardDuty Member '%s' failed: %s", d.Id(), err.Error())
 	}
 	return nil
+}
+
+func decodeGuardDutyMemberID(id string) (accountID, detectorID string, err error) {
+	parts := strings.Split(id, ":")
+	if len(parts) != 2 {
+		err = fmt.Errorf("GuardDuty Member ID must be of the form <Detector ID>:<Member AWS Account ID>")
+		return
+	}
+	accountID = parts[1]
+	detectorID = parts[0]
+	return
 }
--- a/aws/resource_aws_guardduty_member.go
+++ b/aws/resource_aws_guardduty_member.go
@@ -1,0 +1,121 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/guardduty"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsGuardDutyMember() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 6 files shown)


aws/resource_aws_guardduty_member.go
```diff
@@ -0,0 +1,121 @@
...
+
+	input := guardduty.GetMembersInput{
+		AccountIds: []*string{aws.String(accountID)},
+		DetectorId: aws.String(detectorID),
+	}
+
+	log.Printf("[DEBUG] Reading GuardDuty Member: %s", input)
+	gmo, err := conn.GetMembers(&input)
+	if err != nil {
+		if isAWSErr(err, guardduty.ErrCodeBadRequestException, "The request is rejected because the input detectorId is not owned by the current account.") {
```
radeksimko: 😭

aws/resource_aws_guardduty_member.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if gmo.Members == nil || (len(gmo.Members) < 1) {
+		log.Printf("[WARN] GuardDuty Member %q not found, removing from state", d.Id())
+		d.SetId("")
+		return nil
+	}
+	member := gmo.Members[0]
+	d.Set("account_id", *member.AccountId)
+	d.Set("detector_id", detectorID)
+	d.Set("email", *member.Email)
```
radeksimko: Dereferencing here is not necessary, or in fact undesired. `Set()` will perform _safe_ dereferencing - meaning that it won't crash on `nil`. We had many crash reports caused by this in the past which led us to this thought.

Do you mind changing it accordingly? 😉

aws/resource_aws_guardduty_member.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return resourceAwsGuardDutyMemberRead(d, meta)
+}
+
+func resourceAwsGuardDutyMemberRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).guarddutyconn
+
+	idParts := strings.Split(d.Id(), ":")
+	accountID := idParts[1]
+	detectorID := idParts[0]
```
radeksimko: Q: What happens if the user passes only one ID? Isn't it going to crash here because `Split()` won't return `[1]`?

I think it would be nice to have a helper function for this, e.g. `accId, detectorId, err := decodeGuardDutyMemberId(d.Id())` and keep the error message there.

aws/resource_aws_guardduty_test.go
```diff
@@ -10,6 +10,10 @@ func TestAccAWSGuardDuty(t *testing.T) {
 			"basic":  testAccAwsGuardDutyDetector_basic,
 			"import": testAccAwsGuardDutyDetector_import,
 		},
+		"Member": {
+			"basic":  testAccAwsGuardDutyMember_basic,
+			"import": testAccAwsGuardDutyMember_import,
+		},
```
radeksimko: I assume there can be more than 1 member per AWS account, so we don't need to make these tests synchronous? Unless I'm mistaken...

aws/resource_aws_guardduty_member.go
bflad: Yeah its like S3 `NoSuchBucket` in this regard. 😢

aws/resource_aws_guardduty_member.go
bflad: Oi! This is a very good catch! I totally forgot to split it out into a decode function with the error checking after initially writing it.

aws/resource_aws_guardduty_member.go
bflad: 💯  I'll update this code and personally file this under "leave the stars to the night sky"!

aws/resource_aws_guardduty_test.go
bflad: This needs to be like this because its like AWS Config, where it can only be activated once in a region, so we're left with serially doing one test at a time since we turn the GuardDuty detector on and off each test.

bflad:
Updated and verified locally! Thanks for those very important catches. 🚀 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSGuardDuty'
TF_ACC=1 go test ./aws -v -run=TestAccAWSGuardDuty -timeout 120m
=== RUN   TestAccAWSGuardDuty
=== RUN   TestAccAWSGuardDuty/Detector
=== RUN   TestAccAWSGuardDuty/Detector/basic
=== RUN   TestAccAWSGuardDuty/Detector/import
=== RUN   TestAccAWSGuardDuty/Member
=== RUN   TestAccAWSGuardDuty/Member/basic
=== RUN   TestAccAWSGuardDuty/Member/import
--- PASS: TestAccAWSGuardDuty (53.58s)
    --- PASS: TestAccAWSGuardDuty/Detector (31.83s)
        --- PASS: TestAccAWSGuardDuty/Detector/basic (20.85s)
        --- PASS: TestAccAWSGuardDuty/Detector/import (10.97s)
    --- PASS: TestAccAWSGuardDuty/Member (21.75s)
        --- PASS: TestAccAWSGuardDuty/Member/basic (10.05s)
        --- PASS: TestAccAWSGuardDuty/Member/import (11.70s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	53.620s
```

aws/resource_aws_guardduty_member.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return fmt.Errorf("Deleting GuardDuty Member '%s' failed: %s", d.Id(), err.Error())
+	}
+	return nil
+}
+
+func decodeGuardDutyMemberID(id string) (accountID, detectorID string, err error) {
+	parts := strings.Split(id, ":")
+	if len(parts) != 2 {
+		err = fmt.Errorf("GuardDuty Member ID must be of the form <Detector ID>:<Member AWS Account ID>")
```
radeksimko: Nitpick: We could include the real ID in the error message here.

aws/resource_aws_guardduty_test.go
radeksimko: I see, 👌

bflad:

Issue #2493, opened 2017-11-30
Add SageMaker support


jckuester:
Amazon has announced the new service [Sage​Maker](https://aws.amazon.com/sagemaker/) to build, train, and deploy machine learning models at scale.

### Prerequisite
aws-sdk-go v1.12.36 (https://github.com/terraform-providers/terraform-provider-aws/pull/2474)

### Affected Resource(s)

The following resources are needed.

For building model and training:
- aws_sagemaker_notebook_instance (https://github.com/terraform-providers/terraform-provider-aws/pull/2999)
- aws_presigned_notebook_instance_url
- aws_sagemaker_training_job (https://github.com/terraform-providers/terraform-provider-aws/pull/2955)
- aws_sagemaker_lifecycle_config resource (https://github.com/terraform-providers/terraform-provider-aws/pull/7585)

For the deployment/hosting part:
- aws_sagemaker_model (open PR is https://github.com/terraform-providers/terraform-provider-aws/pull/2478)
- aws_sagemaker_endpoint (open PR is https://github.com/terraform-providers/terraform-provider-aws/pull/2479)
- aws_sagemaker_endpoint_configuration (open PR is https://github.com/terraform-providers/terraform-provider-aws/pull/2477)

### Expected Behavior

Create, update, delete, and import for above listed SageMaker resources of  deployment part:

```hcl
resource "aws_sagemaker_endpoint" "foo" {
	name = "endpoint-foo"
	endpoint_config_name = "${aws_sagemaker_endpoint_configuration.ec.name}"
}

resource "aws_sagemaker_endpoint_configuration" "ec" {
	name = "endpoint-config-foo"

	production_variants {
                variant_name            = "variant-1"
		model_name = "${aws_sagemaker_model.m.name}"
                initial_instance_count  = 1
                instance_type           = "m3.xlarge"
                initial_variant_weight  = 1
	}
}

resource "aws_sagemaker_model" "m" {
    name = "my-model"

    primary_container {
       image = "111111111111.ecr.us-west-2.amazonaws.com/my-docker-image:latest"
       model_data_url  = "s3://111111111111-foo/model.tar.gz"
    }
}
```

### Example

A terraform example that shows how to use the above resources to deploy your own model is here: https://github.com/terraform-providers/terraform-provider-aws/pull/2585

### References
- https://github.com/aws/aws-sdk-go/tree/master/service/sagemaker

darrenhaken:
[...]

PR #2924, opened 2018-01-10
Add SageMaker vendor

ddcprg:
This is required for the new SageMaker resources - see #2493

I'm raising a separate PR as stated in README.md


ddcprg:
thanks @radeksimko for looking at this really quickly

PR #2926, opened 2018-01-10
Added option to avoid DNS record overwrite

Erouan50:
The current behavior of the aws_route53_record resource is to overwrite the record if one already exists. For example, we suppose that the notexample.com CNAME to foo.bar is added by hand via the CLI. If the new Terraform resource with the same name but with a different CNAME value is added, the record is going to be overridden. This might be a big problem: a record might be changed by mistake.

So this PR proposes to add an option named allow_overwrite on the resource to protect overwrite. When the option is true the CREATE mode is used instead of UPSERT in the change batch. This means that if an added record already exists on route53, the application will fail. By default, the overwrite is permitted to avoid compatibility issue.

### DIFF
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -232,6 +232,12 @@ func resourceAwsRoute53Record() *schema.Resource {
 				Optional:      true,
 				Set:           schema.HashString,
 			},
+
+			"allow_overwrite": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 		},
 	}
 }
@@ -379,6 +385,16 @@ func resourceAwsRoute53RecordCreate(d *schema.ResourceData, meta interface{}) er
 		return err
 	}
 
+	// Protect existing DNS records which might be managed in another way
+	// Use UPSERT only if the overwrite flag is true or if the current action is an update
+	// Else CREATE is used and fail if the same record exists
+	var action string
+	if d.Get("allow_overwrite").(bool) || !d.IsNewResource() {
+		action = "UPSERT"
+	} else {
+		action = "CREATE"
+	}
+
 	// Create the new records. We abuse StateChangeConf for this to
 	// retry for us since Route53 sometimes returns errors about another
 	// operation happening at the same time.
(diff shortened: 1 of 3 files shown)


bflad:
@Erouan50 thanks for this contribution! I have definitely gotten bit with this behavior in the past. Do you have time to implement an acceptance test for this as well? e.g.

```
func TestAccAWSRoute53Record_allowOverwrite(t *testing.T) {
  resource.Test(t, resource.TestCase{
    PreCheck:      func() { testAccPreCheck(t) },
    IDRefreshName: "aws_route53_record.default",
    Providers:     testAccProviders,
    CheckDestroy:  testAccCheckRoute53RecordDestroy,
    Steps: []resource.TestStep{
      {
        PreConfig: func() {
        // Create new Route53 record ahead of time with AWS SDK and wait until its ready
        },
        Config: testAccRoute53RecordConfig_allowOverride(false),
        ExpectError: regexp.MustCompile("... expected AWS error here ..."),
      },
      {
        Config: testAccRoute53RecordConfig_allowOverride(true),
          testAccCheckRoute53RecordExists("aws_route53_record.default"),
	),
      },
    },
  })
}

func testAccRoute53RecordConfig_allowOverride(allowOverride bool) string {
  return fmt.Sprintf(`
resource "aws_route53_zone" "main" {
  name = "notexample.com"
}

resource "aws_route53_record" "default" {
  allow_override = %v
  zone_id = "${aws_route53_zone.main.zone_id}"
  name = "www.notexample.com"
  type = "A"
  ttl = "30"
  records = ["127.0.0.1"]
}
`, allowOverride)
}
```

Erouan50:
@bflad I just added it:
```
TF_ACC=1 go test -v github.com/terraform-providers/terraform-provider-aws/aws -run AccAWSRoute53Record_allowOverwrite
=== RUN   TestAccAWSRoute53Record_allowOverwrite
--- PASS: TestAccAWSRoute53Record_allowOverwrite (111.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	111.641s
```

Erouan50:
No worries! Thank you for accepting my contribution!

bflad:

PR #2927, opened 2018-01-10
[WIP] create launch template resource

kl4w:
To address #2505 

I've created an initial `LaunchTemplate` resource but unsure how to handle versions and would like input/help.

Creating a new launch template is also conceptually creating the first `LaunchTemplateVersion`. 
If a new `LaunchTemplateVersion` resource is to be created, it seems weird that it may/should also update the `defaultVersion` and `latestVersion` attributes of the `LaunchTemplate`. Deleting a launch template also deletes all existing versions.

I was thinking about combining the concept of `LaunchTemplate` and `LaunchTemplateVersion` such that an update of the `LaunchTemplate` creates a `LaunchTemplateVersion`. This results in a single `LaunchTemplate` in the resulting state. The question is how to handle the versioning of the launch template and how to handle the `sourceVersion` parameter when creating a new `LaunchTemplateVersion`.

Combining `LaunchTemplate` and `LaunchTemplateVersion` also means that it does not allow as much freedom managing versions, and may mean that creating any resource using the `LaunchTemplate` may have to use the `latestVersion`.

Another idea that I was throwing around was create both the `LaunchTemplate` and `LaunchTemplateVersion` resources but remove the `launchTemplateData` attribute from the `LaunchTemplate`. This would mean that the `LaunchTemplate` is used more as a versioning reference, and the data is always held in the `LaunchTemplateVersion` resource. However this would also mean that both resources would be required in order to create a useful launch template.

Thoughts/ideas?


danieladams456:
I don't think having to configure two resources (`LaunchTemplate` and `LaunchTemplateVersion`) would be too bad.  The main thing I would want to do is be able to use Lambda or some other automated process to make a new version of the Launch Template and update the default version to the new without making a re-run of Terraform think it needs to change something.  This would let me do initial infrastructure creation through Terraform and automated maintenance through Lambda.

Having a boolean `manageVersions` or some other similar method would seem to achieve that.  I don't know if Terraform can store attributes in state that are for informational purposes only but not engage on a feedback loop for changing resources based on them?  If not, then maybe that boolean would cause version not to be stored in state so Terraform wouldn't see any changes.

bflad:
Terraform core provides a lifecycle configuration block for all resources that allows you to ignore certain attributes you don't want to track: https://www.terraform.io/docs/configuration/resources.html#ignore_changes

e.g. this will create an ECS service with a desired count of 1 (or import with whatever value) and ignore any changes
```hcl
  # other configuration
  desired_count = 1

  lifecycle {
    ignore_changes = ["desired_count"]
  }
}
```

danieladams456:
@bflad Is that only for explicitly defined fields or computed values as well?

As long as this feature is implemented in a way that the default version attribute can be ignored with this methodology, that sounds great to me.

kl4w:
Would appreciate some eyes on this PR so far.

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (21.18s)
=== RUN   TestAccAWSLaunchTemplate_data
--- PASS: TestAccAWSLaunchTemplate_data (21.27s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	42.490s
```

kl4w:
I think this is ready to remove to the WIP from this PR

kl4w:
rebased and added documentation:
```make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (12.32s)
=== RUN   TestAccAWSLaunchTemplate_data
--- PASS: TestAccAWSLaunchTemplate_data (10.72s)
=== RUN   TestAccAWSLaunchTemplate_update
--- PASS: TestAccAWSLaunchTemplate_update (19.42s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	42.505s```

**Comment by kl4w** [comment]
rebase again

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (10.54s)
=== RUN   TestAccAWSLaunchTemplate_data
--- PASS: TestAccAWSLaunchTemplate_data (10.51s)
=== RUN   TestAccAWSLaunchTemplate_update
--- PASS: TestAccAWSLaunchTemplate_update (18.99s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	40.084s
```

File: aws/resource_aws_launch_template.go [hunk 1]
```diff
@@ -0,0 +1,976 @@
...
+	"strconv"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+const awsSpotInstanceTimeLayout = "2006-01-02T15:04:05Z"
```
bflad: We should use RFC3339 instead of our own custom time layout. 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateLaunchTemplateName,
+			},
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
```
bflad: This can be simplified with:

```go
ValidateFunc: validation.StringLenBetween(0, 255),
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+					value := v.(string)
+					if len(value) > 255 {
+						errors = append(errors, fmt.Errorf(
+							"%q cannot be longer than 255 characters", k))
+					}
+					return
+				},
+			},
+
+			"client_token": {
```
bflad: This attribute is not necessary (nor would I recommend) exposing. 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+						"no_device": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"virtual_name": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"ebs": {
+							Type:     schema.TypeSet,
```
bflad: To simplify the implementation and future maintenance, we can use `schema.TypeList` here 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+				Optional: true,
+			},
+
+			"instance_initiated_shutdown_behavior": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"instance_market_options": {
+				Type:     schema.TypeSet,
```
bflad: To simply the implementation and future maintenance, this can be `schema.TypeList` 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+			},
+
+			"elastic_gpu_specifications": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:     schema.TypeString,
+							Optional: true,
```
bflad: This attribute is listed as required in the [SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#ElasticGpuSpecification). 😄

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"market_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"spot_options": {
+							Type:     schema.TypeSet,
```
bflad: To simply the implementation and future maintenance, this can be schema.TypeList 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+			"instance_market_options": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"market_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
```
bflad: We can provide plan time validation for `market_type`:

```go
ValidateFunc: validation.StringInSlice([]string{ec2.MarketTypeSpot}, false),
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"block_duration_minutes": {
+										Type:     schema.TypeInt,
+										Optional: true,
+									},
+									"instance_interruption_behavior": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
```
bflad: We can provide plan-time validation for `instance_interruption_behavior`:

```go
  ec2.InstanceInterruptionBehaviorHibernate,
  ec2.InstanceInterruptionBehaviorStop,
  ec2.InstanceInterruptionBehaviorTerminate,
, false),
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+									"spot_instance_type": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"valid_until": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
```
bflad: Let's try to standardize `valid_until` on RFC3339 and use plan time validation 👍 

```go
ValidateFunc: validation.ValidateRFC3339TimeString,
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+
+			"image_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"instance_initiated_shutdown_behavior": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
```
bflad: Let's provide plan time validation for the `instance_initiated_shutdown_behavior` attribute 👍 

```go
  ec2.ShutdownBehaviorStop,
  ec2.ShutdownBehaviorTerminate,
}, false),
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+
+			"key_name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"monitoring": {
```
bflad: Since the API nests this attribute, let's nest it as well to prevent issues in the future:

```go
"monitoring": {
  Type: schema.TypeList,
  Optional: true,
  MaxItems: 1,
  Elem: &schema.Resource{
      "enabled": {
        Type: schema.TypeBool,
        Optional: true,
      },
    },
  },
},
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+						},
+						"device_index": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"security_groups": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
```
bflad: You can (and should 😉 ) skip using `Set:` here 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+						},
+						"ipv6_address_count": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
+						"ipv6_addresses": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
```
bflad: You can (and should 😉 ) skip using `Set:` here 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+						},
+						"private_ip_address": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"ipv4_addresses": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
```
bflad: You can (and should 😉 ) skip using `Set:` here 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+							Optional: true,
+						},
+						"spread_domain": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"tenancy": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
```
bflad: We can provide plan-time validation for `tenancy` 👍 

```go
  ec2.TenancyDedicated,
  ec2.TenancyDefault,
  ec2.TenancyHost,
}, false),
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+			"ram_disk_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"security_group_names": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
```
bflad: You can (and should 😉 ) skip using Set: here 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+			},
+
+			"vpc_security_group_ids": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
```
bflad: You can (and should 😉 ) skip using Set: here 👍

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+
+const awsSpotInstanceTimeLayout = "2006-01-02T15:04:05Z"
+
+func resourceAwsLaunchTemplate() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsLaunchTemplateCreate,
+		Read:   resourceAwsLaunchTemplateRead,
+		Update: resourceAwsLaunchTemplateUpdate,
+		Delete: resourceAwsLaunchTemplateDelete,
+		Importer: &schema.ResourceImporter{
```
bflad: Import documentation is missing 😅

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+	log.Printf("[DEBUG] Reading launch template %s", d.Id())
+
+	dlt, err := conn.DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{
+		LaunchTemplateIds: []*string{aws.String(d.Id())},
+	})
+	if err != nil {
+		return fmt.Errorf("Error getting launch template: %s", err)
+	}
+	if len(dlt.LaunchTemplates) == 0 {
+		d.SetId("")
```
bflad: We should also provide a log message when this occurs:
```go
log.Printf("[WARN] launch template (%s) not found - removing from state", d.Id())
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+
+func resourceAwsLaunchTemplateDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	log.Printf("[DEBUG] Launch Template destroy: %v", d.Id())
+	_, err := conn.DeleteLaunchTemplate(&ec2.DeleteLaunchTemplateInput{
+		LaunchTemplateId: aws.String(d.Id()),
+	})
+	if err != nil {
+		return err
```
bflad: If EC2 throws a specific error when its already deleted, we should catch that and return nil, e.g. (I haven't verified this is the correct ErrCode or anything)

```go
if isAWSErr(err, ec2.LaunchTemplateErrorCodeLaunchTemplateIdDoesNotExist, "") {
  return nil
}
return err
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+func resourceAwsLaunchTemplateRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+
+	log.Printf("[DEBUG] Reading launch template %s", d.Id())
+
+	dlt, err := conn.DescribeLaunchTemplates(&ec2.DescribeLaunchTemplatesInput{
+		LaunchTemplateIds: []*string{aws.String(d.Id())},
+	})
+	if err != nil {
+		return fmt.Errorf("Error getting launch template: %s", err)
```
bflad: If EC2 throws a specific error when its missing, we should catch that:

```go
if isAWSErr(err, ec2.LaunchTemplateErrorCodeLaunchTemplateIdDoesNotExist, "") {
  log.Printf("[WARN] launch template (%s) not found - removing from state", d.Id())
  d.SetId("")
  return nil
}
return fmt.Errorf("Error getting launch template: %s", err)
```

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,976 @@
...
+
+	log.Printf("[DEBUG] Launch Template deleted: %v", d.Id())
+	return nil
+}
+
+func getBlockDeviceMappings(m []*ec2.LaunchTemplateBlockDeviceMapping) []interface{} {
+	s := []interface{}{}
+	for _, v := range m {
+		mapping := map[string]interface{}{
+			"device_name":  *v.DeviceName,
```
bflad: Rather than risk panics on nil dereferencing, we should switch anything using `*thing` to the appropriate SDK method that will set it to the zero value for the type: e.g. `aws.StringValue(v.DeviceName)` This should repeated everywhere below.

aws/resource_aws_launch_template_test.go
```diff
@@ -0,0 +1,228 @@
...
+		CheckDestroy: testAccCheckAWSLaunchTemplateDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSLaunchTemplateConfig_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSLaunchTemplateExists(resName, &template),
+					resource.TestCheckResourceAttr(resName, "default_version", "1"),
+					resource.TestCheckResourceAttr(resName, "latest_version", "1"),
+				),
+			},
```
bflad: Let's add a `TestStep` here that imports the resource and ensures the Terraform state matches what is expected 👍 

```go
{
  ResourceName: resName,
  ImportState: true,
  ImportStateVerify: true,
},
```

aws/resource_aws_launch_template_test.go
```diff
@@ -0,0 +1,228 @@
...
+					resource.TestCheckResourceAttrSet(resName, "kernel_id"),
+					resource.TestCheckResourceAttrSet(resName, "key_name"),
+					resource.TestCheckResourceAttrSet(resName, "monitoring"),
+					resource.TestCheckResourceAttr(resName, "network_interfaces.#", "1"),
+					resource.TestCheckResourceAttr(resName, "placement.#", "1"),
+					resource.TestCheckResourceAttrSet(resName, "ram_disk_id"),
+					resource.TestCheckResourceAttr(resName, "vpc_security_group_ids.#", "1"),
+					resource.TestCheckResourceAttr(resName, "tag_specifications.#", "1"),
+				),
+			},
```
bflad: Let's add a `TestStep` here that imports the resource and ensures the Terraform state matches what is expected 👍 

```
{
  ResourceName: resName,
  ImportState: true,
  ImportStateVerify: true,
},
```

aws/resource_aws_launch_template_test.go
```diff
@@ -0,0 +1,228 @@
...
+			{
+				Config: testAccAWSLaunchTemplateConfig_data,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSLaunchTemplateExists(resName, &template),
+					resource.TestCheckResourceAttr(resName, "default_version", "1"),
+					resource.TestCheckResourceAttr(resName, "latest_version", "2"),
+					resource.TestCheckResourceAttrSet(resName, "image_id"),
+				),
+			},
```
bflad: Let's add a `TestStep` here that imports the resource and ensures the Terraform state matches what is expected 👍 

```
{
  ResourceName: resName,
  ImportState: true,
  ImportStateVerify: true,
},
```

aws/resource_aws_launch_template_test.go
```diff
@@ -0,0 +1,228 @@
...
+			LaunchTemplateIds: []*string{aws.String(rs.Primary.ID)},
+		})
+
+		if err == nil {
+			if len(resp.LaunchTemplates) != 0 && *resp.LaunchTemplates[0].LaunchTemplateId == rs.Primary.ID {
+				return fmt.Errorf("Launch Template still exists")
+			}
+		}
+
+		ae, ok := err.(awserr.Error)
```
bflad: This can be simplified with `if isAWSErr(err, "InvalidLaunchTemplateId.NotFound", "") {` 👍  (replacing the middle with an SDK constant if available)

aws/resource_aws_launch_template.go
```diff
@@ -277,34 +279,31 @@ func resourceAwsLaunchTemplate() *schema.Resource {
 							Optional: true,
 						},
 						"security_groups": {
-							Type:     schema.TypeSet,
```
bflad: These (`security_groups`, `ipv6_addresses`, and `ipv4_addresses`) will likely need to remain `Type: TypeSet` (just needed the `Set: schema.HashString` removed) as the API likely does not return them in a consistent order. We'll find out!

aws/resource_aws_launch_template.go
```diff
@@ -358,21 +362,19 @@ func resourceAwsLaunchTemplate() *schema.Resource {
 			},
 
 			"security_group_names": {
-				Type:     schema.TypeSet,
```
bflad: Similar note here about `security_group_names` and `vpc_security_group_ids` -- we shall see!

aws/resource_aws_launch_template.go
```diff
@@ -559,27 +570,27 @@ func getBlockDeviceMappings(m []*ec2.LaunchTemplateBlockDeviceMapping) []interfa
...
 		}
 		if v.Ebs != nil {
 			ebs := map[string]interface{}{
-				"delete_on_termination": *v.Ebs.DeleteOnTermination,
-				"encrypted":             *v.Ebs.Encrypted,
-				"volume_size":           *v.Ebs.VolumeSize,
-				"volume_type":           *v.Ebs.VolumeType,
+				"delete_on_termination": aws.BoolValue(v.Ebs.DeleteOnTermination),
+				"encrypted":             aws.BoolValue(v.Ebs.Encrypted),
+				"volume_size":           aws.Int64Value(v.Ebs.VolumeSize),
```
bflad: Any of these `aws.Int64Value()` calls will need to be wrapped in `int(aws.Int64Value())` since Terraform's `TypeInt` is a regular `int`

kl4w:
Made modifications based off of the review. Can someone please advise on the import? I'm having test failures after adding the import step verification

Without import
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
gofmt -w $(find . -name '*.go' |grep -v vendor)
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (13.15s)
=== RUN   TestAccAWSLaunchTemplate_data
--- PASS: TestAccAWSLaunchTemplate_data (11.08s)
=== RUN   TestAccAWSLaunchTemplate_update
--- PASS: TestAccAWSLaunchTemplate_update (21.39s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws
```

With import:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
gofmt -w $(find . -name '*.go' |grep -v vendor)
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (14.27s)
=== RUN   TestAccAWSLaunchTemplate_data
--- FAIL: TestAccAWSLaunchTemplate_data (12.74s)

		(map[string]string) {
		}


		(map[string]string) (len=5) {
		 (string) (len=12) "monitoring.#": (string) (len=1) "1",
		 (string) (len=20) "monitoring.0.enabled": (string) (len=4) "true",
		 (string) (len=11) "ram_disk_id": (string) (len=12) "ari-a12bc3de",
		 (string) (len=24) "vpc_security_group_ids.#": (string) (len=1) "1",
		 (string) (len=24) "vpc_security_group_ids.0": (string) (len=4) "test"
		}

=== RUN   TestAccAWSLaunchTemplate_update
--- FAIL: TestAccAWSLaunchTemplate_update (23.05s)
	testing.go:518: Step 2 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) {
		}


		(map[string]string) (len=5) {
		 (string) (len=12) "monitoring.#": (string) (len=1) "1",
		 (string) (len=20) "monitoring.0.enabled": (string) (len=4) "true",
		 (string) (len=11) "ram_disk_id": (string) (len=12) "ari-a12bc3de",
		 (string) (len=24) "vpc_security_group_ids.#": (string) (len=1) "1",
		 (string) (len=24) "vpc_security_group_ids.0": (string) (len=4) "test"
		}

FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	50.102s
```

bflad:
@kl4w generally that means `d.Set()` is not being called correctly for those attributes during read. 😄

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,1071 @@
...
+	ltData := dltv.LaunchTemplateVersions[0].LaunchTemplateData
+
+	d.Set("disable_api_termination", ltData.DisableApiTermination)
+	d.Set("ebs_optimized", ltData.EbsOptimized)
+	d.Set("image_id", ltData.ImageId)
+	d.Set("instance_initiated_shutdown_behavior", ltData.InstanceInitiatedShutdownBehavior)
+	d.Set("instance_type", ltData.InstanceType)
+	d.Set("kernel_id", ltData.KernelId)
+	d.Set("key_name", ltData.KeyName)
+	d.Set("monitoring", ltData.Monitoring)
```
bflad: This needs to be updated to err check (since its non-scalar now) as well as correctly handle the new nested structure:

```go
if err := d.Set("monitoring", getMonitoring(ltData.Monitoring)); err != nil {
  return err
}
```

kl4w: ahhh makes sense. Thank you. I'll make those changes now.

aws/resource_aws_launch_template.go
```diff
@@ -0,0 +1,1071 @@
...
+	d.Set("ram_dist_id", ltData.RamDiskId)
```
bflad: Typo: 😅  `d.Set("ram_disk_id", ltData.RamDiskId)` (FYI you can find these by running the acceptance testing with `TF_SCHEMA_PANIC_ON_ERROR=1` as well as the import testing that noticed it)

kl4w:
```
TF_SCHEMA_PANIC_ON_ERROR=1 make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
gofmt -w $(find . -name '*.go' |grep -v vendor)
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_importBasic
--- PASS: TestAccAWSLaunchTemplate_importBasic (14.62s)
=== RUN   TestAccAWSLaunchTemplate_importData
--- PASS: TestAccAWSLaunchTemplate_importData (15.01s)
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (12.00s)
=== RUN   TestAccAWSLaunchTemplate_data
--- PASS: TestAccAWSLaunchTemplate_data (13.97s)
=== RUN   TestAccAWSLaunchTemplate_update
--- PASS: TestAccAWSLaunchTemplate_update (21.29s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws
```

bflad:
Do you have a second to update the `monitoring` documentation as well? Running testing now

website/docs/r/launch_template.html.markdown
```diff
@@ -0,0 +1,243 @@
...
+    market_type = "test"
+  }
+
+  instance_type = "t2.micro"
+
+  kernel_id = "test"
+
+  key_name = "test"
+
+  monitoring = true
```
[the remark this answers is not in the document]
kl4w: ahh thanks. I was going to update the documentation as well on this branch. Do you want me to hold off?

Issue #2303, opened 2017-11-15
Modifying AWS IAM user name deletes state


hashibot:
_This issue was originally opened by @alvin-huang as hashicorp/terraform#16646. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
v0.10.8

### Terraform Configuration Files
main.tf
```hcl
variable "aws_region" {}
variable "my_user" {}

provider "aws" {
  region = "${var.aws_region}"
}

resource "aws_iam_user" "myuser" {
  name = "${var.my_user}"
  path = "/"
}
```

createuser.tfvars
```hcl
aws_region  = "us-east-1"
my_user = "createuser"
```

modifyuser.tfvars
```hcl
aws_region  = "us-east-1"
my_user = "modifyuser"
```

### Expected Behavior
The `createuser` AWS user should be renamed to `modifyuser` and the state file should reflect this change.

### Debug
https://gist.github.com/alvin-huang/e07b5e15b5837401b811db1acff84435

### Actual Behavior
The AWS user does get renamed but the terraform state file is empty. 

### Steps to Reproduce
1. `terraform init`
2. `terraform apply -var-file=createuser.tfvars` to create the `createuser` user
3. `terraform show` to show `createuser` is in created and in the state file
4. `terraform apply -var-file=modifyuser.tfvars` to change the name of `createuser` to `modifyuser`
5. `terraform show` shows an empty state file

### Important Factoids
I see the same issue with a local state file and remote backend (S3). 

### References
terraform-providers/terraform-provider-aws#213
hashicorp/terraform#3227

If this should go in https://github.com/terraform-providers/terraform-provider-aws let me know and I can move it over there or hashibot will take care of it :)

jbardin:
It looks like `aws_iam_user` uses the name field as the id, except that the name field is modifiable 😮 

The example above updates the name field, but then when the user is re-read the "id" no longer exists because the name was changed, and the resource is deleted from the state.

loivis:
There is an attribute as `unique_id`. It could potentially be used as `Id()`, as mentioned in current codes.

https://github.com/terraform-providers/terraform-provider-aws/blob/dddc588b53d0de7421040313d2ec7d54ca55bd4a/aws/resource_aws_iam_user.go#L30-L41

bflad:
[...]

PR #2940, opened 2018-01-11
resource/aws_iam_user: Fix issues with update

julienduchesne:
1. Path can be updated without creating a new user
2. When updating a user's name, the Read query fetched the user with the old name, resulting in a 404.

### DIFF
--- a/aws/resource_aws_iam_user.go
+++ b/aws/resource_aws_iam_user.go
@@ -48,7 +48,6 @@ func resourceAwsIamUser() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "/",
-				ForceNew: true,
 			},
 			"force_destroy": &schema.Schema{
 				Type:        schema.TypeBool,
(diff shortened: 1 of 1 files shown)


julienduchesne:
Fixes #2303

bflad:
@julienduchesne thanks so much for the pull request here! I will likely be pulling this in together with #2979 since it contains an additional acceptance test. One piece we're hoping to add here as well is the following code in the resource definition to allow other downstream resources to see the ID change and not require a second plan/apply (untested):

```go
CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
    if diff.HasChange("name") || diff.HasChange("path") {
        return diff.SetNewComputed("id")
    }
    return nil
}
```

More soon!

bflad:
This is merged into master and will be released with [v1.7.1](https://github.com/terraform-providers/terraform-provider-aws/milestone/9), hopefully later this week. FYI I added an additional acceptance test [TestAccAWSUser_pathChange](https://github.com/terraform-providers/terraform-provider-aws/commit/e42beb2a4d088f7e19f39a74564e955a28026918) to ensure we are successfully updating the user's path.

PR #2947, opened 2018-01-11
New Data Source: aws_security_groups

### LINKED ISSUE #2740: aws_security_group_ids data source
Hi there,

We would love to have the ability of querying all the security groups in a VPC. The idea is to have a `aws_security_group_ids` data source, equivalent to `aws_subnet_ids`.

I have multiple Kubernetes clusters running on the same VPC, and need to Terraform a VPC peering which will open all `nodes` security groups. Using the `aws_security_group` data source, I can query the security groups, however, it fails if there is more than one match. 

Is there any reason for [constraining the results](https://github.com/terraform-providers/terraform-provider-aws/blob/63b287b4d32163a4909c8b7f866c2922a8651d43/aws/data_source_aws_security_group.go#L83-L85)?

```
data "aws_security_group" "nodes" {
  vpc_id = "${var.vpc_id}"
  tags {
    Name = "*nodes*"
  }
}
```
Produces the following error:
```
Error: Error refreshing state: 1 error(s) occurred:

* data.aws_security_group.nodes: 1 error(s) occurred:

* data.aws_security_group.nodes: data.aws_security_group.nodes: multiple Security Groups matched; use additional constraints to reduce matches to a single Security Group

```

Terraform v0.11.1

ian-d:
Resolves #2740.

Using the data sources `aws_instance`/`aws_instances` naming convention, this creates an `aws_security_groups` data source for multiple security groups to complement the `aws_security_group` data source that fetches single security groups.

The example code in the documentation could use an example of creating a resource actually using the data source but I couldn't think of a concise AWS resource that uses a list of security groups that would make much sense in context.

Tests:
```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsSecurityGroups_'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsSecurityGroups_ -timeout 120m
=== RUN   TestAccDataSourceAwsSecurityGroups_tag
--- PASS: TestAccDataSourceAwsSecurityGroups_tag (24.85s)
=== RUN   TestAccDataSourceAwsSecurityGroups_filter
--- PASS: TestAccDataSourceAwsSecurityGroups_filter (25.24s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	50.130s
```

### DIFF
--- a/aws/data_source_aws_security_groups.go
+++ b/aws/data_source_aws_security_groups.go
@@ -1,0 +1,98 @@
+package aws
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"log"
+)
+
+func dataSourceAwsSecurityGroups() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsSecurityGroupsRead,
+
+		Schema: map[string]*schema.Schema{
+			"filter": dataSourceFiltersSchema(),
+			"tags":   tagsSchemaComputed(),
+
+			"ids": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"vpc_ids": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func dataSourceAwsSecurityGroupsRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ec2conn
+	req := &ec2.DescribeSecurityGroupsInput{}
+
+	filters, filtersOk := d.GetOk("filter")
+	tags, tagsOk := d.GetOk("tags")
+
+	if !filtersOk && !tagsOk {
+		return fmt.Errorf("One of filters or tags must be assigned")
+	}
+
+	if filtersOk {
+		req.Filters = append(req.Filters,
+			buildAwsDataSourceFilters(filters.(*schema.Set))...)
+	}
+	if tagsOk {
+		req.Filters = append(req.Filters, buildEC2TagFilterList(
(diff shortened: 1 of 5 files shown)


luckytaxi:
Any idea as to when this would merge? We need this functionality.

aws/data_source_aws_security_groups.go
```diff
@@ -0,0 +1,98 @@
...
+	if tagsOk {
+		req.Filters = append(req.Filters, buildEC2TagFilterList(
+			tagsFromMap(tags.(map[string]interface{})),
+		)...)
+	}
+
+	log.Printf("[DEBUG] Reading Security Groups with request: %s", req)
+
+	var ids, vpc_ids []string
+	nextToken := ""
```
bflad: This new variable and logic surrounding it is extraneous, you can directly map the new `resp.NextToken` onto `req.NextToken`, e.g.

```go
for {
  resp, err := conn.DescribeSecurityGroups(req)
  if err != nil {
    return fmt.Errorf("error reading security groups: %s", err)
  }
  for _, sg := range resp.SecurityGroups {
    ids = append(ids, aws.StringValue(sg.GroupId))
    vpc_ids = append(vpc_ids, aws.StringValue(sg.VpcId))
  }
  if resp.NextToken == nil {
    break
  }
  req.NextToken = resp.NextToken
}
```

website/aws.erb
```diff
@@ -203,6 +203,9 @@
                         <li<%= sidebar_current("docs-aws-datasource-security-group") %>>
```
bflad: We don't expect most folks to know this, but an odd side effect of having the singular and plural data sources is that both entries will be highlighted in the sidebar if you are on the singular data source page since the `sidebar_current` prefix matches both. Our current convention is to go back to the singular data source and append `-x` to the `sidebar_current` in both the sidebar and singular data source page.

website/aws.erb
```diff
@@ -203,6 +203,9 @@
...
                          <a href="/docs/providers/aws/d/security_group.html">aws_security_group</a>
                         </li>
+                        <li<%= sidebar_current("docs-aws-datasource-security-groups") %>>
+                         <a href="/docs/providers/aws/d/security_groups.html">aws_security_group</a>
```
bflad: Typo 😎 : missing `s` in data source name

bflad:

PR #2971, opened 2018-01-12
Include size of user_data in validation error message

alewando:
Related to #2765 , puts the actual size of the user_data in the error message so you can see how close (or not) you are


bflad:

PR #2973, opened 2018-01-12
Port user_data length validation to aws_launch_configuration

alewando:
Port user_data length validation from #2765 to aws_launch_configuration


bflad:
@alewando interestingly enough, in the [AWS Autoscaling API documentation](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_CreateLaunchConfiguration.html), the limit for the `UserData` attribute is actually 21847. 🙁 

> Length Constraints: Maximum length of 21847.

Can you validate or preferably write an acceptance test to prove that the shorter value of 16384 is also correct for launch configuration user data? If not, can you create a separate validation method for this attribute please? Some people might be using values in between the two.

alewando:
I believe the 21847 byte limit is for base64-encoded data (16384 * 1.3333 = 21846), but will verify manually. Terraform base64-encodes the supplied userdata, so I suspect the limit is actually 16kb of binary data.

alewando:
I have verified manually that the maximum allowable size for user_data on a launch configuration is 16384 (same as for EC2 instances themselves). I verified by creating files of controlled sizes and using them as template data sources for a launch configuration's user_data. Files were created via:

`awk 'BEGIN {while (c++<__DESIRED_SIZE__) printf "x"}'>user-data.tpl`

For any file size larger than 16834, AWS failed with an error message:
`* aws_launch_configuration.lc: Error creating launch configuration: ValidationError: 1 validation error detected: Value 'xxx...xxx' at 'userData' failed to satisfy constraint: Member must have length less than or equal to 21847`

Writing an acceptance test to prove this would be awkward, since the whole point of the PR is to 
apply this validation prior to sending the request to AWS. An acceptance test would need to provide some way of disabling this validation, which seems like it would be complicating the user-facing implementation for the purpose of validating documented (albeit poorly) external system behavior.

bflad:

Zordrak:
I could do with some clarification of what is going on here.

Every project everywhere is suddently having to change its default user data value from an empty string to a a single space character as a workaround for this validation.

Is this intentional? What prevents an empty string from being a valid way to define no userdata requirement?

Zordrak:
I will ask the question on https://github.com/terraform-providers/terraform-provider-aws/pull/4037 . I appreciate that you can omit the user_data attribute in one case, but in every single deployment we have autoscaling groups are part of a microservice module that takes userdata information as a parameter. Some of our microservices have userdata and some don't and as a result of this change, those that don't fail to validate. The workaround of changing the default from "" to " " would work.. but will cause a recreation of all launch configurations that have no userdata.

PR #2980, opened 2018-01-12
resource/aws_lb_target_group: Allow a blank health check path, for TCP healtchecks

### LINKED ISSUE #2978: [aws_lb_target_group] Validation forces invalid path for TCP health check
It is desirable to be able to configure either a TCP or HTTP/HTTPS health check based on variables as in the below example. Currently the validation forces the health check path to begin with a slash("/"), even when the protocol is TCP, preventing this use case. When the protocol is TCP, blank ("") is the only valid path for the AWS API.

```hcl
  health_check {
    healthy_threshold   = 2
    unhealthy_threshold = 2
    timeout             = 10
    port                = 8000
    path                = "${var.protocol == "TCP" ? "" : "/healthz"}"
    protocol            = "${var.protocol}"
    interval            = 30
    matcher             = "${var.protocol == "TCP" ? "" : "200-399"}"
  }
```

### Terraform Version
```
Terraform v0.11.1
+ provider.aws v1.7.0
```

### Affected Resource(s)
- aws_lb_target_group

```hcl
  health_check {
    healthy_threshold   = 2
    unhealthy_threshold = 2
    timeout             = 10
    port                = 8000
    path                = ""
    protocol            = "TCP"
    interval            = 30
    matcher             = ""
  }
```

### Expected Behavior
Expected to be able to use a blank 'path' for a TCP health check.
Or else the path validation should be left to the AWS API.

### Actual Behavior
```
Error: module.nlb.aws_lb_target_group.nlb[0]: "health_check.0.path" must begin with a '/' character: ""
```

### Steps to Reproduce
1. Create a `aws_lb_target_group` with `health_check.protocol="TCP"` and `health_check.path=""`
2. `terraform init`
3. `terraform apply`

### References
Validation was recently improved to enable HTTP/HTTPS health checks to be able to be specified for TCP Target Groups in PR #2906.

whereisaaron:
Fixes #2978 "Validation forces invalid path for TCP health check" 

Currently the validation forces the health check path to begin with a slash("/"), even when the protocol is TCP, in which case the health check path must be blank. The validation is preventing valid use cases. 

Since this validation can only check the one field on its own, the check is amended to only validate the path begins with a "/" if the path is not blank. Note that when the path is blank, it is not passed to AWS at all (see lines 226-228), so I think this change does not weaken validation.

### DIFF
--- a/aws/resource_aws_lb_target_group.go
+++ b/aws/resource_aws_lb_target_group.go
@@ -395,7 +395,7 @@ func validateAwsLbTargetGroupHealthCheckPath(v interface{}, k string) (ws []stri
 		errors = append(errors, fmt.Errorf(
 			"%q cannot be longer than 1024 characters: %q", k, value))
 	}
-	if !strings.HasPrefix(value, "/") {
+	if len(value) > 0 && !strings.HasPrefix(value, "/") {
 		errors = append(errors, fmt.Errorf(
 			"%q must begin with a '/' character: %q", k, value))
 	}


bflad:
This has been released in [terraform-provider-aws version 1.7.1](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#171-january-19-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #2986, opened 2018-01-14
[RFC] Code Style and Schema Structure

drewsonne:
I'm not sure how to structure this quite yet, nor sure to its accuracy, but most of it is lifted from my discussion with @bflad and @radeksimko in #2175.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -253,6 +253,29 @@ into Terraform.
    The PR reviewers can help out on this front, and may provide comments with
    suggestions on how to improve the code.
 
+#### Terraform Schema and Code Idiosyncracies
+
+There are aspects of the terraform code base and models which have a common theme
+and style
+
+ - [ ] __Ignore Timestamps__: Generally, creation and modification dates are not
+   included in schemas.
+ - [ ] __Attribute Update Tests__: Try to add a second test step in at least one
+   test case showing attribute changes propagate during Update operations
+ - [ ] __AWS Errors__: Use the helper function (`isAWSErr(err, ...)`) to check the
+   type of AWS error.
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -253,6 +253,29 @@ into Terraform.
...
+   test case showing attribute changes propagate during Update operations
+ - [ ] __AWS Errors__: Use the helper function (`isAWSErr(err, ...)`) to check the
+   type of AWS error.
+ - [ ] __`Computed`__: The `Computed` attribute is generally used in isolation for
+   any IDs or anything not defined in the config and returned by the API.
+ - [ ] __`Computed` with `Optional`__: The `Computed` attribute is generally used
+   in conjunction with `Optional` when the API automatically sets unpredictable 
+   default value or when the value is generally not static and depends on other 
+   attributes.
+ - [ ] __American Spelling__: Use American spelling for variables, functions, and
```
radeksimko: I think it's 👌  to keep this here. Just note that my comment in your past PR was more about the fact that the word was spelled like that in the AWS API.

.github/CONTRIBUTING.md
```diff
@@ -253,6 +253,29 @@ into Terraform.
...
+   constants.
+ - [ ] __Removed Resources__: If a resources is removed from AWS, make sure to
```
radeksimko: Typo: _If a resource is removed_ (no plural)

I would make it a little bit more explicit to make sure people don't think this is supposed to be done during `Delete()`:

> __Removed Resources__: If a resource is removed from AWS outside of Terraform (e.g. via different tool, API or web UI), make sure to catch this case. Print a `[WARN]` log message, and use `d.SetId("")` to remove the resource from state inside `Read()`.

drewsonne:
nw!

PR #3010, opened 2018-01-16
New Resource: aws_cognito_user_group

tomelliff:
Added support for AWS Cognito User Groups.

Should close https://github.com/terraform-providers/terraform-provider-aws/issues/2865 and I think might be the last thing left on https://github.com/terraform-providers/terraform-provider-aws/issues/232


tomelliff:
```
$ make testacc TEST=./aws TESTARGS="-run=TestAccAWSCognitoUserGroup_"
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoUserGroup_ -timeout 120m
=== RUN   TestAccAWSCognitoUserGroup_basic
--- PASS: TestAccAWSCognitoUserGroup_basic (91.23s)
=== RUN   TestAccAWSCognitoUserGroup_complex
--- PASS: TestAccAWSCognitoUserGroup_complex (75.06s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	166.304s
```

GreenSmile:
Hi, thanks for adding cognito groups!

I was preparing similar pull request (#3025). Please consider possibility to apply the following changes to your PR:
1) Please add documentation
2) Please add fields validation to match AWS requirements
3) I believe debug output before every AWS call could be rather useful
4) I think resource should be named "aws_cognito_user_pool_group", similarly to "aws_cognito_user_pool_client" and "aws_cognito_user_pool_domain". However, I'd like to hear comments from core members on this one
5) It maybe useful to include UserPoolId to resource ID, because group name is not unique. Resource ID like `us-east-1_esGpht5Fh/Admins` looks self-explanatory.
6) It would be really nice to have possibility to import existent group like 
```terraform import aws_cognito_user_pool_group.admin_group us-east-1_esGpht5Fh/Admins```. Not sure, however, how difficult it is.

Please check my PR for example of all points, except the very last one.

**Comment by tomelliff** [comment]
@GreenSmile Hey, sorry, I had added the docs and validation during the week but while I was on the train without internet and apparently forgot to push until I just saw your comment.

I'll go through your review comments now, thanks for the comments, particularly on the Id which is a good catch that I didn't think about clearly enough because the API docs states that it has to be unique but in hindsight it clearly means unique for the user pool.

**Comment by tomelliff** [comment]
@GreenSmile I think I've covered everything you talked about but some in slightly different ways to how you approached things by the looks of it, especially around the import.

I'm unsure about changing from `user_group` to `user_pool_group` but happy to change if others feel it's more obvious that way.

Tests seem okay and I've done a bit of manual testing as well:
```
$ make testacc TEST=./aws TESTARGS="-run=TestAccAWSCognitoUserGroup_"
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoUserGroup_ -timeout 120m
=== RUN   TestAccAWSCognitoUserGroup_basic
--- PASS: TestAccAWSCognitoUserGroup_basic (91.41s)
=== RUN   TestAccAWSCognitoUserGroup_complex
--- PASS: TestAccAWSCognitoUserGroup_complex (79.48s)
=== RUN   TestAccAWSCognitoUserGroup_importBasic
--- PASS: TestAccAWSCognitoUserGroup_importBasic (49.76s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	220.668s
```

File: aws/resource_aws_cognito_user_group_test.go [hunk 1]
```diff
@@ -0,0 +1,220 @@
...
+	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSCognitoUserGroup_basic(t *testing.T) {
+	poolName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	groupName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	updatedGroupName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
```
radeksimko: Do you mind prepending a unique fixed prefix to each name? e.g. `tf_acc_`? This is to make it easy/possible to cleanup any potentially leaked resources from the test and make it easier to identify those leaks in general.

aws/resource_aws_cognito_user_group_test.go
```diff
@@ -0,0 +1,220 @@
...
+				),
+			},
+		},
+	})
+}
+
+func TestAccAWSCognitoUserGroup_complex(t *testing.T) {
+	poolName := fmt.Sprintf("%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+	groupName := fmt.Sprintf("%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+	updatedGroupName := fmt.Sprintf("%s", acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
```
radeksimko: ^ likewise

aws/resource_aws_cognito_user_group_test.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func testAccAWSCognitoUserGroupConfig_basic(poolName, groupName string) string {
+	return fmt.Sprintf(`
+resource "aws_cognito_user_pool" "main" {
+  name = "%s"
+}
+
+resource "aws_cognito_user_group" "main" {
+  name         = "user-group-%s"
```
radeksimko: Nitpick: It's usually better to keep the whole name (incl. the prefix like `user-group-` here) in the test body as we may then reference it in checks, instead of rebuilding it again. 😉

aws/resource_aws_cognito_user_group.go
```diff
@@ -0,0 +1,175 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
+	"github.com/hashicorp/errwrap"
```
radeksimko: Nitpick: I know it's used all over the AWS provider at the moment, but in most places for no particular reason. The library is intended to address the problem of wrapping errors into errors so that we can retain all data about the original error. Here we're just wrapping errors to _print_ them out later. In other words I think we don't need this library unless we use `Contains()` or `ContainsType()` per https://github.com/hashicorp/errwrap#usage

tomelliff:
@radeksimko Thanks for the review. Think I've covered those points now and I've fixed the merge conflict on the validators_test as well.

bflad:

jayanderson:
https://www.terraform.io/docs/providers/aws/r/cognito_user_group.html - the documentation page looks a bit messed up for this feature. It isn't rendering the index sidebar and has some raw text at the top.

radeksimko:
@jayanderson Thanks for noticing. See https://github.com/terraform-providers/terraform-provider-aws/pull/3411

PR #3013, opened 2018-01-16
resource/aws_cognito_identity_pool_roles_attachment: Validate role existence

### LINKED ISSUE #3008: aws_cognito_identity_pool_roles_attachment silently passes with invalid role arn
### Terraform Version
```
⎇  terraform version
Terraform v0.11.2
+ provider.aws v1.7.0
```

### Affected Resource(s)
- aws_cognito_identity_pool_roles_attachment

```hcl
resource "aws_cognito_identity_pool_roles_attachment" "this" {
  identity_pool_id = "${var.id}"

  roles {
    "authenticated"   = "arn:aws:iam::123456789012:role/invalid-auth-role"
    "unauthenticated" = "arn:aws:iam::123456789012:role/invalid-unauth-role"
  }
}
```

### Expected Behavior
It should complain about roles.

### Actual Behavior
Apply succeeds with nothing defined in auth/unauth roles.

### Idea
Just hit the issue and haven't checked how aws sdk behaves in such scenario.

If aws sdk **DOES** behave like this, shall we add logic in terraform to handle the situation?

loivis:
Fixes #3008 

AWS `SetIdentityPoolRoles` silently passes if provided role doesn't exist at all. As a user, I'd prefer to raising en error instead of resetting roles to none.

### DIFF
--- a/aws/resource_aws_cognito_identity_pool_roles_attachment.go
+++ b/aws/resource_aws_cognito_identity_pool_roles_attachment.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"bytes"
@@ -10,6 +11,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/cognitoidentity"
+	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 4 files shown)


loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccAWSCognitoIdentityPoolRolesAttachment_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoIdentityPoolRolesAttachment_ -timeout 120m
=== RUN   TestAccAWSCognitoIdentityPoolRolesAttachment_basic
--- PASS: TestAccAWSCognitoIdentityPoolRolesAttachment_basic (37.83s)
=== RUN   TestAccAWSCognitoIdentityPoolRolesAttachment_invalidRole
--- PASS: TestAccAWSCognitoIdentityPoolRolesAttachment_invalidRole (13.09s)
=== RUN   TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappings
--- PASS: TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappings (59.49s)
=== RUN   TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappingsWithAmbiguousRoleResolutionError
--- PASS: TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappingsWithAmbiguousRoleResolutionError (13.01s)
=== RUN   TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappingsWithRulesTypeError
--- PASS: TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappingsWithRulesTypeError (13.52s)
=== RUN   TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappingsWithTokenTypeError
--- PASS: TestAccAWSCognitoIdentityPoolRolesAttachment_roleMappingsWithTokenTypeError (13.70s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	150.685s
```

bflad:
Hi @loivis 👋  Thanks for your work here and sorry for the lengthy delay in providing a review.

It seems the upstream Cognito API should be updated in this case to perform its own validation. If you haven't done so already, I would suggest opening a feature request with AWS Support in that regard.

In this case though, in lieu of directly referencing a resource, we generally recommend using a data source to verify the existence of an resource rather than hardcoding values if possible. In this case, the [`aws_iam_role` data source](https://www.terraform.io/docs/providers/aws/d/iam_role.html) can prevent this issue during `plan` by verifying IAM role existence, rather than trying to shim similar or duplicate IAM role checking logic that would only be caught during `apply` with this resource.

One other note here is that we tend to ensure resources only talk to their respective service due to reduce code complexity and maintenance burden.

For these reasons, I'm going to close this pull request. If instead it seems appropriate to make a documentation update for this resource to include a **NOTE:** about the lack of IAM role validation in the API, we will gladly take a look. 👍

loivis:
@bflad Thanks for your time to look into this. I've almost forgot this. Totally agree with you and I did use data source later on as a way to validate the iam role.

PR #3038, opened 2018-01-17
New Resource: aws_kms_grant

### LINKED ISSUE #641: Support management of AWS KMS grants
_This issue was originally opened by @dougneal as hashicorp/terraform#13151. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Implement a `aws_kms_grant` resource to manage grants against KMS keys.

jammerful:
Hey,

Basically I think this is close to being ready to merge, but I see two flaws with it:

* Dealing with the `InvalidArnException` upon creation of a grant that depends on a newly created role. I have a retry in my code, but it will just retry in the case that the arn is actually invalid. I don't see a way to distinguish between eventual consistency and a truly invalid arn.
  - I think this maybe okay behavior.
* In a similar vein, if you refresh the state quickly after creating a role and grant (as is the case with the acceptance tests currently). The read function will sometimes get an internal identifier (AROAXXXXX instad of arn:aws:XXX) to the grantee principal instead of an ARN, which would if you naively take the value be written to the state file and cause a new resource to be created since grants have no way of being updated. 
  - What I did, which I saw done in another resource file, is if the read doesn't return valid grant or retiree principals then don't update the state file for those attributes and log a warning. This will only happen in two cases ~1 minute after iam principal creation (aws acknoleged eventual consistency problem) and if one of the principals has been deleted (in cases where it is managed outside of terraform).

If merged this resolves #641

I have run the acceptance tests in their current form, and they pass:

```
$ make testacc TESTARGS='-run=TestAWSKmsGrant'
TF_ACC=1 go test ./... -v -run=TestAWSKmsGrant -timeout 120m
=== RUN   TestAWSKmsGrant_Basic
--- PASS: TestAWSKmsGrant_Basic (39.08s)
=== RUN   TestAWSKmsGrant_withConstraints
--- PASS: TestAWSKmsGrant_withConstraints (30.75s)
=== RUN   TestAWSKmsGrant_withRetiringPrincipal
--- PASS: TestAWSKmsGrant_withRetiringPrincipal (39.39s)
=== RUN   TestAWSKmsGrant_bare
--- PASS: TestAWSKmsGrant_bare (38.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	148.205s
```

I have also run the unit tests and they pass:

```
$ make test
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1.765s
```

### DIFF
--- a/aws/resource_aws_kms_grant.go
+++ b/aws/resource_aws_kms_grant.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"strings"
@@ -9,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
@@ -51,22 +53,25 @@ func resourceAwsKmsGrant() *schema.Resource {
 				ForceNew: true,
 			},
 			"constraints": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
+				Set:      resourceKmsGrantConstraintsHash,
 				Optional: true,
 				ForceNew: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"encryption_context_equals": {
-							Type:          schema.TypeMap,
-							Optional:      true,
-							Elem:          schema.TypeString,
-							ConflictsWith: []string{"constraints.0.encryption_context_subset"},
+							Type:     schema.TypeMap,
+							Optional: true,
+							ForceNew: true,
+							Elem:     schema.TypeString,
+							// ConflictsWith encryption_context_subset handled in Create, see kmsGrantConstraintsIsValid
 						},
 						"encryption_context_subset": {
-							Type:          schema.TypeMap,
-							Optional:      true,
-							Elem:          schema.TypeString,
-							ConflictsWith: []string{"constraints.0.encryption_context_equals"},
+							Type:     schema.TypeMap,
+							Optional: true,
+							ForceNew: true,
+							Elem:     schema.TypeString,
+							// ConflictsWith encryption_context_equals handled in Create, see kmsGrantConstraintsIsValid
 						},
 					},
 				},
@@ -108,7 +113,10 @@ func resourceAwsKmsGrantCreate(d *schema.ResourceData, meta interface{}) error {
 		input.Name = aws.String(v.(string))
 	}
 	if v, ok := d.GetOk("constraints"); ok {
-		input.Constraints = expandKmsGrantConstraints(v.([]interface{}))
+		if !kmsGrantConstraintsIsValid(v.(*schema.Set)) {
+			return fmt.Errorf("[ERROR] A grant constraint can't have both encryption_context_equals and encryption_context_subset set")
+		}
+		input.Constraints = expandKmsGrantConstraints(v.(*schema.Set))
 	}
 	if v, ok := d.GetOk("retiring_principal"); ok {
 		input.RetiringPrincipal = aws.String(v.(string))
@@ -367,14 +375,40 @@ func findKmsGrantById(conn *kms.KMS, keyId string, grantId string, marker *strin
 	return nil, NewKmsGrantMissingError(fmt.Sprintf("[DEBUG] Grant %s not found for key id: %s", grantId, keyId))
 }
 
-func expandKmsGrantConstraints(configured []interface{}) *kms.GrantConstraints {
-	if len(configured) < 1 {
+// Can't have both constraint options set:
+// ValidationException: More than one constraint supplied
+// NB: set.List() returns an empty map if the constraint is not set, filter those out
+// using len(v) > 0
+func kmsGrantConstraintsIsValid(constraints *schema.Set) bool {
+	constraintCount := 0
+	for _, raw := range constraints.List() {
+		data := raw.(map[string]interface{})
+		if v, ok := data["encryption_context_equals"].(map[string]interface{}); ok {
+			if len(v) > 0 {
+				constraintCount += 1
+			}
+		}
+		if v, ok := data["encryption_context_subset"].(map[string]interface{}); ok {
+			if len(v) > 0 {
+				constraintCount += 1
+			}
+		}
+	}
+
+	if constraintCount > 1 {
+		return false
+	}
+	return true
+}
+
+func expandKmsGrantConstraints(configured *schema.Set) *kms.GrantConstraints {
+	if len(configured.List()) < 1 {
 		return nil
 	}
 
 	var constraint kms.GrantConstraints
 
-	for _, raw := range configured {
+	for _, raw := range configured.List() {
 		data := raw.(map[string]interface{})
 		if contextEq, ok := data["encryption_context_equals"]; ok {
 			constraint.SetEncryptionContextEquals(stringMapToPointers(contextEq.(map[string]interface{})))
@@ -387,21 +421,49 @@ func expandKmsGrantConstraints(configured []interface{}) *kms.GrantConstraints {
 	return &constraint
 }
 
-func flattenKmsGrantConstraints(constraint *kms.GrantConstraints) []interface{} {
+func concatStringMap(m map[string]*string, sep string) string {
+	var strList []string
+	for k, v := range m {
+		strList = append(strList, k, *v)
+	}
+	return strings.Join(strList, sep)
+}
+
+// The hash needs to encapsulate what type of constraint it is
+// as well as the keys and values of the constraint.
+func resourceKmsGrantConstraintsHash(v interface{}) int {
+	var buf bytes.Buffer
+	m, castOk := v.(map[string]interface{})
+	if !castOk {
+		return 0
+	}
+
+	if v, ok := m["encryption_context_equals"]; ok {
+		buf.WriteString(fmt.Sprintf("encryption_context_equals-%s", concatStringMap(stringMapToPointers(v.(map[string]interface{})), " ")))
+	}
+	if v, ok := m["encryption_context_subset"]; ok {
+		buf.WriteString(fmt.Sprintf("encryption_context_subset-%s", concatStringMap(stringMapToPointers(v.(map[string]interface{})), " ")))
+	}
+
+	return hashcode.String(buf.String())
+}
+
+func flattenKmsGrantConstraints(constraint *kms.GrantConstraints) *schema.Set {
+	constraints := &schema.Set{F: resourceKmsGrantConstraintsHash}
 	if constraint == nil {
-		return []interface{}{}
+		return constraints
 	}
 
 	m := make(map[string]interface{}, 0)
-
 	if constraint.EncryptionContextEquals != nil {
 		m["encryption_context_equals"] = pointersMapToStringList(constraint.EncryptionContextEquals)
 	}
 	if constraint.EncryptionContextSubset != nil {
 		m["encryption_context_subset"] = pointersMapToStringList(constraint.EncryptionContextSubset)
 	}
+	constraints.Add(m)
 
-	return []interface{}{m}
+	return constraints
 }
 
 // Custom error, so we don't have to rely on
--- a/aws/resource_aws_kms_grant.go
+++ b/aws/resource_aws_kms_grant.go
@@ -1,0 +1,418 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsKmsGrant() *schema.Resource {
+	return &schema.Resource{
+		// There is no API for updating/modifying grants, hence no Update
+		// Instead changes to most fields will force a new resource
+		Create: resourceAwsKmsGrantCreate,
+		Read:   resourceAwsKmsGrantRead,
+		Delete: resourceAwsKmsGrantDelete,
+		Exists: resourceAwsKmsGrantExists,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAwsKmsGrantName,
+			},
+			"key_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"grantee_principal": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateArn,
+			},
+			"operations": {
+				Type: schema.TypeList,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateAwsKmsGrantOperation,
+				},
+				Required: true,
+				ForceNew: true,
+			},
+			"constraints": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"encryption_context_equals": {
+							Type:          schema.TypeMap,
+							Optional:      true,
+							Elem:          schema.TypeString,
+							ConflictsWith: []string{"constraints.0.encryption_context_subset"},
+						},
+						"encryption_context_subset": {
+							Type:          schema.TypeMap,
+							Optional:      true,
+							Elem:          schema.TypeString,
+							ConflictsWith: []string{"constraints.0.encryption_context_equals"},
+						},
+					},
+				},
+			},
+			"retiring_principal": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateArn,
+			},
+			"grant_creation_tokens": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+				ForceNew: true,
+			},
+			"grant_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"grant_token": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceAwsKmsGrantCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).kmsconn
+
+	input := kms.CreateGrantInput{
+		GranteePrincipal: aws.String(d.Get("grantee_principal").(string)),
+		KeyId:            aws.String(d.Get("key_id").(string)),
+		Operations:       expandStringList(d.Get("operations").([]interface{})),
+	}
+
+	if v, ok := d.GetOk("name"); ok {
+		input.Name = aws.String(v.(string))
+	}
+	if v, ok := d.GetOk("constraints"); ok {
+		input.Constraints = expandKmsGrantConstraints(v.([]interface{}))
+	}
+	if v, ok := d.GetOk("retiring_principal"); ok {
+		input.RetiringPrincipal = aws.String(v.(string))
+	}
+	if v, ok := d.GetOk("grant_creation_tokens"); ok {
+		input.GrantTokens = expandStringList(v.([]interface{}))
+	}
+
+	log.Printf("[DEBUG]: Adding new KMS Grant: %s", input)
+
+	var out *kms.CreateGrantOutput
+
+	err := resource.Retry(2*time.Minute, func() *resource.RetryError {
+		var err error
+
+		out, err = conn.CreateGrant(&input)
+
+		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
+				// Error Codes: https://docs.aws.amazon.com/sdk-for-go/api/service/kms/#KMS.CreateGrant
+				// Under some circumstances a newly created IAM Role doesn't show up and causes
+				// an InvalidArnException to be thrown. TODO: Possibly change the aws_iam_role code?
+				if awsErr.Code() == "DependencyTimeoutException" ||
+					awsErr.Code() == "InternalException" ||
+					awsErr.Code() == "InvalidArnException" {
+					return resource.RetryableError(
+						fmt.Errorf("[WARN] Error adding new KMS Grant for key: %s, retrying %s",
+							*input.KeyId, err))
+				}
+			}
+			log.Printf("[ERROR] An error occured creating new AWS KMS Grant: %s", err)
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Created new KMS Grant: %s", *out.GrantId)
+	d.SetId(*out.GrantId)
+	d.Set("grant_id", out.GrantId)
+	d.Set("grant_token", out.GrantToken)
+
+	return resourceAwsKmsGrantRead(d, meta)
+}
+
+func resourceAwsKmsGrantRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).kmsconn
+
+	grantId := d.Id()
+	keyId := d.Get("key_id").(string)
+
+	log.Printf("[DEBUG] Looking for grant id: %s", grantId)
+	grant, err := findKmsGrantByIdWithRetry(conn, keyId, grantId)
+
+	if err != nil {
+		return err
+	}
+	if grant != nil {
+		// The grant sometimes contains principals that identified by their unique id: "AROAJYCVIVUZIMTXXXXX"
+		// instead of "arn:aws:...", in this case don't update the state file
+		if strings.HasPrefix(*grant.GranteePrincipal, "arn:aws") {
+			d.Set("grantee_principal", grant.GranteePrincipal)
+		} else {
+			log.Printf(
+				"[WARN] Unable to update grantee principal state %s for grant id %s for key id %s.",
+				*grant.GranteePrincipal, grantId, keyId)
+		}
+
+		if grant.RetiringPrincipal != nil {
+			if strings.HasPrefix(*grant.RetiringPrincipal, "arn:aws") {
+				d.Set("retiring_principal", grant.RetiringPrincipal)
+			} else {
+				log.Printf(
+					"[WARN] Unable to update retiring principal state %s for grant id %s for key id %s",
+					*grant.RetiringPrincipal, grantId, keyId)
+			}
+		}
+
+		d.Set("operations", grant.Operations)
+		if *grant.Name != "" {
+			d.Set("name", grant.Name)
+		}
+		if grant.Constraints != nil {
+			d.Set("constraints", flattenKmsGrantConstraints(grant.Constraints))
+		}
+	} else {
+		log.Printf("[WARN] %s KMS grant id not found for key id %s, removing from state file", grantId, keyId)
+		d.SetId("")
+	}
+
+	return nil
+}
+
+// Retiring grants requires special permissions (i.e. the
+// caller to be root, retiree principal, or grantee principal with retire grant
+// privileges). So just revoke grants.
+func resourceAwsKmsGrantDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).kmsconn
+
+	grantId := d.Get("grant_id").(string)
+	keyId := d.Get("key_id").(string)
+	input := kms.RevokeGrantInput{
+		GrantId: aws.String(grantId),
+		KeyId:   aws.String(keyId),
+	}
+
+	log.Printf("[DEBUG] Revoking KMS grant: %s", grantId)
+	_, err := conn.RevokeGrant(&input)
+
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Checking if grant is revoked: %s", grantId)
+	err = waitForKmsGrantToBeRevoked(conn, keyId, grantId)
+
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func resourceAwsKmsGrantExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	conn := meta.(*AWSClient).kmsconn
+
+	grantId := d.Id()
+	keyId := d.Get("key_id").(string)
+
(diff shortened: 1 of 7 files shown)


jammerful:
Resolves #641

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,414 @@
...
+		out, err = conn.CreateGrant(&input)
+
+		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
+				// Error Codes: https://docs.aws.amazon.com/sdk-for-go/api/service/kms/#KMS.CreateGrant
+				// Under some circumstances a newly created IAM Role doesn't show up and causes
+				// an InvalidArnException to be thrown. TODO: Possibly change the aws_iam_role code?
+				if awsErr.Code() == "DependencyTimeoutException" ||
+					awsErr.Code() == "InternalException" ||
+					awsErr.Code() == "InvalidArnException" {
```
jammerful: This is the first issue I mentioned.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,414 @@
...
+
+		// The grant sometimes contains principals that identified by their unique id: "AROAJYCVIVUZIMTXXXXX"
+		// instead of "arn:aws:..."
+		// This causes the grant to get into a bad state with the wrong principals that would cause resource re-creation after refresh
+		// https://docs.aws.amazon.com/kms/latest/APIReference/API_CreateGrant.html#API_CreateGrant_RequestSyntax
+		// states only ARNs are valid for both the grantee & retiring principals
+		// Confirmed with AWS that this is expected behavior.
+		// TODO: Handle upstream or here?
+		if grant != nil {
+			if !strings.HasPrefix(*grant.GranteePrincipal, "arn:aws") {
```
jammerful: This is the second issue I mentioned.

jammerful:
@radeksimko @bflad Can you please take a look at this soon?

kikitux:
Sorry for the delay, finding someone for a review

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+	return grant, err
+}
+
+// Used by the tests as well
+func waitForKmsGrantToBeRevoked(conn *kms.KMS, keyId string, grantId string) error {
+	err := resource.Retry(2*time.Minute, func() *resource.RetryError {
+		grant, err := findKmsGrantById(conn, keyId, grantId, nil)
+		if err != nil {
+			if _, ok := err.(KmsGrantMissingError); ok {
+				if grant == nil {
```
paddycarver: In what condition would the grant be non-nil and the error be non-nil?

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+	m := make(map[string]interface{}, 0)
+
+	if constraint.EncryptionContextEquals != nil {
+		m["encryption_context_equals"] = pointersMapToStringList(constraint.EncryptionContextEquals)
+	}
+	if constraint.EncryptionContextSubset != nil {
+		m["encryption_context_subset"] = pointersMapToStringList(constraint.EncryptionContextSubset)
+	}
+
+	return []interface{}{m}
```
paddycarver: Can we only have one constraint? Why does the expand version loop over it then?

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+}
+
+// Custom error, so we don't have to rely on
+// the content of an error message
+type KmsGrantMissingError struct {
```
paddycarver: This could just be a string. `type KmsGrantMissingError string`.

aws/resource_aws_kms_grant_test.go
```diff
@@ -0,0 +1,238 @@
...
+%s
+
+resource "aws_iam_role" "tf-acc-test-role" {
+  name               = "tf-acc-test-kms-grant-role-%s"
+  path               = "/service-role/"
+  assume_role_policy = "${data.aws_iam_policy_document.assumerole-policy-template.json}"
+}
+
+resource "aws_kms_grant" "%s" {
+	name = "%s"
```
paddycarver: does this need to be unique?

aws/validators.go
```diff
@@ -1395,6 +1395,43 @@ func validateAwsKmsName(v interface{}, k string) (ws []string, es []error) {
...
+		"ReEncryptFrom":                   true,
+		"ReEncryptTo":                     true,
+		"CreateGrant":                     true,
+		"RetireGrant":                     true,
+		"DescribeKey":                     true,
+	}
+
+	if _, ok := validTypes[value]; !ok {
+		es = append(es, fmt.Errorf(
+			"%s must be one of the following: [ Decrypt, Encrypt, GenerateDataKey, GenerateDataKeyWithoutPlaintext, ReEncryptFrom, ReEncryptTo, CreateGrant, RetireGrant, DescribeKey ]", k))
```
paddycarver: Could we programmatically generate this list, to prevent drift?

aws/validators.go
```diff
@@ -1395,6 +1395,43 @@ func validateAwsKmsName(v interface{}, k string) (ws []string, es []error) {
 	return
 }
 
+func validateAwsKmsGrantOperation(v interface{}, k string) (ws []string, es []error) {
+	value := v.(string)
+
+	validTypes := map[string]bool{
```
paddycarver: This could be a `map[string]struct{}` with each value set to `struct{}{}`, just to denote that the value of the map has no meaning.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+
+	log.Printf("[DEBUG] Looking for grant id: %s", grantId)
+	grant, err := findKmsGrantByIdWithRetry(conn, keyId, grantId)
+
+	if err != nil {
+		return err
+	}
+	if grant != nil {
+		// The grant sometimes contains principals that identified by their unique id: "AROAJYCVIVUZIMTXXXXX"
+		// instead of "arn:aws:...", in this case don't update the state file
```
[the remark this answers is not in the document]
paddycarver: Oh, I see, you acknowledged this in the PR body. In this case, I think it'd be better to retry until we get an ID in the `arn:aws:` format (or a max threshold is hit) than to not set state.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+				ForceNew: true,
+			},
+			"grantee_principal": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateArn,
+			},
+			"operations": {
+				Type: schema.TypeList,
```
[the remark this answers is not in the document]
jammerful: The answer to all of these is no. Will make a set, I think the real benefit here is the set avoids duplication which could cause errors upon creation.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateAwsKmsGrantOperation,
+				},
+				Required: true,
+				ForceNew: true,
+			},
+			"constraints": {
+				Type:     schema.TypeList,
```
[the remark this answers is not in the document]
jammerful: Will make a set.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"encryption_context_equals": {
+							Type:          schema.TypeMap,
+							Optional:      true,
+							Elem:          schema.TypeString,
+							ConflictsWith: []string{"constraints.0.encryption_context_subset"},
```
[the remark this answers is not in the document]
jammerful: It worked it my testing, I'm trying not allow both encryption_context_subset and encryption_context_equals both set, since the AWS KMS Grant API does not allow that. I'll have to change how that looks with the move to sets. Make sense?

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+				},
+			},
+			"retiring_principal": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateArn,
+			},
+			"grant_creation_tokens": {
+				Type:     schema.TypeList,
```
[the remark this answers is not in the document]
jammerful: Will move to a set, as the tokens should be unique. I don't see this a generating a lot of value, but I'll move it to a set.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+				*grant.GranteePrincipal, grantId, keyId)
+		}
+
+		if grant.RetiringPrincipal != nil {
+			if strings.HasPrefix(*grant.RetiringPrincipal, "arn:aws") {
+				d.Set("retiring_principal", grant.RetiringPrincipal)
+			} else {
+				log.Printf(
+					"[WARN] Unable to update retiring principal state %s for grant id %s for key id %s",
+					*grant.RetiringPrincipal, grantId, keyId)
```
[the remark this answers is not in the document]
jammerful: See the above comment.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+		}
+
+		d.Set("operations", grant.Operations)
```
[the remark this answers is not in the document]
jammerful: Will fix, thanks.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+		if *grant.Name != "" {
+			d.Set("name", grant.Name)
+		}
+		if grant.Constraints != nil {
+			d.Set("constraints", flattenKmsGrantConstraints(grant.Constraints))
```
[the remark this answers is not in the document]
jammerful: Will fix, thanks.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+				d.Set("retiring_principal", grant.RetiringPrincipal)
+			} else {
+				log.Printf(
+					"[WARN] Unable to update retiring principal state %s for grant id %s for key id %s",
+					*grant.RetiringPrincipal, grantId, keyId)
+			}
+		}
+
+		d.Set("operations", grant.Operations)
+		if *grant.Name != "" {
```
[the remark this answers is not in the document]
jammerful: No, the grant name is optional, however if you don't set it AWS will set it to "". However, if I was to  load that into the state it would cause the state to mismatch the specified state which would cause resource re-creation. So if the grant name is empty just ignore it.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+			d.Set("name", grant.Name)
+		}
+		if grant.Constraints != nil {
+			d.Set("constraints", flattenKmsGrantConstraints(grant.Constraints))
+		}
+	} else {
```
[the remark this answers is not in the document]
jammerful: Will fix, thanks.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+
+	grantId := d.Id()
+	keyId := d.Get("key_id").(string)
+
+	log.Printf("[DEBUG] Looking for Grant: %s", grantId)
+	grant, err := findKmsGrantByIdWithRetry(conn, keyId, grantId)
+
+	if err != nil {
+		if grant != nil {
+			return true, err
```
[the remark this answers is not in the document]
jammerful: Will fix, thanks.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+	if grant != nil {
+		return true, err
+	}
+
+	return false, err
+}
+
+func getKmsGrantById(grants []*kms.GrantListEntry, grantIdentifier string) *kms.GrantListEntry {
+	for idx := range grants {
```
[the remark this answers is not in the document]
jammerful: Will fix, thanks.

aws/resource_aws_kms_grant.go
jammerful: There isn't one, I'll remove the innermost if statement and just return nil (see findKmsGrantById).

aws/validators.go
jammerful: Will pull from the above map above thanks. By the way this kind of duplication is pretty common in this file, is probably worth cleaning up.

kikitux: lets follow the suggestion in this PR, your suggestion is welcome but lets focus the conversation on the PR it self 

alvaro

aws/validators.go
jammerful: Why is this better than map[string]bool? Just to denote the value of the map has no meaning?

aws/resource_aws_kms_grant.go
jammerful: Will fix.

aws/resource_aws_kms_grant.go
jammerful: Only one of encryption_context_subset or encrytpion_context_equals can be set, but that itself can be a map. Will have to refactor this as part of the move to sets.

jammerful:
@paddycarver Thanks for the review. I left some comments, most of your concerns will be fixed. Except for the one about not loading in IAM principals into terraform state if it doesn't resolve to an ARN--I left some reasons on why retrying isn't optimal (maybe if we do something like if d.isNewResource but even then you really want that to be on the IAM principal not if the grant is new).

I did have one other concern, about the scalability of this approach. I was stress testing my code, and it can handle about ~200 grants managed in one resource file, if it is more the resource refreshing takes too long. The reason for the limit, is that for each resource/grant we are calling list grants, which will get throttled by AWS. Would it make sense to have a global list of grants? Have you seen other resources implement this (resources that don't have a Describe* call)? I could see this getting messy.

aws/validators.go
jammerful: No problem, was just bringing a small possible issue to your attention.

aws/resource_aws_kms_grant.go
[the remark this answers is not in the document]
paddycarver: Yeah, I think I get what you were trying to do here, the syntax is just a bit :/ It may be worthwhile to just implement that logic manually to throw an error in the Create function.

aws/resource_aws_kms_grant.go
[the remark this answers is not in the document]
paddycarver: The main value is that if the API ever changes the order, or the user changes the order, the grant doesn't need to be recreated for no real purpose.

aws/resource_aws_kms_grant_test.go
paddycarver: Awesome, thanks! I just wanted to be sure we wouldn't run into an error if the resource got orphaned, which we usually try to avoid using randomized names. Sounds like we don't need a randomized name, here, though.

paddycarver:
Thanks for the work on this and the careful explanations! For the scale part... The only solution I can think of at the moment is to introduce user-specified timeouts, like this: https://github.com/terraform-providers/terraform-provider-aws/blob/177a19851fbbf8ec6a54f6733312ae205d7327d3/aws/resource_aws_ami_from_instance.go#L30-L34

But I'm toying with the idea of just waiting to see if people hit that timeout case, and then adding the custom timeouts if that turns into an issue people have. I don't know how common it is, unfortunately, to create 200 grants.

There were murmurs and talks of eventually adding some sort of request batching capabilities to Terraform, but I don't think that's something people were talking about for the near future. Though that would ideally be the "real" solution to this, reusing a single List call across resources.

Unfortunately, I think trying to build batching in using a single List call at the moment would be buggy and error-prone, so I'd advise against it unless we discover we absolutely must implement it.

aws/resource_aws_kms_grant.go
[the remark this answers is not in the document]
jammerful: We could store the internal ID. 
But think of this case: the only thing is that typically Create returns the Read function, which will in cases of new IAM roles cause internal IDs to be stored. Then when lets say some just applies terraform again, the change from internal ID to ARN will cause new resources to be created.

Let me ensure that there isn't something clever I'm missing, I'll ask a few people I know that have dealt with these eventual consistency issues. Would it be possible for you to ask a few co-workers as well?

aws/resource_aws_kms_grant.go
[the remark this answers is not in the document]
jammerful: I'm almost done with my revision, so I'll update the PR tomorrow. Sorry for the delay, the hash function for constraints now that it is a set is causing me a little heartache.

jammerful:
I may add user specific timeouts, is that fine? I think I'll have a use for them myself.

jammerful:
@paddycarver Addressed your comments. Please take a look.

Regarding sometimes not setting the roles in state, the only thing I could think of is to always store the role id in the state file. However, I this won't work if the role is from another account since the caller won't have GetRole permissions on that role from another account.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,479 @@
...
+	var strList []string
+	for k, v := range m {
+		strList = append(strList, k, *v)
+	}
+	return strings.Join(strList, sep)
+}
+
+// The hash needs to encapsulate what type of constraint it is
+// as well as the keys and values of the constraint.
+func resourceKmsGrantConstraintsHash(v interface{}) int {
```
jammerful: I need to fix this up. As of now, I know I at least need to sort the maps so that the constraint gets consistent hashes. I couldn't find any common utilities for this.

jammerful: Fixed.

jammerful:
@paddycarver I think all of the initial comments have been resolved. Please take another look. 

I still don't have a sense if not setting the roles in state is a non-starter/blocker. Keep in mind there is no way to modify a grant, so I wouldn't be extremely concerned about drift between the state file and what is in AWS.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+		GranteePrincipal: aws.String(d.Get("grantee_principal").(string)),
+		KeyId:            aws.String(d.Get("key_id").(string)),
+		Operations:       expandStringSet(d.Get("operations").(*schema.Set)),
+	}
+
+	if v, ok := d.GetOk("name"); ok {
+		input.Name = aws.String(v.(string))
+	}
+	if v, ok := d.GetOk("constraints"); ok {
+		if !kmsGrantConstraintsIsValid(v.(*schema.Set)) {
```
jammerful: Not sure how much value this is delivering, I really liked when we used conflicts with in the schema because it validates this before going out to AWS. Right now this is caught right before AWS would let us know of the conflict, after other resources have already been created.

jammerful:
@paddycarver @kikitux If you can't look at this soon could I please get an indication of whether you are inclined to merge this, or are there major issues preventing this from being merged?

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateArn,
+			},
+			"operations": {
+				Type: schema.TypeSet,
+				Set:  schema.HashString,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateAwsKmsGrantOperation,
```
bflad: We can avoid a custom validation function and associated unit test by using Terraform core's `validation.StringInSlice()` 😄  (it should also use SDK provided constants)

```go
  kms.GrantOperationCreateGrant,
  kms.GrantOperationDecrypt,
  kms.GrantOperationDescribeKey,
  kms.GrantOperationEncrypt,
  kms.GrantOperationGenerateDataKey,
  kms.GrantOperationGenerateDataKeyWithoutPlaintext,
  kms.GrantOperationReEncryptFrom,
  kms.GrantOperationReEncryptTo,
  kms.GrantOperationRetireGrant,
}, false),
```

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+
+	var out *kms.CreateGrantOutput
+
+	err := resource.Retry(3*time.Minute, func() *resource.RetryError {
+		var err error
+
+		out, err = conn.CreateGrant(&input)
+
+		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
```
bflad: You can avoid dealing with `awserr` directly by using the `isAWSErr()` helper function we have. 😄  (similarly the SDK usually provides constants for the error codes)

```go
if isAWSErr(err, kms.ErrCodeDependencyTimeoutException, "") {
```

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+	grant, err := findKmsGrantByIdWithRetry(conn, keyId, grantId)
+
+	if err != nil {
+		return err
+	}
+
+	if grant == nil {
+		log.Printf("[WARN] %s KMS grant id not found for key id %s, removing from state file", grantId, keyId)
+		d.SetId("")
+	}
```
bflad: This should `return nil` so the `if grant != nil {` below (and its nesting) is unnecessary.

copumpkin:
> Unfortunately, I think trying to build batching in using a single List call at the moment would be buggy and error-prone, so I'd advise against it unless we discover we absolutely must implement it.

@paddycarver if I understand correctly, without batching we get basically quadratic (actually `O(m * n)` where `m` is number of TF grant resources and `n` is number of grants in the account, which after the first time is always `>= m`). My concern here is that unlike other resources with similar issues (I think KMS aliases), grants are often quite numerous. One of my keys today has several hundred grants on it, the default limit per CMK is 2500 (and I've had it raised by AWS support in the past), and I'd like to manage many of those grants with KMS. That seems like it'll become untenable pretty quickly, unless you have a small handful of grants. But with a small number of grants, you might as well just encode it in the key policy.

I get that Terraform doesn't have a proper generic solution for this problem yet, but it doesn't seem awful to maintain a bit of state across grant resource calls so they don't each have to list grants from scratch.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+			if err := d.Set("constraints", flattenKmsGrantConstraints(grant.Constraints)); err != nil {
+				log.Printf("[DEBUG] Error setting constraints for grant %s with error %s", grantId, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+// Retiring grants requires special permissions (i.e. the
```
bflad: Seems like this behavior should be noted in the documentation. Better yet, it would probably be best to provide an attribute flag to control the behavior (e.g. `retire_on_delete`) that defaults to `true` so we can always return an error if an expected deletion (retire or just revoke) is not possible.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	log.Printf("[DEBUG] Checking if grant is revoked: %s", grantId)
+	err = waitForKmsGrantToBeRevoked(conn, keyId, grantId)
+
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
```
bflad: Nitpick: `d.SetId("")` is no longer required in delete functions

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+	input := kms.RevokeGrantInput{
+		GrantId: aws.String(grantId),
+		KeyId:   aws.String(keyId),
+	}
+
+	log.Printf("[DEBUG] Revoking KMS grant: %s", grantId)
+	_, err := conn.RevokeGrant(&input)
+
+	if err != nil {
+		return err
```
bflad: We should allow non-existent grants to skip returning an error so its not required to do a `terraform state rm`:

```go
if isAWSErr(err, kms.ErrCodeNotFoundException, "") {
  return nil
}
```

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+		}
+	}
+	constraints.Add(m)
+
+	return constraints
+}
+
+// Custom error, so we don't have to rely on
+// the content of an error message
+type KmsGrantMissingError string
```
bflad: Instead of creating our own error class, can we instead utilize the upstream SDK's `awserr.Error` and set to a code like `kms.ErrCodeNotFoundException`?

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,502 @@
...
+		}
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Created new KMS Grant: %s", *out.GrantId)
+	d.SetId(*out.GrantId)
```
bflad: Given that grant IDs are really under a key ID, we might want to consider making the resource ID in Terraform contain both, e.g. `KeyID:GrantID`. This will allow us to import them.

Saving as part of the resource identifier:

```go
// Using keyID := d.Get("key_id").(string) above input
d.SetId(fmt.Sprintf("%s:%s", keyID, aws.StringValue(out.GrantId)))
```

Parsing anytime necessary:

```
keyID, grantID, err := decodeKmsGrantId(d.Id())
if err != nil {
  return err
}
// ...
func decodeKmsGrantId(id string) (string, string, error) {
	parts := strings.Split(id, ":")
	if len(parts) != 2 {
		return "", "", fmt.Errorf("Unexpected format of ID (%q), expected KeyID:GrantID", id)
	}
	return parts[0], parts[1], nil
}
```

aws/resource_aws_kms_grant.go
jammerful: Done.

aws/resource_aws_kms_grant.go
jammerful: Done.

aws/resource_aws_kms_grant.go
jammerful: Done.

aws/resource_aws_kms_grant.go
jammerful: Done. However, I inverted the default to false. Functionally retiring and revoking are the same, they don't show up in the ListGrants call, however retiring requires special permissions. It seemed to me that defaulting to that would be noisy and cause confusion.

aws/resource_aws_kms_grant.go
jammerful: Done. Didn't realize that.

aws/resource_aws_kms_grant.go
jammerful: Great catch. Done.

aws/resource_aws_kms_grant.go
jammerful: `kms.ErrCodeNotFoundException` is a string, and I wanted a type or code to use so I don't have to key off the error message. Might be missing something here?

aws/resource_aws_kms_grant.go
jammerful: Done.

paultyng:
@copumpkin I think the issue is what if there is a grant that depends on a grant, etc, things like that? If you needed the token to create another grant for example then that cache starts to work against you, which is why it hasn't been used in other places.

I think since the grants are listed by `key` you could just create an `aws_kms_grants` resource that worked in bulk potentially where the `ID` was the `key`?

copumpkin:
@paultyng that's an idea that occurred to us, but is there precedent for it in Terraform?

Also, is there a good place (like a TF contributors Slack) to chat live about this sort of thing? Might help sort through these concerns more quickly. There's a few options that have occurred to us but it's hard to factor in what might work vs. trying to guess what HashiCorp might be willing to merge.

aws/resource_aws_kms_grant.go
paddycarver: I'd love this to fail at plan time, too, but in the interest of moving this forward, I wanted to keep our scope here simple and not introduce complicated things like CustomDiff or ConflictsWith on a Set, which _I've_ never used on a Set before, and don't know the intricacies of how it works. @bflad or @paultyng may have more insight on that.

It's almost certainly possible to have this fail at plan time. I wouldn't consider making that happen a blocker, but if you'd like it to be in the first pass, I'm happy to have it included.

aws/resource_aws_kms_grant.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+	if *out.Truncated {
+		log.Printf("[DEBUG] KMS Grant list truncated, getting next page via marker: %s", *out.NextMarker)
+		return findKmsGrantById(conn, keyId, grantId, out.NextMarker)
+	}
+
+	return nil, NewKmsGrantMissingError(fmt.Sprintf("[DEBUG] Grant %s not found for key id: %s", grantId, keyId))
+}
+
+// Can't have both constraint options set:
```
paddycarver: I'll admit I'm confused by this part. You're only allowed to have 1 constraint, right? If you're only allowed 1 constraint, is the only reason constraints is a TypeSet for the block syntax? If that's the case, we _could_ use `MaxItems: 1` on `constraints` to make sure no more than 1 block is specified. Doesn't obviate the need for this function, but can simplify it a little.

**Note:** This is not a blocker for merge.

website/docs/r/kms_grant.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Optional, Forces new resources) A friendly name for identifying the grant.
+* `key_id` - (Required, Forces new resources) The unique identifier for the customer master key (CMK) that the grant applies to.
+* `grantee_principal` - (Required, Forces new resources) The principal that is given permission to perform the operations that the grant permits.
```
paddycarver: Could we document here the eventual consistency problem and workaround we have in place, with the ARN weirdness?

jammerful: Done.

aws/resource_aws_kms_grant.go
jammerful: Setting `MaxItems` won't work given the way the constraint set is being hashed, it ends up being one item with two different sub-attributes. Maybe I should have left it as a list, mostly had moved it on your suggestion.

jammerful:
@paddycarver Thank you for all the help and comments. I updated the documentation, and replied to your comments. Should be ready to merge. Let me know if you want me to squash the commits.

aws/resource_aws_kms_grant.go
jammerful: Will address in a follow up PR, as I mentioned in another comment. Might make the most sense to move this back to a list? The set adds a lot of additional code, as well as not adding much value I think given that there can only be one element.

paddycarver:
Looks great! Thanks for contributing to Terraform's AWS provider!

copumpkin:
Sweet, thanks! I also have an idea for how to get rid of the `O(m * n)` behavior that I'll propose in a separate issue and tag you on, if you don't mind.

paddycarver:
Feel free to tag me, and I'll do my best to get to it, but to set expectations appropriately: I am responsible for other providers, so it may take me a bit to get to it; this week is already looking unlikely. There's no harm in tagging me on it, but your best bet for getting eyes on it is to open it up to the maximum number of maintainers. :)

bflad:

PR #3042, opened 2018-01-18
add new appscaling validate options.

chrisLeeTW:
Hey,

This PR is just add missing PredefinedMetricSpecification & Scalable Dimension & Service Namespace Options for appscaling relateive config validation based on this [document](https://docs.aws.amazon.com/cli/latest/reference/application-autoscaling/put-scaling-policy.html)


bflad:
It turns out we removed these plan-time validation functions awhile ago in #3141 so no need to update them here 🙁

PR #3059, opened 2018-01-19
resource/aws_dynamodb_table: Detect and forbid changes to GSI keys

mdlavin:
I had a terraform change that was attempting to add a range key to a GSI that had only a hash key. The terraform plan step showed that the key would be added and the terraform apply showed that the addition was successful, but the GSI keys were not modified at all.

After looking at the code, I noticed at that
1. Changes to the index attributes are completely ignored
2. The AWS API doesn't allow changing the keys at all

Because the plan and apply worked without error it was possible for things to look successful when nothing had actually happened. This change catches the problem and alerts the user.

NOTE: There are *many* more cases of things that could be detect, like LSI key changes and other un-modifiable attributes, but I wanted to make sure that the general style of checking was accepted before trying to support every type of validation.

### DIFF
--- a/aws/resource_aws_dynamodb_table.go
+++ b/aws/resource_aws_dynamodb_table.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/awsutil"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/hashicorp/terraform/helper/hashcode"
 )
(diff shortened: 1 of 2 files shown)


mdlavin:
@radeksimko is there interest in merging this if I fix the merge conflicts?

mdlavin:
@radeksimko is there interest in merging this if I fix the merge conflicts?

mdlavin:
@radeksimko Any thoughts on this type of PR? I'm happy to fix the merge conflicts and open more associated validations for DynamoDB if you like this sort of change.

mdlavin:
I'm assuming that this will not be merged. I'm closing it so that it will be remove from my open PR list

PR #3063, opened 2018-01-19
resource/aws_emr_cluster: Add scale_down_behavior

csghuser:
This provides the ability to configure the scale down behavior of a cluster. From the docs:

> ScaleDownBehavior
> The way that individual Amazon EC2 instances terminate when an automatic scale-in activity occurs or an instance group is resized. TERMINATE_AT_INSTANCE_HOUR indicates that Amazon EMR terminates nodes at the instance-hour boundary, regardless of when the request to terminate the instance was submitted. This option is only available with Amazon EMR 5.1.0 and later and is the default for clusters created using that version. TERMINATE_AT_TASK_COMPLETION indicates that Amazon EMR blacklists and drains tasks from nodes before terminating the Amazon EC2 instances, regardless of the instance-hour boundary. With either behavior, Amazon EMR removes the least active nodes first and blocks instance termination if it could lead to HDFS corruption. TERMINATE_AT_TASK_COMPLETION available only in Amazon EMR version 4.1.0 and later, and is the default for versions of Amazon EMR earlier than 5.1.0.
> 
> Type: String
> 
> Valid Values: TERMINATE_AT_INSTANCE_HOUR | TERMINATE_AT_TASK_COMPLETION
> 
> Required: No

Source: https://docs.aws.amazon.com/emr/latest/APIReference/API_JobFlowDetail.html

I've added a test which attempts to set the value to TERMINATE_AT_INSTANCE_HOUR for a cluster version of 5.5.0. This is the default for this version, but I didn't want to interfere with the tests as it may prevent the cluster from shutting down should any tasks be running. 

I've not yet been able to run this against AWS to test.

### DIFF
--- a/aws/resource_aws_emr_cluster.go
+++ b/aws/resource_aws_emr_cluster.go
@@ -496,7 +496,11 @@ func resourceAwsEMRClusterRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.Set("name", cluster.Name)
-	d.Set("scale_down_behavior", cluster.ScaleDownBehavior)
+
+	if d.Get("scale_down_behavior").(string) != "" {
+		d.Set("scale_down_behavior", cluster.ScaleDownBehavior)
+	}
+
 	d.Set("service_role", cluster.ServiceRole)
 	d.Set("security_configuration", cluster.SecurityConfiguration)
 	d.Set("autoscaling_role", cluster.AutoScalingRole)
--- a/aws/resource_aws_emr_cluster.go
+++ b/aws/resource_aws_emr_cluster.go
@@ -236,6 +236,11 @@ func resourceAwsEMRCluster() *schema.Resource {
 				ForceNew: true,
 				Required: true,
 			},
+			"scale_down_behavior": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Optional: true,
+			},
 			"security_configuration": {
 				Type:     schema.TypeString,
 				ForceNew: true,
@@ -363,8 +368,13 @@ func resourceAwsEMRClusterCreate(d *schema.ResourceData, meta interface{}) error
 	if v, ok := d.GetOk("log_uri"); ok {
 		params.LogUri = aws.String(v.(string))
 	}
+
 	if v, ok := d.GetOk("autoscaling_role"); ok {
 		params.AutoScalingRole = aws.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("scale_down_behavior"); ok {
+		params.ScaleDownBehavior = aws.String(v.(string))
 	}
 
 	if v, ok := d.GetOk("security_configuration"); ok {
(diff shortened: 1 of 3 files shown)


csghuser:
I've been able to test this. The tests are failing however. If scale_down_behavior is not set (as it is optional), after cluster creation, when reading the state, it wants to set this value back to an empty string. 

```
 scale_down_behavior: "TERMINATE_AT_TASK_COMPLETION" => "" (forces new resource)
```

So whilst we are not setting this, it gets defined at cluster creation by AWS. I'm just looking through the code to see how this can be avoided.

csghuser:
I've added some logic to check whether or not the user has defined this new variable. If not then we rely on the default for the EMR version. This seems like a better approach, as opposed to hard coding the following logic into the code:

```
if release_label >=  5.1.0 
  default = TERMINATE_AT_INSTANCE_HOUR
else if (release_label >= 4.1.0 AND release_label < 5.1.0)
  default = TERMINATE_AT_TASK_COMPLETION
end
```

As its possible this may change at some point and a new default would be introduced with a later version.

csghuser:
@bflad Many thanks for your pointers, very much appreciated. I've addressed each one and reran the basic test, which seems to pass fine. It also passes if no scale_down_behavior config option is set now that the "computed" attribute is set, as you suggested :smiley:

Would it be possible to check if everything looks ok?

aws/resource_aws_emr_cluster.go
```diff
@@ -486,6 +502,7 @@ func resourceAwsEMRClusterRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.Set("name", cluster.Name)
+
```
bflad: We do still want the `d.Set("scale_down_behavior", cluster.ScaleDownBehavior)` here so Terraform is always updating its state correctly from the API each read as well as on import. We just didn't need the wrapping `if d.Get()` check that was previously in the PR. Sorry I wasn't more clear about what I meant earlier! 😄

aws/resource_aws_emr_cluster_test.go
```diff
@@ -24,7 +24,10 @@ func TestAccAWSEMRCluster_basic(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config: testAccAWSEmrClusterConfig(r),
-				Check:  testAccCheckAWSEmrClusterExists("aws_emr_cluster.tf-test-cluster", &cluster),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEmrClusterExists("aws_emr_cluster.tf-test-cluster", &cluster),
+					resource.TestCheckResourceAttrSet("aws_emr_cluster.tf-test-cluster", "scale_down_behavior"),
```
bflad: Since we know the value (hardcoded in the config below), lets actually check it. 😄 
```
resource.TestCheckResourceAttr("aws_emr_cluster.tf-test-cluster", "scale_down_behavior", "TERMINATE_AT_TASK_COMPLETION"),
```

csghuser:
@bflad I've made the modifications as requested 😃 I've reran the tests, seems to run ok.

bflad:

csghuser:
@bflad Thanks! 😄

PR #3091, opened 2018-01-22
Add missing slash in api_gateway_integration docs

cdimitroulas:
There was a missing `/` in the `Lambda Integration` code example in the docs.


cdimitroulas:
Nevermind this is actually correct.

PR #3123, opened 2018-01-24
resource/aws_api_gateway_integration: Allow update of content_handling attributes

mdlavin:
Instead of always forcing a new resource creation for content_handling changes, the integration:update action allows for in place updates.


bflad:

PR #3140, opened 2018-01-26
[WIP] Add new resource acm request certificate

pawelsocha:


bflad:
@pawelsocha before you get too far since you didn't leave any context in the original PR comment field, how does this relate/differ than the currently open PR #2813?

pawelsocha:
@bflad so, pr #2813 looks better and it's cover my problem. I will close this PR.

PR #3142, opened 2018-01-26
Support AWS cn-northwest-1 Ningxia (fixes #3053)

kwerey:
Hi Terraform team - this adds the new Ningxia region to the list of valid options for the AWS provider, fixing https://github.com/terraform-providers/terraform-provider-aws/issues/3053

A lot of the acceptance tests will fail run in Nginxia, since the region doesn't support a very large number of AWS services, but these changes let the provider can successfully make client connections to the region:

```
$ export AWS_DEFAULT_REGION=cn-northwest-1
$ make testacc TEST=.//aws TESTARGS='-run=TestAccAWSCloudWatchMetricAlarm_basic'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test .//aws -v -run=TestAccAWSCloudWatchMetricAlarm_basic -timeout 120m
=== RUN   TestAccAWSCloudWatchMetricAlarm_basic
--- PASS: TestAccAWSCloudWatchMetricAlarm_basic (55.58s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	56.088s
```

I've not read a single line of go source code before today, so please let me know if there's a neater way to add the region. Is there an "is in [list of regions]" comparison operator for strings? I didn't see one with a cursory google!

Thanks for making a cool project, I've spent a few days prototyping stuff in TF and I like it lot 👍

PR #3146, opened 2018-01-26
docs/CONTRIBUTING: New Region

bflad:
From PR review of #3142

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -253,6 +253,19 @@ into Terraform.
    The PR reviewers can help out on this front, and may provide comments with
    suggestions on how to improve the code.
 
+#### New Region
+
+Implementing a new region gives Terraform the ability to connect and interact
+with endpoints in a new geographic area. While baseline region support is fairly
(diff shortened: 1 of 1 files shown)


kwerey:
Hey, a minor other bit of feedback on these guidelines as a newbie: the example for running acceptance tests looks like it comes from a time when the provider lived in the main terraform repo: 
 
```
$ make testacc TEST=./builtin/providers/azurerm TESTARGS='-run=TestAccAzureRMPublicIpStatic_update'
==> Checking that code complies with gofmt requirements..
```
 
It's no big deal to find an aws alternative, but updating the example to something that'll work out the box with the AWS provider would lower the barrier to entry a touch by saving folks from having to break the command apart and parse how the directory structure's changed!

bflad:
That is excellent feedback! I will submit a PR to get the documentation updated. Thank you!

bflad:
@radeksimko you're absolutely right 👍  we should have proper handling for missing map entries (friendly error not crash). I think we identified at least one place where this definitely is an issue and I'll submit an issue for it and others if necessary. I have updated the documentation here say `if available` rather than adding an empty mapping entry.

PR #2189, opened 2017-11-06
resource/aws_api_gateway_authorizer: Add support for Cognito User Pool authorizer

josselin-c:
WIP/Help Needed

Add support for ProviderARNs parameter for the API Gateway Authorizer object
With this patch it's possible to use a Cognito User Pool as an Authorizer.

API Doc: http://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-authorizer.html

Fix #1106

PR #3156, opened 2018-01-27
resource/api_gateway_authorizer: support authorizer type as COGNITO_USER_POOLS

### LINKED ISSUE #1106: Support COGNITO_USER_POOLS authorizer
### Terraform Version
0.9.11

### Affected Resource(s)
- aws_api_gateway_authorizer
- aws_api_gateway_method

### Expected Behavior
type="COGNITO_USER_POOLS" supported

### Actual Behavior
not supported

### References
- http://docs.aws.amazon.com/cli/latest/reference/apigateway/create-authorizer.html
- http://docs.aws.amazon.com/cli/latest/reference/apigateway/put-method.html

loivis:
Fixes #1106
Duplicate of #2189 

I missed existing issue and pr when I started. Anyway I continued with this new pr.

**QUSTION**: `type` changed to `ForceNew`

Because `COGNITO_USER_POOLS` is different with `TOKEN`/`REQUEST`. But when changing between `TOKEN` and `REQUEST`, it doesn't required new resource. To conditionally `ForceNew`, the only way I can think of is to have some logic in `update` function that will detect `type` change and call `create` function. I was wondering if terraform has already supported conditional `ForceNew` someway or we just go with always `ForceNew`.

### DIFF
--- a/aws/resource_aws_api_gateway_authorizer.go
+++ b/aws/resource_aws_api_gateway_authorizer.go
@@ -81,6 +81,9 @@ func resourceAwsApiGatewayAuthorizerCreate(d *schema.ResourceData, meta interfac
 		Type:           aws.String(d.Get("type").(string)),
 	}
 
+	if err := validateAuthorizerType(d); err != nil {
+		return err
+	}
 	if v, ok := d.GetOk("authorizer_uri"); ok {
 		input.AuthorizerUri = aws.String(v.(string))
 	}
@@ -253,20 +256,6 @@ func resourceAwsApiGatewayAuthorizerDelete(d *schema.ResourceData, meta interfac
 }
 
 func resourceAwsApiGatewayAuthorizerCustomizeDiff(diff *schema.ResourceDiff, v interface{}) error {
-	authType := diff.Get("type").(string)
-	// authorizer_uri is required for authorizer TOKEN/REQUEST
-	if authType == apigateway.AuthorizerTypeRequest || authType == apigateway.AuthorizerTypeToken {
-		if val, ok := diff.GetOk("authorizer_uri"); !ok || val.(string) == "" {
-			return fmt.Errorf("authorizer_uri must be set non-empty when authorizer type is %s", authType)
-		}
-	}
-	// provider_arns is required for authorizer COGNITO_USER_POOLS.
-	if authType == apigateway.AuthorizerTypeCognitoUserPools {
-		if val, ok := diff.GetOk("provider_arns"); !ok || len(val.(*schema.Set).List()) == 0 {
-			return fmt.Errorf("provider_arns must be set non-empty when authorizer type is %s", authType)
-		}
-	}
-
 	// switch type between COGNITO_USER_POOLS and TOKEN/REQUEST will create new resource.
 	if diff.HasChange("type") {
 		o, n := diff.GetChange("type")
@@ -274,6 +263,24 @@ func resourceAwsApiGatewayAuthorizerCustomizeDiff(diff *schema.ResourceDiff, v i
 			if err := diff.ForceNew("type"); err != nil {
 				return err
 			}
+		}
+	}
+
+	return nil
+}
+
+func validateAuthorizerType(d *schema.ResourceData) error {
+	authType := d.Get("type").(string)
+	// authorizer_uri is required for authorizer TOKEN/REQUEST
+	if authType == apigateway.AuthorizerTypeRequest || authType == apigateway.AuthorizerTypeToken {
+		if v, ok := d.GetOk("authorizer_uri"); !ok || v.(string) == "" {
+			return fmt.Errorf("authorizer_uri must be set non-empty when authorizer type is %s", authType)
+		}
+	}
+	// provider_arns is required for authorizer COGNITO_USER_POOLS.
+	if authType == apigateway.AuthorizerTypeCognitoUserPools {
+		if v, ok := d.GetOk("provider_arns"); !ok || len(v.(*schema.Set).List()) == 0 {
+			return fmt.Errorf("provider_arns must be set non-empty when authorizer type is %s", authType)
 		}
 	}
 
--- a/aws/resource_aws_api_gateway_authorizer.go
+++ b/aws/resource_aws_api_gateway_authorizer.go
@@ -19,44 +19,44 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 		Delete: resourceAwsApiGatewayAuthorizerDelete,
 
 		Schema: map[string]*schema.Schema{
-			"authorizer_uri": &schema.Schema{
+			"authorizer_uri": {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-			"identity_source": &schema.Schema{
+			"identity_source": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "method.request.header.Authorization",
 			},
-			"name": &schema.Schema{
+			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"rest_api_id": &schema.Schema{
+			"rest_api_id": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
-			"type": &schema.Schema{
+			"type": {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "TOKEN",
 				ForceNew: true,
 			},
-			"authorizer_credentials": &schema.Schema{
+			"authorizer_credentials": {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-			"authorizer_result_ttl_in_seconds": &schema.Schema{
+			"authorizer_result_ttl_in_seconds": {
 				Type:         schema.TypeInt,
 				Optional:     true,
 				ValidateFunc: validateIntegerInRange(0, 3600),
 			},
-			"identity_validation_expression": &schema.Schema{
+			"identity_validation_expression": {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
-			"provider_arns": &schema.Schema{
+			"provider_arns": {
 				Type:     schema.TypeList,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
--- a/aws/resource_aws_api_gateway_authorizer.go
+++ b/aws/resource_aws_api_gateway_authorizer.go
@@ -21,7 +21,7 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"authorizer_uri": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 			},
 			"identity_source": &schema.Schema{
 				Type:     schema.TypeString,
@@ -41,6 +41,7 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "TOKEN",
+				ForceNew: true,
 			},
 			"authorizer_credentials": &schema.Schema{
 				Type:     schema.TypeString,
@@ -55,6 +56,11 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"provider_arns": &schema.Schema{
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayAuthorizer_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayAuthorizer_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayAuthorizer_basic
--- PASS: TestAccAWSAPIGatewayAuthorizer_basic (50.11s)
=== RUN   TestAccAWSAPIGatewayAuthorizer_cognito
--- PASS: TestAccAWSAPIGatewayAuthorizer_cognito (34.12s)
=== RUN   TestAccAWSAPIGatewayAuthorizer_switchAuthType
--- PASS: TestAccAWSAPIGatewayAuthorizer_switchAuthType (83.78s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	168.058s
```

loivis:
Also introduced `randString` in acceptance test, which makes code change more than needed for this pr. I think it's ok to include it here instead of opening another pr. No worth for that.

loivis:
@bflad Do you have time to take a look? Thanks.

bflad:
Not at the current time. Something that is an immediate red flag is this though:
> I missed existing issue and pr when I started. Anyway I continued with this new pr.

We really do not like ignoring someone's contributions and the original author is asking for help in their issue. Can you help them instead of ignoring their contribution?

loivis:
Ok, understand. I’ll try to see how to commit changes to the other pr and merge my codes there if possible. Anyway I’m closing this pr. Thanks for your comments and explanation.

mvh

loivis:
Back to this PR again with close of #2189.

bflad:
Thanks for the followup @loivis! Sorry if I came off a little mean earlier, that was not my intention.

loivis:
No need sorry, not at all. It's good to know how we're managing issue/prs and follow the same.

loivis:
@bflad do you have time to take a look? Thanks.

aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -55,6 +56,11 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"provider_arns": &schema.Schema{
```
bflad: Minor nitpick: the `&schema.Schema` here (copy paste 😄 ) is extra

aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -55,6 +56,11 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
...
+				Type:     schema.TypeList,
```
bflad: Does the ordering of the values matter for the `provider_arns` attribute? If not, we should probably switch this to `schema.TypeSet` 👍

aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -21,7 +21,7 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"authorizer_uri": &schema.Schema{
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
```
bflad: Sorry I'm unfamiliar with this resource, but is this required for `TOKEN`/`REQUEST`? Just curious if that was an oversight before, if we should add documentation, or `CustomizeDiff` for continued plan-time validation

aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -210,3 +228,38 @@ func resourceAwsApiGatewayAuthorizerDelete(d *schema.ResourceData, meta interfac
 
 	return nil
 }
+
+func diffProviderARNsOp(prefix string, old, new []interface{}) (ops []*apigateway.PatchOperation) {
+	// providerARNs can't be empty, so add first and then remove
```
bflad: Thank you for this comment 👍

aws/resource_aws_api_gateway_authorizer_test.go
```diff
@@ -8,60 +8,140 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/apigateway"
+	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 )
 
 func TestAccAWSAPIGatewayAuthorizer_basic(t *testing.T) {
 	var conf apigateway.Authorizer
+	rString := acctest.RandString(7)
```
bflad: 💯

website/docs/r/api_gateway_authorizer.html.markdown
```diff
@@ -113,3 +113,4 @@ The following arguments are supported:
 	For `TOKEN` type, this value should be a regular expression. The incoming token from the client is matched
 	against this expression, and will proceed if the token matches. If the token doesn't match,
 	the client receives a 401 Unauthorized response.
+* `provider_arns` - (Optional, required for type `COGNITO_USER_POOLS`) A list of the Amazon Cognito user pool ARNs. Each element is of this format: `arn:aws:cognito-idp:{region}:{account_id}:userpool/{user_pool_id}`.
```
bflad: Would you mind taking a look at adding a `CustomizeDiff` function that adds plan-time validation for this situation? Maybe something similar to `aws_dynamodb_table` function can work here too:

https://github.com/terraform-providers/terraform-provider-aws/blob/0cd8e5a0e6572740511f54c44c483a0549ecee2e/aws/validators.go#L2236

website/docs/r/api_gateway_authorizer.html.markdown
```diff
@@ -96,14 +96,14 @@ resource "aws_lambda_function" "authorizer" {
...
 	For `TOKEN` type, this must be a well-formed Lambda function URI in the form of
 	`arn:aws:apigateway:{region}:lambda:path/{service_api}`. e.g. `arn:aws:apigateway:us-west-2:lambda:path/2015-03-31/functions/arn:aws:lambda:us-west-2:012345678912:function:my-function/invocations`
 * `name` - (Required) The name of the authorizer
 * `rest_api_id` - (Required) The ID of the associated REST API
 * `identity_source` - (Optional) The source of the identity in an incoming request.
 	Defaults to `method.request.header.Authorization`. For `REQUEST` type, this may be a comma-separated list of values, including headers, query string parameters and stage variables - e.g. `"method.request.header.SomeHeaderName,method.request.querystring.SomeQueryStringName,stageVariables.SomeStageVariableName"`
-* `type` - (Optional) The type of the authorizer. Possible values are `TOKEN` and `REQUEST`.
+* `type` - (Optional) The type of the authorizer. Possible values are `TOKEN` for a Lambda function using a single authorization token submitted in a custom header, `REQUEST` for a Lambda function using incoming request parameters, or `COGNITO_USER_POOLS` for using an Amazon Cognito user pool.
```
bflad: 💯

aws/resource_aws_api_gateway_authorizer_test.go
```diff
@@ -292,29 +373,73 @@ EOF
...
+}
+`, apiGatewayName, apiGatewayName, apiGatewayName, lambdaName)
 }
-`
 
-const testAccAWSAPIGatewayAuthorizerConfig = testAccAWSAPIGatewayAuthorizerConfig_base + `
-resource "aws_api_gateway_authorizer" "test" {
-  name = "tf-acc-test-authorizer"
-  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
+func testAccAWSAPIGatewayAuthorizerConfig_lambda(apiGatewayName, authorizerName, lambdaName string) string {
```
bflad: 😍  thank you

aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -41,6 +41,7 @@ func resourceAwsApiGatewayAuthorizer() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "TOKEN",
+				ForceNew: true,
```
bflad: Can you add plan-time validation here please? 

```go
  apigateway.AuthorizerTypeCognitoUserPools,
  apigateway.AuthorizerTypeRequest,
  apigateway.AuthorizerTypeToken
}, false),
```

bflad:
To answer your question above:

> I was wondering if terraform has already supported conditional ForceNew someway or we just go with always ForceNew.

You can use a `CustomizeDiff` function to do that. 😄 You can check `diff.HasChange("type")` and if it does use `o,n := diff.GetChange("type")` to selectively call a `diff.ForceNew("type")` appropriately 🎉

aws/resource_aws_api_gateway_authorizer.go
loivis: @bflad Indeed `CustomizeDiff` should be used here. But it doesn't seem to get the argument. Did I miss something? 
https://github.com/loivis/terraform-provider-aws/blob/9726a5187a1fb34a8edbc660e0d3968d8e46a3f0/aws/resource_aws_api_gateway_authorizer.go#L277

bflad: I think you need to perform two checks, one for existence and one for string length:

```go
v, ok := diff.GetOk("authorizer_uri")
if ok && v.(string) == "" {
  return errors.New("authorizer_uri must be non-empty when type is ...")
} else {
  return errors.New("authorizer_uri is required when type is ...")
}
```

You'll need to do the same with `len() == 0` for provider_arns

loivis: I did try that. The problem is that `diff.GetOk("authorizer_uri")` didn't give me `true` for `ok`.

loivis: I added a few lines to debug
```
	args := []string{"authorizer_uri", "name", "rest_api_id", "identity_source", "type", "identity_validation_expression", "authorizer_credentials"}
	for _, arg := range args {
		val, ok := diff.GetOk(arg)
		log.Printf("[DEBUG] %s: #%s#, #%v#", arg, val.(string), ok)
	}
```
It gives me logs:
```
2018/03/05 21:50:54 [DEBUG] authorizer_uri: ##, #false#
2018/03/05 21:50:54 [DEBUG] name: #tf-acctest-igw-authorizer-yddwa40#, #true#
2018/03/05 21:50:54 [DEBUG] rest_api_id: ##, #false#
2018/03/05 21:50:54 [DEBUG] identity_source: #method.request.header.Authorization#, #true#
2018/03/05 21:50:54 [DEBUG] type: #TOKEN#, #true#
2018/03/05 21:50:54 [DEBUG] identity_validation_expression: ##, #false#
2018/03/05 21:50:54 [DEBUG] authorizer_credentials: ##, #false#
```
`name`, `identity_source`, `type` are arguments with default value. It seems like `diff.GetOk()` doesn't return value for arguments without default value. This is what makes me confused.

aws/resource_aws_api_gateway_authorizer.go
```diff
@@ -181,6 +199,12 @@ func resourceAwsApiGatewayAuthorizerUpdate(d *schema.ResourceData, meta interfac
 			Value: aws.String(d.Get("identity_validation_expression").(string)),
 		})
 	}
+	if d.HasChange("provider_arns") {
```
bflad: This can be simplified using `schema.Set` `Difference()`, here's an example in action:

https://github.com/terraform-providers/terraform-provider-aws/blob/90e2e8df03e4ded02c825169bc64b3dae747b9a0/aws/resource_aws_api_gateway_integration.go#L338-L361

aws/resource_aws_api_gateway_authorizer.go
bflad: Ah, this is happening because the values are computed:

https://github.com/hashicorp/terraform/blob/ea50f455edca3acaa0a6ab372b861e6cd2d386aa/helper/schema/resource_diff.go#L339

Unfortunately, I'm not sure of a workaround at the moment. For now, we can keep that logic outside `CustomizeDiff`. 🙁

loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayAuthorizer_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayAuthorizer_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayAuthorizer_basic
--- PASS: TestAccAWSAPIGatewayAuthorizer_basic (54.47s)
=== RUN   TestAccAWSAPIGatewayAuthorizer_cognito
--- PASS: TestAccAWSAPIGatewayAuthorizer_cognito (33.99s)
=== RUN   TestAccAWSAPIGatewayAuthorizer_switchAuthType
--- PASS: TestAccAWSAPIGatewayAuthorizer_switchAuthType (89.65s)
=== RUN   TestAccAWSAPIGatewayAuthorizer_authTypeValidation
--- PASS: TestAccAWSAPIGatewayAuthorizer_authTypeValidation (44.02s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	222.195s
```

aws/resource_aws_api_gateway_authorizer.go
loivis: Nice:+1:

bflad:

PR #3157, opened 2018-01-27
data-source/aws_region: Remove EC2 API call, default to current if no endpoint or name specified

bflad:
This PR accomplishes a few things here:
* Removes `ec2.DescribeRegions` API call. The SDK provides all regions and EC2 endpoint information already and this resource will automatically update when we update the SDK. The provider will not work in a new region without an updated release anyways.
* Fixes `current` to match the documentation that listed it as optional, so `current = true` is no longer necessary if `endpoint` and `name` are omitted.

```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsRegion'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsRegion -timeout 120m
=== RUN   TestAccDataSourceAwsRegion_basic
--- PASS: TestAccDataSourceAwsRegion_basic (6.91s)
=== RUN   TestAccDataSourceAwsRegion_endpoint
--- PASS: TestAccDataSourceAwsRegion_endpoint (8.90s)
=== RUN   TestAccDataSourceAwsRegion_name
--- PASS: TestAccDataSourceAwsRegion_name (10.58s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	26.443s
```

### DIFF
--- a/aws/data_source_aws_region.go
+++ b/aws/data_source_aws_region.go
@@ -2,10 +2,9 @@ package aws
 
 import (
 	"fmt"
-	"log"
+	"strings"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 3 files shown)


website/docs/d/region.html.markdown
```diff
@@ -35,12 +33,9 @@ exported as attributes.
 * `name` - (Optional) The full name of the region to select.
 
 * `current` - (Optional) Set to `true` to match only the region configured
-  in the provider. (It is not meaningful to set this to `false`.)
+  in the provider. Defaults to `true` if `endpoint` or `name` is not given.
```
radeksimko: I'm curious what is the value of `current = true` still. 🤔  Can't we just assume it's always true in such conditions and deprecate the field?

website/docs/d/region.html.markdown
```diff
@@ -51,4 +46,4 @@ The following attributes are exported:
 * `current` - `true` if the selected region is the one configured on the
   provider, or `false` otherwise.
 
-* `endpoint` - The endpoint for the selected region.
+* `endpoint` - The EC2 endpoint for the selected region.
```
radeksimko: Thank you! 👍

aws/data_source_aws_region.go
```diff
@@ -36,48 +35,74 @@ func dataSourceAwsRegion() *schema.Resource {
...
+	current := d.Get("current").(bool)
+	for _, partition := range endpoints.DefaultPartitions() {
+		for _, region := range partition.Regions() {
+			if region.ID() == currentRegion {
+				if matchingRegion == nil {
+					matchingRegion = &region
+					break
+				}
+				if current && (*matchingRegion).ID() != currentRegion {
+					return fmt.Errorf("multiple regions matched; use additional constraints to reduce matches to a single region")
```
radeksimko: Is there any reason we can't just error out here regardless of `current`?

aws/data_source_aws_region.go
```diff
@@ -36,48 +35,74 @@ func dataSourceAwsRegion() *schema.Resource {
...
+				}
+				if strings.TrimPrefix(regionEndpointEc2.URL, "https://") == endpoint {
+					matchingRegion = &region
+					break
+				}
+			}
+		}
+		if matchingRegion == nil {
+			return fmt.Errorf("region not found for endpoint: %s", endpoint)
+		}
```
radeksimko: It took me a while to understand what this block of code does. 😅  Would you mind extracting it into something like `matchingRegion, err := findRegionByEndpoint(endpoint)`?

aws/data_source_aws_region.go
```diff
@@ -36,48 +35,74 @@ func dataSourceAwsRegion() *schema.Resource {
...
+						return fmt.Errorf("multiple regions matched; use additional constraints to reduce matches to a single region")
+					}
+					matchingRegion = &region
+					break
+				}
+			}
+		}
+		if matchingRegion == nil {
+			return fmt.Errorf("region not found for name: %s", name)
+		}
```
radeksimko: Likewise - how do you feel about extracting this into `matchingRegion, err := findRegionByName(name)`?

aws/data_source_aws_region.go
```diff
@@ -36,48 +35,74 @@ func dataSourceAwsRegion() *schema.Resource {
...
 	}
-	if resp == nil || len(resp.Regions) == 0 {
-		return fmt.Errorf("no matching regions found")
-	}
-	if len(resp.Regions) > 1 {
-		return fmt.Errorf("multiple regions matched; use additional constraints to reduce matches to a single region")
-	}
-
-	region := resp.Regions[0]
+	d.Set("endpoint", strings.TrimPrefix(regionEndpointEc2.URL, "https://"))
```
radeksimko: Just out of curiosity: What is the reason for the trimming here?

aws/data_source_aws_region.go
bflad: Ha, I think I was trying to make it so it could support current being false or something. I'm on board with deprecating the `current` attribute completely (making it have no effect in the interim period) so this extra logic is moot and I'll remove it now.

aws/data_source_aws_region.go
bflad: That's a great idea! Let me implement that.

aws/data_source_aws_region.go
bflad: 👍 👍

aws/data_source_aws_region.go
bflad: Previously, the EC2 API response was returning just the hostname (see example here: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRegions.html), however the SDK `ResolvedEndpoint.URL` prepends `https://` to the front. We weren't previously testing the old behavior, but the acceptance testing shows it doing the required trimming (I think).

bflad:
@radeksimko refactored those two functions out, much cleaner. 😄 

```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsRegion'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsRegion -timeout 120m
=== RUN   TestAccDataSourceAwsRegion_basic
--- PASS: TestAccDataSourceAwsRegion_basic (6.37s)
=== RUN   TestAccDataSourceAwsRegion_endpoint
--- PASS: TestAccDataSourceAwsRegion_endpoint (5.74s)
=== RUN   TestAccDataSourceAwsRegion_endpointAndName
--- PASS: TestAccDataSourceAwsRegion_endpointAndName (9.67s)
=== RUN   TestAccDataSourceAwsRegion_name
--- PASS: TestAccDataSourceAwsRegion_name (8.01s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	29.817s

make test TEST=./aws
go test ./aws -timeout=30s -parallel=4
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1.695s
```

bflad:

PR #3159, opened 2018-01-27
provider: Automatically validate new regions from AWS SDK

bflad:
Reference: https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/

This version keeps the stricter `==` matching for `ValidateRegion()`, but we could loosen it to the below, which would also turn on regexp matching.

```
if _, isValidRegion := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), c.Region); isValidRegion {
  return nil
}
```

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -255,11 +255,10 @@ into Terraform.
 
 #### New Region
 
-Implementing a new region gives Terraform the ability to connect and interact
-with endpoints in a new geographic area. While baseline region support is fairly
-trivial to add, new regions are generally limited in which services they support.
+While region validation is automatically added with SDK updates, new regions
+are generally limited in which services they support. Below are some
+manually sourced values from documentation.
 
- - [ ] Add region to `aws/config.go`
  - [ ] Check [Regions and Endpoints ELB regions](https://docs.aws.amazon.com/general/latest/gr/rande.html#elb_region) and add Route53 Hosted Zone ID if available to `aws/data_source_aws_elb_hosted_zone_id.go`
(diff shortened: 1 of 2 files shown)


bflad:
It turns out that code is actually "okay" (at least from crash standpoint): https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/hosted_zones_test.go#L15

The type is `map[string]string` and string literals have a zero value of `""`. We just see crashes other places due to usage of `map[string]interface{}` which has `nil` zero value

radeksimko:
Right, so it won't crash 👌  

You just made me think that maybe for some lists, like this one in particular we should actually silently fail to allow folks use the S3 bucket resource, but I still think such failure should be logged as `WARN` or something.

https://github.com/terraform-providers/terraform-provider-aws/blob/e221786c28ba9dfb0b7794df4031a1feaa03309b/aws/resource_aws_s3_bucket.go#L1035-L1039

https://github.com/terraform-providers/terraform-provider-aws/blob/5f3605b9807c6ba4e57a4cdd6faa9bd787a6dac4/aws/data_source_aws_s3_bucket.go#L94-L97

bflad:
I think there are some nuances to consider. I think cn-northwest-1 for example supports S3 buckets but not S3 website hosting. As long as we "best effort" support some functionality that AWS is inconsistently offering across regions then I'll apply any changes you'd like

bflad:
Changed the S3 website hosted zone ID function signature to `HostedZoneIDForRegion(region string) (string, error)` and it'll now log WARN messages when missing instead of setting it to `""`. Please let me know if you'd like anything else adjusted. 😄

bflad:

PR #3164, opened 2018-01-28
resource/aws_mq_broker: validate user password

loivis:
Annoyed by password length when add data source for `aws_mq_broker`. Add validation function to check password length falls into [12, 250].

### DIFF
--- a/aws/resource_aws_mq_broker.go
+++ b/aws/resource_aws_mq_broker.go
@@ -140,9 +140,10 @@ func resourceAwsMqBroker() *schema.Resource {
 							Optional: true,
 						},
 						"password": {
-							Type:      schema.TypeString,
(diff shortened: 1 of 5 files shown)


bflad:
The API documentation also notes some other restrictions, could you add those as well please while you're in here? 😄  https://docs.aws.amazon.com/amazon-mq/latest/api-reference/rest-api-user.html#rest-api-user-attributes

> The password of the user. This value must be at least 12 characters long, must contain at least 4 unique characters, and must not contain commas.

loivis:
Thanks for you input, @bflad. I didn't even check aws doc, just added the validation according to the return error. Now I've included those requirements mentioned from the link.

bflad:

PR #3168, opened 2018-01-28
Add lambda example

oscr:
I think that this example will be useful for a lot of people. At least when I was starting out with Terraform it took me quite a while to achieve this functionality.

If there is anything I can do to improve it please don't hesitate to ask.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.zip
 *.dll
 *.exe
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-*.zip
 *.dll
 *.exe
 .DS_Store
--- a/examples/lambda/README.md
+++ b/examples/lambda/README.md
@@ -1,6 +1,6 @@
 # Lambda Example
 
-This examples shows how to deploy an AWS Lambda function. 
+This examples shows how to deploy an AWS Lambda function using Terraform only.
 
 To run, configure your AWS provider as described in https://www.terraform.io/docs/providers/aws/index.html
 
--- a/examples/lambda/README.md
+++ b/examples/lambda/README.md
@@ -1,0 +1,9 @@
+# Lambda Example
+
+This examples shows how to deploy an AWS Lambda function. 
+
+To run, configure your AWS provider as described in https://www.terraform.io/docs/providers/aws/index.html
+
+Running the example
+
+run `terraform apply` to see it work.
--- a/examples/lambda/hello_lambda.py
+++ b/examples/lambda/hello_lambda.py
@@ -1,0 +1,6 @@
+import os
+
+def lambda_handler(event, context):
+    # This will show up in CloudWatch
+    print("Value of 'foo': " + os.environ['foo'])
+    return 'Hello from Lambda!'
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+*.zip
```
Ninir: We actually have a few examples where we need those zips, as in: https://github.com/terraform-providers/terraform-provider-aws/tree/master/examples/cognito-user-pool so we should remov this one

examples/lambda/README.md
```diff
@@ -0,0 +1,9 @@
+# Lambda Example
+
+This examples shows how to deploy an AWS Lambda function.
```
Ninir: I would like to enhance this message with: `This examples shows how to deploy an AWS Lambda function using Terraform only.`

Just to expose my point of view: another way to deploy a given lambda is by building the ZIP in your CI, put it on S3, and use a s3 object data source to get the file.
With this logic, you can version your lambdas and benefit from a real build - deploy phase.

Thoughts? :)

examples/lambda/hello_lambda.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+def lambda_handler(event, context):
+    # This will show up in CloudWatch
```
Ninir: I think we should either update this message to `This will log to a CloudWatch Logs group if allowed by the Lambda role` or add the correct permission to the IAM role, as we are not defining the CloudWatch:Logs action to actually log.

In both cases, we should rewrite it to `This will log to a CloudWatch Logs group` to better explain what's going on.

What do you think?

oscr: I will remove logging because it's not that clear cut due to the AWS console. If you run the lambda you'll see the following under **__Log output__**:

    START RequestId: dd6b2bf4-05e8-11e8-8679-a50298ecb88b Version: $LATEST
    Value of 'foo': bar
    END RequestId: dd6b2bf4-05e8-11e8-8679-a50298ecb88b
    REPORT RequestId: dd6b2bf4-05e8-11e8-8679-a50298ecb88b	Duration: 1.13 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 21 MB	

If we were to print `This will log to a CloudWatch Logs group if allowed by the Lambda role` and it appears in the console then it wouldn't be unreasonable to believe that the correct permissions are set. So I'll simplify the example.

.gitignore
oscr: I will fix that mistake.

examples/lambda/README.md
oscr: I think that's an improvement. I'll add it.

oscr:
Hi @Ninir 

Thank you  for your feedback. I have updated the example based on your comments.

oscr:
@Ninir Are there any other changes that you would like to see?

examples/lambda/hello_lambda.py
Ninir: @oscr My question was meant for the comment line only, not the log line that would appear in CloudWatch Logs 😄, sorry for the confusion!

Ninir:
@oscr will check it this day!

PR #3175, opened 2018-01-29
Add rules packages datasource for AWS Inspector

MattiasGees:
This add the rules packages datasource to retrieve all the available AWS inspector rules packages for your region.

This is my first time creating a full data source for Terraform. All feedback is welcome on both code and test coverage. 

I am also not sure about the name of the datasource. At this moment it is `rules_packages` but maybe `inspector_rules_packages` is a better name?


bflad:
Since AWS is very segregated between services and the APIs, we would definitely recommend calling the data source something with the service name in it and would require that change on review so its not confused with many of the other services that offer "rules" like EC2 security groups. So in this case `dataSourceAwsInspectorRulePackages` and `aws_inspector_rule_packages` would be appropriate.

I'll provide a full review shortly.

aws/data_source_aws_rules_packages_test.go
```diff
@@ -0,0 +1,81 @@
...
+			return fmt.Errorf("Can't find Tules Packages resource: %s", n)
+		}
+
+		fmt.Printf("%s", rs)
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("Rules Packages resource ID not set.")
+		}
+
+		actual, err := testAccCheckAwsRulesPackagesBuildAvailable(rs.Primary.Attributes)
```
bflad: We don't need this function or its call, see above.

aws/data_source_aws_rules_packages_test.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			return err
+		}
+
+		expected := actual
```
bflad: I don't think we don't need to acceptance test `sort.Strings()`.

MattiasGees:
@bflad I made the requested changes.

aws/data_source_aws_inspector_rules_packages_test.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func TestAccAWSInspectorRulesPackages_basic(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckAWSInspectorRulesPackagesConfig,
+				Check:  resource.TestCheckResourceAttrSet("data.aws_inspector_rules_packages.test", "arns.#"),
```
bflad: The `.test` here needs to either be updated to `.rules_packages` to match the name in `testAccCheckAWSInspectorRulesPackagesConfig` or the data source to be named `test` in the configuration

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSInspectorRulesPackages_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSInspectorRulesPackages_basic -timeout 120m
=== RUN   TestAccAWSInspectorRulesPackages_basic
--- FAIL: TestAccAWSInspectorRulesPackages_basic (7.84s)
	testing.go:513: Step 0 error: Check failed: Not found: data.aws_inspector_rules_packages.test in [root]
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	7.888s
make: *** [testacc] Error 1
```

aws/provider.go
```diff
@@ -214,6 +214,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_region":                           dataSourceAwsRegion(),
 			"aws_route_table":                      dataSourceAwsRouteTable(),
 			"aws_route53_zone":                     dataSourceAwsRoute53Zone(),
+			"aws_inspector_rules_packages":         dataSourceAwsInspectorRulesPackages(),
```
bflad: Minor nitpick: Can you lexicographically sort this please?

aws/data_source_aws_inspector_rules_packages_test.go
MattiasGees: I renamed it to test

bflad:

PR #3187, opened 2018-01-30
Don't mangle DNS aliases: Trim{Prefix,Suffix} not Trim{Left,Right}.

handlerbot:
## Description

A bug in Route 53 alias processing (where the leading string "dualstack." is removed, if present) will strip off all leading characters from a DNS alias name that are in the set of "dualstack." (or "acdklstu.", when sorted and de-duplicated), until it reaches a character that is not in that set. So, "dualstack.listenapi" becomes "istenapi" (instead of "listenapi") because "l" is in the set to trim, but "i" is not.

[source](https://github.com/terraform-providers/terraform-provider-aws/issues/3185#issuecomment-361410720)

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/3185

### DIFF
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -900,12 +900,9 @@ func nilString(s string) *string {
 }
 
 func normalizeAwsAliasName(alias interface{}) string {
-	input := alias.(string)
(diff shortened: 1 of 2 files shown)


bflad:

PR #3191, opened 2018-01-30
resource/aws_kms_alias: Exposed target_key_id in the documentation

Ninir:
This resolves comment https://github.com/terraform-providers/terraform-provider-aws/issues/3019#issuecomment-359657403

Even though the kms_alias resource code actually write the [`target_key_id` to the state](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_kms_alias.go#L119), the documentation isn't reflecting that.
The related data source is indeed OK for this particular point ([source](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/data_source_aws_kms_alias.go#L80)).

### DIFF
--- a/website/docs/r/kms_alias.html.markdown
+++ b/website/docs/r/kms_alias.html.markdown
@@ -35,7 +35,7 @@ The name must start with the word "alias" followed by a forward slash (alias/). 
 
 ## Attributes Reference
 
-The following additional attributes are exported:
+In addition to the arguments, the following attributes are exported:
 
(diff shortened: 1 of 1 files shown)


bflad:
Hmmm, this is a required attribute noted in the argument reference and we have been adding "additional" to the attributes reference section so it reads `The following additional attributes are exported`. Do you think we should update the argument references everywhere to instead say `The following arguments are supported and exported if supplied:`?

Ninir:
@bflad Updated the sentence to make it clearer.

PR #3197, opened 2018-01-30
resource/aws_dynamodb_table: Allow disabling stream w/ empty view type

### LINKED ISSUE #1134: aws_dynamo_db_table throws validation error when disabling stream where stream_view_type is also specified
### Terraform Version
0.9.8

### Affected Resource(s)
- aws_dynamodb_table

```hcl
resource "aws_dynamodb_table" "table" {
  name = "table"
  read_capacity = 10
  write_capacity = 10
  hash_key = "_id"
  attribute {
    name = "_id"
    type = "N"
  }

  stream_enabled = "${var.stream_enabled}"
  stream_view_type = "${var.stream_view_type}"
}

variable "stream_enabled" {
  default = true
}
variable "stream_view_type" {
  default = "NEW_IMAGE"
}
```

### Error Output
```
* aws_dynamodb_table.table: 1 error(s) occurred:

* aws_dynamodb_table.table: ValidationException: One or more parameter values were invalid: If stream is being disabled, then UpdateViewType must not be specified
```

### Expected Behavior

Assuming previously created dynamodb table with stream enabled
Plan Output:
```
~ aws_dynamodb_table.table
    stream_enabled: "true" => "false"
```
when running apply where `var.stream_enabled` set to `false`, resource should update stream to disabled in AWS

### Actual Behavior
Validation error is thrown due to the presence of `var.stream_view_type` even though the value is the same 

### Steps to Reproduce
1. `terraform apply` to create the dynamodb table and stream in AWS, where variables are defaults 
2. run `terraform apply` where `var.stream_enabled` overridden to `false`

radeksimko:
Fixes #1134

Despite `stream_view_type` being optional the validation is still being triggered which makes it impossible to parametrise the stream spec as described in #1134

## Test results

```
=== RUN   TestAccAWSDynamoDbTable_streamSpecification
--- PASS: TestAccAWSDynamoDbTable_streamSpecification (92.66s)
```

### DIFF
--- a/aws/resource_aws_dynamodb_table_test.go
+++ b/aws/resource_aws_dynamodb_table_test.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"regexp"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


aws/resource_aws_dynamodb_table_test.go
```diff
@@ -950,53 +957,23 @@ resource "aws_dynamodb_table" "basic-dynamodb-table" {
...
 
-func testAccAWSDynamoDbConfigStreamSpecification() string {
+func testAccAWSDynamoDbConfigStreamSpecification(enabled bool, viewType string) string {
 	return fmt.Sprintf(`
 resource "aws_dynamodb_table" "basic-dynamodb-table" {
   name = "TerraformTestStreamTable-%d"
   read_capacity = 10
   write_capacity = 20
   hash_key = "TestTableHashKey"
-  range_key = "TestTableRangeKey"
```
bflad: Thank you for making the configuration more relevant to what we're trying to test here 👍

bflad:
I understand where this is change coming from and the Terraform schema helpers don't necessarily make this easy, but it now allows this incorrect configuration to passthrough to the apply stage and fail:

```
stream_enabled   = true
stream_view_type = ""
```

e.g. `testAccAWSDynamoDbConfigStreamSpecification(true, ""),`

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSDynamoDbTable_streamSpecification'
TF_ACC=1 go test ./aws -v -run=TestAccAWSDynamoDbTable_streamSpecification -timeout 120m
=== RUN   TestAccAWSDynamoDbTable_streamSpecification
--- FAIL: TestAccAWSDynamoDbTable_streamSpecification (4.47s)

		* aws_dynamodb_table.basic-dynamodb-table: 1 error(s) occurred:

		* aws_dynamodb_table.basic-dynamodb-table: ValidationException: 1 validation error detected: Value '' at 'streamSpecification.streamViewType' failed to satisfy constraint: Member must satisfy enum value set: [OLD_IMAGE, KEYS_ONLY, NEW_AND_OLD_IMAGES, NEW_IMAGE]
			status code: 400, request id: BU4C7KSK4LOBT764E77I8FEOCRVV4KQNSO5AEMVJF66Q9ASUAAJG
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	4.517s
make: *** [testacc] Error 1
```

radeksimko:
@bflad Added validation as discussed 😉 

I also removed `Computed` from `stream_enabled` as it doesn't seem to be necessary there at all. I was tempted to do the same for `stream_view_type` and set `Default` instead, but we don't have any helper schemas to help us differentiate between defaults and undefined yet, so I left it as is.

All DynamoDB acceptance tests are still passing.

bflad:

PR #3211, opened 2018-01-31
data-source/aws_ssm_parameter: Fix wrong arn attribute for full path parameter names

seppi91:
This should fix the issue described in https://github.com/terraform-providers/terraform-provider-aws/issues/2593

As I'm not experienced with programming in Go, I couldn't implement a respective testcase for this.


bflad:

PR #3214, opened 2018-01-31
resource/aws_ses_template: Send only specified attributes for update

szilveszter:
Addresses an issue where all text-only emails suddenly turned into emails after an update. Currently the `UpdateTemplate` call will set the `HtmlPart` to an empty string if it's not specified for the resource, which results in an email with an empty body in most mail clients.

### DIFF
--- a/aws/resource_aws_ses_template.go
+++ b/aws/resource_aws_ses_template.go
@@ -110,10 +110,19 @@ func resourceAwsSesTemplateUpdate(d *schema.ResourceData, meta interface{}) erro
 	templateName := d.Id()
 
 	template := ses.Template{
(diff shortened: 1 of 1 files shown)


bflad:
This looks like the right change given what you've described above. Can you verify that if you have both text and html configured, then want to remove one, that simply omitting it from the UpdateTemplate request does the right thing to remove it (e.g. doesn't leave the now omitted one configured on the AWS side)?

szilveszter:
I can confirm, that is the case. I tested a build of this branch locally with my Terraform setup, and after the new update went through, the empty HTML part was removed from the template, and my emails appeared properly as plain text again after sending them out.

bflad:

Issue #1574, opened 2017-09-02
Feature request: return last rule priority from aws_alb_listener data source


bofoy:
Hi there,

Terraform Version

0.10.3

Affected Resource(s)

`aws_alb_listener` data source

Expected Behavior

I would like to be able to get the latest rule priority from the `aws_alb_listener` data source so the next `ecs_service` to deploy can simply use the latest rule priority and increment it by 1

My specific use case is that we have a central ECS cluster and are deploying ECS services independently. each service has its own state file. Having to check/keep track of the latest rule priority is a hassle and something we'd like to see in the data source

labkey-ians:
+1 
Or better yet, make priority on `aws_alb_listener_rule` optional, with insert to next priority as the not present action.

bflad:
This has been released in [version 1.10.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1100-february-24-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #3219, opened 2018-01-31
resource/lb_listener_rule: make priority optional

loivis:
Implement #1574 .

We've been using a script to calculate available priority when registering a target group to listener. Initially it was just a workaround. But as we all know, workaround usually lies there forever as solution. The idea in #1574 to make `priority` optional sounds great to me and drives to the implementation.


loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccAWSLBListenerRule_autoPriority'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLBListenerRule_autoPriority -timeout 120m
=== RUN   TestAccAWSLBListenerRule_autoPriority
--- PASS: TestAccAWSLBListenerRule_autoPriority (466.43s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	466.477s
```

loivis:
Making it optional doesn’t work well if registering multiple target groups to the same listener at the same time. In this case, it’s hard to guarantee unique value for each target group. Maybe setting a dependency chain across target groups, like a->b->c->d?

tomelliff:
Does it not make more sense to expose the highest listener priority on the ALB data source instead? That way you could easily get the highest and then increment from there so if you have multiple target groups to register you just keep increasing the increment on each target group?

loivis:
@tomelliff Indeed and that's exactly the `workaround` that we're using currently, except that we don't fall into the scenario to register multiple target groups at the same time. I just imaged the scenario someone might use.

I created this PR simply based on the votes from #1574, `7` for optional priority and `3` for expose priority

Personally, making `priority` optional fits our needs perfectly because in our pipeline we register only one target group for each service.

Personally, I would also like to implement the other approach to expose next available priority from data source `lb_listener`, as you mentioned, as what our workaround is doing now. Actually I was half on the way but chose to finish this PR first. With both implemented, users will have more freedom to choose either that fits their needs. @bflad What do you think from higher level? I'm not sure if we prefer more flexibility with both or just choose one way to go?

tomelliff:
This works for me so I'm happy with this approach but I wonder if it's not confusing for those who do need to register multiple listeners at the same time. Although I guess you'll get an error if you do attempt to register multiple listeners this way and then hopefully people will realise.

loivis:
Maybe a few more words to document will help those who read document:smile:

bflad:
Seems like if we wanted to handle parallelism here, we could make priority optional by:
* If priority configured: setting the priority and failing normally if unable to create rule
* If priority is not configured: grabbing highest priority as a starting point and continually retrying priority + 1 on API "duplicate rule priority" error up to the max priority if there is one

loivis:
Thanks for pointing out the direction, @bflad . I'll take a look existing code with retryable error. I saw such codes a few times but never thought to use it before.

loivis:
Added retry logic as suggusted, together with acceptance test to register 100 rules to the same listener.

aws/resource_aws_lb_listener_rule_test.go
```diff
@@ -829,3 +879,190 @@ resource "aws_security_group" "alb_test" {
...
+    target_group_arn = "${aws_lb_target_group.test.arn}"
+  }
+
+  condition {
+    field = "path-pattern"
+    values = ["/first/*"]
+  }
+}
+
+resource "aws_lb_listener_rule" "third" {
```
bflad: We test a static rule priority with the `_autoPriorityStatic` configuration below. We should either just use this rule here and remove the static configuration below or remove this rule here.

aws/resource_aws_lb_listener_rule_test.go
```diff
@@ -829,3 +879,190 @@ resource "aws_security_group" "alb_test" {
...
+    values = ["/last/*"]
+  }
+}
+`)
+}
+
+func testAccAWSLBListenerRuleConfig_autoPriority100Rules(lbName, targetGroupName string) string {
+	return testAccAWSLBListenerRuleConfig_autoPriorityStatic(lbName, targetGroupName) + fmt.Sprintf(`
+resource "aws_lb_listener_rule" "limit" {
+  count = 97
```
bflad: Can we make this match the test name or vice versa please?

aws/resource_aws_lb_listener_rule.go
```diff
@@ -112,9 +116,35 @@ func resourceAwsLbListenerRuleCreate(d *schema.ResourceData, meta interface{}) e
...
+		params.Priority = aws.Int64(int64(v.(int)))
+		resp, err = elbconn.CreateRule(params)
+		if err != nil {
+			return fmt.Errorf("Error creating LB Listener Rule: %v", err)
+		}
+	} else {
+		err := resource.Retry(5*time.Minute, func() *resource.RetryError {
+			var err error
+			lastPriority, _, err := getListenerRulePriority(elbconn, listenerArn)
+			log.Println(lastPriority)
```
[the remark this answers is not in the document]
loivis: forgot to remove:sweat_smile:

aws/resource_aws_lb_listener_rule.go
```diff
@@ -112,9 +116,35 @@ func resourceAwsLbListenerRuleCreate(d *schema.ResourceData, meta interface{}) e
...
+			if err != nil {
+				return resource.NonRetryableError(err)
+			}
+			params.Priority = aws.Int64(lastPriority + 1)
+			resp, err = elbconn.CreateRule(params)
+			if err != nil {
+				if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == "PriorityInUse" {
```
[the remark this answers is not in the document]
loivis: nice...good to know

aws/resource_aws_lb_listener_rule.go
```diff
@@ -312,3 +342,45 @@ func isRuleNotFound(err error) bool {
 	elberr, ok := err.(awserr.Error)
 	return ok && elberr.Code() == "RuleNotFound"
 }
+
+func getListenerRulePriority(conn *elbv2.ELBV2, arn string) (last, next int64, err error) {
```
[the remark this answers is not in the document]
loivis: Yes, of course. It was the initial implementation but later synced with #3374

aws/resource_aws_lb_listener_rule_test.go
```diff
@@ -191,6 +191,56 @@ func TestAccAWSLBListenerRule_multipleConditionThrowsError(t *testing.T) {
 	})
 }
 
+func TestAccAWSLBListenerRule_autoPriority(t *testing.T) {
```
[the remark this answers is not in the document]
loivis: I didn't mean to test static priority here since it's already tested. I was testing to switch between explicit priority and auto priority without any problem. I agree to simplify the name to `_priority` then include more error checks here.

aws/resource_aws_lb_listener_rule_test.go
loivis: It was the leftover when I tried to use the gap priority in the beginning. I think it's still useful here as test case of existing priorities with gap in between. And as explained above, static priority here is not intended to test static priority but to test switching `priority` argument.

aws/resource_aws_lb_listener_rule_test.go
loivis: Sorry I'm not sure if I get your point. Did you mean either `count = 100` or `testAccAWSLBListenerRuleConfig_autoPriority97Rules`? If that's the case, maybe change to `testAccAWSLBListenerRuleConfig_autoPriorityRuleNumberLimit`? The intension was to fill the listener with 100 rules limit per load balancer and there was already 3 existing rules.

bflad: Okay I wasn't immediately aware of the rules per listener limit.

If we're just trying to test concurrency of auto-assigning priorities (I think more important here), let's just do 10, which should always pass.

If we're trying to test hitting the max rules per listener limit (I do not think its important to test one below the limit), that should be a separate test of 101 rules and `ExpectError`. AWS could change that limit and it makes it hard to understand why we're testing 3 + 97 rules here without making the test name more explicit like you suggested with `testAccAWSLBListenerRuleConfig_autoPriorityRuleNumberLimit`.

aws/resource_aws_lb_listener_rule_test.go
bflad: Valid point 👍

aws/resource_aws_lb_listener_rule_test.go
loivis: I'll do both then, testing for 10 and 101.

loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccAWSLBListenerRule_priority'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLBListenerRule_priority -timeout 120m
=== RUN   TestAccAWSLBListenerRule_priority
--- PASS: TestAccAWSLBListenerRule_priority (482.18s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	482.236s
```

loivis:
@bflad anything else to change before approve and merge?

bflad:

PR #3243, opened 2018-02-03
New Data Source: aws_workspaces_bundle

atsushi-ishibashi:
This may be useful for #1917
```
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsWorkspaceBundle_basic -timeout 120m
=== RUN   TestAccDataSourceAwsWorkspaceBundle_basic
--- PASS: TestAccDataSourceAwsWorkspaceBundle_basic (82.43s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	82.469s
```

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -80,6 +80,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/aws/aws-sdk-go/service/waf"
 	"github.com/aws/aws-sdk-go/service/wafregional"
+	"github.com/aws/aws-sdk-go/service/workspaces"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/go-cleanhttp"
@@ -206,6 +207,7 @@ type AWSClient struct {
 	athenaconn            *athena.Athena
 	dxconn                *directconnect.DirectConnect
 	mediastoreconn        *mediastore.MediaStore
+	wsconn                *workspaces.WorkSpaces
 }
 
 func (c *AWSClient) S3() *s3.S3 {
@@ -451,6 +453,7 @@ func (c *Config) Client() (interface{}, error) {
 	client.athenaconn = athena.New(sess)
 	client.dxconn = directconnect.New(sess)
 	client.mediastoreconn = mediastore.New(sess)
+	client.wsconn = workspaces.New(sess)
 
 	// Workaround for https://github.com/aws/aws-sdk-go/issues/1376
 	client.kinesisconn.Handlers.Retry.PushBack(func(r *request.Request) {
--- a/aws/data_source_aws_workspaces_bundle.go
+++ b/aws/data_source_aws_workspaces_bundle.go
@@ -1,0 +1,118 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/workspaces"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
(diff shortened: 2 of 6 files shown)


aws/data_source_aws_workspaces_bundle.go
```diff
@@ -0,0 +1,118 @@
...
+	d.SetId(bundleID)
+	d.Set("description", bundle.Description)
+	d.Set("name", bundle.Name)
+	d.Set("owner", bundle.Owner)
+	if bundle.ComputeType != nil {
+		r := map[string]interface{}{
+			"name": *bundle.ComputeType.Name,
+		}
+		ct := []map[string]interface{}{r}
+		d.Set("compute_type", ct)
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("compute_type", ct); err != nil {
  return fmt.Errorf("error setting compute_type: %s", err)
}
```

This also needs to be repeated for `root_storage` and `user_storage` below.

aws/data_source_aws_workspaces_bundle.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	bundle := resp.Bundles[0]
+	d.SetId(bundleID)
+	d.Set("description", bundle.Description)
+	d.Set("name", bundle.Name)
+	d.Set("owner", bundle.Owner)
+	if bundle.ComputeType != nil {
+		r := map[string]interface{}{
+			"name": *bundle.ComputeType.Name,
```
bflad: To prevent potential panics, we should prefer to use the SDK provided function for dereferencing values (e.g. `aws.StringValue(bundle.ComputeType.Name)`) instead of directly dereferencing via `*`

This needs to be repeated for `bundle.RootStorage.Capacity` and `bundle.UserStorage.Capacity` below as well.

website/docs/d/workspaces_bundle.html.markdown
```diff
@@ -0,0 +1,48 @@
+---
+layout: "aws"
+page_title: "AWS: aws_workspaces_bundle"
+sidebar_current: "docs-aws-datasource-workspaces-bundle"
+description: |-  
+  Get information on a Workspaces Bundle.
+---
+
+# aws_workspaces_bundle
```
bflad: Most likely a change that occurred after the creation of this pull request, but can you please prepend the title with `Data Source: `, e.g. `Data Source: aws_workspaces_bundle`

aws/data_source_aws_workspaces_bundle.go
```diff
@@ -0,0 +1,118 @@
...
+
+func dataSourceAwsWorkspaceBundle() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsWorkspaceBundleRead,
+
+		Schema: map[string]*schema.Schema{
+			"bundle_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
```
bflad: `ForceNew` is not required for data sources and should be removed from this attribute and the documentation.

aws/config.go
```diff
@@ -206,6 +207,7 @@ type AWSClient struct {
 	athenaconn            *athena.Athena
 	dxconn                *directconnect.DirectConnect
 	mediastoreconn        *mediastore.MediaStore
+	wsconn                *workspaces.WorkSpaces
```
bflad: Minor nitpick: to prevent any confusion with other services, we should probably name this `workspacesconn`

bflad:

PR #3258, opened 2018-02-05
New Data Source: aws_lb_ssl_policy

atsushi-ishibashi:
```
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsLbSSLPolicy_basic -timeout 120m
=== RUN   TestAccDataSourceAwsLbSSLPolicy_basic
--- PASS: TestAccDataSourceAwsLbSSLPolicy_basic (49.11s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	49.148s
```

The motivation is to follow the default policy which AWS defines.
For example,
```
data "aws_lb_ssl_policy" "default" {}

resource "aws_lb_listener" "example" {
  load_balancer_arn = "hogehoge"
  port              = "443"
  protocol          = "HTTPS"
  ssl_policy        = "${data.aws_lb_ssl_policy.default.name}"
  certificate_arn   = "fugafuga"
}
```

The first element of `SslPolicies` seems to be the default policy.

### DIFF
--- a/aws/data_source_aws_lb_ssl_policy.go
+++ b/aws/data_source_aws_lb_ssl_policy.go
@@ -1,0 +1,46 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsLbSSLPolicy() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsLbSSLPolicyRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsLbSSLPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	elbconn := meta.(*AWSClient).elbv2conn
+
+	input := &elbv2.DescribeSSLPoliciesInput{}
+
+	if v, ok := d.GetOk("name"); ok {
+		input.Names = []*string{aws.String(v.(string))}
+	}
+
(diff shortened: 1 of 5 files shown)


bflad:
@atsushi-ishibashi is this for attribute validation purposes or because its required for HTTPS listeners and there is no default value? If for validation, there are only a small number of the AWS predefined policies so seems like this would be better suited as a validation function instead of a full data source. If its for the lack of a default value, I am not sure we should rely on the first element of the `DescribeSslPolicies` API response to always be consistent or a necessarily "good" security posture.

Reference: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies
* ELBSecurityPolicy-2016-08
* ELBSecurityPolicy-TLS-1-2-2017-01
* ELBSecurityPolicy-TLS-1-1-2017-01
* ELBSecurityPolicy-2015-05
* ELBSecurityPolicy-TLS-1-0-2015-04

Curious to your thoughts and others' opinions.

atsushi-ishibashi:
According to docs, there is no default policy but recommended policy.
However I'm not sure that the first element is the recommened. I'll ask AWS stuff👍

atsushi-ishibashi:
I got the response form aws stuff.
> Now the first element matches the recommended policy but the order may change in the future without announce.

Uhh, so it's not good to rely on the first element😅

bflad:
😅  Its all good, that's why we review each other! I am glad they gave a clear answer even though it was not the one we wanted to hear. 🙉 

How would you like to proceed here? I do not think we should pick a default value as AWS' recommendations will likely change over time as well.

atsushi-ishibashi:
The recommendation will change over time so this data source becomes meaningful.
For example, `aws_ami`
```
data "aws_ami" "amazon_ecs" {
  most_recent = true
  owners      = ["amazon"]

  filter {
    name   = "name"
    values = ["amzn-ami-*-amazon-ecs-optimized"]
  }
}
```
But unfortunately there is no way to pick the recommendation surelt 😢 
I'll close.

PR #3267, opened 2018-02-06
resource/aws_rds_cluster: Adding hosted_zone_id as an exported attribute

ngenator:
The API returns a hosted zone id so you can create alias records in route53. This pr adds the hosted zone id to the aws_rds_cluster resource so that you can use it in route53 to create alias records.

### DIFF
--- a/aws/resource_aws_rds_cluster.go
+++ b/aws/resource_aws_rds_cluster.go
@@ -590,6 +590,7 @@ func flattenAwsRdsClusterResource(d *schema.ResourceData, meta interface{}, dbc 
 	d.Set("reader_endpoint", dbc.ReaderEndpoint)
 	d.Set("replication_source_identifier", dbc.ReplicationSourceIdentifier)
 	d.Set("iam_database_authentication_enabled", dbc.IAMDatabaseAuthenticationEnabled)
+	d.Set("hosted_zone_id", dbc.HostedZoneId)
 
 	var vpcg []string
 	for _, g := range dbc.VpcSecurityGroups {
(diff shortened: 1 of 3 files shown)


ngenator:
Should be noted that while you _can_ currently set route53 aliases for RDS db instances (despite the fact that the documentation claims you _can't_), you _can't actually_ do it for RDS clusters yet. The api does return a hosted zone id, so it is likely planned but not implemented on the route53 side.

I can create some route53 alias acceptance tests for rds db instances and rds clusters if needed

bflad:
@ngenator thanks for submitting this! You're definitely on the right track, there are two other parts we'll need to address before this can get in:

* Add this attribute to the resource `Schema`, which can be found inside `resourceAwsRDSCluster()`, e.g.:

```go
"hosted_zone_id": {
	Type:     schema.TypeString,
	Computed: true,
},
```

* Add a very simple acceptance test that this new attribute is actually set inside `TestAccAWSRDSCluster_basic` (I don't think it makes sense to try to determine its value as it'll be different per region), e.g.:

```go
resource.TestCheckResourceAttrSet("aws_rds_cluster.default", "hosted_zone_id"),
```

We can skip any acceptance testing beyond this for this PR as we've ensured that the `aws_rds_cluster` resource has correctly (at least to a certain degree) done its part. Since Route53 alias usage would be part of a separate resource, a separate PR (we don't need to delay this one) ensuring that RDS Route53 aliases work in the `aws_route53_record` acceptance testing would most certainly be welcome of course 😄 .

If you don't have time to fix the above, just let us know and we can finish this up for you. Thanks again! 🚀

ngenator:
Whoops, nice catch. For some reason I thought the schema was for the input attributes. I'll fix that and add the acceptance test for the attribute.

bflad:

PR #3285, opened 2018-02-07
Exposed the AMI field in instance.html.markdown

automaticgiant:
ami is, in fact, exposed

### DIFF
--- a/website/docs/d/instance.html.markdown
+++ b/website/docs/d/instance.html.markdown
@@ -54,7 +54,7 @@ are exported:
 ~> **NOTE:** Some values are not always set and may not be available for
 interpolation.
 
-* `ami` - The AMI in use on the Instance.
+* `ami` - The ID of the AMI used to launch the instance.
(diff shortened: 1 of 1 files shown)


website/docs/d/instance.html.markdown
```diff
@@ -54,6 +54,7 @@ are exported:
 ~> **NOTE:** Some values are not always set and may not be available for
 interpolation.
 
+* `ami` - The AMI in use on the Instance.
```
Ninir: Hey @automaticgiant 

Do you mind updating this to `The ID of the AMI used to launch the instance.`, which is the description extracted from the API? (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Instance.html under  `imageID`)

Thanks!

Ninir:
Hey @automaticgiant 

Allowed myself to make the change, keeping your original commit.

Thanks!

automaticgiant:
Oh thank you! My bad.

PR #3287, opened 2018-02-08
New Resource: aws_api_gateway_documentation_version

radeksimko:
## Test results

```
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_basic
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_basic (45.77s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_allFields
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_allFields (108.51s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_importBasic
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_importBasic (50.27s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	204.595s
```

### DIFF
--- a/aws/resource_aws_api_gateway_documentation_version.go
+++ b/aws/resource_aws_api_gateway_documentation_version.go
@@ -1,0 +1,137 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/apigateway"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsApiGatewayDocumentationVersion() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsApiGatewayDocumentationVersionCreate,
+		Read:   resourceAwsApiGatewayDocumentationVersionRead,
+		Update: resourceAwsApiGatewayDocumentationVersionUpdate,
+		Delete: resourceAwsApiGatewayDocumentationVersionDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"version": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"rest_api_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func resourceAwsApiGatewayDocumentationVersionCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	restApiId := d.Get("rest_api_id").(string)
+
+	params := &apigateway.CreateDocumentationVersionInput{
+		DocumentationVersion: aws.String(d.Get("version").(string)),
+		RestApiId:            aws.String(restApiId),
+	}
+	if v, ok := d.GetOk("description"); ok {
+		params.Description = aws.String(v.(string))
+	}
+
+	log.Printf("[DEBUG] Creating API Gateway Documentation Version: %s", params)
+
+	version, err := conn.CreateDocumentationVersion(params)
+	if err != nil {
+		return fmt.Errorf("Error creating API Gateway Documentation Version: %s", err)
+	}
+
+	d.SetId(restApiId + "/" + *version.Version)
+
+	return resourceAwsApiGatewayDocumentationVersionRead(d, meta)
+}
+
+func resourceAwsApiGatewayDocumentationVersionRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+	log.Printf("[DEBUG] Reading API Gateway Documentation Version %s", d.Id())
+
+	apiId, docVersion, err := decodeApiGatewayDocumentationVersionId(d.Id())
+	if err != nil {
+		return err
+	}
+
+	version, err := conn.GetDocumentationVersion(&apigateway.GetDocumentationVersionInput{
+		DocumentationVersion: aws.String(docVersion),
+		RestApiId:            aws.String(apiId),
+	})
+	if err != nil {
+		if isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
+			log.Printf("[WARN] API Gateway Documentation Version (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("rest_api_id", apiId)
+	d.Set("description", version.Description)
+	d.Set("version", version.Version)
+
+	return nil
+}
+
+func resourceAwsApiGatewayDocumentationVersionUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+	log.Printf("[DEBUG] Updating API Gateway Documentation Version %s", d.Id())
+
+	_, err := conn.UpdateDocumentationVersion(&apigateway.UpdateDocumentationVersionInput{
+		DocumentationVersion: aws.String(d.Get("version").(string)),
+		RestApiId:            aws.String(d.Get("rest_api_id").(string)),
+		PatchOperations: []*apigateway.PatchOperation{
+			{
+				Op:    aws.String(apigateway.OpReplace),
+				Path:  aws.String("/description"),
+				Value: aws.String(d.Get("description").(string)),
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	log.Printf("[DEBUG] Updated API Gateway Documentation Version %s", d.Id())
+
+	return resourceAwsApiGatewayDocumentationVersionRead(d, meta)
+}
+
+func resourceAwsApiGatewayDocumentationVersionDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+	log.Printf("[DEBUG] Deleting API Gateway Documentation Version: %s", d.Id())
+
+	_, err := conn.DeleteDocumentationVersion(&apigateway.DeleteDocumentationVersionInput{
+		DocumentationVersion: aws.String(d.Get("version").(string)),
+		RestApiId:            aws.String(d.Get("rest_api_id").(string)),
+	})
+	return err
+}
+
+func decodeApiGatewayDocumentationVersionId(id string) (string, string, error) {
+	parts := strings.Split(id, "/")
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("Expected ID in the form of REST-API-ID/VERSION, given: %q", id)
+	}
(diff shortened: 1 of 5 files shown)


aws/resource_aws_api_gateway_documentation_version.go
```diff
@@ -0,0 +1,138 @@
...
+		},
+	}
+}
+
+func resourceAwsApiGatewayDocumentationVersionCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+
+	restApiId := d.Get("rest_api_id").(string)
+
+	params := &apigateway.CreateDocumentationVersionInput{
```
bflad: Question: the [`CreateDocumentationVersionInput` type](https://docs.aws.amazon.com/sdk-for-go/api/service/apigateway/#CreateDocumentationVersionInput) also supports `StageName`, do we need to support it as well?
```
// The stage name to be associated with the new documentation snapshot.
StageName *string `locationName:"stageName" type:"string"`
```

aws/resource_aws_api_gateway_documentation_version.go
```diff
@@ -0,0 +1,138 @@
...
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("version", docVersion)
+	d.Set("rest_api_id", apiId)
+	d.Set("description", version.Description)
+	d.Set("version", version.Version)
```
bflad: Setting `version` twice 😉

aws/resource_aws_api_gateway_documentation_version_test.go
```diff
@@ -0,0 +1,266 @@
...
+					resource.TestCheckResourceAttr(resourceName, "version", version),
+					resource.TestCheckResourceAttr(resourceName, "description", uDescription),
+					resource.TestCheckResourceAttrSet(resourceName, "rest_api_id"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAWSAPIGatewayDocumentationVersion_importBasic(t *testing.T) {
```
bflad: Nitpick: Something I've been trying to remember to do is include `ImportState` `TestStep`s with each of the regular tests. We get the extra benefit of checking the additional attributes associated with each of the test cases and don't need explicit import test cases (which generally get neglected unless there's a 🐛  - too late!).

aws/resource_aws_api_gateway_documentation_version.go
```diff
@@ -0,0 +1,138 @@
...
+	log.Printf("[DEBUG] Deleting API Gateway Documentation Version: %s", d.Id())
+
+	_, err := conn.DeleteDocumentationVersion(&apigateway.DeleteDocumentationVersionInput{
+		DocumentationVersion: aws.String(d.Get("version").(string)),
+		RestApiId:            aws.String(d.Get("rest_api_id").(string)),
+	})
+	return err
+}
+
+func decodeApiGatewayDocVersionId(id string) (string, string, error) {
```
bflad: Worst OCD nitpick maybe of all time 😢 : Can you spell out `Documentation` so we can keep everything consistent with `DocumentationVersion`

aws/resource_aws_api_gateway_documentation_version_test.go
```diff
@@ -0,0 +1,266 @@
...
+func testAccAWSAPIGatewayDocumentationVersionAllFieldsConfig(version, apiName, stageName, description string) string {
+	return fmt.Sprintf(`
+resource "aws_api_gateway_documentation_version" "test" {
+  version = "%s"
+  rest_api_id = "${aws_api_gateway_rest_api.test.id}"
+  description = "%s"
+  depends_on = ["aws_api_gateway_documentation_part.test"]
+}
+
+resource "aws_api_gateway_documentation_part" "test" {
```
bflad: Nitpick: Can we reuse something from the `aws_api_gateway_documentation_part` test configurations instead of duplicating this whole large configuration? 😄

website/docs/r/api_gateway_documentation_version.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+layout: "aws"
+page_title: "AWS: aws_api_gateway_documentation_version"
+sidebar_current: "docs-aws-resource-api-gateway-documentation-version"
+description: |-
+  Provides a settings of an API Gateway Documentation Version.
+---
+
+# aws_api_gateway_documentation_version
+
+Provides a settings of an API Gateway Documentation Version.
```
bflad: Typo 🕶 : Maybe `a resource to manage an` here and slightly above?

aws/resource_aws_api_gateway_documentation_version.go
radeksimko: I intentionally removed it, although it was in my original implementation. CloudFormation also doesn't have it, probably for the same reason https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-documentationversion.html

It has a very problematic workflow. You need to create the stage 1st (for referencing in APIG Documentation Version), but you cannot delete the APIG Documentation Version before deleting the stage. Basically the dependency chain reverts during destruction.

Either way we shouldn't need it, because https://www.terraform.io/docs/providers/aws/r/api_gateway_stage.html#documentation_version allows users to associate the two.

aws/resource_aws_api_gateway_documentation_version_test.go
radeksimko: I'm not sure it's a good idea as tests should ideally be as isolated as possible, at least cross resource, IMO.

aws/resource_aws_api_gateway_documentation_version_test.go
radeksimko: I have no objections against adding more import tests with different configurations, but I don't think it's a good idea to add `ImportState` to existing tests which aren't called `_import`.

If we break import (for any reason) and nothing else I expect `_import` tests to fail, but no other tests, and vice versa, ideally. While in reality this is may not be always true I think we should aim for such isolation so it's easier to debug test failures as they appear.

There's benefit to isolation, I think - to keep us sane and focused 😉

radeksimko:
```
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_basic
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_basic (46.78s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_allFields
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_allFields (97.67s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_importBasic
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_importBasic (44.24s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_importAllFields
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_importAllFields (61.75s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	250.464s
```

aws/resource_aws_api_gateway_documentation_version_test.go
radeksimko: I added a new test - `TestAccAWSAPIGatewayDocumentationVersion_importAllFields` which should address this concern, hopefully? 😃

aws/resource_aws_api_gateway_documentation_version.go
bflad: 👍  sounds good

aws/resource_aws_api_gateway_documentation_version_test.go
bflad: I don't mean to harp on this resource/test specifically, but here's where I'm coming from, using this resource's testing as an example. I'm not asking you to change this one, but I think its something worth thinking about. 😄 

* We now have two of these import tests, duplicating the regular ones except a new `TestStep`. This means the testing is going to take roughly twice as long. Not the biggest deal, but even running parallel these all add up. e.g.

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayDocumentationVersion'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayDocumentationVersion -timeout 120m
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_basic
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_basic (14.69s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_allFields
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_allFields (57.37s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_importBasic
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_importBasic (16.63s)
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_importAllFields
--- PASS: TestAccAWSAPIGatewayDocumentationVersion_importAllFields (43.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	132.701s
```

* The "regular" part (first `TestStep`) of these `_importX` tests can also fail because of other resources. There are plenty of examples of these in the EC2 or KMS daily acceptance testing. So if you're looking at just the test name and not the actual error message (we're looking at the error message right? 😉 ), you can be misled already.

* I think `TestStep.ImportStateVerify` seems to do a pretty good job of calling out that its an import problem by the error message.

```go
// Comment out in resourceAwsApiGatewayDocumentationVersionRead
// d.Set("description", version.Description)

// Add in resourceAwsApiGatewayDocumentationVersionCreate
d.Set("description", d.Get("description").(string))

// Add to TestAccAWSAPIGatewayDocumentationVersion_allFields
			resource.TestStep{
				ImportState:       true,
				ImportStateVerify: true,
			},
```

It returns:
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayDocumentationVersion_allFields'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayDocumentationVersion_allFields -timeout 120m
=== RUN   TestAccAWSAPIGatewayDocumentationVersion_allFields
--- FAIL: TestAccAWSAPIGatewayDocumentationVersion_allFields (34.59s)
	testing.go:513: Step 2 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) {
		}


		(map[string]string) (len=1) {
		 (string) (len=11) "description": (string) (len=40) "Tf Acc Test description updated xjsh6nsw"
		}

FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	34.628s
```

Combined with maybe a little bit better naming of the tests (e.g. `_description` in this case), we'll know from the test name its a `description` problem and from the error message its an import problem. 

Hope this makes sense! 🍻  I guess mentally I am treating an attribute's import just a "feature" of an attribute, which should be logically tested with the other CRUD actions on it.

radeksimko: > mentally I am treating an attribute's import just a "feature" of an attribute, which should be logically tested with the other CRUD actions on it.

That's an interesting POV. I never thought about that like this. We can discuss that on Slack with other folks later. 👌

radeksimko:
> we can discuss my likely crazy opinions of import testing outside this PR

Those opinions are important and useful for a healthy discussion 😉  But yeah, let's discuss this outside of this PR.

bflad:

PR #3304, opened 2018-02-09
data.aws_kms_alias: Use kms:DescribeKey to get target key Id & ARN

handlerbot:
Context/fixes: https://github.com/terraform-providers/terraform-provider-aws/issues/2009#issuecomment-363072588

`make testacc TESTARGS='-run=TestAccDataSourceAwsKmsAlias'` passes.

cc @bflad @trung

### DIFF
--- a/aws/data_source_aws_kms_alias.go
+++ b/aws/data_source_aws_kms_alias.go
@@ -5,7 +5,7 @@ import (
 	"log"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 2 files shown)


bflad:
@handlerbot wow this is awesome, good find! Is there a point to calling `ListAliases` at all? Seems like we should just skip ahead to `DescribeKey` where we allow passing in the name directly (and maybe allowing ARN as an optional argument for cross-account)

aws/data_source_aws_kms_alias.go
```diff
@@ -64,25 +64,27 @@ func dataSourceAwsKmsAliasRead(d *schema.ResourceData, meta interface{}) error {
...
-		}
-		d.Set("target_key_arn", targetKeyARN.String())
 
-		d.Set("target_key_id", alias.TargetKeyId)
+	req := &kms.DescribeKeyInput{
+		KeyId: aws.String(target.(string)),
 	}
+	resp, err := conn.DescribeKey(req)
+	if err != nil {
+		return errwrap.Wrapf("Error calling KMS DescribeKey: {{err}}", err)
```
bflad: FYI I know there are a bunch of places in the repository currently using this, but we don't need the context wrapping of `errwrap` when we're just returning an error message. `fmt.Errorf` is just fine 😄

handlerbot: OK, got it. Want I should fix this, or merge and do a global sweep & cleanup later?

bflad: We'll do it later, no sense holding this up. 👍

bflad:

PR #3316, opened 2018-02-10
resource/aws_lambda_alias: Support Traffic Shifting

eggsbenjamin:
Added `routing_config`property to `aws_lambda_alias` which contains `additional_version_weights` to afford lambda alias traffic shifting.

Acceptance tests:
```
=== RUN   TestAccAWSLambdaAlias_basic
--- PASS: TestAccAWSLambdaAlias_basic (37.61s)
=== RUN   TestAccAWSLambdaAlias_routingConfig
--- PASS: TestAccAWSLambdaAlias_routingConfig (69.14s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       106.787s
```


eggsbenjamin:
Hey @radeksimko, I hope I don't sound impatient but, is there any chance I could get some feedback on this submission please? Cheers.

alikor:
When will this feature be released?

kylegalbraith:
Any idea when this feature will be rolled out? We would like to use it for some blue/green deployments.

aws/resource_aws_lambda_alias.go
```diff
@@ -97,6 +126,7 @@ func resourceAwsLambdaAliasRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("function_version", aliasConfiguration.FunctionVersion)
 	d.Set("name", aliasConfiguration.Name)
 	d.Set("arn", aliasConfiguration.AliasArn)
+	d.Set("routing_config", aliasConfiguration.RoutingConfig)
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("routing_config", aliasConfiguration.RoutingConfig); err != nil {
  return fmt.Errorf("error setting routing_config: %s", err)
}
```

After adding this error checking, you should start receiving type errors as `aliasConfiguration.RoutingConfig` is a SDK object and not a `[]interface{}`. We generally handle this via a "flattenServiceObject()" function. e.g.

```go

func flattenLambdaAliasRoutingConfiguration(arc *lambda.AliasRoutingConfiguration) []interface{} {
	if arc == nil {
		return []interface{}{}
	}

	m := map[string]interface{}{
		"additional_version_weights": aws.Float64ValueMap(arc.AdditionalVersionWeights)
	}

	return []interface{}{m}
}
```

Which can be referenced via:

```
if err := d.Set("routing_config", flattenLambdaAliasRoutingConfiguration(aliasConfiguration.RoutingConfig)); err != nil {
  return fmt.Errorf("error setting routing_config: %s", err)
}
```

website/docs/r/lambda_alias.html.markdown
```diff
@@ -30,10 +35,17 @@ resource "aws_lambda_alias" "test_alias" {
...
+
+* `additional_version_weights` - (Optional) A map that defines the proportion of events that should be sent to different versions of a lambda function.
 
 ## Attributes Reference
 
 * `arn` - The Amazon Resource Name (ARN) identifying your Lambda function alias.
 
 [1]: http://docs.aws.amazon.com/lambda/latest/dg/welcome.html
 [2]: http://docs.aws.amazon.com/lambda/latest/dg/API_CreateAlias.html
+[3]: https://docs.aws.amazon.com/lambda/latest/dg/API_AliasRoutingConfiguration.html
```
bflad: It looks like this documentation reference was added but never referenced above.

eggsbenjamin:
Hi @bflad , thanks for the response. Cool, I'll update the PR!

eggsbenjamin:
@bflad updated with the suggested changes. Cheers.

bflad:

PR #3317, opened 2018-02-10
Force vpc endpoint type to gateway in govcloud partition.

jmcarp:
The aws govcloud partition only supports gateway type vpc endpoints
and doesn't return a type when listing connections. To fix, temporarily
force vpc endpoint types to gateway if partition is govcloud.

After vpc endpoint types were added in 1.9, govcloud vpc endpoints
want to destroy and recreate on every apply, since the aws api doesn't
return an endpoint type.

cc @wjwoodson

### DIFF
--- a/aws/resource_aws_vpc_endpoint.go
+++ b/aws/resource_aws_vpc_endpoint.go
@@ -192,7 +192,7 @@ func resourceAwsVpcEndpointRead(d *schema.ResourceData, meta interface{}) error 
 		return nil
 	}
 
-	return vpcEndpointAttributes(d, vpce.(*ec2.VpcEndpoint), conn, meta.(*AWSClient).IsGovCloud())
+	return vpcEndpointAttributes(d, vpce.(*ec2.VpcEndpoint), meta)
 }
 
 func resourceAwsVpcEndpointUpdate(d *schema.ResourceData, meta interface{}) error {
@@ -341,17 +341,16 @@ func vpcEndpointWaitUntilAvailable(d *schema.ResourceData, conn *ec2.EC2) error 
 	return nil
 }
 
-func vpcEndpointAttributes(d *schema.ResourceData, vpce *ec2.VpcEndpoint, conn *ec2.EC2, isGovCloud bool) error {
+func vpcEndpointAttributes(d *schema.ResourceData, vpce *ec2.VpcEndpoint, meta interface{}) error {
 	d.Set("state", vpce.State)
 	d.Set("vpc_id", vpce.VpcId)
 
 	serviceName := aws.StringValue(vpce.ServiceName)
 	d.Set("service_name", serviceName)
-	if isGovCloud {
+	if meta.(*AWSClient).IsGovCloud() {
 		d.Set("vpc_endpoint_type", ec2.VpcEndpointTypeGateway)
 	} else {
-		vpceType := aws.StringValue(vpce.VpcEndpointType)
-		d.Set("vpc_endpoint_type", vpceType)
+		d.Set("vpc_endpoint_type", vpce.VpcEndpointType)
 	}
 
 	policy, err := structure.NormalizeJsonString(aws.StringValue(vpce.PolicyDocument))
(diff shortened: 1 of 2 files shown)


jmcarp:
By the way, it might be useful to add an integration test that creates a resource, then verifies that the plan is empty.

aws/resource_aws_vpc_endpoint.go
```diff
@@ -341,14 +341,18 @@ func vpcEndpointWaitUntilAvailable(d *schema.ResourceData, conn *ec2.EC2) error
...
 	d.Set("vpc_id", vpce.VpcId)
 
 	serviceName := aws.StringValue(vpce.ServiceName)
 	d.Set("service_name", serviceName)
-	vpceType := aws.StringValue(vpce.VpcEndpointType)
-	d.Set("vpc_endpoint_type", vpceType)
+	if isGovCloud {
+		d.Set("vpc_endpoint_type", ec2.VpcEndpointTypeGateway)
+	} else {
+		vpceType := aws.StringValue(vpce.VpcEndpointType)
```
ewbankkit: No need to call `aws.StringValue` on the VPC endpoint type as `d.Set` will handle `*string` parameters. I only call it above for the service name as the value is used later.

aws/resource_aws_vpc_endpoint.go
```diff
@@ -341,14 +341,18 @@ func vpcEndpointWaitUntilAvailable(d *schema.ResourceData, conn *ec2.EC2) error
 	return nil
 }
 
-func vpcEndpointAttributes(d *schema.ResourceData, vpce *ec2.VpcEndpoint, conn *ec2.EC2) error {
+func vpcEndpointAttributes(d *schema.ResourceData, vpce *ec2.VpcEndpoint, conn *ec2.EC2, isGovCloud bool) error {
```
ewbankkit: Consider changing the function signature to
`func vpcEndpointAttributes(d *schema.ResourceData, meta interface{}, vpce *ec2.VpcEndpoint) error`
as both `conn` and `isGovCloud` can be derived from `meta`?

jmcarp:
Updated. WDYT @ewbankkit ?

jmcarp:
Pinging @radeksimko or @bflad for review when you have time.

jmcarp:
ping @bflad

bflad:
Can you provide a previously failing configuration and the outputs from plan/apply?

jmcarp:
@bflad: terraform plan/apply doesn't fail because of this issue--it recreates the endpoint on every apply. Every `plan` says something like `module.stack.module.base.module.vpc.aws_vpc_endpoint.private-s3 (new resource required)` because it wants to change the endpoint type: `vpc_endpoint_type:          "" => "Gateway" (forces new resource)`. Every `apply` recreates the resource, but since govcloud ignores the endpoint type parameter, we repeat the cycle on the next `plan`.

jmcarp:
@bflad: does that description make sense? Let me know if more output would be helpful.

aws/resource_aws_vpc_endpoint.go
```diff
@@ -341,14 +341,17 @@ func vpcEndpointWaitUntilAvailable(d *schema.ResourceData, conn *ec2.EC2) error
 	return nil
 }
 
-func vpcEndpointAttributes(d *schema.ResourceData, vpce *ec2.VpcEndpoint, conn *ec2.EC2) error {
+func vpcEndpointAttributes(d *schema.ResourceData, vpce *ec2.VpcEndpoint, meta interface{}) error {
```
bflad: Let's revert this, I'll explain below. 😄

aws/resource_aws_vpc_endpoint.go
```diff
@@ -341,14 +341,17 @@ func vpcEndpointWaitUntilAvailable(d *schema.ResourceData, conn *ec2.EC2) error
...
 	d.Set("state", vpce.State)
 	d.Set("vpc_id", vpce.VpcId)
 
 	serviceName := aws.StringValue(vpce.ServiceName)
 	d.Set("service_name", serviceName)
-	vpceType := aws.StringValue(vpce.VpcEndpointType)
-	d.Set("vpc_endpoint_type", vpceType)
+	if meta.(*AWSClient).IsGovCloud() && aws.StringValue(vpce.VpcEndpointType) == "" {
```
bflad: Instead of relying on a `meta.(*AWSClient).IsGovCloud()` check, which one day may become obsolete, let's just perform the second check `aws.StringValue(vpce.VpcEndpointType) == ""` with a comment above it why the logic is in place. 👍

jmcarp:
Sounds good, updated.

bflad:

lorengordon:
Thanks @bflad, tested and it does indeed solve the problem I was having.

PR #3321, opened 2018-02-10
Trim trailing dot when importing route53 records.

### LINKED ISSUE #241: aws_route53_zone imports with a trailing dot
_This issue was originally opened by @TaiSHiNet as hashicorp/terraform#8511. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi, 

I'm on version 0.7.1 and when running an aws_route_53 import it'll end up in the state file with a trailing dot (as in bind, and as shown in the console).
This causes the resource to try and be recreated due to a mismatch with the terraform file.
How to reproduce:

Create a Route53 zone manually
Create the TF resource in a file
Import Route53 zone into your state
Run terraform plan

I could have just bypassed this by by adding a trailing dot to the resource, but documentation shows an example without it and it might be misleading to have both ways.

Bests

jmcarp:
We already trim trailing dots from names in https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_route53_record.go#L42-L45, so we should also trim trailing dots on import, so that we can import a resource, plan, and expect no changes.

[Resolves #241]

cc @wjwoodson

### DIFF
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -922,5 +922,6 @@ func parseRecordId(id string) [4]string {
 			}
 		}
 	}
+	recName = strings.TrimSuffix(recName, ".")
 	return [4]string{recZone, recName, recType, recSet}
 }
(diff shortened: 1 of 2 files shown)


bflad:

PR #3323, opened 2018-02-10
Fix docs for codebuild, so the markdown parses properly

nicwise:
Also includes @atsushi-ishibashi doc addition from https://github.com/terraform-providers/terraform-provider-aws/pull/2811 - they kinda need to go together or not at all.


bflad:
@nicwise if you remove `type` here (since the PR might take some time to get fixed up) and just add the newline I'll get this in. Git should do the right thing in the other PR and not touch this change.

PR #3324, opened 2018-02-11
CodeBuild in VPC continuing Issue #2547

moofish32:
updating [2547](https://github.com/terraform-providers/terraform-provider-aws/pull/2547)

@bflad -- I have made your requested changes here. 
@drsolaris happy to hand this back over, just trying to get it merged. I did squash your commits down during the conflict resolution.

@bflad -- the sourceAuth tests obviously fail locally, I'm not sure if it's worth making that test more intelligent.

### DIFF
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -182,7 +182,7 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 			},
 			"tags": tagsSchema(),
 			"vpc_config": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
@@ -192,15 +192,17 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 							Required: true,
 						},
 						"subnets": {
-							Type:     schema.TypeList,
+							Type:     schema.TypeSet,
 							Required: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
+							// Set:      schema.HashString,
 							MaxItems: 16,
 						},
 						"security_group_ids": {
-							Type:     schema.TypeList,
+							Type:     schema.TypeSet,
 							Required: true,
 							Elem:     &schema.Schema{Type: schema.TypeString},
+							// Set:      schema.HashString,
 							MaxItems: 5,
 						},
 					},
@@ -240,9 +242,8 @@ func resourceAwsCodeBuildProjectCreate(d *schema.ResourceData, meta interface{})
 		params.TimeoutInMinutes = aws.Int64(int64(v.(int)))
 	}
 
-	if _, ok := d.GetOk("vpc_config"); ok {
-		vpcConfig := expandVpcConfig(d)
-		params.VpcConfig = vpcConfig
+	if v, ok := d.GetOk("vpc_config"); ok {
+		params.VpcConfig = expandCodeBuildVpcConfig(v.([]interface{}))
 	}
 
 	if v, ok := d.GetOk("tags"); ok {
@@ -360,27 +361,19 @@ func expandProjectEnvironment(d *schema.ResourceData) *codebuild.ProjectEnvironm
 	return projectEnv
 }
 
-func expandVpcConfig(d *schema.ResourceData) *codebuild.VpcConfig {
-	configs := d.Get("vpc_config").(*schema.Set).List()
-	data := configs[0].(map[string]interface{})
+func expandCodeBuildVpcConfig(rawVpcConfig []interface{}) *codebuild.VpcConfig {
+	vpcConfig := codebuild.VpcConfig{}
+	if len(rawVpcConfig) == 0 {
+		return &vpcConfig
+	} else {
 
-	vpcConfig := codebuild.VpcConfig{
-		VpcId: aws.String(data["vpc_id"].(string)),
+		data := rawVpcConfig[0].(map[string]interface{})
+		vpcConfig.VpcId = aws.String(data["vpc_id"].(string))
+		vpcConfig.Subnets = expandStringList(data["subnets"].(*schema.Set).List())
+		vpcConfig.SecurityGroupIds = expandStringList(data["security_group_ids"].(*schema.Set).List())
+
+		return &vpcConfig
 	}
-
-	var vpcConfigSubnets []*string
-	for _, v := range data["subnets"].([]interface{}) {
-		vpcConfigSubnets = append(vpcConfigSubnets, aws.String(v.(string)))
-	}
-	vpcConfig.Subnets = vpcConfigSubnets
-
-	var vpcSecurityGroupIds []*string
-	for _, s := range data["security_group_ids"].([]interface{}) {
-		vpcSecurityGroupIds = append(vpcSecurityGroupIds, aws.String(s.(string)))
-	}
-	vpcConfig.SecurityGroupIds = vpcSecurityGroupIds
-
-	return &vpcConfig
 }
 
 func expandProjectSource(d *schema.ResourceData) codebuild.ProjectSource {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_codebuild_project.go
```diff
@@ -548,6 +607,31 @@ func flattenAwsCodebuildProjectSource(source *codebuild.ProjectSource) []interfa
...
+	if vpcConfig != nil {
+		values := map[string]interface{}{}
+
+		values["vpc_id"] = *vpcConfig.VpcId
+
+		var subnets []string
+		for _, s := range vpcConfig.Subnets {
+			subnets = append(subnets, *s)
+		}
+		sort.Strings(subnets)
```
moofish32: @bflad -- I'm still getting random sort ordering issues. Where am I going wrong?

aws/resource_aws_codebuild_project.go
```diff
@@ -380,15 +431,19 @@ func resourceAwsCodeBuildProjectRead(d *schema.ResourceData, meta interface{}) e
...
+	if err := d.Set("environment", schema.NewSet(resourceAwsCodeBuildProjectEnvironmentHash, flattenAwsCodeBuildProjectEnvironment(project.Environment))); err != nil {
 		return err
 	}
 
-	if err := d.Set("source", flattenAwsCodebuildProjectSource(project.Source)); err != nil {
+	if err := d.Set("source", flattenAwsCodeBuildProjectSource(project.Source)); err != nil {
+		return err
+	}
+
+	if err := d.Set("vpc_config", flattenAwsCodeBuildVpcConfig(project.VpcConfig)); err != nil {
```
moofish32: No matter how I sort subnets (and I've tried a few). I seem to get failing scenarios, but this line seems to be the indicator of the failure:
```
vpc_config is set to [map[security_group_ids:[sg-19a81766] vpc_id:vpc-61a22318 subnets:[subnet-d503a08f subnet-953c9fcf]]]
vpc_config is set to [map[vpc_id:vpc-61a22318 subnets:[subnet-d503a08f subnet-953c9fcf] security_group_ids:[sg-19a81766]]]
```

The object has the same values. The subnets are in the same order, but I get this error:

```
--- FAIL: TestAccAWSCodeBuildProject_basic (43.42s)
        testing.go:513: Step 0 error: After applying this step, the plan was not empty:

                DIFF:

                UPDATE: aws_codebuild_project.foo
                  vpc_config.0.subnets.0: "subnet-d503a08f" => "subnet-953c9fcf"
                  vpc_config.0.subnets.1: "subnet-953c9fcf" => "subnet-d503a08f"

                STATE:
```

Is it easier just go back to Set and calculate the hash for the difference?

bflad: The `subnets` and `security_group_ids` can be `schema.TypeSet`, but you don't need a special hash function for them. Terraform core will "do the right thing" when comparing simple string sets.

bflad:
> @bflad -- the sourceAuth tests obviously fail locally, I'm not sure if it's worth making that test more intelligent.

Are you on the latest master code? What is the error you are seeing? We should address this separately if its an issue as it passes in our daily acceptance testing.

```
--- PASS: TestAccAWSCodeBuildProject_sourceAuth (30.17s)
```

moofish32:
@bflad sourceAuth is generating this for me: 
```
--- FAIL: TestAccAWSCodeBuildProject_sourceAuth (9.65s)
        testing.go:513: Step 1 error: Error applying: 1 error(s) occurred:

                * aws_codebuild_project.foo: 1 error(s) occurred:

                * aws_codebuild_project.foo: [ERROR] Error creating CodeBuild project: InvalidInputException: No Access token found, please visit AWS CodeBuild console to connect to GitHub
                        status code: 400, request id: 1be10e94-1012-11e8-91d8-f7a26fdbda1c

                Error: Check failed: Default error in CodeBuild Test

                State: <no state>
```

I expect this is due to my git setup having some unique-ness because I have public github, work github enterprise, and experimental Code Commit (work). All being used in various flavors on my laptop. Is this worth a separate issue?

moofish32:
Regarding the in VPC config. I've made the changes back to string and `Set` please let me know if we need any other work to get this into the merge queue.

aws/resource_aws_codebuild_project.go
```diff
@@ -181,6 +181,33 @@ func resourceAwsCodeBuildProject() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"vpc_id": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"subnets": {
+							Type:     schema.TypeSet,
+							Required: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+							Set:      schema.HashString,
```
bflad: `Set` function should not be required for a simple `TypeString` set (same with security_group_ids attribute). Although you might have to do this since its nested, I think I ran into something similar recently but would love confirmation.

aws/resource_aws_codebuild_project.go
```diff
@@ -548,6 +606,19 @@ func flattenAwsCodebuildProjectSource(source *codebuild.ProjectSource) []interfa
 	return l
 }
 
+func flattenAwsCodeBuildVpcConfig(vpcConfig *codebuild.VpcConfig) []interface{} {
+	if vpcConfig != nil {
+		values := map[string]interface{}{}
+
+		values["vpc_id"] = *vpcConfig.VpcId
+		values["subnets"] = schema.NewSet(schema.HashString, flattenStringList(vpcConfig.Subnets))
```
bflad: Does this work with just `values["subnets"] = flattenStringList(vpcConfig.Subnets)`?

aws/resource_aws_codebuild_project_test.go
```diff
@@ -350,8 +365,27 @@ func testAccCheckAWSCodeBuildProjectDestroy(s *terraform.State) error {
 	return fmt.Errorf("Default error in CodeBuild Test")
 }
 
-func testAccAWSCodeBuildProjectConfig_basic(rName string) string {
+func testAccAWSCodeBuildProjectConfig_basic(rName string, subnet string) string {
```
bflad: Same with the mention above about the test itself, the VPC configuration should be separated as well. Thanks!

aws/resource_aws_codebuild_project.go
moofish32:
```--- FAIL: TestAccAWSCodeBuildProject_basic (44.01s)

                * aws_codebuild_project.foo: 1 error(s) occurred:

                * aws_codebuild_project.foo: Invalid address to set: []string{"vpc_config", "0", "subnets"}

                Error: Error refreshing: 1 error(s) occurred:

                * aws_codebuild_project.foo: 1 error(s) occurred:

                * aws_codebuild_project.foo: aws_codebuild_project.foo: Invalid address to set: []string{"vpc_config", "0", "subnets"}

                State: <nil>
FAIL
```
I don't think so sadly :(

moofish32: It appears `d.Set` has the magic buried in there to make this work. I tried a few tricks to pull it out so it would work for subresources, but I couldn't quite make it work. I've stopped and decided that perhaps core should some features to set and get with subresources instead of my hack-ish attempts. If you think we should invest more here let me know, but right now I didn't want to block this with that.

moofish32:
@bflad -- one more review please? Thanks for the feedback so far.

moofish32:
@bflad -- did we lose momentum?

aws/resource_aws_codebuild_project_test.go
```diff
@@ -686,3 +737,40 @@ resource "aws_codebuild_project" "foo" {
 }
 `, rName, authResource, authType)
 }
+
+func hclVpcResources() string {
```
bflad: FYI these function names are global across the entire provider so we would generally prefer sticking with the long for `testAccAWSCodeBuildProjectConfig_X`

bflad:

PR #3327, opened 2018-02-12
New Resource: aws_gamelift_fleet

radeksimko:
## Test results

```
=== RUN   TestAccAWSGameliftFleet_basic
--- PASS: TestAccAWSGameliftFleet_basic (1111.72s)
=== RUN   TestAccAWSGameliftFleet_allFields
--- PASS: TestAccAWSGameliftFleet_allFields (1262.53s)
```

Related: https://github.com/terraform-providers/terraform-provider-aws/issues/2546

### DIFF
--- a/aws/resource_aws_gamelift_fleet.go
+++ b/aws/resource_aws_gamelift_fleet.go
@@ -1,0 +1,596 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"reflect"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/gamelift"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsGameliftFleet() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsGameliftFleetCreate,
+		Read:   resourceAwsGameliftFleetRead,
+		Update: resourceAwsGameliftFleetUpdate,
+		Delete: resourceAwsGameliftFleetDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(70 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"build_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"ec2_instance_type": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 1024),
+			},
+			"description": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringLenBetween(1, 1024),
+			},
+			"ec2_inbound_permission": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 50,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"from_port": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(1, 60000),
+						},
+						"ip_range": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"protocol": {
+							Type:     schema.TypeString,
+							Required: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"TCP", "UDP",
+							}, false),
+						},
+						"to_port": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(1, 60000),
+						},
+					},
+				},
+			},
+			"metric_groups": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringLenBetween(1, 255),
+				},
+			},
+			"new_game_session_protection_policy": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  gamelift.ProtectionPolicyNoProtection,
+				ValidateFunc: validation.StringInSlice([]string{
+					gamelift.ProtectionPolicyNoProtection,
+					gamelift.ProtectionPolicyFullProtection,
+				}, false),
+			},
+			"operating_system": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"resource_creation_limit_policy": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"new_game_sessions_per_creator": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntAtLeast(0),
+						},
+						"policy_period_in_minutes": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntAtLeast(0),
+						},
+					},
+				},
+			},
+			"runtime_configuration": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"game_session_activation_timeout_seconds": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntBetween(1, 600),
+						},
+						"max_concurrent_game_session_activations": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntBetween(1, 2147483647),
+						},
+						"server_process": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 50,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"concurrent_executions": {
+										Type:         schema.TypeInt,
+										Required:     true,
+										ValidateFunc: validation.IntAtLeast(1),
+									},
+									"launch_path": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringLenBetween(1, 1024),
+									},
+									"parameters": {
+										Type:         schema.TypeString,
+										Optional:     true,
+										ValidateFunc: validation.StringLenBetween(1, 1024),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceAwsGameliftFleetCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).gameliftconn
+
+	input := gamelift.CreateFleetInput{
+		BuildId:         aws.String(d.Get("build_id").(string)),
+		EC2InstanceType: aws.String(d.Get("ec2_instance_type").(string)),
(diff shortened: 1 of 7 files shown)


aws/resource_aws_gamelift_fleet.go
```diff
@@ -0,0 +1,596 @@
...
+	return resourceAwsGameliftFleetRead(d, meta)
+}
+
+func resourceAwsGameliftFleetDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).gameliftconn
+
+	log.Printf("[INFO] Deleting Gamelift Fleet: %s", d.Id())
+	// It can take ~ 1 hr as Gamelift will keep retrying on errors like
+	// invalid launch path and remain in state when it can't be deleted :/
+	err := resource.Retry(60*time.Minute, func() *resource.RetryError {
```
bflad: Not sure if I have a good answer here, but it would be nice to tie this to the schema timeout because up to an extra hour of apply is an awful long time 🙁

aws/resource_aws_gamelift_fleet.go
```diff
@@ -0,0 +1,596 @@
...
+func getGameliftFleetFailures(conn *gamelift.GameLift, id string) ([]*gamelift.Event, error) {
+	var events []*gamelift.Event
+	err := _getGameliftFleetFailures(conn, id, nil, &events)
+	return events, err
+}
+
+func _getGameliftFleetFailures(conn *gamelift.GameLift, id string, nextToken *string, events *[]*gamelift.Event) error {
+	eOut, err := conn.DescribeFleetEvents(&gamelift.DescribeFleetEventsInput{
+		FleetId:   aws.String(id),
+		NextToken: nextToken,
```
bflad: It'd be so nice if the SDK provided paginators for everything 🤷‍♂️

aws/resource_aws_gamelift_fleet.go
```diff
@@ -0,0 +1,596 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isGameliftEventFailure(event *gamelift.Event) bool {
+	failureCodes := []string{
+		"FLEET_STATE_ERROR",
```
bflad: Attack of the SDK constants! 👽  (it'd be nice to lexicographically sort them too)

```
gamelift.EventCodeFleetStateError,
gamelift.EventCodeFleetInitializationFailed,
gamelift.EventCodeFleetBinaryDownloadFailed,
gamelift.EventCodeFleetValidationLaunchPathNotFound,
gamelift.EventCodeFleetValidationExecutableRuntimeFailure,
gamelift.EventCodeFleetValidationTimedOut,
gamelift.EventCodeFleetActivationFailed,
gamelift.EventCodeFleetActivationFailedNoInstances,
gamelift.EventCodeServerProcessInvalidPath,
gamelift.EventCodeServerProcessSdkInitializationTimeout,
gamelift.EventCodeServerProcessProcessReadyTimeout,
gamelift.EventCodeServerProcessCrashed,
gamelift.EventCodeServerProcessTerminatedUnhealthy,
gamelift.EventCodeServerProcessForceTerminated,
gamelift.EventCodeServerProcessProcessExitTimeout,
gamelift.EventCodeGameSessionActivationTimeout,
gamelift.EventCodeFleetVpcPeeringFailed,
```

aws/resource_aws_gamelift_fleet_test.go
```diff
@@ -0,0 +1,625 @@
...
+	if err != nil {
+		return fmt.Errorf("Error listing Gamelift Fleets: %s", err)
+	}
+
+	if len(resp.FleetIds) == 0 {
+		log.Print("[DEBUG] No Gamelift Fleets to sweep")
+		return nil
+	}
+
+	out, err := conn.DescribeFleetAttributes(&gamelift.DescribeFleetAttributesInput{
```
bflad: We should probably check `NextToken` here for good measure like `_getGameliftFleetFailures`

aws/resource_aws_gamelift_fleet_test.go
```diff
@@ -0,0 +1,625 @@
...
+
+	log.Printf("[INFO] Found %d Gamelift Fleets", len(out.FleetAttributes))
+
+	for _, attr := range out.FleetAttributes {
+		if !strings.HasPrefix(*attr.Name, testAccGameliftFleetPrefix) {
+			continue
+		}
+
+		log.Printf("[INFO] Deleting Gamelift Fleet %q", *attr.FleetId)
+		err := resource.Retry(1*time.Minute, func() *resource.RetryError {
```
bflad: Does this need to be 60 minutes to match the delete function? Otherwise the sweeper could trigger CI failures consistently.

aws/resource_aws_gamelift_fleet_test.go
```diff
@@ -0,0 +1,625 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSGameliftFleetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSGameliftFleetBasicConfig(fleetName, launchPath, params, buildName, bucketName, key, roleArn),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSGameliftFleetExists("aws_gamelift_fleet.test", &conf),
+					resource.TestCheckResourceAttrSet("aws_gamelift_fleet.test", "build_id"),
```
bflad: Missing `arn` attribute check:
```go
resource.TestMatchResourceAttr("aws_gamelift_fleet.test", "arn", regexp.MustCompile(`^arn:[^:]+:gamelift:[^:]+:[^:]+:.+$`))
```

aws/resource_aws_gamelift_fleet_test.go
```diff
@@ -0,0 +1,625 @@
...
+		out, err := conn.DescribeFleetAttributes(&gamelift.DescribeFleetAttributesInput{
+			FleetIds: aws.StringSlice([]string{rs.Primary.ID}),
+		})
+		if err != nil {
+			return err
+		}
+
+		attributes := out.FleetAttributes
+
+		if len(attributes) > 0 {
```
bflad: Do we need to check its status for `gamelift.FleetStatusDeleting` or `gamelift.FleetStatusTerminated`?

website/docs/r/gamelift_fleet.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `build_id` - (Required) ID of the Gamelift Build to be deployed on the fleet.
+* `ec2_instance_type` - (Required) Name of an EC2 instance type. e.g. `t2.micro`
+* `name` - (Required) The name of the fleet.
+* `description` - (Optional) Human-readable description of the fleet.
+* `ec2_inbound_permission` - (Optional) Range of IP addresses and port settings that permit inbound traffic to access server processes running on the fleet. See below.
+* `metric_groups` - (Optional) Name of a metric group to add this fleet to. A metric group tracks metrics across all fleets in the group. Defaults to `default`.
```
bflad: `List of names` to match schema and API, but max of one 👍

website/docs/r/gamelift_fleet.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+* `new_game_session_protection_policy` - (Optional) Game session protection policy to apply to all instances in this fleet. e.g. `FullProtection`. Defaults to `NoProtection`.
+* `resource_creation_limit_policy` - (Optional) Policy that limits the number of game sessions an individual player can create over a span of time for this fleet. sSee below.
```
bflad: Typo 🕶 : `sSee`

aws/resource_aws_gamelift_fleet.go
```diff
@@ -0,0 +1,596 @@
...
+				Optional: true,
+				MaxItems: 50,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"from_port": {
+							Type:         schema.TypeInt,
+							Required:     true,
+							ValidateFunc: validation.IntBetween(1, 60000),
+						},
+						"ip_range": {
```
bflad: I think we can validate this with `ValidateFunc: validateCIDRNetworkAddress,`

aws/resource_aws_gamelift_fleet.go
```diff
@@ -0,0 +1,596 @@
...
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{
+			gamelift.FleetStatusActivating,
+			gamelift.FleetStatusBuilding,
+			gamelift.FleetStatusDownloading,
+			gamelift.FleetStatusNew,
+			gamelift.FleetStatusValidating,
+		},
+		Target:  []string{gamelift.FleetStatusActive},
```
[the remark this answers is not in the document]
radeksimko: So I think we provide as much details as we can in such case. I'm intentionally _not_ catching `FleetStatusError` and let it fall through as error, so I can catch it below and ask for reasons of the state.

We also save some API calls that way since we don't ask for failures with each retry, but only when we _know_ things have failed.

aws/resource_aws_gamelift_fleet.go
radeksimko: Generally speaking we should only let users customize the timeout when we know that _intentionally_ different config (like bigger instance) requires higher timeout. This typically _doesn't_ involve error retries, because these are not affected by instance sizes or anything like that, these are always going to be the same for every user.

I see what you're saying here though and I have mixed feelings about that, but is returning error early any better? Basically you would _already know_ that the fleet has errors which would be raised during creation (that has sufficient timeout). Here the user is _knowingly_ deleting a fleet that is already in such broken state, so I'd argue that waiting is desired? 🤔

aws/resource_aws_gamelift_fleet_test.go
radeksimko: So the assumption I'm making here is that since we're performing ID-based lookup we're always going to get `1` or `0` results, so pagination should never be necessary?

aws/resource_aws_gamelift_fleet_test.go
radeksimko: `waitForGameliftFleetToBeDeleted` should guarantee that we don't as the fleet should be completely gone by this time. If it's not I'd rather treat that as bug.

bflad: Ah sorry, I missed this when looking at `waitForGameliftFleetToBeDeleted` - we're going for zero returned results instead of just `gamelift.FleetStatusTerminated` status. I thought you'd be getting the one result with the terminated status 👍

aws/resource_aws_gamelift_fleet_test.go
bflad: Sorry I meant to put this on `conn.ListFleets` above 😞  (just so this goes and sweeps everything there are additional results). Really just a minor nitpick though.

bflad:
Let me know when you'd like me to give another look! 👀

radeksimko:
@bflad PTAL

Hinidu:
Do you plan to implement fleet autoscaling policies? It is essential part of the fleet for us and many other developers.

radeksimko:
@Hinidu supporting autoscaling policy sounds like a good idea, although it should certainly be a separate resource, e.g. `aws_gamelift_autoscaling_policy` which maps to these API methods:

 - https://docs.aws.amazon.com/gamelift/latest/apireference/API_PutScalingPolicy.html
 - https://docs.aws.amazon.com/gamelift/latest/apireference/API_DescribeScalingPolicies.html
 - https://docs.aws.amazon.com/gamelift/latest/apireference/API_DeleteScalingPolicy.html

It's not on my near-term plan though, so I think we'd be happy to accept PRs for such resource.

Hinidu:
@radeksimko got it. I hope I will find some time to make these PRs :slightly_smiling_face:

aws/resource_aws_gamelift_fleet.go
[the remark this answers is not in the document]
bflad: Okay that makes sense, and is pretty awesome. 👍

aws/resource_aws_gamelift_fleet.go
bflad: Your argument is perfectly right - I think where I'm coming from would be fixed more with the Terraform core enhancement to be able to provide apply-time status messaging during (long) running operations.

bflad:

PR #3337, opened 2018-02-12
docs/aws_db_instance: Add doc notes regarding cross-region replication of encrypted RDS instances

mpilar:
Not directly but might help with this: https://github.com/terraform-providers/terraform-provider-aws/issues/518


mpilar:
@bflad sorry for the ping, but could you look at this really old PR as well?

Issue #2296, opened 2017-11-15
Applying bucket changes without terraform doesn't reflect in state after terraform refresh


srikiraju:
### Terraform Version
Terraform v0.10.8

### Affected Resource(s)
- aws_s3_bucket

### Terraform Configuration Files
```hcl
 resource "aws_s3_bucket" "www" {
     bucket = "main_bucket"
     acl    = "private"
    replication_configuration {
       role = "${aws_iam_role.s3_replication.arn}"

       rules {
         id     = "All"
         prefix = ""
         status = "Enabled"

         destination {
           bucket        = "arn:aws:s3:::otherbucket"
           storage_class = "STANDARD"
         }
      }
    }
 }
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Remove replication from AWS console
3. `terraform plan` - Same, plan does not detect that replication was removed
4. `terraform refresh` - Off-band changes not detected

I think a lot of the other options in s3_bucket have the same behaviour

devonbleak:
Confirming that changing ACLs also doesn't show up - had a bucket set to public today and my terraform with `acl = "private"` showed no changes required in the plan.

Chhed13:
I can't reproduce exactly this code, but I definitely found an issue with replication_configuration update function - it refreshes only if `replication.ReplicationConfiguration != nil` instead of set state to default value.

Chhed13:
Added pull request for replication.
As I can see in code there is now **read** function for **ACL** implemented at all.
For **cors_rule** block - the same situation as for replication - it updates only when not nil.
And I double check **request_payer**, **versioning** and  **lifecycle_rule**. They can work not correct on remove too.

bflad:
This has been released in [version 1.10.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1100-february-24-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

bflad:
(Since a commit already closed this ticket for replication_config, let's address each case individually)

PR #3349, opened 2018-02-13
#2296 aws_s3_bucket-replication-config-update-fix

Chhed13:
Always update replication_configuration fields.
Fix for #2296

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -337,6 +337,7 @@ func resourceAwsS3Bucket() *schema.Resource {
 			"replication_configuration": {
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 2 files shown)


aws/resource_aws_s3_bucket.go
```diff
@@ -337,6 +337,7 @@ func resourceAwsS3Bucket() *schema.Resource {
 			"replication_configuration": {
 				Type:     schema.TypeList,
 				Optional: true,
+				Computed: true,
```
bflad: Hey quick question -- what was the motivation for adding this?

Chhed13: Good question. The reason was - failed test

bflad: In this case, we don't want to change this and the test failure you mention is unfortunately a red herring. S3 eventual consistency is a beast (we get failures regularly in our daily acceptance testing 😬 ) so you're not alone. I believe setting `Computed: true` here basically means that we will ignore out of band changes to the replication configuration in AWS if its not defined in the Terraform configuration, which is a change in behavior for the resource.

Removing it and running the tests worked fine for me 👍 
```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSS3Bucket_Replication'
TF_ACC=1 go test ./aws -v -run=TestAccAWSS3Bucket_Replication -timeout 120m
=== RUN   TestAccAWSS3Bucket_Replication
--- PASS: TestAccAWSS3Bucket_Replication (87.72s)
=== RUN   TestAccAWSS3Bucket_ReplicationWithoutStorageClass
--- PASS: TestAccAWSS3Bucket_ReplicationWithoutStorageClass (57.96s)
=== RUN   TestAccAWSS3Bucket_ReplicationExpectVersioningValidationError
--- PASS: TestAccAWSS3Bucket_ReplicationExpectVersioningValidationError (32.61s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	178.355s
```

Chhed13: Thanks for the great explanation, now I revert it.
But for me the red flag was - that this test always pass before I made other changes. I explained it 2 step bucket creation process (first `Create` without replication, next `Update` and than we `Read` it several times while it come to state). On the first `Read` replication can absent, but what's changed - I added default line for `replication` to state and it can be read. 
While debug output shows that all is ok, looks like test read first version of state if the section is present

bflad:

Issue #167, opened 2017-06-13
aws_cloudfront_distribution shouldn't require restrictions key


hashibot:
_This issue was originally opened by @cweagans as hashicorp/terraform#6721. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

v0.6.15
### Affected Resource(s)
- aws_cloudfront_distribution
### Terraform Configuration Files

``` hcl
resource "aws_cloudfront_distribution" "sitename_cloudfront" {
  origin {
    domain_name = "${aws_s3_bucket.bucket_id.website_endpoint}"
    origin_id = "${aws_s3_bucket.bucket_id.id}"
  }

  enabled = true
  default_root_object = "index.html"

  aliases = [
    "domain1.com",
    "domain2.com"
  ]

  default_cache_behavior {
    allowed_methods = ["GET", "HEAD", "OPTIONS"]
    cached_methods = ["GET", "HEAD", "OPTIONS"]
    target_origin_id = "${aws_s3_bucket.bucket_id.id}"

    forwarded_values {
      query_string = false
      cookies {
        forward = "none"
      }
    }

    viewer_protocol_policy = "redirect-to-https"
    min_ttl = 0
    default_ttl = 3600
    max_ttl = 86400
  }

  # Use only edge servers in USA and EU.
  price_class = "PriceClass_100"

  viewer_certificate {
    acm_certificate_arn = "[mycert]"
    minimum_protocol_version = "TLSv1"
    ssl_support_method = "sni-only"
  }

}
```
### Expected Behavior

The cloudfront distribution should have been created with no restrictions (geo or otherwise).
### Actual Behavior

```
Errors:

  * aws_cloudfront_distribution.sitename_cloudfront: "restrictions": required field is not set
```
### Important Factoids

If the `restrictions` key isn't set, perhaps just considering this the default would work:

``` hcl
  restrictions {
    geo_restriction {
      restriction_type = "none"
    }
  }
```

PR #3364, opened 2018-02-14
added fix for issue #167 - defaulting restrictions to none for cloudf…

dpmerron-ltd:
…ront distributions

In reference to: https://github.com/terraform-providers/terraform-provider-aws/issues/167

Currently the aws_cloudfront_distribtuion requires 

`  restrictions {
    geo_restriction {
      restriction_type = "none"
    }
  }`

The proposed changes allows this section to be computed. 

Acceptance tests output

ok  	github.com/terraform-providers/terraform-provider-aws/aws	3.911s

First PR so go easy on me 👍

### DIFF
--- a/aws/resource_aws_cloudfront_distribution.go
+++ b/aws/resource_aws_cloudfront_distribution.go
@@ -432,16 +432,15 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 			},
 			"restrictions": {
 				Type:     schema.TypeSet,
-				Computed: true,
+				Optional: true,
 				Set:      restrictionsHash,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"geo_restriction": {
 							Type:     schema.TypeSet,
-							Computed: true,
+							Optional: true,
 							Set:      geoRestrictionHash,
-							MaxItems: 1,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"locations": {
@@ -452,7 +451,7 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 									"restriction_type": {
 										Type:     schema.TypeString,
 										Optional: true,
-										Default:  "none",
+										Default:  cloudfront.GeoRestrictionTypeNone,
 									},
 								},
 							},
--- a/aws/resource_aws_cloudfront_distribution.go
+++ b/aws/resource_aws_cloudfront_distribution.go
@@ -432,14 +432,14 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 			},
 			"restrictions": {
 				Type:     schema.TypeSet,
-				Required: true,
+				Computed: true,
 				Set:      restrictionsHash,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"geo_restriction": {
 							Type:     schema.TypeSet,
-							Required: true,
+							Computed: true,
 							Set:      geoRestrictionHash,
 							MaxItems: 1,
 							Elem: &schema.Resource{
@@ -451,7 +451,8 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 									},
 									"restriction_type": {
 										Type:     schema.TypeString,
-										Required: true,
+										Optional: true,
+										Default:  "none",
 									},
 								},
 							},
--- a/aws/resource_aws_cloudfront_distribution_test.go
+++ b/aws/resource_aws_cloudfront_distribution_test.go
@@ -44,9 +44,9 @@ func TestAccAWSCloudFrontDistribution_S3Origin(t *testing.T) {
 	})
 }
 
-func TestAccAWSCloudFrontDistribution_S3OriginNoRestrictions(t *testing.T) {
+func TestAccAWSCloudFrontDistribution_S3Origin_Restrictions(t *testing.T) {
 	ri := acctest.RandInt()
-	testConfig := fmt.Sprintf(testAccAWSCloudFrontDistributionS3ConfigWithNoRestrictions, ri, originBucket, logBucket, testAccAWSCloudFrontDistributionRetainConfig())
+	testConfig := fmt.Sprintf(testAccAWSCloudFrontDistributionS3ConfigRestrictions, ri, originBucket, logBucket, testAccAWSCloudFrontDistributionRetainConfig())
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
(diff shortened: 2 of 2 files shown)


aws/resource_aws_cloudfront_distribution.go
```diff
@@ -432,14 +432,14 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 			},
 			"restrictions": {
 				Type:     schema.TypeSet,
-				Required: true,
+				Computed: true,
```
bflad: Is there a reason you chose `Computed: true` here instead of `Optional: true`? Same with below

aws/resource_aws_cloudfront_distribution.go
```diff
@@ -451,7 +451,8 @@ func resourceAwsCloudFrontDistribution() *schema.Resource {
 									},
 									"restriction_type": {
 										Type:     schema.TypeString,
-										Required: true,
+										Optional: true,
+										Default:  "none",
```
bflad: Nitpick: we should use the SDK provided constant here: `Default: cloudfront.GeoRestrictionTypeNone`

aws/resource_aws_cloudfront_distribution_test.go
```diff
@@ -44,6 +44,31 @@ func TestAccAWSCloudFrontDistribution_S3Origin(t *testing.T) {
 	})
 }
 
+func TestAccAWSCloudFrontDistribution_S3OriginNoRestrictions(t *testing.T) {
```
bflad: Since we're changing the attribute from required to optional, I think it would make more sense to actually remove it from all the other test configurations and just ensure we have "`_restrictions`" test(s) that do actually try to set it.

bflad:
Also FYI:
> Acceptance tests output

> ==> Checking that code complies with gofmt requirements...
> go test ./... -timeout=30s -parallel=4
> ? github.com/terraform-providers/terraform-provider-aws	[no test files]
> ok github.com/terraform-providers/terraform-provider-aws/aws	3.911s

This output looks like the Go unit testing framework, which runs really fast across our codebase. The provider acceptance testing suite that actually calls AWS takes a _very_ long time for CloudFront distributions since they take a long time to disable and delete (10-15 minutes each). In this case to run all the CloudFront distribution acceptance tests, I would be running:

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSCloudFrontDistribution_'
```

aws/resource_aws_cloudfront_distribution.go
danmerron-sig: Just because I thought that it may require computing so that the default could be set to none further down the chain. Reading the aws docs again, it seems that is not the case so I have updated that.

aws/resource_aws_cloudfront_distribution.go
danmerron-sig: Fair Nitpick. Updated.

danmerron-sig:
In reference to all of the above comments @bflad.. I think it is not possible to have optional: true for those options. See the output of the acceptance test below when using Optional: 


`--- FAIL: TestAccAWSCloudFrontDistribution_importBasic (1239.71s)
	testing.go:513: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		UPDATE: aws_cloudfront_distribution.s3_distribution
		  restrictions.#:                                                      "1" => "0"
		  restrictions.1097372288.geo_restriction.#:                           "1" => "0"
		  restrictions.1097372288.geo_restriction.2625240281.locations.#:      "0" => "0"
		  restrictions.1097372288.geo_restriction.2625240281.restriction_type: "none" => ""`


When using the format I had before (with Computed: True) the tests pass:

`=== RUN   TestAccAWSCloudFrontDistribution_importBasic
--- PASS: TestAccAWSCloudFrontDistribution_importBasic (1240.00s)`

Happy to listen to further input you may have.

bflad:
@danmerron-sig thanks for the followup! That makes sense. Can you please address the other two items in my feedback or let me know if you do not have time for them? 😄  Then this should hopefully be good to go.

danmerron-sig:
I have addressed them I believe but havent had chance yet to let the acceptance test run all the way through. Hoping to do this tomorrow 👍 

Dan

danmerron-sig:
Hi @bflad I wonder if you can help here. When running the acceptance tests I get the following; 

> 		  restrictions.#:                                                      "1" => "0"
> 		  restrictions.1097372288.geo_restriction.#:                           "1" => "0"
> 		  restrictions.1097372288.geo_restriction.2625240281.locations.#:      "0" => "0"
> 		  restrictions.1097372288.geo_restriction.2625240281.restriction_type: "none" => ""

I believe this is due to a 0 input for restrictions, still actually creates a restriction but sets to state of disabled. Ideally. what I need to do here is set the default to '1' however with a TypeSet this is not possible. Any other ways around this? 

Dan

bflad:
This is complicated since ideally we want to be able to know when the CloudFront distribution restriction configuration diverges from the default "no geo restrictions" one when its not defined in the Terraform configuration. This is not possible when using `Computed: true` and why we are trying to avoid it here.

I can think of some hacky ways we can accomplish this today, but let me talk with my colleagues to see if we cannot come up with a good way to handle this situation of suppressing the difference in plan.

As an aside, we should leave `geo_restrictions` as `Required: true` and `MaxItems: 1`. Terraform will do the right thing if the parent attribute is `Optional` by only requiring it when necessary. If I remember correctly, its not valid to declare `restrictions` without `geo_restrictions` and its not valid to have multiple `geo_restrictions` configuration blocks.

danmerron-sig:
Thanks. I am keen to get this sorted and merged to tidy up some of my code... Will be interesting to know what your colleagues have to say about this!

danmerron-sig:
@bflad Was there any update on this?

zhelding:
Pull request #21306 has significantly refactored the AWS Provider codebase. As a result, most PRs opened prior to the refactor now have merge conflicts that must be resolved before proceeding.

Specifically, PR #21306 relocated the code for all AWS resources and data sources from a single `aws` directory to a large number of separate directories in `internal/service`, each corresponding to a particular AWS service. This separation of code has also allowed for us to simplify the names of underlying functions -- while still avoiding namespace collisions.

We recognize that many pull requests have been open for some time without yet being addressed by our maintainers. Therefore, we want to make it clear that resolving these conflicts _in no way affects the prioritization_ of a particular pull request. Once a pull request has been prioritized for review, the necessary changes will be made by a maintainer -- either directly or in collaboration with the pull request author.

For a more complete description of this refactor, including examples of how old filepaths and function names correspond to their new counterparts: please refer to issue #20000.

For a quick guide on how to amend your pull request to resolve the merge conflicts resulting from this refactor and bring it in line with our new code patterns: please refer to our [Service Package Refactor Pull Request Guide](https://github.com/hashicorp/terraform-provider-aws/blob/main/docs/contributing/service-package-pullrequest-guide.md).

PR #3365, opened 2018-02-14
resource/aws_iam_user_group_membership: allow non-exclusive group memberships

### LINKED ISSUE #2881: State being written incorrectly for IAM group membership?
### Terraform Version
Terraform v0.11.1
+ provider.aws v1.6.0
+ provider.template v1.0.0

### Affected Resource(s)
aws_iam_group_membership

```hcl
module "dev-user1" {
  "source" = "setup-users"
  "local_dev_group" = "${var.iam_dev_group}"
  "local_dev_username" = "user1"
}
module "dev-user2" {
  "source" = "setup-users"
  "local_dev_group" = "${var.iam_dev_group}"
  "local_dev_username" = "user2"
}

# In module setup-users

variable "dev_username" {}
variable "dev_group" {}

resource "aws_iam_user" "user" {
  name = "${var.dev_username}"
}

resource "aws_iam_group_membership" "group_membership" {
  name = "dev"
  users = ["${aws_iam_user.user.name}"]
  group = "${var.dev_group}"
}
```

### Debug Output
Unless this is absolutely necessary, can we skip this for now? Providing this would be difficult since the terraform config provided is sanitized, and I can't really run it without creating a bunch of bogus config in our AWS account.

### Panic Output
n/a

### Expected Behavior
Users get added to group once, and stay in group.

### Actual Behavior
If you run `terraform apply` several times in a row, the first time, it adds everyone to the group. Then (without changing any config), it removes everyone the second time you run terraform apply. Then, when you run it again, it adds everyone back.

### Steps to Reproduce
1. `terraform apply`
2. `terraform apply`

### Important Factoids
- Nothing special.

### References
None that I could find.

devonbleak:
Resolves #2881 - allows multiple aws_iam_group_memberships to co-exist peacefully by adding a new computed field `group_members` to manage adding/removing users.

Adds a `remove_unmanaged_members` property that can be set to true to restore previous functionality and documentation caveat about flapping with that setting.

### DIFF
--- a/aws/resource_aws_iam_user_group_membership_test.go
+++ b/aws/resource_aws_iam_user_group_membership_test.go
@@ -163,7 +163,7 @@ func testAccAWSUserGroupMembershipCheckGroupListForUser(userName string, groups 
 		for _, group := range groupsNeg {
 			for _, groupFound := range userGroupList.Groups {
 				if group == *groupFound.GroupName {
-					return fmt.Errorf("Required negative group found for %s: %s", userName, group)
+					return fmt.Errorf("Unexpected group found for %s: %s", userName, group)
 				}
 			}
 		}
--- a/aws/resource_aws_iam_user_group_membership_test.go
+++ b/aws/resource_aws_iam_user_group_membership_test.go
@@ -1,0 +1,331 @@
+package aws
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/iam"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSUserGroupMembership_basic(t *testing.T) {
+	rString := acctest.RandString(8)
+	userName1 := fmt.Sprintf("tf-acc-ugm-basic-user1-%s", rString)
+	userName2 := fmt.Sprintf("tf-acc-ugm-basic-user2-%s", rString)
+	groupName1 := fmt.Sprintf("tf-acc-ugm-basic-group1-%s", rString)
+	groupName2 := fmt.Sprintf("tf-acc-ugm-basic-group2-%s", rString)
+	groupName3 := fmt.Sprintf("tf-acc-ugm-basic-group3-%s", rString)
+	membershipName := fmt.Sprintf("tf-acc-ugm-basic-membership-%s", rString)
+
+	usersAndGroupsConfig := testAccAWSUserGroupMembershipConfigUsersAndGroups(userName1, userName2, groupName1, groupName2, groupName3)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccAWSUserGroupMembershipDestroy,
+		Steps: []resource.TestStep{
+			// simplest test
+			resource.TestStep{
+				Config: usersAndGroupsConfig + testAccAWSUserGroupMembershipConfigInit(membershipName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("aws_iam_user_group_membership.user1_test1", "user", userName1),
+					testAccAWSUserGroupMembershipCheckGroupListForUser(userName1, []string{groupName1}, []string{groupName2, groupName3}),
+				),
+			},
+			// test adding an additional group to an existing resource
+			resource.TestStep{
+				Config: usersAndGroupsConfig + testAccAWSUserGroupMembershipConfigAddOne(membershipName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("aws_iam_user_group_membership.user1_test1", "user", userName1),
+					testAccAWSUserGroupMembershipCheckGroupListForUser(userName1, []string{groupName1, groupName2}, []string{groupName3}),
+				),
+			},
+			// test adding multiple resources for the same user, and resources with the same groups for another user
+			resource.TestStep{
+				Config: usersAndGroupsConfig + testAccAWSUserGroupMembershipConfigAddAll(membershipName),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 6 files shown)


aws/resource_aws_iam_group_membership.go
```diff
@@ -35,6 +36,18 @@ func resourceAwsIamGroupMembership() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+
+			"remove_unmanaged_members": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
bflad: This is a breaking change for the resource as written. It should default to `true` to keep backwards compatibility.

aws/resource_aws_iam_group_membership.go
```diff
@@ -35,6 +36,18 @@ func resourceAwsIamGroupMembership() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+
+			"remove_unmanaged_members": &schema.Schema{
```
bflad: Minor nitpick: As of Go 1.7+, the `&schema.Schema` after the attribute name declarations are unnecessary

aws/resource_aws_iam_group_membership_test.go
```diff
@@ -242,3 +324,14 @@ resource "aws_iam_user" "user" {
 }
 `, groupName, membershipName, userNamePrefix)
 }
+
+func testAccAWSGroupMemberAddUnmanagedUser(groupName string, userName string) func() {
```
bflad: We should probably create a complimentary `testAccAWSGroupMemberRemoveManagedUser`

devonbleak:
Thanks @bflad, this makes sense and I'll get the feedback incorporated.

devonbleak:
@bflad thinking about this more, would it make sense to just create a new `aws_iam_user_group_membership` resource rather than having a single resource that may conflict with itself even with the changes to user vs users?

devonbleak:
@bflad the more I think about having a 1:1 user-group membership resource the more it breaks down.  Consider the case where we have a list of users and a list of groups - there wouldn't be a way to add all the users to all the groups if the lengths of those lists has a common factor.  I'm going to go ahead with creating a new resource that will take a list of users and a single group and will be non-exclusive with group memberships.

bflad:
@devonbleak sorry for my lack of reply earlier. Do you need any help?

Just to confirm the use case you're thinking about, it is something like this?

```hcl
variable "users" {
  default = [
    "a",
    "b",
  ]
}

variable "groups" {
  default = [
    "z",
    "y",
    "x",
  ]
}

resource "aws_iam_user" "example" {
  count = "${length(var.users)}"
  # ... other configuration ...
}

resource "aws_iam_group" "example" {
  count = "${length(var.groups)}"
  # ... other configuration ...
}

resource "aws_iam_user_group_membership" "example" {
  # yikes what would we put here? :)
}
```

What do you think about a `groups` attribute on `aws_iam_user` instead? The above becomes something like:

```hcl
variable "users" {
  default = [
    "a",
    "b",
  ]
}

variable "groups" {
  default = [
    "z",
    "y",
    "x",
  ]
}

resource "aws_iam_user" "example" {
  count = "${length(var.users)}"
  # ... other configuration ...
  # add user to all groups
  groups = ["${aws_iam_group.example.*.id}"]
}

resource "aws_iam_group" "example" {
  count = "${length(var.groups)}"
  # ... other configuration ...
}
```

devonbleak:
@bflad The `aws_iam_user_group_membership` would basically work the same as `aws_iam_group_membership` except it wouldn't touch user memberships that were added outside of it (similar to how the original PR works here with `remove_unmanaged_users = false` but without requiring the extra logic in CustomizeDiff or the sidecar).  It would probably need a better name.

Adding a groups argument to `aws_iam_user` is an interesting solution but I think we'll still be missing a resource that just adds a list of users to a group and keeps them there without doing anything else (or vice versa adding a list of groups to a user).

bflad:
@devonbleak I guess we have come full circle. 😄  Do you think you could implement this PR without the sidecar and defaulting the flag attribute to keep existing behavior (exclusive management)?

devonbleak:
@bflad been thinking about this again the past couple days and thinking about implementing something like an `aws_iam_user_group_memberships` that would do the non-exclusive thing without the sidecar or the CustomizeDiff.  The analogy I have in my head is like `aws_iam_policy_attachment` vs `aws_iam_{group,role,user}_policy_attachment` - the former behaves like `aws_iam_group_membership` while the latter ignores things it doesn't know about.

My original implementation of this basically just ignored things it didn't know about in the `Read` function and worked great except that it took two applies for a change to `remove_unmanaged_users` to actually complete which is why I ended up going the sidecar/CustomizeDiff route.  It looks like we have what we need from the Go SDK in [ListGroupsForUser](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.ListGroupsForUser).

I'll start working on this approach, but let me know if you'd like to see something different.

```hcl
resource "aws_iam_user_group_memberships" "test" {
  user   = "${aws_iam_user.test.name}"
  groups = [
    "${aws_iam_group.test1.name}", 
    "${aws_iam_group.test2.name}"
  ]
}
```

bflad:
I'd drop the `s` pluralization on the resource name, but sounds good.

devonbleak:
@bflad implementation is done on resource/aws_iam_user_group_membership.  I also added a note to aws_iam_group_membership about the conflicts behavior so new users aren't caught by surprise with it.  Let me know if you'd like any additional changes!

aws/resource_aws_iam_user_group_membership_test.go
```diff
@@ -0,0 +1,331 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccAWSUserGroupMembershipDestroy,
+		Steps: []resource.TestStep{
+			// simplest test
+			resource.TestStep{
+				Config: usersAndGroupsConfig + testAccAWSUserGroupMembershipConfigInit(membershipName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("aws_iam_user_group_membership.user1_test1", "user", userName1),
+					testAccAWSUserGroupMembershipCheckGroupListForUser(userName1, []string{groupName1}, []string{groupName2, groupName3}),
```
bflad: It'd probably be good to also test that `userName2` is not attached to anything 👍

bflad:

PR #3371, opened 2018-02-14
resource/aws_instance: Expose reason of 'shutting-down' during creation

radeksimko:
This is to improve the following message:

```
=== RUN   TestAccAWSSSMAssociation_basic
--- FAIL: TestAccAWSSSMAssociation_basic (71.48s)
        
        * aws_instance.foo: Error waiting for instance (i-0d22a4b3d279c4f44) to become ready: unexpected state 'shutting-down', wanted target 'running'. last error: %!s(<nil>)
        
        Error: Check failed: Default error in SSM Association Test
        
        State: <no state>
```

It probably won't be much more helpful for the user in all cases, in our case above it was specifically `Server.InternalError: Internal error on launch` 🤷‍♂️ Either way all we can/should do is to expose whatever the API exposes.


bflad:

PR #3374, opened 2018-02-14
data-source/lb_listener: export rule priorities for lb listener

loivis:
Implementation for #1574 on `lb_listener`

Compared with #3219, this approach provides more control to users in case of registering multiple target groups at the same time to the same listener.

Both PRs will use the same function to extract listener rule priorities. If either get merged, we could remove duplicated function in the other.


loivis:
```
⎇  make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAWSLBListener_'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAWSLBListener_ -timeout 120m
=== RUN   TestAccDataSourceAWSLBListener_basic
--- PASS: TestAccDataSourceAWSLBListener_basic (221.54s)
=== RUN   TestAccDataSourceAWSLBListener_https
--- PASS: TestAccDataSourceAWSLBListener_https (213.18s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	434.778s
```

bflad:
Architectural question: does this support creating two new rules at the same time based on referencing the last/next priority value? Sorry its not immediately clear to me how this handles the default parallelism of Terraform, where the new rules would read the same value and one would fail on creation.

loivis:
The difference between last and next is that last is the highest priority number used and next is the lowest available priority number. For example current priorties in use as `[1, 2, 4, 5]`, then last would be `5` and next would be `3`.

I chose to export both for the sake of different preferences.
If only one rule is to be created, either of them would work. I personally prefer minimum number to utilize all possible priorities.
In case of multiple rules required, `next` would be better choice together with simple math support. When creating rules, priority would be like `${data.aws_lb_listener.https.last_rule_priority + 1}`, `${data.aws_lb_listener.https.last_rule_priority + 7}`, and etc. It also helps if people want to control which rule evaluates first/last.

bflad:
My personal opinion here is that the implementation in #3219 seems more approachable from a user experience standpoint. If you're getting into needing math and choosing between two different attributes to figure out a priority number between multiple services, it sounds like you should just use hard set priority values, locals (e.g. a map), or setup explicit dependencies between the rules to get the automated ordering right (when priority is made optional).

That's just my two cents though. I'm very curious about the use case that would require this level of implementation detail that cannot be solved with the above solutions. 😄

loivis:
#3219 is perfect for my use case. It's the reason why I started there. Implementation here is actually just imaginary scenarios from my mind. I'm not sure how people would prefer for those user cases. Maybe explicit dependency/map is much obvious and clear to control:+1:.

bflad:
@loivis thanks so much for working with me on the other PR to make `priority` optional on the `aws_lb_listener_rule` resource! Since we just merged that in (releasing in v1.10.0 of the AWS provider), let's close this one unless there is some good reason that improvement doesn't fit most use cases. 👍

PR #3426, opened 2018-02-17
resource/aws_instance: Remove extra API call on creation for SGs

### LINKED ISSUE #3399: Superfluous call to ModifyNetworkInterfaceAttribute
### Terraform Version
- TF 0.10.8
- AWS provider v1.6.0

### Affected Resource(s)
- aws_instance

```hcl
  ami               = "${module.ami.ami_id}"
  availability_zone = "us-east-1a"
  instance_type     = "r4.large"
  ebs_optimized = "true"
  iam_instance_profile = "${aws_iam_instance_profile.default.name}"
  vpc_security_group_ids = ["sg-xxx","sg-xyx","sg-xyz"]
  subnet_id = "subnet-xxx"

  root_block_device {
    delete_on_termination   = "true"
    volume_size = "12"
  }
}
```

### Debug Output
2018/02/07 18:40:22 [INFO] Modifying vpc_security_group_ids on Instance “i-xxxxxxxxxx”
2018/02/07 18:40:23 [DEBUG] [aws-sdk-go] DEBUG: Request ec2/ModifyNetworkInterfaceAttribute

### Expected Behavior
I'm not sure why these are getting called. I wouldn't expect these. It feels like a race condition.

### Actual Behavior
Given our ACLs we cannot do anything using the permission ec2/ModifyNetworkInterfaceAttribute and the TF job will fail. However, the instance gets created and works fine. Since the modifyNetworkAttribute action returns 403, I would expect the instance to be improperly configured but it seems to have no affect on the outcome. So why is that call needed? When I run this with volume attachments, ebs_volumes and other configs etc, since the job fails, I can't create all the resources I need.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
- Our AWS security and build pipelines prevent us from doing anything needing this permission ec2/ModifyNetworkInterfaceAttribute.

brettahale:
Stops extra calls to update vpc_security_groups and also a call to modifyNetworkInterface when not needed.

Closes #3399

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -914,7 +914,7 @@ func resourceAwsInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
-	if d.HasChange("vpc_security_group_ids") {
+	if d.HasChange("vpc_security_group_ids") && !d.IsNewResource() {
 		var groups []*string
 		if v := d.Get("vpc_security_group_ids").(*schema.Set); v.Len() > 0 {
 			for _, v := range v.List() {


radeksimko:
Thanks @brettahale for this PR! 👍

brettahale:
Woop woop! Np and thank you :)

bflad:

PR #3427, opened 2018-02-17
documentation for acm alternative names dns validation

craiggenner:
I know a few people I've spoken to have been wondering how to do alternative names dns validation with the new `aws_acm_certificate_validation` resource, it may be obvious to some but it's not to others.

This is a feature that came out in 1.9.0 so if you are happy with this doc improvement it can go out now.


craiggenner:
Any reason this hasn't been merged?

PR #3429, opened 2018-02-18
Fix for "invalid Schema.Elem 1; need *Schema or *Resource"

squidfunk:
I'm currently trying to extract the schema from the AWS provider to build an autocomplete provider for Atom, but Terraform panics at this location: https://github.com/hashicorp/terraform/blob/3253367c64ace91aa0f6c3abaaae66721f9b88a3/helper/schema/core_schema.go#L42-L50

I tracked the error down and found that some element types don't have schema definitions attached to them but reference types directly, like this:

``` golang
Elem: schema.TypeString,
```

This fixes the error and seems consistent with the rest of the library:

``` golang
Elem: &schema.Schema{Type: schema.TypeString},
```

This PR fixes all resources that trigger this error.

### DIFF
--- a/aws/data_source_aws_ecs_container_definition.go
+++ b/aws/data_source_aws_ecs_container_definition.go
@@ -53,12 +53,12 @@ func dataSourceAwsEcsContainerDefinition() *schema.Resource {
 			"docker_labels": {
 				Type:     schema.TypeMap,
 				Computed: true,
-				Elem:     schema.TypeString,
+				Elem:     &schema.Schema{Type: schema.TypeString},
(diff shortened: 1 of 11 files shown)


ewbankkit:
See also https://github.com/terraform-providers/terraform-provider-aws/pull/2914.

squidfunk:
@ewbankkit Great, so this is already a known issue. Note that the linked PR misses the fix for `aws/resource_aws_glue_catalog_database.go` which is included in this one. Any ETA for a fix to land in `master`?

PR #3435, opened 2018-02-19
resource/aws_spot_instance_request: Bump delete timeout to 20mins

radeksimko:
This is mainly to address the following test failure:

```
=== RUN   TestAccAWSSpotInstanceRequest_NetworkInterfaceAttributes
--- FAIL: TestAccAWSSpotInstanceRequest_NetworkInterfaceAttributes (679.58s)
        
        * aws_spot_instance_request.foo (destroy): 1 error(s) occurred:
        
        * aws_spot_instance_request.foo: Error terminating spot instance: Error waiting for instance (i-094860585f9ba7567) to terminate: timeout while waiting for state to become 'terminated' (last state: 'shutting-down', timeout: 10m0s)
```

I was playing with the idea of setting custom, higher timeout inside the test config, but since we get some guarantees that the instance is _actually_ shutting down (i.e. transitioning from `shutting-down` to `terminated`) I can't think of any downside of bumping this for everyone.

I'm generally against cranking timeouts so high if it is error-based retry where the risk of false positive is much higher. This isn't the case though.

### DIFF
--- a/aws/resource_aws_spot_instance_request.go
+++ b/aws/resource_aws_spot_instance_request.go
@@ -21,7 +21,7 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
 
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(10 * time.Minute),
-			Delete: schema.DefaultTimeout(10 * time.Minute),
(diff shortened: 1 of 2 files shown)


radeksimko:
@bflad done, PTAL. 😉

bflad:

PR #3454, opened 2018-02-20
tests: Unify test names

radeksimko:


radeksimko:
Well, I generally opted for the path of least resistance as `AWS` makes the majority 🤷‍♂️ 
I have no strong opinions about it 😃

PR #3472, opened 2018-02-21
Consider replication destination for modifications in S3 bucket

julienduchesne:
Currently, if you modify the destination in a replication_config, it is not listed in a plan and applies do not modify it.


julienduchesne:
Fixed in another PR

PR #3478, opened 2018-02-21
docs/instance: Update docs to reflect API behavior

kmacoskey:
The documentation stated that `delete_on_termination` field for each `network_interface` could be set to `true` or `false`. However the behavior of the API does not allow this, specifically the API will respond with an error if `delete_on_termination` is set to true with a `network_interface_id` being specified:

Example:
```
...
  network_interface {
    network_interface_id  = <interface_id>
    device_index          = 0
    delete_on_termination = true
  }
...
```

Error:
```
* aws_instance.node.1: Error launching source instance: InvalidParameterCombination: A network interface may not specify a network interface ID and delete on termination as true
```

Currently it is documented that `network_interface_id` is a required field, which implies that there is no way to successfully apply with `delete_on_termination` set to true.

### DIFF
--- a/website/docs/r/instance.html.markdown
+++ b/website/docs/r/instance.html.markdown
@@ -173,7 +173,6 @@ Each `network_interface` block supports the following:
 
 * `device_index` - (Required) The integer index of the network interface attachment. Limited by instance type.
 * `network_interface_id` - (Required) The ID of the network interface to attach.
-* `delete_on_termination` - (Optional) Whether or not to delete the network interface on instance termination. Defaults to `false`.
 
 ### Example
 
--- a/website/docs/r/instance.html.markdown
+++ b/website/docs/r/instance.html.markdown
@@ -173,6 +173,7 @@ Each `network_interface` block supports the following:
 
 * `device_index` - (Required) The integer index of the network interface attachment. Limited by instance type.
 * `network_interface_id` - (Required) The ID of the network interface to attach.
+* `delete_on_termination` - (Optional) Whether or not to delete the network interface on instance termination. Defaults to `false`. Currently, the only valid value is `false`, as this is only supported when creating new network interfaces when launching an instance.
 
 ### Example
 


bflad:
@kmacoskey does the combination work if you set `network_interface_id = ""`? While the current resource implementation is less than ideal, the API might ignore an empty value and allow `delete_on_termination = true` in that case.

Regardless, I think we should always document available arguments and instead note any problematic conditions between others where necessary, especially in this case where it might get fixed in the future. `delete_on_termination = false` is currently valid and potentially used so the argument should remain documented.

kmacoskey:
@bflad The API fails with a different error for that combination:
```
* aws_instance.web: Error launching source instance: InvalidParameterValue: Each network interface requires either a subnet or a network interface ID.
```
Is it correct to document an available argument when no combination would allow you to actually include the argument?

bflad:
@kmacoskey the [EC2 InstanceNetworkInterfaceSpecification SDK documentation](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#InstanceNetworkInterfaceSpecification) states:

```
    // If set to true, the interface is deleted when the instance is terminated.
    // You can specify true only if creating a new network interface when launching
    // an instance.
    DeleteOnTermination *bool `locationName:"deleteOnTermination" type:"boolean"`
```

It looks like we just don't currently set or allow setting other InstanceNetworkInterfaceSpecification attributes that would allow that parameter to be set to `true`:

https://github.com/terraform-providers/terraform-provider-aws/blob/80e4ae53941c6fc87239f1d704e2dfe6a4dec3b6/aws/resource_aws_instance.go#L1406-L1416

I think simply noting "Currently, the only valid value is `false`, as this is only supported when creating new network interfaces when launching an instance." is sufficient for now. At some point we may support passing in the other parameters that would allow this to be properly set to `true` like we currently do in `aws_launch_template` and `aws_spot_fleet_request`. Sorry if this seems pedantic but we will likely be automatically generating all attribute/argument documentation from the resources in which this would require some sort of description.

kmacoskey:
That's alright, pedantic is good. I wasn't entirely sure whether removing it in the documentation was the best option. I was just hoping it would bring up the topic.

kmacoskey:
You can set it to any value you want... as long as that value is `false` 😄

Issue #305, opened 2017-06-13
Terraform does not detect S3 bucket policy deletion


hashibot:
_This issue was originally opened by @jeffreyckchau as hashicorp/terraform#9288. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hello,

I am having an issue whereby changes to the policy line on an s3 bucket module are not reflected in terraform plan/apply. 
### Terraform Version

v0.7.3
### Terraform Configuration Files

``` hcl
variable "bucket_name" {
    default = ""
}

variable "logging_bucket_name" {
    default = ""
}

variable "vpc_endpoint_id" {
    default = ""
}

data "template_file" "bucket_policy" {
  template = "${file("${path.module}/bucket_policy.json")}"

  vars {
    bucket_name = "${var.bucket_name}"
    vpc_endpoint_id = "${var.vpc_endpoint_id}"
  }
}

resource "aws_s3_bucket" "bucket" {
    bucket = "${var.bucket_name}"
    acl = "public-read"
    versioning {
        enabled = true
    }
    logging {
       target_bucket = "${var.logging_bucket_name}"
       target_prefix = "${var.bucket_name}/"
    }
}


output "s3-bucket" {
    value = "${aws_s3_bucket.bucket.id}"
}


```

In the above code, the aws_s3_bucket module had a policy line that I have since removed. Upon removal, no change was noticed. 
### Expected Behavior

Terraform should have detected a stripped policy with the S3 bucket and marked for change
### Actual Behavior

No S3 buckets were marked for change

Chhed13:
I can't reproduce the exact behavior, but I found close one and PR it.
Mine reproduction:
```
resource "aws_s3_bucket" "bucket" {
    bucket = "${var.bucket_name}"
}
```
1. `terraform apply`
2. add policy to bucket via UI or CLI
3. `terraform apply` and no diff found.

With this PR it's fixed.

bflad:
Hi folks 👋 

[...]

PR #3479, opened 2018-02-21
resource/aws_s3_bucket policy change not detected

Chhed13:
fix for #305
It fixes case when there is no `policy` block in Terraform described, but policy has changed some other way (UI, CLI, etc.).


grahamlyons:
Ooh, yes please. Anything preventing this from being accepted and merged?

Vermyndax:
Still experiencing this issue... any hope this will get merged?

subratkprusty:
It seems like this PR is not yet merged and the issue still persists.
Terraform never compares the original state of the bucket with the state file, hence it never applies the ACL configuration mentioned in your terraform code.
Badly need this PR to be merged and publish.

bflad:
Hi folks 👋 Apologies for the long silence on this pull request and thank you for submitting this @Chhed13.

This type of change is valid as we should generally prefer always refreshing the Terraform state, but only with the addition of `Computed: true` on the attribute, which effectively negates the original intent mentioned in the PR description.

The Terraform AWS Provider supports two methods for configuring a S3 Bucket policy, the `policy` argument of the `aws_s3_bucket` resource and the separate [`aws_s3_bucket_policy` resource](https://www.terraform.io/docs/providers/aws/r/s3_bucket_policy.html). Merging this as-is will prevent operators from effectively using the `aws_s3_bucket_policy` resource without also configuring this `aws_s3_bucket` argument. This requirement, while undesirable, may also not be possible if the policy is referencing attributes from the `aws_s3_bucket` resource.

To perform drift detection with this argument and ensure no policy has been applied, the solution is to set the policy to an empty string, e.g.

```hcl
resource "aws_s3_bucket" "example" {
  # ... other configuration ...
  policy = ""
}
```

Otherwise this argument or the `aws_s3_bucket_policy` resource should be configured with the expected policy.

As for #305 (removing the `policy` argument completely and expecting the policy to be deleted), this is only supported by the same empty string usage noted above or the usage of `aws_s3_bucket_policy` resource. The existing and current resource logic or the addition of `Computed: true` on the policy would still prevent the removal of the argument completely from ever removing the policy as expected due to how Terraform works. In the future, we may deprecate and remove these policy arguments from existing resources and instead prefer the separate resources which can better model intentions.

Since this pull request will have breaking changes for supported Terraform configurations, I'm going to close this for now, but if you wish to either improve the documentation around this or also include these changes with `Computed: true` enabled for the attribute, please submit a new pull request and we'll take a fresh look. Thanks.

Chhed13:
Hey,
sad resume after two years of waiting. That's easy to improve, but not in my plans right now. If any body wanna take a flag, feel free to take this diff as a starting point

PR #3480, opened 2018-02-21
Fix aws_appautoscaling_policy scale in regression

nathanielks:
I hit the regression mentioned in https://github.com/terraform-providers/terraform-provider-aws/issues/2063 and have applied the fix @toddlucas suggested in https://github.com/terraform-providers/terraform-provider-aws/issues/2063#issuecomment-358836553. 

A blocker I'm experiencing is the acceptance tests won't pass because they can't find nested resources with random ids and I am unsure of how to generate those deterministically. How does one go about doing that?


nathanielks:
I've tested the aws provider outside of acceptance tests and it's working as desired. Scaling in using negative values requires setting one of the steps' upper bound to 0, but aside from that it's working as expected. I believe this to be an issue AWS side than a terraform issue, but please correct me if I am wrong!

nathanielks:
@bflad @radeksimko any idea when this can get a review? Also curious if there's any documentation on how PR's are prioritized. I'm seeing lots of new resource PR's and others getting merged, while there are other PR's fixing (critical) bugs or adding functionality that sit dormant for weeks upon weeks. Not criticizing, just curious the thoughts behind it! I know y'all are having to balance a lot on this project.

solomonjames:
@bflad I see you are a big contributor on this repo.  Can we please merge this PR or somehow resolve the issue this is trying to solve?

solomonjames:
@nathanielks the fork/branch worked perfectly.  thank you!

bflad:
Sorry this bugfix got lost in the shuffle 😅 There's two things I'm curious about here:
* Does this just "work" if you take an existing Terraform configuration and its state, switching to the newer version? I'm not familiar if changing the schema type from `TypeFloat` to `TypeString` requires any sort of state migration to properly upgrade existing configurations to not show as a difference.
* The acceptance test added here does seem to be applying an update step to ensure we do not hit this error message mentioned in the original issue: `aws_appautoscaling_policy.reduce_capacity: Failed to update scaling policy: ValidationException: Both lower and upper bounds of a step adjustment cannot be left unspecified`

Can someone please verify the above?

solomonjames:
@nathanielks Likely has more context.

@bflad I was able to use this forked branch, compile and use the plugin with an existing config and state file and it did indeed just work as expected.

nathanielks:
> Does this just "work" if you take an existing Terraform configuration and its state,

Aye, it sure does! Using the [config in the original issue](https://github.com/terraform-providers/terraform-provider-aws/issues/2063#issue-268771635) (and modified for ECS) I tested this by `terraform init`  normally and applying. The change failed as expected. I then `rm -rf .terraform`, compiled this copy of the aws provider and re-applied. The change was able to be applied successfully.

> The acceptance test added here does seem to be applying an update step to ensure we do not hit this error message mentioned in the original issue

That's also correct. The test isn't passing currently, however, because I'm unsure of how to generate the ID of the attribute, so there's a mismatch. Could you fill me in on how to do that, @bflad?

djtarazona:
Any word on getting this fix merged?

@nathanielks Do you mind rebasing and fixing the latest conflicts? Then @bflad can you please take another look?

nathanielks:
@djtarazona can do. 

@bflad hopefully this will make it through all the noise you get! I've got tests written, but I can't get them to pass because I'm not sure how to set their attribute ids. Could you provide some help?

rv-vmalhotra:
@bflad I would like to lend support for this PR as a critical component of app auto-scaling is stuck for me. I am not able to scale-in ecs service based on scale-in policies because of this bug. 

This solution prima-face seems to work and it would be great if it can be prioritized.

nathanielks:
@rv-vhalhotra I'm the author of the PR. All that I'm waiting for is an answer to how to determine what ID's Terraform assigns to attributes so that I can make the tests pass. They _should_ right now but don't because the attribute ID's don't line up.

nathanielks:
@bflad 	pinging! I've got tests written, but I can't get them to pass because I'm not sure how to set their attribute ids. Could you provide some help? See the failng test below:

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSAppautoScalingPolicy_scaleOutAndIn'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAppautoScalingPolicy_scaleOutAndIn -timeout 120m
k=== RUN   TestAccAWSAppautoScalingPolicy_scaleOutAndIn
--- FAIL: TestAccAWSAppautoScalingPolicy_scaleOutAndIn (48.72s)
	testing.go:527: Step 0 error: Check failed: Check 5/36 error: aws_appautoscaling_policy.foobar_out: Attribute 'step_scaling_policy_configuration.0.step_adjustment.2252990027.metric_interval_upper_bound' not found
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	48.760s
make: *** [testacc] Error 1
```

It only references the attribute I'm trying to test and I'm unsure of how to find the resources actually created.

bflad:
Hey @nathanielks 😅  Indeed I do get a large chunk of notifications so apologies for the delayed response. Let's see if I can help hopefully!

When `Type: schema.TypeSet` attributes are "stringified" in the Terraform state, its from the hash value of the set (via either the  `Set:` function on the attribute or a default hashing function if `Set:` is not specified for the attribute).

Just to be clear, in this example:

```
step_scaling_policy_configuration.0.step_adjustment.2252990027.metric_interval_upper_bound
```

The `2252990027` value. 😄 

Actually getting that value is less than ideal and has some caveats on what attributes exist within the `TypeSet` and whether they are stable. For example, if the `TypeSet` contains an attribute that has a value that is something like a randomly generated identifier from another resource (e.g. `"${aws_security_group.example.id}"`), you'll never be able to get a stable hash value for the `TypeSet`. 

If the hash value should be stable and does make sense to use in the acceptance testing, the way we (potentially just me!) currently do it is actually pretty manual by getting it from the debug logging of the acceptance test. 😖 

Running something like this will generate an `aws/log.txt` debug log file:

```
$ rm -f aws/log.txt; TF_LOG_PATH=log.txt TF_LOG=debug make testacc TEST=./aws TESTARGS='-run=TestAccAWSAppautoScalingPolicy_scaleOutAndIn'
```

Inside the log, you can search for `DESTROY:` which will contain the Terraform state with all the resources and their attribute values for destruction. You should see the attribute(s) you're looking for there.

The other option in these cases is to take the `applicationautoscaling.ScalingPolicy` object returned by `testAccCheckAWSAppautoscalingPolicyExists()` and query for information inside of it. A contrived example of something like this for another resource is `testAccCheckAWSRolePermissionsBoundary()`.

At some point we may get rid of or drastically reduce the usage of `TypeSet` attributes in Terraform as their usage and testing is very complicated as seen here.

Does that help answer what you were looking for? I'll try to make sure I stay better on top of this PR.

nathanielks:
@bflad that was exactly what I was looking for, thank you!! Acceptance test is now passing 😄 	
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSAppautoScalingPolicy_scaleOutAndIn -timeout 120m
=== RUN   TestAccAWSAppautoScalingPolicy_scaleOutAndIn
--- PASS: TestAccAWSAppautoScalingPolicy_scaleOutAndIn (43.16s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	43.208s
```

nathanielks:
Wooooohoo! Thanks for circling back around to this, @bflad. Glad to see this getting merged 😄

bflad:

Issue #2946, opened 2018-01-11
Root block device not terminated by default


hashibot:
_This issue was originally opened by @BeatMaster as hashicorp/terraform#17084. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
```
0.10.8
```

### Terraform Configuration Files
```hcl
resource "aws_instance" {
  ami                    = "..."
  instance_type          = "..."
  vpc_security_group_ids = "..."
  subnet_id              = "..."
  key_name               = "..."
  user_data              = "..."
  iam_instance_profile   = "..."
}
```

### Expected Behavior
When an aws_instance is destroyed, the root block device should be terminated, by default when the `root_block_termination = true` **has** been omitted .
### Actual Behavior
The root block device is orphaned after destroying the aws_instance.

### Steps to Reproduce
1. `terraform init`
2. `terraform apply`
3. `terraform destroy`
4. EBS still remains but orphaned.

### Additional Context
If you specify the following (what is considered the default), the EBS is deleted as expected.
```
root_block_device = {
    delete_on_termination = true
  }
```

### Reference
1. https://www.terraform.io/docs/providers/aws/r/instance.html#delete_on_termination

dmrzzz:
For me (using Terraform v0.11.8, provider.aws v1.36.0), the empirical default result with no root_block_device stanza present is actuall AMI-dependent.

```hcl
provider "aws" {
  region = "us-east-2"
}

# amzn-ami-hvm-2018.03.0.20180811-x86_64-gp2
resource "aws_instance" "amazon" {
  ami           = "ami-0b59bfac6be064b78"
  instance_type = "t2.nano"
}

# CentOS Linux 7 x86_64 HVM EBS ENA 1804_1
resource "aws_instance" "centos" {
  ami           = "ami-08b08d6d"
  instance_type = "t2.nano"
}

# another CentOS with a root_block_device stanza
resource "aws_instance" "another_centos" {
  ami           = "ami-08b08d6d"
  instance_type = "t2.nano"
  root_block_device {
    volume_size = 16
  }
}
```

After a terraform apply,

```bash
$ terraform state show aws_instance.amazon | grep root_block_device
root_block_device.#                       = 1
root_block_device.0.delete_on_termination = true
root_block_device.0.iops                  = 100
[...]

PR #3483, opened 2018-02-22
Make delete on termination for root disk to be default

ageekymonk:
Fix for #2946.  By default for root disk delete on termination is set to true by default.

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -1411,6 +1411,8 @@ func readBlockDeviceMappingsFromConfig(
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


bflad:
Hi @ageekymonk 👋 Thanks for submitting this and apologies for the really long delay in review.

I'm fairly certain this behavior is AMI dependent. You can create AMIs that have it enabled or disabled and EC2 will respect that behavior when launching them. This is also a change to default behavior that risks unexpected data loss for operators that might not notice the behavior change (even in a major provider version update).

Given the risks, I think we are going to hold off on this type of change. We can discuss this further in an issue if you would like. Thanks again for writing this up.

PR #3485, opened 2018-02-22
Add wait_for_steady_state attribute to aws_ecs_service

### LINKED ISSUE #3107: Wait until ECS deployment is complete?
Hello,

I have Terraform scripts which setup the following resources:

- Autoscaling Group
- Launch Configuration
- Application Load Balancer
- ECS cluster
- Service
- Task definition
- ECR repository

When I run the script, those resources get provisioned accordingly. However, when I want to update/deploy a service, is there a way to tell Terraform to wait until a health check is okay? (like in Elastic Beanstalk Deployments)

Right now, it will complete the `terraform apply` command, even if the service for some reason is unable to spawn new tasks into the cluster.

mmdriley:
This allows for cases where the old and new service tasks have different resource dependencies -- for example, an old S3 bucket or database -- and the old resources should only be deleted once all old service tasks that depended on it are gone.

In the case that I hit, our service relied on an auxiliary `aws_ecs_task_definition` that became inactive while there were still service tasks that expected to use it with `RunTask`. Now we can set `wait_for_steady_state` and know the old task definition will stay alive as long as it's needed.

This also allows authors to make the success of their deployment conditional on the success of the service update.

To the extent possible, I modeled this after the logic of `aws ecs wait services-stable`. That includes the definition of "stable" as well as the timeout (10m). 

Fixes #3107

### DIFF
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -547,9 +547,6 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 		var err error
 		out, err = conn.DescribeServices(&input)
 		if err != nil {
-			if d.IsNewResource() && isAWSErr(err, ecs.ErrCodeServiceNotFoundException, "") {
-				return resource.RetryableError(err)
-			}
 			return resource.NonRetryableError(err)
 		}
 
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -353,6 +353,12 @@ func resourceAwsEcsService() *schema.Resource {
 				},
 			},
 			"tags": tagsSchema(),
+
+			"wait_for_steady_state": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 	}
 }
@@ -528,6 +534,12 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 	log.Printf("[DEBUG] ECS service created: %s", aws.StringValue(service.ServiceArn))
 	d.SetId(aws.StringValue(service.ServiceArn))
 
+	if d.Get("wait_for_steady_state").(bool) {
+		if err := waitForSteadyState(conn, d); err != nil {
+			return err
+		}
+	}
+
 	return resourceAwsEcsServiceRead(d, meta)
 }
 
@@ -547,6 +559,9 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 		var err error
 		out, err = conn.DescribeServices(&input)
 		if err != nil {
+			if d.IsNewResource() && isAWSErr(err, ecs.ErrCodeServiceNotFoundException, "") {
+				return resource.RetryableError(err)
+			}
 			return resource.NonRetryableError(err)
 		}
 
@@ -1032,6 +1047,12 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	if d.Get("wait_for_steady_state").(bool) {
+		if err := waitForSteadyState(conn, d); err != nil {
+			return err
+		}
+	}
+
 	if d.HasChange("tags") {
 		o, n := d.GetChange("tags")
 
@@ -1160,4 +1181,18 @@ func buildFamilyAndRevisionFromARN(arn string) string {
 // arn:aws:ecs:us-west-2:0123456789:cluster/radek-cluster
 func getNameFromARN(arn string) string {
 	return strings.Split(arn, "/")[1]
+}
+
+func waitForSteadyState(conn *ecs.ECS, d *schema.ResourceData) error {
+	input := &ecs.DescribeServicesInput{
+		Services: aws.StringSlice([]string{d.Id()}),
+	}
+	if v, ok := d.GetOk("cluster"); ok {
+		input.Cluster = aws.String(v.(string))
+	}
+
+	if err := conn.WaitUntilServicesStable(input); err != nil {
+		return fmt.Errorf("error waiting for service (%s) to reach a steady state: %w", d.Id(), err)
+	}
+	return nil
 }
--- a/aws/resource_aws_ecs_service_test.go
+++ b/aws/resource_aws_ecs_service_test.go
@@ -894,6 +894,78 @@ func TestAccAWSEcsService_withLaunchTypeFargateAndPlatformVersion(t *testing.T) 
 	})
 }
 
+func TestAccAWSEcsService_withLaunchTypeFargateAndWaitForSteadyState(t *testing.T) {
+	var service ecs.Service
+	resourceName := "aws_ecs_service.test"
+	rString := acctest.RandString(8)
+	rName := fmt.Sprintf("tf-acc-svc-w-ltf-ss-%s", rString)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEcsServiceDestroy,
+		Steps: []resource.TestStep{
+			{
+				// Wait for the ECS Cluster to reach a steady state w/specified count
+				Config: testAccAWSEcsServiceWithLaunchTypeFargateAndWait(rName, 1, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEcsServiceExists(resourceName, &service),
+					resource.TestCheckResourceAttr(resourceName, "desired_count", "1"),
+					resource.TestCheckResourceAttr(resourceName, "wait_for_steady_state", "true"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportStateId:     fmt.Sprintf("%s/%s", rName, rName),
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyState(t *testing.T) {
+	var service ecs.Service
+	resourceName := "aws_ecs_service.test"
+	rString := acctest.RandString(8)
+
+	rName := fmt.Sprintf("tf-acc-svc-w-ltf-ss-%s", rString)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEcsServiceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEcsServiceWithLaunchTypeFargateWithoutWait(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEcsServiceExists(resourceName, &service),
+					resource.TestCheckResourceAttr(resourceName, "desired_count", "1"),
+					resource.TestCheckResourceAttr(resourceName, "wait_for_steady_state", "false"),
+				),
+			},
+			{
+				// Modify desired count and wait for the ECS Cluster to reach steady state
+				Config: testAccAWSEcsServiceWithLaunchTypeFargateAndWait(rName, 2, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEcsServiceExists(resourceName, &service),
+					resource.TestCheckResourceAttr(resourceName, "desired_count", "2"),
+					resource.TestCheckResourceAttr(resourceName, "wait_for_steady_state", "true"),
+				),
+			},
+			{
+				// Modify desired count without wait
+				Config: testAccAWSEcsServiceWithLaunchTypeFargateAndWait(rName, 1, false),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEcsServiceExists(resourceName, &service),
+					resource.TestCheckResourceAttr(resourceName, "desired_count", "1"),
+					resource.TestCheckResourceAttr(resourceName, "wait_for_steady_state", "false"),
+				),
+			},
+		},
+	})
+}
+
 func TestAccAWSEcsService_withLaunchTypeEC2AndNetworkConfiguration(t *testing.T) {
 	var service ecs.Service
 	rString := acctest.RandString(8)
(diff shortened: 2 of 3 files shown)


aws/resource_aws_ecs_service.go
```diff
@@ -568,6 +577,22 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 		return err
 	}
 
+	if d.Get("wait_for_steady_state").(bool) {
+		log.Println("[INFO] Waiting for service to reach a steady state")
+
+		steadyStateConf := &resource.StateChangeConf{
+			Pending:    []string{"false"},
+			Target:     []string{"true"},
```
mmdriley: Returning a `bool` state as a string feels a bit weird, but there's precedent:

https://github.com/terraform-providers/terraform-provider-aws/blob/490897d8f64608a336ab43564c5dfe681509b213/aws/resource_aws_network_interface.go#L205-L208

I considered returning the service state or "STABLE", but then we're relying on ECS never introducing another state.

website/docs/r/ecs_service.html.markdown
```diff
@@ -63,6 +63,7 @@ into consideration during task placement. The maximum number of
 * `placement_constraints` - (Optional) rules that are taken into consideration during task placement. Maximum number of
 `placement_constraints` is `10`. Defined below.
 * `network_configuration` - (Optional) The network configuration for the service. This parameter is required for task definitions that use the awsvpc network mode to receive their own Elastic Network Interface, and it is not supported for other network modes.
+* `wait_for_steady_state` - (Optional) If `true`, Terraform will wait for the service to reach a steady state (like [`aws ecs wait services-stable`](https://docs.aws.amazon.com/cli/latest/reference/ecs/wait/services-stable.html)) before continuing. Default `false`.
```
mmdriley: I wanted to write an acceptance test for when this is `true`, but I couldn't come up with a reliable, repeatable way to test we *weren't* hitting a race condition.

This would also be the first test case that required container instances (or which used Fargate and was locked to `us-east-1`.)

I have verified manually that this waits for a new deployment to finish.

aws/resource_aws_ecs_service.go
```diff
@@ -709,3 +734,44 @@ func validateAwsEcsServiceHealthCheckGracePeriodSeconds(v interface{}, k string)
 	}
 	return
 }
+
+// Returns a StateRefreshFunc for a given service. That function will return "true" if the service is in a
+// steady state, "false" if the service is running but not in a steady state, and an error otherwise.
+func resourceAwsEcsServiceIsSteadyStateFunc(d *schema.ResourceData, meta interface{}) resource.StateRefreshFunc {
```
tomelliff: Any reason you're not just using https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#ECS.WaitUntilServicesStable here?

website/docs/r/ecs_service.html.markdown
tomelliff: What race conditions were you thinking of?

I would have thought that you could simply create an ECS service in one step (using your `wait_for_steady_state` to make sure it doesn't complete until there is a task running somewhere) and then another step that increases the `desired_count` and then another step that changes the task definition by setting an environment variable or something.

Creating container instances shouldn't be difficult, I'd just use the latest ECS optimised AMI and set the user data so that it joins the ECS cluster you create for the acceptance test.

aws/resource_aws_ecs_service.go
tomelliff: You probably also want a configurable timeout period here. Although I can't see it in the Go SDK the botocore link lower down shows a 600s total wait time (15s * 40 attempts) which seems like a reasonable default.

mmdriley: > Any reason you're not just using https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#ECS.WaitUntilServicesStable here?

It seemed more in keeping with the Terraform code to use `StateChangeConf`, especially since that code has more elaborate and configurable timeout logic than the AWS SDK -- which will help implementing configurable timeouts, as you describe.

```bash
$ grep -R . -e "\.WaitUntil" --exclude './vendor/*'
./aws/resource_aws_ebs_snapshot.go:	err := conn.WaitUntilSnapshotCompleted(req)
./aws/resource_aws_iam_instance_profile.go:	err = iamconn.WaitUntilInstanceProfileExists(waiterRequest)
```

> Although I can't see it in the Go SDK the botocore link lower down shows a 600s total wait time (15s * 40 attempts) which seems like a reasonable default.

That's exactly where I got `10 * time.Minute` from above. :)

mmdriley:
Despite claiming to wait on creates and updates, my implementation previously only waited on updates. Sorry about that! I've updated the PR.

mmdriley:
We've seen a problem on create where the ECS service seems to have disappeared due to eventual consistency, similar to the problem addressed in https://github.com/terraform-providers/terraform-provider-aws/commit/7551f9293f5f92b57e506cefe7ee8968347feb72. Change coming.

mmdriley:
Now updated with code to handle stale reads. Ready for review.

bitbrain:
Any news on this? Having `wait_for_steady_state` in Terraform would be helpful, especially for deployment pipelines.

tomelliff:
@bgiorgini yep, exactly like that.

As a workaround for now I'm using a null resource to shell out to the ECS waiter:
```hcl
resource "null_resource" "wait_for_service_deploy" {
  triggers {
    task_definition = "${aws_ecs_service.web_service.task_definition}"
  }

  provisioner "local-exec" {
    command = "aws ecs wait services-stable --services ${aws_ecs_service.web_service.name} --cluster ${var.cluster_name} --region ${data.aws_region.current.name}"
  }
}
```

website/docs/r/ecs_service.html.markdown
borgstrom: If we create an ALB and Target Group then the acceptance test can be something as simple as launching `busybox` `sleep 600` in that service.

Since it will never pass health checks the service will never become stable.

PaulAtkins:
Thanks for the workaround @tomelliff!

We found when testing that the `ecs wait services-stable` command works slightly differently to the ECS `service stable` event message, and [doesn't wait for new tasks to become healthy in a ALB](https://github.com/boto/botocore/issues/1021). 

So we added a second `null_resource` using `elbv2 wait target-in-service` with a trigger on the first one & the task definition. 

The triggers cause Terraform to only run `elbv2 wait target-in-service` after `service-stable` completes, and prevents the task definition being destroyed until the new instances are healthy. (one downside is that an `apply` will fail if existing Targets aren't already healthy)

```
resource "null_resource" "wait_for_service_deploy" {
  triggers {
    task_definition = "${aws_ecs_service.web_service.task_definition}"
  }

  provisioner "local-exec" {
    command = "aws ecs wait services-stable --services ${aws_ecs_service.web_service.name} --cluster ${var.cluster_name} --region ${data.aws_region.current.name}"
  }
}


resource "null_resource" "wait_for_service_alb_healthy" {
  triggers {
    task_definition = "${aws_ecs_service.web_service.task_definition}"
    service_stable = "${null_resource.wait_for_service_deploy.id}"
  }

  provisioner "local-exec" {
    command = "aws elbv2 wait target-in-service --target-group-arn ${aws_alb_target_group.web_service_tg.arn} --region ${data.aws_region.current.name}"
  }
}
```

tomelliff:
In my experience it does seem to catch when a service is fundamentally unhealthy (container is missing something so it fails to properly start) and this fails our deployment CI jobs when it happens.

What is unfortunate is that if another deployment happens that still doesn't fix things then it seems to return as stable despite the new task definition still being broken. This seems to happen the majority of the time with back to back failing deploys and it will also usually stop the original healthy tasks then. Oddly this isn't 100% reproducible and sometimes ECS keeps the original healthy tasks running until a new task definition is genuinely healthy.

I'm not sure what's exactly happening in these situations but they're rare enough for us that I've not really looked into it other than a cursory glance when alerting points out 0 tasks are running. And it's not been an issue for us because deploy failures always get flushed out in staging because we aren't yet (maybe never) just going straight to production.

derickson82:
I've been looking for this exact feature, and would love to use this instead of the work around. Any clue when this will be ready?

I'd be happy to help. Let me know if there is anything I can do to help finish this.

mmdriley:
If this gets attention from the Terraform folks I can probably try to help shepherd it through, but otherwise I've pretty much forgotten about it.

To be practically deployable, this probably needs to allow configurable timeouts to accommodate large service deployments that will take a long time to update.

tomelliff:
@mmdriley first passes on a feature normally add the minimal thing to get it working and then if people have a requirement for something else (eg for needing to wait more than 10 minutes for an insanely slow starting service to start) then it can be made configurable in a follow up PR.

@bflad any chance this can get a review? Got a few people interested in it and we've had a couple of issues raised looking for this functionality.

mmdriley:
Rebased and ready for review.

tomelliff:
As mentioned before we probably do need to think about how to test this
before it can be merged but I've not had a chance to really think about it
beyond that initial discussion.

If I get some time to sit down with it I'll try to think of some useful
tests. If anyone else wants to weigh in on the test discussion or
suggest/push some tests then that would also help.

jones2026:
Any more progress on this PR? Is there anything I can do to help get this added?

barrytam20:
@ewbankkit ill close mine in favor of this one. I was not aware this PR existed, and looks like this one is further along

bassrock:
What is needed to get this pr through? Happy to help rebase, refactor anything that is needed to get this functionality through.

dekimsey:
Is there any significant difference between this being implemented in the resource itself, or defining a separate resource ala `aws_acm_certificate_validation` to enforce the wait for stable condition?

jones2026:
@dekimsey besides having to add another resource I don't see any major difference. Are you thinking one way is better or worse than another?

dekimsey:
@jones2026 I'm not sure TBH, simply an observation. It's just a methodology that's seen use in the aws provider already and I'm wondering if it offers any advantage. Other than _perhaps_ making it possible to create other validation methods in the future. Frankly, I'll be pleased as peach just to have wait_for_steady_state on my ecs serivce as is. That is going to be _awesome_.

I do know we've seen issues where for instance ECS service itself sees the app is healthy. But the LB registration (for instance the LB lacks a SG rule to hit the healthcheck port) doesn't and starts recycling the service usually at a very large delay. I don't recall if this shows up as a stable deployment or not since I haven't had any real visibility into it. But if memory serves (because we couldn't figure out what was going wrong), it is considered stable from ECS's perspective.

tomelliff:
The `aws_acm_certificate_validation` resource exists mostly because it's an extra API call that does something in the AWS account. This is just running a waiter equivalent to waiting for an EC2 instance to reach the running status or the autoscaling group resource waiting for instances to reach a minimum number of healthy instances.

As such I don't think it's appropriate to create an entirely new resource for this. As far as I know there's nothing contentious about this change, it just hasn't had a review yet and has fallen off the radar.

There is an open question about how this is tested (as mentioned in https://github.com/terraform-providers/terraform-provider-aws/pull/3485/files#r170058398) but I've not seen other waiters have tests added for them as far as I can remember. There's also a potential small refactor needed because this would be expected to live in `aws/internal/service/ecs/waiter/waiter.go` nowadays (although it didn't exist back when this PR was first raised).

jrstarke:
@maryelizbeth There appears to be a bunch of interest in this. Any chance we could get someone to have a look at it?

dekimsey:
@PaulAtkins It should be noted, the shell-out for `aws ecs wait services-stable` and `aws elbv2 wait target-in-service` won't work for anyone using the TF remote backend with remote execution since the environment doesn't actually have the `aws` command-line toolkit installed. And getting even a subset of that API running in a simple python script is not possible. Without proper support, this is actually really difficult to work-around.

dekimsey:
@tomelliff As I understand it, it looks like there are two separate waits that need to happen. One for the ECS services-stable itself and one for the ELB/ALB target-in-service. (Assuming of course this service is associated with a load-balancer). How would this PR be extended to implement or support the latter? As far as I can tell, it doesn't look like it is implemented as is.

My only interest here is that we can easy get support for both waits when appropriate for the aws_ecs_service and report accurately on it.

maryelizbeth:
> @maryelizbeth There appears to be a bunch of interest in this. Any chance we could get someone to have a look at it?

Thanks for the ping on this, we're aware of the issue and keeping an eye on it. Right now, the team is paused on releases until we wrap up [3.0.0 ](https://github.com/terraform-providers/terraform-provider-aws/milestone/70) which is due at the end of the month. We'll revisit this and other PRs for prioritization once we get that release out. 

Feel free to reach out to me directly, if y'all need anything in the meantime!

mmdriley:
@maryelizbeth can you provide a date by which we should expect the next update from a maintainer -- either "we will start reviewing this now and we will respond promptly until it is ready to be merged" or "we are not ready to review this yet, but we will be back with an update by <date>"?

maryelizbeth:
Thanks for following up @mmdriley. At this time we cannot provide a date or estimate for when this will be merged. However, we will return to this issue after we wrap up the v3.0 release with an update. As a reminder v3.0 is due out on Jul 31.

mmdriley:
@maryelizbeth thanks for the reply. I definitely understand you can't commit to a date to merge this, but I was asking a slightly different question -- when can we next expect engagement, of any kind, from a maintainer? That way we know we don't need to ping the issue for updates because it will be seen in the natural course.

dekimsey:
@maryelizbeth How are we looking for getting support for this functionality? I know my team is really looking forward to having this feature to improve our ECS deployments. We are deploying many ECS services via terraform and are running in the terraform remote cloud to do it. Unfortunatley, we simply cannot shell out to `aws ecs wait services-stable` on that platform means I have no alternative but to wait for this feature.

Please let us know if there's anything we can do to get movement on this.

mmdriley:
For my part, I'm setting a reminder to close this PR on October 1 unless I've seen significant engagement from the maintainers. There's no purpose in letting people believe this implementation will be merged if the PR is going to languish for *years*.

mmdriley:
Merge conflicts resolved. Looking forward to seeing a first round of review this week.

aws/resource_aws_ecs_service.go
```diff
@@ -1043,6 +1053,12 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	if d.Get("wait_for_steady_state").(bool) {
+		if err := resourceAwsEcsWaitForServiceSteadyState(d, meta); err != nil {
```
anGie44, line 1057: This method definitely follows the construct used across the provider when implementing custom `StateChangeConf` functionality, but in this case since the AWS Go SDK provides this method with `ECS.WaitUntilServicesStable` and `ECS.WaitUntilServicesStableWithContext` to add configurable waiter options, I would also suggest adopting the SDK's method in place of this custom logic

JaniszM:
Is there any change to get this merged? :)

mmdriley:
The PR has been open for two and a half years. The last two weeks or so
have been a bit busy and I haven’t had a chance to address the most recent
feedback.

Would *love* to hand this PR over to someone who e.g. still uses AWS
professionally. Otherwise I will hope to come back to this before the end
of the month.

tomelliff:
@mmdriley If you don't have any more use for this functionality then I'd be happy to pick it up from where you've left it? We'd still like to use it and are still shelling out to the AWS CLI to do the wait for us so that would tidy that bit up for us. From a quick grep it might also be the last thing we have for Terraform in CI that needs the AWS CLI so we'd be able to drop the AWS CLI and the Python interpreter out of our Terraform Docker image too :smiley_cat:

Of course if you'd prefer to get this over the line yourself after such a long wait then we can definitely wait a few more weeks to see this land :smile_cat:

anGie44:
Hi @tomelliff, I think given @mmdriley's last response, it shouldn't be a problem to pick-up where was left off if you are still interested in doing so at this time 👍

aws/resource_aws_ecs_service.go
```diff
@@ -547,9 +560,6 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 		var err error
 		out, err = conn.DescribeServices(&input)
 		if err != nil {
-			if d.IsNewResource() && isAWSErr(err, ecs.ErrCodeServiceNotFoundException, "") {
-				return resource.RetryableError(err)
-			}
```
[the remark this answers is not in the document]
mmdriley: As mentioned in the commit message for https://github.com/terraform-providers/terraform-provider-aws/pull/3485/commits/97140b338e65f69516fa519a5f0a60231c4a1777, this code wasn't actually handling errors from `DescribeServices`.

aws/resource_aws_ecs_service.go
mmdriley: Ah, fair enough -- I don't think I noticed the `WithContext` variants. Agree this seems like the right approach.

mmdriley:
Yes, @tomelliff or anyone else sufficiently motivated can take this on with my blessing. Thanks!

aws/resource_aws_ecs_service_test.go
```diff
@@ -1350,6 +1380,96 @@ resource "aws_ecs_service" "mongo" {
...
+
+  tags = [
+    {
+      key                 = "foo"
+      value               = "bar"
+      propagate_at_launch = true
+    },
+  ]
+}
+
```
anGie44, line 1429: added b/c `DescribeServices` was throwing `was unable to place a task because no container instance met all of its requirements. Reason: No Container Instances were found in your capacity provider` error

aws/resource_aws_ecs_service_test.go
```diff
@@ -1350,6 +1422,233 @@ resource "aws_ecs_service" "mongo" {
...
+  name            = %[1]q
+  cluster         = aws_ecs_cluster.test.id
+  task_definition = aws_ecs_task_definition.mongo.arn
+  desired_count   = %d
+  launch_type     = "FARGATE"
+
+  network_configuration {
+    security_groups  = [aws_security_group.test.id]
+    subnets          = aws_subnet.test[*].id
+    assign_public_ip = true
```
anGie44, line 1645: required for the task to reach `RUNNING` state -- else task throws error `CannotPullContainerError: Error response from daemon:Get https://registry-name/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

aws/resource_aws_ecs_service_test.go
```diff
@@ -1350,6 +1422,233 @@ resource "aws_ecs_service" "mongo" {
...
+   from_port = 0
+   to_port   = 0
+   protocol  = "-1"
+  
+   cidr_blocks = [
+     "0.0.0.0/0",
+   ]
+  }
+}
+
```
anGie44, line 1609: required for task to run else returns error in comment below

aws/resource_aws_ecs_service_test.go
```diff
@@ -1350,6 +1422,235 @@ resource "aws_ecs_service" "mongo" {
...
+    cidr_block = "0.0.0.0/0"
+    gateway_id = aws_internet_gateway.test.id
+  }
+}
+
+resource "aws_route_table_association" "test" {
+  count          = 2
+  subnet_id      = element(aws_subnet.test.*.id, count.index)
+  route_table_id = aws_route_table.test.id
+}
```
anGie44, line 1585: required to run task else receives the error below and task reaches `STOPPED` state

anGie44:
Output of added tests:
```
--- PASS: TestAccAWSEcsService_withLaunchTypeFargateAndWaitForSteadyState (188.43s)
--- PASS: TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyState (212.50s)
```

bflad (approved):
Looks good! 🚀

PR #3487, opened 2018-02-22
doc/resource/waf_rule: correct predicate  type value

### LINKED ISSUE #3486: Documentation bug: aws_waf_rule contains incorrect predicates
### Affected Resource(s)
- aws_waf_rule

The documentation states for predicates:
- predicates \- (Optional) One of ByteMatchSet, IPSet, SizeConstraintSet, SqlInjectionMatchSet, or XssMatchSet objects to include in a rule.

But when terraform plan is run the error:
```
Error: aws_waf_rule.waf_sqli_rule: "predicates.0.type" must be one of IPMatch | ByteMatch | SqlInjectionMatch | SizeConstraint | XssMatch
```

The documentation should reflect what is expected by the resource.

loivis:
Fix #3486 

The initial intent is only to update document to remove confusion. 

```
Type
The type of predicate in a Rule, such as ByteMatchSet or IPSet.

Type: String

Valid Values: IPMatch | ByteMatch | SqlInjectionMatch | GeoMatch | SizeConstraint | XssMatch | RegexMatch

Required: Yes
```

### DIFF
--- a/website/docs/r/waf_rule.html.markdown
+++ b/website/docs/r/waf_rule.html.markdown
@@ -50,11 +50,11 @@ The following arguments are supported:
 #### Arguments
 
 * `negated` - (Required) Set this to `false` if you want to allow, block, or count requests
-  based on the settings in the specified `ByteMatchSet`, `IPSet`, `SqlInjectionMatchSet`, `XssMatchSet`, or `SizeConstraintSet`.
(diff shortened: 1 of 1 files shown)


website/docs/r/waf_rule.html.markdown
```diff
@@ -50,11 +50,11 @@ The following arguments are supported:
 #### Arguments
 
 * `negated` - (Required) Set this to `false` if you want to allow, block, or count requests
-  based on the settings in the specified `ByteMatchSet`, `IPSet`, `SqlInjectionMatchSet`, `XssMatchSet`, or `SizeConstraintSet`.
+  based on the settings in the specified `IPMatch`, `ByteMatch`, `SqlInjectionMatch`, `SizeConstraint` or `XssMatch`.
```
radeksimko: I agree the list here may not be very helpful, but the patch here looks rather like a mistake? Because the context here is actually resource, not type, AFAICT.

I think we could either leave it as it was or change it to human-readable names (e.g. Byte Match Set, IP Set, ...) or to Terraform resource names (e.g. `aws_waf_byte_match_set`, `aws_waf_ipset`) possibly with links to docs of those resources.

What do you think?

loivis: updated with links to terraform resources

PR #3490, opened 2018-02-22
Show exact AssumeRole error

redbaron:
Chained provider conseals real provider error unless verbose flag is
given. When assome_role is specified, only single provider is used,
so it make sense to get credentials straight from it and report
exact error to users.

### DIFF
--- a/aws/auth_helpers.go
+++ b/aws/auth_helpers.go
@@ -223,21 +223,10 @@ func GetCredentials(c *Config) (*awsCredentials.Credentials, error) {
 		assumeRoleProvider.Policy = aws.String(c.AssumeRolePolicy)
 	}
 
-	providers = []awsCredentials.Provider{assumeRoleProvider}
-
-	assumeRoleCreds := awsCredentials.NewChainCredentials(providers)
+	assumeRoleCreds := awsCredentials.NewCredentials(assumeRoleProvider)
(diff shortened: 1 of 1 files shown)


redbaron:
@bflad, can it go to 1.11 milestone  please?

bflad:
@redbaron is there a reason this should be prioritized over the other already upvoted pull requests? https://github.com/terraform-providers/terraform-provider-aws/pulls?q=is%3Apr+is%3Aopen+sort%3Areactions-%2B1-desc

You can always create a custom build of the provider if necessary in your environment.

redbaron:
@bflad , sorry I am not familiar with release process and how fixes end up in final releases. I figured this change is small and safe enough to be in next release. I spent quite some time trying to debug failing assume role,  so find it useful in case other people end up in the same situation.

If it automatically goes into some release  in the future, I am fine with it, as long as it doesn't left forgotten :)

redbaron:
any chance it can be merged? :-/  whats the process?

aeschright:
Hi @redbaron! Thanks for your work on this. We ended up splitting this section of the code into a separate package and [doing some work on error handling](https://github.com/hashicorp/aws-sdk-go-base/pull/5) that I think addresses your concerns. Please open a new issue if you run into additional problems, and we'll try to sort out any remaining bugs. Thanks again!

PR #3491, opened 2018-02-22
Use session.NewSession instead of depricated session.New

oscr:
session.New() is deprecated in aws-sdk-go ([source](https://github.com/aws/aws-sdk-go/blob/ecc70eb46d61cd1c5b54972d9f61a94bc94b93d9/aws/session/session.go#L53)) and is replaced with session.NewSession(). 

This commit replaces New() with NewSession() and wraps it with Must() for error checking.


oscr:
@bflad Hi Brian! I noticed that this pull request was still open so I updated it just to make sure it still works. Is it still of interest or should I close it?

bflad:
Hi @oscr 👋 Thank you for submitting this and sorry for the delayed review.

At some point since this pull request was opened, we migrated the shared authentication pieces of the Terraform AWS Provider and Terraform S3 Backend (e.g. `auth_helpers.go`) into its own library: https://github.com/hashicorp/aws-sdk-go-base -- I believe this change is already represented in that library, but otherwise a PR would certainly be welcome.

As for `resource_aws_iam_user_login_profile_test.go`, that still looks like its using `session.New()`. We would prefer this to be implemented with error checking instead of using `session.Must()` though since a panic is less than ideal (dangling test resources, etc).

Due to the age of this PR and its current merge conflicts, I'm going to close this PR for now, but updating that particular usage would certainly be appreciated in a new PR and I will keep an eye out for it. Thanks again!

PR #3297, opened 2018-02-08
Added CodeBuild flag 'badge_enabled'

vpadronblanco:
Closes #2836 

- Added 'badge_enabled' resource property
- Added 'badge_url' resource property to be visible for attribute referencing
- Added properties in the documentation
- Added acceptance test

Test output:
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSCodeBuildProject_buildBatchUrlValidation -timeout 120m
=== RUN   TestAccAWSCodeBuildProject_buildBatchUrlValidation
--- PASS: TestAccAWSCodeBuildProject_buildBatchUrlValidation (25.30s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	25.341s

PR #3504, opened 2018-02-23
Added badge_enabled argument, exported badge_url parameter

vpadronblanco:
Solves 

Resolves my problem with branches, reason why I closed #3297 

Testing Output:

TF_ACC=1 go test ./aws -v -run=TestAccAWSCodeBuildProject_buildBadgeUrlValidation -timeout 120m
=== RUN   TestAccAWSCodeBuildProject_buildBadgeUrlValidation
--- PASS: TestAccAWSCodeBuildProject_buildBadgeUrlValidation (24.83s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	24.868s


aws/resource_aws_codebuild_project.go
```diff
@@ -180,6 +180,16 @@ func resourceAwsCodeBuildProject() *schema.Resource {
...
 			},
+			"badge_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+			"badge_url": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
bflad: Since we cannot set the badge URL via the API, `Optional: true` should be removed from this attribute. More information: https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#optional

aws/resource_aws_codebuild_project.go
```diff
@@ -180,6 +180,16 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 				Default:      "60",
 				ValidateFunc: validateAwsCodeBuildTimeout,
 			},
+			"badge_enabled": {
```
bflad: In order for Terraform to determine that this parameter has changed from the API, we need to call `d.Set("badge_enabled", project.Badge.BadgeEnabled)` in the read function.

aws/resource_aws_codebuild_project.go
```diff
@@ -451,6 +465,7 @@ func resourceAwsCodeBuildProjectRead(d *schema.ResourceData, meta interface{}) e
 	d.Set("name", project.Name)
 	d.Set("service_role", project.ServiceRole)
 	d.Set("build_timeout", project.TimeoutInMinutes)
+	d.Set("badge_url", project.Badge.BadgeRequestUrl)
```
bflad: In order to prevent potential nil dereference panic, we need to wrap these in a nil check:

```go
if project.Badge != nil {
  d.Set("badge_enabled", project.Badge.BadgeEnabled)
  d.Set("badge_url", project.Badge.BadgeRequestUrl)
} else {
  d.Set("badge_enabled", false)
  d.Set("badge_url", "")
}
```

aws/resource_aws_codebuild_project_test.go
```diff
@@ -774,3 +793,87 @@ func testAccAWSCodeBuildProjectConfig_vpcConfig(subnets string) string {
...
+        "logs:CreateLogStream",
+        "logs:PutLogEvents"
+      ]
+    }
+  ]
+}
+POLICY
+}
+
+resource "aws_iam_policy_attachment" "codebuild_policy_attachment" {
```
bflad: Minor nitpick: We generally discourage the usage of `aws_iam_policy_attachment` in preference of `aws_iam_role_policy_attachment` in this case as the former resource often provides confusing/unexpected behavior for operators if they are unfamiliar with all the caveats of using that resource.

bflad:

PR #3513, opened 2018-02-24
Closes #3004: adds support for modifying Beanstalk tags

### LINKED ISSUE #3004: Add capability to update Elastic Beanstalk tags
### Terraform Version
Terraform v0.11.2

### Affected Resource(s)
- aws_elastic_beanstalk_environment

In the documentation for `aws_elasticbeanstalk_environment` is the following (see https://www.terraform.io/docs/providers/aws/r/elastic_beanstalk_environment.html#tags )

> tags – (Optional) A set of tags to apply to the Environment. Note: at this time the Elastic Beanstalk API does not provide a programatic way of changing these tags after initial application

I'm not sure when it was added but the Elastic Beanstalk API now /does/ provide a programatic way of changing tags after initial application. See:

* https://docs.aws.amazon.com/cli/latest/reference/elasticbeanstalk/update-tags-for-resource.html
* https://docs.aws.amazon.com/elasticbeanstalk/latest/api/API_UpdateTagsForResource.html

Would be good to get this functionality incorporated in the AWS provider.

ctreatma:
The Elastic Beanstalk environment resource did not support updating
tags on an existing environment, forcing the use of other tools to
manage environment tags.

The Elastic Beanstalk API now supports updating tags via a separate
`UpdateTagsForResource` call.  This updates the terraform resource
to make an `UpdateTagsForResource` call whenever there are changes
to the `tags` attribute.

### DIFF
--- a/aws/resource_aws_elastic_beanstalk_environment.go
+++ b/aws/resource_aws_elastic_beanstalk_environment.go
@@ -317,26 +317,9 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
 	envId := d.Id()
 
 	var hasChange bool
-	var hasTagChange bool
 
 	updateOpts := elasticbeanstalk.UpdateEnvironmentInput{
 		EnvironmentId: aws.String(envId),
-	}
-
-	updateTags := elasticbeanstalk.UpdateTagsForResourceInput{
-		ResourceArn: aws.String(d.Get("arn").(string)),
-	}
-
-	if d.HasChange("tags") {
-		hasTagChange = true
-		o, n := d.GetChange("tags")
-		oldTags := tagsFromMapBeanstalk(o.(map[string]interface{}))
-		newTags := tagsFromMapBeanstalk(n.(map[string]interface{}))
-
-		tagsToAdd, tagNamesToRemove := diffTagsBeanstalk(oldTags, newTags)
-
-		updateTags.TagsToAdd = tagsToAdd
-		updateTags.TagsToRemove = tagNamesToRemove
 	}
 
 	if d.HasChange("description") {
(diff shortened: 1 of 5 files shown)


aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -496,6 +562,10 @@ func resourceAwsElasticBeanstalkEnvironmentRead(d *schema.ResourceData, meta int
 		return err
 	}
 
+	if err := d.Set("arn", env.EnvironmentArn); err != nil {
```
bflad: Nitpick: Unfortunately the code around it does not follow the convention, but for simple `string` attributes, we do not need the `err` handling logic.

```go
d.Set("arn", env.EnvironmentArn)
```

aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -313,11 +317,28 @@ func resourceAwsElasticBeanstalkEnvironmentUpdate(d *schema.ResourceData, meta i
...
 
 	updateOpts := elasticbeanstalk.UpdateEnvironmentInput{
 		EnvironmentId: aws.String(envId),
 	}
 
+	updateTags := elasticbeanstalk.UpdateTagsForResourceInput{
+		ResourceArn: aws.String(d.Get("arn").(string)),
+	}
+
+	if d.HasChange("tags") {
```
bflad: Can you move all the tag update logic inside this conditional here? We do not need to make this more complicated by spreading the logic everywhere like the other update call (which needs to build up arguments).

aws/resource_aws_elastic_beanstalk_environment.go
```diff
@@ -56,6 +56,10 @@ func resourceAwsElasticBeanstalkEnvironment() *schema.Resource {
 		MigrateState:  resourceAwsElasticBeanstalkEnvironmentMigrateState,
 
 		Schema: map[string]*schema.Schema{
+			"arn": {
```
bflad: We should ensure this new attribute has an associated test of some sort in `TestAccAWSBeanstalkEnv_basic`:

```go
resource.TestCheckResourceAttrSet("aws_elastic_beanstalk_environment.tfenvtest", "arn"),
# or preferably
resource.TestMatchResourceAttr("aws_elastic_beanstalk_environment.tfenvtest", "arn", regexp.MustCompile(fmt.Sprintf("^arn:[^:]+:elasticbeanstalk:[^:]+:[^:]+:environment/%s/%s$", appName, envName)),
```

ctreatma:
Thanks for the feedback!  I should be able to address your suggestions in the next couple days.

ctreatma:
Hi, @bflad.  I _think_ I got everything in place for the acceptance tests.  Please let me know if there's anything I missed or if there are things I can clean up.  Thanks!

aws/resource_aws_elastic_beanstalk_environment_test.go
```diff
@@ -283,6 +287,49 @@ func TestAccAWSBeanstalkEnv_resource(t *testing.T) {
 	})
 }
 
+func TestAccAWSBeanstalkEnv_tags(t *testing.T) {
```
bflad: This is currently failing with:
```
=== RUN   TestAccAWSBeanstalkEnv_tags
--- FAIL: TestAccAWSBeanstalkEnv_tags (474.54s)
    testing.go:513: Step 0 error: Check failed: Check 1/1 error: InvalidParameter: 1 validation error(s) found.
        - missing required field, ListTagsForResourceInput.ResourceArn.
```

The first test step needs to include this as the first check so `app` is correctly populated: `testAccCheckBeanstalkEnvExists("aws_elastic_beanstalk_environment.tfenvtest", &app),`

FYI you can run the acceptance test locally with `make testacc TEST=./aws TESTARGS='-run=TestAccAWSBeanstalkEnv_tags` to verify

bflad:

PR #3522, opened 2018-02-25
fixes #2825 supports classic EIP’s in data.aws_eip

### LINKED ISSUE #2825: aws_eip data provider does not support Classic EIPs
The following snippet leads to a crash when the EIP is on ec2-classic. It does work if it is a VPC EIP:

### Terraform Version
Terraform v0.11.1
+ provider.aws v1.6.0

### Affected Resource(s)
- aws_eip

```hcl
provider "aws" {
    version = "1.6"
}

data "aws_eip" "this" {
  public_ip = "10.10.10.10"
}
```

### Debug Output
https://gist.github.com/Ashex/7035dc2d335dfc00650495e59816b979

### Panic Output
https://gist.github.com/Ashex/7035dc2d335dfc00650495e59816b979

### Expected Behavior
EIP allocation Id is retriever

### Actual Behavior
Terraform crashes with a EOF error

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Verified that this occurs on aws provider 1.5.0

When I run this with a VPC EIP it succeeds, this appears to only fail with a Classic EIP

richardbowden:


bflad:
@richardbowden do you think it makes more sense to instead deprecate the unclear `id` attribute and instead use a new `allocation_id` attribute? I think that would simplify the data source usage and documentation. 😄

richardbowden:
@bflad not to sure I follow what you mean? Do you mean replace the usage of `id` with `allocation_id ` in the argument ref / schema  or something else ?

Ashex:
@bflad Could you clarify the inquiry? This bug is a bit disruptive and I've had to wrap a module around an external data source as a workaround.

bflad:
> Do you mean replace the usage of id with allocation_id in the argument ref / schema or something else ?

Sorry, yes, I mean we should probably mark the `id` attribute in the schema as `Deprecated: "Use allocation_id (VPC EIP) or public_ip (EC2-Classic EIP) instead"` then create then `allocation_id` attribute with its appropriate handling.

PR #3524, opened 2018-02-25
New Resource: aws_organizations_account

asedge:
A new resource to implement AWS Organizations member accounts: #571.

I updated the branch based on feedback from #903.  There's probably more to fix up because this branch hadn't been touched in quite a while.  This is quite a bit more difficult to test because once you create a member account in an organization that organization cannot be deleted.  The member accounts themselves cannot be deleted, only suspended, and each account creation consumes an email address that cannot be reused.

Originally attempted a PR at: hashicorp/terraform#14147.

CC: @bflad


aws/resource_aws_organizations_account_test.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		params := &organizations.DescribeAccountInput{
+			AccountId: &rs.Primary.ID,
+		}
+
+		resp, err := conn.DescribeAccount(params)
+
+		if err != nil {
+			return err
```
bflad: We might need to check for `AccountNotFoundException` as the SDK documentation notes:

> We can't find an AWS account with the AccountId that you specified, or the
account whose credentials you used to make this request is not a member of
an organization.

I believe in our testing at this point the testing account will be removed from the organization again, so this would be the case:

```go
if isAWSErr(err, organizations.ErrCodeAccountNotFoundException, "") {
  return nil
}
```

website/aws.erb
```diff
@@ -1377,6 +1377,9 @@
                         <li<%= sidebar_current("docs-aws-resource-organizations-organization") %>>
                             <a href="/docs/providers/aws/r/organizations_organization.html">aws_organizations_organization</a>
                         </li>
+                        <li<%= sidebar_current("docs-aws-resource-organizations-account") %>>
+                            <a href="/docs/providers/aws/r/organizations_account.html">aws_organizations_account</a>
```
bflad: Minor nitpick: Can you sort this alphabetically please?

website/docs/r/organizations_account.html.markdown
```diff
@@ -0,0 +1,42 @@
+---
+layout: "aws"
+page_title: "AWS: aws_organizations_account
+sidebar_current: "docs-aws-resource-organizations-account|"
+description: |-
+  Provides a resource to create a member account in the current AWS Organization.
+---
+
+# aws\_organizations\_account
```
bflad: Minor nitpick: Backslashes are no longer required in documentation titles

website/docs/r/organizations_account.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+ 
+-> **Note:** Account creation must be done from the organization's master account.
+
+-> **Note:** AWS member accounts must be deleted manually by following these steps: 1) Perform a root account password recovery for the email address that was specified for the account in Organizations. 2) Login to the account as that root user. 3) Navigate to "My Organization" in the account menu top-right. 4) Leave the organization. 5) Once the account has successfully left the organization, delete the account as usual.
```
bflad: We should link to the AWS Organizations documentation for this information, e.g. 

```
~> **Note:** Deleting this Terraform resource will only remove an AWS account from an organization. Terraform will not close the account. The member account must be prepared to be a standalone account beforehand. See the [AWS Organizations documentation](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_accounts_remove.html) for more information.

File: website/docs/r/organizations_account.html.markdown [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+layout: "aws"
+page_title: "AWS: aws_organizations_account
+sidebar_current: "docs-aws-resource-organizations-account|"
+description: |-
+  Provides a resource to create a member account in the current AWS Organization.
+---
+
+# aws\_organizations\_account
+ 
+-> **Note:** Account creation must be done from the organization's master account.
```
bflad: We should imply this for all account management, e.g.
```
~> **Note:** Account management must be done from the organization's master account.
```

aws/resource_aws_organizations_account.go
```diff
@@ -0,0 +1,228 @@
...
+
+	d.Set("arn", resp.Account.Arn)
+	d.Set("joined_method", resp.Account.JoinedMethod)
+	d.Set("joined_timestamp", resp.Account.JoinedTimestamp)
+	d.Set("name", resp.Account.Name)
+	d.Set("status", resp.Account.Status)
+	return nil
+}
+
+func resourceAwsOrganizationsAccountDelete(d *schema.ResourceData, meta interface{}) error {
```
bflad: Due to how the AWS Organizations service and closing AWS accounts in general works, we should probably set up the Terraform deletion function here to call [`RemoveAccountFromOrganization`](https://docs.aws.amazon.com/sdk-for-go/api/service/organizations/#Organizations.RemoveAccountFromOrganization) and pass along the error to the Terraform user. I put a review comment to update the documentation appropriately as well. e.g.

```go
func resourceAwsOrganizationsAccountDelete(d *schema.ResourceData, meta interface{}) error {
  conn := meta.(*AWSClient).organizationsconn

  input := &organizations.RemoveAccountFromOrganizationInput{
    AccountId: aws.String(d.Id()),
  }
  log.Printf("[DEBUG] Removing AWS account from organization: %s", input)
  _, err := conn.RemoveAccountFromOrganization(input)
  if err != nil {
    if isAWSErr(err, organizations.ErrCodeAccountNotFoundException, "") {
      return nil
    }
    return err
  }
  return nil
}
```

aws/resource_aws_organizations_account.go
```diff
@@ -0,0 +1,228 @@
...
+	log.Printf("[DEBUG] Account create config: %#v", createOpts)
+
+	var err error
+	var resp *organizations.CreateAccountOutput
+	err = resource.Retry(4*time.Minute, func() *resource.RetryError {
+		resp, err = conn.CreateAccount(createOpts)
+
+		if err != nil {
+			ec2err, ok := err.(awserr.Error)
+			if !ok {
```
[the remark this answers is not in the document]
asedge: This may have been a mistake from when I was trying to clean up.

aws/resource_aws_organizations_account.go
```diff
@@ -0,0 +1,228 @@
...
+
+	log.Printf("[DEBUG] Account create config: %#v", createOpts)
+
+	var err error
+	var resp *organizations.CreateAccountOutput
+	err = resource.Retry(4*time.Minute, func() *resource.RetryError {
+		resp, err = conn.CreateAccount(createOpts)
+
+		if err != nil {
+			ec2err, ok := err.(awserr.Error)
```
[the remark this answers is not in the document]
asedge: This may have been a mistake to leave in.

aws/resource_aws_organizations_account.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if iam_user, ok := d.GetOk("iam_user_access_to_billing"); ok {
+		createOpts.IamUserAccessToBilling = aws.String(iam_user.(string))
+	}
+
+	log.Printf("[DEBUG] Account create config: %#v", createOpts)
+
+	var err error
+	var resp *organizations.CreateAccountOutput
+	err = resource.Retry(4*time.Minute, func() *resource.RetryError {
```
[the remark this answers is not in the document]
asedge: I don't remember 100% because this code is nearly a year old.  After looking at it for a little bit I think the issue was that there's no API action to check the status of `CreateOrganization`.  I didn't realize this could be a problem until I wrote this resource so I put the poll here.  Allowing `organizations.ErrCodeFinalizingOrganizationException` as a `RetryableError` seems to back that up.

aws/resource_aws_organizations_account.go
asedge: OK, I can do that.  In fact, since you wrote the code feel free to update the PR.  Let me know either way, thanks.

asedge:
@bflad I took care of nearly all the feedback you provided.  I provided a response about the 4 minute retry timeout but haven't made any changes on it just yet.

aws/resource_aws_organizations_account.go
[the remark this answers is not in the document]
bflad: Awesome information! If that's the case, maybe we (more like I 😄 ) should setup the organization resource to wait until the organization is finalized before returning that its done on creation. I'll make a note to check that myself.

bflad:
@asedge thank you so much again for your work here! You've really gone above and beyond here. 😄 

At this point, I need to temporarily put this PR on hold while I work with AWS support on how we can regularly acceptance test this in a designated AWS account. No clue what their response will be like yet :joy:.

If people are feeling brave, maybe they could try this out themselves in a custom build? `terraform state rm` is always an option to back out the account management with Terraform. At the current time, I don't think there will be much in the line of "breaking changes" to the resource (e.g. attribute changes), but I cannot make a 100% promise yet until I get more details from AWS.

aws/resource_aws_organizations_account.go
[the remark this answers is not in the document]
asedge: If you create an issue and assign it to me, I will get it done.

lorengordon:
I might be able to play with this. Would be awesome to be able to manage accounts as resources (and data sources) through Terraform. Would someone be able to create a Windows binary from this branch?

asedge:
@bflad Thanks for the review and merge!

bflad:

Flowman:
Can we extend this with add account to specific OU?

PR #3525, opened 2018-02-25
Adds filter to data.aws_eip 

### LINKED ISSUE #3423: aws_eip data source missing support for tags
_This issue was originally opened by @MaximF as hashicorp/terraform#17361. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
> terraform -v
Terraform v0.11.3
+ provider.aws v1.9.0
```

```hcl
data "aws_eip" "sample" {
  tags {
		Name ="A valuable IP"
	}
}

```

### Crash Output
```hcl
Error: data.aws_eip.sample: : invalid or unknown key: tags
```

### Expected Behavior
Terraform would return an IP with a described "Name" tag value

### Actual Behavior
Error: data.aws_eip.sample: : invalid or unknown key: tags

### Steps to Reproduce
1. `terraform init`
2. `terraform apply`

### References
The same issue was fixed for `aws_eip` resource just over a month ago at hashicorp/terraform#16993.
Now it is a turn for `aws_eip` data source

richardbowden:
fixes issue #3423 

this is based of PR "https://github.com/terraform-providers/terraform-provider-aws/pull/3522"

Had to fix the failing tests before I could submit this PR.

this addes the ability to search for EIP’s via tags or other attribute that EIP’s can use in a filter.

This did highlight a known issue with regrads to how terraform eval’s data sources before they should if be the resourse is a computed value. This is only an issue if the data block is referencing a resource that is created at the same time. If referencing a pre exiting resource this does not happen.

the test config testAccDataSourceAwsEipFilterConfig makes a interpol ref to force terraform to eval the data block after the resource has either been created or read in `values = ["${aws_eip.test.tags.Name}”]` see the following links

hashicorp/terraform#10603
hashicorp/terraform#17173

### DIFF
--- a/aws/data_source_aws_eip.go
+++ b/aws/data_source_aws_eip.go
@@ -55,7 +55,13 @@ func dataSourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
 
 	eip := resp.Addresses[0]
 
-	d.SetId(*eip.AllocationId)
+	if *eip.Domain == "vpc" {
+		d.SetId(*eip.AllocationId)
+	} else {
+		log.Printf("[DEBUG] Reading EIP, has no AllocationId, this means we have a Classic EIP, the id will also be the public ip : %s", req)
+		d.SetId(*eip.PublicIp)
+	}
+
 	d.Set("public_ip", eip.PublicIp)
 
 	return nil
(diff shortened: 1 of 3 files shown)


crohr:
Any chance to see this merged? In our case it would be very valuable to be able to setup a fixed pool of EIPs, and assign them to instances via a data source. This makes sure we always keep the same EIPs in case we have to open firewalls etc.

msays2000:
curious when will this make it in ?

bflad:

PR #3526, opened 2018-02-25
fixes #2917 inconsistant enable_dns_hostnames

### LINKED ISSUE #2917: aws_vpc enable_dns_hostnames inconsistent between import and create
### Terraform Version
Terraform v0.11.1
+ provider.aws v1.5.0
+ provider.template v1.0.0

### Affected Resource(s)
- aws_vpc

```hcl
resource "aws_vpc" "default" {
  cidr_block = "172.31.0.0/16"
}

resource "aws_vpc" "dmz" {
  cidr_block = "172.31.0.0/16"
}
```

### Actual Behavior
`default` already existed in my AWS account, and I imported it into terraform. `dmz` was created by terraform.

Terraform knows that they have different values for `enable_dns_hostnames`:

```
$ terraform state show aws_vpc.default | grep enable_dns_hostnames
enable_dns_hostnames             = true
$ terraform state show aws_vpc.dmz | grep enable_dns_hostnames
enable_dns_hostnames             = false
```

(and according to the AWS console these values reflect reality, i.e. terraform is reading their values correctly)

But `terraform plan` shows no changes to be made:

```
No changes. Infrastructure is up-to-date.

This means that Terraform did not detect any differences between your
configuration and real physical resources that exist. As a result, no
actions need to be performed.
```

### Expected Behavior
`terraform plan` should reflect this as a change to be made.

### Workaround
Explicitly setting `enable_dns_hostnames = "true"` for both resources causes `terraform plan` to make the change:

```

  ~ aws_vpc.dmz
      enable_dns_hostnames: "false" => "true"
```

richardbowden:


bchallenor:
Confirm that this fixes my issue.

I'm not as familiar with the other fields, but there are a number of other computed fields in that file. Could you check them to see if the same mistake was made with those?

bchallenor:
Also, is this code shared with `aws_default_vpc`, or does that need looking at separately?

richardbowden:
the only other option that would seem to give the same behaviour as this issue would be
"enable_classiclink": {
				Type:     schema.TypeBool,
				Optional: true,
				Computed: true,
			},

changed to 

"enable_classiclink": {
				Type:     schema.TypeBool,
				Optional: true,
				Default: false,
			},

will test at some point to day if I get time, unless you have 5 ?

richardbowden:
`aws_default_vpc` will be ok as in inherits from `resource_aws_vpc.go` so when thats fixed it will be good. But any setting that is only for the default_vpc can be overridden in `resource_aws_default_vpc.go`

bchallenor:
What about `enable_classiclink_dns_support`, too?

I think ClassicLink is only available in legacy (old) AWS accounts though, right? I don't think I can test that with mine.

bchallenor:
Is there anything blocking this being merged?

richardbowden:
not from my point of view!

bchallenor:
Can we merge it then?

The ClassicLink issue could be fixed separately (though probably not many people use ClassicLink anyway).

bflad:
While I think this is probably a good fix in principal, it could represent a breaking change for certain configurations, like default VPCs and those created via the web console.

The [VPC documentation](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html#vpc-dns-support) has this note:
> By default, both attributes are set to true in a default VPC or a VPC created by the VPC wizard. By default, only the enableDnsSupport attribute is set to true in a VPC created on the Your VPCs page of the VPC console or using the AWS CLI, API, or an AWS SDK.

Since the `aws_default_vpc` resource is inherited from the `aws_vpc` resource schema, it should set the attribute correctly according to the documentation:

```go
dvpc.Schema["enable_dns_hostnames"].Default = true
```

Can you please update that resource?

richardbowden:
updated in last commit - https://github.com/terraform-providers/terraform-provider-aws/pull/3526/commits/908b67acec2579b9491f422db54c2524d7bf64a8

gdavison:
Hi @richardbowden thanks for submitting this PR and sorry for the delay in responding to it.

I've closed the related issue. When a resource parameter isn't set and doesn't have a default value defined by the provider, Terraform doesn't perform any change tracking for the parameter, so it leaves it untouched, regardless of the value returned by the AWS API. If you want to enforce a value, your configuration should specify the value you want.

When no value is supplied, the AWS API defaults the Default VPC to enabling DNS hostnames, but other VPCs do not enabled DNS hostnames, which is why you're seeing the two values returned by the API.

PR #3533, opened 2018-02-26
consolidate tag and tags diff during asg update

enxebre:
Because of https://github.com/terraform-providers/terraform-provider-aws/issues/3532:
After importing and applying an asg, Terraform tries to destroy and recreate tags with same values yielding into additional aws permissions needed to manipulate tags, which are not needed for creating the asg.
This consolidates "tag" and "tags" at the time of making the diff to build the payload for the aws request.
Note that this is intended to be an improvement to avoid an unneeded extra aws request, but the root problem in https://github.com/terraform-providers/terraform-provider-aws/issues/3532 is still `terraform import` generated stated always setting `tag` field even if the config contains `tags`

### DIFF
--- a/aws/autoscaling_tags.go
+++ b/aws/autoscaling_tags.go
@@ -60,37 +60,31 @@ func setAutoscalingTags(conn *autoscaling.AutoScaling, d *schema.ResourceData) e
 		oraw, nraw := d.GetChange("tag")
 		o := setToMapByKey(oraw.(*schema.Set), "key")
 		n := setToMapByKey(nraw.(*schema.Set), "key")
-
-		old, err := autoscalingTagsFromMap(o, resourceID)
+		oldForTag, err := autoscalingTagsFromMap(o, resourceID)
 		if err != nil {
 			return err
 		}
 
-		new, err := autoscalingTagsFromMap(n, resourceID)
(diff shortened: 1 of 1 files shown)


bflad:
Hi @enxebre 👋  Thank you for contributing this and sorry it never got a proper review in the meantime. The Auto Scaling service tags are certainly a pain. 😖 

We plan on fixing the `aws_autoscaling_group` resource tags in two ways:

- Centralizing the conversion and difference logic to match all other AWS services that support resource tagging to remove the custom code, see also: https://github.com/terraform-providers/terraform-provider-aws/pull/13868
- Deprecating the `tags` set of maps argument (an argument type rarely, if ever, used anywhere else in the Terraform AWS Provider and was only present in this resource as a workaround for Terraform 0.11 and earlier configurations) in preference for the `tag` configuration blocks. This will allow us to consolidate all the logic to a single difference operation.

The combination of these should supersede the intent of this pull request, as #13868 should cover a similar difference collapse and the future deprecation will allow the `aws_autoscaling_group` resource tagging to behave like all other resources without needing to handle two tagging arguments. Importing and state saving of the tags will only be in the single schema attribute.

Please reach out if you have any questions and thank you again for submitting this.

PR #3536, opened 2018-02-26
WIP: Set application autoscaling policy type by the configuration used

tomelliff:
Just a failing test for now but should fix https://github.com/terraform-providers/terraform-provider-aws/issues/3520 when tests pass.

No idea why this doesn't work right now but step scaling for ECS and target tracking for DDB does.
Raised an AWS support ticket to see if they can explain what's wrong there in the mean time.

TODO:
- deprecate `policy_type` and simply get it from the type of scaling policy configuration used
- expand the acceptance test to check the returned target tracking scaling configuration

### DIFF
--- a/aws/resource_aws_appautoscaling_policy_test.go
+++ b/aws/resource_aws_appautoscaling_policy_test.go
@@ -74,6 +74,29 @@ func TestAccAWSAppautoScalingPolicy_nestedSchema(t *testing.T) {
 	})
 }
 
+func TestAccAWSAppautoScalingPolicy_targetTrackingEcs(t *testing.T) {
+	var policy applicationautoscaling.ScalingPolicy
+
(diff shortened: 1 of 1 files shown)


tomelliff:
I've fleshed out the acceptance test to cover a target tracking policy on an ECS service to match some amount of requests per target and added the currently necessary `policy_type` parameter for now before I refactor it out and deprecate it later tonight.

bflad:
Hi @tomelliff 👋  Are you planning on continuing work on this? If so, just let us know. If not, let's close this out to clean up the open pull requests to only those actually work in progress. Thanks!

bflad:
It's been a long time since we heard from you here so we're going to close this out to leave only actionable pull requests in the queue. If you or someone wants to continue this work please feel free to open a new PR (keeping the original commits 😄 ).

PR #3537, opened 2018-02-26
resource/aws_lb: Add Cross Zone Load Balancing support for NLBs

### LINKED ISSUE #3574: Feature request: cross_zone_load_balancing support for aws_lb
Can you add the following new directive and support for `cross_zone_load_balancing` in the [aws_lb](https://www.terraform.io/docs/providers/aws/r/lb.html) resource? This was recently released https://aws.amazon.com/about-aws/whats-new/2018/02/network-load-balancer-now-supports-cross-zone-load-balancing/.

mpilar:
Closes #3574 

Enable Cross Zone load balancing support for NLBs:
Just allows setting the `load_balancing.cross_zone.enabled` Attribute. 
Docs: 
https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancerAttribute.html
https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#LoadBalancerAttribute

```
TF_ACC=1 go test ./aws -v -run=TestAccAWSLB_networkLoadbalancer -timeout 120m 
=== RUN   TestAccAWSLB_networkLoadbalancer
--- PASS: TestAccAWSLB_networkLoadbalancer (245.89s)
=== RUN   TestAccAWSLB_networkLoadbalancerEIP
--- PASS: TestAccAWSLB_networkLoadbalancerEIP (262.47s)
=== RUN   TestAccAWSLB_networkLoadbalancer_subnet_change
--- PASS: TestAccAWSLB_networkLoadbalancer_subnet_change (277.25s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	786.288s```

### DIFF
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -372,7 +372,7 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 	if d.Get("load_balancer_type").(string) == "network" && d.HasChange("enable_cross_zone_load_balancing") {
 		attributes = append(attributes, &elbv2.LoadBalancerAttribute{
 			Key:   aws.String("load_balancing.cross_zone.enabled"),
-			Value: aws.String(fmt.Sprintf("%d", d.Get("enable_cross_zone_load_balancing").(int))),
+			Value: aws.String(fmt.Sprintf("%t", d.Get("enable_cross_zone_load_balancing").(bool))),
 		})
 	}
 
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -385,7 +385,7 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 		log.Printf("[DEBUG] ALB Modify Load Balancer Attributes Request: %#v", input)
 		_, err := elbconn.ModifyLoadBalancerAttributes(input)
 		if err != nil {
-			return fmt.Errorf("Failure configuring ALB attributes: %s", err)
+			return fmt.Errorf("Failure configuring LB attributes: %s", err)
 		}
 	}
 
@@ -398,7 +398,7 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 		_, err := elbconn.SetSecurityGroups(params)
 		if err != nil {
-			return fmt.Errorf("Failure Setting ALB Security Groups: %s", err)
+			return fmt.Errorf("Failure Setting LB Security Groups: %s", err)
 		}
 
 	}
@@ -417,7 +417,7 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 
 		_, err := elbconn.SetSubnets(params)
 		if err != nil {
-			return fmt.Errorf("Failure Setting ALB Subnets: %s", err)
+			return fmt.Errorf("Failure Setting LB Subnets: %s", err)
 		}
 	}
 
@@ -430,7 +430,7 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 
 		_, err := elbconn.SetIpAddressType(params)
 		if err != nil {
-			return fmt.Errorf("Failure Setting ALB IP Address Type: %s", err)
+			return fmt.Errorf("Failure Setting LB IP Address Type: %s", err)
 		}
 
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: aws/resource_aws_lb_test.go [hunk 1]
```diff
@@ -157,6 +157,7 @@ func TestAccAWSLBBackwardsCompatibility(t *testing.T) {
 					resource.TestCheckResourceAttr("aws_alb.lb_test", "tags.%", "1"),
 					resource.TestCheckResourceAttr("aws_alb.lb_test", "tags.Name", "TestAccAWSALB_basic"),
 					resource.TestCheckResourceAttr("aws_alb.lb_test", "enable_deletion_protection", "false"),
+					resource.TestCheckResourceAttr("aws_alb.lb_test", "enable_cross_zone_load_balancing", "false"),
```
bflad: We should create a test that ensures enabling/disabling works as expected 😄

mpilar: Any hints as to where I could look for an example? It's my first time digging into a largeish golang codebase.

Is it sufficient to add the directive to `testAccAWSLBConfig_networkLoadBalancerEIP` as `true` and assert on that test?

mpilar:
Oh god I screwed up the PR.

aws/resource_aws_lb_test.go
mpilar: I was unable to run the acceptance tests after the commit. As soon as I have a break I will.

mpilar:
@bflad there's a bug I ran into when running the new acceptance test. I will hunt it down.

aws/resource_aws_lb_test.go
bflad: My general workflow does usually involve adding a directive/parameter to the test configuration function for the new attribute. Generally speaking though, its usually easiest to just create a whole new acceptance test and configuration function so you don't wind up with needing to pass in way too many directives to a function. Copy paste of an existing is okay and generally encouraged for this situation as the testing is more important than any (potential) later refactoring. 👍

aws/resource_aws_lb_test.go
```diff
@@ -980,6 +984,7 @@ resource "aws_route_table_association" "a" {
 resource "aws_lb" "test" {
   name            = "%s"
   load_balancer_type = "network"
+  enable_cross_zone_load_balancing = true
```
bflad: See my note above -- we'll probably want to just copy paste the handling of the new attribute to a new test and config or at least pass the true/false value for this attributes is a directive into the above config function and use `enable_cross_zone_load_balancing = %v` 😄

mpilar:
@bflad I made a new test and am running into some odd behavior, I thought I had a hang of the harness and how it works but this is baffling me.
https://gist.github.com/mpilar/dd0d9dcbf828e5b7fd3778fcce95ec5c

I captured a TRACE to see if I can figure out what's going on, but when the test runs it successfully creates the LB then destroys everything and doesn't create the second environment (or as I would prefer, update the second envionment). This happens regardless of whether I make it cross-zone first or second.

I am fairly sure it's some mistake I am making with the test harness but can't figure it out.

mpilar:
Flagged as WIP to add HTTP2 as well.

bflad:
I would prefer to treat that functionality in a separate PR please (smaller is easier to review)

mpilar:
@bflad then the PR should be good as is, or do you think it needs more?

The next commit has:
- added `enable_http2` as a flag
- added tests for the above
- made the boolean flag tests have 3 steps: on, off, on. to test creation, flag and unflag.
- increased timeout and delay for aws_internet_gateway detachment refresh 
- increased wait time to wait for network interfaces to be destroyed after deleting an NLB (this, combined with the above, caused the EIP hanging resource I saw yesterday)
- made all test fixtures use `lb_test` as the name for the `aws_lb` (to prevent the typo issue I had)

It's ready just waiting for all Acceptance tests to finish running (10 down, 5 to go). I decided to run them all after changing the timeouts and making the test fixtures more consistent.

aws/resource_aws_lb.go
```diff
@@ -362,6 +368,14 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
 		})
 	}
 
+	// It's important to know that Idle timeout is only supported for Network Loadbalancers
```
appilon: Comment references Idle timeout (probably meant to be cross zone flag)

bflad:

PR #3553, opened 2018-02-28
resource/aws_emr_cluster: Add Kerberos support

### LINKED ISSUE #3477: aws_emr_cluster enhancement to support KerberosAttributes
### Terraform Version
0.10.8
aws_provider 1.9.0

### Affected Resource(s)
- aws_emr_cluster

### Expected Behavior
Support the following attributes of the RunJobFlow API documented at https://docs.aws.amazon.com/emr/latest/APIReference/API_KerberosAttributes.html
```
   "KerberosAttributes": { 
      "ADDomainJoinPassword": "string",
      "ADDomainJoinUser": "string",
      "CrossRealmTrustPrincipalPassword": "string",
      "KdcAdminPassword": "string",
      "Realm": "string"
   }
```

bflad:
Closes #3477 

This will require a lot more time to setup a proper Active Directory acceptance test.

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSEMRCluster_Kerberos_ClusterDedicatedKdc'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEMRCluster_Kerberos_ClusterDedicatedKdc -timeout 120m
=== RUN   TestAccAWSEMRCluster_Kerberos_ClusterDedicatedKdc
--- PASS: TestAccAWSEMRCluster_Kerberos_ClusterDedicatedKdc (406.15s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	406.202s
```

### DIFF
--- a/aws/resource_aws_emr_cluster.go
+++ b/aws/resource_aws_emr_cluster.go
@@ -802,7 +802,7 @@ func flattenEc2Attributes(ia *emr.Ec2InstanceAttributes) []map[string]interface{
 func flattenEmrKerberosAttributes(d *schema.ResourceData, kerberosAttributes *emr.KerberosAttributes) []map[string]interface{} {
 	l := make([]map[string]interface{}, 0)
 
-	if kerberosAttributes == nil {
+	if kerberosAttributes == nil || kerberosAttributes.Realm == nil {
 		return l
 	}
 
@@ -813,7 +813,7 @@ func flattenEmrKerberosAttributes(d *schema.ResourceData, kerberosAttributes *em
 
 	m := map[string]interface{}{
 		"kdc_admin_password": d.Get("kerberos_attributes.0.kdc_admin_password").(string),
-		"realm":              aws.StringValue(kerberosAttributes.Realm),
+		"realm":              *kerberosAttributes.Realm,
 	}
 
 	if v, ok := d.GetOk("kerberos_attributes.0.ad_domain_join_password"); ok {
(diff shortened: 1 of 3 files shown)


FireballDWF:
If you can give me a test build which includes this code change, I can run a test which supplies the same parameters as the cluster I successfully created via cli and got working with with AD server.

bflad:
This was (manually) verified in an environment with Active Directory and the relevant AD attributes.

aws/resource_aws_emr_cluster.go
```diff
@@ -532,6 +576,10 @@ func resourceAwsEMRClusterRead(d *schema.ResourceData, meta interface{}) error {
 		log.Printf("[ERR] Error setting EMR Ec2 Attributes: %s", err)
 	}
 
+	if err := d.Set("kerberos_attributes", flattenEmrKerberosAttributes(d, cluster.KerberosAttributes)); err != nil {
+		log.Printf("[ERR] Error setting EMR Kerberos Attributes: %s", err)
```
paultyng: I know the others aren't doing it, but it feels weird to me that this doesn't result in an error returning out of the read?

bflad: Updated!

bflad:

copumpkin:
Is there a nice way to handle static passwords like these in terraform? The KDC admin password can't really be changed after the fact so it seems pointless to keep it in the state file (that's what happens right now), and makes me pretty uncomfortable to see it sitting there. Would be nice to be able to tag values as "ephemeral" somehow.

PR #3561, opened 2018-02-28
increase timeout in spot instance creation to 1 min - fix #3554

### LINKED ISSUE #3554: IAM Role creation race stil present in spot instance requests
_This issue was originally opened by @stadeltobi as hashicorp/terraform#17457. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.10.2
```
resource "aws_spot_instance_request" "quick-sir" {
  ami = "ami-5652ce39"
  instance_type = "t2.xlarge"
  spot_price = ".15"
  wait_for_fulfillment = true
  vpc_security_group_ids = <...>
  key_name = <...>
  iam_instance_profile = "${aws_iam_instance_profile.foo.id}"
  user_data = "${data.template_file.cloud-init.rendered}"
}
```

...
### Additional Context
I'm having the equivalent issue that is described in hashicorp/terraform#15341 when, instead of trying to create an aws_instance, I use a aws_spot_instance_request.

The same error ("Invalid Parameter Value") is thrown, so I'm assuming that the race condition described above was fixed for regular (on-demand) instances still prevails for spot_instance_requests.

Also, the workaround suggested by @spanktar makes the problem go away.

The problem goes away if I switch the spot request resource with an aws_instance resource, using the exact same parameters (except for the ones like spot_price that don't apply for aws_instance

### References

- hashicorp/terraform#15341

stadeltobi:
increase timeout from 15 secs to 1 minute, which matches the timeouts
for other places where we deal with IAM "eventually consistent"
semantics

### DIFF
--- a/aws/resource_aws_spot_instance_request.go
+++ b/aws/resource_aws_spot_instance_request.go
@@ -136,7 +136,7 @@ func resourceAwsSpotInstanceRequestCreate(d *schema.ResourceData, meta interface
 	log.Printf("[DEBUG] Requesting spot bid opts: %s", spotOpts)
 
 	var resp *ec2.RequestSpotInstancesOutput
-	err = resource.Retry(15*time.Second, func() *resource.RetryError {
+	err = resource.Retry(1*time.Minute, func() *resource.RetryError {
 		var err error
 		resp, err = conn.RequestSpotInstances(spotOpts)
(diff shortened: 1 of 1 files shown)


bflad:

PR #3573, opened 2018-02-28
Allow proxy for http transport

pawelsocha:
## Overview

This PR try to solve problem described here: https://github.com/hashicorp/terraform/issues/17386#

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


aws/config.go
```diff
@@ -300,10 +302,23 @@ func (c *Config) Client() (interface{}, error) {
 		opt.Config.Logger = awsLogger{}
 	}
 
-	if c.Insecure {
+	if c.Insecure || c.ProxyURL != "" {
+
 		transport := opt.Config.HTTPClient.Transport.(*http.Transport)
```
paultyng: This isn't part of this PR, but this is pretty fragile, we will probably need to rethink this down the road if we need to modify the transport in other ways or add a `RoundTripper`.

pawelsocha:
@paultyng thanks for your time to review this PR.
I learn something new. 

and I close this PR because NO_PROXY looks good to me :-)

Issue #3511, opened 2018-02-24
aws_lb does not support all loadbalancer attributes


ameir:
The `aws_lb` resource support access logs and deletion protection, but there are a few other attributes that are not exposed via this provider [0][1].  This resource exposes the two available attributes as top-level arguments, while, for example, CloudFormation asks for a list of attributes [2], which I think is more flexible, although maybe this resource tries to keep things similar to `aws_elb`.

I need to enable cross-zone load balancing and adjust the timeout, but this must currently be done out-of-band.

### Affected Resource(s)
- aws_lb

### References

0: https://www.terraform.io/docs/providers/aws/r/lb.html
1:
 https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancerAttribute.html
2: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html#cfn-elasticloadbalancingv2-loadbalancer-loadbalancerattributes

bflad:
Support for NLB cross-zone load balancing has been merged into master and will release with v1.11.0 of the AWS provider, likely in a week. It looks like @appilon and @mpilar will be working out HTTP/2 in a followup PR. 😄

ameir:
Thanks for the hard work, everyone!

octplane:
Hello,

In our tests:

- Creating an ALB with `enable_http2` set to false does not work: it creates an ALB with the flag set to true. The state contains `"enable_http2": "false"`.
- Modifying the value in the terraform code trigger the change up to AWS
- Modifying the value in AWS has no consequence: terraform does not attempt to fix that.

```
Terraform v0.11.2
+ provider.aws v1.11.0
```

juhtie01:
We seem to be having exactly the same issue as @octplane is having. Should this ticket be re-opened or a new one created?

appilon:
I've traced through the code and think I've found the issue, I should have a PR with a fix up soon

appilon:
The bug appears to be a bit more elusive than I thought. Working on a fix. I have made a new issue to track progress.
https://github.com/terraform-providers/terraform-provider-aws/issues/3848

PR #3578, opened 2018-03-01
[WIP] Fix 3511 support all elbv2 attributes

appilon:
First PR, implement missing load balancer attributes for elbv2, small refactor to better group lbtype specific attributes.

- [ ] Acceptance coverage
- [ ] Documentation update

### DIFF
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -157,6 +157,18 @@ func resourceAwsLb() *schema.Resource {
 				Default:  60,
 			},
 
+			"enable_cross_zone_load_balancing": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
+			"enable_http2": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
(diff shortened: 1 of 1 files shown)


bflad:
FYI cross-zone load balancing is being implemented in #3537

appilon:
Dang, well it was a good exercise, I'll change it to just add support for http2 (which is on by default and I don't know that is is any advantage to turning it off.. but at least it will close off covering all the attributes).

appilon:
I will wait for the mentioned PR to be merged and rebase from there, there is a small piece of refactoring I still think is a good idea

mpilar:
@appilon I have a commit ready with your refactor and adding http2 with tests. Once my PR is merged I will rebase the commit.

Thinking about the refactor, it might be a good idea now but if AWS adds more LB types and rules apply to certain combinations it might get unruly, but it can just be refactored again when it gets to that.

appilon:
@mpilar nice! yeah I agree things could get out of hand with more permutations. When your future PR gets merged I'll close this one. I was working off #3511 for reference for your next PR so you can trigger that issue being closed.

PR #3590, opened 2018-03-01
iam_account_alias: Pass awserr in returned error

Dominik-K:
This will display helpful error messages like

    The account alias ... already exists. (EntityAlreadyExists)

The error message is not red as it is common and useful in Terraform. Is there a setting for a `resource` that all returned `error`s are displayed in red? 
Always setting the [colorstring](https://github.com/hashicorp/terraform/blob/master/vendor/github.com/mitchellh/colorstring/README.md)'s `[red]`, etc. on each error message is unnecessary boilerplate, I think.

### DIFF
--- a/aws/resource_aws_iam_account_alias.go
+++ b/aws/resource_aws_iam_account_alias.go
@@ -41,7 +41,7 @@ func resourceAwsIamAccountAliasCreate(d *schema.ResourceData, meta interface{}) 
 	_, err := conn.CreateAccountAlias(params)
 
 	if err != nil {
-		return fmt.Errorf("Error creating account alias with name %s", account_alias)
+		return fmt.Errorf("Error creating account alias with name '%s': %s", account_alias, err)
 	}
 
 	d.SetId(account_alias)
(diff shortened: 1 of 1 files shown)


Dominik-K:
@bflad Printing `awserr.Error` also shows the `returned status code` and `request ID` in an extra line. That's what I want to avoid. I think both should only be shown as a debug output. Most times it's not necessary. This would help to make Terraform output less (unnecessary) verbose, i.e. more actionable.

bflad:
Some folks need this information for opening AWS support cases when its not possible to reproduce errors again with debug logging enabled.

Dominik-K:
Right, I didn't think about Terraform used in a CI system. If you use it manually, you just could repeat the call with `plan` or `apply`. Anyway, it doesn't hurt that much ;-) Ok, I deleted `fmtAwsError()`.

What about the red coloring output?

bflad:
> What about the red coloring output?

Probably worth a separate ticket. 😄  I'd be curious to see your opinion of what the implementation should look like and how it should be controlled. It most likely should be handled in Terraform core though since it sounds like it would apply for all Terraform providers/resources (AWS, GCP, Azure, etc.): https://github.com/hashicorp/terraform/

bflad:

Issue #3279, opened 2018-02-07
Support for Cognito Identity Provider


mdolian:
I believe the following documentation is no longer accurate.  I'm not sure when in the past an IAM SAML Provider was used as a Cognito Identity Provider, but Cognito now has it's own Identity Provider.  Please see the following API documentation:
https://www.terraform.io/docs/providers/aws/r/cognito_identity_pool.html

I don't believe Terraform has support for this at the moment.

Ninir:
Hi @mdolian 

Thanks for reporting this!

Cognito Identity Pool is using Authentication providers which are one/many from: Cognito, Amazon, Facebook, Google+, Twitter / Digits, OpenID, SAML or a Custom one.
SAML do is a correct *authentication* provider. 

> ... Cognito now has it's own Identity Provider

I'm wondering if you are referring to the Cognito *authentication* provider, which is using Cognito User Pool plus an Application... 🤔 
Could you better explain what you have in mind, providing a news from the blog or documentation link?

Thanks!

mdolian:
Sorry, I meant to include the actual API reference and forgot.  I believe this is where the gap exists:

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateIdentityProvider.html

mdolian:
Did the above reply provide enough information?  I'm more curious to know if this is a gap or if I'm potentially just doing it wrong.  I wouldn't mind trying to work on adding this feature but I don't want to dive in unless it's confirmed that it's needed.  Thanks!

Ninir:
Hey @mdolian 

Sorry for the late response. You can find the related SAML provider resource at https://www.terraform.io/docs/providers/aws/r/iam_saml_provider.html.

Thus, you should be able to use the [cognito identity pool](https://www.terraform.io/docs/providers/aws/r/cognito_identity_pool.html) freely and fully. 

Hope it fixes your issue!

mdolian:
@Ninir 

The IAM SAML Provider is not supported in Cognito.  The example in your documentation above will produce an error.  I tried it out and that is what led me to finding the new Cognito Identity Provider.  

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateIdentityProvider.html

If you can point me to where Terraform creates the resource above, then that solves the issue.  I believe Terraform is missing it.

Thanks

Ninir:
Hey @mdolian 

[...]

PR #3601, opened 2018-03-02
New Resource: aws_cognito_identity_provider

pawelsocha:
## Overview
New resource to create identity provider for Cognito user pool. 

This PR solves #3279

## Tests
```==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./... -v -run=TestAccAWSCognitoIdentityProvider_basic -timeout 120m
=== RUN   TestAccAWSCognitoIdentityProvider_basic
--- PASS: TestAccAWSCognitoIdentityProvider_basic (42.71s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	42.747s```

### DIFF
--- a/aws/resource_aws_cognito_identity_provider.go
+++ b/aws/resource_aws_cognito_identity_provider.go
@@ -56,6 +56,7 @@ func resourceAwsCognitoIdentityProvider() *schema.Resource {
 			"user_pool_id": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 		},
 	}
@@ -136,7 +137,7 @@ func resourceAwsCognitoIdentityProviderUpdate(d *schema.ResourceData, meta inter
 	log.Print("[DEBUG] Updating Cognito Identity Provider")
 
 	params := &cognitoidentityprovider.UpdateIdentityProviderInput{
-		UserPoolId:   aws.String(d.Get("UserPoolId").(string)),
+		UserPoolId:   aws.String(d.Get("user_pool_id").(string)),
 		ProviderName: aws.String(d.Id()),
 	}
 
--- a/aws/resource_aws_cognito_identity_provider.go
+++ b/aws/resource_aws_cognito_identity_provider.go
@@ -1,0 +1,185 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCognitoIdentityProvider() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCognitoIdentityProviderCreate,
+		Read:   resourceAwsCognitoIdentityProviderRead,
+		Update: resourceAwsCognitoIdentityProviderUpdate,
+		Delete: resourceAwsCognitoIdentityProviderDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"attribute_mapping": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"idp_identifiers": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"provider_details": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"provider_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"provider_type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"user_pool_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
+	}
+}
+
+func resourceAwsCognitoIdentityProviderCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+	log.Print("[DEBUG] Creating Cognito Identity Provider")
+
+	name := aws.String(d.Get("provider_name").(string))
+	params := &cognitoidentityprovider.CreateIdentityProviderInput{
+		ProviderName: name,
+		ProviderType: aws.String(d.Get("provider_type").(string)),
+		UserPoolId:   aws.String(d.Get("user_pool_id").(string)),
+	}
+
+	if v, ok := d.GetOk("attribute_mapping"); ok {
+		params.AttributeMapping = expandCognitoIdentityProviderMap(v.(map[string]interface{}))
+	}
+
+	if v, ok := d.GetOk("provider_details"); ok {
+		params.ProviderDetails = expandCognitoIdentityProviderMap(v.(map[string]interface{}))
+	}
+
+	if v, ok := d.GetOk("idp_identifiers"); ok {
+		params.IdpIdentifiers = expandStringList(v.([]interface{}))
+	}
+
+	_, err := conn.CreateIdentityProvider(params)
+	if err != nil {
+		return fmt.Errorf("Error creating Cognito Identity Provider: %s", err)
+	}
+
+	d.SetId(*name)
+
+	return resourceAwsCognitoIdentityProviderRead(d, meta)
+}
+
+func resourceAwsCognitoIdentityProviderRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+	log.Printf("[DEBUG] Reading Cognito Identity Provider: %s", d.Id())
+
+	ret, err := conn.DescribeIdentityProvider(&cognitoidentityprovider.DescribeIdentityProviderInput{
+		ProviderName: aws.String(d.Id()),
+		UserPoolId:   aws.String(d.Get("user_pool_id").(string)),
+	})
+
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	ip := ret.IdentityProvider
+	d.Set("provider_name", ip.ProviderName)
+	d.Set("provider_type", ip.ProviderType)
+	d.Set("user_pool_id", ip.UserPoolId)
+
+	if err := d.Set("attribute_mapping", flattenCognitoIdentityProviderMap(ip.AttributeMapping)); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting attribute_mapping error: %#v", err)
+	}
+
+	if err := d.Set("provider_details", flattenCognitoIdentityProviderMap(ip.ProviderDetails)); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting provider_details error: %#v", err)
+	}
+
+	if err := d.Set("idp_identifiers", flattenStringList(ip.IdpIdentifiers)); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting idp_identifiers error: %#v", err)
+	}
+
+	return nil
+}
+
+func resourceAwsCognitoIdentityProviderUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+	log.Print("[DEBUG] Updating Cognito Identity Provider")
+
+	params := &cognitoidentityprovider.UpdateIdentityProviderInput{
+		UserPoolId:   aws.String(d.Get("UserPoolId").(string)),
+		ProviderName: aws.String(d.Id()),
+	}
+
+	if d.HasChange("attribute_mapping") {
+		params.AttributeMapping = expandCognitoIdentityProviderMap(d.Get("attribute_mapping").(map[string]interface{}))
+	}
+
+	if d.HasChange("provider_details") {
+		params.ProviderDetails = expandCognitoIdentityProviderMap(d.Get("provider_details").(map[string]interface{}))
+	}
+
+	if d.HasChange("idp_identifiers") {
+		params.IdpIdentifiers = expandStringList(d.Get("supported_login_providers").([]interface{}))
+	}
+
+	_, err := conn.UpdateIdentityProvider(params)
+	if err != nil {
+		return fmt.Errorf("Error updating Cognito Identity Provider: %s", err)
+	}
+
+	return resourceAwsCognitoIdentityProviderRead(d, meta)
+}
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by pawelsocha** [comment]
@bflad maybe you know how to handle perpetual diffs for TypeMap variable?

**Comment by leonfs** [comment]
Any idea on when this PR will get merged?

At the moment as soon as you change/add/remove a Provider through the console terraform, for some unexpected reason, tries to recreate the user pool.

**Comment by pawelsocha** [comment]
@leonfs thanks for your feedback. I will check this :-)

**Comment by leonfs** [comment]
@pawelsocha No worries, thanks for making this PR. 

Though the bug I mentioned earlier does not have anything to do with your code but with the current implementation of the `aws_cognito_user_pool_client`.

I suspect that after adding an identity provider (let's say Google) to the user pool through AWS console, the api call to AWS from Terraform returns an array with `Google` as value in the `suppoted-Idenityt_providers` field. Therefore,  terraform state is different from what it is in AWS. 

For some reason Terraform tries to recreate the whole resource (it shouldn't as it's a value that can change after creation). 

I'm wondering if with this PR the problem will actually be solved. Or it's a different problem.

**Comment by pawelsocha** [comment]
@leonfs Terraform recreate aws_cognito_user_pool_client resource every apply/plan because it's affected by TypeSet implementation in TF: https://github.com/hashicorp/terraform/issues/15420#issuecomment-311810069 


And also in resource source attribute have ForceNew set to true:
```
		"schema": {
				Type:     schema.TypeSet,
				Optional: true,
				Computed: true,
				ForceNew: true,
```

**Comment by leonfs** [comment]
@pawelsocha That's interesting. What I don't understand this is when a change in the schema attributes happen. 

What triggered for me the change was adding/changing and identity provider (Google). What link could exist between those 2 things?

**Comment by pawelsocha** [comment]
@leonfs change is forced by resource definition in code.

**Comment by leonfs** [comment]
@pawelsocha - Not really sure if I get what you are saying. 

I can see that there is a field `ForceNew: true` but that means that if the `schema` attribute has changed then the whole resource `aws_cognito_user_pool` would be recreated.

But when I change/add the identity provider I don't change the schema attribute. So why would that force the recreation?

**Comment by pawelsocha** [comment]
@leonfs when tf recreating ```aws_cognito_user_pool``` generate new resource ID and this id is used in identity provider.

**Comment by leonfs** [comment]
@pawelsocha Not really sure about what you are saying. Does not really make sense. TF should not recreate the `aws_cognito_user_pool` in the first place, and that's the problem. But the good news is that I've found a solution to the problem by digging a bit more into the plan output.

It looks like AWS adds a custom attribute called `identities` as soon as you add an identity provider to the user pool. This is what's making the set of schemas to change.

The problem is fixed by adding the following schema attribute. 

```
schema {
    attribute_data_type      = "String"
    developer_only_attribute = false
    mutable                  = true
    name                     = "identities"

    string_attribute_constraints {}
  }
```

Notice the empty `string_attribute_constraint`. This is a bug in the cognito module. You can't not include it as that would generate a change in the plan on every apply. 

In any case - thanks for the feedback and looking forward to your PR getting merged.

**Comment by pawelsocha** [comment]
@leonfs :-) Today I had only a few hours to investigate this problem. But, in my case terraform recreate the whole resource and create new ID. 
And I found this info: https://github.com/terraform-providers/terraform-provider-aws/blob/b8857d88b3cc0fe2cb208c8a36508ec531251155/aws/structure.go#L3005-L3007

**Comment by skoop22** [comment]
@pawelsocha 
I had the same issue when creating a workaround for this not yet existing feature.
AWS adds identities to the schema "Automagically" could not find any doc on it ...
As a workaround I added identities to my aws_cognito_user_pool.

```
  resource "aws_cognito_user_pool" "cognito_userpool" {
  name = "${var.cognito_user_pool_name}"

  tags {
    Name = "Cognito Userpool"
    CreatedBy = "terraform"
    Environment = "${var.environment}"
    Application = "${var.application}"
  }
  schema {
    attribute_data_type = "String"
    name = "email"
    required = true
    mutable = true
    string_attribute_constraints {
      min_length = 0
      max_length = 2048
    }
  }
  schema {
    attribute_data_type = "String"
    name = "identities"
    mutable = true
    string_attribute_constraints {
    }
  }
}

resource "aws_cognito_user_pool_domain" "cognito_domain" {
  user_pool_id = "${aws_cognito_user_pool.cognito_userpool.id}"
  domain = "${var.cognito_user_pool_name}"
}


data "template_file" "cognito_userpool_template" {
  template = "${file("cognito_userpool_identityprovider_template/cognito_userpool_identityprovider_template.js")}"
  vars {
    userPoolId = "${aws_cognito_user_pool.cognito_userpool.id}"
    environment = "${var.environment}"
    metaDataUrl = "${var.saml_metadataurl}"

  }
}

resource "null_resource" "create_cognito_userpool_federation" {
  triggers {
    cognito_user_pool_id = "${aws_cognito_user_pool.cognito_userpool.id}"
  }

  provisioner "local-exec" {
    command = "aws cognito-idp create-identity-provider --region ${var.aws_region} --cli-input-json '${data.template_file.cognito_userpool_template.rendered}'"
  }
}

pawelsocha:
@skoop22 Thanks for your research. 
Now I thinking how to solve this problem. The first thing is filter schema attributes in the aws_cognito_user_pool resource. Second, add identity provider directly to the aws_cognito_user_pool resource.

But, I'm not a terraform developer.

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return nil
+}
+
+func resourceAwsCognitoIdentityProviderUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+	log.Print("[DEBUG] Updating Cognito Identity Provider")
+
+	params := &cognitoidentityprovider.UpdateIdentityProviderInput{
+		UserPoolId:   aws.String(d.Get("UserPoolId").(string)),
```
ldenman: Just wondering... is "d.Get("UserPoolId") correct? Or should it be: d.Get("user_pool_id")?

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,180 @@
...
+			},
+
+			"provider_type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"user_pool_id": {
+				Type:     schema.TypeString,
+				Required: true,
```
ldenman: Should user_pool_id be use ForceNew? If the pool id changes, wouldn't a new provider need to be created?

aws/resource_aws_cognito_identity_provider.go
pawelsocha: Yes, I will change it.

aws/resource_aws_cognito_identity_provider.go
pawelsocha: 👍

skoop22:
Any news on this one I want to remove my workaround :)

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"provider_details": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Computed: true,
```
bflad: Not sure if this was intentional, but it does not seem like we should be ignoring this from the API if its not in the Terraform configuration. I'll double check from the acceptance testing. Also, the API/SDK documentation seems to imply this should be `Required: true`.

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+	if v, ok := d.GetOk("idp_identifiers"); ok {
+		params.IdpIdentifiers = expandStringList(v.([]interface{}))
+	}
+
+	_, err := conn.CreateIdentityProvider(params)
+	if err != nil {
+		return fmt.Errorf("Error creating Cognito Identity Provider: %s", err)
+	}
+
+	d.SetId(*name)
```
bflad: In order to easily support import, let's prepend the resource ID with the user pool ID:

```go
d.SetId(fmt.Sprintf("%s:%s", userPoolID, providerName))
```

Then with configuring the passthrough `Importer` in the resource, we can support import via `terraform import aws_cognito_identity_provider.example xxx_yyyyyy:example`

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+	conn := meta.(*AWSClient).cognitoidpconn
+	log.Printf("[DEBUG] Reading Cognito Identity Provider: %s", d.Id())
+
+	ret, err := conn.DescribeIdentityProvider(&cognitoidentityprovider.DescribeIdentityProviderInput{
+		ProviderName: aws.String(d.Id()),
+		UserPoolId:   aws.String(d.Get("user_pool_id").(string)),
+	})
+
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" {
```
bflad: Minor nitpick: we have a helper function to simplify this logic: `isAWSErr(err, cognitoidentity.ErrCodeResourceNotFoundException, "")`

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+			d.SetId("")
```
bflad: When removing a resource from the state, we prefer to provide a warning log, e.g.

```go
log.Printf("[WARN] Cognito Identity Provider %q not found, removing from state", d.Id())
```

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil
+		}
+		return err
+	}
+
+	ip := ret.IdentityProvider
```
bflad: We should perform a `nil` check here in the unlikely scenario that the API does not error and also omits this attribute.

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+	d.Set("provider_name", ip.ProviderName)
+	d.Set("provider_type", ip.ProviderType)
+	d.Set("user_pool_id", ip.UserPoolId)
+
+	if err := d.Set("attribute_mapping", flattenCognitoIdentityProviderMap(ip.AttributeMapping)); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting attribute_mapping error: %#v", err)
```
bflad: Minor nitpick: these error messages will be returned to the operator, so they are not `DEBUG` and should return the normal error string via `%s`

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return resourceAwsCognitoIdentityProviderRead(d, meta)
+}
+
+func resourceAwsCognitoIdentityProviderDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+	log.Printf("[DEBUG] Deleting Cognito Identity Provider: %s", d.Id())
+
+	return resource.Retry(d.Timeout(schema.TimeoutDelete), func() *resource.RetryError {
```
bflad: This retry loop looks extraneous as there is no `return resource.RetryableError()`. It should be removed along with the customizable deletion timeout.

aws/resource_aws_cognito_identity_provider.go
```diff
@@ -0,0 +1,181 @@
...
+		_, err := conn.DeleteIdentityProvider(&cognitoidentityprovider.DeleteIdentityProviderInput{
+			ProviderName: aws.String(d.Id()),
+			UserPoolId:   aws.String(d.Get("user_pool_id").(string)),
+		})
+
+		if err == nil {
+			d.SetId("")
```
bflad: Minor nitpick: `d.SetId("")` is extraneous in delete functions.

aws/resource_aws_cognito_identity_provider_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func TestAccAWSCognitoIdentityProvider_basic(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSCognitoIdentityProviderDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSCognitoIdentityProviderConfig_basic(),
+				Check: resource.ComposeAggregateTestCheckFunc(
```
bflad: This is missing an `testAccCheckAWSCognitoIdentityProviderExists()` function, which we usually call first in a set of checks to help determine if the physical resource exists in the API as well as pass back the API response object back as a variable we can later reference in other checks.

aws/resource_aws_cognito_identity_provider_test.go
```diff
@@ -0,0 +1,95 @@
...
+				return nil
+			}
+			return err
+		}
+	}
+
+	return nil
+}
+
+func testAccAWSCognitoIdentityProviderConfig_basic() string {
```
bflad: Minor nitpick: the formatting is fairly inconsistent throughout this test configuration which makes it harder to read

aws/structure.go
```diff
@@ -3035,6 +3035,23 @@ func flattenCognitoIdentityPoolRoles(config map[string]*string) map[string]strin
...
+func expandCognitoIdentityProviderMap(config map[string]interface{}) map[string]*string {
+	m := map[string]*string{}
+	for k, v := range config {
+		s := v.(string)
+		m[k] = &s
+	}
+	return m
+}
+
+func flattenCognitoIdentityProviderMap(config map[string]*string) map[string]string {
```
bflad: The AWS SDK already provides a helper function to perform this: `aws.StringValueMap()`

aws/structure.go
```diff
@@ -3035,6 +3035,23 @@ func flattenCognitoIdentityPoolRoles(config map[string]*string) map[string]strin
 	return m
 }
 
+func expandCognitoIdentityProviderMap(config map[string]interface{}) map[string]*string {
```
bflad: We already have a (admittedly hard to find) helper function to perform this already: `stringMapToPointers()`

website/docs/r/cognito_identity_provider.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+
+Provides a Cognito User Identity Provider resource.
+
+## Example Usage
+
+### Basic configuration
+
+```hcl
+resource "aws_cognito_user_pool" "example" {
+  name 						= "example-pool"
```
bflad: Minor nitpick: the formatting is slightly off in the example configuration here and below with email

bflad:
After implementing the above feedback, I was able to get the acceptance testing working by ignoring the `identities` schema attribute in the user pool resource -- I believe this will be the correct path forward rather than forcing everyone to define that attribute in their schema if they also implement this resource.

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSCognitoIdentityProvider_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoIdentityProvider_basic -timeout 120m
=== RUN   TestAccAWSCognitoIdentityProvider_basic
--- PASS: TestAccAWSCognitoIdentityProvider_basic (16.53s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	16.567s
```

bflad:

pawelsocha:
@bflad thanks for your review. it's lots of knowledge for me.

PR #3607, opened 2018-03-02
Amend the RDS cluster docs in reference to snapshot_identifier

HistoireDeBabar:
Hey 👋 

I recently tried to create an RDS cluster from a snapshot using the identifier/name (e.g. `rds:production-2015-06-26-06-05`), and received an error message saying a snapshot ARN was required to create an RDS cluster from a snapshot.

I switched to an ARN, and it all worked as expected. With a `db_instance` a name was sufficient.  Anyways, I thought I'd submit this to try and make it a little clearer for the next reader.  fyi, I'm on Terraform v0.11.3.

Cheers!

### DIFF
--- a/website/docs/r/rds_cluster.html.markdown
+++ b/website/docs/r/rds_cluster.html.markdown
@@ -102,7 +102,7 @@ Default: A 30-minute window selected at random from an 8-hour block of time per 
 * `port` - (Optional) The port on which the DB accepts connections
 * `vpc_security_group_ids` - (Optional) List of VPC security groups to associate
   with the Cluster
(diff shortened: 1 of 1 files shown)


bflad:
Hi @HistoireDeBabar! 👋  It turns out, both are actually supported according to the SDK depending on your situation. 😄 

https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#RestoreDBClusterFromSnapshotInput
```
    // You can use either the name or the Amazon Resource Name (ARN) to specify
    // a DB cluster snapshot. However, you can use only the ARN to specify a DB
    // snapshot.
```

What do you think, maybe we should update this to the SDK's wording? Better yet, we technically could add plan-time validation for this as well (maybe worth its own issue).

HistoireDeBabar:
hey @bflad ah that makes sense! Yeah, I'll update it, thanks for checking it out 👍

PR #3609, opened 2018-03-02
resource/aws_lb: enable_http2 flag for LBs and small refactor

mpilar:
Closes: #3511

- refactored to use a switch statement a la #3578
- added enable_http2 as a flag with tests and docs
- made the boolean flag tests have 3 steps: on, off, on. to test creation, removal and enabling of a running LB
- increased wait time to wait for network interfaces to be destroyed after deleting an NLB (this caused the EIP hanging resource I saw yesterday)
- made all test fixtures use lb_test as the name for the aws_lb (to prevent the typo issue I had)

Note about the last two: The fail in acceptance tests seemed to happen more with cross-zone enabled, it's not an issue of the acceptance test but the speed at which AWS can successfully destroy the resources related to the NLB, seems that cross-zone makes that destruction slower somewhat.

Test results:
```
$ TF_ACC=1 go test ./aws -v -run=TestAccAWSLB_ -timeout 120m
=== RUN   TestAccAWSLB_basic
--- PASS: TestAccAWSLB_basic (232.18s)
=== RUN   TestAccAWSLB_networkLoadbalancerBasic
--- PASS: TestAccAWSLB_networkLoadbalancerBasic (264.53s)
=== RUN   TestAccAWSLB_networkLoadbalancerEIP
--- PASS: TestAccAWSLB_networkLoadbalancerEIP (270.10s)
=== RUN   TestAccAWSLB_generatedName
--- PASS: TestAccAWSLB_generatedName (252.17s)
=== RUN   TestAccAWSLB_generatesNameForZeroValue
--- PASS: TestAccAWSLB_generatesNameForZeroValue (232.66s)
=== RUN   TestAccAWSLB_namePrefix
--- PASS: TestAccAWSLB_namePrefix (260.69s)
=== RUN   TestAccAWSLB_tags
--- PASS: TestAccAWSLB_tags (297.39s)
=== RUN   TestAccAWSLB_networkLoadbalancer_updateCrossZone
--- PASS: TestAccAWSLB_networkLoadbalancer_updateCrossZone (370.76s)
=== RUN   TestAccAWSLB_networkLoadbalancer_updateHttp2
--- PASS: TestAccAWSLB_networkLoadbalancer_updateHttp2 (368.44s)
=== RUN   TestAccAWSLB_updatedSecurityGroups
--- PASS: TestAccAWSLB_updatedSecurityGroups (318.00s)
=== RUN   TestAccAWSLB_updatedSubnets
--- PASS: TestAccAWSLB_updatedSubnets (318.37s)
=== RUN   TestAccAWSLB_updatedIpAddressType
--- PASS: TestAccAWSLB_updatedIpAddressType (318.56s)
=== RUN   TestAccAWSLB_noSecurityGroup
--- PASS: TestAccAWSLB_noSecurityGroup (227.11s)
=== RUN   TestAccAWSLB_accesslogs
--- PASS: TestAccAWSLB_accesslogs (441.91s)
=== RUN   TestAccAWSLB_networkLoadbalancer_subnet_change
--- PASS: TestAccAWSLB_networkLoadbalancer_subnet_change (500.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	4673.668s
```

### DIFF
--- a/aws/resource_aws_internet_gateway.go
+++ b/aws/resource_aws_internet_gateway.go
@@ -236,12 +236,13 @@ func resourceAwsInternetGatewayDetach(d *schema.ResourceData, meta interface{}) 
 	// Wait for it to be fully detached before continuing
 	log.Printf("[DEBUG] Waiting for internet gateway (%s) to detach", d.Id())
 	stateConf := &resource.StateChangeConf{
-		Pending:        []string{"detaching"},
-		Target:         []string{"detached"},
-		Refresh:        detachIGStateRefreshFunc(conn, d.Id(), vpcID.(string)),
-		Timeout:        15 * time.Minute,
-		Delay:          10 * time.Second,
-		NotFoundChecks: 30,
+		Pending:                   []string{"detaching"},
+		Target:                    []string{"detached"},
+		Refresh:                   detachIGStateRefreshFunc(conn, d.Id(), vpcID.(string)),
+		Timeout:                   15 * time.Minute,
+		Delay:                     20 * time.Second,
+		NotFoundChecks:            30,
+		ContinuousTargetOccurence: 2,
 	}
 	if _, err := stateConf.WaitForState(); err != nil {
 		return fmt.Errorf(
--- a/aws/resource_aws_internet_gateway.go
+++ b/aws/resource_aws_internet_gateway.go
@@ -236,13 +236,12 @@ func resourceAwsInternetGatewayDetach(d *schema.ResourceData, meta interface{}) 
 	// Wait for it to be fully detached before continuing
 	log.Printf("[DEBUG] Waiting for internet gateway (%s) to detach", d.Id())
 	stateConf := &resource.StateChangeConf{
-		Pending:                   []string{"detaching"},
-		Target:                    []string{"detached"},
-		Refresh:                   detachIGStateRefreshFunc(conn, d.Id(), vpcID.(string)),
-		Timeout:                   15 * time.Minute,
-		Delay:                     20 * time.Second,
-		NotFoundChecks:            30,
-		ContinuousTargetOccurence: 2,
+		Pending:        []string{"detaching"},
+		Target:         []string{"detached"},
+		Refresh:        detachIGStateRefreshFunc(conn, d.Id(), vpcID.(string)),
+		Timeout:        15 * time.Minute,
+		Delay:          10 * time.Second,
+		NotFoundChecks: 30,
 	}
 	if _, err := stateConf.WaitForState(); err != nil {
 		return fmt.Errorf(
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -163,6 +163,12 @@ func resourceAwsLb() *schema.Resource {
 				Default:  false,
 			},
 
+			"enable_http2": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
 			"ip_address_type": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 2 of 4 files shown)


aws/resource_aws_lb_test.go
```diff
@@ -1026,7 +1139,7 @@ resource "aws_lb" "test" {
 }
 
 resource "aws_eip" "lb" {
-  count = "${length(data.aws_availability_zones.available.names)}"
+  count = "${length(data.aws_availability_zones.available.names)>2?2:length(data.aws_availability_zones.available.names)}"
```
bflad: Are there any regions with only 1 AZ? Seems like this should just be `count = 2`

aws/resource_aws_lb_test.go
```diff
@@ -874,6 +919,75 @@ resource "aws_security_group" "alb_test" {
 }`, lbName)
 }
 
+func testAccAWSLBConfig_basicWithHttp2Toggle(lbName string, http2 bool) string {
```
bflad: Minor nitpick: generally we would just name this `testAccAWSLBConfig_enableHttp2`

aws/resource_aws_lb.go
```diff
@@ -324,56 +330,62 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
...
+			})
+		}
+	case "network":
+		if d.HasChange("enable_cross_zone_load_balancing") {
+			attributes = append(attributes, &elbv2.LoadBalancerAttribute{
+				Key:   aws.String("load_balancing.cross_zone.enabled"),
+				Value: aws.String(fmt.Sprintf("%t", d.Get("enable_cross_zone_load_balancing").(bool))),
+			})
+		}
+	default:
```
bflad: When would we hit this scenario? This currently will never hit. Also, I'm not sure the benefit of postulating if/when there may be new load balancer types and having any sort of implicit support of them. We should force a provider update with acceptance testing.

aws/resource_aws_lb.go
```diff
@@ -324,56 +330,62 @@ func resourceAwsLbUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		if d.HasChange("enable_deletion_protection") {
```
bflad: Given the note above about `default:`, shouldn't this be a part of both application and network load balancers? It can be moved outside the switch statement. Also, this likely would've been caught in an acceptance test, which looks like it should be added to enable and then disable this (so it can actually be destroyed).

aws/resource_aws_internet_gateway.go
```diff
@@ -236,12 +236,13 @@ func resourceAwsInternetGatewayDetach(d *schema.ResourceData, meta interface{})
...
-		Timeout:        15 * time.Minute,
-		Delay:          10 * time.Second,
-		NotFoundChecks: 30,
+		Pending:                   []string{"detaching"},
+		Target:                    []string{"detached"},
+		Refresh:                   detachIGStateRefreshFunc(conn, d.Id(), vpcID.(string)),
+		Timeout:                   15 * time.Minute,
+		Delay:                     20 * time.Second,
+		NotFoundChecks:            30,
+		ContinuousTargetOccurence: 2,
```
bflad: We should probably PR and acceptance test this separately 😄

aws/resource_aws_lb.go
mpilar: I will add an acceptance test.

aws/resource_aws_internet_gateway.go
mpilar: I will remove it for now.

aws/resource_aws_lb.go
appilon: This was a mistake in my PR (which was coded offline on plane), I forgot Go's switch doesn't fallthrough by default

mpilar:
```
=== RUN   TestAccAWSLB_applicationLoadBalancer_updateDeletionProtection
--- PASS: TestAccAWSLB_applicationLoadBalancer_updateDeletionProtection (373.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	373.854s
```
After last commit.

mpilar:
@appilon I would love a review from you as well.

appilon:
Looks good to me! (I'm still new to the project and can only really comment on general code/go things)

bflad:

PR #3613, opened 2018-03-03
Write partial state on rest_api create/update failures

lukehoban:
It is possible for the API Gateway RestAPI to create successfully, but for the `PutRestApi` call to fail.  This change ensures that the partial state is written out even in case of this failure so that the creation of the RestAPI is recorded.

The new test validates error paths.  I'm not sure whether it is possible to explicitly test that the partial state was written.

```
$ go test -v -timeout 1h github.com/terraform-providers/terraform-provider-aws/aws -run ^TestAccAWSAPIGatewayRestApi
=== RUN   TestAccAWSAPIGatewayRestApi_basic
--- PASS: TestAccAWSAPIGatewayRestApi_basic (21.68s)
=== RUN   TestAccAWSAPIGatewayRestApi_openapi
--- PASS: TestAccAWSAPIGatewayRestApi_openapi (35.18s)
=== RUN   TestAccAWSAPIGatewayRestApi_openapi_partial
--- PASS: TestAccAWSAPIGatewayRestApi_openapi_partial (29.98s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	86.884s
```

### DIFF
--- a/aws/resource_aws_api_gateway_rest_api.go
+++ b/aws/resource_aws_api_gateway_rest_api.go
@@ -81,6 +81,10 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(*gateway.Id)
 
+	d.Partial(true)
+	d.SetPartial("description")
+	d.SetPartial("binary_media_types")
+
 	if body, ok := d.GetOk("body"); ok {
 		log.Printf("[DEBUG] Initializing API Gateway from OpenAPI spec %s", d.Id())
 		_, err := conn.PutRestApi(&apigateway.PutRestApiInput{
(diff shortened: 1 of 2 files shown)


bflad:
Hi @lukehoban 👋 Thank you for submitting this and apologies about the lack of review sooner.

In practice, we have found `d.SetPartial()` behavior to be sporadically implemented (usually in older Terraform resources) and its functionality to be of limited use. We are beginning to shy away from implementing it in new logic and likely will start phasing its usage out of existing logic.

When encountering resource errors, by design Terraform will prevent dependent resources from completing actions because the upstream resource has entered an unknown state. Without fully refreshing the Terraform state (e.g. the `Read` function of a resource), we cannot be fully assured that previous API action(s) performed the intended action(s), which would be represented by the API state matching the configuration state. While its _likely_ those previously intended actions would represent the correct resource state for the successful portion of multiple updates (since we do fairly comprehensive acceptance testing of the resource code), the lack of a 100% guarantee in these situations generally means we should err on the side of uncertainty.

Given this, we are going to opt to close this PR. Thanks again for contributing this.

Issue #1681, opened 2017-09-15
data.aws_ecs_service should return running task definition revision


joeig:
I deploy new Docker containers to ECS using one task definition per container release (this is usually invoked by a CI job). Since `aws_ecs_service` requires a valid task definition name containing the revision ID, which is increased by the CI job, it is not possible to dynamically adapt the `aws_ecs_service` resource to the latest running task definition.

Wouldn't it be awesome if there was a data source for `aws_ecs_service` which is able to detect the running task definition revision?

```
data "aws_ecs_service" "running_voting" {
  cluster_name = "mycluster"
  service_name = "voting"
}

resource "aws_ecs_service" "voting" {
  name            = "voting"
  task_definition = "${data.aws_ecs_service.running_voting.arn}"
  desired_count   = 3
}
```

Terraform v0.10.4

joeig:
One possible solution is to use `aws_ecs_task_definition` for that, but this just returns the latest active revision and not the actually running revision.

bflad:
The new `aws_ecs_service` data source has been merged and will release with version 1.22.0 of the AWS provider, likely sometime middle of next week.

bflad:
This has been released in [version 1.22.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1220-june-05-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

n1te1337:
@joeig isn't this essentially creating a circular dependency and prohibits the service from being created in the first place?

PR #3617, opened 2018-03-05
New Data Source: ECS Service

elbuo8:
Closes https://github.com/terraform-providers/terraform-provider-aws/issues/1681

### DIFF
--- a/aws/data_source_aws_ecs_service.go
+++ b/aws/data_source_aws_ecs_service.go
@@ -1,0 +1,83 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsEcsService() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsEcsServiceRead,
+
+		Schema: map[string]*schema.Schema{
+			"service_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"cluster_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"desired_count": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+			"launch_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"task_definition": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ecsconn
+
+	params := &ecs.DescribeServicesInput{
+		Cluster:  aws.String(d.Get("cluster_arn").(string)),
+		Services: []*string{aws.String(d.Get("service_name").(string))},
+	}
+
+	log.Printf("[DEBUG] Reading ECS Service: %s", params)
+	desc, err := conn.DescribeServices(params)
+
+	if err != nil {
+		return err
+	}
+
+	for _, service := range desc.Services {
+		if aws.StringValue(service.ClusterArn) != d.Get("cluster_arn").(string) {
+			continue
+		}
+		if aws.StringValue(service.ServiceName) != d.Get("service_name").(string) {
+			continue
+		}
+		d.SetId(aws.StringValue(service.ServiceArn))
+		d.Set("service_name", service.ServiceName)
+		d.Set("arn", service.ServiceArn)
+		d.Set("cluster_arn", service.ClusterArn)
+		d.Set("desired_count", service.DesiredCount)
+		d.Set("launch_type", service.LaunchType)
+		d.Set("task_definition", service.TaskDefinition)
+	}
+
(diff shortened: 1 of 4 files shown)


kwyjibos:
Would be very useful if we could get this reviewed/merged

n1te1337:
Is there any help need to get this across the line?

aws/data_source_aws_ecs_service.go
```diff
@@ -0,0 +1,83 @@
...
+				Computed: true,
+			},
+			"cluster_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"desired_count": {
+				Type:     schema.TypeInt,
+				Optional: true,
```
bflad: This attribute and the two below are not configurable (`Optional: true`) but rather read only and meant to just be set in the Terraform state (`Computed: true`). https://www.terraform.io/docs/extend/schemas/schema-behaviors.html

aws/data_source_aws_ecs_service.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	log.Printf("[DEBUG] Reading ECS Service: %s", params)
+	desc, err := conn.DescribeServices(params)
+
+	if err != nil {
+		return err
+	}
+
+	for _, service := range desc.Services {
```
bflad: We generally prefer to check and immediately return errors for zero returned items or more than one returned item instead of using `for` loops in singular data sources. e.g.

```go
if len(desc.Services) == 0 {
  return fmt.Errorf("service with name %q in cluster %q not found", serviceName, clusterArn)
}

if len(desc.Services) > 1 {
  return fmt.Errorf("multiple services with name %q found in cluster %q", serviceName, clusterArn)
}

service := desc.Service[0]
d.SetId(aws.StringValue(service.ServiceArn))
// ...
```

aws/data_source_aws_ecs_service_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func TestAccAWSEcsDataSource_ecsService(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckAwsEcsServiceDataSourceConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.aws_ecs_service.default", "service_name", "mongodb"),
```
bflad: Nitpick: We can use [`resource.TestCheckResourceAttrPair()`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#TestCheckResourceAttrPair) to ensure the resource and data source attributes match as expected.

website/docs/d/ecs_service.html.markdown
```diff
@@ -0,0 +1,65 @@
+---
+layout: "aws"
+page_title: "AWS: aws_ecs_service"
+sidebar_current: "docs-aws-datasource-ecs-service"
```
bflad: This documentation page is missing a sidebar link in `website/aws.erb`

website/docs/d/ecs_service.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+
+## Example Usage
+
+```hcl
+data "aws_ecs_service" "ecs-mongo" {
+  service_name = "${aws_ecs_service.mongo.name}"
+  cluster_arn  = "${aws_ecs_cluster.foo.arn}"
+}
+
+resource "aws_ecs_cluster" "foo" {
```
bflad: Normally we try to not show an unexpected configuration for newer operators (e.g. the data source directly referencing its resource counterpart) -- in this case its fine to simply just list the example as:

```hcl
data "aws_ecs_service" "example" {
   cluster_arn  = "${data.aws_ecs_cluster.example.arn}"
   service_name = "example"
 }
```

bflad:
🚀 

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSEcsServiceDataSource_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEcsServiceDataSource_basic -timeout 120m
=== RUN   TestAccAWSEcsServiceDataSource_basic
--- PASS: TestAccAWSEcsServiceDataSource_basic (34.14s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	34.179s
```

bflad:
This has been released in [version 1.22.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1220-june-05-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

n1te1337:
@elbuo8 RE: #1681 I was really excited about this PR hoping it would solve my CI/CD issues but now when I think about it, it creates a different issue. 

By using it like so:
```
data "aws_ecs_service" "main" {
  cluster_name = "mycluster"
  service_name = "voting"
}

resource "aws_ecs_service" "main" {
  name            = "voting"
  task_definition = "${data.aws_ecs_service.main.task_definition}"
  desired_count   = 1
}
```
You solve the issue where the CI/CD updates the task and terraform is still in sync because the data source will pull the currently active task revision for the service (which is great). 

Unfortunately, by doing so you also create a dependency on the service itself because you expect the service to be up and running. This pretty much prohibits the use of the data source for new services, the only way to do it is to create a task and a service manually and only after that change it to use the data source (so you can't really automate it for infrastructure deployments).

Thoughts?

PR #3620, opened 2018-03-05
handle GetBucketWebsite <NotImplemented> errors gracefully

### LINKED ISSUE #3603: CreateBucket fails on non-aws endpoint due to non implemented feature
### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

Terraform version: 0.11.3  

### Affected Resource(s)
- aws_s3_bucket

Terraform template:

https://gist.github.com/tlawrence/9ce8f48af687630ff44029369e733313

 

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior

Bucket should create

### Actual Behavior

GetBucketWebsite returns 501 "NotImplemented" error. AWS-SDK retries 10 times then fails. 
501 should be handled gracefully

### Steps to Reproduce

1. `terraform apply`

### Important Factoids
Object storage platform is EMC ECS

tried to update the code in aws/resource_aws_s3_bucket.go to catch the error but think it may need catching in the SDK first?

tlawrence:
Fixes #3603

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -708,6 +708,15 @@ func resourceAwsS3BucketRead(d *schema.ResourceData, meta interface{}) error {
 		})
 	})
 	ws := wsResponse.(*s3.GetBucketWebsiteOutput)
+
+	if err != nil {
+		if !isAWSErr(err, "NotImplemented", "") {
+			return err
+		}
+		log.Printf("[WARN] S3 bucket: %s, website configuration not supported by storage server.", d.Id())
+		return nil
+	}
+
 	var websites []map[string]interface{}
 	if err == nil {
 		w := make(map[string]interface{})
--- a/aws/config.go
+++ b/aws/config.go
@@ -468,6 +468,13 @@ func (c *Config) Client() (interface{}, error) {
 	client.mediastoreconn = mediastore.New(sess)
 	client.appsyncconn = appsync.New(sess)
 
+	// Workaround for https://github.com/aws/aws-sdk-go/issues/1819
+	client.s3conn.Handlers.Retry.PushBack(func(r *request.Request) {
+		if isAWSErr(r.Error, "NotImplemented", "") {
+			r.Retryable = aws.Bool(false)
+		}
+	})
+
 	// Workaround for https://github.com/aws/aws-sdk-go/issues/1376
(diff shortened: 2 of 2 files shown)


tlawrence:
This PR broke creation of native AWS S3 buckets. commit a900f5f resolves this. 
We have now tested against both S3 & non-S3 endpoints.

aws/resource_aws_s3_bucket.go
```diff
@@ -708,6 +708,15 @@ func resourceAwsS3BucketRead(d *schema.ResourceData, meta interface{}) error {
 		})
 	})
 	ws := wsResponse.(*s3.GetBucketWebsiteOutput)
+
+	if err != nil {
+		if !isAWSErr(err, "NotImplemented", "") && !isAWSErr(err, "NoSuchWebsiteConfiguration", "") {
+			return err
+		}
+		log.Printf("[WARN] S3 bucket: %s, website configuration not supported by storage server.", d.Id())
+		return nil
```
bflad: Returning here will prevent the rest of the Read function from working (versioning, acceleration, request payer, logging, lifecycle rules, etc.)

bflad: The line shouldn't be present at all 😄

tlawrence: surely if we remove it all together then the error object will be passed back and cause an exception?

bflad: What error are you seeing? I briefly looked at the code below this and it seemed fine.

tlawrence: Oops, I thought I was returning from the 'if' rather than the function.
I think rather than this I probably need to reset the error object to nil to prevent it being caught later on. 

so rather than `return nil`
we should have `err = nil`

Is that correct?

bflad: You could do that, but its unnecessary. The line does not need to be there at all. The function will continue just fine since line 721 already does the right thing and `err` is overwritten at line 774.

Issue #513, opened 2017-06-13
Support Kinesis Firehose Data Transformation


hashibot:
_This issue was originally opened by @mlimaloureiro as hashicorp/terraform#11723. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi Guys,

I would like to request Terraform to support `Kinesis Firehose` data transformation.

More info here

https://aws.amazon.com/about-aws/whats-new/2016/12/amazon-kinesis-firehose-can-now-prepare-and-transform-streaming-data-before-loading-it-to-data-stores/ 

Thanks!

saiaman:
Even with extended_s3_configuration, we shall have processing_configuration in each destination possibility ( elasticsearch / redshift / s3 ) .

http://docs.aws.amazon.com/firehose/latest/APIReference/API_CreateDeliveryStream.html

CannibalVox:
This is kind of messed up atm because redshift doesn't support extended_s3_configuration, only s3_configuration.

joeig:
The Elasticsearch destination requires the legacy s3_configuration as well.
Compared with the AWS CLI's "aws firehose describe-delivery-stream", the processing configuration should apply to the delivery stream instead of the "extended_s3_configuration".

cgarvis:
Would love to have this for my ElasticSearch firehose.

kchristensen:
Having to manually enable data transformation on our redshift firehoses is a bummer, I'd love to see support for extended_s3_configuration in Redshift firehoses.

ligustah:
I'd also like to see this. From the [API](http://docs.aws.amazon.com/firehose/latest/APIReference/API_RedshiftDestinationConfiguration.html) it appears to be set on the redshift destination configuration, so it should be completely independent of any s3 config.

jseiser:
I'm reading through the docs, and this issue, and I'm still not clear.  I have an elastic search firehose, that I need to recreate with terraform that uses a lambda to convert syslogs to json. 

Is there no way to do this in terraform?

jseiser:
OK, confirming that this doesnt work.


>  When destination is elasticsearch, s3_configuration is required

lhsoeren:
[...]

PR #3621, opened 2018-03-05
Adding processing_configuration to elasticsearch section

jamesrenfro:
This is a partial fix for the issue listed below. The problem currently is
that you can add an extended_s3_configuration to the
aws_kinesis_firehose_delivery_stream and set the destination to 'extended_s3'
if you want data transformation to be enabled for S3, but you can't do the
same if your destination is 'elastisearch' - this commit fixes that.

Support Kinesis Firehose Data Transformation #513
https://github.com/terraform-providers/terraform-provider-aws/issues/513

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ BUG FIXES:
 
 * resource/aws_api_gateway_vpc_link: Ensure `target_arns` is properly read [GH-3569]
 * resource/aws_spot_instance_request: Retry for 1 minute instead of 15 seconds for IAM eventual consistency [GH-3561]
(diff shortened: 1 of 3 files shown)


bflad:
Can you please add an acceptance test that exercises this new code? You can take a look at #3944 for inspiration

bflad:
Actually do not worry about the acceptance testing, I will add it before I merge, as I would like to submit a followup PR to prevent some crashes that will interfere with the implementation of this PR as-is.

jamesrenfro:
Ok - thanks very much!

bflad:

PR #3639, opened 2018-03-06
Add configurable timeouts to aws_route

domdom82:
Turns out the 2 minutes timeout for route creation is pretty tight as we are hitting it quite a lot lately.

Instead of bumping it I just made it configurable for others.

Cheers!


bflad:

PR #3642, opened 2018-03-06
resource/aws_api_gateway_rest_api: Add support for content encoding

### LINKED ISSUE #3144: [Feature Request] Support for Content Encoding in API Gateway
Hi there,

AWS recently announced the possibility of configuring API Gateway to allow clients to request content to be compressed before being sent back in the response - better late than never :)

Reference: https://aws.amazon.com/about-aws/whats-new/2017/12/amazon-api-gateway-supports-content-encoding-for-api-responses/

This would probably need a separate `aws_api_gateway_rest_api_settings` resource (similar to `aws_api_gateway_method_settings`) OR two configurations on the `aws_api_gateway_rest_api` resource:

- `content_encoding_enabled` - Flag to enable/disable
- `minimum_compression_size` - Minimum body size required for compression

appilon:
Fix #3144 

It appears there is a bug with the method `GetOkExists` which prevents it from being useful in this situation (can't detect set --> unset/disable in config). Regardless it was decided using a special value of `-1` keeps the code to 1 variable, and makes the implementation quite clean.

~~The tests finally pass by manually waiting and then reading the apigw. I imagine there might be a more "official" way of getting around the eventual consistency.~~ No eventual consistency issues

### DIFF
--- a/aws/resource_aws_api_gateway_rest_api.go
+++ b/aws/resource_aws_api_gateway_rest_api.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -42,6 +43,13 @@ func resourceAwsApiGatewayRestApi() *schema.Resource {
 				Optional: true,
 			},
 
+			"minimum_compression_size": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      -1,
+				ValidateFunc: validateIntegerInRange(-1, 10485760),
+			},
+
 			"root_resource_id": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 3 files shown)


radeksimko:
> however I found ResourceData could not distinguish set from unset (after a value has been set). I imagine with more internal knowledge there might have been a way to pull it off. Overall I think I prefer the flag being user driven anyways, the implementation feels simpler and the experience is much more similar to that of the AWS web console (where you toggle content_encoding_enabled and then set a minimum size). A minimum size of 0 is sane imo too it just means all content is compressed.

Generally speaking we always tend to stick to API structure and naming conventions if possible and avoid building any abstractions above it as we burnt our fingers by doing so in the past where AWS changes API in a way that it broke our assumptions about such abstractions.

We should be able to implement this without having to introduce a new field. Before `d.GetOkExists` was a thing I'd recommend using `-1` as "unset" placeholder in this context, but we shouldn't even need to do that.

`d.GetOkExists("minimum_compression_size")` should do the job. I assume it may not be working as expected because the field has `Default` set? Either way we don't usually set inherent (like empty string for `TypeString`, zero for `TypeInt` etc.) defaults in the schema. `Default` is there to avoid diffs when the API sets some defaults automatically after creation, typically non-zero value for TypeInt.

Let me know if you need any help in debugging this.

appilon:
@radeksimko When I switch my acceptance test to wait a few seconds and manually read the api, the data is validated (suspected eventual consistency)

aws/resource_aws_api_gateway_rest_api.go
```diff
@@ -74,6 +82,10 @@ func resourceAwsApiGatewayRestApiCreate(d *schema.ResourceData, meta interface{}
 		params.BinaryMediaTypes = expandStringList(binaryMediaTypes.([]interface{}))
 	}
 
+	if minimumCompressionSize := d.Get("minimum_compression_size").(int); minimumCompressionSize > -1 {
```
radeksimko: Nitpick: Do you mind putting the assignment on a separate line above as we're not checking result of casting? (which is AFAIK the common and often only reasonable use-case for these if one-liners)

bflad:

PR #3655, opened 2018-03-07
Fix filename typo in lambda resource

ldenman:


loivis:
I think it's not a typo. `filename` is the path to your local package, so it could be anything valid path string.
Even if we want to change, something like `filename = "path/to/your/local/package.zip"` makes more sense?

ldenman:
@loivis Okay, that's fine. I assumed the examples would be able to apply.

radeksimko:
@loivis is correct.

Thanks @ldenman for the PR, but there's currently nothing to change/fix, so I'm going to close it.

PR #3657, opened 2018-03-07
data-source/iam_server_certificate: drop custom ValidateFunc for name and name_prefix

loivis:

This PR is for:

- [x] data_source_aws_iam_server_certificate

### DIFF
--- a/aws/data_source_aws_iam_server_certificate.go
+++ b/aws/data_source_aws_iam_server_certificate.go
@@ -24,28 +24,15 @@ func dataSourceAwsIAMServerCertificate() *schema.Resource {
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"name_prefix"},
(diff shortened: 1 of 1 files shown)


aws/data_source_aws_iam_server_certificate.go
```diff
@@ -24,28 +24,15 @@ func dataSourceAwsIAMServerCertificate() *schema.Resource {
...
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 102 characters, name is limited to 128", k))
-					}
-					return
-				},
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"name"},
+				ValidateFunc:  validateMaxLength(128 - 26),
```
radeksimko: Nice one, I like how you used code to explain the background behind the number! ❤️

PR #3661, opened 2018-03-07
resource/api_gateway_*: update validate functions

loivis:

This PR is for:

- [x] resource/api_gateway_integration
- [x] resource/api_gateway_integration_response
- [x] resource/api_gateway_method
- [x] resource/api_gateway_method_response

### DIFF
--- a/aws/resource_aws_api_gateway_integration.go
+++ b/aws/resource_aws_api_gateway_integration.go
@@ -4,9 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
-	"time"
-
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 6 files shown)


aws/resource_aws_api_gateway_integration.go
```diff
@@ -6,12 +6,14 @@ import (
 	"log"
 	"time"
 
+	"strings"
+
```
radeksimko: Do you mind aligning this import with the group above?

aws/resource_aws_api_gateway_integration.go
```diff
@@ -89,15 +97,19 @@ func resourceAwsApiGatewayIntegration() *schema.Resource {
...
-				ValidateFunc: validateApiGatewayIntegrationPassthroughBehavior,
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"WHEN_NO_MATCH",
+					"WHEN_NO_TEMPLATES",
+					"NEVER",
+				}, false),
```
radeksimko: Is there any particular reason you decided to make this validation inline instead of keeping it decoupled like all the others?

aws/resource_aws_api_gateway_integration.go
loivis: I didn't notice it at all. vscode did it:smile:

aws/resource_aws_api_gateway_integration.go
loivis: My goal is to make validation inline as default in case of functions in `validation` package. The exception is only when it's used exactly the same by multiple arguments, like `validateHTTPMethod()` and `validateApiGatewayIntegrationContentHandling()` in this PR.

radeksimko: Oh, I see - reusability, doh! Obviously - good point! 👍 😄

PR #3666, opened 2018-03-07
resourc/codebuild_project: drop custom ValidateFunc

loivis:

This PR is for:

- [x] resourc/codebuild_project
+ artifacts.namespace_type
+ artifacts.type
+ environment.compute_type
+ environment.type
+ source.auth.type
+ source.type
+ build_timeout

### DIFF
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"regexp"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


bflad:
Looks like the acceptance test needs to be updated for this one:
```
--- FAIL: TestAccAWSCodeBuildProject_sourceAuth (0.87s)
    testing.go:506: Step 0, expected error:
        
        config is invalid: aws_codebuild_project.foo: expected source.0.auth.0.type to be one of [OAUTH], got INVALID
        
        To match:
        
        Source Auth Type can only be
```

loivis:
I hope it should be fixed now. I was not able to test though because I don't have access setup for codebuild.

johnnorton:
This also fixed this error:  "Error: aws_codebuild_project.foo: CodeBuild: Source Type can only be one of: CODEPIPELINE, GITHUB, S3, BITBUCKET, CODECOMMIT", when using GITHUB_ENTERPRISE

PR #3677, opened 2018-03-08
resource/cognito_user_pool_*: drop custom ValidateFuncs

loivis:

This PR is for:

- [x] resource/cognito_user_pool
+ admin_create_user_config.unused_account_validity_days
+ alias_attributes
+ auto_verified_attributes
+ mfa_configuration
+ password_policy.minimum_length
+ verification_message_template.default_email_option

- [x] resource/cognito_user_pool_client
+ refresh_token_validity


Puneeth-n:
@loivis Any idea why changes to `auto_verified_attributes` are not detected during plan and apply? I see that the attribute `AutoVerifiedAttributes` from the `aws describe pool` call is not set when initially, it was set by terraform and a user removed it via the AWS console. Terraform is unable to detect this change. see #3780

PR #3688, opened 2018-03-08
resource/elb: drop custom ValidateFunc for protocol and int

loivis:

This PR is for:

- [x] resource/elb

### DIFF
--- a/aws/resource_aws_elb.go
+++ b/aws/resource_aws_elb.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsElb() *schema.Resource {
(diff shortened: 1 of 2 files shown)


aws/resource_aws_elb.go
```diff
@@ -994,6 +983,15 @@ func isValidProtocol(s string) bool {
 	return true
 }
 
+func validateListenerProtocol() schema.SchemaValidateFunc {
+	return validation.StringInSlice([]string{
+		"HTTP",
+		"HTTPS",
+		"SSL",
+		"TCP",
+	}, false)
```
bflad: This needs to be `true` to accept lowercase values. All the tests failed similarly:

```
=== RUN   TestAccAWSELBAttachment_basic
--- FAIL: TestAccAWSELBAttachment_basic (2.44s)
    testing.go:513: Step 0 error: config is invalid: 2 problems:
        
        - aws_elb.bar: expected listener.0.instance_protocol to be one of [HTTP HTTPS SSL TCP], got http
        - aws_elb.bar: expected listener.0.lb_protocol to be one of [HTTP HTTPS SSL TCP], got http
```

PR #3696, opened 2018-03-09
resource/emr_*: drop custom ValidateFuncs

loivis:

Acceptance tests might fail since I didn't run any of them at all. I'll fix it if you find any error during review.

This PR is for:

- [x] resource/emr_cluster
- [x] resource/emr_instance_group
- [x] resource/emr_instance_security_configuration

### DIFF
--- a/aws/resource_aws_emr_security_configuration.go
+++ b/aws/resource_aws_emr_security_configuration.go
@@ -1,7 +1,6 @@
 package aws
 
 import (
-	"fmt"
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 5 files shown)


aws/resource_aws_emr_security_configuration.go
```diff
@@ -26,27 +25,13 @@ func resourceAwsEMRSecurityConfiguration() *schema.Resource {
...
-					if len(value) > 10000 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 10000 characters, name is limited to 10280", k))
-					}
-					return
-				},
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateMaxLength(10000),
```
paultyng: You mind adding a comment here about the 10280 as the actual limit? And maybe doing it as the subtraction like you have done elsewhere?

loivis: synced with new const 😏

PR #3702, opened 2018-03-09
resource/iam_*: drop custom ValidateFuncs

loivis:

This PR is for:

- [x] resource/iam_policy_attachment
- [x] resource/iam_server_certificate
- [x] resource/iam_user_login_profile
- [x] resource/iam_user_policy
- [x] resource/iam_user_ssh_key
- [x] resource/instance(included here by mistake)

### DIFF
--- a/aws/resource_aws_iam_server_certificate.go
+++ b/aws/resource_aws_iam_server_certificate.go
@@ -62,28 +62,14 @@ func resourceAwsIAMServerCertificate() *schema.Resource {
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"name_prefix"},
-				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
-					value := v.(string)
-					if len(value) > 128 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 128 characters", k))
-					}
-					return
-				},
+				ValidateFunc:  validateMaxLength(128),
 			},
 
 			"name_prefix": {
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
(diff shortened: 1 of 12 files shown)


aws/resource_aws_iam_server_certificate.go
```diff
@@ -62,28 +62,14 @@ func resourceAwsIAMServerCertificate() *schema.Resource {
...
-					if len(value) > 102 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 102 characters, name is limited to 128", k))
-					}
-					return
-				},
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateMaxLength(128 - 26),
```
paultyng: can you add a comment here about why its subtraction?

loivis: synced with the new const :smirk:

aws/resource_aws_launch_configuration.go
```diff
@@ -99,7 +99,7 @@ func resourceAwsLaunchConfiguration() *schema.Resource {
 						return ""
 					}
 				},
-				ValidateFunc: validateInstanceUserDataSize,
+				ValidateFunc: validateMaxLength(16384),
```
paultyng: I feel like this may be problematic if unicode is used (not that it was correct before). If the user data limit is on the actual encoded byte length, not the string character length, it could be exhausted before this limit is hit (sorry I didn't catch this the first time around). Can you confirm if it is actually bytes like the previous error message said, and if so, do you mind possibly adding a validation for an encoded byte length instead of just string length?

paultyng: Nevermind, I was confused between byte length from `len` and rune counts. Assuming user data is in UTF-8, this will be fine.

PR #3707, opened 2018-03-09
resource/key_pair: drop custom ValidateFunc for name_prefix

loivis:

This PR is for:

- [x] resource/key_pair

### DIFF
--- a/aws/resource_aws_key_pair.go
+++ b/aws/resource_aws_key_pair.go
@@ -26,27 +26,21 @@ func resourceAwsKeyPair() *schema.Resource {
 		MigrateState:  resourceAwsKeyPairMigrateState,
 
 		Schema: map[string]*schema.Schema{
-			"key_name": &schema.Schema{
+			"key_name": {
 				Type:          schema.TypeString,
 				Optional:      true,
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"key_name_prefix"},
+				ValidateFunc:  validateMaxLength(255),
 			},
-			"key_name_prefix": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
-				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
-					value := v.(string)
-					if len(value) > 100 {
-						errors = append(errors, fmt.Errorf(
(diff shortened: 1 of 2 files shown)


aws/resource_aws_key_pair.go
```diff
@@ -32,19 +32,13 @@ func resourceAwsKeyPair() *schema.Resource {
...
-					if len(value) > 100 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 100 characters, name is limited to 255", k))
-					}
-					return
-				},
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateMaxLength(100),
```
paultyng: can you add a comment here about the 255 limit

loivis: I would rather change it to `validateMaxLength(255 - 26)` but was unsure if `100` was there for some reason. What do you think?

paultyng:
FYI I put up a PR to [helper/resource](https://github.com/hashicorp/terraform/pull/17551) that would remove the need to hard code these lengths and the comments, but probably won't land for a little.

paultyng:
Also, thanks a lot for cleaning all this up!

loivis:
That's neat. It's always better with self-explanatory codes than comments. I'll wait to sync the change when it's merged.

loivis:
sync with merged pr https://github.com/hashicorp/terraform/pull/17551 and https://github.com/terraform-providers/terraform-provider-aws/pull/3798

paultyng:
LGTM, I'll let @bflad merge.

aws/resource_aws_key_pair.go
```diff
@@ -26,27 +26,21 @@ func resourceAwsKeyPair() *schema.Resource {
...
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 100 characters, name is limited to 255", k))
-					}
-					return
-				},
+			"key_name_prefix": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateMaxLength(255 - resource.UniqueIDSuffixLength),
```
bflad: 😍

PR #3708, opened 2018-03-09
allow some aws_db_instance changes on "storage-full" state

jordane:
Currently, terraform will error if the following are true:

* The user is increasing the `allocated_storage` of an aws_db_instance
* The instance is in the `storage-full` state
* `apply_immediately` is set to false

However, this is a perfectly valid operation to perform for most db engines.

See https://aws.amazon.com/premiumsupport/knowledge-center/rds-out-of-storage/

This change allows `allocated_storage` and `iops` to be changed (and
not applied immediately).

--- 
For reference, here is the change and error produced (resource name and ID are changed):

```
  ~ aws_db_instance.db-name
      allocated_storage: "100" => "120"

aws_db_instance.db-name: Modifying... (ID: <removed>)
  allocated_storage: "100" => "120"
...
1 error(s) occurred:

* aws_db_instance.db-name: 1 error(s) occurred:

* aws_db_instance.db-name: unexpected state 'storage-full', wanted target 'available, storage-optimization'. last error: %!s(<nil>)
```

There are currently no tests in this PR. I don't have a good idea of how to write an Acceptance test for this, and I didn't see any tests for the `storage-optimization` state that was added previously.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -1180,6 +1180,7 @@ var resourceAwsDbInstanceDeletePendingStates = []string{
 	"modifying",
 	"starting",
 	"stopping",
+	"storage-full",
 	"storage-optimization",
 }
 
@@ -1195,5 +1196,6 @@ var resourceAwsDbInstanceUpdatePendingStates = []string{
 	"resetting-master-credentials",
 	"starting",
 	"stopping",
+	"storage-full",
 	"upgrading",
 }
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -1177,6 +1177,7 @@ var resourceAwsDbInstanceDeletePendingStates = []string{
 	"configuring-enhanced-monitoring",
 	"creating",
 	"deleting",
+	"incompatible-parameters",
 	"modifying",
 	"starting",
 	"stopping",


bflad:
Hi @jordane! 👋  Thanks for submitting this. I think we'll want to understand the RDS API behavior here a little more before making this adjustment.

Some questions I have:
* Does RDS return immediate errors if you try to make modifications not relating to increasing the disk? This could influence whether it should be a pending or target state.
* Does RDS return immediate error if you try to make disk modification for SQL Server? Most likely we actually just want to return this error rather than masking it with the cryptic `unexpected state 'storage-full'` Terraform error, meaning removing the custom sqlserver logic. AWS may update their behavior to allow said change.

Regardless of the answers above, we probably want to:
* Add `storage-full` to `resourceAwsDbInstanceDeletePendingStates` - to cover the case of trying to delete a full database instance

jordane:
Hi @bflad,

* Yes, attempting to modify non-storage attributes produces the following error:

```
The specified database instance is currently in storage-full state. Please allocate more storage by modifying the DB instance. (Service: AmazonRDS; Status Code: 400; Error Code: InvalidDBInstanceState; Request ID: 4c1281e2-6857-4af5-b9ce-35a9cb989bde)
```

`aws rds describe-valid-db-instance-modifications` shows only `StorageType`, `StorageSize`, `IopsToStorageRatio`, and `ProvisionedIops`.

---

It turns out that the RDS documentation was out-of-date for sqlserver, and resizing is now a valid operation there as well. I've removed the guard on that piece of code.

---

I added `storage-full` to `resourceAwsDbInstanceDeletePendingStates`. I noticed there are other valid states not listed there. I know at least `incompatible-parameters` is missing, not sure on the others [1]. I can file a separate issue for the rest if you'd like.

[1] https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Status.html

bflad:
💯  Thanks again for all your work here. 

> Yes, attempting to modify non-storage attributes produces the following error
> It turns out that the RDS documentation was out-of-date for sqlserver, and resizing is now a valid operation there as well.

Fantastic on both accounts! Given the behavior I think in this case we can just add `storage-full` to `resourceAwsDbInstanceUpdatePendingStates` as well instead of the custom logic around the target state and letting AWS return its error instead of the Terraform one (e.g. making this PR two lines difference from master from both PendingStates additions, less is more!). I'd ✅  that unless you think that's a bad idea. 😉 

> I noticed there are other valid states not listed there. I know at least incompatible-parameters is missing, not sure on the others [1]. I can file a separate issue for the rest if you'd like.

This is certainly a cat and mouse game for us -- I would definitely create a separate issue for them if they make sense. Is there a reason we should let an instance sit waiting for `incompatible-parameters` status to change rather than just failing immediately like it would today? 🤔

jordane:
> Fantastic on both accounts! Given the behavior I think in this case we can just add storage-full to resourceAwsDbInstanceUpdatePendingStates as well instead of the custom logic around the target state and letting AWS return its error instead of the Terraform one (e.g. making this PR two lines difference from master from both PendingStates additions, less is more!). I'd ￼ that unless you think that's a bad idea. ￼

That makes sense to me. I've made the suggested change, and squashed things down.

> This is certainly a cat and mouse game for us -- I would definitely create a separate issue for them if they make sense. Is there a reason we should let an instance sit waiting for incompatible-parameters status to change rather than just failing immediately like it would today? ￼

What I meant was that an instance in the `incompatible-parameters` state can still be deleted (delete is a valid operation on most states). Wouldn't that mean it also should be allowed in `resourceAwsDbInstanceDeletePendingStates`?

bflad:
> What I meant was that an instance in the incompatible-parameters state can still be deleted (delete is a valid operation on most states). Wouldn't that mean it also should be allowed in resourceAwsDbInstanceDeletePendingStates?

Ah yes, of course! That would be great. Feel free to also include it here since its a trivial change.

jordane:
> Ah yes, of course! That would be great. Feel free to also include it here since its a trivial change.

I have done so. I'll go through some of the other states and file a future issue/pr for those. Thanks!

bflad:

PR #3714, opened 2018-03-09
resource/launch_configuration: drop custom ValidateFunc for name/name_prefix

loivis:

This PR is for:

- [x] resource/launch_configuration

### DIFF
--- a/aws/resource_aws_launch_configuration.go
+++ b/aws/resource_aws_launch_configuration.go
@@ -34,31 +34,14 @@ func resourceAwsLaunchConfiguration() *schema.Resource {
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"name_prefix"},
-				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
-					// https://github.com/boto/botocore/blob/9f322b1/botocore/data/autoscaling/2011-01-01/service-2.json#L1932-L1939
-					value := v.(string)
-					if len(value) > 255 {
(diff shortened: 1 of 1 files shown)


aws/resource_aws_launch_configuration.go
```diff
@@ -34,31 +34,14 @@ func resourceAwsLaunchConfiguration() *schema.Resource {
...
-					if len(value) > 229 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be longer than 229 characters, name is limited to 255", k))
-					}
-					return
-				},
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateMaxLength(255 - 26),
```
paultyng: can you add a comment about why its subtraction?

loivis: synced with the new const 😏

aws/resource_aws_launch_configuration.go
```diff
@@ -99,7 +82,7 @@ func resourceAwsLaunchConfiguration() *schema.Resource {
 						return ""
 					}
 				},
-				ValidateFunc: validateInstanceUserDataSize,
+				ValidateFunc: validateMaxLength(16384),
```
paultyng: See my comment in #3702

paultyng: Nevermind, this is byte length in utf-8, so should be ok I hope

PR #3716, opened 2018-03-09
resource/lb_*: drop custom ValidateFuncs

loivis:

This PR is for:

- [x] resource/lb_cookie_stickiness_policy
- [x] resource/lb_listener
- [x] resource/lb_listener_rule
- [x] resource/lb_target_group

### DIFF
--- a/aws/resource_aws_lb_target_group.go
+++ b/aws/resource_aws_lb_target_group.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsLbTargetGroup() *schema.Resource {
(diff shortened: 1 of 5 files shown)


aws/resource_aws_lb_target_group.go
```diff
@@ -46,27 +47,27 @@ func resourceAwsLbTargetGroup() *schema.Resource {
...
 				ConflictsWith: []string{"name_prefix"},
-				ValidateFunc:  validateAwsLbTargetGroupName,
+				ValidateFunc:  validateMaxLength(32),
 			},
 			"name_prefix": {
 				Type:         schema.TypeString,
 				Optional:     true,
 				ForceNew:     true,
-				ValidateFunc: validateAwsLbTargetGroupNamePrefix,
+				ValidateFunc: validateMaxLength(32 - 26),
```
paultyng: add a comment about why this calculation

loivis: Can we leave it without the comment for now? When all PRs get merged, I can add comment to such ValidateFuncs all together.

Ninir: Hey @loivis 

I'm also in favor in embedding this kind of comments right away not to forget about it afterwards.

loivis: I understand and agree as well. Comments are needed across multiple branches for exactly the same thing. In this case, adding comments after all merges sounds better to me. Furthermore, @paultyng has created a pr adding const in `validation` package. Eventually it would end up like `validateMaxLength(32 - validation.UniqueIDSuffixLength)`. I won't leave it behind.

loivis: synced with merged pr 😏

PR #3735, opened 2018-03-09
resource/waf_*: drop custom ValidateFuncs

loivis:

This PR is for:

- [x] resource/vpc_endpoint(fmt only)
- [x] resource/waf_rate_based_rule
- [x] resource/waf_rule
- [x] resource/waf_web_acl

### DIFF
--- a/aws/resource_aws_waf_rate_based_rule.go
+++ b/aws/resource_aws_waf_rate_based_rule.go
@@ -8,6 +8,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/waf"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsWafRateBasedRule() *schema.Resource {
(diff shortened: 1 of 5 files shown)


aws/resource_aws_waf_rate_based_rule.go
```diff
@@ -70,15 +57,9 @@ func resourceAwsWafRateBasedRule() *schema.Resource {
...
-				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
-					value := v.(int)
-					if value < 2000 {
-						errors = append(errors, fmt.Errorf("%q cannot be less than 2000", k))
-					}
-					return
-				},
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validation.IntAtMost(2000),
```
bflad: Shouldn't this be `validation.IntAtLeast(2000)`?

aws/resource_aws_waf_web_acl.go
```diff
@@ -68,14 +69,11 @@ func resourceAwsWafWebAcl() *schema.Resource {
...
-								if value != waf.WafRuleTypeRegular && value != waf.WafRuleTypeRateBased {
-									errors = append(errors, fmt.Errorf(
-										"%q must be one of %s | %s", k, waf.WafRuleTypeRegular, waf.WafRuleTypeRateBased))
-								}
-								return
-							},
+							ValidateFunc: validation.StringInSlice([]string{
+								waf.WafRuleTypeRegular,
+								waf.WafRuleTypeRateBased,
+								waf.WafRuleTypeGroup,
```
bflad: `waf.WafRuleTypeGroup` was added here. Do we actually support group rules?

aws/validators.go
```diff
@@ -1614,6 +1615,18 @@ func validateWafMetricName(v interface{}, k string) (ws []string, errors []error
 	return
 }
 
+func validateWafPredicatesType() schema.SchemaValidateFunc {
+	return validation.StringInSlice([]string{
+		waf.PredicateTypeIpmatch,
+		waf.PredicateTypeByteMatch,
+		waf.PredicateTypeSqlInjectionMatch,
+		waf.PredicateTypeGeoMatch,
```
bflad: `waf.PredicateTypeGeoMatch` and `waf.PredicateTypeRegexMatch` where added here, do we actually support them (e.g. they do not require additional resource code/attributes)?

aws/resource_aws_waf_rate_based_rule.go
loivis: 😓

Issue #3736, opened 2018-03-09
AWS Provider docs - Organizations page - formatting is broken - missing sidebar


kmcquade:
https://www.terraform.io/docs/providers/aws/r/organizations_organization.html

![image](https://user-images.githubusercontent.com/3422255/37235307-9b284208-23cb-11e8-9f3e-cd4b6f4d0fff.png)

bflad:
Thanks for this report! The fix is now live: https://www.terraform.io/docs/providers/aws/r/organizations_organization.html

PR #3738, opened 2018-03-10
Fix broken documentation for organisation

derBroBro:
This PR should fix issue #3736 which brokes the menu in the docs for the ressource `organizations_organization`


bflad:
Thanks for submitting this. It was actually fixed in #3722 (including fixing the sidebar as well) and is now live: https://www.terraform.io/docs/providers/aws/r/organizations_organization.html

Issue #2188, opened 2017-11-06
Error while enabling enhanced monitoring on AWS Maria DB RDS.


hashibot:
_This issue was originally opened by @bcavarun as hashicorp/terraform#16568. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

### Terraform Version
0.9.5

### Terraform Configuration Files

resource "aws_iam_role" "rds-enhanced-monitoring-role" {
    name = "-rds-monitoring-role"
    assume_role_policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "",
      "Effect": "Allow",
      "Principal": {
        "Service": "monitoring.rds.amazonaws.com"
      },
      "Action": "sts:AssumeRole"
    }
  ]
}
EOF
}

resource "aws_iam_role_policy" "rds-enhanced-monitoring-policy" {
  depends_on = ["aws_iam_role.rds-enhanced-monitoring-role"]
  name = "enhanced-monitoring-attachment"
  role = "${aws_iam_role.rds-enhanced-monitoring-role.name}"
  policy = <<EOF
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "EnableCreationAndManagementOfRDSCloudwatchLogGroups",
            "Effect": "Allow",
            "Action": [
                "logs:CreateLogGroup",
                "logs:PutRetentionPolicy"
            ],
            "Resource": [
                "arn:aws:logs:*:*:log-group:RDS*"
            ]
        },
        {
            "Sid": "EnableCreationAndManagementOfRDSCloudwatchLogStreams",
            "Effect": "Allow",
            "Action": [
                "logs:CreateLogStream",
                "logs:PutLogEvents",
                "logs:DescribeLogStreams",
                "logs:GetLogEvents"
            ],
            "Resource": [
                "arn:aws:logs:*:*:log-group:RDS*:log-stream:*"
            ]
        }
    ]
}
EOF
}

### Debug Output
![enhancedmonitoring](https://user-images.githubusercontent.com/10197011/32445731-b9d7429e-c32c-11e7-93f7-35a36ed2e4aa.png)

### Expected Behavior
Database should have got Enhanced Monitoring turned on.

### Actual Behavior
It threw error that Role Value is required but the role got created in AWS and I am able to attach that role using AWS Console.

[...]

PR #3747, opened 2018-03-12
RDS Enhanced Monitoring Failures (non MySQL engines)

moofish32:
attempting a reproduction issue for Oracle of #2188

@bflad any chance you can verify this reproduction test for Oracle?


moofish32:
@bflad -- this passing makes me think I do not have reproduction. I was failing locally. Let me know if you see anything incorrect.

bflad:
My hunch is that its not specific to non-MySQL (e.g. should be reproducible on MySQL) and its this code (I was planning on fixing this awhile ago, but have been quite sidetracked):

https://github.com/terraform-providers/terraform-provider-aws/blob/3fb1c1132aeb56dee4257f1ea90f9b8045477520/aws/resource_aws_db_instance.go#L972-L982

I think it needs to just set both parameters in the `req` if its being enabled (and maybe disabled). 

You might be able to reproduce this simpler by just taking the existing acceptance test configuration in master and making the monitoring_interval configurable in the fmt.Sprintf so it can be toggled from 0 to 5 and back.

moofish32:
thanks @bflad, I may not get to this until the weekend. Your hunch seems much more likely.

moofish32:
@bflad -- I have tried various step ordering, but I'm not able to get the reproduction. We have some Oracle RDS instances that we hit this bug on, I'm now questioning the reproduction steps. If you have a minute can look at this test case to see if you see a stupid mistake on my part?

moofish32:
@bflad -- hold on let me re-test, it may be multi-az (don't spend a ton of time here I will try to test today), but our local runs only error on multi-az configs.

PR #3728, opened 2018-03-09
Implementation of acl grants

Chhed13:
It implements ACL policy grants requested in #989.
Type `AmazonCustomerByEmail` is not implemented due to it can only be set, but get is only in form of CannonicalUser.
Canned `acl` also can only be set, but not get and it also add some conditionals.

PR #3757, opened 2018-03-12
Support bucket ACLs.

ameir:
I picked this up from my previous PR at https://github.com/hashicorp/terraform/pull/13448 .

It has also been requested at https://github.com/terraform-providers/terraform-provider-aws/issues/989 .

It takes a JSON-formatted ACL and applies it as a bucket policy.  Since there are many unused/optional fields, it filters nil values in the struct to avoid noop diffs.

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -56,9 +56,17 @@ func resourceAwsS3Bucket() *schema.Resource {
 			},
 
 			"acl": {
-				Type:     schema.TypeString,
-				Default:  "private",
-				Optional: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"acl_policy"},
+			},
+
+			"acl_policy": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Computed:         true,
+				ValidateFunc:     validateJsonString,
+				DiffSuppressFunc: suppressEquivalentJsonDiffs,
 			},
 
 			"policy": {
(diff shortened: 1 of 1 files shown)


Chhed13:
Did you check it on flapping without any `acl` ? Like this:
Twice `terraform apply` this
```
resource "aws_s3_bucket" "bucket" {
	bucket = "tf-test-bucket-%d"
}
```
I faced with this in my realization #3728. That's because you can't completely delete permissions from bucket owner

ameir:
I did notice that, but found that adding `Computed: true` fixed it.  The `acl_policy` is then added to the state file, and is not overwritten until you attempt to in the config.

ameir:
@bflad did you have any thoughts on this or https://github.com/terraform-providers/terraform-provider-aws/pull/3728 ?

tomelliff:
This PR is unlikely to be merged without some accompanying tests. Did you still want to finish this off? If not I can see if I can find the time to work on this as I'm also keen to enable Cloudfront access logs to an S3 bucket.

ameir:
I'd be willing to work on this further (with tests) if I had some sort of
reassurance that it will get merged. I have very little spare time these
days, so I want to make sure the investment is fruitful.

tomelliff:
#3728 looks to be further along but also being actively worked on so I'll wait for that instead of duplicating efforts further here.

ewbankkit:
It looks like this has been superseded by https://github.com/terraform-providers/terraform-provider-aws/pull/3728.

bflad:
Hi @ameir 👋 Thank you for submitting this and apologies that this went a long time without review. #3728 was recently merged and appears to provide similar functionality. If there is anything missing from that implementation, please submit a new pull request.

PR #3762, opened 2018-03-13
resource/cognito_user_pool_client: fix import with user pool id

### LINKED ISSUE #3753: Unable to import cognito user pool client
### Terraform Version
```
Terraform v0.11.3
+ provider.aws v1.11.0
```

### Affected Resource(s)
aws_cognito_user_pool_client

resource "aws_cognito_user_pool" "pool" {
&nbsp;&nbsp;&nbsp;&nbsp;name = "pool"
}
resource "aws_cognito_user_pool_client" "client" {
&nbsp;&nbsp;&nbsp;&nbsp;name = "client"
&nbsp;&nbsp;&nbsp;&nbsp;user_pool_id = "${aws_cognito_user_pool.pool.id}"
}

### Debug Output
https://gist.github.com/repl-chris/7215ee65e99e7ead2caf8100b0b1101d

### Expected Behavior
It should import successfully

### Actual Behavior
It appears to import successfully, but then errors out on 'Refreshing state...'
```
C:\temp> terraform import aws_cognito_user_pool_client.client 3v1ugef2oda5amrfsv9faegkog
aws_cognito_user_pool_client.client: Importing from ID "3v1ugef2oda5amrfsv9faegkog"...
aws_cognito_user_pool_client.client: Import complete!
  Imported aws_cognito_user_pool_client (ID: 3v1ugef2oda5amrfsv9faegkog)
aws_cognito_user_pool_client.client: Refreshing state... (ID: 3v1ugef2oda5amrfsv9faegkog)

Error: aws_cognito_user_pool_client.client (import id: 3v1ugef2oda5amrfsv9faegkog): 1 error(s) occurred:

* import aws_cognito_user_pool_client.client result: 3v1ugef2oda5amrfsv9faegkog: aws_cognito_user_pool_client.client: InvalidParameter: 1 validation error(s) found.
- minimum field size of 1, DescribeUserPoolClientInput.UserPoolId.
```

### Steps to Reproduce
1. Comment out the aws_cognito_user_pool_client resource from the config listed above
2. `terraform apply`
3. Add a client manually (via the AWS console or whatever)...take note of its ID
4. Uncomment out the aws_cognito_user_pool_client
5. `terraform import aws_cognito_user_pool_client.client <your client ID here>`

loivis:
fix #3753 

When describing user pool client, `user_pool_id` is required.

1. Currently `id` is set as `client_id`, which makes it unable to import by default. Changing `id` to `user_pool_id/client_id` should be the best as I think. But that would be a breaking change?

2. Instead, I updated import function to parse `user_pool_id` and `client_id` during import.

### DIFF
--- a/aws/resource_aws_cognito_user_pool_client_test.go
+++ b/aws/resource_aws_cognito_user_pool_client_test.go
@@ -35,6 +35,30 @@ func TestAccAWSCognitoUserPoolClient_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSCognitoUserPoolClient_importBasic(t *testing.T) {
+	userPoolName := fmt.Sprintf("tf-acc-cognito-user-pool-%s", acctest.RandString(7))
+	clientName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+
+	resourceName := "aws_cognito_user_pool_client.client"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEcsServiceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSCognitoUserPoolClientConfig_basic(userPoolName, clientName),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportStateId:     "user_pool_id/client_id", // how to get both ids here?
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
 func TestAccAWSCognitoUserPoolClient_allFields(t *testing.T) {
 	userPoolName := fmt.Sprintf("tf-acc-cognito-user-pool-%s", acctest.RandString(7))
 	clientName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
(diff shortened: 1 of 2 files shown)


aws/resource_aws_cognito_user_pool_client_test.go
```diff
@@ -35,6 +35,68 @@ func TestAccAWSCognitoUserPoolClient_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSCognitoUserPoolClient_importBasic(t *testing.T) {
+	userPoolName := fmt.Sprintf("tf-acc-cognito-user-pool-%s", acctest.RandString(7))
+	clientName := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+
+	resourceName := "aws_cognito_user_pool_client.client"
+
+	getStateId := func(*terraform.State) (string, error) {
```
bflad: We could shorten this lookup function if we set `*client = ...` in `testAccCheckAWSCognitoUserPoolClientExists()` from the return value of `DescribeUserPoolClient()` 

Then we can simply grab the right IDs from an object we've already fetched in there 👍

loivis: Thanks for input. Not sure if I get your point exactly.

I've mostly copied the function and changed return values. Instead of parsing return value from `DescribeUserPoolClient()`, I directly return `userPoolId` and `clientId` set in `DescribeUserPoolClientInput` since there is error checking on `DescribeUserPoolClient()`. I assume that if there is no error then both IDs should be considered as valid.

PR #3769, opened 2018-03-13
Add support for timeouts on aws_eip resource

ctso:
Adds the ability to override timeouts for the `aws_eip` resource.

Would be nice to have https://github.com/terraform-providers/terraform-provider-aws/pull/3586 along with this change, as we're running into something similar as reported in https://github.com/terraform-providers/terraform-provider-aws/issues/3128 but with EIPs instead of security groups.

Open to suggestions on the timeouts here.


bflad:

PR #3771, opened 2018-03-13
#3626 Display Invalid Value During InvalidParameterValue

oscr:
Hi, 

This commit adds the aws_db_instance name argument to the error message as requested in issue #3626

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -674,7 +674,7 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			return nil
 		})
 		if err != nil {
-			return fmt.Errorf("Error creating DB Instance: %s", err)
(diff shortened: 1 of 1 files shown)


aws/resource_aws_db_instance.go
```diff
@@ -674,7 +674,7 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			return nil
 		})
 		if err != nil {
-			return fmt.Errorf("Error creating DB Instance: %s", err)
+			return fmt.Errorf("Error creating DB Instance: provided name: %v, %s", *opts.DBName, err)
```
bflad: This will likely cause a panic as usually the response from the SDK when there is an error is `(nil, error)`.

bflad:
This pull request has been superseded by #4803 which verbosely passes back all provided input options and would not panic.

PR #3404, opened 2018-02-15
Update `resource_aws_cognito_user_pool_client` to bind updates to `callback_urls` to the correct param

nick-woodward:
### Description

Fixes an issue where it was impossible to update a `resource_aws_cognito_user_pool_client`.

I'm not overly familiar with go, so I don't really know the best way to regression test this.

### Related Issues

closes #3384

PR #3783, opened 2018-03-14
Fix copy-paste bug in Cognito User Pool Client

erothmayer:


bflad:
Hi @erothmayer! 👋  Thanks for this contribution! While this was a great catch and fix, an earlier PR (#3404) was merged in already to handle it.

erothmayer:
Roger, as long as it's fixed I'm happy :-)

Thanks @bflad !

Issue #518, opened 2017-06-13
provider/aws: db_instance - Encrypted read-replica in a different region


hashibot:
_This issue was originally opened by @gdowmont as hashicorp/terraform#11784. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,

It would be great if terraform supported creating encrypted RDS read-replicas in a different region than the parent instance inside the same/different account. AWS have recently added support for it and Go AWS SDK has been extended to include that functionality via 'DestinationRegion' [https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#CreateDBInstanceReadReplicaInput](url).

This is very useful if you need to keep real-time backups away from your main region or in a different account.

Thanks for considering this.

Greg

rfletcher:
I don't know if this will actually complicate things, but something to be aware of: RDS instance identifiers only [need to be unique *per-region*](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Aurora.Replication.CrossRegion.html#Aurora.Replication.CrossRegion.Creating) (see the table in bullet no. 5):

> **DB Instance Identifier**
>
> It must be unique for all DB instances for each AWS account, for each region.

Terraform uses that value as the aws_db_instance ID.

handlerbot:
I think this issue can be closed, FTR; I have successfully used Terraform to create encrypted RDS read replicas in a different region, by setting the provider to the desired region, setting `replicate_source_db` to the full ARN (not the identifier) of the source RDS instance, and setting `kms_key_id`.

That said: there is some small user friendliness that could happen here by detecting if you are trying to make an encrypted cross-region RDS read replica (`replicate_source_db` is an ARN, not an identifier + `storage_encrypted` is true) but if you don't have `kms_key_id` set, and returning a error from Terraform saying that you need to set `kms_key_id`. If you don't set `kms_key_id`, Terraform passes through the (confusing) message from AWS saying that "you can't make an unencrypted replica in that region" or something, when what it means is "you said you wanted it encrypted, and you need to provide as KMS key id when you say that".

So, that could happen, but for other folks following the breadcrumbs to here as I did, it does work...

handlerbot:
[...]

PR #865, opened 2017-06-14
resource/rds_db_instance: Correctly create cross-region encrypted replica

micahhausler:
* Bumped AWS SDK to v1.8.41

The AWS SDK generates and sets the `rds.CreateDBInstanceReadReplicaInput.PreSignedUrl` for you if you [set the `.SourceRegion`](https://github.com/aws/aws-sdk-go/blob/v1.8.41/service/rds/customizations.go#L58-L60). 

Closes #864

PR #3795, opened 2018-03-15
resource/db_instance: set SourceRegion based on source_db ARN

omeid:
As Per AmazonRDS/API/CreateDBInstanceReadReplica docs

		If you create an encrypted Read Replica in a different AWS Region,
		then you must specify a KMS key for the destination AWS Region.
		KMS encryption keys are specific to the AWS Region that they are
		created in, and you can't use encryption keys from one AWS Region
		in another AWS Region.

This commit sets the Read Replica Source Region based on the
replicate_source_db arn.

kms_key_id is managed as usual.
Fixes:  #518
Related: #3337, #865

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -535,11 +535,12 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			opts.Iops = aws.Int64(int64(attr.(int)))
 		}
 
+		if arnParts := strings.Split(v.(string), ":"); len(arnParts) >= 4 {
+			opts.SourceRegion = aws.String(arnParts[3])
+		}
+
 		if attr, ok := d.GetOk("kms_key_id"); ok {
 			opts.KmsKeyId = aws.String(attr.(string))
-			if arnParts := strings.Split(v.(string), ":"); len(arnParts) >= 4 {
-				opts.SourceRegion = aws.String(arnParts[3])
-			}
 		}
 
 		if attr, ok := d.GetOk("maintenance_window"); ok {
--- a/website/docs/r/db_instance.html.markdown
+++ b/website/docs/r/db_instance.html.markdown
@@ -115,8 +115,7 @@ identifier beginning with the specified prefix. Conflicts with `identifier`.
 * `instance_class` - (Required) The instance type of the RDS instance.
 * `iops` - (Optional) The amount of provisioned IOPS. Setting this implies a
 storage_type of "io1".
-* `kms_key_id` - (Optional) The ARN for the KMS encryption key. If creating an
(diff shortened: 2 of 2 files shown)


omeid:
@catsby I believe you have done most of the `db_instance` read replica, so if you could review this, that would be excellent.

omeid:
Requires aws/aws-sdk-go#1847 to land.

jasdel:
The AWS SDK for Go's aws/aws-sdk-go#1847 PR has been replaced with aws/aws-sdk-go#2631 adding the outstanding unit tests and will be merged into master once reviewed.

omeid:
aws/aws-sdk-go#2631 just landed, so this is good to go.

omeid:
Any chance this could be merged in? maintaining a fork because of pending bug fixes is rather painful.

aws/resource_aws_db_instance.go
```diff
@@ -535,11 +535,12 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			opts.Iops = aws.Int64(int64(attr.(int)))
 		}
 
+		if arnParts := strings.Split(v.(string), ":"); len(arnParts) >= 4 {
+			opts.SourceRegion = aws.String(arnParts[3])
+		}
```
bflad: This check would be safer and easier to understand using the AWS Go SDK provided ARN handlers [`arn.IsARN()`](https://github.com/aws/aws-sdk-go/blob/9c9db07028aeb7d607e54b2b5aa97bbbbebcbbfd/aws/arn/arn.go#L78-L82) and [`arn.Parse()`](https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/#Parse), e.g.

```go
		if arn.IsARN(v.(string)) {
			replicaSourceDbArn, err := arn.Parse(v.(string))
			if err != nil {
				return fmt.Errorf("error parsing replica_source_db as ARN: %s", err)
			}
			opts.SourceRegion = aws.String(replicaSourceDbArn.Region)
		}
```

bflad:
Hi again 👋 Since we haven't heard back, we are going to close this pull request for now. If anyone is still interested in this change, please submit a new pull request and we will take a fresh look. Thanks.

omeid:
I am still interested in fixing this. However, since we have been waiting for this for almost two years before we heard anything from Terrraform, we can't prioritize putting time into this as a matter of priorities.

Also, I personally find it a bit distasteful that you leave a contributor waiting for an answer for two years but then promptly close the issue once you don't get a reply in a month or so.

aws/resource_aws_db_instance.go
omeid: Okay, I can make that change. Just shows how long this pr has been sitting here. (;

PR #3815, opened 2018-03-16
Document additional allowed source type GITHUB_ENTERPRISE

errriclee:
Now that the most recent commit snuck in support for source type GITHUB_ENTERPRISE...

https://github.com/terraform-providers/terraform-provider-aws/commit/e69f97543e09ed42da63678d7516e175d32724cf#diff-c96a6fa59f39978bec1013238a523f60R184
vs
https://github.com/terraform-providers/terraform-provider-aws/commit/e69f97543e09ed42da63678d7516e175d32724cf#diff-c96a6fa59f39978bec1013238a523f60L794

... document that it's a valid source type.

### DIFF
--- a/website/docs/r/codebuild_project.html.markdown
+++ b/website/docs/r/codebuild_project.html.markdown
@@ -153,7 +153,7 @@ The following arguments are supported:
 
 `source` supports the following:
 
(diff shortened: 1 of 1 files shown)


bflad:
Indeed, great catch! Thanks!

johnnorton:
How does one supply the Auth Personal Access Token  to use GITHUB_ENTERPRISE

PR #3827, opened 2018-03-18
New Resource: aws_appsync_api_key

lucidprogrammer:
Supporting the creation of appsync api key. Either a specific date or number of days.

resource "aws_appsync_graphql_api" "example" {
  authentication_type = "API_KEY"
  name = "example"
}

resource "aws_appsync_api_key" "ex1" {
 appsync_api_id = "${aws_appsync_graphql_api.example.id}"
 validity_period_days = 364
}

resource "aws_appsync_api_key" "ex2" {
 appsync_api_id = "${aws_appsync_graphql_api.example.id}"
 valid_till_date = "30/11/2018"
}

### DIFF
--- a/aws/resource_aws_appsync_api_key.go
+++ b/aws/resource_aws_appsync_api_key.go
@@ -144,7 +144,7 @@ func resourceAwsAppsyncApiKeyUpdate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	return resourceAwsAppsyncApiKeyRead(d, meta)
-	return nil
+
 }
 
 func resourceAwsAppsyncApiKeyDelete(d *schema.ResourceData, meta interface{}) error {
--- a/aws/resource_aws_appsync_api_key.go
+++ b/aws/resource_aws_appsync_api_key.go
@@ -2,13 +2,14 @@ package aws
 
 import (
 	"fmt"
-	"regexp"
+	"log"
 	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/appsync"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsAppsyncApiKey() *schema.Resource {
@@ -25,32 +26,14 @@ func resourceAwsAppsyncApiKey() *schema.Resource {
 				Optional: true,
 				Default:  "Managed by Terraform",
 			},
-			"appsync_api_id": {
+			"api_id": {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"valid_till_date": {
-				Type:          schema.TypeString,
-				ConflictsWith: []string{"validity_period_days"},
-				Optional:      true,
-				ValidateFunc: func(v interface{}, k string) (ws []string, es []error) {
-					value := v.(string)
-					// reference - http://www.regexlib.com/REDetails.aspx?regexp_id=409
-					if !regexp.MustCompile(`^(((0[1-9]|[12]\d|3[01])\/(0[13578]|1[02])\/((1[6-9]|[2-9]\d)\d{2}))|((0[1-9]|[12]\d|30)\/(0[13456789]|1[012])\/((1[6-9]|[2-9]\d)\d{2}))|((0[1-9]|1\d|2[0-8])\/02\/((1[6-9]|[2-9]\d)\d{2}))|(29\/02\/((1[6-9]|[2-9]\d)(0[48]|[2468][048]|[13579][26])|((16|[2468][048]|[3579][26])00))))$`).MatchString(value) {
-						es = append(es, fmt.Errorf(
-							"only dd/mm/yyyy in %q", k))
-					}
-					return
-				},
-			},
-			"validity_period_days": {
-				Type:          schema.TypeInt,
-				ConflictsWith: []string{"valid_till_date"},
-				Optional:      true,
-			},
-			"expiry_date": {
-				Type:     schema.TypeString,
-				Computed: true,
+			"expires": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.ValidateRFC3339TimeString,
 			},
 			"key": {
 				Type:      schema.TypeString,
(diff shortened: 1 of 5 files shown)


aws/resource_aws_appsync_api_key.go
```diff
@@ -0,0 +1,166 @@
...
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "Managed by Terraform",
+			},
+			"appsync_api_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"valid_till_date": {
```
bflad: A few notes here:
* We should match the AWS API and create a single attribute `expires` rather than provide multiple conflicting attributes for the same information (drop `valid_till_date`, `validity_period_days`, and `expiry_date`)
* It should accept a RFC3339 timestamp for now and can be validated via `ValidateFunc: validation.ValidateRFC3339TimeString`

aws/resource_aws_appsync_api_key.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	d.SetId(*resp.ApiKey.Id)
+	return resourceAwsAppsyncApiKeyRead(d, meta)
+}
+
+func resourceAwsAppsyncApiKeyRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).appsyncconn
+
+	input := &appsync.ListApiKeysInput{
```
bflad: We need to account for `NextToken` when the list is larger than the returned results or we could potentially miss fetching the API key.

aws/resource_aws_appsync_api_key.go
```diff
@@ -0,0 +1,166 @@
...
+		ApiId: aws.String(d.Get("appsync_api_id").(string)),
+	}
+
+	resp, err := conn.ListApiKeys(input)
+	if err != nil {
+		return err
+	}
+	var key appsync.ApiKey
```
bflad: This code is not accounting for if the key is not found in the `ListApiKeys` output(s). I would recommend changing this to `var key *appsync.ApiKey` and performing a nil check before dereferencing the values (otherwise it will cause a Terraform panic):

```go
// After exhausting all ListApiKeys searching
if key == nil {
  log.Printf("[WARN] AppSync API Key %q not found, removing from state", d.Id())
  d.SetId("")
  return nil
}
d.Set("key", key.Id)
// ...
```

aws/resource_aws_appsync_api_key.go
```diff
@@ -0,0 +1,166 @@
...
+	d.Set("description", key.Description)
+	d.Set("expiry_date", time.Unix(*key.Expires, 0).String())
+	return nil
+}
+
+func resourceAwsAppsyncApiKeyUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).appsyncconn
+
+	params := &appsync.UpdateApiKeyInput{
+		ApiId: aws.String(d.Get("appsync_api_id").(string)),
```
bflad: Given the API key seems a subset of the API ID, we should probably add this ID to the Terraform resource ID.

In the Create function (if `:` is a safe delimiter):

```go
d.SetId(fmt.Sprintf("%s:%s"), d.Get("appsync_api_id").(string), *resp.ApiKey.Id)
```

We can then parse it with a helper function, e.g.

```go
func decodeAppSyncApiKeyId(id string) (string, string, error) {
	parts := strings.Split(id, ":")
	if len(parts) != 2 {
		return "", "", fmt.Errorf("Unexpected format of ID (%q), expected API-ID:API-KEY-ID", id)
	}
	return parts[0], parts[1], nil
}
```

And anywhere necessary to call that:

```go
apiID, apiKeyID, err := decodeAppSyncApiKeyId(d.Id())
if err != nil {
	return err
}
```

This will allow us to support importing easily with:

```go
},
```

aws/resource_aws_appsync_api_key.go
```diff
@@ -0,0 +1,166 @@
...
+		Update: resourceAwsAppsyncApiKeyUpdate,
+		Delete: resourceAwsAppsyncApiKeyDelete,
+
+		Schema: map[string]*schema.Schema{
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "Managed by Terraform",
+			},
+			"appsync_api_id": {
```
bflad: I don't think we need to prepend `appsync_` to this attribute name. 👍

lucidprogrammer:
@bflad I will update docs tomorrow related to the resource changes later tonight

lucidprogrammer:
@bflad anything else to be done on this, let me know your thoughts. tks

lucidprogrammer:
@bflad i know it could be hectic with lot going on at your end. doing this for a customer, and pressure is mounting, where other resources for appsync also to be created.(like schema creation, types etc). any feedback to merge this will be much appreciated. tks

aws/resource_aws_appsync_api_key.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func resourceAwsAppsyncApiKeyRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).appsyncconn
+	var listKeys func(*appsync.ListApiKeysInput) (*appsync.ApiKey, error)
+	ApiId, Id, er := decodeAppSyncApiKeyId(d.Id())
+	if er != nil {
+		return er
+	}
+	listKeys = func(input *appsync.ListApiKeysInput) (*appsync.ApiKey, error) {
```
bflad: This should be made into its own top level function, so it can be used by `testAccCheckAwsAppsyncApiKeyDestroy` and `testAccCheckAwsAppsyncApiKeyExists` so they properly handle `NextToken`

aws/resource_aws_appsync_api_key_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		resp, err := conn.ListApiKeys(input)
+		if err != nil {
+			return err
+		}
+
+		var key appsync.ApiKey
+		for _, v := range resp.ApiKeys {
+			if *v.Id == *aws.String(Id) {
+				key = *v
```
bflad: I would recommend splitting this logic into two functions:

```go
func testAccCheckAwsAppsyncApiKeyExists(resourceName string, apiKey *appsync.ApiKey)
```

which calls `*apiKey = *v` right here or otherwise returns an error that the key could not be found.

Then separately:

```go
func testAccCheckAwsAppsyncApiKeyTillDate(apiKey *appsync.ApiKey, date time.Time)
```

which handles the date/time logic below.

bflad:
Finished the implementation in a commit after the others. Passes acceptance testing:

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSAppsyncApiKey'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAppsyncApiKey -timeout 120m
=== RUN   TestAccAWSAppsyncApiKey_basic
--- PASS: TestAccAWSAppsyncApiKey_basic (12.30s)
=== RUN   TestAccAWSAppsyncApiKey_Description
--- PASS: TestAccAWSAppsyncApiKey_Description (18.26s)
=== RUN   TestAccAWSAppsyncApiKey_Expires
--- PASS: TestAccAWSAppsyncApiKey_Expires (19.80s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	50.401s
```

bflad:

PR #3858, opened 2018-03-21
New Resource: aws_sns_sms_preferences

lbrucher:
There is currently no way to set the SMS preferences in SNS from Terraform.

This PR is an attempt at providing this feature.

### DIFF
--- a/aws/resource_aws_sns_sms_preferences.go
+++ b/aws/resource_aws_sns_sms_preferences.go
@@ -8,7 +8,24 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/sns"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
+
+func validateMonthlySpend(v interface{}, k string) (ws []string, errors []error) {
+	vInt, _ := strconv.Atoi(v.(string))
+	if vInt < 0 {
+		errors = append(errors, fmt.Errorf("Error setting SMS preferences: monthly spend limit value [%d] must be >= 0!", vInt))
+	}
+	return
+}
+
+func validateDeliverySamplingRate(v interface{}, k string) (ws []string, errors []error) {
+	vInt, _ := strconv.Atoi(v.(string))
+	if vInt < 0 || vInt > 100 {
+		errors = append(errors, fmt.Errorf("Error setting SMS preferences: default percentage of success to sample value [%d] must be between 0 and 100!", vInt))
+	}
+	return
+}
 
 func resourceAwsSnsSmsPreferences() *schema.Resource {
 	return &schema.Resource{
@@ -19,8 +36,9 @@ func resourceAwsSnsSmsPreferences() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"monthly_spend_limit": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateMonthlySpend,
 			},
 
 			"delivery_status_iam_role_arn": {
@@ -29,8 +47,9 @@ func resourceAwsSnsSmsPreferences() *schema.Resource {
 			},
 
 			"delivery_status_success_sampling_rate": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validateDeliverySamplingRate,
 			},
 
 			"default_sender_id": {
@@ -39,8 +58,9 @@ func resourceAwsSnsSmsPreferences() *schema.Resource {
 			},
 
 			"default_sms_type": {
-				Type:     schema.TypeString,
-				Optional: true,
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"Promotional", "Transactional"}, false),
 			},
 
 			"usage_report_s3_bucket": {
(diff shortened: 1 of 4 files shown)


lbrucher:
Hello, I was wondering if anyone has had a chance to look at this PR? Any blocker or changes to be made? Any potential of it being merged soon? Thanks!

bcornils:
Hello @lbrucher apologies for the delay. We are looking into several PR's and requests atm. Let me work on where we are and get back to you with something concrete. I'm the new PM for the project so I believe the slow down is my fault.

aws/resource_aws_sns_sms_preferences.go
```diff
@@ -0,0 +1,159 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+		},
+	}
+}
+
+const resourceId = "aws_sns_sms_id"
+
+var SMSAttributeMap = map[string]string{
```
paultyng: no need to export this I don't think? (same with the var below as well)

aws/resource_aws_sns_sms_preferences.go
```diff
@@ -0,0 +1,159 @@
...
+	monthlySpendLimit := d.Get("monthly_spend_limit").(string)
+	monthlySpendLimitInt, _ := strconv.Atoi(monthlySpendLimit)
+	deliveryStatusIamRoleArn := d.Get("delivery_status_iam_role_arn").(string)
+	deliveryStatusSuccessSamplingRate := d.Get("delivery_status_success_sampling_rate").(string)
+	deliveryStatusSuccessSamplingRateInt, _ := strconv.Atoi(deliveryStatusSuccessSamplingRate)
+	defaultSenderId := d.Get("default_sender_id").(string)
+	defaultSmsType := d.Get("default_sms_type").(string)
+	usageReportS3Bucket := d.Get("usage_report_s3_bucket").(string)
+
+	// Validation
```
paultyng: Please move validation in to `ValidateFunc` in the schema above like this: https://github.com/terraform-providers/terraform-provider-aws/blob/3f9a830038f20f819ba947e90770c50d470d164b/aws/resource_aws_s3_bucket.go#L249

(for these other validations as well, more validations available in the `validation` package, int ranges, string in slice, etc)

aws/resource_aws_sns_sms_preferences.go
```diff
@@ -0,0 +1,159 @@
...
+func resourceAwsSnsSmsPreferences() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSnsSmsPreferencesSet,
+		Read:   resourceAwsSnsSmsPreferencesGet,
+		Update: resourceAwsSnsSmsPreferencesSet,
+		Delete: resourceAwsSnsSmsPreferencesDelete,
+
+		Schema: map[string]*schema.Schema{
+			"monthly_spend_limit": {
+				Type:     schema.TypeString,
```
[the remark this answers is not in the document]
lbrucher: This is tricky because the AWS API treats all those attributes as Strings, including monthly spend limit and delivery sampling rate, and an empty string "" (not defined) is not the same as a "0" (explicit 0).
Using int values in TF won't allow us to distinguish between the 2 cases and will always set those values as "0" when retrieved as "", which would be incorrect.

aws/resource_aws_sns_sms_preferences.go
```diff
@@ -0,0 +1,159 @@
...
+				Optional: true,
+			},
+
+			"delivery_status_iam_role_arn": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"delivery_status_success_sampling_rate": {
+				Type:     schema.TypeString,
```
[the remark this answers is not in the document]
lbrucher: See above comment

aws/resource_aws_sns_sms_preferences.go
lbrucher: Indeed, corrected

lbrucher:
@paultyng thanks for the feedback, please see my comments above, thanks.

bflad:
I'm seeing a bunch of test failures but they are likely from our testing being parallelized:
```
=== RUN   TestAccAWSSNSSMSPreferences_almostAll
--- FAIL: TestAccAWSSNSSMSPreferences_almostAll (6.24s)
    testing.go:518: Step 0 error: After applying this step and refreshing, the plan was not empty:
        
        DIFF:
        
        UPDATE: aws_sns_sms_preferences.test_pref
          default_sms_type:                      "" => "Transactional"
          delivery_status_iam_role_arn:          "arn:aws:iam::*******:role/test_smsdelivery_role" => ""
          delivery_status_success_sampling_rate: "75" => ""
          monthly_spend_limit:                   "" => "1"
          usage_report_s3_bucket:                "" => "some-bucket"
=== RUN   TestAccAWSSNSSMSPreferences_defaultSMSType
--- FAIL: TestAccAWSSNSSMSPreferences_defaultSMSType (6.70s)
    testing.go:518: Step 0 error: After applying this step and refreshing, the plan was not empty:
        
        DIFF:
        
        UPDATE: aws_sns_sms_preferences.test_pref
          default_sms_type:                      "" => "Transactional"
          delivery_status_iam_role_arn:          "arn:aws:iam::*******:role/test_smsdelivery_role" => ""
          delivery_status_success_sampling_rate: "75" => ""

=== RUN   TestAccAWSSNSSMSPreferences_deliveryRole
--- FAIL: TestAccAWSSNSSMSPreferences_deliveryRole (9.55s)
        
        Error: Check failed: SNS SMS Preferences resource exists when it should be destroyed!
        
        State: <no state>

=== RUN   TestAccAWSSNSSMSPreferences_empty
--- FAIL: TestAccAWSSNSSMSPreferences_empty (6.10s)
    testing.go:518: Step 0 error: After applying this step and refreshing, the plan was not empty:
        
        DIFF:
        
        UPDATE: aws_sns_sms_preferences.test_pref
          delivery_status_iam_role_arn:          "arn:aws:iam::*******:role/test_smsdelivery_role" => ""
          delivery_status_success_sampling_rate: "75" => ""
```

This will need to be adjusted on merge

rahuljain1311:
@bflad @paultyng
Thanks for approving this PR. It will significantly help us while playing with SMSs. If you guys can tell possible time frame for releasing v1.18.0 it would be super helpful

bflad:

PR #3862, opened 2018-03-21
resource/aws_batch_*: Add test sweepers

bflad:
♻️  To help prevent:
```
=== RUN   TestAccAWSBatchComputeEnvironment_updateInstanceType
--- FAIL: TestAccAWSBatchComputeEnvironment_updateInstanceType (7.69s)
        
        * aws_batch_compute_environment.ec2: 1 error(s) occurred:
        
        * aws_batch_compute_environment.ec2: : Error executing request, Exception : Limits Error : Only 18 Compute Environments allowed, RequestId: 7206eae0-2cec-11e8-9d0d-891d1a5a6818
            status code: 400, request id: 7206eae0-2cec-11e8-9d0d-891d1a5a6818
```

```
$ aws batch describe-compute-environments | jq '.computeEnvironments | length'
13

$ make sweep SWEEPARGS='-sweep-run=aws_batch_compute_environment'
...
2018/03/21 13:57:30 Sweeper Tests ran:
	- aws_batch_job_queue
	- aws_batch_compute_environment
ok  	github.com/terraform-providers/terraform-provider-aws/aws	40.023s

$ aws batch describe-compute-environments | jq '.computeEnvironments | length'
0
```

### DIFF
--- a/aws/resource_aws_batch_compute_environment.go
+++ b/aws/resource_aws_batch_compute_environment.go
@@ -228,7 +228,7 @@ func resourceAwsBatchComputeEnvironmentCreate(d *schema.ResourceData, meta inter
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{batch.CEStatusCreating},
 		Target:     []string{batch.CEStatusValid},
-		Refresh:    resourceAwsBatchComputeEnvironmentStatusRefreshFunc(d, meta),
+		Refresh:    resourceAwsBatchComputeEnvironmentStatusRefreshFunc(computeEnvironmentName, conn),
 		Timeout:    d.Timeout(schema.TimeoutCreate),
 		MinTimeout: 5 * time.Second,
 	}
@@ -320,7 +320,7 @@ func resourceAwsBatchComputeEnvironmentDelete(d *schema.ResourceData, meta inter
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{batch.CEStatusUpdating},
 		Target:     []string{batch.CEStatusValid},
-		Refresh:    resourceAwsBatchComputeEnvironmentStatusRefreshFunc(d, meta),
+		Refresh:    resourceAwsBatchComputeEnvironmentStatusRefreshFunc(computeEnvironmentName, conn),
 		Timeout:    d.Timeout(schema.TimeoutDelete),
 		MinTimeout: 5 * time.Second,
 	}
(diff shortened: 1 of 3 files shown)


aws/resource_aws_batch_compute_environment.go
```diff
@@ -390,12 +390,8 @@ func resourceAwsBatchComputeEnvironmentUpdate(d *schema.ResourceData, meta inter
 	return resourceAwsBatchComputeEnvironmentRead(d, meta)
 }
 
-func resourceAwsBatchComputeEnvironmentStatusRefreshFunc(d *schema.ResourceData, meta interface{}) resource.StateRefreshFunc {
+func resourceAwsBatchComputeEnvironmentStatusRefreshFunc(computeEnvironmentName string, conn *batch.Batch) resource.StateRefreshFunc {
```
radeksimko: ❤️ 👍 🚿 💯

aws/resource_aws_batch_compute_environment_test.go
```diff
@@ -12,6 +15,93 @@ import (
...
+			Pending:    []string{batch.CEStatusDeleting},
+			Target:     []string{batch.CEStatusDeleted},
+			Refresh:    resourceAwsBatchComputeEnvironmentDeleteRefreshFunc(*name, conn),
+			Timeout:    20 * time.Minute,
+			MinTimeout: 5 * time.Second,
+		}
+		if _, err := stateConfForDelete.WaitForState(); err != nil {
+			log.Printf("[ERROR] Failed to wait for deletion of Batch Compute Environment %s: %s", *name, err)
+		}
+	}
```
radeksimko: I assume this is more or less copy of https://github.com/terraform-providers/terraform-provider-aws/blob/09d4f4ce96100b682a83cf3ae8bf589e6f36b65a/aws/resource_aws_batch_compute_environment.go#L307-L348 - which makes me wonder if we should move that logic into a function and reuse it here and in the Delete function?

aws/resource_aws_batch_job_queue_test.go
```diff
@@ -6,13 +6,101 @@ import (
...
+			Target:     []string{batch.JQStatusDeleted},
+			Refresh:    resourceAwsBatchComputeEnvironmentDeleteRefreshFunc(*name, conn),
+			Timeout:    10 * time.Minute,
+			Delay:      10 * time.Second,
+			MinTimeout: 3 * time.Second,
+		}
+		if _, err := stateConfForDelete.WaitForState(); err != nil {
+			log.Printf("[ERROR] Failed to wait for deletion of Batch Job Queue %s: %s", *name, err)
+		}
+	}
```
radeksimko: Same question here about reusability of the deletion logic ^

bflad:
I think we'd run into a similar problem:

```
# signature
func deleteBatchComputeEnvironment(name, logLevel string, timeout time.Duration)
# resource call
deleteBatchComputeEnvironment(name, "DEBUG", d.Timeout(schema.TimeoutDelete))
# sweeper call
deleteBatchComputeEnvironment(name, "INFO", 20 * time.Minute)
```

Otherwise we need get the resource default delete timeout in the sweeper call (if you know how to do this I'd love to know).

I don't mind deduping to the above, but I really think we should treat the timeout issue separately though, there is other precedence for it elsewhere:

https://github.com/terraform-providers/terraform-provider-aws/blob/0b7a4ff2fcf0b0276e884e712f9a734598b40bd6/aws/resource_aws_db_instance_test.go#L67

PR #3881, opened 2018-03-22
changed docs of aws_appautoscaling_target

GriffithBeliever:
changed docs of aws_appautoscaling_target to show which parts of the resource_id need to be inputed by the user

### DIFF
--- a/website/docs/r/appautoscaling_target.html.markdown
+++ b/website/docs/r/appautoscaling_target.html.markdown
@@ -18,7 +18,7 @@ Provides an Application AutoScaling ScalableTarget resource.
 resource "aws_appautoscaling_target" "dynamodb_table_read_target" {
   max_capacity       = 100
   min_capacity       = 5
-  resource_id        = "table/tableName"
+  resource_id        = "table/${aws_dynamodb_table.table_name.name}"
   role_arn           = "${data.aws_iam_role.DynamoDBAutoscaleRole.arn}"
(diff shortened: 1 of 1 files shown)


website/docs/r/appautoscaling_target.html.markdown
```diff
@@ -18,7 +18,7 @@ Provides an Application AutoScaling ScalableTarget resource.
 resource "aws_appautoscaling_target" "dynamodb_table_read_target" {
   max_capacity       = 100
   min_capacity       = 5
-  resource_id        = "table/tableName"
+  resource_id        = "table/${aws_dynamodb_table.table_name.name}"
```
bflad: Can you change this and the below reference to `${aws_dynamodb_table.example.name}` please?

website/docs/r/appautoscaling_target.html.markdown
```diff
@@ -44,7 +44,7 @@ resource "aws_appautoscaling_target" "dynamodb_index_read_target" {
 resource "aws_appautoscaling_target" "ecs_target" {
   max_capacity       = 4
   min_capacity       = 1
-  resource_id        = "service/clusterName/serviceName"
+  resource_id        = "service/${var.cluster_name}/${var.service_name}"
```
bflad: We should probably provide the resource references here as well: `service/${aws_ecs_cluster.example.name}/${aws_ecs_service.example.name}`

PR #3894, opened 2018-03-23
resource/aws_sns_platform_application: Remove hashing StateFunc on platform credential and principal

rhodrid:
When a APNS certificate is renewed and the private key remains the same the new certificate is detected and an attempt is made to update the SNS platform application. This fails because the hashed value for private key (platform credential) is sent along with the new certificate (platform principal) causing the AWS API to reject the request.

### DIFF
--- a/aws/resource_aws_sns_platform_application.go
+++ b/aws/resource_aws_sns_platform_application.go
@@ -1,7 +1,6 @@
 package aws
 
 import (
-	"crypto/sha256"
 	"fmt"
 	"log"
 	"strings"
@@ -60,7 +59,7 @@ func resourceAwsSnsPlatformApplication() *schema.Resource {
 			"platform_credential": {
 				Type:      schema.TypeString,
 				Required:  true,
-				StateFunc: hashSum,
+				Sensitive: true,
 			},
 			"arn": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


bflad:
Hi @rhodrid 👋 Hmmm. We'll likely need a different way to handle this as this PR will currently be a breaking change for anyone who is currently using the resource.

rhodrid:
Howdy 🖖I couldn't see another way to do it... but then I'm not very familiar with the code base.

rhodrid:
I have come across another fail state. If the certificate (platform principal) changes but the private key (platform credential) remains the same then the API call will fail.

```
Error updating SNS platform application: InvalidParameter: Invalid parameter: 
Attributes Reason: PlatformPrincipal attribute provided without PlatformCredential
```

My "solution" to this is to delete the resource and re-create it. It's not a problem for me at the moment as the resource is not customer facing at this stage of its life but it's a less than ideal way of dealing with it. 

Is there anyway to force sending both PlatformPrincipal & PlatformCredential?

rhodrid:
I've updated the PR so if platform_principal changes then platform_credential will be added to the request payload. As it stands this PR fixes the issues I was experiencing with this resource.

bflad:
Thanks so much for this, @rhodrid, its now been pulled in as part of our version 3.0.0. 👍

PR #3896, opened 2018-03-23
resource/aws_api_gateway_deployment: Do not delete stages in use by other deployments

mdlavin:
When a deployment has been updated in a way that forces a create/delete, the newly deployed deployment can be left in a bad state. Here how the code used to work:

1.  A new deployment is created that points to stage 'test'
2. The old deployment is marked for deletion and was associated with 'test'
3. The 'test' stage is deleted
4. The deposed deployment is deleted
5. The new deployment is now pointing at a stage that does not exist

Here is how the code works now:

1.  A new deployment is created that points to stage 'test'
2. The old deployment is marked for deletion and was associated with 'test'
3. The 'test' stage is not deleted because it is no longer associated with the deposed deployment
4. The deposed deployment is deleted
5. The new deployment is still happy =)

### DIFF
--- a/aws/resource_aws_api_gateway_deployment.go
+++ b/aws/resource_aws_api_gateway_deployment.go
@@ -172,14 +172,32 @@ func resourceAwsApiGatewayDeploymentDelete(d *schema.ResourceData, meta interfac
 
 	return resource.Retry(5*time.Minute, func() *resource.RetryError {
 		log.Printf("[DEBUG] schema is %#v", d)
-		if _, err := conn.DeleteStage(&apigateway.DeleteStageInput{
+
+		stage, err := conn.GetStage(&apigateway.GetStageInput{
 			StageName: aws.String(d.Get("stage_name").(string)),
 			RestApiId: aws.String(d.Get("rest_api_id").(string)),
-		}); err == nil {
-			return nil
+		})
+		if err != nil {
+			return resource.RetryableError(err)
 		}
 
-		_, err := conn.DeleteDeployment(&apigateway.DeleteDeploymentInput{
+		// If the stage has been updated to point at a different deployment, then
+		// the stage should not be removed then this deployment is deleted.
+		var shouldDeleteStage = true
+		if *stage.DeploymentId != d.Id() {
+			shouldDeleteStage = false
+		}
+
+		if shouldDeleteStage {
+			if _, err := conn.DeleteStage(&apigateway.DeleteStageInput{
+				StageName: aws.String(d.Get("stage_name").(string)),
+				RestApiId: aws.String(d.Get("rest_api_id").(string)),
+			}); err == nil {
+				return nil
+			}
+		}
+
+		_, err = conn.DeleteDeployment(&apigateway.DeleteDeploymentInput{
 			DeploymentId: aws.String(d.Id()),
 			RestApiId:    aws.String(d.Get("rest_api_id").(string)),
 		})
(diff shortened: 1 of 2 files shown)


mdlavin:
I just double checked and this PR still cleanly applies to master and passes the tests. Any chance of a merge soon before a conflict appears?

mdlavin:
If anybody is interested in testing out this feature in a patched v1.21.0 version, I've made some Alpine Linux x64 binaries available here: https://github.com/lifeomic/terraform-provider-aws/releases/tag/v1.21.0_patched_5f7d0def

mdlavin:
@bflad is there anything I can do to help this get merged into an official release?

bflad:
At first glance, I believe this change seems appropriate. We don't have too much precedent of resources that create other unmanaged resources though. 😅 

More forward looking: I am not convinced that we should allow the deployment resource to create stages at all (and hence delete them too), as it leaves Terraform management in a precarious state with unmanaged resources or resources that need to be imported. Unfortunately, I do not believe the API Gateway API provides any mechanism to prevent this behavior beyond only sending the deployment to the "default" stage.

aws/resource_aws_api_gateway_deployment.go
```diff
@@ -172,14 +172,32 @@ func resourceAwsApiGatewayDeploymentDelete(d *schema.ResourceData, meta interfac
...
-		if _, err := conn.DeleteStage(&apigateway.DeleteStageInput{
+
+		stage, err := conn.GetStage(&apigateway.GetStageInput{
 			StageName: aws.String(d.Get("stage_name").(string)),
 			RestApiId: aws.String(d.Get("rest_api_id").(string)),
-		}); err == nil {
-			return nil
+		})
+		if err != nil {
+			return resource.RetryableError(err)
```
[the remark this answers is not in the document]
mdlavin: I'll make the changes you suggest

mdlavin:
@bflad I believe that I have addressed all of your feedback. Let me know if there is anything else I should change.

aws/resource_aws_api_gateway_deployment_test.go
```diff
@@ -69,6 +112,32 @@ func testAccCheckAWSAPIGatewayDeploymentExists(n string, res *apigateway.Deploym
...
+
+		deploymentResource, ok := s.RootModule().Resources[deploymentName]
+		if !ok {
+			return fmt.Errorf("Deployment not found: %s", deploymentName)
+		}
+
+		restApiId := aws.String(deploymentResource.Primary.Attributes["rest_api_id"])
+
+		req := &apigateway.GetStageInput{
+			StageName: &stageName,
```
bflad: Nitpick: the stage name can be fetched from attributes as well (`deploymentResource.Primary.Attributes["stage_name"]`) instead of a parameter passed in from the acceptance test

aws/resource_aws_api_gateway_deployment.go
```diff
@@ -172,14 +172,38 @@ func resourceAwsApiGatewayDeploymentDelete(d *schema.ResourceData, meta interfac
...
+		stage, err := conn.GetStage(&apigateway.GetStageInput{
 			StageName: aws.String(d.Get("stage_name").(string)),
 			RestApiId: aws.String(d.Get("rest_api_id").(string)),
-		}); err == nil {
-			return nil
+		})
+		if err != nil {
+			// If the stage is already deleted then there is no need to attempt deletion
+			apigatewayErr, _ := err.(awserr.Error)
+			if apigatewayErr.Code() == "NotFoundException" {
```
bflad: Nitpick: we have a helper function for this type of logic and prefer the AWS Go SDK constants, e.g. `isAWSErr(err, apigateway.ErrCodeNotFoundException, "")`

PR #3934, opened 2018-03-27
Fix password generation in resourceAwsIamUserLoginProfile

KellerFuchs:
The current `generatePassword()` has massive issues:
- it uses the PRNG from `math/rand`, which is *not* secure;
- it seeds said PRNG with non-random data (the current time);
- the password generation method samples one character from each charset, then shuffles the results; this results in suboptimal password entropy, as for a 20-character long (for example), one would know it contains exactly 5 lowercase letters, 5 uppercase letters, 5 digits and 5 “special characters”.

I strongly suggest publishing an advisory after merging this (and updating the vendor-ed source in terraform), as it means that Terraform users have had IAM accounts provisionned with weak, easily-bruteforceable password.

### DIFF
--- a/aws/resource_aws_iam_user_login_profile.go
+++ b/aws/resource_aws_iam_user_login_profile.go
@@ -59,27 +59,14 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
 // characters that are likely to satisfy any possible AWS password policy
 // (given sufficient length).
 func generatePassword(length int) string {
-	charsets := []string{
-		"abcdefghijklmnopqrstuvwxyz",
-		"ABCDEFGHIJKLMNOPQRSTUVWXYZ",
-		"012346789",
-		"!@#$%^&*()_+-=[]{}|'",
-	}
+	charset := `abcdefghijklmnopqrstuvwxyz
+ABCDEFGHIJKLMNOPQRSTUVWXYZ
+012346789
+!@#$%^&*()_+-=[]{}|'`
 
-	// Use all character sets
-	components := make(map[int]byte, length)
-	for i := 0; i < length; i++ {
-		charset := charsets[i%len(charsets)]
-		components[i] = charset[rand.Int(len(charset))]
-	}
-
-	// Randomise the ordering so we don't end up with a predictable
-	// lower case, upper case, numeric, symbol pattern
 	result := make([]byte, length)
-	i := 0
-	for _, b := range components {
-		result[i] = b
-		i = i + 1
+	for i := 0; i < length; i++ {
+		result[i] = charset[rand.Int(len(charset))]
 	}
 
 	return string(result)
--- a/aws/resource_aws_iam_user_login_profile.go
+++ b/aws/resource_aws_iam_user_login_profile.go
@@ -1,9 +1,9 @@
 package aws
 
 import (
+	"crypto/rand"
 	"fmt"
 	"log"
-	"math/rand"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -67,11 +67,10 @@ func generatePassword(length int) string {
 	}
 
 	// Use all character sets
-	random := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
 	components := make(map[int]byte, length)
 	for i := 0; i < length; i++ {
 		charset := charsets[i%len(charsets)]
-		components[i] = charset[random.Intn(len(charset))]
+		components[i] = charset[rand.Int(len(charset))]
 	}
 
 	// Randomise the ordering so we don't end up with a predictable


aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,27 +59,14 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
-
-	// Randomise the ordering so we don't end up with a predictable
-	// lower case, upper case, numeric, symbol pattern
 	result := make([]byte, length)
-	i := 0
-	for _, b := range components {
-		result[i] = b
-		i = i + 1
+	for i := 0; i < length; i++ {
+		result[i] = charset[rand.Int(len(charset))]
```
RyanSquared: It's technically possible to have a password now that is exactly one character, if somehow `rand.Int()` always returns 0. I am not sure if there is a `crypto/rand` alternative to `rand.Perm()`, which returns a random permutation.

KellerFuchs: If by “exactly one character”, you mean the same, repeated characted, then yes it is, though it is vanishingly unlikely.

As `crypto/rand` uses OS-provided randomness (either a syscall like `getentropy(3)` or `/dev/urandom`), if it provides you with non-random numbers, you have a bigger problem.

KellerFuchs:
PS: Crediting @lrvick for pointing out the code fragment.

aws/resource_aws_iam_user_login_profile.go
RyanSquared: There's still the case where it could avoid required characters, like all uppercase, or required symbols.

KellerFuchs:
Clearly, I don't know how to use `crypto/rand` before my second coffee; fixing this in a jiffy.

KellerFuchs:
@bflad Can you escalate this to whoever deals with security at @hashicorp ?  It's a pretty bad one.

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,22 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
+!@#$%^&*()_+-=[]{}|'`
 
-	// Randomise the ordering so we don't end up with a predictable
-	// lower case, upper case, numeric, symbol pattern
 	result := make([]byte, length)
-	i := 0
-	for _, b := range components {
-		result[i] = b
-		i = i + 1
+	for i := 0; i < length; i++ {
```
jbardin: use `range` over `result` here

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,22 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
+		r, err := rand.Int(rand.Reader, big.NewInt(int64(len(charset))))
```
jbardin: create the  `big.Int` outside of the for loop.

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,22 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
+		if err != nil {
+			panic(err)
+		}
+		if r.IsInt64() {
```
jbardin: Invert this boolean and remove the else block

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,24 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
-	// Randomise the ordering so we don't end up with a predictable
-	// lower case, upper case, numeric, symbol pattern
 	result := make([]byte, length)
-	i := 0
-	for _, b := range components {
-		result[i] = b
-		i = i + 1
+	charset_size := big.NewInt(int64(len(charset)))
+
+	for i, _ := range result {
```
jbardin: One last thing -- the `, _` can be left out entirely here

KellerFuchs:
@jbardin Done  :)

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,24 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
+!@#$%^&*()_+-=[]{}|'`
 
-	// Randomise the ordering so we don't end up with a predictable
-	// lower case, upper case, numeric, symbol pattern
 	result := make([]byte, length)
-	i := 0
-	for _, b := range components {
-		result[i] = b
-		i = i + 1
+	charset_size := big.NewInt(int64(len(charset)))
```
jbardin: while we're at it, might as well use `charsetSize` here

jbardin:
Hi @KellerFuchs,

Thanks for taking the time to submit this!
I'd just like to point out for reference, our [Security Policy](https://www.hashicorp.com/security) and [Community Guidelines](https://www.hashicorp.com/community-guidelines), for others wishing to report security issues in the future, or to get in contact directly with our security team.

paultyng:
FWIW the character class code is probably there because someone could have an IAM password policy that enforces those checks, so this may start breaking for people, need to make sure that use case is considered.

https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_account-policy.html

KellerFuchs:
@jbardin Thanks a lot; I looked for a proper security contact before dumping this in a PR, but couldn't find it.
Perhaps add a link in `README.md` or `.github/*.md` ?

aws/resource_aws_iam_user_login_profile.go
paultyng: Yeah, I called this out as well, that was what I suspect the reasoning for the division of the character sets previously.

paultyng:
Something like this would satisfy the character class issues and uses `crypto/rand`: https://github.com/sethvargo/go-password

jbardin:
@paultyng, Yeah, I misread the previous code from the diff -- looking at the old code directly it seems that it was attempting to take each specific component, but it was not a fair distribution. 

I think it's probably better to take a normally distributed password generated like this and verify it it against the policy after. (or we can vet the package linked above)

paultyng:
I think the main issue with that package is that it allows for too many special characters, IAM only allows for `! @ # $ % ^ & * ( ) _ + - = [ ] { } | '` per https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_account-policy.html#password-policy-details

So possibly would need to fork that I guess to HC?

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,24 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
-	// Use all character sets
-	random := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
-	components := make(map[int]byte, length)
-	for i := 0; i < length; i++ {
-		charset := charsets[i%len(charsets)]
-		components[i] = charset[random.Intn(len(charset))]
-	}
+	charset := `abcdefghijklmnopqrstuvwxyz
+ABCDEFGHIJKLMNOPQRSTUVWXYZ
+012346789
```
paultyng: lol this is missing 5

paultyng: I guess that was missed quite a while ago :(

KellerFuchs:
@paultyng FWIW, the core of `go-password` does pretty much the same thing as this:
return a random element (as in, the elemant at a random index) from a “charset” buffer, and fill another buffer with the result, then return it as a string.

It does rejection sampling around that, to enforce “at least x {alphanum; symbols; ...}” constraints, but as far as I understand it's not needed here?  If it _really_ is, rejection sampling is the way to go, but that sort of policy can dangerousely reduce password entropy (and, BTW, NIST also advises against it).

jbardin:
Thanks @KellerFuchs,

I'm also not strongly in favor of the go-password algorithm, since it only inserts a set number of "optional" characters, meaning you don't get much additional entropy as from a normal distribution. I'm OK with rejection sampling of the results in this case, since we have no choice but to conform to the set policy in some way.

Thanks!

paultyng:
In this case we are unfortunately at the mercy of how IAM is configured, but it is possible to query IAM to see what the policy is before forcing the rejections.

https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.GetAccountPasswordPolicy

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -59,28 +59,24 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
-	random := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
-	components := make(map[int]byte, length)
-	for i := 0; i < length; i++ {
-		charset := charsets[i%len(charsets)]
-		components[i] = charset[random.Intn(len(charset))]
-	}
+	charset := `abcdefghijklmnopqrstuvwxyz
+ABCDEFGHIJKLMNOPQRSTUVWXYZ
+0123456789
+!@#$%^&*()_+-=[]{}|'`
```
mwhooker: I think you may want to make this something like

```
	charsets := strings.Join([]string{
		"abcdefghijklmnopqrstuvwxyz",
		"ABCDEFGHIJKLMNOPQRSTUVWXYZ",
		"012346789",
		"!@#$%^&*()_+-=[]{}|'",
	}, "")
```
since this has 3 newlines in it

https://play.golang.org/p/UkuPLAvmbuS

KellerFuchs:
Sorry for the slow replies, my laptop died earlier in the week so it took a while to be back online.
FYI, this issues was assigned CVE-2018-9057.

aws/resource_aws_iam_user_login_profile.go
KellerFuchs: Oops, thanks a lot.
I did pretty much what you did to test, but got luckier (and had a larger non-`\n` charset)  :/

KellerFuchs:
@paultyng As far as I understand, there is no default password policy, and users setting one can break the current code too?
If that's correct, shouldn't that go in a separate issue?

jbardin:
@KellerFuchs,

No worries, thanks for the update.

Yes, the only default in the policy is the minimum length of 6, and this particular PR doesn't need to deal with the policy directly. The old code would probably pass any reasonable policy, but that was at the expense of not generating good passwords.  I'll follow up with a separate PR regarding the policy checks.

jbardin:
Hi @KellerFuchs,

Thanks again for your work on this.
Since the charset needed to be updated, I moved your commit into a new branch and made the modifications there, along with the policy checking.  The new PR is #3989.

Closing this since it's been superseded.

jbardin:
@KellerFuchs, we are intending to merge the security related commit and the policy checks together to prevent breaking existing users. The policy work isn't going to delay the release.

KellerFuchs:
Mkay

PR #3942, opened 2018-03-28
resource/aws_autoscaling_group: Fix updating of service_linked_role

ungureanuvladvictor:
Currently terraform plan/apply show changes to the ASG but no real modifications are done to the infrastructure because the `opts` struct is unused at line 729.

### DIFF
--- a/aws/resource_aws_autoscaling_group.go
+++ b/aws/resource_aws_autoscaling_group.go
@@ -616,6 +616,10 @@ func resourceAwsAutoscalingGroupUpdate(d *schema.ResourceData, meta interface{})
 		}
 	}
 
+	if d.HasChange("service_linked_role_arn") {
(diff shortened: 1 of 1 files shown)


ungureanuvladvictor:
@bflad thx for getting this in! What is the usual cadence of releases for the AWS provider?

bflad:
@ungureanuvladvictor official releases are usually every two weeks: https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md

bflad:
As soon as I say that we then cut an out of band release. 😅  This has been released in [version 1.13.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1130-march-28-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #3944, opened 2018-03-28
resource/aws_kinesis_firehose_delivery_stream: Add Splunk destination…

WhileLoop:
… lambda transformation support

Partial fix for https://github.com/terraform-providers/terraform-provider-aws/issues/513

### DIFF
--- a/aws/resource_aws_kinesis_firehose_delivery_stream.go
+++ b/aws/resource_aws_kinesis_firehose_delivery_stream.go
@@ -121,9 +121,9 @@ func processingConfigurationSchema() *schema.Schema {
 											Required: true,
 											ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
 												value := v.(string)
-												if value != "LambdaArn" && value != "NumberOfRetries" {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_kinesis_firehose_delivery_stream.go
```diff
@@ -321,6 +321,10 @@ func flattenKinesisFirehoseDeliveryStream(d *schema.ResourceData, s *firehose.De
 				"retry_duration":             *destination.SplunkDestinationDescription.RetryOptions.DurationInSeconds,
 			}
 
+			if v := destination.SplunkDestinationDescription.ProcessingConfiguration; v != nil {
+				splunkConfiguration["processing_configuration"] = v
```
bflad: I missed this before merging. This should be `= flattenProcessingConfiguration(*v, roleArn)`. I will fix this in an upcoming PR.

WhileLoop: I was wondering about that, where is RoleArn supposed to come from in the Splunk Configuration?

bflad: Its only used as a default value to filter out a difference if it is undefined in the Terraform configuration under the parameters. In this case, passing it `""` is fine. I have a PR almost ready that includes this fix.

bflad:

PR #3960, opened 2018-03-28
resource/aws_dynamodb_table: Fix DynamoDB TTL state detection

### LINKED ISSUE #3463: DynamoDB TTL Infinitely Out of Sync
DynamoDB tables that have a `ttl` block with `enabled = false` will never be able to have their state in sync with the infrastructure.

### Terraform Version
```
Terraform v0.11.3
+ provider.aws v1.9.0
+ provider.template v1.0.0
```

### Affected Resource(s)
- aws_dynamodb_table

```hcl
provider "aws" {
  region = "us-west-2"
}

resource "aws_dynamodb_table" "dynamodb-composite-gsi-composite" {
  count = 1
  name  = "glen-test-ignore"

  read_capacity  = 10
  write_capacity = 10
  hash_key       = "hi"
  range_key      = "me"

  attribute {
    name = "hi"
    type = "S"
  }

  attribute {
    name = "me"
    type = "N"
  }

  attribute {
    name = "x"
    type = "S"
  }

  attribute {
    name = "y"
    type = "N"
  }

  ttl {
    enabled        = false
    attribute_name = ""
  }

  global_secondary_index {
    name            = "glen-test"
    hash_key        = "x"
    range_key       = "y"
    write_capacity  = "10"
    read_capacity   = "10"
    projection_type = "ALL"
  }
}
```

### Debug Output
https://gist.github.com/goakley/c682bf633544bc71cfbe4367df30bfbf

### Expected Behavior
Terraform should have determined that no changes needed to be applied, since the DyanmoDB TTL feature is already disabled on the table.

### Actual Behavior
Terraform wants to change the TTL to DISABLED, which will end up being a noop.  If this is run (and it will succeed), Terraform will determine that the exact same set of changes will need to be made on the next Terraform run.  That is, it will never think that the DynamoDB TTL feature is disabled even though it is.

### Steps to Reproduce
1. `terraform apply`
2. `terraform apply`

### Important Factoids
This example is a simplified version of our production code, which references variables to set the `enabled` property of the `ttl` block.  Many of our tables don't use TTL, so `enabled` is set to `false`, but this causes Terraform to forever think that the DynamoDB table is out of sync with the Terraform state.

### References
None.

goakley:
Fixes #3463 

DynamoDB has TTL state always set to DISABLED unless it is otherwise enabled.  The terraform logic decided that this meant if TTL is set to DISABLED, then there should be no `ttl {}` block at all.  This is incorrect, as there can be a `ttl {}` block with the setting `enabled = false`.  This change fixes that issue.

DynamoDB acceptance tests have been updated and run.

### DIFF
--- a/aws/resource_aws_dynamodb_table.go
+++ b/aws/resource_aws_dynamodb_table.go
@@ -134,6 +134,7 @@ func resourceAwsDynamoDbTable() *schema.Resource {
 						},
 					},
 				},
+				DiffSuppressFunc: suppressEquivalentDynamodbTableTtlDiffs,
 			},
 			"local_secondary_index": {
 				Type:     schema.TypeSet,
--- a/aws/data_source_aws_dynamodb_table_test.go
+++ b/aws/data_source_aws_dynamodb_table_test.go
@@ -25,6 +25,7 @@ func TestAccDataSourceAwsDynamoDbTable_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "range_key", "GameTitle"),
 					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "attribute.#", "3"),
 					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "global_secondary_index.#", "1"),
+					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "ttl.#", "1"),
 					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "tags.%", "2"),
 					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "tags.Name", "dynamodb-table-1"),
 					resource.TestCheckResourceAttr("data.aws_dynamodb_table.dynamodb_table_test", "tags.Environment", "test"),
@@ -66,6 +67,11 @@ func testAccDataSourceAwsDynamoDbTableConfigBasic(tableName string) string {
     read_capacity      = 10
     projection_type    = "INCLUDE"
     non_key_attributes = ["UserId"]
+  }
+
+  ttl {
+    enabled        = "true"
+    attribute_name = "GameTimeTTL"
   }
 
   tags {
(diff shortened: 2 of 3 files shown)


goakley:
@radeksimko Is there anything that could be improved to make this change more amicable?

danieladams456:
bump

aws/resource_aws_dynamodb_table.go
```diff
@@ -111,6 +111,7 @@ func resourceAwsDynamoDbTable() *schema.Resource {
 			"ttl": {
 				Type:     schema.TypeSet,
 				Optional: true,
+				Computed: true,
```
bflad: Generally speaking: setting `Computed: true` is not an ideal situation because it prevents Terraform from being able to detect any sort of API drift when an attribute is not defined in the configuration. In this case, people would most likely be expecting that Terraform let's them know when TTL is enabled on a DynamoDB table outside of Terraform when its disabled by default and its not defined in their configuration.

We'll need to figure out how to support this without `Computed: true` most likely. An option might be setting this attribute up like how the `aws_codebuild_project` `cache` attribute is with a `DiffSuppressFunc` and changing `enabled` to `Optional: true` (and potentially `Default: false`).

goakley: Are the `Computed`, `DiffSuppressFunc`, etc properties documented anywhere?  I could not find them and was not actually sure what `Computed` means.  I didn't know about `DiffSuppressFunc` - I copied this strategy from the `aws_s3_bucket` resource, but could take that approach instead.

goakley:
@bflad the CR now uses `DiffSuppressFunc` and still works as expected.

goakley:
Any chance this is acceptable now?

goakley:
@bflad Are there any issues with this approach?

goakley:
Is this something that shouldn't be fixed by design?

danieladams456:
This fix would be very useful.

pshastry:
This would be very helpful

goakley:
@bflad is there any chance this can get looked at again?

goakley:
Thanks @bflad!  I've been away until this week and I'm glad to see this was addressed.

bflad:
This has been released in [version 2.1.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#210-march-07-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

inge4pres:
Added some info regarding the TTL problem [here](https://github.com/terraform-providers/terraform-provider-aws/issues/10304#issuecomment-557541148)

PR #3964, opened 2018-03-28
docs: Fix broken WAF links

bflad:
Will push into `stable-website` after merge.


bflad:
Self-merging for immediate release.

Issue #3967, opened 2018-03-28
Expose execution ARN from aws_api_gateway_rest_api resource


mewa:
Hi,

Currently, in order to create for example a Lambda permission for API Gateway one has to create an `aws_api_gateway_deployment` resource and obtain the `execution_arn` attribute.

```hcl
resource "aws_lambda_permission" "lambda_permission" {
  statement_id  = "AllowAPIGatewayInvoke"
  action        = "lambda:InvokeFunction"
  function_name = "example_lambda"
  principal     = "apigateway.amazonaws.com"

  source_arn = "${aws_api_gateway_deployment.example_deployment.execution_arn}/*/*"
}
```
However, this approach couples the deployment (stage) with specific permission.

Alternatively one could use a block like this
```hcl
  source_arn = "${replace(aws_api_gateway_deployment.example_deployment.execution_arn, format("/%s$/", aws_api_gateway_deployment.example_deployment.stage_name), "*")}/*/*"
```
to calculate the ARN not restricted to specific stage, or even create the ARN by hand according to the specification.

But either way we end up with messy expressions and/or permission coupled to the deployment.

I feel that the `aws_api_gateway_rest_api` resource should expose `execution_arn` in a fashion similar to `aws_api_gateway_deployment`.

tdmalone:
I'd also love to see the `arn` exposed. Currently the `id` is exposed, but to build the `arn` you need to do something like:

    "arn:aws:apigateway:${data.aws_region.current.name}::/restapis/${aws_api_gateway_rest_api.api.id}"

bflad:
Support for this has been merged in via #3968 and will release with v1.19.0 of the AWS provider, likely middle of next week.

@tdmalone yes please 😄  Should be a very similar PR to this one

bflad:
This has been released in [version 1.19.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1190-may-16-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #3968, opened 2018-03-28
aws_api_gateway_rest_api: Add execute_arn computed attribute

mewa:
Added execute_arn attribute as described in #3967


aws/resource_aws_api_gateway_rest_api.go
```diff
@@ -152,6 +156,16 @@ func resourceAwsApiGatewayRestApiRead(d *schema.ResourceData, meta interface{})
 	d.Set("name", api.Name)
 	d.Set("description", api.Description)
 	d.Set("binary_media_types", api.BinaryMediaTypes)
+
+	restApiId := d.Id()
+	region := meta.(*AWSClient).region
+	accountId := meta.(*AWSClient).accountid
+	arn, err := buildApiGatewayExecutionARN(restApiId, region, accountId)
```
bflad: This function no longer exists as we have moved to prefer using the SDK `arn` package. Can you please build the ARN like the following (modifying as appropriate):

```go
executionArn := arn.ARN{
		Partition: meta.(*AWSClient).partition,
		Service:   "execute-api",
		Region:    meta.(*AWSClient).region,
		AccountID: meta.(*AWSClient).accountid,
		Resource:  fmt.Sprintf("%s/%s", restApiId, stageName),
	}.String()
	d.Set("execution_arn", executionArn)
```

website/docs/r/api_gateway_rest_api.html.markdown
```diff
@@ -12,13 +12,30 @@ Provides an API Gateway REST API.
...
+### Defining REST API
+
 ```hcl
 resource "aws_api_gateway_rest_api" "MyDemoAPI" {
   name        = "MyDemoAPI"
   description = "This is my API for demonstration purposes"
 }
 ```
 
+### Using REST API execution_arn to specify Lambda permissions
```
bflad: Can you please move this example to the `aws_lambda_permission` resource documentation instead?

mewa:
Hi @bflad, thank you for your response (and sorry for replying late). I have applied your suggestions and merged changes from master, so hopefully the PR should be good to go.

bflad:

Issue #3969, opened 2018-03-29
Add support for configuring CLI/API session durations on aws_iam_role


devonbleak:
AWS has recently added a setting for specifying the maximum session duration of API/CLI credentials for a role: 
- https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
- https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use.html#id_roles_use_view-role-max-session
- https://docs.aws.amazon.com/IAM/latest/APIReference/API_UpdateRole.html
- https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html

### Terraform Version
All versions.

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_iam_role

### Terraform Configuration Files
Recommend adding max_session_duration property to aws_iam_role
```hcl
resource "aws_iam_role" "role" {
  name = "role"
  max_session_duration = 43200
  assume_role_policy = <<EOF
    ...
EOF
}
```

### Expected Behavior
Role ends up with its max session duration as specified in code

### Actual Behavior
Property is not supported.

bflad:
This has been released in [version 1.14.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1140-april-06-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #3977, opened 2018-03-29
Add MaxSesessionDuration to iam_role for federated users

stephencoe:
# Description
Adding `max_session_duration` to `iam_role` fixes https://github.com/terraform-providers/terraform-provider-aws/issues/3976

# Changes
- bump aws provider
- add `max_session_duration`

### DIFF
--- a/aws/resource_aws_iam_role.go
+++ b/aws/resource_aws_iam_role.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsIamRole() *schema.Resource {
@@ -107,21 +108,10 @@ func resourceAwsIamRole() *schema.Resource {
 			},
 
 			"max_session_duration": {
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  3600,
-				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
-					value := v.(int64)
-					if value > 43200 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be greater than 43200", k))
-					}
-					if value < 3600 {
-						errors = append(errors, fmt.Errorf(
-							"%q cannot be less than than 3600", k))
-					}
-					return
-				},
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      3600,
+				ValidateFunc: validation.IntBetween(3600, 43200),
 			},
 		},
 	}
@@ -267,15 +257,15 @@ func resourceAwsIamRoleUpdate(d *schema.ResourceData, meta interface{}) error {
 	if d.HasChange("max_session_duration") {
 		roleMaxDurationInput := &iam.UpdateRoleInput{
 			RoleName:           aws.String(d.Id()),
-			MaxSessionDuration: aws.Int64(d.Get("assume_role_policy").(int64)),
+			MaxSessionDuration: aws.Int64(int64(d.Get("max_session_duration").(int))),
 		}
 		_, err := iamconn.UpdateRole(roleMaxDurationInput)
 		if err != nil {
-			if iamerr, ok := err.(awserr.Error); ok && iamerr.Code() == "NoSuchEntity" {
+			if isAWSErr(err, iam.ErrCodeNoSuchEntityException, "") {
 				d.SetId("")
 				return nil
 			}
-			return fmt.Errorf("Error Updating IAM Role (%s) Assume Role Policy: %s", d.Id(), err)
+			return fmt.Errorf("Error Updating IAM Role (%s) Max Session Duration: %s", d.Id(), err)
 		}
 	}
 
--- a/aws/resource_aws_iam_role.go
+++ b/aws/resource_aws_iam_role.go
@@ -105,6 +105,24 @@ func resourceAwsIamRole() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"max_session_duration": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  3600,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(int64)
+					if value > 43200 {
+						errors = append(errors, fmt.Errorf(
+							"%q cannot be greater than 43200", k))
+					}
+					if value < 3600 {
+						errors = append(errors, fmt.Errorf(
+							"%q cannot be less than than 3600", k))
+					}
+					return
+				},
+			},
 		},
 	}
 }
@@ -135,6 +153,10 @@ func resourceAwsIamRoleCreate(d *schema.ResourceData, meta interface{}) error {
 
 	if v, ok := d.GetOk("description"); ok {
 		request.Description = aws.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("max_session_duration"); ok {
+		request.MaxSessionDuration = aws.Int64(v.(int64))
 	}
 
 	var createResp *iam.CreateRoleOutput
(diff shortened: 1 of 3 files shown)


stephencoe:
this is my first PR into TF, I am unsure why CI failed 😕

bflad:
My recommendation is to split the SDK update into it's own PR first (generally we merge these really fast)

vendor/vendor.json
```diff
@@ -39,884 +39,892 @@
 			"revisionTime": "2017-07-27T15:54:43Z"
 		},
 		{
-			"checksumSHA1": "y1bmVkrw6lKwgq8z+S3GzO1F5JE=",
+			"checksumSHA1": "N+8JQQqczoq0UfwXrcVqqRDcb5E=",
```
bflad: Did you accidentally govendor the root? I think it'll complain about that. See any of the recent `Deps: Bump aws-sdk-go` PRs for how it's done (we're hoping to automate this shortly)

stephencoe: thanks, I will split that out

aws/resource_aws_iam_role.go
```diff
@@ -105,6 +105,24 @@ func resourceAwsIamRole() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"max_session_duration": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  3600,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					value := v.(int64)
```
loivis: how about replace custom function with existing validation.IntBetween() instead?

aws/resource_aws_iam_role.go
```diff
@@ -239,6 +264,21 @@ func resourceAwsIamRoleUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	if d.HasChange("max_session_duration") {
+		roleMaxDurationInput := &iam.UpdateRoleInput{
+			RoleName:           aws.String(d.Id()),
+			MaxSessionDuration: aws.Int64(d.Get("assume_role_policy").(int64)),
```
bflad: This looks like it has two issues:
* Copy-paste typo of which attribute to grab, should be `"max_session_duration"`
* `schema.TypeInt` returns a regular `int`, so I believe this needs to be `aws.Int64(int64(d.Get("max_session_duration").(int)))`

aws/resource_aws_iam_role.go
```diff
@@ -239,6 +264,21 @@ func resourceAwsIamRoleUpdate(d *schema.ResourceData, meta interface{}) error {
...
+		}
+		_, err := iamconn.UpdateRole(roleMaxDurationInput)
+		if err != nil {
+			if iamerr, ok := err.(awserr.Error); ok && iamerr.Code() == "NoSuchEntity" {
```
bflad: Two notes here:
* This can be simplified and use the SDK provided constant with `if isAWSErr(err, iam.ErrCodeNoSuchEntityException, "") {`
* The error message needs to be updated

aws/resource_aws_iam_role_test.go
```diff
@@ -158,6 +158,40 @@ func TestAccAWSIAMRole_force_detach_policies(t *testing.T) {
 	})
 }
 
+func TestAccAWSIAMRole_testMaxSessionDuration(t *testing.T) {
```
bflad: You can drop the second `test` before `MaxSessionDuration`

aws/resource_aws_iam_role_test.go
```diff
@@ -261,6 +295,39 @@ func testAccAddAwsIAMRolePolicy(n string) resource.TestCheckFunc {
 	}
 }
 
+func testMaxSessionDuration(rName string) string {
```
bflad: The signature here should be:
```go
func testAccCheckIAMRoleConfig_MaxSessionDuration(rName string, maxSessionDuration int) string {
```

Then accept the parameter in the HCL configuration:
```hcl
  max_session_duration = %d
```

So we can remove the other two test configurations.

website/docs/r/iam_role.html.markdown
```diff
@@ -49,6 +49,8 @@ The following arguments are supported:
   See [IAM Identifiers](https://docs.aws.amazon.com/IAM/latest/UserGuide/Using_Identifiers.html) for more information.
 * `description` - (Optional) The description of the role.
 
+* `max_session_duration` - (Optional) Max duration of a federated role
```
bflad: This should probably note that the duration is in seconds and defaults to 3600. The [SDK documentation](https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#CreateRoleInput) sums it up pretty well:
> The maximum session duration (in seconds) that you want to set for the specified role. If you do not specify a value for this setting, the default maximum of one hour is applied. This setting can have a value from 1 hour to 12 hours.

aws/resource_aws_iam_role.go
bflad: As noted above, this should be updated to `ValidateFunc: validation.IntBetween(3600, 43200)`

aws/resource_aws_iam_role_test.go
```diff
@@ -261,6 +295,39 @@ func testAccAddAwsIAMRolePolicy(n string) resource.TestCheckFunc {
...
+}
+`, rName)
+}
+
+func testMaxSessionDuration_too_long(rName string) string {
+	return fmt.Sprintf(`
+resource "aws_iam_role" "role" {
+  name   = "test-role-%s"
+	path = "/"
+	max_session_duration = 46800
```
bflad: When the tests are updated, this should test the hard edges of the validation at 43201 and 3599.

aws/resource_aws_iam_role_test.go
```diff
@@ -158,6 +158,40 @@ func TestAccAWSIAMRole_force_detach_policies(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testMaxSessionDuration(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSRoleExists("aws_iam_role.test", &conf),
+					testAccAddAwsIAMRolePolicy("aws_iam_role.test"),
+				),
+			},
+			{
+				Config: testMaxSessionDuration_too_long(rName),
```
bflad: We expect these configurations to fail, which can be accomplished with:

```go
{
  Config:      testAccCheckIAMRoleConfig_MaxSessionDuration(rName, 43201),
  ExpectError: regexp.MustCompile(`error message here`),
},
```

jonotto:
This will also fix duplicate enhancement request #3969

aws/resource_aws_iam_role.go
```diff
@@ -137,6 +145,10 @@ func resourceAwsIamRoleCreate(d *schema.ResourceData, meta interface{}) error {
 		request.Description = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("max_session_duration"); ok {
+		request.MaxSessionDuration = aws.Int64(v.(int64))
```
bflad:
```
=== RUN   TestAccAWSIAMRole_MaxSessionDuration
panic: interface conversion: interface {} is int, not int64

goroutine 227 [running]:
github.com/terraform-providers/terraform-provider-aws/aws.resourceAwsIamRoleCreate(0xc4202c3810, 0x3542900, 0xc4208a0500, 0xc4202c3810, 0x0)
	/Users/bflad/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_iam_role.go:149 +0x6da
```

I will fix this on merge.

aws/resource_aws_iam_role_test.go
```diff
@@ -158,6 +158,42 @@ func TestAccAWSIAMRole_force_detach_policies(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSRoleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckIAMRoleConfig_MaxSessionDuration(rName, 3700),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSRoleExists("aws_iam_role.test", &conf),
```
bflad:
```
--- FAIL: TestAccAWSIAMRole_MaxSessionDuration (6.39s)
	testing.go:518: Step 0 error: Check failed: Check 1/2 error: Not found: aws_iam_role.test
```

The resource is named `role` in the test configuration. I will fix this on merge.

bflad:

PR #3988, opened 2018-03-29
resource/aws_lambda_function: Add missing IAM retry in update

### LINKED ISSUE #3972: IAM eventual consistency race condition for aws_iam_role leaves lambda function in defunct state
To ensure continued functioning of my system through changes which require role replacement, I have used create_before_destroy = true with my lambda function lifecycles. However, I have found that in general, changes which require the role to be replaced fail on terraform apply with the error
```
InvalidParameterValueException: The role defined for the function cannot be assumed by Lambda
```

A second run a short time later it succeeds. When I look in the AWS console, between usages of apply, it appears to leave my Lambda in an indeterminate state with no execution role. When I look at Terraform's output, what has happened is:
* The new role is successfully created to replace the old one.
* The old role has been destroyed.
* The change of role for the Lambda function fails, citing "InvalidParameterValueException: The role defined for the function cannot be assumed by Lambda"

It looks like this might be because AWS returns from the role creation call before the new role is completely available to trusted AWS services such as Lambda.

I have been able to work around this issue by adding a sleep provisioner to every one of my roles, but this is not a tidy workaround:
```hcl
resource "aws_iam_role" "lambda_execution_role" {
  name_prefix = "${var.role_name_prefix}"
  assume_role_policy = "${data.aws_iam_policy_document.entrust_lambda_to_assume_function_role.json}"
  description = "${var.role_description}"
  path = "${var.path}"
  lifecycle {
    create_before_destroy = true
  }
  # AWS returns success for IAM change but change is not yet available for a few seconds. Sleep to miss the race condition failure.
  provisioner "local-exec" {
    interpreter = ["bash", "-c"]
    command = "sleep 30"
  }
}
```

### Terraform Version
Terraform v0.11.5
+ provider.archive v1.0.3
+ provider.aws v1.11.0
+ provider.local v1.1.0
+ provider.null v1.0.0
+ provider.template v1.0.0

### Affected Resource(s)
- aws_iam_role

### Expected Behavior
IAM role replacement with create_before_destroy = true provides zero downtime for AWS Lambda.

### Actual Behavior
Lambda function left in indeterminate state, unable to execute.

### Steps to Reproduce
```hcl
data "aws_iam_policy_document" "entrust_lambda_to_assume_function_role" {
  statement {
    effect = "Allow"
    principals {
      identifiers = ["lambda.amazonaws.com"]
      type        = "Service"
    }
    actions = ["sts:AssumeRole"]
  }
}

resource "aws_iam_role" "lambda_execution_role" {
  name_prefix = "testing"
  assume_role_policy = "${data.aws_iam_policy_document.entrust_lambda_to_assume_function_role.json}"
  lifecycle {
    create_before_destroy = true
  }
}
```
Refer to this IAM role resource with a aws_lambda_function resource. Create and apply this infrastructure. Now change the name_prefix field in aws_iam_role.

bflad:
Closes #3972 

This IAM retry is present during creation, but missing from updates:

https://github.com/terraform-providers/terraform-provider-aws/blob/0938826e1f34e9bfaabb3e03c9791035ed37529d/aws/resource_aws_lambda_function.go#L369-L372

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -710,6 +710,10 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
 			if err != nil {
 				log.Printf("[DEBUG] Received error modifying Lambda Function Configuration %s: %s", d.Id(), err)
 
+				if isAWSErr(err, "InvalidParameterValueException", "The role defined for the function cannot be assumed by Lambda") {
+					log.Printf("[DEBUG] Received %s, retrying UpdateFunctionConfiguration", err)
+					return resource.RetryableError(err)
+				}
 				if isAWSErr(err, "InvalidParameterValueException", "The provided execution role does not have permissions") {
 					log.Printf("[DEBUG] Received %s, retrying UpdateFunctionConfiguration", err)
 					return resource.RetryableError(err)


bflad:

PR #3989, opened 2018-03-29
update iam password generation

jbardin:
There are 2 major issues with the previous generatePassword() function:
- It does not use a cryptographically-secure RNG, meaning that its output
  is not guaranteed to be secret;
- It is seeded only from the current time, making it very easy to bruteforce
  the password for an adversary with (limited) information on the creation
  time of the account.

resource_aws_iam_user_login_profile: Make uniformly random passwords

This maximises the entropy of the resulting password, given a fixed
charset and length; given that the charset contains 2*26 + 10 + 20
distinct charactes, using length-21 passwords results in password
entropy above 128 bits.

Check that generated IAM passwords pass any specified
iam.PasswordPolicy. IAM password policies cannot set specific numbers of each characters
classes, nor can the forbid specific characters. Since the policy
restrictions are minimal, rejection sampling will quickly find even the
shortest passwords containning all characters classes. This can be
extended to lookup the account policy, but in practice that is probably
not required. The only real unknown in this case is the minimum password
length, which will be validated when setting the password, and catching
it earlier won't do any good.

Update the validation for password length, since the minimum is 6
characters.

### DIFF
--- a/aws/resource_aws_iam_user_login_profile.go
+++ b/aws/resource_aws_iam_user_login_profile.go
@@ -1,10 +1,10 @@
 package aws
 
 import (
+	"crypto/rand"
 	"fmt"
 	"log"
-	"math/rand"
-	"time"
+	"math/big"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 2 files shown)


jbardin:
@KellerFuchs, thanks for reminding me! I only moved it out of habit, and I need to remember that feature ;)

aws/resource_aws_iam_user_login_profile.go
```diff
@@ -55,35 +56,62 @@ func resourceAwsIamUserLoginProfile() *schema.Resource {
...
-	i := 0
-	for _, b := range components {
-		result[i] = b
-		i = i + 1
+	panic("failed to generate acceptable password")
+}
+
+// Check the generated password contains all character classes listed in the
+// IAM password policy.
+func checkPwdPolicy(pass []byte) bool {
```
KellerFuchs: Is “the password contains all character classes” actually required by AWS?
If not, please do not enforce that, as it reduces password entropy.

KellerFuchs:
Should generatePassword check whether the requested length and policy yield at least a (configurable?) level of entropy, and error-out if not?
Otherwise, you could easily run into the case where you want a 10 characters password, and the policy states it should have at least 4 lowercase characters, 3 uppercases, 3 digits, and 3 symbols, and run into the same results as the current code (minus the broken PRNG)

KellerFuchs:
OK, I did not realise the options in the policy were Booleans, and not arbitrary numbers; should be Good Enough© then, though I would rather run the numbers on the entropy of generated passwords, but I am not doing that at 1am.

aws/resource_aws_iam_user_login_profile_test.go
```diff
@@ -20,6 +21,17 @@ import (
 	"github.com/hashicorp/vault/helper/pgpkeys"
 )
 
+func TestGenerateIAMPassword(t *testing.T) {
+	p := generatePassword(6)
+
```
paultyng: add a check here that it is actually 6 chars :)

bflad:

KellerFuchs:
@jbardin @paultyng Thanks for getting this finished while I was unavailable  :)

FYI, I did a back-of-the-enveloppe computation, and the entropy of the generated passwords (with the policy) should be >= 6 bits/character, which isn't much lower than when generating without constraints (where it's ≃6.36b/char)

PR #3999, opened 2018-03-30
New Resource: aws_default_internet_gateway

ewbankkit:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/2436.

The default Internet Gateway in an AWS region is the sole attached Internet Gateway in that region's default VPC.

Acceptance tests:
```
 make testacc TEST=./aws/ TESTARGS='-run=TestAccAwsDefaultInternetGateway_'==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws/ -v -run=TestAccAwsDefaultInternetGateway_ -timeout 120m
=== RUN   TestAccAwsDefaultInternetGateway_basic
--- PASS: TestAccAwsDefaultInternetGateway_basic (14.24s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	14.255s
```

### DIFF
--- a/website/docs/r/default_internet_gateway.html.markdown
+++ b/website/docs/r/default_internet_gateway.html.markdown
@@ -1,0 +1,49 @@
+---
+layout: "aws"
+page_title: "AWS: aws_default_internet_gateway"
+sidebar_current: "docs-aws-resource-default-internet-gateway"
+description: |-
+  Provides a resource to manage a default VPC Internet Gateway.
+---
+
+# aws_default_internet_gateway
+
+Provides a resource to manage a [default VPC Internet Gateway](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/default-vpc.html#default-vpc-components)
+in the current region.
+
+The `aws_default_internet_gateway` behaves differently from normal resources, in that
+Terraform does not _create_ this resource, but instead "adopts" it into management.
+
+## Example Usage
+
+Basic usage with tags:
+
+```hcl
+resource "aws_default_internet_gateway" "default" {
+  tags {
+    Name = "Default IGW"
+  }
+}
+```
+
+## Argument Reference
+
+The arguments of an `aws_default_internet_gateway` differ from [`aws_internet_gateway`](internet_gateway.html) resources.
+Namely, the `vpc_id` argument is computed.
+The following arguments are still supported:
+
+* `tags` - (Optional) A mapping of tags to assign to the resource.
+
+### Removing `aws_default_internet_gateway` from your configuration
+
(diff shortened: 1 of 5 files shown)


website/docs/r/default_internet_gateway.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+layout: "aws"
+page_title: "AWS: aws_default_internet_gateway"
+sidebar_current: "docs-aws-resource-default-internet-gateway"
+description: |-
+  Provides a resource to manage a default VPC Internet Gateway.
+---
+
+# aws_default_internet_gateway
+
+Provides a resource to manage a [default VPC Internet Gateway](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/default-vpc.html#default-vpc-components)
```
jen20: Do you think we should provide a more strict warning, akin to https://www.terraform.io/docs/providers/aws/r/default_route_table.html, about the fact that Terraform cannot destroy this resource?

jen20:
Hi @ewbankkit! Thanks for the pull request, and sorry for the delay in getting it reviewed. The code looks good to me - I've left one minor comment on the documentation. If you could let me know your thoughts there we can get this merged.

website/docs/r/default_internet_gateway.html.markdown
ewbankkit: @jen20 I think that's already taken care of in the _Removing aws_default_internet_gateway from your configuration_ section below.

ewbankkit:
After thinking about this some more I think the most sustainable solution is to use `terraform import` to manage the default Internet Gateway.
See remarks here: https://github.com/terraform-providers/terraform-provider-aws/issues/2436#issuecomment-410259953.

e-moshaya:
@ewbankkit any reason why this wasn't merged? Terraform import into each state file is not really an option for us as this a manual step we'd like to avoid.

ewbankkit:
@e-moshaya As I remember my reasoning was:
* Limit resource sprawl
* There's a "workaround" via `terraform import`
* Unlike the other VPC default resources there's no concept in the AWS API of default IGW - Once the resource is imported it behaves exactly like any other IGW

e-moshaya:
@ewbankkit sorry to bring this up again but we've now integrated amazon config and have a rule to check all resources for specific tags. Unfortunately, this is the only resource that is not tagged. importing the resource is really not an option seeing as it's a manual step which defeats the purpose of terraform. Please could you reconsider this pull request?

ewbankkit:
@e-moshaya I'm happy for someone else to pick up the code.

PR #4018, opened 2018-04-03
Issue #3530 Add valid_from and valid_until for aws_spot_instance_request

saravanan30erd:


aws/resource_aws_spot_instance_request.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
+const awsSpotInstanceRequestTimeLayout = "2006-01-02T15:04:05Z"
```
bflad: Instead of creating our own time layout, can we please use `time.RFC3339` instead?

aws/resource_aws_spot_instance_request.go
```diff
@@ -84,6 +86,16 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
 				Default:  "terminate",
 				ForceNew: true,
 			}
+			s["valid_from"] = &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			}
```
bflad: Switching to RFC3339 gives us an easy plan-time validation function for both of these:

```go
ValidateFunc: validation.ValidateRFC3339TimeString,
```

aws/resource_aws_spot_instance_request.go
```diff
@@ -132,6 +144,22 @@ func resourceAwsSpotInstanceRequestCreate(d *schema.ResourceData, meta interface
 		spotOpts.LaunchGroup = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("valid_from"); ok {
+		valid_from, err := time.Parse(awsSpotInstanceRequestTimeLayout, v.(string))
+		if err != nil {
+			return err
+		}
+		spotOpts.ValidFrom = &valid_from
```
bflad: Minor nitpick: we prefer to use the AWS SDK provided functions for these, e.g. `aws.Time(valid_from)`

website/docs/r/spot_instance_request.html.markdown
```diff
@@ -61,6 +61,11 @@ Spot Instance Requests support all the same arguments as
   The duration period starts as soon as your Spot instance receives its instance ID. At the end of the duration period, Amazon EC2 marks the Spot instance for termination and provides a Spot instance termination notice, which gives the instance a two-minute warning before it terminates.
   Note that you can't specify an Availability Zone group or a launch group if you specify a duration.
 * `instance_interruption_behavior` - (Optional) Indicates whether a Spot instance stops or terminates when it is interrupted. Default is `terminate` as this is the current AWS behaviour.
+* `valid_until` - The end date and time of the request, in UTC ISO8601 format
```
bflad: Once we switch to RFC3339, we should link to its documentation, e.g. `in [RFC3339 format](https://tools.ietf.org/html/rfc3339#section-5.8)`

aws/resource_aws_spot_instance_request.go
saravanan30erd: Added the validation function from the same package(validators.go).

aws/resource_aws_spot_instance_request.go
saravanan30erd: Used the aws.Time function for pointer reference.

website/docs/r/spot_instance_request.html.markdown
saravanan30erd: Updated the document with RFC3339 change.

saravanan30erd: Hi @bflad Thank you for your detailed instructions. I want to contribute more for terraform, I would like to discuss with you directly. Is there any chat or slack group?

saravanan30erd:
Hi @bflad added the acceptance test for valid_until. All of the custom tests are passed. But the general test is failing.

make testacc TEST=./aws TESTARGS='-run=TestAccAWSSpotInstanceRequest_validUntil'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSpotInstanceRequest_validUntil -timeout 120m
=== RUN   TestAccAWSSpotInstanceRequest_validUntil
--- FAIL: TestAccAWSSpotInstanceRequest_validUntil (212.13s)
	testing.go:518: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		DESTROY/CREATE: aws_spot_instance_request.foo
                 .........
                 valid_from:                      "0001-01-01T00:00:00Z" => "" (forces new resource)
		 valid_until:                     "2018-04-07T03:16:45Z" => "2018-04-07T03:16:45Z"

Since terraform always expect empty(i.e no changes) during terraform plan after terraform apply.
But here valid_from forcing it to create new resource, because the default value is current time.
In real time if we configure both valid_from and valid_until then terraform plan & destroy works properly, because terraform tracks the same resource due to the value in valid_from. 

Not only this test, all the acceptance tests for aws_spot_instance_request are failing since the aws introduced these parameters(valid_from & valid_until).

make testacc TEST=./aws TESTARGS='-run=TestAccAWSSpotInstanceRequest_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSSpotInstanceRequest_basic -timeout 120m
=== RUN   TestAccAWSSpotInstanceRequest_basic

--- FAIL: TestAccAWSSpotInstanceRequest_basic (212.76s)
	testing.go:518: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		DESTROY/CREATE: aws_spot_instance_request.foo
                .............
                valid_from:                      "0001-01-01T00:00:00Z" => "" (forces new resource)
		  valid_until:                     "0001-01-01T00:00:00Z" => "" (forces new resource)

We can fix this by setting valid_from during acceptance test, but it seems test always expecting the spot request result as "fulfilled" and if we set valid_from, actual result of the request is "not-scheduled-yet", so the test is failing.

bflad:
@saravanan30erd for now add `Computed: true,` to both attributes in the schema 👍

saravanan30erd:
@bflad thanks. Changed valid_from and valid_until as computed attributes, so tests are also passed.

bflad:

PR #4034, opened 2018-04-03
Removed unneccesary s3_bucket_object tagging  error

scottwinkler:
It is possible to put an object in an s3 bucket but not be able to read from it, if the person creating the object is in a different account than the one the s3 bucket exists in, and depending on the policy of the s3 bucket and so forth. In this particular case, the resource would create properly, but then fail on the last step of the resourceAwsS3BucketObjectRead() - there is an unnecessary error about not being able to read the tags, which doesn't even matter because the tag output was set earlier in resourceAwsS3BucketObjectPut(). I changed this to print a debug statement, and then continue as normal. This follows the behavior of the aws cli and makes sense from a user perspective.

### DIFF
--- a/aws/resource_aws_s3_bucket_object.go
+++ b/aws/resource_aws_s3_bucket_object.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
-	"github.com/mitchellh/go-homedir"
+	homedir "github.com/mitchellh/go-homedir"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -325,9 +325,10 @@ func resourceAwsS3BucketObjectRead(d *schema.ResourceData, meta interface{}) err
 				Key:    aws.String(key),
 			})
 		if err != nil {
-			return fmt.Errorf("Failed to get object tags (bucket: %s, key: %s): %s", bucket, key, err)
+			log.Printf("[DEBUG] Failed to get object tags (bucket: %s, key: %s): %s", bucket, key, err)
+		} else {
+			d.Set("tags", tagsToMapS3(tagResp.TagSet))
 		}
-		d.Set("tags", tagsToMapS3(tagResp.TagSet))
 	}
 
 	return nil


Fasian1:
pls fix

bflad:
This change would allow Terraform to silently not detect drift of the object tags from a Terraform configuration, breaking a main tenet of Terraform resource expected functionality.

Take the following scenario:
* Create `aws_s3_bucket_object`
* Outside of Terraform (web console, CLI, etc) adjust object tags
* Run `terraform plan`

The operator will not receive any indication of `GetObjectTagging` errors (without looking through `DEBUG` logs, which most people don't do) and will not see `aws_s3_bucket_object` differences with tags. For most (all?) Terraform resources we should be returning errors when Terraform cannot fully read resource information from the API.

As an edge case, we could consider accepting a flag to this behavior, but this becomes a one-off scenario and implementation for that behavior.

Thoughts?

scottwinkler:
@bflad I just think it's odd that the behavior of this resource is different than what I would see if I created and deleted the exact same resource through the command line or cloudformation. I understand your concern about object tagging drift, but according to the api documentation, the only reason this api call would fail is if the invoker doesn't have permissions to read the object taggings, which corresponds with the conditions of this edge case. Therefore I would think it would be fine to silently fail in this case. If you would like, we could test for a 403 status code of the error to determine whether to throw it or print a debug statement. I do not think creating a special flag  for this would be necessary.

bflad:
> I just think it's odd that the behavior of this resource is different than what I would see if I created and deleted the exact same resource through the command line or cloudformation.

There is an important distinction here -- the AWS CLI is fire and forget by design. [As noted in the Terraform introduction documentation](https://www.terraform.io/intro/index.html#what-is-terraform-):

> Terraform generates an execution plan describing what it will do to reach the desired state, and then executes it to build the described infrastructure. As the configuration changes, Terraform is able to determine what changed and create incremental execution plans which can be applied.

The wording "configuration changes" there applies to both the Terraform configuration as well as the "physical" resource state. Its important to be able to track the lifecycle of a resource by refreshing its current state.

CloudFormation is a much more apt example for comparison. I was not able to find a [CloudFormation resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) to compare the Terraform resource behavior to CloudFormation.

scottwinkler:
I notice in other terraform resources of the aws provider it is typical behavior to return an error if the tags cannot be read. I don't agree with this, but I think it is better to be consistent than inconsistent. In this case I will work around the problem simply by granting permissions to read tags to the account that is creating the s3_bucket_object.

PR #4037, opened 2018-04-04
resource/aws_launch_configuration: Clean up code

radeksimko:
This is just a first step in bigger refactoring of the Launch Configuration resource with the aim to unblock https://github.com/terraform-providers/terraform-provider-aws/pull/2132


Zordrak:
Per https://github.com/terraform-providers/terraform-provider-aws/pull/2973#issuecomment-386282890 I am very concerned that it is no longer possible to specify the user_data attribute of a launch configuration without also providing a value. This prevents user_data from being conditionally specified in a module unless a non-zero default is specified; however changing the default will not only force a recreation of all launch configurations without userdata, but it also removes a perfectly valid AWS configuration from being supported by terraform unless you omit the user_data parameter. Since terraform does not and cannot support the conditional presence of parameters, it completely relies on the ability to detect an empty string as a null value where the parameter is optional in the API call.

PR #4062, opened 2018-04-05
Adding documentation for SSM Command as a Cloudwatch event target

stephencoe:
This PR is adding documentation around using SSM RunCommand as a `cloudwatch_event_target`

[There is an open issue](https://github.com/terraform-providers/terraform-provider-aws/issues/3437) for it to become available but it is possible old as the work to add the functionality was available shortly after the date of the original issue

### DIFF
--- a/website/docs/r/cloudwatch_event_target.html.markdown
+++ b/website/docs/r/cloudwatch_event_target.html.markdown
@@ -124,7 +124,7 @@ DOC
 }
 
 resource "aws_cloudwatch_event_rule" "stop_instances" {
-  name                = "StopInstance
+  name                = "StopInstance"
   description         = "Stop instances nightly"
   schedule_expression = "cron(0 0 * * ? *)"
 }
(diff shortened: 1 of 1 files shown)


stephencoe:
actually, having looked back at this I realise that my implementation is using a custom document opposed to an existing RunCommand. 

I have updated the documentation to include both methods it solves the issue around using RunCommands along with using custom documents. Maybe a little verbose.

website/docs/r/cloudwatch_event_target.html.markdown
```diff
@@ -54,6 +54,120 @@ resource "aws_kinesis_stream" "test_stream" {
...
+          }
+        ]
+      }
+    }
+  }
+DOC
+}
+
+resource "aws_cloudwatch_event_rule" "stop_instances" {
+  name                = "StopInstance
```
masstamike: Missing a `"` at the end of this line.

PR #4086, opened 2018-04-06
Set subnet_mapping element attributes to ForceNew

mmdriley:
The way `diffSet` in Terraform works on `Resource`, the `ForceNew` property
on the Set has no effect unless `ForceNew` is also set on the resource's
attributes.

Without `ForceNew` on these attributes, changing `subnet_mappings` just
silently does nothing.

### DIFF
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -103,10 +103,12 @@ func resourceAwsLb() *schema.Resource {
 						"subnet_id": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 						"allocation_id": {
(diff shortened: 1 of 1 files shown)


mmdriley:
Example output without this change:

```
  ~ update in-place

  ~ aws_lb.test
      enable_http2:                            "" => "true"
      idle_timeout:                            "" => "60"
      subnet_mapping.2493713080.allocation_id: "" => ""
      subnet_mapping.2493713080.subnet_id:     "" => "subnet-d3d764ab"
      subnet_mapping.3213777275.allocation_id: "" => ""
      subnet_mapping.3213777275.subnet_id:     "subnet-d3d764aa" => ""


```

bflad:

PR #4106, opened 2018-04-06
ForceNew on Lambda alias name change

### LINKED ISSUE #4090: Lambda alias name being updated in-line
Opening this issue as discussed is https://github.com/terraform-providers/terraform-provider-aws/issues/3192#issuecomment-371764787 where the provider attempts to perform a lambda alias' rename, which is not even an allowed operation.

### Terraform Version
0.11.5

### Affected Resource(s)
aws_lambda_alias

### Terraform configuration files
Please see https://github.com/terraform-providers/terraform-provider-aws/issues/3192#issue-292762854

### Expected Behavior
A new alias should be created - the name of an alias should be treated as an unique identifier.

### Actual Behavior
An in line update is attempted, which fails with a 404 from the API:
```
* lambda.aws_lambda_alias.version: 1 error(s) occurred:

* aws_lambda_alias.version: Error updating Lambda alias: ResourceNotFoundException: Alias not found: arn:aws:lambda:us-east-1:999999999999:function:sample-lambda:MASTER-121
        status code: 404, request id: ca3401b9-237d-11e8-8e0e-05a5757ee33f
```

### Steps to Reproduce
1. Create and apply a lambda and tag it with aliases according to the scheme shown at the top.
2. Change the lambda_version variable.
3. terraform apply

### References
https://github.com/terraform-providers/terraform-provider-aws/issues/3192
https://github.com/terraform-providers/terraform-provider-aws/issues/3192#issuecomment-371764787

catsby:
When updating a Lamda Alias, the API uses the `name` as the ID of the alias to update. When updating the `name` of `aws_lambda_alias`, we're sending the new `name` from the config, which isn't found, so Terraform errors. 

The next `plan` reads the new `name` from the config, and shows an addition to be made (not an update). Applying then creates a new alias, and orphans the old one.

91d8b16 is a regression test that will fail on the current `master` (84353e40d03ac9a66168ea07a9d4368c9e45189e)

Fixes #4090

### DIFF
--- a/aws/resource_aws_lambda_alias_test.go
+++ b/aws/resource_aws_lambda_alias_test.go
@@ -40,6 +40,44 @@ func TestAccAWSLambdaAlias_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSLambdaAlias_nameupdate(t *testing.T) {
+	var conf lambda.AliasConfiguration
+
+	rString := acctest.RandString(8)
+	roleName := fmt.Sprintf("tf_acc_role_lambda_alias_basic_%s", rString)
+	policyName := fmt.Sprintf("tf_acc_policy_lambda_alias_basic_%s", rString)
+	attachmentName := fmt.Sprintf("tf_acc_attachment_%s", rString)
+	funcName := fmt.Sprintf("tf_acc_lambda_func_alias_basic_%s", rString)
+	aliasName := fmt.Sprintf("tf_acc_lambda_alias_basic_%s", rString)
+	aliasNameUpdate := fmt.Sprintf("tf_acc_lambda_alias_basic_%s", acctest.RandString(8))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsLambdaAliasDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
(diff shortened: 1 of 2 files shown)


catsby:
Tests passing:

screen shot 2018-04-06 at 1 13 15 pm

aws/resource_aws_lambda_alias_test.go
```diff
@@ -40,6 +40,44 @@ func TestAccAWSLambdaAlias_basic(t *testing.T) {
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsLambdaAliasDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAwsLambdaAliasConfig(roleName, policyName, attachmentName, funcName, aliasName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsLambdaAliasExists("aws_lambda_alias.lambda_alias_test", &conf),
+					testAccCheckAwsLambdaAttributes(&conf),
+					resource.TestMatchResourceAttr("aws_lambda_alias.lambda_alias_test", "arn",
+						regexp.MustCompile(`^arn:aws:lambda:[a-z]+-[a-z]+-[0-9]+:\d{12}:function:`+funcName+`:`+aliasName+`$`)),
```
bflad: Future note: When we start testing outside the `aws` partition (e.g. US Government) these will break. We can go with something like `^arn:[^:]+:lambda:`..., create a partition regex constant, or create our own test check wrapper function for AWS ARNs.

bflad:

PR #4108, opened 2018-04-06
New Resource: aws_ses_domain_identity_verification

vexingcodes:
# Description

This PR introduces a new resource that ensures a domain identity in SES has passed verification. This can be used to create SES domain identities in a fully automated fashion as described in the documentation updates. This resource is quite similar to `aws_acm_certificate_validation`.

Since this requires modification of records in Route 53 corresponding to a real domain, a new environment variable `SES_DOMAIN_IDENTITY_ROOT_DOMAIN` has been introduced for the acceptance tests to specify the domain against which the records should be created. If the environment variable is not specified the acceptance test will be skipped with a warning. The basic acceptance test relies on propagation of DNS records, so it can take quite a while to complete.

# Test Results

```
TF_ACC=1 go test ./aws -v -run=TestAccAwsSesDomainIdentityVerification_ -timeout 120m
=== RUN   TestAccAwsSesDomainIdentityVerification_basic
--- PASS: TestAccAwsSesDomainIdentityVerification_basic (550.43s)
=== RUN   TestAccAwsSesDomainIdentityVerification_timeout
--- PASS: TestAccAwsSesDomainIdentityVerification_timeout (12.41s)
=== RUN   TestAccAwsSesDomainIdentityVerification_nonexistent
--- PASS: TestAccAwsSesDomainIdentityVerification_nonexistent (3.71s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       566.569s
```

I built the documentation and it looks good to me.

### DIFF
--- a/aws/resource_aws_ses_domain_identity_verification.go
+++ b/aws/resource_aws_ses_domain_identity_verification.go
@@ -1,0 +1,117 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/service/ses"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsSesDomainIdentityVerification() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesDomainIdentityVerificationCreate,
+		Read:   resourceAwsSesDomainIdentityVerificationRead,
+		Delete: resourceAwsSesDomainIdentityVerificationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"domain": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(45 * time.Minute),
+		},
+	}
+}
+
+func getAttributes(conn *ses.SES, domainName string) (*ses.IdentityVerificationAttributes, error) {
+	input := &ses.GetIdentityVerificationAttributesInput{
+		Identities: []*string{
+			aws.String(domainName),
+		},
+	}
+
+	response, err := conn.GetIdentityVerificationAttributes(input)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting identity verification attributes: %s", err)
+	}
+
+	return response.VerificationAttributes[domainName], nil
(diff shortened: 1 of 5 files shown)


aws/resource_aws_ses_domain_identity_verification.go
```diff
@@ -0,0 +1,117 @@
...
+		att, err := getAwsSesIdentityVerificationAttributes(conn, domainName)
+		if err != nil {
+			return resource.NonRetryableError(fmt.Errorf("Error getting identity verification attributes: %s", err))
+		}
+
+		if att == nil {
+			return resource.NonRetryableError(fmt.Errorf("SES Domain Identity %s not found in AWS", domainName))
+		}
+
+		if *att.VerificationStatus != "Success" {
```
bflad: To prevent a panic in the scenario that `att.VerificationStatus` is `nil`, we should wrap this in `aws.StringValue(att.VerificationStatus)`

aws/resource_aws_ses_domain_identity_verification.go
```diff
@@ -0,0 +1,117 @@
...
+		Resource:  fmt.Sprintf("identity/%s", d.Id()),
+	}.String()
+	d.Set("arn", arn)
+
+	return nil
+}
+
+func resourceAwsSesDomainIdentityVerificationDelete(d *schema.ResourceData, meta interface{}) error {
+	// No need to do anything, domain identity will be deleted when aws_ses_domain_identity is deleted
+	d.SetId("")
```
bflad: `d.SetId("")` is unnecessary in delete functions 👍  (see also #4191)

aws/resource_aws_ses_domain_identity_verification_test.go
```diff
@@ -0,0 +1,176 @@
...
+			return fmt.Errorf("Incorrect ARN: expected %q, got %q", expected, rs.Primary.Attributes["arn"])
+		}
+
+		return nil
+	}
+}
+
+func testAccAwsSesDomainIdentityVerification_basic(rootDomain string, domain string) string {
+	return fmt.Sprintf(`
+resource "aws_route53_zone" "zone" {
```
bflad: Since we need to wire this to an existing public Route53 zone that is tied to an actual registrar with the correct NS records, this should be the data source instead of the resource. 👍

website/docs/r/ses_domain_identity_verification.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+* `arn` - The ARN of the domain identity.
+
+## Timeouts
+
+`acm_ses_domain_identity_verification` provides the following [Timeouts](/docs/configuration/resources.html#timeouts)
+configuration options:
+
+- `create` - (Default `45m`) How long to wait for a domain identity to be verified.
+
+## Example Usage
```
bflad: Nitpick: usually the Example Usage is above the arguments (one day we'll have automated documentation 😄 )

PR #4111, opened 2018-04-07
Feature/add enabled cloudwatch logs exports param for DB Instances

kterada0509:
# Description
This PR implementation a new param `enabled_cloudwatch_logs_exports` to `aws_db_instance` for managing [Exporting Cloudwatch logs Configuration](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/AuroraMySQL.Integrating.CloudWatch.html).

# Test Results

```
$ TF_ACC=1 go test ./aws -v -run=TestAccAWSDBInstance_cloudwatchLogsExportConfiguration -timeout 120m
=== RUN   TestAccAWSDBInstance_cloudwatchLogsExportConfiguration
--- PASS: TestAccAWSDBInstance_cloudwatchLogsExportConfiguration (484.14s)
=== RUN   TestAccAWSDBInstance_cloudwatchLogsExportConfigurationUpdate
--- PASS: TestAccAWSDBInstance_cloudwatchLogsExportConfigurationUpdate (755.77s)
PASS
ok     	github.com/terraform-providers/terraform-provider-aws/aws      	1239.953s
```

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsDbInstance() *schema.Resource {
@@ -351,6 +352,21 @@ func resourceAwsDbInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"enabled_cloudwatch_logs_exports": {
+				Type:     schema.TypeList,
+				Computed: false,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
+						"audit",
+						"error",
+						"general",
+						"slowquery",
+					}, false),
+				},
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
@@ -408,6 +424,10 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			opts.DBSubnetGroupName = aws.String(attr.(string))
 		}
 
+		if attr, ok := d.GetOk("enabled_cloudwatch_logs_exports"); ok && len(attr.([]interface{})) > 0 {
+			opts.EnableCloudwatchLogsExports = expandStringList(attr.([]interface{}))
+		}
+
 		if attr, ok := d.GetOk("kms_key_id"); ok {
 			opts.KmsKeyId = aws.String(attr.(string))
 			if arnParts := strings.Split(v.(string), ":"); len(arnParts) >= 4 {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_db_instance.go
```diff
@@ -351,6 +352,21 @@ func resourceAwsDbInstance() *schema.Resource {
 				Computed: true,
 			},
 
+			"enabled_cloudwatch_logs_exports": {
+				Type:     schema.TypeList,
+				Computed: false,
```
bflad: Minor nitpick: `Computed: false` is the default 👍

aws/resource_aws_db_instance.go
```diff
@@ -1151,10 +1189,55 @@ func resourceAwsDbInstanceStateRefreshFunc(id string, conn *rds.RDS) resource.St
 	}
 }
 
+func buildRDSARN(identifier, partition, accountid, region string) (string, error) {
```
bflad: This likely snuck back in during a rebase. I can remove it on merge 👍

aws/resource_aws_db_instance_test.go
```diff
@@ -1436,6 +1521,238 @@ resource "aws_db_instance" "bar" {
...
 `, acctest.RandInt())
 
+func testAccAWSDBInstanceConfigCloudwatchLogsExportConfiguration(rInt int) string {
+	return fmt.Sprintf(`
+
+	resource "aws_vpc" "foo" {
+		cidr_block           = "10.1.0.0/16"
+		enable_dns_hostnames = true
+		tags {
+		  Name = "terraform-testacc-db-instance-mssql-timezone"
```
bflad: Likely copy paste from other tests, will fix on merge. 👍

aws/resource_aws_db_instance.go
```diff
@@ -775,6 +803,10 @@ func resourceAwsDbInstanceRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("monitoring_role_arn", v.MonitoringRoleArn)
 	}
 
+	if v.EnabledCloudwatchLogsExports != nil {
+		d.Set("enabled_cloudwatch_logs_exports", v.EnabledCloudwatchLogsExports)
```
bflad: This is silently failing as `d.Set()` does not know how to handle a `[]*string` directly. You can see this from the import test:

```
--- FAIL: TestAccAWSDBInstance_importBasic (535.11s)
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=3) {
         (string) (len=33) "enabled_cloudwatch_logs_exports.#": (string) (len=1) "2",
         (string) (len=33) "enabled_cloudwatch_logs_exports.0": (string) (len=5) "audit",
         (string) (len=33) "enabled_cloudwatch_logs_exports.1": (string) (len=5) "error"
        }
```

A general recommendation is to always perform error checking when setting non-scalar types. This can be caught adding an import test step when working with new attributes or using `TF_SCHEMA_PANIC_ON_ERROR=1` while running the acceptance testing.

This should be something like:

```go
if err := d.Set("enabled_cloudwatch_logs_exports", flattenStringList(v.EnabledCloudwatchLogsExports)); err != nil {
  return fmt.Errorf("error setting enabled_cloudwatch_logs_exports: %s", err)
}
```

I will fix this on merge 👍

PR #4122, opened 2018-04-09
website: Standalone website

radeksimko:
## How

```
$ make website
==> Starting website in Docker...
...
```

and voilà!

screen shot 2018-04-09 at 09 27 19

## Caveats

Any links outside of AWS provider (more or less obviously) will 404.

## Motivation

This is to make it possible to serve the provider part of docs without having to pull other 70+ providers as submodules into one giant repository.

It should lower the barrier for maintainers/contributors for checking rendered website before submitting PR or as part of review, which seems to be a thing not many people do since provider split 🙈 , which in turn leads to rendering errors, broken links etc. 💥 

I'm raising this PR to AWS provider first, but the idea is to add it to all other provider repos afterwards (just with different volume mounts).

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,6 +1,7 @@
 SWEEP?=us-east-1,us-west-2
 TEST?=./...
 GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
+WEBSITE_REPO="github.com/hashicorp/terraform-website"
 
 default: build
 
@@ -46,5 +47,12 @@ test-compile:
 	fi
 	go test -c $(TEST) $(TESTARGS)
 
-.PHONY: build sweep test testacc vet fmt fmtcheck errcheck vendor-status test-compile
+website:
+ifneq (,$(wildcard "$(GOPATH)/src/$WEBSITE_REPO"))
+	echo "$(WEBSITE_REPO) not found in your GOPATH (necessary for layouts and assets), get-ting..."
+	go get $WEBSITE_REPO
+endif
+	@$(MAKE) -C $(GOPATH)/src/$(WEBSITE_REPO) website-provider PROVIDER_PATH=$(shell pwd) PROVIDER_NAME=aws
+
+.PHONY: build sweep test testacc vet fmt fmtcheck errcheck vendor-status test-compile website
 


bflad:
I'm generally 👍  this idea although I am wondering a few things:
* Some parts of this (e.g. middleman version, Docker flags) could be really hard to consistently maintain across all the provider repositories. How do we keep them in sync?
* We don't mount Gemfile/Gemfile.lock into the container. Do we need them?

Could this instead live in the terraform-website repository as a separate Makefile target there for building just a provider (e.g. `PROVIDER=aws make provider-website`). We already have a hard dependency on that repository and it would centralize the complexity.

radeksimko:
> Some parts of this (e.g. middleman version, Docker flags) could be really hard to consistently maintain across all the provider repositories. How do we keep them in sync?

I totally agree. This is related, although separate problem we need to address too. I don't have a solution in mind (yet), but I will say that the main goal would be to extract `/scripts/*` from all provider repos and eventually move more logic there which may include these docker commands too.

I believe that makefiles themselves should remain in provider repos with all their targets though, mostly just referencing shared shell scripts (kind of like `errcheck` or `fmtcheck` targets today).

> We don't mount Gemfile/Gemfile.lock into the container. Do we need them?

I believe we do, as part of the `website` directory we mount?

```
--volume "$(shell pwd)/website:/website"
```

> Could this instead live in the terraform-website repository as a separate Makefile target there for building just a provider (e.g. PROVIDER=aws make provider-website). We already have a hard dependency on that repository and it would centralize the complexity. The provider Makefile could call go get and make against that repository.

As a user/developer I'd rather prefer if I can just spin up the relevant website/docs straight from the repo where that relevant documentation lives instead of having to learn about another repo and having to clone it and learn how to spin up the website for a particular provider. Also, most importantly I think that having to commit & push & update submodule in order to just see what the website looks like is a bit heavy dev experience with long feedback cycles which is exactly the reason why people don't do it, IMO.

You could argue we can avoid submodules the same way I did here, but I just took advantage of the fact that `hashicorp/terraform-website` is A) changing significantly less often and/or B) is changing in a way that it in most cases doesn't affect the relevant provider section, so the risk of having outdated `hashicorp/terraform-website` is much lower compared to the risk of having outdated provider repos in `hashicorp/terraform-website`. Also note that I intentionally don't run `go get -u` as I expect the developer the maintain it and avoid overwriting anything.

paultyng:
Instead of a hardcoded docker run, why not just write this up as a `Dockerfile` to build instead?

bflad:
I had suggested offline that we instead enhance the existing terraform-website Makefile target, then each repository can just call it. I verified this locally.

terraform-provider-aws Makefile:
```
WEBSITE_REPO=github.com/hashicorp/terraform-website
website:
ifneq (,$(wildcard "$(GOPATH)/src/$(WEBSITE_REPO)"))
    @echo "$(WEBSITE_REPO) not found in your GOPATH (necessary for layouts and assets), get-ting..."
    go get $(WEBSITE_REPO)
endif
    @$(MAKE) -C $(GOPATH)/src/$(WEBSITE_REPO) website
```

terraform-website Makefile:
```
INSTALLED_PROVIDERS_BASE_DIR=$(GOPATH)/src/github.com/terraform-providers
INSTALLED_PROVIDERS=$(wildcard $(INSTALLED_PROVIDERS_BASE_DIR)/terraform-providers-*)

ifneq (,$(INSTALLED_PROVIDERS))
INSTALLED_PROVIDERS_NAMES=$(subst $(INSTALLED_PROVIDERS_BASE_DIR)/terraform-provider-,,$(INSTALLED_PROVIDERS))
INSTALLED_PROVIDERS_VOLUMES=$(foreach INSTALLED_PROVIDER_NAME,$(INSTALLED_PROVIDERS_NAMES),--volume "$(INSTALLED_PROVIDERS_BASE_DIR)/terraform-provider-$(INSTALLED_PROVIDER_NAME)/website:/ext/providers/$(INSTALLED_PROVIDER_NAME)/website")
else
INSTALLED_PROVIDERS_NAMES=
INSTALLED_PROVIDERS_VOLUMES=
endif

website:
    @echo "==> Starting website in Docker..."
    @docker run \
        --interactive \
        -e "SHOW_EXTEND=${EXTEND}" \
        --rm \
        --tty \
        --publish "4567:4567" \
        --publish "35729:35729" \
        --volume "$(shell pwd)/ext:/ext" \
        $(INSTALLED_PROVIDERS_VOLUMES) \
        --volume "$(shell pwd)/content:/website" \
        hashicorp/middleman-hashicorp:${VERSION}
```

bflad:
I do like the idea of having a public Docker image that can accept a volume mount -- gets rid of the awkward dealings of a separate repository and worrying about Go development environment. 😄

radeksimko:
> Instead of a hardcoded docker run, why not just write this up as a Dockerfile to build instead?

So we can certainly do that, but it won't make the `docker run` command much shorter, certainly not in the part where we deal with volumes, because you still have to tell it where to mount the volumes from. Dockerfile would only specify the mount _destination_ inside the container.

https://docs.docker.com/engine/reference/builder/#notes-about-specifying-volumes

> The host directory is declared at container run-time: The host directory (the mountpoint) is, by its nature, host-dependent.

Maybe you're suggesting we also bake in the whole `hashicorp/terraform-website` repository? Which would allow us to reduce it by 3 lines/args, but it brings another tricky question - is that really how you want to be developing the website locally? What if you need to test changes in `hashicorp/terraform-website` **and** some provider docs changes? You'll then end up mounting both anyway.

> I had suggested offline that we instead enhance the existing terraform-website Makefile target, then each repository can just call it.

Yeah, that sounds like a good idea - I just misunderstood your original comment. Thanks for further explanation via Slack.

radeksimko:
Blocked by https://github.com/hashicorp/terraform-website/pull/240

PR #4131, opened 2018-04-09
provide working CloudWatch Alarm Example

relifeted:
provide working CloudWatch Alarm Example


bflad:
@relifeted I believe the existing example is there to show how to create a Route53 health check based on a CloudWatch metric alarm, rather than the other way around as you have. I think you should instead create your example under the `aws_cloudwatch_metric_alarm` resource page rather than here.

relifeted:
@bflad oh, i misunderstanding the existing example, i thought that is about how to setup an alarm to notify a health check is failing.

bflad:
Closing - but we'll gladly accept anything to make it more clear. Thanks!

PR #4134, opened 2018-04-09
resource/aws_codebuild_project: Properly handle setting cache type NO_CACHE

### LINKED ISSUE #4113: aws_codebuild_project: cache config does not stick when cache type set to "NO_CACHE"
### Terraform Version
0.11.6

I've been trying to build a `aws_codebuild_project` module where cache can be enabled/disabled dynamically.

Although not officially documented, `cache.type` can be set to `NO_CACHE`, which rather intuitively made me thing i could use this value to indicate cache should not be enabled, which it does.

The caveat is that this value does not stick and reapplying the module causes terraform to plan modifications on this value over and over.

### Affected Resource(s)
aws_codebuild_project

```hcl
resource "aws_codebuild_project" "codebuild_project" {
  # ...
  cache {
    type = "NO_CACHE"
    location = ""
  }
}
```

### Expected Behavior
Once applied, a new terraform apply should not trigger any modification

### Actual Behavior
```
  ~ update in-place

  ~ aws_codebuild_project.codebuild_project
      cache.0.type: "" => "NO_CACHE"
```

### Reference: 
I traced the discussion here
https://github.com/terraform-providers/terraform-provider-aws/pull/2860/files#r160149934 

and while i get the rationale that drove to this behavior, I do think the `NO_CACHE` value should be available to users in a way that would stick, so that modules could conditionally enable caching.

cc: @kaofelix, @bflad

bflad:
Removes the initial workaround from #2860 to keep backwards compatibility with `Computed: true` on the `cache` attribute.

Closes #4113 

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSCodeBuildProject'
TF_ACC=1 go test ./aws -v -run=TestAccAWSCodeBuildProject -timeout 120m
=== RUN   TestAccAWSCodeBuildProject_basic
--- PASS: TestAccAWSCodeBuildProject_basic (22.21s)
=== RUN   TestAccAWSCodeBuildProject_vpc
--- PASS: TestAccAWSCodeBuildProject_vpc (54.93s)
=== RUN   TestAccAWSCodeBuildProject_cache
--- PASS: TestAccAWSCodeBuildProject_cache (61.97s)
--- PASS: TestAccAWSCodeBuildProject_sourceAuth (21.89s)
=== RUN   TestAccAWSCodeBuildProject_default_build_timeout
--- PASS: TestAccAWSCodeBuildProject_default_build_timeout (29.48s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	190.521s
```

### DIFF
--- a/website/docs/r/codebuild_project.html.markdown
+++ b/website/docs/r/codebuild_project.html.markdown
@@ -151,8 +151,8 @@ The following arguments are supported:
 
 `cache` supports the following:
 
-* `type` - (Required) The type of storage that will be used for the AWS CodeBuild project cache. The only valid value is `S3`.
-* `location` - (Required) The location where the AWS CodeBuild project stores cached resources. This value must be a valid S3 bucket name/prefix.
+* `type` - (Optional) The type of storage that will be used for the AWS CodeBuild project cache. Valid values: `NO_CACHE` and `S3`. Defaults to `NO_CACHE`.
+* `location` - (Required when cache type is not `NO_CACHE`) The location where the AWS CodeBuild project stores cached resources. For type `S3` the value must be a valid S3 bucket name/prefix.
 
 `environment` supports the following:
 
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/codebuild"
+	"github.com/hashicorp/terraform/helper/customdiff"
 	"github.com/hashicorp/terraform/helper/hashcode"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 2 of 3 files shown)


website/docs/r/codebuild_project.html.markdown
```diff
@@ -151,8 +151,8 @@ The following arguments are supported:
 
 `cache` supports the following:
 
-* `type` - (Required) The type of storage that will be used for the AWS CodeBuild project cache. The only valid value is `S3`.
-* `location` - (Required) The location where the AWS CodeBuild project stores cached resources. This value must be a valid S3 bucket name/prefix.
+* `type` - (Optional) The type of storage that will be used for the AWS CodeBuild project cache. Valid values: `NO_CACHE` and `S3`. Defaults to `NO_CACHE`.
+* `location` - (Required when cache type is not `NO_CACHE`) The location where the AWS CodeBuild project stores cached resources. For type `S3` the value must be a valid S3 bucket name/prefix.
```
katbyte: Might be slightly clearer to state 'Required when cache type is `S3`'. On my first read through my eyes jumped over the not.

bflad: Great catch, updated to: Required when cache type is `S3` 👍

bflad:

PR #4139, opened 2018-04-10
aws_rds_cluster is not setting engine_version for aurora postgres

miromad:
aurora for postgres added support for postgres 9.6.6 version in addition to 9.6.3 (default).

Attempt to set version 9.6.6 revealed that engine_version parameter was not added to all code paths in original enhancement.


bflad:

Issue #2828, opened 2018-01-02
aws_api_gateway_stage Tag support


FireballDWF:
The aws_api_gateway_stage resource should be enhanced to support the new Tag support documented at https://docs.aws.amazon.com/apigateway/api-reference/resource/stage/ and announced at https://aws.amazon.com/about-aws/whats-new/2017/12/amazon-api-gateway-supports-tagging-api-stages/

bflad:
This functionality has been merged into master and will release with v1.16.0 of the AWS provider, likely mid next week.

PR #2858, opened 2018-01-04
r/api_gateway_stage: Support tags

atsushi-ishibashi:
#2828 
- [x] Update tags
- [x] test
- [x] docs

PR #4152, opened 2018-04-10
Feature/api stage tags

garylb2:
#2828 Implemented tags for both individual stages and deployment's convenience stage


bflad:
Hi @garylb2 👋 thanks for submitting this! It turns out there was an older PR for handling this #2858 which just got merged. Sorry for the hassle, but good news is the functionality is now merged in (at least for `aws_api_gateway_stage`) and will release in v1.16.0 of the AWS provider, likely mid next week.

If you're still interested in doing this for `aws_api_gateway_deployment`, I'd suggest creating a new PR based on master. 👍

Issue #4146, opened 2018-04-10
Feature Request: Data source to get API Gateway REST API by name


timborrowdale:
### Use case: 
We would like to be able to independently deploy resource integrations to a single API Gateway. e.g.
1. Main pipeline creates API Gateway
2. Resource 1 pipeline creates lambda function and sets up integration on API Gateway for resource `/1`
3. Resource 2 pipeline creates lambda function and sets up integration on API Gateway for resource `/2`

In order to do this the following terraform resources need knowledge of the existing API Gateway:
* [aws_api_gateway_resource](https://www.terraform.io/docs/providers/aws/r/api_gateway_resource.html)
* [aws_api_gateway_method](https://www.terraform.io/docs/providers/aws/r/api_gateway_method.html)
* [aws_api_gateway_integration](https://www.terraform.io/docs/providers/aws/r/api_gateway_integration.html)

### Suggested data source

```hcl
data "aws_api_gateway_rest_api" "by_name" {
  name = "my-gateway"
}
```

which would return `id` and `root_resource_id` as computed attributes

I can have a go at implementing this if it is considered useful, but I have a concern that currently it is possible to create two API Gateways with the same name. What would be the expected output of a query where there is more than one match?

catsby:
Seems possible, unfortunately there is no "filter" for the [GetRestAPIs](https://docs.aws.amazon.com/apigateway/api-reference/resource/rest-apis/) endpoint. You send the query and get a paginated result, we'd have to loop through and pattern match on the name 😦 

Still, a reasonable request. I don't believe anyone on the Terraform team will be able to get to this, but if someone is reading this and wants to give it a try, I would recommend copying the [AutoScaling Group Data Source](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/data_source_aws_availability_zone.go) and test file, and modifying to do as I mentioned:

- grab name from config
- call `GetRestApis`
- loop until we find a match

The SDK includes `GetRestApisPages `:

- https://github.com/terraform-providers/terraform-provider-aws/blob/6097dd5c25599fe0ca9c2064be48dcdbce5f3256/vendor/github.com/aws/aws-sdk-go/service/apigateway/api.go#L6760-L6779

bflad:
[...]

PR #4172, opened 2018-04-11
add new data source aws_api_gateway_rest_api

timborrowdale:
#4146
```
make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsApiGatewayRestApi'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsApiGatewayRestApi -timeout 120m
=== RUN   TestAccDataSourceAwsApiGatewayRestApi
--- PASS: TestAccDataSourceAwsApiGatewayRestApi (28.58s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       29.862s
```

I've had issues trying to get the terraform-website project working to check the docs locally - I think they're ok but would be good if someone could check.

### DIFF
--- a/aws/data_source_aws_api_gateway_rest_api.go
+++ b/aws/data_source_aws_api_gateway_rest_api.go
@@ -1,0 +1,88 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/apigateway"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsApiGatewayRestApi() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsApiGatewayRestApiRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"root_resource_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceAwsApiGatewayRestApiRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+	params := &apigateway.GetRestApisInput{}
+
+	target := d.Get("name")
+	var matchedApis []*apigateway.RestApi
+	log.Printf("[DEBUG] Reading API Gateway REST APIs: %s", params)
+	err := conn.GetRestApisPages(params, func(page *apigateway.GetRestApisOutput, lastPage bool) bool {
+		for _, api := range page.Items {
+			if *api.Name == target {
+				matchedApis = append(matchedApis, api)
+			}
+		}
+		return true
+	})
+	if err != nil {
+		return errwrap.Wrapf("error describing API Gateway REST APIs: {{err}}", err)
+	}
+
+	if len(matchedApis) == 0 {
+		return fmt.Errorf("no REST APIs with name %q found in this region", target)
+	}
+	if len(matchedApis) > 1 {
+		return fmt.Errorf("multiple REST APIs with name %q found in this region", target)
+	}
+
+	match := matchedApis[0]
+
+	d.SetId(*match.Id)
+
+	if err = dataSourceAwsApiGatewayRestApiRefreshResources(d, meta); err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


aws/data_source_aws_api_gateway_rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+
+func dataSourceAwsApiGatewayRestApi() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsApiGatewayRestApiRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"id": {
```
bflad: `id` does not need to be declared as a schema attribute 👍

aws/data_source_aws_api_gateway_rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+	target := d.Get("name")
+	var matchedApis []*apigateway.RestApi
+	log.Printf("[DEBUG] Reading API Gateway REST APIs: %s", params)
+	err := conn.GetRestApisPages(params, func(page *apigateway.GetRestApisOutput, lastPage bool) bool {
+		for _, api := range page.Items {
+			if *api.Name == target {
+				matchedApis = append(matchedApis, api)
+			}
+		}
+		return true
```
bflad: Should use `return !lastPage` here

aws/data_source_aws_api_gateway_rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+func dataSourceAwsApiGatewayRestApiRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).apigateway
+	params := &apigateway.GetRestApisInput{}
+
+	target := d.Get("name")
+	var matchedApis []*apigateway.RestApi
+	log.Printf("[DEBUG] Reading API Gateway REST APIs: %s", params)
+	err := conn.GetRestApisPages(params, func(page *apigateway.GetRestApisOutput, lastPage bool) bool {
+		for _, api := range page.Items {
+			if *api.Name == target {
```
bflad: To prevent a potential crash, we should wrap this with `aws.StringValue(api.Name)` in the unlikely scenario that its returned as `nil`

aws/data_source_aws_api_gateway_rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+				matchedApis = append(matchedApis, api)
+			}
+		}
+		return true
+	})
+	if err != nil {
+		return errwrap.Wrapf("error describing API Gateway REST APIs: {{err}}", err)
```
bflad: We do not need the context of the error when returning a message back to the user here. We can use the simpler `fmt.Errorf("error describing API Gateway REST APIs: %s", err)` here 👍

aws/data_source_aws_api_gateway_rest_api.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	if len(matchedApis) > 1 {
+		return fmt.Errorf("multiple REST APIs with name %q found in this region", target)
+	}
+
+	match := matchedApis[0]
+
+	d.SetId(*match.Id)
+
+	if err = dataSourceAwsApiGatewayRestApiRefreshResources(d, meta); err != nil {
```
bflad: We do not need the extra function here, its contents can be included in the read function 👍

aws/data_source_aws_api_gateway_rest_api_test.go
```diff
@@ -0,0 +1,80 @@
...
+				apiGatewayRestApiResources.Primary.Attributes["root_resource_id"],
+			)
+		}
+
+		return nil
+	}
+}
+
+const testAccDataSourceAwsApiGatewayRestApiConfig = `
+provider "aws" {
```
bflad: We do not need to declare the `provider` block in the testing framework, it defaults to `us-west-2` if not provided via AWS_DEFAULT_REGION

aws/data_source_aws_api_gateway_rest_api_test.go
```diff
@@ -0,0 +1,80 @@
...
+  region = "us-west-2"
+}
+
+resource "aws_api_gateway_rest_api" "tf_wrong1" {
+  name        = "wrong1"
+}
+
+resource "aws_api_gateway_rest_api" "tf_test" {
+  name        = "tf_test"
```
bflad: These should have randomized names to prevent hitting the multiple match error condition after test panics/AWS failures. Usually we achieve this by wrapping the test configuration in `fmt.Sprintf()` and passing a randomized name from the test function, e.g. `rName := acctest.StringWithPrefix("tf_acc_test_")`

timborrowdale:
@bflad Thanks for the feedback - I've had a go at addressing those issues.

bflad:

PR #4173, opened 2018-04-11
Add contraints link for aws_db_instance attribute

vkatsikaros:
The attribute db_subnet_group_name for resource aws_db_instance should not be
specified if read replicas are in the same region as the master. This change adds
a clarification in the docs.

Sample error from Terraform v0.11.5, provider.aws v1.11.0:
Error creating DB Instance: DBSubnetGroupNotAllowedFault: DbSubnetGroupName
should not be specified for read replicas that are created in the same region as the master

Moreover, the attribute has a number of constraints and this change adds a link to the AWS documentation.


website/docs/r/db_instance.html.markdown
```diff
@@ -86,7 +86,9 @@ for more information.
 Default is `false`.
 * `db_subnet_group_name` - (Optional) Name of DB subnet group. DB instance will
 be created in the VPC associated with the DB subnet group. If unspecified, will
-be created in the `default` VPC, or in EC2 Classic, if available.
+be created in the `default` VPC, or in EC2 Classic, if available. It needs to be
```
bflad: Can you please qualify this "only if" sentence with `For creating read replica instances, it needs to`... I think otherwise this can be confusing when you're not working with read replicas.

vkatsikaros: Thanks for spotting this. Fixed

Issue #3579, opened 2018-03-01
TF keeps recreating a '@' TXT record


FransUrbo:
I've been tasked with moving all domains that the company owns from register.com to AWS Route53.

In a few domains, there's a TXT record named `@`. Moving those into TF and then have TF apply it leads to TF wanting to *create* the record each time.

### Terraform Version
0.11.3

### Affected Resource(s)
- aws_route53_record

### Terraform Configuration Files
```hcl
resource "aws_route53_zone" "bnf_org_uk" {
  name                 = "bnf.org.uk"
  comment              = "bnf.org.uk"
  force_destroy        = "true"
}

resource "aws_route53_record" "bnf_org_uk-at_txt" {
  zone_id                     = "${aws_route53_zone.bnf_org_uk.zone_id}"
  name                        = "@.bnf.org.uk"
  type                        = "TXT"
  ttl                         = 3600
  records                     = [
    "Domain for:tisl"
  ]
}
```

### Expected Behavior
Created once and never any changes again.

### Actual Behavior
TF want's to *add* the resource every time.

### Steps to Reproduce
```
[Turbo-Fredrikssons-MacBook-Pro]$ terraform plan -target=aws_route53_record.bnf_org_uk-at_txt
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

aws_route53_zone.bnf_org_uk: Refreshing state... (ID: SOMETHING)

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  + aws_route53_record.bnf_org_uk-at_txt
      id:                 <computed>
      allow_overwrite:    "true"
      fqdn:               <computed>
      name:               "@.bnf.org.uk"
      records.#:          "1"
      records.4097833795: "Domain for:tisl"
      ttl:                "3600"
      type:               "TXT"
      zone_id:            "SOMETHING"


Plan: 1 to add, 0 to change, 0 to destroy.

------------------------------------------------------------------------

Note: You didn't specify an "-out" parameter to save this plan, so Terraform
can't guarantee that exactly these actions will be performed if
"terraform apply" is subsequently run.

[Turbo-Fredrikssons-MacBook-Pro]$ terraform apply -target=aws_route53_record.bnf_org_uk-at_txt
[...]

PR #4183, opened 2018-04-12
Make sure route53 records really exist after creation and add support for all ASCII characters

Crazybus:
In #3579 we found some weird behaviour where records starting with `@` were always showing up as a change for every `terrafrom apply`. The weird part about this was that the record was being created correctly and not showing up in state. In https://github.com/terraform-providers/terraform-provider-aws/issues/3579#issuecomment-369590392 @bflad pointed out that `@` doesn't need to be used with AWS and they actually give you a warning when you try to use it inside the AWS console. With this answer I was happy and able to move these `@` records to the apex domain and move on with my life. However I was still very curious why the records weren't showing up in state properly. 

Fast forward a few days and I submitted a job to move the values of the `@` records to the apex domains for each zone. It worked! However a little while later someone reported an unrelated DNS record was no longer resolving. I had only just recently moved all of our production DNS into terraform so I was worried that the automation for deploying changes had accidentally deleted something. In https://github.com/terraform-providers/terraform-provider-aws/issues/3579#issuecomment-380064285 you can see more detail about what actually happened. 

Luckily all DNS changes are now automated by a jenkins job running terraform. Looking at the Route53 logs I could see that the unrelated record was removed exactly when terraform was removing the `@` record. Another terraform run fixed things up and added the record back but I felt that I needed to figure out what happened. Luckily I was able to reproduce this locally however I still hadn't explained why it only happened for 1 of the 3 zones which had `@` records.

# Findings

## How did a random record get deleted?

I figured out very quickly that this was only happening for zones which had more than 100 records (the default maxItems per page for the AWS api). I noticed this when I tried to reproduce it with a limited dataset and couldn't trigger the issue. So my focus shifted to the code which was looping of the paginated results from AWS.

The root cause was the line https://github.com/terraform-providers/terraform-provider-aws/blob/c77020f539eed9d3f1dab36c2d2d5a624ed3007b/aws/resource_aws_route53_record.go#L646

So the current record in the response was being compared to the `*lopts.StartRecordName`. Which for the first page of results was the correct record. However if the first result didn't match for some reason (In my case because `@.domain.com` != `\\100.domain.com` it would continue looping until it got to the next page. At this stage `*lopts.StartRecordName` now was set to the first result in the second page of results and then the record matched perfectly!

So this bug would only be triggered if the record being deleted didn't match with how AWS returned the record, and if there were more than 100 records in the zone. 

P.S. I tried to write an acceptance test for this behaviour but with my other fixes I now can't trigger a scenario where terraform can create a record it can't find. 

## Why didn't terraform give an error that it couldn't find the record after creating it?

The above wouldn't have been possible if terraform had correctly detected that it couldn't read the record it had created. When creating and updating records the final line is https://github.com/terraform-providers/terraform-provider-aws/blob/c77020f539eed9d3f1dab36c2d2d5a624ed3007b/aws/resource_aws_route53_record.go#L454

Since this function is also used for checking if a record exists or not it silently will drop the record and remove it from state if the record doesn't exist.

https://github.com/terraform-providers/terraform-provider-aws/blob/c77020f539eed9d3f1dab36c2d2d5a624ed3007b/aws/resource_aws_route53_record.go#L520-L530

So what was happening is terraform was trying to create `@.domain.com`, after successfully creating it then called `resourceAwsRoute53RecordRead` which didn't return an error so Terraform didn't give an error. 

# Other improvements

## Maxitems

My first reaction when seeing the looping over the first 100 results was....why? I then discovered the reason and added it as an inline comment so nobody else gets confused in the future. I have now optimised this to only return a single result if we know that there can be only one match. This saves terraform needing to pull in 100 resources everytime it wants to check if a single record exists. 

## Escaped ASCII characters

AWS supports a lot more special characters apart from `*` and `@` which I have now added support for. You can see a full list at the [Route53 Domain Name Format](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html) docs. I decided to just convert all of them instead of trying to maintain a list of supported characters. If a user does somehow pick a character that doesn't convert correctly they will now correctly receive an error when terraform tries to read back the record after creating it.

### DIFF
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -20,8 +20,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/route53"
 )
 
-var r53NoRecordsFound = errors.New("No matching Hosted Zone found")
-var r53NoHostedZoneFound = errors.New("No matching records found")
+var r53NoRecordsFound = errors.New("No matching records found")
+var r53NoHostedZoneFound = errors.New("No matching Hosted Zone found")
 var r53ValidRecordTypes = regexp.MustCompile("^(A|AAAA|CAA|CNAME|MX|NAPTR|NS|PTR|SOA|SPF|SRV|TXT)$")
 
 func resourceAwsRoute53Record() *schema.Resource {
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -374,7 +374,8 @@ func resourceAwsRoute53RecordUpdate(d *schema.ResourceData, meta interface{}) er
 		return err
 	}
 
-	return resourceAwsRoute53RecordRead(d, meta)
+	_, err = findRecord(d, meta)
+	return err
 }
 
 func resourceAwsRoute53RecordCreate(d *schema.ResourceData, meta interface{}) error {
@@ -451,7 +452,8 @@ func resourceAwsRoute53RecordCreate(d *schema.ResourceData, meta interface{}) er
 		return err
 	}
 
-	return resourceAwsRoute53RecordRead(d, meta)
+	_, err = findRecord(d, meta)
+	return err
 }
 
 func changeRoute53RecordSet(conn *route53.Route53, input *route53.ChangeResourceRecordSetsInput) (interface{}, error) {
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -632,27 +632,49 @@ func findRecord(d *schema.ResourceData, meta interface{}) (*route53.ResourceReco
 	log.Printf("[DEBUG] Expanded record name: %s", en)
 	d.Set("fqdn", en)
 
+	recordName := FQDN(strings.ToLower(en))
+	recordType := d.Get("type").(string)
+	recordSetIdentifier := d.Get("set_identifier")
+
+	// If this isn't a Weighted, Latency, Geo, or Failover resource with
+	// a SetIdentifier we only need to look at the first record in the response since there can be
+	// only one
+	maxItems := "1"
+	if recordSetIdentifier != "" {
+		maxItems = "100"
+	}
+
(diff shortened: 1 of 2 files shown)


aws/resource_aws_route53_record.go
```diff
@@ -630,27 +633,49 @@ func findRecord(d *schema.ResourceData, meta interface{}) (*route53.ResourceReco
...
+		StartRecordType: aws.String(recordType),
+		MaxItems:        aws.String(maxItems),
 	}
 
 	log.Printf("[DEBUG] List resource records sets for zone: %s, opts: %s",
 		zone, lopts)
 
 	var record *route53.ResourceRecordSet
+
+	// We need to loop over all records starting from the record we are looking for because
```
bflad: ❤️  thank you!

aws/resource_aws_route53_record.go
```diff
@@ -630,27 +633,49 @@ func findRecord(d *schema.ResourceData, meta interface{}) (*route53.ResourceReco
...
+	// called SetIdentifier which allows multiple entries with the same name and type but
+	// a different SetIdentifier
+	// For all other records we are setting the maxItems to 1 so that we don't return extra
+	// unneeded records
 	err = conn.ListResourceRecordSetsPages(lopts, func(resp *route53.ListResourceRecordSetsOutput, lastPage bool) bool {
 		for _, recordSet := range resp.ResourceRecordSets {
-			name := cleanRecordName(*recordSet.Name)
-			if FQDN(strings.ToLower(name)) != FQDN(strings.ToLower(*lopts.StartRecordName)) {
+
+			responseName := strings.ToLower(cleanRecordName(*recordSet.Name))
```
bflad: Minor nitpick: To prevent crashes in the very unlikely scenario that AWS returns nil for either of these, it would be safer to wrap them in `aws.StringValue()` 👍  e.g. `aws.StringValue(recordSet.Name)`

bflad:

PR #4198, opened 2018-04-12
[WIP] Adds a flag that will keep old task definitions active

binarymatt:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/258

I'm not crazy on the name and would love some suggestions.

### DIFF
--- a/aws/resource_aws_ecs_task_definition.go
+++ b/aws/resource_aws_ecs_task_definition.go
@@ -135,6 +135,12 @@ func resourceAwsEcsTaskDefinition() *schema.Resource {
 				ForceNew: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+
+			"deregister_old": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


bflad:
Hi @binarydud! 👋  While I sympathize with the use case, I think the implementation might belong more in Terraform core rather than an individual implementation in each resource.

For example, we have a [resource meta-parameter called `prevent_destroy`](https://www.terraform.io/docs/configuration/resources.html#lifecycle) that serves a similar, but different purpose and is available for all resources. Speaking to that parameter directly, its probably worth mentioning https://github.com/hashicorp/terraform/issues/3874#issuecomment-161715361, which maybe something like you're alluding to here might fit the same bill.

Does that upstream issue look like what you're trying to accomplish? I know its not currently implemented, but given that this problem seems like it should be globally solved instead of piecemeal I think we should close this and hope that type of lifecycle management can happen upstream.

binarydud:
@bflad i'm not entirely sure that the prevent_destroy fix would get at what i would like to happen. The issue for my use case is that I want a new task definition to be created, say when i update with a new docker container; however, i don't want the previous one to be deactivated. Another way of saying this is that I want terraform to update my task definition only, no deactivation of previous revisions.

binarydud:
@bflad i realized one of the reasons that the current behavior seems odd to me and the reason i added this flag. The `aws_lambda_function` has the exact behavior i'd expect of the task definition resource. When a new version of a function is uploaded, the aws provider doesn't delete the older versions, it leaves them there. ECS task definitions should have the same behavior imo; however, i realize that would be a breaking change so a flag made sense to me.

binarydud:
@bflad any thoughts on my comments above?

ewbankkit:
Closing as stale due to inactivity.

PR #4209, opened 2018-04-14
resource/aws_iam_role: Suppress NoSuchEntity errors while detaching policies from role

spirius:
`aws_iam_role` with `force_detach_policies = true` sometimes gets outdated list of attached policies from AWS. This happens when terraform is removing `aws_iam_role` right after removing `aws_iam_role_policy_attachment` resources.


bflad:

PR #4212, opened 2018-04-15
New Data Source: cognito_user_pools

loivis:
Complete missing part between cognito user pools and api gateway authorizer.

There is no option to filter cognito user pools, so a simple loop is used to match given name.

```
⎇  make fmt; and echo > aws/debug.log ; and make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsCognitoUserPoolIds_basic'
gofmt -w $(find . -name '*.go' |grep -v vendor)
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsCognitoUserPoolIds_basic -timeout 120m
=== RUN   TestAccDataSourceAwsCognitoUserPoolIds_basic
--- PASS: TestAccDataSourceAwsCognitoUserPoolIds_basic (18.81s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	18.858s
```

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -173,6 +173,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_cloudformation_stack":             dataSourceAwsCloudFormationStack(),
 			"aws_cloudtrail_service_account":       dataSourceAwsCloudTrailServiceAccount(),
 			"aws_cloudwatch_log_group":             dataSourceAwsCloudwatchLogGroup(),
+			"aws_cognito_user_pools":               dataSourceAwsCognitoUserPools(),
 			"aws_db_instance":                      dataSourceAwsDbInstance(),
 			"aws_db_snapshot":                      dataSourceAwsDbSnapshot(),
 			"aws_dynamodb_table":                   dataSourceAwsDynamoDbTable(),
--- a/website/aws.erb
+++ b/website/aws.erb
@@ -73,6 +73,9 @@
                         <li<%= sidebar_current("docs-aws-datasource-cloudwatch-log-group") %>>
                             <a href="/docs/providers/aws/d/cloudwatch_log_group.html">aws_cloudwatch_log_group</a>
                         </li>
+                        <li<%= sidebar_current("docs-aws-datasource-cognito-user-pools") %>>
+                            <a href="/docs/providers/aws/d/cognito_user_pools.html">aws_cognito_user_pools</a>
+                        </li>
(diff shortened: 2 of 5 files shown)


aws/provider.go
```diff
@@ -173,6 +173,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_cloudformation_stack":             dataSourceAwsCloudFormationStack(),
 			"aws_cloudtrail_service_account":       dataSourceAwsCloudTrailServiceAccount(),
 			"aws_cloudwatch_log_group":             dataSourceAwsCloudwatchLogGroup(),
+			"aws_cognito_user_pool_ids":            dataSourceAwsCognitoUserPoolIds(),
```
bflad: Let's opt to call it `aws_cognito_user_pools` (we may add attributes other than just the ID) - will require refactoring the function names and documentation

website/aws.erb
```diff
@@ -72,6 +72,8 @@
                         </li>
                         <li<%= sidebar_current("docs-aws-datasource-cloudwatch-log-group") %>>
                             <a href="/docs/providers/aws/d/cloudwatch_log_group.html">aws_cloudwatch_log_group</a>
+                        <li<%= sidebar_current("docs-aws-datasource-cognito-user-pool-ids") %>>
+                            <a href="/docs/providers/aws/d/cognito_user_pool_ids.html">aws_cognito_user_pool_ids</a>
```
bflad: Missing `</li>`

aws/data_source_aws_cognito_user_pool_ids.go
```diff
@@ -0,0 +1,78 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"ids": {
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
```
bflad: The `Set: schema.HashString` is unnecessary here

loivis:
Opt to `aws_cognito_user_pools`.

Just noticed there are dangling resources after acceptance test. Tried to add sweeper but it didn't seem to work. Did I miss something?

bflad:
The test sweeper should be part of the resource and not the data source. Let's address it separately.

loivis:
All right. I've removed related code now.

bflad:

PR #4215, opened 2018-04-16
Set engine version when calling RestoreDBClusterFromSnapshot

steveh:
Hopefully fixes https://github.com/terraform-providers/terraform-provider-aws/issues/4214


bflad:

PR #4222, opened 2018-04-16
New Data Source: aws_lambda_invocation

spirius:
Data source for using lambda function as data provider, similar to [External Data Source](https://www.terraform.io/docs/providers/external/data_source.html).


spirius:
Test results

```
$ TF_ACC=1 go test -v -run=TestAccDataSourceAwsLambdaInvoke_* -timeout 120m
=== RUN   TestAccDataSourceAwsLambdaInvoke_basic
--- PASS: TestAccDataSourceAwsLambdaInvoke_basic (51.64s)
=== RUN   TestAccDataSourceAwsLambdaInvoke_qualifier
--- PASS: TestAccDataSourceAwsLambdaInvoke_qualifier (50.15s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws   101.834s
```

aws/data_source_aws_lambda_invoke.go
```diff
@@ -0,0 +1,91 @@
...
+			"input": {
+				Type:     schema.TypeMap,
+				Required: true,
+				ForceNew: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+
+			"result": {
```
bflad: Same here -- we should convert this to `TypeString` and remove the JSON unmarshal handling/errors from the data source.

aws/data_source_aws_lambda_invoke.go
```diff
@@ -0,0 +1,91 @@
...
+
+	payload, err := json.Marshal(d.Get("input"))
+
+	if err != nil {
+		return err
+	}
+
+	res, err := conn.Invoke(&lambda.InvokeInput{
+		FunctionName:   aws.String(functionName),
+		InvocationType: aws.String(lambda.InvocationTypeRequestResponse),
```
bflad: In the future, we can allow a new `invocation_type` attribute for calling Lambda functions asynchronously with `Event` 👍

aws/provider.go
```diff
@@ -211,6 +211,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_kms_ciphertext":                   dataSourceAwsKmsCiphertext(),
 			"aws_kms_key":                          dataSourceAwsKmsKey(),
 			"aws_kms_secret":                       dataSourceAwsKmsSecret(),
+			"aws_lambda_invoke":                    dataSourceAwsLambdaInvoke(),
```
bflad: Can you please rename the data source `aws_lambda_function_invocation`? Thanks!

aws/data_source_aws_lambda_invoke.go
spirius: If we use `TypeString` for `result`, how the returned data will be used later in terraform code? As far as I know there is no `jsondecode` or similar function available.

aws/data_source_aws_lambda_invoke.go
spirius: Exactly, I wanted to keep the first version simple, but I think we can pass some *random* s3 key  to lambda during invocation and expect to get result stored there.

PauloMigAlmeida:
@spirius Apparently, there are conflicts with the master branch that need to be solved before anyone from Hashicorp can take further action on this. Would you mind addressing them?

@bflad Apart from these conflicts, what else needs to be done in order to get this approved/merged into master? I'm happy to contribute to it in any way possible.

Cheers!

spirius:
@PauloMigAlmeida conflict fixed.

@bflad I've updated code according to your comments. The only question I have is regarding `result` field. If we use `TypeString` for it, how the result can be used later in terraform since it needs to be unmarshalled somehow and there is no builtin function in terraform for doing that.

bflad:
For handling the response, what do you think about the below? 😄 

* Renaming the existing `result` attribute to `result_map`
* Creating a new `result` attribute using `TypeString` with the "full" JSON string
* Ensuring the logic is setup so it does not completely fail when the JSON cannot be unmarshalled into the `map[string]interface{}` (and document the behavior that the attribute will show as missing downstream if the JSON cannot be unmarshalled)

This enables these use cases:
* Advanced operators that are comfortable with using `null_resource` or one of the community "JSON decoding" providers, can utilize the `TypeString` attribute today
* When `jsondecode` is implemented as an interpolation function upstream in Terraform core, people can immediately use the `TypeString` attribute without AWS provider code changes. The `jsondecode` function is being tracked upstream at: https://github.com/hashicorp/terraform/issues/10363
* Operators can still use the limited `TypeMap` implementation today, if their Lambda response fits the mold

spirius:
Hi @bflad, renamed `result` to `result_map` and added raw result to data source.

Test results:
```
$ TF_ACC=1 go test -v -run=TestAccDataSourceAwsLambdaInvocation_* -timeout 120m ./aws
=== RUN   TestAccDataSourceAwsLambdaInvocation_basic
--- PASS: TestAccDataSourceAwsLambdaInvocation_basic (51.32s)
=== RUN   TestAccDataSourceAwsLambdaInvocation_qualifier
--- PASS: TestAccDataSourceAwsLambdaInvocation_qualifier (53.58s)
=== RUN   TestAccDataSourceAwsLambdaInvocation_complex
--- PASS: TestAccDataSourceAwsLambdaInvocation_complex (50.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	155.076s
```

bflad:

PR #4241, opened 2018-04-17
S3 bucket: Allow enabling expired_object_delete_marker only

errriclee:
https://docs.aws.amazon.com/AmazonS3/latest/dev/lifecycle-configuration-examples.html#lifecycle-config-conceptual-ex7
Under "Current object versions don't have a well-defined lifecycle."

Allows enabling expired_object_delete_marker without setting days, so no current version expiration.

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -246,7 +246,7 @@ func resourceAwsS3Bucket() *schema.Resource {
 									"days": {
 										Type:         schema.TypeInt,
 										Optional:     true,
-										ValidateFunc: validation.IntAtLeast(1),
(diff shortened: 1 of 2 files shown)


errriclee:
Resulting XML for lifecycle:
```
<?xml version="1.0" encoding="UTF-8"?>
<LifecycleConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <Rule>
      <Expiration>
         <ExpiredObjectDeleteMarker>true</ExpiredObjectDeleteMarker>
      </Expiration>
      <Filter>
         <Prefix>path1/</Prefix>
      </Filter>
      <ID>id1</ID>
      <Status>Enabled</Status>
   </Rule>
</LifecycleConfiguration>
```

PR #4248, opened 2018-04-18
Remove principal type AWS from IAM policy wildcard normalization

spirius:
The normalized IAM policies for wildcard principals are not accepted by IAM when used as trust policy for roles. According to [AWS IAM doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html) it should be fine, but in practice IAM returns `MalformedPolicyDocument: AssumeRolepolicy contained an invalid principal: "STAR":"*"` error.

Basically following terraform code doesn't work, because it normalizes principal to `"Principal": "*"`, and it is not possible to get `{"Principal": {"AWS": *}}` in rendered policy when using `aws_iam_policy_document` data source
```hcl
data "aws_iam_policy_document" "policy" {
  statement {
    effect = "Deny"
    actions = ["sts:AssumeRole"]
    principals {
      type = "AWS"
      identifiers = ["*"]
    }
  }
}

resource "aws_iam_role" "role" {
  name = "test_role"
  assume_role_policy = "${data.aws_iam_policy_document.policy.json}"
}
```

Without normalization it works as expected:
```hcl
resource "aws_iam_role" "role" {
  name = "test_role"
  assume_role_policy = <<EOS
{
  "Statement": [
    {
      "Effect": "Deny",
      "Action": "sts:AssumeRole",
      "Principal": {
        "AWS": "*"
      }
    }
  ]
}
EOS
}
```

The PR removes principal type "AWS" from normalization.


bflad:
I _think_ this is a good idea, but we should add some additional user-facing documentation surrounding the various principal behaviors on the [`aws_iam_policy_document` data source page](https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html). Can you include some additional verbiage in the principal(/type) section and/or bottom examples section please? We should call out the support for `type = "*"` explicitly at least.

*MAINTAINER NOTE: This will require a NOTE callout in the CHANGELOG, preferably linking to the above documentation enhancements*

spirius:
@bflad, added new section in docs about wildcard principals

bflad:

PR #4249, opened 2018-04-18
New Resource: aws_organizations_policy

bflad:
Reference: #571 

```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsOrganizationsPolicy'
TF_ACC=1 go test ./aws -v -run=TestAccAwsOrganizationsPolicy -timeout 120m
=== RUN   TestAccAwsOrganizationsPolicy_basic
--- PASS: TestAccAwsOrganizationsPolicy_basic (20.05s)
=== RUN   TestAccAwsOrganizationsPolicy_description
--- PASS: TestAccAwsOrganizationsPolicy_description (19.12s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	39.211s
```

### DIFF
--- a/website/docs/r/organizations_policy.html.markdown
+++ b/website/docs/r/organizations_policy.html.markdown
@@ -1,0 +1,49 @@
+---
+layout: "aws"
+page_title: "AWS: aws_organizations_policy"
+sidebar_current: "docs-aws-resource-organizations-policy"
+description: |-
+  Provides a resource to manage an AWS Organizations policy.
+---
+
+# aws_organizations_policy
+
+Provides a resource to manage an [AWS Organizations policy](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_policies.html).
+
+## Example Usage
+
+```hcl
+data "aws_iam_policy_document" "example" {
+  statement {
+    actions   = ["*"]
+    resources = ["*"]
+  }
+}
+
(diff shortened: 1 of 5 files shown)


website/docs/r/organizations_policy.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+```hcl
+data "aws_iam_policy_document" "example" {
+  statement {
+    actions   = ["*"]
+    resources = ["*"]
+  }
+}
+
+resource "aws_organizations_policy" "example" {
+  content = "${data.aws_iam_policy_document.example.json}"
```
afeld: I wonder if it will be confusing to be including an IAM policy as a Service Control Policy, even though they are compatible...?

bflad:
> Would it be worthwhile to try and share code between this and the aws_iam_policy resource?

What do you propose? Usually the services are different enough it does not make sense to share code across them.

website/docs/r/organizations_policy.html.markdown
bflad: I think its better to show the "easier" way to manage policies rather than error-prone heredoc syntax personally.

afeld:
I was thinking a subset of the `Schema`, but maybe not worth it.

bflad:

PR #4256, opened 2018-04-18
data-source/cognito_user_pools: add attribute of arns

loivis:
This should have been in the first implement with #4212. I missed to check even though an example was provided to serve as api gateway authorizer. As it's not returned directly from aws api, arn is populated from id.


bflad:

Issue #4067, opened 2018-04-05
🚀 Feature Request: Config Aggregator Support


gazoakley:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AWS Config now supports aggregating resources across accounts/regions: https://aws.amazon.com/blogs/aws/aws-config-update-aggregate-compliance-data-across-accounts-regions/

### New or Affected Resource(s)

- aws_config_aggregator
- aws_config_authorization

### Potential Terraform Configuration

```hcl
# Example implementation, may change when developed

resource "aws_config_aggregator" "example" {
  name = "example" # Required

  account_aggregation_source {
    account_ids     = ["123456789012"] # Required
    all_aws_regions = true             # Optional
    aws_regions     = ["eu-west-2"]    # Optional
  }

  organization_aggregation_source {
    all_aws_regions = true          # Optional
    aws_regions     = ["eu-west-2"] # Optional
    role_arn        = ""            # Required
  }
}

resource "aws_config_authorization" "example" {
  authorized_account_id = "123456789012" # Required
  authorized_aws_region = "eu-west-2"    # Required
}

```

### References

- https://docs.aws.amazon.com/config/latest/APIReference/API_PutConfigurationAggregator.html
- https://docs.aws.amazon.com/config/latest/APIReference/API_DeleteConfigurationAggregator.html
- https://docs.aws.amazon.com/config/latest/APIReference/API_PutAggregationAuthorization.html
- https://docs.aws.amazon.com/config/latest/APIReference/API_DeleteAggregationAuthorization.html

marcotesch:
> ```
> Terraform Configuration Files
> 
> # Example implementation, may change when developed
> 
> resource "aws_config_aggregator" "example" {
>   name = "example" # Required
> 
>   account_aggregation_source {
>     account_ids     = ["123456789012"] # Required
>     all_aws_regions = true             # Optional
>     aws_regions     = ["eu-west-2"]    # Optional
>   }
> 
>   organization_aggregation_source {
>     all_aws_regions = true          # Optional
>     aws_regions     = ["eu-west-2"] # Optional
[...]

PR #4263, opened 2018-04-19
New Resource: r/aws_config_authorization

gazoakley:
Partially implements #4067 

New Resources:

* `aws_config_authorization`

```
$ AWS_PROFILE=personal make testacc TESTARGS='-run=TestAccConfigAuthorization'
TF_ACC=1 go test ./... -v -run=TestAccConfigAuthorization -timeout 120m
=== RUN   TestAccConfigAuthorization_import
--- PASS: TestAccConfigAuthorization_import (24.76s)
=== RUN   TestAccConfigAuthorization_basic
--- PASS: TestAccConfigAuthorization_basic (21.88s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	46.688s
```

### DIFF
--- a/aws/import_aws_config_authorization_test.go
+++ b/aws/import_aws_config_authorization_test.go
@@ -1,0 +1,30 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccConfigAuthorization_import(t *testing.T) {
+	resourceName := "aws_config_authorization.example"
+	rString := acctest.RandStringFromCharSet(12, "0123456789")
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckConfigAuthorizationDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccConfigAuthorizationConfig_basic(rString),
+			},
+
+			resource.TestStep{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -303,6 +303,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_cloudwatch_log_resource_policy":           resourceAwsCloudWatchLogResourcePolicy(),
 			"aws_cloudwatch_log_stream":                    resourceAwsCloudWatchLogStream(),
 			"aws_cloudwatch_log_subscription_filter":       resourceAwsCloudwatchLogSubscriptionFilter(),
+			"aws_config_authorization":                     resourceAwsConfigAuthorization(),
 			"aws_config_config_rule":                       resourceAwsConfigConfigRule(),
 			"aws_config_configuration_recorder":            resourceAwsConfigConfigurationRecorder(),
 			"aws_config_configuration_recorder_status":     resourceAwsConfigConfigurationRecorderStatus(),
--- a/aws/resource_aws_config_authorization.go
+++ b/aws/resource_aws_config_authorization.go
@@ -1,0 +1,123 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/configservice"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsConfigAuthorization() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsConfigAuthorizationPut,
+		Read:   resourceAwsConfigAuthorizationRead,
+		Delete: resourceAwsConfigAuthorizationDelete,
+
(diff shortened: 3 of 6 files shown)


aws/resource_aws_config_authorization.go
```diff
@@ -0,0 +1,123 @@
...
+		AuthorizedAccountId: aws.String(accountId),
+		AuthorizedAwsRegion: aws.String(region),
+	}
+
+	_, err = conn.DeleteAggregationAuthorization(req)
+	if err != nil {
+		return fmt.Errorf("Error deleting authorization: %s", err)
+	}
+
+	d.SetId("")
```
bflad: Minor nitpick: `d.SetId("")` is extraneous in delete functions

aws/provider.go
```diff
@@ -303,6 +303,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_cloudwatch_log_resource_policy":           resourceAwsCloudWatchLogResourcePolicy(),
 			"aws_cloudwatch_log_stream":                    resourceAwsCloudWatchLogStream(),
 			"aws_cloudwatch_log_subscription_filter":       resourceAwsCloudwatchLogSubscriptionFilter(),
+			"aws_config_authorization":                     resourceAwsConfigAuthorization(),
```
bflad: Sorry this is going to be very pedantic but the Config API refers to these as "aggregate" authorization and the resource/function naming should reflect that for clarity and in case they introduce other forms/types of authorization.

```go
"aws_config_aggregate_authorization": resourceAwsConfigAggregateAuthoriation(),
```

aws/resource_aws_config_authorization.go
```diff
@@ -0,0 +1,123 @@
...
+
+	accountId, region, err := resourceAwsConfigAuthorizationParseID(d.Id())
+	if err != nil {
+		return err
+	}
+
+	d.Set("account_id", accountId)
+	d.Set("region", region)
+
+	res, err := conn.DescribeAggregationAuthorizations(&configservice.DescribeAggregationAuthorizationsInput{})
```
bflad: This call does not currently handle `NextToken` so it can incorrectly report errors/missing resources in large or separate API responses.

aws/resource_aws_config_authorization_test.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func testSweepConfigAuthorizations(region string) error {
+	client, err := sharedClientForRegion(region)
+	if err != nil {
+		return fmt.Errorf("Error getting client: %s", err)
+	}
+	conn := client.(*AWSClient).configconn
+
+	resp, err := conn.DescribeAggregationAuthorizations(&configservice.DescribeAggregationAuthorizationsInput{})
```
bflad: This call does not currently handle `NextToken` so it can potentially miss authorizations in large or separate API responses.

aws/resource_aws_config_authorization_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_config_authorization" {
+			continue
+		}
+
+		resp, err := conn.DescribeAggregationAuthorizations(&configservice.DescribeAggregationAuthorizationsInput{})
+
+		if err == nil {
+			if len(resp.AggregationAuthorizations) != 0 &&
```
bflad: There may be additional other authorizations present, especially with testing concurrency or in non-testing accounts. This needs to be updated to specifically handle `NextToken` and loop through all authorizations, only returning an error if the account ID _and_ region is found.

website/docs/r/config_authorization.markdown
```diff
@@ -0,0 +1,41 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `account_id` - (Required) Account ID
+* `region` - (Required) Region
+
+## Attributes Reference
+
+The following attributes are exported:
```
bflad: We're in the process of updating all these for clarity (#4685): `In addition to all arguments above, the following attributes are exported:`

aws/import_aws_config_authorization_test.go
```diff
@@ -0,0 +1,30 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func TestAccConfigAuthorization_import(t *testing.T) {
```
bflad: Minor nitpick: we're going to move away from separate import test files and tests (#4705). We can simply add the final TestStep the existing `_basic` test 👍

gazoakley:
@bflad Could we have at least stuck with `aws_config_aggregator_authorization`? Currently after https://github.com/terraform-providers/terraform-provider-aws/pull/4262 there'd be:

- `aws_config_aggregate_authorization`
- `aws_config_aggregator`

That doesn't gel well together - it suggests they might be related but not for certain (the console always refers to aggregator(s))

bflad:
@gazoakley sorry for any initial confusion -- we wound up renaming both of the resources to match the API more closely. If the link between the two is not obvious we should probably cross-link them accordingly in the resource documentation for each of them.

This has been released in [version 1.22.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1220-june-05-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #4284, opened 2018-04-20
(Fix) Mark auth.resource as `Sensitive`

Stretch96:
* https://github.com/terraform-providers/terraform-provider-aws/issues/4261


bflad:

PR #4288, opened 2018-04-20
Add new data source aws_batch_job_queue

gheine:


aws/data_source_aws_batch_job_queue.go
```diff
@@ -0,0 +1,108 @@
...
+	d.Set("name", jobQueue.JobQueueName)
+	d.Set("status", jobQueue.Status)
+	d.Set("status_reason", jobQueue.StatusReason)
+	d.Set("state", jobQueue.State)
+	d.Set("priority", jobQueue.Priority)
+
+	ceos := make([]map[string]interface{}, 0)
+	for _, v := range jobQueue.ComputeEnvironmentOrder {
+		ceo := map[string]interface{}{}
+		ceo["compute_environment"] = *v.ComputeEnvironment
```
bflad: To prevent provider panics, we should opt to safely read these:
```go
ceo["compute_environment"] = aws.StringValue(v.ComputeEnvironment),
ceo["order"] = int(aws.Int64Value(v.Order)),
```

aws/data_source_aws_batch_job_queue.go
```diff
@@ -0,0 +1,108 @@
...
+		ceo["order"] = *v.Order
+		ceos = append(ceos, ceo)
+	}
+	d.Set("compute_environment_order", ceos)
```
bflad: When calling `d.Set()` with non-scalar values, we should implement the error checking to ensure we're :
```go
if err := d.Set("compute_environment_order", ceos); err != nil {
  return fmt.Errorf("error setting compute_environment_order: %s", err)
}
```

I can tell you from looking at the the existing `aws_batch_job_queue` resource that it does not properly set `compute_environments`  into the Terraform state (and therefore cannot detect drift for that attribute) 😓

bflad:

PR #4289, opened 2018-04-20
ForceNew OpsWorks Stack on ChefVersion Change

hnryjms:
Hi there 👋 

I've only worked in Golang a few times in my life, and I only recently started using Terraform at my workplace for managing our servers...but I realized that my OpsWorks stack can't be changed to the newer version right now.

From AWS Management Console, it looks like a new resource is required to make changes to the Chef version.

tfopsworksinvalid
opsworkscheflocked

### DIFF
--- a/aws/resource_aws_opsworks_stack.go
+++ b/aws/resource_aws_opsworks_stack.go
@@ -77,6 +77,7 @@ func resourceAwsOpsworksStack() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Default:  "11.10",
+				ForceNew: true
 			},
 
 			"manage_berkshelf": {


hnryjms:
Hey, I'm new, and I'm not quite sure what the next steps are for this PR. Happy to help with what's needed!

bflad:
Hi @hnryjms 👋 Thank you for submitting this and sorry for the long delay in review.

We should not add `ForceNew: true` to this attribute, since the [UpdateStack API](https://docs.aws.amazon.com/opsworks/latest/APIReference/API_StackConfigurationManager.html#opsworks-Type-StackConfigurationManager-Version) and Terraform resource's update implementation supports passing this value to the UpdateStack API. Otherwise, this would prevent legitimate in-place stack updates by recreating the resource. If we want to try and prevent errant changes to this attributes value during Terraform's plan phase, we would need to implement a much more complex `CustomizeDiff` function, although the usage of `CustomizeDiff` is also discouraged generally due to some of its limitations.

Apologies again for the long silence and thank you again for the contribution.

hnryjms:
Ah, thanks for the answer :P

I changed jobs and now don't use OpsWorks, but thanks for the answer on this repo, I'll know next time what to try if I have another PR into a Terraform provider :)

Issue #4064, opened 2018-04-05
Add application/x-x509-ca-cert as allowable MIME type for aws_s3_bucket_object


dalesit:
We would like to store keys in an encrypted S3 bucket so that a provisioned Jenkins server is able to provision other instances without having to store the key in Git. Unfortunately, the auto-detected MIME type is application/x-x509-ca-cert which is not in the whitelist of allowable MIME types to read. 

Can workaround by installing awscli, but would be useful to have available directly from Terraform.

https://github.com/terraform-providers/terraform-provider-aws/issues/1323

https://github.com/hashicorp/terraform/issues/15474

saravanan30erd:
@bflad whitelisted application/x-x509-ca-cert (#4298 ).

PR #4298, opened 2018-04-21
Issue #4064 Add application/x-x509-ca-cert as allowable MIME type for aws_s3_bucket_object

saravanan30erd:


saravanan30erd:
@bflad added acceptance test,

$ make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAWSS3BucketObject_readableBodyXX509CACert'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAWSS3BucketObject_readableBodyXX509CACert -timeout 120m
=== RUN   TestAccDataSourceAWSS3BucketObject_readableBodyXX509CACert
--- PASS: TestAccDataSourceAWSS3BucketObject_readableBodyXX509CACert (158.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	159.008s

saravanan30erd:
@bflad is there anything I can do to push this?

bflad:
@saravanan30erd you might want to take a look at the link in the comment for the `isContentTypeAllowed()` function there: https://github.com/hashicorp/terraform/pull/3858#issuecomment-156856738

We cannot guarantee that this MIME type is UTF-8 safe and it could corrupt the Terraform state given the right binary content. This would need to be Base64 encoded in another attribute for safety.

bflad:
Closing due to lack of response and because the change as implemented could introduce Terraform state corruption due to the content type. We might accept a different solution that instead works with Base64 encoding content of binary content before it enters the Terraform state.

PR #4303, opened 2018-04-21
Iot Endpoint Data Source

scottwinkler:
The iot endpoint is unique per account and is needed for connecting things to the iot registry. By having a data source for the iot endpoint, now it is possible to configure your things as part of a regular terraform deployment.

### DIFF
--- a/aws/data_source_aws_iot_endpoint_test.go
+++ b/aws/data_source_aws_iot_endpoint_test.go
@@ -6,7 +6,7 @@ import (
 	"github.com/hashicorp/terraform/helper/resource"
 )
 
-func TestAccAWSIotEndpoint(t *testing.T) {
+func TestAccAWSIotEndpointDataSource(t *testing.T) {
 	resource.Test(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
@@ -14,8 +14,8 @@ func TestAccAWSIotEndpoint(t *testing.T) {
 			resource.TestStep{
 				Config: testAccAWSIotEndpointConfig,
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttrSet("aws_iot_endpoint.example", "id"),
-					resource.TestCheckResourceAttrSet("aws_iot_endpoint.example", "endpoint_address"),
+					resource.TestCheckResourceAttrSet("data.aws_iot_endpoint.example", "id"),
+					resource.TestCheckResourceAttrSet("data.aws_iot_endpoint.example", "endpoint_address"),
 				),
 			},
 		},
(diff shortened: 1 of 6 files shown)


aws/data_source_aws_iot_endpoint_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+func TestAccAWSIotEndpoint(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSIotEndpointConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("aws_iot_endpoint.example", "id"),
```
bflad: The "resource" name here should be `data.aws_iot_endpoint.example`

```
=== RUN   TestAccAWSIotEndpoint
--- FAIL: TestAccAWSIotEndpoint (4.88s)
    testing.go:518: Step 0 error: Check failed: Check 1/2 error: Not found: aws_iot_endpoint.example in [root]
```

scottwinkler:
Thanks, @bflad . I made the changes you requested. I am wondering if I should include the "endpoint-type" parameter for the data source. I could not find any documentation about what this type is, I think it might be something that will be used in the future but is not currently used. Should I include this parameter which doesn't appear to do anything to have a complete solution or leave this for a future enhancement? https://docs.aws.amazon.com/cli/latest/reference/iot/describe-endpoint.html

bflad:
> Should I include this parameter which doesn't appear to do anything to have a complete solution or leave this for a future enhancement?

Interesting - its not well documented. I think we're okay leaving it out for now and can always add it in later.

aws/data_source_aws_iot_endpoint_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+func TestAccAWSIotEndpointDataSource(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccAWSIotEndpointConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("data.aws_iot_endpoint.example", "id"),
```
bflad: Sorry I missed this the first review, we don't need to explicitly check `id`, its implicitly expected to be there as part of being a Terraform resource/data source. I'll remove this on merge 👍

website/docs/d/iot_endpoint.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+}
+```
+
+## Argument Reference
+
+N/A
+
+## Attributes Reference
+
+* `endpoint_address` - The endpoint. The format of the endpoint is as follows: identifier .iot.*region* .amazonaws.com.
```
bflad: Formatting got a little wonky here, I'll fix on merge 👍

bflad:

Issue #4267, opened 2018-04-19
Add support for launch templates to aws_spot_fleet_request


WheresWardy:
Now that support for EC2 launch templates has been added to Terraform via https://github.com/terraform-providers/terraform-provider-aws/pull/2927, I'd like to request support for this be added to the `aws_spot_fleet_request` resource.

At the moment, the only way to add multiple variations of instances to spot fleet requests is with individual `launch_specification` parameters. This involves a lot of duplication of `aws_instance` parameters that are the same when you may just want to change the instance type across multiple launch specifications for example.

Launch templates will add the ability to supply the majority of consistent parameters via the template.

### Terraform Version
```
Terraform v0.11.7
+ provider.aws v1.15.0
+ provider.template v1.0.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- aws_spot_fleet_request

ingwarsw:
@gordonbondon Working on that maybe?
Would be nice to have that options..

gordonbondon:
@ingwarsw I'm waiting for #4543 to get merged so we can settle on how this launch_template option is implemented and then I'll add the same for spot fleet.

yanivomc:
Thank you @gordonbondon this would be awesome!

gordonbondon:
Looks like someone got to it before me #4866

lossanarch:
Oh sorry @gordonbondon, hadn't seen this issue and that you were working on it. If anyone has any ideas on how to get the testing assertions against the random-ish numbers in the state working (see WIP section of #4866) I can finish up the PR.

Jamie-BitFlight:
Would you be down for having the overrides being a list of maps?
So they can be passed in dynamically in version 0.11?

four43:
The Launch Template support in EC2 Fleets seems to work really well. Is that direction AWS seems to be going with provisioning like this (instead of spot fleets?) It sounds like Auto Scaling isn't working for those groups yet.

Either way I think that syntax would work well in for `aws_spot_fleet_request`s

almlys:
Is there are update or workaround to have it working with Terraform? I have searched a lot on the web and seems that the only solution is to keep with the "launch_specification" and we would prefer to use the "launch_template" as it suits better with our use case.

WheresWardy:
[...]

Issue #4264, opened 2018-04-19
Feature: Add support for launch_template in aws_instance resource


gordonbondon:
Please add support to provide Launch Template on instance creation. This feature is already available in go sdk https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#RunInstancesInput and `aws_launch_template` resource was recently implemented #2927

ohaiwalt:
Really interested in this, as it would solve a number of issues for us. What's the current state?

majdhamdan:
Hey. Any updates?

julian-alarcon:
I also have this question, any updates on this? PR seems to be OK.

wandrew26:
I would be happy to see this in the next release! ⚠️

dtserekhman-starz:
Yes, really waiting for this feature. Thank you.

sepulvedablanco:
When will be available this feature?

dvir-frey:
any news on it??

dvir-frey:
https://www.terraform.io/docs/providers/aws/r/spot_fleet_request.html look here

PR #4305, opened 2018-04-21
r/aws_autoscaling_group: Add support for launch template

gordonbondon:
Adds support to launch ASG from [template](https://docs.aws.amazon.com/autoscaling/ec2/userguide/create-asg-launch-template.html)

I'll file separate PRs for #4267 and #4264 . They are using the same structure.

Tests:
```
→ vaulted -n my -- make testacc TEST=./aws TESTARGS='-run=TestAccAWSAutoScalingGroup_launchTemplate'
Password:
Enter your MFA code: 272073
TF_ACC=1 go test ./aws -v -run=TestAccAWSAutoScalingGroup_launchTemplate -timeout 120m
=== RUN   TestAccAWSAutoScalingGroup_launchTemplate
--- PASS: TestAccAWSAutoScalingGroup_launchTemplate (103.86s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	104.296s

→ vaulted -n my -- make testacc TEST=./aws TESTARGS='-run=TestAccAWSAutoScalingGroup_WithLoadBalancer'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAutoScalingGroup_WithLoadBalancer -timeout 120m
=== RUN   TestAccAWSAutoScalingGroup_WithLoadBalancer
--- PASS: TestAccAWSAutoScalingGroup_WithLoadBalancer (535.68s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	536.105s

→ vaulted -n my -- make testacc TEST=./aws TESTARGS='-run=TestAccAWSAutoScalingGroup_launchTemplate_toLaunchConfig'
Password:
TF_ACC=1 go test ./aws -v -run=TestAccAWSAutoScalingGroup_launchTemplate_toLaunchConfig -timeout 120m
=== RUN   TestAccAWSAutoScalingGroup_launchTemplate_toLaunchConfig
--- PASS: TestAccAWSAutoScalingGroup_launchTemplate_toLaunchConfig (144.08s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	144.516s
```


aws/resource_aws_autoscaling_group_test.go
```diff
@@ -1290,10 +1341,20 @@ resource "aws_elb" "bar" {
...
+    name   = "name"
+    values = ["bitnami-nginxstack-*-linux-debian-9-x86_64-hvm-ebs"]
+  }
+}
+
 resource "aws_launch_configuration" "foobar" {
-  // need an AMI that listens on :80 at boot, this is:
-  // bitnami-nginxstack-1.6.1-0-linux-ubuntu-14.04.1-x86_64-hvm-ebs-ami-99f5b1a9-3
-  image_id = "ami-b5b3fc85"
+  image_id = "${data.aws_ami.test_ami.id}"
```
bflad: 😍

gordonbondon:
@bflad almost done. I found a couple of edge cases and tried to resolve them.

After my last changes acceptance test is failing like this:
```
→ vaulted -n my -- make testacc TEST=./aws TESTARGS='-run=TestAccAWSAutoScalingGroup_launchTemplate_update'
Password:
TF_ACC=1 go test ./aws -v -run=TestAccAWSAutoScalingGroup_launchTemplate_update -timeout 120m
=== RUN   TestAccAWSAutoScalingGroup_launchTemplate_update
--- FAIL: TestAccAWSAutoScalingGroup_launchTemplate_update (60.55s)
	testing.go:518: Step 2 error: Check failed: Check 2/3 error: aws_autoscaling_group.bar: Attribute 'launch_configuration' found when not expected
FAIL
exit status 1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	61.000s
make: *** [testacc] Error 1
```

But when I run the same steps manually I see `"launch_configuration": ""` in `terraform.tfstate`. Maybe you can help me to sort this out.

My other concern is that I don't really like how  `expandLaunchTemplateSpecification` is done, but I don't have an idea on how to do it better. Basic flow should be like this:
- Only one of name or id can be set for LaunchTemplateSpecification but DescribeAustoScaling group returns both of them.
- We need to check that only one of them is set
- When resource is updated check which one has changed and send it
- When none of them has changed - send first available

gordonbondon:
And maybe expandLaunchTemplateSpecification should return interface instead of `*autoscaling.LaunchTemplateSpecification`  so I can somehow make it work for https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#FleetLaunchTemplateSpecification which has the exact structure but another type.

bflad:
I wouldn't try to get fancy merging types. We prefer (mostly) duplicated code over magic/indirection as it makes community maintenance easier.

bflad:
> Only one of name or id can be set for LaunchTemplateSpecification but DescribeAustoScaling group returns both of them.

`Optional: true` and `Computed: true` should handle this situation fine.

> We need to check that only one of them is set

Why? Does the API return an error when both are specified or is this an issue during updates? If it does not error when both are specified, I would assume we're fine passing whichever is defined. If it does error on update because of ID/name mismatch, then we might need to introduce a `CustomizeDiff` function that calls `d.SetNewComputed()` on the opposite attribute

gordonbondon:
On describe it returns:
```
"LaunchTemplate": {
                "LaunchTemplateId": "lt-057f89ebf426f3765",
                "LaunchTemplateName": "test_template",
                "Version": "$Default"
            }
```

gordonbondon:
On describe it returns:
```
"LaunchTemplate": {
                "LaunchTemplateId": "lt-057f89ebf426f3765",
                "LaunchTemplateName": "test_template",
                "Version": "$Default"
            }
```

But if you set both on Create/Update request it will return error:
```
An error occurred (ValidationError) when calling the UpdateAutoScalingGroup operation: You must use a valid fully-formed launch template. A launch template ID and a launch template name cannot be specified in the same request.
```

gordonbondon:
CustomizedDiff and SetnewComputed is great! I was looking for something like this but did not succeeded. I'll try to switch to this approach

gordonbondon:
I've also tried to add plan time validation for attributes like this:
```
func(diff *schema.ResourceDiff, v interface{}) error {
                // Plan time validation for asg instance configuration
                if diff.Id() == "" || !diff.HasChange("launch_configuration") && !diff.HasChange("launch_template") {
                    return nil
                }
                launchConfigurationValue, launchConfigurationOk := diff.GetOk("launch_configuration")
                launchTemplateValue, launchTemplateOk := diff.GetOk("launch_template")
                lc := launchConfigurationOk && launchConfigurationValue.(string) != ""
                lt := launchTemplateOk && len(launchTemplateValue.([]interface{})) > 0
                if lc || lt {
                    return nil
                }
                // return nil
                return errors.New("One of `launch_configuration` or `launch_template` must be set for an autoscaling group")
            },
```

But it does not seem to work when attributes are passed from resources that does not exist yet. Dor axample ``launch_configuration = "${aws_launch_configuration.test.name}"

bflad:

PR #4314, opened 2018-04-23
d/aws_instance: add missing disable_api_termination attribute

kl4w:
@bflad The `disable_api_termination` attribute was missing from the data source as was noticed in https://github.com/terraform-providers/terraform-provider-aws/pull/2619


bflad:

PR #4321, opened 2018-04-23
r/aws_launch_template: Fix various bugs

stefansundin:
These three commits fixes the following errors:

- > Error: aws_launch_template.default: "name_prefix" can only alphanumeric characters and ()./_ symbols
- > * aws_launch_template.default: InvalidSnapshotID.Malformed: The snapshot ID '' is not valid. The expected format is snap-xxxxxxxx or snap-xxxxxxxxxxxxxxxxx.
  >  status code: 400, request id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
- > * aws_launch_template.default: block_device_mappings.0.ebs: '': source data must be an array or slice, got map

I am not sure if there are more things to fix for the middle one.. The problem is that empty strings were used when the properties were not defined (such as `snapshot_id`). I tried to find other places in the code that had similar conventions, and I just used the first one that worked for me.

I still haven't learned how to run the tests, so if anyone thinks there needs to be tests added, then please feel free to add them to this PR (edits from maintainers are allowed).

### DIFF
--- a/aws/resource_aws_launch_template.go
+++ b/aws/resource_aws_launch_template.go
@@ -597,7 +597,7 @@ func getBlockDeviceMappings(m []*ec2.LaunchTemplateBlockDeviceMapping) []interfa
 				ebs["snapshot_id"] = aws.StringValue(v.Ebs.SnapshotId)
 			}
 
-			mapping["ebs"] = ebs
+			mapping["ebs"] = []interface{}{ebs}
 		}
 		s = append(s, mapping)
 	}
--- a/aws/resource_aws_launch_template.go
+++ b/aws/resource_aws_launch_template.go
@@ -882,16 +882,16 @@ func buildLaunchTemplateData(d *schema.ResourceData, meta interface{}) (*ec2.Req
 func readBlockDeviceMappingFromConfig(bdm map[string]interface{}) *ec2.LaunchTemplateBlockDeviceMappingRequest {
 	blockDeviceMapping := &ec2.LaunchTemplateBlockDeviceMappingRequest{}
 
-	if v := bdm["device_name"]; v != nil {
-		blockDeviceMapping.DeviceName = aws.String(v.(string))
+	if v := bdm["device_name"].(string); v != "" {
+		blockDeviceMapping.DeviceName = aws.String(v)
 	}
 
-	if v := bdm["no_device"]; v != nil {
-		blockDeviceMapping.NoDevice = aws.String(v.(string))
+	if v := bdm["no_device"].(string); v != "" {
+		blockDeviceMapping.NoDevice = aws.String(v)
 	}
 
-	if v := bdm["virtual_name"]; v != nil {
-		blockDeviceMapping.VirtualName = aws.String(v.(string))
+	if v := bdm["virtual_name"].(string); v != "" {
+		blockDeviceMapping.VirtualName = aws.String(v)
 	}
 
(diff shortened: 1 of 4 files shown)


stefansundin:
I updated the docs to provide some help on the things I had trouble with.. there's an unconventional comment in there that I am not sure how to write in the best way. Please let me know what you think.

aws/validators.go
```diff
@@ -1751,8 +1751,8 @@ func validateLaunchTemplateName(v interface{}, k string) (ws []string, errors []
 		errors = append(errors, fmt.Errorf("%q cannot be longer than 99 characters, name is limited to 125", k))
 	} else if !strings.HasSuffix(k, "prefix") && len(value) > 125 {
 		errors = append(errors, fmt.Errorf("%q cannot be longer than 125 characters", k))
-	} else if !regexp.MustCompile(`^[0-9a-zA-Z()./_]+$`).MatchString(value) {
-		errors = append(errors, fmt.Errorf("%q can only alphanumeric characters and ()./_ symbols", k))
+	} else if !regexp.MustCompile(`^[0-9a-zA-Z()./_\-]+$`).MatchString(value) {
```
bflad: Adding a dash here looks good according to the EC2 API documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateLaunchTemplate.html

```
LaunchTemplateName
A name for the launch template.

Type: String

Length Constraints: Minimum length of 3. Maximum length of 128.

Pattern: [a-zA-Z0-9\(\)\.-/_]+

Required: Yes
```

website/docs/r/launch_template.html.markdown
```diff
@@ -10,16 +10,24 @@ description: |-
...
-
 ## Example Usage
 
 ```hcl
 resource "aws_launch_template" "foo" {
   name = "foo"
 
   block_device_mappings {
-    device_name = "test"
+    # to change the type or size of the root volume, override the ami's root device name
```
bflad: We should add this helper information to the `block_device_mappings` documentation below.  I'll adjust on merge 👍  More importantly its better to get the updated `device_name` information in the example.

website/docs/r/launch_template.html.markdown
```diff
@@ -10,16 +10,24 @@ description: |-
 
 Provides an EC2 launch template resource. Can be used to create instances or auto scaling groups.
 
--> **Note:** All arguments are optional except for either `name`, or `name_prefix`.
```
bflad: I'm curious why this was removed.

aws/resource_aws_launch_template.go
```diff
@@ -597,7 +597,7 @@ func getBlockDeviceMappings(m []*ec2.LaunchTemplateBlockDeviceMapping) []interfa
 				ebs["snapshot_id"] = aws.StringValue(v.Ebs.SnapshotId)
 			}
 
-			mapping["ebs"] = ebs
+			mapping["ebs"] = []interface{}{ebs}
```
bflad: An acceptance test implementing this part of the resource would've caught this. I'll add one.

website/docs/r/launch_template.html.markdown
stefansundin: As noted in the commit message, terraform will generate a name if the name is missing. This is also noted in the description of the attribute:
> The name of the launch template. If you leave this blank, Terraform will auto-generate a unique name.

bflad:

Issue #4325, opened 2018-04-24
aws_rds_cluster_instance is not able to disable MultiAZ


eduherraiz:
### Terraform Version
Terraform v0.11.7
+ provider.aws v1.15.0

### Affected Resource(s)
- aws_rds_cluster_instance

### Terraform Configuration Files
```hcl
resource "aws_rds_cluster_instance" "cluster-instances" {
  count = 2

  identifier          = "stage-${count.index}"
  cluster_identifier  = "${aws_rds_cluster.stage.id}"
  engine              = "aurora-postgresql"
  engine_version      = "9.6.6"
  instance_class      = "db.r4.large"
  publicly_accessible = false

  apply_immediately            = true
  monitoring_role_arn          = "${aws_iam_role.enhanced_policy_role.arn}"
  monitoring_interval          = "5"
  performance_insights_enabled = true
}
```
### Expected Behavior
The instance could be configured with or without the MultiAZ deployment.
![screenshot_20180424_125433](https://user-images.githubusercontent.com/717200/39182914-a3d4c32e-47be-11e8-83fa-ed5a04146d22.png)

### Actual Behavior
The actual behavior of the module aws_rds_cluster_instance is create the instance always in MultiAZ.
The Go SDK is able to manage the parameter:

https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#CreateDBInstanceInput

But we don't use it in the resourceAwsRDSClusterInstanceCreate:

https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_rds_cluster_instance.go#L204

eduherraiz:
Read my comment at: https://github.com/terraform-providers/terraform-provider-aws/pull/4329#issuecomment-385969591

PR #4329, opened 2018-04-24
Add MultiAZ parameter in rds_instances

eduherraiz:
Initial work trying to solve: #4325 
I can say the parameter in my actual case was true by default (aurora postgresql 9.6.6).

### DIFF
--- a/aws/resource_aws_rds_cluster_instance.go
+++ b/aws/resource_aws_rds_cluster_instance.go
@@ -196,6 +196,12 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
 				ValidateFunc: validateArn,
 			},
 
+			"multiaz": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+
 			"tags": tagsSchema(),
 		},
 	}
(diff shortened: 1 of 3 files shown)


aws/resource_aws_rds_cluster_instance.go
```diff
@@ -196,6 +196,12 @@ func resourceAwsRDSClusterInstance() *schema.Resource {
 				ValidateFunc: validateArn,
 			},
 
+			"multiaz": {
```
bflad: A few things with this new attribute:
* Can you please document this new argument in `website/docs/r/rds_cluster_instance.html.markdown`?
* Can you please adjust this to be named `multi_az` to match the `aws_db_instance` resource? https://www.terraform.io/docs/providers/aws/r/db_instance.html#multi_az
* Can you please add a new acceptance test in `aws/resource_aws_rds_cluster_instance_test.go` with this new argument set to `false`? Copy paste of an existing test and test configuration is a great way to get started 👍

eduherraiz:
**TLDR**: My mistake, The MultiAZ in Aurora is only a flag saying we have two instances in different zones.

Doing more tests I can see I forgot something.
Trying to enable the multi_az for a Aurora Postgresql was created with a false value I get the error:
`InvalidParameterCombination: VPC Multi-AZ DB Instances are not available for engine: aurora-postgresql`

I can say now it's only my mistake understanding the module. 
On mark the "MultiAZ" option on the web console create 2 separated instances in the cluster. 
It's no used by the API, when create the instance.

PR #4330, opened 2018-04-24
Rename codecommit documentation pages

stefansundin:
Rename codecommit documentation pages to be consistent with their resource names.

I am generating links to the docs based on the resource names, and these two are not consistent with the rest.

Maybe we want to have some kind of redirect (or temporary notice?) for people who have already linked to the current URLs?


bflad:
Upstream PR: https://github.com/hashicorp/terraform-website/pull/258

PR #4365, opened 2018-04-26
Support for proxy protocol 2 attribute

ddtmachado:
Just adding support for the attribute proxy_protocol_v2 on target groups.

I need this for our services to work and can't afford to do it manually every time we create a new environment.

I don't know if I got the test's right, feel free to change it or point the right way.

### DIFF
--- a/aws/resource_aws_lb_target_group.go
+++ b/aws/resource_aws_lb_target_group.go
@@ -83,6 +83,12 @@ func resourceAwsLbTargetGroup() *schema.Resource {
 				ValidateFunc: validation.IntBetween(0, 3600),
 			},
 
+			"proxy_protocol_v2": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
 			"target_type": {
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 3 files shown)


aws/resource_aws_lb_target_group_test.go
```diff
@@ -156,6 +157,13 @@ func TestAccAWSLBTargetGroup_networkLB_TargetGroup(t *testing.T) {
 				ExpectNonEmptyPlan: true,
 				ExpectError:        regexp.MustCompile("Health check interval cannot be updated"),
 			},
+			{
```
bflad: Can you please move this to its own acceptance test?

aws/resource_aws_lb_target_group.go
```diff
@@ -83,6 +83,12 @@ func resourceAwsLbTargetGroup() *schema.Resource {
 				ValidateFunc: validation.IntBetween(0, 3600),
 			},
 
+			"proxy_protocol_v2": {
```
bflad: Does this attribute support updates? If so, we should add an acceptance test step for that. If not, it should be marked as `ForceNew: true,`

bflad:
Oh! I forgot to mention -- this new attribute should also be documented in `website/docs/r/lb_target_group.html.markdown`

bflad:

PR #4344, opened 2018-04-25
r/aws_launch_template: Multiple fixes

gordonbondon:
Commits fix various stuff i've encountered while using this resource:
- documented that  user_data should be base64 encrypted, otherwise you get this:
> InvalidUserData.Malformed: Invalid BASE64 encoding of user data.

- Fix iops check, if not set in terraform it was set to 0 in template causing error when launching instance from template:
> The parameter iops is not supported for gp2 volumes.

- Set that iam arn conflicts with iam name. If set togather you get this when launching instance:
> 'iamInstanceProfile.name' may not be used in combination with 'iamInstanceProfile.arn'

- Set ConflictsWith for `security_group_names` and `vpc_security_group_ids`. [Docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestLaunchTemplateData.html) say that `You cannot specify both a security group ID and security name in the same request.` and you can get this error when trying to launch instance from template in non-default VPC:
> 'The parameter groupName cannot be used with the parameter subnet'

- Add some validators to attributes

Tests:
```
vaulted -n my -- make testacc TEST=./aws TESTARGS='-run=TestAccAWSLaunchTemplate_'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSLaunchTemplate_ -timeout 120m
=== RUN   TestAccAWSLaunchTemplate_importBasic
--- PASS: TestAccAWSLaunchTemplate_importBasic (24.06s)
=== RUN   TestAccAWSLaunchTemplate_importData
--- PASS: TestAccAWSLaunchTemplate_importData (22.25s)
=== RUN   TestAccAWSLaunchTemplate_basic
--- PASS: TestAccAWSLaunchTemplate_basic (20.07s)
=== RUN   TestAccAWSLaunchTemplate_data
--- PASS: TestAccAWSLaunchTemplate_data (19.96s)
=== RUN   TestAccAWSLaunchTemplate_update
--- PASS: TestAccAWSLaunchTemplate_update (35.79s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	122.487s
```

PR #4367, opened 2018-04-26
r/aws_launch_template: some more fixes

stefansundin:
The first fix is trivial: `device_index` can be 0. I don't know if it's even desirable to keep it optional, since the console defaults the value to 0 if you do not fill in the field. If you add multiple devices and leave the index empty, the console will also increment them automatically, i.e. "eth0", "eth1", etc. Maybe we should do the same? I haven't tried launching an instance with a network interface without an index, so if that breaks then this is probably desirable.

The second fix is very incomplete. The thing is that almost all of the boolean values are really optional, but terraform defaults them to false. There is no way to leave them undefined. So terraform is not able to fully configure those properties. I tried making it a string, that way we can have the values `""`, `"true"`, and `"false"`. But then we lose the ability to just use `true` or `false` without quotes in the terraform code, which is kind of confusing for users.

I first tried this to only set it if it was defined as either true or false, but it did not work.
```
if v, ok := ni["associate_public_ip_address"]; ok {
	networkInterface.AssociatePublicIpAddress = aws.Bool(v.(bool))
}
```

I tried to use `Default: nil` like this, but it did not work. Perhaps the schema could be enhanced to allow this to make the code above work?
```
"associate_public_ip_address": {
	Type:     schema.TypeBool,
	Optional: true,
	Default:  nil,
},
```

But it still defaulted to `false`.

I then tried this, which did work, but there must be a better way.

```
"associate_public_ip_address": {
	Type:     schema.TypeString,
	Optional: true,
		"true",
		"false",
	}, false),
},
```

So this affects any property that can be true, false, or undefined ("Do not include in template"). See screenshot:
screen shot 2018-04-26 at 07 45 27

I was going to fix the network interface security groups, but thankfully @kl4w opened his PR first which is much better than my fix. https://github.com/terraform-providers/terraform-provider-aws/pull/4364

### DIFF
--- a/aws/resource_aws_launch_template.go
+++ b/aws/resource_aws_launch_template.go
@@ -952,7 +952,7 @@ func readNetworkInterfacesFromConfig(ni map[string]interface{}) *ec2.LaunchTempl
 		networkInterface.Description = aws.String(v)
 	}
 
-	if v, ok := ni["device_index"].(int); ok && v != 0 {
+	if v, ok := ni["device_index"].(int); ok {
 		networkInterface.DeviceIndex = aws.Int64(int64(v))
 	}
 


bflad:
@stefansundin during create/update does `if v, ok := d.GetOkExists("attribute_name"); ok {`workaround the boolean issue? We can do full `if apiAttribute != nil {` checks during read to prevent setting the attribute instead of just using `aws.BoolValue()` (which will default to false)

bflad:
I'm guessing neither will help in this situation though 🙁

bflad:
I think we might want to making each fix/enhancement individual PRs. Starting to see lots of overlapping work in this resource. See also #4344

aws/resource_aws_launch_template.go
```diff
@@ -952,7 +955,7 @@ func readNetworkInterfacesFromConfig(ni map[string]interface{}) *ec2.LaunchTempl
 		networkInterface.Description = aws.String(v)
 	}
 
-	if v, ok := ni["device_index"].(int); ok && v != 0 {
+	if v, ok := ni["device_index"].(int); ok {
```
bflad: Can you please switch to just this change so we can close #4475 for now?

bflad: Did it automatically assign `device_index` 0? I believe that's the behavior when its not supplied (auto incrementing from there).

joernheissler: No, it was missing completely. DescribeLaunchTemplateVersions returned:
```json
"NetworkInterfaces": [
    {
        "AssociatePublicIpAddress": false,
        "DeleteOnTermination": false,
        "Ipv6AddressCount": 0,
        "SecondaryPrivateIpAddressCount": 0,
        "SubnetId": "subnet-xxxxxxxx"
    }
],
```

stefansundin: @bflad Done, sorry about the delay. I haven't had time to check back on this.

ayarmoluk17:
Can we merge this one in? launch template is still unusable with ASG because of this. Using template with ASG in VPC requires device index of 0 https://docs.aws.amazon.com/autoscaling/ec2/userguide/create-asg-launch-template.html

With fix merged it works perfectly for me.

bflad:

PR #4369, opened 2018-04-26
r/aws_api_gateway_stage: implement access logging

### LINKED ISSUE #2406: Amazon API Gateway Access Logging Support
AWS has announced an enhancement to API Gateway to write access logs in CloudWatch: https://aws.amazon.com/about-aws/whats-new/2017/11/amazon-api-gateway-supports-access-logging/

Prerequisite: #2403

### Terraform Version

```
terraform 0.10+
terraform-provider-aws 1.3.1
```

### Affected Resource(s)

- aws_api_gateway_stage

### Expected Behavior

```hcl
resource "aws_api_gateway_stage" "example" {
  # ... other configuration ...  new enhancement:
  access_log {
    destination_arn = "..."
    format          = "..." 
  }
}
```

### References

Still waiting on the aws-sdk-go API documentation to update, but appears to be new `AccessLogSettings` struct inside `Stage`.

appilon:
Fix #2406 

implement access logging for apigw stage
add acceptance test with different accepted log formats
make note of sanitization performed on destination_arn
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayStage_accessLogSettings -timeout 120m
=== RUN   TestAccAWSAPIGatewayStage_accessLogSettings
--- PASS: TestAccAWSAPIGatewayStage_accessLogSettings (331.21s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	331.260s
```

### DIFF
--- a/aws/resource_aws_api_gateway_stage.go
+++ b/aws/resource_aws_api_gateway_stage.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -21,6 +22,27 @@ func resourceAwsApiGatewayStage() *schema.Resource {
 		Delete: resourceAwsApiGatewayStageDelete,
 
 		Schema: map[string]*schema.Schema{
+			"access_log_settings": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"destination_arn": {
+							Type:     schema.TypeString,
+							Required: true,
+							StateFunc: func(arn interface{}) string {
+								// arns coming from a TF reference to a log group contain a trailing `:*` which is not valid
+								return strings.TrimSuffix(arn.(string), ":*")
+							},
+						},
+						"format": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+			},
 			"cache_cluster_enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -153,6 +175,9 @@ func resourceAwsApiGatewayStageCreate(d *schema.ResourceData, meta interface{}) 
 	if _, ok := d.GetOk("client_certificate_id"); ok {
 		return resourceAwsApiGatewayStageUpdate(d, meta)
 	}
+	if _, ok := d.GetOk("access_log_settings"); ok {
+		return resourceAwsApiGatewayStageUpdate(d, meta)
+	}
 	return resourceAwsApiGatewayStageRead(d, meta)
 }
 
@@ -176,6 +201,10 @@ func resourceAwsApiGatewayStageRead(d *schema.ResourceData, meta interface{}) er
 		return err
 	}
 	log.Printf("[DEBUG] Received API Gateway Stage: %s", stage)
+
+	if err := d.Set("access_log_settings", flattenApiGatewayStageAccessLogSettings(stage.AccessLogSettings)); err != nil {
+		return fmt.Errorf("error setting access_log_settings: %s", err)
+	}
 
 	d.Set("client_certificate_id", stage.ClientCertificateId)
 
@@ -275,6 +304,27 @@ func resourceAwsApiGatewayStageUpdate(d *schema.ResourceData, meta interface{}) 
 		oldV := o.(map[string]interface{})
 		newV := n.(map[string]interface{})
 		operations = append(operations, diffVariablesOps("/variables/", oldV, newV)...)
+	}
+	if d.HasChange("access_log_settings") {
+		accessLogSettings := d.Get("access_log_settings").([]interface{})
+		if len(accessLogSettings) == 1 {
+			operations = append(operations,
+				&apigateway.PatchOperation{
+					Op:   aws.String("replace"),
+					Path: aws.String("/accessLogSettings/destinationArn"),
+					// arns coming from a TF reference to a log group contain a trailing `:*` which is not valid
+					Value: aws.String(strings.TrimSuffix(d.Get("access_log_settings.0.destination_arn").(string), ":*")),
+				}, &apigateway.PatchOperation{
+					Op:    aws.String("replace"),
+					Path:  aws.String("/accessLogSettings/format"),
+					Value: aws.String(d.Get("access_log_settings.0.format").(string)),
+				})
+		} else if len(accessLogSettings) == 0 {
+			operations = append(operations, &apigateway.PatchOperation{
+				Op:   aws.String("remove"),
+				Path: aws.String("/accessLogSettings"),
+			})
+		}
 	}
 
 	input := apigateway.UpdateStageInput{
@@ -383,4 +433,15 @@ func resourceAwsApiGatewayStageDelete(d *schema.ResourceData, meta interface{}) 
 	}
 
 	return nil
+}
+
+func flattenApiGatewayStageAccessLogSettings(accessLogSettings *apigateway.AccessLogSettings) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, 1)
+	if accessLogSettings != nil {
+		result = append(result, map[string]interface{}{
+			"destination_arn": aws.StringValue(accessLogSettings.DestinationArn),
+			"format":          aws.StringValue(accessLogSettings.Format),
+		})
+	}
+	return result
 }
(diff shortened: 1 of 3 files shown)


aws/resource_aws_api_gateway_stage_test.go
```diff
@@ -58,6 +60,67 @@ func TestAccAWSAPIGatewayStage_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr("aws_api_gateway_stage.test", "access_log_settings.#", "1"),
+					testAccCheckAWSAPIGatewayStageAccessLogSettingsDestinationArnIsValid("aws_api_gateway_stage.test"),
+					resource.TestCheckResourceAttr("aws_api_gateway_stage.test", "access_log_settings.0.format", csv),
+				),
+			},
+			resource.TestStep{
+				Config: testAccAWSAPIGatewayStageConfig_basic(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAPIGatewayStageExists("aws_api_gateway_stage.test", &conf),
+					testAccCheckAWSAPIGatewayStageAccessLogSettingsIsNil(&conf),
```
bflad: You can combine both this check and the below with `resource.TestCheckResourceAttr("aws_api_gateway_stage.test", "access_logs_settings.#", "0")`

aws/resource_aws_api_gateway_stage_test.go
```diff
@@ -58,6 +60,67 @@ func TestAccAWSAPIGatewayStage_basic(t *testing.T) {
...
+	xml := `<request id="$context.requestId"> <ip>$context.identity.sourceIp</ip> <caller>$context.identity.caller</caller> <user>$context.identity.user</user> <requestTime>$context.requestTime</requestTime> <httpMethod>$context.httpMethod</httpMethod> <resourcePath>$context.resourcePath</resourcePath> <status>$context.status</status> <protocol>$context.protocol</protocol> <responseLength>$context.responseLength</responseLength> </request>`
+	csv := `$context.identity.sourceIp,$context.identity.caller,$context.identity.user,$context.requestTime,$context.httpMethod,$context.resourcePath,$context.protocol,$context.status,$context.responseLength,$context.requestId`
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSAPIGatewayStageDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
```
bflad: Minor nitpick: The `resource.TestStep` declarations here are extraneous as its already defined by `[]resource.TestStep{` above

aws/resource_aws_api_gateway_stage_test.go
```diff
@@ -58,6 +60,67 @@ func TestAccAWSAPIGatewayStage_basic(t *testing.T) {
...
+				Config: testAccAWSAPIGatewayStageConfig_accessLogSettings(rName, clf),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAPIGatewayStageExists("aws_api_gateway_stage.test", &conf),
+					resource.TestCheckResourceAttr("aws_api_gateway_stage.test", "access_log_settings.#", "1"),
+					testAccCheckAWSAPIGatewayStageAccessLogSettingsDestinationArnIsValid("aws_api_gateway_stage.test"),
```
bflad: Something for consideration instead of writing a new function:

```go
resource.TestMatchResourceAttr("aws_api_gateway_stage.test", "access_log_settings.0.destination_arn", regexp.MustCompile(fmt.Sprintf("^arn:[^:]+:logs:[^:]+:[^:]+:foo-bar-%s$", rName))),
```

website/docs/r/api_gateway_stage.html.markdown
```diff
@@ -69,6 +69,7 @@ The following arguments are supported:
 * `rest_api_id` - (Required) The ID of the associated REST API
 * `stage_name` - (Required) The name of the stage
 * `deployment_id` - (Required) The ID of the deployment that the stage points to
+* `access_log_settings` - (Optional) enables access logs for the api stage
```
bflad: Should probably mention: ...`API stage. Detailed below.`

website/docs/r/api_gateway_stage.html.markdown
```diff
@@ -78,6 +79,15 @@ The following arguments are supported:
 * `variables` - (Optional) A map that defines the stage variables
 * `tags` - (Optional) A mapping of tags to assign to the resource.
 
+### Nested Blocks
+
+#### `access_log_settings`
+
+* `destination_arn` - (Required) ARN of the log group to send the logs to.
+~> **Note:** Referencing a log group arn from a Terraform resource will normally include a `:*` suffix which is not valid for the destination arn, Terraform automatically strips this suffix when configuring this argument.
```
bflad: I don't think we need this bold callout that makes it sound like Terraform is doing something wrong or this is Terraform specific behavior (as the ARN value with `:*` coming from the CloudWatch API). I think a simple ``Automatically removes trailing `:*` if present.`` in the line above would suffice. 👍

aws/resource_aws_api_gateway_stage.go
```diff
@@ -384,3 +432,17 @@ func resourceAwsApiGatewayStageDelete(d *schema.ResourceData, meta interface{})
 
 	return nil
 }
+
+func flattenApiGatewayStageAccessLogSettings(accessLogSettings *apigateway.AccessLogSettings) []map[string]interface{} {
+	if accessLogSettings == nil {
+		return nil
+	}
+	item := make(map[string]interface{})
+	if accessLogSettings.DestinationArn != nil {
```
bflad: Something to consider when working in the AWS provider is that the SDK provides helper functions so you don't need these wrapping `nil` checks (in most cases anyways): `item["destination_arn"] = aws.StringValue(accessLogSettings.DestinationArn)`

aws/resource_aws_api_gateway_stage_test.go
appilon: your regex skills are lit 🔥

appilon: sadly that rule didn't work I was inspired by you though and tweaked it let me know if you think it's robust enough.
```
logGroupArnRegex := regexp.MustCompile(fmt.Sprintf("^(arn:aws:logs:)(.*)(log-group:foo-bar-%s)$", rName))
```

aws/resource_aws_api_gateway_stage_test.go
appilon: because my flattener returns `nil` if the sdk struct is `nil`, `access_logs_settings.#` doesn't even exist in the state, should I just check that `access_logs_settings` is not set?

aws/resource_aws_api_gateway_stage_test.go
bflad: Looks like it just needed `log-group:` before `foo-bar-%s`:
```
resource.TestMatchResourceAttr("aws_api_gateway_stage.test", "access_log_settings.0.destination_arn", regexp.MustCompile(fmt.Sprintf("^arn:[^:]+:logs:[^:]+:[^:]+:log-group:foo-bar-%s$", rName))),
```

Whatever you think is best. I try to avoid wildcards.

I will say we should to try to not hardcode `arn:aws` at the beginning because at some point these tests will be running in US GovCloud (`aws-us-gov`) and maybe China partitions (`aws-cn`).

aws/resource_aws_api_gateway_stage.go
```diff
@@ -276,6 +305,27 @@ func resourceAwsApiGatewayStageUpdate(d *schema.ResourceData, meta interface{})
 		newV := n.(map[string]interface{})
 		operations = append(operations, diffVariablesOps("/variables/", oldV, newV)...)
 	}
+	if d.HasChange("access_log_settings") {
+		accessLogSettings := d.Get("access_log_settings").([]interface{})
```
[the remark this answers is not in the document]
bflad: As far as I know `helper/schema.Schema` should return the appropriate Go type `[]interface{}` here with either 0 or 1 elements: https://godoc.org/github.com/hashicorp/terraform/helper/schema#Schema

aws/resource_aws_api_gateway_stage_test.go
bflad: This does bring up a valid point that maybe we should have a simple ARN matching `TestCheckFunc` instead of handcrafting these.

appilon: 👍 to arn matcher and to your points on harcoded `arn:aws` I didn't even think of that!

bflad:

PR #4381, opened 2018-04-27
make: Extract PKG_NAME to a variable

radeksimko:


radeksimko:
@bflad it should indeed - updated.

PR #4383, opened 2018-04-28
Always set elb access log bucket prefix.

### LINKED ISSUE #4361: AWS ELB ModifyLoadBalancerAttributes omits required parameter "s3BucketPrefix" on modify
When configuring an Elastic Load Balancer to emit access logs to an S3 bucket, if you wish to emit the logs to the root of the target bucket, you specify the parameter bucket_prefix to be an empty string, or you do not specify the parameter as it is an optional parameter.

When a bucket is created using this configuration, everything work as expected, or if the load balancer does not currently have an access log configuration and you add it, it also works. The ELB is created with the CreateLoadBalancer API call (if a new elb), and then the bucket is modified with the ModifyLoadBalancerAttributes API call like with this information:
```json
"accessLog":{"s3BucketPrefix":"","enabled":true,"s3BucketName":"<bucket_name>","emitInterval":5},"
```

If however you are modifying a bucket that already exists, and already has an access log configuration to have this new access log configuration where the logs are to be stored in the root of the bucket, the subsequent ModifyLoadBalancerAttributes API call passes this information, because it thinks the bucket_prefix being empty means it does not need to be specified:
```json
"accessLog":{"enabled":true,"s3BucketName":"<bucket_name>","emitInterval":5},"
```

Because the `s3BucketPrefix` key is missing from the API call, the API thinks you are not wanting to change the s3BucketPrefix from the old value to EmptyString, and so given your new bucket configuration does not allow the writing of logs to the old prefix location, the Load Balancer rejects the configuration with the following message:

`Access Denied for bucket: <bucket_name>. Please check S3bucket permission`

This has been confirmed with the AWS ELB Service Team as a direct result of omitting the s3BucketPrefix key which is a required parameter if you are attempting to modify the prefix.

Terraform respects this requirement at Create time, or when adding a logging configuration but does not respect it at Modify time when changing an existing logging configuration.

*If you don't specify the new prefix, you won't change the prefix, especially when you want to change the prefix to ""*

### Terraform Version
0.11.3

### Provider Version
1.13.0

### Affected Resource(s)
- aws_elb

```hcl
resource "aws_elb" "elb" {
  name = "elb"

  access_logs {
    bucket        = "${aws_s3_bucket.lb_logs.id}"
    interval      = "5"
    enabled       = "true"
  }
```
OR
```hcl
resource "aws_elb" "elb" {
  name = "elb"

  access_logs {
    bucket        = "${aws_s3_bucket.lb_logs.id}"
    bucket_prefix = ""
    interval      = "5"
    enabled       = "true"
  }
```

### Debug Output
https://gist.github.com/Zordrak/e3cfaa83f2dcc486403304b5aa77d21d

### Expected Behavior
```json
{
    "LoadBalancerAttributes": {
        "CrossZoneLoadBalancing": {
            "Enabled": true
        }, 
        "ConnectionSettings": {
            "IdleTimeout": 400
        }, 
        "AccessLog": {
            "S3BucketPrefix": "", 
            "EmitInterval": 5, 
            "Enabled": true, 
            "S3BucketName": "<bucket_name>"
        }
    }, 
    "LoadBalancerName": "elb"
}
```

### Actual Behavior
An error occurred (InvalidConfigurationRequest) when calling the ModifyLoadBalancerAttributes operation: Access Denied for bucket: <bucket_name>. Please check S3bucket permission

### Steps to Reproduce

1. Define an elastic load balancer with access logging configuration pointint to an S3 bucket with a specific non-null prefix location.
2. Define the bucket to allow the ELB service account to write to that prefix location.
3. `terraform apply`
4. Change the bucket configuration to allow the ELB Service Account to write to the bucket root.
5. Change the access log configuration to point to the root of the bucket by removing the bucket_prefix parameter or by speciying the bucket_prefix as an empty string.
6. `terraform apply`

jmcarp:
If a user sets the elb access log bucket prefix to a non-empty value,
then sets it to empty, we have to set the bucket prefix in the
`ModifyLoadBalancerAttributes` request so that AWS sets an empty bucket
prefix rather than ignoring the value.

[Fixes #4361]

cc @zordrak

### DIFF
--- a/aws/resource_aws_elb.go
+++ b/aws/resource_aws_elb.go
@@ -584,17 +584,12 @@ func resourceAwsElbUpdate(d *schema.ResourceData, meta interface{}) error {
 		logs := d.Get("access_logs").([]interface{})
 		if len(logs) == 1 {
 			l := logs[0].(map[string]interface{})
-			accessLog := &elb.AccessLog{
-				Enabled:      aws.Bool(l["enabled"].(bool)),
-				EmitInterval: aws.Int64(int64(l["interval"].(int))),
-				S3BucketName: aws.String(l["bucket"].(string)),
+			attrs.LoadBalancerAttributes.AccessLog = &elb.AccessLog{
+				Enabled:        aws.Bool(l["enabled"].(bool)),
+				EmitInterval:   aws.Int64(int64(l["interval"].(int))),
+				S3BucketName:   aws.String(l["bucket"].(string)),
+				S3BucketPrefix: aws.String(l["bucket_prefix"].(string)),
 			}
-
-			if l["bucket_prefix"] != "" {
-				accessLog.S3BucketPrefix = aws.String(l["bucket_prefix"].(string))
-			}
-
-			attrs.LoadBalancerAttributes.AccessLog = accessLog
 		} else if len(logs) == 0 {
 			// disable access logs
 			attrs.LoadBalancerAttributes.AccessLog = &elb.AccessLog{
--- a/aws/resource_aws_lb.go
+++ b/aws/resource_aws_lb.go
@@ -1,13 +1,12 @@
 package aws
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"regexp"
 	"strconv"
 	"time"
-
-	"bytes"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -139,13 +138,11 @@ func resourceAwsLb() *schema.Resource {
 						"prefix": {
 							Type:             schema.TypeString,
 							Optional:         true,
-							Computed:         true,
 							DiffSuppressFunc: suppressIfLBType("network"),
 						},
 						"enabled": {
 							Type:             schema.TypeBool,
 							Optional:         true,
-							Computed:         true,
 							DiffSuppressFunc: suppressIfLBType("network"),
 						},
 					},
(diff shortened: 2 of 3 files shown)


jmcarp:
I think there were two bugs responsible for #4361. First, the provider logic only set the bucket prefix if its value isn't the empty string. Second, terraform core doesn't recognize a diff if a value is the zero value for its type. So to fix this behavior, I had to both modify the provider logic to always set the bucket prefix, as well as creating a null value for the bucket prefix that isn't the empty string. I picked "/", since I think that's a reasonable null value for a path, and stripped leading and trailing slashes from the prefix in the provider so that we don't try to create invalid bucket paths in the load balancer attributes. Does that make sense @bflad ?

bflad:
> as well as creating a null value for the bucket prefix that isn't the empty string

Can you please explain why this was necessary?

jmcarp:
Looks like I misunderstood something about zero values in terraform. I noticed that setting the access logs bucket prefix to the empty string, or deleting the value entirely, was ignored and didn't lead to a diff in the plan. I thought that that was caused by https://github.com/hashicorp/terraform/issues/5471, but now that I'm looking at this again, I think the issue is that the prefix field is marked as `Computed`. If I drop the `Computed` flag on the `prefix` property, then terraform recognizes a diff when I set the `prefix` to the empty string. Would it make sense to drop `Computed` and add a default of empty string?

bflad:
This needs to be rebased now that the other testing PR got merged 👍

bflad:

PR #4386, opened 2018-04-28
Add support for AWS Batch job definition timeout

hectcastro:
Update the existing `aws_batch_job_definition` resource so that it supports the newly added timeout feature. The timeout configuration is supplied through a `timeout` block inside `aws_batch_job_definition` with a single `attempt_duration_seconds` attribute.

This was tested via `make test` and by building a `terraform-provider-aws` binary. The binary was used with an existing Terraform project after `timeout {}` blocks were added to the existing `aws_batch_job_definition` resources.

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/4186

See:

- https://docs.aws.amazon.com/sdk-for-go/api/service/batch/#RegisterJobDefinitionInput
- https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#timeout

### DIFF
--- a/aws/resource_aws_batch_job_definition.go
+++ b/aws/resource_aws_batch_job_definition.go
@@ -56,6 +56,20 @@ func resourceAwsBatchJobDefinition() *schema.Resource {
 					},
 				},
 			},
+			"timeout": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"attempt_duration_seconds": {
+							Type:     schema.TypeInt,
+							Required: true,
+						},
+					},
+				},
+			},
 			"type": {
 				Type:         schema.TypeString,
 				Required:     true,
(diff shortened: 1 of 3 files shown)


aws/resource_aws_batch_job_definition.go
```diff
@@ -56,6 +56,20 @@ func resourceAwsBatchJobDefinition() *schema.Resource {
...
+			"timeout": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"attempt_duration_seconds": {
+							Type:     schema.TypeInt,
+							Required: true,
```
gheine: According to the AWS documentation, this is an optional field, see https://docs.aws.amazon.com/batch/latest/APIReference/API_JobTimeout.html

gheine: Correct, both timeout and attempt_duration_seconds are optional. I guess the idea is that at some point in the future AWS might add more ways to specify a timeout (eg. attempt_duration_minutes), in which case the individual fields can not be required.

website/docs/r/batch_job_definition.html.markdown
```diff
@@ -70,6 +71,12 @@ The following arguments are supported:
 
 * `attempts` - (Required) The number of times to move a job to the `RUNNABLE` status. You may specify between `1` and `10` attempts.
 
+## timeout
+
+`timeout` supports the following:
+
+* `attempt_duration_seconds` - (Required) The time duration in seconds after which AWS Batch terminates your jobs if they have not finished. The minimum value for the timeout is `60` seconds.
```
gheine: This still says `Required` (and likewise for `attempts` above). Both are now `Optional`

bflad:

PR #4388, opened 2018-04-28
Update alb tests to run on any partition.

jmcarp:
This patch gets alb acceptance tests running in govcloud. If this approach makes sense, we can apply it to other acceptance tests--although it looks like a few tests are hard-coded to run in us-west-1 only. WDYT @bflad ?

### DIFF
--- a/aws/resource_aws_lb_test.go
+++ b/aws/resource_aws_lb_test.go
@@ -1769,6 +1769,8 @@ resource "aws_s3_bucket" "logs" {
   }
 }
 
+data "aws_partition" "current" {}
+
 data "aws_caller_identity" "current" {}
 
 data "aws_elb_service_account" "current" {}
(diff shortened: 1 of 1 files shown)


bflad:
These look like they could be replaced with `${aws_s3_bucket.logs.arn}` and `${data. aws_elb_service_account.current.arn}` respectively.

jmcarp:
@bflad: I replaced the second arn, but not the first, since that would create a circular reference between the logging bucket and its policy document. Does that make sense?

bflad:
Sorry yes it would 😓  -- lets use the `aws_partition` data source then rather than parsing a field out of the STS caller identity.

jmcarp:
Thanks, I somehow hadn't noticed that data source type. Updated.

PR #4117, opened 2018-04-08
resource/cloudfront_distribution: Added ordered cache behaviors

Ninir:
## Description

This allows configuring CloudFront cache behavior precedence.
This is an alternative approach to https://github.com/terraform-providers/terraform-provider-aws/pull/1269 and heavily inspired by it.

While the other PR is a breaking change and thus does not allow a quick merge, this approach introduces a new attribute called `ordered_cache_behavior`, deprecating `cache_behavior`.
This work is backward-compatible as it keeps the old behaviour and adds a new one, mainly benefiting from new methods. The previous methods are marked and named as `XXXDeprecated`.

It also fixes the sub-options `allowed_methods` and `cached_methods` which were lists whereas it should have been set.

Alternative & better names for the new attribute are welcomed :)

Sidenote: panic when running tests after 2hours, main tests + ordered cache behaviors are passing but might be a good idea to run a TC test.

## Related issues
https://github.com/terraform-providers/terraform-provider-aws/issues/188
https://github.com/terraform-providers/terraform-provider-aws/pull/1269

## Test

```
$ TF_LOG=DEBUG TF_LOG_PATH=terraform.log make testacc TEST=./aws TESTARGS='-run=TestAccAWSCloudFrontDistribution_'                    
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSCloudFrontDistribution_ -timeout 120m
=== RUN   TestAccAWSCloudFrontDistribution_importBasic
--- PASS: TestAccAWSCloudFrontDistribution_importBasic (1247.34s)
=== RUN   TestAccAWSCloudFrontDistribution_S3Origin
--- PASS: TestAccAWSCloudFrontDistribution_S3Origin (1323.08s)
=== RUN   TestAccAWSCloudFrontDistribution_S3OriginWithTags
--- PASS: TestAccAWSCloudFrontDistribution_S3OriginWithTags (1382.72s)
=== RUN   TestAccAWSCloudFrontDistribution_customOrigin
--- PASS: TestAccAWSCloudFrontDistribution_customOrigin (1373.03s)
=== RUN   TestAccAWSCloudFrontDistribution_multiOrigin
--- PASS: TestAccAWSCloudFrontDistribution_multiOrigin (1042.85s)
=== RUN   TestAccAWSCloudFrontDistribution_orderedCacheBehavior
--- PASS: TestAccAWSCloudFrontDistribution_orderedCacheBehavior (798.87s)
[...]

PR #4389, opened 2018-04-30
[WIP] Allow updates of ebs volumes nested within instances.

### LINKED ISSUE #4219: increase in ebs volume size doesn't work when used in aws_instance
_This issue was originally opened by @amjad489 as hashicorp/terraform#17868. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
Terraform v0.11.4
```

```hcl
resource "aws_instance" "test_instance" {
  ...
  root_block_device {
    volume_type           = "gp2"
    volume_size           = "30"
    delete_on_termination = "false"
  }

....
  ebs_block_device {
    device_name           = "/dev/sdg"
    volume_type           = "gp2"
    volume_size           = "30"
  }

....
}
```

### Expected Behavior

when the template is run for the first time it will create instance with root and ebs volume which is fine. later if we want to increase the ebs volume size it should increase the size
### Actual Behavior

doesn't increase the size
### Steps to Reproduce

1. `terraform init`
2. `terraform apply`
3. Increase the ebs_volume size
4. `terraform apply`
nothing changes

jmcarp:
[Fixes #4219]

Note: this isn't complete--I'm submitting to get feedback before going farther.

Currently, any updates to `ebs_block_device` properties are silently ignored, as described in #4219. It looks like updates are ignored because the `Set` function on the `ebs_block_device` property ignores all properties but `device_name` and `snapshot_id`. As described in https://github.com/hashicorp/terraform/issues/10520#issuecomment-264681350, `Set` is used both for identity and to detect changes, so any changes to properties that aren't used in the `Set` function are ignored on update.

To get around that issue, I'm proposing that we set `ebs_block_device` to `TypeList` instead. Then, on update, we can change properties like `volume_size` and `delete_on_termination` with `ModifyVolume` and `ModifyInstanceAttribute` on update. For now, I'm not allowing the addition or deletion of volumes from the instance without a recreate, but we could do that too if we want to, by creating and mounting (or unmounting and deleting) volumes as requested.

FWIW, I think we should recommend not nesting ebs volumes within the instance config, and using the `aws_ebs_volume` resource instead. When using nested volumes, we don't have a unique, user-defined label for each volume other than list order, which isn't ideal. The semantics of `ForceNew` are unclear too: flipping a volume to encrypted requires a recreate of the volume, but not the ec2 instance it's nested within--should that property be marked `ForceNew` or not? IMO it's much simpler to use separate resources.

Does this approach make sense @bflad ? I can flesh this out and add tests if so. For now, I've just been testing manually.

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -283,7 +283,7 @@ func resourceAwsInstance() *schema.Resource {
 			},
 
 			"ebs_block_device": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
 				Optional: true,
 				Computed: true,
 				Elem: &schema.Resource{
@@ -292,7 +292,6 @@ func resourceAwsInstance() *schema.Resource {
 							Type:     schema.TypeBool,
 							Optional: true,
 							Default:  true,
-							ForceNew: true,
 						},
 
 						"device_name": {
@@ -312,7 +311,6 @@ func resourceAwsInstance() *schema.Resource {
 							Type:             schema.TypeInt,
 							Optional:         true,
 							Computed:         true,
-							ForceNew:         true,
 							DiffSuppressFunc: iopsDiffSuppressFunc,
 						},
 
@@ -327,14 +325,12 @@ func resourceAwsInstance() *schema.Resource {
 							Type:     schema.TypeInt,
 							Optional: true,
 							Computed: true,
-							ForceNew: true,
 						},
 
 						"volume_type": {
 							Type:     schema.TypeString,
 							Optional: true,
 							Computed: true,
-							ForceNew: true,
 						},
 
 						"volume_id": {
@@ -342,13 +338,6 @@ func resourceAwsInstance() *schema.Resource {
 							Computed: true,
 						},
 					},
-				},
-				Set: func(v interface{}) int {
-					var buf bytes.Buffer
-					m := v.(map[string]interface{})
-					buf.WriteString(fmt.Sprintf("%s-", m["device_name"].(string)))
-					buf.WriteString(fmt.Sprintf("%s-", m["snapshot_id"].(string)))
-					return hashcode.String(buf.String())
 				},
 			},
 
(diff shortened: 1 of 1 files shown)


aws/resource_aws_instance.go
```diff
@@ -283,7 +283,7 @@ func resourceAwsInstance() *schema.Resource {
 			},
 
 			"ebs_block_device": {
-				Type:     schema.TypeSet,
+				Type:     schema.TypeList,
```
bflad: I am personally not aware of a good way to migrate the Terraform state from `TypeSet` to `TypeList`. When we are running into this situation its easier to implement something similar to #4117. A lot of investigative work has been done in this regard, see also #1099 and #1476.

jmcarp: Got it--so we'd add an `ordered_ebs_block_device` field using `TypeList` that `ConflictsWith` the existing field? But if that's the right way to do this, then users are going to have to update their configs to get modifiable inline ebs volumes--and if they're going to have to update anyway, maybe it's better just to recommend using separate `aws_ebs_volume` resources instead. WDYT?

bflad:
Thanks again for submitting this @jmcarp. I'm going to close out this PR since its not necessarily desirable or actionable in its current state unless it occurs during a major version update. Even then, the device ordering in the list becomes a problem as the API I believe is not guaranteed to be consistent.

The behavior of supporting EBS volumes inline should probably have more warnings attached to it since its behavior has many quirks. As you mention, the more desirable solution will be to use separate resources that can appropriately manage their lifecycle instead of trying to replicate all that behavior within this resource. Documentation updates are probably the way to go here for now until we can see if there are better ways we can handle this in Terraform as a whole as these situations are generally annoying to work with for the "inline" configurations. Maybe we should also consider deprecating this support inline, although that deprecation would certainly not happen until after 2.0 of the provider is released.

I'm not sure if we have future design plans to support some sort of "nested" resource model within Terraform, e.g. a `resource "aws_ebs_volume"` declaration inside `aws_instance` -- Any discussion about that type of support would need to happen upstream in Terraform core in a design issue before anything could be done in the AWS provider. 👍

PR #4393, opened 2018-04-30
Support timeout in aws_batch_job_definition

gheine:


bflad:
🙁  hmm looks like we already have an in progress pull request for this: #4386 

Are you okay if we close this in preference of the earlier one? Thanks.

gheine:
@bflad ah balls didn't see that. Thx, will close this and comment in the other.

PR #4406, opened 2018-05-01
 Add support for MongoDB as a DMS source endpoint 

microamp:
This PR is to add support for MongoDB as a DMS source endpoint.

It re-uses some of the attributes from the top-level namespace (`server_name`, `port`, `username`, `password` and `database_name`), but it also introduces MongoDB-specific attributes under `mongodb_settings` closely mimicking the output of `aws dms create-endpoint --generate-cli-skeleton`.

```
{
    ...
    "Username": "",
    "Password": "",
    "ServerName": "",
    "Port": 0,
    "DatabaseName": "",
    ...
    "MongoDbSettings": {
        "Username": "",
        "Password": "",
        "ServerName": "",
        "Port": 0,
        "DatabaseName": "",
        "AuthType": "no",
        "AuthMechanism": "scram_sha_1",
        "NestingLevel": "one",
        "ExtractDocId": "",
        "DocsToInvestigate": "",
        "AuthSource": ""
    }
}
```

Example usage:
```
resource "aws_dms_endpoint" "dms_endpoint" {
	...
	endpoint_type = "source"
	engine_name = "mongodb"
	...
	server_name = "..."
	port = ...
	username = "..."
	password = "..."
	database_name = "..."
	...
	mongodb_settings {
		auth_type = "PASSWORD"
		auth_mechanism = "DEFAULT"
		nesting_level = "NONE"
		extract_doc_id = "false"
		docs_to_investigate = "1000"
		auth_source = "admin"
	}
}
```

Here's an acceptance test for the new endpoint support.

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSDmsEndpointMongoDb'
```
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSDmsEndpointMongoDb -timeout 120m
=== RUN   TestAccAWSDmsEndpointMongoDb
--- PASS: TestAccAWSDmsEndpointMongoDb (60.58s)
PASS
ok     github.com/microamp/terraform-provider-aws/aws 60.596s
```

Thanks. Please let me know if I missed anything. Still very new to Terraform. :)

### DIFF
--- a/aws/resource_aws_dms_endpoint.go
+++ b/aws/resource_aws_dms_endpoint.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"fmt"
 	"log"
 	"strings"
 	"time"
@@ -71,6 +72,7 @@ func resourceAwsDmsEndpoint() *schema.Resource {
 					"redshift",
 					"sybase",
 					"sqlserver",
+					"mongodb",
 				}, false),
 			},
 			"extra_connection_attributes": {
@@ -117,6 +119,52 @@ func resourceAwsDmsEndpoint() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			// With default values as per https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.MongoDB.html
+			"mongodb_settings": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					if old == "1" && new == "0" {
+						return true
+					}
+					return false
+				},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"auth_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "PASSWORD",
+						},
+						"auth_mechanism": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "DEFAULT",
+						},
+						"nesting_level": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "NONE",
+						},
+						"extract_doc_id": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "false",
+						},
+						"docs_to_investigate": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "1000",
+						},
+						"auth_source": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "admin",
+						},
+					},
+				},
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


aws/resource_aws_dms_endpoint.go
```diff
@@ -333,13 +409,41 @@ func resourceAwsDmsEndpointSetState(d *schema.ResourceData, endpoint *dms.Endpoi
...
 		}
-	} else {
+	case "mongodb":
+		if endpoint.MongoDbSettings != nil {
+			d.Set("username", endpoint.MongoDbSettings.Username)
+			d.Set("server_name", endpoint.MongoDbSettings.ServerName)
+			d.Set("port", endpoint.MongoDbSettings.Port)
+			d.Set("database_name", endpoint.MongoDbSettings.DatabaseName)
+
+			d.Set("mongodb_settings.0.auth_type", endpoint.MongoDbSettings.AuthType)
```
bflad: To set `TypeList` or `TypeSet` nested attributes into the Terraform state, we need to create a function that outputs a slice. These current `d.Set()` calls are likely returning an error that is not caught and silently ignoring the issue, which means that Terraform is going to "pass through" the Terraform configuration as-is into the state and it will never get updated from the API.

Here's an example implementation:

```go
// Always perform error checking when setting non-scalar (TypeList, TypeSet, TypeMap) attributes
if err := d.Set("mongo_settings", flattenDmsMongoDbSettings(endpoint.MongoDbSettings)); err != nil {
  return fmt.Errorf("error setting mongo_settings: %s", err)
}

// Outside the read function
func flattenDmsMongoDbSettings(settings *dms.MongoDbSettings) []map[string]interface{} {
  if settings == nil {
  }

  m := map[string]interface{}{
    "auth_mechanism": aws.StringValue(settings.AuthMechanism),
    "auth_source": aws.StringValue(settings.AuthSource),
    "docs_to_investigate": aws.StringValue(settings.DocsToInvestigate),
    "extract_doc_id": aws.StringValue(settings.ExtractDocId),
    "nesting_level": aws.StringValue(settings.NestingLevel),
  }

  return []map[string]interface{}{m}
}
```

aws/resource_aws_dms_endpoint.go
```diff
@@ -333,13 +409,41 @@ func resourceAwsDmsEndpointSetState(d *schema.ResourceData, endpoint *dms.Endpoi
...
+			d.Set("mongodb_settings.0.extract_doc_id", endpoint.MongoDbSettings.ExtractDocId)
+			d.Set("mongodb_settings.0.docs_to_investigate", endpoint.MongoDbSettings.DocsToInvestigate)
+			d.Set("mongodb_settings.0.auth_source", endpoint.MongoDbSettings.AuthSource)
+		} else {
+			d.Set("username", endpoint.Username)
+			d.Set("server_name", endpoint.ServerName)
+			d.Set("port", endpoint.Port)
+			d.Set("database_name", endpoint.DatabaseName)
+
+			// Default values as per https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.MongoDB.html
```
bflad: We should move these to the `Default` parameter in the resource schema: https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#default

e.g. 
```go
"auth_type": {
  Type:     schema.TypeString,
  Optional: true,
  Default: "PASSWORD",
},
```

aws/resource_aws_dms_endpoint.go
```diff
@@ -117,6 +118,39 @@ func resourceAwsDmsEndpoint() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"mongodb_settings": {
```
bflad: This attribute is missing documentation. 😄 

We may start seeing differences that are something like this if the argument is not present in the Terraform configuration (regardless of whether or not all the children attributes are the default values):

```
mongodb_settings.#: "1" => "0"
```

Due to some limitations with Terraform's helper/schema, we will likely need to workaround this with a [DiffSuppressFunc](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#diffsuppressfunc) on the `mongodb_settings` attribute:

```go
  if old == "1" && new == "0" {
    return true
  }
  return false
},
```

microamp:
Hi @bflad :wave: 

Thanks a lot for the comments, super helpful! I've included the suggested changes then squashed the commits afterwards. Please let me know if everything is in its place now.

bflad:

PR #4420, opened 2018-05-02
Change ECS service grace time to 7200 seconds

LordCrocket:
The ecs service grace time now supports a value up to 7200 seconds.


LordCrocket:
AWS had the value documented as 1800 seconds, but it did not match the implementation. The statement has been removed from the aws documentation. The number 7200 comes from the error message received when trying to set it higher.

image

bflad:

PR #3549, opened 2018-02-27
Make spot price optional to use AWS default

YakDriver:
Fixes #3548

PR #4424, opened 2018-05-02
Spot price optional for SIR, SFR

### LINKED ISSUE #3548: Spot price should be optional in aws_spot_instance_request resource
In version 0.11.3, spot_price is **required**. However, according to the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestSpotInstances.html), the spot price is not required and defaults to the On-Demand price. This is _very convenient_ because you don't need to provide any price if you are willing to have the On-Demand price be your maximum price. (Of course, you are hoping to come in below that and usually you will.)

### Affected Resource(s)
- aws_spot_instance_request
- aws_spot_fleet_request

### Expected Behavior
You don't provide a `spot_price` and AWS will default to the On-Demand price (max price, not price you are necessarily going to pay).

### Actual Behavior
Throws an error because `spot_price` is required.

### Steps to Reproduce
1. Create an aws_spot_instance_request/aws_spot_fleet_request without spot_price
2. When you `terraform init`, error is thrown

YakDriver:
Code + tests + docs
Fixes #3548, formerly #3549 (closed)

### DIFF
--- a/aws/resource_aws_spot_fleet_request_test.go
+++ b/aws/resource_aws_spot_fleet_request_test.go
@@ -312,6 +312,31 @@ func TestAccAWSSpotFleetRequest_overriddingSpotPrice(t *testing.T) {
 	})
 }
 
+func TestAccAWSSpotFleetRequest_withoutSpotPrice(t *testing.T) {
+	var sfr ec2.SpotFleetRequestConfig
+	rName := acctest.RandString(10)
+	rInt := acctest.RandInt()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSpotFleetRequestDestroy,
+		Steps: []resource.TestStep{
+			{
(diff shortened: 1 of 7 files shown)


aws/resource_aws_spot_fleet_request_test.go
```diff
@@ -1511,6 +1511,87 @@ resource "aws_spot_fleet_request" "foo" {
 `, rName, rInt, rInt, rName)
 }
 
+func testAccAWSSpotFleetRequestConfigWithoutSpotPrice(rName string, rInt int) string {
```
bflad: This PR is in great shape -- only item is that this new test configuration does not look like its being called. If you fix that up then we should be able to get this in 👍

YakDriver:
@bflad I added the call and squashed the commits. (But, I find I get 100% pass on tests that are never called. 👍 )

bflad:
> I find I get 100% pass on tests that are never called. 👍 

You should get this made into a poster, I would buy it! 😄

YakDriver:
@bflad This is phenomenal! Thanks for your help and merge. When I get those t-shirts made, you'll get the first.

bflad:

PR #4436, opened 2018-05-03
Add account ID to the SSM document ARN

tomelliff:
Without the account ID being specified if the ARN output is used as an allowed resource in an IAM policy then the IAM policy doesn't match the document and you get a permissions error from IAM when executing the document.

### DIFF
--- a/aws/resource_aws_ssm_document_test.go
+++ b/aws/resource_aws_ssm_document_test.go
@@ -20,7 +20,7 @@ func TestAccAWSSSMDocument_basic(t *testing.T) {
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckAWSSSMDocumentDestroy,
 		Steps: []resource.TestStep{
-			resource.TestStep{
+			{
 				Config: testAccAWSSSMDocumentBasicConfig(name),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAWSSSMDocumentExists("aws_ssm_document.foo"),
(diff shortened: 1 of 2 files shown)


tomelliff:
Not a clue why `make vendor-status` is failing in Travis there. Works locally and I've not touched anything with `govendor` so shouldn't be an issue.

aws/resource_aws_ssm_document_test.go
```diff
@@ -19,11 +20,13 @@ func TestAccAWSSSMDocument_basic(t *testing.T) {
...
 		CheckDestroy: testAccCheckAWSSSMDocumentDestroy,
 		Steps: []resource.TestStep{
-			resource.TestStep{
+			{
 				Config: testAccAWSSSMDocumentBasicConfig(name),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAWSSSMDocumentExists("aws_ssm_document.foo"),
 					resource.TestCheckResourceAttr("aws_ssm_document.foo", "document_format", "JSON"),
+					resource.TestMatchResourceAttr("aws_ssm_document.foo", "arn",
+						regexp.MustCompile(`^arn:aws:ssm:[a-z]{2}-[a-z]+-\d{1}:\d{12}:document/.*$`)),
```
bflad: Thanks for this very accurate test check for the Commercial partition -- we'll be making these more generic so we can also support testing in GovCloud (US) and China at some point in the near future:

```
regexp.MustCompile(`^arn:[^:]+:ssm:[^:]+:\d{12}:document/.*$`))
```

tomelliff: Ha, I originally had the regex match gov-cloud as well and figured it was a waste of time because we didn't run the tests in gov-cloud/china and so took it out to simplify the regex. Didn't think about matching all non colon characters for the region and partition though.

I'll keep that in mind for anything else I add in future.

tomelliff:
@bflad thanks for the review and merge again :)

bflad:

PR #4454, opened 2018-05-04
r/aws_security_group_rule: prevent flapping of account prefixed references

### LINKED ISSUE #141: Terraform EC2 security group ingress rule flapping
_This issue was originally opened by @anosulchik as hashicorp/terraform#6135. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,

When I create security group ingress rule that opens access from SG in another aws account, `terraform apply` always recreates this rule even if no changes in terraform config files were made:

```
-/+ aws_security_group_rule.apigw_access
    from_port:                "443" => "443"
    protocol:                 "tcp" => "tcp"
    security_group_id:        "sg-ba999ac2" => "sg-ba999ac2"
    self:                     "false" => "0"
    source_security_group_id: "sg-8c777ceb" => "373682980000/sg-8c777ceb" (forces new resource)
    to_port:                  "443" => "443"
    type:                     "ingress" => "ingress"
```

SG ingress rule config:

```
resource "aws_security_group_rule" "apigw_access" {
  type = "ingress"
  from_port = 443
  to_port = 443
  protocol = "tcp"
  security_group_id = "${aws_security_group.ecs_security_group.id}"
  source_security_group_id = "373682980000/sg-8c777ceb"
}
```

Please advise. Thank you.

P.S. terraform 0.6.14

appilon:
Fix #141 

This PR could be handled a few ways and I've made some changes we might want to omit from the PR. The perpetual diff was caused by 2 spots in the code I will highlight, the issue in the wild is made worse by the fact that the attribute has `ForceNew:true` causing so many destroys/recreates. Please see my self-made PR comments to better understand the changes/gotchas.

UPDATE: currently the change from `p` to `rule` breaks the acceptance tests, I will investigate and see if its worth trying to fix or if I should revert that change. Please see comments below

ANOTHER UPDATE: the tests that fail when I change `p` to `rule` don't make a whole lot of sense why... even worse when I ran 1 of the failing tests individually it passed so.... yeah. My gut feel is that this resource might be a little cruftier than others, maybe for the sake of fixing #141 I just put in the DiffSupressor and call it a day but there are parts about this resource that feel wrong as mentioned in my many comments. Any feedback or suggestions would be great.

FINAL UPDATE: couldn't figure out what goes wrong with the state when I make the change described above, but to keep things simple, leaving out that change all tests pass and the flapping is suppressed.
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSSecurityGroupRule -timeout 120m
=== RUN   TestAccAWSSecurityGroupRule_Ingress_VPC
--- PASS: TestAccAWSSecurityGroupRule_Ingress_VPC (23.46s)
=== RUN   TestAccAWSSecurityGroupRule_Ingress_Protocol
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Protocol (45.91s)
=== RUN   TestAccAWSSecurityGroupRule_Ingress_Ipv6
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Ipv6 (44.74s)
=== RUN   TestAccAWSSecurityGroupRule_Ingress_Classic
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Classic (16.57s)
=== RUN   TestAccAWSSecurityGroupRule_MultiIngress
--- PASS: TestAccAWSSecurityGroupRule_MultiIngress (26.67s)
=== RUN   TestAccAWSSecurityGroupRule_Egress
--- PASS: TestAccAWSSecurityGroupRule_Egress (23.87s)
=== RUN   TestAccAWSSecurityGroupRule_SelfReference
--- PASS: TestAccAWSSecurityGroupRule_SelfReference (42.08s)
=== RUN   TestAccAWSSecurityGroupRule_ExpectInvalidTypeError
--- PASS: TestAccAWSSecurityGroupRule_ExpectInvalidTypeError (1.13s)
=== RUN   TestAccAWSSecurityGroupRule_ExpectInvalidCIDR
--- PASS: TestAccAWSSecurityGroupRule_ExpectInvalidCIDR (1.66s)
=== RUN   TestAccAWSSecurityGroupRule_PartialMatching_basic
--- PASS: TestAccAWSSecurityGroupRule_PartialMatching_basic (45.59s)
=== RUN   TestAccAWSSecurityGroupRule_PartialMatching_Source
--- PASS: TestAccAWSSecurityGroupRule_PartialMatching_Source (47.32s)
=== RUN   TestAccAWSSecurityGroupRule_Issue5310
--- PASS: TestAccAWSSecurityGroupRule_Issue5310 (15.61s)
=== RUN   TestAccAWSSecurityGroupRule_Race
--- PASS: TestAccAWSSecurityGroupRule_Race (387.14s)
=== RUN   TestAccAWSSecurityGroupRule_SelfSource
--- PASS: TestAccAWSSecurityGroupRule_SelfSource (45.57s)
=== RUN   TestAccAWSSecurityGroupRule_PrefixListEgress
--- PASS: TestAccAWSSecurityGroupRule_PrefixListEgress (57.43s)
=== RUN   TestAccAWSSecurityGroupRule_IngressDescription
--- PASS: TestAccAWSSecurityGroupRule_IngressDescription (24.29s)
=== RUN   TestAccAWSSecurityGroupRule_EgressDescription
--- PASS: TestAccAWSSecurityGroupRule_EgressDescription (23.57s)
=== RUN   TestAccAWSSecurityGroupRule_IngressDescription_updates
--- PASS: TestAccAWSSecurityGroupRule_IngressDescription_updates (36.76s)
=== RUN   TestAccAWSSecurityGroupRule_EgressDescription_updates
--- PASS: TestAccAWSSecurityGroupRule_EgressDescription_updates (34.52s)
=== RUN   TestAccAWSSecurityGroupRule_MultiDescription
--- PASS: TestAccAWSSecurityGroupRule_MultiDescription (93.02s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1036.944s
```

### DIFF
--- a/aws/resource_aws_security_group_rule.go
+++ b/aws/resource_aws_security_group_rule.go
@@ -298,7 +298,7 @@ func resourceAwsSecurityGroupRuleRead(d *schema.ResourceData, meta interface{}) 
 	log.Printf("[DEBUG] Found rule for Security Group Rule (%s): %s", d.Id(), rule)
 
 	d.Set("type", ruleType)
-	if err := setFromIPPerm(d, sg, rule); err != nil {
+	if err := setFromIPPerm(d, sg, p); err != nil {
 		return errwrap.Wrapf("Error setting IP Permission for Security Group Rule: {{err}}", err)
 	}
 
--- a/aws/resource_aws_security_group_rule.go
+++ b/aws/resource_aws_security_group_rule.go
@@ -98,6 +98,22 @@ func resourceAwsSecurityGroupRule() *schema.Resource {
 				ForceNew:      true,
 				Computed:      true,
 				ConflictsWith: []string{"cidr_blocks", "self"},
+				// The following documentation https://docs.aws.amazon.com/AmazonVPC/latest/PeeringGuide/vpc-peering-security-groups.html
+				// demonstrates cross account SG referencing only by ID, implying all security group IDs are unique
+				// therefore the account prefix should not be needed for a diff
+				// users can then supply 123456/sg-1234 or just sg-1234 and the diff will be suppressed
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					oldSG := old
+					if items := strings.Split(old, "/"); len(items) > 1 {
+						oldSG = items[1]
+					}
+					newSG := new
+					if items := strings.Split(new, "/"); len(items) > 1 {
+						newSG = items[1]
+					}
+
+					return newSG == oldSG
+				},
 			},
 
 			"self": {
@@ -282,7 +298,7 @@ func resourceAwsSecurityGroupRuleRead(d *schema.ResourceData, meta interface{}) 
 	log.Printf("[DEBUG] Found rule for Security Group Rule (%s): %s", d.Id(), rule)
 
 	d.Set("type", ruleType)
-	if err := setFromIPPerm(d, sg, p); err != nil {
+	if err := setFromIPPerm(d, sg, rule); err != nil {
 		return errwrap.Wrapf("Error setting IP Permission for Security Group Rule: {{err}}", err)
 	}
 
@@ -689,10 +705,15 @@ func setFromIPPerm(d *schema.ResourceData, sg *ec2.SecurityGroup, rule *ec2.IpPe
 	if len(rule.UserIdGroupPairs) > 0 {
 		s := rule.UserIdGroupPairs[0]
 
+		accountPrefix := aws.StringValue(s.UserId)
+		if accountPrefix != "" {
+			accountPrefix += "/"
+		}
+
 		if isVPC {
-			d.Set("source_security_group_id", *s.GroupId)
+			d.Set("source_security_group_id", accountPrefix+*s.GroupId)
 		} else {
-			d.Set("source_security_group_id", *s.GroupName)
+			d.Set("source_security_group_id", accountPrefix+*s.GroupName)
 		}
 	}
 


aws/resource_aws_security_group_rule.go
```diff
@@ -98,6 +98,22 @@ func resourceAwsSecurityGroupRule() *schema.Resource {
 				ForceNew:      true,
 				Computed:      true,
 				ConflictsWith: []string{"cidr_blocks", "self"},
+				// The following documentation https://docs.aws.amazon.com/AmazonVPC/latest/PeeringGuide/vpc-peering-security-groups.html
```
appilon: This suppression function is built on the premise that all security groups have a unique ID. I couldn't find any explicit piece of documentation online that this was true, however the documentation for making cross account SG references does not prefix the reference with an account number... which to me means AWS ensures uniqueness of SG ids else things would be super broken/collisions galore in the wild.

aws/resource_aws_security_group_rule.go
```diff
@@ -282,7 +298,7 @@ func resourceAwsSecurityGroupRuleRead(d *schema.ResourceData, meta interface{})
 	log.Printf("[DEBUG] Found rule for Security Group Rule (%s): %s", d.Id(), rule)
 
 	d.Set("type", ruleType)
-	if err := setFromIPPerm(d, sg, p); err != nil {
+	if err := setFromIPPerm(d, sg, rule); err != nil {
```
appilon: This change is very subtle, but after reading the code up and down, I realized the current implementation actually would build an struct from the local state: `p` and then try and match all its parameters with the list of remotely fetched rules `rules` with function `findRuleMatch` returning `rule`. We then set the local state from `p`.... which is very strange because we build `p` from the local state... furthermore I feel its more "right" to set the local state from the remotely fetched data during the `Read` operation

aws/resource_aws_security_group_rule.go
```diff
@@ -689,10 +705,15 @@ func setFromIPPerm(d *schema.ResourceData, sg *ec2.SecurityGroup, rule *ec2.IpPe
 	if len(rule.UserIdGroupPairs) > 0 {
 		s := rule.UserIdGroupPairs[0]
 
+		accountPrefix := aws.StringValue(s.UserId)
```
appilon: Now that we are setting local state from the remotely fetched rule, we should populate the local state with the full data `1234567/sg-12345`. What I realize though is with the `DiffSuppressFunc` this code becomes inert really.... we regardless of what is stored we will only diff the sg-ids... to me it feels more correct to include the information though

appilon:
One final comment which I can't place on the code (because I didn't change anything). Is that the `Create` function does not `return Read` which is unusual from what I've seen. This will actual cause some drift in terms of localState vs what AWS would return, fortunately the `DiffSuppressFunc` and the fact that `plan/apply` refresh local state with a `Read` means the drift is typically corrected right away.

The circumstances are the following: on create the user has specified a security group without account prefix (which is very common, more common than with the prefix)
```
source_security_group_id = "sg-12345"
```
Which since we do not perform a `Read` on `Create` at the comment the state is exactly as the user has configured. Any subsequent `Read` operation would set the local state to `12345678/sg-12345`. Like i said there is no real problem because of the SuppressFunc but I thought I would document all the side effects at play here.

aws/resource_aws_security_group_rule.go
appilon: This is where the flapping was caused, because previously we saved to local state only the SG ID. So for users who specified `1234567/sg-1245` in their config, TF thought the field was constantly different --> `1234567/sg-1245 != sg-1245`. Ultimately though once I added a SupressFunc as mentioned none of this particular change matters

aws/resource_aws_security_group_rule.go
appilon: This change actually breaks the acceptance tests, I might take time to investigate which ones, if it becomes a mess I can confirm reverting the change fixes the tests, but as mentioned above something about this code feels very wrong.

bflad:
Hey @appilon 👋 Apologies for the long silence here and thank you for giving this a go awhile ago. The EC2 API and the Security Group resources are a super hassle, especially with canonicalization issues like this.

After reviewing the related https://github.com/terraform-providers/terraform-provider-aws/pull/11809, it brought some further reflection on the implications of introducing the "always add the value, but ignore differences" method here. Terraform 0.12 is super sensitive to attribute values being changed between the planned value (configuration) and applied value (`d.Set()`), returning an `provider produced inconsistent results between plan and apply` type error. We would need to guess the account ID if it was not provided and set the new difference during `CustomizeDiff`, which makes this implementation quite difficult or impossible to properly implement.

We are going to try the solution in #11809 and I'm going to close this for now. Should that solution be problematic, we can see if we there are good ways to move forward with this. Thanks again.

PR #4462, opened 2018-05-06
Add import instructions to default VPC

tdmalone:
Almost identical to [the way non-default VPCs work](https://www.terraform.io/docs/providers/aws/r/vpc.html#import), but documented here too for completeness.

I checked first that this works with the latest version of Terraform and the AWS provider.


ewbankkit:
See also https://github.com/terraform-providers/terraform-provider-aws/pull/4382/files#r186301001.

PR #4466, opened 2018-05-07
Refactor s3 acceptance tests to run in multiple partitions.

jmcarp:

* Drop hard-coded regions in s3 acceptance tests
* Ditto for hard-coded partitions
* Skip multi-region tests in govcloud, since govcloud only has one region


aws/resource_aws_s3_bucket_test.go
```diff
@@ -109,6 +109,11 @@ func TestAccAWSS3Bucket_generatedName(t *testing.T) {
 
 func TestAccAWSS3Bucket_region(t *testing.T) {
 	rInt := acctest.RandInt()
+	partition := testAccGetPartition()
+
+	if partition == "aws-us-gov" {
+		t.Skip("skipping replication tests; govcloud only includes a single region")
```
bflad: We should probably implement this as a `PreCheck` function (`testAccMultipleRegionsPreCheck(t)`?) rather than duplicating the logic multiple places. This is especially relevant as it will pop up in other services where we have replication testing as well, like RDS.

aws/resource_aws_s3_bucket_notification_test.go
```diff
@@ -364,7 +366,7 @@ resource "aws_sns_topic" "topic" {
 		"Effect": "Allow",
 		"Principal": {"AWS":"*"},
 		"Action": "SNS:Publish",
-		"Resource": "arn:aws:sns:*:*:%s",
+		"Resource": "arn:%s:sns:*:*:%s",
```
bflad: Rather than requiring this new `partition` string to be passed around as function inputs, which gets messy, I think we should just add a consistently named `aws_partition` data source anywhere its required, e.g. `${data.aws_partition.current.partition}`. This applies to all the testing configurations that have these manually constructed ARNs. 👍

aws/resource_aws_s3_bucket_test.go
bflad: In fact further down the road, coupled with a new `aws_regions` data source that pulls regions from the `endpoints` SDK package, we could have these automatically start running if/when the `endpoints` SDK package gets updated with additional regions in the GovCloud partition. 😉

aws/resource_aws_s3_bucket_notification_test.go
jmcarp: Good call--I'll use the `aws_partition` data source where I can. There are a few places where it looks like we'll need to look up partition directly, like in https://github.com/terraform-providers/terraform-provider-aws/pull/4466/files/e80abe873bb68f23cce3ae545e44db2f6687ad57#diff-2f90975e6a0864163009708350852a9aR19, where we're making assertions about a bucket policy and need to build the expected document.

aws/resource_aws_s3_bucket_test.go
```diff
@@ -1484,15 +1517,8 @@ EOF
 
 func testAccAWSS3BucketConfigWithAcceleration(randInt int) string {
 	return fmt.Sprintf(`
-provider "aws" {
-	alias = "west"
-	region = "eu-west-1"
```
bflad: It looks like `testAccAWSS3BucketConfigWithoutAcceleration` needs to be updated as well

```
=== RUN   TestAccAWSS3Bucket_acceleration
--- FAIL: TestAccAWSS3Bucket_acceleration (243.29s)
    testing.go:518: Step 1 error: Error refreshing: 1 error(s) occurred:
        
        * aws_s3_bucket.bucket: 1 error(s) occurred:
        
        * aws_s3_bucket.bucket: aws_s3_bucket.bucket: error reading S3 bucket "tf-test-bucket-9115369584085629874": BucketRegionError: incorrect region, the bucket is not in 'eu-west-1' region
```

jmcarp:
Updated. FWIW, the bucket acceleration tests are failing for me on master and on this branch, so not sure what's going on there.

PR #4471, opened 2018-05-07
resource/aws_sns_platform_application: documentation - fix inconsistent key name

Druotic:
Got this error while trying to plan:

```
Error: aws_sns_platform_application.my_platform_application: : invalid or unknown key: event_endpoint_updated_topic
```

Extremely simple fix to documentation for the following key name: `event_endpoint_updated_topic` -> `event_endpoint_updated_topic_arn`. See key here: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_sns_platform_application.go#L28

The other keys have `_arn` as expected. Only this one was incorrectly documented.


Druotic:
@bflad Np! Glad I could help.

PR #4473, opened 2018-05-07
Add aws_codebuild_webhook resource for creating GitHub webhooks for CodeBuild projects

joestump:

Introduces a resource called `aws_codebuild_webhook` that [creates GitHub webhook](https://docs.aws.amazon.com/codebuild/latest/APIReference/API_CreateWebhook.html#CodeBuild-CreateWebhook-request-branchFilter) for a CodeBuild project that has its source stored on GitHub.

This works builds on #2814 and adds branch filtering.

```
$ make testacc TESTARGS='-run=TestAccAwsCodeBuildWebhook_basic'
TF_ACC=1 go test ./... -v -run=TestAccAwsCodeBuildWebhook_basic -timeout 120m
=== RUN   TestAccAwsCodeBuildWebhook_basic
--- FAIL: TestAccAwsCodeBuildWebhook_basic (41.83s)
		
		* aws_codebuild_webhook.test: 1 error(s) occurred:
		
		* aws_codebuild_webhook.test: ResourceNotFoundException: Could not find access token for server type github
			status code: 400, request id: fe4a925b-523e-11e8-8f47-a9109ac40b1a
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	41.863s
make: *** [testacc] Error 1
```

The reason because this fails is because:

* The AWS user Terraform runs as must have already done an OAuth login with GitHub on an existing CodeBuild project.
* The GitHub user that the AWS user logs in as must have access to the GitHub repository listed in the `source`. 

I've tried to test this, but passing `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to tests appears to be ignored. Creating a webhook with the credentials I created works via the CLI, but not via the API call this resource makes.

How does one use their own AWS credentials when running `make test`?

**NOTE:** Once this passes, the upstream Terraform test environment will need to address this in order to run these tests. Alternatively, I could implement something like the [Heroku provider has](https://github.com/terraform-providers/terraform-provider-heroku/blob/master/heroku/resource_heroku_app_test.go#L214-L222) for skipping tests.

### DIFF
--- a/aws/resource_aws_codebuild_project_test.go
+++ b/aws/resource_aws_codebuild_project_test.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"os"
 	"regexp"
 	"strings"
 	"testing"
@@ -290,6 +291,18 @@ func testAccCheckAWSCodeBuildProjectDestroy(s *terraform.State) error {
 }
 
 func testAccAWSCodeBuildProjectConfig_basic(rName, vpcConfig, vpcResources string) string {
+	// This is used for testing aws_codebuild_webhook as well as aws_codebuild_project.
+	// In order for that resource to work the Terraform AWS user must have done a GitHub
+	// OAuth dance.
+	//
+	// Additionally, the GitHub user that the Terraform AWS user logs in as must have
+	// access to the GitHub repository. This allows others to run tests for the webhook
+	// without having to have access to the Packer GitHub repository.
+	sourceURL, ok := os.LookupEnv("CODEBUILD_GITHUB_SOURCE_URL")
+	if !ok {
+		sourceURL = "https://github.com/hashicorp/packer.git"
+	}
+
 	return fmt.Sprintf(`
 resource "aws_iam_role" "codebuild_role" {
   name = "codebuild-role-%s"
@@ -375,7 +388,7 @@ resource "aws_codebuild_project" "foo" {
 
   source {
     type     = "GITHUB"
-    location = "https://github.com/hashicorp/packer.git"
+    location = "%s"
   }
 
   tags {
@@ -384,7 +397,7 @@ resource "aws_codebuild_project" "foo" {
 	%s
 }
 %s
-`, rName, rName, rName, rName, vpcConfig, vpcResources)
+`, rName, rName, rName, rName, sourceURL, vpcConfig, vpcResources)
 }
 
 func testAccAWSCodeBuildProjectConfig_basicUpdated(rName string) string {
--- a/aws/resource_aws_codebuild_webhook.go
+++ b/aws/resource_aws_codebuild_webhook.go
@@ -11,6 +11,7 @@ func resourceAwsCodeBuildWebhook() *schema.Resource {
 		Create: resourceAwsCodeBuildWebhookCreate,
 		Read:   resourceAwsCodeBuildWebhookRead,
 		Delete: resourceAwsCodeBuildWebhookDelete,
+		Update: resourceAwsCodeBuildWebhookUpdate,
 
 		Schema: map[string]*schema.Schema{
 			"name": {
@@ -66,7 +67,28 @@ func resourceAwsCodeBuildWebhookRead(d *schema.ResourceData, meta interface{}) e
 	}
 
 	project := resp.Projects[0]
+	d.Set("branch_filter", project.Webhook.BranchFilter)
 	d.Set("url", project.Webhook.Url)
+	return nil
+}
+
+func resourceAwsCodeBuildWebhookUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codebuildconn
+
+	if !d.HasChange("branch_filter") {
+		return nil
+	}
+
+	_, err := conn.UpdateWebhook(&codebuild.UpdateWebhookInput{
+		ProjectName:  aws.String(d.Id()),
+		BranchFilter: aws.String(d.Get("branch_filter").(string)),
+		RotateSecret: aws.Bool(false),
+	})
+
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/aws/resource_aws_codebuild_webhook.go
+++ b/aws/resource_aws_codebuild_webhook.go
@@ -18,6 +18,10 @@ func resourceAwsCodeBuildWebhook() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"branch_filter": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 			"url": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -30,13 +34,15 @@ func resourceAwsCodeBuildWebhookCreate(d *schema.ResourceData, meta interface{})
 	conn := meta.(*AWSClient).codebuildconn
 
 	resp, err := conn.CreateWebhook(&codebuild.CreateWebhookInput{
-		ProjectName: aws.String(d.Get("name").(string)),
+		ProjectName:  aws.String(d.Get("name").(string)),
+		BranchFilter: aws.String(d.Get("branch_filter").(string)),
 	})
 	if err != nil {
 		return err
 	}
 
 	d.SetId(d.Get("name").(string))
+	d.Set("branch_filter", d.Get("branch_filter").(string))
 	d.Set("url", resp.Webhook.Url)
 	return nil
 }
--- a/aws/resource_aws_codebuild_webhook_test.go
+++ b/aws/resource_aws_codebuild_webhook_test.go
@@ -70,7 +70,7 @@ func testAccCheckAwsCodeBuildWebhookExists(name string) resource.TestCheckFunc {
 }
 
 func testAccCodeBuildWebhookConfig_basic(rName string) string {
-	return fmt.Sprintf(testAccAWSCodeBuildProjectConfig_basic(rName) + `
+	return fmt.Sprintf(testAccAWSCodeBuildProjectConfig_basic(rName, "", "") + `
 resource "aws_codebuild_webhook" "test" {
   name = "${aws_codebuild_project.foo.name}"
 }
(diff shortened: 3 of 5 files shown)


gthole:
Not sure this will help much without tracking the `secret` and `payload_url` returned from AWS and making them available as attributes.

For example, this is the syntax I'd expect:

```
resource "aws_codebuild_webhook" "my_webhook" {
    name = "${aws_codebuild_project.some_project.name}"
    branch_filter = "master"
}

resource "github_repository_webhook" "my_webhook" {
    name = "some_project_deploy"

    configuration {
        url = "${aws_codebuild_webhook.my_webhook.payload_url}"
        secret = "${aws_codebuld_webhook.my_webhook.secret}"
        content_type = "json"
        insecure_ssl = false
    }

    active = true
    events = ["push"]
}
```

Unfortunately, it doesn't look like the `secret` is ever returned from `batchGetProjects`, despite what the documentation says.  (Otherwise I'd just create a provisioner and external data source to work around this.)

So if the secret can be stored in the state, that'd be ideal.

Awesome to see progress on this - it'd be a huge win for our devops stack.

bflad:
I'll be taking a look at this tomorrow. 👍

aws/resource_aws_codebuild_project_test.go
```diff
@@ -290,6 +291,18 @@ func testAccCheckAWSCodeBuildProjectDestroy(s *terraform.State) error {
...
 
 func testAccAWSCodeBuildProjectConfig_basic(rName, vpcConfig, vpcResources string) string {
+	// This is used for testing aws_codebuild_webhook as well as aws_codebuild_project.
+	// In order for that resource to work the Terraform AWS user must have done a GitHub
+	// OAuth dance.
+	//
+	// Additionally, the GitHub user that the Terraform AWS user logs in as must have
+	// access to the GitHub repository. This allows others to run tests for the webhook
+	// without having to have access to the Packer GitHub repository.
+	sourceURL, ok := os.LookupEnv("CODEBUILD_GITHUB_SOURCE_URL")
```
bflad: This is a great idea -- I'll make sure this gets ported over in some fashion when I rebase this PR.

aws/resource_aws_codebuild_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+
+func resourceAwsCodeBuildWebhook() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCodeBuildWebhookCreate,
+		Read:   resourceAwsCodeBuildWebhookRead,
+		Delete: resourceAwsCodeBuildWebhookDelete,
+		Update: resourceAwsCodeBuildWebhookUpdate,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
```
bflad: This might seem pedantic for now, but on merge will rename this attribute to `project_name` to better align with the API and help operators understand the association with projects.

aws/resource_aws_codebuild_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+	resp, err := conn.CreateWebhook(&codebuild.CreateWebhookInput{
+		ProjectName:  aws.String(d.Get("name").(string)),
+		BranchFilter: aws.String(d.Get("branch_filter").(string)),
+	})
+	if err != nil {
+		return err
+	}
+
+	d.SetId(d.Get("name").(string))
+	d.Set("branch_filter", d.Get("branch_filter").(string))
```
bflad: `d.Set()` should only be called in the read function 😄 -- will fix this on merge.

aws/resource_aws_codebuild_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+	conn := meta.(*AWSClient).codebuildconn
+
+	resp, err := conn.BatchGetProjects(&codebuild.BatchGetProjectsInput{
+		Names: []*string{
+			aws.String(d.Id()),
+		},
+	})
+
+	if err != nil {
+		return err
```
bflad: We should check for "resource not found" type errors and `d.SetId("")` similar to how the project resource does this. Will fix on merge. 👍

aws/resource_aws_codebuild_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	if len(resp.Projects) == 0 {
+		d.SetId("")
```
bflad: When we're removing a resource from the state, we should log an appropriate warning message in the logs, e.g.

```go
log.Printf("[WARN] CodeBuild Project %q not found, removing from state", d.Id())
```

aws/resource_aws_codebuild_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+func resourceAwsCodeBuildWebhookDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codebuildconn
+
+	_, err := conn.DeleteWebhook(&codebuild.DeleteWebhookInput{
+		ProjectName: aws.String(d.Id()),
+	})
+
+	if err != nil {
+		if isAWSErr(err, codebuild.ErrCodeResourceNotFoundException, "") {
+			d.SetId("")
```
bflad: `d.SetId("")` is extraneous in delete functions and can be removed. 👍

aws/resource_aws_codebuild_webhook_test.go
```diff
@@ -0,0 +1,78 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsCodeBuildWebhookDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCodeBuildWebhookConfig_basic(acctest.RandString(5)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsCodeBuildWebhookExists("aws_codebuild_webhook.test"),
+					resource.TestCheckResourceAttrSet("aws_codebuild_webhook.test", "url"),
```
bflad: We should probably be a little more cautious and at least partially validate the value as `resource.TestCheckResourceAttrSet()` will pass with empty strings and I believe even `d.Set("XXX", nil)`

e.g.

```go
resource.TestMatchResourceAttr("aws_codebuild_webhook.test", "url", regexp.MustCompile(`^https://`))
```

aws/resource_aws_codebuild_webhook_test.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		resp, err := conn.BatchGetProjects(&codebuild.BatchGetProjectsInput{
+			Names: []*string{
+				aws.String(rs.Primary.ID),
+			},
+		})
+
+		if err != nil {
+			return err
```
bflad: A resource not found exception here would be a good thing and should `return nil` -- will fix on merge.

aws/resource_aws_codebuild_webhook_test.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func testAccCheckAwsCodeBuildWebhookExists(name string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		_, ok := s.RootModule().Resources[name]
+		if !ok {
+			return fmt.Errorf("Not found: %s", name)
+		}
+
```
bflad: Similar to how the destroy acceptance test function above is checking against the API, we should do the same thing here to ensure the "physical" webhook actually exists.

website/docs/r/codebuild_webhook.html.markdown
```diff
@@ -0,0 +1,35 @@
...
+sidebar_current: "docs-aws-resource-codebuild-webhook"
+description: |-
+  Provides a CodeBuild Webhook resource.
+---
+
+# aws_codebuild_webhook
+
+Provides a CodeBuild Webhook resource.
+
+~> **Note:** The AWS account that Terraform uses to create this resource *must* have authorized CodeBuild to access GitHub's OAuth API. This is a manual step that must be done *before* creating webhooks with this resource. If OAuth is not configured, AWS will return an error similar to `ResourceNotFoundException: Could not find access token for server type github`.
```
bflad: This manual step must also be done in every applicable region where the resource will be used. We should probably also point to the AWS documentation surrounding project webhooks here. Will update on merge.

aws/resource_aws_codebuild_webhook.go
```diff
@@ -0,0 +1,112 @@
+package aws
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/codebuild"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCodeBuildWebhook() *schema.Resource {
```
bflad: We can easily add import support for this resource as the ID is already handled properly in the read function. Will add the passthrough importer to this resource and document how to import on merge.

bflad:
> Not sure this will help much without tracking the secret and payload_url returned from AWS and making them available as attributes.

AWS is using "your" GitHub OAuth credentials to manage the actual GitHub webhook (physically creating, updating, and deleting on the GitHub side with this resource). I'm not sure they expect you to be managing the webhook outside of their workflow.

bflad:
The new `aws_codebuild_webhook` resource will be available in version 1.21.0 of the AWS provider, likely releasing middle of next week. 🎉

joestump:
@bflad thanks for the notes and merge! 🎉

bflad:

salvianreynaldi:
I think the [documentation](https://www.terraform.io/docs/providers/aws/r/codebuild_webhook.html#argument-reference) needs to be updated. It still list only `name` and `branch_filter` as the supported arguments, while the [code](https://github.com/terraform-providers/terraform-provider-aws/blob/21cc00c044bcb6fc2f6a9a6d08d588d7719f1a14/aws/resource_aws_codebuild_webhook.go#L11) has more than that (and `name` should be `project_name` too).

PR #4480, opened 2018-05-08
Data source does not fail if hosted zone not found.

genevieve:
We are hoping to use the aws route53 hosted zone data source conditionally on the presence of a requested name. Given the zone does not exist, then create the resource. While this works when the zone does exist, it fails when it does not with `no matching Route53Zone found`. We hoped to use the data sources similarly across the IaaSes after being able to do this conditional logic with the (azure dns zone](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/data_source_dns_zone.go#L53-L56).

* Return nil instead of failing when the hosted zone name is not found.

```
--- PASS: TestAccDataSourceAwsRoute53Zone (78.70s)
PASS
```

### DIFF
--- a/aws/data_source_aws_route53_zone.go
+++ b/aws/data_source_aws_route53_zone.go
@@ -158,7 +158,7 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 	if hostedZoneFound == nil {
-		return fmt.Errorf("no matching Route53Zone found")
+		return nil
 	}
 
 	idHostedZone := cleanZoneID(*hostedZoneFound.Id)


bflad:
Thanks for submitting this @genevieve! While this is a simple code change, I think this may warrant a larger discussion about its implications with this and other data sources.

Most data sources (at least in the AWS provider) are currently designed to return an immediate error when they cannot read their respective resource. The AzureRM provider seems to be a mixed bag mostly leaning towards returning no error. The Google provider seems to be a mixed bag mostly leaning towards returning the error. Ideally all the data sources across all "official" providers should handle this similarly rather than each one having its own behavior.

I could see this as an argument to allow some sort of resource `lifecycle` configuration in Terraform core to allow "adopting" resources if they already exist (if implemented in the resource), rather than forcing everyone to write custom conditionals or worry about the return value of an associated data source.

So that said, I'm not sure what we should do here. 😅 I'm not opposed to the idea, but I think we should be consistent. Preferably, we would also handle this more holistically, whether making this change to all the data sources in the provider or saving this for a (potentially more appropriate) implementation in Terraform core. I'd love for some other maintainers to chime in on this manner with their thoughts.

bflad:
Hi again @genevieve -- it looks like we are going to lean towards returning an error when data source cannot find its associated physical resource for now unless some generic support is added for "conditional" data sources. The Azure data sources should now follow that behavior after the next release. Any "official" (terraform-providers GitHub organization) provider data sources that are not doing this correctly should be considered a bug. 👍 

Rather than leaving this PR lingering when its not likely to get merged (currently), I'm going to opt to close it for now. We can always revisit this topic in the future should there be good reason to do this holistically for all the data sources!

If you are interested in having better "create resource X if it does not exist" or "conditional" data source support, its probably worth creating an upstream issue in [Terraform core](https://github.com/hashicorp/terraform/) -- that team/community might have better ideas or future enhancements surrounding this topic. Thanks again for bringing this up though!

bflad:
For future travelers to this pull request, there is an open Terraform core issue discussing at least parts of this: https://github.com/hashicorp/terraform/issues/16380

PR #4484, opened 2018-05-08
New resource - Glue crawlers

### LINKED ISSUE #3875: Feature Request: Resource for managing AWS Glue Crawlers
Split feature request from #1416. Please note there are relevant comments in #1416 about this feature request. For issues not immediately being worked on, please use :+1: upvotes on this original issue comment to help guage community interest.

### Terraform Version

```
terraform 0.10+
terraform-provider-aws (new feature)
```

### Affected Resource(s)

- `aws_glue_crawler`

```hcl
# New Resource - not implemented
resource "aws_glue_crawler" "example" {
  # ...
}
```

### Expected Behavior

Create, read, update, and delete support for AWS Glue crawlers.

### Actual Behavior

Resource not implemented.

### References

* https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-crawler-crawling.html
* https://docs.aws.amazon.com/sdk-for-go/api/service/glue/#Glue.CreateCrawler

darrenhaken:
Closes #3875 

* Add support for a new resource, Glue Crawlers.

```
$ make testacc TESTARGS='-run=TestAccAWSGlueCrawler'

```


darrenhaken:
@bflad I am still having issues having this work for me.

If you run the test do you also see failures around IAM and assuming the role? 
I have been stumped by this and it has meant I haven't really come back to the PR to carry on with it.

Could you help fix this with me and I'll carry on with the rest of the PR?

darrenhaken:
@bflad still not heard back, do you think it's something you can look at?

We're using the Crawlers quite a lot now and getting it integrated into TF would be fantastic.

darrenhaken:
Looks like it's working @f0rk and I've included a few more of the fields to the crawler.

jordanfowler:
How's this looking? We'd love to get this into our terraform stack.

jordanfowler:
@mabushey 🌮

darrenhaken:
Hey all, having an issue with JDBC crawlers and IAM permissions. 

Running `TestAccAWSGlueCrawler_jdbcCrawler` causes the following error:

```
* aws_glue_catalog_crawler.test: error creating Glue crawler: InvalidInputException: Unable to retrieve connection tf-connection: User: arn:aws:sts::{account_id}:assumed-role/tf-glue-service-role/AWS-Crawler is not authorized to perform: glue:GetConnection on resource: * (Service: AmazonDataCatalog; Status Code: 400; Error Code: AccessDeniedException; Request ID: a7364009-7215-11e8-82ef-bd3deae1d88b)
			status code: 400, request id: a65247e5-7215-11e8-838e-479a14bd8b1e
```

I've tried a few different options but I'm stumped, can anyone take a look?

The crawler is almost finished 👍

darrenhaken:
Strangely it seems to be working now, I am not sure why. Maybe some kind of raise condition?

darrenhaken:
OK this looks finished to me now with all tests passing. 

@bflad could you or someone else review the PR?

darrenhaken:
I thought I'd give a bit more detail on where this is up to.

I have now finished writing the AWS Glue Crawler resource and I have included all fields which are available at this point in time from the SDK.

I still need to do the documentation but I'd appreciate a core contributor to at least look up to this point to let me know if you are happy to accept the changes.

My company would love to have this ASAP by the way :)

aws/provider.go
```diff
@@ -412,6 +412,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_glue_catalog_table":                       resourceAwsGlueCatalogTable(),
 			"aws_glue_classifier":                          resourceAwsGlueClassifier(),
 			"aws_glue_connection":                          resourceAwsGlueConnection(),
+			"aws_glue_catalog_crawler":                     resourceAwsGlueCatalogCrawler(),
```
bflad: We should probably remove `_catalog` from the resource/file names, and `Catalog` from the function names everywhere to match the API, which just refers to these objects without including catalog in the naming.

You know this already but this resource is missing documentation in `website/docs/r/glue_crawler.html.markdown` and a sidebar link in `website/aws.erb` 😉

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+			"schema_change_policy": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"delete_behavior": {
+							Type:     schema.TypeString,
+							Optional: true,
+							//ValidateFunc: validateDeletion,
```
bflad: This should help:

```go
  glue.DeleteBehaviorDeleteFromDatabase,
  glue.DeleteBehaviorDeprecateInDatabase,
  glue.DeleteBehaviorLog,
}, false),
```

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+							//TODO: Write a validate function to ensure value matches enum
+						},
+						"update_behavior": {
+							Type:     schema.TypeString,
+							Optional: true,
+							//ValidateFunc: validateUpdate,
```
bflad: This should help:

```go
  glue.UpdateBehaviorLog,
  glue.UpdateBehaviorUpdateInDatabase,
}, false),
```

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+							Optional: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+					},
+				},
+			},
+			"configuration": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: suppressEquivalentAwsPolicyDiffs,
```
bflad: We should prefer to use the generic JSON DiffSuppressFunc, `suppressEquivalentJsonDiffs`, here as this does not look like an IAM policy according to the SDK documentation:

```
// Example: '{ "Version": 1.0, "CrawlerOutput": { "Partitions": { "AddOrUpdateBehavior":
// "InheritFromTable" } } }'
```

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+
+	input := &glue.GetCrawlerInput{
+		Name: aws.String(name),
+	}
+
+	crawlerOutput, err := glueConn.GetCrawler(input)
+	if err != nil {
+		if isAWSErr(err, glue.ErrCodeEntityNotFoundException, "") {
+			log.Printf("[WARN] Glue Crawler (%s) not found, removing from state", d.Id())
+			d.SetId("")
```
bflad: This should `return nil` immediately afterwards 👍

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+		}
+
+		return fmt.Errorf("error reading Glue crawler: %s", err.Error())
+	}
+
```
bflad: To prevent potential panics, we should check that `crawlerOutput.Crawler` is not `nil` beforehand:

```go
if crawlerOutput.Crawler == nil {
  log.Printf("[WARN] Glue Crawler (%s) not found, removing from state", d.Id())
  d.SetId("")
  return nil
}
```

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+	d.Set("database_name", crawlerOutput.Crawler.DatabaseName)
+	d.Set("role", crawlerOutput.Crawler.Role)
+	d.Set("description", crawlerOutput.Crawler.Description)
+	d.Set("schedule", crawlerOutput.Crawler.Schedule)
+	d.Set("classifiers", flattenStringList(crawlerOutput.Crawler.Classifiers))
+	d.Set("table_prefix", crawlerOutput.Crawler.TablePrefix)
+
+	if crawlerOutput.Crawler.SchemaChangePolicy != nil {
+		schemaPolicy := map[string]string{
+			"delete_behavior": *crawlerOutput.Crawler.SchemaChangePolicy.DeleteBehavior,
```
bflad: To prevent potential panics we should prefer to use the SDK provided `aws.StringValue()` function here instead of direct `*` dereferencing, e.g.

```go
"delete_behavior": aws.StringValue(crawlerOutput.Crawler.SchemaChangePolicy.DeleteBehavior),
"update_behavior": aws.StringValue(crawlerOutput.Crawler.SchemaChangePolicy.UpdateBehavior),
```

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+			"update_behavior": *crawlerOutput.Crawler.SchemaChangePolicy.UpdateBehavior,
+		}
+		d.Set("schema_change_policy", schemaPolicy)
```
bflad: We should [perform error checking when working with aggregate attribute types](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types), e.g.

```go
if err := d.Set("schema_change_policy", schemaPolicy); err != nil {
  return fmt.Errorf("error setting schema_change_policy: %s")
}
```

I believe you'll find that its silently ignoring a type error currently because `schema_change_policy` is a `TypeList` and you're trying to set a `map`. You should be able to resolve this with something like:

```go
if err := d.Set("schema_change_policy", []map[string]interface{}{schemaPolicy}); err != nil {
  return fmt.Errorf("error setting schema_change_policy: %s")
}
```

(Although we generally prefer to leave typing logic like this in `flattenXXX()` functions)

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
```
bflad: To prevent potential panics, we should `nil` check `crawlerOutput.Crawler.Targets` here.

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+	var jdbcTargets = crawlerOutput.Crawler.Targets.JdbcTargets
+	if crawlerOutput.Crawler.Targets.JdbcTargets != nil {
+		if err := d.Set("jdbc_target", flattenJdbcTargets(jdbcTargets)); err != nil {
+			log.Printf("[ERR] Error setting Glue JDBC Targets: %s", err)
+		}
+	}
+
+	// AWS provides no other way to read back the additional_info
```
bflad: The schema does not define anything relating to `additional_info`. Did you mean `configuration`? `crawlerOutput.Configuration` will have the information.

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+		info, err := structure.NormalizeJsonString(v)
+		if err != nil {
+			return fmt.Errorf("Additional Info contains an invalid JSON: %v", err)
+		}
+		d.Set("additional_info", info)
+	}
+	return nil
+}
+
+func flattenS3Targets(s3Targets []*glue.S3Target) []map[string]interface{} {
```
bflad: Since this repository currently uses a very flat Go package structure, we should prefer to include `Glue` in the naming of the `flatten`/`expand` function names.

aws/resource_aws_glue_catalog_crawler_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/glue"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"testing"
+)
+
+func TestAccAWSGlueCrawler_basic(t *testing.T) {
+	const name = "aws_glue_catalog_crawler.test"
+	resource.Test(t, resource.TestCase{
```
bflad: All the `TestCase` below are missing the `CheckDestroy` function. See other resource acceptance tests and the [Terraform documentation on CheckDestroy](https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html#checkdestroy) for more information.

aws/resource_aws_glue_catalog_crawler_test.go
```diff
@@ -0,0 +1,276 @@
...
+					resource.TestCheckResourceAttr(name, "schema_change_policy.0.update_behavior", "UPDATE_IN_DATABASE"),
+					resource.TestCheckResourceAttr(name, "s3_target.#", "2"),
+					resource.TestCheckResourceAttr(name, "s3_target.#", "2"),
+				),
+			},
+		},
+	})
+}
+
+func checkGlueCatalogCrawlerExists(name string, crawlerName string) resource.TestCheckFunc {
```
bflad: The `crawlerName` parameter here is extraneous, its available in `rs.Primary.ID`

aws/resource_aws_glue_catalog_crawler_test.go
```diff
@@ -0,0 +1,276 @@
...
+			return fmt.Errorf("no Glue Crawler found")
+		}
+
+		return nil
+	}
+}
+
+const testAccGlueCrawlerConfigBasic = `
+	resource "aws_glue_catalog_database" "test_db" {
+  		name = "test_db"
```
bflad: For all resources requiring a name, we should randomize it due to:
* Acceptance test coding errors leaving dangling resources
* API errors leaving dangling resources

Generally we achieve this by adding something like the following in the acceptance test function:

```go
rName := acctest.RandomWithPrefix("tf-acc-test")
```

Then we change these test configurations to be a function themselves that call `fmt.Sprintf()`, e.g.

```go
func testAccGlueCrawlerConfigBasic(rName string) string {
  return fmt.Sprintf(`
resource "aws_glue_catalog_database" "test_db" {
  name = %q
}
# ...
`, rName)
}
```

Then back in the acceptance test function change the `Config` line to a proper function call, e.g. `Config: testAccGlueCrawlerConfigBasic(rName),`

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+	result := make([]map[string]interface{}, 0)
+
+	for _, s3Target := range s3Targets {
+		attrs := make(map[string]interface{})
+		attrs["path"] = *s3Target.Path
```
bflad: To prevent potential panics, we should use `aws.StringValue()` here as well.

aws/resource_aws_glue_catalog_crawler.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return result
+}
+
+func flattenJdbcTargets(jdbcTargets []*glue.JdbcTarget) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0)
+
+	for _, jdbcTarget := range jdbcTargets {
+		attrs := make(map[string]interface{})
+		attrs["path"] = *jdbcTarget.Path
```
bflad: Same note about `aws.StringValue()` here and the line below.

aws/provider.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Good point, I've renamed both.

FYI my 2 cents - I think this should be changed from a single package ;-)

darrenhaken:
@bflad thanks for getting back to me with the changes.

I'll do my best to change what I can today but if you can finish some of it off I'd certainly appreciate it. I'm quite swamped this week 👎

I have already pushed the rename and I'll do see what I can do on the rest.

darrenhaken:
@bflad I have made a commit which fixes most of the issues you've raised. Do you think you could do the last few? I ran out of time.

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler.go
darrenhaken: Done

aws/resource_aws_glue_catalog_crawler_test.go
darrenhaken: Done

bflad:
@darrenhaken you rock 🥇  I'll add one last commit after yours to get this across the finish line. We should be able to release this today! More soon.

darrenhaken:
Sweet! I've got a commit on my laptop to do the test with random names
which I can commit in an hour or so. I didn't get chance to do the docs and
if I missed anything then I'm sorry!

aws/resource_aws_glue_catalog_crawler_test.go
darrenhaken: done

darrenhaken:
@bflad Sadly I had to revert my commit on the random acceptance test names. The `basic` test failed with a diff error and I do not have any time left today to take a look.

If you could finish the last bits off I'd *really* appreciate it. We will be really pleased when this goes in :)

Do you think you can?

P.S I didnt get chance to do the docs.

darrenhaken:
@bflad sorry to mess you around with this but I did some more investigation.

The `basic` test was failing by adding in one of your suggested changes:
```
if err := d.Set("schema_change_policy", []map[string]string{schemaPolicy}); err != nil {
			return fmt.Errorf("error setting schema_change_policy: %s", schemaPolicy)
		}
```

I have commented this out for now, see line 342.

This does means I have been able to keep the commits in which allow the random AC test names.

darrenhaken:
@bflad any luck?

bflad:
Thanks so much for your work here @darrenhaken! All commits through https://github.com/terraform-providers/terraform-provider-aws/commit/e5832d7404d343dfd50f9b117271261b1b256750 are merged (sorry thought you were done pushing already earlier) and https://github.com/terraform-providers/terraform-provider-aws/commit/6f41ed5f7dd777be6449f541cd3ccc90c26cb22c finishes the implementation including testing all the attributes individually with updates and documentation. I'll need to manually close this PR since GitHub will notice the last two commits in this branch missing from earlier today.

bflad:

PR #4486, opened 2018-05-08
[WIP] res/vpc_peering_connection_accepter: add Importer

Dominik-K:
## Changes proposed in this pull request

* adds `import` functionality to `aws_vpc_peering_connection_accepter`

## Questions

- [ ] How to get the provider's `account_id` and `region` inside this code? I want to add a usage validating function that will check that
    
        account_id != peer_owner_id OR
        region != peer_region

  If it's true, i.e. only VPC's differ, you'll need to use **one** `aws_vpc_peering_connection` instead. Otherwise the options on both resources can be ambiguous.

## Remarks

* Simply using `schema.ImportStatePassthrough` forces new resource on next `plan`. The `ID` is the same on both sides of the peering connection.


bluemalkin:
Hi - has this been released ?

tyrken:
While not necessary (can just apply it over the top) adding import would be a much nicer experience

tyrken:
Would like to add to please DONT add the assertion against use with the same region/account as your "Question" implies.  I'm using explicit accepter & separate option resources for a VPC module with combined peering I'm working on...

Instead please just test/merge as is.

bflad:
Hi @Dominik-K 👋 Since we haven't seen any updates to this pull request in a very long time, we went ahead and finished the implementation (testing and documentation) in a followup commit. Thanks for your contribution!

Issue #4247, opened 2018-04-18
aws_emr_cluster swallows JSON config errors


peay:
### Terraform Version

Terraform v0.10.8, AWS provider 1.14

### Affected Resource(s)

- `aws_emr_cluster`

### Expected Behavior

The `configurations` field allows passing in a JSON configuration for the cluster.

When the JSON is invalid (e.g., parse error), an error should occur during `plan` indicating why the JSON configuration cannot be parsed.

### Actual Behavior

The error is silently ignored, and an EMR cluster with an empty configuration is deployed.

During `apply`, debug logs have:

```
2018-04-18T12:33:54.202+0200 [DEBUG] plugin.terraform-provider-aws_v1.14.1_x4: 2018/04/18 12:33:54 [ERROR] parsing JSON invalid character '"' after object key
2018-04-18T12:33:54.202+0200 [DEBUG] plugin.terraform-provider-aws_v1.14.1_x4: 2018/04/18 12:33:54 [ERR] Error reading body JSON: invalid character '"' after object key
2018-04-18T12:33:54.202+0200 [DEBUG] plugin.terraform-provider-aws_v1.14.1_x4: 2018/04/18 12:33:54 [DEBUG] Expanded EMR Configurations []
```

Notice how these are all `DEBUG` messages, manually formatted to have `[ERR]` in their message. This means the messages are only visible at `DEBUG` level.

Then, Terraform proceeds to apply the change and deploy the cluster. As can be seen from the `Expanded EMR configuration` log line, the configuration will be left empty.

In https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_emr_cluster.go#L1326, the error is not propagated, and the error is silently ignored instead of being propagated.

bflad:
Version 1.30.0 of the AWS provider, releasing later today, will have a new `configurations_json` argument that accepts the actual JSON contents and can correctly determine drift from the EMR API. In an upcoming release we'll mark the existing `configurations` argument as deprecated.

To convert to the new attribute, given this Terraform configuration:

```hcl
resource "aws_emr_cluster" "example" {
  # ... other configuration ...
  configurations = "example.json"
}
```

Update it to:

```hcl
resource "aws_emr_cluster" "example" {
  # ... other configuration ...
  configurations_json = "${file("example.json")}"
}
```

bflad:
[...]

Issue #3652, opened 2018-03-07
Investigate aws_emr_cluster d.Set() errors only logging instead of returning errors


bflad:
It seems like all instances of:

```go
	if err := d.Set("...", ...); err != nil {
		log.Printf("[ERR] Error setting ... for cluster (%s): %s", d.Id(), err)
	}
```

Should really be:

```go
	if err := d.Set("...", ...); err != nil {
		return fmt.Errorf("error setting ... for cluster (%s): %s", d.Id(), err)
	}
```

There might be a historical reason for this current behavior though.

### Terraform Version

Terraform 0.10+
All AWS provider versions

### Affected Resource(s)

- aws_emr_cluster

### References

https://github.com/terraform-providers/terraform-provider-aws/blob/7a6a106542fadb4d2a52883e673d3d6dbd63219e/aws/resource_aws_emr_cluster.go#L566

bflad:
Resolved by #4488 / #6498!

Issue #1385, opened 2017-08-10
Updating a file pointed to in aws_emr_cluster.configurations should trigger instance recreation


synhershko:
E.g.

```hcl
resource "aws_emr_cluster" "querying" {
  name          = "Querying"
  release_label = "emr-5.7.0"
  applications  = ["Hadoop", "Hive", "Presto", "Ganglia"]

  configurations = "querying-emr-configurations.json"

  master_instance_type = "m4.large"
  core_instance_type   = "m4.2xlarge"
  core_instance_count  = 2
}
```

Updating the file `querying-emr-configurations.json` isn't detected and doesn't trigger new instance creation while it should.

Using v0.10.0

bflad:
The `configurations` attribute is quite an odd implementation in this older resource. Its taking in a file path/URL as an input while the API only ever would know how to return the JSON content. There are two issues here:

* During the resource read function, this attribute is never getting set into the Terraform state properly (the `d.Set()` call for it will always fail with trying to write the wrong `[]*emr.Configuration` type as `schema.TypeString`)
* Even if we fix the above, we cannot detect drift between the original input (a file path or URL) and what the API returns (JSON content)

It seems like the `configurations` attribute needs to be deprecated and a new attribute (`configurations_json`?) created that properly only accepts JSON input so it can be properly compared to what the API returns.

bflad:
Related: #543, which would also be fixed with JSON <-> JSON attribute (as it wouldn't matter where the content was coming from to Terraform).

bflad:
Version 1.30.0 of the AWS provider, releasing later today, will have a new `configurations_json` argument that accepts the actual JSON contents and can correctly determine drift from the EMR API. In an upcoming release we'll mark the existing `configurations` argument as deprecated.

To convert to the new attribute, given this Terraform configuration:

```hcl
resource "aws_emr_cluster" "example" {
  # ... other configuration ...
  configurations = "example.json"
}
```

Update it to:

```hcl
resource "aws_emr_cluster" "example" {
  # ... other configuration ...
  configurations_json = "${file("example.json")}"
}
```

bflad:
This has been released in [version 1.30.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1300-august-02-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

Issue #543, opened 2017-06-13
EMR configurations field forcing new EMR cluster when path changes


hashibot:
_This issue was originally opened by @cpower as hashicorp/terraform#12047. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi there,

Running `terraform apply` on a Terraform managed EMR cluster that uses a path.module reference to configurations.json with remote state, when the path to the file changes, causes Terraform to destroy the EMR cluster and create a new one because the path to the configurations.json file is stored in the Terraform state which forces a new resource.

### Terraform Version
Terraform v0.8.5

### Affected Resource(s)
- aws_emr_cluster

### Terraform Configuration Files
```
resource "aws_emr_cluster" "emr" {
  name = "${var.cluster_name}"
  release_label = "${var.release_label}"
  applications  = ["${split(",",var.applications)}"]
  service_role  = "${var.service_role}"
  log_uri  = "s3n://${var.s3_bucket}"
  master_instance_type = "${var.master_instance_type}"
  core_instance_type   = "${var.core_instance_type}"
  core_instance_count  = "${var.core_instance_count}"
  ec2_attributes {
    key_name = "${var.ec2_key_name}"
    subnet_id = "${data.aws_subnet.subnet.id}"
    instance_profile = "${var.ec2_instance_profile}"
    service_access_security_group = "${var.security_groups["emr_service_access"]}"
    emr_managed_master_security_group = "${var.security_groups["emr_managed_master"]}"
    emr_managed_slave_security_group  = "${var.security_groups["emr_managed_slave"]}"
    additional_master_security_groups = "${var.security_groups["emr_additional_master"]}"
    additional_slave_security_groups  = "${var.security_groups["emr_additional_slave"]}"
  }
  configurations = "${path.module}/${var.configurations}"
}
```

### Expected Behavior
Updating a Terraform managed EMR cluster that uses a path.module reference to a configurations JSON file from a different path/machine when using remote state should not create and destroy a new EMR cluster.

### Actual Behavior
Running `terraform apply` on a Terraform managed EMR cluster that uses a path.module reference, causes Terraform to destroy the EMR cluster and create a new one because the path to the configurations.json file is stored in the Terraform state.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
[...]

PR #4488, opened 2018-05-09
aws_emr_cluster importer

stefansundin:
This is my first attempt at making an importer, but I tried this and it seems to work fine.

Not sure if `aws_emr_instance_group` would benefit from an importer as well? We don't use that resource so I don't have the ability to immediately test it.

Also fixed some typos in the docs for `aws_emr_security_configuration`.

Thanks!


aws/resource_aws_emr_cluster.go
```diff
@@ -25,6 +25,10 @@ func resourceAwsEMRCluster() *schema.Resource {
 		Read:   resourceAwsEMRClusterRead,
 		Update: resourceAwsEMRClusterUpdate,
 		Delete: resourceAwsEMRClusterDelete,
+		Importer: &schema.ResourceImporter{
```
bflad: This can be acceptance tested by adding the following as a `TestStep` e.g. after the existing step of `TestAccAWSEMRCluster_basic`:

```go
{
	ResourceName:      "aws_emr_cluster.tf-test-cluster",
	ImportState:       true,
	ImportStateVerify: true,
},
```

stefansundin: Done. Thanks for the tip!

bflad:
This isn't as straightforward as it seems -- this resource is pretty old and does not properly set attributes in the Terraform state during the read function, which is now detected by the import test:

```
=== RUN   TestAccAWSEMRCluster_basic
--- FAIL: TestAccAWSEMRCluster_basic (606.94s)
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=6) {
         (string) (len=14) "configurations": (string) (len=37) "test-fixtures/emr_configurations.json",
         (string) (len=19) "core_instance_count": (string) (len=1) "1",
         (string) (len=18) "core_instance_type": (string) (len=8) "c4.large",
         (string) (len=33) "keep_job_flow_alive_when_no_steps": (string) (len=4) "true",
         (string) (len=20) "master_instance_type": (string) (len=8) "c4.large",
         (string) (len=22) "termination_protection": (string) (len=5) "false"
        }
```

In some of these cases, the fix is as simple as adding `d.Set()` from the API response in the read function. For others though its more complicated. Had I investigated #3652 more, definitely would have discovered these sooner. 😓 

In the `configurations` attribute case, the resource handling of the attribute is completely incorrect to detect resource drift as its taking a file path/URL as input but the API is returning the actual JSON (of course it doesn't know or care about how the JSON content got there), yikes. You can see a related issues for this attribute in #543 and #1385. That attribute also has very weird behavior where it doesn't properly return errors as noted in #4247. Likely the only viable solution there is to create a new attribute (`configurations_json`?) that accepts a JSON string and deprecates the old attribute.

For now you can add this to the new import `TestStep`:

```go
ImportStateVerifyIgnore: []string{"configurations"},
```

Sorry you found a can of worms! Are you able to see about fixing some of these? If not, no big deal, we can add them to the ignore list for now and deal with bug reports later (which should have already been the case as they were missing drift detection). Some of these are thornier than others to properly get the information out of the API.

stefansundin:
I'll give the easy ones a shot. Thanks for the write up.

bflad:
Hey @stefansundin! 👋  Any luck so far? No worries if not, we're just trying to reduce the number of lingering pull requests and can make this work as-is. I didn't want to start doing anything here if you indeed had anything else to add to it. Thanks so much!

stefansundin:
Hey @bflad. I gave it a try but I had some issues.

- I haven't found any way to get `keep_job_flow_alive_when_no_steps` by describing stuff.. so it seems that there is a legacy "job flow" API that terraform is using for some things, and it is hard to make sense of the documentation.. maybe this value is returned with another name in the new API.
- `core_instance_type ` is already being set here: https://github.com/terraform-providers/terraform-provider-aws/blob/70417de771d89f3830277b907c0f0221600bfcec/aws/resource_aws_emr_cluster.go#L635
  - but for some reason the test still fails on it?
  - I was going to add `d.Set("core_instance_count", coreGroup.RequestedInstanceCount)` just below it.
- Think I can add `d.Set("termination_protection", cluster.TerminationProtected)` below this line: https://github.com/terraform-providers/terraform-provider-aws/blob/70417de771d89f3830277b907c0f0221600bfcec/aws/resource_aws_emr_cluster.go#L653
- I think I can add the following:
   ```
   masterGroup := findGroup(instanceGroups, "MASTER")
   if masterGroup != nil {
   	d.Set("master_instance_type", masterGroup.InstanceType)
   	d.Set("master_instance_count", masterGroup.RequestedInstanceCount)
   }
   ```
   right below this line: https://github.com/terraform-providers/terraform-provider-aws/blob/70417de771d89f3830277b907c0f0221600bfcec/aws/resource_aws_emr_cluster.go#L636
   For some reason there is no warning about `master_instance_count`, not sure why.

Anyway, the mixing of the legacy "job flow" API and the new cluster API made it a bit confusing to me. And I'm not an EMR expert by any means.

I have managed to run the tests now, but I haven't run it yet with the proposed changes above.

bflad:
Can you double check the other acceptance tests by adding the import `TestStep`? We need to verify that any configuration can successfully import without the `core_instance` ignores as well.

I think last time I ran the other acceptance tests with the additional `TestStep` I discovered that properly reading the attributes is actually more complicated since `core_instance_count` is actually 1 off from the actual value from the API (due to the master instance) and `core_instance_type` not being returned when there are no core instances created. I was hoping to provide some suggestions then got sidetracked and never commented here, sorry about that. 🙁

stefansundin:
Thanks. I don't think I have enough area-expertise to solve this. If someone else wants to finish this, please feel free!

bflad:

PR #4497, opened 2018-05-10
fix: update us-east-1 S3 bucket naming validations

fahrradflucht:
Since March 1, 2018 AWS applies the same naming conventions for all
regions. [0] This means that the special validation for us-east-1 is obsolete.

This enabled refactoring the validation to a normal schema validator which has
the benefit of plan-time validation.

[[0](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules)](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules)

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -6,8 +6,6 @@ import (
 	"fmt"
 	"log"
 	"net/url"
-	"regexp"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -39,7 +37,9 @@ func resourceAwsS3Bucket() *schema.Resource {
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"bucket_prefix"},
+				ValidateFunc:  validateS3BucketName,
 			},
+
 			"bucket_prefix": {
 				Type:     schema.TypeString,
 				Optional: true,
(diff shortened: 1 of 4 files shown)


bflad:
@fahrradflucht won't this break anyone with existing buckets that follow the old convention? I do not believe we have a way to disable `ValidateFunc` in the resource `lifecycle` configuration.

fahrradflucht:
@bflad lol yes you are correct. I must have been completely brain afk when I did this today in the morning. - Closing

fahrradflucht:
Thinking about it: What we really need isn't a way for the user to disable a validation but having validations that only run on resource that will be created.

This is actually similar to what terraform does right now but the difference would be that it could be executed at schema validation/plan time.

bflad:
I think that's a fantastic idea for a feature request in [Terraform core](https://github.com/hashicorp/terraform/). Sometimes an attribute's validation can be incorrect or prevent working behavior for newer values (e.g. API added new support for a value, but not updated in the provider yet).

If I was going to submit something, it'd be to support something like:

```hcl
resource "XXX" "example" {
  example_attribute = "incorrectly-validated-value"

  lifecycle {
    ignore_validation = ["example_attribute"]
  }
}
```

Similar to `ignore_changes` but hopefully much simpler to implement.

bflad:
Oh ha! I re-read what you meant. Interesting idea you have as well!

Issue #1763, opened 2017-09-27
aws_sns_topic Invalid parameter: TopicArn


hashibot:
_This issue was originally opened by @smailli as hashicorp/terraform#16148. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi guys! 

I have a problem !!!! 

When I use aws_sns_topic + aws_sns_topic_policy + aws_sns_topic_subscription and remove the configuration file, terraform returns Invalid parameter: TopicArn error. 

### Terraform Version
```
$ terraform version
Terraform v0.10.6
```

### Prerequisite
Create two SQS named `SQS_NAME0` and `SQS_NAME1`

### Terraform Configuration Files
**modules/aws/test/main.tf:**
```hcl
resource "aws_sns_topic" "default_topic" {
  name     = "TOPIC_NAME"
}

resource "aws_sns_topic_policy" "default_policy" {
  depends_on = ["aws_sns_topic.default_topic"]
  arn = "${aws_sns_topic.default_topic.arn}"
  policy = <<POLICY
{
  "Version": "2008-10-17",
  "Id": "__default_policy_ID",
  "Statement": [
    {
      "Sid": "__default_statement_ID",
      "Effect": "Allow",
      "Principal": {
        "AWS": "*"
      },
      "Action": [
        "SNS:GetTopicAttributes",
        "SNS:SetTopicAttributes",
        "SNS:AddPermission",
        "SNS:RemovePermission",
        "SNS:DeleteTopic",
        "SNS:Subscribe",
        "SNS:ListSubscriptionsByTopic",
        "SNS:Publish",
        "SNS:Receive"
      ],
      "Resource": "${aws_sns_topic.default_topic.arn}",
      "Condition": {
        "StringEquals": {
          "AWS:SourceOwner": "000000000000"
        }
      }
    }
  ]
}
POLICY
}

resource "aws_sns_topic_subscription" "default_topic_subscription" {
  count = "2"
  depends_on = ["aws_sns_topic.default_topic"]
  topic_arn = "${aws_sns_topic.default_topic.arn}"
  protocol  = "sqs"
  endpoint  = "arn:aws:sqs:us-west-2:000000000000:SQS_NAME${count.index}"
}
```

### Environment
In environment directory create a module file

**environment/test.tf:**
```hcl
module "sns-test" {
[...]

PR #4505, opened 2018-05-10
Better documentation for SNS Subscriptions

### LINKED ISSUE #981: Incorrect Documentation for aws_sns_topic_subscription
_This issue was originally opened by @seemeroland as hashicorp/terraform#15385. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

This note is incorrect:

"NOTE: If SNS topic and SQS queue are in different AWS regions it is important to place the "aws_sns_topic_subscription" into the terraform configuration of the region with the SQS queue. If "aws_sns_topic_subscription" is placed in the terraform configuration of the region with the SNS topic terraform will fail to create the subscription."

The subscription should be in the region of the SNS topic, not the SQS queue. The third is example is correct (using the SNS topic region), but this note is contradictory.

mikegrima:
The documentation about SNS topic subscriptions is incorrect.

This PR attempts to clarify this by adding wording about the AWS provider and also correctly indicates that the SNS topic subscription needs to be using an AWS provider that is in the same region as the topic.

Fixes #981

May also address #1763.

### DIFF
--- a/website/docs/r/sns_topic_subscription.html.markdown
+++ b/website/docs/r/sns_topic_subscription.html.markdown
@@ -13,11 +13,11 @@ This resource allows you to automatically place messages sent to SNS topics in S
 to a given endpoint, send SMS messages, or notify devices / applications. The most likely use case for Terraform users will
 probably be SQS queues.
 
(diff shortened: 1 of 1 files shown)


mikegrima:
Woah -- github messed up the branch.

Updating...

mikegrima:
Ok... now it's good :)

vvondra:
just got bitten by this, this helped me!

PR #4506, opened 2018-05-11
Add support for Azure SQL as a DMS endpoint

microamp:

* `azuredb` added to the list of engines supported in DMS
* Documentation updated accordingly

Please see `aws dms create-endpoint help`, in particular its `engine-name` flag. No additional acceptance tests were added as it only depends on top-level attributes that are already present.

Thanks so much for reviewing.


microamp:
> Looks like we need `s3` in the docs too, I'll update them after.

Good catch! Thanks for doing that. :clap: I recall documenting it. I suspect it was removed accidentally when I was resolving merge conflicts before. :disappointed_relieved:

bflad:

Issue #4449, opened 2018-05-04
no possibility to set iam_instance_profile for the spot fleet request as a global setting


TattiQ:
Hi guys,

Apologies if it's something simple, I am new to Terraform. 
I have been testing the resource "aws_spot_fleet_request" . The goal is to automate the creation of spot instance requests using terraform. Interesting observation  - one can specify iam_instance_profile in the launch_specification for the instances (and then the iam role for the instance is set just fine) , but it will not be visible in the spot instances request settings

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.
Terraform version: 0.11.7 

### Affected Resource(s)
Please list the resources as a list, for example:
resource "aws_spot_fleet_request"

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Debug Output
Please provider a link to a GitHub Gist containing the complete debug output: https://www.terraform.io/docs/internals/debugging.html. Please do NOT paste the debug output in the issue; just paste a link to the Gist.
Neither DEBUG nor TRACE logging showed  anything out of the ordinary

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.
no panic

### Expected Behavior
What should have happened?
if one creates a request manually, once it's fulfilled the properties look like this 
![image](https://user-images.githubusercontent.com/23435230/39628007-54a4f766-4fb0-11e8-9a65-ebb61b5d639f.png)

Note IAM role is filled in. 

### Actual Behavior
What actually happened?
![image](https://user-images.githubusercontent.com/23435230/39628019-5dce3ca8-4fb0-11e8-8ed7-52fb63c9858e.png)

No IAM role, but the instances have the IAM role assigned (since launch config contained it)

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:

```
resource "aws_sws_spot_fleet_request" "cheap_compute" {
  iam_fleet_role      = "arn:aws:iam::xxxxxxxxxxxxxx:role/ec2-spot-fleet"
  spot_price          = "0.888"
  allocation_strategy = "lowestPrice"
  target_capacity     = 2
  terminate_instances_with_expiration = true
  valid_until         = "2019-05-04T07:44:20Z"


[...]

PR #4511, opened 2018-05-11
issue #4449 Fix the issue using Arn field instead of Name in IamInstanceProfileSpecification

saravanan30erd:
Used Arn field instead of Name in Type IamInstanceProfileSpecification.

### DIFF
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -372,7 +372,7 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
 
 	if v, ok := d["iam_instance_profile"]; ok {
 		opts.IamInstanceProfile = &ec2.IamInstanceProfileSpecification{
-			Name: aws.String(v.(string)),
+			Arn: aws.String(v.(string)),
 		}
 	}
 
@@ -935,8 +935,8 @@ func launchSpecToMap(l *ec2.SpotFleetLaunchSpecification, rootDevName *string) m
 		m["monitoring"] = aws.BoolValue(l.Monitoring.Enabled)
 	}
 
-	if l.IamInstanceProfile != nil && l.IamInstanceProfile.Name != nil {
-		m["iam_instance_profile"] = aws.StringValue(l.IamInstanceProfile.Name)
+	if l.IamInstanceProfile != nil && l.IamInstanceProfile.Arn != nil {
+		m["iam_instance_profile"] = aws.StringValue(l.IamInstanceProfile.Arn)
 	}
 
 	if l.UserData != nil {
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -191,6 +191,11 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
 							ForceNew: true,
 							Optional: true,
 						},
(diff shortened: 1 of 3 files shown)


bflad:
Changing the parameter it uses will likely break existing configurations. 😄  Can you please introduce a separate Terraform attribute into the schema? e.g. `iam_instance_profile_arn` Thanks!

aws/resource_aws_spot_fleet_request.go
```diff
@@ -191,6 +191,11 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
 							ForceNew: true,
 							Optional: true,
 						},
+						"iam_instance_profile_arn": {
```
bflad: Can you please
* Document the new argument in the argument reference near the existing `iam_instance_profile` argument
* Add a simple acceptance test to cover this new argument? Copy pasting an existing test and test configuration that uses `iam_instance_profile` is fine 👍

website/docs/r/spot_fleet_request.html.markdown
```diff
@@ -27,13 +27,15 @@ resource "aws_spot_fleet_request" "cheap_compute" {
     ami               = "ami-1234"
     spot_price        = "2.793"
     placement_tenancy = "dedicated"
+    iam_instance_profile_arn = "arn:aws:iam::12345678:instance-profile/webserver-role"
```
bflad: We should either fix this so its a properly formatted ARN (account ID is 12 digits `123456789012`) or just reference a "fake" resource (`"${aws_iam_instance_profile.example.arn}"`). Also as a nitpick, if these are going to be added in here, please line up all the equals symbols similar to how `terraform fmt` would e.g.

```
ami                      = "ami-1234"
spot_price               = "2.793"
placement_tenancy        = "dedicated"
iam_instance_profile_arn = "arn:aws:iam::12345678:instance-profile/webserver-role"
```

aws/resource_aws_spot_fleet_request.go
saravanan30erd: Documented the new argument under `launch_specification` section and added the acceptance test.

website/docs/r/spot_fleet_request.html.markdown
saravanan30erd: Done 👍

bflad:

jekh:
I recently upgraded to 1.19.0 without changing my existing `iam_instance_profile`s to `iam_instance_profile_arn`s. As a result, none of the launched instances had IAM roles assigned at all. So this definitely will break existing configurations.

Changing to the new _arn attribute fixed the problem, and using the ARN seems like a fine idea. Would it be better to simply remove the existing `iam_instance_profile` attribute, so at least folks' terraform breaks, rather than silently not working properly in a non-obvious way?

saravanan30erd:
when using `iam_instance_profile`, `iam_instance_profile_arn` assigned to empty string and looks like it overrides the `IamInstanceProfile ` finally so that launched instances had no IAM roles assigned at all.

bflad:
This has been released in [version 1.20.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1200-may-23-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #4527, opened 2018-05-13
Add import to aws_iam_role_policy_attachment

### LINKED ISSUE #544: provider/aws: Add import support for aws_iam_role_policy_attachment
_This issue was originally opened by @jzbruno as hashicorp/terraform#12100. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Please add import support for the aws_iam_role_policy_attachment resource.

### Terraform Version

All

### Affected Resource(s)

- aws_iam_role_policy_attachment

### Expected Behavior

Running the import command for the aws_iam_role_policy_attachment resource should update the state with all aws_iam_role_policy_attachment's associated with the specified role.

### Steps to Reproduce

1. `terraform import aws_iam_role_policy_attachment.test <ROLE_NAME>`

mindfulmonk:
Fixes #544
Based on @jzbruno https://github.com/hashicorp/terraform/pull/12101 

* Add import functionality to role_policy_attachment

```
make testacc TESTARGS='-run=TestAccAwsIamRolePolicyAttachmentImport'
TF_ACC=1 go test ./... -v -run=TestAccAwsIamRolePolicyAttachmentImport -timeout 120m
=== RUN   TestAccAwsIamRolePolicyAttachmentImport
--- PASS: TestAccAwsIamRolePolicyAttachmentImport (39.39s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	39.450s
```

The test is not very robust, I had to have an empty "test-role" in my AWS account to get it to pass. Would appreciate some help on that front.

### DIFF
--- a/aws/import_aws_iam_role_policy_attachment.go
+++ b/aws/import_aws_iam_role_policy_attachment.go
@@ -1,0 +1,38 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsIamRolePolicyAttachmentImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	conn := meta.(*AWSClient).iamconn
+	roleName := d.Id()
+
+	resp, err := conn.ListAttachedRolePolicies(&iam.ListAttachedRolePoliciesInput{
+		RoleName: aws.String(roleName),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	results := []*schema.ResourceData{}
+
+	for _, policy := range resp.AttachedPolicies {
+		id := fmt.Sprintf("%s-%s", roleName, *policy.PolicyArn)
+
+		attachment := resourceAwsIamRolePolicyAttachment()
+		ad := attachment.Data(nil)
+		ad.SetId(id)
+		ad.SetType("aws_iam_role_policy_attachment")
+		ad.Set("role", roleName)
+		ad.Set("policy_arn", policy.PolicyArn)
+
+		results = append(results, ad)
+	}
+
+	return results, nil
+}
--- a/aws/resource_aws_iam_role_policy_attachment.go
+++ b/aws/resource_aws_iam_role_policy_attachment.go
@@ -7,7 +7,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
-	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
@@ -16,6 +15,9 @@ func resourceAwsIamRolePolicyAttachment() *schema.Resource {
 		Create: resourceAwsIamRolePolicyAttachmentCreate,
 		Read:   resourceAwsIamRolePolicyAttachmentRead,
 		Delete: resourceAwsIamRolePolicyAttachmentDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceAwsIamRolePolicyAttachmentImport,
+		},
 
 		Schema: map[string]*schema.Schema{
 			"role": &schema.Schema{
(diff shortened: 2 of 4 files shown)


robkinyon-roivant:
Bump. Could this be merged?

aws/resource_aws_iam_role_policy_attachment.go
```diff
@@ -43,7 +45,7 @@ func resourceAwsIamRolePolicyAttachmentCreate(d *schema.ResourceData, meta inter
 		return fmt.Errorf("[WARN] Error attaching policy %s to IAM Role %s: %v", arn, role, err)
 	}
 
-	d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", role)))
```
bflad: While its unlikely that operators are currently using the ID of this resource due to its format, we do not currently have a precedent for changing the format of resource IDs and changing them is considered a major breaking change: https://www.terraform.io/docs/extend/best-practices/versioning.html#example-major-number-increments

We can support import without changing this. 👍

aws/import_aws_iam_role_policy_attachment.go
```diff
@@ -0,0 +1,38 @@
...
+	resp, err := conn.ListAttachedRolePolicies(&iam.ListAttachedRolePoliciesInput{
+		RoleName: aws.String(roleName),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	results := []*schema.ResourceData{}
+
+	for _, policy := range resp.AttachedPolicies {
```
bflad: We strongly prefer to have resources only import resources 1:1 currently -- otherwise this can cause dangerous behavior if the operator is not completely vigilant after import. See also: #5631

aws/resource_aws_iam_role_policy_attachment.go
```diff
@@ -16,6 +15,9 @@ func resourceAwsIamRolePolicyAttachment() *schema.Resource {
 		Create: resourceAwsIamRolePolicyAttachmentCreate,
 		Read:   resourceAwsIamRolePolicyAttachmentRead,
 		Delete: resourceAwsIamRolePolicyAttachmentDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceAwsIamRolePolicyAttachmentImport,
```
bflad: Instead of implementing a "complex" import function (importing all attached policies), which can be dangerous, we should instead support importing the resource 1:1.

This can be implemented via (which expects a format ROLENAME_POLICYARN):

```go
			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
				idParts := strings.Split(d.Id(), "_")
				if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
					return nil, fmt.Errorf("unexpected format of ID (%q), expected ROLENAME_POLICYARN", d.Id())
				}
				roleName := idParts[0]
				policyARN := idParts[1]
				d.Set("policy_arn", policyARN)
				d.Set("role_name", roleName)
				d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", roleName)))
				return []*schema.ResourceData{d}, nil
			},
		},
```

And tested via:

```go
// new function
func testAccAWSIAMRolePolicyAttachmentImportStateIdFunc(resourceName string) resource.ImportStateIdFunc {
	return func(s *terraform.State) (string, error) {
		rs, ok := s.RootModule().Resources[resourceName]
		if !ok {
			return "", fmt.Errorf("not found: %s", resourceName)
		}

		return fmt.Sprintf("%s_%s", rs.Primary.Attributes["role_name"], rs.Primary.Attributes["policy_arn"]), nil
	}
}

// As a new TestStep in TestAccAWSRolePolicyAttachment_basic
{
	ResourceName:      "aws_iam_role_policy_attachment.test-attach",
	ImportState:       true,
	ImportStateIdFunc: testAccAWSIAMRolePolicyAttachmentImportStateIdFunc("aws_iam_role_policy_attachment.test-attach"),
	ImportStateVerify: true,
},
```

aws/import_aws_iam_role_policy_attachment.go
bflad: I should also note this behavior of multiple imports can be very problematic if the resources are implemented across separate modules. It will force the operator to `terraform state rm` the other attachments repeatedly for each module (and likely unexpectedly).

jzbruno:
Thanks for the review. I will take a look at this tonight.

jzbruno:
@bflad Got this working. Thanks for the review and suggestions. I don't think `ImportStateIdFunc` existed when I created the PR over a year ago.

I have a couple questions.

- This is no longer my original fork and PR. Should I create a new fork and PR? Suggestions on the best way to handle this?
- Both roles and policies allow an arbitrary number of _ in their names and ARNs can have `/`. How about `//` for the delimiter of the import ID?

> Names of users, groups, roles, policies, instance profiles, and server certificates must be alphanumeric, including the following common characters: plus (+), equal (=), comma (,), period (.), at (@), underscore (_), and hyphen (-).  

[Limitations on IAM Entities and Objects - AWS Identity and Access Management](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html)

bflad:
Superseded by #5910, which has been merged and will release with version 1.37.0 of the AWS provider, likely tomorrow. 👍

PR #4530, opened 2018-05-14
Feature/aws cognito resource server

### LINKED ISSUE #3616: AWS Cognito : Resource Server - New Feature
_This issue was originally opened by @mdinuth as hashicorp/terraform#17495. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

**Feature Request**
Terraform currently doesn't support AWS Cognito User Pool Resource Server configurations. 
(AWS Cognito User Pool > (Left Panel) App Integration > Resource servers)
> **aws_cognito_user_pool_resource_server** 

Is it still under development? When will it be in production?

tomjanne:
Closes #3616

### DIFF
--- a/aws/resource_aws_cognito_resource_server.go
+++ b/aws/resource_aws_cognito_resource_server.go
@@ -1,0 +1,178 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cognitoidentityprovider"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsCognitoResourceServer() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCognitoResourceServerCreate,
+		Read:   resourceAwsCognitoResourceServerRead,
+		Update: resourceAwsCognitoResourceServerUpdate,
+		Delete: resourceAwsCognitoResourceServerDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		// https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateResourceServer.html
+		Schema: map[string]*schema.Schema{
+			"identifier": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"scope": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				MaxItems: 25,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"scope_description": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateCognitoResourceServerScopeDescription,
+						},
+						"scope_name": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateCognitoResourceServerScopeName,
+						},
+						"scope_identifier": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"user_pool_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"scope_identifiers": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+		},
+	}
+}
+
+func resourceAwsCognitoResourceServerCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+
+	params := &cognitoidentityprovider.CreateResourceServerInput{
+		Identifier: aws.String(d.Get("identifier").(string)),
+		Name:       aws.String(d.Get("name").(string)),
+		UserPoolId: aws.String(d.Get("user_pool_id").(string)),
+	}
+
+	if v, ok := d.GetOk("scope"); ok {
+		configs := v.(*schema.Set).List()
+		params.Scopes = expandCognitoResourceServerScope(configs)
+	}
+
+	log.Printf("[DEBUG] Creating Cognito Resource Server: %s", params)
+
+	resp, err := conn.CreateResourceServer(params)
+
+	if err != nil {
+		return errwrap.Wrapf("Error creating Cognito Resource Server: {{err}}", err)
+	}
+
+	d.SetId(*resp.ResourceServer.Identifier)
+
+	return resourceAwsCognitoResourceServerRead(d, meta)
+}
+
+func resourceAwsCognitoResourceServerRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).cognitoidpconn
+
+	params := &cognitoidentityprovider.DescribeResourceServerInput{
+		Identifier: aws.String(d.Id()),
+		UserPoolId: aws.String(d.Get("user_pool_id").(string)),
+	}
+
+	log.Printf("[DEBUG] Reading Cognito Resource Server: %s", params)
+
+	resp, err := conn.DescribeResourceServer(params)
+
+	if err != nil {
+		if isAWSErr(err, "ResourceNotFoundException", "") {
+			log.Printf("[WARN] Cognito Resource Server %s is already gone", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.SetId(*resp.ResourceServer.Identifier)
+	d.Set("name", *resp.ResourceServer.Name)
+	d.Set("user_pool_id", *resp.ResourceServer.UserPoolId)
+
+	scopes := flattenCognitoResourceServerScope(*resp.ResourceServer.Identifier, resp.ResourceServer.Scopes)
+	if err := d.Set("scope", scopes); err != nil {
+		return fmt.Errorf("Failed setting schema: %s", err)
+	}
+
+	var scope_identifiers []string
+	for _, elem := range scopes {
+
+		scope_identifier := elem["scope_identifier"].(string)
+		scope_identifiers = append(scope_identifiers, scope_identifier)
+	}
+	d.Set("scope_identifiers", scope_identifiers)
+	return nil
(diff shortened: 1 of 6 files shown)


aws/resource_aws_cognito_resource_server_test.go
```diff
@@ -0,0 +1,149 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSCognitoResourceServerDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSCognitoResourceServerConfig_full(identifier, name, poolName),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckAWSCognitoResourceServerExists("aws_cognito_resource_server.main"),
+					resource.TestCheckResourceAttr("aws_cognito_resource_server.main", "identifier", identifier),
+					resource.TestCheckResourceAttr("aws_cognito_resource_server.main", "name", name),
+					resource.TestCheckResourceAttrSet("aws_cognito_resource_server.main", "scope_identifiers"),
```
bflad: This is currently failing:

```
=== RUN   TestAccAWSCognitoResourceServer_full
--- FAIL: TestAccAWSCognitoResourceServer_full (5.86s)
    testing.go:518: Step 0 error: Check failed: 1 error(s) occurred:
        
        * Check 4/5 error: aws_cognito_resource_server.main: Attribute 'scope_identifiers' expected to be set
```

Did you mean to do this?

```go
resource.TestCheckResourceAttr("aws_cognito_resource_server.main", "scope.#", "2"),
```

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+
+func resourceAwsCognitoResourceServer() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsCognitoResourceServerCreate,
+		Read:   resourceAwsCognitoResourceServerRead,
+		Update: resourceAwsCognitoResourceServerUpdate,
+		Delete: resourceAwsCognitoResourceServerDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
bflad: The current resource ID (only Identifier) does not have enough information to work with the passthrough importer. We also need the user pool ID so the read function can work directly from the ID.

In this case, we should probably prepend the user pool ID to the resource ID, e.g.

```go
# Choosing pipe delimiter here as the identifier might be a URL
d.SetId(fmt.Sprintf("%s|%s", userPoolID, identifier))
```

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		if isAWSErr(err, "ResourceNotFoundException", "") {
+			log.Printf("[WARN] Cognito Resource Server %s is already gone", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.SetId(*resp.ResourceServer.Identifier)
```
bflad: We should not call `d.SetId()` in read functions 👍

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+	d.Set("name", *resp.ResourceServer.Name)
```
bflad: To prevent panics, we should ensure that `resp` and `resp.ResourceServer` are not `nil` before trying to dereference them.

Also `d.Set()` is able to handle `nil` values directly, so we should prefer to _not_ dereference values to prevent potential panics.

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"scope": {
```
bflad: Pedantic nitpick 😅 : The API refers to this as `Scopes` and using the plural form would hopefully also signify to operators its list-like nature.

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+
+	params := &cognitoidentityprovider.UpdateResourceServerInput{
+		Identifier: aws.String(d.Id()),
+		Name:       aws.String(d.Get("name").(string)),
+		UserPoolId: aws.String(d.Get("user_pool_id").(string)),
+	}
+
+	log.Printf("[DEBUG] Updating Cognito Resource Server: %s", params)
+
+	_, err := conn.UpdateResourceServer(params)
```
bflad: Scopes can be updated but its not being added to the update parameters, so currently they would generate a perpetual difference.

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+		Identifier: aws.String(d.Id()),
+		UserPoolId: aws.String(d.Get("user_pool_id").(string)),
+	}
+
+	log.Printf("[DEBUG] Deleting Resource Server: %s", params)
+
+	_, err := conn.DeleteResourceServer(params)
+
+	if err != nil {
+		return errwrap.Wrapf("Error deleting Resource Server: {{err}}", err)
```
bflad: We should check for `isAWSErr(err, cognitoidentityprovider.ErrCodeResourceNotFoundError, "")` here and `return nil` so we can skip errors if the resource is deleted outside of Terraform.

aws/resource_aws_cognito_resource_server_test.go
```diff
@@ -0,0 +1,149 @@
...
+					resource.TestCheckResourceAttr("aws_cognito_resource_server.main", "identifier", identifier),
+					resource.TestCheckResourceAttr("aws_cognito_resource_server.main", "name", name),
+					resource.TestCheckResourceAttr("aws_cognito_user_pool.main", "name", poolName),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAWSCognitoResourceServer_full(t *testing.T) {
```
bflad: We should descriptively name this `_scopes` to signify what its actually testing

aws/validators.go
```diff
@@ -1386,6 +1386,33 @@ func validateCognitoUserPoolClientURL(v interface{}, k string) (ws []string, es
 	return
 }
 
+func validateCognitoResourceServerScopeDescription(v interface{}, k string) (ws []string, errors []error) {
```
bflad: This can be replaced with `validation.StringLenBetween(1, 256)`: https://godoc.org/github.com/hashicorp/terraform/helper/validation#StringLenBetween

aws/validators.go
```diff
@@ -1386,6 +1386,33 @@ func validateCognitoUserPoolClientURL(v interface{}, k string) (ws []string, es
...
+	value := v.(string)
+
+	if len(value) < 1 {
+		errors = append(errors, fmt.Errorf("%q cannot be less than 1 character", k))
+	}
+	if len(value) > 256 {
+		errors = append(errors, fmt.Errorf("%q cannot be longer than 256 character", k))
+	}
+	if !regexp.MustCompile(`[\x21\x23-\x2E\x30-\x5B\x5D-\x7E]+`).MatchString(value) {
+		errors = append(errors, fmt.Errorf("%q must satisfy regular expression pattern: [\\x21\\x23-\\x2E\\x30-\\x5B\\x5D-\\x7E]+", k))
```
bflad: Minor nitpick: We generally prefer to return friendlier validation errors to operators (e.g. `should only contain letters, numbers, ...`)

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("Failed setting schema: %s", err)
+	}
+
+	var scopeIdentifiers []string
+	for _, elem := range scopes {
+
+		scopeIdentifier := elem["scope_identifier"].(string)
+		scopeIdentifiers = append(scopeIdentifiers, scopeIdentifier)
+	}
+	d.Set("scope_identifiers", scopeIdentifiers)
```
bflad: Since this attribute is a non-scalar type, we should perform `d.Set()` error checking similar to scopes above.

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+		configs := v.(*schema.Set).List()
+		params.Scopes = expandCognitoResourceServerScope(configs)
+	}
+
+	log.Printf("[DEBUG] Creating Cognito Resource Server: %s", params)
+
+	resp, err := conn.CreateResourceServer(params)
+
+	if err != nil {
+		return errwrap.Wrapf("Error creating Cognito Resource Server: {{err}}", err)
```
bflad: Minor nitpick: We do not need to use `errwrap.Wrapf()` when simply returning a single error back to the operator -- `fmt.Errorf()` is preferred

aws/resource_aws_cognito_resource_server.go
```diff
@@ -0,0 +1,178 @@
...
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateCognitoResourceServerScopeDescription,
+						},
+						"scope_name": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validateCognitoResourceServerScopeName,
+						},
+						"scope_identifier": {
```
bflad: This attribute is a little awkward as its not defined by the API and is seemingly only used to build the computed `scope_identifiers` attribute below. Instead of creating this intermediate attribute, let's just directly create `scope_identifiers` from the scopes returned by the API to remove any confusion. 👍

bflad:
After addressing the above feedback and ensuring that updates properly worked for names and scopes (previously all scopes were always removed during update as they were not included in the update call), everything is green.

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSCognitoResourceServer'
TF_ACC=1 go test ./aws -v -run=TestAccAWSCognitoResourceServer -timeout 120m
=== RUN   TestAccAWSCognitoResourceServer_basic
--- PASS: TestAccAWSCognitoResourceServer_basic (24.04s)
=== RUN   TestAccAWSCognitoResourceServer_scope
--- PASS: TestAccAWSCognitoResourceServer_scope (42.89s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	66.971s
```

bflad:

tomjanne:
Thanks for the feedback and the help to get this done.

PR #4555, opened 2018-05-16
Fix incorrect resource reference for acm certificate

jmassey-datadog:

* Change the documentation to specify the `aws_acm_certificate` resource rather than `aws_acm_certificate_validation`. The `aws_acm_certificate_validation` resource does not export the `arn` or `certificate_arn` attribute.

### DIFF
--- a/website/docs/r/acm_certificate_validation.html.markdown
+++ b/website/docs/r/acm_certificate_validation.html.markdown
@@ -48,7 +48,7 @@ resource "aws_acm_certificate_validation" "cert" {
 
 resource "aws_lb_listener" "front_end" {
   # [...]
-  certificate_arn   = "${aws_acm_certificate_validation.cert.certificate_arn}"
+  certificate_arn   = "${aws_acm_certificate.cert.arn}"
 }
 ```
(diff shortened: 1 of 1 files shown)


bflad:
Hi @jmassey-datadog 👋 I'm not sure about the behavior you are seeing.

In Terraform, any provided arguments are automatically available as attributes. Also, the `aws_acm_certificate_validation` resource `certificate_arn` attribute is the recommended way to setup the following infrastructure creation ordering:
1. Create an ACM certificate
2. Wait until validation of the ACM certificate is completed (DNS or email)
3. Create downstream infrastructure (e.g. ELBv2 listeners) based on the now valid ACM certificate, otherwise they will generate an error

If you're not seeing this behavior, it sounds like a bug or an issue with the configuration implementation. Can you please explain what you are seeing?

jmassey-datadog:
Ahh I see the issue now. I assumed the `arn` was an attribute of `aws_acm_certificate_validation`. Whereas it should be `certificate_arn`. All good now. Thanks! No issues with the docs.

Issue #4540, opened 2018-05-15
Add s3 domain names with region so cloudfront origins outside of us-east-1 won't cause redirects immediately after setup


kitsunde:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Buckets created outside of us-east-1 will redirect to s3 when set as origins for cloudfront until DNS propagation has happened *unless* using the regional domain name. We need to set:

```hcl
${aws_s3_bucket.default.bucket}.s3-${aws_s3_bucket.default.region}.amazonaws.com
```

It would be nice if we could get a `regional_domain_name` or maybe `canonical_domain_name` from the `aws_s3_bucket` resource, and even make it recommended for origin since it's rather surprising especially if the redirects get cached.

### New or Affected Resource(s)`

* aws_s3_bucket

### Potential Terraform Configuration

This

```hcl
  origin {
    domain_name = "${aws_s3_bucket.default.domain_name}"
    origin_id = "s3-${var.environment}"
    s3_origin_config {
      origin_access_identity = "${aws_cloudfront_origin_access_identity.default.cloudfront_access_identity_path}"
    }
  }
```

Needs to be rewritten as:

```hcl
  origin {
    domain_name = "${aws_s3_bucket.default.bucket}.s3-${aws_s3_bucket.default.region}.amazonaws.com"
    origin_id = "s3-${var.environment}"
    s3_origin_config {
      origin_access_identity = "${aws_cloudfront_origin_access_identity.default.cloudfront_access_identity_path}"
    }
  }
```

### References

* https://forums.aws.amazon.com/thread.jspa?threadID=216814

saravanan30erd:
@bill-rich I would like to work on this issue.

saravanan30erd:
PR submitted: #4556

bflad:
The `bucket_regional_domain_name` attribute will be available in the `aws_s3_bucket` resource in version 1.21.0 of the AWS provider, releasing later today.

bflad:
[...]

PR #4556, opened 2018-05-16
Issue #4540 Add new attribute bucket_regional_domain_name in aws_s3_bucket

saravanan30erd:
Added new attribute bucket_regional_domain_name in aws_s3_bucket.

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -1085,7 +1085,7 @@ func resourceAwsS3BucketRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	// Add the bucket_regional_domain_name as an attribute
-	regionalEndpoint, err := bucketRegionalDomainName(d.Get("bucket").(string), region)
+	regionalEndpoint, err := BucketRegionalDomainName(d.Get("bucket").(string), region)
 	if err != nil {
 		return err
 	}
@@ -1454,7 +1454,7 @@ func bucketDomainName(bucket string) string {
 	return fmt.Sprintf("%s.s3.amazonaws.com", bucket)
 }
 
-func bucketRegionalDomainName(bucket string, region string) (string, error) {
+func BucketRegionalDomainName(bucket string, region string) (string, error) {
 	// https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region
 
 	for _, partition := range endpoints.DefaultPartitions() {
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform/helper/hashcode"
@@ -1084,7 +1085,11 @@ func resourceAwsS3BucketRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	// Add the bucket_regional_domain_name as an attribute
-	d.Set("bucket_regional_domain_name", bucketRegionalDomainName(d.Get("bucket").(string), region))
+	regionalEndpoint, err := bucketRegionalDomainName(d.Get("bucket").(string), region)
+	if err != nil {
+		return err
+	}
+	d.Set("bucket_regional_domain_name", regionalEndpoint)
 
 	// Add the hosted zone ID for this bucket's region as an attribute
 	hostedZoneID, err := HostedZoneIDForRegion(region)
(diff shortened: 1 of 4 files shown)


aws/resource_aws_s3_bucket_test.go
```diff
@@ -126,6 +126,27 @@ func TestAccAWSS3Bucket_region(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSS3BucketDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSS3BucketConfigRegionalDomainName(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSS3BucketExists("aws_s3_bucket.bucket"),
+					resource.TestCheckResourceAttr("aws_s3_bucket.bucket", "region", "ap-south-1"),
+					resource.TestCheckResourceAttr(
```
bflad: As mentioned in the other PR with a `Computed: true` attribute that is always available, this can be moved as a single check under the `_basic` test.

We also explicitly need the S3 testing to be able to run from AWS Commercial, AWS GovCloud (US), and AWS China which means that we cannot hard code a specific partition region. I believe there is a `testAccProviderRegion()` helper function which can aid us here.

If we want to ensure that specific regions are returning their expected domain names (e.g. `cn-north-1` actually returning `BUCKET.s3.cn-north-1.amazonaws.com.cn`) I would highly recommend setting up unit testing for `bucketRegionalDomainName` as well.

website/docs/r/s3_bucket.html.markdown
```diff
@@ -450,6 +450,7 @@ The following attributes are exported:
 * `id` - The name of the bucket.
 * `arn` - The ARN of the bucket. Will be of format `arn:aws:s3:::bucketname`.
 * `bucket_domain_name` - The bucket domain name. Will be of format `bucketname.s3.amazonaws.com`.
+* `bucket_regional_domain_name` - The bucket region-specific domain name. Will be of format `bucketname.s3-eu-west-1.amazonaws.com`. Note: The AWS CloudFront allows specifying S3 region-specific endpoint when creating S3 origin, it will prevent [redirect issues](https://forums.aws.amazon.com/thread.jspa?threadID=216814) from CloudFront to S3 Origin URL.
```
bflad: Rather than specifying `bucketname.s3-eu-west-1.amazonaws.com` here we should probably instead point to the [AWS regions and endpoints documentation](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region) mentioning the domain name including the region name.

aws/resource_aws_s3_bucket.go
```diff
@@ -1441,6 +1449,18 @@ func bucketDomainName(bucket string) string {
 	return fmt.Sprintf("%s.s3.amazonaws.com", bucket)
 }
 
+func bucketRegionalDomainName(bucket string, region string) string {
```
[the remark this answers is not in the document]
saravanan30erd: used `endpoints` package to retrieve region-specific endpoint.

aws/resource_aws_s3_bucket_test.go
saravanan30erd: Since I need to create separate test config, I used separate acceptance test `TestAccAWSS3Bucket_bucketRegionalDomainName`. I couldn't find `testAccProviderRegion()`, so I used the same `endpoints` package to choose region.

website/docs/r/s3_bucket.html.markdown
saravanan30erd: 👍

saravanan30erd:
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSS3Bucket_bucketRegionalDomainName'
TF_ACC=1 go test ./aws -v -run=TestAccAWSS3Bucket_bucketRegionalDomainName -timeout 120m
=== RUN   TestAccAWSS3Bucket_bucketRegionalDomainName
--- PASS: TestAccAWSS3Bucket_bucketRegionalDomainName (67.64s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	67.687s

aws/resource_aws_s3_bucket_test.go
```diff
@@ -1331,6 +1356,26 @@ func testAccBucketDomainName(randInt int) string {
 	return fmt.Sprintf("tf-test-bucket-%d.s3.amazonaws.com", randInt)
 }
 
+func testAccFindRandomRegion() string {
```
bflad: Generally in our acceptance testing we do not want to insert randomness as it leads to inconsistent testing behavior that makes troubleshooting harder than it needs to be.

Aside from that, as written this currently will enumerate a list of regions across all AWS partitions (Commercial, GovCloud (US), and China) while the testing credentials are only valid for one partition.

saravanan30erd: 👍

saravanan30erd:
@bflad since I couldn't find `testAccProviderRegion()`, I thought it's better to provide region as input in separate test config and test the same. Now `testAccGetRegion()` fixed all the issues.

$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSS3Bucket_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSS3Bucket_basic -timeout 120m
=== RUN   TestAccAWSS3Bucket_basic
--- PASS: TestAccAWSS3Bucket_basic (104.41s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	104.451s

saravanan30erd:
@bflad  is there anything I can do to push this?

bflad:
@saravanan30erd there is currently a merge conflict in `aws/resource_aws_s3_bucket_test.go` and I'd like to add some quick unit testing for `BucketRegionalDomainName`. Fixing these on merge.

bflad:

PR #4568, opened 2018-05-17
r/lb_target_group health chk proto chg req taint

### LINKED ISSUE #4150: Changing health check on TCP listener requires resource to be deleted and re-created
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
```
Terraform v0.11.1
+ provider.aws v1.5.0
``

### Affected Resource(s)
- aws_lb_target_group

```hcl
resource "aws_lb_target_group" "http" {
  name                 = "${ local.resource_name }-80"
  port                 = 80
  protocol             = "TCP"
  vpc_id               = "${ module.network.vpc_id }"
  deregistration_delay = 300

  health_check {
    interval            = 10
    port                = 443
    protocol            = "HTTPS"
    path                = "/health"
    healthy_threshold   = 2
    unhealthy_threshold = 2
    matcher             = "200-399"
  }
}

resource "aws_lb_listener" "http" {
  load_balancer_arn = "${ aws_lb.main.arn }"
  port              = 80
  protocol          = "TCP"

  default_action = {
    target_group_arn = "${ aws_lb_target_group.http.arn }"
    type             = "forward"
  }
}
```

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
AWS Currently does not let you change the health check type from one kind (e.g. TCP, HTTP, HTTPS) to another; the target group resource must be deleted and re-created

### Actual Behavior
Terraform received this error from the AWS API:
> Error modifying Target Group: InvalidConfigurationRequest: You cannot change the health check protocol for a target group with the TCP protocol

### Steps to Reproduce
1. Configure aws_lb_listener as TCP
2. Configure aws_lb_target_group with TCP health check
3. Run `terraform apply`
4. Change aws_lb_target_group created in step 2 above to have a HTTPS health check
5. Run `terraform apply`
6. Receive error.

### Important Factoids
N/A

### References
https://github.com/terraform-providers/terraform-provider-aws/issues/2708#issuecomment-357355482

iancward:
Notify users of TCP lb target groups that changes to health check protocol
require the user to taint the resource and let terraform re-create it.
This commit Fixes #4150.


aws/resource_aws_lb_target_group.go
```diff
@@ -631,6 +631,12 @@ func resourceAwsLbTargetGroupCustomizeDiff(diff *schema.ResourceDiff, v interfac
 	}
 
 	if protocol == "TCP" {
+		if diff.HasChange("health_check.0.protocol") {
+			old, new := diff.GetChange("health_check.0.protocol")
+			return fmt.Errorf("Health check protocol cannot be updated from %d to %d for TCP based Target Group %s"+
```
bflad: Is there a reason to not just call `diff.ForceNew("health_check.0.protocol")` here? https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceDiff.ForceNew

For what its worth, it looks like the others should be the same way to just perform the `(forces new resource)` automatically. 😄

bflad: This also seems to be triggering an acceptance test failure as-is:

```
=== RUN   TestAccAWSLBTargetGroup_stickinessWithTCPDisabled
--- FAIL: TestAccAWSLBTargetGroup_stickinessWithTCPDisabled (70.19s)
    testing.go:518: Step 0 error: Error on follow-up plan: 1 error(s) occurred:
        
        * aws_lb_target_group.test: 1 error(s) occurred:
        
        * aws_lb_target_group.test: Health check protocol cannot be updated from %!d(string=TCP) to %!d(string=HTTP) for TCP based Target Group arn:aws:elasticloadbalancing:us-west-2:*******:targetgroup/tf-20180518010652276200000001/c4d5d741fa72a2ad use 'terraform taint' to recreate the resource if you wish
```

Looking through the last few weeks of acceptance testing, it was previously passing.

iancward: Sorry for the delay on the response. 
Are you meaning the `diff.ForceNew` option in the schema? Most of these messages are in conditionals that only apply for a TCP load balancer type (network LB). I believe this resource is also used for application load balancers (http or https), so I'd have to see if the same restrictions apply there.

iancward: Ah, it looks like I can do something like this: https://github.com/terraform-providers/terraform-provider-aws/blob/37860355dc6210a987819c7d79a699f2b56e15d8/aws/resource_aws_api_gateway_authorizer.go#L263-L265

I guess the only thing to consider is that this kind of change will destroy and re-recreate the load balancer target group, which the end-user might not expect. I assume we trust the end-user to do a `plan` first and notice the change?

bflad: Operators should be checking `plan` first; that is part of the normal Terraform workflow. Really though, I'm more concerned about the new acceptance test failure in `TestAccAWSLBTargetGroup_stickinessWithTCPDisabled` caused by this change. Can you determine why it was successful before and ensure it continues to work?

iancward: I'm trying to run the tests, but I get an error about a miss-matched type. Do you have any ideas?
```
➜  terraform-provider-aws git:(GH-4150) ✗ go test ./... -timeout=30s -parallel=2
# github.com/iancward/terraform-provider-aws
./main.go:10:15: cannot use "github.com/iancward/terraform-provider-aws/vendor/github.com/terraform-providers/terraform-provider-aws/aws".Provider (type func() "github.com/iancward/terraform-provider-aws/vendor/github.com/terraform-providers/terraform-provider-aws/vendor/github.com/hashicorp/terraform/terraform".ResourceProvider) as type "github.com/iancward/terraform-provider-aws/vendor/github.com/hashicorp/terraform/plugin".ProviderFunc in field value
```

iancward: Well, it looks like others have had the same issue https://github.com/terraform-providers/terraform-provider-aws/issues/2077

bflad: @iancward I believe that happens when the codebase lives in a personal directory (`github.com/iancward/terraform-provider-aws`) instead of `github.com/terraform-providers/terraform-provider-aws`

FYI we're hanging out in https://gitter.im/hashicorp-terraform/gardening/ today for the community gardening event if you'd like some more realtime help (or to help close some bugs, including this one!)

ricoli:
any updates on this?

ricoli:
@bflad ?

gordonbondon:
Would be great to get this merged!

iancward:
Thanks! Apologies for not spending the time to sort out the acceptance tests.

PR #4591, opened 2018-05-18
fix for issue in PR #4511

saravanan30erd:
Fixes #4511 

@bflad I believe there will be better fix for these cases, please advise.

### DIFF
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -383,9 +383,11 @@ func buildSpotFleetLaunchSpecification(d map[string]interface{}, meta interface{
 		}
 	}
 
-	if v, ok := d["iam_instance_profile_arn"]; ok {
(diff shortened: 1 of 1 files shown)


saravanan30erd:
@bflad Done, I thought there may be some conflict functions, we can use to prevent these cases.

bflad:
`ConflictsWith` in the schema will prevent the configuration of two conflicting attributes, but does not prevent the `schema.TypeString` zero-value of `""` from still being the value of the attribute. Generally its up to the downstream implementation (like the change here) to worry or not worry about the zero-value.

bflad:
This has been released in [version 1.20.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1200-may-23-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #4605, opened 2018-05-22
travis: Run test jobs in parallel

radeksimko:
**DO NOT MERGE**

-----

This is a test PR I'm going to use to experiment with Travis' Build Stages to see if we could potentially run our make targets in parallel and speed up the success/failure reporting.

https://docs.travis-ci.com/user/build-stages/#What-are-Build-Stages%3F

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,12 @@ install:
 - bash scripts/gogetcookie.sh
 - go get github.com/kardianos/govendor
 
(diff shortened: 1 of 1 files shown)


radeksimko:
Created a separate Asana task for this as it will require a bit more work, e.g. inspecting the significant overhead for each job (~42 seconds).

Issue #4611, opened 2018-05-22
Feature Request - Support Secrets Manager Key/Value in addition to String


rayterrill:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support for Secrets Manager Key/Value in addition to String/plaintext type

### New or Affected Resource(s)

* aws_secretsmanager_secret_version

### Potential Terraform Configuration

Unknown.

### References

https://docs.aws.amazon.com/secretsmanager/latest/userguide/terms-concepts.html

bflad:
@rayterrill does the API provide a different method for storing key/value pairs? The [`SecretString` parameter to the `PutSecretValue` function of the Go SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#PutSecretValueInput) has the following documentation:

```
    // (Optional) Specifies text data that you want to encrypt and store in this
    // new version of the secret. Either SecretString or SecretBinary must have
    // a value, but not both. They cannot both be empty.
    //
    // If you create this secret by using the Secrets Manager console then Secrets
    // Manager puts the protected secret text in only the SecretString parameter.
    // The Secrets Manager console stores the information as a JSON structure of
    // key/value pairs that the default Lambda rotation function knows how to parse.
    //
    // For storing multiple values, we recommend that you use a JSON text string
    // argument and specify key/value pairs. For information on how to format a
    // JSON parameter for the various command line tool environments, see Using
    // JSON for Parameters (http://docs.aws.amazon.com/cli/latest/userguide/cli-using-param.html#cli-using-param-json)
    // in the AWS CLI User Guide.
    SecretString *string `type:"string"`
```

Seemingly you should be able to define a textual JSON string of key/value pairs to accomplish your goal:

```hcl
resource "aws_secretsmanager_secret_version" "example" {
  secret_id     = "${aws_secretsmanager_secret.example.id}"
  secret_string = "{\"key1\": \"value1\"}"
}
```

We can probably improve the documentation to highlight this.

bflad:
[...]

PR #4613, opened 2018-05-22
docs/resource/aws_secretsmanager_secret_version: Provide example key-value pair usage

bflad:
Reference #4611

* Add example key-value pair usage

### DIFF
--- a/website/docs/r/secretsmanager_secret_version.html.markdown
+++ b/website/docs/r/secretsmanager_secret_version.html.markdown
@@ -14,10 +14,34 @@ Provides a resource to manage AWS Secrets Manager secret version including its s
 
 ## Example Usage
 
+### Simple String Value
+
 ```hcl
 resource "aws_secretsmanager_secret_version" "example" {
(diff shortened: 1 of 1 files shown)


bflad:
> would it make sense to add a new property to aws_secretsmanager_secret_version for secret_map?

https://docs.aws.amazon.com/secretsmanager/latest/userguide/terms-concepts.html currently states:

> When storing credentials, different secured services might require different pieces of information. AWS Secrets Manager provides this flexibility by storing the secret as key-value pairs of text strings. 

So theoretically we could support a second argument of `schema.TypeMap` with `schema.TypeString` elements that just did the JSON encoding for you. I'd be a little concerned about:
* The Secrets Manager service supporting other native (non-string) JSON value types in the future (I do not think the Terraform schema can support non-strictly typed elements)
* Downstream modules would only be able to support one or the other arguments (presumably the new argument would have `ConflictsWith`) so they'd likely have to opt for the existing string one either which way, which deflates its usefulness

So certainly not a hard no, but we generally do not have too many precedents for creating two arguments to the same API parameter either. 👍

PR #4654, opened 2018-05-25
Respect lambda function qualifier on data source.

jmcarp:
As written, the lambda function data source always returns the most
recent version and qualified arn, even if a different qualifier is set.
This patch respects the qualifier if provided.

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -527,27 +527,33 @@ func resourceAwsLambdaFunctionRead(d *schema.ResourceData, meta interface{}) err
 		})
 	}
 
-	// List is sorted from oldest to latest
-	// so this may get costly over time :'(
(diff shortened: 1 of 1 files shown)


jen20:
Hi @jmcarp! Thanks for submitting a pull request. This _does_ look like a correct change to me, however it doesn't match the pull request description, which describes the _data source_, whereas the change is to the _resource_ variant of the Lambda function. Is this a change which needs propagating across to the data source also?

jmcarp:
@jen20: this does affect the data source, since the resource and data source share the same read function. It might make sense to refactor to two separate functions for clarity, but I figured I'd try to keep the patch simple to start with.

jen20:
Ah yes, I see now. I'm not particularly sure I like the pattern of multiple resources sharing read functions, but this does fix the issue and that refactor can be made another time. Thanks for contributing!

bflad:

Issue #3560, opened 2018-02-28
Add ability to import a certificate with aws_acm_certificate


grimm26:
Please expand the `aws_acm_certificate` resource to support the ability to import an existing in-hand certificate, a la https://docs.aws.amazon.com/sdk-for-go/api/service/acm/#ACM.ImportCertificate

bflad:
Interesting idea. Maybe this could be implemented via:
* New `certificate`, `certificate_chain`, and `private_key` attributes
* Use `CustomizeDiff` to fail plan time validation if one of the new attributes is defined but another one is missing OR if `validation_method` is missing
* Set `Optional: true` on `validation_method` attribute
* Set `ConflictsWith: []string{"private_key"}` on existing attributes `domain_name`, `subject_alternative_names`, and `validation_method`
* In the Create function, if `d.GetOk("private_key")` then call ImportCertificate

For the acceptance testing, we could probably get away with using the TLS provider to generate self-signed certificates to import

mouellet-coveo:
This would be a really nice improvement since ACM seems to support the in-place update of the certificate.

thisisthetechie:
Yes please, I'd like the ability to import my Certificates using Terraform, otherwise it's a mess of having to perform an outside API call to do it...

mcolley73:
Yes, please. Otherwise we have to bounce out of terraform to do this elsewhere and then come back.

bflad:
This is not a slight against anyone who posted here (I promise!), but we are now including the following note in our feature request issue template for the repository:

---

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

---

I have a feeling this probably would garner more community/maintainer attention if it were not for the current lack of 👍  reactions on the original issue above. The feature request certainly seems valuable. Unfortunately the upvotes on my comment do not show up in the top level GitHub issue sorting. 😄 

As I'm writing this I now feel like we should go back through all the old issues and add that note. It'd
 probably be very helpful especially for those not actively working with this project. 😅

pemcconnell:
per:

>If you are interested in working on this issue or have submitted a pull request, please leave a comment

[...]

PR #4666, opened 2018-05-26
issue #4658 update validation_method attribute schema in aws_acm_certificate

### LINKED ISSUE #4658: importing existing acm certificate in terraform with NONE validation_method
### Community Note

### Terraform Version
* Terraform v0.11.6
* aws_1.20.0

### Affected Resource(s)

* aws_acm_certificate

```hcl
resource "aws_acm_certificate" "my_cert" {
  domain_name       = "*.some.domain"
  validation_method = "NONE"

  tags {
    Creator     = "Terraform"
    Name        = "my-cert"
    Environment = "${var.env}"
  }
}
```

### Expected Behavior

Creates the acm certificate in Terraform based on existing ACM certificate using NONE as validation_method as specified in documentation https://www.terraform.io/docs/providers/aws/r/acm_certificate.html

* validation_method - (Required) Which method to use for validation. DNS or EMAIL are valid, NONE can be used for certificates that were imported into ACM and then into Terraform.

### Actual Behavior

Throws an error

* aws_acm_certificate.my_cert: Error requesting certificate: ValidationException: 1 validation error detected: Value '' at 'validationMethod' failed to satisfy constraint: Member must satisfy enum value set: [DNS, EMAIL]

According to https://github.com/terraform-providers/terraform-provider-aws/blob/80e4ae53941c6fc87239f1d704e2dfe6a4dec3b6/aws/resource_aws_acm_certificate.go#L166 NONE is returned if neither DNS or EMAIL are specified but even if an empty string is set as value for validation_method argument, same error as above. Removing the argument is not possible as it is required.

### Steps to Reproduce

1. `terraform apply`

terraform plan show nothing wrong by the way.

saravanan30erd:
Fixes #4658 

* Change 1
change the validation_method attribute as optional.
https://docs.aws.amazon.com/acm/latest/APIReference/API_RequestCertificate.html

### DIFF
--- a/aws/resource_aws_acm_certificate.go
+++ b/aws/resource_aws_acm_certificate.go
@@ -35,7 +35,7 @@ func resourceAwsAcmCertificate() *schema.Resource {
 			},
 			"validation_method": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 			"arn": {
--- a/website/docs/r/acm_certificate.html.markdown
+++ b/website/docs/r/acm_certificate.html.markdown
@@ -41,7 +41,7 @@ The following arguments are supported:
 
 * `domain_name` - (Required) A domain name for which the certificate should be issued
 * `subject_alternative_names` - (Optional) A list of domains that should be SANs in the issued certificate
-* `validation_method` - (Required) Which method to use for validation. `DNS` or `EMAIL` are valid, `NONE` can be used for certificates that were imported into ACM and then into Terraform.
+* `validation_method` - (Optional) Which method to use for validation. `DNS` or `EMAIL` are valid, `NONE` can be used for certificates that were imported into ACM and then into Terraform.
 * `tags` - (Optional) A mapping of tags to assign to the resource.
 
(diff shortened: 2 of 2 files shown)


bflad:
@saravanan30erd I'm not sure changing this attribute from `Required: true` to `Optional: true` is a good or correct idea at the moment. Currently we only support the following scenarios with the `aws_acm_certificate` resource:

* Creating a certificate with the Terraform resource, which requires either DNS or email verification. I'm guessing without specifying the validation method it'll default to email verification as that was the only verification method originally (or maybe the API errors when the method is not defined in this case, I have not tried it in awhile).
* Importing an existing certificate into ACM, then importing the resource into Terraform. We currently use the `NONE` verification method here.

I believe the intent of #4658 is actually the following scenario, which we currently do not support, but is already a separate feature request #3560:

* Importing a existing certificate with the Terraform resource

So I believe the right answer is we should implement `ImportCertificate` in the resource before a change from `Required: true` should happen. Thoughts?

saravanan30erd:
Yes, we should implement `ImportCertificate ` before this change. I would like to work on #3560 , can I ?

PR #4676, opened 2018-05-29
Change gofmt arguments backquotes to xargs

Hiroyuki-Nagata:

* This resolves error `Argument list too long` caused by gofmt

I have tested building terraform-provider-aws with msys2 on windows. Then I got the error "gofmt: Argument list too long". This is caused by file counts is larger than `ARG_MAX` defined by using OS.


Hiroyuki-Nagata:
Thanks a lot !

Issue #4624, opened 2018-05-23
Add capability to reference the log group name of an aws_cloudwatch_log_group resource


thomasbiddle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

https://www.terraform.io/docs/providers/aws/r/cloudwatch_log_group.html

Currently the aws_cloudwatch_log_group is able to have it's ARN referenced; but not it's name. The name is part of the input; but it would be wonderful to be able to reference it as an attribute so that it could keep code DRY later on and ensure it's tied directly to the resource.

### New or Affected Resource(s)

* aws_cloudwatch_log_group

### Potential Terraform Configuration

Ability to reference as such:
```hcl
aws_cloudwatch_log_group.my_log_group.name
```

### References

Not that I could find.

bflad:
@thomasbiddle some of our documentation pages aren't necessarily clear about this, but all provided resource arguments and some optional ones are available as attributes. Have you tried using `"${aws_cloudwatch_log_group.my_log_group.name}"` somewhere and it does not work?

At some point we will be automatically generating the documentation, but I can submit the documentation page fix here to change the attributes verbiage to `The following additional attributes are exported:`

bflad:
All the resource pages will be updated this week to include the updated wording for the attributes section. Thanks for raising this issue!

PR #4685, opened 2018-05-29
docs: Clarify attributes as additional to arguments

bflad:
Reference (among many others): #4624 

* Change attributes section verbiage to clarify that these attributes are in addition to arguments


bflad:
Let's rip off the bandaid on this one 😄

PR #4694, opened 2018-05-29
r/aws_sns_sms_preferences Fix documentation omission

hwrdprkns:
`r/sns_sms_peferences` is missing from the documentation and also the filename was spelled incorrectly.


bflad:

PR #4697, opened 2018-05-30
Fix force new behavior for aws_batch_job_definition timeout

hectcastro:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/4696

When the `timeout.attempt_duration_seconds` attribute of the `aws_batch_job_definition` is modified, ensure that a new resource is created vs. update-in-place (update-in-place is not support for this resource).

* Add `ForceNew` to `aws_batch_job_definition.timeout.attempt_duration_seconds`


bflad:

Issue #4707, opened 2018-05-31
Support authenticate-oidc and authenticate-cognito for elb_listener


atsushi-ishibashi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

ELB Listener supports new actions to authenticate  token.

### New or Affected Resource(s)

* aws_lb_listener

### Potential Terraform Configuration

```hcl
resource "aws_lb_listener" "front_end" {

  ...

  default_action {
    authenticate_cognito_config {
      ... // modify after research
    }
    order = 1
    type = "authenticate-cognito"
  }
  default_action {
    target_group_arn = "${aws_lb_target_group.front_end.arn}"
    type             = "forward"
  }
}
```

### References

* #4706

bflad:
We should probably create a separate issue to support this in `aws_lb_listener_rule` as well. 😄

bflad:
Thanks to @atsushi-ishibashi and @mikael-lindstrom, initial support for Cognito and OIDC authentication in `aws_lb_listener` and `aws_lb_listener_rule` has been merged in and will release with version 1.40.0 of the AWS provider, likely middle of this week. 🎉

bflad:
This has been released in [version 1.40.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1400-october-10-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #4722, opened 2018-06-01
[WIP, Help]r/lb_listener: Support authenticate action for lb listener

atsushi-ishibashi:
closed: #4707 

- [ ] testacc
- [x] docs

* `aws_lb_listener` supports authenticate action

### DIFF
--- a/aws/resource_aws_lb_listener.go
+++ b/aws/resource_aws_lb_listener.go
@@ -270,12 +270,12 @@ func resourceAwsLbListenerCreate(d *schema.ResourceData, meta interface{}) error
 					action.TargetGroupArn = aws.String(v)
 				}
 			case elbv2.ActionTypeEnumAuthenticateOidc:
-				cfgs := d.Get("authenticate_oidc_config").(*schema.Set).List()
+				cfgs := defaultActionMap["authenticate_oidc_config"].(*schema.Set).List()
 				if len(cfgs) > 0 {
 					action.AuthenticateOidcConfig = expandELbAuthenticateOidcActionConfig(cfgs[0].(map[string]interface{}))
 				}
 			case elbv2.ActionTypeEnumAuthenticateCognito:
-				cfgs := d.Get("authenticate_cognito_config").(*schema.Set).List()
+				cfgs := defaultActionMap["authenticate_cognito_config"].(*schema.Set).List()
 				if len(cfgs) > 0 {
 					action.AuthenticateCognitoConfig = expandELbAuthenticateCognitoActionConfig(cfgs[0].(map[string]interface{}))
 				}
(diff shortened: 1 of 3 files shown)


atsushi-ishibashi:
```
TF_ACC=1 go test ./aws -v -run=TestAccAWSLBListener_cognito -timeout 120m
=== RUN   TestAccAWSLBListener_cognito
--- FAIL: TestAccAWSLBListener_cognito (214.14s)
		
		* aws_lb_listener.test: 1 error(s) occurred:
		
		* aws_lb_listener.test: Invalid address to set: []string{"default_action", "0", "authenticate_cognito_config"}
		
		Error: Error refreshing: 1 error(s) occurred:
		
		* aws_lb_listener.test: 1 error(s) occurred:
		
		* aws_lb_listener.test: aws_lb_listener.test: Invalid address to set: []string{"default_action", "0", "authenticate_cognito_config"}
		
		State: <nil>
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	214.173s
make: *** [testacc] Error 1
```
I got the above error. The below is the return of `flattenElbActions(listener.DefaultActions)`
```
[]interface {}{map[string]interface {}{"order":1, "type":"authenticate-cognito", "authenticate_cognito_config":[]map[string]interface {}{map[string]interface {}{"session_cookie_name":"AWSELBAuthSessionCookie", "session_timeout":604800, "user_pool_arn":"arn:aws:cognito-idp:us-west-2:hogehoge:userpool/us-west-2_nxECAEbsn", "user_pool_client":"3ri67ugsiniu79j3v5tk58ttlh", "user_pool_domain":"e67po", "on_unauthenticated_request":"authenticate", "scope":"openid"}}}, map[string]interface {}{"order":2, "type":"forward", "target_group_arn":"arn:aws:elasticloadbalancing:us-west-2:hogehoge:targetgroup/e67po/369f21393b0ae80b"}}
```

I couldn't find incorrect😅 Please correct me 🙇

mikael-lindstrom:
I tried to figure out this problem and it seems to be that `flattenELbAuthenticateCognitoActionConfig` returns `[]map[string]interface{}` when it expects `schema.Set`. Since I didn't manage to figure out how to fix this in a nice way and since it has `MaxItems: 1` it can be converted to `schema.TypeList` instead. I tested this and it seems to work fine but I found another issue.

It seems that although `DescribeListeners` returns actions in a consistent order, the order they are saved in AWS is seemingly random. Since `default_action` currently is `schema.TypeList` this sometimes causes diffs when running apply multiple times. It could probably be solved by changing `default_action` to `schema.Set` but that would mean we need to handle state migration to be backwards compatible. Instead sorting the actions from `DescribeListeners` based on the order saved in the state seems easier.

For oidc I found that `client_secret` isn't returned from the API since its a secret, so to prevent this from showing as modified every apply we need to use whatever is already stored in the state.

I have a working version that I'm using now and have pushed the code [here](https://github.com/mikael-lindstrom/terraform-provider-aws/tree/issue4707) with added a acceptance test for odic. Let me know if this works for you.

mikael-lindstrom:
@atsushi-ishibashi would you mind taking a look at this and let me know what you think?

mikael-lindstrom:
@bflad Is there anything I can do to speed this up? If you like I can open a new PR with my changes if @atsushi-ishibashi is busy.

bflad:
This PR has been superseded by #6094, which was just merged and will release with version 1.40.0 of the AWS provider, likely middle of this week. Thanks so much to @atsushi-ishibashi and @mikael-lindstrom for their _huge_ efforts here!

PR #4723, opened 2018-06-01
Fix page header and the example code's indentation

KaanErturk:
Page header was showing the wrong resource name, probably forgotten after a copy&paste from another resource documentation.

The indentation in the example code was not right as it was affecting the highlighting.

There was an extra curly bracket in the example code.


KaanErturk:
This is only a documentation fix and I think it's worth pushing it to the website sooner rather than later.

Issue #4713, opened 2018-05-31
Support Neptune


atsushi-ishibashi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* aws_neptune_cluster
* aws_neptune_cluster_instance
* aws_neptune_subnet_group
* aws_neptune_parameter_group
* aws_neptune_cluster_parameter_group
* aws_neptune_event_subscription

### Potential Terraform Configuration

```hcl
resource "aws_neptune_cluster" "example" {
  cluster_identifier = "example"
  engine = "neptune"
  engine_version = "1.0.1"
  cluster_parameter_group_name = "${aws_neptune_cluster_parameter_group.example.name}"
  subnet_group_name = "${aws_neptune_subnet_group.example.name}"
  database_name = "database"
  master_username = "user"
  master_password = "password"
  ...

}

resource "aws_neptune_cluster_instance" "example" {
  identifier = "example"
  cluster_identifier = "${aws_neptune_cluster.example.id}"
  engine = "neptune"
  instance_class = "db.m4.large"
  ...

}

resource "aws_neptune_subnet_group" "example" {
  name = "example"
  description = "example description"
  subnet_ids = ["${aws_subnet.frontend.id}", "${aws_subnet.backend.id}"]
}

resource "aws_neptune_parameter_group" "example" {
  name = "example"
  family = "example"
  description = "example description"
}

resource "aws_neptune_cluster_parameter_group" "example" {
  name = "example"
  family = "example"
  description = "example description"
}

resource "aws_neptune_event_subscription" "example" {
  name = "example"
  sns_topic = "${aws_sns_topic.example.arn}"
  source_type = "db-instance"
  source_ids = ["${aws_neptune_cluster_instance.example.id}"]
  event_categories = [
    "creation",
    "failover",
    ...
  ]
}
```

### Order

1. `aws_neptune_parameter_group`,`aws_neptune_cluster_parameter_group`,`aws_neptune_subnet_group`
2. `aws_neptune_cluster`
3. `aws_neptune_cluster_instance`
4. `aws_neptune_event_subscription`

[...]

PR #4724, opened 2018-06-01
Resource - Neptune: aws_neptune_parameter_group

dav009:
Reference: #4713

* adding a new resource `aws_neptune_parameter_group` as a first step to address #4713

```
$ make testacc TESTARGS='-run=TestAccAWSNeptuneParameterGroup_basic'
=== RUN   TestAccAWSNeptuneParameterGroup_basic
--- PASS: TestAccAWSNeptuneParameterGroup_basic

$ make testacc TESTARGS='-run=TestAccAWSNeptuneParameterGroup_removeParam'
=== RUN   TestAccAWSNeptuneParameterGroup_removeParam
--- PASS: TestAccAWSNeptuneParameterGroup_removeParam

$ make testacc TESTARGS='-run=TestAccAWSNeptuneParameterGroup_only'
=== RUN   TestAccAWSNeptuneParameterGroup_only
--- PASS: TestAccAWSNeptuneParameterGroup_only
```

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -474,6 +474,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_nat_gateway":                              resourceAwsNatGateway(),
 			"aws_network_acl":                              resourceAwsNetworkAcl(),
 			"aws_default_network_acl":                      resourceAwsDefaultNetworkAcl(),
+			"aws_neptune_parameter_group":                  resourceAwsNeptuneParameterGroup(),
 			"aws_network_acl_rule":                         resourceAwsNetworkAclRule(),
 			"aws_network_interface":                        resourceAwsNetworkInterface(),
 			"aws_network_interface_attachment":             resourceAwsNetworkInterfaceAttachment(),
--- a/aws/resource_aws_neptune_parameter_group.go
+++ b/aws/resource_aws_neptune_parameter_group.go
@@ -126,6 +126,10 @@ func resourceAwsNeptuneParameterGroupRead(d *schema.ResourceData, meta interface
 		return err
 	}
 
+	if describeResp == nil {
+		return fmt.Errorf("Unable to get Describe Response for Neptune Parameter Group (%s)", d.Id())
+	}
+
 	if len(describeResp.DBParameterGroups) != 1 ||
 		*describeResp.DBParameterGroups[0].DBParameterGroupName != d.Id() {
 		return fmt.Errorf("Unable to find Parameter Group: %#v", describeResp.DBParameterGroups)
--- a/aws/resource_aws_neptune_parameter_group.go
+++ b/aws/resource_aws_neptune_parameter_group.go
@@ -1,0 +1,266 @@
+package aws
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/neptune"
+)
+
+func resourceAwsNeptuneParameterGroup() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsNeptuneParameterGroupCreate,
+		Read:   resourceAwsNeptuneParameterGroupRead,
+		Update: resourceAwsNeptuneParameterGroupUpdate,
+		Delete: resourceAwsNeptuneParameterGroupDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+				StateFunc: func(val interface{}) string {
+					return strings.ToLower(val.(string))
+				},
+			},
+			"family": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"description": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  "Managed by Terraform",
+			},
+			"parameter": &schema.Schema{
+				Type:     schema.TypeSet,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"value": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"apply_method": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "immediate",
+							// this parameter is not actually state, but a
+							// meta-parameter describing how the RDS API call
+							// to modify the parameter group should be made.
+							// Future reads of the resource from AWS don't tell
+							// us what we used for apply_method previously, so
+							// by squashing state to an empty string we avoid
+							// needing to do an update for every future run.
+							StateFunc: func(interface{}) string { return "" },
+						},
+					},
+				},
+				Set: resourceAwsNeptuneParameterHash,
+			},
+		},
+	}
+}
+
+func resourceAwsNeptuneParameterGroupCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).neptuneconn
+
+	createOpts := neptune.CreateDBParameterGroupInput{
+		DBParameterGroupName:   aws.String(d.Get("name").(string)),
+		DBParameterGroupFamily: aws.String(d.Get("family").(string)),
+		Description:            aws.String(d.Get("description").(string)),
+	}
+
+	log.Printf("[DEBUG] Create Neptune Parameter Group: %#v", createOpts)
+	resp, err := conn.CreateDBParameterGroup(&createOpts)
+	if err != nil {
+		return fmt.Errorf("Error creating Cache Parameter Group: %s", err)
+	}
+
+	d.Partial(true)
+	d.SetPartial("name")
+	d.SetPartial("family")
+	d.SetPartial("description")
+	d.Partial(false)
+
+	d.SetId(*resp.DBParameterGroup.DBParameterGroupName)
+	log.Printf("[INFO] Neptune Parameter Group ID: %s", d.Id())
+
+	return resourceAwsNeptuneParameterGroupUpdate(d, meta)
+}
+
+func resourceAwsNeptuneParameterGroupRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).neptuneconn
+
+	describeOpts := neptune.DescribeDBParameterGroupsInput{
+		DBParameterGroupName: aws.String(d.Id()),
+	}
+
+	describeResp, err := conn.DescribeDBParameterGroups(&describeOpts)
+	if err != nil {
+		return err
+	}
+
+	if len(describeResp.DBParameterGroups) != 1 ||
+		*describeResp.DBParameterGroups[0].DBParameterGroupName != d.Id() {
+		return fmt.Errorf("Unable to find Parameter Group: %#v", describeResp.DBParameterGroups)
+	}
+
+	d.Set("name", describeResp.DBParameterGroups[0].DBParameterGroupName)
+	d.Set("family", describeResp.DBParameterGroups[0].DBParameterGroupFamily)
+	d.Set("description", describeResp.DBParameterGroups[0].Description)
+
+	// Only include user customized parameters as there's hundreds of system/default ones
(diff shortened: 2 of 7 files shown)


aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+func resourceAwsNeptuneParameterGroupRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).neptuneconn
+
+	describeOpts := neptune.DescribeDBParameterGroupsInput{
+		DBParameterGroupName: aws.String(d.Id()),
+	}
+
+	describeResp, err := conn.DescribeDBParameterGroups(&describeOpts)
+	if err != nil {
+		return err
```
bflad: We should check for an error here to indicate that the parameter group has been removed outside of Terraform and remove it from the Terraform state so it can successfully be recreated:

```go
if isAWSErr(err, neptune.ErrCodeDBParameterGroupNotFoundFault, "") {
  log.Printf("[WARN] Neptune Parameter Group (%s) not found, removing from state", d.Id())
  d.SetId("")
  return nil
}
return err
```

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	if len(describeResp.DBParameterGroups) != 1 ||
```
bflad: While we are trying to prevent errors/panics here we should check `describeResp == nil` as well in the (unlikely) scenario that it comes back empty from the API without an error.

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+	d.Set("family", describeResp.DBParameterGroups[0].DBParameterGroupFamily)
+	d.Set("description", describeResp.DBParameterGroups[0].Description)
+
+	// Only include user customized parameters as there's hundreds of system/default ones
+	describeParametersOpts := neptune.DescribeDBParametersInput{
+		DBParameterGroupName: aws.String(d.Id()),
+		Source:               aws.String("user"),
+	}
+
+	describeParametersResp, err := conn.DescribeDBParameters(&describeParametersOpts)
```
bflad: This API call is not currently handling `Marker` for pagination, which means it may not receive all results in large or separated API responses. Instead, we should probably use the available SDK pagination function: [`DescribeDBParametersPages`](https://docs.aws.amazon.com/sdk-for-go/api/service/neptune/#Neptune.DescribeDBParametersPages)

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return err
+	}
+
+	d.Set("parameter", flattenNeptuneParameters(describeParametersResp.Parameters))
```
bflad: When setting the value of a non-scalar attribute into the Terraform state, we should prefer to error check it:

```go
if err := d.Set("parameter", flattenNeptuneParameters(parameters)); err != nil {
  return fmt.Errorf("error setting parameter: %s", err)
}
```

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+func resourceAwsNeptuneParameterGroupDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).neptuneconn
+
+	return resource.Retry(3*time.Minute, func() *resource.RetryError {
+		deleteOpts := neptune.DeleteDBParameterGroupInput{
+			DBParameterGroupName: aws.String(d.Id()),
+		}
+		_, err := conn.DeleteDBParameterGroup(&deleteOpts)
+		if err != nil {
+			awsErr, ok := err.(awserr.Error)
```
bflad: Minor nitpick: we have an available helper function for SDK error handling and should prefer the SDK provided constants:

```go
if isAWSErr(err, neptune.ErrCodeDBParameterGroupNotFoundFault, "") {
  return nil
}
if isAWSErr(err, neptune.ErrCodeInvalidDBParameterGroupStateFault, "") {
}
```

aws/provider.go
```diff
@@ -474,6 +474,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_nat_gateway":                              resourceAwsNatGateway(),
 			"aws_network_acl":                              resourceAwsNetworkAcl(),
 			"aws_default_network_acl":                      resourceAwsDefaultNetworkAcl(),
+			"aws_neptune_parameter_group":                  resourceAwsNeptuneParameterGroup(),
```
bflad: This new resource is missing documentation:
* A new documentation page: `website/docs/r/neptune_parameter_group.html.markdown`
* A sidebar link in `website/aws.erb`

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+					Schema: map[string]*schema.Schema{
+						"name": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"value": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"apply_method": &schema.Schema{
```
bflad: The SDK provides two constants we can use for value validation and to ensure the value is correctly lowercase:

```go
  neptune.ApplyMethodImmediate,
  neptune.ApplyMethodPendingReboot,
}, false),
```

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,266 @@
...
+		toAdd, err := expandNeptuneParameters(ns.Difference(os).List())
+		if err != nil {
+			return err
+		}
+
+		log.Printf("[DEBUG] Parameters to add: %#v", toAdd)
+
+		// We can only modify 20 parameters at a time, so walk them until
+		// we've got them all.
+		maxParams := 20
```
bflad: Minor nitpick: We should probably set this as a top level constant

aws/resource_aws_neptune_parameter_group_test.go
```diff
@@ -0,0 +1,218 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSNeptuneParameterGroupExists("aws_neptune_parameter_group.bar", &v),
+					testAccCheckAWSNeptuneParameterGroupAttributes(&v, rName),
+					resource.TestCheckResourceAttr(
+						"aws_neptune_parameter_group.bar", "name", rName),
+					resource.TestCheckResourceAttr(
+						"aws_neptune_parameter_group.bar", "family", "neptune1"),
+					resource.TestCheckResourceAttr(
+						"aws_neptune_parameter_group.bar", "description", "Managed by Terraform"),
+					resource.TestCheckResourceAttr(
```
bflad: Minor nitpick: when testing nested attributes like TypeList and TypeSet, we should check the associated value count as well: `resource.TestCheckResourceAttr("aws_neptune_parameter_group.bar", "parameter.#", "1")`

We should apply this to any of the other tests below where we also expect a certain number of parameters (especially the removal test).

aws/resource_aws_neptune_parameter_group_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+func TestAccAWSNeptuneParameterGroup_basic(t *testing.T) {
+	var v neptune.DBParameterGroup
+	rName := fmt.Sprintf("parameter-group-test-terraform-%d", acctest.RandInt())
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSNeptuneParameterGroupDestroy,
+		Steps: []resource.TestStep{
```
bflad: To test importing the resource, we should add the following TestStep at the end:

```go
{
  ResourceName:      "aws_neptune_parameter_group.bar",
  ImportState:       true,
  ImportStateVerify: true,
},
```

aws/structure.go
```diff
@@ -740,6 +763,21 @@ func flattenElastiCacheParameters(list []*elasticache.Parameter) []map[string]in
 	return result
 }
 
+// Flattens an array of Parameters into a []map[string]interface{}
+func flattenNeptuneParameters(list []*neptune.Parameter) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, len(list))
+	for _, i := range list {
+		if i.ParameterValue != nil {
+			result = append(result, map[string]interface{}{
+				"name":         strings.ToLower(*i.ParameterName),
```
bflad: To prevent potential panic scenarios, we should use the SDK function here and below to dereference the response objects, rather than `*`: e.g. `aws.StringValue(i.ParameterName)`

dav009:
@bflad 
- addressed your comments
- added docs
🙇

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,280 @@
...
+							Required: true,
+						},
+						"value": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"apply_method": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "immediate",
```
bflad: This seems to be safer to default to `pending-reboot` as that is applicable to both static and dynamic parameters.

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,280 @@
...
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"value": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"apply_method": {
```
bflad: This argument was not documented originally and will be added post merge. Also adding validation:

```go
	neptune.ApplyMethodImmediate,
	neptune.ApplyMethodPendingReboot,
}, false),
```

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,280 @@
...
+							Optional: true,
+							Default:  "immediate",
+							// this parameter is not actually state, but a
+							// meta-parameter describing how the RDS API call
+							// to modify the parameter group should be made.
+							// Future reads of the resource from AWS don't tell
+							// us what we used for apply_method previously, so
+							// by squashing state to an empty string we avoid
+							// needing to do an update for every future run.
+							StateFunc: func(interface{}) string { return "" },
```
bflad: I'm not sure where this `StateFunc` and the comment came from, but the current Neptune API does properly return `ApplyMethod` so we can set it in the Terraform state without modification.

aws/resource_aws_neptune_parameter_group.go
```diff
@@ -0,0 +1,280 @@
...
+						},
+					},
+				},
+				Set: resourceAwsNeptuneParameterHash,
```
bflad: It turns out that this custom `Set` function was extraneous as we want to have `apply_method` computed as a difference if it changed.

aws/structure.go
```diff
@@ -740,6 +763,21 @@ func flattenElastiCacheParameters(list []*elasticache.Parameter) []map[string]in
 	return result
 }
 
+// Flattens an array of Parameters into a []map[string]interface{}
+func flattenNeptuneParameters(list []*neptune.Parameter) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, len(list))
+	for _, i := range list {
+		if i.ParameterValue != nil {
+			result = append(result, map[string]interface{}{
+				"name":         strings.ToLower(aws.StringValue(i.ParameterName)),
```
bflad: The `strings.ToLower()` here seems extraneous. We can always add it back if for some reason its necessary.

aws/structure.go
```diff
@@ -740,6 +763,21 @@ func flattenElastiCacheParameters(list []*elasticache.Parameter) []map[string]in
...
+				"value":        aws.StringValue(i.ParameterValue),
+				"apply_method": strings.ToLower(aws.StringValue(i.ParameterName)),
```
bflad: Two issues:
* Its reading `ParameterName` and not `ApplyMethod`
* The `strings.ToLower()` seems extraneous.

website/docs/r/neptune_parameter_group.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required, Forces new resource) The name of the Neptune parameter group.
+* `family` - (Required) The family of the Neptune parameter group.
+* `description` - (Optional) The description of the Neptune parameter group. Defaults to "Managed by Terraform".
+* `parameter` - (Optional) A list of Neptune parameters to apply.
+* `tags`  - (Optional) A mapping of tags to assign to the resource.
```
bflad: The `tags` attribute is not currently implemented and will be removed from the documentation.

website/docs/r/neptune_parameter_group.html.markdown
```diff
@@ -0,0 +1,64 @@
...
+
+* `name`  - (Required) The name of the Neptune parameter.
+* `value` - (Required) The value of the Neptune parameter.
+
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The dbNeptune parameter group name.
```
bflad: 😎 Small typo `dbNeptune`

bflad:

Issue #3984, opened 2018-03-29
Feature Request: Data source to discover all VPC IDs


erikpaasonen:
Would like to see support for an `aws_vpc_ids` data source which outputs all VPC IDs in the same manner that `aws_subnet_ids` outputs all subnet IDs for a given VPC. The primary use case would interpolate the output into `count` of an `aws_flow_log` resource.

Should work without any arguments; the provider will handle account and region context.

Equivalent of running `aws ec2 describe-vpcs` and selecting the `VpcId` field.

### Suggested Syntax
``` HCL
data "aws_vpc_ids" "foo" {}

output "foo" {
  value = "${data.aws_vpc_ids.foo.ids}"
}

resource "aws_flow_log" "test_flow_log" {
  count = "${length(data.aws_vpc_ids.foo.ids)}"
  ...
  vpc_id = "${element(data.aws_vpc_ids.foo.ids, count.index)}"
  ...
}
```

### Expected Output
With 2 VPCs present:
``` HCL
foo = [
  vpc-a1b2c3d,
  vpc-e4d5f6g
]
```

With zero VPCs:
``` HCL
foo = []
```

### References
https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-vpcs.html

bflad:
A new `aws_vpcs` data source will be available in version 1.23.0 of the AWS provider, likely releasing later today.

bflad:
This has been released in [version 1.23.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1230-june-14-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #4736, opened 2018-06-04
Issue #3984 New Datasource: aws_vpcs

saravanan30erd:
New datasource aws_vpc_ids to discover all VPC IDs(Issue #3984 ).

### DIFF
--- a/aws/data_source_aws_vpcs_test.go
+++ b/aws/data_source_aws_vpcs_test.go
@@ -17,7 +17,7 @@ func TestAccDataSourceAwsVpcs_basic(t *testing.T) {
 			{
 				Config: testAccDataSourceAwsVpcsConfig(),
 				Check: resource.ComposeTestCheckFunc(
-					testCheckResourceAttrGreaterThanValue("data.aws_vpcs.all", "ids.#", "0"),
+					testAccCheckAwsVpcsDataSourceID("data.aws_vpcs.all"),
 				),
 			},
 		},
@@ -86,15 +86,27 @@ func testCheckResourceAttrGreaterThanValue(name, key, value string) resource.Tes
 	}
 }
 
+func testAccCheckAwsVpcsDataSourceID(n string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Can't find aws_vpcs data source: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("aws_vpcs data source ID not set")
+		}
+		return nil
+	}
+}
+
 func testAccDataSourceAwsVpcsConfig() string {
 	return fmt.Sprintf(`
 	resource "aws_vpc" "test-vpc" {
   		cidr_block = "10.0.0.0/24"
 	}
 
-	data "aws_vpcs" "all" {
-		depends_on = ["aws_vpc.test-vpc"]
-	}
+	data "aws_vpcs" "all" {}
 	`)
 }
 


saravanan30erd:
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccDataSourceAwsVpcIDs_'
TF_ACC=1 go test ./aws/ -v -run=TestAccDataSourceAwsVpcIDs_ -timeout 120m
=== RUN   TestAccDataSourceAwsVpcIDs_basic
--- PASS: TestAccDataSourceAwsVpcIDs_basic (81.57s)
=== RUN   TestAccDataSourceAwsVpcIDs_tags
--- PASS: TestAccDataSourceAwsVpcIDs_tags (81.60s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	163.211s

aws/data_source_aws_vpc_ids_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func TestAccDataSourceAwsVpcIDs_basic(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceAwsVpcIDsConfig(),
+				Check: resource.ComposeTestCheckFunc(
+					//datasource aws_vpc_ids will list all the vpcs in a region, so it also includes the Default vpc.
```
bflad: It does not look like anything is reading the `filter` schema attribute into `req.Filter`, which might explain the behavior you're seeing here when you have an invalid filter in the test configuration. It looks like the valid value should be `vpc-id` according to the SDK documentation: https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeVpcsInput

For compatibility in all AWS accounts (which may or may not have other existing VPCs since you can delete the default VPC), I would recommend setting up the `_basic` test to not have any ids/filter/tags arguments, create 1 VPC resource, add `depends_on` in the data source to that VPC resource, then ensure that the value of the attribute is greater than 0. We cannot count on `ids` being `"3"`.

aws/data_source_aws_vpc_ids_test.go
```diff
@@ -0,0 +1,90 @@
...
+					resource.TestCheckResourceAttr("data.aws_vpc_ids.selected", "ids.#", "2"),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataSourceAwsVpcIDsConfig() string {
+	return fmt.Sprintf(`
+	provider "aws" {
```
bflad: Declaring the "default" provider region is extraneous in test configurations and discouraged unless its absolutely necessary (e.g. EC2 Classic testing) -- its handled by the `AWS_DEFAULT_REGION` environment variable (or defaults to `us-west-2`) here: 

https://github.com/terraform-providers/terraform-provider-aws/blob/59ea1c3e859c45a3a19d60b49b7c42fee88e5d7b/aws/provider_test.go#L69

aws/data_source_aws_vpc_ids_test.go
```diff
@@ -0,0 +1,90 @@
...
+  		cidr_block = "10.0.0.0/24"
+
+  		tags {
+  			Name = "testacc-vpc"
+  			Service = "test"
+  		}
+	}
+
+	data "aws_vpc_ids" "selected" {
+		filter {
```
bflad: We should remove this in preference of only testing `tags` in this configuration 👍

aws/data_source_aws_vpc_ids.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	vpcs := make([]string, 0)
+
+	for _, vpc := range resp.Vpcs {
+		vpcs = append(vpcs, *vpc.VpcId)
+	}
+
+	d.SetId(time.Now().UTC().String())
+	d.Set("ids", vpcs)
```
bflad: I would recommend adding error checking for setting a non-scalar attribute:

```go
if err := d.Set("ids", vpcs); err != nil {
  return fmt.Errorf("error setting ids: %s", err)
}
```

aws/data_source_aws_vpc_ids.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	if resp == nil || len(resp.Vpcs) == 0 {
+		return fmt.Errorf("no matching VPC found")
+	}
+
+	vpcs := make([]string, 0)
+
+	for _, vpc := range resp.Vpcs {
+		vpcs = append(vpcs, *vpc.VpcId)
```
bflad: To prevent potential panics we might want to use the SDK helper function here: `aws.StringValue(vpc.VpcId)`

website/docs/d/vpc_ids.html.markdown
```diff
@@ -0,0 +1,57 @@
+---
+layout: "aws"
+page_title: "AWS: aws_vpc_ids"
+sidebar_current: "docs-aws-datasource-vpc-ids"
+description: |-
+    Provides a list of VPC Ids in a region
+---
+
+# Data Source: aws_vpc_ids
```
bflad: This documentation page is missing a sidebar link in `website/aws.erb`

website/docs/d/vpc_ids.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+    service = "production"
+  }
+}
+
+output "foo" {
+  value = "${data.aws_vpc_ids.foo.ids}"
+}
+```
+
+The primary use case would be interpolate the vpc_ids output into `count` of an aws_flow_log resource.
```
bflad: Nitpick: We generally try not to prescribe what might be a "primary" use case -- people use Terraform in all sorts of scenarios. I'd consider saying `An example use case`.

aws/data_source_aws_vpc_ids.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsVpcIDs() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsVpcIDsRead,
+		Schema: map[string]*schema.Schema{
+			"filter": dataSourceFiltersSchema(),
```
bflad: A few items with this attribute:
* It does seem to be read into `req.Filters` so it appears to not currently be doing anything
* Its missing in the data source documentation

I would recommend testing this attribute separately from the `_basic` test (preferably not being passed any arguments) and the `_tags` test (preferably only testing tags filtering).

website/docs/d/vpc_ids.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+layout: "aws"
+page_title: "AWS: aws_vpc_ids"
+sidebar_current: "docs-aws-datasource-vpc-ids"
+description: |-
+    Provides a list of VPC Ids in a region
+---
+
+# Data Source: aws_vpc_ids
+
+`aws_vpc_ids` provides a list of vpc ids
```
bflad: Nitpick: This sentence seems redundant with the one below. Also, the casing of VPC IDs is different between various sentences in the documentation page.

aws/provider.go
```diff
@@ -242,6 +242,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_ssm_parameter":                    dataSourceAwsSsmParameter(),
 			"aws_subnet":                           dataSourceAwsSubnet(),
 			"aws_subnet_ids":                       dataSourceAwsSubnetIDs(),
+			"aws_vpc_ids":                          dataSourceAwsVpcIDs(),
```
bflad: I'd recommend calling this data source `aws_vpcs` -- we may return more than IDs in the future. Unfortunately we have some more explicitly `_ids` named ones (like `aws_subnet_ids` above) that eventually may need to be deprecated should we need to return information other than IDs.

website/docs/d/vpc_ids.html.markdown
saravanan30erd: done

aws/data_source_aws_vpc_ids_test.go
saravanan30erd: sure

aws/data_source_aws_vpc_ids.go
saravanan30erd: Its my mistake, I added the filter in last minute when I was working on test cases, then I forgot to follow on.

aws/data_source_aws_vpc_ids.go
saravanan30erd: Done, can you please explain about non-scalar attribute?

bflad: Technically speaking, we should probably perform the error checking on all attributes during `d.Set()`, but its quite verbose code-wise and we have found nested and list/set attributes to be the most troublesome when it comes to ensuring that the type from the API matches the type defined in the schema (e.g. incorrectly trying to put `[]*string` into a `TypeList` with `TypeString` elements). Generally its worse when there is more attribute nesting happening when you need to be very concerned about lining up `[]map[string]interface{}` correctly.

When this type mismatch occurs, the behavior of Terraform core is to pass the configuration through into the state as-is from the configuration and not provide any feedback about this behavior unless the error check is in place. In the situation where we don't error check and provide incompatible type(s) into `d.Set()`, its not possible to detect drift from the API. It will also show as a plan difference after resource import since the state could not set the real value during the import read.

We have a flag in Terraform core to instead `panic()` when it runs into this situation (you can manually enable it today with `TF_SCHEMA_PANIC_ON_ERROR=1` environment variable), but I'm not sure we'll ever flip it on given the current state of many of the providers and generally poorer user experience that provides even though it would certainly make these bugs way more obvious.

aws/data_source_aws_vpc_ids_test.go
bflad: If you turn on the debug logging, you should see the empty plan output. I generally don't show it during PR submission but I find it very helpful to run my acceptance testing like this:

```
rm -f aws/log.txt; TF_LOG_PATH=log.txt TF_LOG=debug make testacc TEST=./aws TESTARGS='-run=TestAccAWSXXX'
```

Which will turn on debug logging and write that output to a separate log file (`aws/log.txt`) that I can inspect.

aws/data_source_aws_vpc_ids.go
saravanan30erd: thanks

aws/data_source_aws_vpc_ids_test.go
saravanan30erd: @bflad Getting the below error when I using `depends_on` in data source to that VPC resource.
`--- FAIL: TestAccDataSourceAwsVpcs_basic (79.22s)
	testing.go:518: Step 0 error: After applying this step and refreshing, the plan was not empty:

		DIFF:

		CREATE: data.aws_vpcs.all
		  ids.#:  "" => "<computed>"
		  tags.%: "" => "<computed>"

		STATE:

		aws_vpc.test-vpc:
		  ID = vpc-954c44ec
		  provider = provider.aws
		  assign_generated_ipv6_cidr_block = false
		  cidr_block = 10.0.0.0/24
		  default_network_acl_id = acl-61b72c19
		  default_route_table_id = rtb-5ff0e927
		  default_security_group_id = sg-29daa658
		  dhcp_options_id = dopt-f0038698
		  enable_classiclink = false
		  enable_classiclink_dns_support = false
		  enable_dns_hostnames = false
		  enable_dns_support = true
		  instance_tenancy = default
		  main_route_table_id = rtb-5ff0e927
		  tags.% = 0
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	79.264s`

After destroyed the vpc resource and data source and running the plan shows the drift due to `depends_on` (to vpc resource) in data source. In `data_source_aws_vpc_test.go`, `TestAccDataSourceAwsVpc_basic` fails with same error after adding `depends_on` in data source.  Other tests are fine.

aws/data_source_aws_vpcs_test.go
```diff
@@ -0,0 +1,149 @@
...
+				key,
+				value,
+				v)
+		}
+		return nil
+
+	}
+}
+
+func testAccCheckAwsVpcsDataSourceID(n string) resource.TestCheckFunc {
```
bflad: Minor nitpick: We will have the documentation up very shortly (https://github.com/hashicorp/terraform-website/pull/317 😉 ), but even when we're not checking for resource/datasource API object, we prefer to name these functions `testAccCheckXXXExists` for consistency. I'll adjust this on merge.

saravanan30erd: Sure 👍

bflad:

PR #4743, opened 2018-06-05
r/aws_rds_cluster: Document tags

stefansundin:
This documents that it is possible to tag tags `aws_rds_cluster` resources. I also fixed minor punctuation. I do not like double spacing either, it is a sin that should not be practiced.


stefansundin:
6fce618d582875c87c585cda3aebf0e2b970fa9701e3db2076a3213cca3218f3

double-spacing-after-periods-honestly-who-does-that

PR #4798, opened 2018-06-08
AWS: Create composite resource 'aws_iam_policy_with_attachment'

HaraldNordgren:

* I want to create a composite aws resource type 'aws_iam_policy_with_attachment' that creates a policy together with the attachment.

```
$ make testacc TESTARGS='-run=TestAccAWSIAMPolicyWithAttachment'
TF_ACC=1 go test ./... -v -run=TestAccAWSIAMPolicyWithAttachment -timeout 120m
=== RUN   TestAccAWSIAMPolicyWithAttachment
--- PASS: TestAccAWSIAMPolicyWithAttachment (30.22s)
PASS
```

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -426,6 +426,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_iam_instance_profile":                     resourceAwsIamInstanceProfile(),
 			"aws_iam_openid_connect_provider":              resourceAwsIamOpenIDConnectProvider(),
 			"aws_iam_policy":                               resourceAwsIamPolicy(),
+			"aws_iam_policy_with_attachment":               resourceAwsIamPolicyWithAttachment(),
 			"aws_iam_policy_attachment":                    resourceAwsIamPolicyAttachment(),
 			"aws_iam_role_policy_attachment":               resourceAwsIamRolePolicyAttachment(),
 			"aws_iam_role_policy":                          resourceAwsIamRolePolicy(),
--- a/aws/resource_aws_iam_policy.go
+++ b/aws/resource_aws_iam_policy.go
@@ -115,10 +115,13 @@ func resourceAwsIamPolicyCreate(d *schema.ResourceData, meta interface{}) error 
 }
 
 func resourceAwsIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
-	iamconn := meta.(*AWSClient).iamconn
+	return resourceAwsIamPolicyReader(d, d.Id(), meta)
+}
 
+func resourceAwsIamPolicyReader(d *schema.ResourceData, arn string, meta interface{}) error {
+	iamconn := meta.(*AWSClient).iamconn
 	getPolicyRequest := &iam.GetPolicyInput{
-		PolicyArn: aws.String(d.Id()),
+		PolicyArn: aws.String(arn),
 	}
 
 	getPolicyResponse, err := iamconn.GetPolicy(getPolicyRequest)
@@ -131,7 +134,7 @@ func resourceAwsIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	getPolicyVersionRequest := &iam.GetPolicyVersionInput{
-		PolicyArn: aws.String(d.Id()),
+		PolicyArn: aws.String(arn),
 		VersionId: getPolicyResponse.Policy.DefaultVersionId,
 	}
 
@@ -141,7 +144,7 @@ func resourceAwsIamPolicyRead(d *schema.ResourceData, meta interface{}) error {
 			d.SetId("")
 			return nil
 		}
-		return fmt.Errorf("Error reading IAM policy version %s: %s", d.Id(), err)
+		return fmt.Errorf("Error reading IAM policy version %s: %s", arn, err)
 	}
 
 	policy, err := url.QueryUnescape(*getPolicyVersionResponse.PolicyVersion.Document)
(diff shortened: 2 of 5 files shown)


HaraldNordgren:
@bflad Merge this?

bflad:
Hi @HaraldNordgren 👋  Can you please explain the use case here?

We already provide many build-block resources to accomplish this functionality. e.g. with the [`aws_iam_user` resource](https://www.terraform.io/docs/providers/aws/r/iam_user.html), [`aws_iam_policy` resource](https://www.terraform.io/docs/providers/aws/r/iam_policy.html) and [`aws_iam_user_policy_attachment` resource](https://www.terraform.io/docs/providers/aws/r/iam_user_policy_attachment.html)

```hcl
resource "aws_iam_user" "user" {
    name = "test-user"
}

resource "aws_iam_policy" "policy" {
    name        = "test-policy"
    description = "A test policy"
    policy      = "{ ... }"
}

resource "aws_iam_user_policy_attachment" "test-attach" {
    user       = "${aws_iam_user.user.name}"
    policy_arn = "${aws_iam_policy.policy.arn}"
}
```

In many cases, providing these building blocks allows Terraform configurations to be very flexible and maintainable. We would likely not accept something that duplicates existing functionality since it often can provide a confusing operator experience.

HaraldNordgren:
Hi!

I am aware the Terraform implements the policy-attachment workflow already. This is intended as a simplification -- possibly a replacement for the `aws_iam_policy` and `aws_iam_user_policy_attachment` resources.

The implementation makes sure to call the CRUD functions of the underlying resources so that aws_iam_user_policy_with_attachment works as a full replacement for the two resources. The two underlying resources move in sync.

The normal workflow would then be

```tf
resource "aws_iam_user" "user" {
    name = "test-user"
}

resource "aws_iam_user_policy_with_attachment" "test-policy" {
    name        = "test-policy"
    description = "A test policy"
    policy      = "{ ... }"
    user       = "${aws_iam_user.user.name}"
}
```

Notice how `policy_arn` since this is now connected internally.

HaraldNordgren:
Hi @bflad!

I am aware the Terraform implements the policy-attachment workflow already. This is intended as a simplification -- possibly a replacement for the `aws_iam_policy` and `aws_iam_user_policy_attachment` resources.

The implementation makes sure to call the CRUD functions of the underlying resources so that aws_iam_user_policy_with_attachment works as a full replacement for the two resources. The two underlying resources move in sync.

The normal workflow would then be

```tf
resource "aws_iam_user" "user" {
    name = "test-user"
}

resource "aws_iam_user_policy_with_attachment" "test-policy" {
    name        = "test-policy"
    description = "A test policy"
    policy      = "{ ... }"
    user        = "${aws_iam_user.user.name}"
}
```

Notice how `policy_arn` is skipped since this is now connected internally.

bflad:
Hi again, @HaraldNordgren. 👋  We understand what you are trying to accomplish here, but there are a lot of associated costs with introducing another Terraform resource like this, just for the purpose of simplification -- I will explain.

> I am aware the Terraform implements the policy-attachment workflow already. This is intended as a simplification

The provider is meant to give operators the basic building blocks for working with individual AWS API relationships/objects, which are separated to cover all possible use cases. We do not want to prescribe that you must work with your infrastructure in a certain way. There are more subtle complexities when trying to manage multiple-purpose Terraform resources, usually leading to maintenance burden within the provider and operator confusion from multiple ways to do something. So by design, we do not try to manage multiple API relationships/objects within one resource.

Admittedly, there are a few exceptions to this case with a historically problematic one being the `aws_security_group` and `aws_security_group_rule` resources, where they conflict with each other. There is a disproportionately higher number of operator issues opened for conflicting management problems like these than the individual resources themselves, despite large warnings in the documentation.

> possibly a replacement for the aws_iam_policy and aws_iam_user_policy_attachment resources.

Its perfectly valid to have completely separate users/Terraform configurations that create IAM policies in one place and attach them in other. Here are some examples that immediately present problems with using this implementation as the prescribed way:
* An organizational administrative account may handle creating/editing all policies across all member AWS accounts (e.g. for compliance reasons), while those member accounts manage attaching to those policies.
* Creating an IAM policy intended for attaching to groups or roles
* Creating an IAM policy intended for attaching to two separate users

Especially in the last case, [IAM limitations on the number of managed policies](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html) makes it undesirable to create separate IAM policies per user attachment. In quite a few cases, Terraform is used by organizations with thousands of users.

> The implementation makes sure to call the CRUD functions of the underlying resources so that aws_iam_user_policy_with_attachment works as a full replacement for the two resources. The two underlying resources move in sync.

Under the hood, the resource will still be making multiple API calls to manage multiple pieces of infrastructure. Having the code in one resource versus two (or more) resources does not change any of the complexities when working with multiple API calls, especially with [IAM eventual consistency](https://docs.aws.amazon.com/IAM/latest/UserGuide/troubleshoot_general.html#troubleshoot_general_eventual-consistency) in this case. Creating an IAM policy does not immediately make it available for attachment because its within the same Terraform resource.

---

We definitely appreciate your idea, however given that this functionality is not likely to be merged for the reasons above, I am going to close this pull request.

Without any changes to the provider, its also possible to simplify the same end behavior by switching to using in-line IAM user policies instead of managed IAM policies. Below is an example of how to do this via the [`aws_iam_user_policy` resource](https://www.terraform.io/docs/providers/aws/r/iam_user_policy.html):

```hcl
resource "aws_iam_user_policy" "example" {
  name = "test-policy"
  user  = "${aws_iam_user.user.name}"
  policy = "{ ... }"
```

Alternatively and more typically outside this case (since usually there are not two ways to do a similar action/behavior within an API), we would expect any high level simplification like this pull request suggests to be implemented in a [Terraform module](https://www.terraform.io/docs/modules/usage.html), to abstract the configuration details for operators and (optionally) published on https://registry.terraform.io/ for anyone's usage. Below is a very brief example of creating a module for this simplification:

```hcl
variable "description" {
  type = "string"
}

variable "policy" {
  type = "string"
}

variable "policy_name" {
  type = "string"
}

variable "user" {
  type = "string"
}

resource "aws_iam_policy" "policy" {
    name        = "${var.policy_name}"
    description = "${var.description}"
    policy      = "${var.policy}"
}

resource "aws_iam_user_policy_attachment" "test-attach" {
    user       = "${aws_iam_user.user.name}"
    policy_arn = "${aws_iam_policy.policy.arn}"
}
```

Which can be invoked via something like:

```hcl
module "aws_iam_user_policy_with_attachment" {
  source = "..."

  policy_name = "test-policy"
  description = "A test policy"
  policy      = "{ ... }"
  user        = "${aws_iam_user.user.name}"
}
```

I hope either of these alternatives help. 👍

PR #4803, opened 2018-06-10
aws db instance: displaying invalid parameter in error

### LINKED ISSUE #3626: Display Invalid Value During InvalidParameterValue
_This issue was originally opened by @Miserlou as hashicorp/terraform#17502. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

```
Terraform v0.11.3
+ provider.aws v1.10.0
+ provider.local v1.1.0
+ provider.null v1.0.0
+ provider.template v1.0.0
```

```hcl
...
```

### Debug Output

```
* aws_db_instance.postgres-db: Error creating DB Instance: InvalidParameterValue: DBName must begin with a letter and contain only alphanumeric characters.
	status code: 400, request id: 643fee1f-5dcc-4c8f-8250-1aaa2821cdee
```

### Expected Behavior

My value for this field is computed, and I want to make that it is computed correctly. So, I need to be able to see what was computed. I want to see something like:
```
* aws_db_instance.postgres-db: Error creating DB Instance: InvalidParameterValue: DBName must begin with a letter and contain only alphanumeric characters. 
        DBName provided: [123-This-Is-Invalid!]
	status code: 400, request id: 643fee1f-5dcc-4c8f-8250-1aaa2821cdee
```

This is a simple change that would make development and debugging much easier.

dav009:
Fixes #3626

* if a`InvalidParameterValue` error is   returned  when creating DB resource, then opts sent to the requests are shown in the error

None
```

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -908,7 +908,13 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			return nil
 		})
 		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
+				if awsErr.Code() == "InvalidParameterValue" {
+					return fmt.Errorf("Error creating DB Instance: %s, %+v", err, opts)
+				}
+			}
 			return fmt.Errorf("Error creating DB Instance: %s", err)
+
 		}
 	}
 

### CODE REVIEW

File: aws/resource_aws_db_instance.go [hunk 1]
```diff
@@ -908,7 +908,13 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			return nil
 		})
 		if err != nil {
+			if awsErr, ok := err.(awserr.Error); ok {
```
bflad: Minor nitpick: we have a helper function to simplify this logic (and normally prefer to use SDK provided constants but looks like the [rds package](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#pkg-constants) doesn't have one here 😉 )

```go
if isAWSErr(err, "InvalidParameterValue", "") {
```

bflad:

PR #4811, opened 2018-06-11
Add ComputePlatform to aws_codedeploy_app

### LINKED ISSUE #3692: CodeDeploy to Lambda
AWS CodeDeploy applications now support deploying to Lambda functions:
https://docs.aws.amazon.com/codedeploy/latest/userguide/applications-create-lambda.html

As of v1.10.0 setting the "compute platform" of a lambda function is not supproted.

# References
https://www.terraform.io/docs/providers/aws/r/codedeploy_app.html
#3368

cgarvis:
Fixes #3692

* Add `ComputePlatform` to `CodeDeployApp`

```
$ make testacc make testacc TESTARGS='-run=TestAccAWSCodeDeployApp'
TF_ACC=1 go test ./... -v -run=TestAccAWSCodeDeployApp -timeout 120m
=== RUN   TestAccAWSCodeDeployApp_basic
--- PASS: TestAccAWSCodeDeployApp_basic (24.20s)
=== RUN   TestAccAWSCodeDeployApp_computePlatform
--- PASS: TestAccAWSCodeDeployApp_computePlatform (23.75s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       58.844s
``

### DIFF
--- a/aws/resource_aws_codedeploy_app.go
+++ b/aws/resource_aws_codedeploy_app.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/codedeploy"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsCodeDeployApp() *schema.Resource {
@@ -26,6 +27,16 @@ func resourceAwsCodeDeployApp() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"compute_platform": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					codedeploy.ComputePlatformServer,
+					codedeploy.ComputePlatformLambda,
+				}, false),
+				Default: codedeploy.ComputePlatformServer,
+			},
+
 			// The unique ID is set by AWS on create.
 			"unique_id": &schema.Schema{
 				Type:     schema.TypeString,
@@ -40,10 +51,12 @@ func resourceAwsCodeDeployAppCreate(d *schema.ResourceData, meta interface{}) er
 	conn := meta.(*AWSClient).codedeployconn
 
 	application := d.Get("name").(string)
+	computePlatform := d.Get("compute_platform").(string)
 	log.Printf("[DEBUG] Creating CodeDeploy application %s", application)
 
 	resp, err := conn.CreateApplication(&codedeploy.CreateApplicationInput{
 		ApplicationName: aws.String(application),
+		ComputePlatform: aws.String(computePlatform),
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: aws/resource_aws_codedeploy_app.go [hunk 1]
```diff
@@ -26,6 +27,16 @@ func resourceAwsCodeDeployApp() *schema.Resource {
 				ForceNew: true,
 			},
 
+			"compute_platform": &schema.Schema{
```
bflad: There are two subtle, but important bugs with this attribute currently. 😄 
* The read function is not calling `d.Set("") which has two implications: Terraform cannot detect drift of the attribute and in absence of the `d.Set()` call with the value, Terraform core is silently passing the configuration value into the Terraform state (which is why the acceptance testing passes)
* This attribute cannot be updated in `UpdateApplication` and therefore needs to be set as `ForceNew: true` to force new resource creation

One other minor nitpick: (the rest of the file hasn't been updated yet, but) as of Go 1.7, the `&schema.Schema` is extraneous as its defined by `map[string]*schema.Schema` above so we prefer to remove them now :+1:

website/docs/r/codedeploy_app.html.markdown
```diff
@@ -23,6 +23,7 @@ resource "aws_codedeploy_app" "foo" {
 The following arguments are supported:
 
 * `name` - (Required) The name of the application.
+* `compute_platform` - (Optional) The compute platform can either be `Server` or `Lambda`.
```
bflad: We should list the default of `Server` here 👍

aws/resource_aws_codedeploy_app.go
```diff
@@ -26,6 +27,16 @@ func resourceAwsCodeDeployApp() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					codedeploy.ComputePlatformServer,
+					codedeploy.ComputePlatformLambda,
+				}, false),
+				Default: codedeploy.ComputePlatformServer,
```
bflad: We should check the default value in `TestAccAWSCodeDeployApp_basic` as well (when not provided in the test configuration):

```go
resource.TestCheckResourceAttr("aws_codedeploy_app.foo", "compute_platform", "Server"),
```

Jamie-BitFlight:
Looking forward to this PR :)

cgarvis:
@bflad updated per your suggestions.  Thanks!.  Tests look good.

aws/resource_aws_codedeploy_app.go
```diff
@@ -20,14 +21,24 @@ func resourceAwsCodeDeployApp() *schema.Resource {
 		Delete: resourceAwsCodeDeployAppDelete,
 
 		Schema: map[string]*schema.Schema{
-			"name": &schema.Schema{
```
bflad: 😍

cgarvis:
👍

bflad:

PR #4825, opened 2018-06-13
Ecs scheduling strategy

### LINKED ISSUE #4820: Add support for scheduling_strategy on aws_ecs_service
### Community Note

### Description

Support for `daemon` services (1 task per ECS instance) was just added to ECS as a `schedulingStrategy` attribute when creating a service ([api docs](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html#ECS-CreateService-request-schedulingStrategy)). Valid values are:

* `REPLICA`: this is the strategy that all ECS tasks used to be
* `DAEMON`: one task per ECS instance

It would be great to be able to create services of this type with terraform.

* aws_ecs_service

Note: I'm not sure if the `scheduling_strategy` values should be lowercase like most terraform configs or UPPERCASE to match the AWS API values.

```hcl
resource "aws_ecs_service" "datadog" {
  name                = "datadog"
  cluster             = "${aws_ecs_cluster.foo.id}"
  task_definition     = "${aws_ecs_task_definition.datadog.arn}"
  iam_role            = "${aws_iam_role.foo.arn}"
  scheduling_strategy = "daemon"
}
```

### References

* https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html#ECS-CreateService-request-schedulingStrategy

erks:
Fixes #4820
Depends on #4823 

* add support for ECS [`schedulingStrategy`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html#ECS-CreateService-request-schedulingStrategy)

### DIFF
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -329,7 +329,12 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if v, ok := d.GetOk("scheduling_strategy"); ok {
-		input.SchedulingStrategy = aws.String(v.(string))
+		schedulingStrategy := v.(string)
+		input.SchedulingStrategy = aws.String(schedulingStrategy)
+		if schedulingStrategy == "DAEMON" {
+			// unset desired count if DAEMON
+			input.DesiredCount = nil
+		}
 	}
 
 	loadBalancers := expandEcsLoadBalancers(d.Get("load_balancer").(*schema.Set).List())
@@ -500,10 +505,13 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("task_definition", taskDefinition)
 	}
 
-	d.Set("desired_count", service.DesiredCount)
+	d.Set("scheduling_strategy", service.SchedulingStrategy)
+	// Automatically ignore desired count if DAEMON
+	if *service.SchedulingStrategy != "DAEMON" {
+		d.Set("desired_count", service.DesiredCount)
+	}
 	d.Set("health_check_grace_period_seconds", service.HealthCheckGracePeriodSeconds)
 	d.Set("launch_type", service.LaunchType)
-	d.Set("scheduling_strategy", service.SchedulingStrategy)
 
 	// Save cluster in the same format
 	if strings.HasPrefix(d.Get("cluster").(string), "arn:"+meta.(*AWSClient).partition+":ecs:") {
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -17,6 +17,8 @@ import (
 	"github.com/hashicorp/terraform/helper/validation"
 )
 
+const schedulingStrategyDaemon = "DAEMON"
+
 var taskDefinitionRE = regexp.MustCompile("^([a-zA-Z0-9_-]+):([0-9]+)$")
 
 func resourceAwsEcsService() *schema.Resource {
@@ -331,7 +333,7 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 	if v, ok := d.GetOk("scheduling_strategy"); ok {
 		schedulingStrategy := v.(string)
 		input.SchedulingStrategy = aws.String(schedulingStrategy)
-		if schedulingStrategy == "DAEMON" {
+		if schedulingStrategy == schedulingStrategyDaemon {
 			// unset desired count if DAEMON
 			input.DesiredCount = nil
 		}
@@ -507,7 +509,7 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 
 	d.Set("scheduling_strategy", service.SchedulingStrategy)
 	// Automatically ignore desired count if DAEMON
-	if *service.SchedulingStrategy != "DAEMON" {
+	if *service.SchedulingStrategy != schedulingStrategyDaemon {
 		d.Set("desired_count", service.DesiredCount)
 	}
 	d.Set("health_check_grace_period_seconds", service.HealthCheckGracePeriodSeconds)
@@ -729,7 +731,9 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 		Cluster: aws.String(d.Get("cluster").(string)),
 	}
 
-	if d.HasChange("desired_count") {
+	schedulingStrategy := d.Get("scheduling_strategy").(string)
+	// Automatically ignore desired count if DAEMON
+	if schedulingStrategy != schedulingStrategyDaemon && d.HasChange("desired_count") {
 		_, n := d.GetChange("desired_count")
 		input.DesiredCount = aws.Int64(int64(n.(int)))
 	}
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -17,8 +17,6 @@ import (
 	"github.com/hashicorp/terraform/helper/validation"
 )
 
-const schedulingStrategyDaemon = "DAEMON"
-
 var taskDefinitionRE = regexp.MustCompile("^([a-zA-Z0-9_-]+):([0-9]+)$")
 
 func resourceAwsEcsService() *schema.Resource {
@@ -72,7 +70,11 @@ func resourceAwsEcsService() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				Default:  "REPLICA",
+				Default:  ecs.SchedulingStrategyReplica,
+				ValidateFunc: validation.StringInSlice([]string{
+					ecs.SchedulingStrategyDaemon,
+					ecs.SchedulingStrategyReplica,
+				}, false),
 			},
 
 			"iam_role": {
@@ -330,13 +332,11 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 		input.LaunchType = aws.String(v.(string))
 	}
 
-	if v, ok := d.GetOk("scheduling_strategy"); ok {
-		schedulingStrategy := v.(string)
-		input.SchedulingStrategy = aws.String(schedulingStrategy)
-		if schedulingStrategy == schedulingStrategyDaemon {
-			// unset desired count if DAEMON
-			input.DesiredCount = nil
-		}
+	schedulingStrategy := d.Get("scheduling_strategy").(string)
+	input.SchedulingStrategy = aws.String(schedulingStrategy)
+	if schedulingStrategy == ecs.SchedulingStrategyDaemon {
+		// unset desired count if DAEMON
+		input.DesiredCount = nil
 	}
 
 	loadBalancers := expandEcsLoadBalancers(d.Get("load_balancer").(*schema.Set).List())
@@ -509,7 +509,7 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 
 	d.Set("scheduling_strategy", service.SchedulingStrategy)
 	// Automatically ignore desired count if DAEMON
-	if *service.SchedulingStrategy != schedulingStrategyDaemon {
+	if *service.SchedulingStrategy != ecs.SchedulingStrategyDaemon {
 		d.Set("desired_count", service.DesiredCount)
 	}
 	d.Set("health_check_grace_period_seconds", service.HealthCheckGracePeriodSeconds)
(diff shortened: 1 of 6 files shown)


aws/resource_aws_ecs_service.go
```diff
@@ -321,6 +328,15 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 		input.LaunchType = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("scheduling_strategy"); ok {
```
christofferh: Should `deploymentConfiguration` be taken into account too?

> If deploymentConfiguration is specified, the maximum percent parameter must be 100. The default value for a daemon service for maximumPercent is 100%. The default value for a daemon service for minimumHealthyPercent is 0% for the AWS CLI, the AWS SDKs, and the APIs, and 50% for the AWS Management Console.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_services.html#service_scheduler_daemon

bflad:
#4823 has been merged so this is good for rebasing 👍

aws/resource_aws_ecs_service.go
erks: I was thinking about that too, but I'm not sure what the general handlings of this are in other places. I'm obviously relying on AWS doing the validation and returning the error for this. But if the best practice here is to do the validation on the terraform side, I can add it. Let me know.

erks:
@bflad do you have any pointer here on how best I can ignore `desired_count` on create/update/read/etc when `scheduling_strategy` is `DAEMON`? Also, should we validate `deploymentConfiguration` according to https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_services.html#service_scheduler_daemon ?

aws/resource_aws_ecs_service.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/hashicorp/terraform/helper/validation"
 )
 
+const schedulingStrategyDaemon = "DAEMON"
```
tomelliff: You should just be able to use `ecs.SchedulingStrategyDaemon` (and its equivalent `ecs.SchedulingStrategyReplica`) here instead.

aws/resource_aws_ecs_service_test.go
```diff
@@ -1955,3 +2027,41 @@ resource "aws_ecs_service" "test" {
...
+}
+resource "aws_ecs_service" "ghost" {
+  name = "%s"
+  cluster = "${aws_ecs_cluster.default.id}"
+  task_definition = "${aws_ecs_task_definition.ghost.family}:${aws_ecs_task_definition.ghost.revision}"
+  %s
+  %s
+}
+`, clusterName, tdName, svcName, schedulingStrategySnippet, desiredCountSnippet)
+}
```
tomelliff: I waver on this a bit and am personally a little inconsistent but I think I prefer having a bit of repetition in the test config code rather than more complicated dynamic config, particularly when it involves something as slightly convoluted as this.

More of a thought/comment than a critique but I think I find the tests a nice form of documentation about how to use things/how things work but this is a bit harder to follow here.

website/docs/r/ecs_service.html.markdown
```diff
@@ -67,8 +67,9 @@ The following arguments are supported:
 
 * `name` - (Required) The name of the service (up to 255 letters, numbers, hyphens, and underscores)
 * `task_definition` - (Required) The family and revision (`family:revision`) or full ARN of the task definition that you want to run in your service.
-* `desired_count` - (Required) The number of instances of the task definition to place and keep running
+* `desired_count` - (Optional) The number of instances of the task definition to place and keep running. Defaults to 0. Do not specify if using the `DAEMON` scheduling strategy.
 * `launch_type` - (Optional) The launch type on which to run your service. The valid values are `EC2` and `FARGATE`. Defaults to `EC2`.
+* `scheduling_strategry` - (Optional) The scheduling strategy to use for the service. The valid values are `REPLICA` and `DAEMON`. Defaults to `REPLICA`.
```
tomelliff: nit: `scheduling_strategry` -> `scheduling_strategy`

aws/resource_aws_ecs_service.go
```diff
@@ -66,6 +68,13 @@ func resourceAwsEcsService() *schema.Resource {
 				Default:  "EC2",
 			},
 
+			"scheduling_strategy": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  "REPLICA",
+			},
```
tomelliff: This isn't done consistently but I find https://github.com/terraform-providers/terraform-provider-aws/blob/41a726c851769570036f30cd83ffc2b68a47c8dd/aws/resource_aws_lb_listener.go#L50-L52 is a nice convenience for users.

Not sure if it's worth validating this so we can generate failures at plan time because it comes with the trade off that if AWS adds another scheduling strategy we need a code change and release to support it.

erks: Will add the the `StateFunc`. Thanks for the advice @tomelliff .

Regarding the validation, that's also why I'm a bit hesitant about adding. I think it might come down to consistency, i.e. whether this provide performs validations like this somewhere else.

aws/resource_aws_ecs_service_test.go
erks: I'll change it, as it seems other existing functions are also favoring repetition here.

erks:
@bflad any input on the validation question(s) above? or any overall feedback. happy to address them, if any, so we can move forward.

tomelliff:
I wouldn't both validating the deployment configuration. You can't test for it at plan time and the AWS API should return a suitable error that explains what's wrong if it fails.

That said, if for some reason the AWS API error message isn't clear (some of them can be a bit opaque) then it might be worth throwing an error or testing for the error code and providing a different error message.

aws/resource_aws_ecs_service.go
bflad: The `scheduling_strategy` attribute is always going to exist since it has a `Default` -- you can use `input.SchedulingStrategy = d.Get("scheduling_strategy").(string)` instead of working with `d.GetOk()` or add it to the `CreateEcsService` struct directly with the same `d.Get()`

aws/resource_aws_ecs_service.go
```diff
@@ -489,7 +508,11 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("task_definition", taskDefinition)
 	}
 
-	d.Set("desired_count", service.DesiredCount)
```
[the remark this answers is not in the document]
erks: In my testing, it's the number of container instances, which can change periodically.

aws/resource_aws_ecs_service_test.go
```diff
@@ -626,6 +626,78 @@ func TestAccAWSEcsService_withLaunchTypeEC2AndNetworkConfiguration(t *testing.T)
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEcsServiceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEcsServiceWithDefaultSchedulingStrategy(clusterName, tdName, svcName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEcsServiceExists("aws_ecs_service.ghost", &service),
+					resource.TestCheckResourceAttr("aws_ecs_service.ghost", "scheduling_strategy", "REPLICA"),
```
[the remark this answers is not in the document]
erks: Did you mean `_basicImport`? I can't find the `_basic` test here.

aws/resource_aws_ecs_service.go
```diff
@@ -709,7 +732,9 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 		Cluster: aws.String(d.Get("cluster").(string)),
 	}
 
-	if d.HasChange("desired_count") {
```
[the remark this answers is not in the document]
erks: In my testing, it's the number of container instances

aws/resource_aws_ecs_service_test.go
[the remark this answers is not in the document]
bflad: Ah sorry, this resource's testing could use some standardization with everything else. The "basic" test is currently `TestAccAWSEcsService_withARN` (you'll notice the seemingly odd check for `service_registries.#` in there - it should really have all the resource attribute default checks).

aws/resource_aws_ecs_service_test.go
```diff
@@ -1955,3 +2027,88 @@ resource "aws_ecs_service" "test" {
...
+    "cpu": 128,
+    "essential": true,
+    "image": "ghost:latest",
+    "memory": 128,
+    "name": "ghost"
+  }
+]
+DEFINITION
+}
+resource "aws_ecs_service" "ghost" {
```
bflad: This test resource needs to reduce the `deployment_maximum_percent` from the default of 200:

```
=== RUN   TestAccAWSEcsService_withDaemonSchedulingStrategy
--- FAIL: TestAccAWSEcsService_withDaemonSchedulingStrategy (5.71s)
        
        * aws_ecs_service.ghost: 1 error(s) occurred:
        
        * aws_ecs_service.ghost: InvalidParameterException: The daemon scheduling strategy does not support values above 100 for maximumPercent on a deployment configuration. Change the maximumPercent value to 100, and try again
```

e.g. `deployment_maximum_percent = 100`

It might also be worth noting this in the attribute documentation and/or adding a daemon example to the documentation.

TheSimpleZ:
Hi! What is the status on this PR? Is it going to be merged anytime soon or should i find a workaround for setting up daemons in aws?

bflad:
We should be able to get this released tomorrow, more shortly.

aws/resource_aws_ecs_service.go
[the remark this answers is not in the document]
bflad: Adding the following [`DiffSuppressFunc`](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#diffsuppressfunc) to the `desired_count`, `deployment_minimum_percent`, and `deployment_maximum_percent` attributes, we can continue to always call `d.Set("desired_count", service.DesiredCount)` (along with the others) and remove the conditional in the read function. This will remove the breaking change for outputs that depend on the `desired_count` attribute always existing.

```go
	if d.Get("scheduling_strategy").(string) == ecs.SchedulingStrategyDaemon {
		return true
	}
	return false
},
```

aws/resource_aws_ecs_service_test.go
bflad: This acceptance test is still failing:

```
=== RUN   TestAccAWSEcsService_withDaemonSchedulingStrategy
--- FAIL: TestAccAWSEcsService_withDaemonSchedulingStrategy (5.99s)
        
        * aws_ecs_service.ghost: 1 error(s) occurred:
        
        * aws_ecs_service.ghost: InvalidParameterException: Both maximumPercent and minimumHealthyPercent cannot be 100 as this will block deployments.
            status code: 400, request id: d5daf50b-7949-11e8-a362-3ff63025fa34 "tf-acc-svc-w-ss-daemon-8yeejuf7"
```

By changing the create function to never include `DeploymentConfiguration` when set in `DAEMON` scheduling strategy, it then received this error during acceptance testing:

```
=== RUN   TestAccAWSEcsService_withDaemonSchedulingStrategy
--- FAIL: TestAccAWSEcsService_withDaemonSchedulingStrategy (12.56s)

		* aws_ecs_service.ghost (destroy): 1 error(s) occurred:

		* aws_ecs_service.ghost: InvalidParameterException: The daemon scheduling strategy does not support a desired count for services. Remove the desired count value and try again
			status code: 400, request id: d91107b6-794b-11e8-b194-535f28f11fde
```

The deletion function tries to drain any `ACTIVE` status services by setting their `DesiredCount` to `0` before deletion. Ignoring `DAEMON` scheduling strategy for that call, the testing passes as expected:

```
make testacc TEST=./aws TESTARGS='-run=TestAccAWSEcsService_withDaemonSchedulingStrategy'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEcsService_withDaemonSchedulingStrategy -timeout 120m
=== RUN   TestAccAWSEcsService_withDaemonSchedulingStrategy
--- PASS: TestAccAWSEcsService_withDaemonSchedulingStrategy (15.51s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	15.549s
```

aws/resource_aws_ecs_service_test.go
```diff
@@ -1955,3 +2005,61 @@ resource "aws_ecs_service" "test" {
...
+  name = "%s"
+  cluster = "${aws_ecs_cluster.default.id}"
+  task_definition = "${aws_ecs_task_definition.ghost.family}:${aws_ecs_task_definition.ghost.revision}"
+  scheduling_strategy = "DAEMON"
+  deployment_maximum_percent = 100
```
bflad: Instead of requiring a special configuration for min/max, we'll setup the resource to properly ignore those configurations and differences 👍

erks:
thanks, @bflad

bflad:

PR #4830, opened 2018-06-13
Feature/multiple roles in config

dsteinmo:

* Allow multiple assume_role() blocks in provider configuration. This change allows these roles to be assumed one after the other until the last role in the list is reached (order matters).

Motivating use case: We have a user account that has access to assume a role within our organization's root account with permissions delegated to create org accounts and to assume the newly created admin role in new accounts, `OrganizationAccountAccessRole`. However, without first making this change to `terraform-provider-aws`, we were unable to double-hop from the user account's keys, to the account-creation role (this part works already), to the OrganizationAccountAccessRole in the new account (this change enables this, plus any number of additional layers).  

```
--- PASS: TestAccAWSAvailabilityZones_basic (12.93s)
--- PASS: TestAccAWSAvailabilityZones_stateFilter (12.47s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       25.440s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ ENHANCEMENTS:
 BUG FIXES:
 
 * resource/aws_codecommit_repository: Prevent error and trigger recreation when not found during read [GH-4761]
+* resource/aws_eks_cluster: Properly export `arn` attribute [GH-4766] / [GH-4767]
 
 ## 1.22.0 (June 05, 2018)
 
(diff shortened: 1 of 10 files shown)


aeschright:
Hi @dsteinmo! We released an update to how the config files are handled that now allows the roles to be correctly chained (using `source_profile`). Does that resolve your use case? You can see the details of what was addressed here: https://github.com/hashicorp/aws-sdk-go-base/issues/4

dsteinmo:
Looks good @aeschright! We can close off this PR for the record.

PR #4839, opened 2018-06-14
[WIP] Proposed changes to support lambda code deploy

cgarvis:
Before fleshing out the code, want to make sure I was going down the right path with the new attributes.

Relavant docs:
https://docs.aws.amazon.com/sdk-for-go/api/service/codedeploy/#CreateDeploymentConfigInput
https://docs.aws.amazon.com/sdk-for-go/api/service/codedeploy/#TrafficRoutingConfig
https://docs.aws.amazon.com/sdk-for-go/api/service/codedeploy/#TimeBasedCanary
https://docs.aws.amazon.com/sdk-for-go/api/service/codedeploy/#TimeBasedLinear

### DIFF
--- a/website/docs/r/codedeploy_deployment_config.html.markdown
+++ b/website/docs/r/codedeploy_deployment_config.html.markdown
@@ -57,7 +57,9 @@ resource "aws_codedeploy_deployment_group" "foo" {
 The following arguments are supported:
 
 * `deployment_config_name` - (Required) The name of the deployment config.
+* `compute_platform` - (Optional) The compute platform can either be `Server` or `Lambda`. Default is `Server`.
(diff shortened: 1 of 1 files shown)


bflad:
I would suggest splitting out the structure for `time_based_canary` and `time_based_linear` as well. Generally we prefer to follow the nesting of the API for a few reasons:
* Its easier for operators to mentally transition between AWS CLI/CloudFormation/other SDKs if the naming is the same
* It provides us the ability to validate each underlying attribute individually (in case they differ per type)
* If the API introduces other underlying attributes for just one of the types then we do not need to continue creating our own naming like `linear_new_parameter` or need to implement custom plan time validation depending on type

bflad:
Closing this inactive pull request to clean up the open repository issues/PRs as its missing its actual implementation. If/when that occurs, we can reopen this. 👍

PR #4841, opened 2018-06-15
Added data source for aws_route_table_ids

jesseaukeman:
This allows lookup of a list of route table ids similar to the
existing functionality of the aws_subnet_ids data source.

* add a new data source aws_route_table_ids


aws/provider.go
```diff
@@ -233,6 +233,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_region":                           dataSourceAwsRegion(),
 			"aws_route":                            dataSourceAwsRoute(),
 			"aws_route_table":                      dataSourceAwsRouteTable(),
+			"aws_route_table_ids":                  dataSourceAwsRouteTableIDs(),
```
bflad: Sorry this is going to seem real pedantic, but can you please:
* Rename the data source `aws_route_tables` along with the file names
* Rename the functions to match, e.g. `dataSourceAwsRouteTables()` and `dataSourceAwsRouteTablesRead()`

This is mainly for future maintainability as the data source may be adjusted to return information other than just IDs. We unfortunately have some other `_ids` data sources (e.g. `aws_subnet_ids`) that will likely need to be renamed at some point. 😅

aws/data_source_aws_route_table_ids.go
```diff
@@ -0,0 +1,68 @@
...
+func dataSourceAwsRouteTableIDs() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsRouteTableIDsRead,
+		Schema: map[string]*schema.Schema{
+
+			"tags": tagsSchemaComputed(),
+
+			"vpc_id": {
+				Type:     schema.TypeString,
+				Required: true,
```
bflad: According the [EC2 API Reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html), `vpc_id` should not be a required argument. Its valid to search for route tables across multiple VPCs. Can you please:
* Change the attribute to `Optional: true`
* Use `d.GetOk("vpc_id")` instead of expecting it to always exist:

```go
if v, ok := d.GetOk("vpc_id"); ok {
	req.Filters = buildEC2AttributeFilterList(
		map[string]string{
			"vpc-id": v.(string),
		},
	)
}
```

* Change `d.SetId(d.Get("vpc_id").(string))` to `d.SetId(resource.UniqueId())`
* Ensure the documentation is updated for the attribute

aws/data_source_aws_route_table_ids.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	if resp == nil || len(resp.RouteTables) == 0 {
+		return fmt.Errorf("no matching route tables found for vpc with id %s", d.Get("vpc_id").(string))
+	}
+
+	routeTables := make([]string, 0)
+
+	for _, routeTable := range resp.RouteTables {
+		routeTables = append(routeTables, *routeTable.RouteTableId)
```
bflad: To prevent potential panics, we should prefer to use the SDK helper `aws.StringValue(routeTable.RouteTableId)` instead of directly dereferencing via `*`

aws/data_source_aws_route_table_ids.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	d.SetId(d.Get("vpc_id").(string))
+	d.Set("ids", routeTables)
```
bflad: When using `d.Set()` with aggregate types (e.g. TypeList/TypeSet/TypeMap) we should perform error checking: https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types

```go
if err := d.Set("ids", routeTables); err != nil {
  return fmt.Errorf("error setting ids: %s", err)
}
```

jesseaukeman:
Hi @bflad thanks for the feedback.  I believe I have updated as you requested.  please let me know.

aws/data_source_aws_route_tables_test.go
```diff
@@ -0,0 +1,119 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckVpcDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceAwsRouteTablesConfig(rInt),
+			},
+			{
+				Config: testAccDataSourceAwsRouteTablesConfigWithDataSource(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.aws_route_tables.selected", "ids.#", "3"),
```
bflad: This check is currently failing:

```
=== RUN   TestAccDataSourceAwsRouteTables
--- FAIL: TestAccDataSourceAwsRouteTables (11.13s)
    testing.go:518: Step 1 error: Check failed: Check 1/2 error: data.aws_route_tables.selected: Attribute 'ids.#' expected "3", got "4"
```

If I had to venture a guess, its because VPCs include a default route table in addition to the 3 being created in the test.

jesseaukeman: @bflad of course you are correct.  I've updated the test cases to account for that, now it succeeds in my tests.
```
jmpb:aws jaukeman$ TF_ACC=1 go test -v -run TestAccDataSourceAwsRouteTables
=== RUN   TestAccDataSourceAwsRouteTables
--- PASS: TestAccDataSourceAwsRouteTables (54.93s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       54.974s
```

website/docs/d/route_tables.html.markdown
```diff
@@ -0,0 +1,42 @@
+---
+layout: "aws"
+page_title: "AWS: aws_route_tables"
+sidebar_current: "docs-aws-datasource-route-tables"
```
bflad: This page needs a sidebar link added in `website/aws.erb`, will add post-merge.

bflad:

PR #4854, opened 2018-06-16
Fix force new behavior for aws_batch_job_definition attempts

hectcastro:
Fixes https://github.com/terraform-providers/terraform-provider-aws/pull/4697#pullrequestreview-124368089

When the `retry_strategy.attempts` attribute of the
`aws_batch_job_definition` is modified, ensure that a new resource is
created vs. update-in-place.

* Add `ForceNew` to `aws_batch_job_definition.retry_strategy.attempts`


bflad:

PR #4875, opened 2018-06-18
Add enabled_cloudwatch_logs_exports to rds cluster resource

FaHeymann:
Based on the implementation for the `db_instance` resource this is supposed to add the `enabled_cloudwatch_logs_exports` field to cluster resources as well.

### DIFF
--- a/aws/resource_aws_rds_cluster.go
+++ b/aws/resource_aws_rds_cluster.go
@@ -838,8 +838,8 @@ func flattenAwsRdsClusterResource(d *schema.ResourceData, meta interface{}, dbc 
 	d.Set("iam_database_authentication_enabled", dbc.IAMDatabaseAuthenticationEnabled)
 	d.Set("hosted_zone_id", dbc.HostedZoneId)
 
-	if dbc.EnabledCloudwatchLogsExports != nil {
-		d.Set("enabled_cloudwatch_logs_exports", dbc.EnabledCloudwatchLogsExports)
+	if err := d.Set("enabled_cloudwatch_logs_exports", flattenStringList(dbc.EnabledCloudwatchLogsExports)); err != nil {
+		return fmt.Errorf("error setting enabled_cloudwatch_logs_exports: %s", err)
 	}
 
 	var vpcg []string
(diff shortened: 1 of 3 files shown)


aws/resource_aws_rds_cluster.go
```diff
@@ -807,6 +838,10 @@ func flattenAwsRdsClusterResource(d *schema.ResourceData, meta interface{}, dbc
 	d.Set("iam_database_authentication_enabled", dbc.IAMDatabaseAuthenticationEnabled)
 	d.Set("hosted_zone_id", dbc.HostedZoneId)
 
+	if dbc.EnabledCloudwatchLogsExports != nil {
+		d.Set("enabled_cloudwatch_logs_exports", dbc.EnabledCloudwatchLogsExports)
```
bflad: This is currently failing acceptance testing:

```
=== RUN   TestAccAWSRDSCluster_importBasic
--- FAIL: TestAccAWSRDSCluster_importBasic (112.23s)
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=3) {
         (string) (len=33) "enabled_cloudwatch_logs_exports.#": (string) (len=1) "2",
         (string) (len=33) "enabled_cloudwatch_logs_exports.0": (string) (len=5) "audit",
         (string) (len=33) "enabled_cloudwatch_logs_exports.1": (string) (len=5) "error"
        }
        
FAIL
```

We prefer to add error checking when handling aggregate attribute types, which should generate an error here that the types don't match:

```go
if err := d.Set("enabled_cloudwatch_logs_exports", dbc.EnabledCloudwatchLogsExports); err != nil {
  return fmt.Errorf("error setting enabled_cloudwatch_logs_exports: %s", err)
}
```

To fix the error you'll need to convert `dbc.EnabledCloudwatchLogsExports` (a `[]*string`) to `[]string` or `[]interface{}`, which can be accomplished with `flattenStringList()`. This should cover the correct behavior:

```go
if err := d.Set("enabled_cloudwatch_logs_exports", flattenStringList(dbc.EnabledCloudwatchLogsExports)); err != nil {
  return fmt.Errorf("error setting enabled_cloudwatch_logs_exports: %s", err)
}
```

FaHeymann:
Thank you for the feedback! I hope I fixed it. Unfortunately I don't have an easy way to run the tests against an actual AWS account myself currently

bflad:

PR #4888, opened 2018-06-20
Adds support for PRIVATE endpoint type for aws_api_gateway_rest_api

edsonmarquezani:

* Adds support for private API Gateways[1]

[1] https://aws.amazon.com/blogs/compute/introducing-amazon-api-gateway-private-endpoints/

```
$ make testacc TESTARGS='-run=TestAccAWSAPIGatewayRestApi_EndpointConfiguration'
TF_ACC=1 go test ./... -v -run=TestAccAWSAPIGatewayRestApi_EndpointConfiguration -timeout 120m
=== RUN   TestAccAWSAPIGatewayRestApi_EndpointConfiguration
--- PASS: TestAccAWSAPIGatewayRestApi_EndpointConfiguration (136.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	136.921s
...
```

#### Comments
My Golang knowledge is vey limited, but instead of asking for this feature, I thought I could do it myself and contribute. The change is quite simple, yet I'm not sure the acceptance test I added is correct. Please tell me if not.
I built and tested it and it works.

### DIFF
--- a/aws/resource_aws_api_gateway_rest_api_test.go
+++ b/aws/resource_aws_api_gateway_rest_api_test.go
@@ -203,6 +203,24 @@ func TestAccAWSAPIGatewayRestApi_EndpointConfiguration(t *testing.T) {
 			},
 		},
 	})
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSAPIGatewayRestAPIDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSAPIGatewayRestAPIConfig_EndpointConfiguration(rName, "PRIVATE"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAPIGatewayRestAPIExists("aws_api_gateway_rest_api.test", &restApi),
+					resource.TestCheckResourceAttr("aws_api_gateway_rest_api.test", "endpoint_configuration.#", "1"),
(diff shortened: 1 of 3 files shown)


aws/resource_aws_api_gateway_rest_api_test.go
```diff
@@ -203,6 +203,24 @@ func TestAccAWSAPIGatewayRestApi_EndpointConfiguration(t *testing.T) {
 			},
 		},
 	})
+
+	resource.Test(t, resource.TestCase{
```
bflad: We generally prefer to create whole new top level acceptance test functions instead of adding multiple `resource.Test()` within the same function -- this allows us to parallelize our testing better. 😄 

Another complicated piece to our API Gateway testing here is that we run the service testing across multiple AWS partitions (Commercial and GovCloud (US)), which the latter may or may not support the new `PRIVATE` REST API endpoint type. If you look at the `TestStep` above your addition here, we have a `PreConfig` function that ensures the AWS API Gateway service its calling supports the `EDGE` endpoint type or skips the rest of the test. We'll want to add something similar for `PRIVATE` as well.

Do not be discouraged here though! There just happens to be more advanced testing handling with this resource testing than other parts of our resource testing. You actually were on the right track except the tweaks mentioned above, which I can handle in a commit after yours to expedite releasing this enhancement for everyone. 👍

edsonmarquezani: Ok, I get it. I just didn't know private type may not be supported in some regions.

As for the remaing, I understand you're going to make the adjustments yourself, right? Anyway, feel free to ask for it, if you want to. I think I could handle it. ;)

bflad:

PR #4938, opened 2018-06-21
Force all database names to lower case

### LINKED ISSUE #84: aws_db_instance name should be lowercase
_This issue was originally opened by @danosipov as hashicorp/terraform#5286. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Using incorrect symbols in the aws_db_instance name may create the RDS instance, but upon subsequent plans forces RDS instance to be recreated:

```
-/+ aws_db_instance.default
    ...
    name:                       "druid" => "Druid" (forces new resource)
    ...
```

tomelliff:
The Oracle DB engine at least uppercases the database name so if provided with 'foo' will return 'FOO'.

Previously this would then make the next plan show a change from 'FOO' -> 'foo' which required a recreation even though it wasn't actually changing anything.

Instead we now just turn everything into lowercasing and ignoring the casing diffs in case anyone was using uppercase characters at all which would then show a diff for them for no good reason.

I originally came across this when trying to reproduce https://github.com/terraform-providers/terraform-provider-aws/issues/84 and tested all of the engines and could only see a diff for Oracle (as @brandonstevens has also discovered since apparently).

Fixes #84 

* Ignores casing for all database names to prevent the AWS API mangling them for different version names and causing a diff loop.

```
$ make testacc TESTARGS="-run=TestAccAWSDBInstance_nameDiffSuppress"
TF_ACC=1 go test ./... -v -run=TestAccAWSDBInstance_nameDiffSuppress -timeout 120m
=== RUN   TestAccAWSDBInstance_nameDiffSuppress
--- PASS: TestAccAWSDBInstance_nameDiffSuppress (855.88s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	855.901s
$ make testacc TESTARGS="-run=TestAccAWSDBInstance_basic"
TF_ACC=1 go test ./... -v -run=TestAccAWSDBInstance_basic -timeout 120m
=== RUN   TestAccAWSDBInstance_basic
--- PASS: TestAccAWSDBInstance_basic (450.39s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	450.417s
```

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -39,6 +39,13 @@ func resourceAwsDbInstance() *schema.Resource {
 				Optional: true,
 				Computed: true,
 				ForceNew: true,
+				StateFunc: func(v interface{}) string {
+					value := v.(string)
+					return strings.ToLower(value)
+				},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return strings.ToLower(old) == strings.ToLower(new)
+				},
 			},
 
 			"arn": {
@@ -954,7 +961,7 @@ func resourceAwsDbInstanceRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
-	d.Set("name", v.DBName)
+	d.Set("name", strings.ToLower(*v.DBName))
 	d.Set("identifier", v.DBInstanceIdentifier)
 	d.Set("resource_id", v.DbiResourceId)
 	d.Set("username", v.MasterUsername)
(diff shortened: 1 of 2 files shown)


tomelliff:
Hmmm... thinking about it a bit more I'm not sure this is a good idea.

There could easily be people using uppercased database names (PascalCasing is semi common when not lower-casing everything in databases) and this would silently swallow that and force it to lower casing. All to save a reasonably awkward corner case with Oracle when it should just be obvious to the user that they need to update their database name to all uppercase to prevent the diff and subsequent recreation.

I'm going to close this for now unless anyone has any strong objections to that and would like to see it reopened and discussed further.

PR #4943, opened 2018-06-21
Raise an error when failing to retrieve RDS instance tags

### LINKED ISSUE #345: Tags repeatedly marked as not being set correctly for aws_db_instance
_This issue was originally opened by @davedash as hashicorp/terraform#9821. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

In Terraform 0.7.8 (and prior):

terraform plan gives me:

```
~ module.dev.aws_db_instance.db
    tags.%:      "0" => "3"
    tags.Name:   "" => "dbdev"
    tags.dbrole: "" => "master"
    tags.env:    "" => "dev"
```

My tf file looks like:

```

resource "aws_db_instance" "db" {
    ...
    engine               = "postgres"
    engine_version       = "9.5.2"
    instance_class       = "db.m3.medium"
    db_subnet_group_name = "${aws_db_subnet_group.db.name}"
    storage_encrypted    = true
    parameter_group_name = "${var.parameter_group}"  # summit custom
    allow_major_version_upgrade = true

    tags {
      Name = "${var.env}"
      dbrole = "master"
      env = "${var.env}"
    }
...
```

After an apply, the tags are set.  Re-running plan gives me the same output, so I'm always in a dirty state.

This is the case for all of my rds instances.

tomelliff:
Just logging the error means that if the user is unable to call the ListTagsForResource API endpoint then they will see a diff showing that they need to add the tags each time.
Instead it's better to actually error instead of hiding this from the user (unless they happen to check the logs).

Closes https://github.com/terraform-providers/terraform-provider-aws/issues/345

See extended discussion in the original issue: https://github.com/hashicorp/terraform/issues/9821

Fixes #345

* Error instead of just logging when user is unable to call ListTagsForResource.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -1035,7 +1035,7 @@ func resourceAwsDbInstanceRead(d *schema.ResourceData, meta interface{}) error {
 	})
 
 	if err != nil {
-		log.Printf("[DEBUG] Error retrieving tags for ARN: %s", arn)
+		return fmt.Errorf("Error retrieving tags for ARN: %s", arn)
 	}
 
 	var dt []*rds.Tag


bflad:

PR #4946, opened 2018-06-22
Force new route when changing the route table

### LINKED ISSUE #1841: aws_route thinks route_table_id is a modifyable resource - its not
Terraform 0.9.8

This is super important, as terraform basically leaves routes hanging around which completely screws with your networking setup. 

1. Set up an aws_route, with a route_table_id
2. Terraform it.
3. Now change the route to be a different route_table_id

A plan now makes this a modify step (~) - it should be a destroy and create step (-/+)

An apply will error saying it could not find the old routes, (its looking for them on the wrong wroute table), and will then remove them from the state file(!!!!) - the next plan/apply will just apply them on the new route table, leaving the old routes behind on the old route table... this will break your entire networking model and potentially the security of your networks. 

Really hope its been fixed since terraform 0.9.8 but havent had the chance to test yet - screwing up your networking is a pretty horrible thing for terraform to do!

If you're wondering why the route table might change, its because it may well come from a data source, and you've decided to change the VPC peering from A=>B, to A=>C - which is a common thing to do in e.g. blue-green type deployments

tomelliff:
Before this change, changing the route table of a route showed an in place modification for Terraform to perform.
AWS actually creates a new route in the new route table with the ReplaceRoute API call but this left the old route lying around and no longer managed.

The acceptance test provided in this commit was failing before because Terraform was attempting to update the new route's 10.3.0.0/16 route to point to the new target but the 10.3.0.0/16 route didn't exist in that route table and instead wanted Terraform to call the CreateRoute API instead:

> There is no route defined for '10.3.0.0/16' in the route table. Use CreateRoute instead.

Instead we should just use ForceNew to make Terraform create a new route in the new route table and then delete the old one.

Fixes #1841

* Changing the route table of a route should force a new resource

```
$ make testacc TESTARGS="-run=TestAccAWSRoute_changeRouteTable"
TF_ACC=1 go test ./... -v -run=TestAccAWSRoute_changeRouteTable -timeout 120m
=== RUN   TestAccAWSRoute_changeRouteTable
--- PASS: TestAccAWSRoute_changeRouteTable (161.83s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	161.850s
```

### DIFF
--- a/aws/resource_aws_route.go
+++ b/aws/resource_aws_route.go
@@ -98,6 +98,7 @@ func resourceAwsRoute() *schema.Resource {
 			"route_table_id": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 
 			"vpc_peering_connection_id": {
--- a/aws/resource_aws_route_test.go
+++ b/aws/resource_aws_route_test.go
@@ -166,6 +166,33 @@ func TestAccAWSRoute_ipv6ToPeeringConnection(t *testing.T) {
 	})
 }
 
+func TestAccAWSRoute_changeRouteTable(t *testing.T) {
+	var before ec2.Route
+	var after ec2.Route
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		Providers:    testAccProviders,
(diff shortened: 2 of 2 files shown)


bflad:

PR #4969, opened 2018-06-24
issue #4499 data source aws_ssm_parameter returns string for StringList type

### LINKED ISSUE #4499: data source aws_ssm_parameter returns string for StringList type.
### Community Note

### Terraform Version

Terraform 0.11.7

provider.aws v1.13.0

### Affected Resource(s)

* aws_ssm_parameter

```hcl
resource "aws_ssm_parameter" "my_subnets" {
  name      = "/my/my_subnets"
  type      = "StringList"
  overwrite = true
  value     = "1.1.1.1,2.2.2.2"
}

data "aws_ssm_parameter" "my_subnets" {
  name = "/my/my_subnets"
}

output "subnets_string" {
 value = "${data.aws_ssm_parameter.my_subnets.value}"
}

output "subnets_list" {
  value = "${split(",", data.aws_ssm_parameter.private_subnets.value)}"
}
```

### Debug Output

N/A

### Panic Output

N/A

### Expected Behavior

data source aws_ssm_parameter returns list for StringList type.

### Actual Behavior

data source aws_ssm_parameter returns string for StringList type.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

N/A

### References

N/A

saravanan30erd:
Fixes #4499 

* Change 1
Added the attribute value_list to return list of values if type is `StringList`

### DIFF
--- a/aws/data_source_aws_ssm_parameter.go
+++ b/aws/data_source_aws_ssm_parameter.go
@@ -33,6 +33,12 @@ func dataSourceAwsSsmParameter() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+			"value_list": {
+				Type:     schema.TypeSet,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+				Computed: true,
+			},
 			"with_decryption": {
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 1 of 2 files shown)


bflad:
Hi @saravanan30erd, 👋  Thanks for submitting this and sorry I probably should have been more clear in https://github.com/terraform-providers/terraform-provider-aws/issues/4499#issuecomment-388488052 that we likely would not want to implement something like this. A few reasons why we would like to not pursue something like this:
* Its currently possible to achieve the same exact behavior using [`split()`](https://www.terraform.io/docs/configuration/interpolation.html#split-delim-string-), e.g. `split(",", data.aws_ssm_parameter.example.value)`
* The `value_list` attribute is only available during certain situations which makes module creation difficult (`output` references currently require the attribute to always be present, although this might be mitigated with some changes in Terraform 0.12). Its also a little confusing to have two places to get the same information.
* It will not be possible to extend `value_list` beyond strings, should SSM ever support something like an integer list or boolean list. The resource handling in Terraform core currently only works with a single explicit type for list/map elements, in this case strings, which could be confusing in the future.

Ideally, I think we should just document how to handle `StringList` parameters with the split example. This matches how you would interact with the API and saves any future maintenance burden of creating our own implementation here.

PR #4971, opened 2018-06-25
Fixed an example for acm_certificate_validation

adamlc:

* In the example for acm_certificate_validation, zone_id in aws_route53_record.cert_validation was incorrectly a data attribute.

N/A


bflad:
Hi @adamlc 👋  Thanks for submitting this.

It looks like on line 31, it is a data source for `aws_route53_zone` and not a resource that needs to be referenced in the example: `data "aws_route53_zone" "zone" {`

Can you explain why this change is necessary or if I missed something here? Thanks!

adamlc:
@bflad You're totally right! Seems I was having a senior moment 🤦‍♂️

PR #4975, opened 2018-06-25
r/Instance: linking to the Key Pair resource

tombuildsstuff:
It's not immediately obvious that this can be sourced from a Key Pair name - this PR exposes that this can be sourced from another resource.


website/docs/r/instance.html.markdown
```diff
@@ -67,10 +67,15 @@ instance. Amazon defaults this to `stop` for EBS-backed instances and
 instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#Using_ChangingInstanceInitiatedShutdownBehavior) for more information.
 * `instance_type` - (Required) The type of instance to start. Updates to this field will trigger a stop/start of the EC2 instance.
 * `key_name` - (Optional) The key name to use for the instance.
+
+-> **NOTE:** This can be sourced from [the `aws_key_pair` data-source/resource](key_pair.html).
```
bflad: I'm not sure this warrants a full call out box - can we just make it a normal sentence for the attribute and/or include it in the examples with `aws_key_pair.example.name`. Also we do not currently have a data source for `aws_key_pair` so we should probably remove that bit. 😄

tombuildsstuff: 👍 updated

Issue #4633, opened 2018-05-23
Lambda was unable to configure access to your environment  variables because the KMS key is invalid for CreateGrant


kylelaverty:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
Terraform v0.11.7
+ provider.aws v1.17.0
```

### Affected Resource(s)

* aws_lambda_function

### Expected Behavior

Creating a Lambda that uses a role should work.

### Actual Behavior

Instead we are getting the following error sometimes:

```
Error creating Lambda function: InvalidParameterValueException: Lambda was unable to configure access to your environment 
variables because the KMS key is invalid for CreateGrant. Please check your KMS key settings. KMS Exception: InvalidArnException 
KMS Message: ARN does not refer to a valid principal:
arn:aws:sts::[account #]:assumed-role/[build]_[build try #]_[role name]/[build]_[build try #]_[lambda name]
```

### Steps to Reproduce

1. `terraform apply`
2. Luck and a lot of lambdas

### Important Factoids

It is most likely because of this section of code: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lambda_function.go#L367

It appears to be treating the error as a `NonRetryableError` which does not seem to line up with what actually happens.

Trying to rerun the Terraform eventually works, even though everything is recreated each time.

dshmelev:
The problem is also reproduced on provider "aws" (1.23.0)

emmekappa:
Just have had the same issue

denishonig:
Hi!
Is there any information when it is going to be fixed?

Thanks.

simonvanderveldt:
Just ran into this as well, when applying the plan again it does work, so it's probably a dependency/waiting/retrying issues. Can't really reproduce it though, so not sure how to make this report more specific.

bflad:
A potential fix for this has been merged into master and will release with version 1.36.0 of the AWS provider, likely later today. (I say potential fix because as with many eventual consistency issues, they are hard to consistently reproduce. 😄 )

bflad:
[...]

PR #4998, opened 2018-06-26
Adding additional note to kms_key_arn attribute

AHaymond:
I ran into an issue that if you include the ksm_key_arn without adding something for it to actually encrypt (such as ENV vars) the resource will silently fail to build during the plan phase and will instead throw an error if the resource is used as a variable in another declaration. I propose the documentation be updated to help save people hours of confusion.

the offending error returned because of this issue is:

```
aws_lambda_event_source_mapping.my_mapping: Resource 'aws_lambda_function.my_function' not found for variable 'a
ws_lambda_function.my_function'


bflad:
Can you run Terraform with debug logging enabled to determine how the `aws_lambda_function` resource is hanging? e.g. `TF_LOG=debug terraform apply`

I'm wondering if this is related to #4633 and not necessary a lack of environment variable configuration

bflad:
Closing due to lack of response/confirmation. If someone provides additional insight here we can revisit this. 👍

PR #5002, opened 2018-06-27
Import acm certificate (like IAM server certificate)

### LINKED ISSUE #3560: Add ability to import a certificate with aws_acm_certificate

julienduchesne:
Fixes #3560

I tried to do what @bflad proposed in the issue (in this branch https://github.com/julienduchesne/terraform-provider-aws/tree/import-acm-certificates) but ran into a major issue. CustomizeDiff doesn't see fields if their value is from another resource (example: tls certificate resource). Therefore, plan-time field validation did not work in those cases. I opted for a new resource instead since the creation and update methods are different anyways. It also makes for a cleaner field description set (required and optional fields are clearly marked for both resource). I managed to reuse the Read method as well as the Delete.

Let me know what you think!

```
make testacc TESTARGS='-run=TestAccAWSAcmCertificateImported_selfSigned'
TF_ACC=1 go test ./... -v -run=TestAccAWSAcmCertificateImported_selfSigned -timeout 120m
=== RUN   TestAccAWSAcmCertificateImported_selfSigned
--- PASS: TestAccAWSAcmCertificateImported_selfSigned (35.83s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       35.851s

...
```


julienduchesne:
If you approve this, I'll write the doc (don't want to do it for nothing)

bflad:
Thanks for your work here @julienduchesne! 🥇 From an architecture standpoint, I'm not sure we should create a new resource due to lack of plan-time validation (at the moment) or because you interact with the API slightly differently in the separate resource creation cases.

We prefer to handle a singular API "object" for management, in this case an ACM certificate, in one resource where possible. We have plenty of other resources which handle many different types of API calls depending on how you are trying to manage the resource (notably `aws_db_instance` which supports regular creation, restore from S3, restore from backup, etc. in the same resource). It simplifies the management from an operator perspective to keep them all wrapped together in a single resource, although as you mention pushes the burden of implementation details into the code. I personally find its our responsibility as maintainers to create better overall user experience where possible even at the expense of potential bugs. We can iterate as necessary. 😄 

With respect to some of the issues you found while trying to implement this, if there are improvements that need to happen in the Terraform provider SDK or Terraform core, we can submit those issues upstream. 👍 I have found in quite a few cases like this in my experience working in the provider. For example, it would helpful if the resource schema supported some form of "output format" for computed attributes (for plan-time validations in downstream resources, e.g. an incoming computed attribute is actually an ARN format so ValidateFunc can properly work). Another case is specifying that a computed attribute will always be set, which is like what you mention about not being able to use `CustomizeDiff` with a reference from another resource.

I hope this makes sense. Please reach out with any questions, thanks again!

julienduchesne:
Yes, this makes sense. I didn't know if the CustomizeDiff limitation was as designed or something that could be added but that clears it up. I will work on this as soon as I get the time.

Issue #5009, opened 2018-06-27
data aws_pricing_instance


julienduchesne:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Hi, I currently have a python script that retrieves the on-demand price of an ec2 instance with the pricing api that I'd like to use in a terraform context (setting a spot_price that is a fraction of on-demand). It has the following parameters: instance_type string, region string, os string (linux or windows). Would this be an appropriate feature for terraform? I'd implement it myself.

### New or Affected Resource(s)

* aws_pricing_instance

### Potential Terraform Configuration

```hcl
data "aws_pricing_instance" "launch_config_1" {
  region = "us-east-1"
  instance_type = "c5.large"
  os = "LINUX"
}
resource "aws_launch_configuration" "launch_config_1" {
  name_prefix          = "Cluster"
  image_id             = "..."
  instance_type        = "c5.large"
  spot_price           = "${data.aws_pricing_instance.launch_config_1.hourly * 0.5}"
  user_data            = "..."
  key_name             = "..."
  iam_instance_profile = "..."
  security_groups      = ["..."]
}
```

bflad:
From a maintainability standpoint, we may want to consider implementing this with the generic filtering as there are a lot of dimensions associated even with only the `AmazonEC2` service code:

```
$ aws --region us-east-1 pricing describe-services --service-code amazonec2
{
    "Services": [
        {
            "ServiceCode": "AmazonEC2",
            "AttributeNames": [
                "volumeType",
                "maxIopsvolume",
                "instance",
                "instanceCapacity10xlarge",
                "locationType",
                "instanceFamily",
                "operatingSystem",
                "clockSpeed",
                "LeaseContractLength",
                "ecu",
                "networkPerformance",
                "instanceCapacity8xlarge",
                "group",
                "maxThroughputvolume",
                "gpuMemory",
                "ebsOptimized",
                "elasticGpuType",
                "maxVolumeSize",
                "gpu",
                "processorFeatures",
                "intelAvxAvailable",
                "instanceCapacity4xlarge",
                "servicecode",
                "groupDescription",
[...]

PR #4716, opened 2018-05-31
Allow setting endpoint for EFS

joseph-stratoscale:
Adding support for setting user defined end-point

Changes proposed in this pull request:

* provider.go
* config.go
* website/docs/index.html.markdown

Github issue
https://github.com/terraform-providers/terraform-provider-aws/issues/3888

PR #5017, opened 2018-06-28
Add pricing vendor and client

julienduchesne:
Will be used for #5009

* Adds the "pricing" vendor dependency
* Adds a pricingconn
* Adds a custom pricing endpoint (pricing only has an API in us-east-1 and ap-south-1)


aws/provider.go
```diff
@@ -835,6 +835,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		config.KinesisEndpoint = endpoints["kinesis"].(string)
 		config.KmsEndpoint = endpoints["kms"].(string)
 		config.LambdaEndpoint = endpoints["lambda"].(string)
+		config.PricingEndpoint = endpoints["pricing"].(string)
```
bflad: It doesn't look like everything is setup to correctly support the custom endpoint here. There are a few additional things required as shown in #4716. To not setup the customized endpoint, you can simply add handling similar to #4049

julienduchesne:
Oops, sorry about that. I actually just removed the custom endpoint option since the API just refused calls from a provider with region other than the endpoint's region and I don't see much use in anyone implementing their own pricing API.

PR #5019, opened 2018-06-28
New Resource: aws_s3_bucket_inventory

### LINKED ISSUE #532: Add AWS S3 inventory configuration support
_This issue was originally opened by @mfischer-zd as hashicorp/terraform#11943. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Amazon added a new S3 API to configure a new "inventory" feature in November 2016:

http://docs.aws.amazon.com/AmazonS3/latest/dev/storage-inventory.html

This API should be supported by the AWS provider.

danielwhite:
Fixes #532

* Introduces a new resource for managing [Amazon S3 Inventory](https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-inventory.html#storage-inventory-kms-key-policy).

```
$ make testacc TESTARGS='-run TestAccAWSS3BucketInventory'
TF_ACC=1 go test ./... -v -run TestAccAWSS3BucketInventory -timeout 120m
=== RUN   TestAccAWSS3BucketInventory_basic
--- PASS: TestAccAWSS3BucketInventory_basic (63.01s)
=== RUN   TestAccAWSS3BucketInventory_encryptWithSSES3
--- FAIL: TestAccAWSS3BucketInventory_encryptWithSSES3 (41.02s)
		
		* aws_s3_bucket_inventory.test: 1 error(s) occurred:
		
		* aws_s3_bucket_inventory.test: Error putting S3 bucket inventory configuration: MalformedXML: The XML you provided was not well-formed or did not validate against our published schema
			status code: 400, request id: 41C41887B3D56896, host id: ArmdzLF7RPfvH/UNpgddZ86+XmR6OLB+icTo6OkPiHe+k1vyz9vNS5TJ61KBC3VB2taNf+Lk8Pw=
=== RUN   TestAccAWSS3BucketInventory_encryptWithSSEKMS
--- PASS: TestAccAWSS3BucketInventory_encryptWithSSEKMS (99.58s)
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	203.657s
make: *** [testacc] Error 1
```

Pending:

- [x] Limitation in the AWS SDK prevents creating of resources with SSE-S3 encryption. See: https://github.com/aws/aws-sdk-go/issues/2015. **Note:** Resources with this enabled will fail to be imported or created until the underlying issue is resolved.

### DIFF
--- a/aws/resource_aws_s3_bucket_inventory.go
+++ b/aws/resource_aws_s3_bucket_inventory.go
@@ -287,14 +287,26 @@ func resourceAwsS3BucketInventoryRead(d *schema.ResourceData, meta interface{}) 
 	}
 
 	log.Printf("[DEBUG] Reading S3 bucket inventory configuration: %s", input)
-	output, err := conn.GetBucketInventoryConfiguration(input)
-	if err != nil {
-		if isAWSErr(err, s3.ErrCodeNoSuchBucket, "") || isAWSErr(err, "NoSuchConfiguration", "The specified configuration does not exist.") {
-			log.Printf("[WARN] %s S3 bucket inventory configuration not found, removing from state.", d.Id())
-			d.SetId("")
-			return nil
+	var output *s3.GetBucketInventoryConfigurationOutput
+	err = resource.Retry(1*time.Minute, func() *resource.RetryError {
+		var err error
+		output, err = conn.GetBucketInventoryConfiguration(input)
+		if err != nil {
+			if isAWSErr(err, s3.ErrCodeNoSuchBucket, "") || isAWSErr(err, "NoSuchConfiguration", "The specified configuration does not exist.") {
+				if d.IsNewResource() {
+					return resource.RetryableError(err)
+				}
+				return nil
+			}
+			return resource.NonRetryableError(err)
 		}
-		return err
+		return nil
+	})
+
+	if output == nil || output.InventoryConfiguration == nil {
+		log.Printf("[WARN] %s S3 bucket inventory configuration not found, removing from state.", d.Id())
+		d.SetId("")
+		return nil
 	}
 
 	d.Set("enabled", aws.BoolValue(output.InventoryConfiguration.IsEnabled))
--- a/aws/resource_aws_s3_bucket_inventory.go
+++ b/aws/resource_aws_s3_bucket_inventory.go
@@ -1,0 +1,451 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsS3BucketInventory() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsS3BucketInventoryPut,
+		Read:   resourceAwsS3BucketInventoryRead,
+		Update: resourceAwsS3BucketInventoryPut,
+		Delete: resourceAwsS3BucketInventoryDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"bucket": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateMaxLength(64),
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Default:  true,
+				Optional: true,
+			},
+			"filter": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"prefix": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"destination": {
+				Type:     schema.TypeList,
+				Required: true,
+				MaxItems: 1,
+				MinItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"bucket": {
+							Type:     schema.TypeList,
+							Required: true,
+							MaxItems: 1,
+							MinItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"format": {
+										Type:     schema.TypeString,
+										Required: true,
+										ValidateFunc: validation.StringInSlice([]string{
+											s3.InventoryFormatCsv,
+											s3.InventoryFormatOrc,
+										}, false),
+									},
+									"bucket_arn": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validateArn,
+									},
+									"account_id": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"prefix": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"encryption": {
+										Type:     schema.TypeList,
+										Optional: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"sse_kms": &schema.Schema{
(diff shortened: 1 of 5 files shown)


aws/resource_aws_s3_bucket_inventory.go
```diff
@@ -0,0 +1,445 @@
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"prefix": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"destination": s3BucketInventoryDestinationSchema(),
```
bflad: Minor nitpick: We are preferring to statically define the full schema inline in all its nested glory. This is for a few various reasons (future plans for automated code/documentation generation, linting, etc.), but not yet documented on the Terraform documentation website surrounding provider development. Hopefully it will be soon. 👍

aws/resource_aws_s3_bucket_inventory.go
```diff
@@ -0,0 +1,445 @@
...
+
+	input := &s3.GetBucketInventoryConfigurationInput{
+		Bucket: aws.String(bucket),
+		Id:     aws.String(name),
+	}
+
+	log.Printf("[DEBUG] Reading S3 bucket inventory configuration: %s", input)
+	output, err := conn.GetBucketInventoryConfiguration(input)
+	if err != nil {
+		if isAWSErr(err, s3.ErrCodeNoSuchBucket, "") || isAWSErr(err, "NoSuchConfiguration", "The specified configuration does not exist.") {
```
bflad: Using [`d.IsNewResource()`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceData.IsNewResource) as a conditional for guidance, we should setup the resource to allow retries when recently created due to S3 eventual consistency. e.g.

```go
var output s3.GetBucketInventoryConfigurationOutput
err := resource.Retry(1*time.Minute, func() resource.RetryError {
  var err error
  output, err = conn.GetBucketInventoryConfiguration(input)
  if err != nil {
    if isAWSErr(err, s3.ErrCodeNoSuchBucket, "") || isAWSErr(err, "NoSuchConfiguration", "The specified configuration does not exist.") {
      if d.IsNewResource() {
      }
      // output should be nil in this case and caught later (or you can duplicate the NoSuchBucket/Configuration conditional in resource.Retry err check to remove resource from state)
      return nil
    }
  }
  return nil
})
if err != nil {
  return fmt.Errorf("error reading S3 Bucket Inventory Configuration: %s", err)
}

if output == nil || output.InventoryConfiguration == nil {
  log.Printf("[WARN] %s S3 bucket inventory configuration not found, removing from state.", d.Id())
  d.SetId("")
  return nil
}
```

aws/resource_aws_s3_bucket_inventory.go
```diff
@@ -0,0 +1,445 @@
...
+			log.Printf("[WARN] %s S3 bucket inventory configuration not found, removing from state.", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("enabled", *output.InventoryConfiguration.IsEnabled)
```
bflad: To prevent potential panics, we should prefer to use the SDK provided function for dereferencing values (`aws.BoolValue(output.InventoryConfiguration.IsEnabled)`) instead of directly dereferencing via `*`.

Same applies to `included_object_versions` below.

aws/resource_aws_s3_bucket_inventory.go
```diff
@@ -0,0 +1,445 @@
...
+	d.Set("included_object_versions", *output.InventoryConfiguration.IncludedObjectVersions)
+	d.Set("optional_fields", flattenStringList(output.InventoryConfiguration.OptionalFields))
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("optional_fields", flattenStringList(output.InventoryConfiguration.OptionalFields)); err != nil {
  return fmt.Errorf("error setting optional_fields: %s", err)
}
```

This applies to `filter`, `schedule`, and `destination` below as well.

aws/resource_aws_s3_bucket_inventory.go
```diff
@@ -0,0 +1,445 @@
...
+	result = append(result, m)
+
+	return result
+}
+
+func flattenS3InventorySchedule(schedule *s3.InventorySchedule) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, 1)
+
+	m := make(map[string]interface{}, 1)
+	m["frequency"] = *schedule.Frequency
```
bflad: To prevent potential panics, we should prefer to use the SDK provided function for dereferencing values (`aws.StringValue(schedule.Frequency)`) instead of directly dereferencing via `*`.

aws/resource_aws_s3_bucket_inventory.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	return destination
+}
+
+func flattenS3InventoryS3BucketDestination(destination *s3.InventoryS3BucketDestination) []map[string]interface{} {
+	result := make([]map[string]interface{}, 0, 1)
+
+	m := map[string]interface{}{
+		"format":     *destination.Format,
```
bflad: Same with these -- to prevent potential panics, we should prefer to use the SDK provided function for dereferencing values (`aws.StringValue(destination.Format)`) instead of directly dereferencing via `*`.

website/docs/r/s3_bucket_inventory.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `bucket` - (Required) The name of the bucket to put inventory configuration.
+* `name` - (Required) Unique identifier of the inventory configuration for the bucket.
+* `enabled` - (Optional, Default: true) Specifies whether the inventory is enabled or disabled.
+* `filter` - (Optional) Object filtering that accepts a prefix (documented below).
+* `s3_bucket_destination` - (Required) Destination bucket where inventory list files are written (documented below).
```
bflad: Minor nitpick: Can you sort all the Required attributes at the top please?

website/docs/r/s3_bucket_inventory.html.markdown
```diff
@@ -0,0 +1,127 @@
+---
+layout: "aws"
+page_title: "AWS: aws_s3_bucket_inventory"
+sidebar_current: "docs-aws-resource-s3-bucket-inventory"
```
bflad: This resource documentation page needs a sidebar link added to `website/aws.erb` 👍

bflad:
Regarding this issue:

```
		* aws_s3_bucket_inventory.test: Error putting S3 bucket inventory configuration: MalformedXML: The XML you provided was not well-formed or did not validate against our published schema
			status code: 400, request id: 41C41887B3D56896, host id: ArmdzLF7RPfvH/UNpgddZ86+XmR6OLB+icTo6OkPiHe+k1vyz9vNS5TJ61KBC3VB2taNf+Lk8Pw=
```

If you would like, feel free to remove that support for now in the resource or error in that configuration, and we can (hopefully) ship this new resource after the other fixes. 😄

danielwhite:
@bflad, thanks for the quick feedback. I'll start addressing the feedback right away.

danielwhite:
I believe I've addressed each point.

With regards to the SSE-S3 support, I've flagged it as an error when reading to avoid any surprises. If you were to just remove the `sse_s3` option entirely, then it risks quietly importing the resource and later stomping on the option by not having any encryption enabled.

I've also marked the test as permanently skipped since the only thing preventing it from working is the failing PUT request.
If removing the support entirely seems better, then I'm also happy to do that.

website/aws.erb
```diff
@@ -1861,6 +1861,10 @@
                             <a href="/docs/providers/aws/r/s3_bucket.html">aws_s3_bucket</a>
                         </li>
 
+                        <li<%= sidebar_current("docs-aws-resource-s3-bucket-inventory") %>>
+                            <a href="/docs/providers/aws/r/s3_bucket_metric.html">aws_s3_bucket_inventory</a>
```
bflad: Typo 😎  should be `s3_bucket_inventory.html`

danielwhite: D'oh! Thanks for the fix.

bflad:

PR #5024, opened 2018-06-28
Make EventSourceMapping.starting_position optional.

CyrusNajmabadi:
Amazon recently made it possible to have SQS events trigger Lambdas.  See: https://aws.amazon.com/blogs/aws/aws-lambda-adds-amazon-simple-queue-service-to-supported-event-sources/ for more details. 

As part of this they specifically utilize CreateSourceEventMapping to describe the mapping between SQS and Lambda: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

*However*, as part of this, they changed the StartingPosition message property from required to optional.  You can see that it was previous required here: https://web.archive.org/web/20171222204154/https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html

```
StartingPosition
    The position in the stream where Lambda starts reading. For valid values, see CreateEventSourceMapping in the AWS Lambda Developer Guide.
    Required: Yes
    Type: String
    Update requires: Replacement
```

But now is optional in the current set of docs:

```
StartingPosition
    The position in the stream where Lambda starts reading. For valid values, see CreateEventSourceMapping in the AWS Lambda Developer Guide.
    Required: No
    Type: String
    Update requires: Replacement
```

Unfortunately, this makes is (afaict) impossible to use terraform to configure this EventSourceMapping.  If i leave out the starting_position property, terraform complains with:

```
Plan apply failed:
Error creating Lambda event source mapping: ValidationException: 1 validation error detected: Value '' at 'startingPosition' failed to satisfy constraint: Member must satisfy enum value set: [LATEST, AT_TIMESTAMP, TRIM_HORIZON]
```

However, if i supply any of those, AWS itself errors out with:

```
Plan apply failed: 
Error creating Lambda event source mapping: InvalidParameterValueException: StartingPosition is not valid for SQS event sources.
```

This PR attempts to rectify things by making this property optional to match current AWS expectations.

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/5025

```
cyrusn@matebunty ~/go/src/github.com/terraform-providers/terraform-provider-aws[pulumi-master ≡]> make testacc TESTARGS='-run=TestAccAWSAvailabilityZones'
--- PASS: TestAccAWSAvailabilityZones_basic (28.19s)
--- PASS: TestAccAWSAvailabilityZones_stateFilter (29.25s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	57.462s```

### DIFF
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -87,11 +87,14 @@ func resourceAwsLambdaEventSourceMappingCreate(d *schema.ResourceData, meta inte
 	log.Printf("[DEBUG] Creating Lambda event source mapping: source %s to function %s", eventSourceArn, functionName)
 
 	params := &lambda.CreateEventSourceMappingInput{
-		EventSourceArn:   aws.String(eventSourceArn),
-		FunctionName:     aws.String(functionName),
-		StartingPosition: aws.String(d.Get("starting_position").(string)),
-		BatchSize:        aws.Int64(int64(d.Get("batch_size").(int))),
-		Enabled:          aws.Bool(d.Get("enabled").(bool)),
+		EventSourceArn: aws.String(eventSourceArn),
+		FunctionName:   aws.String(functionName),
+		BatchSize:      aws.Int64(int64(d.Get("batch_size").(int))),
+		Enabled:        aws.Bool(d.Get("enabled").(bool)),
+	}
+
+	if startingPosition := d.Get("starting_position"); startingPosition != "" {
+		params.StartingPosition = aws.String(startingPosition.(string))
 	}
 
 	// IAM profiles and roles can take some time to propagate in AWS:
@@ -170,7 +173,19 @@ func resourceAwsLambdaEventSourceMappingDelete(d *schema.ResourceData, meta inte
 		UUID: aws.String(d.Id()),
 	}
 
-	_, err := conn.DeleteEventSourceMapping(params)
+	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
+		_, err := conn.DeleteEventSourceMapping(params)
+		if err != nil {
+			if awserr, ok := err.(awserr.Error); ok {
+				if awserr.Code() == "ResourceInUseException" {
+					return resource.RetryableError(awserr)
+				}
+			}
+			return resource.NonRetryableError(err)
+		}
+		return nil
+	})
+
 	if err != nil {
 		return fmt.Errorf("Error deleting Lambda event source mapping: %s", err)
 	}
@@ -196,7 +211,9 @@ func resourceAwsLambdaEventSourceMappingUpdate(d *schema.ResourceData, meta inte
 		_, err := conn.UpdateEventSourceMapping(params)
 		if err != nil {
 			if awserr, ok := err.(awserr.Error); ok {
-				if awserr.Code() == "InvalidParameterValueException" {
+				if awserr.Code() == "InvalidParameterValueException" ||
+					awserr.Code() == "ResourceInUseException" {
+
 					return resource.RetryableError(awserr)
 				}
 			}
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -35,7 +35,7 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 			},
 			"starting_position": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 			"batch_size": {
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -48,9 +48,12 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					// When AWS repurposed EventSourceMapping for use with SQS they kept
 					// the default for BatchSize at 100 for Kinesis and DynamoDB, but made
-					// the default 10 for SWS.  As such, we had to make batch_size optional.
+					// the default 10 for SQS.  As such, we had to make batch_size optional.
 					// Because of this, we need to ensure that if someone doesn't have
 					// batch_size specified that it is not treated as a diff for those
+					if new != "" && new != "0" {
+						return false
+					}
 
 					eventSourceARN, err := arn.Parse(d.Get("event_source_arn").(string))
 					if err != nil {
@@ -58,11 +61,11 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 					}
 					switch eventSourceARN.Service {
 					case dynamodb.ServiceName, kinesis.ServiceName:
-						if old == "100" && (new == "" || new == "0") {
+						if old == "100" {
 							return true
 						}
 					case sqs.ServiceName:
-						if old == "1" && (new == "" || new == "0") {
+						if old == "10" {
 							return true
 						}
 					default:
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -178,10 +178,8 @@ func resourceAwsLambdaEventSourceMappingDelete(d *schema.ResourceData, meta inte
 	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
 		_, err := conn.DeleteEventSourceMapping(params)
 		if err != nil {
-			if awserr, ok := err.(awserr.Error); ok {
-				if awserr.Code() == "ResourceInUseException" {
-					return resource.RetryableError(awserr)
-				}
+			if isAWSErr(err, lambda.ErrCodeResourceInUseException, "") {
+				return resource.RetryableError(err)
 			}
 			return resource.NonRetryableError(err)
 		}
@@ -212,12 +210,10 @@ func resourceAwsLambdaEventSourceMappingUpdate(d *schema.ResourceData, meta inte
 	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
 		_, err := conn.UpdateEventSourceMapping(params)
 		if err != nil {
-			if awserr, ok := err.(awserr.Error); ok {
-				if awserr.Code() == "InvalidParameterValueException" ||
-					awserr.Code() == "ResourceInUseException" {
+			if isAWSErr(err, lambda.ErrCodeInvalidParameterValueException, "") ||
+				isAWSErr(err, lambda.ErrCodeResourceInUseException, "") {
 
-					return resource.RetryableError(awserr)
-				}
+				return resource.RetryableError(err)
 			}
 			return resource.NonRetryableError(err)
 		}
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -41,7 +41,6 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 			"batch_size": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  100,
 			},
 			"enabled": {
 				Type:     schema.TypeBool,
@@ -89,11 +88,14 @@ func resourceAwsLambdaEventSourceMappingCreate(d *schema.ResourceData, meta inte
 	params := &lambda.CreateEventSourceMappingInput{
 		EventSourceArn: aws.String(eventSourceArn),
 		FunctionName:   aws.String(functionName),
-		BatchSize:      aws.Int64(int64(d.Get("batch_size").(int))),
 		Enabled:        aws.Bool(d.Get("enabled").(bool)),
 	}
 
-	if startingPosition := d.Get("starting_position"); startingPosition != "" {
+	if batchSize, ok := d.GetOk("batch_size"); ok {
+		params.BatchSize = aws.Int64(int64(batchSize.(int)))
+	}
+
+	if startingPosition, ok := d.GetOk("starting_position"); ok {
 		params.StartingPosition = aws.String(startingPosition.(string))
 	}
 
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -6,8 +6,12 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/kinesis"
 	"github.com/aws/aws-sdk-go/service/lambda"
+	"github.com/aws/aws-sdk-go/service/sqs"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by CyrusNajmabadi** [comment]
Running acceptance testing now.

**Comment by CyrusNajmabadi** [comment]
@bflad, i've seen you help out with a lot of PRS.  Could you let me know if i've done this right, and if there's anything else i need to do?  Thanks!

**Comment by bflad** [comment]
Can you please add an acceptance test that shows the resource working when you skip adding this argument and the output of `make testacc TEST=./aws TESTARGS='-run=TestAccAWSLambdaEventSourceMapping'`?

**Comment by CyrusNajmabadi** [comment]
@bflad Done.  results are:

```
cyrusn@matebunty ~/go/src/github.com/terraform-providers/terraform-provider-aws[patch-1 ↑ +0 ~2 -0]> make testacc TEST=./aws TESTARGS='-run=TestAccAWSLambdaEventSourceMapping'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLambdaEventSourceMapping -timeout 120m
=== RUN   TestAccAWSLambdaEventSourceMapping_basic
--- PASS: TestAccAWSLambdaEventSourceMapping_basic (105.09s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqsBasic
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsBasic (92.92s)
=== RUN   TestAccAWSLambdaEventSourceMapping_importBasic
--- PASS: TestAccAWSLambdaEventSourceMapping_importBasic (92.26s)
=== RUN   TestAccAWSLambdaEventSourceMapping_disappears
--- PASS: TestAccAWSLambdaEventSourceMapping_disappears (89.96s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqsDisappears
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsDisappears (124.02s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	504.291s
```

Note: it was good to write these tests.  They revealed the need to add retry logic here as making an event mapping for sqs events can actually take quite a while.  This was fine locally, but revealed problems in testing when the tests wanted to create/update/delete in rapid succession.

**Comment by flosell** [comment]
Just tried this PR on a [toy project](https://github.com/flosell/terraform-sqs-lambda-trigger-example), can confirm that it works fine! 

Thanks @CyrusNajmabadi, this will be a great help on my current project! 
@bflad hope we can get this merged soon

**Comment by ryandeivert** [comment]
hey @CyrusNajmabadi - I could be wrong here, buit I'm fairly certain the [`batch_size` default value](https://github.com/terraform-providers/terraform-provider-aws/pull/5024/files#diff-943f35dca31585fb17f97dbca05290dbL44) for this resource should also be removed now.

The default of 100 worked well for Kinesis and DynamoDB streams. However, the default for SQS (assigned by aws) is actually 1, while the max is 10. It's not incredibly obvious from the docs, but you can see this documented within the [SDK here](https://github.com/aws/aws-sdk-go/pull/2016/files#diff-f4455514cfe1fa73ff46aa8b2e48d530R119).

This means that if someone omits the `batch_size` argument when configuring an SQS event source mapping, a validation error will likely occur with aws.

**Comment by CyrusNajmabadi** [comment]
@ryandeivert You are correct.  However, i'm not sure the right thing to do here.  If we remove te batch_size default, won't that break *existing* customers of terraform who use this for Kinesis/Dynamo, but do not provide a size?  

> This means that if someone omits the batch_size argument when configuring an SQS event source mapping, a validation error will likely occur with aws.

That's true.  But then it's fixable by the user.

@ryandeivert @bflad I don't know enough about terraform to know what the right thing is to do here.  Can either of you shed light on the preferred course of action? Thanks!

**Comment by CyrusNajmabadi** [comment]
> Just tried this PR on a toy project, can confirm that it works fine!

Great!  I also have validated this myself.  But i'm glad to get secondary validation to ensure that there's nothing special about my setup!  Thanks much @flosell !

**Comment by CyrusNajmabadi** [comment]
@ryandeivert Looking at the AWS docs some more, it looks like what you're saying should be safe.  Their docs mention:

```
+	// with all the retrieved records. The default for Amazon Kinesis and Amazon
+	// DynamoDB is 100 records. For SQS, the default is 1.
 	BatchSize *int64 `min:"1" type:"integer"`
```

So it sounds like we can simply remove the default for BatchSize.  Anyone who is using the default today will be fine, since it will continue to be 100 if left out from the call to the actual AWS SDK.

**Comment by CyrusNajmabadi** [comment]
I've made the change to the default, and i've rerun tests:

```
cyrusn@matebunty ~/go/src/github.com/terraform-providers/terraform-provider-aws[patch-1 ≡]> make testacc TEST=./aws TESTARGS='-run=TestAccAWSLambdaEventSourceMapping'
TF_ACC=1 go test ./aws -v -run=TestAccAWSLambdaEventSourceMapping -timeout 120m
=== RUN   TestAccAWSLambdaEventSourceMapping_basic
--- PASS: TestAccAWSLambdaEventSourceMapping_basic (113.63s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqsBasic
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsBasic (100.05s)
=== RUN   TestAccAWSLambdaEventSourceMapping_importBasic
--- PASS: TestAccAWSLambdaEventSourceMapping_importBasic (96.89s)
=== RUN   TestAccAWSLambdaEventSourceMapping_disappears
--- PASS: TestAccAWSLambdaEventSourceMapping_disappears (92.24s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqsDisappears
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsDisappears (128.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	531.044s
```

File: aws/resource_aws_lambda_event_source_mapping_test.go [hunk 1]
```diff
@@ -288,17 +379,18 @@ resource "aws_lambda_function" "lambda_function_test_update" {
 }
 
 resource "aws_lambda_event_source_mapping" "lambda_event_source_mapping_test" {
-		batch_size = 100
-		event_source_arn = "${aws_kinesis_stream.kinesis_stream_test.arn}"
-		enabled = true
-		depends_on = ["aws_iam_policy_attachment.policy_attachment_for_role"]
-		function_name = "${aws_lambda_function.lambda_function_test_create.arn}"
-		starting_position = "TRIM_HORIZON"
+    batch_size = 100
```
bflad: Given that `batch_size` had its default removed, it should also be removed from the existing test configurations and separately tested in its own acceptance test and test configuration.

In the cases where you remove it from the test configurations, I have a hunch you will start seeing test failures with the resource wanting to change it since the API likely always returns a value, but its not specified in the configuration. e.g. `batch_size: "100" => ""`.

If that is the case, we need to add a `DiffSuppressFunc` to the `batch_size` attribute.

aws/resource_aws_lambda_event_source_mapping.go
```diff
@@ -35,13 +35,12 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
...
 			"starting_position": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 			"batch_size": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  100,
```
bflad: As mentioned, this change will likely require adding a `DiffSuppressFunc` to handle which API value is being returned depending on the event source type.

aws/resource_aws_lambda_event_source_mapping.go
```diff
@@ -170,7 +175,19 @@ func resourceAwsLambdaEventSourceMappingDelete(d *schema.ResourceData, meta inte
 		UUID: aws.String(d.Id()),
 	}
 
-	_, err := conn.DeleteEventSourceMapping(params)
+	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
+		_, err := conn.DeleteEventSourceMapping(params)
+		if err != nil {
+			if awserr, ok := err.(awserr.Error); ok {
```
bflad: Minor nitpick: we have a helper function available for this type of logic and also prefer to use the SDK available constants, e.g. `isAWSErr(err, lambda.ErrCodeResourceInUseException, "")`

aws/resource_aws_lambda_event_source_mapping.go
```diff
@@ -196,7 +213,9 @@ func resourceAwsLambdaEventSourceMappingUpdate(d *schema.ResourceData, meta inte
 		_, err := conn.UpdateEventSourceMapping(params)
 		if err != nil {
 			if awserr, ok := err.(awserr.Error); ok {
-				if awserr.Code() == "InvalidParameterValueException" {
+				if awserr.Code() == "InvalidParameterValueException" ||
```
bflad: Minor nitpick: we have a helper function available for this type of logic and also prefer to use the SDK available constants, e.g.

```go
if isAWSErr(err, lambda.ErrCodeInvalidParameterValueException, "") || isAWSErr(err, lambda.ErrCodeResourceInUseException, "") {
```

aws/resource_aws_lambda_event_source_mapping_test.go
```diff
@@ -234,7 +325,7 @@ EOF
 resource "aws_iam_policy" "policy_for_role" {
     name = "%s"
     path = "/"
-    description = "IAM policy for for Lamda event mapping testing"
+    description = "IAM policy for for Lambda event mapping testing"
     policy = <<EOF
```
Puneeth-n: `for for`

aws/resource_aws_lambda_event_source_mapping_test.go
```diff
@@ -322,7 +414,7 @@ EOF
 resource "aws_iam_policy" "policy_for_role" {
     name = "%s"
     path = "/"
-    description = "IAM policy for for Lamda event mapping testing"
+    description = "IAM policy for for Lambda event mapping testing"
```
Puneeth-n: `for for`

aws/resource_aws_lambda_event_source_mapping.go
CyrusNajmabadi: Fixed.

aws/resource_aws_lambda_event_source_mapping.go
CyrusNajmabadi: Fixed.

aws/resource_aws_lambda_event_source_mapping_test.go
CyrusNajmabadi: @bflad Can you expand on this a bit.  I went to take a look.  but i think there's a small snag.  If i'm understanding right, we would want to supress things if ```old``` was ```100```, and ```new``` was ```<empty>```, **and** this was a Kinesis or DynamoDB case.  

However, the problem i'm having is figuring out how to tell if this is Kinesis or DynamoDB.  event_source_arn doesn't really tell me this, as it's just an arbitrary ARN.  Or is the intent that i would somehow get the true value here and try to 'sniff' it to tell what was going on?

Sorry, i've looked around a bit to see if i can figure out the right way to address this.  But it was not clear to me.  So any help would def be appreciated.  Thanks!

aws/resource_aws_lambda_event_source_mapping_test.go
CyrusNajmabadi: Done.   Thanks!

aws/resource_aws_lambda_event_source_mapping_test.go
bflad: Luckily the ARN has the service name in it 😄  Does something like this do the trick?

```go
					eventSourceARN, err := arn.Parse(d.Get("event_source_arn").(string))
					if err != nil {
						return false
					}
					switch eventSourceARN.Service {
					case dynamodb.ServiceID:
						if old == "100" && new == "" {
							return true
						}
					case kinesis.ServiceID:
						if old == "100" && new == "" {
							return true
						}
					case sqs.ServiceID:
						if old == "1" && new == "" {
							return true
						}
					}
					return false
				},
```

bflad: Slightly cleaner version:

```go
					if new != "" {
						return false
					}
					eventSourceARN, err := arn.Parse(d.Get("event_source_arn").(string))
					if err != nil {
						return false
					}
					if eventSourceARN.Service == dynamodb.ServiceID && old == "100" {
						return true
					}
					if eventSourceARN.Service == kinesis.ServiceID && old == "100" {
						return true
					}
					if eventSourceARN.Service == sqs.ServiceID && old == "1" {
						return true
					}
					return false
				},
```

ryandeivert: could have also done similar with multiple case statements:
```go
					eventSourceARN, err := arn.Parse(d.Get("event_source_arn").(string))
					if err != nil {
						return false
					}
					switch eventSourceARN.Service {
					case dynamodb.ServiceID, kinesis.ServiceID:
						if old == "100" && new == "" {
							return true
						}
					case sqs.ServiceID:
						if old == "1" && new == "" {
							return true
						}
					}
					return false
				},
```

CyrusNajmabadi: Thanks guys.  That looks like it will do the trick.  What i was missing was things like arn.Parse and kinesis.ServiceID.  Much better than me manually groveling the string!

CyrusNajmabadi: Ok.  Was able to make this work.  Note: i needed to use .ServiceName not .ServiceID.  Also, i needed to check for hte "0" value, not the "" value as it looks like that's what you get by default for integral values...

CyrusNajmabadi:
@bflad This is ready for review again.  Note, i have added the additional test to make sure batch_size works when removed.  All tests pass:

```
cyrusn@matebunty ~/go/src/github.com/terraform-providers/terraform-provider-aws[patch-1 ≡]> make && make testacc TEST=./aws TESTARGS='-run=TestAccAWSLambdaEventSourceMapping'
go install
TF_ACC=1 go test ./aws -v -run=TestAccAWSLambdaEventSourceMapping -timeout 120m
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_basic
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_basic (107.03s)
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_removeBatchSize
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_removeBatchSize (104.82s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqs_basic
--- PASS: TestAccAWSLambdaEventSourceMapping_sqs_basic (92.34s)
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_import
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_import (91.84s)
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_disappears
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_disappears (89.31s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqsDisappears
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsDisappears (112.39s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	597.853s
```

Hopefully this can go in soon :)    let me know if tehre are more change you want me to make.

aws/resource_aws_lambda_event_source_mapping.go
CyrusNajmabadi: Done.  Thanks!

CyrusNajmabadi:
@bflad Anything else you need from me at this point?  Or is this ready to go in.  Thanks! :)

CyrusNajmabadi:
@bflad Thanks!   Just to make sure:

>  There's two items of feedback I'll handle on merge so we can get this released tonight/tomorrow.

So you don't want me to make any changes here right?  You'll just tweak things on your end?

aws/resource_aws_lambda_event_source_mapping.go
```diff
@@ -35,13 +39,40 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
...
+					case dynamodb.ServiceName, kinesis.ServiceName:
+						if old == "100" {
+							return true
+						}
+					case sqs.ServiceName:
+						if old == "10" {
+							return true
+						}
+					default:
+						panic(eventSourceARN.Service)
```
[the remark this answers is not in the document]
CyrusNajmabadi: Makes sense!

aws/resource_aws_lambda_event_source_mapping.go
CyrusNajmabadi: Makes sense.  Just to check, but you're going to make that change, right?  Or would you prefer i do it?

ryandeivert:
huge thanks to @CyrusNajmabadi and @bflad for expediting this!

CyrusNajmabadi:
Thanks all!

bflad:

PR #5032, opened 2018-06-29
Issue #4266 add Type attribute in aws_spot_fleet_request

### LINKED ISSUE #4266: Add support for the type parameter to aws_spot_fleet_request
Currently Terraform does not support the `Type` parameter when using the `aws_spot_fleet_request` resource.

The parameter as supported by the API (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotFleetRequestConfigData.html) allows you to specify whether you want to just create spot instances according to the current request, or maintain the specified number of targets when the spot instances are terminated.

Because the default is `request`, currently you cannot create spot fleet requests that maintain the target number of spot instances using Terraform.

### Terraform Version
```
Terraform v0.11.7
+ provider.aws v1.15.0
+ provider.template v1.0.0
```

### Affected Resource(s)
- aws_spot_fleet_request

### Expected Behavior
Terraform should support the type parameter as supported by the AWS API for spot fleet requests.

### Actual Behavior
Terraform replies with an error saying that the `type` parameter is not supported.

saravanan30erd:
Fixes #4266

* Change 1
added Type attribute in aws_spot_fleet_request

```
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSSpotFleetRequest_fleetType'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSSpotFleetRequest_fleetType -timeout 120m
=== RUN   TestAccAWSSpotFleetRequest_fleetType
--- PASS: TestAccAWSSpotFleetRequest_fleetType (379.63s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	379.677s
...
```

### DIFF
--- a/aws/resource_aws_spot_fleet_request.go
+++ b/aws/resource_aws_spot_fleet_request.go
@@ -319,6 +319,12 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validateRFC3339TimeString,
 			},
+			"fleet_type": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "maintain",
+				ForceNew: true,
+			},
 			"spot_request_state": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -600,6 +606,7 @@ func resourceAwsSpotFleetRequestCreate(d *schema.ResourceData, meta interface{})
 		TerminateInstancesWithExpiration: aws.Bool(d.Get("terminate_instances_with_expiration").(bool)),
 		ReplaceUnhealthyInstances:        aws.Bool(d.Get("replace_unhealthy_instances").(bool)),
 		InstanceInterruptionBehavior:     aws.String(d.Get("instance_interruption_behaviour").(string)),
+		Type: aws.String(d.Get("fleet_type").(string)),
 	}
 
 	if v, ok := d.GetOk("excess_capacity_termination_policy"); ok {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_spot_fleet_request.go
```diff
@@ -319,6 +319,12 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validateRFC3339TimeString,
 			},
+			"fleet_type": {
```
bflad: Is there a particular reason to call this attribute `fleet_type` and not `type` to match the API?

We can also validate the input via:

```go
  ec2.FleetTypeMaintain,
  ec2.FleetTypeRequest,
}, false),
```

aws/resource_aws_spot_fleet_request.go
```diff
@@ -319,6 +319,12 @@ func resourceAwsSpotFleetRequest() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "maintain",
```
bflad: We should prefer to use the SDK constant: `ec2.FleetTypeMaintain`

saravanan30erd:
@bflad done the changes.

aws/resource_aws_spot_fleet_request.go
bflad: We generally do this because its reduces any differences between the [AWS CLI](https://docs.aws.amazon.com/cli/latest/reference/ec2/request-spot-fleet.html), [CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-spotfleet-spotfleetrequestconfigdata.html), and other SDK implementations. Its probably fine though.

bflad:

PR #5036, opened 2018-06-29
codedeploy documentation update 

marcosdiez:
This is just documentation update.

In case one uses *Load Balancers* with *codedeploy*, I added to the example the necessary IAM permissions for *codedeploy* to be able to add and remove machines from the *load balancer* (or target group, in case it's an *alb*)

This will save time from people googling around the meaning of

`com.amazonaws.services.simpleworkflow.flow.ActivityTaskFailedException: Role does not have correct permissions. role XXXXXXXX
 for activityId="5" of activityType={Name: ExecuteCentralizedCommandOnInstanceActivity.runCentralizedCommand,Version: 1.00}
]`

### DIFF
--- a/website/docs/r/codedeploy_deployment_group.html.markdown
+++ b/website/docs/r/codedeploy_deployment_group.html.markdown
@@ -33,46 +33,9 @@ resource "aws_iam_role" "example" {
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


marcosdiez:
Thanks @bflad ! I did not know that trick. I just updated the PR.

Issue #4137, opened 2018-04-10
Differences between aws_waf_web_acl and aws_wafregional_web_acl


nbrys:
Just for consistency reasons, make aws_waf_web_acl and aws_wafregional_web_acl consistent with each other.

Using rules vs rule makes thinks more complex:
https://www.terraform.io/docs/providers/aws/r/waf_web_acl.html#rules
vs
https://www.terraform.io/docs/providers/aws/r/wafregional_web_acl.html#rule

For both resources, it is possible to attach multiple rules. I suggest changing the parameter from wafregional_web_acl to rules

chroju:
There are other differences between AWS WAF and AWS WAFRegional. In addition to `rule(s)` in aws_waf(regional)_web_acl,

* `byte_match_tuple(s)` in aws_waf(regional)_byte_match_set
* `ip_set_descriptor(s)` in aws_waf(regional)_ipset
* `predicate(s)` in aws_waf(regional)_rate_based_rule
* `predicate(s)` in aws_waf(regional)_rule
* `sql_injection_match_tuple(s)` in aws_waf(regional)_sql_injection_match_set
* `xss_match_tuple(s)` in aws_waf(regional)_xss_match_set

I have made PR about aws_waf(regional)_byte_match_set at #5043 .

chroju:
I have created PRs to fix all differences above.

bflad:
Hi folks 👋 Thanks for your interest here. Its certainly an inconvenience the initial implementation of these WAF vs WAF Regional resources contained these naming inconsistencies for configurations. 😖We will be sure to look for these more closely in future implementations. 

We previously shepherded through https://github.com/terraform-providers/terraform-provider-aws/pull/5043 which started the process of making the `aws_waf_byte_match_set` and `aws_wafregional_byte_match_set` resources consistent. This type of change highlighted the [amount of effort involved for not only the code maintainers](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename) but also the operator burden for converting older configurations and [potential bugs introduced by implementing changes like these](https://github.com/terraform-providers/terraform-provider-aws/pull/5043#issuecomment-419427065). Instead of introducing more burden for the code maintainers and operators with minor benefit in this situation, we will prefer to keep things as they currently exist rather than promise that this may potentially be fixed in the future without a deadline as our focus is on other more impactful features and bug fixes across the rest of the provider.

[...]

PR #5043, opened 2018-07-01
Fix the difference between aws_waf_byte_match_set and aws_wafregional_byte_match_set

chroju:

Some aws waf and aws waf-regional API have same attributes each other. However, Terraform `aws_waf_***` resources and `aws_wafregional_***` resources have same attributes with the different name, one is singular and the other is plural.  Although this is mentioned a little in #4137 , this discrepancy can also be seen in several other resources about AWS WAF(Regional).

This PR will fix such inconsistencies. I think it would be better to unify in singular names, because these attributes type is not `list` but `map` .

WIP

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSWafByteMatchSet*'
TF_ACC=1 go test ./aws -v -run=TestAccAWSWafByteMatchSet* -timeout 120m
=== RUN   TestAccAWSWafByteMatchSet_basic
--- PASS: TestAccAWSWafByteMatchSet_basic (42.07s)
=== RUN   TestAccAWSWafByteMatchSet_changeNameForceNew
--- PASS: TestAccAWSWafByteMatchSet_changeNameForceNew (77.94s)
=== RUN   TestAccAWSWafByteMatchSet_changeTuple
--- PASS: TestAccAWSWafByteMatchSet_changeTuple (72.65s)
=== RUN   TestAccAWSWafByteMatchSet_noTuples
--- PASS: TestAccAWSWafByteMatchSet_noTuples (37.13s)
=== RUN   TestAccAWSWafByteMatchSet_disappears
--- PASS: TestAccAWSWafByteMatchSet_disappears (35.07s)
PASS
ok      github.com/chroju/terraform-provider-aws/aws    264.901s
```

### DIFF
--- a/aws/resource_aws_wafregional_byte_match_set.go
+++ b/aws/resource_aws_wafregional_byte_match_set.go
@@ -25,6 +25,45 @@ func resourceAwsWafRegionalByteMatchSet() *schema.Resource {
 				ForceNew: true,
 			},
 			"byte_match_tuple": &schema.Schema{
+				Type:          schema.TypeSet,
+				Optional:      true,
+				ConflictsWith: []string{"byte_match_tuples"},
+				Deprecated:    "use `byte_match_tuples` instead",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"field_to_match": {
+							Type:     schema.TypeSet,
+							Required: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"data": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"type": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+								},
+							},
+						},
+						"positional_constraint": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"target_string": &schema.Schema{
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"text_transformation": &schema.Schema{
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+			},
+			"byte_match_tuples": &schema.Schema{
 				Type:     schema.TypeSet,
 				Optional: true,
 				Elem: &schema.Resource{
@@ -110,7 +149,11 @@ func resourceAwsWafRegionalByteMatchSetRead(d *schema.ResourceData, meta interfa
 		return err
 	}
 
-	d.Set("byte_match_tuple", flattenWafByteMatchTuplesWR(resp.ByteMatchSet.ByteMatchTuples))
+	if _, ok := d.GetOk("byte_match_tuple"); ok {
+		d.Set("byte_match_tuple", flattenWafByteMatchTuplesWR(resp.ByteMatchSet.ByteMatchTuples))
+	} else {
+		d.Set("byte_match_tuples", flattenWafByteMatchTuplesWR(resp.ByteMatchSet.ByteMatchTuples))
+	}
 	d.Set("name", resp.ByteMatchSet.Name)
 
 	return nil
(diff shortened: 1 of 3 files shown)


aws/resource_aws_waf_byte_match_set.go
```diff
@@ -24,7 +24,7 @@ func resourceAwsWafByteMatchSet() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-			"byte_match_tuples": &schema.Schema{
```
bflad: We need to duplicate the entire existing `byte_match_tuples` attribute into the new `byte_match_tuple` attribute. Then in the existing `byte_match_tupes` attribute, add `ConflictsWith` and `Deprecation` fields, e.g.

```go
ConflictsWith: []string{"byte_match_tuple"},
Deprecation: "use `byte_match_tuple` instead",
```

aws/resource_aws_waf_byte_match_set.go
```diff
@@ -107,7 +107,7 @@ func resourceAwsWafByteMatchSetRead(d *schema.ResourceData, meta interface{}) er
 	}
 
 	d.Set("name", resp.ByteMatchSet.Name)
-	d.Set("byte_match_tuples", flattenWafByteMatchTuples(resp.ByteMatchSet.ByteMatchTuples))
```
bflad: When renaming attributes, we need to ensure both the old and new are still read into the Terraform state.

aws/resource_aws_waf_byte_match_set.go
```diff
@@ -117,8 +117,8 @@ func resourceAwsWafByteMatchSetUpdate(d *schema.ResourceData, meta interface{})
 
 	log.Printf("[INFO] Updating ByteMatchSet: %s", d.Get("name").(string))
 
-	if d.HasChange("byte_match_tuples") {
```
bflad: The existing logic for `byte_match_tuples` must be preserved in addition to the new `byte_match_tuple`

aws/resource_aws_waf_byte_match_set.go
```diff
@@ -24,7 +24,7 @@ func resourceAwsWafByteMatchSet() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-			"byte_match_tuples": &schema.Schema{
+			"byte_match_tuple": &schema.Schema{
```
bflad: The documentation in `website/docs/r/waf_byte_match_set.html.markdown` should add the new attribute and add a **deprecated** notice to the existing attribute.

aws/resource_aws_waf_rule_test.go
```diff
@@ -402,15 +402,15 @@ func testAccAWSWafRuleConfig_changePredicates(name string) string {
 	return fmt.Sprintf(`
 resource "aws_waf_ipset" "ipset" {
   name = "%s"
-  ip_set_descriptors {
```
bflad: Was this change intentional?

bflad:
For what its worth, there can be multiple ByteMatchSet tuples as the attribute is `schema.TypeSet` and the `updateByteMatchSetResource` function works with multiple tuples. It seems this attribute should be plural.

chroju:
@bflad Thanks for your review and comments.

I have noticed that this is so complex and breaking change than I thought. There are much more inconsistent attributes about AWS WAF and AWS WAFRegional, and it is debatable whether each of them should be plural or singular.

So, at first I would fix this PR not about these all inconsistent attributes but only `byte_match_tuples` . Also I will make the list of inconsistent attributes at #4137 once. Is this OK ?

aws/resource_aws_waf_rule_test.go
chroju: Sorry, this is my mistake. Too mach changes were in the commit.

bflad:

nbrys:
After changing 

`byte_match_tuple {
    text_transformation   = "NONE"
    target_string         = "/favicon.png"
    positional_constraint = "EXACTLY"

    field_to_match {
      type = "URI"
    }
  }`

into:
`byte_match_tuples {
    text_transformation   = "NONE"
    target_string         = "/favicon.png"
    positional_constraint = "EXACTLY"

    field_to_match {
      type = "URI"
    }
  }`

I get the following change in terraform:
  byte_match_tuple.511953057.field_to_match.#:                 "1" => "0"
  byte_match_tuple.511953057.field_to_match.3756326843.data:   "" => ""
  byte_match_tuple.511953057.field_to_match.3756326843.type:   "URI" => ""
  byte_match_tuple.511953057.positional_constraint:            "EXACTLY" => ""
  byte_match_tuple.511953057.target_string:                    "/favicon.ico" => ""
  byte_match_tuple.511953057.text_transformation:              "NONE" => ""
  byte_match_tuples.#:                                         "0" => "1"
  byte_match_tuples.511953057.field_to_match.#:                "0" => "1"
  byte_match_tuples.511953057.field_to_match.3756326843.data:  "" => ""
  byte_match_tuples.511953057.field_to_match.3756326843.type:  "" => "URI"
  byte_match_tuples.511953057.positional_constraint:           "" => "EXACTLY"
  byte_match_tuples.511953057.target_string:                   "" => "/favicon.ico"
  byte_match_tuples.511953057.text_transformation:             "" => "NONE"

But I end up with an empty filter in WAF

PR #5056, opened 2018-07-03
Add member_clusters to aws_elasticache_replication_group

haines:

* Add `member_clusters` attribute to Elasticache replication group data source and resource. This is useful when setting up Cloudwatch alarms on the clusters.

_(I wasn't able to get the acceptance tests running on my AWS account)_
```
$ make testacc TESTARGS='-run=TestAcc.*ElasticacheReplicationGroup_basic'
...
--- FAIL: TestAccDataSourceAwsElasticacheReplicationGroup_basic (17.55s)
		
		* aws_elasticache_replication_group.bar: 1 error(s) occurred:
		
		* aws_elasticache_replication_group.bar: Error creating Elasticache Replication Group: CacheParameterGroupNotFound: CacheParameterGroup not found: default.redis3.2
			status code: 404, request id: 9436287e-7ea9-11e8-8f9f-5357cdd1eabd
...
--- FAIL: TestAccAWSElasticacheReplicationGroup_basic (18.06s)
		
		* aws_elasticache_security_group.bar: 1 error(s) occurred:
		
		* aws_elasticache_security_group.bar: Error creating CacheSecurityGroup: InvalidParameterValue: Use of cache security groups is not permitted in this API version for your account.
			status code: 400, request id: 9d97e424-7ea9-11e8-811f-674747b34208
...
```

### DIFF
--- a/aws/data_source_aws_elasticache_replication_group.go
+++ b/aws/data_source_aws_elasticache_replication_group.go
@@ -45,6 +45,12 @@ func dataSourceAwsElasticacheReplicationGroup() *schema.Resource {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
+			"member_clusters": {
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Set:      schema.HashString,
+			},
 			"node_type": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -106,6 +112,7 @@ func dataSourceAwsElasticacheReplicationGroupRead(d *schema.ResourceData, meta i
 		d.Set("primary_endpoint_address", rg.NodeGroups[0].PrimaryEndpoint.Address)
 	}
 	d.Set("number_cache_clusters", len(rg.MemberClusters))
+	d.Set("member_clusters", flattenStringList(rg.MemberClusters))
 	d.Set("node_type", rg.CacheNodeType)
 	d.Set("snapshot_window", rg.SnapshotWindow)
 	d.Set("snapshot_retention_limit", rg.SnapshotRetentionLimit)
--- a/aws/resource_aws_elasticache_replication_group_test.go
+++ b/aws/resource_aws_elasticache_replication_group_test.go
@@ -89,6 +89,8 @@ func TestAccAWSElasticacheReplicationGroup_basic(t *testing.T) {
 						"aws_elasticache_replication_group.bar", "cluster_mode.#", "0"),
 					resource.TestCheckResourceAttr(
 						"aws_elasticache_replication_group.bar", "number_cache_clusters", "2"),
+					resource.TestCheckResourceAttrSet(
+						"aws_elasticache_replication_group.bar", "member_clusters"),
 					resource.TestCheckResourceAttr(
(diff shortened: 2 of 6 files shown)


aws/data_source_aws_elasticache_replication_group.go
```diff
@@ -106,6 +112,7 @@ func dataSourceAwsElasticacheReplicationGroupRead(d *schema.ResourceData, meta i
 		d.Set("primary_endpoint_address", rg.NodeGroups[0].PrimaryEndpoint.Address)
 	}
 	d.Set("number_cache_clusters", len(rg.MemberClusters))
+	d.Set("member_clusters", flattenStringList(rg.MemberClusters))
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("member_clusters", flattenStringList(rg.MemberClusters)); err != nil {
  return fmt.Errorf("error setting member_clusters: %s", err)
}
```

The same applies to the resource.

aws/resource_aws_elasticache_replication_group_test.go
```diff
@@ -89,6 +89,8 @@ func TestAccAWSElasticacheReplicationGroup_basic(t *testing.T) {
 						"aws_elasticache_replication_group.bar", "cluster_mode.#", "0"),
 					resource.TestCheckResourceAttr(
 						"aws_elasticache_replication_group.bar", "number_cache_clusters", "2"),
+					resource.TestCheckResourceAttrSet(
+						"aws_elasticache_replication_group.bar", "member_clusters"),
```
bflad: Understandably since you were not able to run the acceptance testing, you weren't able to discover this test failure:

```
=== RUN   TestAccAWSElasticacheReplicationGroup_basic
--- FAIL: TestAccAWSElasticacheReplicationGroup_basic (1009.96s)
    testing.go:518: Step 0 error: Check failed: Check 4/5 error: aws_elasticache_replication_group.bar: Attribute 'member_clusters' expected to be set
```

The reason this happens is because `schema.TypeSet` attributes do not actually save information at just the attribute name. Specifically for `schema.TypeSet` attributes, the following is set:
* `NAME.#` - the number of elements present
* `NAME.HASH` - the value of an element in the set, where HASH is computed by `schema.HashString` as defined in the schema

Here is an example of the Terraform state from one of my test runs (NOTE: since the identifiers are randomized in the test, the hash values will change):

```
  member_clusters.# = 2
  member_clusters.166139442 = tf-gx1hjkvphs-002
  member_clusters.2431534984 = tf-gx1hjkvphs-001
```

So long story short, to fix the test here, we can switch it to use the following:

```go
resource.TestCheckResourceAttr("aws_elasticache_replication_group.bar", "member_clusters.#", "2"),
```

A similar change needed to be applied in the data source testing.

haines:
@bflad Thanks for the getting this merged so quickly, and for the detailed explanation of the fixes you made. Much appreciated!

bflad:

Issue #5099, opened 2018-07-05
Import of an aws_route_table with propagation erroneously imports propagated routes


richievos:
When importing a route table that has had routes propagated to it, terraform improperly reads the propagated routes as actual configured routes.

My guess is that the `EnableVgwRoutePropagation` logic in https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_route_table.go either isn't being used during the import, or is being ran at a time before the propagation info is available.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```Terraform v0.11.7```

### Affected Resource(s)

* aws_route
* aws_route_table

### Terraform Configuration Files

Here's an example terraform config

```hcl
variable "vpc_cidr" {}
variable "virtual_gateway_amazon_side_asn" {}

resource "aws_vpc" "lz_vpc" {
  cidr_block = "${var.vpc_cidr}"

  enable_dns_support = true
  enable_dns_hostnames = true
}

resource "aws_internet_gateway" "vpc_gateway" {
  vpc_id = "${aws_vpc.lz_vpc.id}"
}

resource "aws_route_table" "public" {
  vpc_id = "${aws_vpc.lz_vpc.id}"

  route {
    cidr_block = "0.0.0.0/0"
    gateway_id = "${aws_internet_gateway.vpc_gateway.id}"
  }
}

resource "aws_vpn_gateway" "test_prop_vpn_gw" {
  vpc_id = "${aws_vpc.lz_vpc.id}"

  amazon_side_asn = "${var.virtual_gateway_amazon_side_asn}"
}

resource "aws_vpn_gateway_route_propagation" "test_prop" {
  vpn_gateway_id = "${aws_vpn_gateway.vpn_gw.id}"
  route_table_id = "${aws_route_table.test_table.id}"
}

resource "aws_route_table" "test_table" {
  vpc_id = "${aws_vpc.lz_vpc.id}"

  route {
    cidr_block = "0.0.0.0/0"
    gateway_id = "${aws_internet_gateway.vpc_gateway.id}"
  }
}
```

### Debug Output

[...]

PR #5100, opened 2018-07-05
Ignore propagated routes during table import

richievos:
Similar to how they're ignored when performing a
`resourceAwsRouteTableRead`.

Relates to #5099

This syncs the continue logic from resourceAwsRouteTableImportState and resourceAwsRouteTableRead. Really these should be using a helper function, but given there's already 3 other places with this code copy-pasted, I wasn't sure what the preferred way of doing this would be. Writing a test for this seems like it'd be pretty difficult, since a full test would require a gateway connected that has routes propagating into it. I assume that's why the existing resourceAwsRouteTableRead doesn't have a comparable test.

Output from test runs below. Apologies I don't have AWS creds in place that I can run this test against. The change proposed here is rather trivial though, and this at least shows it compiles / passes existing *normal* tests.

```
$ make test
ok  	github.com/terraform-providers/terraform-provider-aws/aws	2.009s
```

```
--- FAIL: TestAccAWSAvailabilityZones_basic (1.44s)
	provider_test.go:75: No valid credential sources found for AWS Provider.
			Please see https://terraform.io/docs/providers/aws/index.html for more information on
			providing credentials for the AWS Provider
--- FAIL: TestAccAWSAvailabilityZones_stateFilter (0.96s)
	provider_test.go:75: No valid credential sources found for AWS Provider.
			Please see https://terraform.io/docs/providers/aws/index.html for more information on
			providing credentials for the AWS Provider
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	2.437s
make: *** [testacc] Error 1
```

### DIFF
--- a/aws/import_aws_route_table.go
+++ b/aws/import_aws_route_table.go
@@ -40,6 +40,10 @@ func resourceAwsRouteTableImportState(
 				continue
 			}
 
+			if route.Origin != nil && *route.Origin == "EnableVgwRoutePropagation" {
+				continue
+			}
+
 			if route.DestinationPrefixListId != nil {
 				// Skipping because VPC endpoint routes are handled separately
 				// See aws_vpc_endpoint


bflad:

PR #5106, opened 2018-07-06
Remove tag on create logic for gov cloud since no longer applicable

domwong:

* It used to be that restricted cloud accounts (e.g. gov cloud) cannot create volumes with tags, they can only update it after creation. See [here](https://aws.amazon.com/blogs/aws/new-tag-ec2-instances-ebs-volumes-on-creation/). However, this has recently changed as [announced here](https://aws.amazon.com/about-aws/whats-new/2018/07/tag-on-create-for-amazon-ec2-and-amazon-ebs-is-now-available-in-the-aws-govcloud-region/). This PR removes the logic behind the restriction for gov cloud. *NOTE it is unclear whether China cloud still has this restriction so leaving that logic in there*

> This capability is available at no additional cost in all AWS commercial regions, and the AWS GovCloud (US) Region.

```
$ make testacc TESTARGS='-run=TestAccAWSInstance.*Tag'
--- PASS: TestAccAWSInstance_volumeTags (183.15s)
--- FAIL: TestAccAWSInstance_volumeTagsComputed (18.74s)
		
		* aws_instance.foo: Error launching source instance: InstanceLimitExceeded: You have requested more instances (1) than your current instance limit of 0 allows for the specified instance type. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
			status code: 400, request id: 67bd5f4e-4e29-4ea7-9ca7-27d85acacbfb
--- PASS: TestAccAWSInstance_forceNewAndTagsDrift (225.67s)
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	427.607s
make: *** [testacc] Error 1

```

*Note: test failure due to account limits*

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -514,35 +514,32 @@ func resourceAwsInstanceCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("Only 1 of `ipv6_address_count` or `ipv6_addresses` can be specified")
 	}
 
-	restricted := meta.(*AWSClient).IsGovCloud() || meta.(*AWSClient).IsChinaCloud()
-	if !restricted {
-		tagsSpec := make([]*ec2.TagSpecification, 0)
+	tagsSpec := make([]*ec2.TagSpecification, 0)
 
-		if v, ok := d.GetOk("tags"); ok {
-			tags := tagsFromMap(v.(map[string]interface{}))
+	if v, ok := d.GetOk("tags"); ok {
+		tags := tagsFromMap(v.(map[string]interface{}))
 
-			spec := &ec2.TagSpecification{
-				ResourceType: aws.String("instance"),
-				Tags:         tags,
-			}
-
-			tagsSpec = append(tagsSpec, spec)
+		spec := &ec2.TagSpecification{
+			ResourceType: aws.String("instance"),
+			Tags:         tags,
 		}
 
-		if v, ok := d.GetOk("volume_tags"); ok {
-			tags := tagsFromMap(v.(map[string]interface{}))
+		tagsSpec = append(tagsSpec, spec)
+	}
 
-			spec := &ec2.TagSpecification{
-				ResourceType: aws.String("volume"),
-				Tags:         tags,
-			}
+	if v, ok := d.GetOk("volume_tags"); ok {
+		tags := tagsFromMap(v.(map[string]interface{}))
 
-			tagsSpec = append(tagsSpec, spec)
+		spec := &ec2.TagSpecification{
+			ResourceType: aws.String("volume"),
+			Tags:         tags,
 		}
 
-		if len(tagsSpec) > 0 {
-			runOpts.TagSpecifications = tagsSpec
-		}
+		tagsSpec = append(tagsSpec, spec)
+	}
+
+	if len(tagsSpec) > 0 {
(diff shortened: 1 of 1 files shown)


tomelliff:
Is this still true considering [this recent announcement](https://aws.amazon.com/about-aws/whats-new/2018/07/tag-on-create-for-amazon-ec2-and-amazon-ebs-is-now-available-in-the-aws-govcloud-region/)?

domwong:
@tomelliff perhaps not. I proposed this patch more because as someone working on non gov-cloud AWS the current behaviour results in unneeded CreateTags calls in "non restricted" regions. 

If it's still not available for China cloud then the logic is right in that it needs to be restricted for some cases, but the calculation of `restricted` needs to change to just a check for whether it's China cloud or not.

domwong:
I've made the changes, renamed, and updated the description on this PR. Thanks @tomelliff for the link.

bflad:
I'm not entirely sure the mentioned line below implies AWS China support:

> This capability is available at no additional cost in all AWS commercial regions, and the AWS GovCloud (US) Region.

AWS China is separate from AWS commercial and AWS GovCloud (US).

Unfortunately, we (HashiCorp) do not have an AWS China account to verify this either way so unless someone with that type of account, this pull request cannot be accepted and merged. If the previous behavior with AWS China is present (at least during instance creation -- the update change is fine as that is on the operator to understand the API error), we can get this in.

---

Aside:

The usage of `restricted`/`isGovCloud()`/`isChinaCloud()` represented the first iteration of trying to support the 3 AWS partitions when they offered differing capabilities due to running different service API versions. While this approach of checking the AWS partition provides an immediate fix for these situations, it does not automatically handle if/when those new capabilities are enabled.

In newer iterations of this handling, we attempt to simply ignore read errors based on specific error messages, like `UnsupportedOperation` and passthrough any create/update errors. See https://github.com/terraform-providers/terraform-provider-aws/pull/3794/files for an example of this. I understand this situation is a little different, but I feel like the additional context here is relevant.

I thought I had created a technical debt issue to swap out usage of `restricted`/`isGovCloud()`/`isChinaCloud()` for the future-proofing behavior, but looks like I haven't done that yet. Unfortunately, it would not be a priority for the maintainers to implement, but we would certainly look at pull requests to update that code handling.

bflad:
For what its worth, we could also remove the `isChinaCloud()` handling during tagging on creation in AWS China if we know what error message is getting returned then re-submitting the request without that problematic parameter. This would future-proof the resource for AWS China in that situation.

e.g.

```
runResp, err = conn.RunInstances(runOpts)
// Fake error message handling here -- we need the real one
if isAWSErr(err, "UnsupportedOperation", "TagSpecifications") {
  runOpts.TagSpecifications = nil
}
runResp, err = conn.RunInstances(runOpts)
// Rest of retry logic
```

domwong:
Ah OK, thanks for the input and context. Didn't know the distinction between China et al. If it's unlikely we can get someone  with a China cloud account to verify are we better off amending the PR to still check for `isChinaCloud` only?

bflad:
> are we better off amending the PR to still check for isChinaCloud only?

I think so, we can always remove it later if its actually supported.

domwong:
I've reinstated the china cloud restriction @bflad

bflad:

PR #5108, opened 2018-07-06
diffTags should not identify existing tags as needing creation

domwong:

* `diffTags` incorrectly returns tags that are the same in `oldTags` and `newTags` as needing to be created again. This causes unnecessary calls to `CreateTags`. Fix this to return only tags that need to be deleted and created

```
$ make testacc TESTARGS='-run=TestAccAWSInstance.*Tag'
--- PASS: TestAccAWSInstance_volumeTags (186.28s)
--- FAIL: TestAccAWSInstance_volumeTagsComputed (18.64s)
		
		* aws_instance.foo: Error launching source instance: InstanceLimitExceeded: You have requested more instances (1) than your current instance limit of 0 allows for the specified instance type. Please visit http://aws.amazon.com/contact-us/ec2-request to request an adjustment to this limit.
			status code: 400, request id: ad836a70-ded9-4335-94f4-18871c6c96c0
--- PASS: TestAccAWSInstance_forceNewAndTagsDrift (290.62s)
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	495.584s
make: *** [testacc] Error 1

```
*Note failures due to account limits*

This PR split out from #5072.

### DIFF
--- a/aws/tags.go
+++ b/aws/tags.go
@@ -197,6 +197,9 @@ func diffTags(oldTags, newTags []*ec2.Tag) ([]*ec2.Tag, []*ec2.Tag) {
 		old, ok := create[*t.Key]
 		if !ok || old != *t.Value {
 			remove = append(remove, t)
+		} else if ok {
+			// already present so remove from new
+			delete(create, *t.Key)
 		}
 	}
 
--- a/aws/tags_test.go
+++ b/aws/tags_test.go
@@ -47,6 +47,24 @@ func TestDiffTags(t *testing.T) {
 				"foo": "bar",
 			},
 		},
+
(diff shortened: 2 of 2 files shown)


domwong:
@bflad - continuing on from our discussion on #5072 ...

>Does this logic potentially exist to handle tag value updates for an existing tag key? Some service APIs do not support updating an existing tag so therefore require delete and create.

I misread this in my initial reply. This update scenario is actually covered in the test case I've added. e.g. updating `foo:bar` to `foo:baz` results in corresponding "create" and "remove" calls.

bflad:

PR #5120, opened 2018-07-09
Update ebs acceptance tests to run in multiple regions.

jmcarp:
So that we can run ebs tests on govcloud (and presumably aws-cn). cc @bflad

### DIFF
--- a/aws/resource_aws_ebs_volume_test.go
+++ b/aws/resource_aws_ebs_volume_test.go
@@ -135,7 +135,7 @@ func TestAccAWSEBSVolume_kmsKey(t *testing.T) {
 	var v ec2.Volume
 	ri := acctest.RandInt()
 	config := fmt.Sprintf(testAccAwsEbsVolumeConfigWithKmsKey, ri)
-	keyRegex := regexp.MustCompile("^arn:aws:([a-zA-Z0-9\\-])+:([a-z]{2}-[a-z]+-\\d{1})?:(\\d{12})?:(.*)$")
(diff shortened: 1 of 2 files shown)


aws/resource_aws_ebs_volume_test.go
```diff
@@ -249,8 +251,6 @@ data "aws_ami" "debian_jessie_latest" {
     name   = "root-device-type"
     values = ["ebs"]
   }
-
-  owners = ["379101102735"] # Debian
```
bflad: Removing the account ID filter means that AWS Marketplace AMIs can be selected, which can return the following error:

```
=== RUN   TestAccAWSEBSVolume_updateAttachedEbsVolume
--- FAIL: TestAccAWSEBSVolume_updateAttachedEbsVolume (10.58s)
		
		* aws_instance.test: 1 error(s) occurred:
		
		* aws_instance.test: Error launching source instance: OptInRequired: In order to use this AWS Marketplace product you need to accept terms and subscribe. To do so please visit https://aws.amazon.com/marketplace/pp?sku=3f8t6t8fp5m9xx18yzwriozxi
			status code: 401, request id: fb86043e-96bc-4ac0-9ebc-4fbdaf679b1d
```

We'll need to either filter AWS Marketplace AMIs out or switch to an AMI search to a different AMI completely that will work across partitions (with no Marketplace AMIs).

jmcarp:
Thanks @bflad; I updated the patch to use an amazon-owned image and verified that tests are passing in the `aws` and `aws-us-gov` partitions.

PR #5128, opened 2018-07-09
[WIP] new resource for ses domain identity Policy

gmyers-amfam:

* Adding a new resource for aws_ses_domain_identity_policy

```
$ make testacc TESTARGS='-run=TestAccAWSSESDomainIdentityPolicy'

...
```

This is my first Terraform pull request and I'm a novice when it comes to Go. I tried running the `make testacc` command above but got a lot of "cannot find package" errors. Is anyone able to assist in either getting me up to speed, or getting this PR polished and tested?

A note on this resource: SES Domain Identities can have any number of policies attached to them, similar to how an S3 bucket has a policy. SES Domain Identity Policies allow for things like cross-account sending access (by setting the 'Principals' field in the policy accordingly) which is clearly missing from Terraform at the moment, hence this PR.

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -545,6 +545,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_secretsmanager_secret_version":                resourceAwsSecretsManagerSecretVersion(),
 			"aws_ses_active_receipt_rule_set":                  resourceAwsSesActiveReceiptRuleSet(),
 			"aws_ses_domain_identity":                          resourceAwsSesDomainIdentity(),
+			"aws_ses_domain_identity_policy":                   resourceAwsSesDomainIdentityPolicy(),
 			"aws_ses_domain_identity_verification":             resourceAwsSesDomainIdentityVerification(),
 			"aws_ses_domain_dkim":                              resourceAwsSesDomainDkim(),
 			"aws_ses_domain_mail_from":                         resourceAwsSesDomainMailFrom(),
--- a/aws/resource_aws_ses_domain_identity_policy.go
+++ b/aws/resource_aws_ses_domain_identity_policy.go
@@ -1,0 +1,140 @@
+package aws
+
+import (
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ses"
+	"fmt"
+	"github.com/hashicorp/terraform/helper/resource"
+)
+
+func resourceAwsSesDomainIdentityPolicy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesDomainIdentityPolicyCreate,
+		Read:   resourceAwsSesDomainIdentityPolicyRead,
+		Update: resourceAwsSesDomainIdentityPolicyUpdate,
+		Delete: resourceAwsSesDomainIdentityPolicyDelete,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"policy": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateFunc:     validateJsonString,
+				DiffSuppressFunc: suppressEquivalentAwsPolicyDiffs,
+			},
+		},
+	}
+}
+
+func resourceAwsSesDomainIdentityPolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	arn := d.Get("arn").(string)
+	policyName := d.Get("name").(string)
+	policy := d.Get("policy").(string)
+
+	req := ses.PutIdentityPolicyInput{
+		Identity: aws.String(arn),
+		PolicyName: aws.String(policyName),
+		Policy: aws.String(policy),
+	}
+
+	_, err := conn.PutIdentityPolicy(&req)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(resource.PrefixedUniqueId(fmt.Sprintf("%s-", policyName)))
+	return resourceAwsSesDomainIdentityPolicyRead(d, meta)
+}
+
+func resourceAwsSesDomainIdentityPolicyUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	arn := d.Get("arn").(string)
+	policyName := d.Get("name").(string)
+	policy := d.Get("policy").(string)
+
+	req := ses.PutIdentityPolicyInput{
+		Identity:   aws.String(arn),
+		PolicyName: aws.String(policyName),
+		Policy:     aws.String(policy),
+	}
+
+	_, err := conn.PutIdentityPolicy(&req)
+	if err != nil {
+		return err
+	}
+
+	return resourceAwsSesDomainIdentityPolicyRead(d, meta)
+}
+
+func resourceAwsSesDomainIdentityPolicyRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sesConn
+
+	arn := d.Get("arn").(string)
+	policyName := d.Get("name").(string)
+	policyNames := make([]*string, 1)
+	policyNames[0] = aws.String(policyName)
+
+	policiesOutput, err := conn.GetIdentityPolicies(&ses.GetIdentityPoliciesInput{
+		Identity: aws.String(arn),
+		PolicyNames: policyNames,
+	})
+	if err != nil {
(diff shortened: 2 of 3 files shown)


SijmenHuizenga:
@gmyers-amfam Looking at the code i can make a wild guess what the blocking issues are: It seems that there is no documentation added in this pull-request. Since this change introduces a new resource there should probably be added a documentation page on the website with this pull-request. Also, i can imagine this pull-request becomes easier for the maintainers to merge if there were some tests included. So lets first add some docs and tests, next remove the [wip] and lastly ping the maintainers.

leoblanc:
Hi, any updates? We are terraforming SES and we have SES domain identity policies. Thanks

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+
+func resourceAwsSesDomainIdentityPolicy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSesDomainIdentityPolicyCreate,
+		Read:   resourceAwsSesDomainIdentityPolicyRead,
+		Update: resourceAwsSesDomainIdentityPolicyUpdate,
+		Delete: resourceAwsSesDomainIdentityPolicyDelete,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
```
bflad: This attribute should be changed to `identity` or `identity_arn` to match the expected value. If the API is always setting this to the ARN, we should likely name it `identity_arn` and validate it via:

```go
ValidateFunc: validateArn,
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
```
bflad: We can validate this argument via:

```go
ValidateFunc: validation.All(
	validation.StringLenBetween(1, 64),
	validation.StringMatch(regexp.MustCompile(`^[a-zA-Z0-9\-\_]$`), "must contain only alphanumeric characters, dashes, and underscores"),
),
```

Reference: https://docs.aws.amazon.com/ses/latest/APIReference/API_PutIdentityPolicy.html

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+
+	req := ses.PutIdentityPolicyInput{
+		Identity:   aws.String(arn),
+		PolicyName: aws.String(policyName),
+		Policy:     aws.String(policy),
+	}
+
+	_, err := conn.PutIdentityPolicy(&req)
+	if err != nil {
+		return err
```
bflad: We should return error context for operators and code maintainers, e.g.

```suggestion
		return fmt.Errorf("error creating SES Identity (%s) Policy: %s", identityARN, err)
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
```
bflad: We should return error context here for operators and code maintainers:

```suggestion
		return fmt.Errorf("error updating SES Identity (%s) Policy (%s): %s", identityARN, policyName, err)
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+	conn := meta.(*AWSClient).sesConn
+
+	arn := d.Get("arn").(string)
+	policyName := d.Get("name").(string)
+	policyNames := make([]*string, 1)
+	policyNames[0] = aws.String(policyName)
+
+	policiesOutput, err := conn.GetIdentityPolicies(&ses.GetIdentityPoliciesInput{
+		Identity:    aws.String(arn),
+		PolicyNames: policyNames,
```
bflad: Nit: The above logic can be simplified with:

```suggestion
		PolicyNames: aws.StringSlice([]string{policyName}),
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFound" {
```
bflad: This logic can be simplified with a helper function we have available:

```suggestion
		if isAWSErr(err, "NotFound", "") {
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if policiesOutput.Policies == nil {
+		log.Printf("[WARN] SES Domain Identity Policy (%s) not found (nil)", policyName)
+		d.SetId("")
+		return nil
+	}
+	policies := policiesOutput.Policies
+
+	policy, ok := policies[*aws.String(policyName)]
```
bflad: The casting to a pointer and dereferencing looks extraneous here:

```suggestion
	policy, ok := policies[policyName]
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+		PolicyNames: policyNames,
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "NotFound" {
+			log.Printf("[WARN] SES Domain Identity Policy (%s) not found, error code (404)", policyName)
+			d.SetId("")
+			return nil
+		}
+
+		return err
```
bflad: We should return error context here for operators and code maintainers:

```suggestion
		return fmt.Errorf("error reading SES Identity (%s) Policy (%s): %s", identityARN, policyName, err)
```

aws/resource_aws_ses_domain_identity_policy.go
```diff
@@ -0,0 +1,140 @@
...
+	policyName := d.Get("name").(string)
+
+	req := ses.DeleteIdentityPolicyInput{
+		Identity:   aws.String(arn),
+		PolicyName: aws.String(policyName),
+	}
+
+	log.Printf("[DEBUG] Deleting SES Domain Identity Policy: %s", req)
+	_, err := conn.DeleteIdentityPolicy(&req)
+	return err
```
bflad: We should return error context here for operators and code maintainers:

```suggestion
	if err != nil {
		return fmt.Errorf("error deleting SES Identity (%s) Policy (%s): %s", identityARN, policyName, err)
	}
	
	return nil
```

aws/resource_aws_ses_domain_identity_policy_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestAccAWSSESDomainIdentityPolicy_basic(t *testing.T) {
+	domain := fmt.Sprintf(
+		"%s.terraformtesting.com.",
+		acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsSESDomainIdentityDestroy,
```
bflad: A new `CheckDestroy` function should be created that verifies from the API that all SES Identity Policy configured in the test configurations are deleted. See also: https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html#checkdestroy

aws/resource_aws_ses_domain_identity_policy_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSSESDomainIdentityConfig_withPolicy(domain),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsSESDomainIdentityExists("aws_ses_domain_identity.test"),
```
bflad: A new `testAccCheckAwsSESIdentityPolicyExists()` check function should be created that verifies from the API that the SES Identity Policy referenced by the given resource name was in fact properly created.

aws/provider.go
```diff
@@ -545,6 +545,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_secretsmanager_secret_version":                resourceAwsSecretsManagerSecretVersion(),
 			"aws_ses_active_receipt_rule_set":                  resourceAwsSesActiveReceiptRuleSet(),
 			"aws_ses_domain_identity":                          resourceAwsSesDomainIdentity(),
+			"aws_ses_domain_identity_policy":                   resourceAwsSesDomainIdentityPolicy(),
```
bflad: Two larger items:
* Given that this resource can work with both SES Identity Policies for both domains and emails, we should consider naming it `aws_ses_identity_policy` as there would be no benefit to having separate resources between the two. The file, resource, and resource testing function names will need to be updated to reflect this change.
* This new resource is missing resource documentation in a new `website/docs/r/ses_identity_policy.html.markdown` file and a website sidebar link in `website/aws.erb`

asaghri:
Any update regarding this ? That would be very useful !

bflad:
Since we have not heard anything from the original author regarding this, the implementation has been finished in a followup commit. The new resource is named `aws_ses_identity_policy` (accepting both domain and email identities) and will be released in version 2.16.0 of the Terraform AWS Provider, likely later this week.

bflad:

PR #5167, opened 2018-07-12
aws_lb_listener: wait for listener creation

mildred:
AWS resources can take some time to appear in listings. When we created
a listener before, it could happen that the creation succeeded but the
listing of the resource right after creation would return a resource not
found.

This can me normal on AWS where changes can take some time to propagate.
The correct behaviour in this case is to retry reading the resource
until we find it (because we know that it has been created
successfully).

We don't change the behaviour on resource reads that are not following a
creation where a resource not found is still going to remove the
resource from the tfstate.

This should fix https://github.com/terraform-providers/terraform-provider-aws/issues/2456

### DIFF
--- a/aws/resource_aws_lb_listener.go
+++ b/aws/resource_aws_lb_listener.go
@@ -25,6 +25,10 @@ func resourceAwsLbListener() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		Timeouts: &schema.ResourceTimeout{
+			Read: schema.DefaultTimeout(10 * time.Minute),
+		},
+
 		Schema: map[string]*schema.Schema{
 			"arn": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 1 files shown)


aws/resource_aws_lb_listener.go
```diff
@@ -151,11 +155,26 @@ func resourceAwsLbListenerCreate(d *schema.ResourceData, meta interface{}) error
 func resourceAwsLbListenerRead(d *schema.ResourceData, meta interface{}) error {
 	elbconn := meta.(*AWSClient).elbv2conn
 
-	resp, err := elbconn.DescribeListeners(&elbv2.DescribeListenersInput{
+	var resp *elbv2.DescribeListenersOutput
+	var request = &elbv2.DescribeListenersInput{
 		ListenerArns: []*string{aws.String(d.Id())},
+	}
+
+	err := resource.Retry(d.Timeout(schema.TimeoutRead), func() *resource.RetryError {
```
bflad: When dealing with AWS service eventual consistency, we prefer to hardcode 1 minute timeouts. Generally its not desired to allow that timeout to be user configurable as its a knob that is unlikely to handle the situation properly by waiting longer.

aws/resource_aws_lb_listener.go
```diff
@@ -25,6 +25,10 @@ func resourceAwsLbListener() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		Timeouts: &schema.ResourceTimeout{
```
bflad: If we do decide to add user configurable timeouts for the resource (I would recommend against in this situation), it should be documented in the resource documentation under a `## Timeouts` header.

runtheops:
@mildred , many thanks for the fix, this particular bug is super nasty in my scenario, since I have tons of listeners and rules created and destroyed. Looking forward to get this merged

ngortheone:
+1 for getting this merged. I experience this on a daily basis

bflad:

PR #5177, opened 2018-07-13
provider: Allow AssumeRoleARN and credential validation call to shortcut account ID and partition lookup

bflad:
Potentially supersedes: #5060 

* If provider configuration is set to assume role, use provided ARN to bypass account information calls
* If provider credential validation is enabled, use results to bypass further account information calls
* Refactor GetAccountID back to GetAccountInformation to ensure non-SDK-default partitions can be handled (and make it much more readable 😄 )
* Refactor GetAccountInformation testing

Output from acceptance testing: (provided as a smoke test)

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsAvailabilityZone'
TF_ACC=1 go test ./aws -v -run=TestAccDataSourceAwsAvailabilityZone -timeout 120m
=== RUN   TestAccDataSourceAwsAvailabilityZone
--- PASS: TestAccDataSourceAwsAvailabilityZone (9.98s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	10.021s
```

### DIFF
--- a/aws/auth_helpers_test.go
+++ b/aws/auth_helpers_test.go
@@ -368,7 +368,7 @@ func TestGetAccountIDAndPartitionFromSTSGetCallerIdentity(t *testing.T) {
 	}
 }
 
-func TestAWSparseAccountIDAndPartitionFromARN(t *testing.T) {
+func TestAWSParseAccountIDAndPartitionFromARN(t *testing.T) {
 	var testCases = []struct {
 		InputARN          string
 		ErrCount          int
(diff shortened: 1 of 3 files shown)


ewbankkit:
How about introducing a `struct AccountInformation` to avoid the possibility of inadvertently transposing account ID and partition in the various return values?

bflad:
Updated to use `t.Run()` for new unit tests

aws/auth_helpers_test.go
```diff
@@ -10,304 +10,411 @@ import (
...
+			ExpectedPartition: "aws-us-gov",
+		},
 	}
 
-	invalidArn := "blablah"
-	id, err = parseAccountIDFromArn(invalidArn)
-	if err == nil {
-		t.Fatalf("Expected error when parsing invalid ARN (%q)", invalidArn)
+	for _, testCase := range testCases {
+		accountID, partition, err := parseAccountIDAndPartitionFromARN(testCase.InputARN)
```
paultyng: missing a `t.Run`

bflad: Updated 👍

bflad:
This has been released in [version 1.31.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1310-august-09-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5192, opened 2018-07-14
data-source/aws_acm_certificate:  does not fail if acm certificate is not found

ntmggr:
Reference: https://github.com/hashicorp/terraform/issues/16380

Data source allows search for acm cert and return null if it does not exist.
It will continue and it will not error

Output from acceptance testing: AWS Commercial
```
$make testacc TEST=./aws TESTARGS="-run=TestAccAWSAcmCertificateDataSource_noMatchReturnsError"
TF_ACC=1 go test ./aws -v -run=TestAccAWSAcmCertificateDataSource_noMatchReturnsError -timeout 120m
=== RUN   TestAccAWSAcmCertificateDataSource_noMatchReturnsError
--- PASS: TestAccAWSAcmCertificateDataSource_noMatchReturnsError (30.24s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	30.392s
```

### DIFF
--- a/aws/data_source_aws_acm_certificate.go
+++ b/aws/data_source_aws_acm_certificate.go
@@ -66,7 +66,8 @@ func dataSourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) e
 		return true
 	})
 	if err != nil {
-		return fmt.Errorf("Error listing certificates: %q", err)
+		log.Printf("[DEBUG] No certificate for domain %s found in this region", target)
+		return nil
 	}
 
 	if len(arns) == 0 {


bflad:
Hi @ntmggr 👋 Thanks for submitting this and sorry for not providing any review since it was submitted.

The current design of Terraform AWS provider "singular" data sources is to always error if the intended AWS resource is not found. There can be unintended side effects with other Terraform functionality if we start allowing create-if-non-existent use cases, among others. This behavior is also consistent across all "singular" data sources at the moment in the provider.

After Terraform core 0.12 is released, I believe there are plans to look further into the desired use cases, ensure all workflows continue to work, and come up with a plan to either address it via changes like these to "singular" data sources (once more thoroughly verified), allow "plural" data sources to handle this case, or introduce new concepts into Terraform itself to better support those use cases. See also: https://github.com/hashicorp/terraform/issues/16380#issuecomment-418476841

Rather than leaving this pull request open, which won't get accepted and merged at the moment, I'm going to close this in preference of those continuing discussions upstream and to keep the repository with only actionable pull requests open. We can reopen this pull request if those discussions lead to this type of change across the data sources. 👍 Thanks again.

PR #5198, opened 2018-07-15
resource/aws_route: Print more useful error message when missing valid target type

MattMcNam:

* Update aws_route error message to list which target types are expected when none are given

The current error message does not actually print anything after the colon since `setTarget` is an empty string, and while it's not too hard to figure out the actual problem, this can save a trip to the documentation.

### DIFF
--- a/aws/resource_aws_route.go
+++ b/aws/resource_aws_route.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


bflad:
This has been released in [version 1.28.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1280-july-18-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

Issue #943, opened 2017-06-22
Support for AWS Storage Gateway


wonderphil:
Hi guys,

Just wondering if there is a eta for support of AWS Storage Gateways and related services?

Thanks

alex1x:
Would be great, currently needing to do acrobatics with other scripts to get this all automated and integrating with existing Terraform code.

bflad:
Can you provide which functionality is highest priority for your use cases? It appears that bootstrapping the acceptance test for activating a gateway might take some effort with properly configuring an EC2 instance, etc.

mickeybr:
I could really use it for setting up the file shares more than anything.

lthomas987:
I too would like to configure nfs shares from storage gateway.

scraly:
Any news about it?

bflad:
If someone could share a working `aws_instance` resource configuration that uses the [Storage Gateway AMI](https://docs.aws.amazon.com/storagegateway/latest/userguide/ec2-gateway-file.html), I think that would be helpful for moving forward since that would likely be the basis for all Storage Gateway acceptance testing.

Please note due to other priorities, the Terraform team at HashiCorp won't be able to work on this in the near future, but we'd be happy to review a pull request if you or someone else has the time and motivation to implement it. Alternatively, if others would also like to see this implemented I'd encourage adding a 👍  upvote reaction to the original issue comment (not to this comment), which we use as one of the inputs to prioritize work for the Terraform team.

mickeybr:
I will check with our team for the configuration. I think we ended up delegating to a bash script in terraform for now that passes in the parameters. At this point we connect to the On-Premise VMware appliance instead of using an EC2 instance and it is pretty much a manual process to activate it.

But when an application team requests file shares, that's where we use the automated scripts.
Here's the CLI commands we would like to be supported by Terraform:

- [create-nfs-file-share](https://docs.aws.amazon.com/cli/latest/reference/storagegateway/create-nfs-file-share.html)
- [delete-file-share](https://docs.aws.amazon.com/cli/latest/reference/storagegateway/delete-file-share.html)
- [describe-cache](https://docs.aws.amazon.com/cli/latest/reference/storagegateway/describe-cache.html)
- [describe-gateway-information](https://docs.aws.amazon.com/cli/latest/reference/storagegateway/describe-gateway-information.html)
[...]

PR #5200, opened 2018-07-16
New Data Source: aws_storagegateway_gateway_activation_key

bflad:
Reference: #943 

* New data source for fetching Storage Gateway activation key from on-premises or EC2 instance gateway

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSStorageGatewayGatewayActivationKeyDataSource_FileGateway'
TF_ACC=1 go test ./aws -v -run=TestAccAWSStorageGatewayGatewayActivationKeyDataSource_FileGateway -timeout 120m
=== RUN   TestAccAWSStorageGatewayGatewayActivationKeyDataSource_FileGateway
--- PASS: TestAccAWSStorageGatewayGatewayActivationKeyDataSource_FileGateway (242.01s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	242.048s
```


bflad:
This logic might get folded into the an `aws_storagegateway_gateway` resource instead -- I am going to test the workflow in that type of resource setup.

aws/resource_aws_storagegateway_gateway_activation_key.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func getStorageGatewayGatewayActivationKeyFromIP(activationRegion, ipAddress string) (string, *resource.RetryError) {
+	client := &http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+		Timeout: time.Second * 10,
+	}
+	requestURL := fmt.Sprintf("http://%s/?activationRegion=%s", ipAddress, activationRegion)
```
tombuildsstuff: feels like this wants a "this is a special configuration and we don't support using resources in this manor; please use the AWS SDK where possible" disclaimer?

website/docs/r/storagegateway_gateway_activation_key.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+  Provides a resource to retrieve an activation key from a gateway using the region of the provider
+---
+
+# aws_storagegateway_gateway_activation_key
+
+Provides a resource to retrieve an activation key from a gateway using the region of the provider.
+
+~> **NOTE:** Terraform must be able to make an HTTP (port 80) GET request to the specified IP address from where it is running.
+
+~> **NOTE:** Deleting this Terraform resource does not perform any operations.
```
tombuildsstuff: then should this be a data source?

bflad: Re-implemented as data source after I figured out that data sources also support configurable timeouts 😄

aws/provider.go
```diff
@@ -581,6 +581,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_ssm_patch_group":                              resourceAwsSsmPatchGroup(),
 			"aws_ssm_parameter":                                resourceAwsSsmParameter(),
 			"aws_ssm_resource_data_sync":                       resourceAwsSsmResourceDataSync(),
+			"aws_storagegateway_gateway_activation_key":        resourceAwsStorageGatewayGatewayActivationKey(),
```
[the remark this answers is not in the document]
bflad: Sorry meant to reply to this directly as well -- we're trying to keep the resource/data source names as follows: `aws_SERVICE_THING`. In this case, the service is `storagegateway` to match the SDK and the thing is a gateway activation key (in case they introduce some other "activation key"s) which is used specifically for activating gateways via [ActivateGateway](https://docs.aws.amazon.com/sdk-for-go/api/service/storagegateway/#ActivateGatewayInput).

PR #5201, opened 2018-07-16
New Resource: aws_macie_s3_bucket_association

### LINKED ISSUE #4992: New resource: Amazon Macie S3 Association
Amazon Macie administrative APIs [recently](https://aws.amazon.com/about-aws/whats-new/2018/06/amazon-macie-releases-administrative-APIs-to-enable-IT-management-workflow-automation/) released.
Add initial support with an Amazon Macie S3 Association resource.

ewbankkit:
Fixes #4992.
This is the initial Amazon Macie resource, associating S3 resources with Amazon Macie for monitoring and data classification.

Acceptance tests:

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSMacieS3BucketAssociation_'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSMacieS3BucketAssociation_ -timeout 120m
=== RUN   TestAccAWSMacieS3BucketAssociation_basic
--- PASS: TestAccAWSMacieS3BucketAssociation_basic (67.79s)
=== RUN   TestAccAWSMacieS3BucketAssociation_accountIdAndPrefix
--- PASS: TestAccAWSMacieS3BucketAssociation_accountIdAndPrefix (71.10s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	152.742s
```

### DIFF
--- a/aws/resource_aws_macie_s3_bucket_association.go
+++ b/aws/resource_aws_macie_s3_bucket_association.go
@@ -1,0 +1,235 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/macie"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsMacieS3BucketAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsMacieS3BucketAssociationCreate,
+		Read:   resourceAwsMacieS3BucketAssociationRead,
+		Update: resourceAwsMacieS3BucketAssociationUpdate,
+		Delete: resourceAwsMacieS3BucketAssociationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"bucket_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"prefix": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"member_account_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validateAwsAccountId,
+			},
+			"classification_type": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"one_time": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func resourceAwsMacieS3BucketAssociationCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).macieconn
+
+	req := &macie.AssociateS3ResourcesInput{
+		S3Resources: []*macie.S3ResourceClassification{
+			{
+				BucketName: aws.String(d.Get("bucket_name").(string)),
+			},
+		},
+	}
+	if v, ok := d.GetOk("member_account_id"); ok {
+		req.MemberAccountId = aws.String(v.(string))
+	}
+	if v, ok := d.GetOk("prefix"); ok {
+		req.S3Resources[0].Prefix = aws.String(v.(string))
+	}
+
+	ct := &macie.ClassificationType{
+		Continuous: aws.String(macie.S3ContinuousClassificationTypeFull),
+	}
+	ct.OneTime = aws.String(macieS3BucketAssociationOneTimeClassification(d))
+	req.S3Resources[0].ClassificationType = ct
+
+	log.Printf("[DEBUG] Creating Macie S3 bucket association: %#v", req)
+	resp, err := conn.AssociateS3Resources(req)
+	if err != nil {
+		return fmt.Errorf("Error creating Macie S3 bucket association: %s", err)
+	}
+	if len(resp.FailedS3Resources) > 0 {
+		return fmt.Errorf("Error creating Macie S3 bucket association: %s", resp.FailedS3Resources[0])
+	}
+
+	d.SetId(macieS3BucketAssociationId(d))
+	return resourceAwsMacieS3BucketAssociationRead(d, meta)
+}
+
+func resourceAwsMacieS3BucketAssociationRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).macieconn
+
+	req := &macie.ListS3ResourcesInput{}
+	if v, ok := d.GetOk("member_account_id"); ok {
+		req.MemberAccountId = aws.String(v.(string))
+	}
+
+	bucketName := d.Get("bucket_name").(string)
+	prefix := d.Get("prefix")
+	var res *macie.S3ResourceClassification
+	for {
+		resp, err := conn.ListS3Resources(req)
+		if err != nil {
+			return fmt.Errorf("Error listing Macie S3 bucket associations: %s", err)
+		}
+
+		for _, v := range resp.S3Resources {
+			if aws.StringValue(v.BucketName) == bucketName && aws.StringValue(v.Prefix) == prefix {
+				res = v
+				break
+			}
+		}
+		if res != nil {
+			break
+		}
+
+		if resp.NextToken == nil {
+			break
+		}
+		req.NextToken = resp.NextToken
+	}
+
+	if res == nil {
+		log.Printf("[WARN] Macie S3 bucket association (%s) not found, removing from state", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	m := map[string]interface{}{}
+	if aws.StringValue(res.ClassificationType.OneTime) == macie.S3OneTimeClassificationTypeFull {
+		m["one_time"] = true
+	} else {
+		m["one_time"] = false
+	}
+	if err := d.Set("classification_type", []map[string]interface{}{m}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


aws/resource_aws_macie_s3_bucket_association.go
```diff
@@ -0,0 +1,235 @@
...
+			strings.Contains(aws.StringValue(failed.ErrorMessage), "is not associated with Macie") {
+			return nil
+		}
+		return fmt.Errorf("Error deleting Macie S3 bucket association: %s", failed)
+	}
+
+	return nil
+}
+
+func macieS3BucketAssociationId(d *schema.ResourceData) string {
```
bflad: Nitpick: This indirection is unnecessary as its only used once for a one-liner. It seems like we should also include the member account ID so we can easily support passthrough import.

aws/resource_aws_macie_s3_bucket_association.go
```diff
@@ -0,0 +1,235 @@
...
+func macieS3BucketAssociationOneTimeClassification(d *schema.ResourceData) string {
+	oneTime := false
+	if v := d.Get("classification_type").([]interface{}); len(v) > 0 {
+		if m := v[0].(map[string]interface{}); m["one_time"].(bool) {
+			oneTime = true
+		}
+	}
+	if oneTime {
+		return macie.S3OneTimeClassificationTypeFull
+	} else {
```
bflad: Nitpick: The `else` nesting here is extraneous as the `if` has a return statement.

aws/resource_aws_macie_s3_bucket_association.go
```diff
@@ -0,0 +1,235 @@
...
+	req := &macie.ListS3ResourcesInput{}
+	if v, ok := d.GetOk("member_account_id"); ok {
+		req.MemberAccountId = aws.String(v.(string))
+	}
+
+	bucketName := d.Get("bucket_name").(string)
+	prefix := d.Get("prefix")
+	var res *macie.S3ResourceClassification
+	for {
+		resp, err := conn.ListS3Resources(req)
```
bflad: We should prefer to use the available SDK paginated function: https://docs.aws.amazon.com/sdk-for-go/api/service/macie/#Macie.ListS3ResourcesPages

aws/resource_aws_macie_s3_bucket_association.go
```diff
@@ -0,0 +1,235 @@
...
+		req.NextToken = resp.NextToken
+	}
+
+	if res == nil {
+		log.Printf("[WARN] Macie S3 bucket association (%s) not found, removing from state", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	m := map[string]interface{}{}
```
bflad: To prevent potential panics, we should perform a `nil` check for `res.ClassificationType` before dereferencing.

We also generally prefer to move this type of logic into a "flatten" function, e.g.

```go
func flattenMacieClassificationType(classificationType *macie.ClassificationType) []map[string]interface{} {
  if classificationType == nil {
  }
  m := map[string]interface{}{
    "one_time": false,
  }
  if aws.StringValue(res.ClassificationType.OneTime) == macie.S3OneTimeClassificationTypeFull {
    m["one_time"] = true
  }
  return []map[string]interface{}{m}
}
```

We should also provide a helpful error message when `d.Set()` fails, e.g.

```go
if err := d.Set("classification_type", flattenMacieClassificationType(res.ClassificationType)); err != nil {
  return fmt.Errorf("error setting classification_type: %s", err)
}
```

aws/resource_aws_macie_s3_bucket_association.go
```diff
@@ -0,0 +1,235 @@
...
+			},
+			"classification_type": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"one_time": {
+							Type:     schema.TypeBool,
```
bflad: We might want to consider following the SDK and using a `TypeString` with the constant values in case they introduce additional types. This will also simplify the logic to remove the conditional handling.

aws/resource_aws_macie_s3_bucket_association_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/macie"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSMacieS3BucketAssociation_basic(t *testing.T) {
```
bflad: Is there a way to programatically enable/disable Macie?

```
=== RUN   TestAccAWSMacieS3BucketAssociation_basic
--- FAIL: TestAccAWSMacieS3BucketAssociation_basic (6.07s)
		
		* aws_macie_s3_bucket_association.test: 1 error(s) occurred:
		
		* aws_macie_s3_bucket_association.test: Error creating Macie S3 bucket association: InvalidInputException: The request was rejected. Macie is not enabled for this AWS account *******.
```

aws/resource_aws_macie_s3_bucket_association_test.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		req := &macie.ListS3ResourcesInput{}
+		acctId := rs.Primary.Attributes["member_account_id"]
+		if acctId != "" {
+			req.MemberAccountId = aws.String(acctId)
+		}
+
+		for {
+			resp, err := conn.ListS3Resources(req)
```
bflad: We should prefer to also use the paginated method here 👍

aws/resource_aws_macie_s3_bucket_association_test.go
```diff
@@ -0,0 +1,174 @@
...
+			if resp.NextToken == nil {
+				break
+			}
+			req.NextToken = resp.NextToken
+		}
+	}
+	return nil
+}
+
+func testAccCheckAWSMacieS3BucketAssociationExists(name string) resource.TestCheckFunc {
```
bflad: This function should perform similar logic as the destroy check above to ensure the S3 resources are appropriately allocated in Macie.

ewbankkit:
Yes, should get time to address these today.

aws/resource_aws_macie_s3_bucket_association.go
ewbankkit: 👍

aws/resource_aws_macie_s3_bucket_association.go
ewbankkit: 👍

aws/resource_aws_macie_s3_bucket_association.go
ewbankkit: 👍

aws/resource_aws_macie_s3_bucket_association_test.go
ewbankkit: 👍

aws/resource_aws_macie_s3_bucket_association_test.go
ewbankkit: 👍

aws/resource_aws_macie_s3_bucket_association.go
ewbankkit: Agreed, trying to second guess what might come in the future is guaranteed compatibility pain.
I'll also add in the `continuous` attribute of `TypeString`.

aws/resource_aws_macie_s3_bucket_association_test.go
ewbankkit: I'll see if [`AssociateMemberAccount`](https://docs.aws.amazon.com/sdk-for-go/api/service/macie/#Macie.AssociateMemberAccount) with the current account ID does that.
I just manually enabled Macie via the console.

aws/resource_aws_macie_s3_bucket_association_test.go
ewbankkit: 👍

aws/resource_aws_macie_s3_bucket_association_test.go
ewbankkit: No, it looks like Macie cannot programatically be enabled:

```console
$ aws --region us-east-1 macie list-member-accounts

An error occurred (InvalidInputException) when calling the ListMemberAccounts operation: The request was rejected. Macie is not enabled for this AWS account 999999999999.
$ aws --region us-east-1 macie associate-member-account --member-account-id 999999999999

An error occurred (InvalidInputException) when calling the AssociateMemberAccount operation: The request was rejected. Macie is not enabled for this AWS account 999999999999.
```

I'll add a link to the manual setup instructions to the documentation.

ewbankkit:
Acceptance tests after code review changes:

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSMacieS3BucketAssociation_'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSMacieS3BucketAssociation_ -timeout 120m
=== RUN   TestAccAWSMacieS3BucketAssociation_basic
--- PASS: TestAccAWSMacieS3BucketAssociation_basic (65.16s)
=== RUN   TestAccAWSMacieS3BucketAssociation_accountIdAndPrefix
--- PASS: TestAccAWSMacieS3BucketAssociation_accountIdAndPrefix (72.33s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	138.118s
```

bflad:
This has been released in [version 1.28.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1280-july-18-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5216, opened 2018-07-17
Optimize resource_aws_ssm_parameter

julienduchesne:
* Changed resource_aws_ssm_parameter to use GetParameter instead of GetParameters
* Set MaxResults of DescribeParameters to 50 (maximum value)

```
TESTARGS='-run=TestAccAWSSSMParameter_'  make testacc
TF_ACC=1 go test ./... -v -run=TestAccAWSSSMParameter_ -timeout 120m
=== RUN   TestAccAWSSSMParameter_importBasic
--- PASS: TestAccAWSSSMParameter_importBasic (20.12s)
--- PASS: TestAccAWSSSMParameter_basic (16.35s)
=== RUN   TestAccAWSSSMParameter_disappears
--- PASS: TestAccAWSSSMParameter_disappears (10.70s)
--- PASS: TestAccAWSSSMParameter_update (28.46s)
=== RUN   TestAccAWSSSMParameter_updateDescription
--- PASS: TestAccAWSSSMParameter_updateDescription (26.93s)
--- PASS: TestAccAWSSSMParameter_changeNameForcesNew (29.84s)
=== RUN   TestAccAWSSSMParameter_fullPath
--- PASS: TestAccAWSSSMParameter_fullPath (16.46s)
--- PASS: TestAccAWSSSMParameter_secure (16.27s)
--- PASS: TestAccAWSSSMParameter_secure_with_key (47.42s)
=== RUN   TestAccAWSSSMParameter_secure_keyUpdate
--- PASS: TestAccAWSSSMParameter_secure_keyUpdate (59.86s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       272.432s
```

### DIFF
--- a/aws/resource_aws_ssm_parameter.go
+++ b/aws/resource_aws_ssm_parameter.go
@@ -72,16 +72,18 @@ func resourceAwsSsmParameter() *schema.Resource {
 
 func resourceAwsSmmParameterExists(d *schema.ResourceData, meta interface{}) (bool, error) {
 	ssmconn := meta.(*AWSClient).ssmconn
-
-	resp, err := ssmconn.GetParameters(&ssm.GetParametersInput{
-		Names:          []*string{aws.String(d.Id())},
-		WithDecryption: aws.Bool(true),
+	_, err := ssmconn.GetParameter(&ssm.GetParameterInput{
+		Name:           aws.String(d.Id()),
+		WithDecryption: aws.Bool(false),
 	})
-
 	if err != nil {
+		if isAWSErr(err, ssm.ErrCodeParameterNotFound, "") {
+			return false, nil
+		}
 		return false, err
 	}
-	return len(resp.InvalidParameters) == 0, nil
+
+	return true, nil
 }
 
 func resourceAwsSsmParameterRead(d *schema.ResourceData, meta interface{}) error {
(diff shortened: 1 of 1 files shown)


bflad:
At first glance, we should always read the detailed information. The parameter can be changed outside of Terraform and this would prevent Terraform from detecting those changes.

Should we instead add additional parameter filtering on Type and (if provided) KeyId?

julienduchesne:
The AWS SSM API is really bad. Even with additional filtering, the call is still paginated. It just returns pages with 0 results.

bflad:
That sounds like an issue that should be taken up with AWS Support. We likely won't accept something that removes API drift detection in the resource as that goes against the design philosophy of Terraform, but we would however accept additional throttling error handling if necessary.

julienduchesne:
I understand. Looking at the API ref, I found this: https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameterHistory.html. It returns the whole history for a parameter but it should be a performance improvement overall since it fetches a single parameter instead of the whole list. It would replace both GetParameter and DescribeParameters calls with a single one.

A potential issue is if someone has a very long list of changes to a single parameter.

Is this something that would be accepted anyways?

bflad:
I don't think we should implement a workaround (getting a full list of all the parameter versions and finding the latest) just to get the same information that should be provided with the correct API call on the first result. The SSM team should fix their API if its requiring extraneous pagination that triggers throttling issues.

julienduchesne:
Alright, I understand. I'll modify this PR to only include the GetParameters -> GetParameter change.

julienduchesne:
I've made the changes btw, @bflad

PR #5262, opened 2018-07-20
Fix aws_lambda_event_source_mapping flapping

julienduchesne:

* When creating an event source mapping with a function name instead of an arn, it would show a diff on every plan. function_name supports both arn and name though, so what this does is set the `function_name` to the ARN if it is an import or if an ARN was given as an attribute and it extracts the function's name if a function name was given as the attribute.

```
TESTARGS='-run=TestAccAWSLambdaEventSourceMapping'  make testacc
TF_ACC=1 go test ./... -v -run=TestAccAWSLambdaEventSourceMapping -timeout 120m
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_basic
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_basic (114.88s)
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_removeBatchSize
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_removeBatchSize (111.80s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqs_basic
--- PASS: TestAccAWSLambdaEventSourceMapping_sqs_basic (129.41s)
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_import
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_import (100.45s)
=== RUN   TestAccAWSLambdaEventSourceMapping_kinesis_disappears
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_disappears (96.39s)
=== RUN   TestAccAWSLambdaEventSourceMapping_sqsDisappears
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsDisappears (136.59s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       689.550s
```

### DIFF
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -34,8 +34,16 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 				ForceNew: true,
 			},
 			"function_name": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"function_arn"},
+				Deprecated:    "Use `function_arn` instead.",
+			},
+			"function_arn": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"function_name"},
+				ValidateFunc:  validateArn,
 			},
 			"starting_position": {
 				Type:     schema.TypeString,
@@ -77,10 +85,6 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 				Optional: true,
 				Default:  true,
 			},
-			"function_arn": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
 			"last_modified": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -110,7 +114,10 @@ func resourceAwsLambdaEventSourceMapping() *schema.Resource {
 func resourceAwsLambdaEventSourceMappingCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).lambdaconn
 
-	functionName := d.Get("function_name").(string)
+	functionName, err := getLambdaEventSourceMappingFunctionName(d)
+	if err != nil {
+		return err
+	}
 	eventSourceArn := d.Get("event_source_arn").(string)
 
 	log.Printf("[DEBUG] Creating Lambda event source mapping: source %s to function %s", eventSourceArn, functionName)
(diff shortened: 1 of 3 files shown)


julienduchesne:
Hi @bflad. I actually chose to do it this way on purpose. Let me explain my reasoning and you can tell me if it makes sense or not and if it follows the `spirit` of the provider. 

1. As a user, I expect that resource arguments such as `function_name` will be exactly the same in the state as in my terraform files. What happens right now is that I enter a function name in that field and provider overwrites it with an ARN. If I implement my logic within a `DiffSuppressFunc`, the state will still contain an ARN even though I entered a name, it's inconsistent. Even more so if we consider the fact that the field is called `function_name` and it always contains an ARN.
2. Existing configurations are not broken. You can still enter an ARN or a function name in the `function_name` field. The `Read` will then write the correct value instead of overwriting with an ARN.

If that makes sense, I can of course add a few tests!

Thanks

bflad:
We have plenty of cases in the provider where we flip flop between some form of `kms_key_id`/`kms_key_arn` attribute names, the API accepts both key ID _and_ ARN for the value, and the API generally converts the response to ARN when reading it back. This is great for operators, but not so great for managing declarative configuration. 😄 

In quite a few cases, we went with naming the attributes `X_arn` and plan-time validate ARNs, even if the parameter accepts both forms. In my opinion, forcing ARN usage makes for a more consistent user experience and keeps any cross-account support. The added bonus is that we do not need any complicated logic to munge API responses.

If it instead makes more sense we create a new `function_arn` attribute and deprecate the existing `function_name` one, we can do that.

julienduchesne:
Woops, didn't mean to close this. In a case like this,  can we just stick the `Removed` attribute on `function_name` and make `function_arn` required or do we have to go with the `Deprecated` attribute first before removing.

Because, if we have to have both attributes usable, it gets complicated. We have to specify one of them but we can't set both as `Required`. Also, CustomizeDiff for plan-time validation does not work because function_name could be from another resource's computed field (though my PR here: https://github.com/hashicorp/terraform/pull/18388 would sure help in that case). Am I missing something?

If not, I'm probably going to go with the DiffSupress :)

Thanks for your time on my PRs, btw

bflad:
HashiCorp released Terraform providers follow (or should be following) specific versioning guidelines, as outlined here: https://www.hashicorp.com/blog/hashicorp-terraform-provider-versioning -- we follow the full deprecation cycle for any breaking changes where possible, otherwise it will only be done on major version updates.

If I was deprecating the old attribute, here's what I'd be doing:

* Add new `Optional` attribute with expected behavior and `ConflictsWith` to the old attribute
* In the old attribute add `ConflictsWith` with the new attribute
* In the old attribute add `Deprecated` with a message saying to use the new attribute
* In creation function, return apply-time error if neither argument is defined.
* Wrap `d.Set()` calls on the old attribute with `d.GetOk()`

e.g.

```go
if _, ok := d.GetOk("deprecated_attribute"); ok {
  d.Set("deprecated_attribute", ...)
} else {
  d.Set("new_attribute", ...)
}
```

Or, we could also keep providing the old attribute in state and ignoring differences by marking it `Computed: true`.

The temporary lack of plan-time validation, assuming there is good documentation, is less of a concern rather than moving things forward.

I really need to write up a full deprecation guide in the Extending Terraform section of the documentation... 😅

PR #5288, opened 2018-07-21
Fixed Throttling Lambda Invocations Through Reserved Concurrency 

dthagard:
This PR addresses the issue here (terraform-providers#5287) where setting a value of 0 for the reserved_concurrent_executions will not properly enable Reserved Concurrency for the Lambda.

Using the test Terraform code in the referenced issue, I get the expected output:

```
$ aws lambda get-function --function-name lambda_function_name
{
    "Configuration": {
        "FunctionName": "lambda_function_name",
        "FunctionArn": "arn:aws:lambda:us-east-1:XXXXXXXXXXXX:function:lambda_function_name",
        "Runtime": "go1.x",
        "Role": "arn:aws:iam::XXXXXXXXXXXX:role/iam_for_lambda",
        "Handler": "main",
        "CodeSize": 2661873,
        "Description": "",
        "Timeout": 3,
        "MemorySize": 128,
        "LastModified": "2018-07-21T01:41:36.494+0000",
        "CodeSha256": "VBK96LKeV2/y/h38A2P8yXhiaVGy52e05yxfJcEi4GI=",
        "Version": "$LATEST",
        "TracingConfig": {
            "Mode": "PassThrough"
        }
    },
    "Code": {
        "RepositoryType": "S3",
        "Location": "REDACTED"
    },
    "Concurrency": {
        "ReservedConcurrentExecutions": 0
    }
}
```

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -413,7 +413,7 @@ func resourceAwsLambdaFunctionCreate(d *schema.ResourceData, meta interface{}) e
 
 	d.SetId(d.Get("function_name").(string))
 
-	if reservedConcurrentExecutions > 0 {
+	if reservedConcurrentExecutions >= 0 {
 
 		log.Printf("[DEBUG] Setting Concurrency to %d for the Lambda Function %s", reservedConcurrentExecutions, functionName)
 


bflad:
Hi @dthagard 👋  Thanks for submitting this!

The changes required here are unfortunately a little more nuanced than just a simple logic change. A more completely implementation can be found in the previous PR #3806 that is just waiting for updates. Could you please take a look at that PR and provide feedback or implement commits on top of the previous work?

Some background around this is:
* There are potentially many configurations relying on the current fact that `0` means disabling setting reserved concurrency in Terraform
* Allowing Terraform to actually set `0` for reserved concurrency is a very dangerous change for existing configurations that are depending on the above
* `schema.TypeInt` has a "zero-value"/"default" of `0` -- we will need to change the default of the argument to something like `-1` (we do not currently have a way to specify nullable attribute types in the schema)

The trifecta described above makes enabling this `0` setting potentially reserved for only a major version upgrade since it can very easily cause production downtime. Sorry that this is more difficult than it should be; catching items like this during an initial implementation is tricky. 😅

dthagard:
I understand your concern. When we were looking to implement this feature, one of our developers suggested using a value of `0` as default "to disable concurrency" which caused me to start looking into what the actual code is doing versus what the API expects. 

I agree that the other implementation is more robust, so I will close this PR.

PR #5317, opened 2018-07-25
Route53 import private zone with multiple vpc

### LINKED ISSUE #892: Can't import private Route53 Hosted Zone with more than one VPC association
### Terraform Version
v0.9.8

### Affected Resource(s)
- aws_route53_zone

None, using `terraform import`.

### Debug Output
Full debug output not included due to this not being a `bug` per se.
https://gist.github.com/ttacon/1e26e775e4850211c629c02f44346e7d

### Panic Output
None.

### Expected Behavior
Expected it to import the private Route53 Hosted Zone.

### Actual Behavior
Failed to import private Route53 Hosted Zone with error message.

### Steps to Reproduce
1. `terraform import aws_route53_zone.myzone $route53HostedZone `

### Important Factoids
None.

### References
 - https://github.com/hashicorp/terraform/pull/1526

It seems as though `terraform import` can't import a Route53 private hosted zone that is associated with more than one VPC. The [code clearly shows](https://github.com/terraform-providers/terraform-provider-aws/blob/b8d4e1570fc43a2acee6b6e47f63c9db6b067fa2/aws/resource_aws_route53_zone.go#L183) that this is being explicitly disallowed, but I'm not sure why that is? I understand that there may be some difficulty/schema mis-match since creating a private hosted zone with more than one VPC associated requires some massaging (see linked terraform PR). But I was wondering if perhaps this was something on the roadmap/if I could get some clarification as to logic behind the comment in the linked section of code? I'm fairly certain it's due to the mismatch between the `GetHostedZoneResponse` model and the `CreateHostedZoneRequest` model, but I'd just like clarification.

Thanks!

julienduchesne:
Fixes #892 

* Adds an import handler for the route53 zone resource. This handler will create zone association resources when there are more than one VPC associated to a route53 zone.
* Allow the modification of the route53 zone resource `vpc_id` attribute without forcing a new resource. During import, the VPC ids can end up in the wrong resources (Ex: a vpc_id that you want in the main resource might go in the association resource and vice versa).

```
TESTARGS='-run=TestAccAWSRoute53Zone_importBasic'  make testacc
TF_ACC=1 go test ./... -v -run=TestAccAWSRoute53Zone_importBasic -timeout 120m
=== RUN   TestAccAWSRoute53Zone_importBasic
--- PASS: TestAccAWSRoute53Zone_importBasic (80.99s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       81.002s

TESTARGS='-run=TestAccAWSRoute53Zone_importTwoVpcs'  make testacc
TF_ACC=1 go test ./... -v -run=TestAccAWSRoute53Zone_importTwoVpcs -timeout 120m
        === RUN   TestAccAWSRoute53Zone_importTwoVpcs
--- PASS: TestAccAWSRoute53Zone_importTwoVpcs (214.95s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       214.970s
```


bflad:
> Allow the modification of the route53 zone resource vpc_id attribute without forcing a new resource. During import, the VPC ids can end up in the wrong resources (Ex: a vpc_id that you want in the main resource might go in the association resource and vice versa).

Should we deprecate the `vpc_id` argument on the `aws_route53_zone` resource? If multiple associations are possible and the associations are indeterminately ordered, it sounds like we should remove support for the ambiguous argument.

julienduchesne:
The issue is that you need at least one vpc when you create a private zone. I have another branch where I tried to deprecate `vpc_id` using multiple vpc blocks inside the `aws_route53_zone` but if we handle it that way, we also deprecate the possibility to create zone_associations using a count

bflad:
Ah, bummer! I was being optimistic that they would remove that restriction. 😖 

```
aws_route53_zone.main: InvalidInput: When you're creating a private hosted zone (when you specify true for PrivateZone), you must also specify values for VPCId and VPCRegion.
```

julienduchesne:
Oh, I just saw that Terraform 0.12 will support iteration on blocks https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each. Here's what I would do then:
1. Support vpc_id and vpc_region as is. With the deprecated attribute
2. Implement a new vpc block so that you can add as many vpcs as you want in a zone

Works?

bflad:
We'll also want to look at this in the context of this issue: #384 (PR #2005) -- we'll need to be sure that you can build a Terraform configuration with the proper ordering on a single apply to support cross-account authorizations. I think we'd still need to support a separate association resource in that case to get the [create zone] -> [cross-account authorize] -> [create association] ordering right, but we cannot suggest everyone to use `ignore_changes = ["vpc"]` in the zone for that scenario. 😕 

It'd be so nice if the API was easier to work with! 😄

julienduchesne:
Argh, it might do more harm than good then. At least with this PR, we can change the vpc_id and vpc_region without recreating the zone entirely.

PR #5376, opened 2018-07-29
New resource: aws_neptune_cluster_instance

saravanan30erd:
Reference #4713 

* Change 1
Added new resource aws_neptune_cluster_instance

### DIFF
--- a/aws/resource_aws_neptune_cluster_instance.go
+++ b/aws/resource_aws_neptune_cluster_instance.go
@@ -188,19 +188,6 @@ func resourceAwsNeptuneClusterInstance() *schema.Resource {
 				Default:  0,
 			},
 
-			"performance_insights_enabled": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Computed: true,
-			},
-
-			"performance_insights_kms_key_id": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ValidateFunc: validateArn,
-			},
-
 			"tags": tagsSchema(),
 		},
 	}
--- a/aws/resource_aws_neptune_cluster_instance.go
+++ b/aws/resource_aws_neptune_cluster_instance.go
@@ -205,3 +205,155 @@ func resourceAwsNeptuneClusterInstance() *schema.Resource {
 		},
 	}
 }
+
+func resourceAwsNeptuneClusterInstanceCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).neptuneconn
+	tags := tagsFromMapNeptune(d.Get("tags").(map[string]interface{}))
+
+	createOpts := &neptune.CreateDBInstanceInput{
+		DBInstanceClass:         aws.String(d.Get("instance_class").(string)),
+		DBClusterIdentifier:     aws.String(d.Get("cluster_identifier").(string)),
+		Engine:                  aws.String(d.Get("engine").(string)),
+		PubliclyAccessible:      aws.Bool(d.Get("publicly_accessible").(bool)),
+		PromotionTier:           aws.Int64(int64(d.Get("promotion_tier").(int))),
+		AutoMinorVersionUpgrade: aws.Bool(d.Get("auto_minor_version_upgrade").(bool)),
+		Tags: tags,
+	}
+
+	if attr, ok := d.GetOk("availability_zone"); ok {
+		createOpts.AvailabilityZone = aws.String(attr.(string))
+	}
+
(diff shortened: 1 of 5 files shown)


aws/resource_aws_neptune_cluster_instance.go
```diff
@@ -0,0 +1,560 @@
...
+	if err != nil {
+		if isAWSErr(err, neptune.ErrCodeDBInstanceNotFoundFault, "") {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("Error retrieving Neptune Instances: %s", err)
+	}
+
+	if len(resp.DBInstances) != 1 ||
+		aws.StringValue(resp.DBInstances[0].DBInstanceIdentifier) != id {
+		if err != nil {
```
bflad: Anytime `err != nil`, it is previously `return`ed -- this conditional seems like it should be removed.

aws/resource_aws_neptune_cluster_instance.go
```diff
@@ -0,0 +1,560 @@
...
+				ValidateFunc: validateOnceAWeekWindowFormat,
+			},
+
+			"promotion_tier": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Default:  0,
+			},
+
+			"publicly_accessible": {
```
bflad: Is this attribute updatable? It is not handled by the update function currently and should either be added there or have `ForceNew: true` added

aws/resource_aws_neptune_cluster_instance.go
```diff
@@ -0,0 +1,560 @@
...
+		req.DBParameterGroupName = aws.String(d.Get("neptune_parameter_group_name").(string))
+		requestUpdate = true
+	}
+
+	if d.HasChange("instance_class") {
+		req.DBInstanceClass = aws.String(d.Get("instance_class").(string))
+		requestUpdate = true
+	}
+
+	if d.HasChange("monitoring_role_arn") {
```
bflad: We might want to add an acceptance test to verify that both the `MonitoringRoleArn` and `MonitoringInterval` parameters do not need to specified at the same time during update. See also: #2188

saravanan30erd: @bflad Looks like there is no support for enhanced monitoring for neptune engine 👍 

`aws_neptune_cluster_instance.cluster_instances: error creating Neptune Instance: InvalidParameterCombination: Enhanced Monitoring is not supported for the neptune database engine. Go to the RDS documentation for information about supported database engines.`

saravanan30erd:
@bflad done all the changes

aws/resource_aws_neptune_cluster_instance.go
```diff
@@ -0,0 +1,521 @@
...
+			if aws.BoolValue(m.IsClusterWriter) == true {
+				d.Set("writer", true)
+			} else {
+				d.Set("writer", false)
+			}
+		}
+	}
+
+	if db.Endpoint != nil {
+		d.Set("endpoint", db.Endpoint.Address)
```
bflad: Nitpick: with the `aws_db_instance` resource we have 3 attributes (`address`, `endpoint`, and `port`):

https://github.com/terraform-providers/terraform-provider-aws/blob/2c5c645f5a088c38ac490f032adaf1706e4c09a7/aws/resource_aws_db_instance.go#L987-L995

For a few reasons I think we should follow suit here:
* Provides an attribute with only the hostname
* Matches the API naming for address
* Ancillary: Reduces any confusion working between the two resources

aws/resource_aws_neptune_cluster_instance.go
```diff
@@ -0,0 +1,521 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsNeptuneClusterInstance() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsNeptuneClusterInstanceCreate,
+		Read:   resourceAwsNeptuneClusterInstanceRead,
+		Update: resourceAwsNeptuneClusterInstanceUpdate,
+		Delete: resourceAwsNeptuneClusterInstanceDelete,
+		Importer: &schema.ResourceImporter{
```
bflad: Resource import documentation is missing

bflad:
This has been released in [version 1.30.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1300-august-02-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5394, opened 2018-07-30
Update acm_certificate_validation.html.markdown

umesh55:
Replacing `aws_route53_record.cert_validation.fqdn` with `resource_record_name` as fqdn errors our with `FATAL problem: DomainLabelEmpty (Domain label is empty). `

Fixes #0000

* Change 1
* Change 2

```

...
```

### DIFF
--- a/website/docs/r/acm_certificate_validation.html.markdown
+++ b/website/docs/r/acm_certificate_validation.html.markdown
@@ -43,7 +43,7 @@ resource "aws_route53_record" "cert_validation" {
 
 resource "aws_acm_certificate_validation" "cert" {
   certificate_arn = "${aws_acm_certificate.cert.arn}"
-  validation_record_fqdns = ["${aws_route53_record.cert_validation.fqdn}"]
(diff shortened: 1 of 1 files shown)


bflad:
Hi @umesh55 👋 

The reason for using `aws_route53_record.cert_validation.fqdn` there is to guarantee ordering between the resources. This ensures the DNS record is created before trying to verify the ACM certificate, otherwise the verification may never complete. As such, we cannot accept this change as the example in the documentation is working as intended and verified by many practitioners.

I'd be curious if your `FATAL problem: DomainLabelEmpty (Domain label is empty)` is related to a mismatch between `zone_id` and the domain/zone name, similar to #5237. Version 1.29.0 of the AWS provider should prevent the `DomainLabelEmpty` error and actually create the Route53 record, but it will likely be in the "wrong" domain as that is how you get to that error.  You'll need to update your Terraform configuration to align zone ID to the correct domain name in that case.

If you're still having trouble unrelated to a configuration error, I'd suggest creating a bug report issue. 👍

PR #5395, opened 2018-07-31
Support ena_support

### LINKED ISSUE #206: Add support for Enhanced Networking to aws_instance resource.
_This issue was originally opened by @kwilczynski as hashicorp/terraform#7815. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Amazon at the moment supports two flavours of Enhanced Networking feature:

Overview and [documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html).
- **SR-IOV** (Intel 82599 VF)
  http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sriov-networking.html
- **Elastic Network Adapter** (ENA)
  http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking-ena.html

For any VPC-enabled and HVM-compatible EC2 instances.

We should allow users to enable Enhanced Network support and choose which flavour they would like to use. A very common use case, would be to run and span a Docker overlay network layer across multiple instances participating as hosts in the Docker Swarm. Such nodes do benefit from more linear and consistent performance which Enhanced Networking be bring, especially when latency is concerned and momentary spikes in such and "choking up" network throughput (e.g. exhausting the packer per-second rate, etc.) is not desired.

For this to work an AMI (Amazon Machine Image) needs to have moderately up-to-date Linux kernel and include supported version of both the `ixgbevf`[1] and `ena`[2] kernel driver, thus it would be prudent to to include latest drives and to enable support for the Enhanced Networking when creating (with Packer, etc.) custom AMIs for both the EBS and Instance Store type.

The Elastic Network Adapter is quite new (see the [announcement](https://aws.amazon.com/about-aws/whats-new/2016/06/introducing-elastic-network-adapter-ena-the-next-generation-network-interface-for-ec2-instances/)) and it is not yet supported on Windows due to lack of available drivers for this platform (an on-going development effort by Amazon), plus the only instance type that supports ENA at the moment is X1, but hopefully Amazon would add ENA support to others later.

Worth noting is, that the Amazon Linux HVM-enabled AMIs already ship with both of the drivers included and have Enhanced Networking support enabled, thus are ready to use from the get-go.

API documentation:
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceAttribute.html
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyInstanceAttribute.html

Conceptually, this very similar to the way how the "SourceDestCheck" check is now supported in Terraform, as per: [resource_aws_instance.go#L595-L603](https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_instance.go#L595-L603)

Related to https://github.com/mitchellh/packer/issues/3688 in Packer.
1. Intel 82599 VF (SR-IOV) [drivers](https://sourceforge.net/projects/e1000/files/ixgbevf%20stable/).
2. Amazon Elastic Network Adapter (ENA) [drivers](https://github.com/amzn/amzn-drivers).

nbrownus:
Partially Fixes #206

- Exposes `ena_support` on `aws_ami`

### DIFF
--- a/aws/resource_aws_ami.go
+++ b/aws/resource_aws_ami.go
@@ -59,6 +59,7 @@ func resourceAwsAmiCreate(d *schema.ResourceData, meta interface{}) error {
 		RootDeviceName:     aws.String(d.Get("root_device_name").(string)),
 		SriovNetSupport:    aws.String(d.Get("sriov_net_support").(string)),
 		VirtualizationType: aws.String(d.Get("virtualization_type").(string)),
+		EnaSupport:         aws.Bool(d.Get("ena_support").(bool)),
 	}
 
 	if kernelId := d.Get("kernel_id").(string); kernelId != "" {
@@ -196,6 +197,7 @@ func resourceAwsAmiRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("root_snapshot_id", amiRootSnapshotId(image))
 	d.Set("sriov_net_support", image.SriovNetSupport)
 	d.Set("virtualization_type", image.VirtualizationType)
+	d.Set("ena_support", image.EnaSupport)
 
 	var ebsBlockDevs []map[string]interface{}
 	var ephemeralBlockDevs []map[string]interface{}
@@ -567,6 +569,11 @@ func resourceAwsAmiCommonSchema(computed bool) map[string]*schema.Schema {
 		"manage_ebs_snapshots": {
 			Type:     schema.TypeBool,
 			Computed: true,
+			ForceNew: true,
+		},
+		"ena_support": {
+			Type:     schema.TypeBool,
+			Optional: true,
 			ForceNew: true,
 		},
 	}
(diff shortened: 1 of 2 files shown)


bflad:
This has been released in [version 1.30.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1300-august-02-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

rkhozinov:
Hey, @bflad @nbrownus what about the related change for the `aws_ami_copy ` resource?

`aws_ami_copy.encrypted_ami (new resource required)
      id:                                                 "ami-***" => <computed> (forces new resource)
      architecture:                                       "x86_64" => <computed>
      description:                                        "AMI***" => "AMI***"
      ebs_block_device.#:                                 "9" => <computed>
      ena_support:                                        "true" => "false" (forces new resource)
      encrypted:                                          "true" => "true"
      ephemeral_block_device.#:                           "0" => <computed>`

samjgalbraith:
I think this is causing Terraform to want to replace our AMIs. It states that the state of the current AMI resource has a value ena_enabled = true (even though the flag was never available via Terraform before, it seems to have had true by default) and it wants to replace the AMI to change ena_enabled to false.

PR #5405, opened 2018-07-31
Added arg force_delete on s3 bucket documentation

lde:
Update Documentation added documentation for arg force_delete in s3_bucket page.


bflad:
Hi @lde 👋 Are you sure this attribute is available in the `aws_s3_bucket` data source? The schema does not define this as an available attribute:

https://github.com/terraform-providers/terraform-provider-aws/blob/2c5c645f5a088c38ac490f032adaf1706e4c09a7/aws/data_source_aws_s3_bucket.go#L17-L46

Its currently documented and available in the `aws_s3_bucket` resource: https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#force_destroy

I believe the flag only exists within Terraform so I do not think it would be possible or make sense to make available in the data source.

lde:
Hi,
@bflad 
Sorry I was in bad page. You're right.

PR #5267, opened 2018-07-20
Support Performance Insights for aws_db_instance

rizkidoank:
Fixes #5259

Changes proposed in this pull request:

* Add parameter `performance_insights_enabled` to enabled performance insights in `aws_db_instance`
* Add parameter`performance_insights_kms_key_id` to set kms key used for performance insights in `aws_db_instance`

Output from acceptance testing:

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSDBInstance_performanceInsight'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSDBInstance_performanceInsight -timeout 120m
=== RUN   TestAccAWSDBInstance_performanceInsight
--- PASS: TestAccAWSDBInstance_performanceInsight (524.49s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       524.546s
...
```

PR #5406, opened 2018-07-31
resource/aws_db_instance: Support Performance Insights for aws_db_instance

### LINKED ISSUE #5259: Support performance insights on PostgreSQL RDS (non-aurora)
Hi - 

Performance insights for non-aurora PostgreSQL RDS databases is now in GA. 

https://aws.amazon.com/about-aws/whats-new/2018/04/rds-performance-insights-on-rds-for-postgresql/

It would be great if this could be added to the resource definition.

Thanks!

rizkidoank:
Fixes #5259

* Add parameter `performance_insights_enabled` to enabled performance insights in `aws_db_instance`
* Add parameter `performance_insights_kms_key_id` to set kms key used for performance insights in `aws_db_instance`
* Add parameter `performance_insights_retention_period` to set retention period of performance insights

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSDBInstance_performanceInsights'
TF_ACC=1 go test ./aws -v -run=TestAccAWSDBInstance_performanceInsights -timeout 120m
=== RUN   TestAccAWSDBInstance_performanceInsights
--- PASS: TestAccAWSDBInstance_performanceInsights (1024.11s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       1024.194s
```

I close PR #5267 and open a new one because I was messed up when trying to fix the conflict with master branch.


aws/resource_aws_db_instance.go
```diff
@@ -404,6 +404,25 @@ func resourceAwsDbInstance() *schema.Resource {
 				},
 			},
 
+			"performance_insights_enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
```
bflad: We should not set [`Computed`](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed) on this attribute or any of the others. Operators should know if PI has been configured outside Terraform. See also: https://www.terraform.io/docs/extend/best-practices/detecting-drift.html

aws/resource_aws_db_instance.go
```diff
@@ -404,6 +404,25 @@ func resourceAwsDbInstance() *schema.Resource {
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateArn,
+			},
+
+			"performance_insights_retention_period": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
```
bflad: What is the default retention period if PI is enabled? Instead of using `Computed: true` here we likely need to configure Terraform to ignore the difference of the default value when its enabled.

```go
// Example DiffSuppressFunc if PI defaults to 7 when enabled, otherwise we expect 0 (undefined from API)
	if !d.Get("performance_insights_enabled").(bool) {
		return false
	}
	if old == "7" && new == "0" {
		return true
	}
	return false
},
```

aws/resource_aws_db_instance.go
```diff
@@ -970,6 +1013,9 @@ func resourceAwsDbInstanceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("kms_key_id", v.KmsKeyId)
 	d.Set("port", v.DbInstancePort)
 	d.Set("iam_database_authentication_enabled", v.IAMDatabaseAuthenticationEnabled)
+	d.Set("performance_insights_enabled", v.PerformanceInsightsEnabled)
+	d.Set("performance_insights_kms_key_id", v.PerformanceInsightsKMSKeyId)
+	d.Set("performance_insights_retention_period", v.PerformanceInsightsRetentionPeriod)
```
bflad: We might need to convert this attribute's type before `d.Set()` -- this can be verified with either:

```go
if err := d.Set("performance_insights_retention_period", v.PerformanceInsightsRetentionPeriod); err != nil {
  return fmt.Errorf("error setting performance_insights_retention_period: %s", err)
}
```

Or running the acceptance testing with `TF_SCHEMA_PANIC_ON_ERROR=1`.

If it does have an error (like cannot convert `*int64` to `int`), then the fix is to use: `int(aws.Int64Value(v.PerformanceInsightsRetentionPeriod))`

rizkidoank:
@bflad thanks for the review. I do have time for this. But, my experience in Go is not that much.

```
One other note, the CreateDBInstanceReadReplicaInput can accept the PI parameters so they should be set in that scenario.
```
So for `CreateDBInstanceReadReplicaInput` I have to implement PI parameters like in db instance?
 
```
As for RestoreDBInstanceFromDBSnapshotInput, logic similar to how it triggers security group updates post creation should be used to prevent the need to run terraform apply twice.
```
Did you have reference for these? So I can look into it more.

Thank you!

bflad:
You can find the AWS SDK for Go documentation here, which has all the types and parameters they accept: https://docs.aws.amazon.com/sdk-for-go/api/service/rds/

Specifically, [`CreateDBInstanceReadReplicaInput`](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#CreateDBInstanceReadReplicaInput) accepts the three Performance Insights parameters. 😄 

The note about [`RestoreDBInstanceFromDBSnapshotInput`](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#RestoreDBInstanceFromDBSnapshotInput) comes from those three Performance Insight parameters not being able to set in there, similar to some others we would normally expect to be there like setting the security group. In those cases, to prevent Terraform from needing to be run twice to successfully apply the change we want, we automatically call the update function for the resource.

Here's the logic I'm referencing:

https://github.com/terraform-providers/terraform-provider-aws/blob/e79d81c6f077893afcd18f82483da9bde284ae2d/aws/resource_aws_db_instance.go#L734-L744

It seems we should also perform an update on create when Performance Insights is enabled. Hope this makes sense.

aws/resource_aws_db_instance.go
bflad: Correct 😄

willthames:
@rizkidoank - why was this PR closed? Are there unsalvageable problems?

nunofernandes:
Any info on this?

wagnerone:
Tried unsuccessfully to enable PI for aws_db_instance today w/ TF and ended up doing it manually in the console.

PR #5421, opened 2018-08-01
Fixing aws_dax_cluster example usage.

zabawaba99:

* Fix aws_dax_cluster example usage as it shows using `cluster_id` (which does not exist) instead of `cluster_name`


GrantSheehan:
Since it's using the Elasticache resource's validator (which uses `cluster_id`), should we rename the schema attribute on the DAX resource to match?

PR #5427, opened 2018-08-02
Update api_gateway_method_settings.html.markdown

li0nel:
As per AWS documentation, the method_path parameter must be the full method path relative to the API base URL, without the first forward slash.

Current Terraform documentation only includes the last method path part, which fails to update settings on methods with deeper paths.

Fixes #0000

* Change 1
* Change 2

```

...
```


zoltan-toth-mw:
According to the examples, it starts with a forward slash acutally.
i.e.: https://docs.aws.amazon.com/cli/latest/reference/apigateway/update-stage.html

Anyway the original doc with path_part is totally incorrect.

website/docs/r/api_gateway_method_settings.html.markdown
```diff
@@ -16,7 +16,7 @@ Provides an API Gateway Method Settings, e.g. logging or monitoring.
 resource "aws_api_gateway_method_settings" "s" {
   rest_api_id = aws_api_gateway_rest_api.test.id
   stage_name  = aws_api_gateway_stage.test.stage_name
-  method_path = "${aws_api_gateway_resource.test.path_part}/${aws_api_gateway_method.test.http_method}"
+  method_path = "${trimprefix(aws_api_gateway_resource.test.path, "/)}/${aws_api_gateway_method.test.http_method}"
```
bflad, line 19:
```suggestion
  method_path = "${trimprefix(aws_api_gateway_resource.test.path, "/")}/${aws_api_gateway_method.test.http_method}"
```

bflad (approved):
Thank you for this @li0nel and apologies it has been lingering.

Issue #5118, opened 2018-07-07
resource/aws_elasticache_replication_group: support setting primary/replica AZ atrributes inside NodeGroupConfiguration for RedisClusterModelEnabled


moose007:
Currently I do not have a way to express in an `aws_elasticache_replication_group` resource block which availability zone my primary node gets deployed in and which AZ its read replicas are deployed in. The AWS ElastiCache SDK supports the following attributes:

**PrimaryAvailabilityZone**

    The Availability Zone where ElastiCache launches the node group's primary node.

    Required: No

    Type: String

**ReplicaAvailabilityZones**

    A list of Availability Zones where ElastiCache launches the read replicas. The number of Availability Zones must match the value of the ReplicaCount property or, if you don't specify the ReplicaCount property, the replication group's ReplicasPerNodeGroup property.

    Required: No

    Type: List of String values

(see: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticache-replicationgroup-nodegroupconfiguration.html)

It would be useful for `aws_elasticache_replication_group` resource to provide the same level of control.

moose007:
Some notes from @bflad:

`For Redis Cluster Mode Enabled replication groups (e.g. when using cluster_mode in Terraform), we currently do not have the ability to set availability zones via the NodeGroupConfiguration parameter, which would likely require changing the cluster_mode argument.` (from https://github.com/terraform-providers/terraform-provider-aws/issues/5104)

YakDriver:
Thanks @moose007 for raising this and so sorry it has taken us a minute to get to it! See #44879 for progress on this front.

PR #5452, opened 2018-08-04
Fetching All endpoints in Elastic cache Redis cluster

abhijitpati:
Fetching All endpoints in Elastic cache Redis cluster


gdavison:
Thanks for the PR, @abhijitpati. As part of the work related to #5118, we're adding a `node_groups` structure which will contain all the information related to each node group, not just the endpoints. The list of endpoints will then be accessible  from the members of the list.

I'm going to close this PR since we are working on an alternative design. Thanks again!

PR #5468, opened 2018-08-07
Support aurora engine type in performance insights

w32-blaster:

* Allow turning on performance insights for the RDS cluster engine type "aurora"
Since yesterday, performance insights is supported on aurora clusters with mysql version 5.6 (aurora 1.x). 
On AWS API level there is no change, we just have to remove the limitation we implemented in the provider.
Since other engine types are still not supported, we allow the following 2 engines to enable this attribute:
- aurora-postgresql
- aurora


bflad:
This has been released in [version 1.32.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1320-august-16-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5472, opened 2018-08-08
Fix read for non-pending aws_acm_certificate

miquella:
If a DomainValidation indicates a validation status of success or failed there is no reason for it to be re-queried, even if no ResourceRecord or ValidationEmails are present. Re-querying in either of these cases will cause the import of the resource to fail and prevent the certificate from being imported.

Now the certificate is only re-queried if the DomainValidation is in a pending state.

This bug was encountered while attempting to import older, but still in-use, certificates that no longer include domain validation data, preventing us from importing these certificates into Terraform.

### DIFF
--- a/aws/resource_aws_acm_certificate.go
+++ b/aws/resource_aws_acm_certificate.go
@@ -216,7 +216,7 @@ func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]
 				for _, validationEmail := range o.ValidationEmails {
 					emailValidationResult = append(emailValidationResult, *validationEmail)
 				}
-			} else {
+			} else if *o.ValidationStatus == acm.DomainStatusPendingValidation {
 				log.Printf("[DEBUG] No validation options need to retry: %#v", o)
 				return nil, nil, fmt.Errorf("No validation options need to retry: %#v", o)
 			}


aws/resource_aws_acm_certificate.go
```diff
@@ -216,7 +216,7 @@ func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]
 				for _, validationEmail := range o.ValidationEmails {
 					emailValidationResult = append(emailValidationResult, *validationEmail)
 				}
-			} else {
+			} else if *o.ValidationStatus == acm.DomainStatusPendingValidation {
```
bflad: Found during testing: Right after `RequestCertificate`, ACM API will omit `ValidationStatus` completely from the API response. Adjusting this to also retry if `o.ValidationStatus == nil` helped fix the Go panic.

bflad:
This has been released in [version 1.54.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1540-december-21-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5476, opened 2018-08-08
storagegateway: Add cached iSCSI volume resource, retry on busy connection, fix cache resource updating disk ID

bflad:
Reference #943 

Sorry for bundling these changes together, but it was the creation of this resource and its acceptance testing that highlighted these Storage Gateway API issues.

* New Resource: `aws_storagegateway_cached_iscsi_disk`
* storagegatewayconn: Retry on `InvalidGatewayRequestException: The specified gateway proxy network connection is busy.` which occurs when (near) concurrent operations are performed on a gateway
* resource/aws_storagegateway_cache: Ensure resource updates disk ID after creation due to disk ID relabeling (exhibited by `aws-storage-gateway-2.0.10.0` AMI for tape/volume gateways, but not `aws-thinstaller` AMI for file gateways)

Before `aws_storagegateway_cache` code update:

```
=== RUN   TestAccAWSStorageGatewayCache_TapeAndVolumeGateway
--- FAIL: TestAccAWSStorageGatewayCache_TapeAndVolumeGateway (301.11s)
	testing.go:518: Step 0 error: Check failed: Check 1/3 error: Not found: aws_storagegateway_cache.test
```

```
=== RUN   TestAccAWSStorageGatewayCache_TapeAndVolumeGateway
--- PASS: TestAccAWSStorageGatewayCache_TapeAndVolumeGateway (302.07s)

=== RUN   TestAccAWSStorageGatewayCachedIscsiVolume_Basic
--- PASS: TestAccAWSStorageGatewayCachedIscsiVolume_Basic (306.42s)
=== RUN   TestAccAWSStorageGatewayCachedIscsiVolume_SnapshotId
--- PASS: TestAccAWSStorageGatewayCachedIscsiVolume_SnapshotId (301.38s)
=== RUN   TestAccAWSStorageGatewayCachedIscsiVolume_SourceVolumeArn
--- SKIP: TestAccAWSStorageGatewayCachedIscsiVolume_SourceVolumeArn (0.00s)
	resource_aws_storagegateway_cached_iscsi_volume_test.go:146: This test can cause Storage Gateway 2.0.10.0 to enter an irrecoverable state during volume deletion.
```

### DIFF
--- a/aws/resource_aws_storagegateway_cache.go
+++ b/aws/resource_aws_storagegateway_cache.go
@@ -53,6 +53,8 @@ func resourceAwsStorageGatewayCacheCreate(d *schema.ResourceData, meta interface
 		return fmt.Errorf("error adding Storage Gateway cache: %s", err)
 	}
 
+	d.SetId(fmt.Sprintf("%s:%s", gatewayARN, diskID))
+
 	// Depending on the Storage Gateway software, it will sometimes relabel a local DiskId
 	// with a UUID if previously unlabeled, e.g.
 	//   Before conn.AddCache(): "DiskId": "/dev/xvdb",
--- a/aws/resource_aws_storagegateway_cache.go
+++ b/aws/resource_aws_storagegateway_cache.go
@@ -53,6 +53,32 @@ func resourceAwsStorageGatewayCacheCreate(d *schema.ResourceData, meta interface
 		return fmt.Errorf("error adding Storage Gateway cache: %s", err)
 	}
 
+	// Depending on the Storage Gateway software, it will sometimes relabel a local DiskId
+	// with a UUID if previously unlabeled, e.g.
+	//   Before conn.AddCache(): "DiskId": "/dev/xvdb",
+	//   After conn.AddCache(): "DiskId": "112764d7-7e83-42ce-9af3-d482985a31cc",
+	// This prevents us from successfully reading the disk after creation.
+	// Here we try to refresh the local disks to see if we can find a new DiskId.
+
+	listLocalDisksInput := &storagegateway.ListLocalDisksInput{
+		GatewayARN: aws.String(gatewayARN),
+	}
+
+	log.Printf("[DEBUG] Reading Storage Gateway Local Disk: %s", listLocalDisksInput)
+	output, err := conn.ListLocalDisks(listLocalDisksInput)
+	if err != nil {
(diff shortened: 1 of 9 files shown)


aws/resource_aws_storagegateway_cache.go
```diff
@@ -53,6 +53,32 @@ func resourceAwsStorageGatewayCacheCreate(d *schema.ResourceData, meta interface
...
+	// Here we try to refresh the local disks to see if we can find a new DiskId.
+
+	listLocalDisksInput := &storagegateway.ListLocalDisksInput{
+		GatewayARN: aws.String(gatewayARN),
+	}
+
+	log.Printf("[DEBUG] Reading Storage Gateway Local Disk: %s", listLocalDisksInput)
+	output, err := conn.ListLocalDisks(listLocalDisksInput)
+	if err != nil {
+		return fmt.Errorf("error reading Storage Gateway Local Disk: %s", err)
```
catsby: Should we be doing `return` here?

I'm lacking some context, so bare with me. We don't seem to record the results of `AddCache`, what is created? Do we need to track it?

If `conn.AddCache(input)` above is successful, Terraform has _created_ something. If this call errors for whatever reason, valid or not, and we return before calling `d.SetId`, Terraform quits and we've lost track of whatever we created. 

Again, lacking context, sorry 😰

bflad: Sorry if my comment note wasn't clear enough. 😅 

We're basically allocating a disk for a specified usage (cache) within a gateway (e.g. EC2 instance). When allocating said disk, the gateway (sometimes, but annoyingly) changes the disk identifier. Without refreshing the disks to try and find the new disk identifier, Terraform cannot track the resource as the cache is now based on the new disk identifier and there's no way to map between the old and new identifier.

That does make a good point though that it should probably still call the `d.SetId()` with the potentially wrong identifier before this API call just in case it does error and the disk identifier potentially doesn't change. I'll push up a commit momentarily.

bflad:
This has been released in [version 1.32.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1320-august-16-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5508, opened 2018-08-10
r/aws_dax_cluster: Support encryption at rest

ewbankkit:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/5500.
Acceptance tests:

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSDAXCluster_basic'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSDAXCluster_basic -timeout 120m
=== RUN   TestAccAWSDAXCluster_basic
--- PASS: TestAccAWSDAXCluster_basic (669.50s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	669.513s
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSDAXCluster_encryption'
TF_ACC=1 go test ./aws/ -v -run=TestAccAWSDAXCluster_encryption -timeout 120m
=== RUN   TestAccAWSDAXCluster_encryption
--- PASS: TestAccAWSDAXCluster_encryption (596.19s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	596.203s
```

### DIFF
--- a/aws/resource_aws_dax_cluster.go
+++ b/aws/resource_aws_dax_cluster.go
@@ -22,6 +22,7 @@ func resourceAwsDaxCluster() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+		CustomizeDiff: resourceAwsDaxClusterCustomizeDiff,
 
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(45 * time.Minute),
@@ -94,6 +95,21 @@ func resourceAwsDaxCluster() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
+			},
+			"server_side_encryption": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Required: true,
+							ForceNew: true,
+						},
+					},
+				},
 			},
 			"subnet_group_name": {
 				Type:     schema.TypeString,
@@ -186,6 +202,16 @@ func resourceAwsDaxClusterCreate(d *schema.ResourceData, meta interface{}) error
 	if len(preferred_azs) > 0 {
 		azs := expandStringList(preferred_azs)
 		req.AvailabilityZones = azs
+	}
+
+	if v, ok := d.GetOk("server_side_encryption"); ok {
+		options := v.([]interface{})
+		if options[0] == nil {
+			return fmt.Errorf("At least one field is expected inside server_side_encryption")
+		}
+
+		s := options[0].(map[string]interface{})
+		req.SSESpecification = expandDaxEncryptAtRestOptions(s)
 	}
 
 	// IAM roles take some time to propagate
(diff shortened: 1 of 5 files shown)


aws/resource_aws_dax_cluster_test.go
```diff
@@ -239,9 +275,9 @@ EOF
 func testAccAWSDAXClusterConfig(rString string) string {
 	return fmt.Sprintf(`%s
 		resource "aws_dax_cluster" "test" {
- 		  cluster_name       = "tf-%s"
+		  cluster_name       = "tf-%s"
 		  iam_role_arn       = "${aws_iam_role.test.arn}"
-		  node_type          = "dax.r3.large"
+		  node_type          = "dax.t2.small"
```
ewbankkit: Node type changed as encryption at rest is not supported for dax.r3.* node types:
* https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DAXEncryptionAtRest.html

aws/resource_aws_dax_parameter_group.go
```diff
@@ -158,34 +158,3 @@ func resourceAwsDaxParameterGroupDelete(d *schema.ResourceData, meta interface{}
 
 	return nil
 }
-
```
ewbankkit: Moved these 2 functions to `structure.go`.

aws/resource_aws_dax_cluster_test.go
```diff
@@ -137,6 +151,28 @@ func TestAccAWSDAXCluster_resize(t *testing.T) {
 					testAccCheckAWSDAXClusterExists("aws_dax_cluster.test", &dc),
 					resource.TestCheckResourceAttr(
 						"aws_dax_cluster.test", "replication_factor", "1"),
+					resource.TestCheckResourceAttr(
+						"aws_dax_cluster.test", "server_side_encryption.0.enabled", "true"),
```
bflad: It looks like this check was extraneously added 😉 

```
--- FAIL: TestAccAWSDAXCluster_resize (1252.75s)
	testing.go:527: Step 2 error: Check failed: Check 3/3 error: aws_dax_cluster.test: Attribute 'server_side_encryption.0.enabled' expected "true", got "false"
```

aws/resource_aws_dax_cluster_test.go
bflad: 👍

aws/resource_aws_dax_cluster.go
```diff
@@ -95,6 +96,21 @@ func resourceAwsDaxCluster() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
 			},
+			"server_side_encryption": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
```
bflad: I was able to remove the following by changing the schema definition:
* `Computed: true`
* `CustomizeDiff`
* `isDaxClusterOptionDisabled`

Check this out:

```go
"server_side_encryption": {
	Type:     schema.TypeList,
	Optional: true,
	MaxItems: 1,
		if old == "1" && new == "0" {
			return true
		}
		return false
	},
	Elem: &schema.Resource{
			"enabled": {
				Type:     schema.TypeBool,
				Optional: true,
				Default:  false,
			},
		},
	},
},
```

At some point I need to make that `DiffSuppressFunc` into its own function 😅 , its basically just there to ignore the container attribute being unset in the Terraform configuration (all child attribute differences still apply).

ewbankkit: 👍 I reused the logic from [`aws_dynamodb_table`](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_dynamodb_table.go#L43-L51) so we may want to refactor there also.

ewbankkit: Especially as it looks like with [AWS SDK v1.15.16](https://github.com/aws/aws-sdk-go/releases/tag/v1.15.16) `aws_dynamodb_table` SSE can be updated without recreating the table.

bflad:

PR #5522, opened 2018-08-12
Add dhcp_options_id attribute to vpc data docs

borgstrom:
This is missing from the docs.

Defined here: https://github.com/terraform-providers/terraform-provider-aws/blob/8f13308/aws/data_source_aws_vpc.go#L45


radeksimko:
Hi @borgstrom 
thanks for the PR.

The field is actually documented, we just tend to avoid documenting attributes again, if these are already documented as arguments. We may not have been as consistent in doing this historically, but it is generally the approach we take going forward.

See the block of text:

> All of the argument attributes except `filter` blocks are also exported as
 result attributes. This data source will complete the data by populating
 any fields that are not included in the configuration with the data for
 the selected VPC.
 
 > The following attribute is **additionally** exported:

For that reason I'm going to close this PR. Thanks for understanding.

PR #5558, opened 2018-08-15
New resource: aws_dlm_lifecycle_policy

### LINKED ISSUE #5176: Feature request: Lifecycle Management for Amazon EBS Snapshots
As per https://aws.amazon.com/blogs/aws/new-lifecycle-management-for-amazon-ebs-snapshots/.

tomelliff:
Fixes #5176

Adds the `aws_dlm_lifecycle_policy` resource

```
$ make testacc TESTARGS='-run=TestAccAWSDlmLifecyclePolicy'
TF_ACC=1 go test ./... -v -run=TestAccAWSDlmLifecyclePolicy -timeout 120m
=== RUN   TestAccAWSDlmLifecyclePolicyBasic
--- PASS: TestAccAWSDlmLifecyclePolicyBasic (27.65s)
=== RUN   TestAccAWSDlmLifecyclePolicyFull
--- PASS: TestAccAWSDlmLifecyclePolicyFull (45.55s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	73.213s
...
```

This PR is currently WIP for a few reasons:
1. The AWS SDK appears to be failing to parse the timestamp for the created/modified time so falls over in a big heap. I've raised an [issue](https://github.com/aws/aws-sdk-go/issues/2100) against the SDK in case there's a way to work around it that I'm unaware of but in the meantime I locally patched the SDK in the vendor folder (not pushed) with the diff mentioned in the aforementioned issue.
2. The resource is failing to read state back from AWS properly for some reason so changes made outside of Terraform don't end up being refreshed back into the state. This is immediately evident with the commented out test for the `policy_details.0.schedule.0.create_rule.0.times` attribute in the basic test as AWS provides a value if you omit it but this doesn't end up back in the state currently. I'm not sure what I'm doing wrong here because there's nothing in the logs and no panics when refreshing state so would love a pointer here. The heavily nested structure of the API is much more complicated than other resources I've worked on so I guess I've ballsed something up here.
3. Need to add docs for the resource.
4. There's a couple of outstanding TODOs on the resource, mostly around validation but also for creating the default service role if it doesn't exist and it's not specified. This could maybe be done in a separate PR but would be a nice thing to do here.

I'll rebase the PR into a single commit after

### DIFF
--- a/aws/resource_aws_dlm_lifecycle_policy.go
+++ b/aws/resource_aws_dlm_lifecycle_policy.go
@@ -1,0 +1,344 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/dlm"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsDlmLifecyclePolicy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsDlmLifecyclePolicyCreate,
+		Read:   resourceAwsDlmLifecyclePolicyRead,
+		Update: resourceAwsDlmLifecyclePolicyUpdate,
+		Delete: resourceAwsDlmLifecyclePolicyDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"description": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile("^[0-9A-Za-z _-]+$"), "see https://docs.aws.amazon.com/cli/latest/reference/dlm/create-lifecycle-policy.html"),
+				//	TODO: https://docs.aws.amazon.com/dlm/latest/APIReference/API_LifecyclePolicy.html#dlm-Type-LifecyclePolicy-Description says it has max length of 500 but doesn't mention the regex but SDK and CLI docs only mention the regex and not max length. Check this
+			},
+			"execution_role_arn": {
+				// TODO: Make this not required and if it's not provided then use the default service role, creating it if necessary
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validateArn,
+			},
+			"policy_details": {
+				Type:     schema.TypeList,
+				Required: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"resource_types": {
+							Type:     schema.TypeList,
+							Required: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+						"schedule": {
+							Type:     schema.TypeList,
+							Required: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"create_rule": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"interval": {
+													Type:     schema.TypeInt,
+													Required: true,
+													ValidateFunc: validateIntegerInSlice([]int{
+														12,
+														24,
+													}),
+												},
+												"interval_unit": {
+													Type:     schema.TypeString,
+													Optional: true,
+													Default:  dlm.IntervalUnitValuesHours,
+													ValidateFunc: validation.StringInSlice([]string{
+														dlm.IntervalUnitValuesHours,
+													}, false),
+												},
+												"times": {
+													Type:     schema.TypeList,
+													Optional: true,
+													Computed: true,
+													MaxItems: 1,
+													Elem: &schema.Schema{
+														Type:         schema.TypeString,
+														ValidateFunc: validation.StringMatch(regexp.MustCompile("^([0-9]|0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$"), "see https://docs.aws.amazon.com/dlm/latest/APIReference/API_CreateRule.html#dlm-Type-CreateRule-Times"),
+													},
+												},
+											},
+										},
+									},
+									"name": {
+										Type:         schema.TypeString,
+										Required:     true,
+										ValidateFunc: validation.StringLenBetween(0, 500),
+									},
+									"retain_rule": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"count": {
+													Type:         schema.TypeInt,
+													Required:     true,
+													ValidateFunc: validation.IntBetween(1, 1000),
+												},
+											},
+										},
+									},
+									"tags_to_add": {
+										Type:     schema.TypeMap,
+										Optional: true,
+									},
+								},
+							},
+						},
+						"target_tags": {
+							Type:     schema.TypeMap,
+							Required: true,
+						},
+					},
+				},
+			},
+			"state": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  dlm.SettablePolicyStateValuesEnabled,
+				ValidateFunc: validation.StringInSlice([]string{
+					dlm.SettablePolicyStateValuesDisabled,
+					dlm.SettablePolicyStateValuesEnabled,
+				}, false),
+			},
+		},
+	}
+}
+
+func resourceAwsDlmLifecyclePolicyCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dlmconn
+
+	input := dlm.CreateLifecyclePolicyInput{
+		Description:      aws.String(d.Get("description").(string)),
+		ExecutionRoleArn: aws.String(d.Get("execution_role_arn").(string)),
+		PolicyDetails:    expandDlmPolicyDetails(d.Get("policy_details").([]interface{})),
+		State:            aws.String(d.Get("state").(string)),
+	}
+
+	log.Printf("[INFO] Creating DLM lifecycle policy: %s", input)
+	out, err := conn.CreateLifecyclePolicy(&input)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(*out.PolicyId)
+
+	return resourceAwsDlmLifecyclePolicyRead(d, meta)
+}
+
+func resourceAwsDlmLifecyclePolicyRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dlmconn
+
+	log.Printf("[INFO] Reading DLM lifecycle policy: %s", d.Id())
+	out, err := conn.GetLifecyclePolicy(&dlm.GetLifecyclePolicyInput{
+		PolicyId: aws.String(d.Id()),
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" && !d.IsNewResource() {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("description", out.Policy.Description)
+	d.Set("execution_role_arn", out.Policy.ExecutionRoleArn)
+	d.Set("state", out.Policy.State)
+	if err := d.Set("policy_details", flattenDlmPolicyDetails(out.Policy.PolicyDetails)); err != nil {
+		return fmt.Errorf("error setting policy details %s", err)
+	}
+
+	return nil
+}
+
+func resourceAwsDlmLifecyclePolicyUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dlmconn
+
+	input := dlm.UpdateLifecyclePolicyInput{
+		PolicyId: aws.String(d.Id()),
+	}
+
+	if d.HasChange("description") {
+		input.Description = aws.String(d.Get("description").(string))
+	}
+	if d.HasChange("execution_role_arn") {
+		input.ExecutionRoleArn = aws.String(d.Get("execution_role_arn").(string))
+	}
+	if d.HasChange("state") {
+		input.State = aws.String(d.Get("state").(string))
+	}
+	if d.HasChange("policy_details") {
+		input.PolicyDetails = expandDlmPolicyDetails(d.Get("policy_details").([]interface{}))
+	}
+
+	log.Printf("[INFO] Updating lifecycle policy %s", d.Id())
+	_, err := conn.UpdateLifecyclePolicy(&input)
+	if err != nil {
+		return err
+	}
+
+	return resourceAwsDlmLifecyclePolicyRead(d, meta)
+}
+
+func resourceAwsDlmLifecyclePolicyDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dlmconn
+
+	log.Printf("[INFO] Deleting DLM lifecycle policy: %s", d.Id())
+	_, err := conn.DeleteLifecyclePolicy(&dlm.DeleteLifecyclePolicyInput{
+		PolicyId: aws.String(d.Id()),
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


bflad:
Sorry I don't have time to give in depth review at the moment but to unblock you (hopefully) you can get a better handle on what's wrong with `d.Set("policy_details", ...)`, if return an error from it. 😄 

```
if err := d.Set("policy_details", flattenDlmPolicyDetails(out.Policy.PolicyDetails)); err != nil {
  return fmt.Errorf("error setting policy_details: %s", err)
}
```

At very quick glance of the flatten functions, here are a few hints:
* You'll need to be returning a `[]map[string]interface{}` not just `map[string]interface{}` to account for the "container" attribute that is a `schema.TypeList`
* You'll need to dereference the API response pointer values (recommended with the SDK functions like `aws.StringValue()`, `aws.Int64Value()`, etc.)
* You'll want to `return []map[string]interface{}{}` if the API response object is `nil` before trying to reference attributes within the object

As an example:

```go
func flattenDlmCreateRule(createRule *dlm.CreateRule) []map[string]interface{} {
 	if createRule == nil {
 	}

 	result := make(map[string]interface{})
 	result["interval"] = aws.Int64Value(createRule.Interval)
 	result["internal_unit"] = aws.StringValue(createRule.IntervalUnit)
 	result["times"] = flattenStringList(createRule.Times)

 	return []map[string]interface{}{result}
}
```

You can also help verify the resource read function by adding import support and using `ImportStateVerify` testing:

```go
// in resource schema definition
},
```

```go
// adding an acceptance testing TestStep
{
	ImportState:       true,
	ImportStateVerify: true,
},
```

tomelliff:
@bflad thanks for the pointers, think I've got it working now.

I need to rerun the tests (not had a chance to run them since adding the import test) and then squash the mess of commits here.

We also still need to wait for AWS to fix the API response as per [the previously linked issue on the AWS SDK](https://github.com/aws/aws-sdk-go/issues/2100) before this can be merged.

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,341 @@
...
+									"create_rule": {
+										Type:     schema.TypeList,
+										Required: true,
+										MaxItems: 1,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"interval": {
+													Type:     schema.TypeInt,
+													Required: true,
+													// TODO: Validate this is 12 or 24
```
tomelliff: Is there a nice integer equivalent for `StringInSlice` validation function? Or do I need to write a small validation function for this? Couldn't see one anywhere but might have missed it.

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,341 @@
...
+													// TODO: Make this optional and default to 'HOURS'. Update docs too.
+												},
+												"times": {
+													Type:     schema.TypeList,
+													Optional: true,
+													Computed: true,
+													MaxItems: 1,
+													Elem:     &schema.Schema{Type: schema.TypeString},
+													// TODO: Validate each element in a list?
+													//ValidateFunc: validation.StringMatch(regexp.MustCompile("^([0-9]|0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$"), "see https://docs.aws.amazon.com/dlm/latest/APIReference/API_CreateRule.html#dlm-Type-CreateRule-Times"),
```
tomelliff: Is there a good way to validate each element in a list here?

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,341 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"description": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile("^[0-9A-Za-z _-]+$"), "see https://docs.aws.amazon.com/cli/latest/reference/dlm/create-lifecycle-policy.html"),
+				//	TODO: https://docs.aws.amazon.com/dlm/latest/APIReference/API_LifecyclePolicy.html#dlm-Type-LifecyclePolicy-Description says it has max length of 500 but doesn't mention the regex but SDK and CLI docs only mention the regex and not max length. Check this
+			},
+			"execution_role_arn": {
+				// TODO: Make this not required and if it's not provided then use the default service role, creating it if necessary
```
tomelliff: Does this want doing in first pass or should it be done later? I'm tempted to do the work for this if AWS take a long time fixing the API response which is blocking this.

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,341 @@
...
+	result := make(map[string]interface{})
+	result["interval"] = aws.Int64Value(createRule.Interval)
+	result["interval_unit"] = aws.StringValue(createRule.IntervalUnit)
+	result["times"] = flattenStringList(createRule.Times)
+
+	return []map[string]interface{}{result}
+}
+
+func expandDlmRetainRule(cfg []interface{}) *dlm.RetainRule {
+	// TODO: This is pretty basic so might be possible to collapse this into the caller instead of creating a separate function for it
```
tomelliff: I'm leanings towards keeping this method (and the equivalent flatten method) and just dropping the TODO here because it keeps to the same pattern elsewhere here and will be necessary if the `RetainRule` adds more fields at a later date.

aws/resource_aws_dlm_lifecycle_policy.go
bflad: Not currently 😅  https://godoc.org/github.com/hashicorp/terraform/helper/validation

For now we can add one to `aws/validators.go` 👍 

```go
// IntInSlice returns a SchemaValidateFunc which tests if the provided value
// is of type int and matches the value of an element in the valid slice
func IntInSlice(validInts []int, ignoreCase bool) schema.SchemaValidateFunc {
	return func(i interface{}, k string) (s []string, es []error) {
		v, ok := i.(int)
		if !ok {
			es = append(es, fmt.Errorf("expected type of %s to be int", k))
			return
		}

		for _, validInt := range validInts {
			if v == validInt {
				return
			}
		}

		es = append(es, fmt.Errorf("expected %s to be one of %v, got %s", k, valid, v))
		return
	}
}
```

ryan-dyer-sp:
Any update on this?

tomelliff:
I'm still waiting for AWS to fix the response of the API. I've not checked again recently but the linked issue is still open so I'm leaving it until it's closed.

After that I'll fix up the remaining TODOs on this PR and then it should be good for review.

tomelliff:
I've been really busy at work lately so not had any chance to come back and look at the review comments until now and no incentive to do so either with the DLM SDK/API response still being broken (chased AWS about it recently but it's still not fixed).

I've just pushed fixes for all the feedback now and think it's good to go once the DLM API response is fixed (which should be announced in the linked issue or I can respond here when my support ticket is finally closed).

I've implemented the `IntInSlice` validation func here but think it probably should go upstream. I'll raise a PR for that over in core and if it gets merged we can remove it from here.

bflad:
@tomelliff let us know if you need anything!

tomelliff:
Had a bit of a monster rebase there, let me know if I screwed anything up.

Acceptance test output after rebase:
```
$ make testacc TESTARGS='-run=TestAccAWSDlmLifecyclePolicy'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSDlmLifecyclePolicy -timeout 120m
=== RUN   TestAccAWSDlmLifecyclePolicyBasic
--- PASS: TestAccAWSDlmLifecyclePolicyBasic (36.54s)
=== RUN   TestAccAWSDlmLifecyclePolicyFull
--- PASS: TestAccAWSDlmLifecyclePolicyFull (52.69s)
PASS
```

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,344 @@
...
+		Delete: resourceAwsDlmLifecyclePolicyDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"description": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile("^[0-9A-Za-z _-]+$"), "see https://docs.aws.amazon.com/cli/latest/reference/dlm/create-lifecycle-policy.html"),
```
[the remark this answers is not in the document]
tomelliff: As mentioned in the TODO, the docs for the API and the CLI/SDK differ. This is also the case with the `policyDetails.schedule.name` which also appears to be regex bound (not that I can even find any docs that suggest this) as well as limited to 500 characters.

I can add the validation for both `description` and `policyDetails.schedule.name` for both length and characters to give people a plan time failure rather than apply time but I don't think I can really link the AWS docs due to the inconsistent/incomplete information unfortunately :(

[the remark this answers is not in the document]
chenrui333: yeah, understood. Is AWS aware of this kind of issue?

aws/resource_aws_dlm_lifecycle_policy.go
tomelliff: Looks like @ewbankkit got around to adding this so I dropped my commit that added it after seeing it get pulled in on the rebase against master.

aws/resource_aws_dlm_lifecycle_policy.go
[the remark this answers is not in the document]
tomelliff: I've mentioned it in my support ticket I had open with them over this and also in the linked AWS issue so hopefully they can sort that out.

goochi1:
has this been merged? cant see it on the site
https://www.terraform.io/docs/providers/aws

aws/resource_aws_dlm_lifecycle_policy.go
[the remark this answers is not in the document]
bflad: We can fix this (if necessary) after initial release since plan-time validation is generally a "nice to have" enhancement.

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,344 @@
...
+
+func resourceAwsDlmLifecyclePolicyRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dlmconn
+
+	log.Printf("[INFO] Reading DLM lifecycle policy: %s", d.Id())
+	out, err := conn.GetLifecyclePolicy(&dlm.GetLifecyclePolicyInput{
+		PolicyId: aws.String(d.Id()),
+	})
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "ResourceNotFoundException" && !d.IsNewResource() {
```
bflad: Two things here:
* We have a helper for handling this type of logic, e.g. `isAWSErr(err, dlm.ErrCodeResourceNotFoundException, "")`
* We should not create a special case during Read after Create that returns an error -- instead if this resource has eventual consistency issues, we should implement a `resource.Retry()` loop that retries for a minute or two (then can use `d.IsNewResource()` to determine if its a retryable error).

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,344 @@
...
+			d.SetId("")
```
bflad: Nit: When triggering resource recreation, we should return a warning message in the logs so operators and code maintainers can more easily diagnose why it occurred, e.g.

```go
log.Printf("[WARN] DLM Lifecycle Policy (%s) not found, removing from state", d.Id())
d.SetId("")
```

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,344 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func expandDlmPolicyDetails(cfg []interface{}) *dlm.PolicyDetails {
+	policyDetails := &dlm.PolicyDetails{}
+	m := cfg[0].(map[string]interface{})
```
bflad: To prevent potential panics, we should length and nil check the slice before accessing its first element, e.g.

```go
if len(cfg) == 0 || cfg[0] == nil {
  return nil
}
m := cfg[0].(map[string]interface{})
```

aws/resource_aws_dlm_lifecycle_policy.go
```diff
@@ -0,0 +1,344 @@
...
+		})
+	}
+
+	return result
+}
+
+func flattenDlmTags(tags []*dlm.Tag) map[string]string {
+	result := make(map[string]string)
+	for _, t := range tags {
+		result[*t.Key] = *t.Value
```
bflad: To prevent potential panics, we should prefer to use the SDK helpers in this case, e.g.

```go
result[aws.StringValue(t.Key)] = aws.StringValue(t.Value)
```

aws/resource_aws_dlm_lifecycle_policy.go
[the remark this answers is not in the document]
chenrui333: 👍

aws/resource_aws_dlm_lifecycle_policy_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+		if out.Policy != nil {
+			return fmt.Errorf("DLM lifecycle policy still exists: %#v", out)
+		}
+	}
+
+	return nil
+}
+
+func checkDlmLifecyclePolicyExists(name string) resource.TestCheckFunc {
```
bflad: This function should implement a similar (but opposite) check to the CheckDestroy, where it verifies that the expected resource does actually exist in the API. Admittedly, this is a little buried in the provider development documentation, here: https://www.terraform.io/docs/extend/testing/acceptance-tests/teststep.html#custom-check-functions

```go
  rs, ok := s.RootModule().Resources[name]

  if !ok {
    return fmt.Errorf("Not found: %s", name)
  }

  if rs.Primary.ID == "" {
    return fmt.Errorf("Resource %s ID not found", name)
  }

  input := dlm.GetLifecyclePolicyInput{
    PolicyId: aws.String(rs.Primary.ID),
  }

  out, err := conn.GetLifecyclePolicy(&input)

  if err != nil {
    return fmt.Errorf("error getting DLM Lifecycle Policy (%s): %s", rs.Primary.ID, err)
  }

   return nil
}
```

aws/resource_aws_dlm_lifecycle_policy_test.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/dlm"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSDlmLifecyclePolicyBasic(t *testing.T) {
+	resourceName := "aws_dlm_lifecycle_policy.basic"
+
+	resource.Test(t, resource.TestCase{
```
bflad: Nit: This was written before `resource.ParallelTest()` was available so we'll update these. 👍

aws/resource_aws_dlm_lifecycle_policy_test.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+
+		return nil
+	}
+}
+
+func dlmLifecyclePolicyBasicConfig() string {
+	return fmt.Sprint(`
+resource "aws_iam_role" "dlm_lifecycle_role" {
+  name = "tf-acc-basic-dlm-lifecycle-role"
```
bflad: Nit: This would've been found with `resource.ParallelTest()` (had it existed back then!), but resource names like these need to be randomized (check out other testing with `rName` variables).

website/docs/r/dlm_lifecycle_policy.markdown
```diff
@@ -0,0 +1,113 @@
+---
+layout: "aws"
+page_title: "AWS: aws_dlm_lifecycle_policy"
+sidebar_current: "docs-aws-resource-dlm-lifecycle-policy"
```
bflad: Merge note: Missing sidebar link in `website/aws.erb`

aws/resource_aws_dlm_lifecycle_policy.go
tomelliff: Ack yeah, I keep missing these. Is there a linting rule or something we can use to check for this?

bflad:
This has been released in [version 1.43.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1430-november-07-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

Issue #558, opened 2017-06-13
provider/aws: Add import support for API Gateway resources


hashibot:
_This issue was originally opened by @Vinnl as hashicorp/terraform#12250. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

(I wasn't sure if I had to create separate issues for all of these - most of them are only useful together with some of the others. If you prefer separate issues, do let me know.)

### Terraform Version
At least up to v0.8.7

### Affected Resource(s)
The following resources are required to set up a proxy integration with an AWS Lambda function for API Gateway, and do not support importability yet:

- aws_api_gateway_resource
- aws_api_gateway_method
- aws_api_gateway_integration
- aws_lambda_permission
- aws_api_gateway_method_response
- aws_api_gateway_integration_response
- aws_api_gateway_deployment

### Expected Behavior
Running the import command for these resources should update the state.

### Actual Behavior
Message `resource <resource> doesn't support import`.

tdmalone:
A few more for the API Gateway import list - would love to see these! =]
* aws_api_gateway_rest_api
* aws_api_gateway_stage
* aws_api_gateway_gateway_response
* aws_api_gateway_domain_name

bflad:
The above mentioned PRs have been merged and released in [version 1.33.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1330-august-22-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

erikpaasonen:
@bflad Could you please also add import support for `aws_lambda_permission`?  @Vinnl mentioned it in the original list of resources, but it seems to have slipped through the cracks. Saves me from opening a brand-new issue. :)

kladiv:
+1 for "aws_api_gateway_deployment" import

vinischeidegger:
Please don't forget the **aws_api_gateway_deployment** import functionality.

natblundell:
+1 for "aws_api_gateway_deployment" import

qingvincentyin:
Re:

- `aws_api_gateway_deployment `

[...]

PR #5572, opened 2018-08-16
resource/aws_api_gateway_model: Support resource import

bflad:
Reference: #558 

* Support, test, and document resource import

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayModel_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayModel_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayModel_basic
--- PASS: TestAccAWSAPIGatewayModel_basic (14.34s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	14.981s
```

### DIFF
--- a/aws/resource_aws_api_gateway_model.go
+++ b/aws/resource_aws_api_gateway_model.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


aws/resource_aws_api_gateway_model.go
```diff
@@ -18,6 +19,33 @@ func resourceAwsApiGatewayModel() *schema.Resource {
...
+				output, err := conn.GetModel(&apigateway.GetModelInput{
+					ModelName: aws.String(name),
+					RestApiId: aws.String(restApiID),
+				})
+
+				if err != nil {
+					return nil, err
+				}
+
+				d.SetId(aws.StringValue(output.Id))
```
tombuildsstuff: why not call d.SetId and then the Read function? 🤔

bflad: In this case, the model ID is actually worthless for read. At some point we should switch the resource ID to be the REST API ID and model name instead of the model ID, but that could be perceived as a breaking change.

bflad:

PR #5573, opened 2018-08-16
resource/aws_api_gateway_request_validator: Support resource import

bflad:
Reference: #558 

* Support, test, and document resource import

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSAPIGatewayRequestValidator_'
TF_ACC=1 go test ./aws -v -run=TestAccAWSAPIGatewayRequestValidator_ -timeout 120m
=== RUN   TestAccAWSAPIGatewayRequestValidator_basic
--- PASS: TestAccAWSAPIGatewayRequestValidator_basic (20.93s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	21.568s
```

### DIFF
--- a/aws/resource_aws_api_gateway_request_validator.go
+++ b/aws/resource_aws_api_gateway_request_validator.go
@@ -2,12 +2,13 @@ package aws
 
 import (
 	"fmt"
+	"log"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/apigateway"
 	"github.com/hashicorp/terraform/helper/schema"
-	"log"
-	"strings"
 )
 
 func resourceAwsApiGatewayRequestValidator() *schema.Resource {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_api_gateway_request_validator.go
```diff
@@ -16,6 +17,20 @@ func resourceAwsApiGatewayRequestValidator() *schema.Resource {
...
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				idParts := strings.Split(d.Id(), "/")
+				if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+					return nil, fmt.Errorf("Unexpected format of ID (%q), expected REST-API-ID/REQUEST-VALIDATOR-ID", d.Id())
+				}
+				restApiID := idParts[0]
+				requestValidatorID := idParts[1]
+				d.Set("request_validator_id", requestValidatorID)
+				d.Set("rest_api_id", restApiID)
+				d.SetId(requestValidatorID)
```
tombuildsstuff: why not just call d.SetID and then the Read function? 🤔

bflad: In this case, the read function needs the REST API ID as well as the request validator ID to perform the read, but the existing resource ID is missing the REST API ID. At some point we should adjust the resource ID to include the REST API ID so we can perform reads correctly from ID, but for now we can support import without adjusting other code.

bflad:

Issue #4773, opened 2018-06-07
Importing aws_cloudfront_distribution uses the deprecated cache_behavior instead of ordered_cache_behavior


LHCGreg:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
Terraform v0.11.7
+ provider.aws v1.22.0
```

### Affected Resource(s)

* aws_cloudfront_distribution

### Expected Behavior

Running terraform import on a cloudfront distribution with one default cache behavior and one non-default cache behavior should result in the state file having

```
"ordered_cache_behavior.#": "1",
"ordered_cache_behavior.0.allowed_methods.#": "2",
...
```

and

```
"default_cache_behavior.#": "1",
"default_cache_behavior.2334655761.allowed_methods.#": "2",
...
```

### Actual Behavior

Running terraform import on a cloudfront distribution with one default cache behavior and one non-default cache behavior resulted in the state file having

```
"cache_behavior.#": "1",
"cache_behavior.2295505492.allowed_methods.#": "2",
...
```

and

```
"default_cache_behavior.#": "1",
"default_cache_behavior.2334655761.allowed_methods.#": "2",
...
```

Note that the import used "cache_behavior" for the non-default cache behavior instead of ordered_cache_behavior. https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html says cache_behavior is "Deprecated, use ordered_cache_behavior instead".

### Steps to Reproduce

1. Create a cloudfront distribution in AWS with one non-default cache behavior in addition to the default cache behavior.
2. Write the terraform configuration for it, using ordered_cache_behavior for the non-default cache behavior.
3. Use `terraform import` to import it
4. Run a plan. Terraform wants to remove the non-default cache behavior and add it back.

### Workaround

Do the import and then manually edit the state file, changing "cache_behavior.#" to "ordered_cache_behavior.#" and cache_behavior.XXXXXXXXXX.* to ordered_cache_behavior.0.*. If you have more than one non-default cache behavior, you can order them appropriately.

bflad:
Bug fix pull request submitted: #5586

[...]

PR #5586, opened 2018-08-17
resource/aws_cloudfront_distribution: Import into ordered_cache_behavior instead of deprecated cache_behavior

bflad:
Fixes: #5580 
Fixes: #4773

Previously:

```
--- FAIL: TestAccAWSCloudFrontDistribution_orderedCacheBehavior (739.56s)
	testing.go:527: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

		(map[string]string) (len=59) {
		 (string) (len=16) "cache_behavior.#": (string) (len=1) "2",
...
		 (string) (len=48) "cache_behavior.1203603591.viewer_protocol_policy": (string) (len=9) "allow-all"
		}

		(map[string]string) (len=59) {
		 (string) (len=24) "ordered_cache_behavior.#": (string) (len=1) "2",
...
		 (string) (len=47) "ordered_cache_behavior.1.viewer_protocol_policy": (string) (len=9) "allow-all"
		}
```

* Use `d.GetOk()` against the old attribute during read instead of the new attribute, which defaults imports to the new attribute

```
--- PASS: TestAccAWSCloudFrontDistribution_orderedCacheBehavior (820.16s)
```

### DIFF
--- a/aws/cloudfront_distribution_configuration_structure.go
+++ b/aws/cloudfront_distribution_configuration_structure.go
@@ -144,10 +144,10 @@ func flattenDistributionConfig(d *schema.ResourceData, distributionConfig *cloud
 		}
 	}
 	if distributionConfig.CacheBehaviors != nil {
-		if _, ok := d.GetOk("ordered_cache_behavior"); ok {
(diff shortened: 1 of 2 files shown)


bflad:

PR #5602, opened 2018-08-18
d/Adds a url_suffix attribute to data_source_aws_partition

### LINKED ISSUE #1960: AWS missing "pseudo" parameters
Terraform v0.10.7

data_source_aws_partition.go implements the aws "partition" pseudo parameters. However, other pseudo parameters are missing. For example, URLSuffix - which is correlated to the partition - is not available within terraform.

For details of this and other pseudo parameters, see the [CloudFormation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-urlsuffix) documentation

lorengordon:
Fixes #1960

* Adds a `urlsuffix` attribute to the `AWSClient` struct
* Infers the urlsuffix value from the STS endpoint associated with the provider region
* Uses the new attribute to add a `url_suffix` attribute to `data_source_aws_partition`

```
$ make testacc TESTARGS='-run=TestAccAWSPartition_basic'
TF_ACC=1 go test ./... -v -run=TestAccAWSPartition_basic -timeout 120m
=== RUN   TestAccAWSPartition_basic
--- PASS: TestAccAWSPartition_basic (4.34s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       4.379s
```

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -196,6 +196,7 @@ type AWSClient struct {
 	redshiftconn          *redshift.Redshift
 	r53conn               *route53.Route53
 	partition             string
+	urlsuffix             string
 	accountid             string
 	supportedplatforms    []string
 	region                string
@@ -458,6 +459,14 @@ func (c *Config) Client() (interface{}, error) {
 		}
 	}
 
+	// Infer URL Suffix from STS Endpoint
+	resolver := endpoints.DefaultResolver()
+	endpoint, err := resolver.EndpointFor(endpoints.StatesServiceID, client.region)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to resolve endpoint. Errors: %s", err)
+	}
+	client.urlsuffix = parseURLSuffixFromEndpoint(endpoint.URL, client.region)
+
 	client.ec2conn = ec2.New(awsEc2Sess)
 
 	if !c.SkipGetEC2Platforms {
@@ -608,6 +617,18 @@ func (c *Config) Client() (interface{}, error) {
 	})
 
 	return &client, nil
+}
+
+func parseURLSuffixFromEndpoint(endpoint string, region string) string {
+	e := strings.Split(endpoint, ".")
+
+	// An endpoint may have the region in the second token
+	if e[1] == region {
+		return strings.Join(e[2:], ".")
+	}
+
+	// Without a region, the URL Suffix is everything past the first token
+	return strings.Join(e[1:], ".")
 }
 
 func hasEc2Classic(platforms []string) bool {
--- a/aws/config.go
+++ b/aws/config.go
@@ -465,7 +465,13 @@ func (c *Config) Client() (interface{}, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Failed to resolve endpoint. Errors: %s", err)
 	}
-	client.urlsuffix = parseURLSuffixFromEndpoint(endpoint.URL, client.region)
+
+	urlsuffix, err := parseURLSuffixFromEndpoint(endpoint.URL, client.region)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to resolve URL Suffix. Errors: %s", err)
+	}
+
+	client.urlsuffix = aws.StringValue(urlsuffix)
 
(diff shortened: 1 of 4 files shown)


aws/config.go
```diff
@@ -196,6 +196,7 @@ type AWSClient struct {
 	redshiftconn          *redshift.Redshift
 	r53conn               *route53.Route53
 	partition             string
+	urlsuffix             string
```
bflad: Can you please name this `dnsSuffix`? 

```suggestion
	dnsSuffix             string
```

aws/data_source_aws_partition.go
```diff
@@ -16,6 +16,10 @@ func dataSourceAwsPartition() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"url_suffix": {
```
bflad: Can you please name this `dns_suffix`? We will likely be introducing this as an argument to the provider configuration to override all endpoints and the "URL" naming could be potentially confusing to operators since URLs can contain HTTP-like paths as a suffix.

```suggestion
			"dns_suffix": {
```

website/docs/d/partition.html.markdown
```diff
@@ -37,3 +38,4 @@ There are no arguments available for this data source.
 ## Attributes Reference
 
 `partition` is set to the identifier of the current partition.
+`url_suffix` is set to the URL Suffix associated with the current partition.
```
bflad: Can you please add Markdown list syntax for these attributes and clarify what this attribute contains? 👍 

```suggestion
* `dns_suffix` is set to the base DNS domain name for the current partition (e.g. `amazonaws.com` in AWS Commercial, `amazonaws.com.cn` in AWS China).
```

aws/config.go
```diff
@@ -458,6 +459,20 @@ func (c *Config) Client() (interface{}, error) {
 		}
 	}
 
+	// Infer URL Suffix from STS Endpoint
+	resolver := endpoints.DefaultResolver()
+	endpoint, err := resolver.EndpointFor(endpoints.StatesServiceID, client.region)
```
bflad: The comment notes inferring this information from the STS endpoint, however its using the states service.

That said, the AWS Go SDK documentation lists the `XXXServiceID` constants as deprecated: https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants

> Deprecated: Use client package's EndpointsID value instead of these ServiceIDs. These IDs are not maintained, and are out of date.

This should be swapped out according to that deprecation.

Ideally though, we should fetch this information directly from the AWS Go SDK endpoints information first, then only fallback to parsing an endpoint should that fail since endpoint hostnames may not always follow a consistent implementation. I have submitted an upstream feature request in the AWS to directly provide the DNS suffix information, which can help prevent us from hardcoding these cases in the future: https://github.com/aws/aws-sdk-go/issues/2710

aws/config.go
lorengordon: done

aws/data_source_aws_partition.go
lorengordon: done

website/docs/d/partition.html.markdown
lorengordon: done

aws/config.go
lorengordon: This one I'm not sure when I'll have time to research and figure out the alternative. If anyone wants to take it, please be my guest!

bflad:
Thanks so much for your patience, @lorengordon -- I was able to get the DNS suffix information visible from the AWS Go SDK (), rebase this PR, default to using that new information, and add verifying acceptance testing when setting the region to AWS Commercial, GovCloud (US), and China 👍 

```
--- PASS: TestAccAWSPartition_basic (6.97s)

--- PASS: TestAccAWSProvider_Region_AwsCommercial (2.88s)
--- PASS: TestAccAWSProvider_Region_AwsGovCloudUs (2.93s)
--- PASS: TestAccAWSProvider_Region_AwsChina (2.98s)
```

lorengordon:
Brilliant! Thanks @bflad!

Issue #5599, opened 2018-08-17
aws_dynamodb_table is not respecting the timeout that I define or even the default timeout 


jeffultra:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version
Terraform v0.11.8
+ provider.aws v1.32.0

### Affected Resource(s)

* aws_dynamodb_table

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```
locals {
  common_tags = {
    Company = "abc"
    Segment = "abc-geral"
    Service = "abc-bla"
    Area    = "abc-bla-bla"
  }

  system_tags = {
    Owner       = "mine"
    Environment = "${var.environment_name}"
    System      = "xpto"
    Brand       = "${var.brand}"
    ManagedBy   = "terraform"
  }
}

resource "aws_dynamodb_table" "mda-table" {
  name           = "MDA"
  read_capacity  = "${var.mda_read_capacity}"
  write_capacity = "${var.mda_write_capacity}"
  hash_key       = "mda"

  attribute {
    name = "mda"
    type = "S"
  }

  attribute {
    name = "mda"
    type = "S"
  }

  timeouts {
    create = "1h"
    update = "1h"
    delete = "1h"
  }

  tags = "${merge(
    local.common_tags,
    local.system_tags,
    map(
      "System", "xpto-mda",
      "Type", "dynamodb",
    )
  )}"
}

### Debug Output
https://gist.github.com/jeffultra/f507b3396c7c57ea9349a86f69b88b42

### Expected Behavior

Terraform should have created my DynamoDB tables respecting the timeout defined instead of giving up the creation of them just after 2 minutes.

### Actual Behavior

Terraform complained that the tables took more than 2 minutes to be created and it stopped the creation.

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

bflad:
Hi @jeffultra 👋 

[...]

PR #5614, opened 2018-08-20
provider: Disable AWS SDK retries faster by default for connection refused errors

bflad:
When using `resource.Retry()`, this error is sometimes never displayed except for the Terraform debug logs due to the AWS SDK never returning the response before the `resource.Retry()` timeout:

```
* aws_dynamodb_table.mda-table: timeout while waiting for state to become 'success' (timeout: 2m0s)
```

This type of error is more permanent in nature and should be returned quicker than the default MaxRetries configuration of the provider.

Reference: #5599 

* Disable `Retryable` for `connection refused` errors sooner if using default `MaxRetries` configuration


bflad:

PR #5615, opened 2018-08-20
Do not delete version history on s3 object delete

jocgir:
On versioned bucket, we should not delete the whole history of a file if the file is renamed/deleted.

We added an optional parameter to enforce bucket full object deletion if desired, but that should not be the default behaviour.

We modified our terraform code to change the key name of an S3 object. We then needed to retrieve the previous version of the file and expecting to see a delete marker but we realized that all previous versions of the file were gone.

We feel that destructive operation such as deleting all versions of an object should not be the default behaviour and the terraform S3 object delete should behave as manual S3 object changes done through the UI or the API as it does on S3 object update.


ewbankkit:
Related:
* https://github.com/terraform-providers/terraform-provider-aws/issues/4500

PR #5616, opened 2018-08-20
Update S3 object on mismatched etag

jocgir:
If the saved etag from state file does not match the actual etag on the S3 object, that means that the resource has been modified outside of terraform, so we force the update of the resource.

With encrypted object, we cannot use the etag parameter and compute a local md5 to force the update of the resource since the actual etag will be different from our locally computed md5. However, if the saved etag stored in the state file is not the same as the actual etag, that probably signify that the stored resource has been manually altered outside of terraform. In that case, we should force the update.

We tested our implementation and it works very well to detect manual change of the S3 object. However, if you have better suggestion than adding a dummy computed attribute such as `etag_changed`, your comment are welcome.


bflad:
Hi @jocgir 👋 Thank you for submitting this.

Generally we discourage the usage of `d.Get()` in Terraform resource `Read` functions since it goes against the typical Terraform resource flow and may not work as expected in certain scenarios (e.g. import).

Our recommendation for fixing this issue would likely be an implementation of #6668, which would introduce a secondary hash/checksum argument to the resource, where operators can optionally configure its value to appropriately trigger updates, rather than depending on the potentially unsatisfying `etag` argument.

Since we would prefer an implementation of #6668 and there has not been any community upvotes of this particular implementation, I'm going to close it, but please do not hesitate to contribute in the future! Thanks again.

PR #5641, opened 2018-08-22
provider: Add gometalinter to Make targets and additionally enable errcheck, interfacer, and structcheck linting for TravisCI

bflad:
Keeps backwards compatibility with existing Make targets. We can enable others in the future, but they currently are not passing. 😉 

* Add `tools` Make target for installing testing dependencies, similar to upstream and other HashiCorp projects
* Add `lint` Make target for running all source code linters
* Add `website-lint` Make target for future running of website linters
* Initial `.gometalinter.json` configuration

```
$ make lint
==> Checking source code against linters...
$ 
```


bflad:
Followup linting issues created:
* `goimports`: #5642 
* `goimports`: #5642 
* `goimports`: #5642

.gometalinter.json
```diff
@@ -0,0 +1,20 @@
+{
+    "Deadline": "5m",
+    "Enable": [
+        "errcheck",
+        "gofmt",
```
paultyng: May want to also include `goimports`? Not sure, I think it covers slightly different things from `gofmt`, and goes a little further, but I don't know if it does the simplifications as well.

.gometalinter.json
```diff
@@ -0,0 +1,20 @@
...
+        "interfacer",
+        "structcheck",
+        "vet"
+    ],
+    "EnableGC": true,
+    "Linters": {
+        "errcheck": {
+            "Command": "errcheck -abspath {not_tests=-ignoretests} -ignore github.com/hashicorp/terraform/helper/schema:ForceNew|Set -ignore io:Close"
```
paultyng: May be worth splitting errcheck ignores to a separate file if the list may get longer.

.gometalinter.json
paultyng: No golint?

paultyng: FWIW here is the config I was playing with on archive: https://github.com/terraform-providers/terraform-provider-archive/compare/f/makelint#diff-df70a2d0f734c21a1ebe861e6e198c2fR25

You can see some of the excludes I added, like ignoring missing errcheck in `defer` calls, etc.

.gometalinter.json
```diff
@@ -0,0 +1,20 @@
...
+        }
+    },
+    "Sort": [
+        "path",
+        "line"
+    ]
```
paultyng: I think you also want:
```json
{
  "Vendor": true,
  "WarnUnmatchedDirective": true
}
```

GNUmakefile
```diff
@@ -19,23 +19,23 @@ test: fmtcheck
...
-		exit 1; \
-	fi
-
 fmt:
-	gofmt -w $(GOFMT_FILES)
+	@echo "==> Fixing source code with gofmt..."
+	gofmt -s -w ./$(PKG_NAME)
 
+# Currently required by tf-deploy compile
 fmtcheck:
```
paultyng: You could probably remove this as its checked by metalinter now.

GNUmakefile
```diff
@@ -19,23 +19,23 @@ test: fmtcheck
...
-	@sh -c "'$(CURDIR)/scripts/gofmtcheck.sh'"
+	@echo "==> Checking source code against gofmt..."
+	@gofmt -l -s ./$(PKG_NAME)
```
paultyng: I think the reason this was in a script though was that the `gofmt` exit code doesn't actually fail if its not formatted.

.gometalinter.json
bflad: I'm not including the others because they are broken at the moment and will be added later: https://github.com/terraform-providers/terraform-provider-aws/pull/5641#issuecomment-415016000

`golint` especially is a nightmare since it wants to massively fix capitalization (e.g. `Id` -> `ID`) and remove underscores for acceptance tests for a bunch of names.

.gometalinter.json
bflad: Agreed. 👍  Let me know if you want that done now (the extra file seemed like too much overhead for this currently small list).

.gometalinter.json
bflad: Yes we do! Adding. 👍

GNUmakefile
bflad: See my Slack comments from earlier today. This Make target is required by our release process.

GNUmakefile
bflad: You're correct -- fixed by reinstating the old script for now.

```
$ gofmt -l -s ./aws
aws/resource_aws_api_gateway_domain_name.go
$ echo $?
0
```

GNUmakefile
```diff
@@ -25,8 +25,7 @@ fmt:
 
 # Currently required by tf-deploy compile
 fmtcheck:
-	@echo "==> Checking source code against gofmt..."
-	@gofmt -l -s ./$(PKG_NAME)
+	@sh -c "'$(CURDIR)/scripts/gofmtcheck.sh'"
```
paultyng: I thought about this a bit more, and I think you should just do:

```Makefile
fmtcheck: lint
```

As that was probably the intent of the fmt check.

bflad: @paultyng what do you think?

PR #5663, opened 2018-08-23
Add missing docs for aws instance

ken5scal:
Add documentation related to  #1980

* In the above pr, aws_instance resource supports a new parameter `instance_state`, but the documentation seems not to be updated.

n/a


bflad:
Hi @ken5scal 👋 The referenced PR was never merged. To confirm, you can see that `instance_state` is currently only a [`Computed` attribute](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed):

https://github.com/terraform-providers/terraform-provider-aws/blob/17780a8d48809cd5f72bbd9e58097499ca015b00/aws/resource_aws_instance.go#L230-L233

It does look like the documentation is missing under the attributes section though if you would like to add it there.

ken5scal:
oh, ic. Thanks for checking it out.

bflad:
Closing this as its invalid but feel free to resubmit `instance_state` under attributes if you wish. 👍

PR #5667, opened 2018-08-23
S3 bucket api gateway integration (Issue #5652)

### LINKED ISSUE #5652: Add example S3 bucket integration for api_gateway_integration
### Community Note

### Description

See this article on integrating S3 with API Gateway: https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html

It would be nice to have an example replicating this in Terraform.

hasheddan:
Fixes #5652 

Example for creating an AWS API Gateway with S3 integration. Follows [this article](https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html), but only builds to the point of listing buckets belonging to caller. Article goes on to allow GET, PUT, and DELETE of S3 resources, but this example demonstrates all functionality needed to move forward with the article.

### DIFF
--- a/examples/s3-api-gateway-integration/README.md
+++ b/examples/s3-api-gateway-integration/README.md
@@ -1,1 +1,16 @@
 # S3 Bucket Integration for API Gateway
+
+This example demonstrates how to create an S3 Proxy using AWS API Gateway. It takes you through listing the buckets of the API caller, but provides an example of all of the resources needed to extend the API to manipulating bucket contents. It follows [this article](https://docs.aws.amazon.com/apigateway/latest/developerguide/integrating-api-with-aws-services-s3.html) on AWS.
+
+## Running this Example
+
+*Note: In order to see the API Gateway that this configuration creates you must navigate to the correct region in your AWS console. To ensure the API works as desired, the region you use to create the API should be different than the S3 buckets being queried. If not, you may encounter a 500 Internal Server Error response. This limitation does not apply to any deployed API.*
+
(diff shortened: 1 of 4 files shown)


HashedDan:
@catsby so the diff does not occur if the `selection_pattern = "-"` argument is removed from the 200IntegrationResponse stanza, and the 200 response is still set to the default selection pattern automatically. It appears as though the diff that occurs when explicitly defining the selection_pattern as default is the result of a bug. I would be happy to look into it! Would you prefer that I go ahead and remove the 'selection_pattern' parameter for this example guide, or fix the error first and then add this example with the 'selection_pattern'?

HashedDan:
@catsby just wanted to follow up on this older PR and see if there is still interest. Thanks!

ryndaniels:
@hasheddan thanks for this and sorry for the delay! A couple comments - 
- For the `selection_pattern`, I think it would make sense to remove it from the example for now, and we can address the error as a separate issue
- When I ran this, the initial plan and apply gave me some 404 errors (`Error creating API Gateway Integration Response: NotFoundException: No integration defined for method`) on the integration responses, which I was able to fix by adding `depends_on = ["aws_api_gateway_integration.S3Integration"]` to the response resources. 

If you have the time to still work on this and make those changes we can go ahead and get this merged, otherwise we could pick it up on our end. Thanks!

hasheddan:
@catsby thanks for following up! I will make sure to get these changes completed ASAP!

catsby:
👋 @hasheddan you probably meant to thank @ryndaniels 😆 

Thank you @ryndaniels for picking up on something I dropped the ball on 😦

ryndaniels:
@catsby of course, happy to help out! 😄

hasheddan:
Oh yes! Sorry about that @catsby! Hope you are doing well!

ryndaniels:
This looks great @hasheddan, thanks!

PR #5680, opened 2018-08-26
Specify to use security group id in aws instance.

uLan08:
I am a total beginner in AWS and Terraform and so as I was playing around I encountered an error when supplying the `security_groups` field under `aws_instance`. I initially used `["${aws_security_group.terraform.name}"]` and this is the error
```
* aws_instance.example: Error launching source instance: InvalidGroup.NotFound: The security group 'terraform_example' does not exist in VPC 'vpc-id'
```
 Changing it to `["${aws_security_group.terraform.id}"]` fixed it.

The changes in this PR will specify in the docs to use the security group ID.

```
$ terraform -v
Terraform v0.11.7
+ provider.aws v1.31.0
```

### DIFF
--- a/website/docs/r/instance.html.markdown
+++ b/website/docs/r/instance.html.markdown
@@ -77,7 +77,7 @@ instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/Use
 
 * `get_password_data` - (Optional) If true, wait for password data to become available and retrieve it. Useful for getting the administrator password for instances running Microsoft Windows. The password data is exported to the `password_data` attribute. See [GetPasswordData](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetPasswordData.html) for more information.
 * `monitoring` - (Optional) If true, the launched EC2 instance will have detailed monitoring enabled. (Available since v0.6.0)
-* `security_groups` - (Optional, EC2-Classic and default VPC only) A list of security group names to associate with.
+* `security_groups` - (Optional, EC2-Classic and default VPC only) A list of security group names or IDs to associate with.
 
 -> **NOTE:** If you are creating Instances in a VPC, use `vpc_security_group_ids` instead.
 
--- a/website/docs/r/instance.html.markdown
+++ b/website/docs/r/instance.html.markdown
@@ -77,7 +77,7 @@ instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/Use
 
 * `get_password_data` - (Optional) If true, wait for password data to become available and retrieve it. Useful for getting the administrator password for instances running Microsoft Windows. The password data is exported to the `password_data` attribute. See [GetPasswordData](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetPasswordData.html) for more information.
 * `monitoring` - (Optional) If true, the launched EC2 instance will have detailed monitoring enabled. (Available since v0.6.0)
-* `security_groups` - (Optional, EC2-Classic and default VPC only) A list of security group names or IDs to associate with.
+* `security_groups` - (Optional, EC2-Classic and default VPC only) A list of security group names (EC2-Classic) or IDs (default VPC) to associate with.
 
 -> **NOTE:** If you are creating Instances in a VPC, use `vpc_security_group_ids` instead.
 


bflad:
@uLan08 did you intend to change the documentation for the [`aws_instance` resource](https://www.terraform.io/docs/providers/aws/r/instance.html) page instead of the [`aws_instance` data source](https://www.terraform.io/docs/providers/aws/d/instance.html) page? The first is intended for managing an EC2 instance and the second is intended for just reading information about an EC2 instance.

In the existing [`security_groups` attribute](https://www.terraform.io/docs/providers/aws/r/instance.html#security_groups) documentation for the resource we currently note:
> security_groups - (Optional, EC2-Classic and default VPC only) A list of security group names to associate with.

As far as I know, names are required with EC2-Classic, but it might be good to note IDs can be specified with the default VPC case if that is valid. Can you please update the other page with that instead? https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/instance.html.markdown

PS: I would also double check after creating your EC2 instance that running `terraform plan` does not want to instead have the security group IDs in `vpc_security_group_ids`. The EC2 API is pretty strange with its use of the two similar attributes.

uLan08:
You're totally right. The `aws_instance` data source page was the one I was reading the whole time but it should have been the `aws_instance` resource page. I guess what brought me there is that when I initially searched for "instance" the one under data sources came up first.  

Anyway, should I undo my previous commit? Or should I just create another PR? 

Also I don't really understand what you mean in your PS. Do you mean the `vpc_security_group_ids` field only accepts IDs?

bflad:
> You're totally right. The aws_instance data source page was the one I was reading the whole time but it should have been the aws_instance resource page. I guess what brought me there is that when I initially searched for "instance" the one under data sources came up first.

We have been debating splitting out the documentation into separate "data source" and "resource" sections. I'll note this ticket to help prioritize that effort. 👍 

> Anyway, should I undo my previous commit? Or should I just create another PR?

That's up to you. 😄  If you want to make the appropriate change and `git rebase` on top of this commit, go for it and just let us know when the pull request is updated. Otherwise feel free to close this pull request and create another.

> Also I don't really understand what you mean in your PS. Do you mean the vpc_security_group_ids field only accepts IDs?

Sorry, I meant to just double check that `terraform plan` does not want to show a change to your `aws_instance` resource after running `terraform apply`.

uLan08:
Yeah, splitting them out would probably be better. Also I suggest to make the tree collapsable, it's hard to see other options when one has to scroll down past the other ones.

I have updated the PR btw. Please tell me if you have anything else to add.


>Sorry, I meant that you should just double check that terraform plan does not want to change to your aws_instance resource after running terraform apply.

So you're saying that running `terraform plan` should show potential errors when running `terraform_apply`? Or do you just mean that as an advice I should just always do `terraform_plan` first before `terraform_apply`?

website/docs/r/instance.html.markdown
```diff
@@ -77,7 +77,7 @@ instances. See [Shutdown Behavior](https://docs.aws.amazon.com/AWSEC2/latest/Use
 
 * `get_password_data` - (Optional) If true, wait for password data to become available and retrieve it. Useful for getting the administrator password for instances running Microsoft Windows. The password data is exported to the `password_data` attribute. See [GetPasswordData](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_GetPasswordData.html) for more information.
 * `monitoring` - (Optional) If true, the launched EC2 instance will have detailed monitoring enabled. (Available since v0.6.0)
-* `security_groups` - (Optional, EC2-Classic and default VPC only) A list of security group names to associate with.
+* `security_groups` - (Optional, EC2-Classic and default VPC only) A list of security group names or IDs to associate with.
```
bflad: I think we should specifically call out when to use each 👍  `A list of security group names (EC2-Classic) or IDs (default VPC) to associate with.`

uLan08: Just a quick question, what does `default VPC` mean? Are these the VPC's that are already created?

bflad: Information about default VPCs can be found at: https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html

> If you created your AWS account after 2013-12-04, it supports only EC2-VPC. In this case, you have a default VPC in each AWS Region. A default VPC is ready for you to use so that you don't have to create and configure your own VPC.

In general, yes, they will already exist in newer accounts, but they can also be deleted (usually for security reasons).

uLan08: Ah. It's just that I actually created another VPC so I am confused if that still classifies as a "default VPC".

bflad: Its only possible to have one default VPC per region. If you have created another VPC its likely that it is not the default VPC for the region. You can verify this in the AWS web console by going to Services > VPC > Your VPCs and looking at the `Default VPC` column.

uLan08: Ah yes. It is not classified as default. I guess what I am leading to is that the VPC I created is not default but using it's ID still worked.

bflad: If you run `terraform plan` again, does it report any changes that need to happen? If not, that's actually new behavior.

uLan08: Hmm actually yes it does. It says that it needs to recreate my aws instance. 
```
-/+ aws_instance.example (new resource required)
      id:                           "i-0344ac8a8f27efcc2" => <computed> (forces new resource)
```

I actually have a [repo for this](https://github.com/uLan08/play-terraform) if you want to take a look.

uLan08: From the looks of it, what I am doing is wrong and maybe I should have just used the default VPC. I would appreciate it if you could explain why though. Nevertheless, I updated the docs as per your request.

bflad: > From the looks of it, what I am doing is wrong and maybe I should have just used the default VPC. I would appreciate it if you could explain why though.

The AWS provider is not intended to be opinionated about how you structure your AWS infrastructure, we just follow the API implementation provided by AWS. The code maintainers here will not tell you whether using a default or custom VPC is a better solution for whatever problem you are trying to solve as issues in this GitHub repository are intended only for bugs or feature requests with Terraform's implementation with AWS. For general questions about using AWS, I would reach out to the other [Terraform community forums](https://www.terraform.io/docs/extend/community/index.html) or the [AWS forums](https://forums.aws.amazon.com/).

uLan08: Alright Thanks!

PR #5709, opened 2018-08-28
vendor: blang/semver@v3.5.1

bflad:
Point to same tag as revision.

* `govendor fetch github.com/blang/semver/...@v3.5.1`


bflad:
Reference: #5773

PR #5728, opened 2018-08-30
r/aws_db_parameter_group, r/aws_rds_cluster_parameter_group: Support resetting parameters

stefansundin:
If you create a parameter group like so:

```hcl
resource "aws_rds_cluster_parameter_group" "cluster" {
  name        = "test-cluster"
  family      = "aurora-mysql5.7"

  parameter {
    name         = "binlog_format"
    value        = "MIXED"
    apply_method = "pending-reboot"
  }
}
```

And then uncomment the parameter block and apply again, it doesn't actually do anything. This PR fixes that.

It is sending the parameter value in the reset request, but this doesn't seem to be causing any problems. I inspected a request in the RDS web console, and it also sends the value (and a lot of other stuff).

I tried setting 22 parameters and then resetting them all in one request, and it worked. I was able to re-produce the error of setting more than 20 parameters at the same time.. Is there anywhere else that this limit is documented?

Let me know what you think!! :)

(Relates: #661)

### DIFF
--- a/aws/resource_aws_db_parameter_group.go
+++ b/aws/resource_aws_db_parameter_group.go
@@ -295,6 +295,26 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
 			}
 			d.SetPartial("parameter")
 		}
+
+		// Reset parameters that have been removed
+		parameters, err = expandParameters(os.Difference(ns).List())
+		if err != nil {
+			return err
+		}
+		if len(parameters) > 0 {
+			parameterGroupName := d.Get("name").(string)
+			resetOpts := rds.ResetDBParameterGroupInput{
+				DBParameterGroupName: aws.String(parameterGroupName),
+				Parameters:           parameters,
+			}
+
+			log.Printf("[DEBUG] Reset DB Parameter Group: %s", resetOpts)
+			_, err = rdsconn.ResetDBParameterGroup(&resetOpts)
+			if err != nil {
+				return fmt.Errorf("Error resetting DB Parameter Group: %s", err)
+			}
+			d.SetPartial("parameter")
+		}
 	}
 
 	if err := setTagsRDS(rdsconn, d, d.Get("arn").(string)); err != nil {
--- a/aws/resource_aws_db_parameter_group_test.go
+++ b/aws/resource_aws_db_parameter_group_test.go
@@ -291,6 +291,15 @@ func TestAccAWSDBParameterGroup_basic(t *testing.T) {
 						"aws_db_parameter_group.bar", "arn", regexp.MustCompile(fmt.Sprintf("^arn:[^:]+:rds:[^:]+:\\d{12}:pg:%s", groupName))),
 				),
 			},
+			{
+				Config: testAccAWSDBParameterGroupConfig(groupName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSDBParameterGroupExists("aws_db_parameter_group.bar", &v),
+					testAccCheckAWSDBParameterGroupAttributes(&v, groupName),
+					testAccCheckAWSDBParameterNotUserDefined("aws_db_parameter_group.bar", "collation_connection"),
+					testAccCheckAWSDBParameterNotUserDefined("aws_db_parameter_group.bar", "collation_server"),
+				),
+			},
 		},
 	})
 }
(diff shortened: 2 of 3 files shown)


stefansundin:
Oh, and I will write tests for this tomorrow!

stefansundin:
I have been trying to add a test that would detect when the code fails to reset the parameter, but I am having issues. So this is with my new code commented out.

I added [a new commit](https://github.com/terraform-providers/terraform-provider-aws/pull/5728/commits/cfec3c0f95e96d68fdac53ec58e01b5d20a5a112) with my test code. With that test, and the new code commented out, I expect the test to fail, but it does not.

So my understanding is that in the test steps in `TestAccAWSDBParameterGroup_basic`, they all use the same parameter group, and it isn't destroyed until after the last test step. So the purpose of my new test step is to try to undo the new parameters that were added in the step before, and then check that `collation_connection` and `collation_server` are no longer user defined.

Any help on this would be awesome! Thank you!

aws/resource_aws_db_parameter_group.go
```diff
@@ -295,6 +295,26 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
 			}
 			d.SetPartial("parameter")
 		}
+
+		// Reset parameters that have been removed
+		parameters, err = expandParameters(os.Difference(ns).List())
```
ryndaniels: I'm seeing in the `ResetDBParameterGroupInput` SDK documentation [here](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#ResetDBParameterGroupInput) that there is a documented 20 item max, so it would probably be best to add that logic in here. There's an example of doing this with a for-loop [here](https://github.com/terraform-providers/terraform-provider-aws/pull/10161/files#diff-497d6ad0ec0d61809a57fc6ff6ffaadbR280-R281). 🙂

aws/resource_aws_db_parameter_group.go
```diff
@@ -295,6 +295,26 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
...
+		if err != nil {
+			return err
+		}
+		if len(parameters) > 0 {
+			parameterGroupName := d.Get("name").(string)
+			resetOpts := rds.ResetDBParameterGroupInput{
+				DBParameterGroupName: aws.String(parameterGroupName),
+				Parameters:           parameters,
```
ryndaniels: It looks like by default, all parameters are reset so we should add this in to prevent that:

```suggestion
				Parameters:           parameters,
				ResetAllParameters:   aws.Bool(false),
```

aws/resource_aws_db_parameter_group_test.go
```diff
@@ -291,6 +291,15 @@ func TestAccAWSDBParameterGroup_basic(t *testing.T) {
 						"aws_db_parameter_group.bar", "arn", regexp.MustCompile(fmt.Sprintf("^arn:[^:]+:rds:[^:]+:\\d{12}:pg:%s", groupName))),
 				),
 			},
+			{
+				Config: testAccAWSDBParameterGroupConfig(groupName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSDBParameterGroupExists("aws_db_parameter_group.bar", &v),
+					testAccCheckAWSDBParameterGroupAttributes(&v, groupName),
+					testAccCheckAWSDBParameterNotUserDefined("aws_db_parameter_group.bar", "collation_connection"),
+					testAccCheckAWSDBParameterNotUserDefined("aws_db_parameter_group.bar", "collation_server"),
```
ryndaniels: To add some additional checking to this, you could try adding something like: 
```suggestion
					testAccCheckAWSDBParameterNotUserDefined("aws_db_parameter_group.bar", "collation_server"),
					testAccCheckNoResourceAttr("aws_db_parameter_group.bar", "parameter.2475805061.value"),
```

for additional testing to make sure the parameters were reset.

aws/resource_aws_db_parameter_group_test.go
```diff
@@ -539,6 +548,45 @@ func testAccCheckAWSDBParameterGroupExists(n string, v *rds.DBParameterGroup) re
...
+			return fmt.Errorf("No DB Parameter Group ID is set")
+		}
+
+		conn := testAccProvider.Meta().(*AWSClient).rdsconn
+
+		opts := rds.DescribeDBParametersInput{
+			DBParameterGroupName: aws.String(rs.Primary.ID),
+			Source:               aws.String("user"),
+		}
+		fmt.Printf("test: %s\n", rs.Primary.ID)
```
ryndaniels: In general, we'd prefer not to have `fmt.Printf`s in the code. For potential debug information, a `log.Printf` could work, but if these were just testing messages, could you go ahead and remove them?

aws/resource_aws_rds_cluster_parameter_group.go
```diff
@@ -234,6 +234,26 @@ func resourceAwsRDSClusterParameterGroupUpdate(d *schema.ResourceData, meta inte
 			}
 			d.SetPartial("parameter")
 		}
+
+		// Reset parameters that have been removed
```
ryndaniels: Could you go ahead and add a test for this change as well? Thanks!

nathanielks:
@stefansundin Hey there! I was curious if you're still interested in patching this? If not, I might try picking up the torch and finishing it.

stefansundin:
@nathanielks Feel free to continue this work.

ryndaniels:
@nathanielks thanks for volunteering to pick this up! I've actually just started working on this myself and have got it just about ready to go, so if there's other stuff that's higher priority for you to work on, that's totally fine and we should be getting this going on our end shortly! 🙂

nathanielks:
Oh brilliant, thanks @ryndaniels! Carry on, then 😄

ryndaniels:
(Pulled these commits into another branch (#11540) just to simplify things on our end - thanks for getting this started @stefansundin - I'm going to close this PR to consolidate conversation on the new one! 🙂)

PR #4610, opened 2018-05-22
New Resource: aws_guardduty_invite_accepter

bflad:
Closes #2489 

Changes proposed in this pull request:

* Adds alternate AWS account handling in the provider acceptance testing framework
* Adds new resource for managing a GuardDuty member account invitation from a master account and disassociation from a master account

Output from acceptance testing:

```
    --- PASS: TestAccAWSGuardDuty/InviteAccepter (18.16s)
        --- PASS: TestAccAWSGuardDuty/InviteAccepter/basic (18.16s)
```

PR #5748, opened 2018-08-31
resource/aws_ecr_repository: Use RepositoryUri for repository_url attribute

### LINKED ISSUE #5741: ecr_registry: repositry_url output doesn't support china/gov region
### Terraform Version
0.11.3

### Affected Resource(s)

* aws_ecr_repository

output "e2e_ecr_url" {
  value = "${aws_ecr_repository.attribution-cf-e2e.repository_url}"
}

### Expected Behavior

URL ends with com.cn for cn-north-1 region

### Actual Behavior
URL ends with com for cn-north-1 region

### Steps to Reproduce
Create ECR repository in china
Try to output the URL

bflad:
Fixes #5741 

* Use API `RepositoryUri` instead of custom built URL for `aws_ecr_repository` resource `repository_url` attribute
* Refactors acceptance testing to remove `import_*_test.go` file and check `registry_id`/`repository_url` attribute values in Terraform state
* Adds `testAccCheckResourceAttrRegionalARN` to start reducing usage of `TestMatchResourceAttr` for ARN attributes

AWS Commercial:

```
--- PASS: TestAccAWSEcrRepository_basic (13.26s)
```

AWS GovCloud (US):

```
--- PASS: TestAccAWSEcrRepository_basic (19.02s)
```


aws/provider_test.go
```diff
@@ -78,6 +80,44 @@ func testAccPreCheck(t *testing.T) {
...
+// testAccAwsProviderAccountID returns the account ID of an AWS provider
+func testAccAwsProviderAccountID(provider *schema.Provider) string {
+	if provider == nil {
+		log.Print("[DEBUG] Unable to read account ID from test provider: empty provider")
+		return ""
+	}
+	if provider.Meta() == nil {
+		log.Print("[DEBUG] Unable to read account ID from test provider: unconfigured provider")
+		return ""
+	}
```
tombuildsstuff: _minor_ feels like these could end up all over the place, may be worth putting into a central method somewhere that panics for tests?

bflad: We chatted about this out of band -- currently this will output like this if done incorrectly, which is friendlier than a panic (at least in my opinion):

```
TF_LOG=debug make testacc TEST=./aws TESTARGS='-run=TestAccAWSEcrRepository_basic'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEcrRepository_basic -timeout 120m
=== RUN   TestAccAWSEcrRepository_basic
2018/08/30 23:05:03 [DEBUG] Unable to read account ID from test provider: empty provider
2018/08/30 23:05:03 [DEBUG] Unable to read account ID from test provider: empty provider
--- FAIL: TestAccAWSEcrRepository_basic (9.20s)
    testing.go:527: Step 0 error: Check failed: Check 4/5 error: aws_ecr_repository.default: Attribute 'registry_id' expected "", got "187416307283"
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	10.447s
```

Luckily we don't have too many `AWSClient` attributes we'll need to fetch off the provider after its initialized, so should just be one or two of these functions added into `provider_test.go` as the current acceptance testing framework is written.

I think we both agree that we should try to fix the ordering for the provider initialization if possible though. The delayed provider configuration in the `PreCheck` is what makes this code worse than it needs to be. I'll take this as a separate, more involved task (likely part of #4610), since for these new functions in `provider_test.go` it should be an implementation detail and should not not fundamentally affect all the resources themselves (hopefully). 👍

bflad:
This has been released in [version 1.35.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1350-september-06-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5749, opened 2018-08-31
Add working URL for ICMP types and codes

roooms:

* Add new URL for ICMP types and codes as existing one 404s


roooms:
Hey @bflad, sorry I didn't check for that before. The second commit updates all instances of the broken link.

PR #5755, opened 2018-08-31
vendor: hashicorp/go-cleanhttp@v0.5.0

bflad:
Preparation for migrating to Go modules. Notable change: `net.Dialer` has `DualStack` enabled. Documentation: https://golang.org/pkg/net/#Dialer

```
        // DualStack enables RFC 6555-compliant "Happy Eyeballs"
        // dialing when the network is "tcp" and the host in the
        // address parameter resolves to both IPv4 and IPv6 addresses.
        // This allows a client to tolerate networks where one address
        // family is silently broken.
        DualStack bool
```

* Updated via `govendor fetch github.com/hashicorp/go-cleanhttp/...@v0.5.0`


bflad:
Reference: #5773

PR #5764, opened 2018-09-04
This addresses the oddity of the configuration 

### LINKED ISSUE #2796: Putting GITHUB_TOKEN in terraform config for aws_codepipeline
When using AWS CodePipeline, if you want to fetch your source code from GitHub, you need to run `terraform apply` with the environment variable `GITHUB_TOKEN` (whose value should be a previously generated OAuth token).

This kinda goes against Terraform's philosophy. Would it be more sensible to have a parameter for the resource `aws_codepipeline` to set that token? Is there a compelling reason to use an environment variable instead?

I'm starting the discussion. If we end up deciding on a parameter I'm willing to have a go at a PR.

psteininger:
This addresses the oddity of the configuration in #2796. This simply removes the code that deletes the OAuthToken passed in and arbitrary validation condition looking for GITHUB_TOKEN env variable. This behavior goes against Terraform convention, and was absolutely NOT well documented, it also drove me insane for a few days until I found an off-hand mention and code reference. There was also no rationale provided for why the code should delete what was provided and replace it with the value of arbitrary ENV variable. 

A much better way to address this is to improve the schema of the in the project definition to mark OAuthToken as sensitive.

Fixes #2796

* Remove validation code requiring GITHUB_TOKEN env var
* Remove code that deletes the value provided in the configuration per [AWS official documentation](https://docs.aws.amazon.com/codepipeline/latest/userguide/GitHub-rotate-personal-token-CLI.html)

```
I can't get it to run, never worked with Go.
...
```

### DIFF
--- a/aws/resource_aws_codepipeline.go
+++ b/aws/resource_aws_codepipeline.go
@@ -166,11 +166,7 @@ func resourceAwsCodePipeline() *schema.Resource {
 }
 
 func validateAwsCodePipelineStageActionConfiguration(v interface{}, k string) (ws []string, errors []error) {
-	for k := range v.(map[string]interface{}) {
-		if k == "OAuthToken" {
-			errors = append(errors, fmt.Errorf("CodePipeline: OAuthToken should be set as environment variable 'GITHUB_TOKEN'"))
-		}
-	}
+
 	return
 }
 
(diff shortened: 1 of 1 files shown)


gdavison:
Thanks for the PR, @psteininger. Unfortunately, the AWS API doesn't return the OAuthToken for security reasons, so the change is a bit more involved than not deleting the API response. We've created #14175 to address the issue.

PR #5795, opened 2018-09-05
provider: Return error instead of WARN log with AWS account ID lookup failure

bflad:
Followup to: https://github.com/terraform-providers/terraform-provider-aws/pull/5794

This change is already noted in the Version 2 Upgrade Guide.


bflad:
Closing this PR due to #7653 -- we will still be making this change though, tracking issue: #7674

PR #5800, opened 2018-09-06
Fix error creating aws_db_instance from snapshot where snapshot uses storage_type io1

dannosaur:
There's no issue number for this, this is something I've stumbled across when trying to build an environment in a very specific way.

When restoring a DB instance from a snapshot, where the snapshot uses io1 storage type, the initial create succeeds, but the second step of the create within terraform that modifies the DB instance to sync it up with the resource state fails due to the fact that the iops config item isn't passed. The error it crashes out with is:

```
* aws_db_instance.db: error modifying DB Instance (terraform-2018090607083682360000000e): InvalidParameterCombination: Both storage size and iops must be specified when modifying storage size or iops on an DB Instance that has iops
```

* Move the iops config element from the initial create to the secondary "modify" stage of the create (the initial iops value gets inherited from the snapshot, so this isn't required in the initial stage


bflad:
Hi @dannosaur 👋 Thanks for submitting this!

Can you please provide an acceptance test or an example of your configuration so we can verify this as well as prevent future regressions?

dannosaur:
I'm actually unsure as to how one might build an acceptance test for this. The issue being that in order to hit this piece of code, one must restore a database from a snapshot. As I understand it (please excuse the fact that I have extremely little knowledge of Go), these tests work by creating real resources and testing that the result is as expected? If this is the case, then whoever's testing has a prerequisite in that they need a snapshot ARN to hand in order to restore it from. Public snapshots are a thing, but I can't find a list of anything to use (and the removal of said snapshot would break the test, obviously not what we want).

How would you actually go about this? Is there another method where you can "fake" the calls somehow and test that the calls made are what's expected and in the right order?

Help appreciated on this one. I'd love to contribute more.

bflad:
@dannosaur no worries -- we have existing acceptance testing similar to what needs to be done here where we create a database, snapshot it, then create a second database that is restored from snapshot. Check out:

https://github.com/terraform-providers/terraform-provider-aws/blob/2bcb2446484e2f9a8b98af37918f2b4c21698cb8/aws/resource_aws_db_instance_test.go#L709-L734

and

https://github.com/terraform-providers/terraform-provider-aws/blob/2bcb2446484e2f9a8b98af37918f2b4c21698cb8/aws/resource_aws_db_instance_test.go#L3554-L3579

Then to run it:

```
export AWS_PROFILE=... # or export AWS_ACCESS_KEY_ID=... + export AWS_SECRET_ACCESS_KEY=...
make testacc TEST=./aws TESTARGS='-run=TestAccAWSDBInstance_SnapshotIdentifier_AllocatedStorage'
```

bvargheseamplify:
Our RDS sandbox builds are failing due to this.  Any ETA on the release of the fix?

hylaride:
We've been getting hit with this bug for a few months now.  Current setup: 

```
Terraform v0.11.10
+ provider.aws v1.51.0
+ provider.null v1.0.0
+ provider.template v1.0.0
```

sklarsa:
Is this not being merged because of a lack of tests?  If so, I can take on this task based on @bflad's comments from above.  But if someone else is working on it, I don't want to duplicate effort...

This is pretty critical for my company's workflow, so it would be great to see it fixed!

dannosaur:
@sklarsa yes, that's correct. I simply lack the time to sit and study how the test suite works, and learn enough of Go's syntax to write correct and valid tests that pass.

sgript:
Any updates? Running into the same issue.

sklarsa:
My PR #7426 addresses this problem, but it needs review

hernan82arg:
hitting same issue here... can we please merge? thanks

sgript:
Would be amazing. This is the only thing seemingly stopping us from terraforming our database with an existing snapshot.

sgript:
cc @bflad

bflad:
As noted here: https://github.com/terraform-providers/terraform-provider-aws/pull/7426#pullrequestreview-206675306

Merging this first so @dannosaur gets proper attribution then merging #7426. Thanks so much @dannosaur and @sklarsa

bflad:
This has been released in [version 1.60.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1600-february-22-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5815, opened 2018-09-08
Issue #5808 aws_iam_policy does not have a description attribute

### LINKED ISSUE #5808: aws_iam_policy does not have a description attribute
The aws_iam_policy resource does not have a description attribute yet it is listed in the [terraform docs](https://www.terraform.io/docs/providers/aws/r/iam_policy.html#description-1).

### Terraform Version

```
$ terraform -v
Terraform v0.11.8
+ provider.aws v1.35.0
```

### Affected Resource(s)

aws_iam_policy

```hcl
provider "aws" {
  region = "eu-west-3"
}

data "aws_iam_policy_document" "example" {
  statement {
    actions = [
      "ec2:Describe*",
    ]
    resources = ["*"]
  }
}

resource "aws_iam_policy" "example" {
  name   = "example"
  policy = "${data.aws_iam_policy_document.example.json}"
}

output "description" {
  value = "${aws_iam_policy.example.description}"
}
```

### Debug Output

```
$ terraform apply
data.aws_iam_policy_document.example: Refreshing state...
aws_iam_policy.example: Refreshing state... (ID: arn:aws:iam::REDACTED:policy/example)

Error: Error running plan: 1 error(s) occurred:

* output.description: Resource 'aws_iam_policy.example' does not have attribute 'description' for variable 'aws_iam_policy.example.description'
```

### Expected Behavior

Unsure if `description` is an AWS IAM policy attribute. If it is I would expect a default value (perhaps "") or if it is not an attribute remove this attribute from the aws_iam_policy resource.

### Actual Behavior

Error when running plan.

### Steps to Reproduce

1. `terraform apply`

saravanan30erd:
Fixes #5808 

* Change 1
 added default value to description attribute.

```
$ make testacc TEST=./aws TESTARGS='-run=TestAWSPolicy_checkDescription'
TF_ACC=1 go test ./aws -v -run=TestAWSPolicy_checkDescription -timeout 120m
=== RUN   TestAWSPolicy_checkDescription
--- PASS: TestAWSPolicy_checkDescription (28.46s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	28.503s
...
```

### DIFF
--- a/aws/resource_aws_iam_policy.go
+++ b/aws/resource_aws_iam_policy.go
@@ -28,7 +28,6 @@ func resourceAwsIamPolicy() *schema.Resource {
 				Type:     schema.TypeString,
 				ForceNew: true,
 				Optional: true,
-				Default:  "Managed by Terraform",
 			},
 			"path": {
 				Type:     schema.TypeString,
@@ -279,12 +278,7 @@ func iamPolicyListVersions(arn string, iamconn *iam.IAM) ([]*iam.PolicyVersion, 
 
 func readIamPolicy(d *schema.ResourceData, policy *iam.Policy) error {
 	d.SetId(*policy.Arn)
-	if policy.Description != nil {
-		// the description isn't present in the response to CreatePolicy.
-		if err := d.Set("description", policy.Description); err != nil {
-			return err
-		}
-	}
+	d.Set("description", policy.Description)
 	if err := d.Set("path", policy.Path); err != nil {
 		return err
 	}
--- a/aws/resource_aws_iam_policy.go
+++ b/aws/resource_aws_iam_policy.go
@@ -28,6 +28,7 @@ func resourceAwsIamPolicy() *schema.Resource {
 				Type:     schema.TypeString,
 				ForceNew: true,
 				Optional: true,
+				Default:  "Managed by Terraform",
 			},
 			"path": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


bflad:
Hi @saravanan30erd 👋  Thanks for submitting this.

Changing the attribute default in this manner is a breaking change for existing Terraform configurations (even with a state migration) and while it accomplishes the goal of having `d.Set("description", ...)` always called (since it will not be returned as `nil` by default from the API), its not necessarily the correct way to accomplish this.

This is the problematic line, which dates back 3 years:

https://github.com/terraform-providers/terraform-provider-aws/blob/7be15b7d6273e2badbd0a7d2453eee452bcecd7c/aws/resource_aws_iam_policy.go#L281

My best guess is that this was written before `d.Set()` was setup to handle pointers for scalar types (e.g. `*string`) and that the other fields were fine without that check because they are more or less "required" from a service standpoint. Nowadays, we can pass a `*string` directly into `d.Set()` and it'll set `""` in the state if its an empty string.

Instead of making the broader changes here, I would suggest simply trying to just remove that `if policy.Description != nil` line and adding a check to an existing acceptance test like this (where `description` isn't being set in test configuration):

```go
resource.TestCheckResourceAttr("aws_iam_policy.policy", "description", ""),
```

Theoretically, the above check should fail on master for test configurations missing that argument.

saravanan30erd:
@bflad thanks, done 👍 
I referred how `description` handled in some other resources and followed the same.

bflad:
This has been released in [version 1.36.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1360-september-13-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5862, opened 2018-09-13
[WIP] Add initial container definitions data source.

jmcarp:
This is an early WIP that I'm submitting for feedback--don't use it yet! I would like to define ecs container definitions in hcl directly, without writing templated json, much like https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html for iam policies. This new resource allows users to define container definitions like so:

```
data "aws_ecs_container_definitions" "test" {
  container_definitions {
    name = "container1"
    image = "image1"
    port_mappings {
      container_port = 5000
    }
  }
  container_definitions {
    name = "container2"
    image = "image2"
  }
}

output "container_json" {
  value = "${data.aws_ecs_container_definitions.test.json}"
}
```

If the use case makes sense, I'll flesh this out, write tests, make the code nicer, etc. WDYT @bflad ?

### DIFF
--- a/aws/data_source_aws_ecs_container_definitions.go
+++ b/aws/data_source_aws_ecs_container_definitions.go
@@ -1,0 +1,106 @@
+package aws
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceAwsEcsContainerDefinitions() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsEcsContainerDefinitionsRead,
+
+		Schema: map[string]*schema.Schema{
+			"container_definitions": {
+				Type:     schema.TypeSet,
+				Required: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"image": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"memory": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"port_mappings": {
+							Type:     schema.TypeSet,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"container_port": {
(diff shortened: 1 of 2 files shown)


jmcarp:
@bflad I see you tagged this--thanks! Let me know if you'd be open to the feature, and I'll clean up the patch if so.

bflad:
Hey @jmcarp 👋  I think in principal this is a good idea, but it might be helpful to see a design sketch of the planned configuration options/usage laid out and discussed in an issue first. There was a similar ask in https://github.com/terraform-providers/terraform-provider-aws/issues/3153#issuecomment-396359751, but this is deserving of its own feature request for "Defining ECS Container Definitions in HCL". (I'll likely close out the YAML issue anyways since its not supported by the ECS API and the AWS provider should not concern itself with YAML -> JSON conversion.)

HCL with configuration validation and converting to JSON is a good start, but I think we'll want to look at the problem more holistically as well, e.g. how do I easily add container definition X to Y container definitions?

```
# Example design sketch - ignoring current aws_ecs_container_definition data source implementation
# Arguably, the existing aws_ecs_container_definition should be migrated to something like aws_ecs_task_definition_container_definition (what a gross name)
# Or preferrably, part of the aws_ecs_task_definition data source in a new container_definitions map with name as the key

data "aws_ecs_container_definition" "monitoring" {
  # name, image, etc.
}

data "aws_ecs_container_definition" "application1" {
  # name, image, etc.
}

data "aws_ecs_container_definition" "application2" {
  # name, image, etc.
}

data "aws_ecs_container_definitions" "service1" {
  container_definitions = [
    "${data.aws_ecs_container_definitions.monitoring.json}",
    "${data.aws_ecs_container_definitions.application1.json}",
  ]
}

data "aws_ecs_container_definitions" "service2" {
  container_definitions = [
    "${data.aws_ecs_container_definitions.monitoring.json}",
    "${data.aws_ecs_container_definitions.application2.json}",
  ]
}
```

Let's design and discuss this further in the new issue before trying to continue with any implementation. 👍

bflad:
As mentioned in https://github.com/terraform-providers/terraform-provider-aws/pull/5862#issuecomment-425446332 lets work towards a design proposal and then we can consider an appropriate implementation. 👍 

That said, Terraform 0.12 does include a new [`jsonencode()` function](https://www.terraform.io/docs/configuration/functions/jsonencode.html), which likely alleviates some of the need for a new data source in general except for validation purposes. 😄

steveh:
This would still be very useful in order to `ignore_changes` to only select fields, e.g. if you want to deploy a new version of an image outside of Terraform.

PR #5875, opened 2018-09-13
Add a aws_codepipeline_webhook resource for managing CodePipeline webhooks

joestump:
## What does this add?

Introduces the `aws_codepipeline_webhook` resource to manage [AWS CodePipeline webhooks](https://docs.aws.amazon.com/codepipeline/latest/APIReference/API_PutWebhook.html). 

## Example HCL

```hcl
# A shared secret between GitHub and AWS that allows AWS
# CodePipeline to authenticate the request came from GitHub.
# Would probably be better to pull this from the environment
# or something like SSM Parameter Store.
locals {
  webhook_secret = "super-secret"
}

resource "aws_codepipeline_webhook" "bar" {
  name            = "test-webhook-github-bar" 
  authentication  = "GITHUB_HMAC" 
  target_action   = "Source"
  target_pipeline = "${aws_codepipeline.bar.name}"

  authentication_configuration {
    secret_token = "${local.webhook_secret}"
  }

  filter {
    json_path    = "$.ref"
    match_equals = "refs/heads/{Branch}"
  }
}

# Wire the CodePipeline webhook into a GitHub repository.
resource "github_repository_webhook" "bar" {

  name = "web"

  configuration {
    url          = "${aws_codepipeline_webhook.bar.url}"
    content_type = "form"
    insecure_ssl = true 
    secret       = "${local.webhook_secret}"
  }
  events = ["push"]
}
```

## Test output

```
make testacc TESTARGS='-run=TestAccAWSCodePipelineWebhook'
TF_ACC=1 go test ./... -v -run=TestAccAWSCodePipelineWebhook -timeout 120m
=== RUN   TestAccAWSCodePipelineWebhook_basic
--- PASS: TestAccAWSCodePipelineWebhook_basic (31.78s)
=== RUN   TestAccAWSCodePipelineWebhook_ipAuth
--- PASS: TestAccAWSCodePipelineWebhook_ipAuth (33.43s)
=== RUN   TestAccAWSCodePipelineWebhook_unauthenticated
--- PASS: TestAccAWSCodePipelineWebhook_unauthenticated (29.54s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	95.993s
```

## TODO

- [x] Create an acceptance test
- [x] Write documentation
- [x] Clean up PR with HCL example


joestump:
Here's a complete working example. Push a `buildspec.yml` to kick off a pipeline run/build.

https://gist.github.com/joestump/cac3abb94050186fcba1c57c8a880a71

joestump:
@bflad I took another pass at this one after your comments on #6117. Would love to get 👀 on this one as well. 😄

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	"github.com/aws/aws-sdk-go/service/codepipeline"
+)
+
+func resourceAwsCodePipelineWebhook() *schema.Resource {
+	return &schema.Resource{
+		Create:        resourceAwsCodePipelineWebhookCreate,
+		Read:          resourceAwsCodePipelineWebhookRead,
+		Update:        nil,
+		Delete:        resourceAwsCodePipelineWebhookDelete,
+		SchemaVersion: 1,
```
bflad: `SchemaVersion` should be omitted for new resources (it defaults to `0`)

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+			},
+			"authentication_configuration": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				MinItems: 1,
+				Required: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"secret_token": {
```
bflad: Should this be marked `Sensitive: true`?

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+			"target_pipeline": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+		},
+	}
+}
+
+func extractCodePipelineWebhookAttr(d *schema.ResourceData, attr string) (map[string]interface{}, error) {
```
bflad: We prefer to allow the schema and any API errors to dictate this behavior rather than implementing additional error handling in the resource functions.  `authentication_configuration` is marked as `Required: true` and we can allow the API (or the handling in `extractCodePipelineWebhookAuthConfig`) to return errors if both `Optional` nested arguments are omitted.

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+		secretToken := authConfig["secret_token"].(string)
+		if secretToken == "" {
+			return nil, fmt.Errorf("Secret token must be set when using GITHUB_HMAC")
+		}
+
+		conf.SecretToken = &secretToken
+		break
+	case codepipeline.WebhookAuthenticationTypeUnauthenticated:
+		break
+	default:
```
bflad: The schema handling for the `auth_type` argument prevents this case from ever getting hit.

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	return rules, nil
+}
+
+func extractCodePipelineWebhookAuthConfig(authType string, authConfig map[string]interface{}) (*codepipeline.WebhookAuthConfiguration, error) {
+	var conf codepipeline.WebhookAuthConfiguration
+	switch authType {
+	case codepipeline.WebhookAuthenticationTypeIp:
+		ipRange := authConfig["allowed_ip_range"].(string)
+		if ipRange == "" {
+			return nil, fmt.Errorf("An IP range must be set when using IP-based auth")
```
bflad: If the API returns a sufficient error message, we should remove this 👍

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+
+		conf.AllowedIPRange = &ipRange
+
+		break
+	case codepipeline.WebhookAuthenticationTypeGithubHmac:
+		secretToken := authConfig["secret_token"].(string)
+		if secretToken == "" {
+			return nil, fmt.Errorf("Secret token must be set when using GITHUB_HMAC")
```
bflad: If the API returns a sufficient error message, we should remove this 👍

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	webhook, err := conn.PutWebhook(request)
+	if err != nil {
+		return fmt.Errorf("Error creating webhook: %s", err)
+	}
+
+	arn := *webhook.Webhook.Arn
+	d.SetId(arn)
+
+	url := *webhook.Webhook.Url
+	d.Set("url", url)
```
bflad: `d.Set()` should only be called in the `Read` function, unless this information is not available then. 👍

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	return webhooks, nil
+}
+
+func setCodePipelineWebhookFilters(webhook codepipeline.WebhookDefinition, d *schema.ResourceData) error {
+	filters := []interface{}{}
+	for _, filter := range webhook.Filters {
+		f := map[string]interface{}{
+			"json_path":    *filter.JsonPath,
```
bflad: To prevent potential panics, all dereferences should be handled by the AWS Go SDK provided helper functions or have `nil` checks. e.g. `"json_path": aws.StringValue(filter.JsonPath),`

This is true of many of the `*` dereferences below, but I'll withhold from commenting on every single one of them.

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func resourceAwsCodePipelineWebhookRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).codepipelineconn
+	arn := d.Id()
+	webhooks, err := getAllCodePipelineWebhooks(conn)
```
bflad: Instead of finding and returning all the webhooks then searching for the one we want, we should search until we find the expected one and exit early with it (or return `nil`). 👍

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return fmt.Errorf("Error fetching webhooks: %s", err)
+	}
+
+	if len(webhooks) == 0 {
+		return fmt.Errorf("No webhooks returned!")
```
bflad: Instead of returning an error when the resource is not found, we should trigger resource recreation, e.g.

```go
log.Printf("[WARN] CodePipeline Webhook (%s) not found, removing from state", d.Id())
d.SetId("")
return nil
```

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+			break
+		}
+
+		nextToken = aws.StringValue(out.NextToken)
+	}
+
+	return webhooks, nil
+}
+
+func setCodePipelineWebhookFilters(webhook codepipeline.WebhookDefinition, d *schema.ResourceData) error {
```
bflad: Nit: In this project we tend to have "flatten" functions which perform the first half of this function, then leave the `ResourceData` handling in the `Read` function. e.g.

```go
func flattenCodePipelineWebhookFilters(filters []*codepipeline.WebhookFilterRule) []interface{} {
// ...
}

// in Read
if err := d.Set("filter", flattenCodePipelineWebhookFilters(webhook.Filters)); err != nil {
  return fmt.Errorf("error setting filter: %s", err)
}
```

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	name := *found.Name
+	if name == "" {
+		return fmt.Errorf("Webhook not found: %s", arn)
+	}
+
+	d.Set("name", name)
+
+	targetAction := *found.TargetAction
```
bflad: Other than the recommendation to use the AWS Go SDK helper functions for these, its also important to note that `d.Set()` can also directly handle pointers to simplify things and not require dereferencing (if the types match), e.g.

```go
if err := d.Set("target_action", found.TargetAction); err != nil {
  return fmt.Errorf("error setting target_action: %s", err)
}
```

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+	input := codepipeline.DeleteWebhookInput{
+		Name: &name,
+	}
+	_, err := conn.DeleteWebhook(&input)
+
+	if err != nil {
+		return fmt.Errorf("Could not delete webhook: %s", err)
+	}
+
+	d.SetId("")
```
bflad: Calling `d.SetId("")` during the `Delete` function is extraneous. 👍

aws/resource_aws_codepipeline_webhook_test.go
```diff
@@ -0,0 +1,195 @@
...
+      configuration {
+        ProjectName = "test"
+      }
+    }
+  }
+}
+
+resource "aws_codepipeline_webhook" "bar" {
+    name            = "test-webhook-%s" 
+    authentication  = "GITHUB_HMAC"
```
bflad: We should add testing for the other `authentication` types as well. 👍 You may find its easiest to just make the "base" configuration above its own function so it doesn't need to be duplicated each test configuration.

aws/resource_aws_codepipeline_webhook.go
```diff
@@ -0,0 +1,336 @@
...
+					codepipeline.WebhookAuthenticationTypeGithubHmac,
+					codepipeline.WebhookAuthenticationTypeIp,
+					codepipeline.WebhookAuthenticationTypeUnauthenticated,
+				}, false),
+			},
+			"authentication_configuration": {
+				Type:     schema.TypeList,
+				MaxItems: 1,
+				MinItems: 1,
+				Required: true,
```
bflad: Is this required for `authentication = "UNAUTHENTICATED"`?

aws/resource_aws_codepipeline_webhook.go
joestump: Good to know. For clarification, which is preferred?

```go
d.Set("target_action", found.TargetAction)
```

or...

```go
d.Set("target_action", aws.String(found.TargetAction))
```

aws/resource_aws_codepipeline_webhook.go
joestump: It's not. I'll mark it as `Optional:  true` instead.

joestump:
@bflad I believe all comments have been addressed. There's a test for each `authentication_configuration` iteration now. 👍

PR #5912, opened 2018-09-18
Data Source: aws_ami_ids, add sort_ascending flag

sargun:
This adds the sort ascending flag by creation time to
aws_ami_ids, to allow for a list which only ever adds
items at the end.

This is useful for example, when you have something like this:

```hcl

data "aws_ami_ids" "agent_amis" {
  owners = ["self"]

  filter {
    name   = "name"
    values = ["foo-*"]
  }
}

resource "aws_ami_launch_permission" "launch_permissions_1234" {
  count      = "${length(data.aws_ami_ids.agent_amis.ids)}"
  image_id   = "${data.aws_ami_ids.agent_amis.ids[count.index]}"
  account_id = "1234"
}
```

Otherwise, every time a new AMI is added that matches "foo-*",
it must "recreate" the launch permissions. Although this may
be trivial for launch permissions, on something like AMI
copy, it can wreak havoc.

* Add `sort_ascending` flag to data source `ami_ids` to optionally sort ascending. It will not 

```
$ make testacc TESTARGS="-run=TestAccDataSourceAwsAmiIds_sorted"
TF_ACC=1 go test ./... -v -run=TestAccDataSourceAwsAmiIds_sorted -timeout 120m
=== RUN   TestAccDataSourceAwsAmiIds_sorted
--- PASS: TestAccDataSourceAwsAmiIds_sorted (9.65s)
PASS
```

### DIFF
--- a/aws/data_source_aws_ami.go
+++ b/aws/data_source_aws_ami.go
@@ -277,7 +277,7 @@ func dataSourceAwsAmiRead(d *schema.ResourceData, meta interface{}) error {
 
 // Returns the most recent AMI out of a slice of images.
 func mostRecentAmi(images []*ec2.Image) *ec2.Image {
-	return sortImages(images)[0]
+	return sortImages(images, false)[0]
 }
 
 // populate the numerous fields that the image description returns.
--- a/aws/data_source_aws_ami_ids.go
+++ b/aws/data_source_aws_ami_ids.go
@@ -41,6 +41,11 @@ func dataSourceAwsAmiIds() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"sort_ascending": {
+				Type:     schema.TypeBool,
+				Default:  false,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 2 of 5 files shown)


sargun:
Since adding a reverse function to Terraform is in purgatory (https://github.com/hashicorp/terraform/pull/18887), I figured might as well add the capability natively here.

sargun:
This is the same kind of issue that's presented in https://github.com/terraform-providers/terraform-provider-aws/issues/155, but because AMIs can be "captured" in time, it avoids this problem, as long as you never delete an AMI.

sargun:
Also, let me know if you want me to change the acceptance test back, or split that into a separate PR. Waiting for a VM to spin up, and for snapshotting is a pain for waiting for the test to pass.

bflad:
This has been released in [version 1.37.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1370-september-19-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5926, opened 2018-09-19
New Resource: Snowball Instance

cb17:
This PR serves as more of a conversation starter than a hard submission for code I would consider production-ready, however, it does work as submitted. Snowball Edge EC2 is a unique beast, from an EC2 point of view, in that each instance is essentially a different provider (a unique endpoint and set of credentials). 

For this stub, I'm using a static local file to provide provider configuration, but I would imagine for it to be production ready, there would have to be an external lookup somewhere for this info. Ideally the user of the template would only have to include a job ID in the template and the lookup for the rest of the info would happen behind the scenes.

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -578,6 +578,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_s3_bucket_metric":                             resourceAwsS3BucketMetric(),
 			"aws_s3_bucket_inventory":                          resourceAwsS3BucketInventory(),
 			"aws_security_group":                               resourceAwsSecurityGroup(),
+			"aws_snowball_instance":                            resourceAwsSnowballInstance(),
 			"aws_network_interface_sg_attachment":              resourceAwsNetworkInterfaceSGAttachment(),
 			"aws_default_security_group":                       resourceAwsDefaultSecurityGroup(),
 			"aws_security_group_rule":                          resourceAwsSecurityGroupRule(),
--- a/aws/resource_aws_snowball_instance.go
+++ b/aws/resource_aws_snowball_instance.go
@@ -1,0 +1,1075 @@
+package aws
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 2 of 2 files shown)


cb17:
I'd specifically like to highlight lines 25-49 in the resource file

bflad:
Hi @cb17 👋 Thanks for submitting this! Snowball Edge is pretty neat.

I'm wondering if we should take a step back and maybe try to work on the design of this in an issue first? It would be nice to try and think through some possibilities and desired workflows before settling on a concrete implementation.

At first glance, it seems like we might want to setup something like the following:
* Figure out how to get the existing `aws` provider to properly connect to a "Snowball Edge EC2" endpoint
* Potentially using the existing `aws_instance`, `aws_s3_bucket`, and `aws_lambda_function` resources (adding Snowball Edge specific handling if necessary, although since they claim it is a subset of functionality, we generally just need to bypass `UnsupportedOperation` read errors in those cases)
* Potentially creating Snowball related data sources and resources to support the above

I believe the above would reduce the maintenance burden of supporting Snowball. What do you think?

cb17:
That all sounds good to me. I just submitted it as a separate resource because I was trying not to break stuff :)

Agreed, most of the trick is going to be figuring out how to get the provider to talk to more than one Snowball.

Something else to think about, I didn't mention here, is how to support Snowball [VNICs](https://docs.aws.amazon.com/snowball/latest/developer-guide/using-ec2-edge-client.html). It may or may not make sense to expect Terraform to handle the creation of VNICs because it requires special Snowball credentials and the Snowball CLI (there's no publicly documented API). However, once the VNICs are created, you attach their "public IPs" (really just local private IPs) to EC2 the same way you'd attach an EIP in the cloud (associate-address).

Anyways, thanks for responding.

bflad:
Hi again @cb17 👋 We're going through older pull requests to close out ones that we won't be merged in their current state. As [mentioned previously](https://github.com/terraform-providers/terraform-provider-aws/pull/5926#issuecomment-434750706), let's open a new feature request issue so we can potentially work through the design and see if its feasible to support Snowball via the existing AWS provider/resources or if an entirely separate Terraform Provider should be considered in this case. 👍

PR #5929, opened 2018-09-19
Deprecate SSM Parameter `overwrite` attribute

### LINKED ISSUE #3181: Changing aws_ssm_parameter.overwrite trigger a resource update
### Terraform Version
Terraform v0.10.6

### Affected Resource(s)
- aws_ssm_parameter

### Expected Behavior
Changing the `overwrite` parameter from `true` to `false` or vice versa should not trigger an update

### Actual Behavior
Changing `overwrite` from `true` to `false` trigger an update which then fails because the value can not be updated with `overwrite` set to `false`

### Steps to Reproduce
1. Create a `aws_ssm_parameter` with `overwrite` set to `true`
2. Run `terraform apply`
3. change `overwrite` set to `false`
4. Run `terraform apply` again

### References
- GH-1481
- GH-1556

julienduchesne:
Fixes #3181

* Completely remove the weird SSM parameter overwrite logic
* It doesn't follow the terraform "logic", you shouldn't be able to overwrite existing resources that are not handled by terraform, there's import for that and there's no reason why you would want an error on apply when there's a simple change in the plan
* I have found no way to fix it properly, if you remove the `overwrite` attribute, it reverts back to `false` and seems to always remain in the state afterwards, if you DiffSuppress it, it acts weird and incoherent. It is just confusing right now.

```
TF_ACC=1 go test ./... -v -run=TestAccAWSSSMParameter -timeout 120m
=== RUN   TestAccAWSSSMParameter_importBasic
--- PASS: TestAccAWSSSMParameter_importBasic (23.20s)
--- PASS: TestAccAWSSSMParameter_basic (17.97s)
=== RUN   TestAccAWSSSMParameter_disappears
--- PASS: TestAccAWSSSMParameter_disappears (12.87s)
--- PASS: TestAccAWSSSMParameter_update (41.22s)
=== RUN   TestAccAWSSSMParameter_updateDescription
--- PASS: TestAccAWSSSMParameter_updateDescription (41.69s)
--- PASS: TestAccAWSSSMParameter_changeNameForcesNew (41.32s)
=== RUN   TestAccAWSSSMParameter_fullPath
--- PASS: TestAccAWSSSMParameter_fullPath (23.55s)
--- PASS: TestAccAWSSSMParameter_secure (20.03s)
--- PASS: TestAccAWSSSMParameter_secure_with_key (65.78s)
=== RUN   TestAccAWSSSMParameter_secure_keyUpdate
--- PASS: TestAccAWSSSMParameter_secure_keyUpdate (80.49s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       368.128s
```


website/docs/r/ssm_parameter.html.markdown
```diff
@@ -62,7 +62,6 @@ The following arguments are supported:
 * `value` - (Required) The value of the parameter.
 * `description` - (Optional) The description of the parameter.
 * `key_id` - (Optional) The KMS key id or arn for encrypting a SecureString.
-* `overwrite` - (Optional) Overwrite an existing parameter. If not specified, will default to `false` if the resource has not been created by terraform to avoid overwrite of existing resource and will default to `true` otherwise (terraform lifecycle rules should then be used to manage the update behavior).
```
joestump: I'd consider leaving this in the docs for a bit with a deprecation notice and a note to use the import functionality.

avoidik:
Currently we're affected by the overwrite edge-case issue, this PR could help us to address it. We had initially created SSM value with Terraform, then we have been rewriting the same SSM key with another automation script. In this case subsequent Terraform provisioning leads to the error below

```
* aws_ssm_parameter.secret: error creating SSM parameter: ParameterAlreadyExists: The parameter already exists. To overwrite this value, set the overwrite option in the request to true.
```

Lifecycle rule hasn't helped us much

```
  lifecycle {
    ignore_changes = ["value", "overwrite"]
  }
```

rmclaughlin-nelnet:
Will this allow having the parameter in Terraform, but update the value using a separate process?

avoidik:
YES!

bflad:
Hi @julienduchesne 👋  Thank you for submitting this and apologies for the lingering pull request review here.

The `overwrite` argument of the `aws_ssm_parameter` resource was introduced as a pragmatic solution for higher-than-usual use cases of conflicting management (e.g. by external systems). Practitioners are given the choice whether to accept the risks associated with the atypical resource management in Terraform. While this additional workflow option is atypical in the Terraform AWS Provider, we do have other precedence with similar functionality in the `aws_route53_record` resource, for similar reasons.

Rather than leaving this pull request open in a state where its unlikely to be merged in its current or similar form, I'm going to close this for now. The maintainers would be more willing to accept a proposed solution to #3181 that keeps the existing functionality but prevents unnecessary API calls if `overwrite` is the only state value that is updated. Otherwise, if you or others would like to further discuss this, please submit a proposal [feature request](https://github.com/terraform-providers/terraform-provider-aws/issues/new?labels=enhancement&template=Feature_Request.md) issue that highlights why this functionality needs to be removed.

PR #5933, opened 2018-09-19
Fixes #2384 aws_cloudwatch_log_metric_filter: default_value is automatically set to 0

### LINKED ISSUE #2384: aws_cloudwatch_log_metric_filter: default_value is automatically set to 0

ahl:
Fixes #2384

Note I looked at #2759 and #2902 which were both old and incomplete
cc: @jocelynthode @modax 

* use sentinel value of NaN to indicate that no value was specified
* clean up misspelling and cruft

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSCloudWatchLogMetricFilter_*'
TF_ACC=1 go test ./aws -v -run=TestAccAWSCloudWatchLogMetricFilter_* -timeout 120m
=== RUN   TestAccAWSCloudWatchLogMetricFilter_basic
--- PASS: TestAccAWSCloudWatchLogMetricFilter_basic (18.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	18.743s

```


ahl:
I should also have noted that a better solution is blocked on this issue: https://github.com/hashicorp/terraform/issues/16796

ahl:
maybe @bflad you could take a look?

paultyng:
I think we should implement this instead as a `TypeString` so that empty string can represent the unset value. We have precedent for this with `TypeBool`'s https://github.com/terraform-providers/terraform-provider-aws/blob/502d7b4cfa74d5a0715e6ba32f60d4ad58e2cbd7/aws/resource_aws_launch_template.go#L92-L95

It's definitely an annoying limitation of the current type system, and potentially something we can address post 0.12, but the backwards compatible implementation will probably just use a `TypeString` as well.

ahl:
Makes sense; I'll update the PR

ahl:
@jocelynthode I didn't know that was possible; thanks.

aws/validators.go
```diff
@@ -54,6 +54,22 @@ func validateTypeStringNullableBoolean(v interface{}, k string) (ws []string, es
...
+// validateTypeStringNullableFloat provides custom error messaging for TypeString floats
+// Some arguments require a floating point value or an unspecified, empty field.
+func validateTypeStringNullableFloat(v interface{}, k string) (ws []string, es []error) {
+	value, ok := v.(string)
+	if !ok {
+		es = append(es, fmt.Errorf("expected type of %s to be string", k))
+		return
+	}
+
+	if _, err := strconv.ParseFloat(value, 64); err != nil {
```
paultyng: Should there be a check for empty string in here? May help to do a little unit testing on this as there are some edges. This seems to error if empty string.

https://play.golang.org/p/iBa1Pq5V2tt

paultyng:
@ahl thanks for following up on this one!

ahl:
@paultyng it's pretty critical to us so I was motivated; thanks for sticking with me on it.

bflad:

Issue #4990, opened 2018-06-26
Amazon Pinpoint


marcoreni:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Amazon Pinpoint has been available for a while, even if only in in `us-east-1`.

Is it possible to add support for it into Terraform?

### New or Affected Resource(s)

I tried to extract a list of the resources that TF should support, taken from the aws-sdk-go api reference.

Basic resources:
* `aws_pinpoint_app` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/5956
* `aws_pinpoint_campaign`
* `aws_pinpoint_segment`
* `aws_pinpoint_event_stream` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6069

Channels: 
* `aws_pinpoint_adm_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6038
* `aws_pinpoint_apns_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/5956
* `aws_pinpoint_apns_sandbox_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6233
* `aws_pinpoint_apns_voip_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6234
* `aws_pinpoint_apns_voip_sandbox_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6235
* `aws_pinpoint_baidu_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6111
* `aws_pinpoint_email_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6110
* `aws_pinpoint_gcm_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6089
* `aws_pinpoint_sms_channel` :white_check_mark: https://github.com/terraform-providers/terraform-provider-aws/pull/6088

### References

[...]

PR #5956, opened 2018-09-21
New resource: aws_pinpoint_app

marcoreni:
This is the first step to address #4990 . 

More will (slowly) follow.

* New resource aws_pinpoint_app with related acceptance tests and docs

NOTE: `cloudwatch_metrics_enabled` is currently not returned by `GetApplicationSettings` requests so it's not currently handled inside the resource.

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSPinpointApp'
TF_ACC=1 go test ./aws -v -run=TestAccAWSPinpointApp -timeout 120m
=== RUN   TestAccAWSPinpointApp_basic
--- PASS: TestAccAWSPinpointApp_basic (19.40s)
=== RUN   TestAccAWSPinpointApp_CampaignHookLambda
--- PASS: TestAccAWSPinpointApp_CampaignHookLambda (41.61s)
=== RUN   TestAccAWSPinpointApp_Limits
--- PASS: TestAccAWSPinpointApp_Limits (17.05s)
=== RUN   TestAccAWSPinpointApp_QuietTime
--- PASS: TestAccAWSPinpointApp_QuietTime (24.47s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       103.733s
```


aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+
+	req := &pinpoint.CreateAppInput{
+		CreateApplicationRequest: &pinpoint.CreateApplicationRequest{
+			Name: aws.String(name),
+		},
+	}
+
+	output, err := pinpointconn.CreateApp(req)
+	if err != nil {
+		return fmt.Errorf("[ERROR] creating Pinpoint app: %s", err)
```
bflad: Nitpick: Using the hclog style `[ERROR]` can be replaced with the human readable `error` as it will be placed in the middle of the error messaging from Terraform to operators

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+}
+
+func resourceAwsPinpointAppDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).pinpointconn
+
+	log.Printf("[DEBUG] Pinpoint Delete App: %s", d.Id())
+	_, err := conn.DeleteApp(&pinpoint.DeleteAppInput{
+		ApplicationId: aws.String(d.Id()),
+	})
+	if err != nil {
```
bflad: We should add the following above this to ignore if its already been deleted:

```go
if isAWSErr(err, pinpoint.ErrCodeNotFoundException, "") {
  return nil
}
```

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+		return err
+	}
+	return nil
+}
+
+func expandCampaignHook(d *schema.ResourceData) *pinpoint.CampaignHook {
```
bflad: Function signature nitpicks (which apply to any of these expand/flatten methods):
* We generally prefer to include the service name in expand/flatten functions as the Go package structure of this project is very flat and may conflict with other service "campaign hooks". e.g. `expandPinpointCampaignHook`
* We prefer to reduce the input parameters where possible. In this case we should try to accept `[]interface{}` (once its been converted to `schema.TypeList`), e.g.

```go
appSettings.CampaignHook = expandPinpointCampaignHook(d.Get("campaign_hook").([]interface{}))
// ...
func expandPinpointCampaignHook(configs []interface{}) *pinpoint.CampaignHook {
  if configs == nil || len(configs) == 0 {
  // ...
```

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	l = append(l, m)
+
+	return l
+}
+
+// Assemble the hash for the aws_pinpoint_app campaignHook
+// TypeSet attribute.
+func campaignHookHash(v interface{}) int {
```
bflad: These `Hash` functions go away once the attributes are converted to `schema.TypeList` 👍

aws/resource_aws_pinpoint_app_test.go
```diff
@@ -0,0 +1,249 @@
+package aws
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/service/pinpoint"
```
bflad: Nitpick: We prefer to run [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) to automatically format imports, which would move this entry to the bottom imports section. Many editors support automatically running it on save.

aws/resource_aws_pinpoint_app_test.go
```diff
@@ -0,0 +1,249 @@
...
+					resource.TestCheckResourceAttr(resourceName, "limits.#", "0"),
+					resource.TestCheckResourceAttr(resourceName, "quiet_time.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "quiet_time.537668748.start", "00:00"),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckAWSPinpointAppExists(n string, attributes map[string]string) resource.TestCheckFunc {
```
bflad: It appears `attributes` is unused in this function. We prefer to use the second parameter for `Exists` test functions to return the API object returned by the get/describe/read function of the SDK so we can use it in later acceptance test check functions if necessary. e.g.

```go
func TestAccAWSPinpointApp_basic(t *testing.T) {
  var application pinpoint.ApplicationResponse
// ...
  testAccCheckAWSPinpointAppExists(resourceName, &application),
```

And its definition:

```go
func testAccCheckAWSPinpointAppExists(n string, application *pinpoint.ApplicationResponse) resource.TestCheckFunc {
// ...
params := &pinpoint.GetAppInput{
  ApplicationId: aws.String(rs.Primary.ID),
}
output, err := conn.GetApp(params)

if err != nil {
  return err
}

*application = *output.ApplicationResponse

return nil
```

aws/resource_aws_pinpoint_app_test.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		// Check if the topic exists by fetching its attributes
+		params := &pinpoint.GetAppInput{
+			ApplicationId: aws.String(rs.Primary.ID),
+		}
+		_, err := conn.GetApp(params)
+		if err != nil {
+			if isAWSErr(err, pinpoint.ErrCodeNotFoundException, "") {
+				return nil
```
bflad: Nitpick: While generally the testing configuration will only include one `aws_pinpoint_app`, we should use `continue` here instead of `return nil` in case it includes multiple applications. 👍

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+	configs := d.Get("campaign_hook").(*schema.Set).List()
+	if configs == nil || len(configs) == 0 {
+		return nil
+	}
+
+	m := configs[0].(map[string]interface{})
+
+	ch := &pinpoint.CampaignHook{}
+
+	if m["lambda_function_name"] != nil {
```
bflad: If you would to simplify these a little you can use this convention:

```go
if v, ok := m["lambda_function_name"]; ok {
  ch.LambdaFunctionName = aws.String(v.(string))
}
```

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+
+	return ch
+}
+
+func flattenCampaignHook(ch *pinpoint.CampaignHook) []interface{} {
+	l := make([]interface{}, 0)
+
+	m := map[string]interface{}{}
+
+	if ch.LambdaFunctionName != nil {
```
bflad: You can remove the `nil` checks by using the AWS Go SDK helpers here, e.g.

```go
m["lambda_function_name"] = aws.StringValue(ch.LambdaFunctionName)
```

Especially for `ch.Mode` below, which does not need the `!= ""` check.

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+			return nil
+		}
+
+		return err
+	}
+
+	d.Set("name", app.ApplicationResponse.Name)
+	d.Set("application_id", app.ApplicationResponse.Id)
+
+	d.Set("campaign_hook", flattenCampaignHook(settings.ApplicationSettingsResource.CampaignHook))
```
bflad: We should error check `d.Set()` when attempting to set more complex types like lists and maps to ensure they are properly being read into the Terraform state:

```go
if err := d.Set("campaign_hook", flattenCampaignHook(settings.ApplicationSettingsResource.CampaignHook)); err != nil {
  return fmt.Errorf("error setting campaign_hook: %s", err)
}
if err := d.Set("limits", flattenCampaignLimits(settings.ApplicationSettingsResource.Limits)); err != nil {
  return fmt.Errorf("error setting limits: %s", err)
}
if err := d.Set("quiet_time", flattenQuietTime(settings.ApplicationSettingsResource.QuietTime)); err != nil {
  return fmt.Errorf("error setting campaign_hook: %s", err)
}
```

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+	return &qt
+}
+
+func flattenQuietTime(qt *pinpoint.QuietTime) []interface{} {
+	l := make([]interface{}, 0)
+
+	m := map[string]interface{}{}
+
+	if qt.End != nil {
+		m["end"] = qt.End
```
bflad: I believe this would be found by the `d.Set()` error checking, but this should be converting the string pointer to a string:

```go
m["end"] = aws.StringValue(qt.End)
```

website/docs/r/pinpoint_app.markdown
```diff
@@ -0,0 +1,72 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `application_id` - The Application ID of the Pinpoint App.
+
+## Import
+
+CodeBuild Project can be imported using the `application-id`, e.g.
```
bflad: Copy-paste typo (`CodeBuild Project`) 😎

aws/resource_aws_pinpoint_app_test.go
```diff
@@ -0,0 +1,249 @@
...
+		IDRefreshName: "aws_pinpoint_app.test_app",
+		Providers:     testAccProviders,
+		CheckDestroy:  testAccCheckAWSPinpointAppDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSPinpointAppConfig_withGeneratedName,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSPinpointAppExists("aws_pinpoint_app.test_app", attributes),
+				),
+			},
```
bflad: We should add import `TestStep` to each of the acceptance tests, e.g.

```go
{
  ResourceName: resourceName,
  ImportState: true,
  ImportStateVerify: true,
},
```

aws/resource_aws_pinpoint_app.go
marcoreni: I've been struggling with this while testing the resource. The issue I've encountered is that `GetApplicationSettings` returns an empty object for `campaign_hook`, `limits` and `quiet_time` even if none of the children attributes is set. 

In this case, if I used 
```
m["lambda_function_name"] = aws.StringValue(ch.LambdaFunctionName)
m["mode"] = aws.StringValue(ch.Mode)
m["web_url"] = aws.StringValue(ch.WebUrl)
```
I'd get a map filled with empty values, but that would cause failures on the state diff (for example, on the `_basic` test):
```
        DIFF:

        UPDATE: aws_pinpoint_app.test_app
          campaign_hook.#: "1" => "0"

        STATE:

        aws_pinpoint_app.test_app:
          ID = ff3857dd45734e7ab07d9cb5141f0221
          provider = provider.aws
          application_id = ff3857dd45734e7ab07d9cb5141f0221
          campaign_hook.# = 1
          campaign_hook.0.lambda_function_name =
          campaign_hook.0.mode =
          campaign_hook.0.web_url =
          name = terraform-20180928131703355600000001
```

Any suggestions on how to handle this scenario?

website/docs/r/pinpoint_app.markdown
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/96756bf5c0cbb84b433765c3f0b76ac32261de06

aws/resource_aws_pinpoint_app_test.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/293a621cbff28eb95487ebda815f724fd57381a3 and https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/ea1e963c22edd9cc3720899eecedc4bd13f36c6d

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/commit/ea1e963c22edd9cc3720899eecedc4bd13f36c6d

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/1ea7cdae069f31c5179a80f274ba1282ba06785d

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/fd77a5b6e11241f502ea8741ff2fa12aef2410bf

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/fd77a5b6e11241f502ea8741ff2fa12aef2410bf

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+						},
+						"total": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"quiet_time": {
+				Type:     schema.TypeSet,
```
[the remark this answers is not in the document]
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/fd77a5b6e11241f502ea8741ff2fa12aef2410bf

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+						},
+						"web_url": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+					},
+				},
+			},
+			"limits": {
+				Type:     schema.TypeSet,
```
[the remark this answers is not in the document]
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/fd77a5b6e11241f502ea8741ff2fa12aef2410bf

aws/resource_aws_pinpoint_app.go
```diff
@@ -0,0 +1,435 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			//"cloudwatch_metrics_enabled": {
+			//	Type:     schema.TypeBool,
+			//	Optional: true,
+			//	Default:  false,
+			//},
+			"campaign_hook": {
+				Type:     schema.TypeSet,
```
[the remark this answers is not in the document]
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/fd77a5b6e11241f502ea8741ff2fa12aef2410bf

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/0b9d8ff9d5db243ccb6e43a65adcb1b36cbcc7b5

aws/resource_aws_pinpoint_app_test.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/be7dc0047e5e3ec5d966f2a51c702546a27cfb95

aws/resource_aws_pinpoint_app_test.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/be7dc0047e5e3ec5d966f2a51c702546a27cfb95

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/be7dc0047e5e3ec5d966f2a51c702546a27cfb95

aws/resource_aws_pinpoint_app.go
bflad: Great question! We'll be looking to come up with a better solution to this problem in the future, but for now, you can add the following `DiffSuppressFunc` to the `campaign_hook` attribute that ignores that the API is always returning the "container" attribute:

```go
  if old == "1" && new == "0" {
    return true
  }
  return false
},
```

aws/resource_aws_pinpoint_app.go
marcoreni: Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/d66ab63a1fa5f3ac8ae3598bc8d9717dd9f6a908

aws/resource_aws_pinpoint_app.go
marcoreni: Thanks! Totally missed `DiffSuppressFunc` while exploring the code looking for a workaround :D Fixed in https://github.com/terraform-providers/terraform-provider-aws/pull/5956/commits/4d398847e3e86408f66734b574bedfc086b6d493

marcoreni:
Hey @bflad , thanks for the thorough review and all the inputs! 

I applied all the requested changes. I committed them separately to ease the review. If you think it's better, I can squash them. 

Regarding `DiffSuppressFunc` I decided to repeat it on all the attributes to make it more explicit to a future reader. 

Let me know if there's anything else to be improved :)

PR #6033, opened 2018-10-01
tests/resource/aws_iam_user_login_profile: Remove usage of acctest.RandIntRange()

bflad:
Not sure why `acctest.RandIntRange()` would generate an integer outside its defined range 🤔 , but we also do not need a random integer for this test.

* Hardcode expected password length in `TestAccAWSUserLoginProfile_PasswordLength`

Previously:

```
--- FAIL: TestAccAWSUserLoginProfile_PasswordLength (0.50s)
    testing.go:527: Step 0 error: config is invalid: aws_iam_user_login_profile.user: expected password_length to be in the range (5 - 128), got 1
```

```
--- PASS: TestAccAWSUserLoginProfile_PasswordLength (13.13s)
```

### DIFF
--- a/aws/resource_aws_iam_user_login_profile_test.go
+++ b/aws/resource_aws_iam_user_login_profile_test.go
@@ -3,7 +3,6 @@ package aws
 import (
 	"errors"
 	"fmt"
-	"strconv"
 	"testing"
 	"time"
 
@@ -136,7 +135,6 @@ func TestAccAWSUserLoginProfile_notAKey(t *testing.T) {
 func TestAccAWSUserLoginProfile_PasswordLength(t *testing.T) {
 	var conf iam.GetLoginProfileOutput
 
-	passwordLength := acctest.RandIntRange(4, 128)
 	username := fmt.Sprintf("test-user-%d", acctest.RandInt())
 
 	resource.Test(t, resource.TestCase{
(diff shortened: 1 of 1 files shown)


paultyng:
Should we add a test that checks the minimum value as well? Or maybe a value outside the range?

bflad:
Generally speaking for these. I see both sides of it, but its always going to be a cost vs effort discussion, especially given this would be a new requirement for all future pull requests along with backfilling this testing (in most resources, we do not have the acceptance tests structured for handling every single attribute). `ValidateFunc` implementations generally fall under a more of a "nice to have" category more than a "this must absolutely work" category given its more of a UX helper.  😄  I can certainly add these acceptance test/TestStep if you wish though in the future.

> Should we add a test that checks the minimum value as well?

We typically do not require multiple acceptance tests strictly for attribute `ValidateFunc` testing unless the values will implicitly have other consequences. e.g. its encouraged to check an `aws_directory_service_directory` resource `type = "MicrosoftAD"` vs `type = "SimpleAD"` since this will likely have multiple code paths/configuration options, but its not currently required to check an integer value for something that supports an integer range.

In this case, whether the password length is 5, 42, 99, or 128 _should_ (😅 ) be inconsequential in terms of this resource, otherwise one might argue that we should be checking _every_ value in the range should the code or upstream validation function somehow be updated to only accept 5 and 128 or have some sort of bug in between. 🙃 

> Or maybe a value outside the range?

I'm not sure we need to _acceptance_ test that `ValidateFunc` returns an error (based on `regexp`) properly in every resource/attribute that implements a `ValidateFunc` function, especially those that singularly implement a function from the upstream `helper/validation` package. That is, unless we want to be very pedantic about our test suite and always verify upper + 1 and lower - 1 boundaries in the case of `validation.IntBetween()` and always force a test update/failure should it change. 🤔

bflad:
I think its important to note above as well that _unit_ testing every attribute `ValidateFunc` is actually a very reasonable ask, even against those using `helper/validation` functions. 💯

PR #6035, opened 2018-10-01
tests/resource/aws_iam_saml_provider: Refactor and enhance testing

bflad:

* Combine importBasic acceptance test and basic acceptance test
* Randomize naming
* Additionally check Terraform state for name and saml_metadata_document attributes

Previously:

```
--- FAIL: TestAccAWSIAMSamlProvider_basic (1.81s)
    testing.go:527: Step 0 error: Error applying: 1 error occurred:
        	* aws_iam_saml_provider.salesforce: 1 error occurred:
        	* aws_iam_saml_provider.salesforce: EntityAlreadyExists: Provider with name tf-salesforce-test already exists.
```

```
--- PASS: TestAccAWSIAMSamlProvider_basic (16.06s)
```

### DIFF
--- a/aws/resource_aws_iam_saml_provider_test.go
+++ b/aws/resource_aws_iam_saml_provider_test.go
@@ -5,14 +5,15 @@ import (
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 )
 
-func TestAccAWSIAMSamlProvider_importBasic(t *testing.T) {
-	resourceName := "aws_iam_saml_provider.salesforce"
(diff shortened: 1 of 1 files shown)


aws/resource_aws_iam_saml_provider_test.go
```diff
@@ -5,51 +5,42 @@ import (
...
-			{
-				ResourceName:      resourceName,
-				ImportState:       true,
-				ImportStateVerify: true,
-			},
-		},
-	})
-}
-
 func TestAccAWSIAMSamlProvider_basic(t *testing.T) {
```
paultyng: super minor nit, but I think `SAML` should be capitalized like `AWS` and `IAM`

bflad: Enabling `golint` would be an awesome goal (likely something for a 2.0+ cycle), but we'll need to look into ways we can enforce our own initialisms into it after fixing the current failures and ignoring purposeful things like the acceptance test function naming with underscores 😅 .  Without some form of enforcement, it could become a never ending project or very manual pain point during PR reviews. `SAML` unfortunately isn't one of the builtin initialisms and we'd also really want to add `AWS`, `EC2`, etc. 

I think we'll get there eventually, but we need to decide on a path forward that makes sense with the available tooling or work towards improving the tooling. 👍

PR #6041, opened 2018-10-02
Use current instance types in examples.

jmcarp:
For users who hypothetically copy and paste config from docs, let's reference current instance types so that nobody is using `r3`s instead of `r4`s.

### DIFF
--- a/website/docs/r/elasticache_cluster.html.markdown
+++ b/website/docs/r/elasticache_cluster.html.markdown
@@ -27,7 +27,7 @@ See the AWS Docs on [Modifying an ElastiCache Cache Cluster][2] for more informa
 resource "aws_elasticache_cluster" "example" {
   cluster_id           = "cluster-example"
   engine               = "memcached"
-  node_type            = "cache.m3.medium"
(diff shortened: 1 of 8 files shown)


bflad:
@jmcarp thanks for submitting this. Have you verified all these examples? In particular, I'd be a little concerned with:
* `aws_elasticache_replication_group` and t2 (t2 has many restrictions in Elasticache in general) 
* m5 against specific AMI IDs

website/docs/r/elasticache_cluster.html.markdown
```diff
@@ -27,7 +27,7 @@ See the AWS Docs on [Modifying an ElastiCache Cache Cluster][2] for more informa
 resource "aws_elasticache_cluster" "example" {
   cluster_id           = "cluster-example"
   engine               = "memcached"
-  node_type            = "cache.m3.medium"
+  node_type            = "cache.m4.medium"
```
bflad: `cache.m4.medium` is not a valid instance type -- `cache.m4.large` is however

PR #6059, opened 2018-10-03
resource/aws_waf_web_acl: Add logging configuration

### LINKED ISSUE #5760: Feature Request: Support Comprehensive Logging Functionality for AWS WAF
### Community Note

### Description

AWS released a feature for AWS WAF which allows wiring up Kinesis Firehose stream to route all the logs about incoming requests and matched rules. 

https://aws.amazon.com/about-aws/whats-new/2018/08/aws-waf-launches-new-comprehensive-logging-functionality/

<details>
<summary>Screenshot</summary>

image

</details>

* aws_waf_web_acl
* aws_wafregional_web_acl

```hcl
resource "aws_waf_web_acl" "waf_acl" {
  depends_on = [
    "aws_waf_ipset.ipset",
    "aws_waf_rule.wafrule",
  ]

  name        = "tfWebACL"
  metric_name = "tfWebACL"

  logging {
    firehose_stream_arn = "${aws_kinesis_firehose_delivery_stream.test.arn}"

    redacted_fields {
      http_method  = true
      query_string = true
      uri          = true

      headers = [
        "Accept",
        "Content-Type",
      ]
    }
  }

  default_action {
    type = "ALLOW"
  }

  rules {
    action {
      type = "BLOCK"
    }

    priority = 1
    rule_id  = "${aws_waf_rule.wafrule.id}"
    type     = "REGULAR"
  }
}

```

WhileLoop:
Fixes #5760

* Add support for logging configuration for WAF Web ACL.

**Potential Terraform Configuration**
```
resource "aws_waf_web_acl" "waf_acl" {
  name        = "test"
  metric_name = "test"

  logging {
    log_destination = "${aws_kinesis_firehose_delivery_stream.waf_stream.arn}"

    redacted_fields {

      field_to_match {
        type = "QUERY_STRING"
      }

      field_to_match {
        type = "METHOD"
      }

      field_to_match {
        type = "URI"
      }

      field_to_match {
        type = "HEADER"
        data = "referer"
      }

    }
  }

  default_action {
    type = "ALLOW"
  }

  rules {
    action {
      type = "BLOCK"
    }

    priority = 1
    rule_id  = "${aws_waf_rule.wafrule.id}"
    type     = "REGULAR"
  }
}
```

### DIFF
--- a/aws/resource_aws_waf_web_acl.go
+++ b/aws/resource_aws_waf_web_acl.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/waf"
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
@@ -21,6 +22,10 @@ func resourceAwsWafWebAcl() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
@@ -44,6 +49,44 @@ func resourceAwsWafWebAcl() *schema.Resource {
 				Required:     true,
 				ForceNew:     true,
 				ValidateFunc: validateWafMetricName,
+			},
+			"logging_configuration": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"log_destination": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"redacted_fields": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"field_to_match": {
+										Type:     schema.TypeSet,
+										Required: true,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"data": {
+													Type:     schema.TypeString,
+													Optional: true,
+												},
+												"type": {
+													Type:     schema.TypeString,
+													Required: true,
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
 			},
 			"rules": {
 				Type:     schema.TypeSet,
@@ -120,6 +163,16 @@ func resourceAwsWafWebAclCreate(d *schema.ResourceData, meta interface{}) error 
 	}
 	resp := out.(*waf.CreateWebACLOutput)
 	d.SetId(*resp.WebACL.WebACLId)
+
+	arn := arn.ARN{
+		Partition: meta.(*AWSClient).partition,
+		Service:   "waf",
+		AccountID: meta.(*AWSClient).accountid,
+		Resource:  fmt.Sprintf("webacl/%s", d.Id()),
+	}.String()
+
+	// Set for update
+	d.Set("arn", arn)
 	return resourceAwsWafWebAclUpdate(d, meta)
 }
 
@@ -146,6 +199,14 @@ func resourceAwsWafWebAclRead(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 
+	arn := arn.ARN{
+		Partition: meta.(*AWSClient).partition,
+		Service:   "waf",
+		AccountID: meta.(*AWSClient).accountid,
+		Resource:  fmt.Sprintf("webacl/%s", d.Id()),
+	}.String()
+	d.Set("arn", arn)
+
 	if err := d.Set("default_action", flattenWafAction(resp.WebACL.DefaultAction)); err != nil {
 		return fmt.Errorf("error setting default_action: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


WhileLoop:
@bflad WIP, I have some questions about this before I keep working on it. 

- When does it make sense to use TypeSet vs. TypeList, can you take a look at the schema?
- How clean up expandLoggingConfiguration?
- Should the field types use the all upper case AWS values and expandFieldToMatch or keep the current way?
- Can you show me an example of how to implement flattenLoggingConfiguration in this case?

Thanks

bflad:
> When does it make sense to use TypeSet vs. TypeList, can you take a look at the schema?

The great debate! We have some information about this here: https://www.terraform.io/docs/extend/schemas/schema-types.html#aggregate-types

But essentially it boils down to:
* Is it a "container" attribute (e.g. so you can make a HCL configuration block containing attributes of various `Type`s inside it) -- if so it should be `Type: schema.TypeList` and `MaxItems: 1`
* Does ordering matter? If unsure, err on the side that it should, choose `Type: schema.TypeList` (we cannot automatically migrate `TypeSet` -> `TypeList` should there be issues). If yes, choose `Type: schema.TypeList`. If no, _maybe_ choose `Type: schema.TypeSet`.

Basically, just prefer to use `TypeList` unless necessary (I didn't give other exhaustive examples for `TypeSet` on purpose). 😅 

> How clean up expandLoggingConfiguration?

Sorry I don't have time to provide an extensive review right now, but I would take a look at other existing `expand` functions. If the SDK object contains other SDK objects (e.g. `waf.Thing1` has an attribute with `waf.Thing2`), that's generally a good sign that another `expand` function is needed. 👍 

> Should the field types use the all upper case AWS values and expandFieldToMatch or keep the current way?

Without deeply looking into this, we generally _prefer_ to follow the API/SDK unless it really enhances user experience. It can become a maintenance burden to create our own translations of concepts, especially if additional API fields are added in the future. I would recommend making the `expandWAFFieldToMatch` method if the implementation is switched to something like a `field_to_match` list attribute.

> Can you show me an example of how to implement flattenLoggingConfiguration in this case?

It will depend on the implementation of the schema. I would recommend looking at other existing `flatten` functions for inspiration.

bflad:
@WhileLoop did you have other questions or will you have time to keep working on this?

WhileLoop:
@bflad Forgot all about this tbh. Almost done. On line 204 I use `resp2`, is there a better way to do that?

WhileLoop:
@bflad I'm stuck Can't get past `error setting logging configuration: Invalid address to set: []string{"logging_configuration", "0", "redacted_fields"}`. Any pointers?

bflad:
@WhileLoop it looks like the `redacted_fields` attribute is currently `Type: schema.TypeSet` -- in order to correctly set it in the state you would need to use [`schema.NewSet()`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#NewSet)

My recommendation would be to change that attribute to `Type: schema.TypeList` and `MaxItems: 1` if it is only used to define a single configuration block.

WhileLoop:
`redacted_fields` can have multiple `field_to_match` blocks.

anilkasu:
Any update on this enhancement, when can we expect to release this? Any chances including this feature in v0.12?

aws/resource_aws_waf_web_acl.go
```diff
@@ -219,3 +308,48 @@ func resourceAwsWafWebAclDelete(d *schema.ResourceData, meta interface{}) error
...
+	}
+
+	m := make(map[string]interface{})
+	m["log_destination"] = *n.LogDestinationConfigs[0]
+
+	m["redacted_fields"] = flattenRedactedFields(n.RedactedFields)
+	return []map[string]interface{}{m}
+}
+
+func flattenRedactedFields(ts []*waf.FieldToMatch) []interface{} {
```
bflad: Here is a working example of this function for configurations with and without `redacted_fields`:

```go
func flattenRedactedFields(fieldToMatches []*waf.FieldToMatch) []interface{} {
	if len(fieldToMatches) == 0 {
		return []interface{}{}
	}

	fieldToMatchResource := &schema.Resource{
			"data": {
				Type:     schema.TypeString,
				Optional: true,
			},
			"type": {
				Type:     schema.TypeString,
				Required: true,
			},
		},
	}
	l := make([]interface{}, len(fieldToMatches))

	for i, fieldToMatch := range fieldToMatches {
		l[i] = flattenFieldToMatch(fieldToMatch)[0]
	}

	m := map[string]interface{}{
		"field_to_match": schema.NewSet(schema.HashResource(fieldToMatchResource), l),
	}

	return []interface{}{m}
}
```

aws/resource_aws_waf_web_acl_test.go
```diff
@@ -275,6 +275,80 @@ resource "aws_waf_web_acl" "test" {
...
+      "Effect": "Allow",
+      "Sid": ""
+    }
+  ]
+}
+EOF
+}
+
+resource "aws_kinesis_firehose_delivery_stream" "test_stream" {
+  name        = "terraform-kinesis-firehose-waftest-%d"
```
bflad: The name for WAF logging Kinesis Firehose Delivery Streams must begin with `aws-waf-logs-`

```suggestion
  name        = "aws-waf-logs-%s"
```

bflad:
If you need further examples of anything, you might find the WAF Regional equivalent of this pull request (https://github.com/terraform-providers/terraform-provider-aws/pull/7480) helpful. 👍

WhileLoop:
Thanks @bflad. The expand/flatten functions for loggingConfiguration and redactedFields can probably be shared by regional and global WAF resource.

bflad:
> The expand/flatten functions for loggingConfiguration and redactedFields can probably be shared by regional and global WAF resource.

@WhileLoop most likely, but we can handle that after getting the support in. 😄 The WAF service code has quite a few issues with code structure and naming due to a large portion of it being migrated from hashicorp/terraform pull requests right after Terraform providers were split from the single repository.

---

Let us know when you think this is ready for another review and we'll take a look. 👍

WhileLoop:
@bflad this is ready to go.

```$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSWafWebAcl_LoggingConfiguration'
TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccAWSWafWebAcl_LoggingConfiguration -timeout 120m
=== RUN   TestAccAWSWafWebAcl_LoggingConfiguration
=== PAUSE TestAccAWSWafWebAcl_LoggingConfiguration
=== CONT  TestAccAWSWafWebAcl_LoggingConfiguration
--- PASS: TestAccAWSWafWebAcl_LoggingConfiguration (100.49s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	101.497s```

bflad:

PR #6070, opened 2018-10-04
Adding secret_binary to aws_secretsmanager_secret_version

### LINKED ISSUE #4571: add secret_binary attribute to aws_secretsmanager_secret_version data source
### Community Note

### Description

Thank you for providing the aws_secretsmanager_secret_version data source for us to use. I have found that I cannot store multiline strings, such as ssh keys in the SecretString property as aws mangles them. I can however store them in the SecretBinary property without issue but Terraform doesn't support that property yet.

Please add a secret_binary attribute to the aws_secretsmanager_secret_version data source. 

The resource with the same name could also be updated but frankly I don't think terraform should be managing secret values because that requires storing the secrets somewhere else before giving them to terraform which not only duplicates some of the functionality of secrets manager but breaks other aspects of it like secret rotation.

* data aws_secretsmanager_secret_version
* resource aws_secretsmanager_secret_version

```hcl
variable "ssh_key" {}

resource "aws_secretsmanager_secret" "ssh_key" {
  name        = "ssh-key"
  description = "the ssh private key for logging into the bastion as the sqitch user"
  secret_binary = "${variable.ssh_key.value}"
}

data "aws_secretsmanager_secret" "ssh_key" {
  name = "ssh-key"
}

data "aws_secretsmanager_secret_version" "ssh_key" {
  secret_id = "${data.aws_secretsmanager_secret.ssh_key.id}"
}

output "ssh_key" {
  value = "${data.aws_secretsmanager_secret_version.ssh_key.secret_binary}"
}
```

### References

* https://docs.aws.amazon.com/cli/latest/reference/secretsmanager/create-secret.html
* https://docs.aws.amazon.com/cli/latest/reference/secretsmanager/get-secret-value.htm

Eraac:
Fixes #4571

* Adding data `secret_binary` for data "aws_secretsmanager_secret_version"
* Adding input `secret_binary` for resource "aws_secretsmanager_secret_version"

```
$ make testacc TESTARGS='-run=TestAccAwsSecretsManagerSecretVersion'
TF_ACC=1 go test ./aws -v -run=TestAccAwsSecretsManagerSecretVersion -timeout 120m
=== RUN   TestAccAwsSecretsManagerSecretVersion_BasicString
--- PASS: TestAccAwsSecretsManagerSecretVersion_BasicString (12.46s)
=== RUN   TestAccAwsSecretsManagerSecretVersion_BasicBinary
--- PASS: TestAccAwsSecretsManagerSecretVersion_BasicBinary (10.40s)
=== RUN   TestAccAwsSecretsManagerSecretVersion_VersionStages
--- PASS: TestAccAwsSecretsManagerSecretVersion_VersionStages (26.18s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	49.086s
...
```

First time i contribute to a terraform provider, i hope to do well. 

I have just few doubt about the code for the resource (for the cast []byte <-> string). For the test i only copy/paste the test for `secret_string`

### DIFF
--- a/aws/resource_aws_secretsmanager_secret_version.go
+++ b/aws/resource_aws_secretsmanager_secret_version.go
@@ -31,10 +31,18 @@ func resourceAwsSecretsManagerSecretVersion() *schema.Resource {
 				ForceNew: true,
 			},
 			"secret_string": {
-				Type:      schema.TypeString,
-				Required:  true,
-				ForceNew:  true,
-				Sensitive: true,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				Sensitive:     true,
+				ConflictsWith: []string{"secret_binary"},
+			},
+			"secret_binary": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				Sensitive:     true,
+				ConflictsWith: []string{"secret_string"},
 			},
 			"version_id": {
 				Type:     schema.TypeString,
@@ -55,8 +63,15 @@ func resourceAwsSecretsManagerSecretVersionCreate(d *schema.ResourceData, meta i
 	secretID := d.Get("secret_id").(string)
 
 	input := &secretsmanager.PutSecretValueInput{
-		SecretId:     aws.String(secretID),
-		SecretString: aws.String(d.Get("secret_string").(string)),
+		SecretId: aws.String(secretID),
+	}
+
+	if v, ok := d.GetOk("secret_string"); ok {
+		input.SecretString = aws.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("secret_binary"); ok {
+		input.SecretBinary = []byte(v.(string))
 	}
 
 	if v, ok := d.GetOk("version_stages"); ok {
@@ -105,6 +120,7 @@ func resourceAwsSecretsManagerSecretVersionRead(d *schema.ResourceData, meta int
 
 	d.Set("secret_id", secretID)
 	d.Set("secret_string", output.SecretString)
+	d.Set("secret_binary", fmt.Sprintf("%s", output.SecretBinary))
 	d.Set("version_id", output.VersionId)
 	d.Set("arn", output.ARN)
 
--- a/aws/resource_aws_secretsmanager_secret_version.go
+++ b/aws/resource_aws_secretsmanager_secret_version.go
@@ -76,7 +76,7 @@ func resourceAwsSecretsManagerSecretVersionCreate(d *schema.ResourceData, meta i
 		vs := []byte(v.(string))
 
 		if !isBase64Encoded(vs) {
-			fmt.Errorf("expected base64 in secret_binary")
+			return fmt.Errorf("expected base64 in secret_binary")
 		}
 
 		var err error
--- a/aws/resource_aws_secretsmanager_secret_version.go
+++ b/aws/resource_aws_secretsmanager_secret_version.go
@@ -5,6 +5,8 @@ import (
 	"log"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


bflad:
Hi @Eraac 👋  Thanks for submitting this! When working with binary data in Terraform, it should be base64 encoded and decoded to prevent Terraform state corruption. Are you able to update the handling to include that? Otherwise this is looking pretty good so far. 👍

Eraac:
@bflad https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#GetSecretValueOutput according to documentation is automatically base64 encoded/decoded by the SDK

Eraac:
I probably misunderstood, you mean for the terraform himself ?

So user should insert data as base64 and I decode it in the plugin ? What about output, should be decoded or not ?

bflad:
> So user should insert data as base64 and I decode it in the plugin ?

Yes, sorry if I was unclear! Basically, we should require something like `"${base64encode(file("path/to/binaryfile"))}"` or `"${base64gzip(file("path/to/binaryfile"))}"` in configurations so that there are no issues with parsing configurations or saving the binary data in the Terraform state as a UTF-8 string.

If the AWS SDK already handles the conversion correct, that's great! We'll just want to make sure the acceptance test is updated to perform the above on some binary content (there are some zip files in `aws/test-fixtures` that can be used here) and that the documentation notes the required base64 input  and likely mentioning the [`base64encode()` function](https://www.terraform.io/docs/configuration/interpolation.html#base64encode-string-) available in Terraform.

> What about output, should be decoded or not ?

It should be base64 encoded. If people need to get the actual binary contents, they can use [`base64decode()` function](https://www.terraform.io/docs/configuration/interpolation.html#base64decode-string-)

Eraac:
I take a look to other resources using base64 and your response confirm. Thanks for the precision 

I push new commit with all the changes needed (code, test and documentation), should be good now :proud:

website/docs/d/secretsmanager_secret_version.html.markdown
```diff
@@ -42,4 +42,5 @@ data "aws_secretsmanager_secret_version" "by-version-stage" {
 * `arn` - The ARN of the secret.
 * `id` - The unique identifier of this version of the secret.
 * `secret_string` - The decrypted part of the protected secret information that was originally provided as a string.
+* `secret_binary` - The decrypted part of the protected secret information that was originally provided as a binary.
```
bflad: We should note this is base64 encoded. I'll make the very minor edit post-merge. 😄

aws/resource_aws_secretsmanager_secret_version.go
```diff
@@ -55,8 +65,26 @@ func resourceAwsSecretsManagerSecretVersionCreate(d *schema.ResourceData, meta i
...
+	}
+
+	if v, ok := d.GetOk("secret_string"); ok {
+		input.SecretString = aws.String(v.(string))
+	}
+
+	if v, ok := d.GetOk("secret_binary"); ok {
+		vs := []byte(v.(string))
+
+		if !isBase64Encoded(vs) {
```
bflad: Not a blocker for merging, but I believe we could enhance this check to actually be done during plan with a `ValidateFunc` on the `secret_binary` attribute:

```go
ValidateFunc: func(v interface{}, name string) (warns []string, errs []error) {
	s := v.(string)
	if !isBase64Encoded([]byte(s)) {
		errs = append(errs, fmt.Errorf("%s: must be base64-encoded", name))
	}
	return
},
```

At some point we can migrate the existing attributes that use this (or a similar) ValidateFunc for ensure Base64 content into its own, easy to use common function like `validateBase64String`. 👍

bflad:

PR #6083, opened 2018-10-05
tests/resource/aws_db_instance: Consolidate TestAccAWSDBInstance_basic and enhance

bflad:
I was updating the test as part of investigating a potential bug, which didn't turn out to be anything, but might as well keep the updated test. 😄 

* Consolidate `_importBasic` acceptance test into `_basic` acceptance test
* Check additional attributes in Terraform state

```
--- PASS: TestAccAWSDBInstance_basic (438.47s)
```

### DIFF
--- a/aws/resource_aws_db_instance_test.go
+++ b/aws/resource_aws_db_instance_test.go
@@ -81,7 +81,8 @@ func testSweepDbInstances(region string) error {
 	return nil
 }
 
-func TestAccAWSDBInstance_importBasic(t *testing.T) {
+func TestAccAWSDBInstance_basic(t *testing.T) {
+	var dbInstance1 rds.DBInstance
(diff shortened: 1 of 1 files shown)


aws/resource_aws_db_instance_test.go
```diff
@@ -1932,6 +1924,17 @@ func testAccCheckAWSDBInstanceExists(n string, v *rds.DBInstance) resource.TestC
 	}
 }
 
+func testAccAWSDBInstanceImportStateIdFuncARN(resourceName string) resource.ImportStateIdFunc {
```
tombuildsstuff: I don't see this method used anywhere?

bflad: Heh whoops -- that was related to what I was looking into. I'll remove it! 👍

PR #6087, opened 2018-10-08
Support for setting of certificate in `aws_codebuild_project`

### LINKED ISSUE #4742: Can not set certificate with aws_codebuild_project?
## Description
In aws code build, when the source type is github enterprise, It is necessary to set the github enterprise certificate,
I can not set a certificate & github enterprise access token in aws_codebuild_project of terraform.
Is there any solution?

## Affected Resource
aws_codebuild_project

tkbky:
Fixes #4742 

This change allows the certificate to be set for an `aws_codebuild_project`.

```
$ make testacc TESTARGS='-run= TestAccAWSCodeBuildProject_Environment_Certificate'

TF_ACC=1 go test ./... -v -run=TestAccAWSCodeBuildProject_Environment_Certificate -timeout 120m
=== RUN   TestAccAWSCodeBuildProject_Environment_Certificate
--- PASS: TestAccAWSCodeBuildProject_Environment_Certificate (79.98s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	80.019s
...
```

### DIFF
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -992,7 +992,7 @@ func flattenAwsCodeBuildProjectEnvironment(environment *codebuild.ProjectEnviron
 	envConfig["type"] = *environment.Type
 	envConfig["compute_type"] = *environment.ComputeType
 	envConfig["image"] = *environment.Image
-	envConfig["certificate"] = *environment.Certificate
+	envConfig["certificate"] = aws.StringValue(environment.Certificate)
 	envConfig["privileged_mode"] = *environment.PrivilegedMode
 
 	if environment.EnvironmentVariables != nil {
@@ -1074,13 +1074,11 @@ func resourceAwsCodeBuildProjectEnvironmentHash(v interface{}) int {
 	environmentType := m["type"].(string)
 	computeType := m["compute_type"].(string)
 	image := m["image"].(string)
-	certificate := m["certificate"].(string)
 	privilegedMode := m["privileged_mode"].(bool)
 	environmentVariables := m["environment_variable"].([]interface{})
 	buf.WriteString(fmt.Sprintf("%s-", environmentType))
 	buf.WriteString(fmt.Sprintf("%s-", computeType))
 	buf.WriteString(fmt.Sprintf("%s-", image))
-	buf.WriteString(fmt.Sprintf("%s-", certificate))
 	buf.WriteString(fmt.Sprintf("%t-", privilegedMode))
 	for _, e := range environmentVariables {
 		if e != nil { // Old statefiles might have nil values in them
--- a/aws/resource_aws_codebuild_project.go
+++ b/aws/resource_aws_codebuild_project.go
@@ -177,6 +177,10 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 							Optional: true,
 							Default:  false,
 						},
+						"certificate": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
 					},
 				},
 				Set: resourceAwsCodeBuildProjectEnvironmentHash,
@@ -619,6 +623,10 @@ func expandProjectEnvironment(d *schema.ResourceData) *codebuild.ProjectEnvironm
 		projectEnv.Type = aws.String(v.(string))
 	}
 
+	if v := envConfig["certificate"]; v != nil {
+		projectEnv.Certificate = aws.String(v.(string))
+	}
+
 	if v := envConfig["environment_variable"]; v != nil {
 		envVariables := v.([]interface{})
 		if len(envVariables) > 0 {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_codebuild_project.go
```diff
@@ -984,6 +992,7 @@ func flattenAwsCodeBuildProjectEnvironment(environment *codebuild.ProjectEnviron
 	envConfig["type"] = *environment.Type
 	envConfig["compute_type"] = *environment.ComputeType
 	envConfig["image"] = *environment.Image
+	envConfig["certificate"] = *environment.Certificate
```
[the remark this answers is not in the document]
tkbky: Sure, will fix this up.

tkbky:
@bflad I've managed to set up the acceptance test for the certificate test case. Previously, it fails other acceptance tests due to the certificate was set to empty string which raises the invalid extension error.

aws/resource_aws_codebuild_project.go
```diff
@@ -177,6 +177,10 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 							Optional: true,
 							Default:  false,
 						},
+						"certificate": {
```
bflad: Since the CodeBuild API seems to validate against `.pem` and `.zip` endings, we can implement plan-time validation for that via:

```go
ValidateFunc: validation.StringMatch(regexp.MustCompile(`\.(pem|zip)$`), "must end in .pem or .zip"),
```

Also, since this attribute lives inside a `Type: schema.TypeSet` attribute, there is a `Set` function which is used to detect changes. `resourceAwsCodeBuildProjectEnvironmentHash` in this case.

We can add the following there:

```go
if v, ok := m["certificate"]; ok && v.(string) != "" {
	buf.WriteString(fmt.Sprintf("%s-", v.(string)))
}
```

At some point we'll likely be converting many of these `TypeSet` attributes, due to their unnecessary complexity.

aws/resource_aws_codebuild_project_test.go
```diff
@@ -286,6 +286,29 @@ func TestAccAWSCodeBuildProject_Environment_EnvironmentVariable_Type(t *testing.
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSCodeBuildProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSCodeBuildProjectConfig_Environment_Certificate(rName, bName, oName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSCodeBuildProjectExists(resourceName, &project),
+					resource.TestCheckResourceAttr(resourceName, "environment.1974383098.certificate", fmt.Sprintf("%s/%s", bName, oName)),
```
bflad: Since the `TypeSet` value wasn't dependent on the `certificate` value previously, the `1974383098` value was static. Now that `resourceAwsCodeBuildProjectEnvironmentHash` has been updated and since we randomly generate the S3 bucket name, the hash will never be consistent. Rather than trying to recompute the hash value for this test, we can instead just verify the API is set correctly via:

```go
testAccCheckAWSCodeBuildProjectCertificate(&project, fmt.Sprintf("%s/%s", bName, oName)),
```

and its definition:

```go
func testAccCheckAWSCodeBuildProjectCertificate(project *codebuild.Project, expectedCertificate string) resource.TestCheckFunc {
		if aws.StringValue(project.Environment.Certificate) != expectedCertificate {
			return fmt.Errorf("CodeBuild Project certificate (%s) did not match: %s", aws.StringValue(project.Environment.Certificate), expectedCertificate)
		}
		return nil
	}
}
```

aws/resource_aws_codebuild_project.go
```diff
@@ -619,6 +623,10 @@ func expandProjectEnvironment(d *schema.ResourceData) *codebuild.ProjectEnvironm
 		projectEnv.Type = aws.String(v.(string))
 	}
 
+	if v := envConfig["certificate"]; v != nil {
```
bflad: To prevent the Terraform resource from always triggering this error when not set in the Terrraform configuration:

```
--- FAIL: TestAccAWSCodeBuildProject_basic (6.15s)
    testing.go:527: Step 0 error: Error applying: 1 error occurred:
        	* aws_codebuild_project.test: 1 error occurred:
        	* aws_codebuild_project.test: Error creating CodeBuild project: InvalidInputException: Invalid extension: certificate must be either .pem or .zip
        	status code: 400, request id: bc162507-cbe4-11e8-aeb7-db9c8472a837
```

We can use this instead:

```go
if v, ok := envConfig["certificate"]; ok && v.(string) != "" {
	projectEnv.Certificate = aws.String(v.(string))
}
```

tkbky:
@bflad Thank you for your time go through this PR and all your comments are valuable to my learning, really appreciate that. 😃

bflad:

PR #6102, opened 2018-10-09
Fix AWS lb listener authentication documentation

fotos:
Followup to #6094 which was merged and [will be released in `1.40.0`](https://github.com/terraform-providers/terraform-provider-aws/issues/4707#issuecomment-428039238).

* Added `authenticate-cognito` action example
* Added `authenticate-oidc` action example
* Added `authenticate-cognito` and `authenticate-oidc` in valid values for `type`
* Renamed `user_pool_client` to `user_pool_client_id` to match the code

The examples are based on the tests of #6094.

### DIFF
--- a/website/docs/r/lb_listener.html.markdown
+++ b/website/docs/r/lb_listener.html.markdown
@@ -101,7 +101,7 @@ The following arguments are supported:
 
 Action Blocks (for `default_action`) support the following:
 
-* `type` - (Required) The type of routing action. Valid values are `forward`, `redirect` and `fixed-response`.
+* `type` - (Required) The type of routing action. Valid values are `forward`, `redirect`, `fixed-response` and `authenticate-cognito`.
 * `target_group_arn` - (Optional) The ARN of the Target Group to which to route traffic. Required if `type` is `forward`.
 * `redirect` - (Optional) Information for creating a redirect action. Required if `type` is `redirect`.
 * `fixed_response` - (Optional) Information for creating an action that returns a custom HTTP response. Required if `type` is `fixed-response`.
--- a/website/docs/r/lb_listener.html.markdown
+++ b/website/docs/r/lb_listener.html.markdown
@@ -131,6 +131,43 @@ resource "aws_lb_listener" "front_end" {
 }
 ```
 
+### Authenticate-oidc Action
+
+```hcl
+resource "aws_lb" "front_end" {
+  # ...
+}
+
+resource "aws_lb_target_group" "front_end" {
+  # ...
+}
+
+resource "aws_lb_listener" "front_end" {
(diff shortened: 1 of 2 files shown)


website/docs/r/lb_listener.html.markdown
```diff
@@ -85,6 +85,67 @@ resource "aws_lb_listener" "front_end" {
...
+resource "aws_cognito_user_pool_domain" "domain" {
+  # ...
+}
+
+resource "aws_lb_listener" "front_end" {
+  load_balancer_arn = "${aws_lb.front_end.arn}"
+  port              = "80"
+  protocol          = "HTTP"
+
+  default_action {
```
bflad: Doesn't this require two `default_action` with `order` set? I believe the second action is required for what to do with the request after its authenticated.

website/docs/r/lb_listener.html.markdown
```diff
@@ -85,6 +85,67 @@ resource "aws_lb_listener" "front_end" {
...
+resource "aws_lb" "front_end" {
+  # ...
+}
+
+resource "aws_lb_listener" "front_end" {
+  load_balancer_arn = "${aws_lb.front_end.arn}"
+  port              = "80"
+  protocol          = "HTTP"
+
+  default_action {
```
bflad: Same here, I believe this requires two actions.

website/docs/r/lb_listener_rule.html.markdown
```diff
@@ -97,6 +97,67 @@ resource "aws_lb_listener_rule" "health_check" {
 
 ```
 
+### Authenticate-cognito Action
```
bflad: Same comments for these examples as well.

website/docs/r/lb_listener.html.markdown
fotos: Good point! 👀I haven't tested those example. 🤦‍♂️  I'll do this now.

website/docs/r/lb_listener_rule.html.markdown
fotos: Yes, I just noticed that the documentation in `lb_listener_rule.html.markdown` is in a bit different format. I was updating this while you've been reviewing. 😄 Should be 👌 now, can you check again?

website/docs/r/lb_listener.html.markdown
fotos: You are absolutely right. 🙌 I missed this in the tests (and my local setup is a bit different).

Should be 🆗 now.

website/docs/r/lb_listener.html.markdown
fotos: 2 (default_)actions have been added. Should be 🙆‍♂️ now.

PR #6117, opened 2018-10-10
Add KMS encryption support to the aws_athena_database resource.

joestump:
## What does this add?

Adds an `encryption_key` attribute to the `aws_athena_database` resource. This allows TF to create databases that read from encrypted S3 buckets.

## Example HCL

```hcl
resource "aws_kms_key" "test" {
  name = "athenta-test"
}

resource "aws_athena_database" "test" {
  name          = "athena-test"
  bucket        = "${aws_s3_bucket.test.bucket}"
  force_destroy = true

  encryption_key {
    type = "SSE_KMS"
    id   = "${aws_kms_key.test.arn}" 
  }
}
```

## Test output

```
make testacc TESTARGS='-run=TestAccAWSAthenaDatabase'
TF_ACC=1 go test ./... -v -run=TestAccAWSAthenaDatabase -timeout 120m
=== RUN   TestAccAWSAthenaDatabase_basic
--- PASS: TestAccAWSAthenaDatabase_basic (49.53s)
=== RUN   TestAccAWSAthenaDatabase_encryption
--- PASS: TestAccAWSAthenaDatabase_encryption (79.31s)
=== RUN   TestAccAWSAthenaDatabase_nameStartsWithUnderscore
--- PASS: TestAccAWSAthenaDatabase_nameStartsWithUnderscore (49.75s)
=== RUN   TestAccAWSAthenaDatabase_nameCantHaveUppercase
--- PASS: TestAccAWSAthenaDatabase_nameCantHaveUppercase (0.92s)
=== RUN   TestAccAWSAthenaDatabase_destroyFailsIfTablesExist
--- PASS: TestAccAWSAthenaDatabase_destroyFailsIfTablesExist (61.16s)
=== RUN   TestAccAWSAthenaDatabase_forceDestroyAlwaysSucceeds
--- PASS: TestAccAWSAthenaDatabase_forceDestroyAlwaysSucceeds (72.15s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	314.027s
```

### DIFF
--- a/aws/resource_aws_athena_database.go
+++ b/aws/resource_aws_athena_database.go
@@ -59,7 +59,7 @@ func resourceAwsAthenaDatabase() *schema.Resource {
 	}
 }
 
-func getResultConfig(d *schema.ResourceData) (athena.ResultConfiguration, error) {
+func getResultConfig(d *schema.ResourceData) (*athena.ResultConfiguration, error) {
 	e := d.Get("encryption_key").([]interface{})
 	data := e[0].(map[string]interface{})
 	keyType := data["type"].(string)
@@ -71,7 +71,7 @@ func getResultConfig(d *schema.ResourceData) (athena.ResultConfiguration, error)
 
 	if len(keyType) > 0 {
 		if strings.HasSuffix(keyType, "_KMS") && len(keyID) <= 0 {
-			return fmt.Errorf("Key type %s requires a valid KMS key ID", keyType)
+			return nil, fmt.Errorf("Key type %s requires a valid KMS key ID", keyType)
 		}
 
 		encryptionConfig := athena.EncryptionConfiguration{
@@ -85,7 +85,7 @@ func getResultConfig(d *schema.ResourceData) (athena.ResultConfiguration, error)
 		resultConfig.EncryptionConfiguration = &encryptionConfig
 	}
 
-	return resultConfig, nil
+	return &resultConfig, nil
 }
 
 func resourceAwsAthenaDatabaseCreate(d *schema.ResourceData, meta interface{}) error {
@@ -98,7 +98,7 @@ func resourceAwsAthenaDatabaseCreate(d *schema.ResourceData, meta interface{}) e
 
 	input := &athena.StartQueryExecutionInput{
 		QueryString:         aws.String(fmt.Sprintf("create database `%s`;", d.Get("name").(string))),
-		ResultConfiguration: &resultConfig,
+		ResultConfiguration: resultConfig,
 	}
 
 	resp, err := conn.StartQueryExecution(input)
@@ -121,10 +121,9 @@ func resourceAwsAthenaDatabaseRead(d *schema.ResourceData, meta interface{}) err
 		return err
 	}
 
-	bucket := d.Get("bucket").(string)
 	input := &athena.StartQueryExecutionInput{
 		QueryString:         aws.String(fmt.Sprint("show databases;")),
-		ResultConfiguration: &resultConfig,
+		ResultConfiguration: resultConfig,
 	}
 
 	resp, err := conn.StartQueryExecution(input)
@@ -151,7 +150,6 @@ func resourceAwsAthenaDatabaseDelete(d *schema.ResourceData, meta interface{}) e
 	}
 
 	name := d.Get("name").(string)
-	bucket := d.Get("bucket").(string)
 
 	queryString := fmt.Sprintf("drop database `%s`", name)
 	if d.Get("force_destroy").(bool) {
@@ -161,7 +159,7 @@ func resourceAwsAthenaDatabaseDelete(d *schema.ResourceData, meta interface{}) e
 
 	input := &athena.StartQueryExecutionInput{
 		QueryString:         aws.String(queryString),
-		ResultConfiguration: &resultConfig,
+		ResultConfiguration: resultConfig,
 	}
 
 	resp, err := conn.StartQueryExecution(input)
--- a/aws/resource_aws_athena_database.go
+++ b/aws/resource_aws_athena_database.go
@@ -59,9 +59,7 @@ func resourceAwsAthenaDatabase() *schema.Resource {
 	}
 }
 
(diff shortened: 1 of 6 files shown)


aws/resource_aws_athena_database.go
```diff
@@ -38,18 +38,73 @@ func resourceAwsAthenaDatabase() *schema.Resource {
...
+							Required: true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
 
+func getResultConfig(d *schema.ResourceData) (*athena.ResultConfiguration, error) {
```
bflad: Since this project has a very flat package structure, we prefer to include the service name in the function name. We also refer to these generally as "expand" functions and it would be nice to reduce the signature to not use `d` directly. e.g.

```go
func expandAthenaResultConfiguration(bucket string, encryptionKeyList []interface{}) (*athena.ResultConfiguration, error) {
```

aws/resource_aws_athena_database.go
```diff
@@ -38,18 +38,73 @@ func resourceAwsAthenaDatabase() *schema.Resource {
...
+	e := d.Get("encryption_key").([]interface{})
+	if len(e) <= 0 {
+		return &resultConfig, nil
+	}
+
+	data := e[0].(map[string]interface{})
+	keyType := data["type"].(string)
+	keyID := data["id"].(string)
+
+	if len(keyType) <= 0 {
```
bflad: Since the schema already has this set as `Required: true` this error handling is extraneous.

aws/resource_aws_athena_database.go
```diff
@@ -38,18 +38,73 @@ func resourceAwsAthenaDatabase() *schema.Resource {
...
+		return nil, fmt.Errorf("An encryption key type is required")
+	}
+
+	if strings.HasSuffix(keyType, "_KMS") && len(keyID) <= 0 {
+		return nil, fmt.Errorf("Key type %s requires a valid KMS key ID", keyType)
```
bflad: Does the API not provide a helpful error message already? Since this error handling occurs during apply time anyways, we should avoid adding logic the API already handles unless necessary.

aws/resource_aws_athena_database.go
```diff
@@ -38,18 +38,73 @@ func resourceAwsAthenaDatabase() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"encryption_key": {
```
bflad: The API refers to this as `encryption_configuration` -- is there any reason to diverge from their naming?

aws/resource_aws_athena_database.go
```diff
@@ -38,18 +38,73 @@ func resourceAwsAthenaDatabase() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
```
bflad: The API refers to this as `kms_key` -- is there any reason to diverge from their naming?

aws/resource_aws_athena_database.go
```diff
@@ -38,18 +38,73 @@ func resourceAwsAthenaDatabase() *schema.Resource {
...
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"type": {
```
bflad: Two things:
* The API refers to this as `encryption_option` -- is there any reason to diverge from their naming?
* We should validate this attribute using the SDK provided constants:

```go
  athena.EncryptionOptionCseKms,
  athena.EncryptionOptionSseKms,
  athena.EncryptionOptionSseS3,
}, false),
```

aws/resource_aws_athena_database.go
joestump: It does; just felt dirty making an AWS call for known errors. If that’s the pattern, I’ll update though.

aws/resource_aws_athena_database.go
joestump: I'd modeled it after the CodePipeline's HCL, which makes a bit more sense, but if the pattern is to copy the API directly to HCL I'll update.

aws/resource_aws_athena_database.go
joestump: Good call on the validate. I mentioned why I diverged from the API above. I think the HCL I've proposed would look cleaner than this:

```hcl
encryption_configuration {
  encryption_option = "CSE_KMS"
  kms_key = "${aws_kms_key.foo.id}"
}
```

It feels verbose to me. I think the CodePipeline API looks/feels more correct.

joestump:
@bflad I've addressed most of the comments related to the Golang. Let me know what you think about my response on the HCL. I'm new to the AWS provider and will defer to the prevailing patterns. 👍 

Thanks for the review!

aws/resource_aws_athena_database.go
bflad: It might make a bit more sense today, but not necessarily in the future should the API change.  Here are a few reasons we prefer to stick with the API:
* We have gotten bitten in the past by trying to create abstractions and the API changes in a way that doesn't match those abstractions.
* Less of a maintenance burden to mentally work with the codebase if there are no abstractions/conversions.
* It keeps Terraform in line with the AWS API documentation, SDKs, and CloudFormation so operators coming from other documentation or systems can easily reference or convert between them.
* Different AWS services don't always line up with each other in terms of implementations. While this may look similar to CodePipeline today, Athena or CodePipeline may change tomorrow.

Hope this helps!

joestump:
@bflad I took a stab at adding your notes into the `CONTRIBUTING.md`. I believe all comments have been addressed now. 👍

bflad:
This has been released in [version 1.41.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1410-october-18-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #6150, opened 2018-10-14
aws_ecs_service support deployment_minimum_healthy_percent for DAEMON strategy

### LINKED ISSUE #5236: aws_ecs_service - deployment_minimum_healthy_percent for DAEMON
_This issue was originally opened by @nirpod as hashicorp/terraform#18480. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

why wouldn't you support  deployment_minimum_healthy_percent for DAEMON (for aws_ecs_service rss)?
AWS support this feature.

mschurenko:
This PR aims to fix #5236 

To summarize, I would like to be able to set `deployment_minimum_healthy_percent` to 50 when using a scheduling strategy of `DAEMON`. Currently this is not possible as this value being set to nil. The result of this is that `deployment_minimum_healthy_percent` is set to 0 (AWS default) which results in downtime for a deployment.

This is my first PR to this project, so I might be missing something obvious, but there doesn't appear to be a way to conditionally set a default value in `github.com/hashicorp/terraform/helper/schema` https://github.com/terraform-providers/terraform-provider-aws/blob/master/vendor/github.com/hashicorp/terraform/helper/schema/schema.go#L37. 

Ideally the default value for `deployment_minimum_healthy_percent` would be 200 for REPLICA strategy and 0 for DAEMON, but I was unable to do that.

As a result if one uses the DAEMON strategy but does not set a value for `deployment_minimum_healthy_percent` the default of 100 is used and you get this error:

```
 aws_ecs_service.my_service: InvalidParameterException: Both maximumPercent and minimumHealthyPercent cannot be 100 as this will block deployments.
	status code: 400, request id: 848ec013-cf6e-11e8-8638-07c29e49208f "my_service"
```

IMO this is better then the current behavior where one cannot set `deployment_minimum_healthy_percent` at all. 

Having said this, I'm hoping there's a better approach to doing this.

* `aws_ecs_service` resource will use `deployment_minimum_healthy_percent` when `scheduling_strategy` is set to `DAEMON`

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSEcsService_withDaemonSchedulingStrategy'
TF_ACC=1 go test ./aws -v -run=TestAccAWSEcsService_withDaemonSchedulingStrategy -timeout 120m
=== RUN   TestAccAWSEcsService_withDaemonSchedulingStrategy
--- PASS: TestAccAWSEcsService_withDaemonSchedulingStrategy (27.18s)
PASS
```


mschurenko:
Hi @bflad. Not sure if anyone has had a chance to look at this yet.

Thanks

mschurenko:
I'm not trying to sound like a jerk or anything, but is this normal to have no response for a PR after 11 days? If this won't be accepted for whatever reason that's fine. I would just like to know. If it's not I'll most likely need to use Cloudformation for this.

bflad:
@mschurenko its likely to be reviewed and accepted (potentially with some changes) given my understanding of the situation. A bunch of HashiCorp and community members have been busy with preparation, travel, and attendance of HashiConf this week.

mschurenko:
Hey @bflad thanks a lot for the response. Sorry to be so pushy. I didn't realize HashiConf was next week. 

Thanks!

aws/resource_aws_ecs_service.go
```diff
@@ -106,12 +106,6 @@ func resourceAwsEcsService() *schema.Resource {
 				Type:     schema.TypeInt,
 				Optional: true,
 				Default:  100,
```
bflad: Instead of leaving a `Default: 100` here that only works for one of the scheduling strategies, should we instead remove `Default: 100` and implement a `DiffSuppressFunc` that covers suppressing the difference depending on the strategy? A similar update should likely occur for maximum as well.

aws/resource_aws_ecs_service.go
```diff
@@ -354,7 +348,7 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 	input.SchedulingStrategy = aws.String(schedulingStrategy)
 	if schedulingStrategy == ecs.SchedulingStrategyDaemon {
 		// unset these if DAEMON
-		input.DeploymentConfiguration = nil
+		input.DeploymentConfiguration.MaximumPercent = aws.Int64(100)
```
bflad: Is there a reason to keep this logic here at all? Since the comment eludes to "unsetting" values due to `DAEMON` we should either remove this line or add a comment why its being hardcoded to `100`

aws/resource_aws_ecs_service_test.go
```diff
@@ -2092,6 +2092,7 @@ resource "aws_ecs_service" "ghost" {
   cluster = "${aws_ecs_cluster.default.id}"
   task_definition = "${aws_ecs_task_definition.ghost.family}:${aws_ecs_task_definition.ghost.revision}"
   scheduling_strategy = "DAEMON"
+  deployment_minimum_healthy_percent = "50"
```
bflad: We should implement this in a separate acceptance test and test configuration to ensure both the unconfigured and configured cases either work or are covered by `ExpectError` testing as appropriate. Copy-pasting most of the existing test is fine. 👍

aws/resource_aws_ecs_service.go
mschurenko: If we remove the `Default` wouldn't that be backwards incompatible for those who are using a `replica` strategy and didn't set `deployment_minimum_healthy_percent`?

Similar for `deployment_maximum_percent`, if we remove `Default` won't that be a problem for backwards compatibility?

aws/resource_aws_ecs_service.go
mschurenko: I updated this to be `input.DeploymentConfiguration.MaximumPercent = nil` as well

mschurenko:
I think the fundamental problem here is that `replica` and `daemon` should have different default values for both `deployment_minimum_healthy_percent` and `deployment_maximum_percent` but not sure how to accomplish that the way the schema is currently implemented. (I could be missing something obvious here though).

So one solution, as @bflad has already alluded to, is that we have no default values for `deployment_minimum_healthy_percent` and `deployment_maximum_percent`. One potential issue with this is this behavior is not backwards compatible for `daemon` strategies (maybe this isn't much of an issue if we mention this in the release notes?)

aws/resource_aws_ecs_service_test.go
mschurenko: Having an acceptance test for daemon that does not set `deployment_minimum_healthy_percent` to something less than 100 results in:
```
InvalidParameterException: Both maximumPercent and minimumHealthyPercent cannot be 100 as this will block deployments
```

aws/resource_aws_ecs_service.go
mschurenko: Ok. So when you say `we allow two defaults chosen by the API`, do you mean by the AWS API? If so then for `deployment_maximum_percent` and `deployment_minimum_healthy_percent` we should set these as `Optional`?

bflad:

PR #6154, opened 2018-10-14
aws_lb_listener_rule: Add eventualy consistent read after creation

gseyetik:
It is possible that during the creation of a elbv2 listener rule that the immediate read results in the newly created rule being removed from the state as it can not be found.

This adds a standard 1 minute time out for newly created resources.

Similar to #5167 .

* Add a 1 minute eventual consistency read on creation of a elbv2 listener rule.

```
None to run.

```


bflad:
This has been released in [version 1.41.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1410-october-18-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #5860, opened 2018-09-13
Adding a note to help people avoid a common mistake

josephwilson:
I'm sure there have been plenty of people who struggled with this, I wasted a few hours on it myself. Might be helpful to put a note in to help save people the suffering. The problem is described here: https://github.com/hashicorp/terraform/issues/16285, but it would be easier to not have to find that in the first place.

Fixes #16285

Changes proposed in this pull request:

* Adding one simple note to the documentation.

PR #6177, opened 2018-10-16
updated iam policy docs w/here doc syntax

tracypholmes:
Fixes #5860


bflad:
The Terraform 0.12 documentation for heredoc has been released: https://www.terraform.io/docs/configuration/expressions.html#string-literals

If we'd like to change anything here, lets create a new pull request linking to that. 👍

PR #6185, opened 2018-10-17
Add the custom domain name feature in aws_cognito_user_pool_domain resource

### LINKED ISSUE #5026: Feature Request - support custom domain in aws_cognito_user_pool_domain
### Community Note

### Description

Custom domain support was launced by AWS in June 2018 but not supported by Terraform. There is a new arn param for a certificate and if there is a certificate provided then the domain param will be the full domain. Right now validateCognitoUserPoolDomain function won't accept that.

* aws_cognito_user_pool_domain

```hcl
resource "aws_cognito_user_pool_domain" "main" {
  domain          = "auth.example.com"
  certificate_arn = "${aws_acm_certificate_validation.cert.certificate_arn}"
  user_pool_id    = "${aws_cognito_user_pool.pool.id}"
}
```

### References

* https://aws.amazon.com/about-aws/whats-new/2018/06/amazon-cognito-now-supports-custom-domains-for-a-unified-login-experience/

adelamarre:
Fixes #5026

* Add the `certificate_arn` parameter to let use a custom domain 

You need to set environment variable for this test : 
 * **AWS_COGNITO_USER_POOL_DOMAIN_ROOT_DOMAIN**  
  This environment variable must be set to the fqdn of an ISSUED ACM certificate in us-east-1 to enable this test. The fqdn must have a A record in the zone.
 * **AWS_COGNITO_USER_POOL_DOMAIN_CERTIFICATE_ARN**
  This environment variable must be set to the ARN of an ISSUED ACM certificate in us-east-1 to enable this test.

```
$ make testacc TESTARGS='-run=TestAccAWSCognitoUserPoolDomain_custom'
TF_ACC=1 go test ./... -v -run=TestAccAWSCognitoUserPoolDomain_custom -timeout 120m
=== RUN   TestAccAWSCognitoUserPoolDomain_custom
--- PASS: TestAccAWSCognitoUserPoolDomain_custom (2013.27s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	2013.299s
```

### DIFF
--- a/aws/resource_aws_cognito_user_pool_domain.go
+++ b/aws/resource_aws_cognito_user_pool_domain.go
@@ -22,10 +22,15 @@ func resourceAwsCognitoUserPoolDomain() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"domain": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"certificate_arn": {
 				Type:         schema.TypeString,
-				Required:     true,
+				Optional:     true,
 				ForceNew:     true,
-				ValidateFunc: validateCognitoUserPoolDomain,
+				ValidateFunc: validateArn,
 			},
 			"user_pool_id": {
 				Type:     schema.TypeString,
@@ -57,10 +62,21 @@ func resourceAwsCognitoUserPoolDomainCreate(d *schema.ResourceData, meta interfa
 
 	domain := d.Get("domain").(string)
 
+	timeout := 1 * time.Minute //Default timeout for a basic domain
+
 	params := &cognitoidentityprovider.CreateUserPoolDomainInput{
 		Domain:     aws.String(domain),
 		UserPoolId: aws.String(d.Get("user_pool_id").(string)),
 	}
+
+	if v, ok := d.GetOk("certificate_arn"); ok {
+		customDomainConfig := &cognitoidentityprovider.CustomDomainConfigType{
+			CertificateArn: aws.String(v.(string)),
+		}
+		params.CustomDomainConfig = customDomainConfig
+		timeout = 60 * time.Minute //Custom domains take more time to become active
+	}
+
 	log.Printf("[DEBUG] Creating Cognito User Pool Domain: %s", params)
 
 	_, err := conn.CreateUserPoolDomain(params)
@@ -78,7 +94,8 @@ func resourceAwsCognitoUserPoolDomainCreate(d *schema.ResourceData, meta interfa
 		Target: []string{
 			cognitoidentityprovider.DomainStatusTypeActive,
 		},
-		Timeout: 1 * time.Minute,
+		MinTimeout: 1 * time.Minute,
+		Timeout:    timeout,
 		Refresh: func() (interface{}, string, error) {
 			domain, err := conn.DescribeUserPoolDomain(&cognitoidentityprovider.DescribeUserPoolDomainInput{
 				Domain: aws.String(d.Get("domain").(string)),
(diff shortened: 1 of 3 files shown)


aws/resource_aws_cognito_user_pool_domain.go
```diff
@@ -22,10 +22,15 @@ func resourceAwsCognitoUserPoolDomain() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"domain": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
bflad: Can you please ensure `ValidateFunc: validateCognitoUserPoolDomain,` remains on the `domain` attribute? Thanks!

aws/resource_aws_cognito_user_pool_domain.go
```diff
@@ -119,6 +136,7 @@ func resourceAwsCognitoUserPoolDomainRead(d *schema.ResourceData, meta interface
 	desc := domain.DomainDescription
 
 	d.Set("domain", d.Id())
+	d.Set("certificate_arn", desc.CustomDomainConfig.CertificateArn)
```
bflad: To prevent potential panics, we should ensure `CustomDomainConfig` is not `nil` first, e.g.

```suggestion
	d.Set("certificate_arn", "")
	if desc.CustomDomainConfig != nil {
		d.Set("certificate_arn", desc.CustomDomainConfig.CertificateArn)
	}
```

aws/resource_aws_cognito_user_pool_domain.go
adelamarre: Hi @bflad , thanks for your comments :-)
The problem here is that we don't know what to do in the validator context because the rules are differents if we have a custom domain or not and the other property of the resource is not accessible to determine the regex to use... So i decided to remove the validator and let the api throwing the error instead of terraform... The solution should be to validate the domain name against the two rules but that makes no sens...

bflad: @adelamarre hmm, I was going off https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPoolDomain.html still showing `Pattern: ^[a-z0-9](?:[a-z0-9\-]{0,61}[a-z0-9])?$` for domain. Is that not accurate anymore?

bflad: If the `validateCognitoUserPoolDomain` usage here is being removed, then the function and its testing should be removed as well since it doesn't look like its used anywhere else.

adelamarre: @bflad , No it's not accurate... as you can see the regex do not permit dots (.) in the name however there is always dots in a custom domain name...
Yes if the ```validateCognitoUserPoolDomain``` is not used anymore we could remove it and the affiliated test...

bflad: Please note, on merge I am removing the now unused `validateCognitoUserPoolDomain` function and added 63 character string length validation back to the `domain` attribute as I encountered this error from the SDK/API in my testing:

```
--- FAIL: TestAccAWSCognitoUserPoolDomain_custom (5.54s)
                * aws_cognito_user_pool_domain.main: 1 error occurred:
                * aws_cognito_user_pool_domain.main: Error creating Cognito User Pool Domain: InvalidParameterException: 1 validation error detected: Value 'tf-acc-test-domain-305748293312753547.--OMITTED--' at 'domain' failed to satisfy constraint: Member must have length less than or equal to 63
```

bflad:

PR #6190, opened 2018-10-17
aws_egress_only_internet_gateway: Adjust eventualy consistent logic

gseyetik:
Fixes eventually consistent creates not being captured by the current logic.

* Remove state change function that is overly verbose for our situation.
* Move eventually consistent logic in to read function.
  * The current logic does an immediate read after the first success which could still lead to the new object being removed from the state.
* Remove the check for a non existent error.
  * The error ```InvalidEgressInternetGatewayID.NotFound``` is not a real error. Instead the behaviour is an empty array. https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#CommonErrors

Other:
I kept the weird logic of checking the array for the id being found but I'm not sure that is really needed either.

```
$ make testacc TESTARGS='-run=TestAccAWSEgressOnlyInternetGateway_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSEgressOnlyInternetGateway_basic -timeout 120m
=== RUN   TestAccAWSEgressOnlyInternetGateway_basic
=== PAUSE TestAccAWSEgressOnlyInternetGateway_basic
=== CONT  TestAccAWSEgressOnlyInternetGateway_basic
--- PASS: TestAccAWSEgressOnlyInternetGateway_basic (32.46s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	32.481s

```
This is the first time I've tried to run the tests. If this wasn't done correctly let me know.

### DIFF
--- a/aws/resource_aws_egress_only_internet_gateway.go
+++ b/aws/resource_aws_egress_only_internet_gateway.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -40,18 +39,6 @@ func resourceAwsEgressOnlyInternetGatewayCreate(d *schema.ResourceData, meta int
 
 	d.SetId(*resp.EgressOnlyInternetGateway.EgressOnlyInternetGatewayId)
 
-	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
-		igRaw, _, err := EIGWStateRefreshFunc(conn, d.Id())()
-		if igRaw != nil {
-			return nil
-		}
-		if err == nil {
-			return resource.RetryableError(err)
-		} else {
-			return resource.NonRetryableError(err)
-		}
-	})
-
 	if err != nil {
 		return fmt.Errorf("%s", err)
 	}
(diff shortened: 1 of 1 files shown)


aws/resource_aws_egress_only_internet_gateway.go
```diff
@@ -40,69 +39,42 @@ func resourceAwsEgressOnlyInternetGatewayCreate(d *schema.ResourceData, meta int
...
-		if *igw.EgressOnlyInternetGatewayId == d.Id() {
-			found = true
+	err := resource.Retry(1*time.Minute, func() *resource.RetryError {
+		resp, err := conn.DescribeEgressOnlyInternetGateways(req)
+		if err != nil {
+			return resource.NonRetryableError(err)
+		}
+		if resp != nil && len(resp.EgressOnlyInternetGateways) > 0 {
+			for _, igw := range resp.EgressOnlyInternetGateways {
+				if *igw.EgressOnlyInternetGatewayId == d.Id() {
```
bflad: To prevent potential panics, we should use the SDK provided helper function here, e.g.

```suggestion
				if aws.StringValue(igw.EgressOnlyInternetGatewayId) == d.Id() {
```

aws/resource_aws_egress_only_internet_gateway.go
```diff
@@ -40,69 +39,42 @@ func resourceAwsEgressOnlyInternetGatewayCreate(d *schema.ResourceData, meta int
...
+					found = true
+					break
+				}
+			}
+		}
+		if d.IsNewResource() && !found {
+			return resource.RetryableError(nil)
```
bflad: Nit: We should return context about the error here:

```suggestion
			return resource.RetryableError(fmt.Errorf("Egress Internet Gateway (%s) not found", d.Id()))
```

PR #6196, opened 2018-10-18
vendor: aws/aws-sdk-go@v1.15.57 and add aws/aws-sdk-go/service/resourcegroups@v1.15.57

bflad:
Reference: https://github.com/terraform-providers/terraform-provider-aws/issues/2972#issuecomment-430854855

Updates:
* `govendor fetch github.com/aws/aws-sdk-go/...@v1.15.57`
* `govendor fetch github.com/aws/aws-sdk-go/service/resourcegroups@v1.15.57`

Output from acceptance testing: Handled via daily acceptance testing

PR #6217, opened 2018-10-19
Add an `aws_resource_group` resource for managing AWS resource groups

joestump:
## What is the purpose of this PR?

Introduces an `aws_resource_group` resource.

## Example HCL

```hcl
resource "aws_resource_group" "test" {
  name        = "%s"
  description = "Hello World"

  resource_query {
    query = <<JSON
{
  "ResourceTypeFilters": [
    "AWS::EC2::Instance"
  ],
  "TagFilters": [
    {
      "Key": "Stage",
      "Values": ["Test"]
    }
  ]
}
JSON
  }
}
```

## Test outputs

```
make testacc TESTARGS='-run=TestAccAWSResourceGroup'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSResourceGroup -timeout 120m
=== RUN   TestAccAWSResourceGroup_importBasic
=== PAUSE TestAccAWSResourceGroup_importBasic
=== RUN   TestAccAWSResourceGroup_basic
=== PAUSE TestAccAWSResourceGroup_basic
=== CONT  TestAccAWSResourceGroup_importBasic
=== CONT  TestAccAWSResourceGroup_basic
--- PASS: TestAccAWSResourceGroup_basic (17.90s)
--- PASS: TestAccAWSResourceGroup_importBasic (19.07s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	20.443s
```

## TODO

- [x] Merge in #6196
- [x] Documentation


aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+				ForceNew: true,
+			},
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"resource_query": {
+				Type:     schema.TypeSet,
```
bflad: For "container" attributes with `MaxItems: 1` we prefer to use `TypeList` instead of `TypeSet` to simplify a lot of things. 👍 

```suggestion
				Type:     schema.TypeList,
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+
+	input := resourcegroups.CreateGroupInput{
+		Description:   aws.String(d.Get("description").(string)),
+		Name:          aws.String(d.Get("name").(string)),
+		ResourceQuery: extractResourceGroupResourceQuery(d.Get("resource_query").(*schema.Set).List()),
+	}
+
+	res, err := conn.CreateGroup(&input)
+	if err != nil {
+		return err
```
bflad: Nit: We should provide operators and code maintainers some context about when/why the error occurred, e.g. 

```suggestion
		return fmt.Errorf("error creating Resource Groups Group: %s", err)
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+
+func resourceAwsResourceGroupRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).resourcegroupsconn
+
+	g, err := conn.GetGroup(&resourcegroups.GetGroupInput{
+		GroupName: aws.String(d.Id()),
+	})
+
+	if err != nil {
+		return err
```
bflad: Nit: Same here, regarding context. 

```suggestion
		return fmt.Errorf("error reading Resource Groups Group (%s): %s", d.Id(), err)
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+	return resourceAwsResourceGroupRead(d, meta)
+}
+
+func resourceAwsResourceGroupRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).resourcegroupsconn
+
+	g, err := conn.GetGroup(&resourcegroups.GetGroupInput{
+		GroupName: aws.String(d.Id()),
+	})
+
```
bflad: According to the [SDK documentation](https://docs.aws.amazon.com/sdk-for-go/api/service/resourcegroups/#ResourceGroups.GetGroup), the `GetGroup` call will return a not found error code, which we can use to trigger Terraform resource recreation, e.g.

```suggestion

	if isAWSErr(err, resourcegroups.ErrCodeNotFoundException, "") {
		log.Printf("[WARN] Resource Groups Group (%s) not found, removing from state", d.Id())
		d.SetId("")
		return nil
	}

```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+	d.Set("name", aws.StringValue(g.Group.Name))
+	d.Set("description", aws.StringValue(g.Group.Description))
+	d.Set("arn", aws.StringValue(g.Group.GroupArn))
+
+	q, err := conn.GetGroupQuery(&resourcegroups.GetGroupQueryInput{
+		GroupName: aws.String(d.Id()),
+	})
+
+	if err != nil {
+		return err
```
bflad: Nit: Same here, regarding context. 

```suggestion
		return fmt.Errorf("error reading Resource Groups Group (%s) Query: %s", d.Id(), err)
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	resultQuery := map[string]interface{}{}
+	resultQuery["query"] = aws.StringValue(q.GroupQuery.ResourceQuery.Query)
+	resultQuery["type"] = aws.StringValue(q.GroupQuery.ResourceQuery.Type)
+	d.Set("resource_query", []map[string]interface{}{resultQuery})
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```suggestion
	if err := d.Set("resource_query", []map[string]interface{}{resultQuery}); err != nil {
		return fmt.Errorf("error setting resource_query: %s", err)
	}
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if d.HasChange("description") {
+		input := resourcegroups.UpdateGroupInput{
+			GroupName:   aws.String(d.Id()),
+			Description: aws.String(d.Get("description").(string)),
+		}
+
+		_, err := conn.UpdateGroup(&input)
+		if err != nil {
+			return err
```
bflad: Nit: Same here, regarding context. 

```suggestion
			return fmt.Errorf("error updating Resource Groups Group (%s): %s", d.Id(), err)
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if d.HasChange("resource_query") {
+		input := resourcegroups.UpdateGroupQueryInput{
+			GroupName:     aws.String(d.Id()),
+			ResourceQuery: extractResourceGroupResourceQuery(d.Get("resource_query").([]interface{})),
+		}
+
+		_, err := conn.UpdateGroupQuery(&input)
+		if err != nil {
+			return err
```
bflad: Nit: Same here, regarding context. 

```suggestion
			return fmt.Errorf("error updating Resource Groups Group (%s) Query: %s", d.Id(), err)
```

aws/resource_aws_resource_group.go
```diff
@@ -0,0 +1,169 @@
...
+func resourceAwsResourceGroupDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).resourcegroupsconn
+
+	input := resourcegroups.DeleteGroupInput{
+		GroupName: aws.String(d.Id()),
+	}
+
+	_, err := conn.DeleteGroup(&input)
+	if err != nil {
+		return err
```
bflad: Nit: Same here, regarding context. 

```suggestion
		return fmt.Errorf("error deleting Resource Groups Group (%s) Query: %s", d.Id(), err)
```

aws/resource_aws_resource_group_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/resourcegroups"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSResourceGroup_importBasic(t *testing.T) {
```
bflad: Instead of creating a separate acceptance test that essentially duplicates the regular test, we prefer to add the import `TestStep` to all other acceptance tests. 👍

aws/resource_aws_resource_group_test.go
```diff
@@ -0,0 +1,103 @@
...
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccAWSResourceGroup_basic(t *testing.T) {
+	resourceName := "aws_resource_group.test"
+	n := fmt.Sprintf("test-group-%d", acctest.RandInt())
```
bflad: We have a provider SDK helper function for this 😄 

```suggestion
	n := acctest.RandomWithPrefix("tf-acc-test")
```

aws/resource_aws_resource_group_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+		return nil
+	}
+}
+
+func testAccAWSResourceGroupConfig_basic(rName string) string {
+	return fmt.Sprintf(`
+resource "aws_resource_group" "test" {
+  name        = "%s"
+  description = "Hello World"
```
bflad: We should add separate acceptance tests that attempt to perform an update of `description` and `query`

aws/provider.go
```diff
@@ -393,297 +393,298 @@ func Provider() terraform.ResourceProvider {
 			"aws_dx_hosted_private_virtual_interface_accepter": resourceAwsDxHostedPrivateVirtualInterfaceAccepter(),
 			"aws_dx_hosted_public_virtual_interface":           resourceAwsDxHostedPublicVirtualInterface(),
 			"aws_dx_hosted_public_virtual_interface_accepter":  resourceAwsDxHostedPublicVirtualInterfaceAccepter(),
-			"aws_dx_lag":                                       resourceAwsDxLag(),
```
bflad: If you upgrade to Go 1.11, this issue should go away. 😄 

```
$ make test
gofmt needs running on the following files:
./aws/provider.go
You can use the command: `make fmt` to reformat code.
make: *** [fmtcheck] Error 1
```

bflad:
Oh! I would also recommend naming the resource (and updating file names) to `aws_resourcegroups_group` -- it looks uglier but we have been trying to stick with `aws_SERVICE_THING` where possible to prevent namespace collisions in the future. See also: `aws_organizations_organization` 😅

joestump:
@bflad I think comments have been addressed. 👍

aws/resource_aws_resourcegroups_group_test.go
```diff
@@ -0,0 +1,115 @@
...
+  "TagFilters": [
+    {
+      "Key": "Hello",
+      "Values": ["World"]
+    }
+  ]
+}
+`
+
+	resource.ParallelTest(t, resource.TestCase{
```
bflad: This `TestCase` should include a `CheckDestroy` function that verifies the removal of the resource from the API: https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html#checkdestroy

website/docs/r/resourcegroups_group.html.markdown
```diff
@@ -0,0 +1,62 @@
...
+
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `arn` - The ARN assigned by AWS for this resource group.
+
+## Import
+
+IAM Users can be imported using the `name`, e.g.
```
bflad: Copypasta here and the example command 🍝 😄

joestump: Ha! Literally had that committed about 30 minutes ago and forgot to `push`. 🤦‍♂️

joestump:
@bflad I've got an interesting issue with this provider. Turns out you can set tags when you call `CreateGroup`, but `GetGroup` and `ListGroups` don't return the groups tags. This is causing the import test to fail if tags are present as it expects tags to be instate, but I can't pull them from the API.

My proposal would be to:

* Remove import functionality for now.
* Mark `tags` as `ForceNew`.

The other option would be to remove this attribute entirely I guess? I've sent an email to our AWS TAM asking they make the tags retrievable.

bflad:
For reading `tags`, I believe [`GetTags`](https://docs.aws.amazon.com/ARG/latest/APIReference/API_GetTags.html) is what you are looking for, passing in the ARN of the group. For updating `tags`, I believe [`Tag`](https://docs.aws.amazon.com/ARG/latest/APIReference/API_Tag.html) and [`Untag`](https://docs.aws.amazon.com/ARG/latest/APIReference/API_Untag.html) 😄 

Otherwise feel free to remove the group tagging bits out for now and we can get in separately.

joestump:
I'll give `GetTags` a try with a resource group's ARN. For some reason it hadn't dawned on me that might work.

joestump:
@bflad okay, so looking over these endpoints I've got a new proposal for this. I think we should remove tags from this resource and create two additional resources:

* `aws_resourcegroups_tags` - Data resource that takes an ARN and returns a map of tags for ARN.
* `aws_resourcegroups_tags` - Regular resource that takes an ARN and tags it.

The nice thing about this is that it works for any ARN. 

Thoughts?

bflad:
The API for tagging resources via resource groups is separate and can be found here:
* https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/Welcome.html
* https://docs.aws.amazon.com/sdk-for-go/api/service/resourcegroupstaggingapi/

joestump:
@bflad BRB learning to read good.

bflad:
Hey @joestump 👋 Will you have time to circle back and finish this up? If not, no problem.

joestump:
@bflad I should be able to. Sorry for the delay, the usual things have kept me from revisiting lately.

bflad:

joestump:
@bflad sorry for going MIA. Got slammed with a big deadline at work. I've got tagging with proper tests and whatnot locally. I'll try and clean it up for PR soon!

Issue #6219, opened 2018-10-19
Fix website HCL formatting


paultyng:
The following files have issues when their examples are run through `terraform fmt`:

* [x] [website/docs/d/acm_certificate.html.markdown](../blob/master/website/docs/d/acm_certificate.html.markdown)
* [x] [website/docs/d/autoscaling_groups.html.markdown](../blob/master/website/docs/d/autoscaling_groups.html.markdown)
* [x] [website/docs/d/cloudhsm_v2_cluster.html.markdown](../blob/master/website/docs/d/cloudhsm_v2_cluster.html.markdown)
* [x] [website/docs/d/cognito_user_pools.markdown](../blob/master/website/docs/d/cognito_user_pools.markdown)
* [x] [website/docs/d/db_cluster_snapshot.html.markdown](../blob/master/website/docs/d/db_cluster_snapshot.html.markdown)
* [x] [website/docs/d/db_snapshot.html.markdown](../blob/master/website/docs/d/db_snapshot.html.markdown)
* [x] [website/docs/d/efs_file_system.html.markdown](../blob/master/website/docs/d/efs_file_system.html.markdown)
* [x] [website/docs/d/efs_mount_target.html.markdown](../blob/master/website/docs/d/efs_mount_target.html.markdown)
* [x] [website/docs/d/eks_cluster.html.markdown](../blob/master/website/docs/d/eks_cluster.html.markdown)
* [x] [website/docs/d/elastic_beanstalk_solution_stack.html.markdown](../blob/master/website/docs/d/elastic_beanstalk_solution_stack.html.markdown)
* [x] [website/docs/d/glue_script.html.markdown](../blob/master/website/docs/d/glue_script.html.markdown)
* [x] [website/docs/d/iam_policy_document.html.markdown](../blob/master/website/docs/d/iam_policy_document.html.markdown)
* [x] [website/docs/d/inspector_rules_packages.html.markdown](../blob/master/website/docs/d/inspector_rules_packages.html.markdown)
* [x] [website/docs/d/instances.html.markdown](../blob/master/website/docs/d/instances.html.markdown)
* [x] [website/docs/d/internet_gateway.html.markdown](../blob/master/website/docs/d/internet_gateway.html.markdown)
* [x] [website/docs/d/iot_endpoint.html.markdown](../blob/master/website/docs/d/iot_endpoint.html.markdown)
[...]

PR #6221, opened 2018-10-19
Add script to check example hcl formatting

paultyng:
We probably can't add this to the build process unless we are sure NPM will be present.

This script was used to generate #6219


paultyng:
Closing this, but something we should look in to potentially down the road for both docs and examples.

PR #6231, opened 2018-10-20
d/aws_nat_gateway: Fix filtering and reading of tags

flosell:
Currently, `tags` can be defined as an argument for the `aws_nat_gateway` data source and are [documented](https://www.terraform.io/docs/providers/aws/d/nat_gateway.html) this way but are silently ignored when searching for the NAT Gateway. When reading NAT Gateway details, tag-values aren't populated in terraform. This PR attempts to fix this.

* filter for given tags when provided as attribute
* read tags-value from API response

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsNatGateway'
TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccDataSourceAwsNatGateway -timeout 120m
=== RUN   TestAccDataSourceAwsNatGateway
=== PAUSE TestAccDataSourceAwsNatGateway
=== CONT  TestAccDataSourceAwsNatGateway
--- PASS: TestAccDataSourceAwsNatGateway (244.10s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	244.155s
```

### DIFF
--- a/aws/data_source_aws_nat_gateway.go
+++ b/aws/data_source_aws_nat_gateway.go
@@ -89,6 +89,12 @@ func dataSourceAwsNatGatewayRead(d *schema.ResourceData, meta interface{}) error
 		)...)
 	}
 
+	if tags, ok := d.GetOk("tags"); ok {
+		req.Filter = append(req.Filter, buildEC2TagFilterList(
+			tagsFromMap(tags.(map[string]interface{})),
+		)...)
+	}
+
 	req.Filter = append(req.Filter, buildEC2CustomFilterList(
 		d.Get("filter").(*schema.Set),
 	)...)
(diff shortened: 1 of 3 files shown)


aws/data_source_aws_nat_gateway.go
```diff
@@ -116,6 +122,7 @@ func dataSourceAwsNatGatewayRead(d *schema.ResourceData, meta interface{}) error
 	d.Set("state", ngw.State)
 	d.Set("subnet_id", ngw.SubnetId)
 	d.Set("vpc_id", ngw.VpcId)
+	d.Set("tags", tagsToMap(ngw.Tags))
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("tags", tagstoMap(ngw.Tags)); err != nil {
}
```

Since the `tagsToMap()` function has well established usage across other resources I won't treat this as a blocker here though. 👍

bflad:

PR #6237, opened 2018-10-22
[wip] resource/aws_ses_receipt_rule: support import functionality

acburdine:

* Add import functionality for aws_ses_receipt_rule

Haven't run it yet, waiting on response to the below comment.

### DIFF
--- a/aws/resource_aws_ses_receipt_rule.go
+++ b/aws/resource_aws_ses_receipt_rule.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"sort"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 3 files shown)


acburdine:
After doing some more looking at other resources with imports, it appears this isn't _quite_ the right approach since it won't import the whole state (with all of the subproperties and whatnot).

Ideally, we could use `schema.ImportStatePassthrough` to lookup the rule, but as far as I can tell the resource currently uses just the rule name as an ID, rather than a composite id in the form `<ruleset name>:<rule name>`. Based on the fact that rules are specific to a ruleset, I think it would be better to have the ID be a composite of the ruleset and rule (similar to the iam_role_policy resource).

I'm not sure what the policy is for changing the ID format, so I'd appreciate some feedback on what the best approach to implement this is 😄

bflad:
Hi @acburdine! 👋 Thanks for submitting this. We currently denote changing resource IDs as a major version breaking change since some folks might be relying on the attribute in other references. See also: https://www.terraform.io/docs/extend/best-practices/versioning.html#example-major-number-increments

During import, we only need to `d.Set()` enough information to make the `Read` function succeed, which is immediately called. This looks like its in pretty good shape to implement the import functionality without changing the ID minus the CI errors. Once that's fixed up we can verify the acceptance testing. 👍

bflad:

PR #6239, opened 2018-10-22
add import docs for ses_receipt_rule_set and ses_domain_identity

acburdine:
Noticed these resources support importing, but it was undocumented

* Add import docs for `ses_receipt_rule_set` and `ses_domain_identity`

Output from acceptance testing: n/a


acburdine:
looks like travis is hung up for some reason?

PR #6243, opened 2018-10-23
WIP: Allow in place upgrades of ES domains

### LINKED ISSUE #5554: Add support for Elasticsearch in place upgrades
### Community Note

### Description

AWS has [announced](https://aws.amazon.com/about-aws/whats-new/2018/08/amazon_elasticsearch_service_now_supports_zero_downtime_in-place_version_upgrades/) in place upgrades of ES domains.

Currently the `aws_elasticsearch_domain` resource has a `ForceNew: true` on the `elasticsearch_version` parameter.

It would be good if we could detect if the upgrade path that Terraform is asked to perform is doable (see https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-version-migration.html for migration paths) and then do an in place upgrade if possible.

I'm not sure how to keep the current behaviour of forcing a new resource if it's not possible (with `CustomizeDiff` maybe?) but that would be preferable to just throwing an error if the upgrade path is not possible.

* aws_elasticsearch_domain

tomelliff:
AWS has [announced support](https://aws.amazon.com/about-aws/whats-new/2018/08/amazon_elasticsearch_service_now_supports_zero_downtime_in-place_version_upgrades/) for in place upgrades of some version of ES domains.

For the list of all valid upgrade paths see https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-version-migration.html

Fixes #5554

I'm on a train right now so not yet had a chance to actually run this/the accompanying acceptance test so leaving it as WIP until I get a chance to test it.


rifelpet:
I was messing around with `GetCompatbileElasticsearchVersions` to determine at plan time whether a version change should `ForceNew` or not. It is untested but it may be worth integrating into this PR. Thanks for working on it!

https://github.com/rifelpet/terraform-provider-aws/commit/a94b46ea5192cd8638af3c75aa13c6eb22c01172

tomelliff:
I'm not sure it's necessary to be introduced but it would be a welcome addition to this PR. Unfortunately after cherry picking your change across onto my work it doesn't seem to work and the plan is always showing it can upgrade in place:

```
$ aws es get-compatible-elasticsearch-versions --domain-name tf-test-2
{
    "CompatibleElasticsearchVersions": [
        {
            "SourceVersion": "5.5", 
            "TargetVersions": [
                "5.6"
            ]
        }
    ]
}
$ TF_VAR_version=6.2 terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

aws_elasticsearch_domain.example2: Refreshing state... (ID: arn:aws:es:eu-west-1:340242599322:domain/tf-test-2)

  ~ update in-place

  ~ aws_elasticsearch_domain.example2
      elasticsearch_version: "5.5" => "6.2"


Note: You didn't specify an "-out" parameter to save this plan, so Terraform
can't guarantee that exactly these actions will be performed if
"terraform apply" is subsequently run.
```
I'm finished for the day now so can't take a longer look until probably next week if you're still stuck but if you do get it working I'd be more than happy to cherry pick your commits on to this change.

rifelpet:
I just updated that commit based on your feedback and forked your branch and it appears to be working as expected. https://github.com/rifelpet/terraform-provider-aws/tree/es_upgrade

```
  ~ aws_elasticsearch_domain.valid_in_place
      elasticsearch_version: "5.5" => "5.6"
-/+ aws_elasticsearch_domain.force_new (new resource required)
      id:                    "arn:aws:es:us-east-1:0000000000:domain/abc123" => <computed> (forces new resource)
      elasticsearch_version: "5.5" => "6.3" (forces new resource)
-/+ aws_elasticsearch_domain.invalid_version (new resource required)
      id:                    "arn:aws:es:us-east-1:0000000000:domain/abcdefg" => <computed> (forces new resource)
      elasticsearch_version: "6.3" => "6.4" (forces new resource)
```

tomelliff:
Awesome, had a quick play with the branch and it seems to work now. I've left some feedback on your commit https://github.com/rifelpet/terraform-provider-aws/commit/0ad0cc77d16e2dbcdedb503dfeefe8d095dcc667 that it would be cool to see if you agree with it or not.

I'd also like to think about the acceptance test(s) a little bit more and if we can test that it actually does do in place/not in place upgrades for expected paths as right now all it proves is that the version does in fact change but doesn't tell you if it did an in place upgrade or destroyed and rebuilt the ES domain.

@bflad is this something you'd want to see added before this is merged? Without spending much time thinking about it I'm not sure what the best way is to detect this right now but can take a longer look at it or if you happen to know of a good way to do that then a pointer in the right direction would be much appreciated. I know the `aws_instance` resource has a similar test but that's made easier by EC2 instances having a unique ID that would change on each rebuild (not yet sure if the `endpoint` on the ES domain changes on rebuild or if it's like RDS and the "random looking" part of the endpoint is actually seeded by account ID and region or something).

bflad:
@tomelliff generally by preference, we can check if:
* ID changed (e.g. `testAccCheckAwsSubnetNotRecreated()`)
* Created/updated timestamp changed (e.g. `testAccCheckAWSEc2FleetNotRecreated()`)
* Some other unique, computed attribute changed

For Elasticsearch domains, it looks like `DomainId` might be a good candidate: https://docs.aws.amazon.com/sdk-for-go/api/service/elasticsearchservice/#ElasticsearchDomainStatus

Here's an example implementation of one:

https://github.com/terraform-providers/terraform-provider-aws/blob/5aaf64f80342d36f55185b778a372bf50175813d/aws/resource_aws_ec2_fleet_test.go#L1105-L1113

tomelliff:
`DomainId` is sadly just the account ID and the domain name concatenated together with a `/` separator so that won't work. I just checked and it looks like the endpoint also comes back as the same when recreating the ES domain just like RDS does too :cry:

I can't see anything that exposes anything unique in the calls we're making right now but we could look at the `DescribeElasticsearchDomainConfig` call which has a `CreationDate` under `out.DomainConfig.ElasticsearchClusterConfig.Status`.

Want to block this until I've had a chance to work through necessary changes for the tests?

rifelpet:
@tomelliff I addressed most of your feedback in a new commit to my branch. I can squash it when we're ready. Regarding returning an error if the GetCompatibleElasticsearchVersions call fails, I'm not sure we can do that. The function signature doesn't return an error type, and none of the other instances of `customdiff.ForceNewIf` in the provider use any error handling. I wasnt sure whether in that case it should default to forcing a new resource or not, or if theres a different function we can use that does return an error.

Regarding the GetCompatibleElasticsearchVersions response:

If a version has no compatible upgrades, it returns an empty array. This is a 2.3 cluster:
```json
{
    "CompatibleElasticsearchVersions": []
}
```

If it does have compatible upgrades, the response looks like:
```json
{
    "CompatibleElasticsearchVersions": [
        {
            "SourceVersion": "5.5",
            "TargetVersions": [
                "5.6"
            ]
        }
    ]
}
```

I believe `CompatibleElasticsearchVersions` is a list because the `DomainName` parameter is optional which results in returning a list of all source versions and their possible target versions. Since we're always specifying a specific `DomainName`, I believe the `CompatibleElasticsearchVersions` list will always have a length 0 or 1, though no documentation confirms this. The for loop in the `ForceNewIf`function should handle possible future scenarios where a `SourceVersion` can have many `TargetVersion`s.

tomelliff:
Derp, misread the thing you were length checking, that's fine, sorry.

Changes look good to me if you want to squash and I'll cherry pick those changes on to this branch and actually push it up to my fork.

rifelpet:
I squashed my branch, feel free to cherry pick the commit.

tomelliff:
@rifelpet done, thanks for that, really nice addition.

I'll take a look at the acceptance test(s) to test when we do/don't delete the cluster when updating the version when I get a chance later this week.

rifelpet:
Do we know how terraform will behave if an upgrade takes longer than the 1h timeout? Will it try to re-upgrade, which will presumably fail? There is a `GetUpgradeStatus` API call to get the intermediate state of the cluster, maybe we can utilize that somehow?

Most of the clusters I manage with terraform are over a terrabyte of data so I think I'll be hitting the timeout but I wont be able to test it until this gets merged and released.

tomelliff:
I don't have a good way to test that without writing a ton of useless data to an index that I'm happy to murder during testing but if you could snap an existing large cluster and restore it into another domain that you are happy to play with that would be good.

I think Terraform will time out and throw an error. If the user then attempts to run an upgrade while the existing upgrade is in process then the AWS API will throw an error. If they wait until the upgrade is complete then the refresh will then show nothing to do on the plan.

Probably worth checking that assumption though and yeah we do need to think about configurable timeouts but I'd rather leave that to a separate, smaller PR for now to minimise the change scope.

rifelpet:
@bflad what work is remaining to get this merged?

tomelliff:
I need to make the acceptance test actually verify we didn't delete the ES cluster when upgrading. It requires a bit of rework of how we're testing things and unfortunately I've only had time to contribute to open source projects on my commute where I don't have a stable enough internet connection to be building and deleting ES clusters because they take so damn long to apply and destroy.

Hopefully I'll get some time over the Christmas break or possibly even this weekend.

tomelliff:
Sigh. Just realised I was comparing the same created time to itself by making a copy/paste error.

Force pushed fix back over the top and rerunning the insanely long acceptance test now.

tomelliff:
Acceptance test completed fine:
```
make testacc TESTARGS="-run=TestAccAWSElasticSearchDomain_update_version"
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSElasticSearchDomain_update_version -timeout 120m
=== RUN   TestAccAWSElasticSearchDomain_update_version
=== PAUSE TestAccAWSElasticSearchDomain_update_version
=== CONT  TestAccAWSElasticSearchDomain_update_version
--- PASS: TestAccAWSElasticSearchDomain_update_version (3789.38s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	3789.394s
```

bflad:

PR #6247, opened 2018-10-23
resouce/aws_waf_rule: add import ability

acburdine:

* Add import ability to waf rule (following similar approach as https://github.com/terraform-providers/terraform-provider-aws/commit/742357c2701fd4d63e4a6d7e030f7c27e6ea98d5)

```
=== RUN   TestAccAWSWafRule_basic
=== PAUSE TestAccAWSWafRule_basic
=== CONT  TestAccAWSWafRule_basic
--- PASS: TestAccAWSWafRule_basic (16.65s)
PASS
```

### DIFF
--- a/aws/resource_aws_waf_rule_test.go
+++ b/aws/resource_aws_waf_rule_test.go
@@ -34,6 +34,11 @@ func TestAccAWSWafRule_basic(t *testing.T) {
 						"aws_waf_rule.wafrule", "metric_name", wafRuleName),
 				),
 			},
+			{
+				ResourceName:      "aws_waf_rule.wafrule",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
 		},
 	})
 }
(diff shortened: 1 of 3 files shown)


aws/resource_aws_waf_rule_test.go
```diff
@@ -34,6 +34,11 @@ func TestAccAWSWafRule_basic(t *testing.T) {
 						"aws_waf_rule.wafrule", "metric_name", wafRuleName),
 				),
 			},
+			{
```
bflad: Could you add this `TestStep` to all the acceptance tests? That helps us verify that all resource configurations are imported correctly. Thanks!

bflad: Actually since most of the tests aren't verifying additional configurations other than `TestAccAWSWafRule_geoMatchSetPredicate`, we can probably just get this in as-is.

PR #6285, opened 2018-10-26
Fix bad advice on using ".name" property of Role while ignoring path

rehevkor5:

- When creating an IAM Role with a non-default path (anything other than
"/"), it is not sufficient to use only the "name" property of the Role
when referring to it from the Glue Crawler resource (and probably
others as well). Therefore, in the examples I have replaced usage of
"name" with usage of "arn", and I updated the documentation on the
"role" argument of the aws_glue_crawler resource to clarify the meaning
of "The IAM role" which was not well-defined.
- Relevant Terraform version: This update can be deployed to the site
immediately, it is not specific to an upcoming version of Terraform.

Not performed.

### DIFF
--- a/website/docs/r/glue_crawler.html.markdown
+++ b/website/docs/r/glue_crawler.html.markdown
@@ -18,7 +18,7 @@ Manages a Glue Crawler. More information can be found in the [AWS Glue Developer
 resource "aws_glue_crawler" "example" {
   database_name = "${aws_glue_catalog_database.example.name}"
   name          = "example"
-  role          = "${aws_iam_role.example.name}"
+  role          = "${aws_iam_role.example.arn}"
 
   dynamodb_target {
     path = "table-name"
@@ -32,7 +32,7 @@ resource "aws_glue_crawler" "example" {
 resource "aws_glue_crawler" "example" {
   database_name = "${aws_glue_catalog_database.example.name}"
   name          = "example"
-  role          = "${aws_iam_role.example.name}"
+  role          = "${aws_iam_role.example.arn}"
 
   jdbc_target {
     connection_name = "${aws_glue_connection.example.name}"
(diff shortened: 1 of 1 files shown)


rehevkor5:
Related: https://github.com/terraform-providers/terraform-provider-aws/pull/6287

bflad:
So while this change is for the better, it turns out it does not work immediately out of the box. 😅 It looks like even if you pass in the IAM role via ARN, the Glue API will return it with just the name. I found this by adjusting the `role` references in `aws/resource_glue_crawler_test.go` and running the acceptance testing via: `make testacc TEST=./aws TESTARGS='-run=TestAccAWSGlueCrawler_'`

(other test results similar)
```
--- FAIL: TestAccAWSGlueCrawler_DynamodbTarget (18.09s)
    testing.go:538: Step 0 error: After applying this step, the plan was not empty:

        DIFF:

        UPDATE: aws_glue_crawler.test
          role: "tf-acc-test-5817417459117994664" => "arn:aws:iam::123456789012:role/tf-acc-test-5817417459117994664"
```

I will file this as a separate bug as we'll need to suppress that difference in both the case of a role with and without a path. I'm not sure if we should merge in this documentation update immediately though since at least a role name without path should work in some cases for new practitioners while ARN will not work in any case unless `ignore_changes` is used.

bflad:
Bug report submitted: #6292

I'll try to get in a fix for that right now.

bflad:
This pull request can be merged after #6293 👍

rehevkor5:
@bflad Sorry about that. Thanks for checking & helping. I guess there were some aspects of this I missed due to not running the acceptance tests.

bflad:
@rehevkor5 are your IAM roles in a different AWS account? Maybe the API takes that into consideration, where it attempts to simplify its response when the role is in the same account as the Glue crawler. 😄 Regardless though, #6293 is merged which covers suppressing the Name -> ARN difference for whatever situation our testing account against us-west-2 was running into so this is good to go now!

PR #6287, opened 2018-10-26
Add "name_with_path" to iam_role data source and resource

rehevkor5:

Add "name_with_path" to iam_role data source and resource

- Provides the "friendly name" plus the path without leading slash,
which in some cases must be used instead of "id"/friendly name when the
role has a non-default path (anything but "/").

Not run (sorry).

### DIFF
--- a/aws/data_source_aws_iam_role.go
+++ b/aws/data_source_aws_iam_role.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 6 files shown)


bflad:
Hi again, @rehevkor5! 👋 Was this intended for any use case beyond the one we just "fixed" with Glue crawlers? I would imagine most, if not all, AWS service calls that accept a IAM role name also accept a IAM role ARN in some fashion for the path problem.

While this is a fairly simple addition, I think we should try to not introduce additional code if it does not have strong use cases. If there are others, we would love to know. Thanks!

rehevkor5:
@bflad My main concern was the issue with Glue Crawlers. I don't know whether there may be other places in the AWS API which work the same way or not. I assume it's possible to use the built-in functions of the config language to calculate this value manually, so now that Crawler Role ARN works properly & the documentation has been updated I think it's ok. I understand the motivation for omitting something that might not get a lot of use. Feel free to close, thanks for your attention.

PR #6294, opened 2018-10-29
Update default_network_acl.html.markdown

GMartinez-Sisti:
The `aws_default_network_acl` is a resource users tend to struggle with. In environments where one can't control all the subnets, adding them all to the default ACL is not practical or sometimes possible.

Using this lifecycle configuration will prevent a plan to show a change that won't be applied.


GMartinez-Sisti:
Thanks @bflad!

PR #6309, opened 2018-10-31
Update cognito_user_pool.markdown

ehault:
Fix documentation error on the verify by link for Cognito Pool.

With {##VERIFY EMAIL##} we get:
 
```
* module.cognito.aws_cognito_user_pool.pool: 1 error(s) occurred:

* aws_cognito_user_pool.pool: Error updating Cognito User pool: InvalidParameterException: The message is invalid.
        status code: 400, request id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxx
```

But with this change ( {##Click Here##} ) we get the proper response:

```
Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

### DIFF
--- a/website/docs/r/cognito_user_pool.markdown
+++ b/website/docs/r/cognito_user_pool.markdown
@@ -115,7 +115,7 @@ The following arguments are supported:
 
   * `default_email_option` (Optional) - The default email option. Must be either `CONFIRM_WITH_CODE` or `CONFIRM_WITH_LINK`. Defaults to `CONFIRM_WITH_CODE`.
   * `email_message` (Optional) - The email message template. Must contain the `{####}` placeholder. **NOTE:** - If `email_verification_message` and `verification_message_template.email_message` are specified and the values are different, either one is prioritized and updated.
-  * `email_message_by_link` (Optional) - The email message template for sending a confirmation link to the user, it must contain the `{##VERIFY EMAIL##}` placeholder.
+  * `email_message_by_link` (Optional) - The email message template for sending a confirmation link to the user, it must contain the `{##Click Here##}` placeholder.
   * `email_subject` (Optional) - The subject line for the email message template. **NOTE:** - If `email_verification_subject` and `verification_message_template.email_subject` are specified and the values are different, either one is prioritized and updated.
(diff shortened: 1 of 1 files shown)


bflad:
Hi @ehault 👋 Could you provide some additional context about your configuration? I just verified that `{##VERIFY EMAIL##}` successfully wrote to Cognito by updating this line:

https://github.com/terraform-providers/terraform-provider-aws/blob/d095a2c72b372b405fdc1ed337024c8f7dc4bc71/aws/resource_aws_cognito_user_pool_test.go#L1126

And running the acceptance test (without updating the test condition, so it fails as expected):

```
--- FAIL: TestAccAWSCognitoUserPool_withVerificationMessageTemplate (9.12s)
    testing.go:538: Step 0 error: Check failed: 1 error occurred:
        	* Check 4/7 error: aws_cognito_user_pool.pool: Attribute 'verification_message_template.0.email_message_by_link' expected "{##foobar##}", got "{##VERIFY EMAIL##}"
```

Is your message almost 20,000 characters long by chance? The [API reference for `VerificationMessageTemplateType`](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_VerificationMessageTemplateType.html) notes these restrictions (which we do not currently validate during plan time), however the content within `{##...##}` shouldn't matter either which way between the old and new configuration you provided.

> **EmailMessageByLink**
>
> The email message template for sending a confirmation link to the user.
> 
> Type: String
> 
> Length Constraints: Minimum length of 6. Maximum length of 20000.
> 
> Pattern: [\p{L}\p{M}\p{S}\p{N}\p{P}\s*]*\{##[\p{L}\p{M}\p{S}\p{N}\p{P}\s*]*##\}[\p{L}\p{M}\p{S}\p{N}\p{P}\s*]*
> 
> Required: No

ehault:
I updated my pool to go from "CONFIRM_WITH_CODE" to "CONFIRM_WITH_LINK",
and updated the email to have {##VERIFY EMAIL##} and that failed with the
above error message. Then after some google-fu I found an old post saying
to use the {##CLICK HERE##} approach, which then worked.

Not sure what that could be about though...

Thanks for your help, do not hesitate to drop this altogether.

Sincerely

Thibault LEPEU

Issue #5646, opened 2018-08-22
Tech Debt: Fix deadcode Lint Issues and Enable in TravisCI Testing


bflad:
- [ ] Fix existing `gometalinter --disable-all --enable deadcode ./aws` issues

```
aws/autoscaling_tags.go:252:1:warning: autoscalingTagsToMap is unused (deadcode)
aws/autoscaling_tags.go:267:1:warning: autoscalingTagDescriptionsToMap is unused (deadcode)
aws/network_acl_entry.go:61:1:warning: flattenNetworkAclEntries is unused (deadcode)
aws/resource_aws_acmpca_certificate_authority.go:694:1:warning: listAcmpcaCertificateAuthorities is unused (deadcode)
aws/resource_aws_batch_job_queue.go:242:1:warning: batchJobQueueRefreshStateFunc is unused (deadcode)
aws/resource_aws_codebuild_project_migrate.go:11:1:warning: resourceAwsCodebuildMigrateState is unused (deadcode)
aws/resource_aws_codepipeline.go:168:1:warning: validateAwsCodePipelineStageActionConfiguration is unused (deadcode)
aws/resource_aws_cognito_identity_pool_roles_attachment.go:267:1:warning: cognitoRoleMappingHash is unused (deadcode)
aws/resource_aws_cognito_identity_pool_roles_attachment.go:275:1:warning: cognitoRoleMappingValueHash is unused (deadcode)
aws/resource_aws_cognito_identity_pool_roles_attachment.go:286:1:warning: cognitoRoleMappingRulesConfigurationHash is unused (deadcode)
aws/resource_aws_dynamodb_table.go:704:1:warning: readDynamoDbPITR is unused (deadcode)
aws/resource_aws_ecs_service.go:951:1:warning: parseTaskDefinition is unused (deadcode)
aws/resource_aws_elasticache_replication_group_migrate.go:11:1:warning: resourceAwsElasticacheReplicationGroupMigrateState is unused (deadcode)
aws/resource_aws_emr_cluster.go:1096:1:warning: loadGroups is unused (deadcode)
aws/resource_aws_gamelift_fleet.go:435:1:warning: flattenGameliftIpPermissions is unused (deadcode)
aws/resource_aws_route53_health_check.go:383:1:warning: createChildHealthCheckList is unused (deadcode)
[...]

PR #6331, opened 2018-11-02
Removed/moved all dead code

mbfrahry:
This PR removes or moves all deadcode from ./aws

```
$gometalinter --disable-all --enable deadcode ./aws
$
```


bflad:
This can also have Closes #5646 added to the PR comment so it does just that. 😄

bflad:
Post-merge LGTM 🚀

PR #5397, opened 2018-07-31
New Resource aws_gamelift_queue

monrostar:
https://github.com/terraform-providers/terraform-provider-aws/issues/2546

Changes proposed in this pull request:

* New Resource `aws_gamelift_queue` with `aws_gamelift_queue_test`

```bash
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSGameliftQueue_basic'
```

Output from acceptance testing:
```bash
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -run=TestAccAWSGameliftQueue_basic -timeout 120m
=== RUN   TestAccAWSGameliftQueue_basic
--- PASS: TestAccAWSGameliftQueue_basic (46.79s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	46.804s
```

Issue #2546, opened 2017-12-05
Amazon GameLift support


Hinidu:
Hello guys! Thank you for your work!
There are no resources for [Amazon GameLift](https://aws.amazon.com/gamelift) and I can't find any related issues or discussions here or in Google.
I would like to implement some GameLift resources and contribute them back if needed. I think I will have some questions especially because some of GameLift resources are immutable and can't be deleted - I didn't use any terraform resources with such properties before and I'm not sure what would be the convenient way to work with such resources.
Does anybody have some thoughts about it?

Hinidu:
> If you could just sum up for every resources you would like to manage how this works on the AWS side, we would better help & guide you :)

Thanks for the feedback! I will post more detailed information about all types of resources that I need soon.

> If those are immutable, how can one object disappear on the AWS side? expiration?

For now matchmaking rule sets just can't be deleted and they are never expired AFAIK. It is quite new part of the service and they have the deletion feature in the backlog.

alanmur:
Hello @Hinidu, I am a Solutions Architect at AWS and I support customers using Amazon GameLift. Please contact me by email at alanmur at amazon dot com for more help using Amazon GameLift. You can also view the documentation at: https://aws.amazon.com/documentation/gamelift/

radeksimko:
@Hinidu FYI: We support `aws_gamelift_build` since [`1.8.0`](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#180-january-29-2018) and I just opened #3327 for `aws_gamelift_fleet`.

Once that gets merged I have `aws_gamelift_alias` ready for PR in a branch. These 3 resources should give us at least some baseline to work with.

Hinidu:
Hi @radeksimko! It looks very nice! Thank you very much!
We don't upload our builds to S3 before uploading them to GameLift but we can add this step for sure. Actually I didn't think whether we will actually upload game server builds using terraform but this opportunity looks interesting.
That's the list of GameLift resource types we are using now: builds, fleets, matchmaking rule sets, matchmaking configurations and queues.
Probably the hardest thing is that matchmaking rule sets are immutable and can't be deleted.
I hope I will have some time soon to use your new resources.

bflad:
Hi Everyone 👋 

The Terraform AWS Provider currently supports multiple GameLift resources for managing this service:

[...]

PR #6335, opened 2018-11-02
Gamelift session queue

Hinidu:
Replacement for #5397 
Original issue #2546

* New resource `aws_gamelift_session_queue`

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSGameliftQueue_basic'
TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccAWSGameliftQueue_basic -timeout 120m
=== RUN   TestAccAWSGameliftQueue_basic
--- PASS: TestAccAWSGameliftQueue_basic (38.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	38.630s
```

The only think I didn't changed is `playerLatencyPolicies`:
> Can there be multiple of these configurations? If not, we should set MaxItems: 1 for this attribute +1

Because there can be many policies.

Also I'm going to write the docs for this resource soon.

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -442,6 +442,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_gamelift_alias":                               resourceAwsGameliftAlias(),
 			"aws_gamelift_build":                               resourceAwsGameliftBuild(),
 			"aws_gamelift_fleet":                               resourceAwsGameliftFleet(),
+			"aws_gamelift_queue":                               resourceAwsGameliftQueue(),
 			"aws_glacier_vault":                                resourceAwsGlacierVault(),
 			"aws_glue_catalog_database":                        resourceAwsGlueCatalogDatabase(),
 			"aws_glue_catalog_table":                           resourceAwsGlueCatalogTable(),
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -442,7 +442,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_gamelift_alias":                               resourceAwsGameliftAlias(),
 			"aws_gamelift_build":                               resourceAwsGameliftBuild(),
 			"aws_gamelift_fleet":                               resourceAwsGameliftFleet(),
-			"aws_gamelift_queue":                               resourceAwsGameliftQueue(),
+			"aws_gamelift_session_queue":                       resourceAwsGameliftSessionQueue(),
 			"aws_glacier_vault":                                resourceAwsGlacierVault(),
 			"aws_glue_catalog_database":                        resourceAwsGlueCatalogDatabase(),
 			"aws_glue_catalog_table":                           resourceAwsGlueCatalogTable(),
--- a/aws/resource_aws_gamelift_queue.go
+++ b/aws/resource_aws_gamelift_queue.go
@@ -11,10 +11,10 @@ import (
 
 func resourceAwsGameliftQueue() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceAwsGameliftQueuesCreate,
-		Read:   resourceAwsGameliftQueuesRead,
-		Update: resourceAwsGameliftQueuesUpdate,
-		Delete: resourceAwsGameliftQueuesDelete,
+		Create: resourceAwsGameliftQueueCreate,
+		Read:   resourceAwsGameliftQueueRead,
+		Update: resourceAwsGameliftQueueUpdate,
+		Delete: resourceAwsGameliftQueueDelete,
 
 		Schema: map[string]*schema.Schema{
 			"destinations": {
@@ -58,7 +58,7 @@ func resourceAwsGameliftQueue() *schema.Resource {
 	}
 }
 
-func resourceAwsGameliftQueuesCreate(d *schema.ResourceData, meta interface{}) error {
+func resourceAwsGameliftQueueCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).gameliftconn
 
 	input := getFullInputCreate(d)
@@ -74,10 +74,10 @@ func resourceAwsGameliftQueuesCreate(d *schema.ResourceData, meta interface{}) e
 	d.SetId(*out.GameSessionQueue.GameSessionQueueArn)
 	d.Set("name", out.GameSessionQueue.Name)
 
-	return resourceAwsGameliftQueuesRead(d, meta)
+	return resourceAwsGameliftQueueRead(d, meta)
 }
 
-func resourceAwsGameliftQueuesRead(d *schema.ResourceData, meta interface{}) error {
+func resourceAwsGameliftQueueRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).gameliftconn
 	log.Printf("[INFO] Describing Gamelift Session Queues: %s", d.Get("name"))
 	limit := int64(1)
@@ -112,7 +112,7 @@ func resourceAwsGameliftQueuesRead(d *schema.ResourceData, meta interface{}) err
 	return nil
 }
 
-func resourceAwsGameliftQueuesUpdate(d *schema.ResourceData, meta interface{}) error {
+func resourceAwsGameliftQueueUpdate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).gameliftconn
 
 	name := d.Get("name").(string)
@@ -130,10 +130,10 @@ func resourceAwsGameliftQueuesUpdate(d *schema.ResourceData, meta interface{}) e
 		}
 	}
 
-	return resourceAwsGameliftQueuesRead(d, meta)
+	return resourceAwsGameliftQueueRead(d, meta)
 }
 
-func resourceAwsGameliftQueuesDelete(d *schema.ResourceData, meta interface{}) error {
+func resourceAwsGameliftQueueDelete(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).gameliftconn
 	name := d.Get("name").(string)
 	log.Printf("[INFO] Deleting Gamelift Session Queue: %s", name)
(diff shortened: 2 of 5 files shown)


Hinidu:
@bflad I think I did everything you asked for and this PR is ready for another review.

Hinidu:
I've started to implement a resource for matchmaking configurations but to properly test it I need a game session queue because game session queue ARN is one of the required parameters.

aws/resource_aws_gamelift_game_session_queue.go
```diff
@@ -0,0 +1,204 @@
...
+		PlayerLatencyPolicies: expandGameliftGameSessionPlayerLatencyPolicies(d.Get("player_latency_policy").([]interface{})),
+		TimeoutInSeconds:      aws.Int64(int64(d.Get("timeout_in_seconds").(int))),
+	}
+	log.Printf("[INFO] Creating Gamelift Session Queue: %s", input)
+	out, err := conn.CreateGameSessionQueue(&input)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(*out.GameSessionQueue.GameSessionQueueArn)
```
bflad: Since the `Read`, `Update`, and `Delete` functions can all work off the `name` attribute, I'd recommend switch setting the ID to that and referencing via `d.Id()` elsewhere. If/when we want to support resource import, then we can simply use this in the schema definition:

```go
},
```

Tested by adding a `TestStep` like this to each acceptance test:

```go
{
  ImportState:       true,
  ImportStateVerify: true,
},
```

And document via:

``````markdown
## Import

Gamelift Game Session Queues can be imported by their `name`, e.g.

```
$ terraform import aws_gamelift_game_session_queue.example example
```
``````

File: aws/resource_aws_gamelift_game_session_queue.go [hunk 2]
```diff
@@ -0,0 +1,204 @@
...
+	d.Set("name", out.GameSessionQueue.Name)
```
bflad: If the ID is changed, this `d.Set()` during the `Create` function is extraneous. We generally only recommend `d.Set()` during `Read` for simplicity.

aws/resource_aws_gamelift_game_session_queue.go
```diff
@@ -0,0 +1,204 @@
...
+	if len(sessionQueues) != 1 {
+		return fmt.Errorf("expected exactly 1 Gamelift Session Queues, found %d under %q",
+			len(sessionQueues), d.Get("name"))
+	}
+	sessionQueue := sessionQueues[0]
+
+	d.Set("arn", sessionQueue.GameSessionQueueArn)
+	d.Set("name", sessionQueue.Name)
+	d.Set("timeout_in_seconds", sessionQueue.TimeoutInSeconds)
+	d.Set("destinations", sessionQueue.Destinations)
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("destinations", sessionQueue.Destinations); err != nil {
  return fmt.Errorf("error setting destinations: %s", err)
}
```

After adding this error checking, you should start receiving type errors as `sessionQueue.Destinations` is a `[]*GameSessionQueueDestination` SDK object and not a `[]interface{}`. We generally handle this via a "flattenServiceObject()" function. e.g.

```go
func flattenGameliftGameSessionQueueDestinations(destinations []*gamelift.GameSessionQueueDestination) []interface{} {
	l := make([]interface{}, 0)

	for _, destination := range destinations {
		if destination == nil {
			continue
		}
		l = append(l, aws.StringValue(destination.DestinationArn))
	}

	return l
}
```

Which can be referenced via:

```go
if err := d.Set("destinations", flattenGameliftGameSessionQueueDestinations(sessionQueue.Destinations)); err != nil {
  return fmt.Errorf("error setting destinations: %s", err)
}
```

aws/resource_aws_gamelift_game_session_queue.go
```diff
@@ -0,0 +1,204 @@
...
+	input := gamelift.CreateGameSessionQueueInput{
+		Name:                  aws.String(d.Get("name").(string)),
+		Destinations:          expandGameliftGameSessionQueueDestinations(d.Get("destinations").([]interface{})),
+		PlayerLatencyPolicies: expandGameliftGameSessionPlayerLatencyPolicies(d.Get("player_latency_policy").([]interface{})),
+		TimeoutInSeconds:      aws.Int64(int64(d.Get("timeout_in_seconds").(int))),
+	}
+	log.Printf("[INFO] Creating Gamelift Session Queue: %s", input)
+	out, err := conn.CreateGameSessionQueue(&input)
+	if err != nil {
+		return err
```
bflad: Nit: We should return context about when errors occur for operators and code maintainers, e.g.

```suggestion
		return fmt.Errorf("error creating Gamelift Game Session Queue: %s", err)
```

aws/resource_aws_gamelift_game_session_queue.go
```diff
@@ -0,0 +1,204 @@
...
+		Names: aws.StringSlice([]string{d.Get("name").(string)}),
+		Limit: &limit,
+	})
+	if err != nil {
+		if isAWSErr(err, gamelift.ErrCodeNotFoundException, "") {
+			log.Printf("[WARN] Gamelift Session Queues (%s) not found, removing from state", d.Get("name"))
+			d.SetId("")
+			return nil
+		}
+		return err
```
bflad: Nit: We should return context about when errors occur for operators and code maintainers, e.g.

```suggestion
		return fmt.Errorf("error reading Gamelift Game Session Queue (%s): %s", d.Id(), err)
```

aws/resource_aws_gamelift_game_session_queue.go
```diff
@@ -0,0 +1,204 @@
...
+	input := gamelift.UpdateGameSessionQueueInput{
+		Name:                  aws.String(d.Get("name").(string)),
+		Destinations:          expandGameliftGameSessionQueueDestinations(d.Get("destinations").([]interface{})),
+		PlayerLatencyPolicies: expandGameliftGameSessionPlayerLatencyPolicies(d.Get("player_latency_policy").([]interface{})),
+		TimeoutInSeconds:      aws.Int64(int64(d.Get("timeout_in_seconds").(int))),
+	}
+
+	_, err := conn.UpdateGameSessionQueue(&input)
+	if err != nil {
+		return err
```
bflad: Nit: We should return context about when errors occur for operators and code maintainers, e.g.

```suggestion
		return fmt.Errorf("error updating Gamelift Game Session Queue (%s): %s", d.Id(), err)
```

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: Is it ok to use `name` as ID even if `name` can be changed?

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: Just curious - while we should skip `nil` values in `destinations`? How they can get there? Why it is not the error if we got such destination?

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: :+1:

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: :+1:

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: :+1:

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: :+1:

aws/resource_aws_gamelift_game_session_queue.go
bflad: Can the `name` be changed? 😄 Usually they have a `NewName` or something parameter in the update API call (or `Name` with some other more specific ID), but I'm not seeing that here: https://docs.aws.amazon.com/gamelift/latest/apireference/API_UpdateGameSessionQueue.html

If you're correct though, then we would likely want to stick with ARN. 👍

aws/resource_aws_gamelift_game_session_queue.go
bflad: This is just a safety check in case the API does something unexpected. Terraform will completely stop running (panic) if you try to perform a dereference against a `nil` value.

aws/resource_aws_gamelift_game_session_queue.go
Hinidu: Yeah, it’s just the `Name`. It can be changed both using UI or API and it is even changing in the test case.

Hinidu: Though what about the import in this case? I don’t mind to implement it as well. If I understand correctly we should implement import by ARN?

bflad: The attribute in Terraform appears to be correctly marked `ForceNew: true` so it should be deleting the old queue and recreating the new one. When I create a queue in the UI, the option to edit the queue name is grayed out. Am I missing something?

screen shot 2018-11-14 at 12 42 47 am

Hinidu: Oh, sorry! It’s my bad. I will replace ARN by name soon. Thank you!

Hinidu:
Hello @bflad! I've replaced `ARN` by `Name` as `ID` and added the `import`.

Hinidu:
Nice! Thank you @bflad for your feedback and merge! Hopefully I will try to implement other GameLift resources soon.

bflad:

Issue #6363, opened 2018-11-05
Differences in Tags are not calculated correctly


fbreckle:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Hi,

terraform seems to recreate all Tags on an instance, even Tags that were unchanged.
This is a problem if recreation of a tag is not allowed by IAM permissions and in general unnecessary behavior. The aws API seems to silently succeed when creating duplicate tags.

### Terraform Version
$ terraform -v
Terraform v0.11.10
+ provider.aws v1.42.0

### Affected Resource(s)

* aws_instance (but anything that is getting tagged probably)

### Terraform Configuration Files

Start with

```hcl
provider "aws" {
  region = "eu-west-1"
  version = "1.42"
}

resource "aws_instance" "web" {
  ami           = "ami-0a5e707736615003c"
  instance_type = "t2.micro"

  tags {
    Origin = "Terraform"
  }
}
```

then add another tag "Testtag" to the instance as per 

```hcl
tags {
    Origin = "Terraform"
    Testtag = "Testy"
  }
```
### Debug Output
`
2018-11-05T18:21:17.617+0100 [DEBUG] plugin.terraform-provider-aws_v1.42.0_x4: 2018/11/05 18:21:17 [DEBUG] Creating tags: [{
2018-11-05T18:21:17.617+0100 [DEBUG] plugin.terraform-provider-aws_v1.42.0_x4:   Key: "Origin",
2018-11-05T18:21:17.617+0100 [DEBUG] plugin.terraform-provider-aws_v1.42.0_x4:   Value: "Terraform"
2018-11-05T18:21:17.617+0100 [DEBUG] plugin.terraform-provider-aws_v1.42.0_x4: } {
2018-11-05T18:21:17.617+0100 [DEBUG] plugin.terraform-provider-aws_v1.42.0_x4:   Key: "Testtag",
2018-11-05T18:21:17.617+0100 [DEBUG] plugin.terraform-provider-aws_v1.42.0_x4:   Value: "Testy"
[...]

PR #6370, opened 2018-11-06
Fix diffTags to actually create a diff instead of unnecessarily recre…

fbreckle:
…ating existing tags.

Solves Issue  [#6363](https://github.com/terraform-providers/terraform-provider-aws/issues/6363)

Modify diffTags to behave as explained in its comment:

// diffTags takes our tags locally and the ones remotely and returns
// the set of tags that must be created, and the set of tags that must
// be destroyed.

Many similar functions exist for other aws elements, which all seem to have the same erroneous behavior and can be fixed analogously.


fbreckle:
done!

bflad:

PR #6432, opened 2018-11-11
New Resource: aws_glacier_vault_lock

bflad:
This resource is a little odd because it can only be "destroyed" when in a testing mode (e.g. before the lock is completed). We provide a `ignore_deletion_error` argument to workaround cases of  deleting/recreating the Glacier Vault.

```
--- PASS: TestAccAWSGlacierVaultLock_basic (18.63s)
--- PASS: TestAccAWSGlacierVaultLock_CompleteLock (19.02s)
```


bflad:

PR #6437, opened 2018-11-12
resource/aws_lambda_event_source_mapping: Add starting_position_timestamp argument

bflad:
This argument, like `starting_position`, is only used during resource creation and therefore cannot be updated or read from the Lambda API.

* resource/aws_lambda_event_source_mapping: Add `starting_position_timestamp` argument
* docs/resource/aws_lambda_event_source_mapping: Show DynamoDB, Kinesis, and SQS example usage

```
--- PASS: TestAccAWSLambdaEventSourceMapping_sqs_withFunctionName (43.90s)
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_disappears (85.32s)
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_import (87.35s)
--- PASS: TestAccAWSLambdaEventSourceMapping_StartingPositionTimestamp (87.99s)
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_removeBatchSize (92.56s)
--- PASS: TestAccAWSLambdaEventSourceMapping_kinesis_basic (94.66s)
--- PASS: TestAccAWSLambdaEventSourceMapping_sqs_basic (108.48s)
--- PASS: TestAccAWSLambdaEventSourceMapping_sqsDisappears (121.69s)
--- PASS: TestAccAWSLambdaEventSourceMapping_changesInEnabledAreDetected (122.55s)
```

### DIFF
--- a/aws/resource_aws_lambda_event_source_mapping.go
+++ b/aws/resource_aws_lambda_event_source_mapping.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsLambdaEventSourceMapping() *schema.Resource {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_lambda_event_source_mapping.go
```diff
@@ -136,6 +148,11 @@ func resourceAwsLambdaEventSourceMappingCreate(d *schema.ResourceData, meta inte
 		params.StartingPosition = aws.String(startingPosition.(string))
 	}
 
+	if startingPositionTimestamp, ok := d.GetOk("starting_position_timestamp"); ok {
```
mbfrahry: Do we also need to confirm that `starting_position` is also `lambda.EventSourcePositionAtTimestamp` or can this be specified without it?

bflad: The API returns an error if its incorrectly set (e.g. `starting_position = "LATEST"` and `starting_position_timestamp = "..."): `InvalidParameterValueException: Timestamp only valid with StartingPosition AT_TIMESTAMP`

Since the API returns a helpful error and `CustomizeDiff` has a lot of issues for cross-attribute validation, we try to avoid the extra coding, but I can add it in there if you wish.

mbfrahry: No need! That makes sense to me!

bflad:

PR #6472, opened 2018-11-15
Add name and description for maintenance window targets

nbrys:
Add name and description for maintenance window targets

This pull requests addresses issue 6471

### DIFF
--- a/aws/validators.go
+++ b/aws/validators.go
@@ -1580,6 +1580,19 @@ func validateAwsSSMMaintenanceWindowTaskName(v interface{}, k string) (ws []stri
 	return
 }
 
+func validateAwsSSMMaintenanceWindowTargetName(v interface{}, k string) (ws []string, errors []error) {
+	// https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_RegisterTaskWithMaintenanceWindow.html#systemsmanager-RegisterTaskWithMaintenanceWindow-request-Name
+	value := v.(string)
+
+	if !regexp.MustCompile(`^[a-zA-Z0-9_\-.]{3,128}$`).MatchString(value) {
+		errors = append(errors, fmt.Errorf(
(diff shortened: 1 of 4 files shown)


aws/validators.go
```diff
@@ -1580,6 +1580,19 @@ func validateAwsSSMMaintenanceWindowTaskName(v interface{}, k string) (ws []stri
 	return
 }
 
+func validateAwsSSMMaintenanceWindowTargetName(v interface{}, k string) (ws []string, errors []error) {
+	// https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_RegisterTaskWithMaintenanceWindow.html#systemsmanager-RegisterTaskWithMaintenanceWindow-request-Name
+	value := v.(string)
+
+	if !regexp.MustCompile(`^[a-zA-Z0-9_\-.]{3,128}$`).MatchString(value) {
+		errors = append(errors, fmt.Errorf(
+			"Only alphanumeric characters, hyphens, dots & underscores allowed in %q: %q (Must satisfy regular expression pattern: ^[a-zA-Z0-9_\\-.]{3,128}$)",
```
nywilken: I think we can shorten this message to 

```
errors = append(errors, fmt.Errorf("%q is not a valid value for %q (only alphanumeric characters, dots, underscores and hyphens are allowed)", value, k))
```

nywilken: Actually looking at this function a little further it looks like you can just use the function `validateAwsSSMName` defined on line 1157 and update the error message there.

ryndaniels:
@nbrys since it's been a while since we've seen any update on this, we're going to pick up where you left off to get this finished up. Thanks so much for this contribution!

ryndaniels:
To avoid force-pushing to @nbrys' master branch, opened up #8671 with this commit and my updates instead, with both of us getting contribution credit.

PR #6510, opened 2018-11-18
 ECS Fargate Platform Version Support

trevorlauder:

* This adds support for setting the ECS Fargate Platform Version

This is my first time diving into this code, please let me know if I'm missing anything.  It works in my testing but the one item I'm not sure how to handle is when it's run against a service that isn't using the AWS default of `LATEST`, this change will cause terraform runs to trigger a change which people may not be expecting.

```
$ make testacc TESTARGS='-run=TestAccAWSEcsService_withLaunchTypeFargateAndPlatformVersion'

TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSEcsService_withLaunchTypeFargateAndPlatformVersion -timeout 120m
=== RUN   TestAccAWSEcsService_withLaunchTypeFargateAndPlatformVersion
--- PASS: TestAccAWSEcsService_withLaunchTypeFargateAndPlatformVersion (303.77s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       307.168s
```

### DIFF
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -69,6 +69,13 @@ func resourceAwsEcsService() *schema.Resource {
 				Default:  "EC2",
 			},
 
+			"platform_version": {
+				Type:     schema.TypeString,
+				ForceNew: false,
+				Optional: true,
+				Default:  "LATEST",
+			},
+
 			"scheduling_strategy": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -360,6 +367,10 @@ func resourceAwsEcsServiceCreate(d *schema.ResourceData, meta interface{}) error
 		input.LaunchType = aws.String(v.(string))
 	}
 
+	if v, ok := d.GetOk("platform_version"); ok {
+		input.PlatformVersion = aws.String(v.(string))
+	}
+
 	loadBalancers := expandEcsLoadBalancers(d.Get("load_balancer").(*schema.Set).List())
 	if len(loadBalancers) > 0 {
 		log.Printf("[DEBUG] Adding ECS load balancers: %s", loadBalancers)
@@ -538,6 +549,7 @@ func resourceAwsEcsServiceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("desired_count", service.DesiredCount)
 	d.Set("health_check_grace_period_seconds", service.HealthCheckGracePeriodSeconds)
 	d.Set("launch_type", service.LaunchType)
+	d.Set("platform_version", service.PlatformVersion)
 
 	// Save cluster in the same format
 	if strings.HasPrefix(d.Get("cluster").(string), "arn:"+meta.(*AWSClient).partition+":ecs:") {
@@ -795,6 +807,11 @@ func resourceAwsEcsServiceUpdate(d *schema.ResourceData, meta interface{}) error
 				MinimumHealthyPercent: aws.Int64(int64(d.Get("deployment_minimum_healthy_percent").(int))),
 			}
 		}
+	}
+
+	if d.HasChange("platform_version") {
+		updateService = true
+		input.PlatformVersion = aws.String(d.Get("platform_version").(string))
 	}
 
 	if d.HasChange("health_check_grace_period_seconds") {
--- a/aws/resource_aws_ecs_service_test.go
+++ b/aws/resource_aws_ecs_service_test.go
@@ -594,7 +594,7 @@ func TestAccAWSEcsService_withLaunchTypeFargateAndPlatformVersion(t *testing.T) 
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAWSEcsServiceExists("aws_ecs_service.main", &service),
 					resource.TestCheckResourceAttr("aws_ecs_service.main", "launch_type", "FARGATE"),
-					resource.TestCheckResourceAttr("aws_ecs_service.main", "platform_version", "1.2.0"),
+					resource.TestCheckResourceAttr("aws_ecs_service.main", "platform_version", "LATEST"),
 					resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.assign_public_ip", "false"),
 					resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.security_groups.#", "2"),
 					resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.subnets.#", "2"),
@@ -1322,7 +1322,7 @@ data "aws_availability_zones" "available" {}
 resource "aws_vpc" "main" {
   cidr_block = "10.10.0.0/16"
   tags {
-    Name = "terraform-testacc-ecs-service-with-launch-type-fargate"
+    Name = "terraform-testacc-ecs-service-with-launch-type-fargate-and-platform-version"
   }
 }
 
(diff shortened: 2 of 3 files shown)


aws/resource_aws_ecs_service_test.go
```diff
@@ -574,6 +574,64 @@ func TestAccAWSEcsService_withLaunchTypeFargate(t *testing.T) {
...
+	var service ecs.Service
+	rString := acctest.RandString(8)
+
+	sg1Name := fmt.Sprintf("tf-acc-sg-1-svc-ltf-w-pv-%s", rString)
+	sg2Name := fmt.Sprintf("tf-acc-sg-2-svc-ltf-w-pv-%s", rString)
+	clusterName := fmt.Sprintf("tf-acc-cluster-svc-ltf-w-pv-%s", rString)
+	tdName := fmt.Sprintf("tf-acc-td-svc-ltf-w-pv-%s", rString)
+	svcName := fmt.Sprintf("tf-acc-svc-ltf-w-pv-%s", rString)
+
+	resource.Test(t, resource.TestCase{
```
bflad: We can parallelize this testing with the others via `resource.ParallelTest()` 😄 (dependent on the changes below to remove the extraneous other `resource.Test()` calls)

```suggestion
	resource.ParallelTest(t, resource.TestCase{
```

aws/resource_aws_ecs_service_test.go
```diff
@@ -574,6 +574,64 @@ func TestAccAWSEcsService_withLaunchTypeFargate(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSEcsServiceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSEcsServiceWithLaunchTypeFargateAndPlatformVersion(sg1Name, sg2Name, clusterName, tdName, svcName, "false"),
```
bflad: Ideally, `testAccAWSEcsServiceWithLaunchTypeFargateAndPlatformVersion()` should be accepting the platform version as its final parameter instead of assign public IP, which can just be hardcoded to true/false in the test configuration. This will allow us to properly test updating the platform version across multiple `TestStep` (e.g. `1.0.0` in the first step and `1.1.0` in the second step). 👍

aws/resource_aws_ecs_service_test.go
```diff
@@ -574,6 +574,64 @@ func TestAccAWSEcsService_withLaunchTypeFargate(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSEcsServiceExists("aws_ecs_service.main", &service),
+					resource.TestCheckResourceAttr("aws_ecs_service.main", "launch_type", "FARGATE"),
+					resource.TestCheckResourceAttr("aws_ecs_service.main", "platform_version", "LATEST"),
+					resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.assign_public_ip", "false"),
```
bflad: Since we're not concerned too much with the network configuration when adjusting the platform version, these `network_configuration` checks can be removed in preference of just ensuring we're checking the `platform_version`. 👍

aws/resource_aws_ecs_service_test.go
```diff
@@ -574,6 +574,64 @@ func TestAccAWSEcsService_withLaunchTypeFargate(t *testing.T) {
...
+					resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.security_groups.#", "2"),
+					resource.TestCheckResourceAttr("aws_ecs_service.main", "network_configuration.0.subnets.#", "2"),
+				),
+			},
+		},
+	})
+	resource.Test(t, resource.TestCase{
```
bflad: I'm guessing this acceptance testing was copied from the Fargate acceptance testing above which was unfortunately using multiple invocations of `resource.Test()`, a pattern not used anywhere else in the codebase (in preference of a single `resource.ParallelTest()` with multiple `TestStep`). 😅  Can you move the `TestStep` in these two `resource.Test()` calls to the `Steps` in the first (soon-to-be) `resource.ParallelTest()` above? Thanks! I'll submit a pull request to fix up the other Fargate acceptance test separately.

aws/resource_aws_ecs_service_test.go
```diff
@@ -1254,6 +1312,97 @@ resource "aws_ecs_service" "main" {
 `, sg1Name, sg2Name, clusterName, tdName, svcName, assignPublicIP)
 }
 
+func testAccAWSEcsServiceWithLaunchTypeFargateAndPlatformVersion(sg1Name, sg2Name, clusterName, tdName, svcName, assignPublicIP string) string {
+	return fmt.Sprintf(`
+provider "aws" {
```
bflad: We prefer to not define the AWS provider configuration in the test configurations unless necessary. The below test configuration should work in other regions beyond `us-east-1`, so this provider configuration should be removed. 👍

aws/resource_aws_ecs_service_test.go
```diff
@@ -1254,6 +1312,97 @@ resource "aws_ecs_service" "main" {
...
+DEFINITION
+}
+
+resource "aws_ecs_service" "main" {
+  name = "%s"
+  cluster = "${aws_ecs_cluster.main.id}"
+  task_definition = "${aws_ecs_task_definition.mongo.arn}"
+  desired_count = 1
+  launch_type = "FARGATE"
+  platform_version = "LATEST"
```
bflad: As mentioned above, we should swap the assign public IP parameter for this one here, e.g.

```suggestion
  platform_version = %q
```

website/docs/r/ecs_service.html.markdown
```diff
@@ -80,6 +80,7 @@ The following arguments are supported:
 * `task_definition` - (Required) The family and revision (`family:revision`) or full ARN of the task definition that you want to run in your service.
 * `desired_count` - (Optional) The number of instances of the task definition to place and keep running. Defaults to 0. Do not specify if using the `DAEMON` scheduling strategy.
 * `launch_type` - (Optional) The launch type on which to run your service. The valid values are `EC2` and `FARGATE`. Defaults to `EC2`.
+* `platform_version` - (Optional) The platform version on which to run your service. Defaults to `LATEST`.
```
bflad: Could you please add a link to the AWS documentation and mention this only applies to the `launch_type` being set to `FARGATE`? e.g.

```suggestion
* `platform_version` - (Optional) The platform version on which to run your service. Only applicable for `launch_type` set to `FARGATE`. Defaults to `LATEST`. More information about Fargate platform versions can be found in the [AWS ECS User Guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/platform_versions.html).
```

aws/resource_aws_ecs_service.go
```diff
@@ -69,6 +69,13 @@ func resourceAwsEcsService() *schema.Resource {
 				Default:  "EC2",
 			},
 
+			"platform_version": {
+				Type:     schema.TypeString,
+				ForceNew: false,
+				Optional: true,
+				Default:  "LATEST",
```
bflad: Defaulting this to `LATEST` breaks all the other tests with `launch_type` set to `EC2` (or unset, which defaults to `EC2`) with the following:

```
--- FAIL: TestAccAWSEcsService_withARN (4.53s)
        	* aws_ecs_service.mongo: 1 error occurred:
        	* aws_ecs_service.mongo: InvalidParameterException: The platform version must be null when specifying an EC2 launch type.
```

My recommendation would be to adjust this attribute's schema to the following:

```go
"platform_version": {
	Type:     schema.TypeString,
	Optional: true,
	Computed: true,
},
```

The `Computed: true` here will address your original concern by allowing Terraform plans to ignore any difference if the argument is not defined in the Terraform configuration.

PR #6528, opened 2018-11-20
r/aws_ram_resource_share: Add new resource

### LINKED ISSUE #6527: Feature Request: AWS Resource Access Manager
### Community Note

### Description

Use AWS Resource Access Manager to share AWS resources between AWS accounts. To share a resource, you create a resource share, associate the resource with the resource share, and specify the principals that can access the resource

* `aws_ram_resource_share`

```hcl
resource "aws_ram_resource_share" "example" {
  name                      = "example"
  allow_external_principals = true

  resource_arns = [
    "..."
  ]

  principals = [
    "..."
  ]

  tags {
    Environment = "Production"
  }
}
```

### References

* https://aws.amazon.com/blogs/aws/new-aws-resource-access-manager-cross-account-resource-sharing/
* https://docs.aws.amazon.com/cli/latest/reference/ram/index.html#cli-aws-ram

gazoakley:
Fixes #6527 

```
$ make testacc TESTARGS='-run=TestAccAWSRAMResourceShare'

...
```


bflad:
Hey @gazoakley 👋 I'm guessing this is work in progress due to the acceptance testing? Anything we can help with? Since we now support EC2 Transit Gateway and License Manager Configurations, maybe we can use either of those for testing? Thanks so much for all your great contributions!

aws/config.go
```diff
@@ -143,6 +144,7 @@ type Config struct {
 	KinesisAnalyticsEndpoint string
 	KmsEndpoint              string
 	LambdaEndpoint           string
+	RamEndpoint              string
```
bflad: Since we're not wiring up endpoint customization in the provider configuration (see `aws/provider.go`) we should either implement it or remove it here, e.g. removing this line, `awsRamSess` and using `client.ramconn = ram.New(sess)` 👍

aws/resource_aws_ram_resource_share.go
```diff
@@ -0,0 +1,267 @@
...
+func resourceAwsRamResourceShareCreate(d *schema.ResourceData, meta interface{}) error {
+	ramconn := meta.(*AWSClient).ramconn
+
+	request := &ram.CreateResourceShareInput{
+		Name:                    aws.String(d.Get("name").(string)),
+		AllowExternalPrincipals: aws.Bool(d.Get("allow_external_principals").(bool)),
+	}
+
+	if v, ok := d.GetOk("resource_arns"); ok {
+		request.ResourceArns = expandStringList(v.([]interface{}))
```
bflad: When testing is turned on you'll probably notice immediately this should be `expandStringSet(v.(*schema.Set))` 😄 Same with `principals` below.

gazoakley:
Hi @bflad - at the time I started writing this AWS hadn't announced what you could share using it 😄. I'm planning to write acceptance tests using VPCs since they can be constructed/torn down quickly during acceptance tests.

What I've got kind of works, but I'm thinking it would be better to split `resource_arns` and `principals` out of this resource into some kind of `aws_ram_association`. It looks like adding/removing from/to those can take time, and separating them out would allow Terraform to track the progress of them individually much better. The code currently doesn't wait for associations to move to the "ASSOCIATED"/"DISASSOCIATED" status: https://docs.aws.amazon.com/ram/latest/APIReference/API_ResourceShareAssociation.html

The design then becomes something along the lines of:

```hcl
resource "aws_ram_resource_share" "example" {
  name = "Example"
  allow_external_principals = true
}

resource "aws_ram_association" "example" {
  resource_share_arn = "${aws_ram_resource_share.example.arn}"
  association_type = "PRINCIPAL"
  associated_entity = "123456789012"
}

resource "aws_ram_association" "example" {
  resource_share_arn = "${aws_ram_resource_share.example.arn}"
  association_type = "RESOURCE"
  associated_entity = "<some_vpc_arn>"
}
```

Or better still in my mind:

```hcl
resource "aws_ram_resource_share" "example" {
  name = "Example"
  allow_external_principals = true
}

resource "aws_ram_principal_association" "example" {
  resource_share_arn = "${aws_ram_resource_share.example.arn}"
  principal = "123456789012"
  # external - calculated field that is only applicable to principals associated to a resource share
}

resource "aws_ram_resource_association" "example" {
  resource_share_arn = "${aws_ram_resource_share.example.arn}"
  resource_arn = "<some_vpc_arn>"
}
```

Any thoughts?

bflad:
@gazoakley looking at the API, I would agree with your assessment and the second approach with two associations resources. The time tracking is likely pretty important from an infrastructure management standpoint to allow a single Terraform configuration to perform downstream actions when the associations are actually ready. 👍

If you could split those new resources from this PR, we can likely get this resource merged then work on the association bits separately. 🚀

gazoakley:
@bflad I've done the split - will open PRs for the associations later

gazoakley:
Unfortunately there's an issue in `aws-sdk-go` or the RAM API that prevents us from easily checking if a resource share isn't found: https://github.com/aws/aws-sdk-go/issues/2377

aws/config.go
bflad: This still needs to be addressed either which way. 😄

aws/resource_aws_ram_resource_share.go
```diff
@@ -0,0 +1,241 @@
...
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsRamResourceShare() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsRamResourceShareCreate,
+		Read:   resourceAwsRamResourceShareRead,
+		Update: resourceAwsRamResourceShareUpdate,
```
bflad: We should add at least one acceptance test that exercises `Update` 😄

aws/resource_aws_ram_resource_share.go
```diff
@@ -0,0 +1,241 @@
...
+	_, err := conn.DeleteResourceShare(deleteResourceShareInput)
+	if err != nil {
+		if isAWSErr(err, ram.ErrCodeUnknownResourceException, "") {
+			return nil
+		}
+		return fmt.Errorf("Error deleting RAM resource share %s: %s", d.Id(), err)
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{ram.ResourceShareStatusDeleting},
```
bflad: We may also want to include `ram.ResourceShareStatusActive` in the `Pending` list in case there is any delay the start of the process. 👍

bflad:

piersf:
@bflad the Terraform documentation shows only about `aws_ram_resource_share`. Has there been anything else released?

PR #6569, opened 2018-11-23
Several improvements for eks getting started guide

runningman84:

* Added Mixed spot fleet (to save costs an add availability)
* Changed to 3az (added availability espicially for spot stuff)
* Added 8ks provider to apply configmap

```

...
```


bflad:
Hi @runningman84 👋 Thanks for submitting this!

Generally speaking, the `examples` are provided to give a relatively straightforward/simple combination of multiple Terraform resources to achieve an end goal and we do not particularly try to optimize their configuration or match any specific environment requirements. Also of note for the EKS getting started example, is that is should almost exactly match its guide as they are designed to be updated together: https://www.terraform.io/docs/providers/aws/guides/eks-getting-started.html

Switching from Launch Configurations to Launch Templates is definitely a great change, but the others  seem to fall more under more environmental specific optimizations. If you could change this pull request to only updating the example and guide with a working Launch Template configuration we can get this in. 👍

runningman84:
Hmm the launch config does not work the same way as launch configs therefore I needed to make some changes in vpc and due to spot issues provide a wider list of possible instances. The k8 provider might be too much but if you want to show the power of Terraform why not including these few lines. My example will just work in more cases out of the box... The old one was regularly failing if no specific spot instance was available.

bflad:
Hi again @runningman84 👋 As mentioned previously, we would like to ensure the configuration in the getting started guide focuses only on the bare essentials of setting up an EKS Cluster so new practitioners are not overwhelmed with extra concepts. For this changeset we would only accept the enhancement to switch from the `aws_launch_configuration` resource to `aws_launch_template` resource. Since we are not willing to accept this pull request in its current state, I'm going to close it for now, but if the appropriate changes are made we'll be happy to reopen and review it.

If you're interested in submitting a more advanced example EKS Cluster configuration, I would encourage submitting a pull request to the `examples` directory in this repository, which we can link out to from the provider documentation sidebar or resource pages.

PR #6597, opened 2018-11-26
resource/aws_db_instance: allow configuring-iam-database-auth as pending state

MetricMike:
Currently, terraform errors if you try to enable iam db auth and it takes over 30 seconds. This triggers the default waiter, which throws:
```
* aws_db_instance.default: error waiting for DB Instance (dbname) to be available: unexpected state 'configuring-iam-database-auth', wanted target 'available, storage-optimization'. last error: %!s(<nil>)
```

This state is not listed in the AWS documentation, but I've opened a separate [issue](https://github.com/awsdocs/amazon-rds-user-guide/issues/44) for that.

---

I tried to follow along #3708 which was the most similar change I could find - likewise I didn't see any good place to write tests for this.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -1612,6 +1612,7 @@ func diffCloudwatchLogsExportConfiguration(old, new []interface{}) ([]interface{
 var resourceAwsDbInstanceCreatePendingStates = []string{
 	"backing-up",
 	"configuring-enhanced-monitoring",
+	"configuring-iam-database-auth",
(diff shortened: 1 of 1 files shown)


bflad:
This has been released in [version 1.51.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1510-december-05-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #6679, opened 2018-12-01
docs/resource/aws_elastic_transcoder_present: Fix typo in example for InterlacedMode

### LINKED ISSUE #6615: InterlaceMode can not work well in AWS Transcode
_This issue was originally opened by @shuwenli as hashicorp/terraform#19479. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I follow the document (https://www.terraform.io/docs/providers/aws/r/elastic_transcoder_preset.html)
to create aws trancode,  it keep showing me the re-cycle  destroy `InterlaceMode` and create `InterlacedMode`  when I run `terraform apply`
The root cause is we should use **InterlacedMode** instead of **InterlaceMode**

bflad:
Argument reference documentation below this has it correct. (see also: https://docs.aws.amazon.com/elastictranscoder/latest/developerguide/create-preset.html)

Fixes #6615 

* On the tin

### DIFF
--- a/website/docs/r/elastic_transcoder_preset.html.markdown
+++ b/website/docs/r/elastic_transcoder_preset.html.markdown
@@ -48,7 +48,7 @@ resource "aws_elastictranscoder_preset" "bar" {
     Profile                  = "main"
     Level                    = "2.2"
     MaxReferenceFrames       = 3
-    InterlaceMode            = "Progressive"
+    InterlacedMode           = "Progressive"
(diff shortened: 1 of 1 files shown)


bflad:
I see #6623 fixes this exactly the same and precedes mine. Closing to merge that PR!

PR #6687, opened 2018-12-03
Update AWS Lightsail Instance documentation

MitchellCash:
Generally clarifying or providing more detail around the AWS Lightsail instances.

- Update existing and add missing Blueprint IDs
- Add missing Bundle IDs
- More detail/clarification around Availability Zones and also how Availability Zones and Bundle IDs work together

### DIFF
--- a/website/docs/r/lightsail_instance.html.markdown
+++ b/website/docs/r/lightsail_instance.html.markdown
@@ -48,24 +48,30 @@ Lightsail currently supports the following Blueprint IDs:
 
 ### OS Only
 
-- `amazon_linux_2017_03_1_1`
-- `ubuntu_16_04_1`
+- `amazon_linux_2018_03_0_2`
+- `centos_7_1805_01`
 - `debian_8_7`
-- `freebsd_11`
+- `debian_9_5`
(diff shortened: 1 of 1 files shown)


bflad:
Hi, @MitchellCash, this is pretty fantastic! I'm wondering if instead of explicitly listing these out, which will slowly become out of date again in the Terraform documentation, are there AWS documentation links which contain these lists? I'd also be a proponent of keeping the explicit lists if they mentioned them as an example and included the links as well. What do you think?

MitchellCash:
@bflad I like that idea better! Let me see if AWS documents this somewhere.

MitchellCash:
@bflad I cannot find definitive lists of everything needed, most of the AWS documentation will point to AWS CLI/API commands to pull this info. I could put notes to use the AWS CLI if the user wants the up to date information?

On another note, because the info can be pulled from an API it means that it could potentially be automated through Terraform itself (user specifies `large` bundle and Terraform corrects it to `large_2_2` based on API response) or if there were a Terraform documentation bot of some sort it could detect changes in the API response and updated docs.

PR #6718, opened 2018-12-04
Create new resource for changed aws_iam_user_ssh_key public_key and encoding

### LINKED ISSUE #378: aws_iam_user_ssh_key incorrectly reports updating of the key value
_This issue was originally opened by @hamstah as hashicorp/terraform#10160. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version

terraform 0.7.7
(looking at the source in master and changelog this should affect 0.7.11 as well)

### Affected Resource(s)
`aws_iam_user_ssh_key`

First generate 2 ssh keys and save them as `first` and `second` in the current folder
```
ssh-keygen -t rsa -c first
ssh-keygen -t rsa -c second
```

```hcl
resource "aws_iam_user" "nico" {
    name = "nico"
}

resource "aws_iam_user_ssh_key" "nico" {
  username   = "${aws_iam_user.nico.name}"
  encoding   = "PEM"
  public_key = "${file("./first.pub")}"
}
```

Apply the change.
The key has now been uploaded and attached to the user.

Now change the file from `first.pub` to `second.pub`

Run terraform plan
It will show that the resource is going to be updated because the `public_key` field is different. You can see the `first -> second` in the plan, as expected.

Run terraform apply
It will show you the change has been applied successfully.

Running terraform plan shows that there are no more change to apply.

Fetch the key with the aws cli
```
aws iam list-public-ssh-key --user-name nico
aws iam get-public-ssh-key --user-name nico --ssh-public-key-id <id>
```
and see that the key is the `first` one, not the second.

This is because the AWS api doesn't allow an update to the `public_key` field, only the status, as documented here https://docs.aws.amazon.com/cli/latest/reference/iam/update-ssh-public-key.html

Looking at the terraform code https://github.com/hashicorp/terraform/blob/master/builtin/providers/aws/resource_aws_iam_user_ssh_key.go#L102 You can see that the resource is only updated if `status` has changed. The debug log shows 

```
2016/11/16 15:18:22 [DEBUG] Checking resource noop: aws_iam_user_ssh_key.nico2
2016/11/16 15:18:22 [DEBUG] No diff, not a noop
```

### Debug Output
I have logs but they have a lot of resources in them and scrubbing them will take time.
Given the code and the fact that the bug is easy to reproduce I don't think my logs are required, but let me know if you need them.

### Expected Behavior
Since the field can't be updated, terraform should flag the resource in the plan and force the user to delete and recreate it. Automatically deleting and recreating the key is another option but would result in a different id generated which might not be safe?

In any case it shouldn't look like the key was updated successfully.

### Actual Behavior
See above

### Steps to Reproduce
See above

AMeng:
Fixes #378

* Create a new `aws_iam_user_ssh_key` resource when either `public_key` or `encoding` changes.

```
$ make testacc TESTARGS='-run=TestAccAWSUserSSHKey'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSUserSSHKey -timeout 120m
=== RUN   TestAccAWSUserSSHKey_basic
=== PAUSE TestAccAWSUserSSHKey_basic
=== RUN   TestAccAWSUserSSHKey_pemEncoding
=== PAUSE TestAccAWSUserSSHKey_pemEncoding
=== CONT  TestAccAWSUserSSHKey_basic
=== CONT  TestAccAWSUserSSHKey_pemEncoding
--- PASS: TestAccAWSUserSSHKey_pemEncoding (13.64s)
--- PASS: TestAccAWSUserSSHKey_basic (15.12s)
PASS
```

According to the [AWS API docs](https://docs.aws.amazon.com/cli/latest/reference/iam/update-ssh-public-key.html), you cannot edit `public_key` or `encoding`. Therefore Terraform should create a new resource.

### DIFF
--- a/aws/resource_aws_iam_user_ssh_key.go
+++ b/aws/resource_aws_iam_user_ssh_key.go
@@ -36,11 +36,13 @@ func resourceAwsIamUserSshKey() *schema.Resource {
 			"public_key": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 			},
 
 			"encoding": {
 				Type:     schema.TypeString,
 				Required: true,
+				ForceNew: true,
 				ValidateFunc: validation.StringInSlice([]string{
 					iam.EncodingTypeSsh,
 					iam.EncodingTypePem,


bflad:
This has been released in [version 1.51.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1510-december-05-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #6725, opened 2018-12-05
add warning about currently immutable maintenance window

kiwivogel:

* Add a comment to docs warning user that maintenance window currently cannot be changed.

```
None, documentation only
```


kiwivogel:
Figured a warning might be nice as most other managed services DO allow updates to the maintenance window.

kiwivogel:
I did get in touch with AWS support and they'll pitch it as a feature request internally. Will update provider when/if they decide to fix this omission.

PR #6734, opened 2018-12-05
deps: aws/aws-sdk-go@v1.16.0

bflad:
Updated via: `govendor fetch github.com/aws/aws-sdk-go/...@v1.16.0`

NOTE: Marking as `breaking-change` until we can verify the `StringValue()` changes.


bflad:
Looking at the upstream changes, they seem like they should be fine, but I'm going to kick off a full acceptance test run and look for new test failures.

PR #6770, opened 2018-12-07
Do not throw error when lambda permission is deleted after function

### LINKED ISSUE #6769: Terraform throws error when deleting lambda permission after lambda function has been deleted
Terraform throws an error when there is a race condition between deleting lambda function and deleting lambda permission and it tries to delete a lambda permission after the lambda function has been already deleted.

Puneeth-n:
Fixes #6769

### DIFF
--- a/aws/resource_aws_lambda_permission.go
+++ b/aws/resource_aws_lambda_permission.go
@@ -301,6 +301,13 @@ func resourceAwsLambdaPermissionDelete(d *schema.ResourceData, meta interface{})
 	log.Printf("[DEBUG] Removing Lambda permission: %s", input)
 	_, err := conn.RemovePermission(&input)
 	if err != nil {
+		// Missing whole policy or Lambda function (API error)
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() == "ResourceNotFoundException" {
+				log.Printf("[WARN] No Lambda Permission Policy found: %v", input)
+				return nil
+			}
+		}
 		return err
 	}
 


aws/resource_aws_lambda_permission.go
```diff
@@ -301,6 +301,13 @@ func resourceAwsLambdaPermissionDelete(d *schema.ResourceData, meta interface{})
 	log.Printf("[DEBUG] Removing Lambda permission: %s", input)
 	_, err := conn.RemovePermission(&input)
 	if err != nil {
+		// Missing whole policy or Lambda function (API error)
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() == "ResourceNotFoundException" {
```
bflad: Nit: We prefer to use our helper function to determine these and use the SDK provided constants:

```go
if isAWSErr(err, lambda.ErrCodeResourceNotFoundException, "") {
```

aws/resource_aws_lambda_permission.go
```diff
@@ -301,6 +301,13 @@ func resourceAwsLambdaPermissionDelete(d *schema.ResourceData, meta interface{})
...
+				log.Printf("[WARN] No Lambda Permission Policy found: %v", input)
```
bflad: Nit: We prefer to skip logging that an API object is missing when Terraform is already trying to delete it.

Puneeth-n: @bflad Thanks for the heads up. Will remember it :)

bflad:
This has been released in [version 1.52.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1520-december-13-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #6771, opened 2018-12-07
deps: aws/aws-sdk-go@v1.16.2

Puneeth-n:
`govendor fetch github.com/aws/aws-sdk-go/...@v1.16.2`


CyrusNajmabadi:
Hey @bflad @puneet-h.  This is unfortunately causing a problem for us (details at https://github.com/pulumi/pulumi-aws/issues/396).

The core crux of the problem is that this version of the aws-sdk doesn't play nicely with another change Terraform took recently.  

The two specific issues are:

1. aws-sdk now does stricter argument checking when making certain api calls.  this change came in https://github.com/aws/aws-sdk-go/pull/2323, which you have now picked up.  As an example: apigateway.GetStage will now throw if you pass an empty "stagename" to it.
2. Terraform aws recently changed how deployment deletion works in https://github.com/terraform-providers/terraform-provider-aws/commit/8973a045c4185a107e16b9a240bba039585d5cf5.  

In the terraform PR this code as added:

```go
	stage, err := conn.GetStage(&apigateway.GetStageInput{
		StageName: aws.String(d.Get("stage_name").(string)),
		RestApiId: aws.String(d.Get("rest_api_id").(string)),
	})

	if err != nil && !isAWSErr(err, apigateway.ErrCodeNotFoundException, "") {
		return fmt.Errorf("error getting referenced stage: %s", err)
	}
```

This now ends up throwing if you have a Deployment with an empty stage_name (which is legal, and which we use ourselces).  Can this change be rolled back?  Or, alternatively if you want this version of the awssdk, can the appropriate codepaths be updated to not throw here when dealing with legal situations?  

This could be done in the above case by also adding ErrParamMinLen as the type of error that is ok to get back from aws now that they're stricter about the data passed to them.

Thanks!

PR #6780, opened 2018-12-10
Vpc Peering auto accept

jsenon:

* Be able to auto-accept connexion on same account form to different region

Example of tf for this feature:
```
provider "aws" {
  alias   = "that"
  version = ">=2.0.4"
  region  = "eu-central-1"
}

resource "aws_vpc" "that_vpc" {
  cidr_block = "10.1.0.0/16"
  provider = "aws.that"
}

provider "aws" {
  alias   = "this"
  version = ">=2.0.4"
  region  = "eu-west-3"
}

resource "aws_vpc" "this_vpc" {
  cidr_block = "10.0.0.0/16"
  provider = "aws.this"
}

resource "aws_vpc_peering_connection" "test" {
  provider = "aws.that"
  vpc_id        = "${aws_vpc.that_vpc.id}"
  peer_vpc_id   = "${aws_vpc.this_vpc.id}"
  peer_region   = "eu-west-3"
  auto_accept   = "true"
}
```

### DIFF
--- a/aws/resource_aws_vpc_peering_connection.go
+++ b/aws/resource_aws_vpc_peering_connection.go
@@ -85,7 +85,6 @@ func resourceAwsVPCPeeringCreate(d *schema.ResourceData, meta interface{}) error
 		if _, ok := d.GetOk("auto_accept"); ok {
 			log.Printf("[DEBUG] VPC Peering Auto Accept")
 			createOpts.PeerRegion = aws.String(v.(string))
-			// return fmt.Errorf("peer_region cannot be set whilst auto_accept is true when creating a vpc peering connection")
 			resp, err := conn.CreateVpcPeeringConnection(createOpts)
 			if err != nil {
 				return fmt.Errorf("Error creating VPC Peering Connection: %s", err)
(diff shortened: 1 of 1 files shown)


bflad:
Hi @jsenon 👋 Thanks for submitting this. Was there a particular reason you cc'd me above? While I'm currently an active maintainer for the project, I'm not the only one. 😄 

We typically disallow Terraform AWS provider resources from initializing secondary AWS clients, especially in other regions, since the behavior goes against the explicit provider configuration we currently use (`AWS*_REGION` environment variables, `region = ""` configuration, AWS profile, etc.) and must account for all client authentication methods via the same method when initializing or could generate unexpected credential errors without our wrapping credential error message.

Instead, depending on the error messaging and workflow, we might consider just not setting the problematic parameters in the request if they are errantly present in the Terraform configuration and just logging the behavior. I'm not sure if that is a viable option off the top of my head here though as my focus is elsewhere at the moment. This seems like it should be discussed in an issue with other community member/maintainer input before trying to work out an implementation.

lorengordon:
Hi @jsenon, have you tried this using [aws_vpc_peering_connection_accepter](aws_vpc_peering_connection_accepter)? I suspect you ought to be able to do something like this...

```
provider "aws" {
  alias   = "that"
  version = ">=2.0.4"
  region  = "eu-central-1"
}

resource "aws_vpc" "that_vpc" {
  cidr_block = "10.1.0.0/16"
  provider = "aws.that"
}

provider "aws" {
  alias   = "this"
  version = ">=2.0.4"
  region  = "eu-west-3"
}

resource "aws_vpc" "this_vpc" {
  cidr_block = "10.0.0.0/16"
  provider = "aws.this"
}

resource "aws_vpc_peering_connection" "test" {
  provider = "aws.that"
  vpc_id        = "${aws_vpc.that_vpc.id}"
  peer_vpc_id   = "${aws_vpc.this_vpc.id}"
  peer_region   = "eu-west-3"
}

resource "aws_vpc_peering_connection_accepter" "test" {
  provider = "aws.this"
  vpc_peering_connection_id = "${aws_vpc_peering_connection.test.id}"
  auto_accept               = true
}
```

jen20:
Hi @jsenon! As @bflad points out, operating in a region other than that configured for the provider (or in the global context for region-less services such as IAM) would be a pretty bad idea - not least because the code in this PR doesn't take into account things like role assumption workflows which may be necessary.

I think the way @lorengordon outlined above with explicit configuration of two providers is the correct way to go about this - is there a reason that won't work for your situation?

bflad:
Closing this PR because as it is not likely to be accepted in its current form and hasn't seen a response in the last three weeks. The above mentioned configuration from @lorengordon should allow accepting VPC peering cross region using the typical method of configuring two separate AWS providers. If you would like to discuss this further, please don't hesitate to reach back out. 👍

PR #6829, opened 2018-12-12
RDS:   Added two CW log export values to support postgresql

tylersmith34:
This change allows for CloudWatch log exports to be enabled for PostgreSQL databases v9.6.6 and newer.  Older versions use the existing log exports but newer versions have their own list.  This PR adds those new values to the list and to the documentation.

* Update the list in the `aws_db_instance` code
* Update the list in the documentation

Note, it's in two commits because my company blocks access to GitHub so I had to use the GUI to make the changes.

```

...
```
I didn't run the acceptance tests, I don't have an account where I can incur charges and I only added two items to a list.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -408,6 +408,8 @@ func resourceAwsDbInstance() *schema.Resource {
 						"listener",
 						"slowquery",
 						"trace",
+						"postgresql",
+						"upgrade",
 					}, false),
 				},
 			},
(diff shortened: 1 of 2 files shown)


tylersmith34:
Supporting documentation:
https://aws.amazon.com/about-aws/whats-new/2018/12/amazon-rds-supports-postgresql-logfiles-publish-to-amazon-cloudwatch-logs/
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_LogAccess.Concepts.PostgreSQL.html#USER_LogAccess.PostgreSQL.PublishtoCloudWatchLogs

bflad:
I already added it.

bflad:
This has been released in [version 1.52.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1520-december-13-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

guersam:
I guess these should be allowed in [`aws_rds_cluster`](https://github.com/terraform-providers/terraform-provider-aws/blob/v2.11.0/aws/resource_aws_rds_cluster.go#L383-L396) resource too. WDYT?

ckuehl:
> Nevermind, `aurora-postgresql` engine does not support exporting to CloudWatch Logs yet.

It appears that it does now, so similar changes need to be made to `aws_rds_cluster`.

PR #6835, opened 2018-12-12
r/aws_licensemanager: Add aws_licensemanager_license_configuration resource

### LINKED ISSUE #6654: Feature Request: License Manager
### Community Note

* `aws_licensemanager_license_configuration` (New resource)
* `aws_licensemanager_settings` (New resource)
* `aws_launch_template` (Affected resource)

```hcl
resource "aws_licensemanager_license_configuration" "example" {
  name                     = "Example"
  description              = "Example"
  license_count            = 10
  license_count_hard_limit = true
  license_counting_type    = "Socket"

  license_rules = [
    "#minimumSockets=3"
  ]

  tags {
    foo = "barr"
  }
}
```

### References

* https://docs.aws.amazon.com/license-manager/latest/APIReference/API_CreateLicenseConfiguration.html
* https://docs.aws.amazon.com/license-manager/latest/APIReference/API_GetLicenseConfiguration.html
* https://docs.aws.amazon.com/license-manager/latest/APIReference/API_UpdateLicenseConfiguration.html
* https://docs.aws.amazon.com/license-manager/latest/APIReference/API_DeleteLicenseConfiguration.html

gazoakley:
Reference: #6654

* **New Resource:** `aws_licensemanager_license_configuration`

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSLicenseManagerLicenseConfiguration'
TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccAWSLicenseManagerLicenseConfiguration -timeout 120m
=== RUN   TestAccAWSLicenseManagerLicenseConfiguration_basic
=== PAUSE TestAccAWSLicenseManagerLicenseConfiguration_basic
=== CONT  TestAccAWSLicenseManagerLicenseConfiguration_basic
--- PASS: TestAccAWSLicenseManagerLicenseConfiguration_basic (10.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	10.172s
```

### DIFF
--- a/aws/resource_aws_licensemanager_license_configuration.go
+++ b/aws/resource_aws_licensemanager_license_configuration.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/licensemanager"
@@ -49,7 +50,10 @@ func resourceAwsLicenseManagerLicenseConfiguration() *schema.Resource {
 				Type:     schema.TypeList,
 				Optional: true,
 				ForceNew: true,
-				Elem:     &schema.Schema{Type: schema.TypeString},
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringMatch(regexp.MustCompile("^#([^=]+)=(.+)$"), "Expected format is #RuleType=RuleValue"),
+				},
 			},
 			"name": {
 				Type:     schema.TypeString,
@@ -118,7 +122,9 @@ func resourceAwsLicenseManagerLicenseConfigurationRead(d *schema.ResourceData, m
 	d.Set("license_count", resp.LicenseCount)
 	d.Set("license_count_hard_limit", resp.LicenseCountHardLimit)
 	d.Set("license_counting_type", resp.LicenseCountingType)
-	d.Set("license_rules", flattenStringList(resp.LicenseRules))
+	if err := d.Set("license_rules", flattenStringList(resp.LicenseRules)); err != nil {
+		return fmt.Errorf("error setting license_rules: %s", err)
+	}
 	d.Set("name", resp.Name)
 
 	if err := d.Set("tags", tagsToMapLicenseManager(resp.Tags)); err != nil {
@@ -136,9 +142,8 @@ func resourceAwsLicenseManagerLicenseConfigurationUpdate(d *schema.ResourceData,
 	if d.HasChange("tags") {
 		if err := setTagsLicenseManager(conn, d); err != nil {
 			return err
-		} else {
-			d.SetPartial("tags")
 		}
+		d.SetPartial("tags")
 	}
 
 	d.Partial(false)
@@ -179,7 +184,6 @@ func resourceAwsLicenseManagerLicenseConfigurationDelete(d *schema.ResourceData,
 	_, err := conn.DeleteLicenseConfiguration(opts)
 	if err != nil {
 		if isAWSErr(err, licensemanager.ErrCodeInvalidParameterValueException, "") {
-			log.Printf("[WARN] License Manager license configuration (%s) not found", d.Id())
 			return nil
 		}
 		return fmt.Errorf("Error deleting License Manager license configuration: %s", err)
--- a/aws/resource_aws_licensemanager_license_configuration.go
+++ b/aws/resource_aws_licensemanager_license_configuration.go
@@ -1,0 +1,189 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/licensemanager"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsLicenseManagerLicenseConfiguration() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsLicenseManagerLicenseConfigurationCreate,
+		Read:   resourceAwsLicenseManagerLicenseConfigurationRead,
+		Update: resourceAwsLicenseManagerLicenseConfigurationUpdate,
+		Delete: resourceAwsLicenseManagerLicenseConfigurationDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"license_count": {
+				Type:     schema.TypeInt,
+				Optional: true,
+			},
+			"license_count_hard_limit": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+			"license_counting_type": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					licensemanager.LicenseCountingTypeVCpu,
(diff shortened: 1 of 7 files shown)


aws/resource_aws_licensemanager_license_configuration.go
```diff
@@ -0,0 +1,189 @@
...
+
+func resourceAwsLicenseManagerLicenseConfigurationUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).licensemanagerconn
+
+	d.Partial(true)
+
+	if d.HasChange("tags") {
+		if err := setTagsLicenseManager(conn, d); err != nil {
+			return err
+		} else {
```
bflad: Nit: Since the `if` calls `return` the `else` and nesting are extraneous here 😄

aws/resource_aws_licensemanager_license_configuration.go
```diff
@@ -0,0 +1,189 @@
...
+			return nil
+		}
+		return fmt.Errorf("Error reading License Manager license configuration: %s", err)
+	}
+
+	d.Set("description", resp.Description)
+	d.Set("license_count", resp.LicenseCount)
+	d.Set("license_count_hard_limit", resp.LicenseCountHardLimit)
+	d.Set("license_counting_type", resp.LicenseCountingType)
+	d.Set("license_rules", flattenStringList(resp.LicenseRules))
```
bflad: Nit: This looks okay, but in general if its a `TypeList`, `TypeSet`, or `TypeMap` we should be adding the error checking. 👍

aws/resource_aws_licensemanager_license_configuration.go
```diff
@@ -0,0 +1,189 @@
...
+	conn := meta.(*AWSClient).licensemanagerconn
+
+	opts := &licensemanager.DeleteLicenseConfigurationInput{
+		LicenseConfigurationArn: aws.String(d.Id()),
+	}
+
+	_, err := conn.DeleteLicenseConfiguration(opts)
+	if err != nil {
+		if isAWSErr(err, licensemanager.ErrCodeInvalidParameterValueException, "") {
+			log.Printf("[WARN] License Manager license configuration (%s) not found", d.Id())
```
bflad: Nit: Logging a missing resource during deletion is extraneous. 😄

aws/resource_aws_licensemanager_license_configuration_test.go
```diff
@@ -0,0 +1,161 @@
...
+func testSweepLicenseManagerLicenseConfigurations(region string) error {
+	client, err := sharedClientForRegion(region)
+	if err != nil {
+		return fmt.Errorf("error getting client: %s", err)
+	}
+	conn := client.(*AWSClient).licensemanagerconn
+
+	resp, err := conn.ListLicenseConfigurations(&licensemanager.ListLicenseConfigurationsInput{})
+
+	if err != nil {
```
bflad: Can you please add this check in the test sweeper? Thanks!

```go
if testSweepSkipSweepError(err) {
	log.Printf("[WARN] Skipping License Manager License Configuration sweep for %s: %s", region, err)
	return nil
}
```

website/docs/r/licensemanager_license_configuration.markdown
```diff
@@ -0,0 +1,57 @@
+---
+layout: "aws"
+page_title: "AWS: aws_licensemanager_license_configuration"
+sidebar_current: "docs-aws-resource-licensemanager-license-configuration"
```
bflad: Missing sidebar link in `website/aws.erb`

aws/resource_aws_licensemanager_license_configuration.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/aws/aws-sdk-go/service/licensemanager"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsLicenseManagerLicenseConfiguration() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsLicenseManagerLicenseConfigurationCreate,
+		Read:   resourceAwsLicenseManagerLicenseConfigurationRead,
+		Update: resourceAwsLicenseManagerLicenseConfigurationUpdate,
```
bflad: Update function is not currently exercised in the acceptance testing

gazoakley:
Thanks @bflad - I'll address the review comments shortly. From a design perspective I'm thinking `license_rules` could be turned into a map since it can only contain strings in the format of `#RuleType=RuleValue` (e.g. `#minimumSockets=3`). That feels more user friendly to me, but I think it breaks away from the principal of matching the API. Any preference? At the very least I want to add validation for the strings and document the possible RuleTypes.

There also seems to be a bug in the License Manager API that prevents removing `license_count` - the API doesn't allow resetting the license count to unlimited. The same problem is visible in the AWS console too. I could:

- Try and get AWS to fix the API and console before adding the resource
- Set `ForceNew` on that attribute, but that seems overkill when it otherwise works
- Throw an error when applying a change that removes `license_count` asking the user to taint the resource
- Could we have a `ForceNewFunc` to allow conditionally re-creating resources at plan time?

bflad:
> From a design perspective I'm thinking license_rules could be turned into a map since it can only contain strings in the format of #RuleType=RuleValue (e.g. #minimumSockets=3). That feels more user friendly to me, but I think it breaks away from the principal of matching the API. Any preference?

I would only do this if we can get confirmation from the AWS License Manager service team that the format will always follow a single `#KEY=VALUE`, map structure for the string field. One would hope they would use a map in the API if it would always be a map like that. I could see the potential for other logic additions such as `OR` or other operators. I have also have concerns that trying to inject validation could leave us either with a maintenance burden trying to keep up or needing to remove it if arbitrary rules are introduced. I these are all questions of stability that the service team could also confirm.

> There also seems to be a bug in the License Manager API that prevents removing license_count - the API doesn't allow resetting the license count to unlimited. The same problem is visible in the AWS console too.

Does the API throw a decent error? Is this planned on getting fixed? If it will be fixed, at most we should create a better error message, otherwise we should just pass the existing error through and potentially document the upstream issue. This doesn't seem like it should hold up releasing the resource or block other working update behavior with `ForceNew`.

gazoakley:
> I would only do this if we can get confirmation from the AWS License Manager service team that the format will always follow a single `#KEY=VALUE`, map structure for the string field. One would hope they would use a map in the API if it would always be a map like that. I could see the potential for other logic additions such as `OR` or other operators. I have also have concerns that trying to inject validation could leave us either with a maintenance burden trying to keep up or needing to remove it if arbitrary rules are introduced. I these are all questions of stability that the service team could also confirm.

The API throws an error message if you don't specify it in that format: `Expected format is #RuleType=RuleValue`. I get that if it can't be guaranteed that they won't change the format then using a map could cause problems in the future. I've added a ValidateFunc for now to check the format (which would be easy to strip out if they start to allow logic operators) - maybe that feels like the best compromise?

> Does the API throw a decent error? Is this planned on getting fixed? If it will be fixed, at most we should create a better error message, otherwise we should just pass the existing error through and potentially document the upstream issue. This doesn't seem like it should hold up releasing the resource or block other working update behavior with `ForceNew`.

It just silently returns success if you don't specify a value for `opts.LicenseCount` and then doesn't make a change. The console specifies `"LicenseCount": null` (checking Chrome DevTools), but that doesn't reset the value either. I'm not sure aws-sdk-go allows me to pass that value as null because I can't see the request body for JSON requests in the debug logs (although I can see JSON responses). For now I could document that resetting this value requires tainting the resource?

bflad:
This has been released in [version 1.53.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1530-december-20-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #6880, opened 2018-12-17
Add complex table definition for kinesis parquet integration

mwarkentin:
It took me quite a bit of trial and error to figure out the right format to configure a more complex example like this (a glue table which is used as the schema for kinesis firehose -> s3 conversion and athena queries).

Felt like it would probably save others trying to do something similar a bunch of time so here's a PR to add it to the docs.

Let me know if you have any changes or suggestions you'd like.


mwarkentin:
Any thoughts on this doc update?

dedamico:
Thanks @mwarkentin!!!

PR #6909, opened 2018-12-18
Fix gosimple api issues

vnguyen7:

 - Fix gosimple api issues

### DIFF
--- a/aws/resource_aws_api_gateway_method_test.go
+++ b/aws/resource_aws_api_gateway_method_test.go
@@ -207,7 +207,7 @@ func testAccCheckAWSAPIGatewayMethodAttributes(conf *apigateway.Method) resource
 		if val, ok := conf.RequestParameters["method.request.header.Content-Type"]; !ok {
 			return fmt.Errorf("missing Content-Type RequestParameters")
 		} else {
-			if !*val {
+			if *val {
 				return fmt.Errorf("wrong Content-Type RequestParameters value")
 			}
 		}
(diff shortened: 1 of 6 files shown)


aws/resource_aws_api_gateway_method_test.go
```diff
@@ -207,14 +207,14 @@ func testAccCheckAWSAPIGatewayMethodAttributes(conf *apigateway.Method) resource
 		if val, ok := conf.RequestParameters["method.request.header.Content-Type"]; !ok {
 			return fmt.Errorf("missing Content-Type RequestParameters")
 		} else {
-			if *val != false {
+			if !*val {
```
bflad: It looks like the logic got flipped here.

```
--- FAIL: TestAccAWSAPIGatewayMethod_basic (41.36s)
    testing.go:538: Step 0 error: Check failed: Check 2/5 error: wrong Content-Type RequestParameters value
--- FAIL: TestAccAWSAPIGatewayMethod_customrequestvalidator (34.10s)
    testing.go:538: Step 0 error: Check failed: Check 2/6 error: wrong Content-Type RequestParameters value
```

```suggestion
			if *val {
```

aws/resource_aws_api_gateway_method_test.go
```diff
@@ -234,7 +234,7 @@ func testAccCheckAWSAPIGatewayMethodAttributesUpdate(conf *apigateway.Method) re
 		if val, ok := conf.RequestParameters["method.request.querystring.page"]; !ok {
 			return fmt.Errorf("missing updated page RequestParameters")
 		} else {
-			if *val != false {
+			if !*val {
```
bflad: Same here.

```suggestion
			if *val {
```

PR #6885, opened 2018-12-18
resource/aws_appautoscaling_policy: Fix gosimple linting issues

bflad:
Reference: #6343 

Previously:

```
aws/resource_aws_appautoscaling_policy.go:458:66:warning: should use make([]*applicationautoscaling.MetricDimension, s.Len()) instead (S1019) (gosimple)
aws/resource_aws_appautoscaling_policy.go:638:36:warning: should use make(map[string]interface{}) instead (S1019) (gosimple)
aws/resource_aws_appautoscaling_policy.go:660:29:warning: should use make([]interface{}, len(adjs)) instead (S1019) (gosimple)
aws/resource_aws_appautoscaling_policy.go:663:37:warning: should use make(map[string]interface{}) instead (S1019) (gosimple)
aws/resource_aws_appautoscaling_policy.go:685:36:warning: should use make(map[string]interface{}) instead (S1019) (gosimple)
aws/resource_aws_appautoscaling_policy.go:729:27:warning: should use make([]interface{}, len(ds)) instead (S1019) (gosimple)
```

Output from acceptance testing:

```
--- PASS: TestAccAWSAppautoScalingPolicy_basic (37.33s)
--- PASS: TestAccAWSAppautoScalingPolicy_nestedSchema (37.39s)
--- PASS: TestAccAWSAppautoScalingPolicy_scaleOutAndIn (38.26s)
--- PASS: TestAccAWSAppautoScalingPolicy_spotFleetRequest (70.42s)
--- PASS: TestAccAWSAppautoScalingPolicy_dynamoDb (112.74s)
--- PASS: TestAccAWSAppautoScalingPolicy_multiplePoliciesSameResource (113.80s)
--- PASS: TestAccAWSAppautoScalingPolicy_multiplePoliciesSameName (114.53s)
```

PR #6915, opened 2018-12-19
Fix gosimple app issues

vnguyen7:

Fix gosimple app issues


bflad:
I'll close #6885 since this handles the same failures.

Issue #6674, opened 2018-12-01
Feature Request: Security Hub


gazoakley:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AWS has announced Security Hub: https://aws.amazon.com/security-hub/

### New or Affected Resource(s)

- [X] `aws_securityhub_account`
- [X] `aws_securityhub_action_target`
- [X] `aws_securityhub_member`
- [X] `aws_securityhub_product_subscription`
- [X] `aws_securityhub_standards_subscription`
- [X] `aws_securityhub_organization_admin_account`
- [x] `aws_securityhub_invite_accepter`
- [ ] `aws_securityhub_insight`

### Potential Terraform Configuration

```hcl
# Used to enable AWS Security Hub
resource "aws_securityhub_account" "example" {}

# Subscribe to the CIS AWS Foundations Benchmark
resource "aws_securityhub_standards_subscription" "example" {
  depends_on    = ["aws_securityhub_account.example"]
  standards_arn = "arn:aws:securityhub:::ruleset/cis-aws-foundations-benchmark/v/1.2.0"
}

# Subscribe to a third party provider
data "aws_region" "current" {}

resource "aws_securityhub_product_subscription" "example" {
  depends_on  = ["aws_securityhub_account.example"]
  product_arn = "arn:aws:securityhub:${data.aws_region.current.name}:679703615338:product/armordefense/armoranywhere"
}

# Add a member AWS account
resource "aws_securityhub_member" "example" {
  depends_on = ["aws_securityhub_account.example"]
  account_id = "123456789012"
  email      = "example@example.com"
  invite       = true
}

resource "aws_securityhub_account" "invitee" {
  provider = "aws.invitee"
}

resource "aws_securityhub_invite_accepter" "invitee" {
  provider   = "aws.invitee"
  depends_on = ["aws_securityhub_account.invitee"]
  master_id  = "${aws_securityhub_invitation.example.master_id}"
}

# Create an insight (group of findings)
resource "aws_securityhub_insight" "example" {
  depends_on         = ["aws_securityhub_account.example"]
  name               = "Example"
[...]

PR #6921, opened 2018-12-19
r/aws_securityhub: Add aws_securityhub_product_subscription resource

gazoakley:
Partly addresses #6674 

* **New Resource:** `aws_securityhub_product_subscription`

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSSecurityHub/ProductSubscription'
TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccAWSSecurityHub/ProductSubscription -timeout 120m
=== RUN   TestAccAWSSecurityHub
=== RUN   TestAccAWSSecurityHub/ProductSubscription
=== RUN   TestAccAWSSecurityHub/ProductSubscription/basic
--- PASS: TestAccAWSSecurityHub (46.97s)
    --- PASS: TestAccAWSSecurityHub/ProductSubscription (46.97s)
        --- PASS: TestAccAWSSecurityHub/ProductSubscription/basic (46.97s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	47.021s
```

### DIFF
--- a/aws/resource_aws_securityhub_product_subscription.go
+++ b/aws/resource_aws_securityhub_product_subscription.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/securityhub"
@@ -25,23 +26,29 @@ func resourceAwsSecurityHubProductSubscription() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validateArn,
 			},
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
 
 func resourceAwsSecurityHubProductSubscriptionCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).securityhubconn
-	log.Printf("[DEBUG] Enabling Security Hub product subscription for product %s", d.Get("product_arn"))
+	productArn := d.Get("product_arn").(string)
+
+	log.Printf("[DEBUG] Enabling Security Hub product subscription for product %s", productArn)
 
 	resp, err := conn.EnableImportFindingsForProduct(&securityhub.EnableImportFindingsForProductInput{
-		ProductArn: aws.String(d.Get("product_arn").(string)),
+		ProductArn: aws.String(productArn),
 	})
 
 	if err != nil {
-		return fmt.Errorf("Error enabling Security Hub product subscription for product %s: %s", d.Get("product_arn"), err)
+		return fmt.Errorf("Error enabling Security Hub product subscription for product %s: %s", productArn, err)
 	}
 
-	d.SetId(*resp.ProductSubscriptionArn)
+	d.SetId(fmt.Sprintf("%s,%s", productArn, *resp.ProductSubscriptionArn))
 
 	return resourceAwsSecurityHubProductSubscriptionRead(d, meta)
 }
@@ -49,9 +56,15 @@ func resourceAwsSecurityHubProductSubscriptionCreate(d *schema.ResourceData, met
 func resourceAwsSecurityHubProductSubscriptionRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).securityhubconn
 
+	productArn, productSubscriptionArn, err := resourceAwsSecurityHubProductSubscriptionParseId(d.Id())
+
+	if err != nil {
+		return err
+	}
+
 	log.Printf("[DEBUG] Reading Security Hub product subscriptions to find %s", d.Id())
 
-	exists, err := resourceAwsSecurityHubProductSubscriptionCheckExists(conn, d.Id())
+	exists, err := resourceAwsSecurityHubProductSubscriptionCheckExists(conn, productSubscriptionArn)
 
 	if err != nil {
 		return fmt.Errorf("Error reading Security Hub product subscriptions to find %s: %s", d.Id(), err)
@@ -61,6 +74,9 @@ func resourceAwsSecurityHubProductSubscriptionRead(d *schema.ResourceData, meta 
 		log.Printf("[WARN] Security Hub product subscriptions (%s) not found, removing from state", d.Id())
 		d.SetId("")
 	}
+
+	d.Set("product_arn", productArn)
+	d.Set("arn", productSubscriptionArn)
 
 	return nil
 }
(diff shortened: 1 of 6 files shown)


aws/resource_aws_securityhub_product_subscription.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/aws/aws-sdk-go/service/securityhub"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsSecurityHubProductSubscription() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsSecurityHubProductSubscriptionCreate,
+		Read:   resourceAwsSecurityHubProductSubscriptionRead,
+		Delete: resourceAwsSecurityHubProductSubscriptionDelete,
+		Importer: &schema.ResourceImporter{
```
bflad: This is currently missing acceptance testing and documentation. 😅 I think the current comment in the acceptance testing might be outdated since it looks like the `Read` function is implemented just fine below.

aws/resource_aws_securityhub_product_subscription.go
```diff
@@ -0,0 +1,85 @@
...
+	conn := meta.(*AWSClient).securityhubconn
+
+	log.Printf("[DEBUG] Reading Security Hub product subscriptions to find %s", d.Id())
+	resp, err := conn.ListEnabledProductsForImport(&securityhub.ListEnabledProductsForImportInput{})
+
+	if err != nil {
+		return fmt.Errorf("Error reading Security Hub product subscriptions to find %s: %s", d.Id(), err)
+	}
+
+	productSubscriptions := make([]interface{}, len(resp.ProductSubscriptions))
```
bflad: Nit: We can simplify the logic here, rather than allocating a second slice and adding the (albeit minor) computation work beyond every element past the one we want:

```go
productSubscriptionARN := ""
for _, productSubscription := range resp.ProductSubscriptions {
  if aws.StringValue(productSubscription) == d.Id() {
    productSubscriptionARN = aws.StringValue(productSubscription)
    break
  }
}

if productSubscriptionARN == "" {
  log.Printf("[WARN] Security Hub product subscriptions (%s) not found, removing from state", d.Id())
  d.SetId("")
  return nil
}

d.Set("product_arn", productSubscriptionARN)

return nil
```

It may also be worth moving this logic into its own function since it needs to get repeated in a few places like the acceptance testing.

aws/resource_aws_securityhub_product_subscription.go
```diff
@@ -0,0 +1,85 @@
...
+	d.SetId(*resp.ProductSubscriptionArn)
+
+	return resourceAwsSecurityHubProductSubscriptionRead(d, meta)
+}
+
+func resourceAwsSecurityHubProductSubscriptionRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).securityhubconn
+
+	log.Printf("[DEBUG] Reading Security Hub product subscriptions to find %s", d.Id())
+	resp, err := conn.ListEnabledProductsForImport(&securityhub.ListEnabledProductsForImportInput{})
```
bflad: The `ListEnabledProductsForImport` API call has a paginated counterpart in the AWS Go SDK, [`ListEnabledProductsForImportPages`](https://docs.aws.amazon.com/sdk-for-go/api/service/securityhub/#SecurityHub.ListEnabledProductsForImportPages), which we should use to prevent search issues in large environments, e.g.

```go
input := &securityhub.ListEnabledProductsForImportInput{}
productSubscriptionARN := ""

err := conn.ListEnabledProductsForImportPages(input, func(page *securityhub.ListEnabledProductsForImportOutput, lastPage bool) bool {
  for _, productSubscription := range resp.ProductSubscriptions {
    if aws.StringValue(productSubscription) == d.Id() {
      productSubscriptionARN = aws.StringValue(productSubscription)
      return false
    }
  }
  return !lastPage
})

if err != nil {
  return fmt.Errorf("Error reading Security Hub product subscriptions to find %s: %s", d.Id(), err)
}
```

aws/resource_aws_securityhub_product_subscription_test.go
```diff
@@ -0,0 +1,104 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSecurityHubAccountDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSSecurityHubProductSubscriptionConfig_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSSecurityHubProductSubscriptionExists("aws_securityhub_product_subscription.example"),
+				),
+			},
+			// Import is not supported, since the API to lookup product_arn from a product
```
bflad:
```go
{
  ResourceName: "aws_securityhub_product_subscription.example",
  ImportState: true,
  ImportStateVerify: true,
}
```

website/docs/r/securityhub_product_subscription.markdown
```diff
@@ -0,0 +1,70 @@
...
+* `arn:aws:securityhub:${var.region}:422820575223:product/tenable/tenable-io`
+* `arn:aws:securityhub:${var.region}:679593333241:product/trend-micro/deep-security`
+* `arn:aws:securityhub:${var.region}:453761072151:product/turbot/turbot`
+* `arn:aws:securityhub:${var.region}:496947949261:product/twistlock/twistlock-enterprise`
+
+## Attributes Reference
+
+The following attributes are exported in addition to the arguments listed above:
+
+* `id` - The ARN of a resource that represents your subscription to the product that generates the findings that you want to import into Security Hub.
```
bflad:
``````markdown
* `id` - The ARN of a resource that represents your subscription to the product that generates the findings that you want to import into Security Hub.

## Import

Security Hub Product Subscriptions can be imported via the product ARN, e.g.

```sh
$ terraform import aws_securityhub_product_subscription.example arn:aws:securityhub:us-east-1::product/aws/guardduty
```
``````

File: aws/resource_aws_securityhub_product_subscription.go (hunk 1 above)
Comment (gazoakley) [reply aws/resource_aws_securityhub_product_subscription.go]: I left this in accidentally - but I might have a misunderstanding about the import process. Do resources need to be able to read/populate all attributes solely from the resource ID? The API doesn't have a way to read back the `product_arn` given a product subscription ARN (you can only check a subscription exists). It looks like I might be able to make an import test that works if I specify `ImportStateVerifyIgnore: ["product_arn"]`

File: aws/resource_aws_securityhub_product_subscription_test.go (hunk 4 above)
Comment (gazoakley) [reply aws/resource_aws_securityhub_product_subscription_test.go]: I've added an import test step, but I've need to use `ImportStateVerifyIgnore`:

```go
{
  ResourceName:            "aws_securityhub_product_subscription.example",
  ImportState:             true,
  ImportStateVerify:       true,
  ImportStateVerifyIgnore: []string{"product_arn"},
},
```

website/docs/r/securityhub_product_subscription.markdown
gazoakley: The ID is the subscription ARN since that's the only thing I can test still exists - I've added an Import section describing it.

aws/resource_aws_securityhub_product_subscription.go
gazoakley: I thought about option 3 previously and then ran away scared 🤣 (I think it's doable, but the resource would need to know the current account ID and region to build an ARN and it feels brittle). Option 2 sounds good to me - I'll look at it later today.

gazoakley:
Hi @bflad - updated to address review comments 👍

aws/resource_aws_securityhub_product_subscription_test.go
```diff
@@ -0,0 +1,108 @@
...
+`
+
+const testAccAWSSecurityHubProductSubscriptionConfig_basic = `
+resource "aws_securityhub_account" "example" {}
+
+data "aws_region" "current" {}
+
+resource "aws_securityhub_product_subscription" "example" {
+  depends_on  = ["aws_securityhub_account.example"]
+  product_arn = "arn:aws:securityhub:${data.aws_region.current.name}:733251395267:product/alertlogic/althreatmanagement"
```
bflad: Just noting here for future posterity and not that we don't like our friends over at AlertLogic, but its a little bit of a bummer that the AWS services are automatically subscribed for acceptance testing purposes since they are effectively free, assumed stable for a long while, and doesn't involve an extra EULA. (I'm a little surprised the API doesn't require accepting the EULA first.)

I'll try to switch to using a `PreConfig` to disable something like GuardDuty and re-enable it here in the configuration, but if that doesn't go as well as I'd expect, the pricing for this service is fine for the (presumably single hour) time it would be billed.

bflad: 🎉 7cdbfcf3520d41117bd7509ed42528bea7f2cc14

bflad:
This has been released in [version 1.54.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1540-december-21-2018). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #6925, opened 2018-12-19
Fix gosimple autoscaling issues

vnguyen7:

       Fix gosimple autoscaling issues

### DIFF
--- a/aws/resource_aws_autoscaling_group_waiting.go
+++ b/aws/resource_aws_autoscaling_group_waiting.go
@@ -115,7 +115,7 @@ func waitForASGCapacity(
 	})
 	if aErr == nil {
 		if len(resp.Activities) > 0 {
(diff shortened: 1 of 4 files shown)


aws/resource_aws_autoscaling_group_waiting.go
```diff
@@ -115,7 +115,7 @@ func waitForASGCapacity(
 	})
 	if aErr == nil {
 		if len(resp.Activities) > 0 {
-			recentStatus = fmt.Sprintf("%s", resp.Activities[0])
+			recentStatus = (resp.Activities[0].String())
```
bflad: Nit: We generally prefer to use the AWS Go SDK provided type conversion functions since they will safely convert `nil` pointers to zero values instead of causing a panic, e.g. `aws.StringValue(resp.Activities[0])`, but since this is no worse here, its okay for now.

PR #6941, opened 2018-12-20
Fix gosimple elastic issues

vnguyen7:

     Fix gosimple elastic issues

### DIFF
--- a/aws/resource_aws_elasticache_cluster.go
+++ b/aws/resource_aws_elasticache_cluster.go
@@ -438,7 +438,7 @@ func resourceAwsElasticacheClusterRead(d *schema.ResourceData, meta interface{})
 		if c.ConfigurationEndpoint != nil {
 			d.Set("port", c.ConfigurationEndpoint.Port)
 			d.Set("configuration_endpoint", aws.String(fmt.Sprintf("%s:%d", *c.ConfigurationEndpoint.Address, *c.ConfigurationEndpoint.Port)))
-			d.Set("cluster_address", aws.String(fmt.Sprintf("%s", *c.ConfigurationEndpoint.Address)))
+			d.Set("cluster_address", aws.String((*c.ConfigurationEndpoint.Address)))
 		} else if len(c.CacheNodes) > 0 {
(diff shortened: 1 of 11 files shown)


aws/resource_aws_elasticache_replication_group_test.go
```diff
@@ -614,10 +614,9 @@ func TestAccAWSElasticacheReplicationGroup_NumberCacheClusters_Failover_AutoFail
 				PreConfig: func() {
 					// Simulate failover so primary is on node we are trying to delete
 					conn := testAccProvider.Meta().(*AWSClient).elasticacheconn
-					var input *elasticache.ModifyReplicationGroupInput
 
 					// Must disable automatic failover first
-					input = &elasticache.ModifyReplicationGroupInput{
+					var input *elasticache.ModifyReplicationGroupInput = &elasticache.ModifyReplicationGroupInput{
```
bflad: Nit: Its easier to just use `input := &elasticache.ModifyReplicationGroupInput{` instead of the long form syntax

aws/resource_aws_elasticache_cluster.go
```diff
@@ -438,7 +438,7 @@ func resourceAwsElasticacheClusterRead(d *schema.ResourceData, meta interface{})
 		if c.ConfigurationEndpoint != nil {
 			d.Set("port", c.ConfigurationEndpoint.Port)
 			d.Set("configuration_endpoint", aws.String(fmt.Sprintf("%s:%d", *c.ConfigurationEndpoint.Address, *c.ConfigurationEndpoint.Port)))
-			d.Set("cluster_address", aws.String(fmt.Sprintf("%s", *c.ConfigurationEndpoint.Address)))
+			d.Set("cluster_address", aws.String((*c.ConfigurationEndpoint.Address)))
```
bflad: Nit: Its easier to just passthrough the existing string pointer instead of dereferencing it and then creating a pointer again. `d.Set("cluster_address", c.ConfigurationEndpoint.Address)`

PR #7009, opened 2019-01-02
r/aws_client_vpn_endpoint:  Adding new resource to manage AWS Client VPN endpoints

### LINKED ISSUE #6949: Support for AWS Client VPN
### Community Note

### Description

AWS just released a new VPN service that works with OpenVPN-based clients

### References

https://aws.amazon.com/about-aws/whats-new/2018/12/introducing-aws-client-vpn-to-securely-access-aws-and-on-premises-resources/

slapula:
Fixes #6949

* Adds a new resource that manages AWS Client VPN endpoints: `aws_client_vpn_endpoint`

```
$ make testacc TESTARGS='-run=TestAccAwsEc2ClientVpnEndpoint_'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsEc2ClientVpnEndpoint_ -timeout 120m
=== RUN   TestAccAwsEc2ClientVpnEndpoint_basic
--- PASS: TestAccAwsEc2ClientVpnEndpoint_basic (10.77s)
=== RUN   TestAccAwsEc2ClientVpnEndpoint_msAD
--- PASS: TestAccAwsEc2ClientVpnEndpoint_msAD (1748.41s)
=== RUN   TestAccAwsEc2ClientVpnEndpoint_withLogGroup
--- PASS: TestAccAwsEc2ClientVpnEndpoint_withLogGroup (18.49s)
=== RUN   TestAccAwsEc2ClientVpnEndpoint_withDNSServers
--- PASS: TestAccAwsEc2ClientVpnEndpoint_withDNSServers (16.18s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1793.872s
```


aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+							Required: true,
+						},
+					},
+				},
+			},
+			"dns_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"status": {
```
gazoakley: Terraform doesn't normally expose status attributes - they're normally waited on instead (see below)

aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsClientVpnEndpoint() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsClientVpnEndpointCreate,
+		Read:   resourceAwsClientVpnEndpointRead,
+		Delete: resourceAwsClientVpnEndpointDelete,
+		Update: resourceAwsClientVpnEndpointUpdate,
+		Importer: &schema.ResourceImporter{
```
gazoakley: I'd add a Timeout block for the creation/deletion `WaitForState` calls (see below):

```go
			Create: schema.DefaultTimeout(5 * time.Minute),
			Delete: schema.DefaultTimeout(5 * time.Minute),
		},
```

aws/resource_aws_client_vpn_endpoint_test.go
```diff
@@ -0,0 +1,259 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAwsClientVpnEndpoint_importBasic(t *testing.T) {
```
gazoakley: Nit: New resources are normally written without a specific import test like this (the ImportState step is just added to all other test cases instead)

aws/resource_aws_client_vpn_endpoint_test.go
```diff
@@ -0,0 +1,259 @@
...
+		Providers:    testAccProvidersWithTLS,
+		CheckDestroy: testAccCheckAwsClientVpnEndpointDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccClientVpnEndpointConfig(acctest.RandString(5)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsClientVpnEndpointExists("aws_client_vpn_endpoint.test"),
+				),
+			},
+		},
```
gazoakley: Nit: Add an  ImportState step:

```go
			{
 				ImportState:       true,
 				ImportStateVerify: true,
 			},
```

aws/resource_aws_client_vpn_endpoint_test.go
```diff
@@ -0,0 +1,259 @@
...
+		CheckDestroy: testAccCheckAwsClientVpnEndpointDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccClientVpnEndpointConfigWithLogGroup(acctest.RandString(5)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsClientVpnEndpointExists("aws_client_vpn_endpoint.test"),
+				),
+			},
+		},
+	})
```
gazoakley: Nit: Add an ImportState step (see above)

aws/resource_aws_client_vpn_endpoint_test.go
```diff
@@ -0,0 +1,259 @@
...
+		Providers:    testAccProvidersWithTLS,
+		CheckDestroy: testAccCheckAwsClientVpnEndpointDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccClientVpnEndpointConfigWithDNSServers(acctest.RandString(5)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsClientVpnEndpointExists("aws_client_vpn_endpoint.test"),
+				),
+			},
+		},
```
gazoakley: Nit: Add an ImportState step (see above)

aws/resource_aws_client_vpn_endpoint_test.go
```diff
@@ -0,0 +1,259 @@
...
+	})
+}
+
```
gazoakley: There aren't any test cases here that test updating the resource in `resourceAwsClientVpnEndpointUpdate`

website/docs/r/client_vpn_endpoint.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+  server_certificate_arn = "${aws_acm_certificate.cert.arn}"
+  client_cidr_block = "10.0.0.0/16"
+
+  authentication_options {
+    type = "certificate-authentication"
+    root_certificate_chain_arn = "${aws_acm_certificate.root_cert.arn}"
+  }
+
+  connection_log_options {
+	  enabled = true
```
gazoakley: Nit: Could reformat the example (too many spaces here)

aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+			},
+			"server_certificate_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"transport_protocol": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  "udp",
```
gazoakley: We should check the user has provided a valid value:

```go
					ec2.TransportProtocolTcp,
					ec2.TransportProtocolUdp,
				}, false),
```

aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+			"authentication_options": {
+				Type:     schema.TypeSet,
+				Required: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
```
gazoakley:
```go
					ec2.ClientVpnAuthenticationTypeCertificateAuthentication,
					ec2.ClientVpnAuthenticationTypeDirectoryServiceAuthentication,
				}, false),
```

slapula:
@gazoakley Thanks for the feedback!  Thanks for bringing up endpoint `status` as I wasn't quite sure how to handle it with this first pass.  According to the documentation, "status" for an endpoint appears to behave differently from how we expect it to with other AWS resources.  I'm looking at this as a reference: https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/cvpn-getting-started.html#cvpn-getting-started-target.  A Client VPN endpoint sets its status to `pending-associate` upon creation then has to wait for the operator to initiate an action that associates the CIDR block to the endpoint before it can change its status to `available`.  Based on how I have the resource operating here, this would cause Terraform to wait indefinitely until that CIDR association action completes successfully.  

 Now, here's the fork in the road.  Do we roll this `AssociateClientVpnTargetNetwork` API action into this resource or do we create another resource that specifically manages those associations (like `aws_client_vpn_endpoint_association`)?  From what I can tell, the pattern we already follow with this provider seems to align with the latter option.

Hope that all makes sense :smile:

slapula:
@gazoakley Yup, that was my original plan from the start.  Let me address your review comments and then I'll get started on the resource for CIDR associations.

aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+	conn := meta.(*AWSClient).ec2conn
+
+	_, err := conn.DeleteClientVpnEndpoint(&ec2.DeleteClientVpnEndpointInput{
+		ClientVpnEndpointId: aws.String(d.Id()),
+	})
+	if err != nil {
+		return fmt.Errorf("Error deleting Client VPN endpoint: %s", err)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
slapula: As discussed, these statuses do not lend themselves well to this particular pattern.  I will be writing a CIDR association resource to handle the outstanding issue of getting a created endpoint (which this resource handles) into an `available` state.

aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	log.Printf("[DEBUG] Creating Client VPN endpoint: %#v", req)
+	resp, err := conn.CreateClientVpnEndpoint(req)
+	if err != nil {
+		return fmt.Errorf("Error creating Client VPN endpoint: %s", err)
+	}
+
+	d.SetId(*resp.ClientVpnEndpointId)
+	d.Set("dns_name", resp.DnsName)
```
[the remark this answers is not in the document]
slapula: Understood.  These have been removed.

aws/resource_aws_client_vpn_endpoint_test.go
slapula: Done

website/docs/r/client_vpn_endpoint.html.markdown
slapula: Done

aws/resource_aws_client_vpn_endpoint.go
slapula: Done (using the actual values instead of these example stucts)

aws/resource_aws_client_vpn_endpoint.go
slapula: Done (using the actual values instead of these example stucts)

aws/resource_aws_client_vpn_endpoint_test.go
slapula: I have removed this test and rolled the functions in the import file into the main resource file.

aws/resource_aws_client_vpn_endpoint_test.go
slapula: Done

aws/resource_aws_client_vpn_endpoint_test.go
slapula: I've rolled this into the `withLogGroup` and `withDNSServers` tests.  Two birds with one stone.

aws/resource_aws_client_vpn_endpoint_test.go
slapula: Done

aws/resource_aws_client_vpn_endpoint.go
```diff
@@ -0,0 +1,279 @@
...
+		ClientVpnEndpointIds: []*string{aws.String(d.Id())},
+	})
+
+	if err != nil {
+		return fmt.Errorf("Error reading Client VPN endpoint: %s", err)
+	}
+
+	d.Set("dns_name", result.ClientVpnEndpoints[0].DnsName)
+	d.Set("status", result.ClientVpnEndpoints[0].Status)
+
```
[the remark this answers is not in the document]
slapula: * this has been addressed (see related conversation in this PR)
* I have added those attributes to the Read function

slapula:
@gazoakley I believe I've addressed your comments!  Let me know if you need me to take a look at anything else.

aws/provider.go
```diff
@@ -337,6 +337,7 @@ func Provider() terraform.ResourceProvider {
 			"aws_autoscaling_policy":                           resourceAwsAutoscalingPolicy(),
 			"aws_autoscaling_schedule":                         resourceAwsAutoscalingSchedule(),
 			"aws_budgets_budget":                               resourceAwsBudgetsBudget(),
+			"aws_client_vpn_endpoint":                          resourceAwsClientVpnEndpoint(),
```
bflad: I apologize I don't have time for a full review right now, but I figure its worth mentioning that we are preferring to add EC2 naming to EC2 service resources going forward to help consolidate the resource naming and in this case prevent any future issues where another service might introduce a "Client VPN". 👍 

```suggestion
			"aws_ec2_client_vpn_endpoint":                          resourceAwsEc2ClientVpnEndpoint(),
```

(along with matching function and filename updates)

slapula: @bflad This actually another good question.  The API classifies this as an EC2 service but I see that Terraform also categorizes some VPC services separately even if they are still technically filed under EC2.  If you look at the Client VPN service in the console, you'll see this reflected by the fact that it's accessible in the VPC section.  Would it be preferable to name this `aws_vpc_client_vpn_endpoint`

bflad: I would lean towards `aws_ec2_client_vpn_endpoint`. Between the two (`_ec2_` vs `_vpc_`) we do have precedent of only including `_vpc_` in the resource naming only with [EC2 API operations with "Vpc" in the name](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Operations.html) (except for `aws_vpc_dhcp_options`). This also provides some potential future proofing if some of these EC2 networking related resources work outside VPCs should some other networking concepts be introduced in the API.

The important part to remember here is that the API naming is considerably more stable than the web console, which could (theoretically) change at any time with browser redirects.

slapula: Okay, makes sense to me!  I'll go ahead and make the appropriate changes.

slapula: @bflad This has been done.  See 9f077a7.

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+
+	return resourceAwsEc2ClientVpnEndpointRead(d, meta)
+}
+
+func expandEc2ConnectionLogOptionsRequest(data map[string]interface{}) *ec2.ConnectionLogOptions {
+	req := &ec2.ConnectionLogOptions{
+		Enabled: aws.Bool(data["enabled"].(bool)),
+	}
+
+	if data["enabled"].(bool) == true && data["cloudwatch_log_group"].(string) != "" {
```
bflad: Nit: `gosimple` linter (being enabled soon) will flag this to remove the `== true` since its redundant

```suggestion
	if data["enabled"].(bool) && data["cloudwatch_log_group"].(string) != "" {
```

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+		req.CloudwatchLogGroup = aws.String(data["cloudwatch_log_group"].(string))
+	}
+
+	if data["enabled"].(bool) == true && data["cloudwatch_log_stream"].(string) != "" {
```
bflad: Nit: `gosimple` linter (being enabled soon) will flag this to remove the `== true` since its redundant

```suggestion
	if data["enabled"].(bool) && data["cloudwatch_log_stream"].(string) != "" {
```

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+
+	return req
+}
+
+func expandEc2ClientVpnAuthenticationRequest(data map[string]interface{}) *ec2.ClientVpnAuthenticationRequest {
+	req := &ec2.ClientVpnAuthenticationRequest{
+		Type: aws.String(data["type"].(string)),
+	}
+
+	if data["type"].(string) == "certificate-authentication" {
```
bflad: Nit: AWS Go SDK provided constant available:

```suggestion
	if data["type"].(string) == ec2.ClientVpnAuthenticationTypeCertificateAuthentication {
```

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+		req.MutualAuthentication = &ec2.CertificateAuthenticationRequest{
+			ClientRootCertificateChainArn: aws.String(data["root_certificate_chain_arn"].(string)),
+		}
+	}
+
+	if data["type"].(string) == "directory-service-authentication" {
```
bflad: Nit: AWS Go SDK provided constant available:

```suggestion
	if data["type"].(string) == ec2.ClientVpnAuthenticationTypeDirectoryServiceAuthentication {
```

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ForceNew:     true,
+							ValidateFunc: validation.StringInSlice([]string{"certificate-authentication", "directory-service-authentication"}, false),
+						},
+						"active_directory_id": {
```
bflad: This attribute does not currently have an associated acceptance test.

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+		m["cloudwatch_log_group"] = *lopts.CloudwatchLogGroup
+	}
+	if lopts.CloudwatchLogStream != nil {
+		m["cloudwatch_log_stream"] = *lopts.CloudwatchLogStream
+	}
+	m["enabled"] = *lopts.Enabled
+	return schema.NewSet(connLoggingConfigHash, []interface{}{m})
+}
+
+func connLoggingConfigHash(v interface{}) int {
```
bflad: This can be removed with switching to `TypeList` 👍

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,347 @@
...
+		m["root_certificate_chain_arn"] = *aopts[0].MutualAuthentication.ClientRootCertificateChain
+	}
+	if aopts[0].ActiveDirectory != nil {
+		m["active_directory_id"] = *aopts[0].ActiveDirectory.DirectoryId
+	}
+	m["type"] = *aopts[0].Type
+	return schema.NewSet(authOptsConfigHash, []interface{}{m})
+}
+
+func authOptsConfigHash(v interface{}) int {
```
bflad: This can be removed with switching to `TypeList` 👍

aws/resource_aws_ec2_client_vpn_endpoint.go
slapula: I added a new acceptance test to check for using the AD based authentication option.

aws/resource_aws_ec2_client_vpn_endpoint.go
slapula: I tried using the suggested change above but it never worked for me (both with and without defining the type as string).  Am I just not comparing these values the right way?

slapula:
@bflad Thanks for the feedback! I believe I have address all of your comments.

slapula:
@bflad It appears that gosimple was enabled and is now alerting on a soon-to-be removed check: https://staticcheck.io/changes/2019.1

hussfelt:
Stalled? :)
Awesome resource to have.

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,323 @@
...
+
+		if attrs["enabled"].(bool) && attrs["cloudwatch_log_stream"].(string) != "" {
+			connLogReq.CloudwatchLogStream = aws.String(attrs["cloudwatch_log_stream"].(string))
+		}
+
+		req.ConnectionLogOptions = connLogReq
+	}
+
+	log.Printf("[DEBUG] Creating Client VPN endpoint: %#v", req)
+	resp, err := conn.CreateClientVpnEndpoint(req)
```
bflad: It looks like we need to use a mutex and/or `resource.Retry()` to protect from some AWS concurrency errors:

```
--- FAIL: TestAccAwsEc2ClientVpnEndpoint_basic (4.44s)
        	* aws_ec2_client_vpn_endpoint.test: 1 error occurred:
        	* aws_ec2_client_vpn_endpoint.test: Error creating Client VPN endpoint: OperationNotPermitted: Endpoint cannot be created while another endpoint is being created or the service linked role is being deleted
```

```suggestion
	var resp *ec2.CreateClientVpnEndpointOutput
	err := resource.Retry(1 * time.Minute, func() *resource.RetryError {
		var err error
		resp, err = conn.CreateClientVpnEndpoint(req)

		if isAWSErr(err, "OperationNotPermitted", "Endpoint cannot be created while another endpoint is being created") {
			return resource.RetryableError(err)
		}

		if err != nil {
			return resource.NonRetryableError(err)
		}

		return nil
	})
```

aws/resource_aws_ec2_client_vpn_endpoint.go
```diff
@@ -0,0 +1,323 @@
...
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"client_cidr_block": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"dns_servers": {
```
bflad: This attribute is not being set into the Terraform state in `resourceAwsEc2ClientVpnEndpointRead` so Terraform is unable to appropriately perform drift detection or import the attribute.

```
--- FAIL: TestAccAwsEc2ClientVpnEndpoint_withDNSServers (24.62s)
    testing.go:538: Step 2 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=3) {
         (string) (len=13) "dns_servers.#": (string) (len=1) "2",
         (string) (len=22) "dns_servers.2112610058": (string) (len=7) "8.8.4.4",
         (string) (len=22) "dns_servers.2609393598": (string) (len=7) "8.8.8.8"
        }
```

We'll need to call `d.Set()` with the correct information from the API or if its not possible to get, use `ImportStateVerifyIgnore: []string{"dns_servers"},` in the acceptance testing.

aws/resource_aws_ec2_client_vpn_endpoint_test.go
```diff
@@ -0,0 +1,382 @@
...
+func testAccEc2ClientVpnEndpointConfigWithMicrosoftAD(rName string) string {
+	return fmt.Sprintf(`
+resource "aws_vpc" "test" {
+  cidr_block  = "10.0.0.0/16"
+}
+
+resource "aws_subnet" "test1" {
+  vpc_id     				= "${aws_vpc.test.id}"
+  cidr_block 				= "10.0.1.0/24"
+  availability_zone = "us-east-1a"
```
bflad: While configurable, by default the acceptance testing runs in us-west-2:

https://github.com/terraform-providers/terraform-provider-aws/blob/03eb00938807d553711008e93dacfa0c6b0492fa/aws/provider_test.go#L183-L189

Setting these to us-east-1X availability zones generates an error in the acceptance testing:

```
--- FAIL: TestAccAwsEc2ClientVpnEndpoint_msAD (4.15s)
    testing.go:538: Step 0 error: Error applying: 2 errors occurred:
        	* aws_subnet.test2: 1 error occurred:
        	* aws_subnet.test2: Error creating subnet: InvalidParameterValue: Value (us-east-1d) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-west-2a, us-west-2c, us-west-2b.
        	status code: 400, request id: 62334444-44e2-4655-8adc-00fde6a2b7a3
        
        
        	* aws_subnet.test1: 1 error occurred:
        	* aws_subnet.test1: Error creating subnet: InvalidParameterValue: Value (us-east-1a) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-west-2a, us-west-2c, us-west-2b.
        	status code: 400, request id: 8218ec9d-a281-49cf-998a-1deef24ab6b9
```

Rather than hardcoding values or hardcoding a provider configuration to the `us-east-1` region unnecessarily, we generally use the `aws_availability_zones` data source to find available availability zones, e.g.

```hcl
data "aws_availability_zones" "available" {}
# ...
  availability_zone = "${data.aws_availablility_zones.available.names[0]}"
# ...
  availability_zone = "${data.aws_availablility_zones.available.names[1]}"
```

aws/resource_aws_ec2_client_vpn_endpoint.go
slapula: I'm opting to ignore this attribute due to the fact that the Describe action doesn't include it.

aws/resource_aws_ec2_client_vpn_endpoint.go
slapula: Thanks for pointing this out!  This recommendation has been added.

aws/resource_aws_ec2_client_vpn_endpoint_test.go
slapula: Fixed.

slapula:
@bflad Thanks again for your feedback!  I believe I've addressed your comments. This ready for another round.

bflad:

Bwanabanana:
Thanks for this feature, it saves a lot of messing about!  Some feedback - I couldn't find a way to automatically authorise ingress on the client vpn endpoint.  Also, would be nice to be able to associate one or more custom security groups (perhaps using another resource) rather than being stuck with it using the VPC's default group.

slapula:
@Bwanabanana can you open a new issue and tag me in it? Let get a formal request open and I'll take a look when I have some free cycles.

Bwanabanana:
@slapula - all done, many thanks!

PR #7015, opened 2019-01-02
Fix example usage

tylersmith34:
The `aws_iam_role.test_role.id` reference is not correct, the `id` attribute is only from a data source, it is a `unique_id` from a resource.  

https://www.terraform.io/docs/providers/aws/r/iam_role.html

* Fix an incorrect attribute reference for an IAM Role resource

```

...
```
I did not run acceptance tests, this is a markdown only change.

### DIFF
--- a/website/docs/r/iam_role_policy.html.markdown
+++ b/website/docs/r/iam_role_policy.html.markdown
@@ -15,7 +15,7 @@ Provides an IAM role policy.
 ```hcl
 resource "aws_iam_role_policy" "test_policy" {
   name = "test_policy"
-  role = "${aws_iam_role.test_role.id}"
+  role = "${aws_iam_role.test_role.unique_id}"
 
   policy = <<EOF
 {


bflad:
Hi @tylersmith34 👋 Can you please provide more context about this change?

The resource calls `d.SetId()` here, which sets the `id` attribute:

https://github.com/terraform-providers/terraform-provider-aws/blob/a002fbc25f8fe358a271b0cec7fae8b2179ee7a8/aws/resource_aws_iam_role.go#L183

Also the `unique_id` attribute returns a value like `AROA...` while the `id`/`name` attributes are the IAM role name.

If the `id` attribute is being reported as missing it might be indicative of some other issue with the resource code, potentially due to IAM's eventual consistency model, but we will require other context about this issue to determine the cause.

tylersmith34:
Hi Brian.  I noticed this when trying to create an `aws_iam_role_policy` resource and compare attributes to an internal module I'm using.  When I looked at the require inputs and outputs from the resources I noticed this issue.  The documentation for `aws_iam_role` does not list `id` as an available attribute [as seen here](https://www.terraform.io/docs/providers/aws/r/iam_role.html#attributes-reference).  However, the [aws_iam_role_policy example](https://www.terraform.io/docs/providers/aws/r/iam_role_policy.html#example-usage) shows it used to send the reference the role when creating the policy.

Maybe my update should be to 'name' instead or if the actual 'id' attribute exists, I should update the `aws_iam_role` document?

bflad:
> Maybe my update should be to 'name' instead or if the actual 'id' attribute exists, I should update the `aws_iam_role` document?

We'd prefer adding the missing `id` attribute to `website/docs/r/iam_role.html.markdown` 👍

tylersmith34:
You bet.  I'll open a different PR so this commit doesn't make it into history.

Issue #6971, opened 2018-12-22
Feature request: AWS Step Functions tags


ewbankkit:
AWS Step Functions state machines and activities [now support](https://aws.amazon.com/about-aws/whats-new/2019/01/aws-step-functions-now-supports-resource-tagging/) [resource tagging](https://docs.aws.amazon.com/cli/latest/reference/stepfunctions/tag-resource.html).
Affected resources:
* [`aws_sfn_activity`](https://www.terraform.io/docs/providers/aws/r/sfn_activity.html)
* [`aws_sfn_state_machine`](https://www.terraform.io/docs/providers/aws/r/sfn_state_machine.html)

PR #7024, opened 2019-01-03
Add support for tags for aws_sfn_state_machine and aws_sfn_activity

sunilkumarmohanty:
As part of #6971

* Tag support for `aws_sfn_state_machine` and `aws_sfn_activity`

```
$ make testacc TESTARGS='-run=TestAccAWSSfnStateMachine_Tags'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSSfnStateMachine_Tags -timeout 120m
=== RUN   TestAccAWSSfnStateMachine_Tags
=== PAUSE TestAccAWSSfnStateMachine_Tags
=== CONT  TestAccAWSSfnStateMachine_Tags
--- PASS: TestAccAWSSfnStateMachine_Tags (95.35s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       95.384s...



make testacc TESTARGS='-run=TestAccAWSSfnActivity_Tags'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSSfnActivity_Tags -timeout 120m
=== RUN   TestAccAWSSfnActivity_Tags
=== PAUSE TestAccAWSSfnActivity_Tags
=== CONT  TestAccAWSSfnActivity_Tags
--- PASS: TestAccAWSSfnActivity_Tags (50.14s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       50.171s

```

### DIFF
--- a/aws/resource_aws_sfn_activity.go
+++ b/aws/resource_aws_sfn_activity.go
@@ -55,7 +55,18 @@ func resourceAwsSfnActivityCreate(d *schema.ResourceData, meta interface{}) erro
 
 	d.SetId(*activity.ActivityArn)
 
-	return resourceAwsSfnActivityUpdate(d, meta)
+	if v, ok := d.GetOk("tags"); ok {
+		input := &sfn.TagResourceInput{
+			ResourceArn: aws.String(d.Id()),
+			Tags:        tagsFromMapSfn(v.(map[string]interface{})),
+		}
+		log.Printf("[DEBUG] Tagging SFN Activity: %s", input)
+		_, err := conn.TagResource(input)
+		if err != nil {
+			return fmt.Errorf("error tagging SFN Activity (%s): %s", d.Id(), input)
+		}
+	}
+	return resourceAwsSfnActivityRead(d, meta)
 }
 
 func resourceAwsSfnActivityUpdate(d *schema.ResourceData, meta interface{}) error {
--- a/aws/resource_aws_sfn_activity.go
+++ b/aws/resource_aws_sfn_activity.go
@@ -132,6 +132,20 @@ func resourceAwsSfnActivityRead(d *schema.ResourceData, meta interface{}) error 
 		log.Printf("[DEBUG] Error setting creation_date: %s", err)
 	}
 
+	tagsResp, err := conn.ListTagsForResource(
+		&sfn.ListTagsForResourceInput{
+			ResourceArn: aws.String(d.Id()),
+		},
+	)
+
+	if err != nil {
+		return fmt.Errorf("error listing SFN Activity (%s) tags: %s", d.Id(), err)
+	}
+
+	if err := d.Set("tags", tagsToMapSfn(tagsResp.Tags)); err != nil {
+		return fmt.Errorf("error setting tags: %s", err)
+	}
+
 	return nil
 }
 
--- a/aws/resource_aws_sfn_activity.go
+++ b/aws/resource_aws_sfn_activity.go
@@ -17,6 +17,7 @@ func resourceAwsSfnActivity() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceAwsSfnActivityCreate,
 		Read:   resourceAwsSfnActivityRead,
+		Update: resourceAwsSfnActivityUpdate,
 		Delete: resourceAwsSfnActivityDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
@@ -34,6 +35,7 @@ func resourceAwsSfnActivity() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"tags": tagsSchema(),
 		},
 	}
 }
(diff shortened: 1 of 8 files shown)


aws/resource_aws_sfn_activity.go
```diff
@@ -53,6 +55,48 @@ func resourceAwsSfnActivityCreate(d *schema.ResourceData, meta interface{}) erro
 
 	d.SetId(*activity.ActivityArn)
 
+	return resourceAwsSfnActivityUpdate(d, meta)
```
bflad: This is trivial case now, but we prefer to always call the `Read` function after `Create` to prevent duplicate modifications on creation. When doing `Update` after `Create`, `d.IsNewResource()` conditional checks are required in the `Update` function (oft forgotten or missed during code review) to prevent the duplicate update should another attribute that is supported during `Create` and `Update` be added in the future.

When we get around to adding our own linter for the Terraform AWS provider codebase, `Update` after `Create` will be an item that is flagged.

The `Create` function should instead add the logic to call `TagResource` here, then call `Read` instead of `Update`:

```suggestion
	if v, ok := d.GetOk("tags"); ok {
		input := &sfn.TagResourceInput{
			ResourceArn: aws.String(d.Id()),
			Tags:        tagsFromMapSfn(v.(map[string]interface{})),,
		}

		log.Printf("[DEBUG] Tagging SFN Activity: %s", input)
		_, err := conn.TagResource(input)
		if err != nil {
			return fmt.Errorf("error tagging SFN Activity (%s): %s", d.Id(), input)
		}
	}

	return resourceAwsSfnActivityRead(d, meta)
```

aws/resource_aws_sfn_activity.go
```diff
@@ -34,6 +35,7 @@ func resourceAwsSfnActivity() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"tags": tagsSchema(),
```
bflad: This new attribute is missing drift detection in the `Read` function. It can be added similar to the state machine read logic below (with its update).

aws/resource_aws_sfn_state_machine.go
```diff
@@ -92,7 +93,7 @@ func resourceAwsSfnStateMachineCreate(d *schema.ResourceData, meta interface{})
 
 	d.SetId(*activity.StateMachineArn)
 
-	return resourceAwsSfnStateMachineRead(d, meta)
+	return resourceAwsSfnStateMachineUpdate(d, meta)
```
bflad: Same here as above:

```suggestion
	if v, ok := d.GetOk("tags"); ok {
		input := &sfn.TagResourceInput{
			ResourceArn: aws.String(d.Id()),
			Tags:        tagsFromMapSfn(v.(map[string]interface{})),,
		}

		log.Printf("[DEBUG] Tagging SFN State Machine: %s", input)
		_, err := conn.TagResource(input)
		if err != nil {
			return fmt.Errorf("error tagging SFN State Machine (%s): %s", d.Id(), input)
		}
	}

	return resourceAwsSfnStateMachineRead(d, meta)
```

aws/resource_aws_sfn_state_machine.go
```diff
@@ -121,6 +122,16 @@ func resourceAwsSfnStateMachineRead(d *schema.ResourceData, meta interface{}) er
 	if err := d.Set("creation_date", sm.CreationDate.Format(time.RFC3339)); err != nil {
 		log.Printf("[DEBUG] Error setting creation_date: %s", err)
 	}
+	tagsResp, err := conn.ListTagsForResource(
+		&sfn.ListTagsForResourceInput{
+			ResourceArn: aws.String(d.Id()),
+		},
+	)
+	if err != nil {
+		log.Printf("[DEBUG] Error retrieving tags for Step Function: %s. %s", aws.StringValue(sm.Name), err)
```
bflad: We should always return the error instead of logging a debug message to alert operators that Terraform resource functionality is not working as expected. If there are non-AWS Commercial environments that require special handling because the API does not exist (e.g. "UnsupportedOperation"), we can add that logic with the `err != nil` check. 👍

aws/resource_aws_sfn_activity.go
bflad: Hey @sunilkumarmohanty -- not sure if you are done making changes or didn't push something up, but you marked this as resolved and I don't see any changes to `resourceAwsSfnActivityRead()`, e.g. adding

```go
  tagsResp, err := conn.ListTagsForResource(
    &sfn.ListTagsForResourceInput{
      ResourceArn: aws.String(d.Id()),
    },
  )

  if err != nil {
    return fmt.Errorf("error listing SFN Activity (%s) tags: %s", d.Id(), err)
  }

  if err := d.Set("tags", tagsToMapSfn(tagsResp.Tags)); err != nil {
  }
```

sunilkumarmohanty: I am sorry, I forgot to make these changes. It is done now.

sunilkumarmohanty:
@bflad  I am sorry, I forgot to make these changes. It is done now.

bflad:

PR #7030, opened 2019-01-04
 r/aws_ec2_client_vpn_network_association: Adding resource to manage Client VPN network associations

### LINKED ISSUE #6949: Support for AWS Client VPN

slapula:
Fixes #6949 

* Adds a new resource that manages AWS Client VPN endpoints: `aws_ec2_client_vpn_network_association`

```
$ make testacc TESTARGS='-run=TestAccAwsEc2ClientVpnNetworkAssociation_'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsEc2ClientVpnNetworkAssociation_ -timeout 120m
=== RUN   TestAccAwsEc2ClientVpnNetworkAssociation_basic
=== PAUSE TestAccAwsEc2ClientVpnNetworkAssociation_basic
=== CONT  TestAccAwsEc2ClientVpnNetworkAssociation_basic
--- PASS: TestAccAwsEc2ClientVpnNetworkAssociation_basic (661.40s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	661.424s
```


slapula:
This branch is based off https://github.com/terraform-providers/terraform-provider-aws/pull/7009 since I need that functionality to run tests.

slapula:
@gazoakley Here's where your `WaitForState()` logic comes into play for AWS Client VPNs.

slapula:
I am aware the documentation is missing at the moment. I'll be adding that tomorrow.

aws/resource_aws_ec2_client_vpn_network_association.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsEc2ClientVpnNetworkAssociation() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsEc2ClientVpnNetworkAssociationCreate,
+		Read:   resourceAwsEc2ClientVpnNetworkAssociationRead,
+		Delete: resourceAwsEc2ClientVpnNetworkAssociationDelete,
+		Importer: &schema.ResourceImporter{
```
bflad: This import support is missing:
* Acceptance testing in the form of a `TestStep` that includes `ImportState: true` and `ImportStateVerify: true`
* Documentation in the form of a `## Import` section in the resource documentation.

The support should be fully implemented or removed from the pull request. 👍 

If you are opting to keep the support, you will notice that you will need to either implement a custom import function to that calls `d.Set("client_vpn_endpoint_id", ...)` and trims that part of the ID for `d.SetId()` (then use `ImportStateIdFunc` in acceptance testing) or switch the resource to implementing a "complex" ID that includes both IDs in its string, e.g. `ENDPOINT/ASSOCIATION`. For new implementations, we have generally preferred the latter form where we parse the complex ID and continue to use `ImportStatePassthrough`, but either form is acceptable.

aws/resource_aws_ec2_client_vpn_network_association.go
```diff
@@ -0,0 +1,136 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"vpc_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+
+		Timeouts: &schema.ResourceTimeout{
```
bflad: We should prefer to remove these customizable timeouts as this resource does not seem have a scalable dimension for the timeout (e.g. an amount of data affecting the operation). Otherwise, it should at least be documented with a `## Timeouts` section in the resource documentation.

aws/resource_aws_ec2_client_vpn_network_association.go
```diff
@@ -0,0 +1,136 @@
...
+	conn := meta.(*AWSClient).ec2conn
+
+	_, err := conn.DisassociateClientVpnTargetNetwork(&ec2.DisassociateClientVpnTargetNetworkInput{
+		ClientVpnEndpointId: aws.String(d.Get("client_vpn_endpoint_id").(string)),
+		AssociationId:       aws.String(d.Id()),
+	})
+	if err != nil {
+		return fmt.Errorf("Error deleting Client VPN network association: %s", err)
+	}
+
```
bflad: Given the usage of `resource.StateChangeConf` in the `Create` function, it seems likely that we should also be performing the same type of waiting for disassociation in the `Delete` function here to prevent issues while destroying subnets at the same time as this resource.

aws/resource_aws_ec2_client_vpn_network_association.go
```diff
@@ -0,0 +1,136 @@
...
+		ClientVpnEndpointId: aws.String(d.Get("client_vpn_endpoint_id").(string)),
+		AssociationIds:      []*string{aws.String(d.Id())},
+	})
+
+	if err != nil {
+		return fmt.Errorf("Error reading Client VPN network association: %s", err)
+	}
+
+	d.Set("client_vpn_endpoint_id", result.ClientVpnTargetNetworks[0].ClientVpnEndpointId)
+	d.Set("security_groups", result.ClientVpnTargetNetworks[0].SecurityGroups)
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```suggestion
	if err := d.Set("security_groups", result.ClientVpnTargetNetworks[0].SecurityGroups); err != nil {
		return fmt.Errorf("error setting security_groups: %s", err)
	}
```

aws/resource_aws_ec2_client_vpn_network_association_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/hashicorp/terraform/helper/acctest"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAwsEc2ClientVpnNetworkAssociation_basic(t *testing.T) {
+	var assoc1 ec2.TargetNetwork
+	rStr := acctest.RandString(5)
+
+	resource.Test(t, resource.TestCase{
```
bflad: This acceptance testing should use `resource.ParallelTest()` instead of `resource.Test()`

```suggestion
	resource.ParallelTest(t, resource.TestCase{
```

aws/resource_aws_ec2_client_vpn_network_association.go
slapula: Oh, this is cruft from my copy/paste when I started working on this.  I'm going to opt to *not* include import functionality here since I wasn't planning on adding it initially.  I can be swayed to add it however because I can see the need for it.

aws/resource_aws_ec2_client_vpn_network_association.go
```diff
@@ -0,0 +1,151 @@
...
+		Pending: []string{ec2.AssociationStatusCodeDisassociating},
+		Target:  []string{ec2.AssociationStatusCodeDisassociated},
+		Refresh: clientVpnNetworkAssociationRefreshFunc(conn, d.Id(), d.Get("client_vpn_endpoint_id").(string)),
+		Timeout: d.Timeout(schema.TimeoutDelete),
+	}
+
+	log.Printf("[DEBUG] Waiting for Client VPN endpoint to disassociate with target network: %s", d.Id())
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		if strings.Contains(err.Error(), "couldn't find resource") != true {
```
slapula: @bflad This seems super janky to me. Is there a better way to handle this exception?  I can't seem to find the specific error code that is tripping on this condition so that I can put it in the RefreshFunc (where this logic should actually live IMO).

slapula:
@bflad This is ready for another look.

aws/resource_aws_ec2_client_vpn_network_association.go
bflad: Great question! Yeah, don't fret too much about this as EC2 does things a bit differently than other AWS services for errors and they do not make the error codes available in the AWS Go SDK as constants. 🙁 

For reference, they mostly (I've seen cases where they are missing for a long period of time) can be found here: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

Rather than guessing from the API reference though, I personally tend to look for the error codes by enabling debug logging in Terraform during an acceptance test run. For example, in this case, I ran the below:

```console
$ rm -f aws/log.txt; TF_LOG_PATH=log.txt TF_LOG=debug TF_ACC=1 go test ./aws -v -timeout 120m -parallel 20 -run=TestAccAwsEc2ClientVpnNetworkAssociation_basic
=== RUN   TestAccAwsEc2ClientVpnNetworkAssociation_basic
=== PAUSE TestAccAwsEc2ClientVpnNetworkAssociation_basic
=== CONT  TestAccAwsEc2ClientVpnNetworkAssociation_basic
--- PASS: TestAccAwsEc2ClientVpnNetworkAssociation_basic (690.14s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	691.015s
```

Which tells Terraform to create a log file with the debug output (you can stick it anywhere, I'm just lazy with `log.txt` 😄) and keep the console output "normal". In the debug log it output this:

```
2019/02/07 21:22:56 [DEBUG] [aws-sdk-go] DEBUG: Response ec2/DescribeClientVpnTargetNetworks Details:
---[ RESPONSE ]--------------------------------------
HTTP/1.1 400 Bad Request
Connection: close
Transfer-Encoding: chunked
Date: Fri, 08 Feb 2019 02:22:55 GMT
Server: AmazonEC2


-----------------------------------------------------
2019/02/07 21:22:56 [DEBUG] [aws-sdk-go] <?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidClientVpnEndpointId.NotFound</Code><Message>Endpoint cvpn-endpoint-0f2e85e16167a3465 does not exist</Message></Error></Errors><RequestID>31e4906a-8249-4e2d-9284-a03a2672ff86</RequestID></Response>
2019/02/07 21:22:56 [DEBUG] [aws-sdk-go] DEBUG: Validate Response ec2/DescribeClientVpnTargetNetworks failed, not retrying, error InvalidClientVpnEndpointId.NotFound: Endpoint cvpn-endpoint-0f2e85e16167a3465 does not exist
```

The `InvalidClientVpnEndpointId.NotFound` is the error code we can use for easily checking, e.g.

```go
if isAWSErr(err, "InvalidClientVpnEndpointId.NotFound", "") {
```

🎉

aws/resource_aws_ec2_client_vpn_network_association.go
```diff
@@ -0,0 +1,151 @@
...
+		Pending: []string{ec2.AssociationStatusCodeDisassociating},
+		Target:  []string{ec2.AssociationStatusCodeDisassociated},
+		Refresh: clientVpnNetworkAssociationRefreshFunc(conn, d.Id(), d.Get("client_vpn_endpoint_id").(string)),
+		Timeout: d.Timeout(schema.TimeoutDelete),
+	}
+
+	log.Printf("[DEBUG] Waiting for Client VPN endpoint to disassociate with target network: %s", d.Id())
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		if !strings.Contains(err.Error(), "couldn't find resource") {
```
bflad: Just so we can get this released today, I'm going to fix this last little bit up given my previous comment. I hope you don't mind. 😄

aws/resource_aws_ec2_client_vpn_network_association.go
bflad: Haha and after writing this I notice we should check for the above _and_ `InvalidClientVpnAssociationId.NotFound`

bflad:

PR #6782, opened 2018-12-10
r/lambda_layer_version: Add lambda_layer_version resource

acburdine:
refs #6651 

Changes proposed in this pull request:

* Add `aws_lambda_layer` resource

TODO:
- [x] add docs for resource
- [x] finish adding remainder of tests

Output from acceptance testing (so far):

```
$ make testacc TESTARGS='-run=TestAccAWSLambdaLayer'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./... -v -parallel 20 -run TestAccAWSLambdaLayer -timeout 120m
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
=== RUN   TestAccAWSLambdaLayer_basic
=== PAUSE TestAccAWSLambdaLayer_basic
=== RUN   TestAccAWSLambdaLayer_s3
=== PAUSE TestAccAWSLambdaLayer_s3
=== CONT  TestAccAWSLambdaLayer_basic
=== CONT  TestAccAWSLambdaLayer_s3
--- PASS: TestAccAWSLambdaLayer_basic (17.40s)
--- PASS: TestAccAWSLambdaLayer_s3 (25.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	36.018s
```

Issue #6651, opened 2018-11-29
Support Lambda Layers


popey456963:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AWS has recently released a new feature, "[Lambda Layers](https://aws.amazon.com/blogs/aws/new-for-aws-lambda-use-any-programming-language-and-share-common-components/)", a way of sharing code between Lambda functions.  It would be useful for Terraform to support this.

### New or Affected Resource(s)

* aws_lambda

### Potential Terraform Configuration

```hcl
# An implementation would likely have two aspects to it, an addition lambda parameter:

resource "aws_lambda_function" "example_function" {
  layers = [
    // reference to layer
    // another reference
  ]
}

# And the layer type itself

resource "aws_lambda_layer" "example_layer" {
  # layer metadata
  layer_name = ""
  tags = []
  description = ""

  # location of layer
  s3_bucket = []
  s3_key = []
  s3_object_version = []
}
```

robertpeteuil:
I was going to submit the same request.  I'm glad to see someone has already suggested it.

robertpeteuil:
Might also need a `data` resource to be able to retrieve the latest version of a layer.

It looks like the [GetLayerVersion](https://docs.aws.amazon.com/lambda/latest/dg/API_GetLayerVersion.html) API supports this.

acburdine:
I'll take a look at this one 😄

ispulkit:
Waiting for this to come to life...

popey456963:
@acburdine Wondered if you'd had a chance to look into this yet?

ampopp04:
If I wanted to deploy a simple AWSLambda-Python37-SciPy1x layer to a python lambda, is there a way to do it within terraform until your enhancement gets pushed? I can look through the docs to see if there is some sort of generic mechanism I can use to call a layer create. Great work by the way!

zilman:
@acburdine Thank you for doing this, your PR (#6782) looks very straightforward, any idea if it will make it into 0.12.0?

alixaxel:
When can we expect this to be included in the stable Terraform release?

[...]

PR #7126, opened 2019-01-12
lambda_function: add layers support to resource and data source

acburdine:
refs #6651 (depends on #6782)
- add layers attribute to resource and data source
- update documentation

TODO:
- [x] add tests once #6782 is merged

* Add `layers` attribute to r/lambda_function and d/lambda_function

Output from acceptance testing: (waiting on aforementioned PR)

```
$ make testacc TESTARGS='-run=TestAccAWSLambdaFunction'

...
```

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -105,6 +105,15 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"layers": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 5,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validateArn,
+				},
+			},
 			"memory_size": {
 				Type:     schema.TypeInt,
 				Optional: true,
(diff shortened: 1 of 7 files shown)


acburdine:
@bflad working on tests for this now, should have them done today

acburdine:
Tests should be good now, although I had a bit of an issue running them due to some restrictions on iam role creation/deletion in the AWS account I normally run my tests in 😕

acburdine:
Ran it in a different account:

Resource tests:
```
--- PASS: TestAccAWSLambdaFunction_Layers (28.41s)
--- PASS: TestAccAWSLambdaFunction_LayersUpdate (41.70s)
PASS
```

Data source tests:
```
--- PASS: TestAccDataSourceAWSLambdaFunction_layers (27.65s)
PASS
```

bflad:
@acburdine awesome! Taking a look now. 😄

aws/resource_aws_lambda_function.go
```diff
@@ -105,6 +105,15 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"layers": {
+				Type:     schema.TypeSet,
```
bflad: I think the ordering of layers matters according to the AWS documentation: https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html

> Layers are applied in the order that's specified, merging any folders with the same name. If the same file appears in multiple layers, the version in the last applied layer is used.

Given this, we should switch this to `TypeList` here and in the data source. 👍

acburdine: Gotcha - will do

acburdine: done!

bflad:

PR #2999, opened 2018-01-15
New Resource: aws_sagemaker_notebook_instance

ddcprg:
Helps resolving #2493

PR #7139, opened 2019-01-15
Sagemaker notebook instance

mbfrahry:
This PR starts the process to add the long requested sagemaker resources to the aws provider. Huge shoutout to @ddcprg in #2999 for getting the bulk of this done. 

```
=== RUN   TestAccAWSSagemakerNotebookInstance_basic
=== PAUSE TestAccAWSSagemakerNotebookInstance_basic
=== RUN   TestAccAWSSagemakerNotebookInstance_update
=== PAUSE TestAccAWSSagemakerNotebookInstance_update
=== RUN   TestAccAWSSagemakerNotebookInstance_tags
=== PAUSE TestAccAWSSagemakerNotebookInstance_tags
=== CONT  TestAccAWSSagemakerNotebookInstance_basic
=== CONT  TestAccAWSSagemakerNotebookInstance_tags
=== CONT  TestAccAWSSagemakerNotebookInstance_update
--- PASS: TestAccAWSSagemakerNotebookInstance_tags (298.50s)
--- PASS: TestAccAWSSagemakerNotebookInstance_basic (431.74s)
--- PASS: TestAccAWSSagemakerNotebookInstance_update (483.99s)
```


aws/resource_aws_sagemaker_notebook_instance.go
```diff
@@ -0,0 +1,390 @@
...
+
+func resourceAwsSagemakerNotebookInstanceDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	describeNotebookInput := &sagemaker.DescribeNotebookInstanceInput{
+		NotebookInstanceName: aws.String(d.Id()),
+	}
+	notebook, err := conn.DescribeNotebookInstance(describeNotebookInput)
+	if err != nil {
+		if sagemakerErr, ok := err.(awserr.Error); ok && sagemakerErr.Code() == "ResourceNotFound" {
```
bflad: Ah ha -- this is related to above ResourceNotFound comment. All instances of this logic should likely be:

```suggestion
		if isAWSErr(err, sagemaker.ErrCodeResourceNotFound, "") {
```

aws/resource_aws_sagemaker_notebook_instance.go
```diff
@@ -0,0 +1,390 @@
...
+			return nil
+		}
+		return fmt.Errorf("unable to find sagemaker notebook instance to delete %q: %s", d.Id(), err)
+	}
+	if *notebook.NotebookInstanceStatus != sagemaker.NotebookInstanceStatusFailed {
+		if err := stopSagemakerNotebookInstance(conn, d.Id()); err != nil {
```
bflad: Should we check to see if its already stopped and skip this if it is? It looks like `stopSagemakerNotebookInstance` might already check this so the `DescribeNotebookInstance` call above may be extraneous.

aws/resource_aws_sagemaker_notebook_instance_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+func testAccCheckSagemakerNotebookInstanceDestroy(s *terraform.State) error {
+	conn := testAccProvider.Meta().(*AWSClient).sagemakerconn
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_sagemaker_notebook_instance" {
+			continue
+		}
+
+		resp, err := conn.ListNotebookInstances(&sagemaker.ListNotebookInstancesInput{
```
bflad: Should this be using `DescribeNotebookInstance` to match the `Read` function and `Exists` test function? A name contains check seems more flakey than directly checking.

website/docs/r/sagemaker_notebook_instance.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+## Attributes Reference
+
+The following attributes are exported:
+
+* `id` - The name of the notebook instance.
+* `arn` - The Amazon Resource Name (ARN) assigned by AWS to this notebook instance.
+
+## Import
+
+Models can be imported using the `name`, e.g.
```
bflad: 🍝 

```suggestion
SageMaker Notebook Instances can be imported using the `name`, e.g.
```

aws/resource_aws_sagemaker_notebook_instance.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func sagemakerNotebookInstanceStateRefreshFunc(conn *sagemaker.SageMaker, name string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		describeNotebookInput := &sagemaker.DescribeNotebookInstanceInput{
+			NotebookInstanceName: aws.String(name),
+		}
+		notebook, err := conn.DescribeNotebookInstance(describeNotebookInput)
+		if err != nil {
+			if sagemakerErr, ok := err.(awserr.Error); ok && sagemakerErr.Message() == "RecordNotFound" {
```
bflad: Should this be RecordNotFound or ResourceNotFound to match others?

aws/resource_aws_sagemaker_notebook_instance.go
```diff
@@ -0,0 +1,390 @@
...
+		if isAWSErr(err, "", "RecordNotFound") {
+			d.SetId("")
+			log.Printf("[LOG] Unable to find sageMaker notebook instance %q; removing from state file", d.Id())
+			return nil
+		}
+		return fmt.Errorf("error finding sagemaker notebook instance %q: %s", d.Id(), err)
+
+	}
+
+	d.Set("security_groups", flattenStringList(notebookInstance.SecurityGroups))
```
bflad: Should add error checking here to match everything else 👍 

```suggestion
	if err := d.Set("security_groups", flattenStringList(notebookInstance.SecurityGroups)); err != nil {
		return fmt.Errorf("error setting security_groups for sagemaker notebook instance %q: %s", d.Id(), err)
	}
```

aws/resource_aws_sagemaker_notebook_instance.go
bflad: Ah I guess `stopSagemakerNotebookInstance` returns different errors and whatnot. Adding the conditional for `sagemaker.NotebookInstanceStatusStopped` is probably best here. 👍

aws/resource_aws_sagemaker_notebook_instance.go
```diff
@@ -0,0 +1,390 @@
...
+func resourceAwsSagemakerNotebookInstanceRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	// TODO change this to describeNotebook Instance
+	describeNotebookInput := &sagemaker.DescribeNotebookInstanceInput{
+		NotebookInstanceName: aws.String(d.Id()),
+	}
+	notebookInstance, err := conn.DescribeNotebookInstance(describeNotebookInput)
+	if err != nil {
+		if isAWSErr(err, "", "RecordNotFound") {
```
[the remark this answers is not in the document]
mbfrahry: All other sagemaker resources return ResourceNotFound except for notebook instance. It's definitely odd.

aws/resource_aws_sagemaker_notebook_instance.go
mbfrahry: This is a mistake and it needs to be like the above as RecordNotFound

aws/resource_aws_sagemaker_notebook_instance.go
mbfrahry: RecordNotFound is the correct usage here

aws/resource_aws_sagemaker_notebook_instance.go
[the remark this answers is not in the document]
mbfrahry: This is the error that gets returned when you try and do something with a notebook instance that has been deleted `ValidationException: RecordNotFound`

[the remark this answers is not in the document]
bflad: `isAWSErr()` does a strict check on the second parameter so `isAWSErr(err, "ValidationException", "RecordNotFound")` should do the trick then. 🎉

bflad:

ddcprg:
Awesome @mbfrahry and @bflad !!! Massive thanks for taking over :metal:

ddcprg:
Hi @mbfrahry and @bflad the `aws_sagemaker_training_job` resource was in this branch but was not merged together with the `aws_sagemaker_notebook_instance` resource. The PR https://github.com/terraform-providers/terraform-provider-aws/pull/2955 also contains the resource implementation. Are you planning to include this soon?

mbfrahry:
Hey @ddcprg, apologies for not surfacing this earlier. I have the training job piece cleaned up in a branch but I'm not sure it's the best fit for a resource in Terraform. Resources in Terraform are thought of as immutable infrastructure to manage some application . Training jobs go against that by acting like an application. We fire it off, let it run to completion, and then go over the results of what happened. I believe another tool would be better suited for training jobs but I'm open to adding this if you or others feel strongly about it living inside of Terraform.

ddcprg:
Thanks for the insight @mbfrahry that makes sense and if that's the philosophy of the project then I'm fine with keeping it out or leave the decision to the community. Cheers!

PR #7141, opened 2019-01-15
Remove max_frame_rate default value

### LINKED ISSUE #695: [regression] providers/aws: Setting MaxFrameRate on ElasticTranscoderPreset breaks fixed frame rate
_This issue was originally opened by @svanzoest as hashicorp/terraform#13638. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

@catsby hashicorp/terraform#13422  broke hashicorp/terraform#9847 and undid hashicorp/terraform#11340, I can no longer set a preset with a fixed frame rate in 0.9.3, this was the reason the default was removed in the first place.

I now get:
```
* aws_elastictranscoder_preset.proxy: Error creating Elastic Transcoder Preset: ValidationException: When you specify a value for MaxFrameRate, you must also specify 'FrameRate':'Auto'. Either change the value of FrameRate or remove the MaxFrameRate object, and resubmit the request.
    status code: 400, request id: 8cf3908f-207e-11e7-84a3-01d702018b22
```

cretueusebiu:
Fixes #695

This is the same as https://github.com/hashicorp/terraform/pull/11340, but even if it was merged there, it's still not merged here.

* Remove `max_frame_rate` default value and let AWS set it.

### DIFF
--- a/aws/resource_aws_elastic_transcoder_preset.go
+++ b/aws/resource_aws_elastic_transcoder_preset.go
@@ -216,7 +216,6 @@ func resourceAwsElasticTranscoderPreset() *schema.Resource {
 						"max_frame_rate": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Default:  "30",
 							ForceNew: true,
 						},
 						"max_height": {


pabloxco:
I need this! 😩

bowczarek:
I need that as well :( any ETA for merging the change and releasing nie provider plugin?

thaffenden:
Sorry to bump, but any news on this tiny one?

aeschright:
Hi folks, I had a look at [the AWS docs](https://docs.aws.amazon.com/sdk-for-go/api/service/elastictranscoder/#VideoParameters) for this feature and I agree the `max_frame_rate` default doesn't make sense. In order to move forward, this PR needs to update our own docs and also make sure the tests are passing.

I suspect it would be a good idea for us to send a default `frame_rate` of `auto` here since the field is optional and the AWS docs aren't explicit about what happens if it's blank. If anyone wants to look at [the test cases on this resource](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_elastic_transcoder_preset_test.go) and make sure they are realistic, that would be very helpful. Thanks!

PR #7152, opened 2019-01-16
Documentation fix for cpu_credits

stefansundin:
Fix documentation for `cpu_credits` default.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances-unlimited-mode.html#burstable-performance-instances-unlimited-mode-concepts

### DIFF
--- a/website/docs/r/instance.html.markdown
+++ b/website/docs/r/instance.html.markdown
@@ -198,7 +198,7 @@ Credit specification can be applied/modified to the EC2 Instance at any time.
 
 The `credit_specification` block supports the following:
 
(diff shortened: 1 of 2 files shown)


stefansundin:
In case anyone else encounters this.. The documentation claims that you can switch instances from Unlimited and back, but it appears that your account has to be whitelisted for this to be allowed. I am getting this error message when trying to disable or enable unlimited:

> This account cannot run T2 instances with Unlimited enabled. Please contact AWS Support to enable this feature.

It works fine to re-create the instance in the other mode though! Weird!

Also mentioned here: https://forums.aws.amazon.com/thread.jspa?messageID=852621&tstart=0

PR #7181, opened 2019-01-17
Document VPC based ES clusters

tomelliff:
It's necessary to have already created the `elasticsearch.amazonaws.com` service linked role before attempting to create a VPC based ES cluster.

This can either be done separately from the ES cluster creation (useful if you have multiple ES clusters in an AWS account because the service linked role can only be created once) or the `aws_elasticsearch_domain` resource needs to wait for it to be created by using `depends_on`.

See the discussion on https://github.com/terraform-providers/terraform-provider-aws/issues/5218 for more information.

### DIFF
--- a/website/docs/r/elasticsearch_domain.html.markdown
+++ b/website/docs/r/elasticsearch_domain.html.markdown
@@ -108,6 +108,101 @@ resource "aws_elasticsearch_domain" "example" {
   }
 }
 ```
+### VPC based ES
+
+```hcl
+variable "vpc" {}
+
+variable "domain" {
+  default = "tf-test"
+}
+
+data "aws_vpc" "selected" {
+  tags {
+    Name = "${var.vpc}"
+  }
+}
+
+data "aws_subnet_ids" "selected" {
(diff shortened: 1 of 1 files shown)


website/docs/r/elasticsearch_domain.html.markdown
```diff
@@ -108,6 +108,101 @@ resource "aws_elasticsearch_domain" "example" {
...
+    protocol  = "tcp"
+
+    cidr_blocks = [
+      "${data.aws_vpc.selected.cidr_blocks}",
+    ]
+  }
+}
+
+resource "aws_iam_service_linked_role" "es" {
+  aws_service_name = "elasticsearch.amazonaws.com"
```
stantonk: i'm not sure this is correct? i get this error: 

```* aws_iam_service_linked_role.es: Error creating service-linked role with name elasticsearch.amazonaws.com: AccessDenied: elasticsearch.amazonaws.com is not a valid AWS service name (Service Principal)
	status code: 403```

the docs here say it should be `es.amazonaws.com` ?

PR #7183, opened 2019-01-17
Remove ES service linked role creation

tomelliff:
This is not currently working due to the error message being different than expected ('cannot be found' instead of the expected 'Role not found').

Instead of fixing the bug it was decided that a better approach would be to force users to explicitly create the IAM service linked role instead of doing it for them if it didn't exist.
This more closely matches what we do elsewhere and also means we aren't implicitly creating an IAM role despite the user not trying to do that which is particularly dangerous with IAM even if it is just a service linked role.

This is also reasonably necessary because it's going to be an absolute pain to catch this in acceptance tests as it's something that gets implicitly created once per account and then not managed by Terraform after that.

See the discussion on https://github.com/terraform-providers/terraform-provider-aws/issues/5218 for more information.

* Remove implicit IAM service linked role creation when creating a VPC based ES cluster. This is borderline a breaking change as users _may_ be expecting this to always work in a fresh account without the service linked role being present.

### DIFF
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -10,7 +10,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	elasticsearch "github.com/aws/aws-sdk-go/service/elasticsearchservice"
-	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/hashicorp/terraform/helper/customdiff"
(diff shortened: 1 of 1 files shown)


tomelliff:
I've not ran the acceptance tests because none of them cover this functionality so nothing should change here.

It does mean that if the acceptance tests are ran in a fresh account they will fail until the service linked role is created but I'm not sure if we want to put that role creation into the acceptance tests as it can only be created once in the account so will screw with the parallel tests.

Is there a precedent for this with other things that rely on service linked roles?

PR #7271, opened 2019-01-23
tests/resource/aws_iam_group_membership: Temporarily use expanded users references

bflad:
This change is both backwards (0.11) and forwards (0.12) compatible to allow the configuration on both versions. Eventually the temporary workaround will be removed when we switch the provider test configurations to 0.12-only syntax.

Previous output from Terraform 0.12 acceptance testing:

```
--- FAIL: TestAccAWSGroupMembership_paginatedUserList (1.68s)
    testing.go:568: Step 0 error: errors during plan:

        Error: Incorrect attribute value type

          on /opt/teamcity-agent/temp/buildTmp/tf-test240570565/main.tf line 8:
          (source code not available)

        Inappropriate value for attribute "users": element 0: string required.
```

Output from Terraform 0.12 acceptance testing:

```
--- PASS: TestAccAWSGroupMembership_paginatedUserList (61.05s)
```


aws/resource_aws_iam_group_membership_test.go
```diff
@@ -232,8 +232,110 @@ resource "aws_iam_group" "group" {
 
 resource "aws_iam_group_membership" "team" {
 	name = "%s"
-	users = ["${aws_iam_user.user.*.name}"]
 	group = "${aws_iam_group.group.name}"
+	users = [
+		"${aws_iam_user.user.*.name[0]}",
```
tombuildsstuff: I feel like this /may/ want a TODO?

bflad: One does not simply forget `"${aws_iam_user.user.*.name[100]}",` 😂 I'll add a comment to both the huge expanded lists post-merge explaining them and hopefully they are caught during 0.12 configuration upgrades.

PR #7272, opened 2019-01-23
tests/resource/aws_iam_policy_attachment: Temporarily use expanded users references

bflad:
This change is both backwards (0.11) and forwards (0.12) compatible to allow the configuration on both versions. Eventually the temporary workaround will be removed when we switch the provider test configurations to 0.12-only syntax.

Previous output from Terraform 0.12 acceptance testing:

```
--- FAIL: TestAccAWSIAMPolicyAttachment_paginatedEntities (2.16s)
    testing.go:568: Step 0 error: errors during plan:

        Error: Incorrect attribute value type

          on /opt/teamcity-agent/temp/buildTmp/tf-test032782444/main.tf line 26:
          (source code not available)

        Inappropriate value for attribute "users": element 0: string required.
```

Output from Terraform 0.12 acceptance testing:

```
--- PASS: TestAccAWSIAMPolicyAttachment_paginatedEntities (57.78s)
```


aws/resource_aws_iam_policy_attachment_test.go
```diff
@@ -352,7 +352,109 @@ EOF
 }
 resource "aws_iam_policy_attachment" "test-paginated-attach" {
 	name = "%s"
-	users = ["${aws_iam_user.user.*.name}"]
 	policy_arn = "${aws_iam_policy.policy.arn}"
+	users = [
+		"${aws_iam_user.user.*.name[0]}",
```
nywilken: Is the temporary workaround you speak of the use of the splat operator? Reading the v0.12 preview I found https://www.hashicorp.com/blog/terraform-0-12-generalized-splat-operator#full-splat-operator so just want to confirm for learning purposes.

bflad: The temporary workaround here is explicitly declaring each element in the list as individual string references to maintain compatibility for Terraform 0.11 vendoring (in the likely case we will release more 1.X provider releases off master) and future compatibility for Terraform 0.12 vendoring. After some brief trial and error there does not seem to be an easy/friendly backwards and forwards compatible syntax that allows passing the list of strings directly even with various forms of using the splat operator. While there may actually be some decent syntax to accomplish this for longer-term real world usage (e.g. in public Terraform modules), these updates are only temporarily required in this acceptance testing, so opting for the first reasonable solution seems okay.

Once the Terraform AWS provider is out of its 2.0 beta period (to coincide with Terraform 0.12's beta period) and we are fairly confident we will not need to continue 1.X provider releases, we can update these acceptance test configurations to use the newer syntax and go back to a simplified configuration.

For operators upgrading to Terraform 0.12, they will theoretically be able to use the `terraform 0.12upgrade` command to automatically have their syntax simplified in this case from the existing:

```hcl
users = ["${aws_iam_user.user.*.name}"]
```

To the newer:

```hcl
users = aws_iam_user.user[*].name
```

Utilizing the full splat operator and bypassing errors seen trying to support both Terraform 0.11's configuration parser and Terraform 0.12's configuration parser.

nywilken: Thanks for explaining.

PR #7313, opened 2019-01-24
EKS example: select node AMI specific to cluster version

alexsomesan:
Improves consistency of EKS example cluster nodes.

* Makes use of the configured cluster's version in filtering the AMIs for the node. This avoids old / incompatible AMIs being used, even if they were published more recently than compatible ones.

Doesn't change any provider code.


alexsomesan:
GSG updated.
Let me know if it needs rephrasing.

PR #7349, opened 2019-01-27
Added public_dns and private_dns to aws_eip

### LINKED ISSUE #1149: aws_eip should have a Public DNS attribute
Elastic IPs are assigned a Public DNS name. That name should be accessible in Terraform as an attribute of `aws_eip` objects. The final bullet of [this](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/elastic-ip-addresses-eip.html#eip-basics) Elastic IP doc reads:

> We resolve a public DNS hostname to the public IPv4 address or the Elastic IP address of the instance outside the network of the instance, and to the private IPv4 address of the instance from within the network of the instance.

Using the [DNS name](https://forums.aws.amazon.com/thread.jspa?messageID=99701#99701) is the recommended method to use an eip to reference internal resources.

### Terraform Version
Terraform v0.9.11

### Affected Resource(s)
- aws_eip

### Important Factoids
Although the `describe-addresses` API call does not specifically contain a field with the public DNS, the AWS console does, and the IP does not need to be associated to anything first. I'm not sure if there is another API call that would tell you the DNS name or if the console is just templating out the `ip-0-0-0-0.compute-1.amazonaws.com` format.

Editing to add an additional AWS forum link that's dated but specifies that it is a derivation of the IP and is not available from an API: https://forums.aws.amazon.com/message.jspa?messageID=84856

hamstah:
Fixes #1149

* Added public_dns and private_dns for aws_eip data and resource

Looking at the discussion in https://github.com/terraform-providers/terraform-provider-aws/pull/3731 this seems like a more simple implementation than using the reverse lookup.

I've created/looked up IPs in different regions (us-east-1, us-west-1, us-east-2, eu-west-1) to double check they follow the documentation.

Could you please advise on the best way to add unit tests for the new attributes? 

Thanks

```
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSAvailabilityZones -timeout 120m
=== PAUSE TestAccAWSAvailabilityZones_basic
=== PAUSE TestAccAWSAvailabilityZones_stateFilter
=== CONT  TestAccAWSAvailabilityZones_basic
=== CONT  TestAccAWSAvailabilityZones_stateFilter
--- PASS: TestAccAWSAvailabilityZones_stateFilter (25.13s)
--- PASS: TestAccAWSAvailabilityZones_basic (25.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	25.153s
...
```

### DIFF
--- a/aws/data_source_aws_eip.go
+++ b/aws/data_source_aws_eip.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -44,9 +45,17 @@ func dataSourceAwsEip() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"private_dns": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"public_ip": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
+			},
+			"public_dns": {
+				Type:     schema.TypeString,
 				Computed: true,
 			},
 			"public_ipv4_pool": {
@@ -86,7 +95,6 @@ func dataSourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
 		req.Filters = nil
 	}
 
-	log.Printf("[DEBUG] Reading EIP: %s", req)
 	resp, err := conn.DescribeAddresses(req)
 	if err != nil {
 		return fmt.Errorf("error describing EC2 Address: %s", err)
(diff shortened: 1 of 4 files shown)


aws/resource_aws_eip.go
```diff
@@ -70,11 +70,23 @@ func resourceAwsEip() *schema.Resource {
 				Computed: true,
 			},
 
+			"public_dns": {
+				Type:     schema.TypeString,
+				Optional: true,
```
bflad: `Optional: true` should be removed here as operators cannot specify this argument.

aws/resource_aws_eip.go
```diff
@@ -70,11 +70,23 @@ func resourceAwsEip() *schema.Resource {
...
+			},
+
 			"private_ip": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
+			"private_dns": {
+				Type:     schema.TypeString,
+				Optional: true,
```
bflad: `Optional: true` should be removed here as operators cannot specify this argument.

aws/data_source_aws_eip.go
```diff
@@ -112,8 +120,30 @@ func dataSourceAwsEipRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("instance_id", eip.InstanceId)
 	d.Set("network_interface_id", eip.NetworkInterfaceId)
 	d.Set("network_interface_owner_id", eip.NetworkInterfaceOwnerId)
+
+	region := *conn.Config.Region
```
bflad: Minor nit: We tend to access the provider region across the codebase via `region := meta.(*AWSClient).region` instead of via the client.

bflad:
Also FYI from your acceptance testing output above, it looks like you ran the `aws_availability_zones` data source acceptance tests. If you filter on `TestAccAWSEIP_` and `TestAccDataSourceAwsEip_` it will run the `aws_eip` resource and data source acceptance tests.

bflad:
Hi again @hamstah 👋 Since we didn't hear back from you regarding this, we went ahead and implemented the feedback in a followup commit (ac09b676dcafbdbc9f3bcc879b341bc17e420662). Thanks for your contribution. 🚀 

```
--- PASS: TestAccAWSEIP_disappears (6.81s)
--- PASS: TestAccAWSEIP_PublicIpv4Pool_default (8.29s)
--- PASS: TestAccAWSEIP_basic (8.58s)
--- PASS: TestAccDataSourceAwsEip_Id (9.73s)
--- PASS: TestAccDataSourceAwsEip_PublicIP_VPC (10.02s)
--- PASS: TestAccDataSourceAwsEip_Filter (10.05s)
--- PASS: TestAccDataSourceAwsEip_Tags (11.68s)
--- PASS: TestAccDataSourceAwsEip_PublicIP_EC2Classic (15.80s)
--- PASS: TestAccAWSEIP_tags (16.97s)
--- PASS: TestAccAWSEIP_twoEIPsOneNetworkInterface (24.87s)
--- PASS: TestAccAWSEIP_network_interface (25.61s)
--- PASS: TestAccDataSourceAwsEip_NetworkInterface (27.86s)
--- PASS: TestAccAWSEIP_importVpc (28.10s)
--- PASS: TestAccAWSEIP_importEc2Classic (121.35s)
--- PASS: TestAccAWSEIP_associated_user_private_ip (138.09s)
--- PASS: TestAccDataSourceAwsEip_Instance (192.65s)
--- PASS: TestAccAWSEIP_instance (281.95s)
```

bflad:
This has been released in [version 2.2.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#220-march-15-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #7207, opened 2019-01-18
r/aws_backup_vault: Adding resource to manage AWS Backup vaults

slapula:
Reference #7166 

Changes proposed in this pull request:

* Adding resource to manage AWS Backup vaults: `aws_backup_vault`

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccAwsBackupVault_'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupVault_ -timeout 120m
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
=== RUN   TestAccAwsBackupVault_basic
=== PAUSE TestAccAwsBackupVault_basic
=== RUN   TestAccAwsBackupVault_withKmsKey
=== PAUSE TestAccAwsBackupVault_withKmsKey
=== RUN   TestAccAwsBackupVault_withTags
=== PAUSE TestAccAwsBackupVault_withTags
=== CONT  TestAccAwsBackupVault_basic
=== CONT  TestAccAwsBackupVault_withTags
=== CONT  TestAccAwsBackupVault_withKmsKey
--- PASS: TestAccAwsBackupVault_withTags (8.01s)
--- PASS: TestAccAwsBackupVault_basic (8.10s)
--- PASS: TestAccAwsBackupVault_withKmsKey (33.02s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	33.050s
```

PR #7350, opened 2019-01-28
r/aws_backup_plan: Adding resource for managing AWS Backup plans

### LINKED ISSUE #7166: Support for AWS Backup
### Community Note

### Description

AWS just launched a new service called [AWS Backup](https://aws.amazon.com/blogs/aws/aws-backup-automate-and-centrally-manage-your-backups/), which replaces and significantly extends the AWS Data Lifecycle Manager; by providing finer-grained policies (e.g. cron schedule), cold storage, adding the ability to backup not only EBS volumes but various other services (e.g. RDS databases, DynamoDB tables, EFS file systems, and Storage Gateway volumes).

The community could greatly profit from a Terraform implementation of this resource.

slapula:
Reference: #7166

* Adding resource to manage AWS Backup plans: `aws_backup_plan`

```
$  make testacc TESTARGS='-run=TestAccAwsBackupPlan_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupPlan_basic -timeout 120m
=== RUN   TestAccAwsBackupPlan_basic
=== PAUSE TestAccAwsBackupPlan_basic
=== CONT  TestAccAwsBackupPlan_basic
--- PASS: TestAccAwsBackupPlan_basic (11.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	11.637s

$ make testacc TESTARGS='-run=TestAccAwsBackupPlan_withRules'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupPlan_withRules -timeout 120m
=== RUN   TestAccAwsBackupPlan_withRules
=== PAUSE TestAccAwsBackupPlan_withRules
=== CONT  TestAccAwsBackupPlan_withRules
--- PASS: TestAccAwsBackupPlan_withRules (11.44s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	11.461s

$ make testacc TESTARGS='-run=TestAccAwsBackupPlan_withRuleRemove'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupPlan_withRuleRemove -timeout 120m
=== RUN   TestAccAwsBackupPlan_withRuleRemove
=== PAUSE TestAccAwsBackupPlan_withRuleRemove
=== CONT  TestAccAwsBackupPlan_withRuleRemove
--- PASS: TestAccAwsBackupPlan_withRuleRemove (19.56s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	19.574s

$ make testacc TESTARGS='-run=TestAccAwsBackupPlan_withRuleAdd'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupPlan_withRuleAdd -timeout 120m
=== RUN   TestAccAwsBackupPlan_withRuleAdd
=== PAUSE TestAccAwsBackupPlan_withRuleAdd
=== CONT  TestAccAwsBackupPlan_withRuleAdd
--- PASS: TestAccAwsBackupPlan_withRuleAdd (19.20s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	19.225s
```


slapula:
This is still a WIP.  This PR's test rely on #7207.  I also need to write up a few more tests before this is ready for final review.

slapula:
I think this ready for review now

bflad:
Can you please rebase this with master please? Thanks so much.

slapula:
@bflad Done

denniswebb:
@bflad Really need this resource added. Can it get some ❤️ ?

mooseracer:
I get a type-related panic when trying to use Lifecycle.

`resource "aws_backup_plan" "my_backup_plan" {
  name = "my_backup_plan"

  rule {
    rule_name         = "my_backup_plan_rule"
    target_vault_name = "${aws_backup_vault.my_vault.name}"
    schedule          = "cron(0 10 * * ? *)"

    lifecycle = {
      cold_storage_after = 90
      delete_after       = 120
    }
  }
}`

Passes a terraform plan, but apply crashes:
> panic: interface conversion: interface {} is string, not int

> 2019-03-12T10:00:53.081+1300 [DEBUG] plugin.terraform-provider-aws.exe:         C:/Users/username/Documents/tf_aws_backup/terraform-provider-aws-backup/aws/resource_aws_backup_plan.go:267 +0xc6c

slapula:
@mooseracer thanks for catching that! I'm looking into that right now. Should be an easy fix.

slapula:
@mooseracer I think I fixed your issue.  Could you try that again?

mooseracer:
@slapula Sorry, tried it again but it's throwing the same crash errors.

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+							Optional: true,
+							Default:  60,
+						},
+						"completion_window": {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  180,
+						},
+						"lifecycle": {
+							Type:     schema.TypeMap,
```
bflad: The Terraform Provider SDK does not currently support TypeMap configuration blocks. Did you mean `TypeList` and `MaxItems: 1`?

```suggestion
							Type:     schema.TypeList,
							MaxItems: 1,
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+		BackupPlan: plan,
+	}
+
+	if v, ok := d.GetOk("tags"); ok {
+		input.BackupPlanTags = v.(map[string]*string)
+	}
+
+	resp, err := conn.CreateBackupPlan(input)
+	if err != nil {
+		return err
```
bflad: We should return error context for operators and code maintainers:

```suggestion
		return fmt.Errorf("error creating Backup Plan: %s", err)
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+
+func resourceAwsBackupPlanRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).backupconn
+
+	input := &backup.GetBackupPlanInput{
+		BackupPlanId: aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetBackupPlan(input)
+	if err != nil {
```
bflad: Prior to returning any read error to the operator, we should check if the Backup Plan has been removed and prompt Terraform to recreate the resource automatically instead of the error.

```suggestion
	if isAWSErr(err, backup.ErrCodeResourceNotFoundException, "") {
		log.Printf("[WARN] Backup Plan (%s) not found, removing from state", d.Id())
		d.SetId("")
		return nil
	}

	if err != nil {
```

This can be acceptance tested with something like the following:

```go
func TestAccAwsBackupPlan_disappears(t *testing.T) {
	var backupPlan backup.GetBackupPlanOutput
	rInt := acctest.RandInt()
		CheckDestroy: testAccCheckAwsBackupPlanDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccBackupPlanConfig(rInt),
					testAccCheckAwsBackupPlanExists("aws_backup_plan.test", &backupPlan),
					testAccCheckAwsBackupPlanDisappears(&backupPlan),
				),
			},
		},
	})
}

func testAccCheckAwsBackupPlanDisappears(backupPlan *backup.GetBackupPlanOutput) resource.TestCheckFunc {
		conn := testAccProvider.Meta().(*AWSClient).backupconn

		input := &backup.DeletePlanInput{
			BackupPlanId: backupPlan.BackupPlanId,
		}

		_, err := conn.DeleteBackupPlan(input)

		return err
	}
}
```

I'll note below the required changes for the `Exists` function since it doesn't appear to be implemented.

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+		return err
```
bflad: We should return context for operators and code maintainers:

```suggestion
		return fmt.Errorf("error reading Backup Plan (%s): %s", d.Id(), err)
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	rule := &schema.Set{F: resourceAwsPlanRuleHash}
+
+	for _, r := range resp.BackupPlan.Rules {
+		m := make(map[string]interface{})
+
+		if r.CompletionWindowMinutes != nil {
```
bflad: Many of these `nil` checks can be skipped by using the AWS Go SDK provided conversion functions, e.g.

```go
m["completion_window"] = aws.Int64Value(r.CompletionWindowMinutes)
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+			m["schedule"] = *r.ScheduleExpression
+		}
+		if r.StartWindowMinutes != nil {
+			m["start_window"] = *r.StartWindowMinutes
+		}
+		m["target_vault_name"] = *r.TargetBackupVaultName
+
+		rule.Add(m)
+	}
+	d.Set("rule", rule)
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```suggestion
	if err := d.Set("rule", rule); err != nil {
		return fmt.Errorf("error setting rule: %s", err)
	}
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+	plan.Rules = rules
+
+	input := &backup.UpdateBackupPlanInput{
+		BackupPlanId: aws.String(d.Id()),
+		BackupPlan:   plan,
+	}
+
+	_, err := conn.UpdateBackupPlan(input)
+	if err != nil {
+		return err
```
bflad: We should return context here for operators and code maintainers:

```suggestion
		return fmt.Errorf("error updating Backup Plan (%s): %s", d.Id(), err)
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+
+func resourceAwsBackupPlanDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).backupconn
+
+	input := &backup.DeleteBackupPlanInput{
+		BackupPlanId: aws.String(d.Id()),
+	}
+
+	_, err := conn.DeleteBackupPlan(input)
+	if err != nil {
```
bflad: Similar to the `Read` change noted above, prior to returning any delete error to the operator, we should check if the Backup Plan has been removed and instead allow the function to return without error.

```suggestion
	if isAWSErr(err, backup.ErrCodeResourceNotFoundException, "") {
		return nil
	}

	if err != nil {
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+		return err
```
bflad: We should return context here for operators and code maintainers:

```suggestion
		return fmt.Errorf("error deleting Backup Plan (%s): %s", d.Id(), err)
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	return nil
+}
+
+func gatherPlanRules(d *schema.ResourceData) []*backup.RuleInput {
```
bflad: A few items about this function signature:
* Since this codebase uses a very flat Go package structure, the function name should include the service name
* We tend to refer to these as "expand" functions throughout the codebase
* As `d.Get("rule").(*schema.Set).List()` is the only usage of `d *schema.ResourceData`, the input parameter should be reduced to only its necessary usage

```suggestion
func expandBackupPlanRules(l []interface{}) []*backup.RuleInput {
```

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,311 @@
...
+	rules := []*backup.RuleInput{}
+	planRules := d.Get("rule").(*schema.Set).List()
+
+	for _, i := range planRules {
+		item := i.(map[string]interface{})
+		lifecycle := i.(map[string]interface{})["lifecycle"].(map[string]interface{})
```
bflad: To prevent panics, `["lifecycle"]` should be `nil` checked before type assertion since it is `Optional: true`

aws/resource_aws_backup_plan_test.go
```diff
@@ -0,0 +1,184 @@
...
+			if *resp.BackupPlanId == rs.Primary.ID {
+				return fmt.Errorf("Plane '%s' was not deleted properly", rs.Primary.ID)
+			}
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckAwsBackupPlanExists(name string) resource.TestCheckFunc {
```
bflad: The `Exists` functions in Terraform Provider acceptance testing are designed to check the remote API to ensure the requested resource was actually created (and preferably return a SDK object for other potential acceptance testing usage like we need to create a `Disappears` function). More information can be found in the [Extending Terraform documentation](https://www.terraform.io/docs/extend/testing/acceptance-tests/teststep.html#custom-check-functions).

```go
func testAccCheckAwsBackupPlanExists(resourceName string, backupPlan *backup.GetBackupPlanOutput) resource.TestCheckFunc {
		rs, ok := s.RootModule().Resources[resourceName]

		if !ok {
			return fmt.Errorf("Not found: %s", resourceName)
		}

		if rs.Primary.ID == "" {
			return fmt.Errorf("Resource ID is not set")
		}

		conn := testAccProvider.Meta().(*AWSClient).backupconn

		input := &backup.GetBackupPlanInput{
			BackupPlanId: aws.String(rs.Primary.ID),
		}

		output, err := conn.GetBackupPlan(input)

		if err != nil {
			return err
		}

		*backupPlan = *output

		return nil
	}
}
```

aws/resource_aws_backup_plan_test.go
```diff
@@ -0,0 +1,184 @@
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsBackupPlanDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccBackupPlanConfig(rInt),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsBackupPlanExists("aws_backup_plan.test"),
+					resource.TestCheckResourceAttrSet("aws_backup_plan.test", "arn"),
```
bflad: We have a helper function to better check the Terraform state value here:

```suggestion
					testAccMatchResourceAttrRegionalARN("aws_backup_plan.test", "arn", "backup", regexp.MustCompile(`plan:.+`)),
```

The `arn` and `version` attribute checks in other acceptance test functions below can be removed as they provide no additional value.

aws/resource_aws_backup_plan_test.go
```diff
@@ -0,0 +1,184 @@
...
+resource "aws_backup_plan" "test" {
+	name = "tf_acc_test_backup_plan_%d"
+
+	rule {
+		rule_name 			= "tf_acc_test_backup_rule_%d"
+		target_vault_name 	= "${aws_backup_vault.test.name}"
+		schedule			= "cron(0 12 * * ? *)"
+	}
+}
+`, randInt, randInt, randInt)
```
bflad: You can avoid `fmt` argument sprawl by using indexed formatter references, e.g.

```go
fmt.Sprintf(`
thing1 = %[1]d
thing2 = %[1]d
`, randInt)
```

website/docs/r/backup_plan.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+Provides an AWS Backup plan resource.
+
+## Example Usage
+
+```hcl
+resource "aws_backup_plan" "example" {
+	name = "tf_example_backup_plan"
+
+	rule {
+		rule_name 			= "tf_example_backup_rule"
```
bflad: The equals sign alignment of the example here and some of the testing appears to be an issue. Mixed tabs and spaces?

slapula:
@bflad Thanks for the review!  I believe I've addressed all of your comments:  
```
$ make testacc TESTARGS='-run=TestAccAwsBackupPlan_'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupPlan_ -timeout 120m
=== RUN   TestAccAwsBackupPlan_basic
=== PAUSE TestAccAwsBackupPlan_basic
=== RUN   TestAccAwsBackupPlan_withRules
=== PAUSE TestAccAwsBackupPlan_withRules
=== RUN   TestAccAwsBackupPlan_withRuleRemove
=== PAUSE TestAccAwsBackupPlan_withRuleRemove
=== RUN   TestAccAwsBackupPlan_withRuleAdd
=== PAUSE TestAccAwsBackupPlan_withRuleAdd
=== RUN   TestAccAwsBackupPlan_disappears
=== PAUSE TestAccAwsBackupPlan_disappears
=== CONT  TestAccAwsBackupPlan_basic
=== CONT  TestAccAwsBackupPlan_withRuleAdd
=== CONT  TestAccAwsBackupPlan_disappears
=== CONT  TestAccAwsBackupPlan_withRuleRemove
=== CONT  TestAccAwsBackupPlan_withRules
--- PASS: TestAccAwsBackupPlan_disappears (14.37s)
--- PASS: TestAccAwsBackupPlan_withRules (15.00s)
--- PASS: TestAccAwsBackupPlan_basic (15.14s)
--- PASS: TestAccAwsBackupPlan_withRuleRemove (26.08s)
--- PASS: TestAccAwsBackupPlan_withRuleAdd (27.80s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	27.823s
```

website/docs/r/backup_plan.html.markdown
slapula: Yeah, this is due to me having different tab settings for Go and Terraform.

aws/resource_aws_backup_plan.go
slapula: Is this something that has been depreciated?  I seem to see it used elsewhere in this provider.

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,305 @@
...
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  60,
+						},
+						"completion_window": {
+							Type:     schema.TypeInt,
+							Optional: true,
+							Default:  180,
+						},
+						"lifecycle": {
```
bflad: Can you please add an acceptance test that exercises this configuration block? Thanks! I think line 156 should be a something like `m["lifecycle"] = []interface{}{l}`

aws/resource_aws_backup_plan.go
```diff
@@ -0,0 +1,305 @@
...
+			},
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"version": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"tags": tagsSchema(),
```
bflad: Can you please add covering acceptance testing for tags (add, update, remove)? I don't think it is getting handed during read or update currently.

aws/resource_aws_backup_plan.go
slapula: One question that's been nagging me as I've been working on these resources... How do I test for attributes in a list where the identifier appears to be a random number instead of an ordered number? For example:
```
  rule.# = 1
  rule.160206105.completion_window = 180
  rule.160206105.lifecycle.# = 1
  rule.160206105.lifecycle.0.cold_storage_after = 30
  rule.160206105.lifecycle.0.delete_after = 160
  rule.160206105.recovery_point_tags.% = 0
  rule.160206105.rule_name = tf_acc_test_backup_rule_7563947093289689272
  rule.160206105.schedule = cron(0 12 * * ? *)
  rule.160206105.start_window = 60
  rule.160206105.target_vault_name = tf_acc_test_backup_vault_7563947093289689272
```
How do I test the attributes of `rule` when `160206105` changes with each test run?

aws/resource_aws_backup_plan.go
bflad: A `TypeMap` attribute can only currently[1] be used with a single primative type element (`TypeString`, `TypeInt`, or `TypeFloat`). Any `Elem` usage outside of that with `TypeMap` is an implementation bug and/or a quirk that just happens to work in the current Terraform Provider SDK and the usage may not be supported in the future.

In Terraform 0.12 core, we can start supporting a `TypeMap` configuration block syntax, such as:

```hcl
my_configuration_block "key1" {
  child_argument = ""
}
```

However its implementation is still in its design phases and is not present in the Terraform Provider SDK. See also: https://github.com/hashicorp/terraform/issues/19749#issuecomment-450256714

[1]: The underlying type system in Terraform 0.12 removes this restriction. When the Terraform Provider SDK supports dynamic elements in a future update and we upgrade to that version of the Terraform Provider SDK (requires removing Terraform 0.11 support in a later major version of the Terraform AWS Provider) we can support multiple value types in `TypeMap`.

slapula: Oh wow, TIL. Thanks for the explanation! 😁

aws/resource_aws_backup_plan.go
bflad: Unfortunately in these cases, the randomization gets complicated for the acceptance testing. You'll probably want to make the rule names static (unless the randomization there actually matters). While it is technically possible you could try to calculate the hash of the `TypeSet`, once you're in this situation it is generally easier to take the resultant API object returned by the `Exists` `TestCheckFunc` and check the values within it directly, e.g.

```go
func testAccCheckBackupPlanRules(plan *backup.GetBackupPlanOutput, expectedRules []*backup.Rule) resource.TestCheckFunc {
		if a, e := len(plan.Plan.Rules), len(expectedRules); a != e {
			return fmt.Errorf("expected %d Backup Plan Rules, got: %d", e, a)
		}

		for _, expectedRule := range expectedRules {
			for _, rule := range plan.Plan.Rules {
				if aws.StringValue(rule.RuleName) != aws.StringValue(expectedRule.RuleName) {
					continue
				}

				ruleName := aws.StringValue(rule.RuleName)

				if a, e := aws.StringValue(rule.Schedule), aws.StringValue(expectedRule.Schedule); a != e {
					return fmt.Errorf("expected Backup Plan Rule (%s) schedule to match (%s), got: %s", ruleName, e, a)
				}

				// Check other various attributes as appropriate

				break
			}
		}

		return nil
	}
}

// In a TestStep, filling in rule details as necessary
testAccCheckAwsBackupPlanExists("aws_backup_plan.test", &plan),
testAccCheckBackupPlanRules(&plan, []*backup.Rule{
  {
    RuleName: aws.String("test"),
    Schedule: aws.String("cron(0 12 * * ? *)"),
  },
})
```

As an added assurance to the above, acceptance tests that implement `TestStep`s with `ImportStateVerify: true` will automatically ensure that the resource read function is correctly able to re-read the API object and generate the same Terraform state.

slapula: Ah ok, that makes sense.  There really isn't a requirement for the plan to be unique other than to help me avoid running into dangling resources while testing.  For this test, I'm going to opt to not make it use the random `int` so I can reliably test for the existence of the lifecycle.

slapula:
@bflad I believe I've addressed your last round of comments.  I've implemented the acceptance test for lifecycle blocks.  I've also added the missing read/update functionality for `tags` as well as adding the related acceptance tests for that too.  Let me know if there's anything else I need to look at.

mooseracer:
Thanks for all the work on this! I'd just like to report a logical bug with
the Lifecycle -- it should be able to accept values of "delete_after" when
"cold_storage_after" is left unspecified. Right now it's always passing
"cold_storage_after" to the API and it gets validated against their logic:
lifecycle = {
delete_after = 15
}

Error in rule backup-plan-rule : Invalid lifecycle. DeleteAfterDays cannot
be less than 90 days apart from MoveToColdStorageAfterDays
status code: 400

This represents a simple scenario where you don't want to transition to
cold storage and only keep X days of snapshots. It's supported by the API;
you can set it in the GUI and the following JSON is accepted by AWS CLI:
{
"BackupPlan": {
"BackupPlanName": "myplan",
"Rules": [
{
"RuleName": "myrule",
"TargetBackupVaultName": "Default",
"ScheduleExpression": "cron(0 10 * * ? *)",
"Lifecycle": {
"DeleteAfterDays": 15
}
}
]
}
}

slapula:
@mooseracer Can you submit this as an issue?  I noticed some weirdness with these API calls when I was working on this and if I remember correctly AWS wasn't letting me submit requests with any empty or zero-valued parameters.  As you can see, I made them both required to work around the issue. I'll need to take another look to confirm all of this and think of a better solution.

bflad:
This has been released in [version 2.3.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#230-march-21-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #7382, opened 2019-01-30
 r/aws_backup_selection: Adding resource to manage selections for AWS Backup plans

### LINKED ISSUE #7166: Support for AWS Backup

slapula:
Fixes #7166

* Adding resource to manage AWS Backup plans: `aws_backup_selection`

```
$ make testacc TESTARGS='-run=TestAccAwsBackupSelection_'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupSelection_ -timeout 120m
=== RUN   TestAccAwsBackupSelection_basic
=== PAUSE TestAccAwsBackupSelection_basic
=== RUN   TestAccAwsBackupSelection_withTags
=== PAUSE TestAccAwsBackupSelection_withTags
=== RUN   TestAccAwsBackupSelection_withResources
=== PAUSE TestAccAwsBackupSelection_withResources
=== CONT  TestAccAwsBackupSelection_basic
=== CONT  TestAccAwsBackupSelection_withResources
=== CONT  TestAccAwsBackupSelection_withTags
--- PASS: TestAccAwsBackupSelection_withTags (14.03s)
--- PASS: TestAccAwsBackupSelection_withResources (14.40s)
--- PASS: TestAccAwsBackupSelection_basic (14.41s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	14.430s
```


aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+func resourceAwsBackupSelection() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsBackupSelectionCreate,
+		Read:   resourceAwsBackupSelectionRead,
+		Delete: resourceAwsBackupSelectionDelete,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
```
bflad: We can validate this argument via:

```go
ValidateFunc: validation.All(
	validation.StringLenBetween(1, 50),
	validation.StringMatch(regexp.MustCompile(`^[a-zA-Z0-9\-\_\.]+$`), "must contain only alphanumeric, hyphen, underscore, and period characters"),
),
```

Reference: https://docs.aws.amazon.com/aws-backup/latest/devguide/API_BackupSelection.html

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"plan_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"iam_role": {
```
bflad: This attribute should be named `iam_role_arn` to match the API and denote what value its expecting. It can also be validated via:

```go
ValidateFunc: validateArn,
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"value": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+				Set: resourceAwsConditionTagHash,
```
bflad: We should simplify this attribute by removing the custom `Set` function and allowing Terraform to use its default.

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+				Required: true,
+				ForceNew: true,
+			},
+			"tag": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"type": {
```
bflad: We can validate this argument via:

```go
	backup.ConditionTypeStringequals,
}, false),
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+		},
+	}
+}
+
+func resourceAwsBackupSelectionCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).backupconn
+
+	selection := &backup.Selection{}
+
+	selection.SelectionName = aws.String(d.Get("name").(string))
```
bflad: Nit: these could be set in the struct declaration above, e.g.

```go
selection := &backup.Selection{
	IamRoleArn:    aws.String(d.Get("iam_role_arn").(string)),
	ListOfTags:    expandBackupConditionTags(d.Get("tag").(*schema.Set)),
	Resources:     expandStringList(d.Get("resources").([]interface{})),
	SelectionName: aws.String(d.Get("name").(string)),
}
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+func resourceAwsBackupSelectionRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).backupconn
+
+	input := &backup.GetBackupSelectionInput{
+		BackupPlanId: aws.String(d.Get("plan_id").(string)),
+		SelectionId:  aws.String(d.Id()),
+	}
+
+	resp, err := conn.GetBackupSelection(input)
+	if err != nil {
```
bflad: Instead of returning an error to the operator in the case the resource has been removed outside of Terraform, we should instead trigger resource recreation, e.g.

```suggestion

	if isAWSErr(err, backup.ErrCodeResourceNotFoundException, "") {
		log.Printf("[WARN] Backup Selection (%s) not found, removing from state", d.Id())
		d.SetId("")
		return nil
	}

	if err != nil {
```

This can be acceptance tested by adding a "disappears" test, e.g.

```go
func TestAccAwsBackupSelection_disappears(t *testing.T) {
	var selection1 backup.GetBackupSelectionOutput
	rInt := acctest.RandInt()
		CheckDestroy: testAccCheckAwsBackupSelectionDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccBackupSelectionConfigBasic(rInt),
					testAccCheckAwsBackupSelectionExists("aws_backup_selection.test", &selection1),
					testAccCheckAwsBackupSelectionDisappears(&selection1),
				),
				ExpectNonEmptyPlan: true,
			},
		},
	})
}

func testAccCheckAwsBackupSelectionDisappears(selection *backup.GetBackupSelectionOutput) resource.TestCheckFunc {
		conn := testAccProvider.Meta().(*AWSClient).backupconn

		input := &backup.DeleteBackupSelectionInput{
			BackupPlanId: selection.BackupPlanId,
			SelectionId:  selection.SelectionId,
		}

		_, err := conn.DeleteBackupSelection(input)

		return err
	}
}
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+		return fmt.Errorf("error reading Backup Selection: %s", err)
+	}
+
+	d.Set("plan_id", resp.BackupPlanId)
+	d.Set("name", resp.BackupSelection.SelectionName)
+	d.Set("iam_role", resp.BackupSelection.IamRoleArn)
+
+	if resp.BackupSelection.ListOfTags != nil {
+		tag := &schema.Set{F: resourceAwsConditionTagHash}
```
bflad: We can use a `[]interface{}` variable here instead with `append()` below 👍

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+
+		for _, r := range resp.BackupSelection.ListOfTags {
+			m := make(map[string]interface{})
+
+			m["type"] = *r.ConditionType
```
bflad: To prevent potential panics here and below, we should use the [AWS Go SDK conversion functions](https://docs.aws.amazon.com/sdk-for-go/api/aws/), e.g.

```suggestion
			m["type"] = aws.StringValue(r.ConditionType)
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+			m["key"] = *r.ConditionKey
+			m["value"] = *r.ConditionValue
+
+			tag.Add(m)
+		}
+
+		d.Set("tag", tag)
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```suggestion
		if err := d.Set("tag", tag); err != nil {
			return fmt.Errorf("error setting tag: %s", err)
		}
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if resp.BackupSelection.Resources != nil {
+		d.Set("resources", resp.BackupSelection.Resources)
```
bflad: Same here with regards to error checking:

```suggestion
		if err := d.Set("resources", resp.BackupSelection.Resources); err != nil {
			return fmt.Errorf("error setting resources: %s", err)
		}
```

I believe in this case you may see that it needs `aws.StringValueSlice()` or `flattenStringList()` called on it first.

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,182 @@
...
+
+	_, err := conn.DeleteBackupSelection(input)
+	if err != nil {
+		return fmt.Errorf("error deleting Backup Selection: %s", err)
+	}
+
+	return nil
+}
+
+func gatherConditionTags(d *schema.ResourceData) []*backup.Condition {
```
bflad: For consistency across the codebase and since we have a flat Go package structure, the naming of the function should be `expandBackupConditionTags`. We also should reduce the parameter to its necessary information.

```suggestion
func expandBackupConditionTags(tagSet *schema.Set) []*backup.Condition {
```

aws/resource_aws_backup_selection_test.go
```diff
@@ -0,0 +1,189 @@
...
+			if *resp.SelectionId == rs.Primary.ID {
+				return fmt.Errorf("Selection '%s' was not deleted properly", rs.Primary.ID)
+			}
+		}
+	}
+
+	return nil
+}
+
+func testAccCheckAwsBackupSelectionExists(name string) resource.TestCheckFunc {
```
bflad: The acceptance testing `Exists` functions should be performing an API call to verify the resource creation did occur and preferably return the API response object for potential usage in other checking functions. In this case it should be calling `GetBackupSelection` similar to the `Destroy` function above.

```go
func testAccCheckAwsBackupSelectionExists(name string, selection *backup.GetBackupSelectionOutput) resource.TestCheckFunc {
		_, ok := s.RootModule().Resources[name]
		if !ok {
			return fmt.Errorf("not found: %s, %v", name, s.RootModule().Resources)
		}

		conn := testAccProvider.Meta().(*AWSClient).backupconn

		input := &backup.GetBackupSelectionInput{
			BackupPlanId: aws.String(rs.Primary.Attributes["plan_id"]),
			SelectionId:  aws.String(rs.Primary.ID),
		}

		output, err := conn.DescribeBackupSelection(input)

		if err != nil {
			return err
		}

		*selection = *output

		return nil
	}
}
```

website/docs/r/backup_selection.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+Manages selection conditions for AWS Backup plan resources.
+
+## Example Usage
+
+```hcl
+resource "aws_backup_selection" "example" {
+  plan_id  = "${aws_backup_plan.example.id}"
+
+  name     = "tf_example_backup_selection_%d"
```
bflad: Copypasta including an invalid character 🍝 

```suggestion
  name     = "example"
```

website/docs/r/backup_selection.html.markdown
```diff
@@ -0,0 +1,52 @@
...
+
+Tag conditions (`tag`) support the following:
+
+* `type` - (Required) An operation, such as `StringEquals`, that is applied to a key-value pair used to filter resources in a selection.
+* `key` - (Required) The key in a key-value pair.
+* `value` - (Required) The value in a key-value pair.
+
+## Attributes Reference
+
+All of the above arguments are exported as attributes.
```
bflad:
```suggestion
In addition to all arguments above, the following attributes are exported:

* `id` - Backup Selection identifier

```

asweat-pubfactory:
Just curious is this scheduled for any upcoming releases? We've got the backup_vault, and backup_plan setup, super excited to have backup_selection. 👍

amirgrois:
waiting as well

slapula:
Sit tight guys, I need to make some changes then this will be up for another review. We'll merge this once we get all the kinks worked out.

aws/resource_aws_backup_selection.go
slapula: Does it matter if this resource is using these tags as a parameter to make backup plan selections vs. the normal tag/untag operations that Terraform handles?  That's what I noticed when I first planned this out but I will admit I didn't try using the default functions first because I thought they wouldn't work in this context.

slapula:
@bflad I believe I've addressed all of your comments:
```
$ make testacc TESTARGS='-run=TestAccAwsBackupSelection_'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsBackupSelection_ -timeout 120m
=== RUN   TestAccAwsBackupSelection_basic
=== PAUSE TestAccAwsBackupSelection_basic
=== RUN   TestAccAwsBackupSelection_disappears
=== PAUSE TestAccAwsBackupSelection_disappears
=== RUN   TestAccAwsBackupSelection_withTags
=== PAUSE TestAccAwsBackupSelection_withTags
=== RUN   TestAccAwsBackupSelection_withResources
=== PAUSE TestAccAwsBackupSelection_withResources
=== CONT  TestAccAwsBackupSelection_basic
=== CONT  TestAccAwsBackupSelection_withResources
=== CONT  TestAccAwsBackupSelection_disappears
=== CONT  TestAccAwsBackupSelection_withTags
--- PASS: TestAccAwsBackupSelection_disappears (19.11s)
--- PASS: TestAccAwsBackupSelection_basic (19.12s)
--- PASS: TestAccAwsBackupSelection_withResources (19.84s)
--- PASS: TestAccAwsBackupSelection_withTags (20.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	20.634s
```

aws/resource_aws_backup_selection.go
bflad: I must admit I'm not sure what you're referring to? 😅 All Terraform schema definitions are unrelated and the default `TypeSet` functions apply to any `TypeSet` attribute. We happen to use a `tags` `TypeMap` for most Terraform AWS Provider resources as the resource tagging attribute, but that is only a convention to make it simpler for operators.

If you would like to clear up any confusion between a `tag` configuration block here and the typical `tags` map argument across the provider, you can always switch the naming here to `selection_tag` or something. It won't have any bearing on the `TypeSet` `Set` function either which way though.

bflad:
From quick glance it looks like removing `Set` function from the `TypeSet` configuration block might be the only item, although I think renaming the attribute to `selection_tag` might not be a bad idea either. 👍

aws/resource_aws_backup_selection.go
slapula: Now that I have this working without the custom hash function this all makes sense.  I think I misinterpreted your initial comment to imply that I use `tagSchema()` which doesn't make sense.  Oops!  Anyway, thanks for the clarification :-)

slapula:
@bflad I went ahead and changed the attribute to `selection_tag`.  I like that and that makes it unique compared to the other uses of `tag` in this provider's codebase.

aws/resource_aws_backup_selection_test.go
```diff
@@ -0,0 +1,243 @@
...
+  iam_role_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/service-role/AWSBackupDefaultServiceRole"
+
+  selection_tag {
+    type = "STRINGEQUALS"
+    key = "foo"
+    value = "bar"
+  }
+
+  resources = [
+    "arn:aws:ec2:us-east-1:${data.aws_caller_identity.current.account_id}:volume/"
```
bflad: The acceptance testing runs in `us-west-2` by default:

https://github.com/terraform-providers/terraform-provider-aws/blob/26fea575f2100d68307e063a8b1c590d8de622d6/aws/provider_test.go#L181-L185

```
--- FAIL: TestAccAwsBackupSelection_withResources (13.48s)
        	* aws_backup_selection.test: 1 error occurred:
        	* aws_backup_selection.test: error creating Backup Selection: InvalidParameterValueException: Invalid ARN: arn:aws:elasticfilesystem:us-east-1:--OMITTED--:file-system/. Expected region is us-west-2
        	status code: 400, request id: 8c64a59c-7b8c-46de-8e7e-6fa1a7879b3a

--- FAIL: TestAccAwsBackupSelection_disappears (13.53s)
        	* aws_backup_selection.test: 1 error occurred:
        	* aws_backup_selection.test: error creating Backup Selection: InvalidParameterValueException: Invalid ARN: arn:aws:ec2:us-east-1:--OMITTED--:volume/. Expected region is us-west-2
        	status code: 400, request id: bb483272-0b15-4978-88ab-0a5cf2400302

--- FAIL: TestAccAwsBackupSelection_basic (13.60s)
        	* aws_backup_selection.test: 1 error occurred:
        	* aws_backup_selection.test: error creating Backup Selection: InvalidParameterValueException: Invalid ARN: arn:aws:ec2:us-east-1:--OMITTED--:volume/. Expected region is us-west-2
        	status code: 400, request id: 6e4c8dda-6cc8-4d63-be62-b4b156e4c30b

--- FAIL: TestAccAwsBackupSelection_withTags (13.66s)
        	* aws_backup_selection.test: 1 error occurred:
        	* aws_backup_selection.test: error creating Backup Selection: InvalidParameterValueException: Invalid ARN: arn:aws:ec2:us-east-1:--OMITTED--:volume/. Expected region is us-west-2
        	status code: 400, request id: a1229860-b1f7-4462-9b22-68d46cf62e31
```

We generally prefer using the `aws_partition` and `aws_region` data sources in this case to make the testing region and partition agnostic:

```hcl
data "aws_partition" "current" {}

// ...
"arn:${data.aws_partition.current.partition}:ec2:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:volume/"
```

aws/resource_aws_backup_selection.go
```diff
@@ -0,0 +1,180 @@
...
+						},
+						"value": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+					},
+				},
+			},
+			"resources": {
+				Type:     schema.TypeList,
```
bflad: The API seems to be unordered for this attribute:

```
--- FAIL: TestAccAwsBackupSelection_withResources (14.15s)
    testing.go:538: Step 0 error: After applying this step, the plan was not empty:

        DIFF:

        DESTROY/CREATE: aws_backup_selection.test
          iam_role_arn:                   "arn:aws:iam::123456789012:role/service-role/AWSBackupDefaultServiceRole" => "arn:aws:iam::187416307283:role/service-role/AWSBackupDefaultServiceRole"
          name:                           "tf_acc_test_backup_selection_5994421555613158377" => "tf_acc_test_backup_selection_5994421555613158377"
          plan_id:                        "7e69f127-aab4-4217-90cc-437ec8d9e19a" => "7e69f127-aab4-4217-90cc-437ec8d9e19a"
          resources.#:                    "2" => "2"
          resources.0:                    "arn:aws:ec2:us-west-2:123456789012:volume/" => "arn:aws:elasticfilesystem:us-west-2:123456789012:file-system/" (forces new resource)
          resources.1:                    "arn:aws:elasticfilesystem:us-west-2:123456789012:file-system/" => "arn:aws:ec2:us-west-2:123456789012:volume/" (forces new resource)
          selection_tag.#:                "1" => "1"
          selection_tag.3487478888.key:   "foo" => "foo"
          selection_tag.3487478888.type:  "STRINGEQUALS" => "STRINGEQUALS"
          selection_tag.3487478888.value: "bar" => "bar"
```

So this can be switched to `TypeSet` 👍

bflad:

Issue #7231, opened 2019-01-22
Tech Debt: Enable staticcheck SA4006 Check


bflad:
## Description

Recently a few Go language linting tools (e.g. `gosimple` and `unused`) have merged into [`staticcheck`](https://staticcheck.io). When switching our linting setup to the new combined linter, there are new default checks that began failing. We have explicitly disabled these in a root level `staticcheck.conf` until the fixes can be addressed and acceptance tested.

e.g.

```
aws/resource_aws_route.go:352:2:warning: this value of replaceOpts is never used (SA4006) (staticcheck)
```

## Definition of Done

- [ ] All `SA4006` linting errors from `staticcheck` are addressed
- [ ] The `-SA4006` line in `staticcheck.conf` is removed
- [ ] TravisCI and existing acceptance tests for affected code are passing

## References

* https://staticcheck.io/docs/checks#SA4006

tracypholmes:
This issue was fixed in PR #7392 .

PR #7392, opened 2019-01-31
Staticcheck SA4006

vnguyen7:
#7231 Enable staticcheck SA4006


aws/resource_aws_route.go
```diff
@@ -350,7 +350,7 @@ func resourceAwsRouteUpdate(d *schema.ResourceData, meta interface{}) error {
 		"vpc_peering_connection_id",
 	}
 	replaceOpts := &ec2.ReplaceRouteInput{}
-
+	_ = replaceOpts
```
bflad: @vnguyen7 instead of assigning this for no reason, we should instead convert the `replaceOpts` declaration to a variable declaration:

```
var replaceOpts *ec2.ReplaceRouteInput
```

tracypholmes:
Hi @vnguyen7 ! There was a merge conflict, so we went ahead and fixed it. This has now been merged. Thanks again for contributing!

Issue #7236, opened 2019-01-22
Tech Debt: Enable staticcheck SA9003 Check


bflad:
## Description

Recently a few Go language linting tools (e.g. `gosimple` and `unused`) have merged into [`staticcheck`](https://staticcheck.io). When switching our linting setup to the new combined linter, there are new default checks that began failing. We have explicitly disabled these in a root level `staticcheck.conf` until the fixes can be addressed and acceptance tested.

e.g.

```
aws/resource_aws_spot_fleet_request.go:1155:2:warning: empty branch (SA9003) (staticcheck)
```

## Definition of Done

- [ ] All `SA9003` linting errors from `staticcheck` are addressed
- [ ] The `-SA9003` line in `staticcheck.conf` is removed
- [ ] TravisCI and existing acceptance tests for affected code are passing

## References

* https://staticcheck.io/docs/checks#SA9003

PR #7403, opened 2019-01-31
Enable staticcheck SA9003

vnguyen7:
#7236 tech debt:
         Changes request for:  enable staticcheck SA9003


aws/resource_aws_spot_fleet_request.go
```diff
@@ -1135,7 +1135,7 @@ func rootBlockDeviceToSet(
 
 func resourceAwsSpotFleetRequestUpdate(d *schema.ResourceData, meta interface{}) error {
 	// http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifySpotFleetRequest.html
-	conn := meta.(*AWSClient).ec2conn
+	//Value not used: conn := meta.(*AWSClient).ec2conn
```
bflad: @vnguyen7 instead of keeping commented code in this update function that effectively does nothing, we should completely remove `resourceAwsSpotFleetRequestUpdate` and add `ForceNew: true` to any attribute with `Optional: true` or `Required: true`.

nywilken:
Hi @vnguyen7, thanks for the contribution. Since there hasn’t been any updates on this we are going to carry this change and make the edits ourselves. Here's the link to the continuing PR #7943

PR #4904, opened 2018-06-21
Add EKS cluster auth token data resource

evilmarty:
Allow Terraform to authenticate with an EKS cluster via the Kubernetes provider:

```hcl
resource "aws_eks_cluster" "foo" {
  name = "foo"
}

data "aws_eks_cluster_auth" "foo_auth" {
  name = "foo"
}

provider "kubernetes" {
  host = "${aws_eks_cluster.foo.endpoint}"
  cluster_ca_certificate = "${base64decode(aws_eks_cluster.foo.certificate_authority.0.data)}"
  token = "${data.aws_eks_cluster_auth.foo_auth.token}"
}
```

The auth logic was extracted from
https://github.com/heptio/aws-iam-authenticator because of lack of
documentation from AWS. Basically, the token is a signed URL for the
GetCallerIdentity action with a custom header. The URL is then base64
encoded and prefixed with vendor string.

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccAWSEksClusterAuthDataSource_basic'
 ==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./... -v -run=TestAccAWSEksClusterAuthDataSource_basic -timeout 120m
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
=== RUN   TestAccAWSEksClusterAuthDataSource_basic
--- PASS: TestAccAWSEksClusterAuthDataSource_basic (40.47s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	40.510s
```

PR #7438, opened 2019-02-04
Add EKS cluster auth token data resource continued

mbarrien:
This is a continuation of #4904 to create the aws_eks_cluster_auth data source for authentication tokens. This PR keeps the same interface (removing the duration flag), and on the backend uses the aws-iam-authenticator's token library directly instead of rolling our own implementation. It is also rebased against the head of master, so it includes the changeover to go modules mentioned in the predecessor PR.

Note that we had to specify a non-tagged version of aws-iam-authenticator; the most recent tagged version lacks the GetWithSTS api that we rely on. go.mod/go.sum reflects this.

```
$  make testacc TESTARGS='-run=TestAccAWSEksClusterAuthDataSource_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSEksClusterAuthDataSource_basic -timeout 120m
=== RUN   TestAccAWSEksClusterAuthDataSource_basic
--- PASS: TestAccAWSEksClusterAuthDataSource_basic (7.10s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	7.870s
...
```


aws/data_source_aws_eks_cluster_auth.go
```diff
@@ -0,0 +1,49 @@
...
+
+func dataSourceAwsEksClusterAuth() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsEksClusterAuthRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
```
bflad: Nit: `ForceNew: true` is redundant for data sources as they will always refresh.

```suggestion

```

aws/data_source_aws_eks_cluster_auth_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSEksClusterAuthDataSource_basic(t *testing.T) {
+	dataSourceResourceName := "data.aws_eks_cluster_auth.test"
+
+	resource.Test(t, resource.TestCase{
```
bflad: Nit: We should use `resource.ParallelTest()` instead of `resource.Test()` here.

```suggestion
	resource.ParallelTest(t, resource.TestCase{
```

aws/data_source_aws_eks_cluster_auth_test.go
```diff
@@ -0,0 +1,66 @@
...
+		if !ok {
+			return fmt.Errorf("Can't find EKS Cluster Auth resource: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("EKS Cluster Auth resource ID not set")
+		}
+
+		name := rs.Primary.Attributes["name"]
+		if expected := "foobar"; name != expected {
```
bflad: Checking the Terraform state for a specific attribute value can be performed without custom logic in the `TestStep` via:

```go
resource.TestCheckResourceAttr(dataSourceResourceName, "name", "foobar"),
```

aws/data_source_aws_eks_cluster_auth_test.go
```diff
@@ -0,0 +1,66 @@
...
+			return fmt.Errorf("Incorrect EKS cluster name: expected %q, got %q", expected, name)
+		}
+
+		tok := rs.Primary.Attributes["token"]
+		if tok == "" {
+			return fmt.Errorf("Token expected to not be nil")
+		}
+
+		verifier := token.NewVerifier(name)
+		identity, err := verifier.Verify(tok)
```
bflad: 👍 👍 👍

mbarrien:
All suggestions done. Please re-review.

bflad:
Looks like the build is currently failing according to TravisCI:

```
148.55s$ make test
# github.com/terraform-providers/terraform-provider-aws/aws [github.com/terraform-providers/terraform-provider-aws/aws.test]
aws/data_source_aws_eks_cluster_auth_test.go:43:33: undefined: name
aws/data_source_aws_eks_cluster_auth_test.go:46:66: undefined: name
FAIL	github.com/terraform-providers/terraform-provider-aws/aws [build failed]
make: *** [test] Error 2
The command "make test" exited with 2.
```

mbarrien:
Fixed, and added a missing file.

mbarrien:
I'm going to hate raining on my own parade here, but since token gets written to the data source state, this essentially means that a valid authentication token can get written to a tfstate file. This can be a security risk.

(That said, other solutions like the following also suffer from the same issue.)
```hcl
data "external" "aws-iam-authenticator" {
  program = ["sh", "-c", "result=$(aws-iam-authenticator token -i '${var.name}' --token-only) && echo {\\\"token\\\":\\\"$${result}\\\"}"]
}
```

bflad:

PR #7448, opened 2019-02-05
go.sum: Fix checksum of go-getter

radeksimko:
Fixing the following:

```
$ go get ./...
go: verifying github.com/hashicorp/go-getter@v0.0.0-20180327010114-90bb99a48d86: checksum mismatch
	downloaded: h1:hLYM35twiyKH44g36g+GFYODcrZQetEAY4+zrJtGea0=
	go.sum:     h1:mv3oKLM8sTaxmU/PrT39T35HRnUfchK+vtzXw6Ci9lY=
```


bflad:
Presumably I had an older checksum locally and the new checksum worked after running this:

```
$ go version
go version go1.11.5 darwin/amd64
 # If receiving the checksum mismatch error, go commands will tend to fail until the checksum file is temporarily removed
$ rm -f go.sum
$ go clean -modcache
```

PR #7455, opened 2019-02-06
tests/provider: Use GO111MODULE=off and GOFLAGS=-mod=vendor with gometalinter

bflad:
The Go module issues seem to be getting worse with respect to scanning too many files. This is to test more explicit behavior on TravisCI.


bflad:
Superseded by #7457.

PR #7465, opened 2019-02-06
Add safe_delete option to resource/aws_sns_topic

Xopherus:
`safe_delete` will protect against deletion of SNS topics if
there are other resources subscribed to that topic.

This can be a problem in the following scenario:
  - project A (with it's own terraform state) creates an `aws_sns_topic`
  - project B (with a different state) creates an
`aws_sns_topic_sbubscription`

Project A may not want to delete it's SNS topic if other projects
rely on that subscription. This `safe_delete` flag gives devs an way
to safely protect against this behavior.

Fixes #0000

* Change 1
* Change 2

```

...
```

### DIFF
--- a/aws/resource_aws_sns_topic_test.go
+++ b/aws/resource_aws_sns_topic_test.go
@@ -256,6 +256,26 @@ func TestAccAWSSNSTopic_encryption(t *testing.T) {
 	})
 }
 
+func TestAccAWSSNSTopic_safeDelete(t *testing.T) {
+	attributes := make(map[string]string)
+
+	rName := acctest.RandString(10)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:      func() { testAccPreCheck(t) },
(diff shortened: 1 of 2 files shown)


aws/resource_aws_sns_topic_test.go
```diff
@@ -256,6 +256,26 @@ func TestAccAWSSNSTopic_encryption(t *testing.T) {
 	})
 }
 
+func TestAccAWSSNSTopic_safeDelete(t *testing.T) {
```
Xopherus: Hmm, I'm not sure if this test is going to work as is.

ewbankkit:
@Xopherus Thanks for the contribution 👏.
The new attribute as you have proposed it does not fit well with Terraform's resource dependency and lifecycle management approach. You may want to try using the [`prevent_destroy`](https://www.terraform.io/docs/language/meta-arguments/lifecycle.html#prevent_destroy) meta-argument to achieve the protection you require.

PR #7480, opened 2019-02-08
resource/aws_wafregional_web_acl: Add arn attribute and logging_configuration argument

bflad:
This also enhances the documentation organization and adds an example of group rule usage.

Reference: https://github.com/terraform-providers/terraform-provider-aws/issues/5760

```
--- PASS: TestAccAWSWafRegionalWebAcl_noRules (18.11s)
--- PASS: TestAccAWSWafRegionalWebAcl_createGroup (31.70s)
--- PASS: TestAccAWSWafRegionalWebAcl_basic (36.22s)
--- PASS: TestAccAWSWafRegionalWebAcl_createRateBased (41.71s)
--- PASS: TestAccAWSWafRegionalWebAcl_changeDefaultAction (49.67s)
--- PASS: TestAccAWSWafRegionalWebAcl_changeNameForceNew (51.45s)
--- PASS: TestAccAWSWafRegionalWebAcl_disappears (57.65s)
--- PASS: TestAccAWSWafRegionalWebAcl_changeRules (59.63s)
--- PASS: TestAccAWSWafRegionalWebAcl_LoggingConfiguration (86.47s)
```

### DIFF
--- a/aws/resource_aws_wafregional_web_acl.go
+++ b/aws/resource_aws_wafregional_web_acl.go
@@ -5,6 +5,7 @@ import (
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/waf"
 	"github.com/aws/aws-sdk-go/service/wafregional"
 	"github.com/hashicorp/terraform/helper/schema"
@@ -19,6 +20,10 @@ func resourceAwsWafRegionalWebAcl() *schema.Resource {
 		Delete: resourceAwsWafRegionalWebAclDelete,
 
 		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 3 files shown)


aws/resource_aws_wafregional_web_acl.go
```diff
@@ -37,6 +42,44 @@ func resourceAwsWafRegionalWebAcl() *schema.Resource {
 					},
 				},
 			},
+			"logging_configuration": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
```
nywilken: General question: `MaxItems` implies the number of resources that can be defined within this TypeList, but since this resource happens to be a map of *scheme.Schema the resource has no bounds on the number of keys defined within it. Is that correct?

aws/resource_aws_wafregional_web_acl.go
```diff
@@ -153,6 +224,26 @@ func resourceAwsWafRegionalWebAclRead(d *schema.ResourceData, meta interface{})
...
+	loggingConfiguration := []interface{}{}
+
+	log.Printf("[DEBUG] Getting WAF Regional Web ACL (%s) Logging Configuration: %s", d.Id(), getLoggingConfigurationInput)
+	getLoggingConfigurationOutput, err := conn.GetLoggingConfiguration(getLoggingConfigurationInput)
+
+	if err != nil && !isAWSErr(err, waf.ErrCodeNonexistentItemException, "") {
+		return fmt.Errorf("error getting WAF Regional Web ACL (%s) Logging Configuration: %s", d.Id(), err)
+	}
+
+	if getLoggingConfigurationOutput != nil {
```
nywilken: General question: This seems to imply that the output can be nil even if there is no error. Is that the case or is this more of a guard pattern that is followed in the case base?

aws/resource_aws_wafregional_web_acl.go
bflad: Thats correct, the `MaxItems` on a `schema.TypeList` defines the number of configuration blocks allowed, not the number of nested attributes inside. `TypeList` and `MaxItems: 1` is a common pattern for resources to generate a single configuration block, e.g.

```hcl
resource "TYPE" "NAME" {
  logging_configuration {
    # ... nested attributes ...
  }
}
```

The following configuration will throw a validation error at plan time:

```hcl
resource "TYPE" "NAME" {
  logging_configuration {
    # ... nested attributes ...
  }

  logging_configuration {
    # ... nested attributes ...
  }
}
```

aws/resource_aws_wafregional_web_acl.go
bflad: This is one of those panic protections that is likely to never get used, but just in case the API or SDK does something unexpected in the future. Better safe than sorry; we've gotten burned in the past with API response changes.

nywilken: Thanks for that background. I'll keep that in mind for the future.

aws/resource_aws_wafregional_web_acl.go
nywilken: :pray: thanks for explanation and example case, which I validated as well :wink:

bflad:

PR #7519, opened 2019-02-12
Catch AccessDenied permission errors when creating aws_iam_user_login…

### LINKED ISSUE #7518: iam user login profile: Lacking permissions lead to erratic behavior
### Community Note

### Terraform Version

v0.11.11

aws provider version: 1.58.0

### Affected Resource(s)

* aws_iam_user_login_profile

### Description of the bug

For least privilege, I iteratively run `terraform apply` and add the required missing permissions to the iam policy until the `terraform apply` succeeds.

When doing this for user and initial-password creation (via the aws_iam_user_login_profile resource), I noticed that at one point `terraform apply` succeeds but an initial password is not set.
After some testing, I found that this behavior occurs when the terraform is run without the `iam:GetLoginProfile` permissions.

### Expected Behavior

Terraform should not report creation of the resource as successful. Instead it should show the error message.

### Actual Behavior

Terraform succeeds, but the encrypted_password property is empty.

### Steps to Reproduce

Create a aws_iam_user_login_profile resource with a user/role that does not have the aws permission "iam:GetLoginProfile" and `terraform apply`.

### References

* #7519

fbreckle:
…_profile resources.

Fixes #7518

* AccessDenied errors are no longer silently ignored

I dont understand the rationale of the "error != NoSuchEntity" part below my changes.
My fix merely catches the AccessDenied error issue.

If someone knows why the error handling is written as-is, please explain.

### DIFF
--- a/aws/resource_aws_iam_user_login_profile.go
+++ b/aws/resource_aws_iam_user_login_profile.go
@@ -126,6 +126,10 @@ func resourceAwsIamUserLoginProfileCreate(d *schema.ResourceData, meta interface
 		UserName: aws.String(username),
 	})
 	if err != nil {
+		// Catch AccessDenid errors (e.g. because we have no iam:GetLoginProfile permissions)
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == "AccessDenied" {
+			return err
+		}
 		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() != "NoSuchEntity" {
 			// If there is already a login profile, bring it under management (to prevent
 			// resource creation diffs) - we will never modify it, but obviously cannot
--- a/aws/resource_aws_iam_user_login_profile.go
+++ b/aws/resource_aws_iam_user_login_profile.go
@@ -122,23 +122,24 @@ func resourceAwsIamUserLoginProfileCreate(d *schema.ResourceData, meta interface
 	passwordResetRequired := d.Get("password_reset_required").(bool)
 	passwordLength := d.Get("password_length").(int)
 
+	// DEPRECATED: Automatic import will be removed in a future major version update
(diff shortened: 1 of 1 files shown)


bflad:
Hi @fbreckle 👋 Thanks for submitting this.

It appears this resource was initially written in a way that contradicts how Terraform is designed to work for existing infrastructure. Terraform resources _should_ require `terraform import` to begin managing existing infrastructure while this resource is instead automatically assuming control during resource creation. In a future major version release, we should expect this behavior to change and I have created a tracking issue: #7536

That said, we can still also fix the behavior for the current way this resource works. The resource creation logic in this case should be returning _all_ errors (including networking issues, etc.) and not silently ignoring them. To accomplish this, I would suggest updating the code like below:

```go
// DEPRECATED: Automatic import will be removed in a future major version update
// https://github.com/terraform-providers/terraform-provider-aws/issues/7536
_, err = iamconn.GetLoginProfile(&iam.GetLoginProfileInput{
	UserName: aws.String(username),
})

// If there is already a login profile, bring it under management (to prevent
// resource creation diffs) - we will never modify it, but obviously cannot
// set the password.
if err == nil {
  d.SetId(username)
  d.Set("key_fingerprint", "")
  d.Set("encrypted_password", "")
  return nil
}

if err != nil && !isAWSErr(err, iam.ErrCodeNoSuchEntityException, "") {
  return fmt.Errorf("error checking for existing IAM User Login Profile: %s", err)
}
```

With the above change, we can get the updated error handling in before this resource is rewritten to appropriately behave like other Terraform resources in the next major version.

fbreckle:
Hi @bflad,

thanks for your explanation. I figured already that the implementation on this method is weird/outdated, so I went with a "minimally invasive" implementation for my initial MR and just caught the AccessDenied error.
I changed the code to your snippet now, if it is still to be merged.

bflad:

Issue #4817, opened 2018-06-12
aws_elasticache_replication_group always recreates when cluster_mode replicas_per_node_group is zero


b-dean:
When a `aws_elasticache_replication_group` uses cluster mode and the `replicas_per_node` is 0, the resource will want to recreate every time.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
$ terraform version
Terraform v0.11.7
+ provider.aws v1.22.0
```

### Affected Resource(s)

* aws_elasticache_replication_group

### Terraform Configuration Files

```hcl
variable "size" {
  default = "1"
}

resource "aws_elasticache_replication_group" "foo" {
  replication_group_id          = "foo"
  replication_group_description = "foo"
  node_type                     = "cache.r3.large"
  automatic_failover_enabled    = "${var.size > 1}"

  cluster_mode {
    num_node_groups         = 1
    replicas_per_node_group = "${var.size - 1}"
  }
}
```

### Debug Output
https://gist.github.com/b-dean/2746b1e47ff03544e3c6e92fcf877af6

### Expected Behavior
Second `terraform plan` should show no changes needed

### Actual Behavior
The plan shows:

```
-/+ aws_elasticache_replication_group.foo (new resource required)
      id:                                     "foo" => <computed> (forces new resource)
      apply_immediately:                      "" => <computed>
      at_rest_encryption_enabled:             "false" => "false"
      auto_minor_version_upgrade:             "true" => "true"
      automatic_failover_enabled:             "false" => "false"
      cluster_mode.#:                         "0" => "1"
      cluster_mode.0.num_node_groups:         "" => "1"
      cluster_mode.0.replicas_per_node_group: "" => "0" (forces new resource)
      configuration_endpoint_address:         "" => <computed>
      engine:                                 "redis" => "redis"
      engine_version:                         "3.2.10" => <computed>
      maintenance_window:                     "fri:08:30-fri:09:30" => <computed>
      node_type:                              "cache.r3.large" => "cache.r3.large"
      number_cache_clusters:                  "1" => <computed>
[...]

PR #7571, opened 2019-02-15
Add clarification about cluster mode

kiwivogel:
Explains the issue encountered by users in  #4817

* Add a small warning about failing to set a parameter group

None, documentation only

### DIFF
--- a/website/docs/r/elasticache_replication_group.html.markdown
+++ b/website/docs/r/elasticache_replication_group.html.markdown
@@ -88,6 +88,10 @@ resource "aws_elasticache_replication_group" "baz" {
 }
 ```
 
+~> **Note:** While parameter_group_name is optional not setting it will result in creation of a cluster with ClusterMode off.
+This then results in forced recreation of the resource as AWS will return ClusterEnabled false while Terraform has it set to enabled 
+in the terraform state.
+
 ~> **Note:** We currently do not support passing a `primary_cluster_id` in order to create the Replication Group.
 
 ~> **Note:** Automatic Failover is unavailable for Redis versions earlier than 2.8.6,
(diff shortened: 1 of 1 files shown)


kiwivogel:
Could be worded better, feedback welcome!

website/docs/r/elasticache_replication_group.html.markdown
```diff
@@ -88,6 +88,9 @@ resource "aws_elasticache_replication_group" "baz" {
 }
 ```
 
+~> **Note:** While parameter_group_name is optional not setting it will result in creation of a cluster with ClusterMode off.
+This then results in forced recreation of the resource as AWS will return ClusterEnabled false while Terraform has it set to enabled in the terraform state.
```
masoncj: Maybe instead/in addition you could add the following note below near the `Cluster Mode`:

> Note, if enabling cluster mode via this block, be sure to use a parameter group that includes `cluster-enabled=yes`, for example `default.redis5.0.cluster.on`.  Otherwise, the replication group resource will be recreated each time you apply.

masoncj:
Could we please merge this?  This has caused us some pain recently.

robbiet480:
This should be merged but also it would be good to suggest a parameter group name of `default.redis5.0.cluster.on` which is a default in AWS.

kiwivogel:
> This should be merged but also it would be good to suggest a parameter group name of `default.redis5.0.cluster.on` which is a default in AWS.

I'll see if I can make some time to update.

breathingdust:
Hi @kiwivogel! Its been a while since you raised this and for that I apologize. We have recently done some work on smoothing out on some of the rough edges within the ElastiCache resources and have added new documentation as part of that. If you still think it needs further clarification, please raise a new PR and we will take a look.

kiwivogel:
It's not been an issue for us anymore. I can't remember after which version but somewhere in time the issue went away :D

PR #7585, opened 2019-02-15
New resource: aws_sagemaker_lifecycle_config

jckuester:
This PR adds the resource `aws_sagemaker_endpoint_configuration` (including tests).


aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"name": {
```
bflad: Can you please add validation for this field? e.g.

```go
ValidateFunc: validation.StringLenBetween(1, 63),
```

aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sagemaker"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsSagemakerLifeCycleConfiguration() *schema.Resource {
```
bflad: Its unfortunately verbose, but we should probably follow the API naming here for the resource functions and resource name to clarify this is specifically for managing Notebook Instance Lifecycle Configurations. This is in case the SageMaker API introduces other "Lifecycle Configurations".

```
aws_sagemaker_notebook_instance_lifecycle_configuration
```

```suggestion
func resourceAwsSagemakerNotebookInstanceLifeCycleConfiguration() *schema.Resource {
```

aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+			return nil
+		}
+		return fmt.Errorf("error reading SageMaker notebook instance lifecycle configuration %s: %s", d.Id(), err)
+	}
+
+	if err := d.Set("name", lifecycleConfig.NotebookInstanceLifecycleConfigName); err != nil {
+		return fmt.Errorf("error setting name for SageMaker notebook instance lifecycle configuration (%s): %s", d.Id(), err)
+	}
+
+	if len(lifecycleConfig.OnCreate) > 0 {
```
bflad: To prevent potential panics, this should perform a `nil` check

```suggestion
	if len(lifecycleConfig.OnCreate) > 0 && lifecycleConfig.OnCreate[0] != nil {
```

aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+		if err := d.Set("on_create", lifecycleConfig.OnCreate[0].Content); err != nil {
+			return fmt.Errorf("error setting on_create for SageMaker notebook instance lifecycle configuration (%s): %s", d.Id(), err)
+		}
+	}
+
+	if len(lifecycleConfig.OnStart) > 0 {
```
bflad: To prevent potential panics, this should perform a `nil` check

```suggestion
	if len(lifecycleConfig.OnStart) > 0 && lifecycleConfig.OnStart[0] != nil {
```

aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+
+func resourceAwsSagemakerLifeCycleConfigurationUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	updateOpts := &sagemaker.UpdateNotebookInstanceLifecycleConfigInput{
+		NotebookInstanceLifecycleConfigName: aws.String(d.Get("name").(string)),
+	}
+
+	onCreateHook := &sagemaker.NotebookInstanceLifecycleHook{Content: aws.String(d.Get("on_create").(string))}
+	updateOpts.SetOnCreate([]*sagemaker.NotebookInstanceLifecycleHook{onCreateHook})
```
bflad: Same here as above regarding Set function versus equals

```suggestion
	updateOpts.OnCreate = []*sagemaker.NotebookInstanceLifecycleHook{onCreateHook}
```

aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func resourceAwsSagemakerLifeCycleConfigurationUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	updateOpts := &sagemaker.UpdateNotebookInstanceLifecycleConfigInput{
+		NotebookInstanceLifecycleConfigName: aws.String(d.Get("name").(string)),
+	}
+
+	onCreateHook := &sagemaker.NotebookInstanceLifecycleHook{Content: aws.String(d.Get("on_create").(string))}
```
bflad: Should this be performing a `d.GetOk()` check or `== ""` check first?

aws/resource_aws_sagemaker_lifecycle_configuration.go
```diff
@@ -0,0 +1,168 @@
...
+
+func resourceAwsSagemakerLifeCycleConfigurationDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).sagemakerconn
+
+	deleteOpts := &sagemaker.DeleteNotebookInstanceLifecycleConfigInput{
+		NotebookInstanceLifecycleConfigName: aws.String(d.Id()),
+	}
+	log.Printf("[INFO] Deleting SageMaker notebook instance lifecycle configuration: %s", d.Id())
+
+	return resource.Retry(5*time.Minute, func() *resource.RetryError {
```
bflad: Since this `resource.Retry()` contains no `return resource.RetryableError()` conditions, the `resource.Retry()` logic should be removed.

aws/resource_aws_sagemaker_lifecycle_configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSagemakerLifecycleConfigurationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSagemakerLifecycleConfigurationConfig_Basic(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSSagemakerLifecycleConfigurationExists(resourceName, &lifecycleConfig),
+
+					resource.TestCheckResourceAttr(resourceName, "name", rName),
+					resource.TestCheckResourceAttr(resourceName, "on_create", base64Encode([]byte("echo foo"))),
```
bflad: The "basic" test configuration should omit any optional parameters and instead check for attribute zero-values or empty values, e.g.

```go
resource.TestCheckNoResourceAttr(resourceName, "on_create"),
resource.TestCheckNoResourceAttr(resourceName, "on_start"),
```

aws/resource_aws_sagemaker_lifecycle_configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+	resourceName := "aws_sagemaker_lifecycle_config.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSSagemakerLifecycleConfigurationDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSagemakerLifecycleConfigurationConfig_Basic(rName),
+				Check: resource.ComposeTestCheckFunc(
```
bflad: We should check the value of the `arn` attribute is populated correct in here, e.g.

```go
testAccCheckResourceAttrRegionalARN(resourceName, "arn", "sagemaker", fmt.Sprintf("notebook-instance-lifecycle-config/%s", rName)),
```

aws/resource_aws_sagemaker_lifecycle_configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		conn := testAccProvider.Meta().(*AWSClient).sagemakerconn
+		lifecycleConfig, err := conn.DescribeNotebookInstanceLifecycleConfig(&sagemaker.DescribeNotebookInstanceLifecycleConfigInput{
+			NotebookInstanceLifecycleConfigName: aws.String(rs.Primary.ID),
+		})
+
+		if err != nil {
+			if isAWSErr(err, "ValidationException", "") {
+				return nil
```
bflad: `CheckDestroy` functions should use `continue` instead of `return nil` to allow checking multiple declarations of the resource.

aws/resource_aws_sagemaker_lifecycle_configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+			return err
+		}
+
+		if lifecycleConfig != nil && aws.StringValue(lifecycleConfig.NotebookInstanceLifecycleConfigName) == rs.Primary.ID {
+			return fmt.Errorf("SageMaker notebook instance lifecycle configuration %s still exists", rs.Primary.ID)
+		}
+
+		return nil
```
bflad: Same here although the line can just be removed since it is the end of the `for` loop logic.

aws/resource_aws_sagemaker_lifecycle_configuration.go
jckuester: I'll use the func `validateSagemakerName` here to check also against the name pattern (https://github.com/awsdocs/amazon-sagemaker-developer-guide/blob/master/doc_source/API_CreateNotebookInstanceLifecycleConfig.md#request-parameters)

jckuester:
@bflad I addressed your feedback (see last commit).

bflad:

PR #7587, opened 2019-02-16
Issue #7573 terraform crash - codepipeline

### LINKED ISSUE #7573: resource/aws_codepipeline: Crash when missing artifact_store or using multiple artifact stores
_This issue was originally opened by @MarkAyliffe as hashicorp/terraform#20355. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

As title

### Crash Output

[crash.log](https://github.com/hashicorp/terraform/files/2868670/crash.log)

### Expected Behavior
It should work without crashing like it has for the last couple of months!
### Actual Behavior
Terraform crashes if I try terraform plan or terraform apply.

saravanan30erd:
Fixes #7573 

* Change 1
verifying the output before set to attr  - aws_codepipeline

### DIFF
--- a/aws/resource_aws_codepipeline.go
+++ b/aws/resource_aws_codepipeline.go
@@ -227,8 +227,12 @@ func expandAwsCodePipelineArtifactStore(d *schema.ResourceData) *codepipeline.Ar
 
 func flattenAwsCodePipelineArtifactStore(artifactStore *codepipeline.ArtifactStore) []interface{} {
 	values := map[string]interface{}{}
-	values["type"] = *artifactStore.Type
-	values["location"] = *artifactStore.Location
+	if artifactStore.Type != nil {
+		values["type"] = aws.StringValue(artifactStore.Type)
+	}
+	if artifactStore.Location != nil {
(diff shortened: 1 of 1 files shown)


aws/resource_aws_codepipeline.go
```diff
@@ -227,8 +227,12 @@ func expandAwsCodePipelineArtifactStore(d *schema.ResourceData) *codepipeline.Ar
 
 func flattenAwsCodePipelineArtifactStore(artifactStore *codepipeline.ArtifactStore) []interface{} {
 	values := map[string]interface{}{}
-	values["type"] = *artifactStore.Type
-	values["location"] = *artifactStore.Location
+	if artifactStore.Type != nil {
```
bflad: Performing a `nil` check before using the AWS Go SDK type conversion functions is redundant. They convert `nil` to the zero-value for the type: https://docs.aws.amazon.com/sdk-for-go/api/aws/#StringValue

bflad:
Closing for now given the above and no additional movement within this pull request.

PR #7612, opened 2019-02-20
aws_kinesis_firehose_delivery_stream: Updated key description to include elasticsearch

jurajseffer:

* Better s3_configuration description that mentions that this key can be also used when using elasticsearch as destination as `extended_s3_configuration` cannot be used in that case


jurajseffer:
@bflad makes sense, thanks.

PR #7615, opened 2019-02-20
Added use case for availability_zones in autoscaling docs

unixninja92:

* Fix docs to include new use case for the availability_zones parameter in the auto scaling group resource


unixninja92:
yeah, when you try and create an auto scaling group based on a launch template that attaches a specific ENI via id, if you read the current docs you assume you need to set `vpc_zone_identifier` in the asg config rather than `availability_zones` becasue of the note that the latter is only for ec2-classic. When running 
```
resource "aws_network_interface" "XXXXX" {
  subnet_id = "${data.terraform_remote_state.vpc.public-subnet}"
  security_groups = ["${aws_security_group.XXXXX.id}"]
  source_dest_check = false
}

resource "aws_launch_template" "XXXXX" {
    name = "XXXXX"
    ebs_optimized = true
    image_id = "ami-XXXXX"
    instance_type               = "m5.large"
    key_name = "XXXXX"
    network_interfaces {
      network_interface_id = "${aws_network_interface.XXXXX.id}"
    }
    user_data            = "${base64encode(local.XXXXX-userdata)}"
}
resource "aws_autoscaling_group" "XXXXX" {
  name                 = "XXXXX"
  launch_template {
    id = "${aws_launch_template.XXXXX.id}"
    version = "${aws_launch_template.XXXXX.latest_version}"
  }
  min_size             = 1
  desired_capacity     = 1
  max_size             = 1
  vpc_zone_identifier = ["${data.terraform_remote_state.vpc.public-subnet}"]
}
```
you get the following error:
`Invalid details specified: You must use a valid fully-formed launch template. A network interface may not specify both a network interface ID and a subnet`
This is because the subnet is selected based on where the ENI is and they don't want you defining this twice. When neither `vpc_zone_identifier` or `availability_zones` is specified you get an error saying you need to set one of them (don't have the exact error message on hand). 

I opened a support ticket with aws about this as I was also having this issue with their web console. They confirmed that it was a bug in the web console, but gave me cli commands that work (which sets `availability_zones` instead of `vpc_zone_identifier`) and also noted that their cli docs do not currently cover this use case. I believe fixing this is now on their internal roadmap. 

To summarize, switching 
`vpc_zone_identifier = ["${data.terraform_remote_state.vpc.public-subnet}"]` 
to
`availability_zones = ["${data.aws_subnet.XXXXX-subnet.availability_zone}"]`
makes it work but is not reflected in current docs.

rflume:
> They confirmed that it was a bug in the web console, but gave me cli commands that work [...]

Hey, could you provide that command maybe? I'm having the same issue currently but try not to use terraform to create the ASG.

unixninja92:
> > They confirmed that it was a bug in the web console, but gave me cli commands that work [...]
> 
> Hey, could you provide that command maybe? I'm having the same issue currently but try not to use terraform to create the ASG.

`aws autoscaling create-auto-scaling-group --auto-scaling-group-name <NAME> --min-size 1 --max-size 1 --launch-template LaunchTemplateId=<TemplateID> --availability-zones <az-name>`

PR #7645, opened 2019-02-21
Vpc endpoint suppress policy diff

snakeb1t:
* Suppresses policy diff for vpc endpoint policy when old and new policy is structurally equivalent
  * Lack of this feature is considered a bug on our side, since a change is occurring despite both old and new policy being logically equivalent. We're using PTFE, so we require the upstream change for Terraform runs to be kosher.

### DIFF
--- a/aws/resource_aws_vpc_endpoint.go
+++ b/aws/resource_aws_vpc_endpoint.go
@@ -46,10 +46,10 @@ func resourceAwsVpcEndpoint() *schema.Resource {
 				ForceNew: true,
 			},
 			"policy": {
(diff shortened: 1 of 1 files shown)


bflad:
This has been released in [version 1.60.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#1600-february-22-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #7734, opened 2019-02-26
resource/aws_route53_record: Switch allow_overwrite default from true to false

### LINKED ISSUE #3895: Sugestion: aws_route53_record allow_overwrite should have the default value to false
### Terraform Version
```
Terraform v0.11.5
+ provider.aws v1.11.0
```
### Affected Resource(s)

- aws_route53_record

```hcl
provider "aws" {
  version = "1.11"
  region  = "us-east-1"
}

resource "aws_route53_record" "www_d2" {
  zone_id = "ZZZZZZZZZZZZZZ"
  name    = "marcos_test.AAAAAAAAA.com"
  type    = "A"
  ttl     = "300"

  #   allow_overwrite = "false"
  records = ["10.10.10.20"]
}

```

### Sugestion

One of the nice things about terraform is that it only fiddles with resources it created (unless one explicitly uses `terraform import`)

That means one can be certain that he or she won't step on nobody else's toes while using terraform.

Weirdly, that does not happen with `aws_route53_record`.
It simply overwrites existing records as if they did not matter.
@Erouan50 was nice enough to create a parameter called `allow_overwrite` exactly to prevent that behavior.

Nevertheless the default value is `true`. I deeply believe it should be `false`. The former is counter intuitive and when one notices, it may be too late. Although this is a behavior change, it's the same as adding a sit belt to the existing process. More or less like `terraform --apply` is not automatic now.

Please consider changing the default value of `allow_overwrite` to `false`

### References
- https://github.com/terraform-providers/terraform-provider-aws/pull/2926

bflad:
Closes #3895
Reference: https://github.com/terraform-providers/terraform-provider-aws/pull/2926

Previously, the `aws_route53_record` resource did not follow standard Terraform conventions of requiring existing infrastructure to be imported into Terraform's state for management, which meant operators could unexpectedly affect that existing infrastructure. In version 1.10.0, we introduced the `allow_overwrite` argument so operators could opt into the upcoming import requirement and force the Terraform resource during resource creation to error if it attempted to create a Route53 record that previously existed.

Here we make the breaking change to switch the default resource behavior to error on creation for existing records. Operators can opt out of the new behavior by setting `allow_overwrite = true`, but it is marked as deprecated for removal in the next major version and will display the deprecation warning when used to signal that workflows should be adjusted if necessary.

```
--- PASS: TestAccAWSRoute53Record_Alias_Elb (319.83s)
--- PASS: TestAccAWSRoute53Record_Alias_S3 (123.47s)
--- PASS: TestAccAWSRoute53Record_Alias_Uppercase (184.67s)
--- PASS: TestAccAWSRoute53Record_Alias_VpcEndpoint (450.17s)
--- PASS: TestAccAWSRoute53Record_AliasChange (157.29s)
--- PASS: TestAccAWSRoute53Record_allowOverwrite (365.48s)
--- PASS: TestAccAWSRoute53Record_basic (130.53s)
--- PASS: TestAccAWSRoute53Record_basic_fqdn (146.30s)
--- PASS: TestAccAWSRoute53Record_caaSupport (177.87s)
--- PASS: TestAccAWSRoute53Record_disappears (107.58s)
--- PASS: TestAccAWSRoute53Record_disappears_MultipleRecords (247.77s)
--- PASS: TestAccAWSRoute53Record_empty (115.48s)
--- PASS: TestAccAWSRoute53Record_failover (204.75s)
--- PASS: TestAccAWSRoute53Record_generatesSuffix (180.48s)
--- PASS: TestAccAWSRoute53Record_geolocation_basic (196.58s)
--- PASS: TestAccAWSRoute53Record_importBasic (174.28s)
--- PASS: TestAccAWSRoute53Record_importUnderscored (114.40s)
--- PASS: TestAccAWSRoute53Record_latency_basic (173.99s)
--- PASS: TestAccAWSRoute53Record_longTXTrecord (114.97s)
--- PASS: TestAccAWSRoute53Record_multivalue_answer_basic (197.09s)
--- PASS: TestAccAWSRoute53Record_SetIdentifierChange (206.47s)
--- PASS: TestAccAWSRoute53Record_spfSupport (152.57s)
--- PASS: TestAccAWSRoute53Record_txtSupport (170.00s)
--- PASS: TestAccAWSRoute53Record_TypeChange (220.81s)
--- PASS: TestAccAWSRoute53Record_weighted_alias (278.61s)
--- PASS: TestAccAWSRoute53Record_weighted_basic (112.68s)
--- PASS: TestAccAWSRoute53Record_wildcard (216.04s)
```

### DIFF
--- a/website/docs/guides/version-2-upgrade.html.md
+++ b/website/docs/guides/version-2-upgrade.html.md
@@ -650,7 +650,7 @@ resource, with in-line routes, to the state.
 
 ### allow_overwrite Default Value Change
 
-The resource now requires existing Route 53 Records to be imported into the Terraform state for management unless the `allow_overwrite` argument is enabled. The `allow_overwrite` flag is considered deprecated for removal in the next major version of the Terraform AWS Provider (version 3.0.0).
+The resource now requires existing Route 53 Records to be imported into the Terraform state for management unless the `allow_overwrite` argument is enabled. The `allow_overwrite` flag is considered deprecated for removal in a future major version of the Terraform AWS Provider.
 
 For example, if the `www.example.com` Route 53 Record in the `example.com` Route 53 Hosted Zone existed previously and this new Terraform configuration was introduced:
 
--- a/website/docs/guides/version-2-upgrade.html.md
+++ b/website/docs/guides/version-2-upgrade.html.md
@@ -24,6 +24,7 @@ Upgrade topics:
 - [Data Source: aws_ami_ids](#data-source-aws_ami_ids)
 - [Data Source: aws_iam_role](#data-source-aws_iam_role)
 - [Data Source: aws_kms_secret](#data-source-aws_kms_secret)
+- [Data Source: aws_lambda_function](#data-source-aws_lambda_function)
 - [Data Source: aws_region](#data-source-aws_region)
 - [Resource: aws_api_gateway_api_key](#resource-aws_api_gateway_api_key)
 - [Resource: aws_api_gateway_integration](#resource-aws_api_gateway_integration)
@@ -44,6 +45,7 @@ Upgrade topics:
 - [Resource: aws_network_acl](#resource-aws_network_acl)
 - [Resource: aws_redshift_cluster](#resource-aws_redshift_cluster)
 - [Resource: aws_route_table](#resource-aws_route_table)
+- [Resource: aws_route53_record](#resource-aws_route53_record)
 - [Resource: aws_route53_zone](#resource-aws_route53_zone)
 - [Resource: aws_wafregional_byte_match_set](#resource-aws_wafregional_byte_match_set)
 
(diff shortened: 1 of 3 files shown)


website/docs/guides/version-2-upgrade.html.md
```diff
@@ -644,6 +646,31 @@ Previously, importing this resource resulted in an `aws_route` resource for each
 addition to the `aws_route_table`, in the Terraform state. Support for importing `aws_route` resources has been added and importing this resource only adds the `aws_route_table` 
 resource, with in-line routes, to the state.
 
+## Resource: aws_route53_record
+
+### allow_overwrite Default Value Change
+
+The resource now requires existing Route 53 Records to be imported into the Terraform state for management unless the `allow_overwrite` argument is enabled. The `allow_overwrite` flag is considered deprecated for removal in the next major version of the Terraform AWS Provider (version 3.0.0).
```
paddycarver: Suggestion: don't name a specific version, just say "a future version". Then if you forget for 3.0.0, you didn't break a promise.

bflad: 👍 Great suggestion

osterman:
@bflad just making sure https://github.com/terraform-providers/terraform-provider-aws/pull/7734 is on your radar =) Hoping that `allow_overwrite` **will not** actually get deprecated as this is a very valid use-case for setting the `SOA` record that necessitates `allow_overwrite`.

PR #7753, opened 2019-02-28
deps: Update github.com/hashicorp/terraform@pluginsdk-v0.12-early4

bflad:
Support Terraform 0.12 plugin protocol. Updated via:

```
go get github.com/hashicorp/terraform@pluginsdk-v0.12-early4
go mod tidy
go mod vendor
```


bflad:
There have been some changes upstream which will require an additional dependency update. Rather than trying to update this pull request, which already has merge conflicts, we will submit a new one when the appropriate upstream version is determined.

Issue #7006, opened 2018-12-31
Rename instance_interruption_behaviour attribute in v2.0.0


stefansundin:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Proposal: in `aws_spot_fleet_request` and `aws_spot_instance_request`, rename `instance_interruption_behaviour` to `instance_interruption_behavior` in v2.0.0.

The current name does not match the AWS API, and it is inconsistent with `aws_launch_template`, which is using the proper name. I think it makes sense to resolve this inconsistency in v2.0.0.

I was reading through the Provider Version 2 Upgrade Guide, and I think this request is similar to many of the things already in there.

I could attempt to submit a PR if this is given the Ok. I just don't know which branch is currently used for v2.0.0 development. `v2.0.0-dev` seems to be a month old? And do I just rename it 

### Affected Resources

* `aws_spot_instance_request`
* `aws_spot_fleet_request`

tomelliff:
This was briefly discussed in https://github.com/terraform-providers/terraform-provider-aws/pull/1986 but the breaking change was dropped from that PR in the end and I didn't get around to raising the PR to add the breaking change to fix the naming because I've moved away from individual spot requests since then.

nywilken:
Hi @stefansundin apologies for any frustration here. I reviewed the pull request associated with this issue and, unfortunately could not approve it. Renaming attributes has a [detailed process](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename) that needs to be followed, and can have a large impact on operators down stream. I left more details around this particular issue on your PR. 

Thanks for your help and contributions to the provider. Apologies again for any frustration here and we will take extra precautions to prevent this situation in the future.

stefansundin:
@nywilken Here's an idea. Create the `v3.0.0` tag and assign it to this issue so that this won't be forgotten again. I know it's probably some time before v3.0.0 will be needed, but by just closing this issue you are guaranteeing it will be forgotten (or a more likely scenario is that someone else will open a duplicate of this issue, taking more of your time).

PR #7758, opened 2019-02-28
Rename instance_interruption_behaviour

stefansundin:
A bit to my disappointment, v2.0.0 was released without https://github.com/terraform-providers/terraform-provider-aws/issues/7006 being fixed. So I decided to give it a go.

Initially I tried to deprecate it cleanly using `Deprecated` and `DiffSuppressFunc`, and it kinda worked, but there were some edge cases that I didn't like a lot. For example, my plan would print both attributes because I still had `Default` in there.

Example:
```
  + aws_spot_instance_request.cheap_worker
      id:                              <computed>
      ami:                             "ami-005bdb005fb00e791"
      arn:                             <computed>
      associate_public_ip_address:     <computed>
      availability_zone:               <computed>
      host_id:                         <computed>
      instance_interruption_behavior:  "terminate"
      instance_interruption_behaviour: "stop"
      instance_state:                  <computed>
      instance_type:                   "t1.micro"
```

This didn't look very great to me.

Anyway, I decided to just remove the old attribute instead. Maybe it can still be sneaked in a v2.1.0 release. I think this attribute is not that commonly used, and I hope that not everyone updates to v2.0.0 too quickly so we still have a little bit of time to break things.

I also added some minor things. Validation in `aws_launch_template`, sorted the validation lists, and for whatever reason there's both `ec2.InstanceInterruptionBehavior*` and `ec2.SpotInstanceInterruptionBehavior*` with no clear reasoning so I updated everything to use the constants without `Spot` in the name.

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/7006.

Let me know what you think!

### DIFF
--- a/aws/resource_aws_ec2_fleet.go
+++ b/aws/resource_aws_ec2_fleet.go
@@ -175,11 +175,11 @@ func resourceAwsEc2Fleet() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 							ForceNew: true,
-							Default:  ec2.SpotInstanceInterruptionBehaviorTerminate,
+							Default:  ec2.InstanceInterruptionBehaviorTerminate,
 							ValidateFunc: validation.StringInSlice([]string{
-								ec2.SpotInstanceInterruptionBehaviorHibernate,
-								ec2.SpotInstanceInterruptionBehaviorStop,
-								ec2.SpotInstanceInterruptionBehaviorTerminate,
+								ec2.InstanceInterruptionBehaviorHibernate,
+								ec2.InstanceInterruptionBehaviorStop,
+								ec2.InstanceInterruptionBehaviorTerminate,
 							}, false),
 						},
 						"instance_pools_to_use_count": {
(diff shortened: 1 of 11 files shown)


nywilken:
Hi @stefansundin thanks for your work on this PR and for creating #7006. Unfortunately, it is not a change we can accept at this time. 

The release v2.0.0, which contained a number of breaking changes, was an initiative to remove previously deprecated attributes following our [attribute deprecation guidelines](https://www.terraform.io/docs/extend/best-practices/deprecations.html),  a known security issue, and some heavy impact user experience bugs.

The main change proposed in this PR, while valid, is an implementation detail that we have decided to continue supporting, as changing the attribute name would have a large impact on our users for little gain. A little background on the impact of renaming attributes is best summed up by @bflad on https://github.com/terraform-providers/terraform-provider-aws/issues/4137#issuecomment-470108161

For future contributions we are working hard to ensure inconsistencies are caught during the review phase, but for inconsistencies that already exist the plan forward is to not change what is not broken.

Issue #6246, opened 2018-10-23
EBS root block encryption not available in EC2


georgesvachulay:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I would like the following to work:

```
    root_block_device = {
      volume_type = "gp2"
      volume_size = "10"
      encrypted   = true
    }
```

Current state if used:

`root_block_device.0: invalid or unknown key: encrypted`

Works fine with the `ebs_block_device `resource (https://www.terraform.io/docs/providers/aws/r/instance.html#block-devices). This is a fairly general use case in AWS, without this the most used volume of an instance won't be encrypted.

### New or Affected Resource(s)

* aws_instance
* aws_ebs_volume (potentially)

### Potential Terraform Configuration

```hcl
resource "aws_instance" "xy" {
    root_block_device = {
      volume_type = "gp2"
      volume_size = "10"
      encrypted   = true
    }
}
```

### References

* https://github.com/terraform-aws-modules/terraform-aws-ec2-instance/issues/6

jcreyf:
Please also add support for `kms_key_id`

a-h:
`aws_launch_configuration` would also need updating to allow machines to have their root volumes encrypted.

## Input

```
resource "aws_launch_configuration" "example_config" {
  image_id             = "ami-0274e11dced17bb5b"
  instance_type        = "t2.micro"
  root_block_device {
    encrypted = true
  }
}
```

## Current Output

```
aws_launch_configuration.example_config: root_block_device.0: invalid or unknown key: encrypted
```

The current workaround for this is use an encrypted AMI (see https://github.com/terraform-aws-modules/terraform-aws-ec2-instance/issues/6#issuecomment-387003402).

johnjelinek:
I hacked on this functionality and got it to read the root block encrypted state, but I could not get it to encrypt an AMI that was unencrypted.

```
--- FAIL: TestAccAWSInstance_blockDevices (7.80s)
    testing.go:538: Step 0 error: Error applying: 1 error occurred:
                * aws_instance.foo: 1 error occurred:
[...]

PR #7759, opened 2019-02-28
Add the ability to encrypt root_block_device in aws_launch_configuration

joestump:
Refs #6246
Refs #8624

* Adds the `encrypted` attribute to the `root_block_device` on `aws_launch_configuration`.

**NOTE:** It appears `KmsKeyId` is [not supported by this API](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_Ebs.html).

```
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSLaunchConfiguration_encryptedRootBlockDevice -timeout 120m
=== RUN   TestAccAWSLaunchConfiguration_encryptedRootBlockDevice
=== PAUSE TestAccAWSLaunchConfiguration_encryptedRootBlockDevice
=== CONT  TestAccAWSLaunchConfiguration_encryptedRootBlockDevice
--- PASS: TestAccAWSLaunchConfiguration_encryptedRootBlockDevice (34.95s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	36.220s
```

### DIFF
--- a/aws/resource_aws_launch_configuration.go
+++ b/aws/resource_aws_launch_configuration.go
@@ -260,6 +260,13 @@ func resourceAwsLaunchConfiguration() *schema.Resource {
 							ForceNew: true,
 						},
 
+						"encrypted": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+							ForceNew: true,
+						},
+
 						"iops": {
 							Type:     schema.TypeInt,
 							Optional: true,
@@ -419,6 +426,10 @@ func resourceAwsLaunchConfigurationCreate(d *schema.ResourceData, meta interface
 			bd := v.(map[string]interface{})
 			ebs := &autoscaling.Ebs{
 				DeleteOnTermination: aws.Bool(bd["delete_on_termination"].(bool)),
+			}
+
+			if v, ok := bd["encrypted"].(bool); ok && v {
+				ebs.Encrypted = aws.Bool(v)
 			}
 
 			if v, ok := bd["volume_size"].(int); ok && v != 0 {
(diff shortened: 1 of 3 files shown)


bflad:
The [API documentation](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_Ebs.html) for this parameter is pretty confusing 😅 
> Specifies whether the volume should be encrypted. Encrypted EBS volumes must be attached to instances that support Amazon EBS encryption. Volumes that are created from encrypted snapshots are automatically encrypted. There is no way to create an encrypted volume from an unencrypted snapshot or an unencrypted volume from an encrypted snapshot. If your AMI uses encrypted volumes, you can only launch it on supported instance types. For more information, see Amazon EBS Encryption in the Amazon EC2 User Guide for Linux Instances.

Are we sure this actually does anything for unencrypted root block devices to encrypt them? It might be worth spinning this up in an autoscaling group and verifying. We've seen this a bunch with EC2 Launch Templates where the template will create but its actual usage will cause an error.

joestump:
> Are we sure this actually does anything for unencrypted root block devices to encrypt them?

It will definitely not do this. This _only_ works if the instance type supports it. If you're using a snapshot, it _must_ be encrypted. Also, like I said no KMS support so it'll be encrypted with your `aws/ebs` key. I suspect if you provide an unencrypted snapshot _or_ an instance type that doesn't support encryption, you'd end up with ASG errors. 

I'll run a more manual test and report back. I did verify in the console that `aws_instance` works as advertised. Presumably with the right instance type and no snapshot, you'd be good to go.

jhmartin:
> If you're using a snapshot, it must be encrypted.
https://aws.amazon.com/about-aws/whats-new/2019/05/launch-encrypted-ebs-backed-ec2-instances-from-unencrypted-amis-in-a-single-step/ may have changed that

atiftw:
Yep - AWS now allows you to launch instances with encrypted root block devices from amis with unencrypted ones.Also supports adding a KMS key, i can add support for the latter once this is merged to master.We need this badly.

JoostSaanen:
@atiftw That's awesome. Where did you hear/read it?

jkface:
Can we have this merged now and create a new PR for supporting KMS key ID once that feature is in?

jc-asdf:
NOTE - this seems to have broken the instance store case.

When creating an instance-store launch template, it fails with:
```
Error: Error applying plan:

1 error occurred:
	* aws_launch_configuration.yes_i_used_an_instance_store_ami: 1 error occurred:
	* aws_launch_configuration.yes_i_used_an_instance_store_ami: Instance store backed AMIs do not provide a root device name - Use an EBS AMI
```

Writing up the full bug and will link to this issue.

Issue #1110, opened 2017-07-11
Using Lambda function modules with an empty environment


jsonmaur:
### Terraform Version
0.9.11

### Affected Resource(s)
- aws_lambda_function

###### main.tf
```hcl
module "edge" {
  source = "./modules/function"
  # ...
}
```

###### modules/function/main.tf
```hcl
variable "environment" {
  type = "map"
  default = {}
}

resource "aws_lambda_function" "main" {
  function_name = "hello-world"
  # ...
  environment = {
    variables = "${var.environment}"
  }
}
```

### Debug Output
* module.edge.aws_lambda_function.main: 1 error(s) occurred:
* aws_lambda_function.main: At least one field is expected inside environment

### Expected Behavior
It should allow an empty object for the function environment variables, specifically because Lambda@Edge doesn't allow environment variables. Breaks all the function modules that want to be repurposed for edge functions.

### Actual Behavior
It doesn't allow the empty object. If I try to go one step up as seen below, it produces a different error:

```hcl
resource "aws_lambda_function" "main" {
  function_name = "hello-world"
  # ...
  environment = "${var.environment}"
}
```

* module.edge.aws_lambda_function.main: environment: should be a list

Obviously `environment` is supposed to be a map, but it seems to expect a list. I can't find a workaround to providing a default empty object for the environment, aside from creating a seperate module specifically for edge functions, which would omit `environment` entirely.

jsonmaur:
Also, according to the docs, `aws_lambda_function.environment.variables` is optional. But if you omit it, you get the `At least one field is expected inside environment` error, making it required.

fgimian:
I can also confirm this issue on my end and know of no workarounds to the problem.

sebolabs:
yes, this is problematic in my case as well
I have a lambda module and I pass environment variables to it only if I need to

@Terraform v. 0.10.7

hmnp:
Is there any developments on this?

bflad:
The configuration language of Terraform (HCL) currently has some limitations when it comes to typing and lacks the ability to pass an "empty"/`null` value. The good news is that this is likely being addressed in the next major version of Terraform: https://www.hashicorp.com/blog/terraform-0-1-2-preview

In this situation, the following configuration will be available:

```hcl
# Potential Terraform 0.12 configuration - implementation may change during development
resource "aws_lambda_function" "main" {
  # ... other configuration ...
  environment = var.environment
}
```

[...]

PR #7762, opened 2019-02-28
Refactor Lambda environment variable handling

### LINKED ISSUE #1272: Empty map assignment fails
_This issue was originally opened by @betabandido as hashicorp/terraform#15651. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

I have developed a module that creates AWS lambda functions. When I tried to let users pass environment variables to the module (so that they can be passed to `aws_lambda_function`) I got into an issue where assigning a map to a variable fails if the map is empty.

### Terraform Version
Terraform v0.10.0-rc1-dev

Partial contents of the module file:

```hcl
variable "variables" {
  type = "map"
  default = {}
}

resource "aws_lambda_function" "lambda_function" {
  # ...

  environment {
    variables = "${var.variables}"
  }
}
```

When the code using the lambda module does not define any variables, `var.variables` defaults to an empty map. When terraform tries to assign the map to `environment.variables` I get the following error:

```
Error applying plan:
1 error(s) occurred:
* module.lambda.aws_lambda_function.lambda_function: 1 error(s) occurred:
* aws_lambda_function.lambda_function: At least one field is expected inside environment
```

If I pass some environment variables to the module, or if I define `var.variables` as:

```hcl
variable "variables" {
  type = "map"
  default = { dummy_ = "1" }
}
```

then, I do not get any error and terraform successfully creates the lambda function.

This issue does not seem to be strictly related to the AWS provider, but it seems to affect the terraform core. Could you please have a look into it? The problem itself occurs in the following line of code: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_lambda_function.go#L313

You can find the complete code for the module here: https://github.com/betabandido/TerraformModules/blob/feature/lambda-env-vars/modules/lambda/main.tf

ouroboros8:
This change refactors the `environment` parameter to accept a map of key/value pairs corresponding to environment variable name/values.

Previously, `environment` was a configuration block with one mandatory key, `variables`, which itself was a map of environment variable name/values. This set up, besides being somewhat redundant, also caused some problems for users attempting to pass empty maps to `variables` (see issues #1272 and #1110).

This refactor resolves the core problem in the issues mentioned above whereby users are unable to pass an empty map of variable to the lambda resource; setting `environment` to an empty map will create no environment variables as expected.

However, this change is not backwards compatible with the previous form of `environment`. I'm submitting this as a minimal fix, but I am open to suggestions; perhaps creating a new `environment_variables` parameter and depracating the `environment` parameter would be a preferable approach?  

Fixes #1272 and #1110

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -217,18 +217,8 @@ func resourceAwsLambdaFunction() *schema.Resource {
 				Computed: true,
 			},
 			"environment": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeMap,
 				Optional: true,
-				MaxItems: 1,
-				Elem: &schema.Resource{
-					Schema: map[string]*schema.Schema{
-						"variables": {
-							Type:     schema.TypeMap,
-							Optional: true,
-							Elem:     &schema.Schema{Type: schema.TypeString},
-						},
-					},
-				},
 			},
 			"tracing_config": {
 				Type:     schema.TypeList,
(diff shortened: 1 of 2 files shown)


bflad:
Hi @ouroboros8 👋  Thank you for submitting this.

After long consideration and due to the [`dynamic` block syntax](https://www.terraform.io/docs/configuration/expressions.html#dynamic-blocks) introduced into Terraform 0.12 released a year ago which should resolve issues around dynamically handling configuration blocks, we have decided that causing a breaking change of this type that would impact a large portion of our community is not something we want to impose or support at this time. We also have concerns that the [Lambda API could introduce additional fields into the `Environment` object](https://docs.aws.amazon.com/lambda/latest/dg/API_Environment.html), which would cause further burden on the community in the future should we want to go back to following the API structure. Given the above, I'm going to close this. 

Thanks again for the potential solution. 👍

PR #7774, opened 2019-02-28
Add tagging to aws_resourcegroup_group.

joestump:
Refs #6217.

* Adds `tags` attribute to `aws_resourcegroup_group`.

Import test is currently failing. No idea why.

```
TF_ACC=1 go test ./... -v -parallel 20 -run="TestAccAWSResourceGroup_basic" -timeout 120m
=== RUN   TestAccAWSResourceGroup_basic
=== PAUSE TestAccAWSResourceGroup_basic
=== CONT  TestAccAWSResourceGroup_basic
--- FAIL: TestAccAWSResourceGroup_basic (22.26s)
    testing.go:538: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=2) {
         (string) (len=6) "tags.%": (string) (len=1) "1",
         (string) (len=8) "tags.Foo": (string) (len=3) "bar"
        }
        
FAIL
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	22.525s
make: *** [testacc] Error 1
```


joestump:
@bflad I tried creating an importer based on `resourceAwsResourceGroupsGroupRead` and it still wasn't pulling in the tags. Any pointers on how to fix this import test would be great.

bflad:
Related: #7811

DrFaust92:
Hey @joestump, i have a working PR (#10640) for this with changes including the `keyvaluetags` package, would you like me to merge these to this PR or close this one?

bflad:
Hi @joestump 👋 Thank you for working on this and sorry it was previously lingering.

Relating to the `ImportStateVerify` issue, it is generally is a signal that the `Read` function is not properly saving an attribute (via `d.Set()`) into the Terraform state correctly. One way of ensuring any `d.Set()` call is properly working is to perform error checking on it, e.g.

```go
if err := d.Set("tags", /* ... */); err != nil {
}
```

As @DrFaust92 is mentioning above, there have been some consistency changes and simplifications to how to implement resource tagging into the Terraform AWS Provider codebase, which can be found in the [Contributing Guide](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#adding-resource-tagging-support) using a new internal `keyvaluetags` package.

Since we didn't see a response to the above and it appears this pull request wasn't in a mergeable state, we are going to opt to close this in preference of the one implementing the new methodology. Thanks again for your contribution here and let us know if you have any questions.

PR #7791, opened 2019-03-01
Fix aws_s3_bucket_notification without filters

MickVanDuijn:
Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/6724

* Check properties before de-referencing, to prevent nil pointer exception


bflad:
This has been released in [version 2.1.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#210-march-07-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #7798, opened 2019-03-03
provider: Remove direct dependency on github.com/davecgh/go-spew

bflad:
Updated via:

```
go mod tidy
go mod vendor
```

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -174,7 +174,7 @@ github.com/blang/semver
 github.com/boombuler/barcode
 github.com/boombuler/barcode/qr
 github.com/boombuler/barcode/utils
-# github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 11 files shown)


vendor/modules.txt
```diff
@@ -174,7 +174,7 @@ github.com/blang/semver
 github.com/boombuler/barcode
 github.com/boombuler/barcode/qr
 github.com/boombuler/barcode/utils
-# github.com/davecgh/go-spew v1.1.1
+# github.com/davecgh/go-spew v1.1.0
```
tombuildsstuff: I think these want to be removed?

bflad: Its still an indirect dependency unfortunately. Deep in:

```
aws/aws-sdk-go
jmespath/go-jmespath
davecgh/go-spew
```

PR #7907, opened 2019-03-13
resource/aws_kms_alias: Prevent state removal of resource immediately after creation due to eventual consistency

### LINKED ISSUE #6560: Race condition when creating kms key alias leads to 'Resource x does not have attribute y for variable'
### Community Note

### Terraform Version
* 0.11.3

### Aws Provider Version
* 1.40.0

### Affected Resource(s)

* aws_kms_alias

### Debug Output
* https://gist.github.com/AndreuPlata/e336dd0bcd47e2349aa1d7eae14293b4

### Expected Behaviour

Terraform should have waited for the kms alias to be created

### Actual Behaviour

* Terraform sent a request to create the alias
* it immediately after sent a get aliases request to find the created kms alias
* it did not find it (potential aws delay/eventual consistency)
* removed the alias from the state file which broke all future references to this resource

### Steps to Reproduce

This is hard to reproduce as it is not consistently happening. We believe there is a race condition between terraform requesting the list of aliases and aws confirming alias creation

* Create a kms key and alias
* Have an output referencing that alias.name and alias.arn
* Use that output in a different resource

### LINKED ISSUE #7891: KMS Key Aliases are created but state is not always updated
_This issue was originally opened by @tolidano as hashicorp/terraform#20640. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
```
0.11.7
```

```hcl

resource "aws_kms_key" "key" {
  description             = "${var.description}"
  policy                  = "${data.aws_iam_policy_document.key_policy.json}"
  deletion_window_in_days = "30"
  is_enabled              = "true"
  enable_key_rotation     = "true"

  tags {
    description = "${var.description}"
    cost        = "${var.env == "production" ? "production" : "non-production"}"
    name        = "${var.alias_name}"
    managed     = "Terraform"
  }
}

resource "aws_kms_alias" "key_alias" {
  name          = "alias/${var.alias_name}"
  target_key_id = "${aws_kms_key.key.id}"
}
```

### Debug Output
N/A

### Crash Output
N/A

### Expected Behavior
When I apply the above, it should create a key and an alias and properly update the state in S3

### Actual Behavior
The key and alias were created, but the state was only updated with the key. A subsequent plan against the same environment attempted to recreate the key alias, which resulted in an AWS exception stating the alias already existed (it did).

### Steps to Reproduce
Over multiple environments, the first run (in QA) worked as expected, but the second run (in production) did not. The key and alias were created, but the state was only updated with the key. A subsequent plan against the same environment attempted to recreate the key alias, which resulted in an AWS exception stating the alias already existed (it did). Import is non-trivial, so I deleted the resource in AWS directly (aws kms delete-alias) and ran the apply again, and this time it worked properly (as it did in QA)

### Additional Context
We are using Terragrunt, but that would not explain the erratic behavior.

### References
None found.

bflad:
### Community Note

Closes #7891
Closes #6560

References:

* https://github.com/terraform-providers/terraform-provider-aws/issues/7873
* https://github.com/hashicorp/terraform/pull/17220

The KMS service has eventual consistency considerations and the `aws_kms_alias` resource immediately tries to read the KMS alias after creation, which may not find the KMS alias. When not able to find the KMS alias, the resource logic returns an empty API object instead of an error. Since a `nil` check was already performed on the error, the error will always be `nil`. Invoking `return resource.RetryableError(nil)`  is equivalent to `return nil`. The resource during its Read performs an error check first which will skip because its `nil`, then assumes the resource has been deleted outside Terraform and triggers recreation.

Here when we cannot find a KMS alias after allowing some time for eventual consistency, we return a resource not found error and ensure we handle any timeouts due to automatic AWS Go SDK retries.

```
--- PASS: TestAccAWSKmsAlias_no_name (37.63s)
--- PASS: TestAccAWSKmsAlias_name_prefix (37.80s)
--- PASS: TestAccAWSKmsAlias_multiple (38.38s)
--- PASS: TestAccAWSKmsAlias_importBasic (40.13s)
--- PASS: TestAccAWSKmsAlias_ArnDiffSuppress (43.61s)
--- PASS: TestAccAWSKmsAlias_basic (46.76s)
```

### DIFF
--- a/aws/resource_aws_kms_alias.go
+++ b/aws/resource_aws_kms_alias.go
@@ -182,10 +182,15 @@ func retryFindKmsAliasByName(conn *kms.KMS, name string) (*kms.AliasListEntry, e
 			return resource.NonRetryableError(err)
 		}
 		if resp == nil {
-			return resource.RetryableError(err)
+			return resource.RetryableError(&resource.NotFoundError{})
 		}
 		return nil
 	})
+
+	if isResourceTimeoutError(err) {
+		resp, err = findKmsAliasByName(conn, name, nil)
+	}
+
 	return resp, err
 }
 


bflad:
This has been released in [version 2.2.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#220-march-15-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #8001, opened 2019-03-19
data-source/aws_instance: Add get_user_data argument and user_data_base64 attribute

### LINKED ISSUE #5927: aws_instance data source should expose "user_data" content
### Community Note

### Description

aws_instance data source exposes `user_data` only as hash. There is no option to view/save/use original script. This seems inconsistent with original resource meaning of `user_data`

* data source aws_instance

### Proposed solution

Apart from `user_data` attribute we could expose `user_data_script` or `user_data_content`.
The original field name should be called `user_data_hash` IMHO but that would be a breaking change, so probably not worth it.

bflad:
### Community Note

Closes #5927

As contents may be large and not necessarily safe for the Terraform state, which requires valid UTF-8, we allow retrieving the Base64 contents of EC2 Instance user data into the `user_data_base64` attribute when a new `get_user_data` argument is set to `true`. The SHA-1 hash of the EC2 Instance user data will continue to be set in the `user_data` attribute regardless of the `get_user_data` argument.

```
--- PASS: TestAccAWSInstanceDataSource_basic (149.59s)
--- PASS: TestAccAWSInstanceDataSource_GetUserData (158.67s)
```

### DIFF
--- a/aws/data_source_aws_instance_test.go
+++ b/aws/data_source_aws_instance_test.go
@@ -22,6 +22,7 @@ func TestAccAWSInstanceDataSource_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("data.aws_instance.web-instance", "tags.%", "1"),
 					resource.TestCheckResourceAttr("data.aws_instance.web-instance", "instance_type", "m1.small"),
 					resource.TestMatchResourceAttr("data.aws_instance.web-instance", "arn", regexp.MustCompile(`^arn:[^:]+:ec2:[^:]+:\d{12}:instance/i-.+`)),
+					resource.TestCheckNoResourceAttr("data.aws_instance.web-instance", "user_data_base64"),
 				),
 			},
 		},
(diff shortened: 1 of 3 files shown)


aws/data_source_aws_instance_test.go
```diff
@@ -22,6 +22,7 @@ func TestAccAWSInstanceDataSource_basic(t *testing.T) {
 					resource.TestCheckResourceAttr("data.aws_instance.web-instance", "tags.%", "1"),
 					resource.TestCheckResourceAttr("data.aws_instance.web-instance", "instance_type", "m1.small"),
 					resource.TestMatchResourceAttr("data.aws_instance.web-instance", "arn", regexp.MustCompile(`^arn:[^:]+:ec2:[^:]+:\d{12}:instance/i-.+`)),
+					resource.TestCheckNoResourceAttr("data.aws_instance.web-instance", "user_data_base64"),
```
nywilken: Does it makes sense to add a test case for the case where `get_user_data = true`? I would expect the results to be the same since there is no actual user data provided.

bflad: Added covering test 👍 

```
--- PASS: TestAccAWSInstanceDataSource_GetUserData_NoUserData (210.42s)
```

bflad:
This has been released in [version 2.3.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#230-march-21-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #8078, opened 2019-03-26
resource/aws_emr_instance_group: add autoscaling_policy and bid_price arguments, along with resource importing support

nywilken:
This change updates `aws_emr_instance_group` resource so that it has full parity with the deprecated `instance_group` configuration block in the `aws_emr_cluster` resource. 

```
--- PASS: TestAccAWSEMRInstanceGroup_basic (559.04s)
--- PASS: TestAccAWSEMRInstanceGroup_AutoScalingPolicy (547.62s)
--- PASS: TestAccAWSEMRInstanceGroup_EbsConfig_EbsOptimized (567.56s)
--- PASS: TestAccAWSEMRInstanceGroup_BidPrice (601.52s)
--- PASS: TestAccAWSEMRInstanceGroup_InstanceCount (626.62s)

```


aws/resource_aws_emr_cluster.go
```diff
@@ -1236,10 +1243,6 @@ func flattenInstanceGroup(ig *emr.InstanceGroup) (map[string]interface{}, error)
 		attrs["autoscaling_policy"] = ""
 	}
 
-	if attrs["name"] != nil {
```
bflad: Noting this looks okay since its a duplicate of earlier logic 👍

aws/resource_aws_emr_instance_group.go
```diff
@@ -1,32 +1,73 @@
...
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/emr"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/structure"
+	"github.com/hashicorp/terraform/helper/validation"
 )
 
-var emrInstanceGroupNotFound = errors.New("No matching EMR Instance Group")
+var errEMRInstanceGroupNotFound = errors.New("No matching EMR Instance Group")
```
bflad: Nit: We can use the upstream [`resource.NotFoundError`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#NotFoundError) and `isResourceNotFoundError()` helper instead of our own custom error here.

aws/resource_aws_emr_instance_group.go
```diff
@@ -1,32 +1,73 @@
...
+				return []*schema.ResourceData{d}, nil
+			},
+		},
 		Schema: map[string]*schema.Schema{
+			"autoscaling_policy": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				DiffSuppressFunc: suppressEquivalentJsonDiffs,
+				ValidateFunc:     validation.ValidateJsonString,
+				StateFunc: func(v interface{}) string {
```
bflad: Nit: It'd be preferred to remove this `StateFunc` if possible. 👍 The `DiffSuppressFunc` and `Read` function should cover its usage.

aws/resource_aws_emr_instance_group.go
```diff
@@ -1,32 +1,73 @@
...
+					s, _ := structure.NormalizeJsonString(v.(string))
+					return s
+				},
+			},
+			"bid_price": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
bflad: Not necessarily for this pull request, but if interested, we could potentially make this field optionally `ForceNew` via `CustomizeDiff`, e.g.

```go
CustomizeDiff: customdiff.Sequence(
	customdiff.ForceNewIfChange("bid_price", func(old, new, meta interface{}) bool {
		return old == "" || new == ""
	}),
),
```

aws/resource_aws_emr_instance_group.go
```diff
@@ -1,32 +1,73 @@
...
 				ForceNew: true,
 			},
+			"instance_role": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  "TASK",
+				ValidateFunc: validation.StringInSlice([]string{
+					"CORE",
+					"MASTER",
```
bflad: I'd recommend removing `MASTER` from this list 👍

aws/resource_aws_emr_instance_group.go
```diff
@@ -1,32 +1,73 @@
...
 			"cluster_id": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
 			},
+			"instance_role": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				Default:  "TASK",
```
bflad: Nit: We should prefer to use the [SDK provided constants](https://docs.aws.amazon.com/sdk-for-go/api/service/emr/#pkg-constants) in the `Default` here and `ValidateFunc` below, e.g.

```suggestion
				Default:  emr.InstanceGroupTypeTask,
```

aws/resource_aws_emr_instance_group.go
```diff
@@ -85,197 +126,196 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
+		autoscalingPolicyRulesBytes, err := normalizeEmptyRules(ig.AutoScalingPolicy.Rules)
+		if err != nil {
+			return fmt.Errorf("error parsing EMR Cluster Instance Groups AutoScalingPolicy Rules: %s", err)
+		}
+
+		autoscalingPolicyString = fmt.Sprintf("{\"Constraints\":%s,\"Rules\":%s}", string(autoscalingPolicyConstraintsBytes), string(autoscalingPolicyRulesBytes))
+	}
+	d.Set("autoscaling_policy", autoscalingPolicyString)
+
+	if ig.BidPrice != nil {
```
bflad: Nit: `d.Set()` automatically handles `nil` values so the check is extraneous 👍

aws/resource_aws_emr_instance_group.go
```diff
@@ -85,197 +126,196 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
+					InstanceCount:   aws.Int64(int64(instanceCount)),
+				},
+			},
 		}
-		if respGrps == nil {
-			return nil, fmt.Errorf("Error reading EMR Instance Group for cluster (%s)", clusterId)
+
+		_, err := conn.ModifyInstanceGroups(params)
+		if err != nil {
+			return err
```
bflad: We should return error context here for operators and code maintainers:

```suggestion
			return fmt.Errorf("error modifying EMR Instance Group (%s): %s", d.Id(), err)
```

aws/resource_aws_emr_instance_group.go
```diff
@@ -85,197 +126,196 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
 		}
 
-		if respGrps.InstanceGroups != nil {
-			groups = append(groups, respGrps.InstanceGroups...)
-		} else {
-			log.Printf("[DEBUG] EMR Instance Group list was empty")
+		stateConf := &resource.StateChangeConf{
+			Pending:    []string{"PROVISIONING", "BOOTSTRAPPING", "RESIZING"},
```
bflad: We can use SDK constants here as well 🎉 

```suggestion
			Pending:    []string{
				emr.InstanceGroupStateBootstrapping,
				emr.InstanceGroupStateProvisioning,
				emr.InstanceGroupStateResizing,
			},
```

aws/resource_aws_emr_instance_group.go
```diff
@@ -85,197 +126,196 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
-	}
-
-	_, err = stateConf.WaitForState()
-	if err != nil {
-		return fmt.Errorf(
-			"Error waiting for instance (%s) to terminate: %s", d.Id(), err)
-	}
-
-	return resourceAwsEMRInstanceGroupRead(d, meta)
+	return err
```
bflad: We should return error context here for operators and code maintainers:

```suggestion
	if err != nil {
		return fmt.Errorf("error draining EMR Instance Group (%s): %s", d.Id(), err)
	}
	
	return nil
```

aws/resource_aws_emr_instance_group.go
```diff
@@ -294,18 +334,88 @@ func instanceGroupStateRefresh(conn *emr.EMR, clusterID, igID string) resource.S
...
+func fetchAllEMRInstanceGroups(conn *emr.EMR, clusterID string) ([]*emr.InstanceGroup, error) {
+	req := &emr.ListInstanceGroupsInput{
+		ClusterId: aws.String(clusterID),
+	}
+
+	var groups []*emr.InstanceGroup
+	marker := aws.String("intitial")
+	for marker != nil {
+		log.Printf("[DEBUG] EMR Cluster Instance Marker: %s", *marker)
+		respGrps, errGrps := conn.ListInstanceGroups(req)
```
bflad: We should prefer to use the [SDK provider paginated function `ListInstanceGroupsPages()`](https://docs.aws.amazon.com/sdk-for-go/api/service/emr/#EMR.ListInstanceGroupsPages) instead of implementing our own, e.g.

```go
input := &emr.ListInstanceGroupsInput{
	ClusterId: aws.String(clusterID),
}
var groups []*emr.InstanceGroup

err := conn.ListInstanceGroupsPages(input, func(page *emr.ListInstanceGroupsOutput, lastPage bool) {
	groups = append(groups, page.InstanceGroups...)

	return !lastPage
})

return groups, err
```

aws/resource_aws_emr_instance_group.go
```diff
@@ -294,18 +334,88 @@ func instanceGroupStateRefresh(conn *emr.EMR, clusterID, igID string) resource.S
...
+
+	if len(groups) == 0 {
+		return nil, fmt.Errorf("No instance groups found for EMR Cluster (%s)", clusterID)
+	}
+
+	return groups, nil
+}
+
+func fetchEMRInstanceGroup(conn *emr.EMR, clusterID, groupID string) (*emr.InstanceGroup, error) {
+	// Is this needed or can we consolidate this a bit?
```
bflad: 😛 You tell us!

aws/resource_aws_emr_instance_group.go
```diff
@@ -294,18 +334,88 @@ func instanceGroupStateRefresh(conn *emr.EMR, clusterID, igID string) resource.S
...
+
+	groups, err := fetchAllEMRInstanceGroups(conn, clusterID)
+	if err != nil {
+		return nil, err
+	}
+
+	var group *emr.InstanceGroup
+	for _, ig := range groups {
+		if groupID == *ig.Id {
```
bflad: To prevent potential panics, we should use the [SDK provided pointer conversion functions](https://docs.aws.amazon.com/sdk-for-go/api/aws/), e.g.

```suggestion
		if groupID == aws.StringValue(ig.Id) {
```

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -129,46 +188,23 @@ func testAccCheckAWSEmrInstanceGroupExists(n string, v *emr.InstanceGroup) resou
 	}
 }
 
+func testAccAWSEMRInstanceGroupResourceImportStateIdFunc(resourceName string) resource.ImportStateIdFunc {
```
bflad: 💯

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -129,46 +188,23 @@ func testAccCheckAWSEmrInstanceGroupExists(n string, v *emr.InstanceGroup) resou
...
+	return func(s *terraform.State) (string, error) {
+		rs, ok := s.RootModule().Resources[resourceName]
+		if !ok {
+			return "", fmt.Errorf("Not found: %s", resourceName)
+		}
+
+		log.Println(fmt.Sprintf("%s/%s", rs.Primary.Attributes["cluster_id"], rs.Primary.ID))
```
bflad: Extraneous log line can be removed 👍

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -129,46 +188,23 @@ func testAccCheckAWSEmrInstanceGroupExists(n string, v *emr.InstanceGroup) resou
...
+		return fmt.Sprintf("%s/%s", rs.Primary.Attributes["cluster_id"], rs.Primary.ID), nil
+	}
+}
+
 const testAccAWSEmrInstanceGroupBase = `
 provider "aws" {
```
bflad: Not caused by this pull request, but most provider acceptance test configurations should omit `provider` configurations (unless explicit multi-region testing)

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -231,11 +267,49 @@ resource "aws_main_route_table_association" "a" {
...
+    emr_managed_master_security_group = "${aws_security_group.allow_all.id}"
+    emr_managed_slave_security_group  = "${aws_security_group.allow_all.id}"
+    instance_profile                  = "${aws_iam_instance_profile.emr_profile.arn}"
+  }
+
+  master_instance_type = "m4.large"
+  core_instance_type = "m4.large"
+  core_instance_count = 2
+
+  tags = {
```
bflad: As a general practice, I would recommend removing any optional arguments that are not pertinent to the acceptance test to reduce the test configuration size.

aws/resource_aws_emr_instance_group_test.go
```diff
@@ -396,39 +470,108 @@ resource "aws_iam_policy" "iam_emr_profile_policy" {
...
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonElasticMapReduceforAutoScalingRole"
+}
 `
 
-func testAccAWSEmrInstanceGroupConfig(r int) string {
+func testAccAWSEmrInstanceGroupConfig_basic(r int) string {
 	return fmt.Sprintf(testAccAWSEmrInstanceGroupBase+`
 	resource "aws_emr_instance_group" "task" {
     cluster_id     = "${aws_emr_cluster.tf-test-cluster.id}"
+    instance_role  = "TASK"
```
bflad: For `basic` acceptance tests we should omit optional arguments and ensure they have the correct default 👍

website/docs/r/emr_instance_group.html.markdown
```diff
@@ -32,10 +32,14 @@ The following arguments are supported:
 
 * `name` (Required) Human friendly name given to the instance group. Changing this forces a new resource to be created.
 * `cluster_id` (Required) ID of the EMR Cluster to attach to. Changing this forces a new resource to be created.
+* `instance_role` (Optional) The EC2 instance role for all instances in the instance group. Valid values are MASTER, CORE, TASK. Default `"TASK"`
 * `instance_type` (Required) The EC2 instance type for all instances in the instance group. Changing this forces a new resource to be created.
+* `bid_price` - (Optional) If set, the bid price for each EC2 instance in the instance group, expressed in USD. By setting this attribute, the instance group is being declared as a Spot Instance, and will implicitly create a Spot request. Leave this blank to use On-Demand Instances. `bid_price` can not be set for the `MASTER` instance group, since that group must always be On-Demand
```
bflad: If we won't support `MASTER` instance groups, the extra verbiage can be removed here and above. 👍

aws/resource_aws_emr_cluster.go
nywilken: :+1: :100:

aws/resource_aws_emr_instance_group.go
nywilken: Thanks for providing this snippet. I went ahead and implemented it in this pull request. I am writing test against it now so I should have something new for you to review.

nywilken:
@bflad thanks for the thorough review. I went ahead and implemented the changes and suggestions you called out; making a bit of a refactor to consolidate logic into smaller non shared functions  (speaking abut cluster and instance_group) and move things out of structure.go to start. I'll add a few more test and remove the WIP label from the title shortly.

bflad:
@nywilken I just merged the `aws_emr_cluster` resource pull request. Ping me when this is ready for another review and I'll give it another lookover. 👍 If you have the bid price failure, I'd love to see the output. Thanks!

aws/resource_aws_emr_instance_group.go
```diff
@@ -58,254 +58,369 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
+				{
+					InstanceGroupId: aws.String(d.Id()),
+					InstanceCount:   aws.Int64(int64(instanceCount)),
+				},
+			},
 		}
-		if respGrps == nil {
-			return nil, fmt.Errorf("Error reading EMR Instance Group for cluster (%s)", clusterId)
+
+		// Give me some EBS config stuff.
```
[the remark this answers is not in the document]
nywilken: No longer needed. EBS config stuff has been handled. Good catch :+1:

aws/resource_aws_emr_instance_group.go
```diff
@@ -58,254 +58,369 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
+		if err != nil {
+			return fmt.Errorf("error parsing EMR Cluster Instance Groups AutoScalingPolicy Rules: %s", err)
+		}
+
+		autoscalingPolicyString = fmt.Sprintf("{\"Constraints\":%s,\"Rules\":%s}", string(autoscalingPolicyConstraintsBytes), string(autoscalingPolicyRulesBytes))
+	}
+	d.Set("autoscaling_policy", autoscalingPolicyString)
+
+	d.Set("bid_price", ig.BidPrice)
+	d.Set("ebs_config", flattenEBSConfig(ig.EbsBlockDevices))
```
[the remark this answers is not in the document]
nywilken: :+1:

aws/resource_aws_emr_instance_group.go
```diff
@@ -58,254 +58,369 @@ func resourceAwsEMRInstanceGroup() *schema.Resource {
...
+		return fmt.Errorf("error reading EMR Instance Group (%s): %s", d.Id(), err)
+	}
+
+	var autoscalingPolicyString string
+	if ig.AutoScalingPolicy != nil {
+		// AutoScalingPolicy has an additional Status field and null values that are causing a new hashcode to be generated for `instance_group`.
+		// We are purposefully omitting that field and the null values here when we flatten the autoscaling policy string for the statefile.
+		for i, rule := range ig.AutoScalingPolicy.Rules {
+			for j, dimension := range rule.Trigger.CloudWatchAlarmDefinition.Dimensions {
+				if *dimension.Key == "JobFlowId" {
```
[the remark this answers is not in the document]
nywilken: Good catch :+1:

bflad:
This has been released in [version 2.11.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2110-may-17-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #8092, opened 2019-03-27
docs/provider: Split out endpoint customization into a new Custom Service Endpoints Guide

bflad:
### Community Note

We will be allowing customization for all endpoints soon and the AWS Provider index page is getting quite long with little benefit. Here we split the information into its own guide so we can further provide more helpful information about working in these specialized environments.


bflad:

PR #8099, opened 2019-03-27
Doc update: aws_codecommit_repository does not support default_branch

jwatson0:
The website docs list 'default_branch' as an attribute of aws_codecommit_repository, but this attribute is not in the current code.

```
module.codecommit.output.default_branch: Resource
'aws_codecommit_repository.repo' does not have attribute 
'default_branch' for variable 'aws_codecommit_repository.repo.default_branch'
```

See also: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/data_source_aws_codecommit_repository.go

### Community Note

### DIFF
--- a/website/docs/r/codecommit_repository.html.markdown
+++ b/website/docs/r/codecommit_repository.html.markdown
@@ -29,7 +29,6 @@ The following arguments are supported:
 
 * `repository_name` - (Required) The name for the repository. This needs to be less than 100 characters.
 * `description` - (Optional) The description of the repository. This needs to be less than 1000 characters
-* `default_branch` - (Optional) The default branch of the repository. The branch specified here needs to exist.
 
 ## Attributes Reference
 


bflad:
Hi @jwatson0 👋 Thanks for submitting this. The code you link to is for the data source while the change is for the resource documentation. The resource does indeed support the `default_branch` argument:

https://github.com/terraform-providers/terraform-provider-aws/blob/964e6be60f53e594f833e46c2d993e2cd77028f7/aws/resource_aws_codecommit_repository.go#L57-L60

And is acceptance tested here:

https://github.com/terraform-providers/terraform-provider-aws/blob/964e6be60f53e594f833e46c2d993e2cd77028f7/aws/resource_aws_codecommit_repository_test.go#L80-L124

What you might be seeing instead is that the resource is only making the `default_branch` attribute available if it was specified, which may be an implementation bug:

https://github.com/terraform-providers/terraform-provider-aws/blob/964e6be60f53e594f833e46c2d993e2cd77028f7/aws/resource_aws_codecommit_repository.go#L132-L136

And noted in the testing here with the `...NoResourceAttr`:

https://github.com/terraform-providers/terraform-provider-aws/blob/964e6be60f53e594f833e46c2d993e2cd77028f7/aws/resource_aws_codecommit_repository_test.go#L110-L111

So that said, we likely cannot accept this change since the argument is valid, but indeed the behavior is currently confusing and probably warrants a resource code fix.

bflad:
Closing given the previous comment.

PR #8116, opened 2019-03-28
resource/aws_cloudfront_distribution: Add wait_for_deployment argument

### LINKED ISSUE #8073: Release 2.1.0 (commit #7794) introduces serious and problematic performance degradation for CloudFront Distribution updates
Provisioner release 2.1.0 (#7794 appears to be the relevant commit related to CloudFront distributions) introduces a significant workflow problem with timing for CloudFront Distribution updates.

Prior to this change, updating an origin path, for example, took seconds to complete. Now, however, the TF apply sits and waits until propagation throughout the CloudFront edge network completes.  This in turn destroys automated workflows. 

Prior to 2.1.0, a CloudFront Distribution update (for example, an origin path change), the portion of a TF apply related to making and registering the change in distribution configuration took, on average, 2 seconds to complete.  Beginning with release 2.1.0, the same CloudFront distribution update time increases to an average of 20 plus minutes, with peak times of over an hour and 20 plus minutes.

It would be nice if a significant change like this included a boolean attribute added (comparable to the 'retain_on_destroy' boolean, but for updates) to allow users to _choose_ whether or not to wait for distribution propagation, or to move on as soon as the change is reflected, as was the behavior prior to release 2.1.0 (#7794).

bflad:
### Community Note

Closes #8073

Since we are adding a virtual attribute with a `Default`, we must include a schema state migration to prevent the `"" => "true"` difference on upgrade.

```
--- PASS: TestAccAWSCloudFrontDistribution_Origin_EmptyDomainName (1.25s)
--- PASS: TestAccAWSCloudFrontDistribution_Origin_EmptyOriginID (1.30s)
--- PASS: TestAccAWSCloudFrontDistribution_disappears (571.01s)
--- PASS: TestAccAWSCloudFrontDistribution_noOptionalItemsConfig (1263.05s)
--- PASS: TestAccAWSCloudFrontDistribution_noCustomErrorResponseConfig (1265.06s)
--- PASS: TestAccAWSCloudFrontDistribution_HTTP11Config (1265.34s)
--- PASS: TestAccAWSCloudFrontDistribution_customOrigin (1267.09s)
--- PASS: TestAccAWSCloudFrontDistribution_OriginGroups (1267.68s)
--- PASS: TestAccAWSCloudFrontDistribution_RetainOnDelete (1268.95s)
--- PASS: TestAccAWSCloudFrontDistribution_orderedCacheBehavior (1271.12s)
--- PASS: TestAccAWSCloudFrontDistribution_S3Origin (1271.30s)
--- PASS: TestAccAWSCloudFrontDistribution_multiOrigin (1271.50s)
--- PASS: TestAccAWSCloudFrontDistribution_IsIPV6EnabledConfig (1272.82s)
--- PASS: TestAccAWSCloudFrontDistribution_ViewerCertificate_AcmCertificateArn (1275.16s)
--- PASS: TestAccAWSCloudFrontDistribution_WaitForDeployment (1275.81s)
--- PASS: TestAccAWSCloudFrontDistribution_ViewerCertificate_AcmCertificateArn_ConflictsWithCloudFrontDefaultCertificate (1372.94s)
--- PASS: TestAccAWSCloudFrontDistribution_S3OriginWithTags (1782.07s)
--- PASS: TestAccAWSCloudFrontDistribution_Enabled (1782.38s)
```

### DIFF
--- a/aws/resource_aws_cloudfront_distribution_migrate_test.go
+++ b/aws/resource_aws_cloudfront_distribution_migrate_test.go
@@ -1,0 +1,52 @@
+package aws
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAwsCloudFrontDistributionMigrateState(t *testing.T) {
+	cases := map[string]struct {
+		StateVersion int
+		Attributes   map[string]string
+		Expected     map[string]string
+		Meta         interface{}
+	}{
+		"v0_1": {
+			StateVersion: 0,
+			Attributes: map[string]string{
+				"wait_for_deployment": "",
+			},
+			Expected: map[string]string{
+				"wait_for_deployment": "true",
+			},
+		},
+	}
+
+	for tn, tc := range cases {
+		is := &terraform.InstanceState{
+			ID:         "some_id",
+			Attributes: tc.Attributes,
+		}
+
+		tfResource := resourceAwsCloudFrontDistribution()
+
+		if tfResource.MigrateState == nil {
+			t.Fatalf("bad: %s, err: missing MigrateState function in resource", tn)
+		}
+
+		is, err := tfResource.MigrateState(tc.StateVersion, is, tc.Meta)
+		if err != nil {
+			t.Fatalf("bad: %s, err: %#v", tn, err)
+		}
+
+		for k, v := range tc.Expected {
+			if is.Attributes[k] != v {
+				t.Fatalf(
+					"bad: %s\n\n expected: %#v -> %#v\n got: %#v -> %#v\n in: %#v",
+					tn, k, v, k, is.Attributes[k], is.Attributes)
+			}
+		}
+	}
+}
--- a/aws/resource_aws_cloudfront_distribution_migrate_test.go
+++ b/aws/resource_aws_cloudfront_distribution_migrate_test.go
@@ -7,13 +7,13 @@ import (
 )
 
 func TestAwsCloudFrontDistributionMigrateState(t *testing.T) {
-	cases := map[string]struct {
+	testCases := map[string]struct {
 		StateVersion int
 		Attributes   map[string]string
 		Expected     map[string]string
 		Meta         interface{}
 	}{
-		"v0_1": {
+		"v0_to_v1": {
 			StateVersion: 0,
 			Attributes: map[string]string{
 				"wait_for_deployment": "",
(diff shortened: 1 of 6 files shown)


aws/resource_aws_cloudfront_distribution_migrate_test.go
```diff
@@ -0,0 +1,52 @@
...
+			Attributes: map[string]string{
+				"wait_for_deployment": "",
+			},
+			Expected: map[string]string{
+				"wait_for_deployment": "true",
+			},
+		},
+	}
+
+	for tn, tc := range cases {
```
nywilken: The variable names are a bit confusing here. I would stick with the std. i for index so that we don't have to think what it means sometime in the future. I would apply this thinking to any other variable that makes you think for more than two seconds.

website/docs/r/cloudfront_distribution.html.markdown
```diff
@@ -255,6 +255,10 @@ of several sub-resources - these resources are laid out below.
     deleting it when destroying the resource through Terraform. If this is set,
     the distribution needs to be deleted manually afterwards. Default: `false`.
 
+  * `wait_for_deployment` (Optional) - By default, the resource will wait
+    for creation and updates to fully deploy. Setting this to `false`
+    will skip waiting for deployments to complete. Default: `true`.
```
Nowaker: @bflad I would suggest this to read instead so it's clear what's happening here:

If enabled, the resource will wait for the distribution status to change from `InProgress` to `Deployed`.
Setting this to `false` will skip the process. Default: `true`.

aws/resource_aws_cloudfront_distribution_migrate_test.go
bflad: 👍 will update these. We should consider updating the rest of the codebase too as this was mostly copied from elsewhere.

website/docs/r/cloudfront_distribution.html.markdown
bflad: 👍 great suggestion, will update.

nathanielks:
Thanks for giving us the option to skip! This is great 👍

DJAlPee:
Why is default behavior `true` and not `false`?

Before introducing this new feature, the default behavior was not waiting for the complete deployment to be finished. To stick with this behavior, I now have to change every configuration, where I use cloudfront... :(

bflad:
Hi @DJAlPee 👋 Sorry for the hassle as CloudFront deployments are _particularly_ long as compared with many other AWS resources.

A design philosophy with most Terraform resources is that they should be designed to wait until they have completed any creation/update/deletion action. This is so any downstream Terraform configuration or post-Terraform actions that are intending to interact with the infrastructure does so when the infrastructure is in a stable state. Where this waiting behavior does not happen we typically consider that to be a bug in the Terraform resource and do not consider it a breaking change as we were not previously performing an expected Terraform resource behavior.

If you have suggestions where we might be able to document these concepts better, we would gladly take a look, just please submit a new GitHub issue so it doesn't get lost in a previously closed issue/pull request.

bflad:

PR #8188, opened 2019-04-04
Add options and fix names for EKS example.

squaresurf:
Why:

* It is beneficial to be able to select the size of your autoscaling
  group.
* The names/tags should be built off of the cluster name variable.

This change addresses the need by:

* Add the desired-workers variable
* Add the max-workers variable
* Add the min-workers variable
* Replace static "terraform-eks-demo" values with "${var.cluster-name}"

Side effects:

* The example may not line up anymore with the blog post.

### Community Note

Fixes #0000

* Change 1
* Change 2

```

...
```

### DIFF
--- a/examples/eks-getting-started/eks-cluster.tf
+++ b/examples/eks-getting-started/eks-cluster.tf
@@ -6,7 +6,7 @@
 #
 
 resource "aws_iam_role" "demo-cluster" {
-  name = "terraform-eks-demo-cluster"
+  name = "${var.cluster-name}-cluster"
 
   assume_role_policy = <<POLICY
 {
@@ -35,7 +35,7 @@ resource "aws_iam_role_policy_attachment" "demo-cluster-AmazonEKSServicePolicy" 
 }
 
 resource "aws_security_group" "demo-cluster" {
-  name        = "terraform-eks-demo-cluster"
+  name        = "${var.cluster-name}-cluster"
   description = "Cluster communication with worker nodes"
   vpc_id      = "${aws_vpc.demo.id}"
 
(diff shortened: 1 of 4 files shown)


bflad:
Hi @squaresurf 👋 Thanks for submitting this and putting some abstraction handling Terraform configuration concepts into there. For operators looking to take the example configuration as-is into their environments and potentially scale this configuration by building their own Terraform module, this would likely be a next good step.

The challenge we have here is that this example code is currently intended to be nearly synchronized with the [AWS EKS Introduction Guide](https://learn.hashicorp.com/terraform/aws/eks-intro) configurations so operators can cross reference the fully built code with steps of the guide to get a full picture of how everything might come together while not overburdening with abstractions. The example configuration may be a great starting point for understanding the various components associated with the service, but we don't intend it to necessarily be "production ready" or reusable as many operators will have differing opinions on those requirements.

We likely need to call this out a little better in the guide and example code. They were written in preparation of the EKS service launch day last year and since then the community has made great strides in writing more robust variations of this configuration and published them in public Terraform Registry: https://registry.terraform.io/search?provider=aws&q=eks&verified=false

So all this said, we likely are a little hesitant to introduce these abstractions into this particular example code even though they are sound enhancement to improve the reusability of the code, since they fall outside the scope of the basics of getting started with the EKS service.

If you are still interested in contributing though, we would be more than happy to accept simple improvements along with synchronized changes to the guide source, which is located at `website/docs/guides/eks-getting-started.html.md`. We would also consider accepting a separate example directory that does not have the same "getting started" restrictions with the guide, however we generally encourage the publishing of Terraform Modules and to that end it might be nice properly give a shout out in the guide and example README to the Terraform modules that have been published in the Terraform Registry.

squaresurf:
Thank you for your thoughtful response @bflad. I had a suspicion that would be the case. I decided to take my edits and start a new repo with the same licensing applied: https://github.com/squaresurf/quick-eks-cluster

I made sure to give credit this example. My repo also adds some other niceties in getting an eks cluster really quickly. Let me know what you think :)

PR #8194, opened 2019-04-04
resource/aws_route_table: Set SchemaConfigModeAttr on route configuration block attributes

bflad:
### Community Note

References:

* https://github.com/hashicorp/terraform/issues/20505
* https://github.com/hashicorp/terraform/pull/20626

In Terraform 0.12, behaviors with configuration blocks are more explicit to allow configuration language improvements and remove ambiguities that required various undocumented workarounds in Terraform 0.11 and prior. As a consequence, configuration blocks that are marked `Optional: true` and `Computed: true` will no longer support explicitly zero-ing out the configuration without special implementation.

The `ConfigMode` schema parameter on the configuration block attribute allows the schema to override certain behaviors. In particular, setting `ConfigMode: schema.SchemaConfigModeAttr` will allow practitioners to continue setting the configuration block attribute to an empty list (in the semantic sense, e.g. `attr = []`), keeping this prior behavior in Terraform 0.12. This parameter does not have any effect with Terraform 0.11. This solution may be temporary or revisited in the future.

Previous output from Terraform 0.11:

```
--- PASS: TestAccAWSRouteTable_Route_ConfigMode (66.53s)
```

Previous output from Terraform 0.12:

```
--- FAIL: TestAccAWSRouteTable_Route_ConfigMode (36.11s)
    testing.go:568: Step 2 error: config is invalid: Unsupported argument: An argument named "route" is not expected here. Did you mean to define a block of type "route"?
```

Output from Terraform 0.11:

```
--- PASS: TestAccAWSRouteTable_Route_ConfigMode (67.27s)
```

Output from Terraform 0.12:

```
--- PASS: TestAccAWSRouteTable_Route_ConfigMode (69.03s)
```

### DIFF
--- a/aws/resource_aws_route_table_test.go
+++ b/aws/resource_aws_route_table_test.go
@@ -270,6 +270,40 @@ func TestAccAWSRouteTable_panicEmptyRoute(t *testing.T) {
 	})
 }
 
+func TestAccAWSRouteTable_Route_ConfigMode(t *testing.T) {
+	var routeTable1, routeTable2, routeTable3 ec2.RouteTable
+	resourceName := "aws_route_table.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckRouteTableDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSRouteTableConfigRouteConfigModeBlocks(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouteTableExists(resourceName, &routeTable1),
+					resource.TestCheckResourceAttr(resourceName, "route.#", "2"),
+				),
+			},
+			{
+				Config: testAccAWSRouteTableConfigRouteConfigModeNoBlocks(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouteTableExists(resourceName, &routeTable2),
+					resource.TestCheckResourceAttr(resourceName, "route.#", "2"),
+				),
+			},
+			{
+				Config: testAccAWSRouteTableConfigRouteConfigModeZeroed(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouteTableExists(resourceName, &routeTable3),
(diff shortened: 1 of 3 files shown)


aws/resource_aws_route_table_test.go
```diff
@@ -270,6 +270,40 @@ func TestAccAWSRouteTable_panicEmptyRoute(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouteTableExists(resourceName, &routeTable1),
+					resource.TestCheckResourceAttr(resourceName, "route.#", "2"),
+				),
+			},
+			{
+				Config: testAccAWSRouteTableConfigRouteConfigModeNoBlocks(),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouteTableExists(resourceName, &routeTable2),
+					resource.TestCheckResourceAttr(resourceName, "route.#", "2"),
```
nywilken: Why is the count two here if there are no route attributes in the config?

bflad: This attribute is marked `Computed: true`: https://github.com/terraform-providers/terraform-provider-aws/blob/c5c8b5c69fd40fca3d4b9dfec6d2e546dc1783cb/aws/resource_aws_route_table.go#L46

For better or worse, the historical reason for this is because we support two mutually exclusive configuration styles for EC2 Route Table Routes:
* Inline `route` configuration blocks, each managing a Route, within the `aws_route_table` resource
* Individual Route managing [`aws_route` resources](https://www.terraform.io/docs/providers/aws/r/route.html), usually paired with modularized Terraform configuration

In the case of the latter, the `Computed: true` is there to prevent the need for configuring `ignore_changes`, e.g.

```hcl
# Example required configuration if Computed: true was not set for route attribute
resource "aws_route_table" "example" {
  # ... other configuration, except any route configuration blocks ...

  lifecycle {
    ignore_changes = "route"
  }
}

resource "aws_route" "example" {
  # ... other configuration ...

  route_table_id = "${aws_route_table.example.id}"
}
```

nywilken: Got it! I was looking at it as a separate test :man_facepalming:

 All good here and thanks for the backgound on the two resources `aws_route` and the inline route in `aws_route_table`. That was not at first obvious in reading the documentation for `aws_route_table`.

PR #7634, opened 2019-02-21
resource/aws_ram_resource_share: Add arn attribute

bflad:
While the ARN was already present in the `id` attribute, most resources that have an ARN follow the pattern of having an `arn` attribute.

Output from acceptance testing:

```
--- PASS: TestAccAwsRamResourceShare_basic (9.85s)
--- PASS: TestAccAwsRamResourceShare_Name (15.21s)
--- PASS: TestAccAwsRamResourceShare_AllowExternalPrincipals (15.60s)
--- PASS: TestAccAwsRamResourceShare_Tags (22.71s)
```

PR #8220, opened 2019-04-05
Update ram_resource_association.html.markdown

mrzenka:
### Community Note

Currently the documentation points to use aws_ram_resource_share.example.arn which is incorrect. The output of the aws_ram_resource_share.example is id which contains the ARN value.

### DIFF
--- a/website/docs/r/ram_resource_association.html.markdown
+++ b/website/docs/r/ram_resource_association.html.markdown
@@ -17,7 +17,7 @@ Manages a Resource Access Manager (RAM) Resource Association.
 ```hcl
 resource "aws_ram_resource_association" "example" {
   resource_arn       = "${aws_subnet.example.arn}"
-  resource_share_arn = "${aws_ram_resource_share.example.arn}"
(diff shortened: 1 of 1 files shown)


bflad:
Hi @mrzenka 👋 Thanks for submitting this. The `arn` attribute was added to the `aws_ram_resource_share` resource via #7634 and previously released in [version 2.1.0 of the AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#210-march-07-2019).

We show using the `arn` reference here for clarity against the `id` attribute, which someone skipping the documentation below and hardcoding a value might misconstrue as the shorted identifier. This documentation pattern is very consistent across resources even if the resource `id` attribute happens to contain the ARN and is only a problem here because there was a small gap between the resource creation and `arn` attribute existence. Since we do not have versioned documentation for Terraform Providers yet, the version requirement is not clear at the moment, although the number of environments with only the `id` attribute available will continue to decrease. We would prefer to keep using `arn` just for consistency with other documentation. Apologies for any confusion and thanks for raising this.

PR #8259, opened 2019-04-09
New resource: ram_resource_share_accepter

### LINKED ISSUE #7601: New Resource: aws_resource_share_accepter
### Community Note

### Description
Use AWS Resource Access Manager to share AWS resources between AWS accounts. To share a resource, you create a resource share, associate the resource with the resource share, and specify the principals that can access the resource. The principal must accept an invitation to be associated with the resource share. This last part, accepting the association invite, is the feature that is currently missing in the Terraform AWS provider.

1. Create a resource share (aws_ram_resource_share)
2. Associate target account (aws_ram_principal_association) - this exposed the ARN 
3. Accept the association (new resource required)

After creating a Resource Access Manager share (aws_ram_resource_share), other accounts/principals may be associated with the share. This association (aws_ram_principal_association) establishes an invite that must be accepted by the target account. aws_ram_principal_association.example.id exposes the ARN required for input to (3) aws_ram_principal_accepter.

* aws_ram_principal_accepter

```hcl
resource "aws_ram_principal_accepter" "example" {
  resource_share_invitation_arn = "${aws_ram_principal_association.example.id}"
}
```

### References
* https://github.com/terraform-providers/terraform-provider-aws/issues/6527 (closed)
* https://www.terraform.io/docs/providers/aws/r/ram_principal_association.html
* https://docs.aws.amazon.com/cli/latest/reference/ram/accept-resource-share-invitation.html
* https://aws.amazon.com/blogs/aws/new-aws-resource-access-manager-cross-account-resource-sharing/
* https://docs.aws.amazon.com/cli/latest/reference/ram/index.html#cli-aws-ram

YakDriver:
### Community Note

Fixes #7601 

* r/ram_resource_share_accepter: New resource allowing accepting AWS resource share invitations.

```console
$ make testacc TESTARGS='-run=TestAccAwsRamResourceShareAccepter_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsRamResourceShareAccepter_basic -timeout 120m
=== RUN   TestAccAwsRamResourceShareAccepter_basic
=== PAUSE TestAccAwsRamResourceShareAccepter_basic
=== CONT  TestAccAwsRamResourceShareAccepter_basic
--- PASS: TestAccAwsRamResourceShareAccepter_basic (34.00s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	34.071s
```

## Usage
Example usage:
```hcl
resource "aws_ram_resource_share_accepter" "example" {
  arn = "arn:aws:ram:us-east-1:999999999999:resource-share-invitation/49510c94-3f65-5949-a939-98900e9a5765"
}
```

Slightly more complicated example:
```hcl
resource "aws_ram_resource_share" "test" {
  provider = "aws.alternate"

  name                      = "veryfineshare"
  allow_external_principals = true

  tags = {
	Name = "veryfineshare"
  }
}

resource "aws_ram_principal_association" "test" {
  provider = "aws.alternate"

  principal          = "${data.aws_caller_identity.receiver.account_id}"
  resource_share_arn = "${aws_ram_resource_share.test.arn}"

  depends_on = ["data.aws_caller_identity.receiver"]
}

data "aws_caller_identity" "receiver" {}

resource "aws_ram_resource_share_accepter" "test" {
  share_arn = "${aws_ram_resource_share.test.arn}"

  depends_on = ["aws_ram_resource_share.test", "aws_ram_principal_association.test"]
}
```

### Arguments
One of the following is required:
**share_arn** - (Optional) Resource share ARN. 
**invitation_arn** - (Optional) Resource share invitation ARN. 

### Attributes
**status** - Invitation status (e.g., ACCEPTED, REJECTED).
**receiver_account_id** - Account ID of resource share receiver.
**sender_account_id** - Account ID of resource share sender.
**share_name** - Name of the resource share.
**resources** - List of ARNs of shared resources.

### DIFF
--- a/aws/resource_aws_ram_resource_share_accepter.go
+++ b/aws/resource_aws_ram_resource_share_accepter.go
@@ -193,14 +193,6 @@ func resourceAwsRamResourceShareAccepterRead(d *schema.ResourceData, meta interf
 	return nil
 }
 
-/*
-func resourceAwsRamResourceShareAccepterDelete(d *schema.ResourceData, meta interface{}) error {
-	d.SetId("")
-	log.Printf("[WARN] Will not delete resource share invitation. Terraform will remove this invitation accepter from the state file. However, resources may remain.")
-	return nil
-}
-*/
-
 func resourceAwsRamResourceShareAccepterDelete(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ramconn
 
--- a/aws/resource_aws_ram_resource_share_accepter.go
+++ b/aws/resource_aws_ram_resource_share_accepter.go
@@ -1,0 +1,163 @@
+package aws
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ram"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceAwsRamResourceShareAccepter() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsRamResourceShareAccepterCreate,
+		Read:   resourceAwsRamResourceShareAccepterRead,
+		Delete: resourceAwsRamResourceShareAccepterDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validateArn,
+			},
+
+			"status": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"receiver_account_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"sender_account_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"share_arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"share_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"client_token": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(1, 64),
+			},
+		},
+	}
+}
+
+func resourceAwsRamResourceShareAccepterCreate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ramconn
+
+	requestInput := &ram.AcceptResourceShareInvitationInput{
+		ResourceShareInvitationArn: aws.String(d.Get("arn").(string)),
+	}
+
+	if v, ok := d.GetOk("client_token"); ok {
+		requestInput.ClientToken = aws.String(v.(string))
+	}
+
+	log.Println("[DEBUG] Accept RAM resource share invitation request:", requestInput)
+	requestOutput, err := conn.AcceptResourceShareInvitation(requestInput)
+	if err != nil {
+		return fmt.Errorf("Error accepting RAM resource share invitation: %s", err)
+	}
+
+	d.SetId(aws.StringValue(requestOutput.ResourceShareInvitation.ResourceShareInvitationArn))
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{ram.ResourceShareInvitationStatusPending},
+		Target:  []string{ram.ResourceShareInvitationStatusAccepted},
+		Refresh: resourceAwsRamResourceShareAccepterStateRefreshFunc(conn, d.Id()),
+		Timeout: d.Timeout(schema.TimeoutCreate),
+	}
+
+	_, err = stateConf.WaitForState()
+	if err != nil {
+		return fmt.Errorf("Error waiting for RAM resource share invitation (%s) state: %s", d.Id(), err)
+	}
+
+	return resourceAwsRamResourceShareAccepterRead(d, meta)
+}
+
+func resourceAwsRamResourceShareAccepterRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ramconn
+
+	request := &ram.GetResourceShareInvitationsInput{
+		ResourceShareInvitationArns: []*string{aws.String(d.Id())},
+	}
+
+	output, err := conn.GetResourceShareInvitations(request)
+	if err != nil {
+		if isAWSErr(err, ram.ErrCodeUnknownResourceException, "") {
+			log.Printf("[WARN] No RAM resource share invitation by ARN (%s) found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("Error reading RAM resource share invitation %s: %s", d.Id(), err)
+	}
+
+	if len(output.ResourceShareInvitations) == 0 {
+		log.Printf("[WARN] No RAM resource share invitation by ARN (%s) found, removing from state", d.Id())
+		d.SetId("")
+		return nil
+	}
+
+	invitation := output.ResourceShareInvitations[0]
+
+	d.Set("status", invitation.Status)
+	d.Set("receiver_account_id", invitation.ReceiverAccountId)
+	d.Set("sender_account_id", invitation.SenderAccountId)
+	d.Set("share_arn", invitation.ResourceShareArn)
+	d.Set("share_name", invitation.ResourceShareName)
+
+	return nil
+}
+
+func resourceAwsRamResourceShareAccepterDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[WARN] Will not delete resource share invitation. Terraform will remove this invitation from the state file. However, resources may remain.")
+	return nil
+}
+
+func resourceAwsRamResourceShareAccepterStateRefreshFunc(conn *ram.RAM, invitationArn string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		request := &ram.GetResourceShareInvitationsInput{
+			ResourceShareInvitationArns: []*string{aws.String(invitationArn)},
+		}
+
+		output, err := conn.GetResourceShareInvitations(request)
+
+		if err != nil {
+			return nil, "Unable to get resource share invitations", err
+		}
+
+		if len(output.ResourceShareInvitations) == 0 {
+			return nil, "Resource share invitation not found", nil
+		}
+
+		invitation := output.ResourceShareInvitations[0]
+
+		return invitation, aws.StringValue(invitation.Status), nil
+	}
+}
--- a/aws/resource_aws_ram_resource_share_accepter.go
+++ b/aws/resource_aws_ram_resource_share_accepter.go
@@ -3,14 +3,15 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ram"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
-	"github.com/hashicorp/terraform/helper/validation"
 )
 
 func resourceAwsRamResourceShareAccepter() *schema.Resource {
@@ -19,17 +20,35 @@ func resourceAwsRamResourceShareAccepter() *schema.Resource {
 		Read:   resourceAwsRamResourceShareAccepterRead,
 		Delete: resourceAwsRamResourceShareAccepterDelete,
 
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(5 * time.Minute),
 			Delete: schema.DefaultTimeout(5 * time.Minute),
 		},
 
 		Schema: map[string]*schema.Schema{
-			"arn": {
+			"invitation_arn": {
 				Type:         schema.TypeString,
-				Required:     true,
+				Optional:     true,
 				ForceNew:     true,
+				Computed:     true,
 				ValidateFunc: validateArn,
+			},
+
+			"share_arn": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Computed:     true,
+				ValidateFunc: validateArn,
+			},
+
+			"share_id": {
+				Type:     schema.TypeString,
+				Computed: true,
 			},
 
 			"status": {
@@ -47,21 +66,17 @@ func resourceAwsRamResourceShareAccepter() *schema.Resource {
 				Computed: true,
 			},
 
-			"share_arn": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
-
 			"share_name": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 
-			"client_token": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				ForceNew:     true,
-				ValidateFunc: validation.StringLenBetween(1, 64),
+			"resources": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
 			},
 		},
 	}
@@ -70,21 +85,31 @@ func resourceAwsRamResourceShareAccepter() *schema.Resource {
 func resourceAwsRamResourceShareAccepterCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).ramconn
 
-	requestInput := &ram.AcceptResourceShareInvitationInput{
-		ResourceShareInvitationArn: aws.String(d.Get("arn").(string)),
+	in := &ram.AcceptResourceShareInvitationInput{
+		ClientToken: aws.String(resource.UniqueId()),
 	}
 
-	if v, ok := d.GetOk("client_token"); ok {
-		requestInput.ClientToken = aws.String(v.(string))
+	if v, ok := d.GetOk("invitation_arn"); ok {
+		in.ResourceShareInvitationArn = aws.String(v.(string))
+	} else if v, ok := d.GetOk("share_arn"); ok {
+		// need to find invitation arn
+		invitationARN, err := resourceAwsRamResourceShareGetInvitationARN(conn, v.(string))
+		if err != nil {
+			return err
+		}
+
+		in.ResourceShareInvitationArn = aws.String(invitationARN)
+	} else {
+		return fmt.Errorf("Either an invitation ARN or share ARN are required")
 	}
 
-	log.Println("[DEBUG] Accept RAM resource share invitation request:", requestInput)
-	requestOutput, err := conn.AcceptResourceShareInvitation(requestInput)
+	log.Printf("[DEBUG] Accept RAM resource share invitation request: %s", in)
+	out, err := conn.AcceptResourceShareInvitation(in)
 	if err != nil {
 		return fmt.Errorf("Error accepting RAM resource share invitation: %s", err)
 	}
 
-	d.SetId(aws.StringValue(requestOutput.ResourceShareInvitation.ResourceShareInvitationArn))
+	d.SetId(aws.StringValue(out.ResourceShareInvitation.ResourceShareInvitationArn))
 
 	stateConf := &resource.StateChangeConf{
 		Pending: []string{ram.ResourceShareInvitationStatusPending},
@@ -95,7 +120,7 @@ func resourceAwsRamResourceShareAccepterCreate(d *schema.ResourceData, meta inte
 
 	_, err = stateConf.WaitForState()
 	if err != nil {
-		return fmt.Errorf("Error waiting for RAM resource share invitation (%s) state: %s", d.Id(), err)
+		return fmt.Errorf("Error waiting for RAM resource share (%s) state: %s", d.Id(), err)
 	}
 
 	return resourceAwsRamResourceShareAccepterRead(d, meta)
@@ -108,7 +133,7 @@ func resourceAwsRamResourceShareAccepterRead(d *schema.ResourceData, meta interf
 		ResourceShareInvitationArns: []*string{aws.String(d.Id())},
 	}
 
-	output, err := conn.GetResourceShareInvitations(request)
+	out, err := conn.GetResourceShareInvitations(request)
 	if err != nil {
 		if isAWSErr(err, ram.ErrCodeUnknownResourceException, "") {
 			log.Printf("[WARN] No RAM resource share invitation by ARN (%s) found, removing from state", d.Id())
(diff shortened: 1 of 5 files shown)


lorengordon:
Would love to get this reviewed and merged! Being able to create and accept resource shares in the same config would be awesome (across accounts, especially)!

bflad:
@YakDriver do you need help with the testing/documentation here?

YakDriver:
@bflad maybe. not sure how you'd like to handle a multi-account test.

ewbankkit:
@YakDriver You could take a look at the cross-account tests in [`resource_aws_dx_gateway_association_test.go`](https://github.com/terraform-providers/terraform-provider-aws/blob/04354ca6d376fb0e1aaa4fd68543342cd546df8b/aws/resource_aws_dx_gateway_association_test.go) from https://github.com/terraform-providers/terraform-provider-aws/pull/8528 for usage of the cross-account support available in the acceptance testing framework.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,163 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"share_name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"client_token": {
```
ewbankkit: We don't usually expose `client_token` as a resource attribute but instead use something like `ClientToken:      aws.String(resource.UniqueId()),` in the AWS API call.
See e.g. the [`aws_ram_principal_association`](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_ram_principal_association.go) resource.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,163 @@
...
+		Read:   resourceAwsRamResourceShareAccepterRead,
+		Delete: resourceAwsRamResourceShareAccepterDelete,
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
```
ewbankkit: Probably just me being dumb, but where would this RAM resource share invitation ARN come from in the case where the RAM resource share is managed via Terraform? The [`AssociateResourceShare`](https://docs.aws.amazon.com/ram/latest/APIReference/API_AssociateResourceShare.html) API used by `aws_ram_principal_association` and `aws_ram_resource_association` doesn't seem to return the invitation ARN.

ewbankkit:
Should we also return the ARNs of the resources shared in the invitation? I think this would help when using other resources that depend on those shared resources.

aws/resource_aws_ram_resource_share_accepter.go
lorengordon: Dang, that's kind of messed up. The API that generates the invite does not return the invite ARN? Playing around with it a little, it looks like the invite ARN is available by querying `GetResourceShareInvitations` in the _target_ account.

Does this accepter resource (or perhaps a new data source) then need to use the `GetResourceShareInvitations` API to get the list of invitations, and filter it by `resourceShareArn` to select `resourceShareInvitationArn`? Seems maybe prone to race conditions, between when the association is created and when the invitation can be queried from the target account.

ewbankkit: Since this accepter resource runs in the context of the target account then specifying `resource_share_arn` as a `Required: true` attribute should mean that [`GetResourceShareInvitations`](https://docs.aws.amazon.com/ram/latest/APIReference/API_GetResourceShareInvitations.html) would pick the one and only (hopefully 😄) `resource_shared_invitation_arn` that needs to be [accepted](https://docs.aws.amazon.com/ram/latest/APIReference/API_AcceptResourceShareInvitation.html) - both API calls would happen in the resource's _Create_ function.
Some investigation needs to be done to determine the `id` value for this resource - I think  `resource_shared_invitation_arn` would work.
You'll have to think about what it means to _Read_ and _Delete_ (plus maybe _Update_) this resource.

lorengordon: > GetResourceShareInvitations would pick the one and only (hopefully 😄) resource_shared_invitation_arn that needs to be accepted

Not quite that easy... rejected invites are also returned... 😢 

```
$ aws ram get-resource-share-invitations --profile target --resource-share-arns arn:aws:ram:us-east-1:MASTER:resource-share/SHARE_UUID
{
    "resourceShareInvitations": [
        {
            "resourceShareInvitationArn": "arn:aws:ram:us-east-1:MASTER:resource-share-invitation/INVITE_UUID",
            "resourceShareName": "SHARE_NAME",
            "resourceShareArn": "arn:aws:ram:us-east-1:MASTER:resource-share/SHARE_UUID",
            "senderAccountId": "MASTER",
            "receiverAccountId": "TARGET",
            "invitationTimestamp": 1557673462.185,
            "status": "REJECTED"
        },
        {
            "resourceShareInvitationArn": "arn:aws:ram:us-east-1:MASTER:resource-share-invitation/INVITE_UUID",
            "resourceShareName": "SHARE_NAME",
            "resourceShareArn": "arn:aws:ram:us-east-1:MASTER:resource-share/SHARE_UUID",
            "senderAccountId": "MASTER",
            "receiverAccountId": "TARGET",
            "invitationTimestamp": 1557692314.324,
            "status": "PENDING"
        }
    ]
}
```

lorengordon: Does seem, thankfully, as if only _one_ invite will be PENDING for a given share, though. The API accepts additional calls to associate the same share to the same principal, but there is only one PENDING invite in the target account...

ewbankkit: 👍

aws/resource_aws_ram_resource_share_accepter.go
YakDriver: If I understand the AWS API correctly, this is not an attribute (output) but an argument (input) for users that want to ensure idempotency with their own token. That the AWS provider would handle that for users seems reasonable. I can make the change.

aws/resource_aws_ram_resource_share_accepter.go
YakDriver: I've made it work with either invitation or share ARN.

YakDriver:
> Should we also return the ARNs of the resources shared in the invitation? I think this would help when using other resources that depend on those shared resources.

I added a list `resources` with the ARNs of resources available through the share.

YakDriver:
@ewbankkit you asked about delete and update... Here's what I think:

For delete, I've made it leave the resource share on deleting the accepter. The other option is simply remove the accepter from the state and warn that resources still exist. I prefer the leaving option but I can see arguments both ways.

For update, all arguments are `force_new` so I don't think update makes much sense.

YakDriver:
In order to make pass though import work (not sure why you'd need to import but...), I've made the ID the invitation ARN.

ewbankkit:
@YakDriver In my mind this resource is kind of like the `aws_vpn_peering_connection_accepter` resource where the peering connection (equivalent to the RAM association I think) could be deleted via the AWS API from either the creator or accepter's end but Terraform only allows deletion from the creator's end - Deleting the accepter resource just removes the resource from the state.
Not sure if the RAM association can be deleted from either end but if it can and you choose to allow deletion from the accepter's resource delete method, I think you may find there are problems running end-to-end acceptance tests.

YakDriver:
@ewbankkit I've added an end-to-end acceptance test, cleaned up commented out sections, added docs, and tested the code in a variety of situations. It seems to work well. The state waiting seems especially important since it takes RAM statuses a while to update.

This is ready for another look!

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,298 @@
...
+		return fmt.Errorf("Either an invitation ARN or share ARN are required")
+	}
+
+	log.Println("[DEBUG] Accept RAM resource share invitation request:", in)
+	out, err := conn.AcceptResourceShareInvitation(in)
+	if err != nil {
+		return fmt.Errorf("Error accepting RAM resource share invitation: %s", err)
+	}
+
+	d.SetId(*out.ResourceShareInvitation.ResourceShareInvitationArn)
```
ewbankkit: No need to dereference here - `SetId()` can handle `*string`.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			return err
+		}
+
+		in.ResourceShareInvitationArn = invitationARN
+	} else {
+		return fmt.Errorf("Either an invitation ARN or share ARN are required")
+	}
+
+	log.Println("[DEBUG] Accept RAM resource share invitation request:", in)
```
ewbankkit: Need a `%s` here to print `in`.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,298 @@
...
+		return fmt.Errorf("Error accepting RAM resource share invitation: %s", err)
+	}
+
+	d.SetId(*out.ResourceShareInvitation.ResourceShareInvitationArn)
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{ram.ResourceShareInvitationStatusPending},
+		Target:  []string{ram.ResourceShareInvitationStatusAccepted},
+		Refresh: resourceAwsRamResourceShareAccepterStateRefreshFunc(
+			conn, *out.ResourceShareInvitation.ResourceShareInvitationArn),
```
ewbankkit: Can use `d.Id()` here as it was set to the ARN above.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		if nextToken != "" {
+			input.NextToken = aws.String(nextToken)
+		}
+		out, err := conn.GetResourceShareInvitations(input)
+		if err != nil {
+			return nil, err
+		}
+		for _, invitation := range out.ResourceShareInvitations {
+			if *invitation.Status == ram.ResourceShareInvitationStatusPending {
```
ewbankkit: Use aws.StringValue() instead of dereferencing directly.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	invitation := out.ResourceShareInvitations[0]
+
+	d.Set("status", invitation.Status)
+	d.Set("receiver_account_id", invitation.ReceiverAccountId)
+	d.Set("sender_account_id", invitation.SenderAccountId)
+	d.Set("share_arn", invitation.ResourceShareArn)
+	d.Set("invitation_arn", invitation.ResourceShareInvitationArn)
+	d.Set("share_id", resourceAwsRamResourceShareGetIDFromARN(*invitation.ResourceShareArn))
```
ewbankkit: Use aws.StringValue() instead of dereferencing directly.

YakDriver:
@ewbankkit requested changes made

aws/resource_aws_ram_resource_share_accepter.go
ewbankkit: Actually, I don't see `share_id` in any of the other RAM resources and having to compute the ID from the ARN seems a bit janky, can we just remove this attribute?

YakDriver:
@ewbankkit @bflad anything else I can provide?

ewbankkit:
Once the RAM share has been accepted it seems to me that this resource then has behavior very similar to the [`aws_ram_resource_share`](https://www.terraform.io/docs/providers/aws/r/ram_resource_share.html) resource and maybe the ID should be `ResourceShareInvitation.ResourceShareArn` and `resourceAwsRamResourceShareAccepterRead()` functions similar to `resourceAwsRamResourceShareRead()`?
Interested in other opinions here.

YakDriver:
I originally had the ID as the resource share ARN but then could not do pass through import so changed it back to invitation ARN. I'm not sure why you'd need to import. @lorengordon do you have scenarios where importing of the invitation would be useful? If import is a non-issue, it's easy to change to share ARN.

lorengordon:
I think as long as this resource does the "right thing" when the invite is accepted outside of terraform, then the import is immaterial.

lorengordon:
@ewbankkit @bflad Are we any closer to merging support for the resource share accepter?

ewbankkit:
@YakDriver Are there scenarios where the invitation "expires" and gets recreated forcing a new resource?
See for example https://github.com/terraform-providers/terraform-provider-aws/issues/9209 where cross-account Direct Connect gateway associations have a proposal resource that has a TTL.

lorengordon:
@ewbankkit @YakDriver it does appear that the invites age out after some time period... bother.

YakDriver:
@ewbankkit @lorengordon I don't see that being a big issue for most configurations for several reasons.

1. Terraform _is_ an automation tool after all and automated processes are generally pretty quickly relative to expiring invitations ([15 days](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_manage_accounts_invites.html)).
2. A significant percentage of scenarios using this will be within the commonplace account/subaccount model where the same Terraform instance is going to create and accept the invitation seconds later. (I added wait states because trying to accept before the invitation is fully created seems more likely.)

YakDriver:
@bflad As requested, I've checked this for resource cycling and it is not a problem. See my configuration and results below:

### config
```hcl
provider "aws" {
  region  = "us-east-1"
  profile = "profile2"
}

provider "aws" {
  alias   = "alternate"
  region  = "us-east-1"
  profile = "profile1"
}

resource "aws_ram_resource_share" "test" {
  provider = "aws.alternate"
  name                      = var.resource_share_name
  allow_external_principals = true

  tags = {
    Name = var.test_tag
  }
}

resource "aws_ram_principal_association" "test" {
  provider = "aws.alternate"
  principal          = "${data.aws_caller_identity.receiver.account_id}"
  resource_share_arn = "${aws_ram_resource_share.test.arn}"
  depends_on = ["data.aws_caller_identity.receiver"]
}

data "aws_caller_identity" "receiver" {}

resource "aws_ram_resource_share_accepter" "test" {
  share_arn = "${aws_ram_resource_share.test.arn}"
  depends_on = ["aws_ram_resource_share.test", "aws_ram_principal_association.test"]
}
```

### Initial apply
```console
$ terraform apply
data.aws_caller_identity.receiver: Refreshing state...

  + create

...
Plan: 3 to add, 0 to change, 0 to destroy.

aws_ram_resource_share.test: Creating...
aws_ram_resource_share.test: Creation complete after 3s [id=arn:aws:ram:us-east-1:123456789012:resource-share/f6b624d0-4205-bc4a-1b75-8c8e0f7120f0]
aws_ram_principal_association.test: Creating...
aws_ram_principal_association.test: Creation complete after 4s [id=arn:aws:ram:us-east-1:123456789012:resource-share/f6b624d0-4205-bc4a-1b75-8c8e0f7120f0,987654321987]
aws_ram_resource_share_accepter.test: Creating...
aws_ram_resource_share_accepter.test: Creation complete after 2s [id=arn:aws:ram:us-east-1:123456789012:resource-share-invitation/70b624d0-4793-43f6-f440-ae209ae95bfe]

Apply complete! Resources: 3 added, 0 changed, 0 destroyed.
```

### Re-apply
```console
$ terraform apply
data.aws_caller_identity.receiver: Refreshing state...
aws_ram_resource_share.test: Refreshing state... [id=arn:aws:ram:us-east-1:123456789012:resource-share/f6b624d0-4205-bc4a-1b75-8c8e0f7120f0]
aws_ram_principal_association.test: Refreshing state... [id=arn:aws:ram:us-east-1:123456789012:resource-share/f6b624d0-4205-bc4a-1b75-8c8e0f7120f0,987654321987]
aws_ram_resource_share_accepter.test: Refreshing state... [id=arn:aws:ram:us-east-1:123456789012:resource-share-invitation/70b624d0-4793-43f6-f440-ae209ae95bfe]

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

### Re-re-apply
```console
$ terraform apply
aws_ram_resource_share.test: Refreshing state... [id=arn:aws:ram:us-east-1:123456789012:resource-share/f6b624d0-4205-bc4a-1b75-8c8e0f7120f0]
data.aws_caller_identity.receiver: Refreshing state...
aws_ram_principal_association.test: Refreshing state... [id=arn:aws:ram:us-east-1:123456789012:resource-share/f6b624d0-4205-bc4a-1b75-8c8e0f7120f0,987654321987]
aws_ram_resource_share_accepter.test: Refreshing state... [id=arn:aws:ram:us-east-1:123456789012:resource-share-invitation/70b624d0-4793-43f6-f440-ae209ae95bfe]

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+		return fmt.Errorf("Either an invitation ARN or share ARN are required")
+	}
+
+	log.Printf("[DEBUG] Accept RAM resource share invitation request: %s", in)
+	out, err := conn.AcceptResourceShareInvitation(in)
+	if err != nil {
+		return fmt.Errorf("Error accepting RAM resource share invitation: %s", err)
+	}
+
+	d.SetId(aws.StringValue(out.ResourceShareInvitation.ResourceShareInvitationArn))
```
bflad: Given that the intention of the resource is to ensure that we have accepted a RAM Resource Share from another account and that the invitation could be swapped out from under the Terraform resource, I wonder if we should instead use the resource share ARN we are expecting as the resource and import ID. I have a feeling most folks won't want to be dealing with invitations directly unless really necessary.

```suggestion
	d.SetId(aws.StringValue(out.ResourceShareInvitation.ResourceShareArn))
```

In the `Read` function this will require some tweaks.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func resourceAwsRamResourceShareAccepterRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ramconn
+
+	request := &ram.GetResourceShareInvitationsInput{
+		ResourceShareInvitationArns: []*string{aws.String(d.Id())},
+	}
+
+	out, err := conn.GetResourceShareInvitations(request)
```
bflad: Given the requested ID change above, we should lookup by `ResourceShareArns` instead of `ResourceShareInvitationArns`. This will allow this resource to tolerate the invitation being rescinded and the resource share reaccepted under a new ARN outside of Terraform.

We should use the paginated AWS Go SDK function [`GetResourceShareInvitationsPages`](https://docs.aws.amazon.com/sdk-for-go/api/service/ram/#RAM.GetResourceShareInvitationsPages) here, to prevent any potential issues with paging. 

```go
input := &ram.GetResourceShareInvitationsInput{
	ResourceShareArns: []*string{aws.String(d.Id())},
}

var invitation *ram.ResourceShareInvitation
err := conn.GetResourceShareInvitationsPages(input, func(page *ram.GetResourceShareInvitationsOutput, lastPage bool) bool {
	for _, rsi := range page.ResourceShareInvitations {
		if aws.StringValue(rsi.Status) == ram.ResourceShareInvitationStatusAccepted {
			invitation = rsi
			break
		}
	}

	return !lastPage
})

if isAWSErr(err, ram.ErrCodeUnknownResourceException, "") {
	log.Printf("[WARN] No RAM resource share invitation by ARN (%s) found, removing from state", d.Id())
	d.SetId("")
	return nil
}

if err != nil {
	return fmt.Errorf("Error reading RAM resource share invitation %s: %s", d.Id(), err)
}

if invitation == nil {
	log.Printf("[WARN] No RAM resource share invitation by ARN (%s) found, removing from state", d.Id())
	d.SetId("")
	return nil
}
```

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+
+	d.Set("status", invitation.Status)
+	d.Set("receiver_account_id", invitation.ReceiverAccountId)
+	d.Set("sender_account_id", invitation.SenderAccountId)
+	d.Set("share_arn", invitation.ResourceShareArn)
+	d.Set("invitation_arn", invitation.ResourceShareInvitationArn)
+	d.Set("share_id", resourceAwsRamResourceShareGetIDFromARN(aws.StringValue(invitation.ResourceShareArn)))
+	d.Set("share_name", invitation.ResourceShareName)
+
+	var nextToken string
```
bflad: Same here, we should prefer the paginated AWS Go SDK function [`ListResourcesPages`](https://docs.aws.amazon.com/sdk-for-go/api/service/ram/#RAM.ListResourcesPages) here 👍 

```go
	listResourcesInput := &ram.ListResourcesInput{
		ResourceOwner:     aws.String(ram.ResourceOwnerOtherAccounts),
		ResourceShareArns: []*string{invitation.ResourceShareArn)},
		Principal:         invitation.SenderAccountId,
	}
	var resourceARNs []string

	err = conn.ListResourcesPages(listResourcesInput, func(page *ram.ListResourcesOutput, lastPage bool) bool {
		for _, r := range page.Resources {
			resourceARNs = append(resourceARNs, aws.StringValue(resource.Arn))
		}
		
		return !lastPage
	})

	if err != nil {
		return fmt.Errorf("error listing resources for RAM Resource Share (%s): %s", d.Id(), err)
	}
	
	if err := d.Set("resources", resourceARNs); err != nil {
		return fmt.Errorf("unable to set resources: %s", err)
	}
```

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+		return fmt.Errorf("unable to set resources: %s", err)
+	}
+
+	return nil
+}
+
+func resourceAwsRamResourceShareAccepterDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).ramconn
+
+	v, ok := d.GetOk("share_arn")
```
bflad: This will always be available as `d.Id()` with the above changes 👍

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+	_, err := conn.DisassociateResourceShare(in)
+	if err != nil {
+		return fmt.Errorf("Error leaving RAM resource share: %s", err)
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{ram.ResourceShareAssociationStatusAssociated},
+		Target:  []string{ram.ResourceShareAssociationStatusDisassociated},
+		Refresh: resourceAwsRamResourceShareStateRefreshFunc(conn, shareARN),
+		Timeout: d.Timeout(schema.TimeoutCreate),
```
bflad: 😎 

```suggestion
		Timeout: d.Timeout(schema.TimeoutDelete),
```

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		if awserr, ok := err.(awserr.Error); ok {
```
bflad: This can be simplified with

```go
isAWSErr(err, ram.ErrCodeUnknownResourceException, "")
```

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+			switch awserr.Code() {
+			case ram.ErrCodeUnknownResourceException:
+				// what we want
+				d.SetId("")
```
bflad: Calling `d.SetId("")` during the `Delete` function is extraneous. 👍

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+			}
+		}
+		return fmt.Errorf("Error waiting for RAM resource share (%s) state: %s", d.Id(), err)
+	}
+
+	return nil
+}
+
+func resourceAwsRamResourceShareGetInvitationARN(conn *ram.RAM, resourceShareARN string) (string, error) {
+	var nextToken string
```
bflad: Please see the `GetResourceShareInvitationsPages` example above.

aws/resource_aws_ram_resource_share_accepter.go
```diff
@@ -0,0 +1,297 @@
...
+	return "", fmt.Errorf("Unable to find a pending invitation for resource share %s", resourceShareARN)
+}
+
+func resourceAwsRamResourceShareAccepterStateRefreshFunc(conn *ram.RAM, invitationArn string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		request := &ram.GetResourceShareInvitationsInput{
+			ResourceShareInvitationArns: []*string{aws.String(invitationArn)},
+		}
+
+		out, err := conn.GetResourceShareInvitations(request)
```
bflad: Same here regarding the paginated function. 👍

aws/resource_aws_ram_resource_share_accepter_test.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		request := &ram.GetResourceSharesInput{
+			ResourceShareArns: []*string{aws.String(rs.Primary.Attributes["share_arn"])},
+			ResourceOwner:     aws.String(ram.ResourceOwnerOtherAccounts),
+		}
+
+		resp, err := conn.GetResourceShares(request)
+		if err != nil {
+			if awserr, ok := err.(awserr.Error); ok {
```
bflad: Nit: `isAWSErr(err, ram.ErrCodeUnknownResourceException, "")`

aws/resource_aws_ram_resource_share_accepter_test.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		if len(resp.ResourceShares) == 0 {
+			return nil
+		}
+	}
+	return fmt.Errorf("RAM resource share association found, should be destroyed")
+}
+
+func testAccCheckAwsRamResourceShareAccepterExists(name string) resource.TestCheckFunc {
```
bflad: This function should verify from the API that there is an accepted resource share invitation, similar but opposite to the check destroy function above. 👍

aws/resource_aws_ram_resource_share_accepter_test.go
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+resource "aws_ram_principal_association" "test" {
+  provider = "aws.alternate"
+
+  principal          = "${data.aws_caller_identity.receiver.account_id}"
+  resource_share_arn = "${aws_ram_resource_share.test.arn}"
+
+  depends_on = ["data.aws_caller_identity.receiver"]
```
bflad: Nit: This `depends_on` is extraneous since there is an implicit resource reference above.

aws/resource_aws_ram_resource_share_accepter_test.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+data "aws_caller_identity" "receiver" {}
+
+resource "aws_ram_resource_share_accepter" "test" {
+  share_arn = "${aws_ram_resource_share.test.arn}"
+
+  depends_on = ["aws_ram_resource_share.test", "aws_ram_principal_association.test"]
```
bflad: Nit: The `aws_ram_resource_share.test` within the `depends_on` is extraneous since there is an implicit resource reference above. That said, this resource could be updated to the below, which would not require any `depends_on` usage:

```hcl
resource "aws_ram_resource_share_accepter" "test" {
  share_arn = "${aws_ram_principal_association.test.resource_share_arn}"
}
```

aws/resource_aws_ram_resource_share_accepter_test.go
```diff
@@ -0,0 +1,128 @@
...
+}
+`, shareName, shareName)
```
bflad: Nit: Go format verbs accept 1-based indexing notation for simplification of arguments. e.g. replacing the two `%q` above with `%[1]q` will allow you to remove the second `shareName` argument here.

website/docs/r/ram_resource_share_accepter.markdown
```diff
@@ -0,0 +1,55 @@
...
+layout: "aws"
+page_title: "AWS: aws_ram_resource_share_accepter"
+sidebar_current: "docs-aws-resource-ram-resource-share-accepter"
+description: |-
+  Manages accepting a Resource Access Manager (RAM) Resource Share invitation.
+---
+
+# Resource: aws_ram_resource_share_accepter
+
+Manages accepting a Resource Access Manager (RAM) Resource Share invitation. From a _receiver_ AWS organization, accept an invitation to share resources that were shared by a _sender_ AWS organization. To create a resource share in the _sender_, see the [`aws_ram_resource_share` resource](/docs/providers/aws/r/ram_resource_share.html).
```
bflad: It might be good to clarify in here that this resource is not necessary within the same AWS Organizations organization if organizations sharing is enabled. 👍

website/docs/r/ram_resource_share_accepter.markdown
```diff
@@ -0,0 +1,55 @@
...
+
+## Example Usage
+
+The invitation can be accepted using the Resource Share Amazon Resource Name (ARN).
+```hcl
+resource "aws_ram_resource_share_accepter" "example" {
```
bflad: It might be good to take the full test configuration and use it in here for clarity with the `aws_ram_principal_association` resource.

aws/resource_aws_ram_resource_share_accepter.go
YakDriver: I'm concerned that the issues with the invitation ARN are going to cause problems. If the receiver account has multiple resource shares with the sender account, the read function is going to randomly grab the first invitation AWS returns with the accepted status. When deleting, it can leave the wrong share.

bflad: The API filtering should be limiting to a single resource share, but in case its not, we can be extra cautious with also checking the resource share ARN matches our ID:

```go
var invitation *ram.ResourceShareInvitation
err := conn.GetResourceShareInvitationsPages(input, func(page *ram.GetResourceShareInvitationsOutput, lastPage bool) bool {
	for _, rsi := range page.ResourceShareInvitations {
		if aws.StringValue(rsi.ResourceShareArn) == d.Id() && aws.StringValue(rsi.Status) == ram.ResourceShareInvitationStatusAccepted {
			invitation = rsi
			return false
		}
	}

	return !lastPage
})
```

YakDriver: That extra check doesn't address the concern: you can have multiple invitations for the same share. **But**, after testing it doesn't appear to matter which invitation you accept. Accepting any one of them has the same effect as accepting any other of them. I'm good if you're good with this.

This is the config I used to test it out. Here we don't know which invitation will be accepted since `aws_ram_principal_association.test2.resource_share_arn == aws_ram_principal_association.test1.resource_share_arn`, but, it doesn't appear to matter.

```hcl
resource "aws_ram_principal_association" "test1" {
  provider           = "aws.alternate"
  principal          = "${data.aws_caller_identity.receiver.account_id}"
  resource_share_arn = "${aws_ram_resource_share.test.arn}"
}

resource "aws_ram_principal_association" "test2" {
  provider           = "aws.alternate"
  principal          = "${data.aws_caller_identity.receiver.account_id}"
  resource_share_arn = "${aws_ram_resource_share.test.arn}"
}

data "aws_caller_identity" "receiver" {}

resource "aws_ram_resource_share_accepter" "test" {
  share_arn = "${aws_ram_principal_association.test2.resource_share_arn}"
}
```

website/docs/r/ram_resource_share_accepter.markdown
YakDriver: added

YakDriver:
@bflad I've completed the adjust. I made these changes on top of your recommended changes:

- I could not figure out a good reason why `invitation_arn` was an argument. It adds unnecessary complexity since there is no way to even get this value from Terraform. I made it an attribute and it simplified the code.
- Created one function that gets invitations whether they are `ACCEPTED` or `PENDING` instead of duplicating in `create()` and `read()`.

Let me know if there's any other adjust. It's much cleaner now.

YakDriver:
```console
$ make testacc TESTARGS='-run=TestAccAwsRamResourceShareAccepter_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAwsRamResourceShareAccepter_basic -timeout 120m
=== RUN   TestAccAwsRamResourceShareAccepter_basic
=== PAUSE TestAccAwsRamResourceShareAccepter_basic
=== CONT  TestAccAwsRamResourceShareAccepter_basic
--- PASS: TestAccAwsRamResourceShareAccepter_basic (41.76s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	41.842s
```

YakDriver:
@bflad Any idea why pr-label, labeler 2.0 says that "resource not accessible by integration?"

bflad:
> @bflad Any idea why pr-label, labeler 2.0 says that "resource not accessible by integration?"

Nope haven't seen that yet. We're still new to GitHub Actions v2 ourselves. 😅

PR #8272, opened 2019-04-10
Fix id in Data Source Cognito User Pool

zemanlx:
### Community Note

* Set `data "aws_cognito_user_pools"` key `id` to user pool ID and not name value (which is confusing). Name value is already kept in `name`.

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsCognitoUserPools_basic'
TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccDataSourceAwsCognitoUserPools_basic -timeout 120m
=== RUN   TestAccDataSourceAwsCognitoUserPools_basic
=== PAUSE TestAccDataSourceAwsCognitoUserPools_basic
=== CONT  TestAccDataSourceAwsCognitoUserPools_basic
--- PASS: TestAccDataSourceAwsCognitoUserPools_basic (21.58s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       (cached)
```

### DIFF
--- a/aws/data_source_aws_cognito_user_pools.go
+++ b/aws/data_source_aws_cognito_user_pools.go
@@ -61,7 +61,7 @@ func dataSourceAwsCognitoUserPoolsRead(d *schema.ResourceData, meta interface{})
 		return fmt.Errorf("No cognito user pool found with name: %s", name)
 	}
 
-	d.SetId(name)
+	d.SetId(ids[0])
 	d.Set("ids", ids)
 	d.Set("arns", arns)
 
--- a/aws/data_source_aws_cognito_user_pools_test.go
+++ b/aws/data_source_aws_cognito_user_pools_test.go
@@ -18,6 +18,8 @@ func TestAccDataSourceAwsCognitoUserPools_basic(t *testing.T) {
 			{
 				Config: testAccDataSourceAwsCognitoUserPoolsConfig_basic(rName),
 				Check: resource.ComposeTestCheckFunc(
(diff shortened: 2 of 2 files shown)


bflad:
Hi @zemanlx 👋 Thanks for submitting this. Please note that changing resource IDs falls under the category of [a major version change for Terraform Providers](https://www.terraform.io/docs/extend/best-practices/versioning.html#example-major-number-increments) so I have labeled this as a breaking change that would only be considered for a future major version. We also do not have a way to signal to operators about upcoming resource ID changes as that type of change is very uncommon.

The usage of the singular `id` reference from a plural data source is not a use case we are intending to support or encourage. The `id` value can and likely will change across Terraform runs. Its existence in this case is only because it is a requirement for Terraform resources and data sources so it requires some value, which could have been a random timestamp or some other inconsequential value. In this case the original authors chose `name` as that is the common identifier across all the potential returned Cognito User Pools.

Given the above, I'm going to close this pull request as there is some rationale behind the current implementation and having operators go through the burden of changing their configurations in an upcoming major version release seems unbalanced for the change. Please feel free to reach out if there is something we are missing about the intended use case other than some slight confusion about the attribute naming.

If you would prefer to submit another contribution for this situation, we would be happy to look at a _singular_ data source for Cognito User Pools, that errors with 0 or more than 1 result, in which the `id` attribute could be appropriately set to match the returned User Pool. 👍

PR #8274, opened 2019-04-10
resource/aws_route53_record: Remove deprecation from `allow_overwrite` argument

### LINKED ISSUE #7846: aws_route53_record `allow_overwrite` deprecation will break ability to update `SOA`
### Community Note

### Terraform Version

0.11.12

### Affected Resource(s)

2.0.0+

* `aws_route53_record`

Our best practice is to set a low `SOA` on service discovery zones. The default SOA of 900 seconds leads to a default negative cache TTL which messes with service discovery when services query for the DNS record before the records have been updated/created. Basically, when this happens DNS lookups will fail for 15 minutes =); we change it to 30 seconds, so negative DNS queries are refreshed more quickly.

Since AWS creates this `SOA` record automatically, the *only* way to accomplish this in a single `terraform apply` is to support the `allow_overwrite` flag. Using `terraform import` in this scenario is a poor UX and something which we cannot mitigate; it's a consequence of how AWS creates the zone.

```
resource "aws_route53_record" "soa" {
  allow_overwrite = true
  zone_id         = "XXXXX"
  name            = "myzone
  type            = "SOA"
  ttl             = "30"

  records = [
   ...
  ]
}
```

### Expected Behavior

I understand why this behavior is getting deprecated and it makes sense for non-automatically generated DNS record sets. The `SOA` record, however, is an exception. It's created automatically by AWS.

Not sure the best course of action. Terraform should maybe import the `SOA` automatically since it gets generated by AWS, not by terraform. There should be a way to set the `SOA` without getting the deprecation warning because this is not a user problem. There's no way we can mitigate this without the `allow_overwrite` flag.

### Actual Behavior

```
Warning: module.dns.aws_route53_record.soa: "allow_overwrite": [DEPRECATED] The next major version of the Terraform AWS Provider will require importing existing records

```

### Steps to Reproduce

1. Create a new `aws_route53_zone`
2. Attempt to update the `SOA` via the `aws_route53_record` within the same `terraform apply`
1. `terraform apply`

### References

* https://www.terraform.io/docs/providers/aws/guides/version-2-upgrade.html#allow_overwrite-default-value-change

bflad:
### Community Note

Closes #7846

Route53 Record `UPSERT` on creation does have some minor valid use cases. We will keep this argument for now while discouraging its usage in the documentation.

```
--- PASS: TestAccAWSRoute53Record_allowOverwrite (149.81s)
```

### DIFF
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -235,10 +235,9 @@ func resourceAwsRoute53Record() *schema.Resource {
 			},
 
 			"allow_overwrite": {
-				Type:       schema.TypeBool,
-				Optional:   true,
-				Computed:   true,
-				Deprecated: "The next major version of the Terraform AWS Provider will require importing existing records",
+				Type:     schema.TypeBool,
+				Optional: true,
+				Computed: true,
 			},
 		},
 	}
--- a/website/docs/guides/version-2-upgrade.html.md
+++ b/website/docs/guides/version-2-upgrade.html.md
@@ -754,7 +754,7 @@ resource, with in-line routes, to the state.
 
 ### allow_overwrite Default Value Change
 
-The resource now requires existing Route 53 Records to be imported into the Terraform state for management unless the `allow_overwrite` argument is enabled. The `allow_overwrite` flag is considered deprecated for removal in a future major version of the Terraform AWS Provider.
+The resource now requires existing Route 53 Records to be imported into the Terraform state for management unless the `allow_overwrite` argument is enabled.
 
 For example, if the `www.example.com` Route 53 Record in the `example.com` Route 53 Hosted Zone existed previously and this new Terraform configuration was introduced:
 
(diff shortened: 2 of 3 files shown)


aws/resource_aws_route53_record.go
```diff
@@ -235,10 +235,9 @@ func resourceAwsRoute53Record() *schema.Resource {
 			},
 
 			"allow_overwrite": {
-				Type:       schema.TypeBool,
-				Optional:   true,
-				Computed:   true,
-				Deprecated: "The next major version of the Terraform AWS Provider will require importing existing records",
+				Type:     schema.TypeBool,
```
nywilken: Where there any tests removed when the attribute was deprecated that should be restored?

bflad: Nope we don't tend to remove testing on deprecations as the functionality is still expected to work as it did previously. 👍 Here's the deprecation pull request for verification: #7734

PR #8308, opened 2019-04-12
Rearrange docs to de-emphasize using AWS static credentials

topfunky:
Moves description of static credentials to the bottom of the authentication docs. This will encourage new practitioners to use environment variables, roles, or other authentication methods rather than hard-coding sensitive AWS creds in their Terraform files.

It also adds a warning to the static credentials section in order to educate the practitioner about the risks of hard-coding static credentials.

### Community Note

* Moves description of static credentials to the bottom of the authentication docs. This will encourage new practitioners to use environment variables, roles, or other authentication methods rather than hard-coding sensitive AWS creds in their Terraform files.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -38,10 +38,27 @@ The AWS provider offers a flexible means of providing credentials for
 authentication. The following methods are supported, in this order, and
 explained below:
 
+- Static credentials
 - Environment variables
 - Shared credentials file
 - EC2 Role
-- Static credentials
+
+### Static credentials ###
+
+!> WARNING: Hard-coding credentials into any Terraform file is not recommended and risks secret leakage should this file ever be committed to a public source code control system.
+
+Static credentials can be provided by adding an `access_key` and `secret_key`
+in-line in the AWS provider block:
+
+Usage:
+
+```hcl
+provider "aws" {
+  region     = "us-west-2"
+  access_key = "my-access-key"
+  secret_key = "my-secret-key"
+}
+```
 
 ### Environment variables
 
@@ -131,23 +148,6 @@ provider "aws" {
     session_name = "SESSION_NAME"
     external_id  = "EXTERNAL_ID"
   }
-}
-```
-
-### Static credentials ###
-
-!> WARNING: Hard-coding credentials into any Terraform file is not recommended and risks secret leakage should this file ever be committed to a public source code control system.
-
-Static credentials can be provided by adding an `access_key` and `secret_key` in-line in the
-AWS provider block:
-
-Usage:
-
-```hcl
-provider "aws" {
-  region     = "us-west-2"
-  access_key = "anaccesskey"
-  secret_key = "asecretkey"
 }
 ```
 
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -23,12 +23,9 @@ provider "aws" {
   region = "us-east-1"
 }
 
-# Create a web server
-resource "aws_instance" "web" {
-  ami           = "ami-2757f631"
-  instance_type = "t2.micro"
-
-  # ...
+# Create a VPC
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -36,25 +32,10 @@ The AWS provider offers a flexible means of providing credentials for
 authentication. The following methods are supported, in this order, and
 explained below:
 
-- Static credentials
```
bflad: Whether we recommend their usage or not, this authentication method list is ordered. We need to keep static credentials at the top as they take priority over all others.

You mention in your pull request comment that the goal is to encourage other methods so I believe rather than changing what is here we should just follow through by adding those recommendations in that documentation section (e.g. using `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables instead).

website/docs/index.html.markdown
```diff
@@ -17,15 +17,11 @@ Use the navigation to the left to read about the available resources.
 ## Example Usage
 
 ```hcl
-# Configure the AWS Provider
```
bflad: I believe this Terraform configuration for the provider is important here in some form, to show even a basic example of how it is done for newer practitioners.

I would suggest leaving a simple configuration without the credentials, such as the below:

```hcl
# Configure the AWS Provider
provider "aws" {
  region = "us-east-1"
}
```

website/docs/index.html.markdown
```diff
@@ -17,15 +17,11 @@ Use the navigation to the left to read about the available resources.
...
-provider "aws" {
-  access_key = "${var.aws_access_key}"
-  secret_key = "${var.aws_secret_key}"
-  region     = "us-east-1"
-}
-
 # Create a web server
 resource "aws_instance" "web" {
+  ami           = "ami-2757f631"
```
bflad: I would discourage hardcoding an AMI ID configuration that can/will become out of date or be potentially incorrect for environments (`ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20170307` in `us-east-1`). We may want to consider choosing another resource if we are trying to show a more valid configuration such as:

```hcl
resource "aws_vpc" "example" {
  cidr_block = "10.0.0.0/16"
}
```

website/docs/index.html.markdown
```diff
@@ -147,6 +128,23 @@ provider "aws" {
 }
 ```
 
+### Static credentials ###
+
+!> WARNING: Hard-coding credentials into any Terraform file is not recommended and risks secret leakage should this file ever be committed to a public source code control system.
```
nfagerlund:
```suggestion
!> **Warning:** Hard-coding credentials into any Terraform configuration is not recommended, and risks secret leakage should this file ever be committed to a public version control system.
```

website/docs/index.html.markdown
nfagerlund: Mm, probably good to put a version constraint there too.

bflad:
Hi @topfunky 👋 Are you planning on adjusting this pull request based on the above feedback or would you like us to?

website/docs/index.html.markdown
topfunky: Fixed in 71195a8d3619e13ec17a0059ed6714b7df852a0a with a version constraint.

website/docs/index.html.markdown
topfunky: Fixed in c89ce9076

website/docs/index.html.markdown
topfunky: Fixed in 2b2bcd156

website/docs/index.html.markdown
topfunky: Static credential section has been moved to top as suggested. Fixed in 945b612

topfunky:
Resolved and ready for review or merge.

PR #8318, opened 2019-04-15
Instrument to produce distributed traces for AWS API calls to make troubleshooting modules easier

jbarrick-mesosphere:
### Community Note

* Adds optional support for exporting distributed traces from the Terraform provider.


bflad:
Hi @jbarrick-mesosphere 👋 Thanks for submitting this! Definitely an interesting enhancement.

Normally when we see more impactful enhancements like these to the provider, we prefer some indication of the use cases for the functionality and some of the rationale behind the potential implementation discussed in a proposal issue first so we can ensure its right for the project. While the use case here seems fairly self-explanatory, any discussion on the choice of the implementation is not outlined in the pull request. The lack of this discussion makes it difficult to assess if this underlying library is maintained well or chosen over others for any particular reason.

Another facet of this enhancement is that this is only offered within the context of Terraform AWS Provider. Should this be offered to other Terraform Providers as well? If so, it seems like it might be best to have the majority of this implementation (and therefore a "chosen" underlying library) live in the Terraform Provider SDK, which currently lives upstream in Terraform Core. When that is in place, then any Terraform Provider could then hook directly into the tracing framework with minimal change and any configuration could be handled consistently across providers.

As such, regrettably I am going to close this pull request for now since its not likely we will accept it in its current form until some larger discussions happen, rather than keeping the pull request indefinitely open with no activity. My best recommendation right now would be to submit [a new GitHub issue upstream into Terraform Core](https://github.com/hashicorp/terraform/issues/new?labels=enhancement&template=feature_request.md), which outlines the use case(s) and underlying library selection process. 👍 

Thank you again for spending the time submitting this and hopefully we can see if this feature can land in a form that can be applied to all Terraform Providers.

PR #8347, opened 2019-04-16
tests/resource/aws_redshift_cluster: Add enable_logging removed attribute to ImportStateVerifyIgnore list

nywilken:
### Community Note

In the Terraform 0.12 Provider SDK acceptance testing framework, removed
attributes that do not call ResourceData.Set() are no longer
automatically ignored during ImportStateVerify testing. Here we add the
removed attribute to the ImportStateVerifyIgnore list to match similar
behavior of Terraform 0.11 acceptance testing. This change is backwards
compatible.

Previous output from Terraform 0.12 Provider SDK acceptance testing:

```
--- FAIL: TestAccAWSRedshiftCluster_importBasic (577.26s)
testing.go:568: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

  (map[string]string) {
  }

  (map[string]string) (len=1) {
   (string) (len=14) "enable_logging": (string) (len=5) "false"
  }
```

Output from Terraform 0.12 Provider SDK acceptance testing:

```
--- PASS: TestAccAWSRedshiftCluster_basic (476.37s)
```

### DIFF
--- a/aws/resource_aws_redshift_cluster_test.go
+++ b/aws/resource_aws_redshift_cluster_test.go
@@ -95,29 +95,6 @@ func TestValidateRedshiftClusterDbName(t *testing.T) {
 	}
 }
 
-func TestAccAWSRedshiftCluster_importBasic(t *testing.T) {
-	resourceName := "aws_redshift_cluster.default"
-	config := testAccAWSRedshiftClusterConfig_basic(acctest.RandInt())
-
-	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:     func() { testAccPreCheck(t) },
-		Providers:    testAccProviders,
-		CheckDestroy: testAccCheckAWSRedshiftClusterDestroy,
-		Steps: []resource.TestStep{
-			{
-				Config: config,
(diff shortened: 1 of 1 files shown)


aws/resource_aws_redshift_cluster_test.go
```diff
@@ -140,6 +117,16 @@ func TestAccAWSRedshiftCluster_basic(t *testing.T) {
 					resource.TestMatchResourceAttr("aws_redshift_cluster.default", "dns_name", regexp.MustCompile(fmt.Sprintf("^tf-redshift-cluster-%d.*\\.redshift\\..*", ri))),
 				),
 			},
+			{
+				ResourceName:      "aws_redshift_cluster.default",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"enable_logging",
```
nywilken: Does it make sense to ignore removed attributes or should we be checking that the new attribute is set on import?

PR #8365, opened 2019-04-17
[docs] cloudwatch_metric_alarm

ChrisLundquist:
There is a spurious %d in a resource name that makes it seem like it
will use format, and print a number when used with `count`.
This is not the case.

This removes the spurious %d, and adds an example using `format` with
`count`, as I *suspect* this could attribute to issues like:

- https://github.com/terraform-providers/terraform-provider-aws/issues/422

I along with at least one other user, noted that really they had mistakenly
duplicate alarm names. I suspect this could be the case for others as
well.

My hope is that by providing an example to do something like that above,
fewer issues will be opened.

### Community Note

### DIFF
--- a/website/docs/r/cloudwatch_metric_alarm.html.markdown
+++ b/website/docs/r/cloudwatch_metric_alarm.html.markdown
@@ -57,11 +57,27 @@ resource "aws_cloudwatch_metric_alarm" "bat" {
 }
 ```
 
+## Example for Alarms per Resource
+```
+resource "aws_cloudwatch_metric_alarm" "nat_gateway_egress" {
+  alarm_name          = "${format("terraform-test-nat-gateway-egress-%d", count.index)}"
+  comparison_operator = "LessThanOrEqualToThreshold"
+  evaluation_periods  = "2"
+  metric_name         = "BytesOutToDestination"
+  namespace           = "AWS/NATGateway"
+  period              = "120"
+  statistic           = "Average"
+  threshold           = "0"
+  count               = "${length(aws_nat_gateway.gateways.*.id)}"
+}
+```
+
+
 ## Example with an Expression
 
 ```hcl
 resource "aws_cloudwatch_metric_alarm" "foobar" {
-  alarm_name                = "terraform-test-foobar%d"
+  alarm_name                = "terraform-test-foobar"
   comparison_operator       = "GreaterThanOrEqualToThreshold"
   evaluation_periods        = "2"
   threshold                 = "10"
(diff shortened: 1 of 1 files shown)


ChrisLundquist:
Sure, I’d be happy to.
I agree it shouldn’t be used everywhere, it seemed like this particular
page might benefit though as I saw a few open issues with misleading error
messages, that boiled down to “you have the same alarm name associated with
multiple things”

I think folks looked at the examples here, didn’t generate unique names,
then were confused by the error message along the lines of, “Validation
error: resource is already being modified” and wanted to chalk it up to a
race condition. I know I hit it too, and thought I’d get a planning error
trying to associate a 1:1 relation thing as a 1:N relation. (Even though my
intent was to generate N 1:1 relations)

To recap, I’m happy to delete the new example. It seems like there is some
snag that is tripping up people enough to open issues and up vote them.
My hope was we could do a doc fix, do you think there is a better place to
fix this?

Chris Lundquist

ChrisLundquist:
Updated at requested.

bflad:
> do you think there is a better place to fix this?

Hmm. I did some poking around of the existing Terraform documentation and could not find an immediate good home for this type of documentation. I believe we have two reasonable options here, but both are probably worth additional input from other upstream Terraform maintainers before we pursue either route:

* Create a new Terraform documentation page or guide that shows various randomization strategies, e.g. `count.index` and `random_XXX` resources. With that in place we can implement updates to our arguments in all resources that may require this, such as: `This argument requires unique naming, see the [Terraform guide on naming randomization](/docs/XXX) for various strategies.`
* Link to various randomization strategies in our arguments directly, e.g. `This argument requires unique naming (e.g. [count.index](https://www.terraform.io/docs/configuration/resources.html#count-multiple-resource-instances) or resources from the [Terraform Random Provider](https://www.terraform.io/docs/providers/random/index.html)).`

My worry with the latter though is that it becomes a maintenance burden as the copy can become inconsistent easily.

If you're interested, its probably best to submit an upstream Terraform issue to garner additional input as this affects more resource documentation than just the Terraform AWS Provider: https://github.com/hashicorp/terraform/issues/new?labels=enhancement&template=feature_request.md 👍

nywilken:
This has been released in [version 2.8.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#280-april-26-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #8446, opened 2019-04-25
fix: doc api_gateway_integration example wrong source arn 

nijohando:
source_arn contians duplicate slash.

```
 source_arn = "arn:aws:execute-api:${var.myregion}:${var.accountId}:${aws_api_gateway_rest_api.api.id}/*/${aws_api_gateway_method.method.http_method}/${aws_api_gateway_resource.resource.path}"
```

The value of `${aws_api_gateway_resource.resource.path}` starts with `/`, so `/${aws_api_gateway_resource.resource.path}`  makes incorrect arn with duplicate slash.


bflad:
Hi @aeschright 👋 I noticed this was marked for today's release but wasn't sure if there was a specific reason why it was not merged so I did not merge it myself. For documentation pull requests that are approved, they can usually be merged immediately, especially since they won't include a CHANGELOG update. 👍

So if you think this is good to go, please feel free to merge, otherwise please do reach out if you are waiting for a second opinion.

PR #8495, opened 2019-05-01
Specifying default behavior network interface delete_on_termination for launch templates.

willswank:
Improves the documentation for network interface `delete_on_termination` for launch templates by specifying the default behavior as `False`.  Current documentation can be found at: https://www.terraform.io/docs/providers/aws/r/launch_template.html#network-interfaces

Perhaps the default behavior should be `True` but I believe there is a separate issue opened for that proposal.

This is worth posting to the site immediately for the current version

### DIFF
--- a/website/docs/r/launch_template.html.markdown
+++ b/website/docs/r/launch_template.html.markdown
@@ -252,7 +252,7 @@ Check limitations for autoscaling group in [Creating an Auto Scaling Group Using
 Each `network_interfaces` block supports the following:
 
 * `associate_public_ip_address` - Associate a public ip address with the network interface.  Boolean value.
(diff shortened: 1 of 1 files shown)


savar:
that's actually not the right thing to do. The documentation is correct. The terraform-provider-aws has a bug inside.

bflad:
Hi @willswank 👋  Thank you for submitting this and apologies this was missed previously during related efforts. It appears a similar documentation change was handled via https://github.com/terraform-providers/terraform-provider-aws/pull/13546. We anticipate the behavior fixes for this argument will occur with #8612, to be released with version 3.0.0 of the Terraform AWS Provider.

PR #8517, opened 2019-05-03
Include access_key, secret_key, and session_token in aws_caller_identity

### LINKED ISSUE #8242: Pass credentials to local-exec OR extract credentials via properties
### Community Note

### Description

If you need to execute a local command, via local-exec provisioner - whether part of a standard aws resource or via `null_resource` - the AWS credentials are not passed to that command. Almost all implementations of AWS CLI/SDK accept the usage of `AWS_PROFILE`,`AWS_ACCESS_KEY_ID`,`AWS_SECRET_ACCESS_KEY`, so any could consume it.

Reasoning: I might have something I want to execute outside of terraform due to unsupported resources (e.g. directory upload #3020 ) or to keep some things out of terraform state, or to go through compliance.

In theory, those commands _should_ have access to the AWS vars with which I launched terraform, but that usually isn't good enough for several reasons:

* I might have assumed a role inside my `provider "aws"` clause
* I might have picked a different profile inside my `provider "aws"` clause
* I might have multiple `provider "aws"` clauses, each with a different alias, and would want to run one command with one instance, another command with the other

Further, `provider "aws"` already has all of the support for fixed credentials, profiles, and even assuming roles to get temporary credentials via `sts`. A local command should, in principle, run just like any other resource.

I see two ways of doing this:

1. When executing `local-exec` provisioners, pass the relevant credentials - `AWS_SECRET_ACCESS_KEY` / `AWS_ACCESS_KEY_ID` - into the environment. I am not sure this can be done directly. If it can, I believe it only would work if the `local-exec` provisioner is attached to an `aws_*` resource, and not a `null_resource`.
2. Provide a way to extract the credentials used for the connection. The best candidate appears to be extending `aws_caller_identity`, although there might be a better way.

The first case would be straightforward:

```hcl
resource "aws_launch_configuration" "my_config" {
    # lots of other stuff
    provisioner "local-exec" {
      command = "some-command"
      environment {
        # automatically includes AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
      }
}
```

The second case is just explicit:

```hcl
resource "null_resource" "my_resource" {
  provisioner "local-exec" {
    command = "some-command"
    environment {
      AWS_ACCESS_KEY_ID = "${data.aws_caller_identity.current.access_key}"
      AWS_ACCESS_KEY_ID = "${data.aws_caller_identity.current.secret_key}"
    }
  }
}
```

For this to work, we would need to ensure that `access_key` and `secret_key` are the _actual_ ones used, e.g. temporary credentials via assume role or similar, or ones retrieved via using a profile, and not ones passed in via the `provider "aws"` config, unless those are the ones actually being used.

* provider itself
* possible `aws_caller_identity`

See the examples above.

Thanks!

twang817:
This allows local-exec provisioners to be able to retrieve the
credentials that the aws provider is using.

Ex:
```
provider "aws" {}

resource "null_resource" "my_resource" {
  provisioner "local-exec" {
    command = "some-command"
    environment {
      AWS_ACCESS_KEY_ID = "${data.aws_caller_identity.current.access_key}"
      AWS_SECRET_ACCESS_KEY = "${data.aws_caller_identity.current.secret_key}"
      AWS_SESSION_TOKEN = "${data.aws_caller_identity.current.session_token}"
    }
  }
}
```

### Community Note

Fixes #8242

```release-note
Allows credentials (access_key, secret_key, and session_token) to be fetched from aws_caller_identity to be used within local-exec provisioners.
```

```

...
```

### DIFF
--- a/aws/data_source_aws_caller_identity.go
+++ b/aws/data_source_aws_caller_identity.go
@@ -28,6 +28,21 @@ func dataSourceAwsCallerIdentity() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+
+			"access_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"secret_key": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"session_token": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 		},
 	}
 }
@@ -44,10 +59,18 @@ func dataSourceAwsCallerIdentityRead(d *schema.ResourceData, meta interface{}) e
 
 	log.Printf("[DEBUG] Received Caller Identity: %s", res)
 
+	creds, err := client.Config.Credentials.Get()
+	if err != nil {
+		return fmt.Errorf("Error getting credentials: %v", err)
+	}
+
 	d.SetId(time.Now().UTC().String())
 	d.Set("account_id", res.Account)
 	d.Set("arn", res.Arn)
 	d.Set("user_id", res.UserId)
+	d.Set("access_key", creds.AccessKeyID)
+	d.Set("secret_key", creds.SecretAccessKey)
+	d.Set("session_token", creds.SessionToken)
 
 	return nil
 }
--- a/aws/data_source_aws_caller_identity_test.go
+++ b/aws/data_source_aws_caller_identity_test.go
@@ -64,6 +64,14 @@ func testAccCheckAwsCallerIdentityAccountId(n string) resource.TestCheckFunc {
 			return fmt.Errorf("ARN expected to not be nil")
 		}
 
+		if rs.Primary.Attributes["access_key"] == "" {
+			return fmt.Errorf("Access Key expected to not be nil")
+		}
+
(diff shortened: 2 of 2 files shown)


twang817:
Having a little trouble running tests on my dinky laptop. Will try again on a bigger machine when I get the chance.

jamroks:
Hi ,

Can we have an update on this issue, is it good enough to be merged or is there anything still blocking in order for it to be reviewed and merge ?  

Thanks for clarification

AyliD:
Any update on that?

twang817:
I wrote this PR, but having recently been more involved with my company's security organization, I don't believe that this is a good solution to the problem.

Having keys stored in the tfstate, even if temporary, is a generally bad idea.  Even worse, this implementation doesn't give users that may need aws_caller_identity a choice.  Users could unwittingly publishing their potentially long-term keys through their tfstate into an unencrypted S3 bucket (or worse, Github!).

A slightly better solution would be to break this out into a separate data resource that can be used explicitly to fetch keys for those that are unconcerned with the security implications.  I mentioned that I had done this in my last comment in #8242.  I will close this and open a new PR with those changes.

PR #8535, opened 2019-05-06
Resource xray sampling rules

### LINKED ISSUE #7028: Support for X-Ray Sampling Rules
### Community Note

### Description

AWS xray allows for creating sampling rules and groups to manipulate the data stream. It would be nice to support this configuration in TF.

* aws_xray (new)

### References

https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sampling.html

dedoussis:
### Community Note

Fixes #7028

```release-note
New Resource: aws_xray_sampling_rule (#7028)
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSXraySamplingRule'

TF_ACC=1 go test ./aws -v -parallel 20 -run=TestAccAWSXraySamplingRule -timeout 120m
=== RUN   TestAccAWSXraySamplingRule_basic
=== PAUSE TestAccAWSXraySamplingRule_basic
=== RUN   TestAccAWSXraySamplingRule_update
=== PAUSE TestAccAWSXraySamplingRule_update
=== RUN   TestAccAWSXraySamplingRule_import
=== PAUSE TestAccAWSXraySamplingRule_import
=== CONT  TestAccAWSXraySamplingRule_basic
=== CONT  TestAccAWSXraySamplingRule_update
=== CONT  TestAccAWSXraySamplingRule_import
--- PASS: TestAccAWSXraySamplingRule_basic (18.06s)
--- PASS: TestAccAWSXraySamplingRule_import (19.33s)
--- PASS: TestAccAWSXraySamplingRule_update (37.13s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	37.193s
```
 
### References
https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sampling.html


aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(1, 128),
+			},
+			"resource_arn": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"priority": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.IntBetween(1, 9999),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeInt,
+				Required: true,
+			},
+			"fixed_rate": {
+				Type:     schema.TypeFloat,
+				Required: true,
+			},
+			"reservoir_size": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.IntAtLeast(0),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeInt,
+				Required: true,
+			},
+			"service_name": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.StringLenBetween(0, 64),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"service_type": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.StringLenBetween(0, 64),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"host": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.StringLenBetween(0, 64),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"http_method": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.StringLenBetween(0, 10),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"url_path": {
```
bflad: We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ValidateFunc: validation.StringLenBetween(0, 128),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"version": {
```
bflad: Two notes:
* Rather than return an error during update, this attribute should use [`ForceNew`](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#forcenew) to trigger Terraform resource recreation
* We can add validation for this attribute to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
ForceNew: true,
ValidateFunc: validation.IntAtLeast(1),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeInt,
+				Required: true,
+			},
+			"attributes": {
```
bflad: Two notes:

* While Terraform does not explicitly error during resource schema validation without it currently, we prefer to add the explicit `Elem` declaration to delineate the underlying map value type, as it likely will when a future version of the Terraform Provider SDK supports multiple map value types. I don't see an explicit `Elem` section in the [Extending Terraform documentation section on schemas](https://www.terraform.io/docs/extend/schemas/index.html) so I will make a note to ensure this gets added.
* We can add validation for this attribute's keys and values to match the [API reference](https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html) via:

```go
Elem: &schema.Schema{
	Type: schema.TypeString,
	ValidateFunc: validation.StringLenBetween(1, 32),
},
ValidateFunc: validation.StringLenBetween(1, 32),
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+				Type:     schema.TypeMap,
+				Optional: true,
+			},
+			"rule_arn": {
```
bflad: While `rule_arn` strictly follows the API attribute naming, we typically use `arn` as the attribute name across all Terraform AWS Provider resources to denote the ARN of the AWS resource under management. 👍

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+		samplingRule.Attributes = stringMapToPointers(v.(map[string]interface{}))
+	}
+
+	params := &xray.CreateSamplingRuleInput{
+		SamplingRule: samplingRule,
+	}
+
+	out, err := conn.CreateSamplingRule(params)
+	if err != nil {
+		return err
```
bflad: We should provide additional context about this error here for operators and code maintainers, e.g.

```suggestion
		return fmt.Errorf("error creating XRay Sampling Rule: %s", err)
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+	d.SetId(*out.SamplingRuleRecord.SamplingRule.RuleName)
+
+	return resourceAwsXraySamplingRuleRead(d, meta)
+}
+
+func resourceAwsXraySamplingRuleRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).xrayconn
+	params := &xray.GetSamplingRulesInput{}
+
+	for {
```
bflad: This seems like it should be simplified to use the `getSamplingRule` function (I have some additional notes associated with that function as well below). 😄 e.g.

```go
samplingRule, err := getXraySamplingRule(conn, d.Id())

if err != nil {
	return fmt.Errorf("error reading XRay Sampling Rule (%s): %s", d.Id(), err)
}

if samplingRule == nil {
	log.Printf("[WARN] XRay Sampling Rule (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}

d.Set("rule_name", samplingRule.RuleName)
// ... set other attributes ...
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+		ReservoirSize: aws.Int64(int64(d.Get("reservoir_size").(int))),
+		ServiceName:   aws.String(d.Get("service_name").(string)),
+		ServiceType:   aws.String(d.Get("service_type").(string)),
+		Host:          aws.String(d.Get("host").(string)),
+		HTTPMethod:    aws.String(d.Get("http_method").(string)),
+		URLPath:       aws.String(d.Get("url_path").(string)),
+	}
+
+	if d.HasChange("version") {
+		return fmt.Errorf("Version cannot be modified")
```
bflad: As noted above, we should add `ForceNew: true,` to the attribute and remove this conditional. 👍

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+		samplingRuleUpdate.Attributes = attributes
+	}
+
+	params := &xray.UpdateSamplingRuleInput{
+		SamplingRuleUpdate: samplingRuleUpdate,
+	}
+
+	_, err := conn.UpdateSamplingRule(params)
+	if err != nil {
+		return err
```
bflad: We should provide context for this error here for operators and code maintainers, e.g.

```suggestion
		return fmt.Errorf("error updating XRay Sampling Rule (%s): %s", d.Id(), err)
```

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+	conn := meta.(*AWSClient).xrayconn
+
+	log.Printf("[INFO] Deleting XRay Sampling Rule: %s", d.Id())
+
+	params := &xray.DeleteSamplingRuleInput{
+		RuleName: aws.String(d.Id()),
+	}
+	_, err := conn.DeleteSamplingRule(params)
+	if err != nil {
+		return fmt.Errorf("Error deleting XRay Sampling Rule: %s", d.Id())
```
bflad: This error message should include the error as well 😄 

```suggestion
		return fmt.Errorf("error deleting XRay Sampling Rule (%s): %s", d.Id(), err)
```

aws/resource_aws_xray_sampling_rule_test.go
```diff
@@ -0,0 +1,209 @@
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSXraySamplingRuleDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSXraySamplingRuleConfig_basic(ruleName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckXraySamplingRuleExists(resourceName, &samplingRule),
+					testAccCheckResourceAttrRegionalARN(resourceName, "rule_arn", "xray", fmt.Sprintf("sampling-rule/%s", ruleName)),
+					resource.TestCheckResourceAttrSet(resourceName, "priority"),
```
bflad: For these `TestCheckResourceAttrSet` checks here we should prefer to actually check the value in the state to ensure it is what we expect from the configuration, e.g.

```go
resource.TestCheckResourceAttr(resourceName, "priority", "5"),
// ... repeat for below attributes ...
```

aws/resource_aws_xray_sampling_rule_test.go
```diff
@@ -0,0 +1,209 @@
...
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSIotThingTypeDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSXraySamplingRuleConfig_basic(ruleName),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
```
bflad: Instead of creating a separate full acceptance test which implements import testing for one configuration, we should instead add this `TestStep` to all other tests above and remove this test. This is related to the [Implements ImportState Testing section of the contributing guide](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#acceptance-testing-guidelines) 👍

aws/resource_aws_xray_sampling_rule_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+		if err == nil {
+			return fmt.Errorf("Expected XRay Sampling Rule to be destroyed, %s found", rs.Primary.ID)
+		}
+	}
+
+	return nil
+}
+
+func getSamplingRule(ruleName string) (*xray.SamplingRule, error) {
```
bflad: This function is super handy and should be used in the resource as well. I would recommend moving it to `aws/resource_aws_xray_sampling_rule.go` and updating the signature to `getXraySamplingRule(conn *xray.XRay, ruleName string) (*xray.SamplingRule, error)`

aws/resource_aws_xray_sampling_rule_test.go
```diff
@@ -0,0 +1,209 @@
...
+		if err != nil {
+			return nil, err
+		}
+		for _, samplingRuleRecord := range out.SamplingRuleRecords {
+			samplingRule := samplingRuleRecord.SamplingRule
+			if aws.StringValue(samplingRule.RuleName) == ruleName {
+				return samplingRule, nil
+			}
+		}
+		if out.NextToken == nil {
```
bflad: We have seen APIs or the AWS Go SDK occasionally misbehave and return an empty string so I would recommend checking it as such:

```suggestion
		if aws.StringValue(out.NextToken) == "" {
```

aws/resource_aws_xray_sampling_rule_test.go
```diff
@@ -0,0 +1,209 @@
...
+			break
+		}
+		params.NextToken = out.NextToken
+	}
+	return nil, fmt.Errorf("XRay Sampling Rule: %s not found found", ruleName)
```
bflad: If we return no error here, then we can let downstream functions decide how they would prefer to handle the missing sampling rule rather than trying to catch and parse this error message.

```suggestion
	return nil, nil
```

aws/resource_aws_xray_sampling_rule_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+func testAccCheckAWSXraySamplingRuleDestroy(s *terraform.State) error {
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_xray_sampling_rule" {
+			continue
+		}
+
+		_, err := getSamplingRule(rs.Primary.ID)
+
+		if err == nil {
```
bflad: When `getXraySamplingRule` is updated with the last comment, this logic becomes:

```go
if err != nil {
  return err
}

if samplingRule != nil {
  return fmt.Errorf("Expected XRay Sampling Rule to be destroyed, %s found", rs.Primary.ID)
}
```

website/docs/r/xray_sampling_rule.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+* `http_method` - (Required) Matches the HTTP method of a request.
+* `url_path` - (Required) Matches the path from a request URL.
+* `version` - (Required) The version of the sampling rule format (`1` )
+* `attributes` - (Optional) Matches attributes derived from the request.
+
+## Attributes Reference
+
+In addition to the arguments above, the following attributes are exported:
+
+* `rule_arn` - The ARN of the sampling rule.
```
bflad: Two notes:

* As noted above we should update this attribute to just `arn`
* We should include the `id` attribute here, which is the rule name

dedoussis:
Thanks for reviewing this @bflad ! 😃 
I do have time to implement any needed fixes. Have already picked up some of your comments.

Regarding the vendoring, I do get that SDKs can get outdated but do not understand how a separate PR solves this.

aws/resource_aws_xray_sampling_rule.go
dedoussis: Should we similarly convert `rule_name` to just `name`?

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"rule_name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(1, 128),
+			},
+			"resource_arn": {
```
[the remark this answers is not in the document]
dedoussis: This does not allow `"*"`, which is the default value used by AWS

aws/resource_aws_xray_sampling_rule.go
dedoussis: Added the `Elem` part -  However, `ValidateFunc: validation.StringLenBetween(1, 32),` on the TypeMap level causes this:
Input:
```hcl
...
     attributes = {
         Hello = "World"
     }
...
```
Output:
`testing.go:568: Step 0 error: config is invalid: expected type of attributes to be string`

aws/resource_aws_xray_sampling_rule.go
```diff
@@ -0,0 +1,221 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"priority": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				ValidateFunc: validation.IntBetween(1, 9999),
+			},
+			"fixed_rate": {
```
dedoussis: @bflad Do you think this would require a custom validator as the validation helper does not support anything along the lined of `FloatBetween`?

bflad:
> Regarding the vendoring, I do get that SDKs can get outdated but do not understand how a separate PR solves this.

The maintainers generally prioritize merging vendoring changes before other code reviews as they are a fast moving target with how often the AWS Go SDK is released. If the AWS Go SDK for one service is out of sync version-wise with the rest of the services, it will show as unexpected changes the next time vendoring occurs. Since the AWS Go SDK is generally additive _usually_ this is not a problem, but there have been rare cases in the past where that did not hold. Separating the SDK client addition, e.g. as #8512 did for `managedblockchain`, allows us to immediately get the service SDK into the codebase. 

Once Go Module caching becomes more publicly stable and we feel comfortable removing the vendoring, this will no longer be a problem. 👍

aws/resource_aws_xray_sampling_rule.go
bflad: It appears there are two upstream PRs for adding `FloatBetween` to the `helper/validation` package upstream:

* https://github.com/hashicorp/terraform/pull/18614
* https://github.com/hashicorp/terraform/pull/20877

I cannot imagine there is a specific reason why one of these is not merged, so I will followup with those maintainers to make sure the new validation function gets in. We are likely pulling in a newer version of the upstream dependency later this week anyways as part of some Terraform 0.12 work, so this problem might resolve itself shortly. I would suggest omitting any changes in this regard for now.

aws/resource_aws_xray_sampling_rule.go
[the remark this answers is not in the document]
bflad: If `"*"` is the only exception, then we should be able to do the following with [`validation.Any()`](https://godoc.org/github.com/hashicorp/terraform/helper/validation#Any):

```go
ValidateFunc: validation.Any(
	validateArn,
	validation.StringInSlice([]string{"*"}),
),
```

Otherwise, perfectly fine to ignore validation for now!

aws/resource_aws_xray_sampling_rule.go
bflad: Acceptable either way, honestly. 😄 Its just ARN that is always shortened.

dedoussis:
@bflad Have addressed all your comments. 😺

aws/resource_aws_xray_sampling_rule.go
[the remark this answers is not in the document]
dedoussis: I will ignore validation for now as the resource should accept ARNs such as  `arn:aws:lambda:*:*:function:*-dev`

bflad:
This has been released in [version 2.10.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2100-may-10-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

PR #8656, opened 2019-05-16
💄 Added so aws_lb_listener_rule works with "" on priority

Tazer:
Before when wanting "priority" to be auto-assigned, you would need to create a rule without "priority" not defined at all, this will add so it's the same behavior when having empty string, cause else you need 2 resources if you want to create a "module" that is the rule resource. Cause passing empty string right now breaks the code cause it tries to parse "" to int. 

### Community Note

### DIFF
--- a/aws/resource_aws_lb_listener_rule.go
+++ b/aws/resource_aws_lb_listener_rule.go
@@ -467,7 +467,7 @@ func resourceAwsLbListenerRuleCreate(d *schema.ResourceData, meta interface{}) e
 	}
 
 	var resp *elbv2.CreateRuleOutput
-	if v, ok := d.GetOk("priority"); ok {
+	if v, ok := d.GetOk("priority"); ok && v != "" {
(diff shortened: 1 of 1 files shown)


bflad:
Hi @Tazer 👋 Thanks for submitting this.

These types of workarounds should no longer be a necessary workaround with Terraform 0.12, which was [released last week](https://www.hashicorp.com/blog/announcing-terraform-0-12), and as such we are not likely to continue adding Terraform 0.11 syntax workarounds into resources.

Terraform 0.12 has support for a `null` value, which effectively hides the argument configuration from the resource. e.g. the following contrived example should automatically assign the priority:

```hcl
resource "aws_lb_listener_rule" "example" {
  # ... other configuration
  priority = null
}
```

Can you verify that using/passing `null` in your situation works as expected? Thanks.

bflad:
Closing due to lack of response.

PR #8657, opened 2019-05-16
[WIP] ACM Certificate: sort domain validation options by domain name

jtsaito:
This PR ads a fix for https://github.com/terraform-providers/terraform-provider-aws/issues/8531.

Domain validation options (DVOs) are sorted to guarantee the same order even when the AWS API returns them jumbled up. To order imposed is as follows: (1) the certificate's domain name comes first, (2) the remaining options are sorted by their respective `domain_name` attribute alphabetically.

This PR does not address the issue of unsorted `subject alternative names`. However, these may be ignored by lifecycle `ignore_changes`.

I have included unit tests but I'm not sure if they belong here at all.

I would be grateful if someone could run the acceptance tests.

### Community Note

Fixes: https://github.com/terraform-providers/terraform-provider-aws/issues/8531

```release-note
For several regions [it has been reported](https://github.com/terraform-providers/terraform-provider-aws/issues/8531) that the ACM API returns randomly sorted *subject alternative names* (SANs) and *domain validation options* (DVOs) elements. This resulted in non-empty plans after apply, indefinitely.

The release imposes an ordering on only the DVOs attribute lists. As a consequence, there will be a one-time re-ordering of the elements.

If you apply this change in a region not affected before, this is the first and only time you will notice the change. Dependent resources maybe require re-creating. In regions affected by the API change already, the problem should go away after applying one more time.

You may chose to ignore the changing ordering on the SANs.
```

I did not run the acc tests on this because I did not want to run the email validation. However, I tested the implementation by compiling and using it.

### DIFF
--- a/aws/resource_aws_acm_certificate.go
+++ b/aws/resource_aws_acm_certificate.go
@@ -322,17 +322,17 @@ func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]
 // the remaining options are sorted alphabetically by their domain name.
 // Sorting is necessary because the API endpoint returns the validation options in random
 // order different each time.
-func sortDomainValidationOptions(domainName *string, domainValiationOptions []map[string]interface{}) {
-	sort.Slice(domainValiationOptions, func(i, j int) bool {
-		if domainValiationOptions[i]["domain_name"].(string) == *domainName {
+func sortDomainValidationOptions(domainName *string, domainValidationOptions []map[string]interface{}) {
+	sort.Slice(domainValidationOptions, func(i, j int) bool {
+		if domainValidationOptions[i]["domain_name"].(string) == *domainName {
 			return true
 		}
-		if domainValiationOptions[j]["domain_name"].(string) == *domainName {
+		if domainValidationOptions[j]["domain_name"].(string) == *domainName {
 			return false
 		}
 
-		return domainValiationOptions[i]["domain_name"].(string) <
-			domainValiationOptions[j]["domain_name"].(string)
+		return domainValidationOptions[i]["domain_name"].(string) <
+			domainValidationOptions[j]["domain_name"].(string)
 	})
 }
 
--- a/aws/resource_aws_acm_certificate.go
+++ b/aws/resource_aws_acm_certificate.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"sort"
 	"strings"
 	"time"
 
@@ -215,6 +216,7 @@ func resourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) err
 		}
 
 		domainValidationOptions, emailValidationOptions, err := convertValidationOptions(resp.Certificate)
+		sortDomainValidationOptions(resp.Certificate.DomainName, domainValidationOptions)
 
 		if err != nil {
 			return resource.RetryableError(err)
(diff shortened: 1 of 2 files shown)


aws/resource_aws_acm_certificate.go
```diff
@@ -315,6 +317,25 @@ func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]
 	return domainValidationResult, emailValidationResult, nil
 }
 
+// sortDomainValidationOptions sorts a slice of domain validation options as follows:
+// the option corresponding to the certficate's domain name is first;
+// the remaining options are sorted alphabetically by their domain name.
+// Sorting is necessary because the API endpoint returns the validation options in random
+// order different each time.
+func sortDomainValidationOptions(domainName *string, domainValiationOptions []map[string]interface{}) {
+	sort.Slice(domainValiationOptions, func(i, j int) bool {
```
ddfreyne: Typo: a `d` is missing in `domainValiationOptions` (validation, not valiation)

ddfreyne:
Question: Would it make sense for the domain validations to be ordered by the matching order in `subject_alternative_names`? IMO, that would be more predictable from an end-user point of view.

rifelpet:
It might be worth mentioning in the release notes that users may see a change in ordering the first time they plan and apply after upgrading the provider, especially since the original issue is only impacting certain regions.

tdmalone:
^ Yes, this definitely shouldn’t be `NONE` for the release notes. It’s a bug fix that users need to know about.

_Thank you_ for doing this though!!

jtsaito:
@rifelpet @tdmalone Thanks for your useful feedback. I have added a comment for the release note, please check the updated description above.

jtsaito:
@ddfreyne Thanks for your constructive feedback. I have opened a PR that includes your suggestion to sort *domain validation options* (DVOs) according to the ordering of the *subject alternative names* (SANs) (https://github.com/terraform-providers/terraform-provider-aws/pull/8708). 

The current PR suggests to ignore the problem of unsorted SANs and only sort the *domain validation names (the SANs could be ignored with `ignore_change`).  https://github.com/terraform-providers/terraform-provider-aws/pull/8708 now has to solve sorting the SANs first and then the DVOs accordingly otherwise both keep changing from apply to apply.

I have a slight preference for https://github.com/terraform-providers/terraform-provider-aws/pull/8708 because it keeps both attribute sets consistent and won't require `inogre_changes`. (However, the problem of unordered SANs seems to be less of a problem the way I read the comments in https://github.com/terraform-providers/terraform-provider-aws/issues/8531. Someone will correct me if I'm wrong :) And the first PR has a smaller footprint.)

Please let me know what you think.

jamescarr:
Thanks, we can really benefit from this fix... we have many people running applies many times a day and each time they see the ACM change they hit the breaks and start asking around. Sinks a bit of time throughout the week! :-)

jonseymour:
@jtsaito, @tdmalone  I'd like to comment that I think that a SANs fix is needed as well. 

The `ignore_changes` workaround for that problem is only acceptable if you *never* change the list of SANs, particularly if you need to edit an existing name. If you do, then there is a chance that required changes won't be applied because of the `ignore_changes` declaration.

tdmalone:
@jonseymour I believe the alternative PR mentioned above - #8708 - covers that

jonseymour:
@tdmalone Understood. My point is that i think #8708 should not be rejected in favour of this PR, because I don't think this PR alone is sufficient to address all the issues raised by the recent change in AWS behaviour - the other issues area real and need to be addressed, otherwise manual, out of band, workarounds will be required in a variety of scenarios which would be counter to objectives using terraform in the first place.

jtsaito:
I'm closing this PR. According to https://github.com/mlafeldt a fix for the issue should be rolled out by AWS until about 1. June 2019. Also, https://github.com/terraform-providers/terraform-provider-aws/pull/8708 would now definitely a better solution.

Issue #8531, opened 2019-05-06
aws_acm_certificate subject_alternative_names & domain_validation_options get returned in a different order each time


tdmalone:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
Terraform v0.11.13
+ provider.aws v2.8.0
```

### Affected Resource(s)

* [aws_acm_certificate](https://www.terraform.io/docs/providers/aws/r/acm_certificate.html)

### Terraform Configuration Files

```hcl
resource "aws_acm_certificate" "main" {
  domain_name = "example.com"
  validation_method = "DNS"

  subject_alternative_names = [
    "one.example.com",
    "two.example.com",
    "three.example.com",
    "four.example.com",
    "five.example.com",
    "six.example.com",
    "seven.example.com",
    "eight.example.com",
    "nine.example.com",
  ]
}

output "domain_validation_options" {
  value = "${aws_acm_certificate.main.domain_validation_options}"
}

```

### Expected Behavior

* There should be no diff after creating the certificate.
* The `domain_validation_options` output should contain the same results, in the same order, each time it is refreshed.

### Actual Behavior

The `subject_alternative_names` re-order on each apply, presenting a diff that would recreate the certificate:

```
-/+ aws_acm_certificate.main (new resource required)
      id:                          "arn:aws:acm:ap-southeast-2:xxxxxxxxxx:certificate/4c16e4c2-b77e-46f7-82e4-37aa5145c95f" => <computed> (forces new resource)
      arn:                         "arn:aws:acm:ap-southeast-2:xxxxxxxxxx:certificate/4c16e4c2-b77e-46f7-82e4-37aa5145c95f" => <computed>
      domain_name:                 "example.com" => "example.com"
      domain_validation_options.#: "10" => <computed>
      subject_alternative_names.#: "9" => "9"
      subject_alternative_names.0: "five.example.com" => "one.example.com" (forces new resource)
      subject_alternative_names.1: "seven.example.com" => "two.example.com" (forces new resource)
      subject_alternative_names.2: "nine.example.com" => "three.example.com" (forces new resource)
[...]

PR #8708, opened 2019-05-19
Sort ACM cert subject alternative names and domain validation opts

jtsaito:
This PR is an alternative implementation to https://github.com/terraform-providers/terraform-provider-aws/pull/8657 addressing https://github.com/terraform-providers/terraform-provider-aws/pull/8657#issuecomment-493100649. It ads a fix for https://github.com/terraform-providers/terraform-provider-aws/issues/8531.

The advantage of this implementation in comparison to https://github.com/terraform-providers/terraform-provider-aws/pull/8657 is that the *subject alternative names* (SANs) are also ordered and they will be ordered consistently with the DVOs.

We sort the SANs as specified in the Terraform file and subsequently sort the *domain validation options* (DVOs) such that the option for the certificate's domain name comes first and the remainder is sorted according to the order of the SANs.

I have included unit tests but I'm not sure if they belong here at all. (Obviously, the are passing.)

I would be grateful if someone could run the acceptance tests as I don't want to handle setting up the email validation.

### Community Note

Fixes: https://github.com/terraform-providers/terraform-provider-aws/issues/8531

```release-note
For several regions it has been reported (https://github.com/terraform-providers/terraform-provider-aws/issues/8531) that the ACM API returns randomly sorted *subject alternative names* (SANs) and *domain validation options* (DVOs) elements. This resulted in repeatedly non-empty plans after apply.

This release imposes an ordering on both attribute lists. As a consequence, there will be a one-time re-ordering of the elements in the two attribute lists. If you apply this change in a region not affected before, this is the first and only time you will notice this change. As a consequence, you may be asked to recreate resources depending on SANs or DVOs. In regions affected before already, the problem should go away after applying one more time.
```

I did not run the acc test on this because I did not want to setup the email validation. However, I tested the implementation by compiling and using it.

### DIFF
--- a/aws/resource_aws_acm_certificate.go
+++ b/aws/resource_aws_acm_certificate.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"sort"
 	"strings"
 	"time"
 
@@ -210,15 +211,16 @@ func resourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("domain_name", resp.Certificate.DomainName)
 		d.Set("arn", resp.Certificate.CertificateArn)
 
-		if err := d.Set("subject_alternative_names", cleanUpSubjectAlternativeNames(resp.Certificate)); err != nil {
+		sans := cleanUpSubjectAlternativeNames(resp.Certificate)
+		if err := d.Set("subject_alternative_names", sans); err != nil {
 			return resource.NonRetryableError(err)
 		}
 
 		domainValidationOptions, emailValidationOptions, err := convertValidationOptions(resp.Certificate)
-
 		if err != nil {
 			return resource.RetryableError(err)
 		}
+		sortDomainValidationOptions(d.Get("domain_name").(*string), sans, domainValidationOptions)
 
 		if err := d.Set("domain_validation_options", domainValidationOptions); err != nil {
 			return resource.NonRetryableError(err)
@@ -244,6 +246,7 @@ func resourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) err
 		return nil
 	})
 }
+
 func resourceAwsAcmCertificateGuessValidationMethod(domainValidationOptions []map[string]interface{}, emailValidationOptions []string) string {
 	// The DescribeCertificate Response doesn't have information on what validation method was used
 	// so we need to guess from the validation options we see...
@@ -283,8 +286,8 @@ func cleanUpSubjectAlternativeNames(cert *acm.CertificateDetail) []string {
 			vs = append(vs, aws.StringValue(v))
 		}
 	}
+	sort.Strings(vs)
 	return vs
-
 }
 
 func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]interface{}, []string, error) {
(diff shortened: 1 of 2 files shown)


jtsaito:
The SANs are no longer sorted alphabetically. Instead the order given by the user is preserved if possible. I apologies for the inconvenience caused by the earlier sorting.

Anyway, it seems AWS is going to address the issue! (https://github.com/terraform-providers/terraform-provider-aws/issues/8531#issuecomment-495719079)

tdmalone:
It looks like the SANs are now possibly randomly sorted when the certificate is first created, but that order is then preserved afterwards. ref https://github.com/terraform-providers/terraform-provider-aws/issues/8531#issuecomment-498000860

jtsaito:
@tdmalone Thanks for investigating this matter in such detail. Based on your observation, I think this fix is still helpful because it would make the sorting unnecessary. 

However, I would suggest to wait for one or two weeks more to see if the problem still persists before asking for this PR to be merged.

morancj:
@jtsaito Great work, thanks! Is this expected to handle the case where SAN entries are in different zones?

tdmalone:
Update: in `ap-southeast-2` at least, it appears that the initial addressing of the issue for `domain_validation_options` has regressed again :( It'd be really great to get this fix merged :)

wvidana:
Anything else needed to see this PR merged? People are still facing the issue with `subject_alternative_names`

zacblazic:
@bflad When can we expect this to be merged? 

Even though this is a breaking change, the implicit  changes made by AWS to cause this issue have already broken many users' terraform plans. We've been patiently waiting for a fix for a long while now.

kraduk:
this is still a major headache, when is it getting merged?

aeschright:
Hi everyone, there's some internal discussion we need to have about the underlying issue and preferred fixes. In general, anything labeled `breaking change` would not be merged until the next major release. This is on our radar, though, and we'll post an update as soon as we can. Thanks for digging into the issue! If you want to discuss workarounds, the bug report is #8531

aws/resource_aws_acm_certificate.go
```diff
@@ -210,15 +211,16 @@ func resourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) err
...
+		if err := d.Set("subject_alternative_names", sans); err != nil {
 			return resource.NonRetryableError(err)
 		}
 
 		domainValidationOptions, emailValidationOptions, err := convertValidationOptions(resp.Certificate)
-
 		if err != nil {
 			return resource.RetryableError(err)
 		}
+		sortDomainValidationOptions(d.Get("domain_name").(*string), sans, domainValidationOptions)
```
dshkyra: I tried to use this patch with [terraform-provider-aws v2.23.0](https://github.com/terraform-providers/terraform-provider-aws/commits/v2.23.0), but unfortunately it failed with an error.
I'm not an expert in golang, however it looks to me that `d.Get("domain_name")` returns just `string`, not `*string`.
So changing the type of `domainName` from `*string` to `string` worked for me.

aws/resource_aws_acm_certificate.go
```diff
@@ -275,16 +277,43 @@ func resourceAwsAcmCertificateUpdate(d *schema.ResourceData, meta interface{}) e
...
+// Also, the domain_name is excluded from the result.
+//
+// Sorting keeps the state stable and is necessary because the API may return
+// SANS ordred differetnly from request to request.
+func readSubjectAlternativeNames(cert *acm.CertificateDetail, d *schema.ResourceData) []string {
+	remotes := cert.SubjectAlternativeNames
+	sans := make([]string, 0)
+
+	// intersection of all SANs, user defined (ud) and remote. ordered as user defined (ud) in state.
+	for _, ud := range d.Get("subject_alternative_names").([]interface{}) {
```
dshkyra: I like this idea to have the certificates ordered as defined by user.
Just want to highlight that when I tried to use this solution (with terraform 0.11.4 and based on terraform aws provider 2.23) SANs were not ordered as defined by user because `d.Get("subject_alternative_names")` returned an empty list during the refresh.

kcburge: I tried to use this PR as well, and had issues with it.  I have created a new pull request that should address the problems: #10791

kcburge:
Just wanted to add to this discussion that I addressed the issues with this pull request, here: #10791

Bharathkumarraju:
is this been released already? I am facing this issue ..i.e. if i add additional SAN names it is trying to recreate ACM validation again and tried to delete the old ACM one which has been already used in ALB https listener?

jtsaito:
This is ancient, so I'm closing this.

PR #8724, opened 2019-05-21
Documentation Update: Shows how one can use normal tags in auto scaling groups

marcosdiez:
### Community Note

This is just a documentation update.

It teaches how one can use a normal map of tags/values in the auto scaling group.
This is probably a common question, hence I believe it's a reasonable thing to add to the documentation.

I changed the variable names so they will be more intuitive for beginners. 
It took me hours googling around for a solution and I unfortunately lost the original author.

I know terraform 2 is on the way, but not everybody will update to it in day zero, so this documentation patch should be welcome.

```release-note
NONE
```

```

### DIFF
--- a/website/docs/r/autoscaling_group.html.markdown
+++ b/website/docs/r/autoscaling_group.html.markdown
@@ -124,16 +124,36 @@ resource "aws_autoscaling_group" "example" {
 ## Interpolated tags
 
 ```hcl
-variable "extra_tags" {
+
+variable "extra_tags_in_normal_map_format" {
+  default = {
+    "maptag_key_1" = "maptag_value_1"
+    "maptag_key_2" = "maptag_value_2"
+    "maptag_key_3" = "maptag_value_3"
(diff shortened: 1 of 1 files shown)


bflad:
Hi @marcosdiez 👋 Thank you for submitting this and apologies for the delayed review.

Terraform 0.12 introduces many enhancements to the configuration language which likely change the optimal configuration for the example here. We will be updating all configuration examples shortly to the updated syntax as part of #8950. We would also like to prevent showing examples dependent on other Terraform Providers, such as the `null` provider, so we do not need to keep up with the maintenance of those providers and their functionality. In practice, we would prefer to enhance the configuration language usability instead of recommending potential workarounds in our examples. For these reasons, we are opting to close this documentation update.

If you are interested in updating this for the new features of Terraform 0.12 without the introduction of the separate provider, which should simplify this example, please feel free to submit a new pull request and we will take a fresh look. Thank you.

PR #8771, opened 2019-05-24
EBS default encryption

ewbankkit:
### Community Note

Fixes https://github.com/terraform-providers/terraform-provider-aws/issues/8760.

```release-note
FEATURES:

* New Resource: `aws_ebs_encryption_by_default`
* New Resource: `aws_ebs_default_kms_key`
```

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSEBSEncryptionByDefault_'==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws/ -v -parallel 20 -run=TestAccAWSEBSEncryptionByDefault_ -timeout 120m
=== RUN   TestAccAWSEBSEncryptionByDefault_basic
=== PAUSE TestAccAWSEBSEncryptionByDefault_basic
=== CONT  TestAccAWSEBSEncryptionByDefault_basic
--- PASS: TestAccAWSEBSEncryptionByDefault_basic (24.04s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	24.079s
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSEBSDefaultKmsKey_'
TF_ACC=1 go test ./aws/ -v -parallel 20 -run=TestAccAWSEBSDefaultKmsKey_ -timeout 120m
=== RUN   TestAccAWSEBSDefaultKmsKey_basic
=== PAUSE TestAccAWSEBSDefaultKmsKey_basic
=== CONT  TestAccAWSEBSDefaultKmsKey_basic
--- PASS: TestAccAWSEBSDefaultKmsKey_basic (56.43s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	56.464s
```


aws/resource_aws_ebs_default_kms_key.go
```diff
@@ -0,0 +1,80 @@
...
+
+func resourceAwsEbsDefaultKmsKey() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsEbsDefaultKmsKeyCreate,
+		Read:   resourceAwsEbsDefaultKmsKeyRead,
+		Update: resourceAwsEbsDefaultKmsKeyUpdate,
+		Delete: resourceAwsEbsDefaultKmsKeyDelete,
+
+		Schema: map[string]*schema.Schema{
+			"key_id": {
```
bflad: Since the API requires the ARN instead of just the ID, it seems like we should prefer to naming this more clearly (even though its slightly off from the API itself 👍 )

```suggestion
			"key_arn": {
```

aws/resource_aws_ebs_default_kms_key.go
```diff
@@ -0,0 +1,80 @@
...
+	conn := meta.(*AWSClient).ec2conn
+
+	_, err := conn.ModifyEbsDefaultKmsKeyId(&ec2.ModifyEbsDefaultKmsKeyIdInput{
+		KmsKeyId: aws.String(d.Get("key_id").(string)),
+	})
+	if err != nil {
+		return fmt.Errorf("error creating EBS default KMS key: %s", err)
+	}
+
+	d.SetId(resource.UniqueId())
```
bflad: Should we set this to the key ARN/ID instead of a random identifier so its usable in downstream resources? 😄

aws/resource_aws_ebs_default_kms_key.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsEbsDefaultKmsKey() *schema.Resource {
+	return &schema.Resource{
```
bflad: We should add import support from the get-go, e.g. via the key ARN as the ID 🚀 

```go
},
```

In the testing:

```go
{
	ImportState:       true,
	ImportStateVerify: true,
},
```

And the documentation:

``````md
## Import

EBS Default KMS Key can be imported with the KMS Key ARN, e.g.

```console
$ terraform import aws_ebs_default_kms_key.example arn:aws:kms:us-east-1:123456789012:key/abcd-1234
```
``````

File: aws/resource_aws_ebs_default_kms_key_test.go [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+			},
+		},
+	})
+}
+
+func testAccCheckAwsEbsDefaultKmsKeyDestroy(s *terraform.State) error {
+	ec2conn := testAccProvider.Meta().(*AWSClient).ec2conn
+	kmsconn := testAccProvider.Meta().(*AWSClient).kmsconn
+
+	alias, err := findKmsAliasByName(kmsconn, "alias/aws/ebs", nil)
```
bflad: Nice!

aws/resource_aws_ebs_default_kms_key_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if aws.StringValue(resp.KmsKeyId) != arn.String() {
+		return fmt.Errorf("Default CMK (%s) is not the account's AWS-managed default CMK (%s)", aws.StringValue(resp.KmsKeyId), arn.String())
+	}
+
+	return nil
+}
+
+func testAccCheckEbsDefaultKmsKey(name string) resource.TestCheckFunc {
```
bflad: Is it worth having this function do the inverse of the destroy check?

aws/resource_aws_ebs_encryption_by_default.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceAwsEbsEncryptionByDefault() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceAwsEbsEncryptionByDefaultCreate,
+		Read:   resourceAwsEbsEncryptionByDefaultRead,
+		Update: resourceAwsEbsEncryptionByDefaultUpdate,
+		Delete: resourceAwsEbsEncryptionByDefaultDelete,
```
bflad: If we are going to do nothing in the `Delete` function, this can use `schema.Noop` instead of a custom function. 

```suggestion
		Delete: schema.Noop,
```

The documentation in that case should also explicitly state that it is only removing Terraform's management of the setting.

Instead of an empty `Delete` function though, I think a better user experience would be do disable encryption when this resource is removed. 👍

aws/resource_aws_ebs_encryption_by_default_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccAWSEBSEncryptionByDefault_basic(t *testing.T) {
+	resourceName := "aws_ebs_encryption_by_default.test"
+
+	resource.ParallelTest(t, resource.TestCase{
```
bflad: If the `Delete` function is to remain empty, can you please add an explicit `CheckDestroy: nil,` to this `TestCase`?

Ideally though, this resource seems like it should enable encryption by default when its added and disable encryption when its removed. The `CheckDestroy` in that case would verify that encryption is disabled. 😄

website/docs/r/ebs_default_kms_key.html.markdown
```diff
@@ -0,0 +1,32 @@
...
+# Resource: aws_ebs_default_kms_key
+
+Provides a resource to manage the default customer master key (CMK) that your AWS account uses to encrypt EBS volumes.
+
+Your AWS account has an AWS-managed default CMK that is used for encrypting an EBS volume when no CMK is specified in the API call that creates the volume.
+By using the `aws_ebs_default_kms_key` resource, you can specify a customer-managed CMK to use in place of the AWS-managed default CMK.
+
+~> **NOTE:** Creating an `aws_ebs_default_kms_key` resource does not enable default EBS encryption. Use the [`aws_ebs_encryption_by_default`](ebs_encryption_by_default.html) to enable default EBS encryption.
+
+~> **NOTE:** Destroying this resource will reset the default CMK to the account's AWS-managed default CMK for EBS.
```
bflad: 💯

aws/resource_aws_ebs_encryption_by_default.go
```diff
@@ -0,0 +1,80 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"enabled": {
+				Type:     schema.TypeBool,
+				Required: true,
```
bflad: This feels like it should be `true` by default:

```suggestion
				Optional: true,
				Default:  true,
```

website/docs/r/ebs_encryption_by_default.html.markdown
```diff
@@ -0,0 +1,25 @@
...
+layout: "aws"
+page_title: "AWS: aws_ebs_encryption_by_default"
+sidebar_current: "docs-aws-ebs-encryption-by-default"
+description: |-
+  Manages whether default EBS encryption is enabled for your AWS account in the current AWS region.
+---
+
+# Resource: aws_ebs_encryption_by_default
+
+Provides a resource to manage whether default EBS encryption is enabled for your AWS account in the current AWS region.
```
bflad: Might be worth a cross link to the `aws_ebs_default_kms_key` resource here as well, e.g.

```suggestion
Provides a resource to manage whether default EBS encryption is enabled for your AWS account in the current AWS region. To manage the default KMS key for the region, see the [`aws_ebs_default_kms_key` resource](/docs/providers/aws/r/ebs_default_kms_key.html).
```

aws/resource_aws_ebs_default_kms_key.go
ewbankkit: Currently we are allowing the ARN to change and doing a resource update - If we go with using the ARN as the ID then we need to add `ForceNew: true` to the `key_arn` attribute and remove `resourceAwsEbsDefaultKmsKeyUpdate()`. IMHO this is actually a better way of modeling the resource.

aws/resource_aws_ebs_default_kms_key.go
ewbankkit: Since `resourceAwsEbsDefaultKmsKeyRead()` doesn't actually use the resource's ID for the AWS API call (there's a single default EBS CMK per region per account) do we want to check that the KMS key ARN passed to `terraform import` is really the current default EBS CMK?

ewbankkit:
Review comments addressed.
Re-ran acceptance tests:

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSEBSEncryptionByDefault_'
TF_ACC=1 go test ./aws/ -v -parallel 20 -run=TestAccAWSEBSEncryptionByDefault_ -timeout 120m
=== RUN   TestAccAWSEBSEncryptionByDefault_basic
=== PAUSE TestAccAWSEBSEncryptionByDefault_basic
=== CONT  TestAccAWSEBSEncryptionByDefault_basic
--- PASS: TestAccAWSEBSEncryptionByDefault_basic (25.52s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	25.542s
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSEBSDefaultKmsKey_'
TF_ACC=1 go test ./aws/ -v -parallel 20 -run=TestAccAWSEBSDefaultKmsKey_ -timeout 120m
=== RUN   TestAccAWSEBSDefaultKmsKey_basic
=== PAUSE TestAccAWSEBSDefaultKmsKey_basic
=== CONT  TestAccAWSEBSDefaultKmsKey_basic
--- PASS: TestAccAWSEBSDefaultKmsKey_basic (46.26s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	46.284s
```

aws/resource_aws_ebs_default_kms_key.go
ewbankkit: If the Terraform code corresponding to the imported resource has the same `key_arn` value as the ARN passed to `terraform import` then I think that having `ForceNew` on the `key_arn` attribute will cause the resource to be recreated.

bflad:

PR #8806, opened 2019-05-29
Upgrade lambda example to 0.12

oscr:
### Community Note

Upgrade lambda example to 0.12

```release-note
NONE
```

```

...
```

### DIFF
--- a/examples/lambda/main.tf
+++ b/examples/lambda/main.tf
@@ -1,9 +1,10 @@
 # Specify the provider and access details
 provider "aws" {
-  region = "${var.aws_region}"
+  region = var.aws_region
 }
 
-provider "archive" {}
+provider "archive" {
+}
 
 data "archive_file" "zip" {
   type        = "zip"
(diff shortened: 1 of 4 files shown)


bflad:
Hi @oscr 👋 Thanks for upgrading this configuration!

I'm not sure if we are quite ready to upgrade the examples into Terraform 0.12-only syntax since the syntax is not backwards compatible with Terraform 0.11, which is very prevalent in practice currently.  I also believe we will be looking to do all the examples in the repository at the same time for consistency. We do not currently have a timeline for removing Terraform 0.11 support from the provider (likely also when we would update all the examples syntax), but for now it is a few months out.

I think our safest course of action for now since the changes at first glance do not look _required_ for Terraform 0.12 support is to leave them in their current state. While they are more verbose with the string interpolation, they should continue to work in both versions. 👍  Let me ask around and get gather additional opinions though.

bflad:
Relating to the above comment, please see #8950, which is why this pull request is going to be closed for now. Thanks again for submitting this, @oscr.

PR #8845, opened 2019-06-03
resource/aws_cognito_user_pool: Add TemporaryPasswordValidityDays field to PasswordPolicy

### LINKED ISSUE #8827: aws_cognito_user_pool support temporary password validity days in password policy
### Community Note

### Description

Amazon has changed the way temporary password expiry is specified. See references.

Specifically 

1. `UnusedAccountValidityDays` in `AdminCreateUserConfigType`

is being moved to 

2. `TemporaryPasswordValidityDays` in `PasswordPolicyType`

The way it works is that you can continue to use the first option but as soon as you specify the second option for a user pool the first option is deprecated and unusable for that user pool.

I think we can either make a clean break and move to the second option or we have to make both fields optional with no default provided to support both use cases. I definitely prefer the first option but it definitely breaks terraform API BC and could have implications by "implicitly" deprecating the first option in anybody's user pool that runs terraform.

I am going to take a crack at implementing this myself, probably the optional way because that will make the most people happy. And then once the deprecated field is removed we can remove it from here too 🤷‍♂ .

* aws_cognito_user_pool

```hcl
resource "aws_cognito_user_pool" "_" {
  # current
  admin_create_user_config {
    unused_account_validity_days: 7
  }
  # proposed
  password_policy {
    temporary_password_validity_days: 7
  }
}
```

### References
https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminCreateUserConfigType.html
https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_PasswordPolicyType.html

kradical:
### Community Note

Fixes #8827

```release-note
temporary_password_validity_days can now be specified in a password policy in an aws_cognito_user_pool. One gotcha is that to use the new field in combination with admin_create_user_config fields you will need to specify the deprecated admin_create_user_config.unused_account_validity_days to 0. Or else you will get an AWS API error that advises you to use the new password_policy.temporary_password_validity_days field.
```

Running the acceptance tests costs $$ right? Is there any way for somebody to run them, or to have them run through CI somewhere? Let me know if this change requires other additional tests as well. I did a bunch of manual verification on our terraform project, but only updated a couple automated tests.

```

...
```


imedmar:
Thanks a lot for this man. I'm totally blocked on applying updates to my user pool so hoping this gets merged asap.

kradical:
@imedmar we found a workaround that by manually setting some admin_user_create_config fields (with the aws cli) you can "reset" your cognito user pool to use the old field instead of the new field. Super weird behaviour but it seems to work for us and get our builds running again. If you need to use the new field you might be hooped though.

If you think that could help you out I can send more specific details! Because even once it gets merged, not sure when itll get released 🤷‍♂ and it is really hard to use non-released plugins unfortunately.

omarelsaid:
Really hoping this gets merged too asap, I'm blocked too!

SantoDE:
ping @nywilken :) Maybe you can have a helping hand here too!

omarelsaid:
Shouldn't this be labelled a bug? There's no nice way around this bug, if any

SantoDE:
@kradical - Can you share your cli script maybe? I'm having a similar issue, so that would be an awesome starting point to have!

kradical:
How to recreate the error and then make it go away using the AWS console / CLI: this could definitely interfere with terraform things so I would only recommend it as a last resort:
1. click save on password policy page (this should set `PasswordPolicy.TemporaryPasswordValidityDays`)

2. `aws cognito-idp update-user-pool --user-pool-id us-west-2_PaalXRZtg --admin-create-user-config '{"AllowAdminCreateUserOnly": false,"UnusedAccountValidityDays": 7}'`

Output from 2: 
```
An error occurred (InvalidParameterException) when calling the UpdateUserPool operation: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays
```

Great, we have the error reproduced. Now try:

3. `aws cognito-idp update-user-pool --user-pool-id us-west-2_PaalXRZtg --admin-create-user-config '{"AllowAdminCreateUserOnly": false}'`

4. `aws cognito-idp update-user-pool --user-pool-id us-west-2_PaalXRZtg --admin-create-user-config '{"AllowAdminCreateUserOnly": false,"UnusedAccountValidityDays": 6}'`

I believe `AllowAdminCreateUserOnly` was just set to what we had it set to before 🤷‍♂. I think the AWS API also has similar "reset" behaviour if you send 0 in the new field, I could be wrong though. This worked for getting our builds going again.

kradical:
Also @omarelsaid the only reason I wouldn't classify this as a bug is because in order to get into this pickle I believe you have to do something outside of terraform to get your userpool to start using the new field. It sucks that it is basically unrecoverable from that point on.. but if you stick to only using terraform and never touching the console / cli I think this is completely avoidable.

This does have to get addressed eventually because the field is deprecated, so now is a good time :D.

imedmar:
Yea this sounds right.
>  I believe you have to do something outside of terraform to get your userpool to start using the new field.

I was testing custom emails on my pool since now there's a daily limit of 50 for the default email client Cognito uses, so had to switch to SES and had to do so via the console because TF doesn't offer any way of doing this afaik.
Don't know for sure if this is what forced my pool to deprecate `UnusedAccountValidityDays`, but it is totally possible that I twisted more knobs than I should've while trying to get SES emails to work.

@kradical just steps 3 and 4 and I'm up and running again. Thanks a lot 😄

jerryatwise:
Testing out the fix seemed to have removed `MFA required` on our dev pool and I ended up recreating the pool. Posting, in case anyone else runs into that. Could of been another step as well 🥇

derKrischan:
> I was testing custom emails on my pool since now there's a daily limit of 50 for the default email client Cognito uses, so had to switch to SES and had to do so via the console because TF doesn't offer any way of doing this afaik.
> Don't know for sure if this is what forced my pool to deprecate `UnusedAccountValidityDays`, but it is totally possible that I twisted more knobs than I should've while trying to get SES emails to work.

I've got the same problem and think that https://github.com/terraform-providers/terraform-provider-aws/issues/8434 is the reason why I ended up here. Looks like we're catched in a loop now. Terraform apply resets SES sending at Cognito pool. Manually activating SES again via console or CLI forces us to execute the workaround via CLI described here and and than we can use Terraform again :-).

kradical:
Yeah the CLI hack probably has other unintended consequences, sorry about that, it worked for our case 🤷‍♂

jerryatwise:
@kradical yeah... all good :) Hence having dev environments :P Appreciate the work on this either way.

maticko:
Hit the same issue today. Would be great if we could get this fix merged and released soon.

@kradical thx for the fix!

damianesteban:
I second that @maticko. We have hit a bit of a wall and would love to see this merged 👍 

Thanks!

C1S0:
we need this merged.

bflad:
Hi @kradical  👋 Thanks for submitting this update. It is not currently passing all the existing acceptance tests so this change will need to be updated to ensure existing environments are not broken by the introduction of the new attribute within Terraform:

```
--- PASS: TestAccAWSCognitoUserPool_basic (13.83s)
--- PASS: TestAccAWSCognitoUserPool_importBasic (14.71s)
--- FAIL: TestAccAWSCognitoUserPool_withTags (15.59s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withSmsVerificationMessage (15.67s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withPasswordPolicy (15.96s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withSmsConfigurationUpdated (16.08s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withAdvancedSecurityMode (16.12s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withVerificationMessageTemplate (17.46s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withAdminCreateUserConfiguration (17.88s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withDeviceConfiguration (20.96s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- PASS: TestAccAWSCognitoUserPool_withSchemaAttributes (21.02s)
--- PASS: TestAccAWSCognitoUserPool_withAliasAttributes (21.95s)
--- FAIL: TestAccAWSCognitoUserPool_withEmailVerificationMessage (22.02s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- FAIL: TestAccAWSCognitoUserPool_withLambdaConfig (27.50s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays

--- PASS: TestAccAWSCognitoUserPool_withSmsConfiguration (34.33s)
--- FAIL: TestAccAWSCognitoUserPool_update (37.31s)
    testing.go:568: Step 1 error: errors during apply:

        Error: Error updating Cognito User pool: InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays
```

Regarding your original questions:

> Running the acceptance tests costs $$ right? Is there any way for somebody to run them, or to have them run through CI somewhere?

According to the [Cognito Pricing page](https://aws.amazon.com/cognito/pricing/), only active monthly users are billed so this testing should be free as the test configurations are not adding users. It looks like there is one test including a Lambda Function as well. The [Lambda Pricing page](https://aws.amazon.com/lambda/pricing/) notes that they are billed after a threshold in requests and duration. Since this testing would not be invoking the function itself, or at least anywhere near the threshold, it is also essentially free.

> Let me know if this change requires other additional tests as well.

For new attributes, we generally prefer to leave existing tests as-is and add a new acceptance test function and configuration for the new attributes. Additional information about the contribution process can be found in the [Contributing Guide](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#enhancementbugfix-to-a-resource).

---

Please reach out if you have any questions or do not have time to continue this implementation.

kradical:
@bflad Thanks for the feedback.

I will figure out how to get the tests running, ensure the unchanged tests still pass (or update code so that they do) and add tests specifically for the new property!

SantoDE:
Hey @kradical,

You okay if I take a shot as well? :)

kradical:
@SantoDE absolutely, feel free to work off my fork if you want 🤷‍♂ . I will take a go at it on the weekend if I haven't seen anything from you.

tboock:
> Testing out the fix(steps 3 and 4) seemed to have removed `MFA required` on our dev pool and I ended up recreating the pool.

When I was testing out the fix I also noticed that `aws cognito-idp update-user-pool` resets attributes to their default values when optional arguments are not passed. Be very careful when you do this. Check `terraform plan` output after doing changes with aws cli to ensure you haven't done unintended changes. In my case I had to supply the following arguments: admin-create-user-config, auto-verified-attributes, device-configuration, lambda-config.

SantoDE:
Hey @bflad,

I'm trying to get everything stitched together, based on the feedback you provided. Thanks for that!
However, I'm having troubles to deal with this situation. Let me explain what I figured so far:

The Cognito API tries to be smart and does set the new value automatically, when you set the old value on a pool. When terraform performs the read option to synchronize state, it therefore fetches the new value as well and saves it to state, even tough this has never been intended by the user. Once you do the next update operation, it will now try to set both options which results in the api fail. 

I dont really know where to move from here without impacting the user too much. For me, it feels like having two different states (hcl file and actual state) which somehow need to work together but I dont really see how. Insights appreciated :)

tibbing:
Waiting for this fix as well. Any idea when it can be merged? :)

kradical:
@tibbing in its current form it breaks some things. We need to fix some tests and maybe alter the implementation. Waiting to hear back.

bflad:
This is certainly more complicated than it would seem due to the type of API interaction changes that occur. There are various facets that need to be handled outside just the addition of the new attribute.

---

We want the Terraform resource to not automatically apply any value for this new `temporary_password_validity_days` attribute, _unless_ it is explicitly configured in the Terraform. By default, a `Type: schema.TypeInt` attribute in the resource schema will have a value of `0` (matching the zero-value of the underlying `int` Go type).  Since `0` is a valid value for the `TemporaryPasswordValidityDays` API parameter we are trying to configure, we cannot simply just tell the `expandCognitoUserPoolPasswordPolicy` function to simply ignore that value (otherwise folks will report that they cannot explicitly configure it to `0` 😅 ):

```go
	// This is unfortunately _not_ a valid change in this scenario
	// Do not use this please
	if v, ok := config["temporary_password_validity_days"]; ok && v.(int) != 0 {
		configs.TemporaryPasswordValidityDays = aws.Int64(int64(v.(int)))
	}
```

An option we could employ here without too much effort would be to configure the schema attribute with `Default: -1` (or something outside the valid range) and then apply logic like the above example to ignore `v.(int) != -1`. While this would hopefully workaround this situation, differences presented by Terraform in this case could potentially be confusing for operators, e.g. `temporary_password_validity_days: null -> -1` (what does this mean?).

Rather than causing potential confusion with the value of this attribute when its "undefined" as `TypeInt`, we can use a different technique to help us with the "undefined" case. The current Terraform Provider SDK unfortunately doesn't provide an easy way to declare something akin to an underlying Go type of `*int` (an integer pointer) as a native schema type, however we have had some success using the schema type `TypeString` (with its default zero-value of `""`) to handle odd cases like these with resources like `aws_launch_template` that require three boolean states (true, false, and undefined) while making the attribute feel like the native type. Under the hood, the configuration language parsing within Terraform Core will automatically convert configurations containing a raw integer (`123`) to a string form (`"123"`) for `TypeString` attributes. We can use this to allow the follow value assignments: `null` (always available in Terraform 0.12), `""` (Terraform 0.11 "`null`" support), and integers like `123`

In the schema this becomes:

```go
"temporary_password_validity_days": {
	Optional:     true,
	// The validation can be handled later
},
```

And in the resource logic it would need to handle converting between the types and ignore the `""` value:

```go
// expandCognitoUserPoolPasswordPolicy
	if v, ok := config["temporary_password_validity_days"]; ok && v.(string) != "" {
		// Ignore the strconv error because it should be caught in the ValidateFunc
		vInt, _ := strconv.ParseInt(v.(string), 10, 64)
		configs.TemporaryPasswordValidityDays = aws.Int64(int64(vInt))
	}

// flattenCognitoUserPoolPasswordPolicy
	if s.TemporaryPasswordValidityDays != nil {
		m["temporary_password_validity_days"] = fmt.Sprintf("%d", aws.Int64Value(s.TemporaryPasswordValidityDays))
	}
```

An initial implementation of the validation should at least be a function similar to `validateTypeStringNullableFloat` in `aws/validators.go`, so we do not need to worry about error handling in the resource logic. e.g.

```go
// validateTypeStringNullableInt provides custom error messaging for TypeString integers
// Some arguments require an integer value or an unspecified, empty field.
func validateTypeStringNullableInt(v interface{}, k string) (ws []string, es []error) {
	value, ok := v.(string)
	if !ok {
		es = append(es, fmt.Errorf("expected type of %s to be string", k))
		return
	}

	if value == "" {
		return
	}

	if _, err := strconv.ParseInt(value, 10, 64); err != nil {
		es = append(es, fmt.Errorf("%s: cannot parse '%s' as integer: %s", k, value, err))
	}

	return
}
```

The validation handling to check for values between `0` and `365` would only be a bonus in this scenario, but it is likely a function similar to `validation.IntBetween(0, 365)` could be written that instead handles the string values, converts them to integers, and performs the same greater/less than logic.

---

When we see the `UpdateUserPool` API error saying to use the new parameter (`InvalidParameterException: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays`) we will either need to automatically remove the `UnusedAccountValidityDays` parameter from the API call and retry it or try to catch this error and present a more helpful error for how to fix the Terraform configuration. I lean towards the first option of just automatically removing the parameter because it may be awkward for operators that account for switching between the two arguments within Terraform modules.

A quick implementation of this is the following, underneath where `UpdateUserPool` is being called, which allows the API call to be retried without the first parameter:

```go
		if isAWSErr(err, cognitoidentityprovider.ErrCodeInvalidParameterException, "use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays") {
			params.AdminCreateUserConfig.UnusedAccountValidityDays = nil
		}
```

---

I believe the combination of the `TypeString` schema type and update handling above should help in this odd scenario. Please reach out with any questions or let us know if you do not have time to go down this path.

kradical:
That sounds pretty reasonable, allows us to have a "faux" nil value. I will spend some time fixing this and update the PR this week / over the weekend.

Thanks for the detailed feedback!

kradical:
@bflad I made the requested changes, I tried running the 2 acceptance tests that I updated and they passed individually, not sure about the entire suite 🤷‍♂ . I am going to make a PR against the helpers module to add a max/min validation factory thing and then at some point we can use it here!

bflad:
Ah looks like the linters caught something in my example code:

```
aws/structure.go:2709:58: unnecessary conversion (unconvert)
		configs.TemporaryPasswordValidityDays = aws.Int64(int64(vInt))
		                                                       ^
GNUmakefile:34: recipe for target 'lint' failed
make: *** [lint] Error 1
The command "make lint" exited with 2.
```

And instead can be just:

```go
		configs.TemporaryPasswordValidityDays = aws.Int64(vInt)
```

kradical:
Whoops, if I knew there was a linter I would have ran it 🤦‍♂ . Updated!

Also see this PR: https://github.com/hashicorp/terraform/pull/21826, I am not sure if you manage that repo to? or have some responsibility over it? Either way once that helper gets merged we can use it here 🎉

SantoDE:
Awesome @kradical and @bflad. Thanks a lot for your very detailed Feedback!

@kradical : I just pulled your changes and let the suite run. Here's the result:

`=== RUN   TestAccAWSCognitoUserPool_importBasic
=== PAUSE TestAccAWSCognitoUserPool_importBasic
=== RUN   TestAccAWSCognitoUserPool_basic
=== PAUSE TestAccAWSCognitoUserPool_basic
=== RUN   TestAccAWSCognitoUserPool_withAdminCreateUserConfiguration
=== PAUSE TestAccAWSCognitoUserPool_withAdminCreateUserConfiguration
=== RUN   TestAccAWSCognitoUserPool_withAdvancedSecurityMode
=== PAUSE TestAccAWSCognitoUserPool_withAdvancedSecurityMode
=== RUN   TestAccAWSCognitoUserPool_withDeviceConfiguration
=== PAUSE TestAccAWSCognitoUserPool_withDeviceConfiguration
=== RUN   TestAccAWSCognitoUserPool_withEmailVerificationMessage
=== PAUSE TestAccAWSCognitoUserPool_withEmailVerificationMessage
=== RUN   TestAccAWSCognitoUserPool_withSmsVerificationMessage
=== PAUSE TestAccAWSCognitoUserPool_withSmsVerificationMessage
=== RUN   TestAccAWSCognitoUserPool_withEmailConfiguration
--- SKIP: TestAccAWSCognitoUserPool_withEmailConfiguration (0.00s)
    resource_aws_cognito_user_pool_test.go:281: 'TEST_AWS_SES_VERIFIED_EMAIL_ARN' not set, skipping test.
=== RUN   TestAccAWSCognitoUserPool_withSmsConfiguration
=== PAUSE TestAccAWSCognitoUserPool_withSmsConfiguration
=== RUN   TestAccAWSCognitoUserPool_withSmsConfigurationUpdated
=== PAUSE TestAccAWSCognitoUserPool_withSmsConfigurationUpdated
=== RUN   TestAccAWSCognitoUserPool_withTags
=== PAUSE TestAccAWSCognitoUserPool_withTags
=== RUN   TestAccAWSCognitoUserPool_withAliasAttributes
=== PAUSE TestAccAWSCognitoUserPool_withAliasAttributes
=== RUN   TestAccAWSCognitoUserPool_withPasswordPolicy
=== PAUSE TestAccAWSCognitoUserPool_withPasswordPolicy
=== RUN   TestAccAWSCognitoUserPool_withLambdaConfig
=== PAUSE TestAccAWSCognitoUserPool_withLambdaConfig
=== RUN   TestAccAWSCognitoUserPool_withSchemaAttributes
=== PAUSE TestAccAWSCognitoUserPool_withSchemaAttributes
=== RUN   TestAccAWSCognitoUserPool_withVerificationMessageTemplate
=== PAUSE TestAccAWSCognitoUserPool_withVerificationMessageTemplate
=== RUN   TestAccAWSCognitoUserPool_update
=== PAUSE TestAccAWSCognitoUserPool_update
=== CONT  TestAccAWSCognitoUserPool_importBasic
=== CONT  TestAccAWSCognitoUserPool_withSmsConfigurationUpdated
=== CONT  TestAccAWSCognitoUserPool_withAliasAttributes
=== CONT  TestAccAWSCognitoUserPool_withEmailVerificationMessage
=== CONT  TestAccAWSCognitoUserPool_withTags
=== CONT  TestAccAWSCognitoUserPool_update
=== CONT  TestAccAWSCognitoUserPool_withVerificationMessageTemplate
=== CONT  TestAccAWSCognitoUserPool_withSchemaAttributes
=== CONT  TestAccAWSCognitoUserPool_withLambdaConfig
=== CONT  TestAccAWSCognitoUserPool_withPasswordPolicy
=== CONT  TestAccAWSCognitoUserPool_withDeviceConfiguration
=== CONT  TestAccAWSCognitoUserPool_withSmsConfiguration
=== CONT  TestAccAWSCognitoUserPool_withSmsVerificationMessage
=== CONT  TestAccAWSCognitoUserPool_withAdvancedSecurityMode
=== CONT  TestAccAWSCognitoUserPool_basic
=== CONT  TestAccAWSCognitoUserPool_withAdminCreateUserConfiguration
panic: interface conversion: interface {} is int, not string

goroutine 3213 [running]:
github.com/terraform-providers/terraform-provider-aws/aws.expandCognitoUserPoolPasswordPolicy(0xc00e531b00, 0xc0000ce7c8)
	/Users/manuelzapf/Projects/terraform/terraform-provider-aws/aws/structure.go:2707 +0x4ff
github.com/terraform-providers/terraform-provider-aws/aws.resourceAwsCognitoUserPoolCreate(0xc00f08b5e0, 0x524eea0, 0xc00059b000, 0x2, 0xaa3e9e0)
	/Users/manuelzapf/Projects/terraform/terraform-provider-aws/aws/resource_aws_cognito_user_pool.go:598 +0x1103
github.com/hashicorp/terraform/helper/schema.(*Resource).Apply(0xc000289500, 0xc00cf520f0, 0xc007f83880, 0x524eea0, 0xc00059b000, 0xc00e530601, 0xc00714ebe0, 0x574a700)
	/Users/manuelzapf/go/pkg/mod/github.com/hashicorp/terraform@v0.12.2/helper/schema/resource.go:286 +0x3b4
github.com/hashicorp/terraform/helper/schema.(*Provider).Apply(0xc00071cf80, 0xc008f77a58, 0xc00cf520f0, 0xc007f83880, 0xc0064f2088, 0xc008fcbf48, 0x574dd00)
	/Users/manuelzapf/go/pkg/mod/github.com/hashicorp/terraform@v0.12.2/helper/schema/provider.go:285 +0x18f
github.com/hashicorp/terraform/helper/plugin.(*GRPCProviderServer).ApplyResourceChange(0xc0119bd4c0, 0x6e73a20, 0xc0088315c0, 0xc007c11620, 0xc0119bd4c0, 0xc0088315c0, 0xc002b3abd0)
	/Users/manuelzapf/go/pkg/mod/github.com/hashicorp/terraform@v0.12.2/helper/plugin/grpc_provider.go:851 +0x894
github.com/hashicorp/terraform/internal/tfplugin5._Provider_ApplyResourceChange_Handler(0x6044e00, 0xc0119bd4c0, 0x6e73a20, 0xc0088315c0, 0xc0109c1ef0, 0x0, 0x6e73a20, 0xc0088315c0, 0xc000830000, 0x4f3)
	/Users/manuelzapf/go/pkg/mod/github.com/hashicorp/terraform@v0.12.2/internal/tfplugin5/tfplugin5.pb.go:3217 +0x23e
google.golang.org/grpc.(*Server).processUnaryRPC(0xc004a7b680, 0x6ea4160, 0xc00c0ec000, 0xc0051e9400, 0xc00c5b1f50, 0xa9fd320, 0x0, 0x0, 0x0)
	/Users/manuelzapf/go/pkg/mod/google.golang.org/grpc@v1.18.0/server.go:966 +0x470
google.golang.org/grpc.(*Server).handleStream(0xc004a7b680, 0x6ea4160, 0xc00c0ec000, 0xc0051e9400, 0x0)
	/Users/manuelzapf/go/pkg/mod/google.golang.org/grpc@v1.18.0/server.go:1245 +0xd25
google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc0028ec1e0, 0xc004a7b680, 0x6ea4160, 0xc00c0ec000, 0xc0051e9400)
	/Users/manuelzapf/go/pkg/mod/google.golang.org/grpc@v1.18.0/server.go:685 +0x9f
created by google.golang.org/grpc.(*Server).serveStreams.func1
	/Users/manuelzapf/go/pkg/mod/google.golang.org/grpc@v1.18.0/server.go:683 +0xa1
FAIL	github.com/terraform-providers/terraform-provider-aws/aws	8.026s`

kradical:
I'll make sure those tests pass too! Is there any reuse of use pools between tests? that could lead to some undesirable behaviour. I'll look into it though

kradical:
`TPVD` = `TemporaryPasswordValidityDays`
`UAVD` = `UnusedAccountValidityDays`

Hey @SantoDE it seems like we are in a bit of a pickle, the `TestAccAWSCognitoUserPool_withAdminCreateUserConfiguration` doesn't pass anymore. I think there is some weird behaviour going on with automatically setting the password_policy even when it is not specified.

This is the response from creating a pool:

```
{
  UserPool: {
    AdminCreateUserConfig: {
      AllowAdminCreateUserOnly: true,
      InviteMessageTemplate: {
        EmailMessage: "Your username is {username} and temporary password is {####}. ",
        EmailSubject: "FooBar {####}",
        SMSMessage: "Your username is {username} and temporary password is {####}."
      },
      UnusedAccountValidityDays: 6
    },
    Arn: "somearn",
    CreationDate: 2019-06-21 17:04:16 +0000 UTC,
    EmailConfiguration: {
      EmailSendingAccount: "COGNITO_DEFAULT"
    },
    EstimatedNumberOfUsers: 0,
    Id: "us-west-2_5cMfXmGqF",
    LambdaConfig: {

    },
    LastModifiedDate: 2019-06-21 17:04:16 +0000 UTC,
    MfaConfiguration: "OFF",
    Name: "terraform-test-pool-aiqhb",
    Policies: {
      PasswordPolicy: {
        MinimumLength: 8,
        RequireLowercase: true,
        RequireNumbers: true,
        RequireSymbols: true,
        RequireUppercase: true,
        TemporaryPasswordValidityDays: 6
      }
    },
    SchemaAttributes: [
    ...
}
```

It is the same response on my branch and master upstream. So this `TPVD` is getting created by AWS behind the scenes regardless of if it is specified or not. The problem occurs now when we "read" these values. Now that we have a slot for `TPVD` to be read into it is ignoring the update to `UAVD` in the second half of the test.

I think we tried really hard to do this in a backwards compatible way but I don't know that we can do it without specifically setting `TPVD` to `""` whenever `UAVD` is specified and `TPVD` isn't in configuration. Can you see any way to work around this?

SantoDE:
Phew, I stumpled accross that already as well. I'm also not really sure if there is a work arround.

@bflad - what do you think?

SantoDE:
Hey @kradical,

can you push your current state please? I'd like to give it another shot :)

SantoDE:
hey all,

@kradical: nvm - I see that is your latest version

@bflad I guess I have to agree with kradical on that one. After some more tests, I dont see a way to handle that gracefully. The only option I could think of (and I'm totaly unsure about the validity of that one), is to add another configuration setting like `use_old`, which "forces" a usage either attribute. As you then know which attribute to use (as both are set), you could map the values back and forth. Other than that, I don't really see a way.

That special value should immediately be deprecated tough, as with the next major bump the "use_old" field and the old `UnusedAccountValidityDays` field should be removed.

All imho of course :)

kradical:
@SantoDE just be careful if you have commits, i possibly rewrote git history on my fork at some point, sorry!

johannroux:
Hi guys! Any update on this one? We're eagerly waiting for it to be merged :-)

kradical:
@johannroux it is kind of dead in the water in its current form. Breaks things in non-obvious ways, it breaks the current tests because of the behaviour. I think we may have to be opinionated and use the new field. This would be a "breaking" configuration specification change.. which I was trying to avoid.. but it is a deprecated field 🤷‍♂

fschroder:
Although I'm not sure I fully understand it, I tend to agree with @SantoDE approach of using a "use_old"/selector field. Meaning having a field to explicitly tell TF which field it should set. This provides complete control to the teams in different situations. Once the old field is removed, the selector and the old field can be removed from the interface.
In the meantime the default could be to use the new field when needed.

jorenvh1:
When will this be merged?

kradical:
@jorenvh1 never in its current form. I think we want to add an extra field to configure which field is used so that the end user has control over which field gets sent off with the request.

This implementation is buggy (over the course of multiple runs and configuration changes) and breaks the tests. I might have another go at it soon.. just been busy with work and life.

jorenvh1:
Any updates? We are pretty blocked by this issue, would be awesome if this could be fixed 👍

jorenvh1:
If this won't be merged, can anyone tell us if there's any workaround?

arne21a:
We built a preview of this fix, it worked for us.
If someone wants to try it out, feel free to download our binaries.
https://github.com/adesso-as-a-service/terraform-provider-aws/releases/tag/v2.27.1
Use it at your own risk ;)

kradical:
@jorenvh1 read through the whole thread, a number of people (myself included) have suggested manual workarounds to get terraform builds running again!

DawidMazurek:
Manual workarounds no longer work in eu-central-1 as AWS already dropped support for old field
They still work in eu-west-1 but will probably also stop working

Following cli command fixes for manual action:
```
aws cognito-idp update-user-pool --user-pool-id eu-central-1_2wajDkt7Y --admin-create-user-config 'AllowAdminCreateUserOnly=false' --policies='PasswordPolicy={MinimumLength=6,TemporaryPasswordValidityDays=7}' --region eu-central-1
```

But when running a terraform it still fails.
We need to have new terraform version with this fix bundled to progress

kradical:
Thanks for the update @DawidMazurek . I am going to close this PR because it attempts to do something other than actually fix the underlying problem. I also just ran out of time to work on this. Hopefully somebody can fix the provider to use the new field sooner rather than later!

tlhampton13:
@kradical is there a corresponding issue on the the provider project ?

kradical:
@tlhampton13 original issue: https://github.com/terraform-providers/terraform-provider-aws/issues/8827

shadowfax-n:
tested today... and the manual fix no longer works in eu-west-1 :-(.....

cwcoffman:
+1 on @shadowfax-n - manual fix no longer works and am looking for options besides killing the User Pool.

PR #8899, opened 2019-06-07
Remove flatmap dependency

radeksimko:
As part of some prep work for SDK decoupling I found this provider imports a legacy `flatmap` package - this isn't the way we plan to represent data in the SDK in the future anymore and regardless of that it seems unnecessary to import the whole package for what can practically be done in 1 line of code.

```release-note
NONE
```

```
$ go test ./aws -run=TestExpandStringList -v
=== RUN   TestExpandStringList
--- PASS: TestExpandStringList (0.00s)
=== RUN   TestExpandStringListEmptyItems
--- PASS: TestExpandStringListEmptyItems (0.00s)
```

There's one more place where this provider imports this package, but changing that may require a bit more effort, so I filed a separate issue for that: https://github.com/terraform-providers/terraform-provider-aws/issues/8902

### DIFF
--- a/aws/structure_test.go
+++ b/aws/structure_test.go
@@ -16,32 +16,8 @@ import (
 	"github.com/aws/aws-sdk-go/service/rds"
 	"github.com/aws/aws-sdk-go/service/redshift"
 	"github.com/aws/aws-sdk-go/service/route53"
-	"github.com/hashicorp/terraform/flatmap"
 	"github.com/hashicorp/terraform/helper/schema"
 )
-
-// Returns test configuration
-func testConf() map[string]string {
-	return map[string]string{
-		"listener.#":                   "1",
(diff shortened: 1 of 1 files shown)


nywilken:
This change looks good to me for this particular file. Since `github.com/hashicorp/terraform/flatmap` is a legacy pkg should we also look to remove it as a dependency in any of the resources? Possibly, as a separate PR per resource? 
https://github.com/terraform-providers/terraform-provider-aws/blob/32cf24db2f7ccfa1fe1fab3c22a8a13c43b058bd/aws/cloudfront_distribution_configuration_structure.go#L18

radeksimko:
@nywilken yes, as I mentioned in my comment above 😉 

> There's one more place where this provider imports this package, but changing that may require a bit more effort, so I filed a separate issue for that: #8902

PR #8906, opened 2019-06-07
Update ram_resource_share.html.markdown

MMarulla:
resource_owner is a required field, but it is not included in the example.

### Community Note

Fixes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/d/ram_resource_share.html.markdown
+++ b/website/docs/d/ram_resource_share.html.markdown
@@ -13,6 +13,7 @@ description: |-
 ## Example Usage
 ```hcl
 data "aws_ram_resource_share" "example" {
+  resource_owner = "SELF"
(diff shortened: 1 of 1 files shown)


website/docs/d/ram_resource_share.html.markdown
```diff
@@ -13,6 +13,7 @@ description: |-
 ## Example Usage
 ```hcl
 data "aws_ram_resource_share" "example" {
+  resource_owner = "SELF"
```
nywilken: Good catch! Could you move `resource_owner = "SELF"` down a line so that it comes after the `name` attribute to keep it consistent with the filters example.

MMarulla: Done.

PR #8909, opened 2019-06-07
aws_eks_cluster: Increase create timeout to 30 mins

### LINKED ISSUE #5165: aws_eks_cluster tales longer to create than the timeout period
_This issue was originally opened by @jtravell as hashicorp/terraform#18438. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

The timeout period for `terraform apply` seems to be 15 minutes.  It takes longer than that to create an EKS cluster on AWS using `terraform apply`.  This is causing a terraform program to break because the `aws_eks_cluster` has progressed to RUNNING status before the `terraform apply` throws an error due to the timeout.  

Specifically, the USERDATA that gets imported into the launch configuration of an auto scaling group to be built in the cluster is receiving null values for `${aws_eks_cluster.demo.endpoint}` and the resulting Kubernetes cluster thus does not allow worker nodes to join.  
  
Simply re-running `terraform apply` does not solve the problem, as a new launch configuration is created again before the `aws_eks_cluster` has been created.  I tried adding `depends_on` to the `locals` object that declares the user data, but this has not solved the problem.  

I have NOT tried waiting half an hour after the first `terraform apply` times out to re-run `terraform apply`.  But that seems like an extreme thing to try.  

Is there a better way of dealing with this timeout issue?

jen20:
It seems that as reported in #5165, an EKS cluster frequently takes more to actually bring up than the default create timeout allows for. While the timeout is configurable, the default should be sufficient to actually provision a cluster in most cases.

This commit increases the default timeout to 30 minutes.

Fixes #5165.

### Community Note

```release-note
Increase default create timeout for `aws_eks_cluster` to 30 minutes.
```

### DIFF
--- a/aws/resource_aws_eks_cluster.go
+++ b/aws/resource_aws_eks_cluster.go
@@ -33,7 +33,7 @@ func resourceAwsEksCluster() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(15 * time.Minute),
+			Create: schema.DefaultTimeout(30 * time.Minute),
 			Update: schema.DefaultTimeout(60 * time.Minute),
 			Delete: schema.DefaultTimeout(15 * time.Minute),
 		},
--- a/website/docs/r/eks_cluster.html.markdown
+++ b/website/docs/r/eks_cluster.html.markdown
@@ -98,7 +98,7 @@ In addition to all arguments above, the following attributes are exported:
 `aws_eks_cluster` provides the following
 [Timeouts](/docs/configuration/resources.html#timeouts) configuration options:
 
-* `create` - (Default `15 minutes`) How long to wait for the EKS Cluster to be created.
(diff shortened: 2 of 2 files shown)


aws/resource_aws_eks_cluster.go
```diff
@@ -33,7 +33,7 @@ func resourceAwsEksCluster() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(15 * time.Minute),
+			Create: schema.DefaultTimeout(30 * time.Minute),
```
bflad: This update should include a documentation update in `website/docs/r/eks_cluster.html.markdown` - thanks

jen20: Done.

PR #8915, opened 2019-06-09
Update main.tf

IPvSean:
### Community Note

### Author Notes
I recieved the error "The argument "host" is required, but no definition was found." until I added this.  This example should be updated.

```
Terraform v0.12.1
+ provider.aws v2.14.0
```

```release-note
NONE
```

```

...
```
N/A

### DIFF
--- a/examples/two-tier/main.tf
+++ b/examples/two-tier/main.tf
@@ -108,7 +108,7 @@ resource "aws_instance" "web" {
   connection {
     # The default username for our AMI
     user = "ubuntu"
-    host = self.public_ip
+    host = "${self.public_ip}"
     # The connection will use the local SSH agent for authentication.
   }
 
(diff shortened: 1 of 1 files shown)


IPvSean:
more info here-> https://www.terraform.io/upgrade-guides/0-12.html#default-settings-in-connection-blocks

examples/two-tier/main.tf
```diff
@@ -108,7 +108,7 @@ resource "aws_instance" "web" {
   connection {
     # The default username for our AMI
     user = "ubuntu"
-
+    host = self.public_ip
```
bflad: Hi @IPvSean 👋 Thanks for submitting this change. We would like to keep the examples both Terraform 0.11 and Terraform 0.12 compatible where possible for the next few months until the community has much broader adoption of Terraform 0.12. To that end, could you please use the string interpolation syntax here? I verified it works on both Terraform 0.11 and 0.12. Thanks again.

```suggestion
    host = "${self.public_ip}"
```

IPvSean:
makes sense to me, thanks @bflad change is made now

PR #8934, opened 2019-06-11
Clearer source_code_hash description

oliversalzburg:
Since 0.12, wrapping in "${}" is no longer required, so it should be omitted here.

Additionally, you're required to use a `.zip` file for an AWS Lambda, so using `file()` here is not ideal, since you're reading a binary file. Thus, `filebase64()` is a better example, as the ZIP file contents are unlikely to be valid UTF-8 content.

### Community Note

Fixes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/r/lambda_function.html.markdown
+++ b/website/docs/r/lambda_function.html.markdown
@@ -153,7 +153,7 @@ large files efficiently.
 * `vpc_config` - (Optional) Provide this to allow your function to access your VPC. Fields documented below. See [Lambda in VPC][7]
 * `environment` - (Optional) The Lambda environment's configuration settings. Fields documented below.
 * `kms_key_arn` - (Optional) The ARN for the KMS encryption key.
-* `source_code_hash` - (Optional) Used to trigger updates. Must be set to a base64-encoded SHA256 hash of the package file specified with either `filename` or `s3_key`. The usual way to set this is `base64sha256(filebase64("file.zip"))`, where "file.zip" is the local filename of the lambda function source archive.
+* `source_code_hash` - (Optional) Used to trigger updates. Must be set to a base64-encoded SHA256 hash of the package file specified with either `filename` or `s3_key`. The usual way to set this is `filebase64sha256("file.zip")` (Terraform 0.11.12 and later) or `base64sha256(file("file.zip"))` (Terraform 0.11.11 and earlier), where "file.zip" is the local filename of the lambda function source archive.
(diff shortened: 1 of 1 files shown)


website/docs/r/lambda_function.html.markdown
```diff
@@ -153,7 +153,7 @@ large files efficiently.
 * `vpc_config` - (Optional) Provide this to allow your function to access your VPC. Fields documented below. See [Lambda in VPC][7]
 * `environment` - (Optional) The Lambda environment's configuration settings. Fields documented below.
 * `kms_key_arn` - (Optional) The ARN for the KMS encryption key.
-* `source_code_hash` - (Optional) Used to trigger updates. Must be set to a base64-encoded SHA256 hash of the package file specified with either `filename` or `s3_key`. The usual way to set this is `${base64sha256(file("file.zip"))}`, where "file.zip" is the local filename of the lambda function source archive.
+* `source_code_hash` - (Optional) Used to trigger updates. Must be set to a base64-encoded SHA256 hash of the package file specified with either `filename` or `s3_key`. The usual way to set this is `base64sha256(filebase64("file.zip"))`, where "file.zip" is the local filename of the lambda function source archive.
```
bflad: Hi @oliversalzburg 👋 Thanks for this update -- the equivalent for the old `base64sha256(file())` usage from Terraform 0.11 is actually the single function `filebase64sha256()` in Terraform 0.11.12 and later as noted above in the example configuration:

```hcl
  # The filebase64sha256() function is available in Terraform 0.11.12 and later
  # For Terraform 0.11.11 and earlier, use the base64sha256() function and the file() function:
  # source_code_hash = "${base64sha256(file("lambda_function_payload.zip"))}"
  source_code_hash = "${filebase64sha256("lambda_function_payload.zip")}"
```

We should likely include information about both of these usages just to remove any confusion between Terraform versions. 👍 

```suggestion
* `source_code_hash` - (Optional) Used to trigger updates. Must be set to a base64-encoded SHA256 hash of the package file specified with either `filename` or `s3_key`. The usual way to set this is `filebase64sha256("file.zip")` (Terraform 0.11.12 and later) or `base64sha256(file("file.zip"))` (Terraform 0.11.11 and earlier), where "file.zip" is the local filename of the lambda function source archive.
```

Thanks for catching this here!

oliversalzburg: Even better!

PR #8948, opened 2019-06-11
Use 0.12 syntax for aws_ses_domain_mail_from

oliversalzburg:
The values are no longer required to be wrapped as string explicitly, so the simpler syntax should be used.

### Community Note

Fixes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/r/ses_domain_mail_from.html.markdown
+++ b/website/docs/r/ses_domain_mail_from.html.markdown
@@ -16,7 +16,7 @@ Provides an SES domain MAIL FROM resource.
 
 ```hcl
 resource "aws_ses_domain_mail_from" "example" {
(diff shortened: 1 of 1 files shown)


bflad:
Hi @oliversalzburg  👋 Thanks for submitting this change.

We would like to keep the examples both Terraform 0.11 and Terraform 0.12 compatible where possible for the next few months until the community has much broader adoption of Terraform 0.12, otherwise those running Terraform 0.11 will not be able to use the available examples without figuring out the syntax errors. When the time seems appropriate, we will convert all the documentation, examples, and test configurations to Terraform 0.12 syntax as part of removing support for Terraform 0.11 and earlier in the provider.

I have created an issue that can be used to track this effort here: #8950

Thanks again for raising this and being ahead of the curve!

Issue #6532, opened 2018-11-20
* aws_instance.instance: unexpected EOF


jbanmene:
### Community Note

Hi there,

when i'm trying to launch a new instance,my terraform failed with error:
* aws_instance.Microsoft_Instance: unexpected EOF
A part of my config related to my instance  below 

### Terraform Version

Terraform v0.11.10
+ provider.aws v1.46.0
+ provider.template v1.0.0

### Affected Resource(s)

* aws_instances

### Terraform Configuration Files
my instance module

```
#Setting Provider
provider "aws" {
  region = "${var.region}"
}

#Setting User data
data "template_file" "user_data" {
  template = "${file("..\\..\\..\\Files\\user_data.txt")}"

  vars {
    envVariable     = "Vendor:${var.Vendor}#Product:${var.Product}#Action:${var.Action}#AvailableName:${var.Available_Name}#TargetOU:${var.Target_OU}"
    BootstrapPrefix = "${var.BootstrapPrefix}"
    Environment     = "${var.Environment}"
    SetupScripts    = "${var.SetupScripts}"
    ArchiveType     = "${var.ArchiveType}"
  }
}

#Setting aws Instance
data "aws_ami" "ami_base" {
filter {
  name = "tag:Name"
  values = ["${var.ami}"]
}
most_recent = true
}

resource "aws_instance" "Microsoft_Instance" {
  key_name             = "${var.key_name}"
  iam_instance_profile = "${var.iam_instance_profile}"
  ami                    = "${data.aws_ami.ami_base.id}"
  instance_type          = "${var.instance_type}"
  vpc_security_group_ids = ["${split(",",var.vpc_security_group_ids)}"]
  subnet_id              = "${var.subnet_id}"
  private_ip             = "${var.private_ip}"

  tags {
    Application = "${var.Tag_Application}"
    dependency  = "${var.Tag_Dependency}"
    depends_id  = "${var.depends_id}"
    Name = "${var.Available_Name}"
  }
  root_block_device {
    volume_type = "${var.volume_type}"
    volume_size = "${var.rootVolume_size}"
  }

  credit_specification {
    cpu_credits = "${var.cpu_credit}"
  }
  user_data = "${data.template_file.user_data.rendered}"
}
output "private_ip" {
  value = "${aws_instance.Microsoft_Instance.private_ip}"
}
```

### Panic Output

https://gist.github.com/jbanmene/642f5d8edecbd2abd6f50c92751aec6f

[...]

Issue #4987, opened 2018-06-26
Setup Region/Partition Agnostic aws_instance Acceptance Testing


bflad:
Its not currently possible to run the `aws_instance` resource and data source acceptance testing outside `us-west-2` and get meaningful results. For example, setting `AWS_DEFAULT_REGION=us-east-1` while running the tests today returns: `Tests failed: 61 (59 new), passed: 6`

Many of the failures are similar to:

```
--- FAIL: TestAccAWSInstance_basic (8.76s)
    testing.go:518: Step 0 error: Check failed: InvalidZone.NotFound: The zone 'us-west-2a' does not exist.
```

And:

```
--- FAIL: TestAccAWSInstance_disableApiTermination (17.22s)
    testing.go:518: Step 0 error: Error applying: 1 error(s) occurred:
        
        * aws_instance.foo: 1 error(s) occurred:
        
        * aws_instance.foo: Error launching source instance: InvalidAMIID.NotFound: The image id '[ami-4fccb37f]' does not exist
```

We can switch to using the `aws_region`, `aws_availability_zones`, and `aws_partition` data sources as appropriate to remove hardcoded `us-west-2` pieces. For acceptance tests that cannot occur outside of `us-west-2` (due to very specific AMIs, etc.), we can set them up with `t.Skip()`.

ewbankkit:
Look also at acceptance tests for:
* `aws_ami_from_instance`
* `aws_autoscaling_attachment`
* `aws_autoscaling_lifecycle_hook`
* `aws_autoscaling_notification`
* `aws_autoscaling_schedule`
* `aws_eip_association`
* `aws_elb_attachment`
* `aws_launch_configuration`
* `aws_launch_template`
* `aws_lb_target_group_attachment`
* `aws_network_interface_attachment`
* `aws_network_interface`
* `aws_route_table`
* `aws_route`
* `aws_spot_fleet_request`
* `aws_spot_instance_request`
* `aws_volume_attachment`

as there are some with hard-coded AMIs (look for `ami` or `image_id` attributes).

ewbankkit:
Not all instance types are available in all regions/partitions so there will need to be cleanup of `instance_type` also.

PatMyron:
[...]

PR #9003, opened 2019-06-14
resource/aws_instance: Add check for empty credit_specification block

nywilken:
Fixes: #6532

When given an empty credit_specification block (i.e cpu_credits is missing) the Terraform provider
will crash upon trying to type assert on the empty (nil) block. This check introduces a guard clause around the type assertion
to protect around the empty (nil) block. A warning is logged to indicated that a default value will be used in place of the missing
value.

### Community Note

```release-note
* resource/aws_instance: Add check for empty credit_specification block
```

```
$ make testacc TESTARGS='-run=TestAccAWSInstance_creditSpecification_unknown'

--- PASS: TestAccAWSInstance_creditSpecification_unknownCpuCredits_t2
--- PASS: TestAccAWSInstance_creditSpecification_unknownCpuCredits_t3
```

### DIFF
--- a/aws/resource_aws_instance_test.go
+++ b/aws/resource_aws_instance_test.go
@@ -1693,6 +1693,50 @@ func TestAccAWSInstance_creditSpecification_unlimitedCpuCredits(t *testing.T) {
 	})
 }
 
+func TestAccAWSInstance_creditSpecification_unknownCpuCredits_t2(t *testing.T) {
+	var instance ec2.Instance
+	rInt := acctest.RandInt()
+	resName := "aws_instance.foo"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckInstanceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccInstanceConfig_creditSpecification_unknownCpuCredits(rInt, "t2.micro"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckInstanceExists(resName, &instance),
+					resource.TestCheckResourceAttr(resName, "credit_specification.#", "1"),
+					resource.TestCheckResourceAttr(resName, "credit_specification.0.cpu_credits", "standard"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccAWSInstance_creditSpecification_unknownCpuCredits_t3(t *testing.T) {
+	var instance ec2.Instance
+	rInt := acctest.RandInt()
+	resName := "aws_instance.foo"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckInstanceDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccInstanceConfig_creditSpecification_unknownCpuCredits(rInt, "t3.micro"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckInstanceExists(resName, &instance),
+					resource.TestCheckResourceAttr(resName, "credit_specification.#", "1"),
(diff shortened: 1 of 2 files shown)


aws/resource_aws_instance_test.go
```diff
@@ -3971,6 +4016,33 @@ resource "aws_instance" "foo" {
...
+
+resource "aws_instance" "foo" {
+  ami           = "ami-51537029"               # us-west-2
+  instance_type = %q
+  subnet_id     = "${aws_subnet.my_subnet.id}"
+
+  credit_specification {
+  }
+}
+`, rInt, instance_type)
```
nywilken:
```suggestion
`, rInt, instance)
```

aws/resource_aws_instance_test.go
```diff
@@ -3971,6 +4016,33 @@ resource "aws_instance" "foo" {
 `, rInt)
 }
 
+func testAccInstanceConfig_creditSpecification_unknownCpuCredits(rInt int, instance_type string) string {
```
nywilken:
```suggestion
func testAccInstanceConfig_creditSpecification_unknownCpuCredits(rInt int, instance string) string {
```

aws/resource_aws_instance_test.go
aeschright: Having it match the attribute name makes the most sense to me.

bflad: Drive by comment: we historically have preferred naming these consistent with Terraform configuration argument and using [mixed caps to match the Effective Go documentation](https://golang.org/doc/effective_go.html#mixed-caps). So in this case: `instanceType` 👍 

[`golint`](https://github.com/golang/lint) would catch underscored variable names like this, e.g. one example on master:

```
aws/config_test.go:43:5: don't use underscores in Go names; var `test_ec2_describeAccountAttributes_response` should be `testEc2DescribeAccountAttributesResponse` (golint)
var test_ec2_describeAccountAttributes_response = `<DescribeAccountAttributesResponse xmlns="http://ec2.amazonaws.com/doc/2016-11-15/">
    ^
```

However there are a _lot_ of linting failures we generally do not want to concern ourselves within `golint`, including its non-configurable initialisms and the underscore check for acceptance test function names (among others), which is why its not enabled by default in our pull request `golangci-lint` testing.

bflad: Submitted a pull request to the contributing guide to ensure it includes a link to Effective Go: #9031

nywilken: Drive by commenting - love it. Thanks for the additional context @bflad and good call out on the Effective Go resource.

aws/resource_aws_instance_test.go
```diff
@@ -3971,6 +4016,33 @@ resource "aws_instance" "foo" {
...
+}
+
+resource "aws_subnet" "my_subnet" {
+  vpc_id            = "${aws_vpc.my_vpc.id}"
+  cidr_block        = "172.16.20.0/24"
+  availability_zone = "us-west-2a"
+}
+
+resource "aws_instance" "foo" {
+  ami           = "ami-51537029"               # us-west-2
```
bflad: Same note and references here about being AWS region agnostic. We can handle AMI lookup via the [`aws_ami` data source](https://www.terraform.io/docs/providers/aws/d/ami.html) against Amazon Linux AMIs that are generally available in all AWS partitions and regions. e.g. from the contributing guide

```hcl
data "aws_ami" "amzn-ami-minimal-hvm-ebs" {
  most_recent = true
  owners      = ["amazon"]

  filter {
    name = "name"
    values = ["amzn-ami-minimal-hvm-*"]
  }
  filter {
    name = "root-device-type"
    values = ["ebs"]
  }
}

resource "aws_instance" "foo" {
  ami           = "${data.aws_ami.amzn-ami-minimal-hvm-ebs.id}"
```

nywilken:
@bflad thanks the review and context around testing. I'll circle back on this resource with the details outlined in #4987 and the CONTRIBUTING guide.

bflad:

PR #9008, opened 2019-06-15
Add clarification about lightsail's name uniqueness constraints

awendland:
### Community Note

```release-note
NONE
```

This is purely a documentation change to make it easier to get up and running with Lightsail instances using Terraform.

### DIFF
--- a/website/docs/r/lightsail_instance.html.markdown
+++ b/website/docs/r/lightsail_instance.html.markdown
@@ -31,8 +31,7 @@ resource "aws_lightsail_instance" "gitlab_test" {
 
 The following arguments are supported:
 
-* `name` - (Required) The name of the Lightsail Instance. *This must be unique (within your
-account), and cannot be the same as a name given to a `lightsail_static_ip`.*
+* `name` - (Required) The name of the Lightsail Instance. Names be unique within each AWS Region in your Lightsail account.
 * `availability_zone` - (Required) The Availability Zone in which to create your
 instance (see list below)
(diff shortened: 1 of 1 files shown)


aeschright:
Hi @awendland -- thanks for submitting this! It seems reasonable but I can't find this constraint in the AWS documentation. Can you point me in the right direction?

awendland:
Hi @aeschright, I wish I found the constraint in the AWS docs 😂 it would have saved me the couple hours it took to isolate my deployment issues down to this. I don't want to make this a big deal which requires setting up extensive test cases and such, so maybe we could change the phrasing to ```*This must be unique (within your account), and likely cannot be the same as a name given to a `lightsail_static_ip`.*``` (I added "likely" to the verbiage). My hope is that this will save someone substantial time, with very minimal downside.

In case you were referring to the uniqueness constraint in general, I can't find it explicitly in the docs but it's demonstrated by how it's used as a unique identifier in API calls such as https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lightsail.html#Lightsail.Client.attach_disk.

blckct:
Clarification might be useful but it's wrong. The name's are unique per region not per account. Also it's too specific.
As we see here: https://docs.aws.amazon.com/cli/latest/reference/lightsail/tag-resource.html we can tag a resource just by using it's name and so all the resources that use tags must have unique names. 

That means the name must be unique among : disks, domains, instances, key pairs, load balancers and databases and static ips (which don't support tags, weird). It seems only thing that's exempt is domain entries.

PR #9027, opened 2019-06-17
Automate labeling of new issues for triage

aeschright:
### Community Note

Adds a GitHub action workflow to automatically apply a `needs-triage` label to new issues for filtering.

```release-note
NONE
```

```
N/A
```

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -22,7 +22,7 @@ action "pr-label" {
 }
 
 workflow "Issue triage" {
-  on = "issue"
+  on = "issues"
   resolves = "Apply Triage Label"
 }
 
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -20,3 +20,14 @@ action "pr-label" {
         LABEL_SPEC_FILE = ".github/PULL_REQUEST_LABELS.yml"
     }
 }
+
+workflow "Issue triage" {
+  on = "issue"
+  resolves = "Apply Triage Label"
+}
+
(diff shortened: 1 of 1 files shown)


.github/main.workflow
```diff
@@ -20,3 +20,14 @@ action "pr-label" {
         LABEL_SPEC_FILE = ".github/PULL_REQUEST_LABELS.yml"
     }
 }
+
+workflow "Issue triage" {
+  on = "issue"
```
bflad: I believe the (confusing) correct value for this argument is `issues` according to the [GitHub Actions documentation](https://developer.github.com/actions/managing-workflows/workflow-configuration-options/#events-supported-in-workflow-files):

```suggestion
  on = "issues"
```

It appears the documentation in the https://github.com/actions/github repository has the same issue 😅

.github/main.workflow
```diff
@@ -5,7 +5,7 @@ workflow "Pull Request Updates" {
 
 action "pr-filter-sync" {
     uses = "actions/bin/filter@master"
-    args = "action 'opened|review_requested|synchronize'"
```
bflad: `review_requested` was adding in here due to GitHub Actions errors previously, but maybe those are fixed now. 🤞

.github/main.workflow
aeschright: Ha! Thanks for catching that.

.github/main.workflow
aeschright: 👍

PR #9208, opened 2019-07-02
Allow configuring assume role from env variables

svanharmelen:
This allows for a more dynamic configuration when used in Terraform Enterprise.

### DIFF
--- a/aws/provider.go
+++ b/aws/provider.go
@@ -1072,18 +1072,21 @@ func assumeRoleSchema() *schema.Schema {
 					Type:        schema.TypeString,
 					Optional:    true,
 					Description: descriptions["assume_role_role_arn"],
+					DefaultFunc: schema.EnvDefaultFunc("AWS_ASSUME_ROLE_ARN", nil),
 				},
 
 				"session_name": {
 					Type:        schema.TypeString,
 					Optional:    true,
 					Description: descriptions["assume_role_session_name"],
+					DefaultFunc: schema.EnvDefaultFunc("AWS_ASSUME_ROLE_SESSION_NAME", nil),
 				},
 
 				"external_id": {
 					Type:        schema.TypeString,
 					Optional:    true,
 					Description: descriptions["assume_role_external_id"],
+					DefaultFunc: schema.EnvDefaultFunc("AWS_ASSUME_ROLE_EXTERNAL_ID", nil),
 				},
 
 				"policy": {
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -278,13 +278,16 @@ for more information about connecting to alternate AWS endpoints or AWS compatib
 
 The nested `assume_role` block supports the following:
 
-* `role_arn` - (Required) The ARN of the role to assume.
+* `role_arn` - (Optional) The ARN of the role to assume. It must be provided, but
+  it can also be sourced from the `AWS_ASSUME_ROLE_ARN` environment variable.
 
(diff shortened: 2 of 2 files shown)


aws/provider.go
```diff
@@ -1072,18 +1072,21 @@ func assumeRoleSchema() *schema.Schema {
...
 					Optional:    true,
 					Description: descriptions["assume_role_role_arn"],
+					DefaultFunc: schema.EnvDefaultFunc("AWS_ASSUME_ROLE_ARN", nil),
 				},
 
 				"session_name": {
 					Type:        schema.TypeString,
 					Optional:    true,
 					Description: descriptions["assume_role_session_name"],
+					DefaultFunc: schema.EnvDefaultFunc("AWS_ASSUME_SESSION_NAME", nil),
```
ewbankkit: Maybe `AWS_ASSUME_ROLE_SESSION_NAME` to match the AWS code property name?
https://github.com/terraform-providers/terraform-provider-aws/blob/905ba44ccd6b3e442f709878abadf458e114f0cc/aws/provider.go#L1023-L1025

aws/provider.go
```diff
@@ -1072,18 +1072,21 @@ func assumeRoleSchema() *schema.Schema {
...
 				},
 
 				"external_id": {
 					Type:        schema.TypeString,
 					Optional:    true,
 					Description: descriptions["assume_role_external_id"],
+					DefaultFunc: schema.EnvDefaultFunc("AWS_ASSUME_EXTERNAL_ID", nil),
```
ewbankkit: Maybe `AWS_ASSUME_ROLE_EXTERNAL_ID` to match the AWS code property name?
https://github.com/terraform-providers/terraform-provider-aws/blob/905ba44ccd6b3e442f709878abadf458e114f0cc/aws/provider.go#L1023-L1025

svanharmelen:
@ewbankkit sure, update the env names 👍

bflad:
Hi @svanharmelen 👋 I'm not sure if you saw #8985, but its attempting to solve the same problem and has been through some design cycles. The problem with just adding `DefaultFunc` to the existing attributes is that the provider configuration requires the following:

```hcl
provider "aws" {
  assume_role {}
}
```

Ideally though, we need to work with the AWS Go SDK development team to get standardized environment variable naming. I'm very hesitant about us creating our own standard in the `AWS_` namespace since the AWS Go SDK can introduce unexpected breaking changes such as https://github.com/aws/aws-sdk-go/pull/2667/files#diff-83705f3ccd6052c21c7195e6c8a39664R45

svanharmelen:
Hi @bflad! No, I didn't see #8985 so thanks for pointing that one out!

About the provider needing:

```go
provider "aws" {
  assume_role {}
}
```

That is correct, but in a way I thought it would make sense as it allows you to be explicit about the fact that you want to assume a role while allowing you to specify which role using env vars. 

> Ideally though, we need to work with the AWS Go SDK development team to get standardized environment variable naming.

That would indeed be ideal, but I wonder how long that is going to take and if it makes sense to wait on that?

But after looking at #8985 I realized I just totally didn't think about using Terraform variables. I think I mixed up my thoughts a bit as I thought about interpolation not working in providers. But of course normal Terraform variables work just fine in providers.

While I still prefer a solution to set these values using env vars directly, for now I'll just update my configs and use Terraform vars instead. If you think it makes sense to close this PR in favor of the ongoing discussion in #8985 I'm totally fine by that.

svanharmelen:
I guess this has been open for long enough 😏

PR #9217, opened 2019-07-02
Add support to import terraform aws_iam_group_policy resources

### LINKED ISSUE #8788: IAM group policy "does not support import"
Terraform 0.12.0
provider.aws ~> 1.17

Action:
Importing a existing aws_iam_group as stated in the following documentation:
https://www.terraform.io/docs/providers/aws/r/iam_group_policy.html

Result:
Error: resource aws_iam_group_policy doesn't support import

This issue might be similar to #8785 .

lucioveloso:
### Community Note

Fixes #8788 

```release-note
resource/aws_iam_group_policy: Support resource import
```

```
$ make testacc TESTARGS='-run=TestAccAWSIAMGroupPolicy_importBasic'
=== RUN   TestAccAWSIAMGroupPolicy_importBasic
--- PASS: TestAccAWSIAMGroupPolicy_importBasic (30.81s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       30.845s
...
```

### DIFF
--- a/aws/resource_aws_iam_group_policy.go
+++ b/aws/resource_aws_iam_group_policy.go
@@ -22,6 +22,10 @@ func resourceAwsIamGroupPolicy() *schema.Resource {
 		Read:   resourceAwsIamGroupPolicyRead,
 		Delete: resourceAwsIamGroupPolicyDelete,
 
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
 		Schema: map[string]*schema.Schema{
 			"policy": {
 				Type:     schema.TypeString,
@@ -58,7 +62,13 @@ func resourceAwsIamGroupPolicyPut(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	var policyName string
-	if v, ok := d.GetOk("name"); ok {
+	var err error
+	if !d.IsNewResource() {
+		_, policyName, err = resourceAwsIamGroupPolicyParseId(d.Id())
+		if err != nil {
+			return err
+		}
+	} else if v, ok := d.GetOk("name"); ok {
 		policyName = v.(string)
 	} else if v, ok := d.GetOk("name_prefix"); ok {
 		policyName = resource.PrefixedUniqueId(v.(string))
@@ -78,14 +88,16 @@ func resourceAwsIamGroupPolicyPut(d *schema.ResourceData, meta interface{}) erro
 func resourceAwsIamGroupPolicyRead(d *schema.ResourceData, meta interface{}) error {
 	iamconn := meta.(*AWSClient).iamconn
 
-	group, name := resourceAwsIamGroupPolicyParseId(d.Id())
+	group, name, err := resourceAwsIamGroupPolicyParseId(d.Id())
+	if err != nil {
+		return err
+	}
 
 	request := &iam.GetGroupPolicyInput{
 		PolicyName: aws.String(name),
 		GroupName:  aws.String(group),
 	}
 
-	var err error
 	getResp, err := iamconn.GetGroupPolicy(request)
 	if err != nil {
 		if isAWSErr(err, iam.ErrCodeNoSuchEntityException, "") {
@@ -105,17 +117,22 @@ func resourceAwsIamGroupPolicyRead(d *schema.ResourceData, meta interface{}) err
 		return err
 	}
 
-	d.Set("group", group)
-	d.Set("name", name)
-	d.Set("policy", policy)
-
-	return nil
+	if err := d.Set("policy", policy); err != nil {
+		return err
+	}
+	if err := d.Set("name", name); err != nil {
+		return err
+	}
+	return d.Set("group", group)
 }
 
 func resourceAwsIamGroupPolicyDelete(d *schema.ResourceData, meta interface{}) error {
 	iamconn := meta.(*AWSClient).iamconn
 
-	group, name := resourceAwsIamGroupPolicyParseId(d.Id())
+	group, name, err := resourceAwsIamGroupPolicyParseId(d.Id())
+	if err != nil {
+		return err
+	}
 
 	request := &iam.DeleteGroupPolicyInput{
 		PolicyName: aws.String(name),
@@ -131,8 +148,13 @@ func resourceAwsIamGroupPolicyDelete(d *schema.ResourceData, meta interface{}) e
 	return nil
 }
 
-func resourceAwsIamGroupPolicyParseId(id string) (groupName, policyName string) {
+func resourceAwsIamGroupPolicyParseId(id string) (groupName, policyName string, err error) {
 	parts := strings.SplitN(id, ":", 2)
+	if len(parts) != 2 {
+		err = fmt.Errorf("group_policy id must be of the form <group name>:<policy name>")
+		return
+	}
+
 	groupName = parts[0]
 	policyName = parts[1]
 	return
--- a/aws/resource_aws_iam_group_policy.go
+++ b/aws/resource_aws_iam_group_policy.go
@@ -118,12 +118,18 @@ func resourceAwsIamGroupPolicyRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	if err := d.Set("policy", policy); err != nil {
-		return err
+		return fmt.Errorf("error setting policy: %s", err)
 	}
+
 	if err := d.Set("name", name); err != nil {
-		return err
+		return fmt.Errorf("error setting name: %s", err)
 	}
-	return d.Set("group", group)
+
+	if err := d.Set("group", group); err != nil {
+		return fmt.Errorf("error setting group: %s", err)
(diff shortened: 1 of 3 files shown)


aws/resource_aws_iam_group_policy.go
```diff
@@ -58,7 +62,13 @@ func resourceAwsIamGroupPolicyPut(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	var policyName string
-	if v, ok := d.GetOk("name"); ok {
+	var err error
+	if !d.IsNewResource() {
```
bflad: Is this new logic necessary? I think if it is, rather than introducing this new `d.IsNewResource()` logic here, instead we typically prefer to split the resource `Create` and `Update` functions to remove any confusion. The `Update` function can be created by duplicating the _prior_ `resourceAwsIamGroupPolicyPut()` function and using `resourceAwsIamGroupPolicyParseId()`.

aws/resource_aws_iam_group_policy.go
```diff
@@ -105,17 +117,22 @@ func resourceAwsIamGroupPolicyRead(d *schema.ResourceData, meta interface{}) err
 		return err
 	}
 
-	d.Set("group", group)
-	d.Set("name", name)
-	d.Set("policy", policy)
-
-	return nil
+	if err := d.Set("policy", policy); err != nil {
+		return err
```
bflad: If we are going to add error checking for these attributes, the errors should include context for operators and code maintainers to understand the issue, e.g.

```suggestion
		return fmt.Errorf("error setting policy: %s", err)
```

aws/resource_aws_iam_group_policy.go
```diff
@@ -105,17 +117,22 @@ func resourceAwsIamGroupPolicyRead(d *schema.ResourceData, meta interface{}) err
...
+	}
+	if err := d.Set("name", name); err != nil {
+		return err
```
bflad: Same here regarding error context. 👍 

```suggestion
		return fmt.Errorf("error setting name: %s", err)
```

aws/resource_aws_iam_group_policy.go
```diff
@@ -105,17 +117,22 @@ func resourceAwsIamGroupPolicyRead(d *schema.ResourceData, meta interface{}) err
...
+	}
+	return d.Set("group", group)
```
bflad: Same here regarding error context

```suggestion
	if err := d.Set("group", group); err != nil {
		return fmt.Errorf("error setting group: %s", err)
	}
	return nil
```

aws/resource_aws_iam_group_policy.go
```diff
@@ -131,8 +148,13 @@ func resourceAwsIamGroupPolicyDelete(d *schema.ResourceData, meta interface{}) e
 	return nil
 }
 
-func resourceAwsIamGroupPolicyParseId(id string) (groupName, policyName string) {
+func resourceAwsIamGroupPolicyParseId(id string) (groupName, policyName string, err error) {
 	parts := strings.SplitN(id, ":", 2)
+	if len(parts) != 2 {
```
bflad: This check should also include empty string checks for before and after the colon 😄 

```suggestion
	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
```

aws/resource_aws_iam_group_policy_test.go
```diff
@@ -139,6 +139,28 @@ func TestAccAWSIAMGroupPolicy_generatedName(t *testing.T) {
 	})
 }
 
+func TestAccAWSIAMGroupPolicy_importBasic(t *testing.T) {
```
bflad: Rather than creating a separate import test, which duplicates test code/infrastructure and does not test all resource configurations, we prefer to now add the import `TestStep` to all applicable existing tests. See also: https://github.com/terraform-providers/terraform-provider-aws/issues/8944

Here this means removing the new `TestAccAWSIAMGroupPolicy_importBasic` and `testAccAwsIamGroupPolicyConfig` while copying the code block noted below to each existing acceptance test:

```go
			{
				ImportState:       true,
				ImportStateVerify: true,
			},
```

aws/resource_aws_iam_group_policy_test.go
```diff
@@ -147,23 +169,28 @@ func testAccCheckIAMGroupPolicyDestroy(s *terraform.State) error {
...
 			GroupName:  aws.String(group),
 		}
 
-		_, err := conn.GetGroupPolicy(request)
+		getResp, err := conn.GetGroupPolicy(request)
 		if err != nil {
-			// Verify the error is what we want
-			if ae, ok := err.(awserr.Error); ok && ae.Code() == "NoSuchEntity" {
-				continue
+			if iamerr, ok := err.(awserr.Error); ok && iamerr.Code() == "NoSuchEntity" {
```
bflad: If this logic is being updated, it should use our helper function and the AWS Go SDK provided constant for the error code:

```suggestion
			if isAWSErr(err, iam.ErrCodeNoSuchEntityException, "") {
```

aws/resource_aws_iam_group_policy_test.go
```diff
@@ -147,23 +169,28 @@ func testAccCheckIAMGroupPolicyDestroy(s *terraform.State) error {
...
+				// none found, that's good
+				return nil
```
bflad: The previous logic using `continue` was correct and should be reverted. `CheckDestroy` functions loop through all provisioned resources of a type and all those resources should be checked.

```suggestion
				continue
```

aws/resource_aws_iam_group_policy.go
lucioveloso: Actually I guess we can keep it as before. I reverted it.

lucioveloso:
Hi @bflad. Thanks for the review/feedback. 👍I updated the code following the instructions. Please let me know if something else is needed.

aws/resource_aws_iam_group_policy.go
lucioveloso: Done.

aws/resource_aws_iam_group_policy.go
lucioveloso: Done.

aws/resource_aws_iam_group_policy.go
lucioveloso: Done.

aws/resource_aws_iam_group_policy.go
lucioveloso: Done.

aws/resource_aws_iam_group_policy_test.go
lucioveloso: Done. Just skipped the `_disappears`.

aws/resource_aws_iam_group_policy_test.go
lucioveloso: Done.

aws/resource_aws_iam_group_policy_test.go
lucioveloso: Done. Actually in the line "161".

aws/resource_aws_iam_group_policy_test.go
```diff
@@ -101,6 +110,11 @@ func TestAccAWSIAMGroupPolicy_namePrefix(t *testing.T) {
 					testAccCheckAWSIAMGroupPolicyNameMatches(&groupPolicy1, &groupPolicy2),
 				),
 			},
+			{
+				ResourceName:      "aws_iam_group_policy.test",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
bflad: One minor fix required here:

```
--- FAIL: TestAccAWSIAMGroupPolicy_namePrefix (18.56s)
    testing.go:568: Step 2 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=11) "name_prefix": (string) (len=24) "test-7166041588452991103"
        }
```

At some point in the future we may make the resources `XXX_prefix` "aware" during import by checking for the format of the name, etc. value for the dash and number of numeric characters that would match our generated values from these prefix fields, but for now these prefix attribute only exist within Terraform (and not the API to refresh during `Read`) so we can ignore the difference on import for now.

```suggestion
				ImportStateVerifyIgnore: []string{"name_prefix"},
			},
```

bflad:
This has been released in [version 2.18.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2180-july-05-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

For further feature requests or bug reports with this functionality, please create a [new GitHub issue](https://github.com/terraform-providers/terraform-provider-aws/issues/new/choose) following the template for triage. Thanks!

PR #9247, opened 2019-07-05
docs/provider: Initial Adding Resource Import Support section for Contributing Guide

bflad:
### Community Note

```release-note
NONE
```

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -15,9 +15,9 @@ ability to merge PRs and respond to issues.
 
 - [Issues](#issues)
     - [Issue Reporting Checklists](#issue-reporting-checklists)
-        - [Bug Reports](#bug-reports)
-        - [Feature Requests](#feature-requests)
-        - [Questions](#questions)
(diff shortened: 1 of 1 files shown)


bflad:
> A lot of the information would be helpful in the Extend Terraform Documentation

I agree! It would be great to link out to that documentation where possible. I think the [Resources page](https://www.terraform.io/docs/extend/resources.html) which currently has a little bit of information about `Importer` amongst other functionality, should probably become its own full section with each of the topics on its own page before we expand it even more. Let me take some time to submit the first change of splitting up that large documentation page and then we can circle back on updating the documentation after the lift and shift. 👍

Do you think its worthwhile getting this in as-is now or waiting until we can link out?

bflad:
I submitted a pull request for the splitting the Resources page in the Extend documentation: https://github.com/hashicorp/terraform-website/pull/847 😄

bflad:
Here's the pull request to get the this information into the Extend documentation: https://github.com/hashicorp/terraform-website/pull/848

PR #9265, opened 2019-07-08
fixes #8230

### LINKED ISSUE #8230: 2.4.0 -> 2.5.0 migration -  log_subscription_filter keeps re-planning "distribution" attribute change
### Community Note

### Terraform Version

TF: 0.11.11
AWS provider: 2.5.0 (current infrastructure was stable on version 2.4.0)

### Affected Resource(s)

*  aws_cloudwatch_log_subscription_filter

```hcl
resource "aws_cloudwatch_log_subscription_filter"
"api-svc-ec2-prd-cloudwatch-subscription"
{
  depends_on = [
    "aws_lambda_permission.api-svc-ec2-prd-sumologic-logging-allow-cloudwatch-lambda"
  ]
  name = "api-svc-ec2-prd-cloudwatch-subscription"
  filter_pattern = ""
  log_group_name =
    "${aws_cloudwatch_log_group.api-svc-ec2-prd-loggroup.name}"
  destination_arn =
    "${aws_lambda_function.cloudwatch-sumologic-logging-lambda.arn}"
}
```

### Steps to Reproduce

1. `terraform plan`

Plan output shows that TF is going to set:  `distribution: "ByLogStream" => ""`

2. ` terraform apply`

Apply output shows that TF is setting:  `distribution: "ByLogStream" => ""`

3. `terraform apply`

Plan output shows that TF is going to set:  `distribution: "ByLogStream" => ""`

### Workaround

Update the resource and add attribute `distribution = "ByLogStream"`

dedamico:
### Community Note

Fixes #8230

```release-note
resource/aws_cloudwatch_log_subscription_filter: prevent change being reported for blank distribution attribute
```

### DIFF
--- a/aws/resource_aws_cloudwatch_log_subscription_filter.go
+++ b/aws/resource_aws_cloudwatch_log_subscription_filter.go
@@ -54,7 +54,13 @@ func resourceAwsCloudwatchLogSubscriptionFilter() *schema.Resource {
 			"distribution": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Computed: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// Ignore unsetting default value
+					if old == "ByLogStream" && new == "" {
+						return true
+					}
+					return false
+				},
 			},
 		},
 	}
--- a/aws/resource_aws_cloudwatch_log_subscription_filter.go
+++ b/aws/resource_aws_cloudwatch_log_subscription_filter.go
@@ -54,6 +54,7 @@ func resourceAwsCloudwatchLogSubscriptionFilter() *schema.Resource {
 			"distribution": {
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


bflad:
Hi @dedamico 👋 Would you happen to have a configuration that triggers the behavior so we can add a covering acceptance test? The change to `Computed: true` is likely okay if the value has different defaults depending on other arguments, although it'd be better if we could use a `DiffSuppressFunc` based on those other triggering arguments so Terraform can continue to perform drift detection when its not configured. Thanks!

dedamico:
Hi @bflad!  Thanks for the feedback! 

I have done the following with the latest commit to my branch:

1. Updated the code to use ``DiffSuppressFunc`` rather than ``Computed``.
2. I took a stab at updating the acceptance test to add a Test Step for the slightly different config.  The added config's only difference is that the ``distribution`` attribute is not set, which is the use case which the bug happened with (AWS API sets default distribution value to ``ByLogStream`` if it is not set).

please let me know if anything looks off/wrong and I'll fix it!

samm-git:
@bflad hi, any chance to get this merged?

bflad:
(This was merged in after being rebased - the commits are there despite me closing this 😄 )

PR #9283, opened 2019-07-09
Allow for creating a IoT Core certificate without CSR.

borisroman:
### Community Note

Allow for creating a IoT Core certificate without CSR. Makes the CSR field optional and adds the CreateKeysAndCertificate API call.

### DIFF
--- a/aws/resource_aws_iot_certificate.go
+++ b/aws/resource_aws_iot_certificate.go
@@ -17,7 +17,7 @@ func resourceAwsIotCertificate() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"csr": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 				ForceNew: true,
 			},
 			"active": {
@@ -28,6 +28,21 @@ func resourceAwsIotCertificate() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"certificate_pem": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+			"public_key": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+			"private_key": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


borisroman:
Hi @bflad @ryndaniels 

Does one of you have time to take a look at this PR?

Thanks in advance

aws/resource_aws_iot_certificate.go
```diff
@@ -28,26 +28,58 @@ func resourceAwsIotCertificate() *schema.Resource {
...
 
 func resourceAwsIotCertificateCreate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).iotconn
 
-	log.Printf("[DEBUG] Creating certificate from csr")
-	out, err := conn.CreateCertificateFromCsr(&iot.CreateCertificateFromCsrInput{
-		CertificateSigningRequest: aws.String(d.Get("csr").(string)),
-		SetAsActive:               aws.Bool(d.Get("active").(bool)),
-	})
+	if d.Get("csr") != "" {
```
bflad: Since `d.Get()` returns an interface, we should either check it after type assertion or switch the check to `d.GetOk()` which will automatically check the value against the zero value for the type:

```suggestion
	if _, ok := d.GetOk("csr"); ok {
```

aws/resource_aws_iot_certificate.go
```diff
@@ -28,26 +28,58 @@ func resourceAwsIotCertificate() *schema.Resource {
...
+		})
 
-	if err != nil {
-		log.Printf("[ERROR] %s", err)
-		return err
-	}
-	log.Printf("[DEBUG] Created certificate from csr")
+		if err != nil {
+			log.Printf("[ERROR] %s", err)
+			return err
```
bflad: Terraform will automatically log returned errors so the line above this is extraneous. We should also return some error context here for operators and code maintainers about when this error occurred, e.g.

```suggestion
			return fmt.Errorf("error creating IoT Certificate from CSR: %s", err)
```

aws/resource_aws_iot_certificate.go
```diff
@@ -28,26 +28,58 @@ func resourceAwsIotCertificate() *schema.Resource {
...
+		d.SetId(*out.CertificateId)
+	} else {
+		log.Printf("[DEBUG] Creating keys and certificate")
+		out, err := conn.CreateKeysAndCertificate(&iot.CreateKeysAndCertificateInput{
+			SetAsActive: aws.Bool(d.Get("active").(bool)),
+		})
+
+		if err != nil {
+			log.Printf("[ERROR] %s", err)
+			return err
```
bflad: Same here regarding the log message and error context 👍 

```suggestion
			return fmt.Errorf("error creating IoT Keys and Certificate: %s", err)
```

website/docs/r/iot_certificate.html.markdown
```diff
@@ -11,21 +11,38 @@ description: |-
...
-* `csr` - (Required) The certificate signing request. Review the
-  [IoT API Reference Guide] (http://docs.aws.amazon.com/iot/latest/apireference/API_CreateCertificateFromCsr.html)
-  for more information on creating a certificate from a certificate signing request (CSR).
+* `csr` - (Optional) The certificate signing request. Review
+  [CreateCertificateFromCsr](https://docs.aws.amazon.com/iot/latest/apireference/API_CreateCertificateFromCsr.html)
+  for more information on generating a certificate from a certificate signing request (CSR).
+  If none is specified both the certificate and keys will be generated, review [CreateKeysAndCertificate](https://docs.aws.amazon.com/iot/latest/apireference/API_CreateKeysAndCertificate.html)
+  for more information on generating keys and a certificate.
+
+## Attributes Reference
```
bflad: Please remove the duplicate documentation section below. 😄

svanharmelen:
@bflad all feedback is addressed by the latest commit.

PR #9372, opened 2019-07-16
New resources: Traffic mirroring

### LINKED ISSUE #9140: VPC Traffic Mirroring
### Community Note

### Description

VPC Traffic Mirroring launched today.
~[AWS SDK v1.20.8](https://github.com/aws/aws-sdk-go/releases/tag/v1.20.8) not yet released so exact API details not available.~

* aws_vpc_traffic_mirror_filter
* aws_vpc_traffic_mirror_filter_rule
* aws_vpc_traffic_mirror_session
* aws_vpc_traffic_mirror_target

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* [Announcement](https://aws.amazon.com/about-aws/whats-new/2019/06/announcing-amazon-vpc-traffic-mirroring-for-amazon-ec2-instances/)
* [Blog post](https://aws.amazon.com/blogs/aws/new-vpc-traffic-mirroring/)
* [`CreateTrafficMirrorFilter`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTrafficMirrorFilter.html)
* [`CreateTrafficMirrorFilterRule`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTrafficMirrorFilterRule.html)
* [`CreateTrafficMirrorSession`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTrafficMirrorSession.html)
* [`CreateTrafficMirrorTarget`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTrafficMirrorTarget.html)

johnthedev97:
Adding support for AWS Traffic mirroring resources

### Community Note

Fixes #9140

```release-note
Added the following resources to support AWS Traffic mirroring
* aws_vpc_traffic_mirror_filter
* aws_vpc_traffic_mirror_filter_rule
* aws_vpc_traffic_mirror_session
* aws_vpc_traffic_mirror_target
```

```
make testacc TESTARGS="-run=TestAccAWSTrafficMirror"
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSTrafficMirror -timeout 120m
=== RUN   TestAccAWSTrafficMirrorFilterRule_basic
=== PAUSE TestAccAWSTrafficMirrorFilterRule_basic
=== RUN   TestAccAWSTrafficMirrorFilter_basic
=== PAUSE TestAccAWSTrafficMirrorFilter_basic
=== RUN   TestAccAWSTrafficMirrorSession_basic
=== PAUSE TestAccAWSTrafficMirrorSession_basic
=== RUN   TestAccAWSTrafficMirrorTarget_nlb
=== PAUSE TestAccAWSTrafficMirrorTarget_nlb
=== RUN   TestAccAWSTrafficMirrorTarget_eni
=== PAUSE TestAccAWSTrafficMirrorTarget_eni
=== CONT  TestAccAWSTrafficMirrorFilterRule_basic
=== CONT  TestAccAWSTrafficMirrorTarget_eni
=== CONT  TestAccAWSTrafficMirrorSession_basic
=== CONT  TestAccAWSTrafficMirrorTarget_nlb
=== CONT  TestAccAWSTrafficMirrorFilter_basic
--- PASS: TestAccAWSTrafficMirrorFilter_basic (47.58s)
--- PASS: TestAccAWSTrafficMirrorFilterRule_basic (50.55s)
--- PASS: TestAccAWSTrafficMirrorTarget_eni (111.66s)
--- PASS: TestAccAWSTrafficMirrorTarget_nlb (290.85s)
--- PASS: TestAccAWSTrafficMirrorSession_basic (295.84s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	295.927s

...
```


johnthedev97:
This became a little larger and longer than I thought. I have taken the [WIP] as I assume this will get reviewed only after that. 

One thing I myself was not certain is regarding virtual_network_id in aws_traffic_mirror_session. The AWS API has set this field as optional, but if not passed AWS selects a random value for this. If I read this value and save in state it causes a change in every plan as the value from config comes by default as "0". So as a workaround I am reading this value into state only if it is set in config as well. If there is a better way to handle this let me know. I also had to set this field to ignore in import verify test.

Another doubt I had was regarding the state management. While writing tests I noticed that for many fields which AWS allows removal terraform doesn't remove those attributes from state rather keep the attribute with Zero value in the state. This made me write many tests to use TestCheckResourceAttr though logically it made sense to use TestCheckNoResourceAttr. If terraform is by design storing remove attributes in state with Zero vale, I was wondering it would be a good idea to re-write TestCheckNoResourceAttr such that it passes even if the attribute is present but with Zero value.

johnthedev97:
@bflad Sorry to ping you directly, but since you are the one who merged my previous PR, I thought of asking you what does it take to get this merged?

pbartruff:
Folks,

Is there reason this is lingering?  It doesn't seem to cause and conflicts or is there a different effort underway to address the issue?

R/
Paul

aeschright:
Hi everyone, [per our guidelines](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#new-resource) it will be much easier for us to review and test these changes if each resource is in its own PR. Long PRs tend to linger because we have so many things in the queue, so I recommend you split it and resubmit. Thanks so much for getting this started!

pbartruff:
@aeschright Thanks!
I'm not the requester; however, I understand the need for "small bites" at a time.  Sadly all of these resources would be required to mirror packets.  Individually they don't provide much.  Maybe @johnthedev97 will be able to break it up a bit.

johnthedev97:
@aeschright , like @pbartruff has mentioned the resources are interdependent. traffic mirror filter rule tests will fail without traffic mirror filter. Like traffic mirror session tests will fail without traffic mirror or traffic mirror filter rule. And the tests for session will fail without any of the other three resources. So I feel all these belong together, and should be reviewed together as well. Still if you think this needs to be broke into multiple PRs let me know, I will try my best to do it.

chrislujan:
Just checking in, hoping we haven't forgot about this.

stevesmoot:
It really doesnt seem like splitting and resubmitting will help here, even if better per guidelines.

Will-Low:
Anyone know when this can get a review? It's blocking #9140.

zephyia:
I'd like to echo the request to get this merged, we are using this currently and have to keep building the aws provider module with this patch. How do we get the great work done by @johnthedev97 reviewed and merged?

stevesmoot:
+1

chrismatteson:
This is a blocker for ExtraHop which is impacting numerous of their customers, as well as holding up a potential joint go to market strategy for us. Can we get this reviewed asap?

aws/resource_aws_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	var removeFields []*string
+	if d.HasChange("protocol") {
+		_, n := d.GetChange("protocol")
+		if n == "0" {
+			removeFields = append(removeFields, aws.String("protocol"))
+		} else {
+			input.Protocol = aws.Int64(int64(d.Get("protocol").(int)))
+		}
```
bflad: Only using the second return value of `d.GetChange()` is equivalent to `d.Get()`, so would recommend the following (along with AWS Go SDK constants) here and below:

```go
		v := d.Get("protocol").(int)
		if v == 0 {
			removeFields = append(removeFields, aws.String(ec2.TrafficMirrorFilterRuleFieldProtocol))
		} else {
			input.Protocol = aws.Int64(int64(v))
		}
```

See also: https://github.com/bflad/tfproviderlint/issues/44

aws/resource_aws_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,369 @@
...
+
+	if v, ok := portSchema["to_port"]; ok {
+		portRange.ToPort = aws.Int64(int64(v.(int)))
+		out = &portRange
+	}
+
+	return out
+}
+
+func populateTrafficMirrorFilterRuleResource(d *schema.ResourceData, rule *ec2.TrafficMirrorFilterRule) error {
```
bflad: Since this is only used once, let's remove the indirection and include this directly in the Read function. 👍

aws/resource_aws_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,369 @@
...
+	d.Set("traffic_mirror_filter_id", rule.TrafficMirrorFilterId)
+	d.Set("destination_cidr_block", rule.DestinationCidrBlock)
+	d.Set("source_cidr_block", rule.SourceCidrBlock)
+	d.Set("rule_action", rule.RuleAction)
+	d.Set("rule_number", rule.RuleNumber)
+	d.Set("traffic_direction", rule.TrafficDirection)
+	d.Set("description", rule.Description)
+	d.Set("protocol", rule.Protocol)
+	d.Set("destination_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule.DestinationPortRange))
+	d.Set("source_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule.SourcePortRange))
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("destination_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule.DestinationPortRange)); err != nil {
	return fmt.Errorf("error setting destination_port_range: %s", err)
}
if err := d.Set("source_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule. SourcePortRange)); err != nil {
	return fmt.Errorf("error setting source_port_range: %s", err)
}
```

aws/resource_aws_traffic_mirror_filter_rule_test.go
```diff
@@ -0,0 +1,249 @@
...
+		ruleId := rs.Primary.ID
+		filterId := rs.Primary.Attributes["traffic_mirror_filter_id"]
+
+		out, err := conn.DescribeTrafficMirrorFilters(&ec2.DescribeTrafficMirrorFiltersInput{
+			TrafficMirrorFilterIds: []*string{
+				aws.String(filterId),
+			},
+		})
+
+		if isAWSErr(err, "InvalidTrafficMirrorFilterId.NotFound", "") {
```
bflad: If this is the error that is returned when an EC2 Traffic Mirror Filter cannot be found, we should include this in the filter and filter rule resource Read functions to trigger Terraform resource recreation:

```go
// Resource Read functions, potentially in addition to over nil/length checks from DescribeTrafficMirrorFilters call
if isAWSErr(err, "InvalidTrafficMirrorFilterId.NotFound", "") {
	log.Printf("[WARN] EC2 Traffic Mirror Filter (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}
```

aws/resource_aws_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+		TrafficMirrorSessionIds: []*string{
+			&sessionId,
+		},
+	}
+
+	out, err := conn.DescribeTrafficMirrorSessions(input)
+	if nil != err {
+		d.SetId("")
+		return fmt.Errorf("Error describing traffic mirror session %v", sessionId)
+	}
```
bflad: We should not remove this resource from the Terraform state on API errors and should return the error itself.

```
if err != nil {
	return fmt.Errorf("error describing EC2 Traffic Mirror Session (%s): %w", d.Id(), err)
}
```

aws/resource_aws_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if 0 == len(out.TrafficMirrorSessions) {
+		d.SetId("")
+		return fmt.Errorf("Unable to find traffic mirrir session %v", sessionId)
+	}
```
bflad: When triggering resource recreation we should prefer a warning log and return no error.

```
if out == nil || len(out.TrafficMirrorSessions) == 0 {
	log.Printf("[WARN] EC2 Traffic Mirror Session (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}
```

aws/resource_aws_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+	d.Set("traffic_mirror_target_id", session.TrafficMirrorTargetId)
+	d.Set("description", session.Description)
+	d.Set("packet_length", session.PacketLength)
+
+	// AWS returns a value for VNI even if nothing is set in config. To avoid further plan
+	// showing a change, using a workaround - set this in state only if it was set in config
+	_, vniSet := d.GetOk("virtual_network_id")
+	if vniSet && nil != session.VirtualNetworkId {
+		d.Set("virtual_network_id", *session.VirtualNetworkId)
+	}
```
bflad: We should always prefer to call `d.Set()` in Read functions -- `d.Get()` and `d.GetOk()` logic does not work during import. If this configuration is optional and is defaulted by the API in that case, then setting `Computed: true` on the schema attribute is more appropriate.

```go
			"virtual_network_id": {
				Type:         schema.TypeInt,
				Optional:     true,
				Computed:     true,
				ValidateFunc: validation.IntBetween(1, 16777216),
			},
```

```go
d.Set("virtual_network_id", session.VirtualNetworkId)
```

aws/resource_aws_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+	input := &ec2.DeleteTrafficMirrorSessionInput{
+		TrafficMirrorSessionId: &sessionId,
+	}
+
+	_, err := conn.DeleteTrafficMirrorSession(input)
+	if nil != err {
+		return fmt.Errorf("Error deleting traffic mirror session %v", sessionId)
+	}
+
+	d.SetId("")
```
bflad: Nit:

```suggestion
```

aws/resource_aws_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+				Optional: true,
+				ForceNew: true,
+			},
+			"network_interface_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ConflictsWith: []string{
+					"network_load_balancer_arn",
+				},
```
bflad: If the API requires one or the other of these two attributes, then we can use `ExactlyOneOf` instead of `ConflictsWith` to have Terraform return an error if both are missing or both are declared. 👍 

```go
				ExactlyOneOf: []string{
					"network_interface_id",
					"network_load_balancer_arn",
				},
```

aws/resource_aws_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+			},
+			"network_load_balancer_arn": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ConflictsWith: []string{
+					"network_interface_id",
+				},
```
bflad:
```go
				ExactlyOneOf: []string{
					"network_interface_id",
					"network_load_balancer_arn",
				},
```

aws/resource_aws_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+	targetId := d.Id()
+	input := &ec2.DescribeTrafficMirrorTargetsInput{
+		TrafficMirrorTargetIds: []*string{&targetId},
+	}
+
+	out, err := conn.DescribeTrafficMirrorTargets(input)
+	if err != nil {
+		d.SetId("")
+		return fmt.Errorf("Error describing traffic mirror target %v", targetId)
+	}
```
bflad: Same note as above about not removing from state and returning the error itself.

aws/resource_aws_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if 1 != len(out.TrafficMirrorTargets) {
+		d.SetId("")
+		return fmt.Errorf("Not able to find find target %v", targetId)
+	}
```
bflad: Same note as above about warning logging and returning no error.

aws/resource_aws_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+
+	target := out.TrafficMirrorTargets[0]
+	if nil != target.Description {
+		d.Set("description", *target.Description)
+	}
```
bflad:
```go
d.Set("description", target.Description)
```

aws/resource_aws_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+	input := &ec2.DeleteTrafficMirrorTargetInput{
+		TrafficMirrorTargetId: &targetId,
+	}
+
+	_, err := conn.DeleteTrafficMirrorTarget(input)
+	if nil != err {
+		return fmt.Errorf("Error deleting traffic mirror target %v", targetId)
+	}
+
+	d.SetId("")
```
bflad:
```suggestion
```

aws/resource_aws_traffic_mirror_target_test.go
```diff
@@ -0,0 +1,241 @@
...
+			continue
+		}
+
+		out, err := conn.DescribeTrafficMirrorTargets(&ec2.DescribeTrafficMirrorTargetsInput{
+			TrafficMirrorTargetIds: []*string{
+				aws.String(rs.Primary.ID),
+			},
+		})
+
+		if isAWSErr(err, "InvalidTrafficMirrorTargetId.NotFound", "") {
```
bflad: Same note as above here about including this in the resource Read function to trigger resource recreation.

website/aws.erb
```diff
@@ -2950,6 +2950,31 @@
                         </li>
                     </ul>
                 </li>
+                <li>
+                    <a href="#">Traffic Mirror Resources</a>
```
bflad: These resources should be included in the EC2 resources instead of a new category. 👍

website/docs/r/traffic_mirror_filter.html.markdown
```diff
@@ -0,0 +1,45 @@
+---
+subcategory: "Traffic Mirroring"
```
bflad: For here and others below

```suggestion
subcategory: "EC2"
```

website/docs/r/traffic_mirror_filter.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+layout: "aws"
+page_title: "AWS: aws_traffic_mirror_filter"
+sidebar_current: "docs-aws-resource-traffic-mirror-filter"
```
bflad: The `sidebar_current` frontmatter has been deprecated and is no longer necessary (re-running CI will now return an error if this is still present)

website/docs/r/traffic_mirror_filter.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+Provides an Traffic mirror filter.  
+Read [limits and considerations](https://docs.aws.amazon.com/vpc/latest/mirroring/traffic-mirroring-considerations.html) for traffic mirroring
+
+## Example Usage
+
+To create a basic traffic mirror filter
+
+```hcl
+resource "aws_traffic_mirror_filter" "foo" {
+  description                                 = "traffic mirror filter - terraform example"
```
bflad: Nit: Excessive whitespace to the left of the equals signs here and below.

johnthedev97:
Thanks @bflad I will look into your comments.

aws/provider.go
```diff
@@ -767,6 +767,10 @@ func Provider() terraform.ResourceProvider {
 			"aws_default_subnet":                                      resourceAwsDefaultSubnet(),
 			"aws_subnet":                                              resourceAwsSubnet(),
 			"aws_swf_domain":                                          resourceAwsSwfDomain(),
+			"aws_traffic_mirror_filter":                               resourceAwsTrafficMirrorFilter(),
+			"aws_traffic_mirror_filter_rule":                          resourceAwsTrafficMirrorFilterRule(),
+			"aws_traffic_mirror_target":                               resourceAwsTrafficMirrorTarget(),
+			"aws_traffic_mirror_session":                              resourceAwsTrafficMirrorSession(),
```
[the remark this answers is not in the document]
johnthedev97: done

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,370 @@
...
+	filterId, filterIdSet := d.GetOk("traffic_mirror_filter_id")
+	input := &ec2.DescribeTrafficMirrorFiltersInput{}
+	if filterIdSet {
+		input.TrafficMirrorFilterIds = []*string{aws.String(filterId.(string))}
+		awsMutexKV.Lock(filterId.(string))
+		defer awsMutexKV.Unlock(filterId.(string))
+	} else {
+		awsMutexKV.Lock("DescribeTrafficMirrorFilters")
+		defer awsMutexKV.Unlock("DescribeTrafficMirrorFilters")
+	}
```
bflad: `traffic_mirror_filter_id` is always set since its required, however this logic to use mutexes during read is very uncommon, can you explain why its included here (potentially adding a code comment)? If its not necessary, then the logic can be simplified:

```go
	input := &ec2.DescribeTrafficMirrorFiltersInput{
		TrafficMirrorFilterIds: aws.StringSlice([]string{d.Get("traffic_mirror_filter_id").(string)}),
	}
```

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,370 @@
...
+			} else {
+				input.NextToken = out.NextToken
+			}
+		}
+	}
+
+	if !found {
+		d.SetId("")
+		return fmt.Errorf("Rule %s not found", ruleId)
+	}
```
bflad: The AWS Go SDK provides a paginated function ([`DescribeTrafficMirrorFiltersPages`](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeTrafficMirrorFiltersPages)) so we do not need to manually handle `NextToken` logic.

However, usually this type of logic is unnecessary. The API should be returning all children rules of the EC2 Traffic Mirror Filter in the response including the filter, not returning multiple responses of the same filter. I would suggest simplifying this logic to a single `DescribeTrafficMirrorFilters` call. We should also return `nil` when triggering resource recreation, rather than returning an error.

```go
out, err := conn.DescribeTrafficMirrorFilters(input)

if err != nil {
	return fmt.Errorf("error describing EC2 Traffic Mirror Filter (%s): %w", d.Get("traffic_mirror_filter_id").(string))
}

if out == nil || len(out.TrafficMirrorFilters) == 0 {
	log.Printf("[WARN] EC2 Traffic Mirror Filter (%s) not found, removing from state")
	d.SetId("")
	return nil
}

rule := findEc2TrafficMirrorFilterRule(ruleId, out.TrafficMirrorFilters)

if rule == nil {
	log.Printf("[WARN] EC2 Traffic Mirror Filter Rule (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}
```

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,370 @@
...
+
+	return populateTrafficMirrorFilterRuleResource(d, rule)
```
bflad: Since this function is only used once here, let's remove the indirection and include its `d.Set()`, etc. logic directly here. 👍

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func findRuleInFilters(ruleId string, filters []*ec2.TrafficMirrorFilter) (filter *ec2.TrafficMirrorFilter, rule *ec2.TrafficMirrorFilterRule, err error) {
```
bflad: Since no implementations use the filter return value and any returned error in this logic is the same as the rule being `nil`, suggest removing both of those function return values. Also would update naming since this function is global to the whole `aws` Go package (currently).

```
func findEc2TrafficMirrorFilterRule(ruleId string, filters []*ec2.TrafficMirrorFilter) *ec2.TrafficMirrorFilterRule {
```

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	var removeFields []*string
+	if d.HasChange("protocol") {
+		_, n := d.GetChange("protocol")
+		if n == "0" {
+			removeFields = append(removeFields, aws.String("protocol"))
+		} else {
+			input.Protocol = aws.Int64(int64(d.Get("protocol").(int)))
+		}
```
bflad: Only using the second return value of `d.GetChange()` is equivalent to `d.Get()`, so would recommend the following (along with AWS Go SDK constants) here and below in the rest of the Update functions:

```go
		v := d.Get("protocol").(int)
		if v == 0 {
			removeFields = append(removeFields, aws.String(ec2.TrafficMirrorFilterRuleFieldProtocol))
		} else {
			input.Protocol = aws.Int64(int64(v))
		}
```

See also: https://github.com/bflad/tfproviderlint/issues/44

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,370 @@
...
+	d.Set("traffic_mirror_filter_id", rule.TrafficMirrorFilterId)
+	d.Set("destination_cidr_block", rule.DestinationCidrBlock)
+	d.Set("source_cidr_block", rule.SourceCidrBlock)
+	d.Set("rule_action", rule.RuleAction)
+	d.Set("rule_number", rule.RuleNumber)
+	d.Set("traffic_direction", rule.TrafficDirection)
+	d.Set("description", rule.Description)
+	d.Set("protocol", rule.Protocol)
+	d.Set("destination_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule.DestinationPortRange))
+	d.Set("source_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule.SourcePortRange))
```
bflad: When using `d.Set()` with aggregate types (TypeList, TypeSet, TypeMap), [we should perform error checking to prevent issues where the code is not properly able to set the Terraform state](https://www.terraform.io/docs/extend/best-practices/detecting-drift.html#error-checking-aggregate-types). e.g.

```go
if err := d.Set("destination_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule.DestinationPortRange)); err != nil {
	return fmt.Errorf("error setting destination_port_range: %s", err)
}
if err := d.Set("source_port_range", buildTrafficMirrorFilterRulePortRangeSchema(rule. SourcePortRange)); err != nil {
	return fmt.Errorf("error setting source_port_range: %s", err)
}
```

aws/resource_aws_ec2_traffic_mirror_filter_rule_test.go
```diff
@@ -0,0 +1,249 @@
...
+		ruleId := rs.Primary.ID
+		filterId := rs.Primary.Attributes["traffic_mirror_filter_id"]
+
+		out, err := conn.DescribeTrafficMirrorFilters(&ec2.DescribeTrafficMirrorFiltersInput{
+			TrafficMirrorFilterIds: []*string{
+				aws.String(filterId),
+			},
+		})
+
+		if isAWSErr(err, "InvalidTrafficMirrorFilterId.NotFound", "") {
```
bflad: If this is the error that is returned when an EC2 Traffic Mirror Filter cannot be found, we should include this in the filter and filter rule resource `Read` functions to trigger Terraform resource recreation:

```go
// In the resource Read functions, potentially in addition to over nil/length checks from DescribeTrafficMirrorFilters call
if isAWSErr(err, "InvalidTrafficMirrorFilterId.NotFound", "") {
	log.Printf("[WARN] EC2 Traffic Mirror Filter (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}

if err != nil {
	fmt.Errorf("error describing EC2 Traffic Mirror Filter (%s): %w", d.Id(), err)
}
```

aws/resource_aws_ec2_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+		TrafficMirrorSessionIds: []*string{
+			&sessionId,
+		},
+	}
+
+	out, err := conn.DescribeTrafficMirrorSessions(input)
+	if nil != err {
+		d.SetId("")
+		return fmt.Errorf("Error describing traffic mirror session %v", sessionId)
+	}
```
bflad: We should not remove this resource from the Terraform state on API errors and should return the error itself.

```go
if err != nil {
	return fmt.Errorf("error describing EC2 Traffic Mirror Session (%s): %w", d.Id(), err)
}
```

aws/resource_aws_ec2_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if 0 == len(out.TrafficMirrorSessions) {
+		d.SetId("")
+		return fmt.Errorf("Unable to find traffic mirrir session %v", sessionId)
+	}
```
bflad: When triggering resource recreation we should prefer a warning log and return no error.

```go
if out == nil || len(out.TrafficMirrorSessions) == 0 {
	log.Printf("[WARN] EC2 Traffic Mirror Session (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}
```

aws/resource_aws_ec2_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+	d.Set("traffic_mirror_target_id", session.TrafficMirrorTargetId)
+	d.Set("description", session.Description)
+	d.Set("packet_length", session.PacketLength)
+
+	// AWS returns a value for VNI even if nothing is set in config. To avoid further plan
+	// showing a change, using a workaround - set this in state only if it was set in config
+	_, vniSet := d.GetOk("virtual_network_id")
+	if vniSet && nil != session.VirtualNetworkId {
+		d.Set("virtual_network_id", *session.VirtualNetworkId)
+	}
```
bflad: We should always prefer to call `d.Set()` in Read functions -- `d.Get()` and `d.GetOk()` logic does not work during import. If this configuration is optional and is defaulted by the API in that case, then setting `Computed: true` on the schema attribute is more appropriate.

```go
			"virtual_network_id": {
				Type:         schema.TypeInt,
				Optional:     true,
				Computed:     true,
				ValidateFunc: validation.IntBetween(1, 16777216),
			},
```

And the logic here reduced to just:

```go
d.Set("virtual_network_id", session.VirtualNetworkId)
```

aws/resource_aws_ec2_traffic_mirror_session.go
```diff
@@ -0,0 +1,212 @@
...
+	input := &ec2.DeleteTrafficMirrorSessionInput{
+		TrafficMirrorSessionId: &sessionId,
+	}
+
+	_, err := conn.DeleteTrafficMirrorSession(input)
+	if nil != err {
+		return fmt.Errorf("Error deleting traffic mirror session %v", sessionId)
+	}
+
+	d.SetId("")
```
bflad: Nit: `d.SetId("")` is extraneous is `Delete` functions:

```suggestion
```

aws/resource_aws_ec2_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+				Optional: true,
+				ForceNew: true,
+			},
+			"network_interface_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ConflictsWith: []string{
+					"network_load_balancer_arn",
+				},
```
bflad: If the API requires one or the other of these two attributes, then we can use `ExactlyOneOf` instead of `ConflictsWith` to have Terraform return an error if both are missing or both are declared. 👍 

```go
				ExactlyOneOf: []string{
					"network_interface_id",
					"network_load_balancer_arn",
				},
```

aws/resource_aws_ec2_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+			},
+			"network_load_balancer_arn": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				ConflictsWith: []string{
+					"network_interface_id",
+				},
```
bflad:
```go
				ExactlyOneOf: []string{
					"network_interface_id",
					"network_load_balancer_arn",
				},
```

aws/resource_aws_ec2_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+	targetId := d.Id()
+	input := &ec2.DescribeTrafficMirrorTargetsInput{
+		TrafficMirrorTargetIds: []*string{&targetId},
+	}
+
+	out, err := conn.DescribeTrafficMirrorTargets(input)
+	if err != nil {
+		d.SetId("")
+		return fmt.Errorf("Error describing traffic mirror target %v", targetId)
+	}
```
bflad: Same note here -- we should not remove this resource from the Terraform state on API errors and should return the error itself.

```go
if err != nil {
	return fmt.Errorf("error describing EC2 Traffic Mirror Target (%s): %w", d.Id(), err)
}
```

aws/resource_aws_ec2_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if 1 != len(out.TrafficMirrorTargets) {
+		d.SetId("")
+		return fmt.Errorf("Not able to find find target %v", targetId)
+	}
```
bflad: Same note here -- when triggering resource recreation we should prefer a warning log and return no error.

```go
if out == nil || len(out.TrafficMirrorTargets) == 0 {
	log.Printf("[WARN] EC2 Traffic Mirror Target (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}
```

aws/resource_aws_ec2_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+	if nil != target.Description {
+		d.Set("description", *target.Description)
+	}
+
+	switch *target.Type {
+	case "network-interface":
+		d.Set("network_interface_id", target.NetworkInterfaceId)
+	case "network-load-balancer":
+		d.Set("network_load_balancer_arn", target.NetworkLoadBalancerArn)
+	}
```
bflad: Same notes here about always calling `d.Set()` so Terraform can perform drift detection, `d.Set()` will automatically handle `nil` values.

```go
d.Set("description", target.Description)
d.Set("network_interface_id", target.NetworkInterfaceId)
d.Set("network_load_balancer_arn", target.NetworkLoadBalancerArn)
```

aws/resource_aws_ec2_traffic_mirror_target.go
```diff
@@ -0,0 +1,120 @@
...
+	input := &ec2.DeleteTrafficMirrorTargetInput{
+		TrafficMirrorTargetId: &targetId,
+	}
+
+	_, err := conn.DeleteTrafficMirrorTarget(input)
+	if nil != err {
+		return fmt.Errorf("Error deleting traffic mirror target %v", targetId)
+	}
+
+	d.SetId("")
```
bflad: Nit: `d.SetId("")` is extraneous in `Delete` functions

```suggestion
```

aws/resource_aws_ec2_traffic_mirror_target_test.go
```diff
@@ -0,0 +1,242 @@
...
+			continue
+		}
+
+		out, err := conn.DescribeTrafficMirrorTargets(&ec2.DescribeTrafficMirrorTargetsInput{
+			TrafficMirrorTargetIds: []*string{
+				aws.String(rs.Primary.ID),
+			},
+		})
+
+		if isAWSErr(err, "InvalidTrafficMirrorTargetId.NotFound", "") {
```
bflad: Same note here -- if this is the error that is returned when an EC2 Traffic Mirror Target cannot be found, we should include this target resource `Read` functions to trigger Terraform resource recreation:

```go
// In the resource Read functions, potentially in addition to over nil/length checks from DescribeTrafficMirrorTargets call
if isAWSErr(err, "InvalidTrafficMirrorTargetId.NotFound", "") {
	log.Printf("[WARN] EC2 Traffic Mirror Target (%s) not found, removing from state", d.Id())
	d.SetId("")
	return nil
}

if err != nil {
	fmt.Errorf("error describing EC2 Traffic Mirror Target (%s): %w", d.Id(), err)
}
```

website/docs/r/ec2_traffic_mirror_filter.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+Read [limits and considerations](https://docs.aws.amazon.com/vpc/latest/mirroring/traffic-mirroring-considerations.html) for traffic mirroring
+
+## Example Usage
+
+To create a basic traffic mirror filter
+
+```hcl
+resource "aws_ec2_traffic_mirror_filter" "foo" {
+  description                                 = "traffic mirror filter - terraform example"
+  network_services                            = ["amazon-dns"]
```
bflad: Nit: Excessive whitespace to the left of the equals signs here and in below similar example configurations.

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
johnthedev97: I believe in case of an import the read would be called without an filterId which means we would need to scan all the rules of all filters to find the proper rule. This is the reason I looped through, but ofcourse I should use the paginator.

bflad: In that case, it might be best if the Terraform resource import ID includes both the filter ID and the filter rule ID. Otherwise the looping logic should occur in the `State` import function and leave the `Read` function simple. 👍

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
johnthedev97: What about during import?

bflad: If import requires special logic, would suggest moving to the import function. Otherwise, the more preferred solution is to have all necessary identifiers in the resource import ID.

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
johnthedev97: Can you point me to an example using state import function?

bflad: Example state import function, with multi-part import ID handling:

https://github.com/terraform-providers/terraform-provider-aws/blob/5cffb1d4d654d55f1d4429dfc417c1eed949e925/aws/resource_aws_api_gateway_authorizer.go#L24-L36

johnthedev97:
@bflad I believe I have fixed all your comments.

aws/resource_aws_ec2_traffic_mirror_session.go
```diff
@@ -0,0 +1,206 @@
...
+	conn := meta.(*AWSClient).ec2conn
+
+	sessionId := d.Id()
+	input := &ec2.DeleteTrafficMirrorSessionInput{
+		TrafficMirrorSessionId: &sessionId,
+	}
+
+	_, err := conn.DeleteTrafficMirrorSession(input)
+	if nil != err {
+		return fmt.Errorf("Error deleting traffic mirror session %v", sessionId)
```
bflad: This is not returning the API/SDK error for operators and code maintainers to troubleshoot the issue:

```suggestion
		return fmt.Errorf("error deleting EC2 Traffic Mirror Session (%s): %w", sessionId, err)
```

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,356 @@
...
+	conn := meta.(*AWSClient).ec2conn
+
+	ruleId := d.Id()
+	input := &ec2.DeleteTrafficMirrorFilterRuleInput{
+		TrafficMirrorFilterRuleId: &ruleId,
+	}
+
+	_, err := conn.DeleteTrafficMirrorFilterRule(input)
+	if err != nil {
+		return fmt.Errorf("Error deleting traffic mirror filter rule %v", ruleId)
```
bflad: This is not returning the API/SDK error for operators and code maintainers to troubleshoot the issue:

```suggestion
		return fmt.Errorf("error deleting EC2 Traffic Mirror Filter Rule (%s): %w", ruleId, err)
```

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,356 @@
...
+		input.TrafficDirection = aws.String(d.Get("traffic_direction").(string))
+	}
+
+	if len(removeFields) > 0 {
+		input.SetRemoveFields(removeFields)
+	}
+
+	_, err := conn.ModifyTrafficMirrorFilterRule(input)
+	if err != nil {
+		return fmt.Errorf("Error modifying rule %v", ruleId)
```
bflad: This is not returning the API/SDK error for operators and code maintainers to troubleshoot the issue:

```suggestion
		return fmt.Errorf("error modifying EC2 Traffic Mirror Filter Rule (%s): %w", ruleId, err)
```

aws/resource_aws_ec2_traffic_mirror_filter_rule.go
```diff
@@ -0,0 +1,356 @@
...
+		input.DestinationPortRange = buildTrafficMirrorPortRangeRequest(v.([]interface{}))
+	}
+
+	if v, ok := d.GetOk("source_port_range"); ok {
+		input.SourcePortRange = buildTrafficMirrorPortRangeRequest(v.([]interface{}))
+	}
+
+	out, err := conn.CreateTrafficMirrorFilterRule(input)
+	if err != nil {
+		return fmt.Errorf("Error creating traffic mirror filter rule for %v", filterId)
```
bflad: This is not returning the API/SDK error for operators and code maintainers to troubleshoot the issue:

```suggestion
		return fmt.Errorf("error creating EC2 Traffic Mirror Filter Rule (%s): %w", filterId, err)
```

aws/resource_aws_ec2_traffic_mirror_filter.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		removeServices := o.(*schema.Set).Difference(n.(*schema.Set)).List()
+		if len(removeServices) > 0 {
+			input.RemoveNetworkServices = expandStringList(removeServices)
+		}
+
+		_, err := conn.ModifyTrafficMirrorFilterNetworkServices(input)
+		if err != nil {
+			return fmt.Errorf("Error modifying network services for traffic mirror filter %v", d.Id())
```
bflad: This is not returning the API/SDK error for operators and code maintainers to troubleshoot the issue:

```suggestion
			return fmt.Errorf("error modifying EC2 Traffic Mirror Filter (%s) network services: %w", d.Id(), err)
```

aws/resource_aws_ec2_traffic_mirror_session_test.go
```diff
@@ -0,0 +1,244 @@
...
+					resource.TestCheckResourceAttr(resourceName, "virtual_network_id", "0"),
+				),
+			},
+			// import test without VNI
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"virtual_network_id"},
+			},
```
bflad: Now that `virtual_network_id` is always set in the Terraform state, the following allows this test to pass:

```go
			//create
			{
				Config: testAccTrafficMirrorSessionConfig(lbName, session),
					testAccCheckAWSEc2TrafficMirrorSessionExists(resourceName),
					resource.TestCheckResourceAttr(resourceName, "description", ""),
					resource.TestCheckResourceAttr(resourceName, "packet_length", "0"),
					resource.TestCheckResourceAttr(resourceName, "session_number", strconv.Itoa(session)),
					resource.TestMatchResourceAttr(resourceName, "virtual_network_id", regexp.MustCompile(`\d+`)),
				),
			},
			// update of description, packet length and VNI
			{
				Config: testAccTrafficMirrorSessionConfigWithOptionals(description, lbName, session, pLen, vni),
					testAccCheckAWSEc2TrafficMirrorSessionExists(resourceName),
					resource.TestCheckResourceAttr(resourceName, "description", description),
					resource.TestCheckResourceAttr(resourceName, "packet_length", strconv.Itoa(pLen)),
					resource.TestCheckResourceAttr(resourceName, "session_number", strconv.Itoa(session)),
					resource.TestCheckResourceAttr(resourceName, "virtual_network_id", strconv.Itoa(vni)),
				),
			},
			// removal of description, packet length and VNI
			{
				Config: testAccTrafficMirrorSessionConfig(lbName, session),
					testAccCheckAWSEc2TrafficMirrorSessionExists(resourceName),
					resource.TestCheckResourceAttr(resourceName, "description", ""),
					resource.TestCheckResourceAttr(resourceName, "packet_length", "0"),
					resource.TestCheckResourceAttr(resourceName, "session_number", strconv.Itoa(session)),
					resource.TestMatchResourceAttr(resourceName, "virtual_network_id", regexp.MustCompile(`\d+`)),
				),
			},
			// import test without VNI
			{
				ImportState:       true,
				ImportStateVerify: true,
			},
```

Will handle on merge. 👍 

```
--- PASS: TestAccAWSEc2TrafficMirrorSession_basic (282.32s)
```

johnthedev97:
Thanks @bflad for all the valuable review comments, pointers and the last set of fixes :-)

PR #9390, opened 2019-07-17
data-source/aws_route53_zone: add linked service attributes

jstewmon:
- updated test error strings to better reflect actual vs expected
- fixed `name_servers` check to test the data source, not the resource
- renamed the resource attributes variable and added a data source
  attributes varialbe to improve readability

### Community Note

```release-note
data-source/aws_route53_zone: new attributes `linked_service_principal` and `linked_service_description`
```

```
$ make testacc TESTARGS='-run=TestAccDataSourceAwsRoute53Zone'
=== RUN   TestAccDataSourceAwsRoute53Zone
=== PAUSE TestAccDataSourceAwsRoute53Zone
=== CONT  TestAccDataSourceAwsRoute53Zone
--- PASS: TestAccDataSourceAwsRoute53Zone (128.46s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	128.504s
...
```

### DIFF
--- a/aws/data_source_aws_route53_zone.go
+++ b/aws/data_source_aws_route53_zone.go
@@ -32,12 +32,10 @@ func dataSourceAwsRoute53Zone() *schema.Resource {
 			},
 			"comment": {
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 			"caller_reference": {
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 			"vpc_id": {
@@ -54,6 +52,14 @@ func dataSourceAwsRoute53Zone() *schema.Resource {
 			"name_servers": {
 				Type:     schema.TypeList,
 				Elem:     &schema.Schema{Type: schema.TypeString},
+				Computed: true,
+			},
+			"linked_service_principal": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"linked_service_description": {
+				Type:     schema.TypeString,
 				Computed: true,
 			},
 		},
@@ -169,6 +175,10 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("private_zone", hostedZoneFound.Config.PrivateZone)
 	d.Set("caller_reference", hostedZoneFound.CallerReference)
 	d.Set("resource_record_set_count", hostedZoneFound.ResourceRecordSetCount)
+	if hostedZoneFound.LinkedService != nil {
+		d.Set("linked_service_principal", hostedZoneFound.LinkedService.ServicePrincipal)
+		d.Set("linked_service_description", hostedZoneFound.LinkedService.Description)
+	}
 
 	nameServers, err := hostedZoneNameServers(idHostedZone, conn)
 	if err != nil {
--- a/aws/data_source_aws_route53_zone_test.go
+++ b/aws/data_source_aws_route53_zone_test.go
@@ -2,19 +2,18 @@ package aws
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/acctest"
 	"github.com/hashicorp/terraform/helper/resource"
-	"github.com/hashicorp/terraform/terraform"
 )
 
 func TestAccDataSourceAwsRoute53Zone(t *testing.T) {
 	rInt := acctest.RandInt()
 	publicResourceName := "aws_route53_zone.test"
-	publicDomain := fmt.Sprintf("terraformtestacchz-%d.com.", rInt)
 	privateResourceName := "aws_route53_zone.test_private"
-	privateDomain := fmt.Sprintf("test.acc-%d.", rInt)
+	serviceDiscoveryResourceName := "aws_service_discovery_private_dns_namespace.test_service_discovery"
 
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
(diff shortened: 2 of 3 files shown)


aws/data_source_aws_route53_zone.go
```diff
@@ -56,6 +56,16 @@ func dataSourceAwsRoute53Zone() *schema.Resource {
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Computed: true,
 			},
+			"linked_service_principal": {
+				Type:     schema.TypeString,
+				Optional: true,
```
bflad: Since this attribute is never used as an input argument (e.g. for filtering/searching), it should not be marked `Optional: true` 👍 

```suggestion
```

aws/data_source_aws_route53_zone.go
```diff
@@ -56,6 +56,16 @@ func dataSourceAwsRoute53Zone() *schema.Resource {
...
+				Computed: true,
+			},
+			"linked_service_description": {
+				Type:     schema.TypeString,
+				Optional: true,
```
bflad: Since this attribute is never used as an input argument (e.g. for filtering/searching), it should not be marked `Optional: true` 👍 

```suggestion
```

aws/data_source_aws_route53_zone_test.go
```diff
@@ -25,13 +28,33 @@ func TestAccDataSourceAwsRoute53Zone(t *testing.T) {
...
+						"data.aws_route53_zone.by_vpc",
+						map[string]string{"name": privateDomain},
+					),
 					testAccDataSourceAwsRoute53ZoneCheck(
-						privateResourceName, "data.aws_route53_zone.by_tag", privateDomain),
+						privateResourceName,
+						"data.aws_route53_zone.by_tag",
+						map[string]string{"name": privateDomain},
+					),
+					testAccDataSourceAwsRoute53ZoneCheck(
```
bflad: Since this acceptance testing was originally written, the Terraform Provider SDK provides a much easier way to verify state attribute values match between two resources via [`resource.TestCheckResourceAttrPair()`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#TestCheckResourceAttrPair) and values within a single resource via [`resource.TestCheckResourceAttr()`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#TestCheckResourceAttr) and [`resource.TestMatchResourceAttr()`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#TestMatchResourceAttr).

Instead of changing the `testAccDataSourceAwsRoute53ZoneCheck` function to further increase its complexity, we would prefer to start improving the testing to use the simpler functions instead of the these legacy methods if possible.

e.g. in this case:

```go
resource.TestCheckResourceAttrPair("data.aws_route53_zone.service_discovery_by_vpc", "name", serviceDiscoveryResourceName, "name"),
resource.TestCheckResourceAttr("data.aws_route53_zone.service_discovery_by_vpc", "linked_service_principal", "servicediscovery.amazonaws.com"),
// if the value matches an attribute value in the resource, use TestCheckResourceAttrPair() instead of the below
resource.TestMatchResourceAttr("data.aws_route53_zone.service_discovery_by_vpc", "linked_service_description", regexp.MustCompile(`^arn:[^:]+:servicediscovery:[^:]+:[^:]+:namespace/ns-\\w+$`)),
```

aws/data_source_aws_route53_zone_test.go
```diff
@@ -121,5 +162,15 @@ data "aws_route53_zone" "by_zone_id" {
...
+resource "aws_service_discovery_private_dns_namespace" "test_service_discovery" {
+  name        = "test.acc-sd-%d"
+  vpc         = "${aws_vpc.test.id}"
+}
+
+data "aws_route53_zone" "service_discovery_by_vpc" {
+  name   = "${aws_service_discovery_private_dns_namespace.test_service_discovery.name}"
+  vpc_id = "${aws_vpc.test.id}"
+}
+`, rInt, rInt, rInt, rInt, rInt)
```
bflad: Instead of repeatedly defining the same `fmt.Sprintf()` arguments, the format verbs support a 1-based index argument (e.g. `%[1]d` instead of `%d`) and this list can be simplified to a single `rInt` argument.

After replacing all `%d` above with `%[1]d`:

```suggestion
`, rInt)
```

jstewmon:
@bflad thanks for the excellent feedback. I've incorporated your suggestions and completely replaced the old test check function with the resource helpers.

aws/data_source_aws_route53_zone.go
```diff
@@ -32,12 +32,10 @@ func dataSourceAwsRoute53Zone() *schema.Resource {
 			},
 			"comment": {
 				Type:     schema.TypeString,
-				Optional: true,
```
bflad: 🙇 thank you

aws/data_source_aws_route53_zone_test.go
```diff
@@ -24,52 +23,26 @@ func TestAccDataSourceAwsRoute53Zone(t *testing.T) {
...
 				),
 			},
 		},
 	})
 }
 
-// rsName for the name of the created resource
-// dsName for the name of the created data source
-// zName for the name of the domain
-func testAccDataSourceAwsRoute53ZoneCheck(rsName, dsName, zName string) resource.TestCheckFunc {
```
bflad: ❤️ thank you!

nywilken:
The enhancement to the `aws_route53_zone` data source, has been released in [version 2.20.0 of the Terraform AWS provider](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md#2200-july-19-2019). Please see the [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions) or reach out if you need any assistance upgrading.

For further feature requests or bug reports with this resource, please create a [new GitHub issue](https://github.com/terraform-providers/terraform-provider-aws/issues/new/choose) following the template for triage. Thanks!

Issue #5197, opened 2018-07-14
After import aws_autoscaling_group  terraform plan doesn't show load_balancers diff


savankumargudaas:
Hi!

Let me give a background

I imported ASG against my local terraform resource file. ASG in AWS have couple of ELB(classic) association. I can see that imported state contains ELB related information. 

When I run plan, plan doesn't show any difference of load_balancers. 

```
asg.tf
resource "aws_autoscaling_group" "existing-asg" {
...//without load_balancers value
}

```

`terraform import aws_autoscaling_group.existing-asg existing-asg-name`

```
terraform state show aws_autoscaling_group.existing-asg
...
load_balancers.#                   = 2
load_balancers.1433097967          = first-elb-name
load_balancers.653101771           = second-elb-name
....

```
```
terraform plan
 ~ aws_autoscaling_group.existing-asg
      force_delete:              "" => "false"
```

Why the plan not showing like below?
```
 ~ aws_autoscaling_group.existing-asg
      force_delete:              "" => "false"
      load_balancers.#        2 => 0
      load_balancers.1433097967     "first-elb-name" => ""
      load_balancers.653101771        "second-elb-name" => ""
```

I just wonder, is it a bug? Or it's intended behaviour, why so?

MetricMike:
I'm having a similar issue migrating an `aws_autoscaling_group` from ELB/`load_balancers` to ALB/`target_group_arns`

going from:

```hcl
resource "aws_autoscaling_group" {
    ...
    load_balancers = ["${split(",", var.asg_load_balancer_names)}"]
    ...
}
```

to:

```hcl
resource "aws_autoscaling_group" {
    ...
    target_group_arns = ["${split(",", var.asg_target_group_arns)}"]
    ...
}
```

generates the following diff:

```
  ~ module.auto_scaling_groups.aws_autoscaling_group.default
      target_group_arns.#:                                                                "" => <computed>
```

This causes issue when autoscaling, as the ASG still references an ELB that might have been deleted. Specifying `load_balancers = []` produces the following diff:

```
  ~ module.auto_scaling_groups.aws_autoscaling_group.default
      load_balancers.#:                                                                   "1" => "0"
      load_balancers.3481940168:                                                          "sample-elb" => ""
      target_group_arns.#:                                                                "" => <computed>
```

[...]

PR #9478, opened 2019-07-24
resource/aws_autoscaling_group: Fix issue preventing the removal of `load_balancers` when switching to `target_group_arns`

nywilken:
### Community Note

Closes: #5197

```release-note
* resource/aws_autoscaling_group: Fix issue preventing the removal of `load_balancers` when switching to `target_group_arns`

```

Upon configuring an ASG with load_balancers, it was found that swapping the load_balancers argument with a list of target_group_arns would result in no diff for the load_balancers despite them being removed.This issue being the computed property on `load_balancers`, which refreshes the configuration at plan time resulting in `load_balancers` being added back to the configuration as they are still part of the referenced infrastructure. This change removes the `Computed` property on both `load_balancers` and `target_group_arns` to ensure Terraform always honors the arguments within a provided configuration.

```
$ make testacc TESTARGS='-run=TestAccAWSAutoScalingGroup_WithLoadBalancer'
--- PASS: TestAccAWSAutoScalingGroup_WithLoadBalancer (390.68s)
--- PASS: TestAccAWSAutoScalingGroup_WithLoadBalancer_ToTargetGroup (468.74s)
```

### DIFF
--- a/aws/resource_aws_autoscaling_group.go
+++ b/aws/resource_aws_autoscaling_group.go
@@ -258,7 +258,6 @@ func resourceAwsAutoscalingGroup() *schema.Resource {
 			"load_balancers": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
 			},
@@ -330,7 +329,6 @@ func resourceAwsAutoscalingGroup() *schema.Resource {
 			"target_group_arns": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Set:      schema.HashString,
 			},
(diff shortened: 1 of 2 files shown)


aws/resource_aws_autoscaling_group.go
```diff
@@ -78,6 +78,7 @@ func resourceAwsAutoscalingGroup() *schema.Resource {
 						"version": {
 							Type:         schema.TypeString,
 							Optional:     true,
+							Default:      "$$Default",
```
bflad: I believe this fix should be the in test configuration and not in the resource. 😉 If I recall its a difference in Terraform 0.12's configuration handling regarding `$` escaping.

nywilken: I see so within the testing configuration we should not escape `$` and instead render the configuration such as `Version: "$Latest"`. Is that correct. I see that the mixed_instance_policy.launch_template.version` has an attribute of `Default: "$$Deault"` should that be removed at some point. Not within this PR.

nywilken: I broke this into a separate PR https://github.com/terraform-providers/terraform-provider-aws/pull/9488

bflad:
Please see https://github.com/terraform-providers/terraform-provider-aws/issues/9513 for the unexpected behavior change with this pull request. I'm going to lock this pull request only to encourage following the new issue.

PR #9537, opened 2019-07-29
Remove ambiguous validation on S3 bucket name while creating bucket

hansrajdas:
* This is validating on string length between 0 to 63 and bucket name
can have length from 3 to 63 so this validation is ambiguous
* Validation on bucket name will be done by validateS3BucketName while
creating bucket

### Community Note

```release-note
NONE
```

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -39,7 +39,6 @@ func resourceAwsS3Bucket() *schema.Resource {
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"bucket_prefix"},
-				ValidateFunc:  validation.StringLenBetween(0, 63),
 			},
 			"bucket_prefix": {
 				Type:          schema.TypeString,


bflad:
Hi @hansrajdas 👋  Thank you for submitting this!

The reason for this ambiguous plan-time validation of zero length is that the empty string (`""`) value needs to be supported for Terraform 0.11 and earlier since it has no concept of `null`, which this provider will be supporting until the next major version (version 3.0.0) is released. We would also want to ensure we catch longer than 63 character naming at plan-time when possible to keep that existing functionality, so wholly removing the validation does not seem appropriate. While we could tailor this to only accept either an empty string value or 3-63 characters, the messaging in that case can also be ambiguous/confusing. So rather than changing this, we are going to prefer to wait until https://github.com/terraform-providers/terraform-provider-aws/issues/13943 is potentially addressed.

If you would like to discuss this further, please create a [feature request issue](https://github.com/terraform-providers/terraform-provider-aws/issues/new?labels=enhancement&template=Feature_Request.md) or comment in #13943. 👍

PR #9570, opened 2019-07-31
provider: Use real Terraform version in UA header

radeksimko:
The existing implementation leverages Terraform's `version` package as Terraform currently provides the SDK for interacting with core and is therefore imported by the provider, including the `version` package.

The side effect of this is that provider reports version of vendored Terraform (i.e. plugin SDK), not the version which actually talks to the provider's gRPC server and parses the configuration.

This PR is leveraging recent changes introduced in 0.12 to fix that and report accurate version.

----

I also wanted to change the header from the following

```
APN/1.0 HashiCorp/1.0 Terraform/X.Y.Z (+https://www.terraform.io)
```
to
```
APN/1.0 HashiCorp Terraform/X.Y.Z (+https://www.terraform.io)
```
to align it better with other providers.

The point of versioning of `HashiCorp` was (IMO rightfully) questioned during the implementation of https://github.com/hashicorp/terraform/pull/22272 and the version was dropped from that helper function.

However the upstream AWS SDK helpers used here for building the UA string may need to be tuned to reflect [the RFC](https://tools.ietf.org/html/rfc7231#section-5.5.3) where version of products is optional. Dropping the `Version` part would otherwise produce

```
APN/1.0 HashiCorp/ Terraform/X.Y.Z (+https://www.terraform.io)
```
which is most likely undesired.

An example of a full UA string with this patch follows
```
aws-sdk-go/1.23.0 (go1.12.7; darwin; amd64) APN/1.0 HashiCorp/1.0 Terraform/0.12.7 (+https://www.terraform.io)
```
I think it would be worth exploring ways of moving SDK/Go versions towards the end, to reflect [RFC7231](https://tools.ietf.org/html/rfc7231#section-5.5.3), specifically this part:

> By convention, the product identifiers are listed in decreasing order of their significance for identifying the user agent software.  Each product identifier consists of a name and optional version.

IMO Terraform is more significant in this context than the SDK - but I will leave that for another discussion/PR.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -137,7 +137,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/xray"
 	awsbase "github.com/hashicorp/aws-sdk-go-base"
 	"github.com/hashicorp/terraform/helper/logging"
-	"github.com/hashicorp/terraform/terraform"
 )
 
 type Config struct {
@@ -166,6 +165,8 @@ type Config struct {
 	SkipRequestingAccountId bool
 	SkipMetadataApiCheck    bool
 	S3ForcePathStyle        bool
+
+	terraformVersion string
 }
 
 type AWSClient struct {
@@ -335,7 +336,8 @@ func (c *Config) Client() (interface{}, error) {
 		UserAgentProducts: []*awsbase.UserAgentProduct{
 			{Name: "APN", Version: "1.0"},
 			{Name: "HashiCorp", Version: "1.0"},
-			{Name: "Terraform", Version: terraform.VersionString()},
+			{Name: "Terraform", Version: c.terraformVersion,
+				Extra: []string{"+https://www.terraform.io"}},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


bflad:
Hey @radeksimko 👋 I think this is looking okay -- can you please also fix the other usage of `VersionString()` at `aws/resource_aws_opsworks_stack.go:390`?

```console
$ tf-sdk-migrator check
...
Checking whether provider uses deprecated SDK packages or identifiers...
Deprecated SDK packages in use: [github.com/hashicorp/terraform/flatmap]
Deprecated SDK identifiers in use: [Ident VersionString from package github.com/hashicorp/terraform/terraform is used at [/Users/bflad/src/github.com/terraform-providers/terraform-provider-aws/aws/config.go:338:43 /Users/bflad/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_opsworks_stack.go:390:110]]

Some constraints not satisfied. Please resolve these before migrating to the new SDK.
```

Thanks!

radeksimko:
Yeah, I was looking at it, but I wasn't sure what was the motivation there to have a separate resource-specific logic for handling sessions etc. - happy to look into it, although I hoped we could deal with it in a separate PR 😃

PR #9685, opened 2019-08-08
Fix update of imported ACM certificates

julienduchesne:
The update for the ACM certificates did not work for imported (external) certificates. The reason for that is that only the hash of the certificates was stored in the state. The API call requires all three parts (chain, body and private key) of a certificate when updating. 

Therefore, an update would only work if you changed all three parts, otherwise, it would try to send the hash instead of the actual certificate.

I tried to hack something together with `CustomizeDiff` but I couldn't manage to force Terraform to go re-read the source of the unchanged parts instead of getting them from the state when one of the three parts has changed.

The way I made it work was by putting the certificate in the state

```
TF_ACC=1 go test  github.com/terraform-providers/terraform-provider-aws/aws -run "^(TestAccAWSAcmCertificate_imported_DomainName)$" -v
=== RUN   TestAccAWSAcmCertificate_imported_DomainName
=== PAUSE TestAccAWSAcmCertificate_imported_DomainName
=== CONT  TestAccAWSAcmCertificate_imported_DomainName
--- PASS: TestAccAWSAcmCertificate_imported_DomainName (76.78s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       76.819s
```


aws/resource_aws_acm_certificate_test.go
```diff
@@ -539,20 +539,30 @@ func TestAccAWSAcmCertificate_imported_DomainName(t *testing.T) {
...
 				),
+				ExpectNonEmptyPlan: true, // The certificate body is regenerated every time
+			},
+			{
+				Config: testAccAcmCertificateConfig_selfSigned("example"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "tags.%", "0"),
+					resource.TestCheckResourceAttr(resourceName, "domain_name", "example.com"),
+				),
+				ExpectNonEmptyPlan: true, // The certificate body is regenerated every time
```
julienduchesne: The `tls_locally_signed_cert` is regenerated every time because it's a new request it seems

edahlseng:
Linking everything up to be pedantic, it looks like this should address #10847, and perhaps #9809.

Issue #7873, opened 2019-03-08
Implement isResourceTimeoutError() Conditionals After All resource.Retry() Usage


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When implementing [`resource.Retry()`](https://godoc.org/github.com/hashicorp/terraform/helper/resource#Retry) to allow Terraform AWS Provider resources to retry AWS API requests either for eventual consistency errors or for known concurrency errors, the code immediately following should include a conditional check for `isResourceTimeoutError()` to retry the request outside the timeout.

If the AWS Go SDK encounters an error such as 5XX AWS API responses, networking related errors, or certain throttling errors, it will automatically retry the AWS API requests itself without returning to the calling code. If these automatic AWS Go SDK retries last longer than the `resource.Retry()` timeout and we do not retry the request outside the `resource.Retry()` block before returning the error, operators will receive an "empty" timeout error without any context about the underlying AWS Go SDK error:

```
timeout while waiting for state to become 'success' (timeout: 1m0s)
```

Debugging these with Terraform is only possible by [enabling debug logging](https://www.terraform.io/docs/internals/debugging.html), e.g. `TF_LOG=debug terraform plan/apply`.

Example incorrect implementation in pseudocode:

```go
err := resource.Retry(1 * time.Minute, func() *resource.RetryError {
  _, err := conn.SomeRequest(input)
  // ... other logic including resource.RetryableError() ...
})

if err != nil {
  return err
}
```

Fixed implementation:

```go
err := resource.Retry(1 * time.Minute, func() *resource.RetryError {
  _, err := conn.SomeRequest(input)
  // ... other logic including resource.RetryableError() ...
})

if isResourceTimeoutError(err) {
  // Using = here intentionally to overwrite the err
  _, err = conn.SomeRequest(input)
}

if err != nil {
  return err
}
```

### References

There are numerous bugs and pull requests relating to this behavior, but most recently: #7871

ryndaniels:
HURRAY WE DID IT

PR #9779, opened 2019-08-15
Final retries after internet gateway timeouts

ryndaniels:
### Community Note

Relates #7873

```release-note
BUG FIXES:
* resource/aws_internet_gateway: Final retries after timeouts creating, attaching, and deleting gateways

```

```
$ make testacc TESTARGS="-run=TestAccAWSInternetGateway"
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSInternetGateway -timeout 120m
=== RUN   TestAccAWSInternetGateway_importBasic
=== PAUSE TestAccAWSInternetGateway_importBasic
=== RUN   TestAccAWSInternetGateway_basic
=== PAUSE TestAccAWSInternetGateway_basic
=== RUN   TestAccAWSInternetGateway_delete
=== PAUSE TestAccAWSInternetGateway_delete
=== RUN   TestAccAWSInternetGateway_tags
=== PAUSE TestAccAWSInternetGateway_tags
=== CONT  TestAccAWSInternetGateway_importBasic
=== CONT  TestAccAWSInternetGateway_delete
=== CONT  TestAccAWSInternetGateway_basic
=== CONT  TestAccAWSInternetGateway_tags
--- PASS: TestAccAWSInternetGateway_importBasic (66.69s)
--- PASS: TestAccAWSInternetGateway_delete (93.20s)
--- PASS: TestAccAWSInternetGateway_tags (98.69s)
--- PASS: TestAccAWSInternetGateway_basic (121.75s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       123.228s
```

### DIFF
--- a/aws/resource_aws_internet_gateway.go
+++ b/aws/resource_aws_internet_gateway.go
@@ -51,9 +51,9 @@ func resourceAwsInternetGatewayCreate(d *schema.ResourceData, meta interface{}) 
 	ig := *resp.InternetGateway
 	d.SetId(*ig.InternetGatewayId)
 	log.Printf("[INFO] InternetGateway ID: %s", d.Id())
-
+	var igRaw interface{}
 	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
-		igRaw, _, err := IGStateRefreshFunc(conn, d.Id())()
+		igRaw, _, err = IGStateRefreshFunc(conn, d.Id())()
 		if igRaw != nil {
 			return nil
 		}
@@ -63,9 +63,14 @@ func resourceAwsInternetGatewayCreate(d *schema.ResourceData, meta interface{}) 
 			return resource.NonRetryableError(err)
 		}
 	})
-
+	if isResourceTimeoutError(err) {
+		igRaw, _, err = IGStateRefreshFunc(conn, d.Id())()
+		if igRaw == nil {
+			return fmt.Errorf("Error refreshing internet gateway state: nil state received")
+		}
+	}
 	if err != nil {
-		return fmt.Errorf("%s", err)
+		return fmt.Errorf("Error refreshing internet gateway state: %s", err)
 	}
 
 	err = setTags(conn, d)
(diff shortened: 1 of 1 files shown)


aws/resource_aws_internet_gateway.go
```diff
@@ -63,9 +63,14 @@ func resourceAwsInternetGatewayCreate(d *schema.ResourceData, meta interface{})
 			return resource.NonRetryableError(err)
 		}
 	})
-
+	if isResourceTimeoutError(err) {
+		igRaw, _, err = IGStateRefreshFunc(conn, d.Id())()
+		if igRaw == nil {
+			return fmt.Errorf("Error refreshing internet gateway state: nil state received")
```
bflad: As an operator, I'm thinking this error message might not be so obvious what to do -- I'd suggest changing this to be more indicative of what happened from their perspective so they know how to remediate in the situation, e.g.

```suggestion
			return fmt.Errorf("error finding Internet Gateway (%s) after creation; retry running Terraform", d.Id())
```

Also the retry code above is fairly confusing as well as the logic currently reads:

- if we found the Internet Gateway, good
- if we received no error (and no Internet Gateway either since we are past the above conditional), good (`return resource.RetryableError(err)` is `return resource.RetryableError(nil)` which really is `return nil` -- see also https://github.com/hashicorp/terraform/pull/17220)
- we received an error, bad

It might make more sense for the original logic to also actually retry on a similar error message, e.g.

```go
	err = resource.Retry(5*time.Minute, func() *resource.RetryError {
		igRaw, _, err = IGStateRefreshFunc(conn, d.Id())()

		if err != nil {
		}

		if igRaw == nil {
			return resource.RetryableError(fmt.Errorf("error finding Internet Gateway (%s) after creation; retry running Terraform", d.Id()))
		}

		return nil
	})
```

That said, this retry logic may not solve the real intention here of handling EC2 eventual consistency issues after creation for _later_ API calls. This retry logic may return success while later API calls still fail. Given the current resource code, typically we would prefer to handle the eventual consistency by wrapping `setTags` and `resourceAwsInternetGatewayAttach` with `resource.Retry()` instead of a separate `resource.Retry()` before those calls. Its probably not worth changing this all right now though.

PR #9946, opened 2019-09-01
Update rate based rule limit for WAF

pasali:
https://aws.amazon.com/about-aws/whats-new/2019/08/lower-threshold-for-aws-waf-rate-based-rules/

### DIFF
--- a/aws/resource_aws_waf_rate_based_rule.go
+++ b/aws/resource_aws_waf_rate_based_rule.go
@@ -59,7 +59,7 @@ func resourceAwsWafRateBasedRule() *schema.Resource {
 			"rate_limit": {
 				Type:         schema.TypeInt,
 				Required:     true,
-				ValidateFunc: validation.IntAtLeast(2000),
+				ValidateFunc: validation.IntAtLeast(100),
 			},
 		},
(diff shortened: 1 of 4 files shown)


vad:
I just updated the provider and it gives me the same error as before:

```
aws_wafregional_rate_based_rule.login_rule: Modifying... [id=<CUT>]

Error: Error updating WAF Regional Rate Based Rule Predicates (<CUT>): Error getting WAF regional change token: InvalidParameter: 1 validation error(s) found.
- minimum field value of 2000, UpdateRateBasedRuleInput.RateLimit.
```

Is it just me? Could it be that the AWS client should be updated too?

bflad:
Hi @vad, its not likely just you. I reopened #10058 with details about the upstream issue(s) here: https://github.com/terraform-providers/terraform-provider-aws/issues/10058#issuecomment-534376731

pasali:
Sorry for the error, i assumed aws would have changed their REST API.

praveensastry:
FWIW, I tried updating the rate limit value to 100 through the web console and I could. cli is still throwing similar error though.

PR #10016, opened 2019-09-05
Updated issue templates with questions

tracypholmes:
Added 1 new lines to `feature_request` template:

1. Question asking the affected resource is new or existing

Added 2 new lines to `bug_report` template:

1. Question asking if bug has been tested with latest version of Terraform, and 
2. And if it has not been tested with an updated/the latest version of Terraform, why?

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.md
@@ -30,7 +30,7 @@ If you are running into one of these scenarios, we recommend opening an issue in
 
 <!--- Please run `terraform -v` to show the Terraform core version and provider version(s). If you are not running the latest version of Terraform or the provider, please upgrade because your issue may have already been fixed. [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions). --->
 
-### Has this been tested with the latest version of Terraform?
+### Has this been tested with the latest version of Terraform CLI and Terraform AWS Provider? If you cannot upgrade, please let us know why.
 <!-- Please respond with "yes" or "no". -->
 
 ### If not, please let us know why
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/Bug_Report.md
```diff
@@ -30,6 +30,12 @@ If you are running into one of these scenarios, we recommend opening an issue in
 
 <!--- Please run `terraform -v` to show the Terraform core version and provider version(s). If you are not running the latest version of Terraform or the provider, please upgrade because your issue may have already been fixed. [Terraform documentation on provider versioning](https://www.terraform.io/docs/configuration/providers.html#provider-versions). --->
 
+### Has this been tested with the latest version of Terraform?
```
bflad: If we are going to ask this question, I would suggest a few things:

- Removing the verbiage above about "If you are not running the latest...", or more preferably, moving that comment below this question
- Also asking about the latest version of the Terraform AWS Provider, which is versioned differently. 👍
- The next question seems like it could also be added here for consolidation and to dig deeper into potential issues people may have with upgrading in general.

```suggestion
### Has this been tested with the latest version of Terraform CLI and Terraform AWS Provider? If you cannot upgrade, please let us know why.
```

.github/ISSUE_TEMPLATE/Feature_Request.md
```diff
@@ -24,6 +24,10 @@ labels: enhancement
 
 * aws_XXXXX
 
+### Is the affected resource new or existing
```
bflad: This question may be confusing in the scenario where the feature request is for functionality that requires new resources and also affects existing ones. Is there something we can do with the existing question above to better attempt to get this information?

.github/ISSUE_TEMPLATE/Bug_Report.md
tracypholmes: That makes sense. I was trying to keep it short, but it flows better this way.

bflad:
Thanks for submitting this, @tracypholmes I believe the team has plans to overhaul these templates shortly, so we will defer any changes to that effort. 👍

PR #10018, opened 2019-09-06
internal/keyvaluetags: New Go package for consistently managing resource tags

bflad:
### Community Note

```release-note
NONE
```

The `keyvaluetags` package is designed to provide a consistent interface for handling AWS resource key-value tags. Many of the AWS Go SDK services, implement their own Go struct with `Key` and `Value` fields (e.g. `athena.Tag`) while others simply use a map (e.g. `map[string]string`). These inconsistent implementations and numerous Go types makes the process of correctly working with each of the services a tedius, previously copy-paste-modify process.

This package instead implements a single `KeyValueTags` type, which covers all key-value handling logic such as merging tags and ignoring keys via functions on the single type. Much of the code associated with tagging operations can be code generated. Adding tagging abilities for new AWS services now only requires a few lines of code.

```
--- PASS: TestAccAWSAthenaWorkGroup_disappears (13.62s)
--- PASS: TestAccAWSAthenaWorkGroup_Configuration_ResultConfiguration_EncryptionConfiguration_SseS3 (20.73s)
--- PASS: TestAccAWSAthenaWorkGroup_Configuration_BytesScannedCutoffPerQuery (25.21s)
--- PASS: TestAccAWSAthenaWorkGroup_Description (29.82s)
--- PASS: TestAccAWSAthenaWorkGroup_Configuration_EnforceWorkgroupConfiguration (31.01s)
--- PASS: TestAccAWSAthenaWorkGroup_Configuration_PublishCloudWatchMetricsEnabled (36.91s)
--- PASS: TestAccAWSAthenaWorkGroup_Tags (40.80s)
--- PASS: TestAccAWSAthenaWorkGroup_State (43.45s)
--- PASS: TestAccAWSAthenaWorkGroup_basic (57.10s)
--- PASS: TestAccAWSAthenaWorkGroup_Configuration_ResultConfiguration_EncryptionConfiguration_Kms (59.79s)
--- PASS: TestAccAWSAthenaWorkGroup_Configuration_ResultConfiguration_OutputLocation (60.00s)
```


aws/internal/keyvaluetags/key_value_tags.go
```diff
@@ -0,0 +1,171 @@
...
+
+package keyvaluetags
+
+import (
+	"strings"
+)
+
+const AwsTagKeyPrefix = `aws:`
+const ElasticbeanstalkTagKeyPrefix = `elasticbeanstalk:`
+const NameTagKey = `Name`
```
ewbankkit: Also, [RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Tagging.html#Overview.Tagging) and [Neptune](https://docs.aws.amazon.com/neptune/latest/userguide/tagging.html#Overview.Tagging) tag keys cannot start with `rds:`.

ewbankkit:
SO happy for this 😃 😃 😃.
Will there be support for services (S3 springs to mind) where there are no operation to explictly update tags, just operations to put and delete the full set of tags on a resource?

bflad:
> Will there be support for services (S3 springs to mind) where there are no operation to explictly update tags, just operations to put and delete the full set of tags on a resource?

I think once we get through a first pass of covering most of the "regular" services, we can look to migrate the very customized services into this new package as well. If nothing else, to at least use the same underlying functionality of the package, even if the wrapping code is not automatically generated. 😄

aws/internal/keyvaluetags/key_value_tags.go
bflad: Excellent find! The current functions for these two services are not currently handling this:

- https://github.com/terraform-providers/terraform-provider-aws/blob/b4efb0be0ac4315f2f4ac52955e58fc32ffe05d5/aws/tagsRDS.go#L123
- https://github.com/terraform-providers/terraform-provider-aws/blob/b4efb0be0ac4315f2f4ac52955e58fc32ffe05d5/aws/tagsNeptune.go#L106

Given that there are now a few services that require some special tag key handling, maybe it'll be best to these service-specific additions directly into the `IgnoreAws()` function directly instead of creating bespoke custom functions, which may be missed during implementations.

bflad: Ah, in order encode each service's customizations into a single `IgnoreAws()` function, the `KeyValueTags` type would need to know the service name of the tags it represents. It is certainly doable, but it would increase the implementation complexity considerably for little benefit at this time. This implementation now includes a separate `IgnoreRds()` function for the `KeyValueTags` type which of course can always be refactored later, should this type of implementation be really problematic in the future. 👍

bflad:
Rebased against master to resolve merge conflict with `GNUmakefile`

bflad:
My immediate plan after merging:

- Implement EKS tagging and document the process in the Contributing Guide
- Fix the CloudHSMv2 multiple tags issue
- Switchover `aws_vpc` and `aws_subnet` resources to prepare for a planned proof of concept

PR #10092, opened 2019-09-12
Add mutex and retry for usage plan key

hieven:
We have same issue as https://github.com/terraform-providers/terraform-provider-aws/issues/483 while creating multiple associations between API key and usage plan with default concurrency: 10.

Hence we'd like to introduce mutex lock while creating such resource to prevent from concurrent modification error

As mutex lock couldn't help with simultaneous executions of Terraform from many developers, we also add retry logic.

### Community Note

Relates https://github.com/terraform-providers/terraform-provider-aws/issues/483

```release-note
Add mutex lock to prevent errors from concurrently creating usage plan key
```

```

...
```

### DIFF
--- a/aws/resource_aws_api_gateway_usage_plan_key.go
+++ b/aws/resource_aws_api_gateway_usage_plan_key.go
@@ -3,12 +3,15 @@ package aws
 import (
 	"fmt"
 	"log"
+	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/apigateway"
 	"github.com/hashicorp/terraform/helper/schema"
 )
+
+var resourceAwsApiGatewayUsagePlanKeyMutex = &sync.Mutex{}
 
 func resourceAwsApiGatewayUsagePlanKey() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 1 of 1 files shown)


neemiasjnr:
I'm having the same problem, I created a repo reproducing the problem:
https://github.com/neemiasjnr/terraform-issue

hieven:
Hi @Ninir,
Would you mind to leave some feedback regard this PR? Does introducing mutex lock still a valid solution for API Gateway concurrent modification error?

Similar issue https://github.com/terraform-providers/terraform-provider-aws/issues/483 and PR https://github.com/terraform-providers/terraform-provider-aws/pull/2393 before

Thanks :-)

arjunpunnam:
Team,
I am facing the same problem and seems like this would fix the issue.Could you please consider reviewing the pull request

gertjangaillet:
@gdavison @bflad is this something that could be merged in the near future? We're also having this issue in a situation where we have a configurable amount of api keys to be linked to usage plans, so adding explicit dependencies does not work in our case -- and executing with parallellism 1 slows down our pipeline quite a lot...

bflad:
Hi folks 👋  After creating some additional testing for this resource that included concurrent `aws_api_gateway_usage_plan_key` resources, I was able to reproduce what appears to be the problem here. To speed things up in the future for the maintainers and get bug fixes in, it is helpful to ensure there is a covering issue that describes the problem including a reproducing configuration and typically debug logging.

In this case, we'll be taking a slightly different approach instead of a mutex where we introduce the retry logic directly into the AWS Go SDK service client since it will yield the simplest fix. Thank you @hieven for contributing this and that implementation will be refactored on top of this commit along with the covering testing. 👍

PR #10217, opened 2019-09-24
Retrieve server certificate thumbprint for embedded eks OIDC provider

wdullaer:
### Community Note

### Issue
The server certificate thumbprint is not automatically populated when creating an OIDC provider through terraform.

Closes terraform-providers/terraform-provider-aws#10104

### Description
This adds a list of server certificate thumbprints as a computed attribute
to both the aws_eks_cluster resource and aws_eks_cluster data source.
It implements the instructions from https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc_verify-thumbprint.html
in go.

After the change, the oidc portion of an eks_cluster state looks as follows

```
data "aws_eks_cluster" "test_cluster" {
    identity = [
        {
            oidc = [
                {
                    issuer = "https://oidc.eks.us-east-2.amazonaws.com/id/ABCDEFGHI"
                    thumbprint_list = [
                        "9e99a48a9960b14926bb7f3b02e22da2b0ab7280",
                    ]
                },
            ]
        },
    ]
}
```

The happy path acceptance tests have been updated.

### Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note
ENHANCEMENTS
* resource/aws_eks_cluster: Added thumbprint_list attribute to embedded OIDC provider
* data-source/aws_eks_cluster: Added thumbprint_list attribute to embedded OIDC provider
```

```
$ make testacc TESTARGS='-run=TestAccAWSEksCluster_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSEksCluster_basic -timeout 120m
=== RUN   TestAccAWSEksCluster_basic
=== PAUSE TestAccAWSEksCluster_basic
=== CONT  TestAccAWSEksCluster_basic
--- PASS: TestAccAWSEksCluster_basic (1119.64s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       1119.678s
```

```
$ make testacc TESTARGS='-run=TestAccAWSEksClusterDataSource_basic'
TF_ACC=1 go test ./... -v -parallel 20 -run=TestAccAWSEksClusterDataSource_basic -timeout 120m
=== RUN   TestAccAWSEksClusterDataSource_basic
=== PAUSE TestAccAWSEksClusterDataSource_basic
=== CONT  TestAccAWSEksClusterDataSource_basic
--- PASS: TestAccAWSEksClusterDataSource_basic (1133.96s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       1133.998s
```

### DIFF
--- a/aws/resource_aws_eks_cluster.go
+++ b/aws/resource_aws_eks_cluster.go
@@ -1,8 +1,12 @@
 package aws
 
 import (
+	"crypto/sha1"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
 	"strings"
 	"time"
 
@@ -77,6 +81,13 @@ func resourceAwsEksCluster() *schema.Resource {
 									"issuer": {
 										Type:     schema.TypeString,
 										Computed: true,
+									},
+									"thumbprint_list": {
+										Type:     schema.TypeList,
+										Computed: true,
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
 									},
 								},
 							},
@@ -254,7 +265,11 @@ func resourceAwsEksClusterRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("created_at", aws.TimeValue(cluster.CreatedAt).String())
 	d.Set("endpoint", cluster.Endpoint)
 
-	if err := d.Set("identity", flattenEksIdentity(cluster.Identity)); err != nil {
+	eksIdentity, err := flattenEksIdentity(cluster.Identity)
+	if err != nil {
+		return fmt.Errorf("error obtaining eks cluster identity: %s", err)
+	}
+	if err := d.Set("identity", eksIdentity); err != nil {
 		return fmt.Errorf("error setting identity: %s", err)
 	}
 
@@ -455,28 +470,37 @@ func flattenEksCertificate(certificate *eks.Certificate) []map[string]interface{
 	return []map[string]interface{}{m}
 }
 
-func flattenEksIdentity(identity *eks.Identity) []map[string]interface{} {
+func flattenEksIdentity(identity *eks.Identity) ([]map[string]interface{}, error) {
 	if identity == nil {
-		return []map[string]interface{}{}
+		return []map[string]interface{}{}, nil
 	}
 
+	oidc, err := flattenEksOidc(identity.Oidc)
+	if err != nil {
+		return nil, err
+	}
 	m := map[string]interface{}{
-		"oidc": flattenEksOidc(identity.Oidc),
+		"oidc": oidc,
 	}
 
-	return []map[string]interface{}{m}
+	return []map[string]interface{}{m}, nil
 }
 
-func flattenEksOidc(oidc *eks.OIDC) []map[string]interface{} {
+func flattenEksOidc(oidc *eks.OIDC) ([]map[string]interface{}, error) {
 	if oidc == nil {
-		return []map[string]interface{}{}
+		return []map[string]interface{}{}, nil
 	}
 
+	thumbprint, err := getCertThumbprintList(oidc.Issuer)
+	if err != nil {
+		return nil, err
+	}
 	m := map[string]interface{}{
-		"issuer": aws.StringValue(oidc.Issuer),
+		"issuer":          aws.StringValue(oidc.Issuer),
+		"thumbprint_list": thumbprint,
 	}
 
-	return []map[string]interface{}{m}
+	return []map[string]interface{}{m}, nil
 }
 
 func flattenEksVpcConfigResponse(vpcConfig *eks.VpcConfigResponse) []map[string]interface{} {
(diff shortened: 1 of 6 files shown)


bflad:
Hi @wdullaer 👋 Thank you for submitting this.

It seems that `eksctl` is accomplishing this differently: https://github.com/weaveworks/eksctl/blob/87cf27f54493551381fac016601d8f687788ec0f/pkg/iam/oidc/api.go#L115-L133

I'm not sure if we should at least follow suit with that type of logic to simplify this process.

That said, this problem seems a little out of scope for the `aws_eks_cluster` resource and data source. While the information relates directly to OIDC Issuer URL associated with the EKS Cluster, the additional HTTP connection might create unexpected failures for Terraform runners without direct access to the OIDC endpoint.

In this case, we would probably lean more towards a generic solution for Terraform as outlined in https://github.com/terraform-providers/terraform-provider-tls/issues/52 to create a separate Terraform TLS Provider data source to fetch this information for any TLS certificate. Another option would be asking the EKS service team (via support cases or your TAM) to include this information directly in the EKS API response, which would make it a great candidate for inclusion in this Terraform resource and data source.

What do you think?

bflad:
Another potential issue with implementing this directly would be that older EKS Clusters without OIDC Issuer information could potentially fail. This would need to be verified as working with version 1.12 and older clusters.

wdullaer:
> Another potential issue with implementing this directly would be that older EKS Clusters without OIDC Issuer information could potentially fail

I think this should be fine: the code shortcircuits higher up if there's no OIDC data. The code I wrote will also return before executing any HTTP requests if the issuer is `nil`.

> It seems that `eksctl` is accomplishing this differently

Looking at that code, I think `eksctl` is wrong there. The AWS docs state that the thumbprint of the root certificate of the server which returns the OIDC keys (the json web keyset) needs to be given. The issuer url is the OIDC discovery endpoint. The key endpoint is obtained from the `jwks_uri` attribute that this returns (as per the OIDC spec). This makes sense since the thumbprint ensures you are obtaining any public keys from the correct server.
It is highly likely that for a given OIDC server both endpoints are on the same hostname or at the very least using the same root certificate, so this won't cause much issues in practise, but I do believe my implementation here is the correct one (at the cost of 1 additional HTTP call which might fail)

> Another option would be asking the EKS service team (via support cases or your TAM)

This is definitely something we will do. Having the AWS API return this data is the ideal scenario, since they know exactly which server you should be trusting. Unfortunately there's no telling if and when AWS will do this. It might be next week, it might be 2 years from now.
I view this PR as a stop gap solution until they do. It provides the user experience most people will expect.

> the additional HTTP connection might create unexpected failures for Terraform runners without direct access to the OIDC endpoint

If you can access the AWS API's, it'll be highly likely you'll be able to hit the OIDC endpoint they provide, but orgs do tend to have wonky networking setups.
The higher risk in my opinion is that this method will grap the wrong fingerprint in case the corporate network MITMs TLS connections (but I'm not terribly inclined to make life any easier for any orgs that do this)

> In this case, we would probably lean more towards a generic solution for Terraform as outlined in terraform-providers/terraform-provider-tls#52 

I think both are useful to have. I would expect the embedded OIDC info to have the thumbprints as well (user study with sample size = 1). The additional data source and provider is a bit clunky and hard to discover.
That said, I'm happy to help build that feature if the consensus of the community is that this resource is not the right place to implement this functionality

ashafer01:
👍 to this feature existing somehow or another. This is pretty essential for utilizing [Pod-attached IAM roles](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html)

Perhaps if integrating directly is not ideal for compatibility, another data provider similar to `aws_eks_cluster_auth` could be added? `aws_eks_oidc_thumprint` perhaps?

drmjo:
When creating an OpenID Connect Identity provider on the AWS web UI the fingerprint is retrieved automatically.

with a warning in the docs to make sure you verify this information manually https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc_verify-thumbprint.html 

> We recommend that you also obtain the thumbprint for your OIDC IdP manually and verify that the console fetched the correct thumbprint.


however when doing this via the API
you get
```
1 validation error detected: Value null at 'thumbprintList' failed to satisfy constraint: Member must not be null
```

since there is no means of "validating this thing manually"

*SHOULD* this information be fetched automatically by AWS?
OR
Is this something that *SHOULD* be known prior to configuring a provider?

I am asking because this seems to be a security thing. I don't know what the exact implications are.

would it be safe to assume the "CA thumbprint" for the EKS OIDC JWKS endpoint will always be `9E99A48A9960B14926BB7F3B02E22DA2B0AB7280`? I don't think AWS is going to move that url out of their domain. What are the chances that they will change CAs.

That would cause a larger issue 💥 

I have been just hardcoding that in my EKS terraform module I'd like to know more how this is used by amazon.

wdullaer:
The thumbprint is to prevent a man in the middle attack: you are telling AWS "trust the OIDC endpoint at this URL, but only if it presents a certificate that is signed by this CA".
The underlying assumption is that a CA will never hand out a certificate to a different entity for the same domain: ie you won't get a google.com cert unless you're google.

The reason they're telling you to check is because this is a TOFU (trust on first use): if someone pulls off a man in the middle at the time you are provisioning this trust relationship, they have won.

Hardcoding the fingerprint is relatively low risk: root CA certs stay valid for a long time. However AWS may decide to roll their root CA at some point in the future, at which time this would break.

aws/resource_aws_eks_cluster.go
```diff
@@ -512,6 +536,46 @@ func flattenEksEnabledLogTypes(logging *eks.Logging) *schema.Set {
...
+	jwks, ok := issuerData["jwks_uri"]
+	if !ok {
+		return []string{}, fmt.Errorf("error retrieving jwks_uri from %s: %s", endpoint, issuerData)
+	}
+	jwksUri, ok := jwks.(string)
+	if !ok {
+		return []string{}, fmt.Errorf("error retrieving jwks_uri from %s: %s", endpoint, issuerData)
+	}
+
+	jwksResp, err := http.Get(jwksUri)
```
nairb774: `jwksResp.Body` needs to be closed.

wdullaer: nice catch

bflad:
Hi everyone 👋  Just to provide an update here, it appears the associated issue has been closed and that generic support for this type of lookup functionality will [land in the Terraform TLS Provider](https://github.com/hashicorp/terraform-provider-tls/pull/62) shortly as a new `tls_certificate` data source. Operators will have the choice of whether to follow the AWS recommendations to manually hardcode/verify the thumbprints to potentially curtail man-in-the-middle attacks or to use the new data source in their configurations to handle any (likely very long term) updates.

Thanks to @wdullaer for submitting this contribution and continuing this conversation. If the EKS API does ever implement this information directly, we can certainly consider adding it into the Terraform AWS Provider resources.

kmoe:
The `tls_certificate` resource has landed in TLS provider version 2.2.0.

PR #10267, opened 2019-09-27
Add data source to download a s3 bucket object

### LINKED ISSUE #1323: Download files from S3 bucket

julienduchesne:
Closes #1323

Adds a data source that allows Terraform to download a S3 object to a file

```
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccDataSourceAWSS3DownloadBucketObject_basic -timeout 120m
=== RUN   TestAccDataSourceAWSS3DownloadBucketObject_basic
=== PAUSE TestAccDataSourceAWSS3DownloadBucketObject_basic
=== CONT  TestAccDataSourceAWSS3DownloadBucketObject_basic
--- PASS: TestAccDataSourceAWSS3DownloadBucketObject_basic (113.87s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       113.939s
```


website/docs/d/s3_download_bucket_object.html.markdown
```diff
@@ -0,0 +1,42 @@
...
+description: |-
+    Downloads a bucket object to be used as a local file
+---
+
+# Data Source: aws_s3_download_bucket_object
+
+The S3 Download Bucket Object Data Source allows download of a file to the disk to be used as a local file.
+
+## Example Usage
+
```
AbdulRahmanAlHamali: The description of the example here is wrong. I think you changed the example.

AbdulRahmanAlHamali:
Awesome!
I have two problems with that:
1. Wouldn't it make more sense to just provide a "force_download" parameter to the original data source, instead of having a separate datasource?
2. It seems I cannot access the filename by using `"${data.aws_s3_download_bucket_object.my_data.filename}"`

julienduchesne:
1. We have this PR too: https://github.com/terraform-providers/terraform-provider-aws/pull/1745 but you need a data to write the file and as I remember, if you had a big file, passing the body of your file around in two resources was very slow. Doing it in one resource was much faster.
2. I'll check it out

PR #10293, opened 2019-09-29
tests/resource/aws_ebs_snapshot_copy: Refactoring to remove inconsistent ProviderFactories usage and make region agnostic

bflad:
### Community Note

```release-note
NONE
```

Noticed this was the only testing left in the provider using the direct `map[string]terraform.ResourceProviderFactory{}` setup instead of `testAccProviderFactories()`. While here, also refactored the testing to make it region agnostic and use consistent test resource naming.

```
--- PASS: TestAccAWSEbsSnapshotCopy_disappears (61.46s)
--- PASS: TestAccAWSEbsSnapshotCopy_basic (62.02s)
--- PASS: TestAccAWSEbsSnapshotCopy_withDescription (62.38s)
--- PASS: TestAccAWSEbsSnapshotCopy_withKms (70.19s)
--- PASS: TestAccAWSEbsSnapshotCopy_withRegions (74.09s)
```

### DIFF
--- a/aws/resource_aws_ebs_snapshot_copy_test.go
+++ b/aws/resource_aws_ebs_snapshot_copy_test.go
@@ -2,7 +2,6 @@ package aws
 
 import (
 	"fmt"
-	"regexp"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


aws/resource_aws_ebs_snapshot_copy_test.go
```diff
@@ -13,7 +12,9 @@ import (
 )
 
 func TestAccAWSEbsSnapshotCopy_basic(t *testing.T) {
-	var v ec2.Snapshot
+	var snapshot ec2.Snapshot
```
nywilken: To me: it would help if variable renaming was done as a separate commit so that the actual change can be validated independently of the renames. I'm now sure how the rest of the folks feel about it.

bflad: Aye - good call. Will try to do that in the future.

PR #10313, opened 2019-09-30
docs/provider: Add Adding Resource Tagging Support section

bflad:
### Community Note

```release-note
NONE
```

Includes information how to work with the new `internal/keyvaluetags` generators.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -25,6 +25,7 @@ ability to merge PRs and respond to issues.
         - [Documentation Update](#documentation-update)
         - [Enhancement/Bugfix to a Resource](#enhancementbugfix-to-a-resource)
         - [Adding Resource Import Support](#adding-resource-import-support)
+        - [Adding Resource Tagging Support](#adding-resource-tagging-support)
         - [New Resource](#new-resource)
         - [New Service](#new-service)
         - [New Region](#new-region)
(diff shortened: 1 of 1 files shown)


ewbankkit:
Any guidance as to when the tag support code changes should be made for [new services](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#new-service)?
* In the same PR as the new service client changes
* A separate PR after the new service client PR
* With the first new resource that supports resource tags

bflad:
Hey @ewbankkit, great question. I'm not sure if the maintainers will have too strong of an opinion on this quite yet since working with the library is still very new.

For new services, it seems perfectly fine to include the tag support in the same PR as adding the dependency and service client, if you notice tagging support in the API. Maybe we can add that as a line item to the new service checklist, although it does add some slight complexity there in documenting it (since tagging support in the API may be available from the beginning or much later on, like EKS). Otherwise, I think it might be okay if we accept it separately as well.

If we run into more complicated merging situations in the libray, the pull request review for new services feels too heavy with tagging support, or the process doesn't feel right as we go on, we can certainly come up with some more pragmatic guidelines. 😄

ewbankkit:
@bflad Agree that we should work with this new mechanism for a while before forming string opinions.

.github/CONTRIBUTING.md
```diff
@@ -213,6 +214,201 @@ In addition to the below checklist and the items noted in the Extending Terrafor
 - [ ] _Resource Acceptance Testing Implementation_: In the resource acceptance testing (e.g. `aws/resource_aws_service_thing_test.go`), implementation of `TestStep`s with `ImportState: true`
 - [ ] _Resource Documentation Implementation_: In the resource documentation (e.g. `website/docs/r/service_thing.html.markdown`), addition of `Import` documentation section at the bottom of the page
 
+#### Adding Resource Tagging Support
+
+AWS provides key-value metadata across many services and resources, which can be used for a variety of use cases including billing, ownership, and more. See the [AWS Tagging Stategy page](https://aws.amazon.com/answers/account-management/aws-tagging-strategies/) for more information about tagging at a high level.
```
ewbankkit: Typo: `Stategy` -> `Strategy`.

ewbankkit:
Do we want to explicitly call out what to do for resource that don't (yet) support tag-on-create, where you have to effectively call the `Update` functionality from the `Create` function? 
For [`aws_api_gateway_v2_api`](https://github.com/terraform-providers/terraform-provider-aws/pull/8842) I went with code like:

```go
	if v := d.Get("tags").(map[string]interface{}); len(v) > 0 {
		arn := ...
		if err := keyvaluetags.Apigatewayv2UpdateTags(conn, arn, map[string]interface{}{}, v); err != nil {
			return fmt.Errorf("error creating tags: %s", err)
		}
	}
```

after having successfully created the resource and set its ID.

.github/CONTRIBUTING.md
bflad: Addressed in a732852d4e30eec3cc7923a5e089e6dbe85a35bf 👍

bflad:
Addressed the last comment about tagging on creation vs not in a732852d4e30eec3cc7923a5e089e6dbe85a35bf

@ewbankkit FYI you can simplify the UpdateTags call with `nil` instead of `map[string]interface{}{}` to increase the readability slightly.

PR #10435, opened 2019-10-09
Batch up requests with more than 10 target group changes

beezly:
The AWS API only supports adding/removing 10 requests at a time. 

### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Closes #256 

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note
Batch up requests for Attach/DetachLoadBalancerTargetGroups into groups of 10
```

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

PR #10445, opened 2019-10-10
resource/aws_autoscaling_group: Batch ELB attachments and detachments by 10 to prevent API and rate limiting errors

bflad:
### Community Note

Reference: https://github.com/terraform-providers/terraform-provider-aws/issues/256 (being closed by #10435)
Reference: https://github.com/terraform-providers/terraform-provider-aws/pull/10435
Reference: https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_AttachLoadBalancers.html
Reference: https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_DetachLoadBalancers.html

```release-note
* resource/aws_autoscaling_group: Batch ELB attachments and detachments by 10 to prevent API and rate limiting errors
```

This change is complementary to the target group batching in #10435. The AutoScaling API for `AttachLoadBalancers` and `DetachLoadBalancers` only allow 10 elements at a time. In addition to the batching to split the API requests, we must wait for the batch to fully complete before moving onto the next batch, otherwise the API returns a rate limiting error:

```
--- FAIL: TestAccAWSAutoScalingGroup_LoadBalancers (360.22s)
    testing.go:569: Step 2 error: errors during apply:

        Error: Error updating Load Balancers for AutoScaling Group (tf-asg-2019101000090127270000000d), error: ValidationError: Trying to update too many Load Balancers/Target Groups at once. The limit is 10
```

```
--- PASS: TestAccAWSAutoScalingGroup_LoadBalancers (443.84s)
```

### DIFF
--- a/aws/resource_aws_autoscaling_group.go
+++ b/aws/resource_aws_autoscaling_group.go
@@ -884,22 +884,56 @@ func resourceAwsAutoscalingGroupUpdate(d *schema.ResourceData, meta interface{})
 		add := expandStringList(ns.Difference(os).List())
 
 		if len(remove) > 0 {
-			_, err := conn.DetachLoadBalancers(&autoscaling.DetachLoadBalancersInput{
-				AutoScalingGroupName: aws.String(d.Id()),
-				LoadBalancerNames:    remove,
-			})
-			if err != nil {
-				return fmt.Errorf("Error updating Load Balancers for AutoScaling Group (%s), error: %s", d.Id(), err)
+			// API only supports removing 10 at a time
+			var batches [][]*string
+
+			batchSize := 10
+
+			for batchSize < len(remove) {
+				remove, batches = remove[batchSize:], append(batches, remove[0:batchSize:batchSize])
+			}
+			batches = append(batches, remove)
+
+			for _, batch := range batches {
+				_, err := conn.DetachLoadBalancers(&autoscaling.DetachLoadBalancersInput{
+					AutoScalingGroupName: aws.String(d.Id()),
+					LoadBalancerNames:    batch,
+				})
+
+				if err != nil {
+					return fmt.Errorf("error detaching AutoScaling Group (%s) Load Balancers: %s", d.Id(), err)
+				}
+
+				if err := waitUntilAutoscalingGroupLoadBalancersRemoved(conn, d.Id()); err != nil {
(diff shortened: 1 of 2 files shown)


aws/resource_aws_autoscaling_group.go
```diff
@@ -1493,3 +1527,75 @@ func flattenAutoScalingMixedInstancesPolicy(mixedInstancesPolicy *autoscaling.Mi
 
 	return []interface{}{m}
 }
+
```
ryndaniels: Is there ever a case when we'd want to set a timeout on these functions?

aws/resource_aws_autoscaling_group_test.go
```diff
@@ -1358,6 +1358,56 @@ func TestAccAWSAutoScalingGroup_LaunchTemplate_IAMInstanceProfile(t *testing.T)
...
+			{
+				Config: testAccAWSAutoScalingGroupConfig_LoadBalancers(rName, 11),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSAutoScalingGroupExists(resourceName, &group),
+					resource.TestCheckResourceAttr(resourceName, "load_balancers.#", "11"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
```
ryndaniels: 🎊

aws/resource_aws_autoscaling_group.go
bflad: Great question and yes, there might be! Quickly scanning the recent issues, I don't see any complaints about the similar change to the target group handling (that also does not have timeouts), but for now this is dependent on the AWS side entering some sort of failure state rather than staying in the adding/deleting state forever in those cases. 🤞

Itiho:
The problem still happens here:
Terraform v0.12.18
+ provider.aws v2.42.0


```
Error: Failure attaching AutoScaling Group lc-autf-4-0-20-27-PGWEB-QA-20191214194536076900000001-asg with ALB Target Group: arn:aws:elasticloadbalancing:us-east-1:106431551699:targetgroup/tf-2019110818465657090000000c/20b68e9c4fa34380: ValidationError: Trying to update too many Load Balancers/Target Groups at once. The limit is 10
	status code: 400, request id: 6c62f77e-1eaa-11ea-96a8-0d0d00d2891c
```

Issue #1243, opened 2017-07-26
Feature request: Allow aws_instance resource to create and manage spot instances


tomelliff:
I haven't taken a particularly long look at the code to see how much impact this would have but it would be nice if the `aws_instance` resource could be used to manage spot instances instead of having to explicitly use the `aws_spot_instance_request` resource.

This would simplify being able to switch between spot and on demand requests without duplicating code. As it is I have a module that creates EC2 instances and then a bunch of other dependent resources (Route53 records, Cloudwatch alarms etc) and I'm going to have to duplicate the `aws_instance` resource plus all the resources dependent on it to use the `aws_spot_instance_request` and also make them conditional on whether a spot price has been set.

So as a minimal example I'll have something like:

```hcl
variable "instance_count" {
  default = 1
}

variable "spot_price" {
  default = ""
}

data "aws_ami" "ubuntu" {
  most_recent = true

  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
  }

  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }

  owners = ["099720109477"] # Canonical
}

resource "aws_instance" "instance" {
  count         = "${var.spot_price == "" ? var.instance_count : 0}"
  ami           = "${data.aws_ami.ubuntu.id}"
  instance_type = "t2.micro"

  tags {
    Name = "${var.name}${var.instance_count > 1 ? format("-%v", count.index+1) : ""}.${var.domain}"
  }
}

resource "aws_spot_instance_request" "instance" {
  count         = "${var.spot_price != "" ? var.instance_count : 0}"
  ami           = "${data.aws_ami.ubuntu.id}"
  instance_type = "t2.micro"
  spot_price    = "${var.spot_price}"

  tags {
    Name = "${var.name}${var.instance_count > 1 ? format("-%v", count.index+1) : ""}.${var.domain}"
  }
}

data "aws_vpc" "vpc" {
  tags {
    Name = "myvpc"
  }
}

data "aws_route53_zone" "zone" {
  name   = "example.com"
  vpc_id = "${data.aws_vpc.selected.id}"
}

resource "aws_route53_record" "on_demand_record" {
  count   = "${var.spot_price == "" ? var.instance_count : 0}"
  zone_id = "${data.aws_route53_zone.zone.zone_id}"
[...]

PR #10536, opened 2019-10-17
Allow aws instance create spot instances #1243

dbgeek:
### Community Note

Relates  #1243

This is a Draft pull request of an example of how to make aws_instance resource to support the spot market. 

This is an example of how to do it. Please make comments if this is an good way to do it. I think this makes it easier to switch between normal ec2 instances and spot instances. I do not think we need to expose attributes like spot_request_state,  spot_bid_status, etc or?

I have not yet look into the terraform provider documentation but that needs to be updated.

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -517,6 +517,37 @@ func resourceAwsInstance() *schema.Resource {
 					},
 				},
 			},
+
+			"use_spot_market": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+				Default:  false,
+			},
+			"max_price": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"block_duration_minutes": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+			"instance_interruption_behaviour": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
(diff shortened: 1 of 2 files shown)


pcgeek86:
Can we get this merged?

pcgeek86:
@ryndaniels Any chance you could help out getting this feature merged, so we can provision EC2 Spot instances using the `aws_instance` resource?

stefansundin:
I was just looking to see if I could implement this, but I'm glad I looked if someone else had started already. Thanks for making the effort.

I don't work for HashiCorp, but I think they would tell you to put this stuff in its own block, maybe called `instance_market_options` in order to emulate the API as closely as possible. Then you can get rid off `use_spot_market`. See how `credit_specification` is implemented.

So the end-user would use something like the following:

```hcl
  # ...

  instance_market_options {
    max_price = "0.1"
    spot_instance_type = "persistent"
    # ...
  }
}
```

Hopefully someone will stop by to help you get this merged. :)

stefansundin:
I just realized that the launch template resource has these options already, so this should probably use the exact same structure if possible.

https://github.com/terraform-providers/terraform-provider-aws/blob/2cd7beee14a0746d6e925da718c2121e96379da7/aws/resource_aws_launch_template.go#L265-L309

dbgeek:
Thanks @stefansundin will redo this pull request like they have done it in the launch template.

dbgeek:
Replaced this Pull request with #11427 do make the implementation as @stefansundin suggested.

PR #10584, opened 2019-10-21
#396 fix vpn connection tunnel output sorting

### LINKED ISSUE #396: VPN Connection Tunnels incorrectly ordered
_This issue was originally opened by @kitforbes as hashicorp/terraform#10411. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
Terraform v0.7.13

### Affected Resource(s)
- aws_vpn_connection

```hcl
module "vpc" {
  source = "../../../modules/aws/network/vpc"

  name        = "${var.name}-vpc"
  environment = "${var.environment}"
  cidr        = "${var.vpc_cidr}"
}

resource "aws_customer_gateway" "edinburgh" {
  type       = "${var.vpn_connection_type}"
  bgp_asn    = "${var.vpn_bgs_asn}"
  ip_address = "${var.vpn_local_ip}"

  tags {
    Name        = "${var.name}-customer-gateway"
    Environment = "${var.environment}"
    Creator     = "Terraform"
  }
}

resource "aws_vpn_gateway" "edinburgh" {
  vpc_id = "${module.vpc.vpc_id}"

  tags {
    Name        = "${var.name}-edinburgh"
    Environment = "${var.environment}"
    Creator     = "Terraform"
  }
}

resource "aws_vpn_connection" "edinburgh" {
  vpn_gateway_id      = "${aws_vpn_gateway.edinburgh.id}"
  customer_gateway_id = "${aws_customer_gateway.edinburgh.id}"
  type                = "${var.vpn_connection_type}"
  static_routes_only  = true

  tags {
    Name        = "${var.name}-edinburgh"
    Environment = "${var.environment}"
    Creator     = "Terraform"
  }
}

output "vpn_config" {
  value = <<CONFIGURATION

  Tunnel 1
    Address:     ${aws_vpn_connection.edinburgh.tunnel1_address}
  Tunnel 2
    Address:     ${aws_vpn_connection.edinburgh.tunnel2_address}
CONFIGURATION
}
```

### Expected Behavior
`${aws_vpn_connection.edinburgh.tunnel1_address}` should output the value of Tunnel 1 in the AWS Console.

### Actual Behavior
It actually outputs the value of tunnel 2, and tunnel 2 outputs the value of tunnel 1. Not a critical issue by any means.

### Steps to Reproduce
Create an AWS VPN connection and compare the Terraform output to actual AWS Console information in the VPC section.

jochen42:
### Community Note

Closes #396 

```release-note
BUG FIXES:

* fix incorrect order (tunnel1 and tunnel2) in output of resource aws_vpn_connection
```

```
TF_ACC=1 go test ./... -v -count 1 -parallel 20 -run=TestAccXXX -timeout 120m
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       0.037s [no tests to run]
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws/internal/flatmap      0.007s [no tests to run]
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws/internal/keyvaluetags 0.014s [no tests to run]
```

### DIFF
--- a/aws/resource_aws_vpn_connection.go
+++ b/aws/resource_aws_vpn_connection.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"net"
 	"regexp"
-	"sort"
 	"strings"
 	"time"
 
@@ -591,9 +590,6 @@ func xmlConfigToTunnelInfo(xmlConfig string) (*TunnelInfo, error) {
 	if err := xml.Unmarshal([]byte(xmlConfig), &vpnConfig); err != nil {
 		return nil, fmt.Errorf("Error Unmarshalling XML: %s", err)
 	}
-
-	// don't expect consistent ordering from the XML
-	sort.Sort(vpnConfig)
 
 	tunnelInfo := TunnelInfo{
 		Tunnel1Address:          vpnConfig.Tunnels[0].OutsideAddress,
--- a/aws/resource_aws_vpn_connection_test.go
+++ b/aws/resource_aws_vpn_connection_test.go
@@ -584,35 +584,6 @@ const testAccAwsVpnTunnelInfoXML = `
         <ip_address>123.123.123.123</ip_address>
       </tunnel_outside_address>
       <tunnel_inside_address>
-        <ip_address>SECOND_CGW_INSIDE_ADDRESS</ip_address>
-        <network_mask>255.255.255.252</network_mask>
-        <network_cidr>30</network_cidr>
-      </tunnel_inside_address>
-    </customer_gateway>
-    <vpn_gateway>
-      <tunnel_outside_address>
-        <ip_address>SECOND_ADDRESS</ip_address>
-      </tunnel_outside_address>
-      <tunnel_inside_address>
-        <ip_address>SECOND_VGW_INSIDE_ADDRESS</ip_address>
-        <network_mask>255.255.255.252</network_mask>
-        <network_cidr>30</network_cidr>
-      </tunnel_inside_address>
-      <bgp>
-        <asn>SECOND_BGP_ASN</asn>
-        <hold_time>32</hold_time>
-      </bgp>
-    </vpn_gateway>
-    <ike>
-      <pre_shared_key>SECOND_KEY</pre_shared_key>
-    </ike>
-  </ipsec_tunnel>
-  <ipsec_tunnel>
-    <customer_gateway>
-      <tunnel_outside_address>
-        <ip_address>123.123.123.123</ip_address>
-      </tunnel_outside_address>
-      <tunnel_inside_address>
         <ip_address>FIRST_CGW_INSIDE_ADDRESS</ip_address>
         <network_mask>255.255.255.252</network_mask>
         <network_cidr>30</network_cidr>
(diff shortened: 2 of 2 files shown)


n3ph:
This does not fix the problem itself. You simply sorted the test xml blob yourself.

jochen42:
> This does not fix the problem. You simply sorted the test xml blob yourself.

no! i just removed  the senseless sorting of an correct xml wich is the root-problem. 
i tested this with many real-world-vpn connections and the result is 100% correct.

n3ph:
For the case that the order of tunnels in the XML does not match the order of `options[]` applied to the create request it makes more sense to check whether `vpnConfig.Tunnels[0].VgwInsideAddress` or `vpnConfig.Tunnels[0].CgwInsideAddress` matches the CIDR block already defined by `tunnel1_inside_cidr` and `tunnel2_inside_cidr`. 

Do you agree?

n3ph:
Checked that again..

It seems that getting rid of `sort.Sort(vpnConfig)` leads to the same remote state I have with my fix applied. So it seems to be sufficient to get rid of the sort.

I am undecided if it's okay to expect the right order without checking it.

jochen42:
hi @n3ph ,

i think you [are facing another issue](https://github.com/terraform-providers/terraform-provider-aws/pull/11298) than [me](https://github.com/terraform-providers/terraform-provider-aws/pull/10584). My issue  is just the wrong order of the tunnels itself.
With the removed complex-object-sort i just want to have the correct order like aws-api answers.

it is hard to debug for the operators at the client-side if they are wrong ordered.

f.e. the results of the following command should be in the same order as the provided connection-parameter vom resource-output or statefile.

```sh
aws ec2 describe-vpn-connections --vpn-connection-id <some_connection_idf> --query 'VpnConnections[*].VgwTelemetry[*]'
```
the same for aws console.

your issue seems to be much more complex. for me it looks like you are right with the wrong tunnel*_vgw_inside_address, but your solution:

* will also provide the wrong tunnel-order/sorting (hard to debug)
* tries to fix a problem on aws-side with an assumption of their internal-behavior  (what if you are wrong or they change the wrong behavior next week?)

Perhaps your issue should be reported to aws support?

bflad:
Marking as a breaking change since removing the existing sorting without accounting for what was in the previous Terraform configuration/state was can cause potential differences with existing Terraform environments. My understanding of this issue is that the EC2 API may not return tunnel results consistently, which was why the sorting was originally added a long while ago.

jochen42:
> Marking as a breaking change since removing the existing sorting without accounting for what was in the previous Terraform configuration/state was can cause potential differences with existing Terraform environments. My understanding of this issue is that the EC2 API may not return tunnel results consistently, which was why the sorting was originally added a long while ago.

you are right. it's an breaking change. but the sorting in the current version doesn't make sense at all. a complext object is sorted without an definition which fields are used for wording and how the sorting  of complex strings like ip-networks/addresses are done. so: the sorting was never correct with the line of code. and it's also impossible that this sorting fixed any previous sorting-problem on aws-side.

bflad:
Hi @jochen42 👋  Thank you again for submitting this awhile back. To ensure the tunnel sorting changes are backwards compatible with existing resources, we have opted to take a different approach in https://github.com/hashicorp/terraform-provider-aws/pull/19077, which should be released shortly to fix this issue.

PR #10684, opened 2019-10-30
website: Add subcategories to docs for display on the Terraform Registry

justincampbell:
We will soon be displaying documentation for this provider both on [terraform.io](https://www.terraform.io/docs/providers/index.html) and on [the Terraform Registry](https://registry.terraform.io/providers).

Documentation displayed on the Terraform Registry will not use the ERB layout containing the existing navigation, and will instead build the navigation dynamically based on the directory and YAML frontmatter of a file. For providers that group similar resources and data sources by service or use case (subcategories), we'll need to add this information to the Markdown source file.

This PR modifies resource and data source documentation source files by adding a subcategory to the metadata if those files are currently grouped on terraform.io.

For more information about how documentation is rendered on the Terraform Registry, please see this reference: [Terraform Registry - Provider Documentation](https://www.terraform.io/docs/registry/providers/docs.html).

### Community Note

```release-note
NONE
```

```sh
$ make docscheck
$ # exited 0
```


website/docs/r/launch_configuration.html.markdown
```diff
@@ -1,4 +1,5 @@
 ---
+subcategory: "EC2"
```
bflad: Not for this pull request, but this is confusingly in Autoscaling for the data source and EC2 for the resource. Technically, I think it should be in Autoscaling since that's the service API.

justincampbell:
@bflad Updated this to not rename any files! Let me know if it's ok to merge soon.

Issue #8530, opened 2019-05-06
aws_acm_certificate appears to be waiting for DNS validation to complete when SANs are present


tdmalone:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
Terraform v0.11.13
+ provider.aws v2.8.0
```

### Affected Resource(s)

* [aws_acm_certificate](https://www.terraform.io/docs/providers/aws/r/acm_certificate.html)

### Terraform Configuration Files

```hcl
resource "aws_acm_certificate" "main" {
  domain_name = "example.com"
  validation_method = "DNS"

  subject_alternative_names = [
    "one.example.com",
  ]
}
```

### Expected Behavior

The certificate should be created in a state of pending validation, and Terraform should return success.

### Actual Behavior

The certificate _is_ created in a state of pending validation, but Terraform appears to wait for validation of the SAN to succeed (which it won't, DNS records haven't been retrieved and added yet), and then errors as follows:

```
aws_acm_certificate.main: Creating...
  arn:                         "" => "<computed>"
  domain_name:                 "" => "example.com"
  domain_validation_options.#: "" => "<computed>"
  subject_alternative_names.#: "" => "1"
  subject_alternative_names.0: "" => "one.example.com"
  validation_emails.#:         "" => "<computed>"
  validation_method:           "" => "DNS"
aws_acm_certificate.main: Still creating... (10s elapsed)
aws_acm_certificate.main: Still creating... (20s elapsed)
aws_acm_certificate.main: Still creating... (30s elapsed)
aws_acm_certificate.main: Still creating... (40s elapsed)
aws_acm_certificate.main: Still creating... (50s elapsed)
aws_acm_certificate.main: Still creating... (1m0s elapsed)

Error: Error applying plan:

1 error(s) occurred:

* aws_acm_certificate.main: 1 error(s) occurred:

* aws_acm_certificate.main: No validation options need to retry: {
  DomainName: "one.example.com",
  ValidationMethod: "DNS",
  ValidationStatus: "PENDING_VALIDATION"
}

Terraform does not automatically rollback in the face of errors.
[...]

PR #10791, opened 2019-11-07
Sort ACM cert subject alternative names and domain validation option

kcburge:
This pull request is similar to, and was based on, #8708. However, it resolves a few issues I discovered with that patch.

The certificate creation process is clearly asynchronous, and, given that the provider is attempting to read properties of an asynchronously created object, it must poll, retrying, until all critical information is available. #8530, however, expects that this object creation succeeds BEFORE validation is complete, so, we cannot wait until the certificate is status succeeded, OR, wait until the domain validation is complete; however, Terraform requires the state to be intact before returning successfully from creation (as I understand it), and about the only way to assure the object is created correctly is to retry, which is what this resource does.

My updates:

- I added a retry in case the subject alternate names was empty.

- I wait to Set the subject alternate names until after we've received all of the domain validation options (if any), so as to prevent side-effects from retrying.

- Like #8708, this patch sorts the SANs and DVOs according to the order in the original request / terraform state file, so that the order is predictable.

This should address issue: #8531.

If this patch is applied, users will be required to either recreate their certificates, OR, manually edit the terraform state files to ensure that the order in the state file reflects the order in their terraform code.

If found three places that must be edited:

- Reorder domain_validation_options

```
"domain_validation_options.0.resource_record_name": "domain.com",
"domain_validation_options.0.resource_record_type": "CNAME",
"domain_validation_options.0.resource_record_value": "...",
```

Replace ".N." in the name with the zero-based index of each domain_validation_options.

- Reorder subject_alternative_names

```
"subject_alternative_names.0": "*.domain.com"
```

Replace ".N" in the name with the zero-based index of each subject_alternative_name.

- Reorder aws_route53_record validation resources:

```
"aws_route53_record.validation.1": {
```

Replace ".N" with the zero-based index of each route 53 record's domain.

Kevin Burge
Nice, Inc. (https://nice.com)

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/aws/resource_aws_acm_certificate.go
+++ b/aws/resource_aws_acm_certificate.go
@@ -258,14 +258,25 @@ func resourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) err
 		d.Set("arn", resp.Certificate.CertificateArn)
 		d.Set("certificate_authority_arn", resp.Certificate.CertificateAuthorityArn)
 
-		if err := d.Set("subject_alternative_names", cleanUpSubjectAlternativeNames(resp.Certificate)); err != nil {
-			return resource.NonRetryableError(err)
+		sans, err := cleanUpSubjectAlternativeNames(resp.Certificate, d)
+		if err != nil {
+			return resource.RetryableError(err)
 		}
 
 		domainValidationOptions, emailValidationOptions, err := convertValidationOptions(resp.Certificate)
 
 		if err != nil {
 			return resource.RetryableError(err)
+		}
+		sortDomainValidationOptions(resp.Certificate.DomainName, sans, domainValidationOptions)
+
+		// subject_alternative_names MUST be set after the above
+		// RetryableError, since sometimes the
+		// subject_alternative_names response from AWS is empty. We
+		// must ensure the domain validation options are good before
+		// actually storing this value, since we read its value up above.
+		if err := d.Set("subject_alternative_names", sans); err != nil {
+			return resource.NonRetryableError(err)
 		}
 
 		if err := d.Set("domain_validation_options", domainValidationOptions); err != nil {
@@ -296,6 +307,7 @@ func resourceAwsAcmCertificateRead(d *schema.ResourceData, meta interface{}) err
 		return nil
 	})
 }
+
 func resourceAwsAcmCertificateGuessValidationMethod(domainValidationOptions []map[string]interface{}, emailValidationOptions []string) string {
 	// The DescribeCertificate Response doesn't have information on what validation method was used
 	// so we need to guess from the validation options we see...
@@ -327,16 +339,43 @@ func resourceAwsAcmCertificateUpdate(d *schema.ResourceData, meta interface{}) e
 	return resourceAwsAcmCertificateRead(d, meta)
 }
 
-func cleanUpSubjectAlternativeNames(cert *acm.CertificateDetail) []string {
+func cleanUpSubjectAlternativeNames(cert *acm.CertificateDetail, d *schema.ResourceData) ([]string, error) {
+	if len(cert.SubjectAlternativeNames) == 0 && aws.StringValue(cert.Status) == acm.DomainStatusPendingValidation {
+		log.Printf("[DEBUG] No subject alternative names, need to retry.")
+		return nil, fmt.Errorf("No subject alternative names, need to retry.")
+	}
+
 	sans := cert.SubjectAlternativeNames
 	vs := make([]string, 0)
-	for _, v := range sans {
-		if aws.StringValue(v) != aws.StringValue(cert.DomainName) {
-			vs = append(vs, aws.StringValue(v))
+	dn := aws.StringValue(cert.DomainName)
+
+	// create unique list of remote sans
+	m := make(map[string]bool)
+	for _, r := range sans {
+		v := aws.StringValue(r)
+		if v != dn {
+			m[v] = true
 		}
 	}
-	return vs
 
+	// add sans in existing order, noting sans we saw
+	for _, r := range d.Get("subject_alternative_names").([]interface{}) {
+		v := strings.TrimSuffix(r.(string), ".")
+		_, found := m[v]
+		if found {
+			delete(m, v)
+			vs = append(vs, v)
+		}
+	}
+
+	// add any remote sans not already in subject_alternative_names
+	for v := range m {
+		if v != dn {
+			vs = append(vs, v)
+		}
+	}
+
+	return vs, nil
 }
 
 func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]interface{}, []string, error) {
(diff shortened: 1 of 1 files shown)


jpSimkins:
Does this also solve the issue with duplicate SSL verfications?

I am referring to the duplicate SSL DNS verification entries for a `domain.com` and  wildcard for same domain `*.domain.com`. 

I am expecting the solution for the ordering to solve this bug as well.

aws/resource_aws_acm_certificate.go
```diff
@@ -378,6 +409,29 @@ func convertValidationOptions(certificate *acm.CertificateDetail) ([]map[string]
 	return domainValidationResult, emailValidationResult, nil
 }
 
+// sortDomainValidationOptions sorts a slice of domain validation options
+// according to the order of domain names in the subject alternatnive names slice
```
fcuello-fudo: "alternatnive" -> "alternative"

kcburge:
I have not tested "duplicate SSL DNS verification entries". Is there an open issue that has example terraform? Without that, I don't know if this addresses that issue as well.

aws/resource_aws_acm_certificate.go
kcburge: I must have missed a typo from the source patch. Already fixed a few of those. Will update. Thanks!

jpSimkins:
> I have not tested "duplicate SSL DNS verification entries". Is there an open issue that has example terraform? Without that, I don't know if this addresses that issue as well.

This was reported on this ticket as this really goes belongs to the ordering issue as well: https://github.com/terraform-providers/terraform-provider-aws/issues/8531#issuecomment-520522948

The solution was to use `-1` on the count to prevent the issue with duplicate SSL verification entries due to `domain.com` and `*.domain.com` having the same values for their verification but creating 2 entries to be validated. The `-1` was a way to remove the duplicate but due to random ordering, this proved to be problematic. 

It seems most efficient to provide this solution in this patch due to the nature of the issue. Ideally, remove the duplicates IMHO.

kcburge:
I can confirm that this should resolve the issue you reported, since the ordering is now "guaranteed".  I agree that the duplicate domain validation option should not be included in the response, but, this would be a significant breaking change to not include it, so, using the -1 workaround seems to work, as in below:

```
locals {
  domains = [
    "domain.com",
    "*.cdn.domain.com",
    "*.assets.cdn.domain.com",
    "*.api.domain.com",
    "*.domain.com",
  ]
}

resource "aws_acm_certificate" "this" {
  domain_name               = "${local.domains[0]}"
  subject_alternative_names = "${slice(local.domains, 1, length(local.domains))}"
  validation_method         = "DNS"
}

resource "aws_route53_record" "validation" {
  # exclude *.domain.com from validation since it is a duplicate
  count = "${length(local.domains)-1}"

  name    = "${lookup(aws_acm_certificate.this.domain_validation_options[count.index], "resource_record_name")}"
  type    = "${lookup(aws_acm_certificate.this.domain_validation_options[count.index], "resource_record_type")}"
  zone_id = "${data.terraform_remote_state.account_base.zoneid}"
  records = ["${lookup(aws_acm_certificate.this.domain_validation_options[count.index], "resource_record_value")}"]
  ttl     = 60
}
```

CumpsD:
Looking forward to seeing this merged, certificates are a hell in Terraform right now with

lorengordon:
It seems really weird that even if I sort `subject_alternative_names` myself, terraform state doesn't keep them in the sorted order. I mean, that's a property of a list, order is important.

```
resource "aws_acm_certificate" "this" {
  domain_name               = local.domains[0]
  subject_alternative_names = sort(slice(local.domains, 1, length(local.domains)))  # <-- not stored in sorted order
  validation_method         = "DNS"
}
```

Hopefully this gets merged. It's super annoying that the provider wants to recreate the cert as it is.

kcburge:
> It seems really weird that even if I sort `subject_alternative_names` myself, terraform state doesn't keep them in the sorted order. I mean, that's a property of a list, order is important.
> 
> ```
> resource "aws_acm_certificate" "this" {
>   domain_name               = local.domains[0]
>   subject_alternative_names = sort(slice(local.domains, 1, length(local.domains)))  # <-- not stored in sorted order
>   validation_method         = "DNS"
> }
> ```
> 
> Hopefully this gets merged. It's super annoying that the provider wants to recreate the cert as it is.

It is not a matter of the provider not keeping them in sorted order - the provider is not preserving *any* order.  It is taking what it receives from the AWS API and storing that in the state. The AWS API provides no guarantee as to the order, so, sometimes it is the same as was requested, sometimes it isn't.  This change tries to provide some assurance that you get out what you put in, so that your associated validation aws_route53_records don't keep getting re-ordered.

CumpsD:
What is required to get this merged?

philtrep:
@fcuello-fudo this is blocking our team as well, is there something we can do to help move this along?

fcuello-fudo:
> @fcuello-fudo this is blocking our team as well, is there something we can do to help move this along?

I have no saying on this, just made a review comment. I'd like to see this solved also.

nathanielram:
I'm going to have to bump this as well. Hopefully can be included in a release soon.

CumpsD:
Spent 3 hours yesterday re-ordering my certs in TF files because of this again, because AWS decided to return them differently and TF wanted to re-create them.

kcburge:
> Spent 3 hours yesterday re-ordering my certs in TF files because of this again, because AWS decided to return them differently and TF wanted to re-create them.

Did this patch not work for you?

CumpsD:
> Did this patch not work for you?

I didn't build TF myself if that's what you mean? Using the latest stable

kcburge:
> > Did this patch not work for you?
> 
> I didn't build TF myself if that's what you mean? Using the latest stable

I see.  You don't need to rebuild terraform to use this patch. You only have to patch and build the terraform aws provider, and then ensure that your terraform code uses your built version rather than pulling the latest (utilizing the plugin cache and provider "version" setting).  This is a bit of a pain, to be sure.  But, it is worth it for us to have our plans run cleanly instead of always reporting false changes.

aaronmell:
@kcburge Can you rerequest the review?

@bflad Can you assist here?

bflad:
Hi folks 👋 As part of a spiked effort against the `aws_acm_certificate` resource (#13053), this pull request was holistically reviewed against all open reports and pull requests for the resource. The issues reported as part of #8531 are certainly a major pain and we want to address them with minimal operator changes where possible.

To this end, our typical implementation with schema attributes where the API returns them unordered is to simply switch them from `TypeList` to `TypeSet` and any Go type assertion updates from `.([]interface{})` to `.(*schema.Set).List()`. This tells the Terraform Plugin SDK to ignore ordering differences of elements by hashing each of their values. We have found this to be the easiest to maintain in the resource code with the minimal changes noted above and requires no configuration updates for operators except if the attribute is referenced elsewhere (e.g. adding wrapping function calls to maintain sorted list behavior 
`sort(tolist(aws_acm_certificate.example.subject_alternative_names))`).

It looks like #11300 is on the desired implementation path already, so to reduce confusion, we are going to close this pull request to focus efforts there. We want to thank @kcburge for the effort put into this contribution and apologies that this situation has been confusing and frustrating.

Please note that we will likely wait until the version 3.0.0 release in the coming weeks to merge this type of change, however, so this change that affects a large portion of configurations can be well documented and easily guarded against in Terraform modules. The newer resource documentation and version 3 upgrade guide will include details about how to switch to the newer (unordered) configuration, using Terraform 0.12's features.

Issue #10688, opened 2019-10-31
Refactor Resource Tagging Code to use keyvaluetags Package


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

## Description

Since its inception, the Terraform AWS Provider has supported [resource tagging](https://aws.amazon.com/answers/account-management/aws-tagging-strategies/) on various Terraform resources. While in theory most resource tagging looks similar to practitioners (a simple key-value mapping), in practice there are underlying differences between how each AWS service API implements tagging support. Another challenge within this space is that the AWS Go SDK implementation itself can differ (e.g. whether a key-value map or struct containing a Key and Value field) and that each AWS service has its own Go types in the latter implementation.

The [internal `keyvaluetags` package was introduced into this codebase](https://github.com/terraform-providers/terraform-provider-aws/pull/10018) to provide a consistent interface for handling AWS resource key-value tags. Using this package simplifies the process of [adding support for resource tagging](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#adding-resource-tagging-support) and will allow for much easier implementation of provider-wide enhancements for tagging support, such as the ability to specify default tags and tags to ignore for all resources under a provider (e.g. https://github.com/terraform-providers/terraform-provider-aws/issues/7926). To sustainably support those enhancements, a refactoring of the existing resources to use the new package needs to occur.

The existing tagging code across AWS services in the Terraform AWS Provider varies on a few dimensions:

- Whether the service uses `map[string]*string` versus an AWS Go SDK specific type (e.g. `[]*eks.Tag`)
- Whether the resource supports tagging on creation or a separate API call after creation
- Whether the resource supports tags as part of reading the resource or requires a separate API call
- Whether the tagging logic is directly within resource functions, in separate functions in the resource file, or separate functions in a service-based `aws/tagsXXX.go` file

Each of these will slightly vary the refactoring instructions, however in principle the highest level of the refactoring process is:

- Per data source, switching the implementation in the `Read` function to using the `keyvaluetags` package
[...]

PR #10806, opened 2019-11-08
keyvaluetags - support Waf Regional 

DrFaust92:
### Community Note

Relates #10688

```release-note
NONE
```


aws/internal/keyvaluetags/generators/listtags/main.go
```diff
@@ -178,11 +181,20 @@ func {{ . | Title }}ListTags(conn {{ . | ClientType }}, identifier string{{ if .
 		return New(nil), err
 	}
 
-	return {{ . | Title }}KeyValueTags(output.{{ . | ListTagsOutputTagsField }}), nil
+	return {{ . | KeyValueTagsPrefix }}KeyValueTags(output.{{ . | ListTagsOutputTagsField }}), nil
```
bflad: Instead of creating more customization in this generator, which would apply to any other weird packages in the AWS Go SDK using cross-package types, I think it might be nice to instead support `WafregionalKeyValueTags()` from the `servicetags` generator just like any other service. 👍 

To support that we can introduce a new customization function such as:

```go
// ServiceTagPackage determines the service tagging tag type value field.
func ServiceTagPackage(serviceName string) string {
	switch serviceName {
	case "wafregional":
		return "waf"
	default:
		return serviceName
	}
}
```

And replace the following lines:

```
func (tags KeyValueTags) {{ . | Title }}Tags() []*{{ . }}.{{ . | TagType }} {

func {{ . | Title }}KeyValueTags(tags []*{{ . }}.{{ . | TagType }}) KeyValueTags {
```

With the following:

```
func (tags KeyValueTags) {{ . | Title }}Tags() []*{{ . | TagPackage }}.{{ . | TagType }} {

func {{ . | Title }}KeyValueTags(tags []*{{ . | TagPackage }}.{{ . | TagType }}) KeyValueTags {
```

aws/internal/keyvaluetags/generators/updatetags/main.go
```diff
@@ -110,6 +111,7 @@ func main() {
 	}
 	templateFuncMap := template.FuncMap{
 		"ClientType":                      keyvaluetags.ServiceClientType,
+		"TagFunctionClass":                ServiceTagFunctionClass,
```
bflad: Similar suggestion as above:

```suggestion
		"TagPackage":                      ServiceTagPackage,
```

DrFaust92:
Thanks for the comments @bflad, it made this things much simpler than i intended to make them.

aws/internal/keyvaluetags/generators/servicetags/main.go
```diff
@@ -182,8 +187,10 @@ package keyvaluetags
 import (
 	"github.com/aws/aws-sdk-go/aws"
 {{- range .SliceServiceNames }}
+{{- if . | IncludeImport }}
```
DrFaust92: i had to add this annoying thing is that i had to add a "filter" for imports the the servicetags generation to skip wafregional import as the package itself is not used in this instance

bflad: We can actually remove the custom function with [`eq`](https://golang.org/pkg/text/template/#hdr-Functions) 👍 

```go
{{- if eq . (. | TagPackage) }}
```

aws/internal/keyvaluetags/generators/listtags/main.go
```diff
@@ -102,6 +103,7 @@ func main() {
 	}
 	templateFuncMap := template.FuncMap{
 		"ClientType":                           keyvaluetags.ServiceClientType,
+		"ListTagPackage":                       keyvaluetags.ServiceTagPackage,
```
bflad: Will rename to `TagPackage` for consistency across all generators.

aws/internal/keyvaluetags/generators/updatetags/main.go
```diff
@@ -110,6 +111,7 @@ func main() {
 	}
 	templateFuncMap := template.FuncMap{
 		"ClientType":                      keyvaluetags.ServiceClientType,
+		"TagFunctionClass":                keyvaluetags.ServiceTagPackage,
```
bflad: Same here -- will rename to `TagPackage` for consistency with other generators.

PR #10813, opened 2019-11-08
Set default description for aws_service_discovery_public_dns_namespace

ad-m:
Unify description to match other resources https://github.com/terraform-providers/terraform-provider-aws/search?l=Go&q=%22Managed+by+Terraform%22 . 

I suppose AWS should review all Terraform resources for AWS and standardize the use of the "Description" field.

Copy of suraj-rajan/terraform-provider-aws#1

### DIFF
--- a/aws/resource_aws_athena_named_query.go
+++ b/aws/resource_aws_athena_named_query.go
@@ -44,6 +44,7 @@ func resourceAwsAthenaNamedQuery() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
+				Default:  "Managed by Terraform",
 			},
 		},
 	}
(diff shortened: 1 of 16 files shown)


ad-m:
I bulk adjusted some of the resources. I cannot test the E2E of this code.

bflad:
Hi @ad-m 👋 Thank you for submitting this.

Adjusting `Default` values represents a breaking change for existing Terraform configurations that do not set a value as the next run would show a difference of `description: "" => "Managed By Terraform"` so this would need to be contained in a major version release and appropriately documented in the major update guide so operators know about the change.

The practice of defaulting descriptions to `Managed by Terraform` as you found is not consistent across all resources in this provider, however we lean towards not including this value in more recent resources since it goes against the API defaults which are usually empty. For existing resources that happen to include this default description, we have left them as is due to our compatibility promises and to reduce operator burden for fairly small benefit.

Since this change would likely cause unexpected operator reaction and we do not anticipate making this type of change in the next major version release, I'm going to close this issue for now. If you would like to discuss this topic more with the maintainers and community, please create a new proposal feature request. 👍 Thanks again.

PR #10839, opened 2019-11-11
Add stale issue closer behavior

aeschright:
### Community Note

Adding HashiBot stale issue behavior: close after 90 days, starting with issues created today.

```release-note

```

```

...
```

### DIFF
--- a/.hashibot.hcl
+++ b/.hashibot.hcl
@@ -16,8 +16,9 @@ poll "stale_issue_closer" "closer" {
     no_reply_in_last = "2160h" # 90 days
     max_issues = 500
     sleep_between_issues = "5s"
-    created_after = "2019-11-11"
-    exclude_labels = ["needs-triage"]
+    created_after = "2019-06-01"
+    exclude_labels = ["needs-triage", "technical-debt"]
+    extra_search_params = "reactions:<20 no:milestone no:assignee"
     message = <<-EOF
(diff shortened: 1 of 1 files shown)


.hashibot.hcl
```diff
@@ -11,6 +11,20 @@ poll "closed_issue_locker" "locker" {
   EOF
 }
 
+poll "stale_issue_closer" "closer" {
+    schedule = "0 50 12 * * *"
```
appilon: this is a brittle aspect of the system 🙈, I'd like to avoid polls all waking at this same time, could we just pick some arbitrary point in the day, say 3:10? `0 10 3 * * *`

.hashibot.hcl
```diff
@@ -11,6 +11,20 @@ poll "closed_issue_locker" "locker" {
...
+    no_reply_in_last = "2160h" # 90 days
+    max_issues = 500
+    sleep_between_issues = "5s"
+    created_after = "2019-11-11"
```
appilon: I'm not certain on why we only want to consider issues after today, but that is outside the context of the review, I'm sure this was very intentional.

aeschright: There are some concerns about inflicting this on people who might have other expectations -- I'll ping the rest of the team to make sure we couldn't go earlier.

.hashibot.hcl
aeschright: Sounds fine, updated

PR #10964, opened 2019-11-21
Add missing required argument `subnet_id`

bigwheel:
Document update.

# Detail

I saw below error.

```
 % terraform apply

Error: Missing required argument

  on all.tf line 58, in resource "aws_nat_gateway" "eks":
  58: resource "aws_nat_gateway" "eks" {

The argument "subnet_id" is required, but no definition was found.

 %
```

and by [test code](https://github.com/terraform-providers/terraform-provider-aws/blob/0321db117feaac9f2e48fafef1a6d651b443ffa8/aws/resource_aws_eip_test.go), I guess subnet_id is required certainly.

### DIFF
--- a/website/docs/r/eip.html.markdown
+++ b/website/docs/r/eip.html.markdown
@@ -97,6 +97,7 @@ resource "aws_eip" "byoip-ip" {
 
 The following arguments are supported:
 
+* `subnet_id` - (Required) The VPC subnet ID.
 * `vpc` - (Optional) Boolean if the EIP is in a VPC or not.
 * `instance` - (Optional) EC2 instance ID.
 * `network_interface` - (Optional) Network interface ID to associate with.


ryndaniels:
Hi @bigwheel - thanks for taking the time to open this. Looking at the error message you shared, the error is coming from the `aws_nat_gateway` resource, which has the `subnet_id` documented already [here](https://www.terraform.io/docs/providers/aws/r/nat_gateway.html). The change you submitted is to the documentation for the EIP resource, which does not have a `subnet_id` argument. As such, we're going to go ahead and close this, but if you run into further problems you can always open a new issue. 🙂

bigwheel:
@ryndaniels Sorry I must confused aws_nat_gateway and whose eip. Thanks for your review.

PR #10971, opened 2019-11-21
d/api_gateway_rest_api - add attributes

DrFaust92:
### Community Note

Relates #10688

```release-note
- data/data_source_api_gateway_rest_api: Add `tags`, `arn`, `description`, `policy`, `api_key_source`, `minimum_compression_size`, `binary_media_types`, `endpoint_configuration` and `execution_arn` attributes
```

```
$ make testacc TESTARGS='-run=TestAccDataSourceAwsApiGatewayRestApi'
--- PASS: TestAccDataSourceAwsApiGatewayRestApi (111.73s)
```

### DIFF
--- a/aws/data_source_aws_api_gateway_rest_api.go
+++ b/aws/data_source_aws_api_gateway_rest_api.go
@@ -5,8 +5,10 @@ import (
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/apigateway"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/terraform-providers/terraform-provider-aws/aws/internal/keyvaluetags"
 )
 
 func dataSourceAwsApiGatewayRestApi() *schema.Resource {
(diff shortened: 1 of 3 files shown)


aws/data_source_aws_api_gateway_rest_api.go
```diff
@@ -21,6 +23,54 @@ func dataSourceAwsApiGatewayRestApi() *schema.Resource {
...
+						},
+						"vpc_endpoint_ids": {
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem:     &schema.Schema{Type: schema.TypeString},
+						},
+					},
+				},
+			},
+			"execution_arn": {
```
bflad: New attribute is missing documentation, will add on merge.

DrFaust92: Sorry about that, this happened to me more than once and got me thinking if we can have some sort of validation for this with a linter. ill take a look if i can hack something.

bflad: Its partially possible today, but requires the output of `terraform providers schema -json`, which is not PR friendly without some Terraform CLI and plugin directory setup in CI. We're hoping to allow the provider binary to output its own form of that JSON instead (with all data; Terraform core does not receive the full schema information from providers) to remove the Terraform CLI and ancillary setup, but that's likely a few weeks out.

I have some centralized tooling issues available here (https://github.com/bflad/tfproviderdocs/issues/15 / https://github.com/bflad/tfproviderdocs/issues/16), but we'd likely also want Markdown AST parsing in that tooling (rather than regular expressions, for example) before implementing a lot of the Markdown checking to prevent false positives.

PR #11005, opened 2019-11-25
resource/aws_datasync_agent: Update deletion error to match updated API deletion error

bflad:
### Community Note

```release-note
* resource/aws_datasync_agent: Trigger resource recreation on updated `InvalidRequestException` error for agents deleted outside Terraform
```

Starting around November 19, 2019 the DataSync API in AWS Commercial regions changed the error messaging for non-existent hosts from including "not found" to this new message:

```
--- FAIL: TestAccAWSDataSyncAgent_basic (158.82s)
    testing.go:696: Error destroying resource! WARNING: Dangling resources

        Error: Check failed: InvalidRequestException: host-0c61c0944a13c5fda does not exist or has been deleted.

--- FAIL: TestAccAWSDataSyncAgent_disappears (116.35s)
    testing.go:635: Step 0 error: errors during follow-up refresh:

        Error: error reading DataSync Agent (arn:aws:datasync:us-west-2:*******:agent/agent-027e5766a4963f977): InvalidRequestException: host-027e5766a4963f977 does not exist or has been deleted.
```

It would be better if the API used a special error code instead so we did not need to rely on the contents of the generic InvalidRequestException error code message.

```
--- PASS: TestAccAWSDataSyncAgent_basic (185.94s)
--- PASS: TestAccAWSDataSyncAgent_disappears (217.60s)
--- PASS: TestAccAWSDataSyncAgent_AgentName (234.13s)
--- PASS: TestAccAWSDataSyncAgent_Tags (286.51s)
```

### DIFF
--- a/aws/resource_aws_datasync_agent.go
+++ b/aws/resource_aws_datasync_agent.go
@@ -150,7 +150,7 @@ func resourceAwsDataSyncAgentCreate(d *schema.ResourceData, meta interface{}) er
 	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
 		_, err := conn.DescribeAgent(descAgentInput)
 
-		if isAWSErr(err, "InvalidRequestException", "not found") {
+		if isAWSErr(err, "InvalidRequestException", "does not exist") {
 			return resource.RetryableError(err)
 		}
 
@@ -180,7 +180,7 @@ func resourceAwsDataSyncAgentRead(d *schema.ResourceData, meta interface{}) erro
 	log.Printf("[DEBUG] Reading DataSync Agent: %s", input)
 	output, err := conn.DescribeAgent(input)
 
-	if isAWSErr(err, "InvalidRequestException", "not found") {
+	if isAWSErr(err, "InvalidRequestException", "does not exist") {
 		log.Printf("[WARN] DataSync Agent %q not found - removing from state", d.Id())
 		d.SetId("")
 		return nil
@@ -269,7 +269,7 @@ func resourceAwsDataSyncAgentDelete(d *schema.ResourceData, meta interface{}) er
 	log.Printf("[DEBUG] Deleting DataSync Agent: %s", input)
 	_, err := conn.DeleteAgent(input)
 
-	if isAWSErr(err, "InvalidRequestException", "not found") {
+	if isAWSErr(err, "InvalidRequestException", "does not exist") {
 		return nil
 	}
 
--- a/aws/resource_aws_datasync_agent_test.go
+++ b/aws/resource_aws_datasync_agent_test.go
@@ -56,7 +56,7 @@ func testSweepDataSyncAgents(region string) error {
 
 			_, err := conn.DeleteAgent(input)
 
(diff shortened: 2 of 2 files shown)


aws/resource_aws_datasync_agent.go
```diff
@@ -150,7 +150,7 @@ func resourceAwsDataSyncAgentCreate(d *schema.ResourceData, meta interface{}) er
 	err = resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
 		_, err := conn.DescribeAgent(descAgentInput)
 
-		if isAWSErr(err, "InvalidRequestException", "not found") {
+		if isAWSErr(err, "InvalidRequestException", "does not exist") {
```
gdavison: Would it make sense to encapsulate this check in a function, rather than doing the string comparison in three places in this file and two more in the tests?

bflad: These `isAWSErr()` checks do tend to need to be copy-pasted like this for every resource (in the `Read` to trigger recreation, `Delete` to ignore on deletion, and `testAccCheck...Destroy()` testing).

It might be nice to think about this problem more globally. The flip side is not trying to introduce abstractions for minor gains or introduce this abstraction in just one or few resources so it becomes inconsistent with the wider codebase.

The quickest abstraction would be something like:

```go
func isDatasyncAgentNotFoundError(err error) bool {
  return isAWSErr(err, "InvalidRequestException", "does not exist")
}
```

But that doesn't feel like we're gaining much. Another (arguably better) abstraction would be if we could define each resource with a mapping of "not found" error codes and messages since that is the crux of what we want to declare:

```go
&AwsResource{ // new wrapper type that extends schema.Resource
  NotFoundErrors: map[string]string{
    "InvalidRequestException": "does not exist",
  },
  Create: // ...
}
```

This would play well with resources that have multiple error conditionals like association resources, e.g. `aws_ec2_client_vpn_network_association`:

```go
&AwsResource{ // new wrapper type that extends schema.Resource
  NotFoundErrors: map[string]string{
    "InvalidClientVpnAssociationId.NotFound": "",
    "InvalidClientVpnEndpointId.NotFound", "",
  },
  Create: // ...
}
```

Then we'd need some function to extract that from the schema and loop through those:

```go
func (r *AwsResource) isNotFoundError(err error) bool {
  for code, message := range r.NotFoundErrors {
    if isAWSErr(err, code, message) {
      return true
    }
  }

  return false
}

// usage
_, err := conn.DoSomething(input)

if resourceAwsDataSyncAgent().isNotFoundError(err) {
  // guessing this could be done better
}
```

Any which way, we'd want to ensure it either compliments/enhances the existing Terraform Plugin SDK `resource.NotFoundError` / `isResourceNotFoundError()` or is not named similarly to prevent confusion.

bflad: @gdavison what would you like me to do here?

gdavison: Yeah, I was thinking of the first case. More of an encapsulation of the logic, rather than anything more general. I'm good either way.

PR #11157, opened 2019-12-05
scaling config values are not required

ddollar:
These attributes are not required in the API. In particular, I don't want to put `desired_size` in my template because then every `apply` would force the cluster back to a static size.

https://docs.aws.amazon.com/eks/latest/APIReference/API_NodegroupScalingConfig.html

### Community Note

Relates OR Closes #0000

```release-note
Change `scaling_config` attributes of `eks_node_group` to be not required to match API.
```

```

...
```

### DIFF
--- a/aws/resource_aws_eks_node_group.go
+++ b/aws/resource_aws_eks_node_group.go
@@ -143,17 +143,17 @@ func resourceAwsEksNodeGroup() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"desired_size": {
 							Type:         schema.TypeInt,
-							Required:     false,
+							Optional:     true,
 							ValidateFunc: validation.IntAtLeast(1),
 						},
 						"max_size": {
 							Type:         schema.TypeInt,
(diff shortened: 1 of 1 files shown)


mklauber:
Fixes: https://github.com/terraform-providers/terraform-provider-aws/issues/11296

ddollar:
I looked into modifying the acceptance tests for a bit but had to timebox it. I have currently found a workaround by using `ignore_changes` on the desired size attribute for these resources.

If anyone else affected by this wants to take this over and write an acceptance test please feel free!

bflad:
Hi again 👋 Since there is no active progress on this at the moment, we are going to close this. If someone is able to verify https://github.com/terraform-providers/terraform-provider-aws/pull/11157#pullrequestreview-361901748 via new acceptance testing, then we'll be happy to take a fresh look. In general though, we would expect `ignore_changes` to not necessarily be a workaround in this case, but potentially an expectation to handle automatically scaled values after creation.

Issue #11204, opened 2019-12-09
Missing ARM_CONTAINER as Environment Type


kabej:
Since 18th november, AWS Code Build support ARM.

https://aws.amazon.com/fr/about-aws/whats-new/2019/11/aws-codebuild-adds-support-for-arm--gpu--and-x-large-compute-typ/

Terraform module performs a validation and doesn't accept ARM_CONTAINER value

> expected environment.0.type to be one of [LINUX_CONTAINER LINUX_GPU_CONTAINER WINDOWS_CONTAINER], got ARM_CONTAINER

kabej:
#11206

bflad:
@kabej next time can you please submit feature requests following the [Feature Request issue template](https://github.com/terraform-providers/terraform-provider-aws/issues/new?labels=enhancement&template=Feature_Request.md) so contributors and maintainers have all the relevant details for searching/implementing? e.g. this affects the `aws_codebuild_project` resource and that the configuration being attempted is something like:

```hcl
resource "aws_codebuild_project" "example" {
  # ... other configuration ...

  environment {
    # ... other configuration ...
    type = "ARM_CONTAINER"
  }
}
```

Thanks so much.

bflad:
Support for this has been merged and will release in version 2.42.0 of the Terraform AWS Provider, later this week. Thanks to @kabej for the implementation. 👍

PR #11206, opened 2019-12-09
add CONTAINER_ARM as valid environment type

kabej:
Fix for #11204


aws/resource_aws_codebuild_project.go
```diff
@@ -212,6 +212,7 @@ func resourceAwsCodeBuildProject() *schema.Resource {
 								codebuild.EnvironmentTypeLinuxContainer,
 								codebuild.EnvironmentTypeLinuxGpuContainer,
 								codebuild.EnvironmentTypeWindowsContainer,
+								codebuild.EnvironmentTypeArmContainer,
```
bflad: Can you please add this new value to the [resource documentation](https://www.terraform.io/docs/providers/aws/r/codebuild_project.html#type-2), e.g. `website/docs/r/codebuild_project.html.markdown` and preferably add a new test (you can copy-paste-modify the `TestAccAWSCodeBuildProject_WindowsContainer` and its associated configuration). Thanks.

kabej: Ok, i do it

aws/resource_aws_codebuild_project_test.go
```diff
@@ -2896,6 +2928,30 @@ resource "aws_codebuild_project" "test" {
...
+resource "aws_codebuild_project" "test" {
+  name         = "%s"
+  service_role = "${aws_iam_role.test.arn}"
+
+  artifacts {
+    type = "NO_ARTIFACTS"
+  }
+
+  environment {
+    compute_type = "BUILD_GENERAL1_MEDIUM"
```
bflad: This test configuration is currently failing:

```
--- FAIL: TestAccAWSCodeBuildProject_ARMContainer (11.15s)
    testing.go:635: Step 0 error: errors during apply:

        Error: Error creating CodeBuild project: InvalidInputException: Compute type BUILD_GENERAL1_MEDIUM is not supported for ARM_CONTAINER projects
```

The [CodeBuild User Guide](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html) notes that `ARM_CONTAINER` is supported by `BUILD_GENERAL1_LARGE`

```suggestion
    compute_type = "BUILD_GENERAL1_LARGE"
```

Please see the Contributing Guide section on running acceptance testing if you need assistance with that: https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#running-an-acceptance-test

aws/resource_aws_codebuild_project_test.go
```diff
@@ -985,6 +985,38 @@ func TestAccAWSCodeBuildProject_WindowsContainer(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t); testAccPreCheckAWSCodeBuild(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSCodeBuildProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSCodeBuildProjectConfig_ARMContainer(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSCodeBuildProjectExists(resourceName, &project),
+					resource.TestCheckResourceAttr(resourceName, "environment.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "environment.2306861956.compute_type", "BUILD_GENERAL1_LARGE"),
```
bflad: The `TypeSet` hash value (`2306861956`) copied here from the other test will likely not work since it will be different for the updated `compute_type` -- Please feel free to just remove these `TestCheckResourceAttr()` calls that include that value since its difficult to get this value in the current acceptance testing framework and Terraform will report any unexpected difference without these.

PR #11280, opened 2019-12-13
Add note about default ASG tags

ryndaniels:
### Community Note

```release-note
NONE
```

```
NONE
```

Screenshot:
Screenshot 2019-12-13 at 10 20 50

### DIFF
--- a/website/docs/r/autoscaling_group.html.markdown
+++ b/website/docs/r/autoscaling_group.html.markdown
@@ -299,7 +299,7 @@ Alternatively the `tags` attributes can be used, which accepts a list of maps co
 This allows the construction of dynamic lists of tags which is not possible using the single `tag` attribute.
 `tag` and `tags` are mutually exclusive, only one of them can be specified.
 
-~> **NOTE:** If no tags are specified, the API will add one tag with a key of `AmazonECSManaged` and a value of `true`.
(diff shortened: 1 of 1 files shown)


website/docs/r/autoscaling_group.html.markdown
```diff
@@ -299,6 +299,8 @@ Alternatively the `tags` attributes can be used, which accepts a list of maps co
 This allows the construction of dynamic lists of tags which is not possible using the single `tag` attribute.
 `tag` and `tags` are mutually exclusive, only one of them can be specified.
 
+~> **NOTE:** If no tags are specified, the API will add one tag with a key of `AmazonECSManaged` and a value of `true`.
```
bflad: Found the documentation on the [ECS tagging](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cluster-auto-scaling.html):

> When using capacity providers with Auto Scaling groups, the `autoscaling:CreateOrUpdateTags` permission is needed on the IAM user creating the capacity provider. This is because Amazon ECS adds a tag to the Auto Scaling group when it associates it with the capacity provider.

Maybe this wording here can help clarify this note and also how to remediate?

```suggestion
~> **NOTE:** Other AWS APIs may automatically add special tags to their associated Auto Scaling Group for management purposes, such as ECS Capacity Providers adding the `AmazonECSManaged` tag. To ignore the removal of these automatic tags, see the [`ignore_tags` provider configuration](https://www.terraform.io/docs/providers/aws/index.html#ignore_tags) or the [`ignore_changes` lifecycle argument for Terraform resources](https://www.terraform.io/docs/configuration/resources.html#ignore_changes).
```

It is also worth mentioning that we _may_ want to consider automatically ignoring AWS managed tag keys like `AmazonECSManaged` like we already do for keys prefixed with `aws:` (e.g. CloudFormation Stack tags, etc), since manually working with them via Terraform seems more problematic.

PR #11294, opened 2019-12-14
Clarify aws_cloudwatch_event_target docs with ECS ARNs

jwr0:
### Community Note

```release-note
NONE
```
The way the documentation was written, the Cloudwatch event target would use an ECS task definition ARN which includes the task definition revision, like `arn:aws:ecs:us-west-2:123456789012:task-definition/some-task:4`. It's likely that eventually a new task definition revision will be created at some time in the future (possibly by Terraform, possibly by a CI/CD pipeline out-of-band from Terraform). If/when that happens, the Cloudwatch event target will still be pointed to the old revision (`some-task:4`). This will cause the task to fail to run when Cloudwatch tries to run it, the RunTask API will reject it because `some-task:4` is no longer active.

The change I'm proposing shows the user how to make the event target get an ARN like `arn:aws:ecs:us-west-2:123456789012:task-definition/some-task`, without a revision number, so Cloudwatch will always run the latest revision of the task definition. Which is probably a more commonly desired behavior anyway.

### DIFF
--- a/website/docs/r/cloudwatch_event_target.html.markdown
+++ b/website/docs/r/cloudwatch_event_target.html.markdown
@@ -204,7 +204,7 @@ resource "aws_iam_role_policy" "ecs_events_run_task_with_any_role" {
         {
             "Effect": "Allow",
             "Action": "ecs:RunTask",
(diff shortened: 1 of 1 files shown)


sthulb:
this can just be closed – it was fixed last year.

PR #11352, opened 2019-12-18
Refactor Kinesis Streams resource and data source to use keyvaluetags package

ewbankkit:
### Community Note

Relates #10688.

```release-note
NONE
```

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSKinesisStreamDataSource'
TF_ACC=1 go test ./aws/ -v -count 1 -parallel 1 -run=TestAccAWSKinesisStreamDataSource -timeout 120m
=== RUN   TestAccAWSKinesisStreamDataSource
=== PAUSE TestAccAWSKinesisStreamDataSource
=== CONT  TestAccAWSKinesisStreamDataSource
--- PASS: TestAccAWSKinesisStreamDataSource (172.41s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	172.498s
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSKinesisStream_basic'
TF_ACC=1 go test ./aws/ -v -count 1 -parallel 1 -run=TestAccAWSKinesisStream_basic -timeout 120m
=== RUN   TestAccAWSKinesisStream_basic
=== PAUSE TestAccAWSKinesisStream_basic
=== CONT  TestAccAWSKinesisStream_basic
--- PASS: TestAccAWSKinesisStream_basic (79.57s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	79.671s
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAWSKinesisStream_Tags'
TF_ACC=1 go test ./aws/ -v -count 1 -parallel 1 -run=TestAccAWSKinesisStream_Tags -timeout 120m
=== RUN   TestAccAWSKinesisStream_Tags
=== PAUSE TestAccAWSKinesisStream_Tags
=== CONT  TestAccAWSKinesisStream_Tags
--- PASS: TestAccAWSKinesisStream_Tags (106.87s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	106.976s
```


aws/internal/keyvaluetags/generators/updatetags/main.go
```diff
@@ -422,6 +432,16 @@ func ServiceTagInputTagsField(serviceName string) string {
 	}
 }
 
+// ServiceTagInputResourceTypeField determines the service tagging tags field requires a map[string]*string.
```
ewbankkit: Yet another permutation for the Kinesis Streams service.
Listing tags returns a `[]*kinesis.Tag` but the tags input to the `AddTags` API requires a `map[string]*string`.
Capture that here via another specialization.

ewbankkit:
Remove _WIP_.
Ready for review.

ewbankkit:
Given that https://github.com/terraform-providers/terraform-provider-aws/pull/11368 has a similar problem with a transformation of the tags passed to the tagging function i replaced the `ServiceTagInputRequiresStringMap` customization with the more general `ServiceTagInputRequiresTransformation` customization.
This requires the addition of a hand-coded transformation function to be added to a new file `service_customizations.go`.
@bflad, I'm open to other ideas.

ewbankkit:
Removed _WIP_ (again).

bflad:
Phew! We are getting into the weeds now with these tagging implementations. Thanks for starting this one and nice work shimming this in here so far. A few notes/open questions -- would love to hear your opinions here!

---

The `kinesis` service is one that [requires](https://github.com/terraform-providers/terraform-provider-aws/issues/4562) [batched tagging](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_AddTagsToStream.html) 🙁 (only 10 tags at a time per API call), so we would need to support batching in the generator template. There's actually other services that have this sort of limitation as well, although I'm forgetting exactly which off the top of my head (sorry I know this isn't too helpful), and we don't have any open bug reports I can quickly find at the moment.

I think for this type of handling we can introduce a generator function that just contains the number of tags per batch, e.g.

```go
func ServiceTagFunctionBatchSize(serviceName string) string {
  switch serviceName {
  case "kinesis":
    return "10"
  default:
    return ""
  }
}
```

And the template can if/else the batching logic based off that. The logic I originally had in `tagsKinesis.go` was a little Go code verbose, so totally understandable if it was switched to something more compact like the batching in #10435 with slice indices.

---

For the TagInputRequiresTransformation bits, wow that API difference is annoying. While the extra `{SERVICE}TagInput()` functions would offer the most flexibility, it might be nice to simplify this to not need abstraction layer either via:

- Making the customization value itself the input value in the template from the customization function, e.g. `return "aws.StringMap(updatedTags.IgnoreAws().Map())"` and

```go
			{{- if . | TagInputCustomValue }}
			{{ . | TagInputTagsField }}:         {{ . | TagInputCustomValue }},
			{{- else }}
```

This has the problem of needing to make sure any changes to the `updatedTags.IgnoreAws()` bits occur in both the template and the new customization function though.

- Making the template function be based on something like "tag input requires map" for dropping the real input value directly since we already support `UntagInputRequiresTagType`:

```go
			{{- if . | TagInputRequiresMap }}
			{{ . | TagInputTagsField }}:         aws.StringMap(updatedTags.IgnoreAws().Map()),
			{{- else }}
```

At first blush, this seemed like a quick win, but I see the problem you mention with Pinpoint being that it needs to be `&pinpoint.TagsModel{Tags: updatedTags.IgnoreAws().PinpointTags()},`. That wouldn't translate well to just being handled in the template though. This was a problem that would needed to be solved [for CloudFront](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudfront/#TagResourceInput) as well.

- Generate the `{SERVICE}TagInput()` and `{SERVICE}UntagInput()` functions for _all_ services and update the generator template to just use those functions as values instead of the if/else conditions in the template

This would make the logic potentially easier to understand since it'd be the same for all services instead of having one-off customization functions either in the generator `main.go` or up a level.

---

I'm not necessarily opposed to these one-off TagInput customization functions since they would allow the ultimate customizability, just food for thought about later folks who need to read and understand the generator. 😄 Open to ideas.

ewbankkit:
Thinking...

ewbankkit:
#### Batch size:

I implemented the `ServiceTagFunctionBatchSize` customization function and associated template changes as suggested.

#### Input transformation:

Overall the suggestion of generating `{SERVICE}TagInput()` and `{SERVICE}UntagInput()` functions for _all_ services seems to be most generic (and we can collapse the `ServiceUntagInputRequiresTagType()` logic into this at the same time) but the implementation seemed to be too invasive - The code that generates the default  `{SERVICE}TagInput()` functions needs to know about `TagType` and Map and Slice services from the `listtags` generator.
I ended up implementing the first option (`TagInputCustomValue`) which is the simplest but should be flexible enough to cover the use cases.

aws/internal/keyvaluetags/generators/updatetags/main.go
```diff
@@ -188,6 +191,33 @@ func {{ . | Title }}UpdateTags(conn {{ . | ClientType }}, identifier string{{ if
...
+				{{- end }}
+			}
+
+			_, err := conn.{{ . | UntagFunction }}(input)
+
+			if err != nil {
+				return fmt.Errorf("error untagging resource (%s): %w", identifier, err)
+			}
+		}
+		{{- else }}
```
bflad: Potential optimization: Given this nice `KeyValueTags.Chunks()` implementation (💯), we might be able to skip redeclaring the majority of the copy-paste template content by re-scoping the `removedTags` and `updatedTags` variables in the loop:

```go
		{{- if . | TagFunctionBatchSize }}
		for _, removedTags := range removedTags.Chunks({{ . | TagFunctionBatchSize }}) {
		{{- end }}
```

The template generator should automatically fix up the formatting whether it should be nested for the looping or not. This doesn't seem worth holding up this merge though. 😄 

We could also support always calling `for _, removedTags := range removedTags.Chunks(-1)` for other services and forego the template conditional, but only adding the looping logic when batching is enabled seems more right.

bflad: PR submitted: https://github.com/terraform-providers/terraform-provider-aws/pull/11502

ewbankkit: @bflad Nice,
I didn't know about the formatting fix up with the template generator and had thought about something like:
```go
{{- if . | TagFunctionBatchSize }}
chunks := removedTags.Chunks({{ . | TagFunctionBatchSize }})
{{- else }}
chunks := []KeyValueTags{removedTags}
```

Issue #11207, opened 2019-12-09
Support for importing AWS Batch resource types


jrr:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Today I attempted to import some AWS Batch resources I'd created in AWS console, only to discover import is not supported for them. The [Resource Importability](https://www.terraform.io/docs/import/importability.html) doc says to file an issue on the provider repo, so here we are.

### New or Affected Resource(s)

Specifically, I tried to import the following resource types:

- `aws_batch_compute_environment`
- `aws_batch_job_queue`
- `aws_batch_job_definition`

And received messages like these:
```
Error: resource aws_batch_compute_environment doesn't support import

Error: resource aws_batch_job_queue doesn't support import

Error: resource aws_batch_job_definition doesn't support import
```

bflad:
Support for importing `aws_batch_compute_environment` resources has been merged and will release with version 2.43.0 of the Terraform AWS Provider, tomorrow. Thanks to @prabusah for the implementation. 👍

bflad:
Support for importing `aws_batch_job_queue` resources has been merged and will release with version 2.44.0 of the Terraform AWS Provider, Thursday this week. Thanks again to @prabusha for the implementation.

The `aws_batch_job_definition` resource will require some additional work before it can properly support import, so it may not be implemented as quick as the above.

bflad:
Support for importing `aws_batch_job_definition` resources has been merged and will release with version 2.48.0 of the Terraform AWS Provider, later today.

PR #11406, opened 2019-12-22
Add import support for aws_batch_job_queue

prabusah:
### Community Note

Relates  #11207 

```
resource/aws_batch_job_queue: support resource import 
```

```
Did not run acceptance test

...
```


aws/resource_aws_batch_job_queue_test.go
```diff
@@ -68,10 +69,15 @@ func TestAccAWSBatchJobQueue_basic(t *testing.T) {
...
 				Config: config,
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckBatchJobQueueExists("aws_batch_job_queue.test_queue", &jq),
+					testAccCheckBatchJobQueueExists(resourceName, &jq),
 					testAccCheckBatchJobQueueAttributes(&jq),
 				),
 			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
```
bflad: This new acceptance testing is currently failing for this test and the others:

```
--- FAIL: TestAccAWSBatchJobQueue_basic (71.48s)
    testing.go:640: Step 1 error: : Provided name  is not supported pattern[a-zA-Z_0-9-]{1,128}
        	status code: 400, request id: fb70a8c2-e8e2-4417-a94e-7f357fda0ebe
```

This is happening because the resource `Read` function is dependent on the `name` attribute being set:

https://github.com/terraform-providers/terraform-provider-aws/blob/b4a971dfc9db81c20df443b0274f2893a976a30c/aws/resource_aws_batch_job_queue.go#L89

The easiest fix here should be to use the resource ID instead of trying to read the `name` attribute, since the [Batch DescribeJobQueues API](https://docs.aws.amazon.com/batch/latest/APIReference/API_DescribeJobQueues.html#Batch-DescribeJobQueues-request-jobQueues) notes support for full ARN in addition to just the name, e.g.

```go
jq, err := getJobQueue(conn, d.Id())
```

aws/resource_aws_batch_job_queue_test.go
```diff
@@ -94,6 +100,11 @@ func TestAccAWSBatchJobQueue_disappears(t *testing.T) {
 				),
 				ExpectNonEmptyPlan: true,
 			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
bflad: Since `_disappears` tests are removing the resource in question to verify Terraform triggers recreation of the resource instead of an error, these tests will generate the below error when trying to do `ImportState` testing:

```
--- FAIL: TestAccAWSBatchJobQueue_disappears (80.44s)
    testing.go:640: Step 1 error: Resource specified by ResourceName couldn't be found: aws_batch_job_queue.test_queue
```

The solution here is just to not include the import testing in "disappears" tests like these. 👍

PR #11422, opened 2019-12-25
docs/r: Change minimum_healthy_hosts to required

LeeXun:
Declare a resource with only deployment_config_name would cause `Error: InvalidMinimumHealthyHostValueException: minimum healthy hosts argument is missing` in terraform v0.12.17. This field should be required or maybe terraform should set a proper default value to 0 if it is not configured.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/r/codedeploy_deployment_config.html.markdown
+++ b/website/docs/r/codedeploy_deployment_config.html.markdown
@@ -95,7 +95,7 @@ The following arguments are supported:
 
 * `deployment_config_name` - (Required) The name of the deployment config.
 * `compute_platform` - (Optional) The compute platform can be `Server`, `Lambda`, or `ECS`. Default is `Server`.
-* `minimum_healthy_hosts` - (Optional) A minimum_healthy_hosts block. Minimum Healthy Hosts are documented below.
+* `minimum_healthy_hosts` - (Required) A minimum_healthy_hosts block. Minimum Healthy Hosts are documented below.
 * `traffic_routing_config` - (Optional) A traffic_routing_config block. Traffic Routing Config is documented below.
 
 The `minimum_healthy_hosts` block supports the following:


website/docs/r/codedeploy_deployment_config.html.markdown
```diff
@@ -95,7 +95,7 @@ The following arguments are supported:
 
 * `deployment_config_name` - (Required) The name of the deployment config.
 * `compute_platform` - (Optional) The compute platform can be `Server`, `Lambda`, or `ECS`. Default is `Server`.
-* `minimum_healthy_hosts` - (Optional) A minimum_healthy_hosts block. Minimum Healthy Hosts are documented below.
+* `minimum_healthy_hosts` - (Required) A minimum_healthy_hosts block. Minimum Healthy Hosts are documented below.
```
bflad: Hi @LeeXun 👋 Thank you for submitting this.

According to the [Terraform resource code](https://github.com/terraform-providers/terraform-provider-aws/blob/b6d453f73fc978ce900198f13df646b13e4717ac/aws/resource_aws_codedeploy_deployment_config.go#L42-L44), [passing Terraform resource testing with the `Lambda` compute platform](https://github.com/terraform-providers/terraform-provider-aws/blob/b6d453f73fc978ce900198f13df646b13e4717ac/aws/resource_aws_codedeploy_deployment_config_test.go#L312-L324), and [CodeDeploy API Reference](https://docs.aws.amazon.com/codedeploy/latest/APIReference/API_CreateDeploymentConfig.html#API_CreateDeploymentConfig_RequestSyntax), this argument is indeed optional and the Terraform documentation should still list it as optional to prevent any confusion. It may be required for certain types of compute types though, which may be good to mention in the argument description instead, e.g.

```suggestion
* `minimum_healthy_hosts` - (Optional) A minimum_healthy_hosts block. Required for `Server` compute platform. Minimum Healthy Hosts are documented below.
```

If you can instead list any requirements in the description and leave the initial `Optional` notation, we can get this in. 👍

bflad:
Hi again @LeeXun 👋 Since we haven't heard back from you regarding the PR feedback, we submitted it via a follow up commit.

LeeXun:
@bflad Ok thanks :)

PR #11428, opened 2019-12-27
New Data Source: aws_workspaces_image

Tensho:
### Community Note

**New Data Source:** `aws_workspaces_image` (#11428)

### Example

```hcl
data "aws_workspaces_image" "example" {
  image_id = "wsi-ten5h0y19"
}
```

### TODO

* [ ] Create `aws_workspaces_image` resource before inspecting it with the described data source

### Acceptance Tests

```
$ AWS_WORKSPACES_IMAGE_ID=wsi-vwtykwqyx AWS_DEFAULT_REGION=us-east-1 make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsWorkspacesImage_basic'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccDataSourceAwsWorkspacesImage_basic -timeout 120m
=== RUN   TestAccDataSourceAwsWorkspacesImage_basic
=== PAUSE TestAccDataSourceAwsWorkspacesImage_basic
=== CONT  TestAccDataSourceAwsWorkspacesImage_basic
--- PASS: TestAccDataSourceAwsWorkspacesImage_basic (30.84s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	34.844s
```

### References

- [Support for AWS Workspaces](#434)


Tensho:
Unfortunately, it's not possible to create an image with the API at the moment, I've requested corresponding API action with AWS Support ticket.

aws/data_source_aws_workspaces_image.go
```diff
@@ -0,0 +1,64 @@
...
+
+	imageID := d.Get("image_id").(string)
+	input := &workspaces.DescribeWorkspaceImagesInput{
+		ImageIds: []*string{aws.String(imageID)},
+	}
+
+	resp, err := conn.DescribeWorkspaceImages(input)
+	if err != nil {
+		return err
+	}
```
ewbankkit: Will probably want to check `len(resp.Images) > 0` here.
See https://github.com/terraform-providers/terraform-provider-aws/pull/11837.

Tensho: Good point, will add

aws/data_source_aws_workspaces_image_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+func TestAccDataSourceAwsWorkspacesImage_basic(t *testing.T) {
+	dataSourceName := "data.aws_workspaces_image.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceAwsWorkspacesImageConfig("wsi-vwtykwqyx"),
```
gdavison, line 18: This won't pass in our testing environments, since the image is owned by your account.

We could add an environment variable for the image ID and have a pre-check function that checks for the environment variable. Alternatively, there are API functions to import AMIs as Workspace images.

Tensho: I like the idea with environment variable, will add the pre-check for now. Unfortunately, [ImportWorkspaceImage API](https://docs.aws.amazon.com/workspaces/latest/api/API_ImportWorkspaceImage.html) works only with Windows. I've submitted respective feature request to AWS Support last year. Is it OK to create `aws_workspace_image` Terraform resource only for Windows, while AWS supports Linux images too?

Tensho:
@gdavison Thank you, I've added fixes according to your review, please let me know if I can do anything else here 🙇

aws/data_source_aws_workspaces_image_test.go
```diff
@@ -0,0 +1,50 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceAwsWorkspacesImageConfig(os.Getenv("AWS_WORKSPACES_IMAGE_ID")),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(dataSourceName, "image_id", "wsi-vwtykwqyx"),
+					resource.TestCheckResourceAttr(dataSourceName, "name", "Windows10Baseline"),
+					resource.TestCheckResourceAttr(dataSourceName, "description", "Google Chrome AWS WorkDocs Drive CrowdStrike Falcon Russian/Ukrainian Language Pack"),
+					resource.TestCheckResourceAttr(dataSourceName, "os", "WINDOWS"),
+					resource.TestCheckResourceAttr(dataSourceName, "required_tenancy", "DEFAULT"),
+					resource.TestCheckResourceAttr(dataSourceName, "state", "AVAILABLE"),
```
gdavison, line 29: Since we're retrieving a WorkSpaces Image using an arbitrary ID, we won't know these values beforehand. You might be able to test this by retrieving the API resource by ID, and then comparing the values in the Terraform resource state. `testAccCheckEfsMountTarget()` is an example of the approach you could take

aws/data_source_aws_workspaces_image_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+data "aws_workspaces_image" "test" {
+  image_id = %q
+}
+`, imageID)
+}
+
+func testAccWorkspacesImagePreCheck(t *testing.T) {
+	if os.Getenv("AWS_WORKSPACES_IMAGE_ID") == "" {
+		t.Fatal("AWS_WORKSPACES_IMAGE_ID env var must be set for AWS WorkSpaces image acceptance tests. This is required until AWS provides ubiquitous (Windows, Linux) import image API.")
```
gdavison, line 48: If the environment variable isn't set, this should skip the test instead of failing

aws/data_source_aws_workspaces_image_test.go
Tensho: `testAccCheckEfsMountTarget()` retrieves actual resource attrubtes within AWS API, but doesn't test them against state 😄  But I got the idea, you meant item 2 at [Anatomy of an Acceptance Test](https://github.com/terraform-providers/terraform-provider-aws/blob/master/docs/contributing/running-and-writing-acceptance-tests.md#anatomy-of-an-acceptance-test) – `testAccCheckCloudWatchDashboardExists()`.

Tensho: Fixed

aws/data_source_aws_workspaces_image_test.go
Tensho: Fixed

Tensho:
@gdavison FIxed, check it out again.

aws/data_source_aws_workspaces_image_test.go
gdavison: Yes, we can do that. We'll just need to document that it only supports Windows images because of the AWS API

aws/data_source_aws_workspaces_image_test.go
```diff
@@ -0,0 +1,115 @@
...
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		attrs := rs.Primary.Attributes
+
+		if attrs["id"] != *image.ImageId {
+			return fmt.Errorf("Expect image ID %q, got %q", *image.ImageId, attrs["id"])
+		}
```
gdavison, line 91: We can do something like the following for each attribute, and then we get the standard checking and error messages

```suggestion
		if err := resource.TestCheckResourceAttr(n, "id", *image.ImageId)(s); err != nil {
			return err
		}
```

aws/data_source_aws_workspaces_image_test.go
```diff
@@ -0,0 +1,115 @@
...
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		conn := testAccProvider.Meta().(*AWSClient).workspacesconn
+		resp, err := conn.DescribeWorkspaceImages(&workspaces.DescribeWorkspaceImagesInput{
+			ImageIds: []*string{aws.String(rs.Primary.ID)},
+		})
+		if err != nil {
+			return fmt.Errorf("Failed describe workspaces images: %s", err)
```
gdavison, line 65: We prefer error messages to use the [Go 1.13 error wrapping verb `%w`](https://blog.golang.org/go1.13-errors#TOC_3.3.)

aws/data_source_aws_workspaces_image.go
```diff
@@ -0,0 +1,69 @@
...
+	conn := meta.(*AWSClient).workspacesconn
+
+	imageID := d.Get("image_id").(string)
+	input := &workspaces.DescribeWorkspaceImagesInput{
+		ImageIds: []*string{aws.String(imageID)},
+	}
+
+	resp, err := conn.DescribeWorkspaceImages(input)
+	if err != nil {
+		return fmt.Errorf("Failed describe workspaces images: %s", err)
```
gdavison, line 54: We prefer error messages to use the [Go 1.13 error wrapping verb `%w`](https://blog.golang.org/go1.13-errors#TOC_3.3.)

aws/data_source_aws_workspaces_image.go
```diff
@@ -0,0 +1,69 @@
...
+				Computed: true,
+			},
+			"required_tenancy": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"state": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
gdavison, line 39: We can remove the `state` attribute, since it is more of a run-time property, not a configuration property.

aws/data_source_aws_workspaces_image.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if len(resp.Images) == 0 {
+		return fmt.Errorf("Workspace image %s was not found", imageID)
+	}
+
+	image := resp.Images[0]
+	d.SetId(imageID)
+	d.Set("name", image.Name)
+	d.Set("description", image.Description)
+	d.Set("os", image.OperatingSystem.Type)
```
gdavison, line 64: The `os` attribute should have its name updated to more closely match the API field, something like `operating_system_type`

aws/data_source_aws_workspaces_image.go
Tensho: But this is a data source, not a resource. Don't we need to get info about any possible attribute there?

Tensho:
@gdavison Adjusted according to your comments. Please let me know if you need anything else 🙇 

I also work on `aws_workspaces_workspace` data source, but I've stuck with [one test](https://github.com/terraform-providers/terraform-provider-aws/pull/14135/files#diff-235eac32d83cb3673643d9ecaf111a7eR48). If you have some time to look there, please could you guide me the right direction 🙇

aws/data_source_aws_workspaces_image_test.go
Tensho: Unfortunately, even Windows images could be imported only with [BYOL](https://docs.aws.amazon.com/workspaces/latest/adminguide/byol-windows-images.html), which looks like a totally separate feature.

Tensho:
@gdavison Please could you check this data source again?

Tensho:
@gdavison @ewbankkit Hi ^_^ Please could you check it again?

aws/data_source_aws_workspaces_image.go
gdavison: Yes, typically we'd like the data source to return as many attributes as possible. In this case, though, `state` is something that can change at any time, even potentially could change one second after Terraform is run. So it doesn't fit with what Terraform should be working with.

gdavison (approved):
LGTM 🚀 

--- PASS: TestAccDataSourceAwsWorkspacesImage_basic (54.32s)

Tensho:
@gdavison Thank you 🙇

PR #11620, opened 2020-01-16
set postgresql to postgres so postgresql is a valid input for field e…

hunkeelin:
### Community Note

This is related to closed issue [4492](https://github.com/terraform-providers/terraform-provider-aws/issues/4492). And related to the open issue [11619](https://github.com/terraform-providers/terraform-provider-aws/issues/11619)

**Summary:** 
The end solution for issue [4492](https://github.com/terraform-providers/terraform-provider-aws/issues/4492) was to simply use `postgres`. If it plans, theoretically it should apply.  If `postgresql` is not a valid input then it should error out during `terraform plan`. This is not the current behavior. The changes set `postgresql` as `postgres`.

### DIFF
--- a/aws/resource_aws_db_instance.go
+++ b/aws/resource_aws_db_instance.go
@@ -863,6 +863,11 @@ func resourceAwsDbInstanceCreate(d *schema.ResourceData, meta interface{}) error
 			}
 		}
 
+		// allow postgresql as input as postgres
+		if strings.ToLower(d.Get("engine").(string)) == "postgresql" {
+			d.Set("engine", "postgres")
+		}
+
 		if attr, ok := d.GetOk("allocated_storage"); ok {
 			modifyDbInstanceInput.AllocatedStorage = aws.Int64(int64(attr.(int)))
 			requiresModifyDbInstance = true


bflad:
Hi @hunkeelin 👋 Thank you for submitting this.

When working with Terraform resource code, it is important that the API, configuration, and state values all match. In this case, the API's value for the PostgreSQL engine is `postgres`, which is the value that the API would respond with and would be set into the Terraform state during the resource `Read` function for drift detection. The implication of this is that the next Terraform plan output would show a destroy/create difference of `"postgres" => "postgresql"` or worse fail the Terraform apply creating the database instance with a `provider produced an inconsistent value` error if there are downstream references to the engine attribute.

Since this change would ultimately not be accepted in principal, I'm going to close this pull request, however if you are interested in having this Terraform resource instead perform plan-time validation of the `engine` value to prevent the wrong value from ever being submitted, it can be accomplished via a [`ValidateFunc` on the attribute](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#validatefunc) with all the values listed in the [`CreateDBInstance` API `Engine` parameter reference](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html), e.g.

```go
			"engine": {
				Type:     schema.TypeString,
				Optional: true,
				Computed: true,
				ForceNew: true,
				StateFunc: func(v interface{}) string {
					value := v.(string)
					return strings.ToLower(value)
				},
					"aurora",
					"aurora-mysql",
					"aurora-postgresql",
					"mariadb",
					"mysql",
					"oracle-ee",
					"oracle-se",
					"oracle-se1",
					"oracle-se2",
					"postgres",
					"sqlserver-ee",
					"sqlserver-ex",
					"sqlserver-se",
					"sqlserver-web",
				}, false)},
			},
```

PR #11837, opened 2020-01-31
r/aws_workspaces_directory: Fix panic if previously registered directory is no longer registered

ewbankkit:
### Community Note

Closes https://github.com/terraform-providers/terraform-provider-aws/issues/11814.

```release-note
resource/aws_workspaces_directory: Fix panic if previously registered directory is no longer registered
```

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccAwsWorkspacesDirectory'
TF_ACC=1 go test ./aws/ -v -count 1 -parallel 20 -run=TestAccAwsWorkspacesDirectory -timeout 120m
=== RUN   TestAccAwsWorkspacesDirectory
=== RUN   TestAccAwsWorkspacesDirectory/basic
=== RUN   TestAccAwsWorkspacesDirectory/disappears
=== RUN   TestAccAwsWorkspacesDirectory/subnetIds
--- PASS: TestAccAwsWorkspacesDirectory (2056.41s)
    --- PASS: TestAccAwsWorkspacesDirectory/basic (662.88s)
    --- PASS: TestAccAwsWorkspacesDirectory/disappears (784.63s)
    --- PASS: TestAccAwsWorkspacesDirectory/subnetIds (608.90s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	2056.456s
```

### DIFF
--- a/aws/resource_aws_workspaces_directory.go
+++ b/aws/resource_aws_workspaces_directory.go
@@ -141,17 +141,22 @@ func resourceAwsWorkspacesDirectoryCreate(d *schema.ResourceData, meta interface
 func resourceAwsWorkspacesDirectoryRead(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).workspacesconn
 
-	resp, err := conn.DescribeWorkspaceDirectories(&workspaces.DescribeWorkspaceDirectoriesInput{
-		DirectoryIds: []*string{aws.String(d.Id())},
-	})
+	raw, state, err := workspacesDirectoryRefreshStateFunc(conn, d.Id())()
 	if err != nil {
-		return err
+		return fmt.Errorf("error getting workspaces directory (%s): %s", d.Id(), err)
 	}
-	dir := resp.Directories[0]
+	if state == workspaces.WorkspaceDirectoryStateDeregistered {
+		log.Printf("[WARN] workspaces directory (%s) not found, removing from state", d.Id())
+		d.SetId("")
+		return nil
+	}
 
+	dir := raw.(*workspaces.WorkspaceDirectory)
 	d.Set("directory_id", dir.DirectoryId)
 	d.Set("subnet_ids", dir.SubnetIds)
-	d.Set("self_service_permissions", flattenSelfServicePermissions(dir.SelfservicePermissions))
+	if err := d.Set("self_service_permissions", flattenSelfServicePermissions(dir.SelfservicePermissions)); err != nil {
+		return fmt.Errorf("error setting self_service_permissions: %s", err)
+	}
 
 	tags, err := keyvaluetags.WorkspacesListTags(conn, d.Id())
 	if err != nil {
@@ -205,7 +210,7 @@ func resourceAwsWorkspacesDirectoryDelete(d *schema.ResourceData, meta interface
 
 	err := workspacesDirectoryDelete(d.Id(), conn)
 	if err != nil {
-		return err
+		return fmt.Errorf("error deleting workspaces directory (%s): %s", d.Id(), err)
 	}
 	log.Printf("[DEBUG] Workspaces directory %q is deregistered", d.Id())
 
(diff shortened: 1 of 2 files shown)


ewbankkit:
Verified that `aws_workspaces_ip_group` resource handles the similar case correctly
https://github.com/terraform-providers/terraform-provider-aws/blob/31e3d47fbfd65d39d61a59b228626e34e394d330/aws/resource_aws_workspaces_ip_group.go#L79-L90

aws/resource_aws_workspaces_directory_test.go
```diff
@@ -181,7 +206,13 @@ func testAccCheckAwsWorkspacesDirectoryDestroy(s *terraform.State) error {
 	return nil
 }
 
-func testAccCheckAwsWorkspacesDirectoryExists(n string) resource.TestCheckFunc {
+func testAccCheckAwsWorkspacesDirectoryDisappears(v *workspaces.WorkspaceDirectory) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		return workspacesDirectoryDelete(aws.StringValue(v.DirectoryId), testAccProvider.Meta().(*AWSClient).workspacesconn)
```
Tensho: It's not clear from the test helper name that it just deletes the resource under the test and doesn't check anything actually. But the framework expects `TestCheckFunc` type, so nothing better to do here 🤷‍♂

aws/resource_aws_workspaces_directory_test.go
```diff
@@ -66,6 +67,7 @@ func TestAccAwsWorkspacesDirectory(t *testing.T) {
 }
 
 func testAccAwsWorkspacesDirectory_basic(t *testing.T) {
+	var v workspaces.WorkspaceDirectory
```
Tensho: Good point 👍 I guess this should be added in a separate PR with the actual usage of `v` in checks. I know it's a good pattern to test TF state along with the actual resource one, however, it doesn't make sense without those tests.

aws/resource_aws_workspaces_directory.go
```diff
@@ -141,17 +141,22 @@ func resourceAwsWorkspacesDirectoryCreate(d *schema.ResourceData, meta interface
...
+		return fmt.Errorf("error getting workspaces directory (%s): %s", d.Id(), err)
+	}
+	if state == workspaces.WorkspaceDirectoryStateDeregistered {
+		log.Printf("[WARN] workspaces directory (%s) not found, removing from state", d.Id())
+		d.SetId("")
+		return nil
 	}
-	dir := resp.Directories[0]
 
+	dir := raw.(*workspaces.WorkspaceDirectory)
```
Tensho: Don't we return `workspaces.WorkspaceDirectory` right from the refresh function? Why it's raw?

aws/resource_aws_workspaces_directory.go
```diff
@@ -141,17 +141,22 @@ func resourceAwsWorkspacesDirectoryCreate(d *schema.ResourceData, meta interface
...
 	d.Set("directory_id", dir.DirectoryId)
 	d.Set("subnet_ids", dir.SubnetIds)
-	d.Set("self_service_permissions", flattenSelfServicePermissions(dir.SelfservicePermissions))
+	if err := d.Set("self_service_permissions", flattenSelfServicePermissions(dir.SelfservicePermissions)); err != nil {
```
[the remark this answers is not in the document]
ewbankkit: If the type of the data passed to `Set()` isn't compatible with the schema type declared for the attribute it's possible to get an error (e.g. passing a `map[string]interface{}` when a `*schema.Set` is expected.
To be safe, anytime a "complex" type is being set we should really check errors (although there are many places in the code that isn't done).
Silent failure can lead to attributes not being set correctly in state.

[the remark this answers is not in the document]
Tensho: Got it, thank you 🙇

PR #11842, opened 2020-01-31
d/aws_vpc_endpoint_service: Correctly handle SageMaker Notebook endpoints

ewbankkit:
### Community Note

Closes https://github.com/terraform-providers/terraform-provider-aws/issues/11806.

```release-note
NONE
```

```
n/a
```

### DIFF
--- a/aws/data_source_aws_vpc_endpoint_service.go
+++ b/aws/data_source_aws_vpc_endpoint_service.go
@@ -45,19 +45,19 @@ func dataSourceAwsVpcEndpointService() *schema.Resource {
 				Computed: true,
 			},
 			"service": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				ConflictsWith: []string{"service_name"},
+				Type:         schema.TypeString,
+				Optional:     true,
+				ExactlyOneOf: []string{"service", "service_name"},
 			},
 			"service_id": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 			"service_name": {
-				Type:          schema.TypeString,
-				Optional:      true,
-				Computed:      true,
-				ConflictsWith: []string{"service"},
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ExactlyOneOf: []string{"service", "service_name"},
 			},
 			"service_type": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


aws/data_source_aws_vpc_endpoint_service.go
```diff
@@ -76,13 +76,14 @@ func dataSourceAwsVpcEndpointServiceRead(d *schema.ResourceData, meta interface{
...
-		return fmt.Errorf(
-			"One of ['service', 'service_name'] must be set to query VPC Endpoint Services")
+	if v := d.Get("service_name").(string); v != "" {
+		serviceName = v
+	} else if v := d.Get("service"); v != "" {
+		if v == "sagemaker.notebook" {
+			serviceName = fmt.Sprintf("aws.sagemaker.%s.notebook", meta.(*AWSClient).region)
+		} else {
+			serviceName = fmt.Sprintf("com.amazonaws.%s.%s", meta.(*AWSClient).region, v)
+		}
```
bflad: Hey @ewbankkit 👋 While supporting `sagemaker.notebook` is a nice enhancement, it does not cover the original bug report configuration:

```hcl
data "aws_vpc_endpoint_service" "sagemaker_notebook" {
  service = "aws.sagemaker.us-east-1.notebook"
}
```

It seems like we may want to allow arbitrary `service` configurations to pass through as-is if they contain multiple periods (or some other loose heuristic) so we don't need to special case a bunch of values should other service teams choose their own standards 🙃.

ewbankkit: Yes, this is a tough one.
The `service_name` field is designed to allow the full reverse-DNS name of a service to be specified and `service = "sagemaker.runtime"` for example just happens to work currently as the AWS service ID is `runtime.sagemaker`.
Let me think about this for a while.
Specifying `service = "aws.sagemaker.us-east-1.notebook"` is the workaround for the specific problem.

ewbankkit: I think the best thing here may be to have no code changes and tighten up the documentation for both the data source and corresponding resource. The underlying AWS API has no notion of just `service` (it deals with `ServiceName`) so this argument is really just for convenience.

website/docs/d/vpc_endpoint.html.markdown
```diff
@@ -43,7 +43,7 @@ More complex filters can be expressed using one or more `filter` sub-blocks,
 which take the following arguments:
 
 * `name` - (Required) The name of the field to filter by, as defined by
-  [the underlying AWS API](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcPeeringConnections.html).
+  [the underlying AWS API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpoints.html).
```
ewbankkit: Correct link.

ewbankkit:
@bflad I have changed to clarifying `service` and `service_name` in the documentation rather than any special-case code changes.

PR #11895, opened 2020-02-04
aws_route53_record: add create,update,delete timeout options for the resource

abhinavdahiya:
### Community Note

ChangeRecordSets now uses the AWS SDKs inbuilt retry method instead of StateConf change retry loop for 5 minutes.

```release-note
* ChangeRecordSets now uses the AWS SDKs inbuilt retry method instead of StateConf change retry loop for 5 minutes.
```

```

...
```

Closes https://github.com/terraform-providers/terraform-provider-aws/issues/11896

### DIFF
--- a/aws/resource_aws_route53_record.go
+++ b/aws/resource_aws_route53_record.go
@@ -344,15 +344,15 @@ func resourceAwsRoute53RecordUpdate(d *schema.ResourceData, meta interface{}) er
 		},
 	}
 
-	req := &route53.ChangeResourceRecordSetsInput{
+	input := &route53.ChangeResourceRecordSetsInput{
 		HostedZoneId: aws.String(cleanZoneID(*zoneRecord.HostedZone.Id)),
 		ChangeBatch:  changeBatch,
 	}
 
 	log.Printf("[DEBUG] Updating resource records for zone: %s, name: %s\n\n%s",
-		zone, *rec.Name, req)
+		zone, *rec.Name, input)
 
-	respRaw, err := changeRoute53RecordSet(conn, req)
+	respRaw, err := changeRoute53RecordSet(conn, input)
 	if err != nil {
 		return fmt.Errorf("[ERR]: Error building changeset: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
Hi @bflad ! thanks for the detailed response and the alternative implementation suggestion..

When i started looking into the adding `PriorRequestNotComplete` as retryable error, i found out that it was already part of the throttle code set in SDK.. https://github.com/aws/aws-sdk-go/blame/3acad1210f5c4feb4866ff9874b5f3f5895bd96d/aws/request/retryer.go#L93 and has been for some time now.

So the logic from https://github.com/terraform-providers/terraform-provider-aws/issues/7873 seemed like a better way to solve the timeout issue. Therefore i updated the functions to use that workflow, allows aws eventual consistency to kick in and then depend on SDK's internal backoff for PriorRequestNotComplete.

PTAL.

aws/resource_aws_route53_record.go
```diff
@@ -754,35 +745,25 @@ func resourceAwsRoute53RecordDelete(d *schema.ResourceData, meta interface{}) er
...
+			return nil
+		}
+		return resource.NonRetryableError(err)
+	})
+	if isResourceTimeoutError(err) {
+		out, err = conn.ChangeResourceRecordSets(input)
+		if isAWSErr(err, "InvalidChangeBatch", "") {
+			// This means the record is already gone
+			return out, nil
+		}
```
bflad: This can be simplified without `resource.Retry()` at all since there are no retryable conditions 👍 

```go
out, err := conn.ChangeResourceRecordSets(input)

if isAWSErr(err, "InvalidChangeBatch", "") {
  return out, nil
}

return out, err
```

See also: https://github.com/terraform-providers/terraform-provider-aws/issues/11864

aws/resource_aws_route53_record.go
```diff
@@ -459,30 +459,21 @@ func resourceAwsRoute53RecordCreate(d *schema.ResourceData, meta interface{}) er
...
-		},
+	var out *route53.ChangeResourceRecordSetsOutput
+	err := resource.Retry(1*time.Minute, func() *resource.RetryError {
+		var err error
+		out, err = conn.ChangeResourceRecordSets(input)
+		if isAWSErr(err, "NoSuchHostedZone", "") {
+			log.Print("[DEBUG] Hosted Zone not found, retrying...")
+			return resource.RetryableError(err)
+		}
+		return resource.NonRetryableError(err)
```
bflad: Nit: In the future we will likely require the usage of `resource.NonRetryableError()`/`resource.RetryableError()` with non-nil errors: 

```suggestion
		if err != nil {
			return resource.NonRetryableError(err)
		}

		return nil
```

See also: https://github.com/hashicorp/terraform-plugin-sdk/pull/199

abhinavdahiya: +1
fixed with https://github.com/terraform-providers/terraform-provider-aws/pull/11895/commits/f5cf401f5cec685033d519c03233bdce3f112173

aws/resource_aws_route53_record.go
abhinavdahiya: @bflad 

+1

fixed with https://github.com/terraform-providers/terraform-provider-aws/pull/11895/commits/f5cf401f5cec685033d519c03233bdce3f112173

Issue #8940, opened 2019-06-11
Creating AWS SES Configuration Set sometimes fails in Terraform 0.12


zipbee:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

`Terraform v0.12.1`

### Affected Resource(s)

* aws_ses_configuration_set

### Terraform Configuration Files

```hcl
provider "aws" {
  alias  = "eu-west-1"
  region = "eu-west-1"
}

resource "aws_ses_configuration_set" "ses_config_set" {
  name     = "client-${var.client_id}-${var.env}"
  provider = aws.eu-west-1
}
```

### Debug Output

Snippets of the full debug output: https://gist.github.com/zipbee/19f2fcd8efc40e79901d50ab3e17f61b
It looks like this happens:
* Request to create ConfigurationSet
* 200 OK response from creation request
* Request to list all configuration sets
* 200 OK response from list request
* Newly created ConfigurationSet is not listed in the output of the list command

This might be a timing issue, e.g. AWS creates it but doesn't instantly list it when requesting a list of all configuration sets?

### Expected Behavior

Resource should be created and successfully tracked.

### Actual Behavior
Apply fails with:
```
Error: Provider produced inconsistent result after apply

When applying changes to aws_ses_configuration_set.ses_config_set, provider
"aws" produced an unexpected new value for was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

Re-running the Apply fails because a the resource already exists.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

We use this in various projects and it sometimes works and sometimes doesn't, everything has been upgraded to Terraform v0.12 and we can't find a clear indicator for what breaks it. We never saw this issue in Terraform 0.11 at least.

zipbee:
We've been trying to narrow down what triggers this issue as we have a lot of instances where this works successfully. So far we can only conclude it is down to the name of the configuration set but can't discern a pattern.

We have some examples that work:
* client-exa-dev
* client-goz-dev
* client-nstest8-dev

And some that definitely don't work:
[...]

PR #12024, opened 2020-02-13
Fix for SES configuration sets state and error handling

carthewd:
Relates to: https://github.com/terraform-providers/terraform-provider-aws/issues/8940

In some cases, when the checking the state for SES Configuration Sets the AWS API call can quietly fail and this results in the resource assuming that none exist and clearing out the state accordingly. In reality, things like authentication errors aren't being handled - in my case, my build environment's role was missing ses:ListConfigurationSets permissions and the resource assumed that no configuration sets existed or found the state/resource to be inconsistent. 

I've refactored this to use DescribeConfigurationSets instead of ListConfigurationSets and iterating through the results to find the specific resource - I couldn't see a need to use the ListConfigurationSets action. This will also throw underlying errors like auth exceptions. 

### Community Note

Relates #8940

```release-note
NONE
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSSESConfigurationSet' -R            
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAWSSESConfigurationSet -timeout 120m
=== RUN   TestAccAWSSESConfigurationSet_basic
=== PAUSE TestAccAWSSESConfigurationSet_basic
=== CONT  TestAccAWSSESConfigurationSet_basic
--- PASS: TestAccAWSSESConfigurationSet_basic (66.84s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       66.866s


...
```

### DIFF
--- a/aws/resource_aws_ses_configuration_set.go
+++ b/aws/resource_aws_ses_configuration_set.go
@@ -50,7 +50,7 @@ func resourceAwsSesConfigurationSetCreate(d *schema.ResourceData, meta interface
 }
 
 func resourceAwsSesConfigurationSetRead(d *schema.ResourceData, meta interface{}) error {
-	conn := meta.(*AWSClient).sesConn
+	conn := meta.(*AWSClient).sesconn
 
 	configSetInput := &ses.DescribeConfigurationSetInput{
 		ConfigurationSetName: aws.String(d.Id()),
--- a/aws/resource_aws_ses_configuration_set.go
+++ b/aws/resource_aws_ses_configuration_set.go
@@ -5,7 +5,6 @@ import (
 	"log"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ses"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
@@ -91,13 +90,13 @@ func findConfigurationSet(name string, meta interface{}) (bool, error) {
 	response, err := conn.DescribeConfigurationSet(configSetInput)
 
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "ConfigurationSetDoesNotExist" {
+		if isAWSErr(err, ses.ErrCodeConfigurationSetDoesNotExistException, "") {
 			return configurationSetExists, nil
 		}
 		return false, err
 	}
 
-	if *response.ConfigurationSet.Name == name {
+	if aws.StringValue(response.ConfigurationSet.Name) == name {
 		configurationSetExists = true
 	}
 
--- a/aws/resource_aws_ses_configuration_set.go
+++ b/aws/resource_aws_ses_configuration_set.go
@@ -50,19 +50,24 @@ func resourceAwsSesConfigurationSetCreate(d *schema.ResourceData, meta interface
 }
 
 func resourceAwsSesConfigurationSetRead(d *schema.ResourceData, meta interface{}) error {
-	configurationSetExists, err := findConfigurationSet(d.Id(), meta)
+	conn := meta.(*AWSClient).sesConn
+
+	configSetInput := &ses.DescribeConfigurationSetInput{
+		ConfigurationSetName: aws.String(d.Id()),
+	}
+
+	response, err := conn.DescribeConfigurationSet(configSetInput)
 
 	if err != nil {
+		if isAWSErr(err, ses.ErrCodeConfigurationSetDoesNotExistException, "") {
+			log.Printf("[WARN] SES Configuration Set (%s) not found, removing from state", d.Id())
+			d.SetId("")
+			return nil
+		}
 		return err
 	}
 
-	if !configurationSetExists {
-		log.Printf("[WARN] SES Configuration Set (%s) not found", d.Id())
-		d.SetId("")
-		return nil
-	}
-
-	d.Set("name", d.Id())
+	d.Set("name", aws.StringValue(response.ConfigurationSet.Name))
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


aws/resource_aws_ses_configuration_set.go
```diff
@@ -78,29 +79,27 @@ func resourceAwsSesConfigurationSetDelete(d *schema.ResourceData, meta interface
...
-		}
-	}
-
-	if err != nil && !configurationSetExists && response.NextToken != nil {
-		configurationSetExists, err = findConfigurationSet(name, response.NextToken, meta)
-	}
+	response, err := conn.DescribeConfigurationSet(configSetInput)
 
 	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "ConfigurationSetDoesNotExist" {
```
anGie44, line 94: hi @carthewd! while this is logically correct, in the provider we're transitioning to using the `isAwsError` method so to consolidate this line we can do something like
```suggestion
    if isAWSErr(err, ses.ErrCodeConfigurationSetDoesNotExistException, "") {}
```

aws/resource_aws_ses_configuration_set.go
```diff
@@ -78,29 +79,27 @@ func resourceAwsSesConfigurationSetDelete(d *schema.ResourceData, meta interface
...
+			return configurationSetExists, nil
+		}
 		return false, err
 	}
 
+	if *response.ConfigurationSet.Name == name {
```
anGie44, line 100: here we should use the AWS SDK's conversion function to safely extract the Name value to avoid any unwanted panics
```suggestion
	if aws.StringValue(response.ConfigurationSet.Name) == name 
```

aws/resource_aws_ses_configuration_set.go
```diff
@@ -78,29 +79,27 @@ func resourceAwsSesConfigurationSetDelete(d *schema.ResourceData, meta interface
 	return err
 }
 
-func findConfigurationSet(name string, token *string, meta interface{}) (bool, error) {
+func findConfigurationSet(name string, meta interface{}) (bool, error) {
```
anGie44, line 82: alternatively this method's logic could be moved to the `Read` method @ L54 since this method is only called by the `Read` method atm and as an added bonus, the response to `DescribeConfigurationSet` can then be used to set the `name` attribute @ L66; in doing so, the error received from the `DescribeConfigurationSet` can also be directly evaluated (as described in the comment below) and returned as needed

aws/resource_aws_ses_configuration_set.go
```diff
@@ -50,18 +51,18 @@ func resourceAwsSesConfigurationSetCreate(d *schema.ResourceData, meta interface
 }
 
 func resourceAwsSesConfigurationSetRead(d *schema.ResourceData, meta interface{}) error {
-	configurationSetExists, err := findConfigurationSet(d.Id(), nil, meta)
+	configurationSetExists, err := findConfigurationSet(d.Id(), meta)
```
anGie44, line 54: see comment below

aws/resource_aws_ses_configuration_set_test.go
```diff
@@ -77,16 +51,18 @@ func testAccCheckAwsSESConfigurationSetExists(n string) resource.TestCheckFunc {
...
 			return err
 		}
 
 		found := false
-		for _, element := range response.ConfigurationSets {
-			if *element.Name == fmt.Sprintf("some-configuration-set-%d", escRandomInteger) {
-				found = true
-			}
+
+		if *response.ConfigurationSet.Name == rs.Primary.ID {
```
anGie44, line 64: we can safely extract this `Name` variable with the AWS SDK's conversion: `aws.StringValue()`

anGie44: in addition, the equality logic could be updated to return immediately if the condition is satisfied e.g.
```
if aws.StringValue(response.ConfigurationSet.Name) != rs.Primary.ID {
    return fmt.Errorf("The configuration set was not created")
}
return nil
````

File: aws/resource_aws_ses_configuration_set_test.go [hunk 6]
```diff
@@ -97,9 +73,34 @@ func testAccCheckAwsSESConfigurationSetExists(n string) resource.TestCheckFunc {
...
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aws_ses_configuration_set" {
+			continue
+		}
+
+		_, err := conn.DescribeConfigurationSet(&ses.DescribeConfigurationSetInput{
+			ConfigurationSetName: aws.String(rs.Primary.ID),
+		})
+
+		if err == nil {
```
anGie44, line 88: small change: to validate the error @ L89, we'll want to check that the err object exists with `err != nil`

aws/resource_aws_ses_configuration_set_test.go
```diff
@@ -97,9 +73,34 @@ func testAccCheckAwsSESConfigurationSetExists(n string) resource.TestCheckFunc {
...
+			if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "ConfigurationSetDoesNotExist" {
```
anGie44, line 89: the `isAWSErr` method in combination with the AWS SDK's error constant can alternatively be used here to consolidate the logic
e.g.
```
    if isAWSErr(err, ses.ErrCodeConfigurationSetDoesNotExistException, "") {}
```

aws/resource_aws_ses_configuration_set_test.go
```diff
@@ -97,9 +73,34 @@ func testAccCheckAwsSESConfigurationSetExists(n string) resource.TestCheckFunc {
...
+				return fmt.Errorf("The configuration set still exists")
```
anGie44, line 90: if the if-logic @ L89 is true, we can `return nil` in this case as it validates that the Resource has been deleted as expected

anGie44: in this resources test file, the Destroy method is a good example to refer to as well https://github.com/terraform-providers/terraform-provider-aws/blob/bafd3e6f90f63049ebf3acbc910f660d79b6f21c/aws/resource_aws_cognito_resource_server_test.go#L140

aws/resource_aws_ses_configuration_set.go
carthewd: Hey @anGie44 - thanks for the great feedback, really appreciate the time and effort. I've addressed all of these... I did leave `findConfigurationSet` in place but I can look at moving that too

anGie44: sure thing @carthewd! thanks for addressing the comments :) re: this function, i think it would be a nice-to-have but definitely up to you as it predates this PR..imo moving the logic to the`Read`method of the resource would improve readability, especially with the error handling in place, and as a rule of thumb we should use the API response values to set schema values like `name`

anGie44: also I see github is reporting merge conflicts so would you mind rebasing when you have a chance?

carthewd: Thanks again @anGie44 - that does make sense, I've refactored it and it does seem a lot cleaner to me. Also fixed up the merge conflicts

PR #12112, opened 2020-02-21
resource/aws_rds_cluster_parameter_group and resource/aws_db_parameter_group: Restore ability to change parameter values

### LINKED ISSUE #11846: aws_rds_cluster_parameter_group change in a parameter reset it
### Community Note

### Terraform Version

⇒  terraform -v
Terraform v0.12.20
+ provider.aws v2.47.0
+ provider.external v1.2.0
+ provider.template v2.1.2
+ provider.vault v2.7.1

### Affected Resource(s)

* aws_rds_cluster_parameter_group

I want to update a parameter value, my plan look like :

  # aws_rds_cluster_parameter_group.default will be updated in-place
  ~ resource "aws_rds_cluster_parameter_group" "default" {
        arn         = "..."
        description = "Managed by Terraform"
        family      = "aurora-mysql5.7"
        id          = "..."
        name        = "..."

      - parameter {
          - apply_method = "immediate" -> null
          - name         = "max_connections" -> null
          - value        = "1000" -> null
        }
      + parameter {
          + apply_method = "immediate"
          + name         = "max_connections"
          + value        = "10000"
        }
    }

but when I apply, it change the max_connections to 10000 and immediatly after do a reset on it,
restoring it the aws default (GREATEST({log(DBInstanceClassMemory/805306368)*45},{log(DBInstanceClassMemory/8187281408)*1000}) )

When the apply is run TF_LOG=debug, I see it doing the modify :
2020/01/31 18:21:01 [DEBUG] [aws-sdk-go] DEBUG: Request rds/ModifyDBClusterParameterGroup Details:
and then immediatly after looping until it can do a reset :
2020/01/31 18:21:01 [DEBUG] [aws-sdk-go] DEBUG: Request rds/ResetDBClusterParameterGroup Details:

I see it also in the event log :

Parameter groups | Fri Jan 31 18:24:02 GMT-500 2020 | Updated  parameter max_connections to  GREATEST({log(DBInstanceClassMemory/805306368)*45},{log(DBInstanceClassMemory/8187281408)*1000})  with apply method immediate
Parameter groups | Fri Jan 31 18:22:17 GMT-500 2020 | Updated parameter max_connections to 10000 with apply method immediate

If I run a plan again, I have :

  # aws_rds_cluster_parameter_group.default will be updated in-place
  ~ resource "aws_rds_cluster_parameter_group" "default" {
        arn         = "..."
        description = "Managed by Terraform"
        family      = "aurora-mysql5.7"
        id          = "..."
        name        = "..."

      + parameter {
          + apply_method = "immediate"
          + name         = "max_connections"
          + value        = "10000"
        }
    }

And a apply will work as it only do modifyDBClusterParameterGroup, but no reset after.

### Expected Behavior

I expect my max_connections changed to 10000

### Actual Behavior

it change max_connections to 10000 and then reset it to the aws default formula.

### Steps to Reproduce

Changing a parameter value in the HCL script and ruin plan/apply.

### References

### LINKED ISSUE #13551: aws_db_parameter_group parameter updates not working
### Community Note

### Terraform Version

Terraform v0.12.26
+ provider.aws v2.64.0
+ provider.dns v2.2.0
+ provider.template v2.1.2

### Affected Resource(s)

* aws_db_parameter_group

```hcl
resource "aws_db_parameter_group" "postgres_10" {
  name        = "psql-pg-test-10"
  family      = "postgres10"
  description = "Postgres parameter group test"

  lifecycle {
    create_before_destroy = true
  }

  parameter {
    name         = "rds.pg_stat_ramdisk_size"
    value        = 256
    apply_method = "pending-reboot"
  }
}
```

### Panic Output

N/A

### Expected Behavior

Parameter group value is updated 

### Actual Behavior

Parameter group remains the same

The plan output looks like this:
```
      - parameter {
          - apply_method = "pending-reboot" -> null
          - name         = "rds.pg_stat_ramdisk_size" -> null
          - value        = "0" -> null
        }
      + parameter {
          + apply_method = "pending-reboot"
          + name         = "rds.pg_stat_ramdisk_size"
          + value        = "256"
        }
```

Reading the code, it sees this a removal of an old parameter and the addition of a new parameter, and then modifies the parameter group with the new value, and then resets the old parameter back to the existing value. Even swapping the order of those two operations would likely fix this issue

### Steps to Reproduce

1. Create an empty parameter group using terraform
2. Update the parameter group to use `rds.pg_stat_ramdisk_size = 256`

### Important Factoids

N/A

### References

* #12742 looks similar

camlow325:
### Community Note

Closes #11846
Closes #13551

```release-note
* resource/aws_rds_cluster_parameter_group and resource/aws_db_parameter_group: Restore ability to change parameter values ([#11846](https://github.com/terraform-providers/terraform-provider-aws/issues/11846))
```

In #11540, support was added to the rds parameter group resources for resetting parameter values to AWS defaults when parameters are removed from config. A side-effect of these changes, however, was that parameters which remain in the config but whose values have been changed would also be reset to AWS defaults. This commit restores the ability for parameter values to be updated in the config while retaining the ability for parameters being removed from the config to be reset to AWS defaults.

```
$  make testacc TEST=./aws TESTARGS='-run=TestAccAWSDBClusterParameterGroup_'
...
--- PASS: TestAccAWSDBClusterParameterGroup_disappears (18.01s)
--- PASS: TestAccAWSDBClusterParameterGroup_namePrefix (23.28s)
--- PASS: TestAccAWSDBClusterParameterGroup_only (23.40s)
--- PASS: TestAccAWSDBClusterParameterGroup_generatedName (23.42s)
--- PASS: TestAccAWSDBClusterParameterGroup_generatedName_Parameter (23.77s)
--- PASS: TestAccAWSDBClusterParameterGroup_namePrefix_Parameter (23.79s)
--- PASS: TestAccAWSDBClusterParameterGroup_withApplyMethod (23.89s)
--- PASS: TestAccAWSDBClusterParameterGroup_basic (86.64s)
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSDBParameterGroup_'
...
--- PASS: TestAccAWSDBParameterGroup_Disappears (18.18s)
--- PASS: TestAccAWSDBParameterGroup_generatedName (22.40s)
--- PASS: TestAccAWSDBParameterGroup_namePrefix (22.69s)
--- PASS: TestAccAWSDBParameterGroup_Only (23.40s)
--- PASS: TestAccAWSDBParameterGroup_MatchDefault (24.19s)
--- PASS: TestAccAWSDBParameterGroup_withApplyMethod (25.23s)
--- PASS: TestAccAWSDBParameterGroup_limit (45.08s)
--- PASS: TestAccAWSDBParameterGroup_basic (75.38s)
```

### DIFF
--- a/aws/resource_aws_db_parameter_group.go
+++ b/aws/resource_aws_db_parameter_group.go
@@ -277,8 +277,25 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
 			}
 		}
 
+		toRemove := map[string]*rds.Parameter{}
+
+		for _, p := range expandParameters(os.List()) {
+			if p.ParameterName != nil {
+				toRemove[*p.ParameterName] = p
+			}
+		}
+
+		for _, p := range expandParameters(ns.List()) {
+			if p.ParameterName != nil {
+				delete(toRemove, *p.ParameterName)
+			}
+		}
+
 		// Reset parameters that have been removed
-		resetParameters := expandParameters(os.Difference(ns).List())
+		var resetParameters []*rds.Parameter
+		for _, v := range toRemove {
+			resetParameters = append(resetParameters, v)
+		}
 		if len(resetParameters) > 0 {
 			maxParams := 20
 			for resetParameters != nil {
--- a/aws/resource_aws_db_parameter_group_test.go
+++ b/aws/resource_aws_db_parameter_group_test.go
@@ -688,6 +688,65 @@ func TestAccAWSDBParameterGroup_MatchDefault(t *testing.T) {
 	})
 }
 
+func TestAccAWSDBParameterGroup_updateParameters(t *testing.T) {
+	var v rds.DBParameterGroup
+	resourceName := "aws_db_parameter_group.test"
+	groupName := fmt.Sprintf("parameter-group-test-terraform-%d", acctest.RandInt())
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSDBParameterGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSDBParameterGroupUpdateParametersInitialConfig(groupName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSDBParameterGroupExists(resourceName, &v),
+					testAccCheckAWSDBParameterGroupAttributes(&v, groupName),
+					resource.TestCheckResourceAttr(resourceName, "name", groupName),
+					resource.TestCheckResourceAttr(resourceName, "family", "mysql5.6"),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_results",
+						"value": "utf8",
+					}),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_server",
+						"value": "utf8",
+					}),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_client",
+						"value": "utf8",
+					}),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			{
+				Config: testAccAWSDBParameterGroupUpdateParametersUpdatedConfig(groupName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSDBParameterGroupExists(resourceName, &v),
+					testAccCheckAWSDBParameterGroupAttributes(&v, groupName),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_results",
+						"value": "ascii",
+					}),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_server",
+						"value": "ascii",
+					}),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_client",
+						"value": "utf8",
+					}),
+				),
+			},
+		},
+	})
+}
+
 func testAccCheckAWSDbParamaterGroupDisappears(v *rds.DBParameterGroup) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		conn := testAccProvider.Meta().(*AWSClient).rdsconn
@@ -1367,6 +1426,54 @@ resource "aws_db_parameter_group" "test" {
     name         = "client_encoding"
     value        = "UTF8"
     apply_method = "pending-reboot"
+  }
+}
+`, n)
+}
+
+func testAccAWSDBParameterGroupUpdateParametersInitialConfig(n string) string {
+	return fmt.Sprintf(`
+resource "aws_db_parameter_group" "test" {
+  name   = "%s"
+  family = "mysql5.6"
+
+  parameter {
+    name  = "character_set_server"
+    value = "utf8"
+  }
+
+  parameter {
+    name  = "character_set_client"
+    value = "utf8"
+  }
+
+  parameter {
+    name  = "character_set_results"
+    value = "utf8"
+  }
+}
+`, n)
+}
+
+func testAccAWSDBParameterGroupUpdateParametersUpdatedConfig(n string) string {
+	return fmt.Sprintf(`
+resource "aws_db_parameter_group" "test" {
+  name   = "%s"
+  family = "mysql5.6"
+
+  parameter {
+    name  = "character_set_server"
+    value = "ascii"
+  }
+
+  parameter {
+    name  = "character_set_client"
+    value = "utf8"
+  }
+
+  parameter {
+    name  = "character_set_results"
+    value = "ascii"
   }
 }
 `, n)
--- a/aws/resource_aws_rds_cluster_parameter_group.go
+++ b/aws/resource_aws_rds_cluster_parameter_group.go
@@ -221,16 +221,33 @@ func resourceAwsRDSClusterParameterGroupUpdate(d *schema.ResourceData, meta inte
 			}
 		}
 
+		toRemove := map[string]*rds.Parameter{}
+
+		for _, p := range expandParameters(os.List()) {
+			if p.ParameterName != nil {
+				toRemove[*p.ParameterName] = p
+			}
+		}
+
+		for _, p := range expandParameters(ns.List()) {
+			if p.ParameterName != nil {
+				delete(toRemove, *p.ParameterName)
+			}
+		}
+
 		// Reset parameters that have been removed
-		parameters = expandParameters(os.Difference(ns).List())
-		if len(parameters) > 0 {
-			for parameters != nil {
+		var resetParameters []*rds.Parameter
+		for _, v := range toRemove {
+			resetParameters = append(resetParameters, v)
+		}
+		if len(resetParameters) > 0 {
+			for resetParameters != nil {
 				parameterGroupName := d.Get("name").(string)
 				var paramsToReset []*rds.Parameter
-				if len(parameters) <= rdsClusterParameterGroupMaxParamsBulkEdit {
-					paramsToReset, parameters = parameters[:], nil
+				if len(resetParameters) <= rdsClusterParameterGroupMaxParamsBulkEdit {
+					paramsToReset, resetParameters = resetParameters[:], nil
 				} else {
-					paramsToReset, parameters = parameters[:rdsClusterParameterGroupMaxParamsBulkEdit], parameters[rdsClusterParameterGroupMaxParamsBulkEdit:]
+					paramsToReset, resetParameters = resetParameters[:rdsClusterParameterGroupMaxParamsBulkEdit], resetParameters[rdsClusterParameterGroupMaxParamsBulkEdit:]
 				}
 
 				resetOpts := rds.ResetDBClusterParameterGroupInput{
--- a/aws/resource_aws_rds_cluster_parameter_group_test.go
+++ b/aws/resource_aws_rds_cluster_parameter_group_test.go
@@ -376,6 +376,65 @@ func TestAccAWSDBClusterParameterGroup_only(t *testing.T) {
 	})
 }
 
+func TestAccAWSDBClusterParameterGroup_updateParameters(t *testing.T) {
+	var v rds.DBClusterParameterGroup
+	resourceName := "aws_rds_cluster_parameter_group.test"
+	groupName := fmt.Sprintf("cluster-parameter-group-test-tf-%d", acctest.RandInt())
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSDBClusterParameterGroupDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSDBClusterParameterGroupUpdateParametersInitialConfig(groupName),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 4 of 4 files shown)


deiga:
@lachlancooper Could this PR get a review? This is a very annoying bug

lachlancooper:
> @lachlancooper Could this PR get a review? This is a very annoying bug

Agreed, the bug has a rather high potential for disastrous impact on production systems. This fix looks sensible to me, but I'm not a maintainer. The contributing docs say a maintainer will review it, but not how to request a review.

@bflad: would you mind taking a look at this (since you reviewed the original PR)?

mikkoc:
Any news on this? We're looking at adding parameter groups, but I'm very hesitant now given what I read.

Olliepop:
> Any news on this? We're looking at adding parameter groups, but I'm very hesitant now given what I read.

I recommend holding off on adding parameter groups until this is merged. Parameter values are reset to the default value on each apply, no matter what you specify. Until this PR is merged you must update the parameter value after applying a Terraform plan within the RDS console or with aws-cli.

The effect on production systems can be disastrous if you have a large number of parameter values which are not the RDS default.

aws/resource_aws_db_parameter_group.go
```diff
@@ -290,8 +290,26 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
 			}
 		}
 
+		// Build a map with names from the new parameter set
+		newNames := make(map[string]bool, ns.Len())
```
dooferlad, line 294: Shouldn't we move the removed / reset logic into the above modify loop so there is one place where a parameter change is managed? That way we can:

```
for , parameter := range parameters {
  if parameter in ns and not in os {
   add
  } else if parameter in ns and in os {
   modify
 } else if paremeter not in ns and in os {
  delete / reset to default
 }
```

The way we have the code now, and after this change, acts on modified values. The bug is that `parameters, err := expandParameters(ns.Difference(os).List())` generates the same set of values as `resetParameters, err := expandParameters(os.Difference(ns).List())`.

dooferlad: Sorry, I was wrong, this is a set difference. Order matters - ns.Difference(os) is the values in ns that are not in os.

camlow325: Yeah, the parameters that have been removed shouldn't be caught by the `parameters, err := expandParameters(ns.Difference(os).List())` call which precedes the block which processes modified parameters since none of the removed parameters are in `ns`. As @lachlancooper explained really well in https://github.com/terraform-providers/terraform-provider-aws/issues/11846#issuecomment-586777896, the `resetParameters, err := expandParameters(os.Difference(ns).List())` call would catch any parameters whose name or other attributes (including value) appear in `os` but not in `ns`. This leads to parameters which have the same name in `os` and `ns` but different values getting unintentionally reset, which is what this PR intends to fix.

aws/resource_aws_db_parameter_group.go
```diff
@@ -277,8 +277,26 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
 			}
 		}
 
+		// Build a map with names from the new parameter set
```
dooferlad, line 280: I would be tempted to go with:

```go
		toRemove := map[string]*rds.Parameter{}

		for _, p := range expandParameters(os.List()) {
			toRemove[*p.ParameterName] = p
		}

		for _, p := range expandParameters(ns.List()) {
			delete(toRemove, *p.ParameterName)
		}

		// Reset parameters that have been removed
		var resetParameters []*rds.Parameter
		for _, v := range toRemove {
			resetParameters = append(resetParameters, v)
		}
```

It is a bit more obvious and lets `expandParameters` care about the format of the interface it receives.

Ray-B:
Would be good to see this out as soon as is reasonable. This bug is a show-stopper for us, and I suspect many others who work in larger environments.

Without this fix, it effectively bricks our ability to make configuration changes to databases under RDS, particularily if you have limitations on your ability to make manual configuration changes to production environments / have complex deployment workflows, limitations on rollback, etc.

wgorski:
Any news on this? We are already using parameter groups on production environments and this bug makes it very difficult to apply any changes.

aws/resource_aws_rds_cluster_parameter_group_test.go
```diff
@@ -176,6 +176,25 @@ func TestAccAWSDBClusterParameterGroup_basic(t *testing.T) {
...
+						resourceName, "parameter.3127312348.value", "ascii"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.422535354.name", "character_set_server"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.422535354.value", "ascii"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.2478663599.name", "character_set_client"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.2478663599.value", "utf8"),
+				),
```
anGie44, line 196: these Set indices unfortunately are no longer supported in acceptance testing since the recent upgrade to V2 of the terraform-plugin-sdk, but they can be easily updated to use the new convention:
```go
tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
    "name": "character_set_results",
    "value": "ascii"", 
}),
```
where the attribute that is a Set is provided with an `.*` suffix like `parameter.*` and the map[string]string provided corresponds to the key/value pairs expected

aws/resource_aws_rds_cluster_parameter_group_test.go
```diff
@@ -176,6 +176,25 @@ func TestAccAWSDBClusterParameterGroup_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.3127312348.value", "ascii"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.422535354.name", "character_set_server"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.422535354.value", "ascii"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.2478663599.name", "character_set_client"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.2478663599.value", "utf8"),
```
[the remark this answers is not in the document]
camlow325: Thanks, I made those updates.

aws/resource_aws_db_parameter_group_test.go
```diff
@@ -163,6 +163,25 @@ func TestAccAWSDBParameterGroup_basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.3127312348.value", "ascii"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.422535354.name", "character_set_server"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.422535354.value", "ascii"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.2478663599.name", "character_set_client"),
+					resource.TestCheckResourceAttr(
+						resourceName, "parameter.2478663599.value", "utf8"),
```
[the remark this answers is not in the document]
camlow325: Thanks, I made those updates.

camlow325:
Thanks for all of the review comments, @anGie44.

> Overall, I would just recommend creating a new test with each of the new configs to test the new "update" behavior so we ensure we are not introducing a regression and we can attempt to keep the basic test as one that should only test for the minimum required fields, though understandably over time it has expanded quite a bit.

Sorry, just saw this comment after I reworked my previous test updates to use the `TestCheckTypeSetElemNestedAttrs` helper.`Update` had seemed like a part of `Basic` functionality for the resource but I agree that this test has a fair amount of other assertions that it's doing already. I can split this out into a new test.
 
> With regard to the code additions in both the db_parameter_group and rds_cluster_parameter_group resources, since the debug logs show that the ordering of the `Update` behavior transitions from `Modifying` then to `Reset` of the parameters, we could instead flip around that existing logic and that should allow a user to modify a param as intended without resetting soon after. This wouldn't require additional logic then what is already present in our codebase, just a shift in the ordering, but if you have any additional feedback or questions please let me know!

This would be less turmoil than what I had added for the PR. @lachlancooper mentioned in https://github.com/terraform-providers/terraform-provider-aws/issues/11846#issuecomment-586777896:

> Assuming we want to keep the functionality of resetting parameters that are being removed, one extremely simple fix would be to swap the order of the modify and reset operations so that modify always "wins". That's still not ideal as the operations would happen non-atomically, meaning every changing parameter would get the wrong value for a brief period.

I agree with this comment and this is the reason why I had gone with the approach of ensuring that parameters that have changed are only updated once during an apply. I worry that by just flipping the logic that the initial reset of a parameter to its default value rather than the intended new value could have negative repercussions - especially for "apply immediately" parameters or if a network issue were to happen between the "reset" and "modify" that could indefinitely leave one or more parameters unintentionally reset to their default values.

aws/resource_aws_db_parameter_group.go
camlow325: Thanks for this writeup, @dooferlad. This does seem like an improvement. If @anGie44 agrees, I'd be happy to swap in your implementation.

camlow325:
> Overall, I would just recommend creating a new test with each of the new configs to test the new "update" behavior so we ensure we are not introducing a regression and we can attempt to keep the basic test as one that should only test for the minimum required fields, though understandably over time it has expanded quite a bit.

In 31e7b8f, I moved the new `TestStep` entries out into new test funcs - `TestAccAWSDBParameterGroup_updateParameters` `TestAccAWSDBClusterParameterGroup_updateParameters` - with separate configs for each.

anGie44:
> by just flipping the logic that the initial reset of a parameter to its default value rather than the intended new value could have negative repercussions

I think this is a great point @camlow325 ! and definitely well illustrated in @lachlancooper's comment. I'd go ahead with @dooferlad's https://github.com/terraform-providers/terraform-provider-aws/pull/12112#discussion_r454467536 . I would just nitpick on the use of `*p.ParameterName` , to avoid any unwanted panics (not very likely from the API response, but we have had cases in the past where struct fields can be `nil`) we can use the AWS SDK Go method `aws.StringValue()` to safely get the pointer's value.

aws/resource_aws_rds_cluster_parameter_group.go
```diff
@@ -221,16 +221,33 @@ func resourceAwsRDSClusterParameterGroupUpdate(d *schema.ResourceData, meta inte
...
+
 		// Reset parameters that have been removed
-		parameters = expandParameters(os.Difference(ns).List())
-		if len(parameters) > 0 {
-			for parameters != nil {
+		var resetParameters []*rds.Parameter
+		for _, v := range toRemove {
+			resetParameters = append(resetParameters, v)
+		}
+		if len(resetParameters) > 0 {
```
camlow325, line 243: The changes from here on down in this file just rename the `parameters` variable to `resetParameters` in order to avoid conflicting with the previously defined `parameters` variable and make this code look more like the code in the `resource_aws_db_parameter_group.go` file.

camlow325:
> I'd go ahead with @dooferlad's #12112 (comment) . I would just nitpick on the use of *p.ParameterName , to avoid any unwanted panics (not very likely from the API response, but we have had cases in the past where struct fields can be nil) we can use the AWS SDK Go method aws.StringValue() to safely get the pointer's value.

I swapped in the new implementation in 1296450. For the `p.ParameterName` references, I just wrapped them with some `nil` checks to avoid adding an entry to or referencing an entry in the map for a `nil` value. That seemed a little better than just using a potential empty string key from `aws.StringValue` with the map. In any event, a `nil` value for `p.ParameterValue` would seem to be a sign of something bad/unintended.

aws/resource_aws_db_parameter_group.go
```diff
@@ -277,8 +277,25 @@ func resourceAwsDbParameterGroupUpdate(d *schema.ResourceData, meta interface{})
...
+			if p.ParameterName != nil {
+				toRemove[*p.ParameterName] = p
+			}
+		}
+
+		for _, p := range expandParameters(ns.List()) {
+			if p.ParameterName != nil {
+				delete(toRemove, *p.ParameterName)
+			}
+		}
```
anGie44, line 292: nice 👍 I forgot since technically we're expanding user input here (in the rds cluster resource) and the "name"  field within the `parameter` block is `Required`, we shouldn't ever come across a `nil` (as well i see the `exandParameters` actually skips for cases "name" is an empty string) but i prefer erring on the side of caution when it comes to pointers.

jamesmharvey:
@anGie44 I know it's a long shot, but is there any chance of getting this released as a pointfix in a v2.70.0? There are a significant number of breaking changes in v3.0.0, and we're still working through them.

Issue #11046, opened 2019-11-27
AWS Managed Rules for AWS WAF


ewbankkit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

AWS WAF announces AWS Managed Rules (AMRs), a set of AWS WAF rules curated and maintained by the AWS Threat Research Team.

### New or Affected Resource(s)

* aws_XXXXX

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

[Announcement](https://aws.amazon.com/about-aws/whats-new/2019/11/introducing-aws-managed-rules-for-aws-waf/).
[Blog post](https://aws.amazon.com/blogs/aws/announcing-aws-managed-rules-for-aws-waf/).

New API version, `wafv2`; See https://github.com/aws/aws-sdk-go/pull/2976.

Requires [AWS SDK v1.25.42](https://github.com/aws/aws-sdk-go/releases/tag/v1.25.42):
* https://github.com/terraform-providers/terraform-provider-aws/pull/10955

bflad:
Submitted the wafv2 service client and tagging implementation to kick things off here: https://github.com/terraform-providers/terraform-provider-aws/pull/11172

I believe [WAFv2 APIs are wholly independent of previous WAF "Classic" resources](https://docs.aws.amazon.com/waf/latest/APIReference/Welcome.html), so this issue may need to turn into multiple feature requests to support other WAFv2 resources as well like Web ACL for parity with the previous ones.

bflad:
Split out other related WAFv2 feature requests:
* `aws_wafv2_ip_set` [resource](https://github.com/terraform-providers/terraform-provider-aws/issues/11173) / [data source](https://github.com/terraform-providers/terraform-provider-aws/issues/11178)
[...]

PR #12119, opened 2020-02-21
New resource r/aws_wafv2_ip_set

### LINKED ISSUE #11173: Feature Request: WAFv2 IP Set Resource
### Community Note

### Description

* aws_wafv2_ip_set

```hcl
resource "aws_wafv2_ip_set" "example" {
  addresses          = [""]
  description        = ""
  ip_address_version = ""
  name               = ""
  scope              = ""
  tags               = {}
}
```

### References

pvanbuijtene:
### Community Note

Closes #11173
Relates #11046

```release-note
* **New Resource**: `aws_wafv2_ip_set` (#12119)
```

```
make testacc TEST=./aws TESTARGS='-run=TestAccAwsWafv2IPSet_*'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAwsWafv2IPSet_* -timeout 120m
...
--- PASS: TestAccAwsWafv2IPSet_minimal (14.02s)
--- PASS: TestAccAwsWafv2IPSet_changeNameForceNew (24.21s)
--- PASS: TestAccAwsWafv2IPSet_basic (32.15s)
--- PASS: TestAccAwsWafv2IPSet_tags (36.29s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       36.850s
```


aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+		if err != nil {
+			return resource.NonRetryableError(fmt.Errorf("Error getting lock token: %s", err))
+		}
+
+		u := &wafv2.UpdateIPSetInput{
+			Id:          aws.String(d.Id()),
+			Name:        aws.String(d.Get("name").(string)),
+			Scope:       aws.String(d.Get("scope").(string)),
+			Addresses:   []*string{},
+			Description: aws.String(d.Get("description").(string)),
```
dessite:
```suggestion
```

Field is optional, if no description was provided during create, trying to update will fail with
```
Error: Error updating WAFV2 IPSet: InvalidParameter: 1 validation error(s) found.
- minimum field size of 1, UpdateIPSetInput.Description.
```

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+
+	params := &wafv2.GetIPSetInput{
+		Id:    aws.String(d.Id()),
+		Name:  aws.String(d.Get("name").(string)),
+		Scope: aws.String(d.Get("scope").(string)),
+	}
+
+	resp, err := conn.GetIPSet(params)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == wafv2.ErrCodeWAFNonexistentItemException {
```
DrFaust92: can you refactor this to:
 `if isAWSErr(err, wafv2.ErrCodeWAFNonexistentItemException, ") {`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+
+		return err
+	}
+
+	d.Set("name", resp.IPSet.Name)
+	d.Set("description", resp.IPSet.Description)
+	d.Set("ip_address_version", resp.IPSet.IPAddressVersion)
+	d.Set("arn", resp.IPSet.ARN)
+
+	if err := d.Set("addresses", schema.NewSet(schema.HashString, flattenStringList(resp.IPSet.Addresses))); err != nil {
```
DrFaust92: use `flattenStringSet(resp.IPSet.Addresses)`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+
+func resourceAwsWafv2IPSetUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafv2conn
+	var resp *wafv2.GetIPSetOutput
+	params := &wafv2.GetIPSetInput{
+		Id:    aws.String(d.Id()),
+		Name:  aws.String(d.Get("name").(string)),
+		Scope: aws.String(d.Get("scope").(string)),
+	}
+	log.Printf("[INFO] Updating WAFV2 IPSet %s", d.Id())
```
DrFaust92: you should add a `lock_token` computed field and get it from the state instead of calling `GetIpSet`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+
+	err := resource.Retry(15*time.Minute, func() *resource.RetryError {
+		var err error
+		resp, err = conn.GetIPSet(params)
+		if err != nil {
+			return resource.NonRetryableError(fmt.Errorf("Error getting lock token: %s", err))
+		}
```
DrFaust92: see previous comment on `lock_token`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+
+		u := &wafv2.UpdateIPSetInput{
+			Id:          aws.String(d.Id()),
+			Name:        aws.String(d.Get("name").(string)),
+			Scope:       aws.String(d.Get("scope").(string)),
+			Addresses:   []*string{},
+			Description: aws.String(d.Get("description").(string)),
+			LockToken:   resp.LockToken,
```
DrFaust92: see previous comment on `lock_token`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+	})
+
+	if isResourceTimeoutError(err) {
+		_, err = conn.DeleteIPSet(&wafv2.DeleteIPSetInput{
+			Id:        aws.String(d.Id()),
+			Name:      aws.String(d.Get("name").(string)),
+			Scope:     aws.String(d.Get("scope").(string)),
+			LockToken: resp.LockToken,
+		})
+	}
```
DrFaust92: why delete if update times out?

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,302 @@
...
+
+func resourceAwsWafv2IPSetDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafv2conn
+	var resp *wafv2.GetIPSetOutput
+	params := &wafv2.GetIPSetInput{
+		Id:    aws.String(d.Id()),
+		Name:  aws.String(d.Get("name").(string)),
+		Scope: aws.String(d.Get("scope").(string)),
+	}
+	log.Printf("[INFO] Deleting WAFV2 IPSet %s", d.Id())
```
DrFaust92: see previous comment on `lock_token`

aws/resource_aws_wafv2_ip_set.go
pvanbuijtene: error between trackpad and seat 😬

pvanbuijtene:
@dessite @DrFaust92 thanks for the review

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+				d.Set("scope", scope)
+				return []*schema.ResourceData{d}, nil
+			},
+		},
+
+		Schema: map[string]*schema.Schema{
+			"addresses": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				MinItems: 1,
```
anGie44, line 43:
```suggestion
```
hi @pvanbuijtene 👋, we'll want to allow for an empty list here else if a user provides in a config `addresses = []`, there will an error at plan time to enforce the 1 item minimum

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+		IPAddressVersion: aws.String(d.Get("ip_address_version").(string)),
+		Name:             aws.String(d.Get("name").(string)),
+		Scope:            aws.String(d.Get("scope").(string)),
+	}
+
+	if v, ok := d.GetOk("addresses"); ok && v.(*schema.Set).Len() > 0 {
+		params.Addresses = expandStringSet(d.Get("addresses").(*schema.Set))
+	}
+
+	if d.HasChange("description") {
```
anGie44, line 104: alternatively, a similar `GetOk` pattern used in L100 can be used here

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+		params.Description = aws.String(d.Get("description").(string))
+	}
+
+	if v := d.Get("tags").(map[string]interface{}); len(v) > 0 {
+		params.Tags = keyvaluetags.New(v).IgnoreAws().Wafv2Tags()
+	}
+
+	err := resource.Retry(15*time.Minute, func() *resource.RetryError {
```
anGie44, line 112: in my testing, I found these retries can be omitted across the different CRUD operations but if still needed could we move this usage to `config.go`? will make for better re-usability in this resource as well as future ones to come :)

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+
+	return nil
+}
+
+func resourceAwsWafv2IPSetUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafv2conn
+
+	log.Printf("[INFO] Updating WAFV2 IPSet %s", d.Id())
+
+	err := resource.Retry(15*time.Minute, func() *resource.RetryError {
```
anGie44, line 187: same comment as above re: `resource.Retry`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+		u := &wafv2.UpdateIPSetInput{
```
anGie44, line 188: we can move this assignment block outside this func to make it accessible @L216

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+
+	return resourceAwsWafv2IPSetRead(d, meta)
+}
+
+func resourceAwsWafv2IPSetDelete(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafv2conn
+
+	log.Printf("[INFO] Deleting WAFV2 IPSet %s", d.Id())
+
+	err := resource.Retry(15*time.Minute, func() *resource.RetryError {
```
anGie44, line 243: same comment as above re: `resource.Retry`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+		_, err := conn.DeleteIPSet(&wafv2.DeleteIPSetInput{
```
anGie44, line 244: similar comment to one above re: moving assignment block out of this func

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,277 @@
...
+
+	params := &wafv2.GetIPSetInput{
+		Id:    aws.String(d.Id()),
+		Name:  aws.String(d.Get("name").(string)),
+		Scope: aws.String(d.Get("scope").(string)),
+	}
+
+	resp, err := conn.GetIPSet(params)
+	if err != nil {
+		if isAWSErr(err, wafv2.ErrCodeWAFNonexistentItemException, "AWS WAF couldn’t perform the operation because your resource doesn’t exist") {
```
anGie44, line 151: nice! can we also add a test case that uses `testAccCheckResourceDisappears` to cover this behavior?

aws/resource_aws_wafv2_ip_set.go
pvanbuijtene: I added these retries based on the documentation: https://docs.aws.amazon.com/waf/latest/APIReference/API_CreateIPSet.html
The retry func is added to `config.go`

aws/resource_aws_wafv2_ip_set.go
pvanbuijtene: yes we can :) ... hope it's correct

pvanbuijtene:
@anGie44 thanks for the review, the changes are applied.
```
...
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAwsWafv2IPSet -timeout 120m
...
--- PASS: TestAccAwsWafv2IPSet_Disappears (14.87s)
--- PASS: TestAccAwsWafv2IPSet_Minimal (22.19s)
--- PASS: TestAccAwsWafv2IPSet_Basic (35.51s)
--- PASS: TestAccAwsWafv2IPSet_ChangeNameForceNew (37.66s)
--- PASS: TestAccAwsWafv2IPSet_Tags (53.83s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       54.968s
```

aws/config.go
```diff
@@ -759,3 +761,39 @@ func GetSupportedEC2Platforms(conn *ec2.EC2) ([]string, error) {
 
 	return platforms, nil
 }
+
+type wafv2CallFunc func() (interface{}, error)
+
+// Calls a WAFv2 action and retries on a retryable error
+func callWafv2WithRetry(f wafv2CallFunc) (interface{}, error) {
+	var out interface{}
+	err := resource.Retry(15*time.Minute, func() *resource.RetryError {
```
bflad, line 770: Hi @pvanbuijtene 👋 Thank you for updating this! Our intention was to suggest marking the `wafv2.ErrCodeWAFInternalErrorException` and `wafv2.ErrCodeWAFTagOperationInternalErrorException` errors as retryable at the service client layer, since these seem like they should be globally retryable for any WAFv2 API call. This allows the AWS Go SDK to automatically perform the retries instead of our own retry logic via `resource.Retry()`. For example a few lines up from here:

https://github.com/terraform-providers/terraform-provider-aws/blob/07f8ab992c780c4911832234afbbf67feea2fe66/aws/config.go#L697-L710

Other errors like `wafv2.ErrCodeWAFAssociatedItemException` should keep the `resource.Retry()` logic in the resources directly, so it is clear what we are retrying on and for how long. 👍

We'd recommend not retrying `wafv2.ErrCodeWAFAssociatedItemException` for more than 5 minutes normally, since it may be unrecoverable due to a misconfiguration or lack of configuration. Also it'd be good to know why `wafv2.ErrCodeWAFTagOperationException` is included in this retry logic (potentially saved as a code comment if staying), since that seems like an unusual addition.

pvanbuijtene: @bflad totally missed that, the retry implementation is changed and looks cleaner now.

> Also it'd be good to know why wafv2.ErrCodeWAFTagOperationException is included in this retry logic

WAFv2 supports tag on create and these errors can be the result of a create action according to the documentation, I added this in a comment.

anGie44: @pvanbuijtene, apologies the confusion! my wording on updating the retry logic to the patterns used in the client was unclear. thank you for the quick refactor  🙏

aws/resource_aws_wafv2_ip_set_test.go
```diff
@@ -0,0 +1,372 @@
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSWafv2IPSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAwsWafv2IPSetConfig(ipSetName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSWafv2IPSetExists(resourceName, &r),
+					testAccCheckAWSWafv2IPSetDisappears(resourceName),
```
anGie44, line 77: this looks great! i missed mentioning you can also leverage this generic method defined here: https://github.com/terraform-providers/terraform-provider-aws/blob/07f8ab992c780c4911832234afbbf67feea2fe66/aws/provider_test.go#L590

it's recent to the codebase but a good example can be referred to here:
https://github.com/terraform-providers/terraform-provider-aws/blob/07f8ab992c780c4911832234afbbf67feea2fe66/aws/resource_aws_backup_plan_test.go#L453

aws/config.go
pvanbuijtene: @anGie44 no apologies needed, I didn't notice the already present retry logic in that file.

aws/resource_aws_wafv2_ip_set_test.go
pvanbuijtene: I picked the wrong example :) ... it's updated

bflad: Sorry this is in my backlog to add into the contributing guide, we are just working on refactoring the documentation beforehand this week. Thank you for being flexible.

DrFaust92:
Can you add an ipv6 test case? (compressed and uncompressed ) AWS APIs are really not consistent on that and it can cause a perpetual diff similar to what v1 resources

pvanbuijtene:
@DrFaust92 I have added a test, let me know if you want to change or add something.
```
--- PASS: TestAccAwsWafv2IPSet_Disappears (18.91s)
--- PASS: TestAccAwsWafv2IPSet_IPv6 (20.25s)
--- PASS: TestAccAwsWafv2IPSet_Minimal (33.56s)
--- PASS: TestAccAwsWafv2IPSet_ChangeNameForceNew (33.58s)
--- PASS: TestAccAwsWafv2IPSet_Basic (33.75s)
--- PASS: TestAccAwsWafv2IPSet_Tags (51.29s)
```

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -148,7 +151,7 @@ func resourceAwsWafv2IPSetRead(d *schema.ResourceData, meta interface{}) error {
 
 	resp, err := conn.GetIPSet(params)
 	if err != nil {
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == wafv2.ErrCodeWAFNonexistentItemException {
+		if isAWSErr(err, wafv2.ErrCodeWAFNonexistentItemException, "AWS WAF couldn’t perform the operation because your resource doesn’t exist") {
```
anGie44, line 154: in case the returned error's `Message` string changes in the API, it'll be safer to omit the message string to match against
```suggestion
		if isAWSErr(err, wafv2.ErrCodeWAFNonexistentItemException, "") {
```

aws/resource_aws_wafv2_ip_set_test.go
```diff
@@ -0,0 +1,400 @@
...
+		conn := testAccProvider.Meta().(*AWSClient).wafv2conn
+		resp, err := conn.GetIPSet(
+			&wafv2.GetIPSetInput{
+				Id:    aws.String(rs.Primary.ID),
+				Name:  aws.String(rs.Primary.Attributes["name"]),
+				Scope: aws.String(rs.Primary.Attributes["scope"]),
+			})
+
+		if err == nil {
+			if *resp.IPSet.Id == rs.Primary.ID {
```
anGie44, line 254: we should use the AWS conversion func here on `resp.IPSet.Id`

aws/resource_aws_wafv2_ip_set_test.go
```diff
@@ -0,0 +1,400 @@
...
+				return fmt.Errorf("WAFV2 IPSet %s still exists", rs.Primary.ID)
+			}
+		}
+
+		// Return nil if the IPSet is already destroyed
+		if awsErr, ok := err.(awserr.Error); ok {
```
anGie44, line 260: we should use the `isAWSErr` checking method here instead

aws/resource_aws_wafv2_ip_set_test.go
```diff
@@ -0,0 +1,400 @@
...
+			Id:    aws.String(rs.Primary.ID),
+			Name:  aws.String(rs.Primary.Attributes["name"]),
+			Scope: aws.String(rs.Primary.Attributes["scope"]),
+		})
+
+		if err != nil {
+			return err
+		}
+
+		if *resp.IPSet.Id == rs.Primary.ID {
```
anGie44, line 294: similar comment as above to use AWS SDK's conversion function to extract the response's Id

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,255 @@
...
+		params.Tags = keyvaluetags.New(v).IgnoreAws().Wafv2Tags()
+	}
+
+	resp, err := conn.CreateIPSet(params)
+
+	if err != nil {
+		return fmt.Errorf("Error creating WAFv2 IPSet: %s", err)
+	}
+
+	d.SetId(*resp.Summary.Id)
```
anGie44, line 137: we should use the AWS SDK's conversion function here as well around `resp.Summary.Id`

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,255 @@
...
+	d.Set("description", resp.IPSet.Description)
+	d.Set("ip_address_version", resp.IPSet.IPAddressVersion)
+	d.Set("arn", resp.IPSet.ARN)
+	d.Set("lock_token", resp.LockToken)
+
+	if err := d.Set("addresses", flattenStringSet(resp.IPSet.Addresses)); err != nil {
+		return fmt.Errorf("Error setting addresses: %s", err)
+	}
+
+	tags, err := keyvaluetags.Wafv2ListTags(conn, *resp.IPSet.ARN)
```
anGie44, line 173: similar comment as above but re: this ARN field

aws/resource_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return fmt.Errorf("Error listing tags for WAFv2 IpSet (%s): %s", *resp.IPSet.ARN, err)
```
anGie44, line 175: see above comment

pvanbuijtene:
@anGie44 I fixed and added some extra checks, let me know if I missed something.

website/docs/r/wafv2_ip_set.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - A unique identifier for the set.
+* `arn` - The Amazon Resource Name (ARN) that identifies the cluster.
+
+## Import
+
+WAFv2 IP Sets can be imported using their ID, Name and Scope e.g.
```
anGie44, line 50: Nit to make import syntax into a formatted block w/delimiter (just to extra communicate the syntax if not evident by example)
```suggestion
WAFv2 IP Sets can be imported using `ID/name/scope`
```

aws/resource_aws_wafv2_ip_set.go
anGie44: @pvanbuijtene one last pointer conversion :)

anGie44:
Many thanks @pvanbuijtene! one catch in the resource re: pointer conversion and a suggestion to the documentation to match general formatting used around the provider's resource docs 

Output of acceptance tests:
```
--- PASS: TestAccAwsWafv2IPSet_Minimal
--- PASS: TestAccAwsWafv2IPSet_Disappears
--- PASS: TestAccAwsWafv2IPSet_Basic
--- PASS: TestAccAwsWafv2IPSet_ChangeNameForceNew
--- PASS: TestAccAwsWafv2IPSet_Tags
--- PASS: TestAccAwsWafv2IPSet_IPv6
```

aws/resource_aws_wafv2_ip_set.go
pvanbuijtene: aaah my IDE hid it from me 😬

PR #12221, opened 2020-03-02
[#9209] Fix dx gateway association proposal so it works even if expired

### LINKED ISSUE #9209: aws_dx_gateway_association_proposal recreated
## Terraform bug

It happens when associating a virtual private gateway with a direct connect gateway in a separate account ([See AWS doc](https://docs.aws.amazon.com/directconnect/latest/UserGuide/multi-account-associate-vgw.html)). There is a [terraform resource](https://www.terraform.io/docs/providers/aws/r/dx_gateway_association_proposal.html) for the proposal. But this resource is not permanent, so terraform will try to recreate it.

### Terraform Version

- terraform 0.11.10
- terraform-provider-aws 2.12.0

### Affected Resource(s)

* aws_dx_gateway_association_proposal
* aws_dx_gateway_association

```hcl
resource "aws_dx_gateway_association_proposal" "dcgw_association_proposal" {
  depends_on                  = ["aws_vpn_gateway_attachment.vpn_attachment"]
  dx_gateway_id               = "${local.dcgw_id}"
  dx_gateway_owner_account_id = "${local.shared_services_account_id}"
  associated_gateway_id       = "${aws_vpn_gateway.vpn_gw.id}"
}

resource "aws_dx_gateway_association" "dcgw_association" {
  provider = "aws.shared_services"

  proposal_id                         = "${aws_dx_gateway_association_proposal.dcgw_association_proposal.id}"
  dx_gateway_id                       = "${local.dcgw_id}"
  associated_gateway_owner_account_id = "${local.account_id}"

  timeouts {
    delete = "30m"
    create = "30m"
  }
}
```
### Expected Behavior

- Apply: create proposal, associate in other account, networks are connected
- Any apply after that: No change

### Actual Behavior

- Apply: create proposal, associate in other account
- Apply immediately: No change
- Apply after a few days once the proposal has been removed by AWS: create proposal, force recreation of association, networks are disconnected for up to 20min -> **DOWNTIME**

### Steps to Reproduce

1. `terraform apply`
2. Wait for a few days or delete the proposal ([Example](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/directconnect.html#DirectConnect.Client.delete_direct_connect_gateway_association_proposal))
3. `terraform plan`: shows creation of the proposal and recreation of the association

### Important Factoids

Raised a support case with AWS to understand the behaviour. Here is the response:

*I relayed the same question to the service team and they have informed me that the proposal will be available as long as the "accepter" has not accepted the proposal. Once the proposal is accepted, it will disappear after some time. There is no defined time duration as to how long the proposal is available after acceptance through CLI. As we use a distributed and eventual consistency model, I would highly recommend you to not use the time the proposal remains in the CLI output after acceptance to make any design, architectural, or application decisions.*

blckct:
### Community Note

Currently the proposal expires after some time which causes TF to want to recreate it and recreate everything that depends on it so it's pretty useless. This change makes the resource ignore that API doesn't return anything and keep returning the last known state. It seems to work with the expired proposals I have access to however it would be great if someone could help with testing or give some insight into what may be problems with heavy handed approach like this one. 
Closes #9209

```
resource/aws_dx_gateway_association_proposal: Return last known state for dx_gateway_associaton_proposal if expired
```

I haven't been able to run the tests, as I'm waiting for my second account. Would be nice if someone else could run them.

### DIFF
--- a/aws/resource_aws_dx_gateway_association_proposal.go
+++ b/aws/resource_aws_dx_gateway_association_proposal.go
@@ -31,7 +31,11 @@ func resourceAwsDxGatewayAssociationProposal() *schema.Resource {
 					return false
 				}
 
-				return proposal != nil && aws.StringValue(proposal.ProposalState) == directconnect.GatewayAssociationProposalStateRequested
+				if proposal == nil {
+					return true
+				}
+
+				return aws.StringValue(proposal.ProposalState) == directconnect.GatewayAssociationProposalStateRequested
 			}),
 		),
 
@@ -122,8 +126,7 @@ func resourceAwsDxGatewayAssociationProposalRead(d *schema.ResourceData, meta in
 	}
 
 	if proposal == nil {
-		log.Printf("[WARN] Direct Connect Gateway Association Proposal (%s) not found, removing from state", d.Id())
-		d.SetId("")
+		//The resource may have expired, return the last known state
 		return nil
 	}
 
--- a/website/docs/r/dx_gateway_association_proposal.html.markdown
+++ b/website/docs/r/dx_gateway_association_proposal.html.markdown
@@ -10,6 +10,8 @@ description: |-
 
 Manages a Direct Connect Gateway Association Proposal, typically for enabling cross-account associations. For single account associations, see the [`aws_dx_gateway_association` resource](/docs/providers/aws/r/dx_gateway_association.html).
 
+~> **NOTE:** The proposal expires after some time so the resource keeps returning the last known state. If you need to create it again you have to taint it.
+
 ## Example Usage
 
 ```hcl


ewbankkit:
@blckct Thanks for this.
Have you written acceptance test(s)? If so could you please add them to the PR and I can take a look; otherwise if you need help, let me know.

blckct:
@ewbankkit Hey, I didn't run existing tests as I had a bit of a trouble setting up an alternative account that's needed for it. Do you mind running those for me?
As for the acceptance tests...Well, it's a bit of a problem since from what I know it expires after a few days. I'll have to check if it can be deleted.

Omarimcblack:
@blckct currently this has been working great for me, the only caveat being when you then need to delete an expired(accepted) proposal terraform bombs out. I came up with a fix that works for me https://github.com/blckct/terraform-provider-aws/pull/1

blckct:
@Omarimcblack this has been picked up by @ewbankkit in #12482 so I'm not sure this code will be used. Have you checked if his code is working for you?

Omarimcblack:
> @Omarimcblack this has been picked up by @ewbankkit in #12482 so I'm not sure this code will be used. Have you checked if his code is working for you?

@blckct unfortunately for my use-case handling from the `aws_dx_gateway_association` only is not enough. Where we are using cross account association we must use both resources `aws_dx_gateway_association` & `aws_dx_gateway_association_proposal` therefore both persist in our state and we cannot have the `aws_dx_gateway_association_proposal` wishing to be recreated every time it has been accepted and terraform runs again(regardless of whether the `proposal_id` is ignored by the `aws_dx_gateway_association` or not).

understandable handling the resource this way, if you ever manually delete a proposal or its expired for a reason other than you accepting, you will need to run a `state rm`. i'm very much open to hearing any other ways of handling this but for now i believe this PR to still be relevant

ewbankkit:
@blckct Thanks for the contribution :tada: :clap:.
The solution (🤞) to this problem is being addressed in #19741.

PR #12283, opened 2020-03-06
Fix error handling for alternative errors

eldondev:
### Community Note

```release-note
Improve EBS Snapshot copy error handling
```


aws/resource_aws_ebs_snapshot_copy.go
```diff
@@ -114,6 +114,8 @@ func resourceAwsEbsSnapshotCopyRead(d *schema.ResourceData, meta interface{}) er
 		log.Printf("Snapshot %q Not found - removing from state", d.Id())
 		d.SetId("")
 		return nil
+	} else if err != nil {
+		return err
```
bflad: Since the previous `if` condition has a `return`, the `else` is extraneous. It might also be helpful for operators and code maintainers if there was some contextual information provided about the error as well.

```suggestion
	}

	if err != nil {
		return fmt.Errorf("error describing EC2 Snapshot (%s): %w", d.Id(), err)
```

PR #12284, opened 2020-03-06
New resource r/aws_wafv2_regex_pattern_set

### LINKED ISSUE #11174: Feature Request: WAFv2 Regex Pattern Set Resource
### Community Note

### Description

* aws_wafv2_regex_pattern_set

```hcl
resource "aws_wafv2_regex_pattern_set" "example" {
  description             = ""
  name                    = ""
  regular_expression_list = [""]
  scope                   = ""
  tags                    = {}
}
```

### References

pvanbuijtene:
### Community Note

Closes #11174
Relates #11046

```release-note
* **New Resource**: `aws_wafv2_regex_pattern_set` (#12284)
```

```
--- PASS: TestAccAwsWafv2RegexPatternSet_Disappears (17.37s)
--- PASS: TestAccAwsWafv2RegexPatternSet_Minimal (19.88s)
--- PASS: TestAccAwsWafv2RegexPatternSet_Basic (35.35s)
--- PASS: TestAccAwsWafv2RegexPatternSet_ChangeNameForceNew (35.46s)
--- PASS: TestAccAwsWafv2RegexPatternSet_Tags (42.81s)
```

### DIFF
--- a/aws/resource_aws_wafv2_regex_pattern_set.go
+++ b/aws/resource_aws_wafv2_regex_pattern_set.go
@@ -65,7 +65,7 @@ func resourceAwsWafv2RegexPatternSet() *schema.Resource {
 						"regex_string": {
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringLenBetween(1, 512),
+							ValidateFunc: validation.StringLenBetween(1, 200),
 						},
 					},
 				},
--- a/aws/resource_aws_wafv2_regex_pattern_set.go
+++ b/aws/resource_aws_wafv2_regex_pattern_set.go
@@ -56,7 +56,7 @@ func resourceAwsWafv2RegexPatternSet() *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.StringLenBetween(1, 128),
 			},
-			"regular_expression_list": {
+			"regular_expression": {
 				Type:     schema.TypeSet,
 				Optional: true,
 				MaxItems: 10,
@@ -96,7 +96,7 @@ func resourceAwsWafv2RegexPatternSetCreate(d *schema.ResourceData, meta interfac
 		params.Description = aws.String(v.(string))
 	}
 
-	if v, ok := d.GetOk("regular_expression_list"); ok && v.(*schema.Set).Len() > 0 {
+	if v, ok := d.GetOk("regular_expression"); ok && v.(*schema.Set).Len() > 0 {
 		params.RegularExpressionList = expandWafv2RegexPatternSet(v.(*schema.Set).List())
 	}
 
@@ -110,7 +110,11 @@ func resourceAwsWafv2RegexPatternSetCreate(d *schema.ResourceData, meta interfac
 		return fmt.Errorf("Error creating WAFv2 RegexPatternSet: %s", err)
 	}
 
-	d.SetId(*resp.Summary.Id)
+	if resp == nil || resp.Summary == nil {
+		return fmt.Errorf("Error creating WAFv2 RegexPatternSet")
+	}
+
+	d.SetId(aws.StringValue(resp.Summary.Id))
 
 	return resourceAwsWafv2RegexPatternSetRead(d, meta)
 }
@@ -131,22 +135,25 @@ func resourceAwsWafv2RegexPatternSetRead(d *schema.ResourceData, meta interface{
 			d.SetId("")
 			return nil
 		}
-
 		return err
 	}
 
-	d.Set("name", resp.RegexPatternSet.Name)
-	d.Set("description", resp.RegexPatternSet.Description)
-	d.Set("arn", resp.RegexPatternSet.ARN)
-	d.Set("lock_token", resp.LockToken)
-
-	if err := d.Set("regular_expression_list", flattenWafv2RegexPatternSet(resp.RegexPatternSet.RegularExpressionList)); err != nil {
-		return fmt.Errorf("Error setting regular_expression_list: %s", err)
+	if resp == nil || resp.RegexPatternSet == nil {
+		return fmt.Errorf("Error getting WAFv2 RegexPatternSet")
 	}
 
-	tags, err := keyvaluetags.Wafv2ListTags(conn, *resp.RegexPatternSet.ARN)
+	d.Set("name", aws.StringValue(resp.RegexPatternSet.Name))
+	d.Set("description", aws.StringValue(resp.RegexPatternSet.Description))
+	d.Set("arn", aws.StringValue(resp.RegexPatternSet.ARN))
+	d.Set("lock_token", aws.StringValue(resp.LockToken))
+
+	if err := d.Set("regular_expression", flattenWafv2RegexPatternSet(resp.RegexPatternSet.RegularExpressionList)); err != nil {
+		return fmt.Errorf("Error setting regular_expression: %s", err)
+	}
+
+	tags, err := keyvaluetags.Wafv2ListTags(conn, aws.StringValue(resp.RegexPatternSet.ARN))
 	if err != nil {
-		return fmt.Errorf("Error listing tags for WAFv2 RegexPatternSet (%s): %s", *resp.RegexPatternSet.ARN, err)
+		return fmt.Errorf("Error listing tags for WAFv2 RegexPatternSet (%s): %s", aws.StringValue(resp.RegexPatternSet.ARN), err)
 	}
 
 	if err := d.Set("tags", tags.IgnoreAws().IgnoreConfig(ignoreTagsConfig).Map()); err != nil {
@@ -169,7 +176,7 @@ func resourceAwsWafv2RegexPatternSetUpdate(d *schema.ResourceData, meta interfac
 		RegularExpressionList: []*wafv2.Regex{},
 	}
 
-	if v, ok := d.GetOk("regular_expression_list"); ok && v.(*schema.Set).Len() > 0 {
+	if v, ok := d.GetOk("regular_expression"); ok && v.(*schema.Set).Len() > 0 {
 		params.RegularExpressionList = expandWafv2RegexPatternSet(v.(*schema.Set).List())
 	}
 
@@ -205,7 +212,6 @@ func resourceAwsWafv2RegexPatternSetDelete(d *schema.ResourceData, meta interfac
 	}
 	err := resource.Retry(5*time.Minute, func() *resource.RetryError {
 		_, err := conn.DeleteRegexPatternSet(params)
-
 		if err != nil {
 			if isAWSErr(err, wafv2.ErrCodeWAFAssociatedItemException, "") {
 				return resource.RetryableError(err)
@@ -236,7 +242,6 @@ func expandWafv2RegexPatternSet(l []interface{}) []*wafv2.Regex {
 		if regexPattern == nil {
 			continue
 		}
-
 		regexPatterns = append(regexPatterns, expandWafv2Regex(regexPattern.(map[string]interface{})))
 	}
 
@@ -254,13 +259,20 @@ func expandWafv2Regex(m map[string]interface{}) *wafv2.Regex {
 }
 
 func flattenWafv2RegexPatternSet(r []*wafv2.Regex) interface{} {
-	regexPatterns := make([]interface{}, len(r))
+	if r == nil {
+		return []interface{}{}
+	}
 
-	for i, regexPattern := range r {
-		d := map[string]interface{}{
-			"regex_string": *regexPattern.RegexString,
+	regexPatterns := make([]interface{}, 0)
+
+	for _, regexPattern := range r {
+		if regexPattern == nil {
+			continue
 		}
-		regexPatterns[i] = d
+		d := map[string]interface{}{
+			"regex_string": aws.StringValue(regexPattern.RegexString),
+		}
+		regexPatterns = append(regexPatterns, d)
 	}
 
 	return regexPatterns
--- a/aws/resource_aws_wafv2_regex_pattern_set.go
+++ b/aws/resource_aws_wafv2_regex_pattern_set.go
@@ -58,8 +58,8 @@ func resourceAwsWafv2RegexPatternSet() *schema.Resource {
 			},
 			"regular_expression_list": {
 				Type:     schema.TypeSet,
-				Required: true,
-				MinItems: 1,
+				Optional: true,
+				MaxItems: 10,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"regex_string": {
--- a/aws/resource_aws_wafv2_regex_pattern_set.go
+++ b/aws/resource_aws_wafv2_regex_pattern_set.go
@@ -245,22 +245,11 @@ func resourceAwsWafv2RegexPatternSetUpdate(d *schema.ResourceData, meta interfac
 
 func resourceAwsWafv2RegexPatternSetDelete(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).wafv2conn
-	var resp *wafv2.GetRegexPatternSetOutput
-	params := &wafv2.GetRegexPatternSetInput{
-		Id:    aws.String(d.Id()),
-		Name:  aws.String(d.Get("name").(string)),
-		Scope: aws.String(d.Get("scope").(string)),
-	}
+
 	log.Printf("[INFO] Deleting WAFV2 RegexPatternSet %s", d.Id())
 
 	err := resource.Retry(15*time.Minute, func() *resource.RetryError {
-		var err error
-		resp, err = conn.GetRegexPatternSet(params)
-		if err != nil {
-			return resource.NonRetryableError(fmt.Errorf("Error getting lock token: %s", err))
-		}
-
-		_, err = conn.DeleteRegexPatternSet(&wafv2.DeleteRegexPatternSetInput{
+		_, err := conn.DeleteRegexPatternSet(&wafv2.DeleteRegexPatternSetInput{
 			Id:        aws.String(d.Id()),
 			Name:      aws.String(d.Get("name").(string)),
 			Scope:     aws.String(d.Get("scope").(string)),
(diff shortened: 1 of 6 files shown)


website/docs/r/wafv2_regex_pattern_set.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) A friendly name of the regular expression pattern set.
+* `description` - (Optional) A friendly description of the regular expression pattern set.
+* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the Region US East (N. Virginia).
+* `regular_expression_list` - (Required) One or more blocks of regular expression patterns that you want AWS WAF to search for, such as `B[a@]dB[o0]t`.
```
anGie44, line 43: *set as `Optional` in the current schema

aws/resource_aws_wafv2_regex_pattern_set.go
```diff
@@ -0,0 +1,268 @@
...
+func resourceAwsWafv2RegexPatternSetUpdate(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafv2conn
+
+	log.Printf("[INFO] Updating WAFv2 RegexPatternSet %s", d.Id())
+
+	params := &wafv2.UpdateRegexPatternSetInput{
+		Id:                    aws.String(d.Id()),
+		Name:                  aws.String(d.Get("name").(string)),
+		Scope:                 aws.String(d.Get("scope").(string)),
+		Description:           aws.String(d.Get("description").(string)),
```
anGie44, line 168: since L177 takes care of conditionally setting the the `Description` key below, I think it's safe to omit it here (docs also suggest the `Description` attribute isn't required)

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,389 @@
...
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSWafv2RegexPatternSetDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAwsWafv2RegexPatternSetConfig_Minimal(regexPatternSetName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSWafv2RegexPatternSetExists(resourceName, &v),
+					testAccCheckAWSWafv2RegexPatternSetDisappears(resourceName),
```
anGie44, line 80: similar suggestion made in #12119, to use the generic method now available in the provider https://github.com/terraform-providers/terraform-provider-aws/blob/07f8ab992c780c4911832234afbbf67feea2fe66/aws/provider_test.go#L590 in place of a custom one

anGie44:
a couple minor comments, otherwise looks great @pvanbuijtene 👍 

output of acceptance tests:
```
--- PASS: TestAccAwsWafv2RegexPatternSet_Minimal
--- PASS: TestAccAwsWafv2RegexPatternSet_Disappears
--- PASS: TestAccAwsWafv2RegexPatternSet_Basic
--- PASS: TestAccAwsWafv2RegexPatternSet_ChangeNameForceNew
--- PASS: TestAccAwsWafv2RegexPatternSet_Tags
```

aws/resource_aws_wafv2_regex_pattern_set.go
```diff
@@ -0,0 +1,267 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(1, 128),
+			},
+			"regular_expression_list": {
```
anGie44, line 59: how would we feel about changing this attribute to just a `Set` of `Elem: &schema.Schema{Type: schema.TypeString}` (i'm thinking a la `wafregional_regex_pattern_set` schema)? since the API expects a list of strings in the request, perhaps it could help to simplify things here and not require more complex flattening methods.

anGie44, line 59: how would we feel about updating this attribute to be a `Set ` with `Elem:  &schema.Schema{Type: schema.TypeString}`(I'm thinking a la `wafregional_regex_pattern_set`)? while the API does require a map formatting in a request and responds back with a map, we don't necessarily make use of the keys so I'm wondering if we (and the user) may benefit here by adopting a flatter schema type

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,366 @@
...
+		conn := testAccProvider.Meta().(*AWSClient).wafv2conn
+		resp, err := conn.GetRegexPatternSet(
+			&wafv2.GetRegexPatternSetInput{
+				Id:    aws.String(rs.Primary.ID),
+				Name:  aws.String(rs.Primary.Attributes["name"]),
+				Scope: aws.String(rs.Primary.Attributes["scope"]),
+			})
+
+		if err == nil {
+			if *resp.RegexPatternSet.Id == rs.Primary.ID {
```
anGie44, line 213: we should guard against an unexpected `panic` here with a nil check before retrieving the `RegexPatternSet.Id` just in case :)

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,366 @@
...
+				return fmt.Errorf("WAFv2 RegexPatternSet %s still exists", rs.Primary.ID)
+			}
```
anGie44, line 215: similar to the pattern in the [docs](https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html#checkdestroy), you can also `return nil` after this inner `if` condition

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+
+		// Return nil if the RegexPatternSet is already destroyed
+		if awsErr, ok := err.(awserr.Error); ok {
```
anGie44, line 219: we can use the `isAWSErr()` helper function here instead to help with the error code comparison

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,366 @@
...
+			Id:    aws.String(rs.Primary.ID),
+			Name:  aws.String(rs.Primary.Attributes["name"]),
+			Scope: aws.String(rs.Primary.Attributes["scope"]),
+		})
+
+		if err != nil {
+			return err
+		}
+
+		if *resp.RegexPatternSet.Id == rs.Primary.ID {
```
anGie44, line 253: similar comment as above to guard against an unexpected panic here with a nil check

anGie44:
hi @pvanbuijtene, thank you again for you contributions thus far! I just added some additional comments regarding the schema and some recommendations in the custom check functions. thanks for adding in the ARN and disappears checks already 👍

aws/resource_aws_wafv2_regex_pattern_set.go
```diff
@@ -0,0 +1,267 @@
...
+	d.Set("name", resp.RegexPatternSet.Name)
+	d.Set("description", resp.RegexPatternSet.Description)
+	d.Set("arn", resp.RegexPatternSet.ARN)
+	d.Set("lock_token", resp.LockToken)
+
+	if err := d.Set("regular_expression_list", flattenWafv2RegexPatternSet(resp.RegexPatternSet.RegularExpressionList)); err != nil {
+		return fmt.Errorf("Error setting regular_expression_list: %s", err)
+	}
+
+	tags, err := keyvaluetags.Wafv2ListTags(conn, *resp.RegexPatternSet.ARN)
```
anGie44, line 147: we should use the AWS Go SDK's conversion function here to extract the ARN to prevent possible panics

aws/resource_aws_wafv2_regex_pattern_set.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return fmt.Errorf("Error listing tags for WAFv2 RegexPatternSet (%s): %s", *resp.RegexPatternSet.ARN, err)
```
anGie44, line 149: similar comment as above

aws/resource_aws_wafv2_regex_pattern_set.go
```diff
@@ -0,0 +1,267 @@
...
+		RegexString: aws.String(m["regex_string"].(string)),
+	}
+}
+
+func flattenWafv2RegexPatternSet(r []*wafv2.Regex) interface{} {
+	regexPatterns := make([]interface{}, len(r))
+
+	for i, regexPattern := range r {
+		d := map[string]interface{}{
+			"regex_string": *regexPattern.RegexString,
```
anGie44, line 261: similar comment as above re: using the AWS SDK's conversion function

aws/resource_aws_wafv2_regex_pattern_set.go
pvanbuijtene: Yeah thought about it and wasn't sure what to do and stayed close to the API implementation as is done in general. Will change it.

aws/resource_aws_wafv2_regex_pattern_set_test.go
pvanbuijtene: @anGie44 you mean like this `aws.StringValue(resp.RegexPatternSet.Id) == rs.Primary.ID` or should `resp` and `resp.RegexPatternSet` also be checked?

aws/resource_aws_wafv2_regex_pattern_set.go
anGie44: makes sense @pvanbuijtene.. ah before making any major changes, @bflad, what are your recommendations re: this schema definition?

pvanbuijtene: too late 😬 ... can role it back when needed

aws/resource_aws_wafv2_regex_pattern_set_test.go
anGie44: let's go with the former :) looking at our codebase, seems the pattern is to forgo the initial nil check of an object like `resp.RegexPatternSet`

pvanbuijtene: I don't mind adding the extra checks, but are these values not guaranteed to be non nil by the SDK in case of a successful operation?

bflad: @pvanbuijtene unfortunately, the AWS Go SDK provides no guarantees and its up to the API remaining consistent. I would encourage adding `nil` checks where possible -- see https://github.com/terraform-providers/terraform-provider-aws/issues/3949 (and the plethora of crash reports around that time) for an example where missing `nil` checks caused issues from an API update.

aws/resource_aws_wafv2_regex_pattern_set.go
bflad: We typically recommend following the API schema and naming where possible, even if its looks weird initially with single nested arguments in this case. We generally have to assume the API authors have further intentions of expanding the structures within the API, so then we'd later have to deprecate the simplified configurations to support the expanded structures. Much better to future-proof now where possible then deal with the work later. 👍 

Aside: Certainly curious there was an API choice of including the `List` suffix in the actual field name. Nothing wrong with following that usually, but `_list` could be debatable dropping from the Terraform attribute naming if can be confusing for practitioners writing configurations and expecting it to accept the `list` type (`= []`). This is why we can't have nice things. 😅

aws/resource_aws_wafv2_regex_pattern_set_test.go
pvanbuijtene: @bflad will keep that in mind for the future, thanks for the info 👍

aws/resource_aws_wafv2_regex_pattern_set.go
pvanbuijtene: The schema is back as it was, minus the `_list` post fix.

pvanbuijtene:
@anGie44 the schema and nil checks are updated, hope I didn't miss anything 😅

website/docs/r/wafv2_regex_pattern_set.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+## Attributes Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `id` - A unique identifier for the set.
+* `arn` - The Amazon Resource Name (ARN) that identifies the cluster.
+
+## Import
+
+WAFv2 Regex Pattern Sets can be imported using their ID, Name and Scope e.g.
```
anGie44, line 59: in looking at other documentation docs, I could be helpful to add the format just in case it's not evident by the example 
```suggestion
WAFv2 Regex Pattern Sets can be imported using `ID/name/scope` e.g.
```

website/docs/r/wafv2_regex_pattern_set.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) A friendly name of the regular expression pattern set.
+* `description` - (Optional) A friendly description of the regular expression pattern set.
+* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the Region US East (N. Virginia).
+* `regular_expression` - (Optional) One or more blocks of regular expression patterns that you want AWS WAF to search for, such as `B[a@]dB[o0]t`. See [Visibility Configuration](#regular-expression) below for details.
```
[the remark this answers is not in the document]
pvanbuijtene: 🙈

anGie44:
@pvanbuijtene do you mind resolving conflicts when you get a chance now that #12119 is merged

pvanbuijtene:
@anGie44 Done :)

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,368 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSWafv2RegexPatternSetExists(resourceName, &v),
+					testAccMatchResourceAttrRegionalARN(resourceName, "arn", "wafv2", regexp.MustCompile(`regional/regexpatternset/.+$`)),
+					resource.TestCheckResourceAttr(resourceName, "name", regexPatternSetName),
+					resource.TestCheckResourceAttr(resourceName, "description", regexPatternSetName),
+					resource.TestCheckResourceAttr(resourceName, "scope", wafv2.ScopeRegional),
+					resource.TestCheckResourceAttr(resourceName, "regular_expression.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "regular_expression.1641128102.regex_string", "one"),
+					resource.TestCheckResourceAttr(resourceName, "regular_expression.265355341.regex_string", "two"),
+					resource.TestCheckResourceAttr(resourceName, "tags.%", "2"),
```
DrFaust92, line 36: can you remove tags from basic example and add
`resource.TestCheckResourceAttr(resourceName, "tags.%", "0"),`

aws/resource_aws_wafv2_regex_pattern_set_test.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/wafv2"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func TestAccAwsWafv2RegexPatternSet_Basic(t *testing.T) {
+	var v wafv2.RegexPatternSet
+	regexPatternSetName := fmt.Sprintf("regex-pattern-set-%s", acctest.RandString(5))
```
DrFaust92, line 17: can you please change the pattern names to for all occurances:
`rName := acctest.RandomWithPrefix("tf-acc-test")`

pvanbuijtene: @DrFaust92 I changed it but now I'm getting `Error: Error creating WAFv2 RegexPatternSet: WAFDuplicateItemException: AWS WAF couldn’t perform the operation because some resource in your request is a duplicate of an existing one.`
I see it's based on time and not random enough 🤔

DrFaust92: I get that in windows as well sometimes. I usually lower parallel to 1 to overcome this

aws/resource_aws_wafv2_regex_pattern_set_test.go
pvanbuijtene: Done.

aws/resource_aws_wafv2_regex_pattern_set_test.go
pvanbuijtene: That seems to work indeed, thanks.

website/docs/r/wafv2_regex_pattern_set.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) A friendly name of the regular expression pattern set.
+* `description` - (Optional) A friendly description of the regular expression pattern set.
+* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the Region US East (N. Virginia).
```
anGie44, line 42: could you also please update the wording here like suggested in #12788 ? 
```
To work with CloudFront, you must also specify the region `us-east-1` (N. Virginia) on the AWS provider.
```

pvanbuijtene: @anGie44 done.

anGie44:
hi @pvanbuijtene, do you mind rebasing w/master to correct the `aws.erb` file list items?

PR #12313, opened 2020-03-09
Update index.html.markdown

ASaiun:
“access_key” and "secret_key" must be provided

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -158,11 +158,11 @@ In addition to [generic `provider` arguments](https://www.terraform.io/docs/conf
 (e.g. `alias` and `version`), the following arguments are supported in the AWS
  `provider` block:
 
-* `access_key` - (Optional) This is the AWS access key. It must be provided, but
+* `access_key` - (Required) This is the AWS access key. It must be provided, but
   it can also be sourced from the `AWS_ACCESS_KEY_ID` environment variable, or via
   a shared credentials file if `profile` is specified.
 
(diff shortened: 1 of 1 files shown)


bflad:
Hi @ASaiun 👋 Thank you for submitting this.

Terraform Provider documentation uses the `(Required)` and `(Optional)` denotations on attributes in terms of the Terraform Plugin SDK schema properties `Required: true` and `Optional: true` respectively, which maps to whether or not values must be provided in the Terraform configuration itself.

In this case, both of these provider attributes are optional:

https://github.com/terraform-providers/terraform-provider-aws/blob/2cf06d85441a6d32e708015e1f9e623fed930ea7/aws/provider.go#L20-L32

We allow sourcing this information also via other means such as environment variables and the AWS shared credentials files (for profiles), which is why the configuration is optional. Admittedly this is a little confusing in this case since the value must be given in some fashion, however in preference of being consistent with all other Terraform Provider documentation we would prefer to continue to match the schema definitions of Required/Optional here.

If you have suggestions for how to improve the understanding of the Terraform Providers documentation overall, please file a new issue and we can discuss a broader implementation. 👍 Thanks again for contributing.

ASaiun:
hi @bflad , thank you for feedback for this request.
but, I am confusing with the three parmeters(access_key - (Optional) ,secret_key - (Optional), region - (Required) ).
I think them should stay the same
图片

bflad:
@ASaiun Thank you for clarifying. The `region` argument should be `Optional` like the others, so they are the same. 👍

Issue #7004, opened 2018-12-30
API Gateway- Websocket


bbernays:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

API Gateway now supports WebSocket connections. It would be great to be able to define this resource 

### New or Affected Resource(s)

* aws_api_gateway_v2
* aws_api_gateway_v2_route
* aws_api_gateway_v2_integration
* aws_api_gateway_v2_integration_response
* aws_api_gateway_v2_route_response
* aws_api_gateway_v2_deployment
* aws_api_gateway_v2_stage

### Potential Terraform Configuration

```hcl
#Incomplete!!
resource "aws_api_gateway_v2" "websocket" {
  name = "myWebSocketApi3"
  description    = "${format("CMS V2 %s", var.environment)}"
  protocol_type = "WEBSOCKET"
  route_selection_expression = "$request.body.action'"
}
resource "aws_api_gateway_v2_route" "non_default_route" {
  api_id = "${aws_api_gateway_v2.websocket.id}"
  routeKey = "testing"
}


```

### References

* [Documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api.html)

* [Announcement](https://aws.amazon.com/blogs/compute/announcing-websocket-apis-in-amazon-api-gateway/)

mbonig:
I'd like to contribute to this, but have never done any development on this project before. Can we explain a little further the changes that are needed?

Right now I read this to mean there are new _v2 resources created, and that they will largely be similar to the existing resources but with additional logic. But, I think just adding support to the existing resources, with intelligent defaults, feels like a better fit. 

Either way, I'll try submitting a PR...

bbernays:
I suggested that it be _v2 to stay inline with AWS naming of the new resource. [Link here](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/api-reference.html)

mbonig:
I've created a fork and made some changes to support the first new 'v2' resource. Anyone care to review and let me know things I got wrong before implementing the rest?

[...]

PR #12339, opened 2020-03-10
Updates hashibot config to identify `apigatewayv2` service name

gdavison:
Update hashibot config to include `apigatewayv2`

Relates #7004

```release-note
NONE
```

N/A

### DIFF
--- a/.hashibot.hcl
+++ b/.hashibot.hcl
@@ -99,6 +99,7 @@ behavior "regexp_issue_labeler_v2" "service_labels" {
     ],
     "service/apigatewayv2" = [
       "aws_api_gateway_v2_",
+      "aws_apigatewayv2_",
     ],
     "service/applicationautoscaling" = [
       "aws_appautoscaling_",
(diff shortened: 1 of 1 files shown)


bflad:
Since we're not using the long form `aws_api_gateway_v2_`, should we just remove it? It might also be good to remove the (now extraneous) `([^v]|v[^2]|v2[^_])` handling from `service/apigateway` as well. 👍

gdavison:
There are a number of existing PRs which have/had the old naming. Once they're either updated or merged, I figured we could remove that old naming

.hashibot.hcl
```diff
@@ -584,7 +585,9 @@ behavior "pull_request_path_labeler" "service_labels" {
     ]
     "service/apigatewayv2" = [
       "**/*_api_gateway_v2_*",
+      "**/*_apigatewayv2_*",
       "**/api_gateway_v2_*"
```
bflad: Not sure how sensitive hashibot's HCL is for the missing mid-list comma here:

```suggestion
      "**/api_gateway_v2_*",
```

PR #12342, opened 2020-03-11
DynamoDB V2019.11.21

### LINKED ISSUE #11096: Support Version 2 of DynamoDB Global Tables
### Community Note

### Description

AWS recently released a new version of Global Tables for DynamoDB which changes how the the global table is defined. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.V2.html

There is an updated synchronization protocol that decreases the needed write capacity and a simplified definition of how to create a cross region replication (i.e. a global table).

Where as before to set-up cross region replication the individual tables needed to be created and a seperate globabl table to keep those tables in sync (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.tutorial.html#creategt_console) the new version API (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/V2globaltables.tutorial.html#V2creategt_cli) requires an update to the original table to tell it to replicate to separate regions rather than having separately managed tables connected through a global table. This means the current method of defining a global table (aws_dynamodb_global_table) is outdated in modeling how global tables are defined with this new API.

* aws_dynamodb_global_table
* aws_dynamodb_table

Unsure

### References

* https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.V2.html
* https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.V1.html

cmonty:
Attempts to add support for DynamoDB Global Table V2019.11.21. ~~The PR introduces a new resource `aws_dynamodb_table_2019` as initially discussed in #11096. After implementing though, it may be better to fold `replica` into `aws_dynamodb_table` as you can add `replica` to existing tables (just not existing tables already replicated with V2017).~~

I've decided to change the initial PR to add `replica` to `aws_dynamodb_table`. I think there are benefits if you have an existing DynamoDB table without any GlobalTables, being able to add `replica` to the existing resource and having it add the new version of GlobalTables. Otherwise, you'd need to recreate the resource which would require a new table name or deleting an existing one.

### Community Note

Closes #11096 

```release-note
* Adds support for DynamoDB Global Table V2019.11.21
```

```
$ make testacc TESTARGS='-run=TestAccAWSDynamoDbTable2019_basic'
--- PASS: TestAccAWSDynamoDbTable2019_basic (127.79s)                                                     
PASS                                                                                              
ok      github.com/terraform-providers/terraform-provider-aws/aws       127.814s                                                                                                                         
```


stolfi:
Have you tried importing an existing 2019 replicated table?

Can you drop a sample bit of TF code as an example of the configuration?

cmonty:
> Have you tried importing an existing 2019 replicated table?

If the table is replicated using version 2017, I'm not sure you can use the API to convert the table to a 2019 version. Let me check more on this but I think you may have to do it in the AWS console. Is this what you mean?

> Can you drop a sample bit of TF code as an example of the configuration?

Configuration for what I've implemented so far (`replica` has more attributes that can be configured)

```terraform
resource "aws_dynamodb_table" "test" {
  name         = "MyTestTable"
  hash_key     = "TestTableHashKey"
  billing_mode = "PAY_PER_REQUEST"
  stream_enabled = true
  stream_view_type = "NEW_AND_OLD_IMAGES"
  attribute {
    name = "TestTableHashKey"
    type = "S"
  }
  replica {
    region_name = "us-west-1"
  }
  replica {
    region_name = "au-southeast-2"
  }
}
```

stolfi:
> > Have you tried importing an existing 2019 replicated table?
> 
> If the table is replicated using version 2017, I'm not sure you can use the API to convert the table to a 2019 version. Let me check more on this but I think you may have to do it in the AWS console. Is this what you mean?
>

No, I was more thinking about a table I had in 2017, that I moved to 2019 via the console, and now want to import that change in to my existing statefiles.

Currently I have:
- Table definition in region 1
- Table definition in region 2
- Global table definition

I'll need to remove the table definition in region 2 and global table definition, in favor of an update table definition in region 1

Now that I'm thinking about it, I may not have to import it at all....since the end state will already match the described state.

I'll give it a test and see what happens.

cmonty:
I'm running into issues being able to test additional attributes for `ProvisionedThroughputOverride` on both the table _and_ global secondary indexes. The problem stems from the fact that AWS wants the table to either be "Pay-Per-Request or AutoScaled".

Pay-Per-Request is "easy" in I can set the `billing_mode` to `PAY_PER_REQUEST` _however_ when applying the `UpdateTable` AWS doesn't apply the override until the next day (the AWS Console states "Next available change to on-demand mode" is the next day and keeps the table in "Updating" status.

When adding the app auto-scaling targets and policies I still get a validation error from AWS, `Table write capacity should either be Pay-Per-Request or AutoScaled.` So, I'm not sure this would be a way out for the purposes of testing, but also what are the typical strategies for applying changes that take 24 hours to apply?

Would it be reasonable to accept the PR as-is with only adding / deleting regions and not configuring read capacity overrides?

stolfi:
Typically things that I expect to take a long time to run, like a table convert to PPR, I will do in the console, or via the CLI, so that TF doesn't have to sit and wait for it.

stolfi:
After converting my tables to PROVISIONED, and back to PPR, my plan's are clean....not sure I can explain that.

On one, I did have to do an attribute reorder on a GSI, but that was it.

I've only been operating on my existing tables, I can do a couple table create/update/deletes to ensure those run smoothly.

cmonty:
> Typically things that I expect to take a long time to run, like a table convert to PPR, I will do in the console, or via the CLI, so that TF doesn't have to sit and wait for it.

This is what I expected and makes me think updating `read_capacity` is not needed in the PR in favor of using the console or CLI.

cmonty:
Any update on what's needed for merge?

aws/resource_aws_dynamodb_table.go
```diff
@@ -418,6 +430,12 @@ func resourceAwsDynamoDbTableCreate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	if _, ok := d.GetOk("replica"); ok {
+		if err := createDynamoDbReplicas(d.Id(), d.Get("replica").([]interface{}), conn); err != nil {
```
bflad, line 434: To go with our attribute change to `TypeSet`

```suggestion
		if err := createDynamoDbReplicas(d.Id(), d.Get("replica").(*schema.Set).List(), conn); err != nil {
```

aws/resource_aws_dynamodb_table.go
```diff
@@ -652,7 +715,23 @@ func resourceAwsDynamoDbTableDelete(d *schema.ResourceData, meta interface{}) er
...
+	}
+
+	output, err := conn.DescribeTable(input)
+	log.Printf("[DEBUG] DynamoDB delete describe: %s", output)
+	if err != nil {
+		return fmt.Errorf("error describing DynamoDB Table (%s):  %s", d.Id(), err)
+	}
+
+	if len(output.Table.Replicas) > 0 {
+		if err := deleteDynamoDbReplicas(d.Id(), d.Get("replica").([]interface{}), conn); err != nil {
```
bflad, line 729: We typically avoid additional re-reads in `Delete` functions and prefer to just skip any errors that indicate that any deletion already happened. If there are additional replicas added outside what was previously refreshed into the Terraform state, we _should_ allow the API to return an error so the operator can determine if the deletion is still safe.

```suggestion
	if replicas := d.Get("replica").(*schema.Set).List(); len(replicas) > 0 {
			if err := deleteDynamoDbReplicas(d.Id(), replicas, conn); err != nil {
```

aws/resource_aws_dynamodb_table.go
```diff
@@ -738,6 +863,140 @@ func waitForDynamodbTableDeletion(conn *dynamodb.DynamoDB, tableName string, tim
...
+			var targetReplica *dynamodb.ReplicaDescription
+			for _, replica := range result.Table.Replicas {
+				if *replica.RegionName == region {
+					targetReplica = replica
+				}
+			}
+
+			if targetReplica == nil {
+				return nil, "", nil
+			}
```
bflad, line 899: This logic can be consolidated and also exit early out of the loop of replicas when it matches.

```suggestion
			var targetReplica *dynamodb.ReplicaDescription

			for _, replica := range result.Table.Replicas {
				if aws.StringValue(replica.RegionName) == region {
					targetReplica = replica
					break
				}
			}

			if targetReplica == nil {
				return nil, dynamodb.ReplicaStatusCreating, nil
			}
```

aws/resource_aws_dynamodb_table.go
```diff
@@ -738,6 +863,140 @@ func waitForDynamodbTableDeletion(conn *dynamodb.DynamoDB, tableName string, tim
...
+			for _, replica := range result.Table.Replicas {
+				if *replica.RegionName == region {
+					targetReplica = replica
+				}
+			}
+			log.Printf("[DEBUG] targetReplica: %s", targetReplica)
+
+			if targetReplica == nil {
+				return result, "", nil
+			}
```
bflad, line 944: Same here:

```suggestion
			var targetReplica *dynamodb.ReplicaDescription

			for _, replica := range result.Table.Replicas {
				if aws.StringValue(replica.RegionName) == region {
					targetReplica = replica
					break
				}
			}

			if targetReplica == nil {
				return nil, "", nil
			}
```

aws/resource_aws_dynamodb_table_test.go
```diff
@@ -1382,6 +1382,87 @@ func testAccCheckAWSDynamoDbTableDisappears(table *dynamodb.DescribeTableOutput)
...
+    name = "TestTableHashKey"
+    type = "S"
+  }
+
+	replica {
+	  region_name = "us-west-1"
+	}
+}
+`, rName)
+}
```
bflad, line 1447: A few notes about this testing function:

- There should be an additional `TestStep` like the first that re-adds the replica to the table (this will also ensure the deletion function can delete replicas when deleting the table)
- Rather than hardcoding `us-west-1`, which is only in AWS Commercial and may not be usable for everyone running the testing, this should follow the instructions for [Writing and running Cross-Region Acceptance Tests in the CONTRIBUTING Guide](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#writing-and-running-cross-region-acceptance-tests)
- Nit: The test function prefix for this file uses `TestAccAWSDynamoDbTable_`, so this test should also use that. 👍 

```suggestion
func TestAccAWSDynamoDbTable_Replica(t *testing.T) {
	var conf dynamodb.DescribeTableOutput
	var providers []*schema.Provider
	alternateRegionDataSourceName := "data.aws_region.alternate"
	resourceName := "aws_dynamodb_table.test"
	tableName := acctest.RandomWithPrefix("TerraformTestTable-")

	resource.ParallelTest(t, resource.TestCase{
		PreCheck: func() {
			testAccPreCheck(t)
			testAccMultipleRegionsPreCheck(t)
			testAccAlternateRegionPreCheck(t)
		},
		ProviderFactories: testAccProviderFactories(&providers),
		CheckDestroy:      testAccCheckAWSDynamoDbTableDestroy,
		Steps: []resource.TestStep{
			{
				Config: testAccAWSDynamoDbReplicaUpdates(tableName),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckInitialAWSDynamoDbTableExists(resourceName, &conf),
					resource.TestCheckResourceAttr(resourceName, "name", tableName),
					resource.TestCheckResourceAttr(resourceName, "hash_key", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "attribute.2990477658.name", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "attribute.2990477658.type", "S"),
					resource.TestCheckResourceAttr(resourceName, "replica.#", "1"),
					resource.TestCheckResourceAttrPair(resourceName, "replica.0.region_name", alternateRegionDataSourceName, "name"),
				),
			},
			{
				Config:         testAccAWSDynamoDbReplicaUpdates(tableName),
				ResourceName:      resourceName,
				ImportState:       true,
				ImportStateVerify: true,
			},
			{
				Config: testAccAWSDynamoDbReplicaDeletes(tableName),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckInitialAWSDynamoDbTableExists(resourceName, &conf),
					resource.TestCheckResourceAttr(resourceName, "name", tableName),
					resource.TestCheckResourceAttr(resourceName, "hash_key", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "attribute.2990477658.name", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "attribute.2990477658.type", "S"),
					resource.TestCheckResourceAttr(resourceName, "hash_key", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "replica.#", "0"),
				),
			},
			{
				Config: testAccAWSDynamoDbReplicaUpdates(tableName),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckInitialAWSDynamoDbTableExists(resourceName, &conf),
					resource.TestCheckResourceAttr(resourceName, "name", tableName),
					resource.TestCheckResourceAttr(resourceName, "hash_key", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "attribute.2990477658.name", "TestTableHashKey"),
					resource.TestCheckResourceAttr(resourceName, "attribute.2990477658.type", "S"),
					resource.TestCheckResourceAttr(resourceName, "replica.#", "1"),
					resource.TestCheckResourceAttrPair(resourceName, "replica.0.region_name", alternateRegionDataSourceName, "name"),
				),
			},
		},
	})
}

func testAccAWSDynamoDbReplicaUpdates(rName string) string {
	return composeConfig(
		testAccAlternateRegionProviderConfig(),
		fmt.Sprintf(`
data "aws_region" "alternate" {
  provider = "aws.alternate"
}

resource "aws_dynamodb_table" "test" {
  name             = "%s"
  hash_key         = "TestTableHashKey"
  billing_mode     = "PAY_PER_REQUEST"
  stream_enabled   = true
  stream_view_type = "NEW_AND_OLD_IMAGES"

  attribute {
    name = "TestTableHashKey"
    type = "S"
  }

  replica {
    region_name = data.aws_region.alternate.name
  }
}
`, rName))
}
```

stolfi:
> Any update on what's needed for merge?

Not yet, but I should be doing a bit more testing today. I don't think we can not apply the change assuming that people will do via CLI or console....the state should match the code, unless there is a specific `ignore_changes ` for it.

stolfi:
Any chance we can get the suggested changes merged in so we can do some additional testing?

aws/resource_aws_dynamodb_table.go
```diff
@@ -279,6 +279,18 @@ func resourceAwsDynamoDbTable() *schema.Resource {
 					},
 				},
 			},
+			"replica": {
+				Type:     schema.TypeList,
```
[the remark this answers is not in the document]
cmonty: Thanks!

cmonty:
> Hi @cmonty 👋 Thank you for submitting this -- overall its off to a good start code-wise. Please see the below for some initial feedback and let us know if you have any questions or do not have time to implement the items.
> 
> Some other things to consider:
> 
> * Documentation updates will need to be made within `website/docs/r/dynamodb_table.html.markdown` to add the new `replica` configuration block and its nested argument. The documentation for the configuration block should link to the [AWS documentation for V2 global tables](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.V2.html) and explicitly note that it is for version 2019.11.21 while version 2017.11.29 must be managed via the [`aws_dynamodb_global_table` resource](https://www.terraform.io/docs/providers/aws/r/dynamodb_global_table.html). We should also consider adding something like the replica test configuration as an example to the documentation page.
> * Documentation updates to `website/docs/r/dynamodb_global_table.html.markdown` to denote that it is only for version 2017.11.29 global tables, linking to their [AWS documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/globaltables.V1.html)
> * Potentially refactoring the suggested logic within the loops of `updateDynamoDbReplica` into the `deleteDynamoDbReplicas`/`createDynamoDbReplicas`, then just calling those two functions in the `updateDynamoDbReplica` instead of duplicating the logic both places.
> 
> Thanks again.

Thanks for you feedback! I've incorporated the suggestions and will work on the remaining updates to documentation and refactoring. Sorry for the delay!

cmonty:
> Any chance we can get the suggested changes merged in so we can do some additional testing?

I've just accepted the suggestions. If travis fails I can fix it up tomorrow.

cmonty:
@bflad thanks for your help and merging!

gyoga99:
Trying to create a global dynamodb table, with GSI, provisioned billing mode and using autoscaling policy
required_providers {
    aws = "~> 2.58"
  }
I am seeing issue with creating a table with replica field set. 
`error updating DynamoDB Table (sample_table) replica: error creating DynamoDB Table (sample_table) replicas: ValidationException: Table write capacity should either be Pay-Per-Request or AutoScaled.`

giacomomagini:
I'm having some troubles with Global tables V2 and multiple replicas.
`Error: error updating DynamoDB Table (my_table) replica: error creating DynamoDB Table (my_table) replicas: ValidationException: Update table operation with more than one create or delete replica actions not allowed`

PR #11211, opened 2019-12-09
resource/aws_lambda_function: Publish new version on config-only function updates

nemreid:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

### Changes

Closes #601, Closes #1692, Closes #4088, Closes #9707, Closes #11592

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note
BUG FIXES:
* resource/aws_lambda_function: Publish new version on config-only function updates
```

Updates `resourceAwsLambdaFunctionUpdate` to track changes to config OR code and publish a new version using `PublishVersion`, instead of passing `Publish` into the  `UpdateFunctionCodeInput`.

This has the same end result according to the [Go SDK docs for UpdateFunctionCodeInput](https://docs.aws.amazon.com/sdk-for-go/api/service/lambda/#UpdateFunctionCodeInput):
```go
    // Set to true to publish a new version of the function after updating the code.
    // This has the same effect as calling PublishVersion separately.
    Publish *bool `type:"boolean"`
```

Updates `TestAccAWSLambdaFunction_versionedUpdate` to test for publishing new version when runtime is updated but code is not (one example of config-only update).

This recreates functionality from the AWS console, which allows you to publish a new version if the runtime or any other configuration variable (e.g. envs, layer ARN, etc.) has changed.

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccAWSLambdaFunction_versionedUpdate'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./... -v -count 1 -parallel 20 -run=TestAccAWSLambdaFunction_versionedUpdate -timeout 120m
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
=== RUN   TestAccAWSLambdaFunction_versionedUpdate
=== PAUSE TestAccAWSLambdaFunction_versionedUpdate
=== CONT  TestAccAWSLambdaFunction_versionedUpdate
--- PASS: TestAccAWSLambdaFunction_versionedUpdate (73.00s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	73.065s
testing: warning: no tests to run
PASS
[...]

Issue #12266, opened 2020-03-05
aws_lambda_function publish=true doesn't work as expected (convergence issue)


andresvia:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
terraform version
Terraform v0.12.20
+ provider.aws v2.51.0

Your version of Terraform is out of date! The latest version
is 0.12.21. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Resource(s)

* aws_lambda_* (and others that may use the version spewed out by provider)

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
variable timeout {}

provider "aws" {
  region = "us-east-1"
}

resource "aws_lambda_function" "example" {
  layers = [local.my_layer]
  handler = "whatever"
  role = local.my_role
  function_name = "tf-issue-11211"
  runtime = "java8"
  publish = true
  filename = "something.zip"
  timeout = var.timeout
}

resource "aws_lambda_alias" "rolling" {
  function_version = aws_lambda_function.example.version
  name = "rolling"
  function_name = aws_lambda_function.example.arn
}

resource "aws_lambda_alias" "stable" {
  function_version = aws_lambda_function.example.version
  name = "stable"
  function_name = aws_lambda_function.example.arn
  lifecycle {
    ignore_changes = [function_version]
  }
}
```

### Debug Output

Skip.

### Panic Output

Skip.

### Expected Behavior

`Plan: 0 to add, 0 to change, 0 to destroy.`

### Actual Behavior

`Plan: 0 to add, 1 to change, 0 to destroy.`

### Steps to Reproduce

```
# cleanup
terraform destroy -var timeout=10 -auto-approve

# reproduction
# 1
terraform apply -var timeout=10 -auto-approve
# out: Apply complete! Resources: 3 added, 0 changed, 0 destroyed.

# 2
terraform apply -var timeout=20 -auto-approve
[...]

PR #12417, opened 2020-03-16
resource/aws_lambda_function: recompute versioned attributes on config update

nemreid:
Add conditional to the `updateComputedAttributesOnPublish` function to recompute attributes on config-only updates. This section of the code was missed in #11211, which enabled config-only versioning. It means that the provider does not recompute the `qualified_arn` and `version` attributes after publishing config-only version updates, which breaks the graph for any downstream resources dependent on those attributes, requiring a refresh and reapply.

Confirmed to recompute outputs correctly in ad-hoc testing:

```
~ qualified_arn                  = "arn:aws:lambda:(region):(account-id):function:MyLambdaFunction:4" -> (known after apply)
```

Apologies for missing this in the original PR to @bflad and the community.

### Community Note

```release-note
resource/aws_lambda_function: fix bug where qualified_arn/version attributes were not updated on config-only updates
```

```
N/A (did not run)
```

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -267,7 +267,7 @@ func resourceAwsLambdaFunction() *schema.Resource {
 }
 
 func updateComputedAttributesOnPublish(d *schema.ResourceDiff, meta interface{}) error {
-	if needsFunctionCodeUpdate(d) {
+	if needsFunctionCodeUpdate(d) || needsFunctionConfigUpdate(d) {
 		d.SetNewComputed("last_modified")
 		publish := d.Get("publish").(bool)
 		if publish {
@@ -665,6 +665,10 @@ func needsFunctionCodeUpdate(d resourceDiffer) bool {
 	return d.HasChange("filename") || d.HasChange("source_code_hash") || d.HasChange("s3_bucket") || d.HasChange("s3_key") || d.HasChange("s3_object_version")
 }
 
+func needsFunctionConfigUpdate(d resourceDiffer) bool {
+	return d.HasChange("description") || d.HasChange("handler") || d.HasChange("memory_size") || d.HasChange("role") || d.HasChange("timeout") || d.HasChange("kms_key_arn") || d.HasChange("layers") || d.HasChange("dead_letter_config") || d.HasChange("tracing_config") || d.HasChange("vpc_config") || d.HasChange("runtime") || d.HasChange("environment")
+}
+
 // resourceAwsLambdaFunctionUpdate maps to:
 // UpdateFunctionCode in the API / SDK
 func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) error {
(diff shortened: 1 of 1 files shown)


aws/resource_aws_lambda_function.go
```diff
@@ -685,101 +689,87 @@ func resourceAwsLambdaFunctionUpdate(d *schema.ResourceData, meta interface{}) e
...
+					variables := readEnvironmentVariables(environmentVariables.(map[string]interface{}))
 
+					configReq.Environment = &lambda.Environment{
+						Variables: aws.StringMap(variables),
+					}
+				}
+			} else {
 				configReq.Environment = &lambda.Environment{
-					Variables: aws.StringMap(variables),
+					Variables: aws.StringMap(map[string]string{}),
```
nemreid: This interesting diff is actually just me indenting the shown lines to nest within the conditional block.

aws/resource_aws_lambda_function.go
```diff
@@ -665,6 +665,10 @@ func needsFunctionCodeUpdate(d resourceDiffer) bool {
 	return d.HasChange("filename") || d.HasChange("source_code_hash") || d.HasChange("s3_bucket") || d.HasChange("s3_key") || d.HasChange("s3_object_version")
 }
 
+func needsFunctionConfigUpdate(d resourceDiffer) bool {
+	return d.HasChange("description") || d.HasChange("handler") || d.HasChange("memory_size") || d.HasChange("role") || d.HasChange("timeout") || d.HasChange("kms_key_arn") || d.HasChange("layers") || d.HasChange("dead_letter_config") || d.HasChange("tracing_config") || d.HasChange("vpc_config") || d.HasChange("runtime") || d.HasChange("environment")
```
nemreid: Downside to this PR is checking each of these fields twice worst case, as the existing logic still checks each individual field, although this part short-circuits in most cases.

aws/resource_aws_lambda_function.go
```diff
@@ -267,7 +267,7 @@ func resourceAwsLambdaFunction() *schema.Resource {
 }
 
 func updateComputedAttributesOnPublish(d *schema.ResourceDiff, meta interface{}) error {
-	if needsFunctionCodeUpdate(d) {
+	if needsFunctionCodeUpdate(d) || needsFunctionConfigUpdate(d) {
```
nemreid: Crucially, this was the function where an update was missed in #11211.

nemreid:
I've narrowed it down to `d.HasChange("vpc_config")` which appears to **always** return true, regardless of whether the values are provided through variables or hardcoded into my `main.tf`.

Testing confirms VPC config **is** version-specific, so recalculating the version on a VPC is necessary. 

I'm wondering if there is something funky with the way `d.HasChange("vpc_config")` confirms or whether I should update the `DiffSuppressFunc` section of `resourceAwsLambdaFunction` that deals with VPCs in order to alleviate this.

Any advice would be appreciated.

andresvia:
@nemreid , thanks for this PR, I'm new to this and I was wondering if this PR will also help with  #12266

As for what `d.HasChange("vpc_config")` is actually doing, would need to check, I thought changes could/should be calculated based on the config+statefile it seems like I'm wrong, here the change (`HasChange`) is calculated based on config+statefile+responses from the service?

Thanks.

nemreid:
This issue may have been resolved by https://github.com/hashicorp/terraform-provider-aws/pull/15121. I will do some testing when possible, and if resolved I will close this PR.

nemreid:
After performing testing, I can confirm that I am seeing the desired behavior on config-only updates such as only changing the function `timeout` or `environment` on versioned lambdas, which does result in a new version number. This PR is no longer necessary so I am closing it.

PR #12432, opened 2020-03-17
[WIP] fix bug with PutRestApi removing policy

jamesrwarren:
Relates #5364

[WIP] The issue here is not that the policy is left blank on the first run. What seems to happen is that the policy is created by the CreateRestApi function and then the body is subsequently added with PutRestApi which requires an existing rest api to exist to apply the body against.

As the issue doesn't lie with terraform the only workaround I can see is to reapply the policy after the create as part of the PutRestApi step. PR tested against our infra and creates with the policy.

```release-note
NONE
```

To be updated if there's interest in this fix.


bflad:
Hi @jamesrwarren 👋  Thank you for submitting this. The commit in this pull request will become part of the broader fixes for the `aws_api_gateway_rest_api` resource in #17099, which handles some of the special case logic during creation and updates. 👍  Please follow there for further updates.

PR #12563, opened 2020-03-27
Fix multiline pattern in an example.

dmitriy-drenkalyuk:
### Community Note

```release-note
Fixed a mistake in documentation.
```


dmitriy-drenkalyuk:
@bflad oh, I didn't actually know it would work with any identifier, I don't think this is mentioned in the docs for version 0.11 which I was reading. But yeah, I guess it's still better to stick to the convention proposed by Terraform.

PR #12671, opened 2020-04-03
r/aws_spot_instance_request: Remove metadata_options

stefansundin:
### Community Note

---

Today I was surprised to see this in a plan for an `aws_spot_instance_request`:
```
      + metadata_options {
          + http_endpoint               = (known after apply)
          + http_put_response_hop_limit = (known after apply)
          + http_tokens                 = (known after apply)
        }
```

I was a bit surprised because we only added this to `aws_instance` and `aws_launch_template`, and it's very unlikely that AWS will add support to good old ancient spot instance requests. Opening the code reveals, as I suspected, that it inherits the schema from `aws_instance`.

I couldn't spot any other things that should be removed, but let me know if you know of any.

Changelog may not be required for a tiny change like this?

```release-note
resource/aws_spot_instance_request: Resource should not accept `metadata_options`
```

### DIFF
--- a/aws/resource_aws_spot_instance_request.go
+++ b/aws/resource_aws_spot_instance_request.go
@@ -30,6 +30,9 @@ func resourceAwsSpotInstanceRequest() *schema.Resource {
 			// The Spot Instance Request Schema is based on the AWS Instance schema.
 			s := resourceAwsInstance().Schema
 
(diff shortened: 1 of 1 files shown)


stefansundin:
I just confirmed that this is still an issue. Rebased the fix and confirmed that it still works.

PR #12673, opened 2020-04-04
Docs: Update examples for the post-0.12 world

nfagerlund:
This PR includes two commits. 

- The first one updates all of the example code in the AWS provider to remove "interpolation-only expressions". In Terraform 0.12, interpolating a lone value results in a deprecation warning; this update makes our examples 0.12 compliant (and nicer-looking), at the expense of breaking 0.11 compatibility. Is it time to do that yet? Maybe! It's probably worth taking a temperature check of product and support folks before merging this, but it seems like we might finally be at a point where it's worth making 0.11 usage less convenient. 
- The second commit refactors all the examples I could find that were using string interpolation to build complex IAM JSON documents. Those examples now construct native HCL object and tuple literals and pass them to the `jsonencode()` function. This was suggested by a community member over at https://github.com/hashicorp/terraform-website/issues/1047; the reason I did it now was because these examples were causing a bunch of false positives as I was searching for interpolation-only expressions, so when I went back through to un-break them, it seemed worthwhile to make them a little more robust. 

### Community Note

```release-note
NONE
```

### DIFF
--- a/website/docs/d/api_gateway_resource.html.markdown
+++ b/website/docs/d/api_gateway_resource.html.markdown
@@ -19,7 +19,7 @@ data "aws_api_gateway_rest_api" "my_rest_api" {
 }
 
 data "aws_api_gateway_resource" "my_resource" {
-  rest_api_id = "${data.aws_api_gateway_rest_api.my_rest_api.id}"
+  rest_api_id = data.aws_api_gateway_rest_api.my_rest_api.id
   path        = "/endpoint/path"
 }
 ```
--- a/website/docs/d/autoscaling_groups.html.markdown
+++ b/website/docs/d/autoscaling_groups.html.markdown
@@ -27,7 +27,7 @@ data "aws_autoscaling_groups" "groups" {
 }
 
 resource "aws_autoscaling_notification" "slack_notifications" {
-  group_names = ["${data.aws_autoscaling_groups.groups.names}"]
+  group_names = [data.aws_autoscaling_groups.groups.names]
 
   notifications = [
     "autoscaling:EC2_INSTANCE_LAUNCH",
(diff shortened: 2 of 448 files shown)


nfagerlund:
This was sort of a proof of concept, to see how long it would take to update all that code in one of our biggest providers. Turns out it takes most of a single afternoon.

bflad:
@nfagerlund FYI, we have tooling that can automatically do the equivalent of `terraform 0.12upgrade`: [`terrafmt 012upgrade`](https://github.com/katbyte/terrafmt#upgrade-terraform-in-a-file-to-012)

Related: #8950

maryelizbeth:
@nfagerlund, thanks for submitting this PR!

Terraform AWS provider will deprecate support for TF 0.11 in 3.0.0. The deprecation work can be tracked here: [#8950](https://github.com/terraform-providers/terraform-provider-aws/issues/8950). 

In order to optimize for a friendlier user experience, we will hold off on merging PRs that convert existing examples or documentation to Terraform 0.12 syntax while the provider supports the use of TF 0.11. Terraform 0.12 will give a warning if it detects Terraform 0.11 syntax, whereas TF 0.11 will error if TF 0.12 syntax is used within the configuration. We believe receiving a warning is a better user experience for new users, and therefore will continue using 0.11 syntax styles in the documentation, examples, and test configurations.

We’ll mark this as part of the [3.0.0 milestone](https://github.com/terraform-providers/terraform-provider-aws/milestone/70) and evaluate it during our work on that release.  We expect to release 3.0.0 within the next quarter.

nfagerlund:
@maryelizbeth Grand, thanks! 

@bflad OMG, that's so much nicer than the more primitive semi-automation I was doing. Thank you for the heads-up.

PR #12697, opened 2020-04-06
Update iam_role.html.markdown

brain-is-afk:
Correct the documentation to identify that it is required to remove all policies before destroying a role and not just the ones that Terraform attached.  This should not even be an option and if the intention is to destroy a role, performing a destruction should just simply remove all policies then destroy the role.  I am truly confused as to why this option even exists.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/r/iam_role.html.markdown
+++ b/website/docs/r/iam_role.html.markdown
@@ -50,7 +50,7 @@ The following arguments are supported:
 
 ~> **NOTE:** This `assume_role_policy` is very similar but slightly different than just a standard IAM policy and cannot use an `aws_iam_policy` resource.  It _can_ however, use an `aws_iam_policy_document` [data source](https://www.terraform.io/docs/providers/aws/d/iam_policy_document.html), see example below for how this could work.
 
-* `force_detach_policies` - (Optional) Specifies to force detaching any policies the role has before destroying it. Defaults to `false`.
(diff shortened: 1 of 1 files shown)


bflad:
Hi @brain-is-afk 👋  Thank you for submitting this. In this case, we would prefer to keep the existing documentation since it more accurately reflects the current behavior of the resource and is consistent with our other resource documentation practices. The argument's schema uses `Optional: true` and many operators use this type of API verification on destroy to catch cases where unexpected IAM dependencies require attention. Please feel free to reach out in a feature request issue if you believe there are improvements that can be made with this resource.

brain-is-afk:
To me this logic is a failure because any role created along side of existing infrastructure that requires the removal of only the role must have two executions run in order to clean up the role.  First update must be to toggle the boolean and the second requires the removal of the resource and a second apply.  This is less a safety net than it is unexpected or undesirable behavior when two passes of infrastructure change are required where typically one is expected.  

Documentation may be accurate but the behavior is undesirable due to multiple executions being needed.  This is not CI/CD friendly since multiple code changes are required in a sequence and duplicate executions when the boolean isn't set.  Any large code change would be frustrating and the answer would be to always set that boolean, hence the reason that it should be defaulted to true since it is almost a certain guarantee that there will be individuals who neglect to set the boolean due to lack of knowledge.

PR #12788, opened 2020-04-12
New data source: aws_wafv2_ip_set

### LINKED ISSUE #11178: Feature Request: WAFv2 IP Set Data Source
### Community Note

### Description

### New Data Source

* aws_wafv2_ip_set

```hcl
data "aws_wafv2_ip_set" "example" {
  name = ""
  scope = ""
}
```

### References

pvanbuijtene:
### Community Note

Closes #11178
Relates #11046
Depends on #12119 because of duplicated code for tests.

```release-note
**New Data Source:** `aws_wafv2_ip_set` (#12788)
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccDataSourceAwsWafv2IP'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccDataSourceAwsWafv2IP -timeout 120m
=== RUN   TestAccDataSourceAwsWafv2IPSet_Basic
=== PAUSE TestAccDataSourceAwsWafv2IPSet_Basic
=== CONT  TestAccDataSourceAwsWafv2IPSet_Basic
--- PASS: TestAccDataSourceAwsWafv2IPSet_Basic (22.06s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       22.417s
```

### DIFF
--- a/aws/data_source_aws_wafv2_ip_set.go
+++ b/aws/data_source_aws_wafv2_ip_set.go
@@ -14,11 +14,20 @@ func dataSourceAwsWafv2IPSet() *schema.Resource {
 		Read: dataSourceAwsWafv2IPSetRead,
 
 		Schema: map[string]*schema.Schema{
+			"addresses": {
+				Type:     schema.TypeSet,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
 			"arn": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"ip_address_version": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
@@ -49,7 +58,7 @@ func dataSourceAwsWafv2IPSetRead(d *schema.ResourceData, meta interface{}) error
 	for {
 		output, err := conn.ListIPSets(input)
 		if err != nil {
-			return fmt.Errorf("Error reading WAFV2 IPSets: %s", err)
+			return fmt.Errorf("Error reading WAFv2 IPSets: %s", err)
 		}
 
 		for _, ipSet := range output.IPSets {
@@ -66,12 +75,24 @@ func dataSourceAwsWafv2IPSetRead(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if foundIpSet == nil {
-		return fmt.Errorf("WAFV2 IPSet not found for name: %s", name)
+		return fmt.Errorf("WAFv2 IPSet not found for name: %s", name)
 	}
 
-	d.SetId(aws.StringValue(foundIpSet.Id))
-	d.Set("arn", aws.StringValue(foundIpSet.ARN))
-	d.Set("description", aws.StringValue(foundIpSet.Description))
+	resp, err := conn.GetIPSet(&wafv2.GetIPSetInput{
+		Id:    aws.String(*foundIpSet.Id),
+		Name:  aws.String(*foundIpSet.Name),
+		Scope: aws.String(d.Get("scope").(string)),
+	})
+
+	if err != nil {
+		return fmt.Errorf("WAFv2 IPSet not found for name: %s", name)
+	}
+
+	d.SetId(aws.StringValue(resp.IPSet.Id))
+	d.Set("addresses", flattenStringList(resp.IPSet.Addresses))
+	d.Set("arn", aws.StringValue(resp.IPSet.ARN))
+	d.Set("description", aws.StringValue(resp.IPSet.Description))
+	d.Set("ip_address_version", aws.StringValue(resp.IPSet.IPAddressVersion))
 
 	return nil
 }
--- a/aws/data_source_aws_wafv2_ip_set.go
+++ b/aws/data_source_aws_wafv2_ip_set.go
@@ -54,24 +54,30 @@ func dataSourceAwsWafv2IPSetRead(d *schema.ResourceData, meta interface{}) error
 	var foundIpSet *wafv2.IPSetSummary
 	input := &wafv2.ListIPSetsInput{
 		Scope: aws.String(d.Get("scope").(string)),
+		Limit: aws.Int64(100),
 	}
+
 	for {
-		output, err := conn.ListIPSets(input)
+		resp, err := conn.ListIPSets(input)
 		if err != nil {
 			return fmt.Errorf("Error reading WAFv2 IPSets: %s", err)
 		}
 
-		for _, ipSet := range output.IPSets {
-			if aws.StringValue(ipSet.Name) == name {
+		if resp == nil || resp.IPSets == nil {
+			return fmt.Errorf("Error reading WAFv2 IPSets")
+		}
+
+		for _, ipSet := range resp.IPSets {
+			if ipSet != nil && aws.StringValue(ipSet.Name) == name {
 				foundIpSet = ipSet
 				break
 			}
 		}
 
-		if output.NextMarker == nil || foundIpSet != nil {
+		if resp.NextMarker == nil || foundIpSet != nil {
 			break
 		}
-		input.NextMarker = output.NextMarker
+		input.NextMarker = resp.NextMarker
 	}
 
 	if foundIpSet == nil {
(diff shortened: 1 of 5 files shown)


pvanbuijtene:
@anGie44 sure no problem, will rebase them tomorrow.

aws/data_source_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+	}
+}
+
+func dataSourceAwsWafv2IPSetRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).wafv2conn
+	name := d.Get("name").(string)
+
+	var foundIpSet *wafv2.IPSetSummary
+	input := &wafv2.ListIPSetsInput{
```
anGie44, line 55: we may want to also set the `Limit` attribute in here to the allowed maximum (100) to prevent API throttling

aws/data_source_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		input.NextMarker = output.NextMarker
+	}
+
+	if foundIpSet == nil {
+		return fmt.Errorf("WAFv2 IPSet not found for name: %s", name)
+	}
+
+	resp, err := conn.GetIPSet(&wafv2.GetIPSetInput{
+		Id:    aws.String(*foundIpSet.Id),
```
anGie44, line 82: hmm maybe this is one instance where we can forgo the AWS Go SDK's conversion function as the `Id` is already a string pointer?..else we could get quite nested and do `aws.String(aws.StringValue(foundIpSet.Id))` 😅

aws/data_source_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,98 @@
...
+		Name:  aws.String(*foundIpSet.Name),
+		Scope: aws.String(d.Get("scope").(string)),
+	})
+
+	if err != nil {
+		return fmt.Errorf("WAFv2 IPSet not found for name: %s", name)
+	}
+
+	d.SetId(aws.StringValue(resp.IPSet.Id))
+	d.Set("addresses", flattenStringList(resp.IPSet.Addresses))
```
anGie44, line 92: we should error check this aggregate type with the `if err := d.Set()` pattern

aws/data_source_aws_wafv2_ip_set_test.go
```diff
@@ -0,0 +1,69 @@
...
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccDataSourceAwsWafv2IPSet_NonExistent(name),
+				ExpectError: regexp.MustCompile(`WAFv2 IPSet not found`),
+			},
+			{
+				Config: testAccDataSourceAwsWafv2IPSet_Name(name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair(datasourceName, "addresses", resourceName, "addresses"),
```
anGie44, line 28: we should also use the ARN testing check function in this step as well

anGie44:
hi @pvanbuijtene, thanks again for pushing up the rebase and additions! I've left some follow-up suggestions above for your consideration. also if you could please omit the resource file and the reference to it in `provider.go`, it would be greatly appreciated :)

aws/data_source_aws_wafv2_ip_set.go
pvanbuijtene: Good one 👍

aws/data_source_aws_wafv2_ip_set_test.go
pvanbuijtene: The ARN is tested against the ARN of the resource, does this not guarantee it's correct? 🤔

aws/data_source_aws_wafv2_ip_set.go
```diff
@@ -79,20 +81,23 @@ func dataSourceAwsWafv2IPSetRead(d *schema.ResourceData, meta interface{}) error
...
 	resp, err := conn.GetIPSet(&wafv2.GetIPSetInput{
-		Id:    aws.String(*foundIpSet.Id),
-		Name:  aws.String(*foundIpSet.Name),
+		Id:    foundIpSet.Id,
+		Name:  foundIpSet.Name,
 		Scope: aws.String(d.Get("scope").(string)),
 	})
 
-	if err != nil {
+	if err != nil || resp == nil || resp.IPSet == nil {
```
anGie44, line 89: i think we shouldn't lose the error here in case, so can we keep it as `if err != nil { return fmt.Errorf("message: %s", err) }`  and then the following lines can do the checks similar to `if resp == nil || resp.IPSet == nil { return fmt.Errorf("") }?

aws/data_source_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	for {
+		output, err := conn.ListIPSets(input)
+		if err != nil || output == nil {
+			return fmt.Errorf("Error reading WAFv2 IPSets: %s", err)
+		}
+
+		for _, ipSet := range output.IPSets {
+			if aws.StringValue(ipSet.Name) == name && ipSet != nil {
```
anGie44, line 67: quick adjustment here: we'll want to re-order this `if` condition such that `ipSet` first gets `nil` checked before attempting to access its fields

aws/data_source_aws_wafv2_ip_set.go
pvanbuijtene: I'm not sure how much extra value these specific nil checks add, that's why I grouped them.
Will change it tomorrow in all PRs.

aws/data_source_aws_wafv2_ip_set_test.go
anGie44: ah yes good catch! for ARN checking `resource.TestCheckResourceAttrPair` or `testAccMatchResourceAttrRegionalARN` are valid so either one will suffice

aws/data_source_aws_wafv2_ip_set.go
```diff
@@ -0,0 +1,103 @@
...
+	var foundIpSet *wafv2.IPSetSummary
+	input := &wafv2.ListIPSetsInput{
+		Scope: aws.String(d.Get("scope").(string)),
+		Limit: aws.Int64(100),
+	}
+
+	for {
+		output, err := conn.ListIPSets(input)
+		if err != nil || output == nil {
+			return fmt.Errorf("Error reading WAFv2 IPSets: %s", err)
```
anGie44, line 63: only concern here is that if we satisfy the `if` condition with `output == nil` , the err here will be `nil` ; i'd recommend splitting this OR into 2 separate `if` blocks to help make the return statements relate to the condition in question

aws/data_source_aws_wafv2_ip_set.go
anGie44: grouping the later 2 nil checks make sense 👍 i think the majority of the value comes from the initial `err != nil` and the error we'd get back to give the user context.

pvanbuijtene:
@anGie44 the resource code is removed.

anGie44:
thanks again @pvanbuijtene! LGTM 👍 

Note: holding off on acceptance testing as depends on corresponding resource implementation in https://github.com/terraform-providers/terraform-provider-aws/pull/12119

website/aws.erb
```diff
@@ -3553,6 +3553,19 @@
                         </li>
                     </ul>
                 </li>
+                <li>
+                    <a href="#">WAFv2</a>
```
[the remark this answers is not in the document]
pvanbuijtene: I updated it.

website/docs/d/wafv2_ip_set.html.markdown
```diff
@@ -0,0 +1,37 @@
...
+  scope = "REGIONAL"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the WAFv2 IP Set.
+* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. To work with CloudFront, you must also specify the Region US East (N. Virginia).
```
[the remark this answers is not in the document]
pvanbuijtene: @bflad  It's updated, hope it's more clear now.

PR #12921, opened 2020-04-21
add support for importing aws_network_acl_rule resources

### LINKED ISSUE #10983: Importing aws_network_acl_rule
### Community Note

### Description

I'm trying to import existing network acl rules. This was originally reported in #703 along with importing `aws_route`. import support for `aws_route` was complete but not `aws_network_acl_rule`. I think we should to track each request individually, hence filing this issue.

* aws_network_acl_rule

N/A

### References

* #704

anGie44:
### Community Note

Closes #10983

```
support for network_acl_rule resource importing
```

```
--- PASS: TestAccAWSNetworkAclRule_tcpProtocol (40.08s)
--- PASS: TestAccAWSNetworkAclRule_disappears_NetworkAcl (171.46s)
--- PASS: TestAccAWSNetworkAclRule_missingParam (172.85s)
--- PASS: TestAccAWSNetworkAclRule_basic (186.02s)
--- PASS: TestAccAWSNetworkAclRule_ingressEgressSameNumberDisappears (301.98s)
--- PASS: TestAccAWSNetworkAclRule_ipv6ICMP (303.63s)
--- PASS: TestAccAWSNetworkAclRule_ipv6 (304.05s)
--- PASS: TestAccAWSNetworkAclRule_disappears (304.36s)
--- PASS: TestAccAWSNetworkAclRule_allProtocol (340.56s)
--- PASS: TestAccAWSNetworkAclRule_ipv6VpcAssignGeneratedIpv6CidrBlockUpdate (341.29s)
```

### DIFF
--- a/aws/resource_aws_network_acl_rule.go
+++ b/aws/resource_aws_network_acl_rule.go
@@ -235,8 +235,8 @@ func resourceAwsNetworkAclRuleRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("ipv6_cidr_block", resp.Ipv6CidrBlock)
 	d.Set("egress", resp.Egress)
 	if resp.IcmpTypeCode != nil {
-		d.Set("icmp_code", strconv.FormatInt(*resp.IcmpTypeCode.Code, 10))
-		d.Set("icmp_type", strconv.FormatInt(*resp.IcmpTypeCode.Type, 10))
+		d.Set("icmp_code", strconv.FormatInt(aws.Int64Value(resp.IcmpTypeCode.Code), 10))
+		d.Set("icmp_type", strconv.FormatInt(aws.Int64Value(resp.IcmpTypeCode.Type), 10))
 	}
 	if resp.PortRange != nil {
 		d.Set("from_port", resp.PortRange.From)
--- a/aws/resource_aws_network_acl_rule.go
+++ b/aws/resource_aws_network_acl_rule.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -19,6 +20,38 @@ func resourceAwsNetworkAclRule() *schema.Resource {
 		Create: resourceAwsNetworkAclRuleCreate,
 		Read:   resourceAwsNetworkAclRuleRead,
 		Delete: resourceAwsNetworkAclRuleDelete,
+		Importer: &schema.ResourceImporter{
+			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+				idParts := strings.Split(d.Id(), ":")
+				if len(idParts) != 4 || idParts[0] == "" || idParts[1] == "" || idParts[2] == "" || idParts[3] == "" {
+					return nil, fmt.Errorf("unexpected format of ID (%q), expected NETWORK_ACL_ID:RULE_NUMBER:PROTOCOL:EGRESS", d.Id())
+				}
+				networkAclID := idParts[0]
+				ruleNumber, err := strconv.Atoi(idParts[1])
+				if err != nil {
+					return nil, err
+				}
+				protocol, err := validateProtocolArgumentValue(idParts[2])
+				if err != nil {
+					return nil, err
+				}
+				if protocol == nil {
+					return nil, fmt.Errorf("unknown protocol %s, expected one defined at %s",
+						idParts[2], "https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml)")
+				}
+				egress, err := strconv.ParseBool(idParts[3])
+				if err != nil {
+					return nil, err
+				}
+
+				d.Set("network_acl_id", networkAclID)
+				d.Set("rule_number", ruleNumber)
+				d.Set("egress", egress)
+				d.Set("protocol", protocol)
+				d.SetId(networkAclIdRuleNumberEgressHash(networkAclID, ruleNumber, egress, *protocol))
+				return []*schema.ResourceData{d}, nil
+			},
+		},
 
 		Schema: map[string]*schema.Schema{
 			"network_acl_id": {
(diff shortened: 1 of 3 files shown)


aws/resource_aws_network_acl_rule_test.go
```diff
@@ -632,15 +643,13 @@ resource "aws_network_acl" "test" {
 }
 
 resource "aws_network_acl_rule" "test" {
-  from_port       = -1
```
anGie44, line 635: removed these port ranges as the protocol here is `ipv6-icmp` and port range is supported only for ucp/tcp per [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-network-acl-entry.html). if port ranges are passed to the test, the api returns nil for `resp.PortRange`; response output from `findNetworkAclRule`: 
```
  Egress: false,
  IcmpTypeCode: {
    Code: -1,
    Type: -1
  },
  Ipv6CidrBlock: "::/0",
  Protocol: "58",
  RuleAction: "allow",
  RuleNumber: 150
}
```

aws/resource_aws_network_acl_rule_test.go
```diff
@@ -632,15 +643,13 @@ resource "aws_network_acl" "test" {
...
   icmp_code       = -1
   icmp_type       = -1
   ipv6_cidr_block = "::/0"
   network_acl_id  = "${aws_network_acl.test.id}"
-  protocol        = 58
+  protocol        = "ipv6-icmp"
```
anGie44, line 650: updated here to the string value b/c if the protocol is sent as an int, while the resource ID will be generated with the value provided as input https://github.com/terraform-providers/terraform-provider-aws/blob/c8027b65401471582d1d699b8638efc164840b76/aws/resource_aws_network_acl_rule.go#L167 , the protocol value that gets saved to the resource's state is the string identifier https://github.com/terraform-providers/terraform-provider-aws/blob/c8027b65401471582d1d699b8638efc164840b76/aws/resource_aws_network_acl_rule.go#L227 which then causes difficulty at `ImportState` time b/c `testAccAWSNetworkAclRuleImportStateIdFunc` will return the string value of the protocol and this will inevitably generate the wrong ID

anGie44: since we depend on `testAccAWSNetworkAclRuleImportStateIdFunc` to generate the ID for importing, is a better alternative to fix the output there instead of adjusting the resource params?

aws/resource_aws_network_acl_rule_test.go
```diff
@@ -15,20 +15,34 @@ import (
...
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckAWSNetworkAclRuleDestroy,
 		Steps: []resource.TestStep{
 			{
 				Config: testAccAWSNetworkAclRuleBasicConfig,
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckAWSNetworkAclRuleExists("aws_network_acl_rule.baz", &networkAcl),
-					testAccCheckAWSNetworkAclRuleExists("aws_network_acl_rule.qux", &networkAcl),
-					testAccCheckAWSNetworkAclRuleExists("aws_network_acl_rule.wibble", &networkAcl),
-				),
```
bflad, line 31: For consistency with the rest of the provider codebase, let's keep these `Check`s (here and in the below tests) in the testing, even with import testing. We can discuss their potential removal separately. 👍

aws/resource_aws_network_acl_rule_test.go
```diff
@@ -162,20 +178,15 @@ func TestAccAWSNetworkAclRule_ipv6VpcAssignGeneratedIpv6CidrBlockUpdate(t *testi
...
-					resource.TestCheckResourceAttr(vpcResourceName, "assign_generated_ipv6_cidr_block", "false"),
-					resource.TestCheckResourceAttr(vpcResourceName, "ipv6_cidr_block", ""),
-				),
 			},
 			{
 				Config: testAccAWSNetworkAclRuleConfigIpv6VpcAssignGeneratedIpv6CidrBlockUpdate(true),
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckVpcExists(vpcResourceName, &vpc),
-					resource.TestCheckResourceAttr(vpcResourceName, "assign_generated_ipv6_cidr_block", "true"),
-					resource.TestMatchResourceAttr(vpcResourceName, "ipv6_cidr_block", regexp.MustCompile(`/56$`)),
```
bflad, line 176: This testing might seem a little odd, but it is written this way to ensure assumptions about changes to an upstream resource have the desired outcome in the tested resource and the `Check`s should stay. Please see the reference issue: https://github.com/terraform-providers/terraform-provider-aws/issues/6710

aws/resource_aws_network_acl_rule_test.go
bflad: Ideally the `ImportStateIdFunc` should handle any quirky behavior caused by multiple potential values for an argument, rather than forcing the test configurations to be updated. 👍

aws/resource_aws_network_acl_rule.go
```diff
@@ -210,8 +235,8 @@ func resourceAwsNetworkAclRuleRead(d *schema.ResourceData, meta interface{}) err
 	d.Set("ipv6_cidr_block", resp.Ipv6CidrBlock)
 	d.Set("egress", resp.Egress)
 	if resp.IcmpTypeCode != nil {
-		d.Set("icmp_code", resp.IcmpTypeCode.Code)
-		d.Set("icmp_type", resp.IcmpTypeCode.Type)
+		d.Set("icmp_code", strconv.FormatInt(*resp.IcmpTypeCode.Code, 10))
+		d.Set("icmp_type", strconv.FormatInt(*resp.IcmpTypeCode.Type, 10))
```
bflad, line 239: To prevent potential panics, we should use the [AWS Go SDK provided conversion functions](https://docs.aws.amazon.com/sdk-for-go/api/aws/), e.g. `aws.StringValue()`, here:

```suggestion
		d.Set("icmp_code", strconv.FormatInt(aws.StringValue(resp.IcmpTypeCode.Code), 10))
		d.Set("icmp_type", strconv.FormatInt(aws.StringValue(resp.IcmpTypeCode.Type), 10))
```

aws/resource_aws_network_acl_rule_test.go
anGie44: ah will do!

PR #12927, opened 2020-04-21
autoscaling: set conflict between availability_zones and vpc_zone_identifier

bendrucker:
### Community Note

This PR adds `ConflictsWith` to `vpc_zone_identifier`, ensuring that a user that has passed VPC subnet IDs does not also pass a set of AZ names. This configuration would have previously resulted in an error on apply and now will trigger and error during validation.

Relates to https://github.com/terraform-providers/terraform-provider-aws/issues/9622

```release-note
Prevent `availability_zones` and `vpc_zone_identifier` from being specified together
```

### DIFF
--- a/aws/diff_suppress_funcs.go
+++ b/aws/diff_suppress_funcs.go
@@ -97,6 +97,16 @@ func suppressOpenIdURL(k, old, new string, d *schema.ResourceData) bool {
 	return oldUrl.String() == newUrl.String()
 }
 
+func suppressAutoscalingGroupAvailabilityZoneDiffs(k, old, new string, d *schema.ResourceData) bool {
+	// If VPC zone identifiers are provided then there is no need to explicitly
+	// specify availability zones.
+	if _, ok := d.GetOk("vpc_zone_identifier"); ok {
+		return true
+	}
+
+	return false
+}
+
 func suppressCloudFormationTemplateBodyDiffs(k, old, new string, d *schema.ResourceData) bool {
 	normalizedOld, err := normalizeCloudFormationTemplate(old)
 
--- a/aws/diff_suppress_funcs.go
+++ b/aws/diff_suppress_funcs.go
@@ -97,16 +97,6 @@ func suppressOpenIdURL(k, old, new string, d *schema.ResourceData) bool {
 	return oldUrl.String() == newUrl.String()
 }
 
-func suppressAutoscalingGroupAvailabilityZoneDiffs(k, old, new string, d *schema.ResourceData) bool {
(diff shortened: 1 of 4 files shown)


bflad:
Hi @bendrucker 👋 Thank you for submitting this change. Please note I've marked this as a breaking change, since it currently causes some existing acceptance testing to fail that previously succeeded:

```
--- FAIL: TestAccAWSAutoScalingGroup_classicVpcZoneIdentifier (4.08s)
    testing.go:669: Step 0 error: config is invalid: "vpc_zone_identifier": conflicts with availability_zones

--- FAIL: TestAccAWSAutoScalingGroup_emptyAvailabilityZones (23.59s)
    testing.go:669: Step 0 error: errors during apply:
        
        Error: "vpc_zone_identifier": conflicts with availability_zones
```

The test configurations with these are using empty lists with the arguments, e.g.

```hcl
resource "aws_autoscaling_group" "test" {
  min_size = 0
  max_size = 0

  availability_zones   = []
  launch_configuration = "${aws_launch_configuration.test.name}"
  vpc_zone_identifier  = ["${aws_subnet.test.id}"]
}
```

Which the resource code is designed to only pass elements when there is more than 0, e.g.

https://github.com/terraform-providers/terraform-provider-aws/blob/8a8845ed183d2f084f356cb8631d09870d05d340/aws/resource_aws_autoscaling_group.go#L533

The Terraform Plugin SDK `ConflictsWith` behavior strictly checks argument existence; not the final semantics of the given configuration values. Given that we probably should implement this at the schema level so it does return a plan-time validation error, at the cost of forcing operators to specify `null` for missing lists instead `[]`, I will put this our next major release milestone for future consideration.

bendrucker:
Thank you! Rushed through this a bit and didn't have a chance to run the acceptance tests. Totally understand that this is breaking, I have a similar change or two (adding `ConflictsWith`) similarly queued up for the Google provider.

* Updated `TestAccAWSAutoScalingGroup_classicVpcZoneIdentifier` to not use an empty list for `vpc_zone_identifiers`
* Removed `TestAccAWSAutoScalingGroup_emptyAvailabilityZones`, which is now appears irrelevant

I can set up and actually run the tests later today.

It also seemed like `DiffSuppressFunc` wasn't necessary anymore but then started to doubt myself.

bendrucker:
Verified that the remaining acceptance test passes:

```sh
make testacc TESTARGS='-run=TestAccAWSAutoScalingGroup_classicVpcZoneIdentifier'
```

bendrucker:
Thank you!

PR #13078, opened 2020-04-29
Cognito User Pool - Email Configuration Doc Upddate

mousedownmike:
### Community Note

```release-note
`cognito_user_pool` Email Configuration documentation update
```


website/docs/r/cognito_user_pool.markdown
```diff
@@ -89,9 +89,9 @@ The following arguments are supported:
 #### Email Configuration
 
   * `reply_to_email_address` (Optional) - The REPLY-TO email address.
-  * `source_arn` (Optional) - The ARN of the email source.
-  * `from_email_address` (Optional) - Sender’s email address or sender’s name with their email address (e.g. "john@smith.com" or "John Smith <john@smith.com>")
-  * `email_sending_account` (Optional) - Instruct Cognito to either use its built-in functional or Amazon SES to send out emails.
+  * `source_arn` (Optional) - The ARN of the SES verified email identity to to use. Required if `email_sending_account` is set to `DEVELOPER`.
+  * `from_email_address` (Optional) - Sender’s email address or sender’s name with their email address (e.g. "john@example.com" or "John Smith &lt;john@example.com&gt;")
```
anGie44, line 93: hi @mousedownmike thank you for this PR! the updates to `source_arn` and `email_sending_account` LGTM 👍 
for `from_email_address`, do you mind adding the format backticks around the email address and name+email address as suggested below? in `master`, we have the backticks to emphasize these [examples](https://www.terraform.io/docs/providers/aws/r/cognito_user_pool.html#email-configuration).  the side-effect of this though is that the entity names will not be acknowledged and thus must be reverted to `<` and `>`, respectively. 
```suggestion
  * `from_email_address` (Optional) - Sender’s email address or sender’s name with their email address (e.g. `john@example.com` or `John Smith <john@example.com>`)
```

mousedownmike: Hi @anGie44 

That change looks reasonable. However, after using this configuration more I've discovered that the display name has to be in quotes or a `terraform apply` will register a change.  I'm about to push a change that incorporates that with yours:

```markdown
  * `from_email_address` (Optional) - Sender’s email address or sender’s name with their email address (e.g. `john@example.com` or `\"John Smith\" <john@example.com>`)
```

anGie44: interesting, when you say TF will register a change, what's the diff show as?

anGie44:
thanks @mousedownmike, do you mind rebasing/solving conflicts for the markdown file as well?

website/docs/r/cognito_user_pool.markdown
mousedownmike: It shows "John Smith <john@example.com" -> "\"John Smith\" <john@example.com>".

anGie44: hmm just checking i'm understanding the behavior here: so as input , you provided  `""John Smith" john@example.com"` instead of `"John Smith <john@example.com>"` ? without the `<` and `>`?

mousedownmike: Sorry... that didn't format correctly.  Here's the exact output I was getting from a plan (with the example name/email swapped in:

```
      ~ email_configuration {
            email_sending_account = "DEVELOPER"
          ~ from_email_address    = "\"John Smith\" <john@example.com>" -> "John Smith <john@example.com>"
            source_arn            = "arn:aws:ses:us-east-1:1234567890123:identity/john@example.com"
        }
```

This went away when I added the escaped quotes around John Smith indicating to me that the AWS API would wrap the John Smith in quotes when it stored that value making subsequent plans show that as a change.  I think what I have put in the most recent commit represents this.

anGie44: thanks @mousedownmike for the config snippet! let me get back to you on this b/c this brings to light some interesting behavior we might not be capturing in our acceptance testing for this resource

anGie44: hi again @mousedownmike! so I confirmed locally the diff you commented on earlier, but in cases where the display name has a `.` such as `John A. Smith` and this RFC [example](https://tools.ietf.org/html/rfc5322#appendix-A.1.2) shows a similar use case, explaining that a string with a period character must be enclosed in quotes as it's normally used in parsing as a delimiter. so going back to your latest commit which looks well formatted, I would just suggest modifying your example to showcases these 2 different distinctions with something like:
```
(e.g. `john@example.com` or `John Smith <john@example.com>`). 
To use a sender's name with a period character such as `John J. Smith`, it must be escaped e.g. `\"John J. Smith\" <john@example.com>`
```

anGie44: yep there are definitely some docs in the code base that link to RFCs so this fits right in (e.g ref to RFC timestamp format in https://www.terraform.io/docs/providers/aws/r/lambda_event_source_mapping.html). thanks @mousedownmike!

PR #13248, opened 2020-05-10
tests/provider: Add Terraform Provider GitHub Actions workflow

bflad:
### Community Note

Reference: https://github.com/actions/cache
Reference: https://github.com/bflad/tfproviderdocs
Reference: https://github.com/hashicorp/setup-terraform
Reference: https://www.terraform.io/docs/commands/providers/schema.html

```release-note
NONE
```

This workflow encompasses a large portion of the TravisCI handling, but leaves that in place until this experimental setup is verified with some real world usage. In particular, this workflow enables the enhanced validations available in `tfproviderdocs check`, which requires the JSON schema of the provider and is not setup in TravisCI due to the relative complexity of manually downloading, verifying, and installing Terraform CLI. These validations are able to detect mismatched and missing data source and resource documentation files, which a few existing issues are ignored for future remediation.

Aside: The schema dump process can be vastly improved in Terraform Plugin SDK v2 (or potentially a future SDK v1 release) by no longer requiring Terraform CLI and instead implementing a wrapper utilizing gRPC Reflection and matching the output of `terraform providers schema -json`. For now though, this uses the new `hashicorp/setup-terraform` GitHub Action and the necessary plugin directory setup to accomplish the schema dump.

This workflow heavily utilizes the `actions/cache` GitHub Action to improve feedback time as well as split the jobs into individual tools for readability by contributors and maintainers.

Output from acceptance testing: N/A (CI testing)

### DIFF
--- a/.github/workflows/terraform_provider.yml
+++ b/.github/workflows/terraform_provider.yml
@@ -1,0 +1,258 @@
+name: Terraform Provider Checks
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - .github/workflows/terraform_provider.yml
+      - .golangci.yml
+      - aws/**
+      - awsproviderlint/**
+      - docs/index.md
+      - docs/data-sources/**
+      - docs/guides/**
+      - docs/resources/**
+      - go.sum
+      - GNUmakefile
+      - main.go
+      - staticcheck.conf
+      - website/**
+
+env:
+  AWS_DEFAULT_REGION: us-west-2
+  GO_VERSION: "1.14"
+  GO111MODULE: on
+  TERRAFORM_VERSION: "0.12.24"
+
+jobs:
+  go_mod_download:
+    name: go mod download
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GO_VERSION }}
+    - uses: actions/cache@v1
+      continue-on-error: true
+      id: cache-go-pkg-mod
+      timeout-minutes: 2
+      with:
+        path: ~/go/pkg/mod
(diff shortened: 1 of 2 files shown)


.github/workflows/terraform_provider.yml
```diff
@@ -0,0 +1,258 @@
...
+      - docs/resources/**
+      - go.sum
+      - GNUmakefile
+      - main.go
+      - staticcheck.conf
+      - website/**
+
+env:
+  AWS_DEFAULT_REGION: us-west-2
+  GO_VERSION: "1.14"
```
gdavison, line 25: Should `.go-version` be updated to a 1.14 as well?

.github/workflows/terraform_provider.yml
```diff
@@ -0,0 +1,258 @@
...
+  GO111MODULE: on
+  TERRAFORM_VERSION: "0.12.24"
```
gdavison, line 27: Should this be 0.12.25?

.github/workflows/terraform_provider.yml
bflad: At this point, probably yes! I've been using 1.14.0 through 1.14.2 just fine locally the last few weeks. There's a few things that need to be updated when we do that so let's create a separate tracking issue and ensure the maintaining guide has all those items listed. 👍

.github/workflows/terraform_provider.yml
bflad: Looks like this came out between submission and review -- I can update it here to ensure the PR passes CI still, but maybe the Terraform CLI version can be either:

- Read from a consistent file (similar to `.go-version`)
- Fetched in `hashicorp/setup-terraform` by it supporting the latest of a minor version (e.g. just passing `"0.12") (I'll submit an issue over there or play with this after merging)

Issue #11888, opened 2020-02-04
Fix and enable awsproviderlint AWSAT001 check: prefer resource.TestCheckResourceAttrPair() or ARN check functions


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When checking ARN values in acceptance testing, it is preferred to use `resource.TestCheckResourceAttrPair()` to compare the value against another state value (typical for data sources) or use one of the other available ARN checking functions provided by the Terraform AWS Provider such as `testAccMatchResourceAttrRegionalARN`. See also: https://github.com/terraform-providers/terraform-provider-aws/blob/master/awsproviderlint/passes/AWSAT001/README.md

Due to the amount of reports and potential review complexity, any fixes should likely be done per-resource.

e.g.

```
/Users/bflad/src/github.com/terraform-providers/terraform-provider-aws/aws/data_source_aws_acm_certificate_test.go:44:6: AWSAT001: prefer resource.TestCheckResourceAttrPair() or ARN check functions (e.g. testAccMatchResourceAttrRegionalARN)
/Users/bflad/src/github.com/terraform-providers/terraform-provider-aws/aws/data_source_aws_acm_certificate_test.go:50:6: AWSAT001: prefer resource.TestCheckResourceAttrPair() or ARN check functions (e.g. testAccMatchResourceAttrRegionalARN)
/Users/bflad/src/github.com/terraform-providers/terraform-provider-aws/aws/data_source_aws_acm_certificate_test.go:56:6: AWSAT001: prefer resource.TestCheckResourceAttrPair() or ARN check functions (e.g. testAccMatchResourceAttrRegionalARN)
/Users/bflad/src/github.com/terraform-providers/terraform-provider-aws/aws/data_source_aws_acm_certificate_test.go:62:6: AWSAT001: prefer resource.TestCheckResourceAttrPair() or ARN check functions (e.g. testAccMatchResourceAttrRegionalARN)
[...]

PR #13263, opened 2020-05-11
service/RAM: Update acceptance tests to use ARN testing check functions

gdavison:
### Community Note

Relates #11888

```release-note
NONE
```

Previously:

```
aws/resource_aws_ram_resource_share_accepter_test.go:34:6: AWSAT001: prefer resource.TestCheckResourceAttrPair() or ARN check functions (e.g. testAccMatchResourceAttrRegionalARN)
aws/resource_aws_ram_resource_share_accepter_test.go:35:6: AWSAT001: prefer resource.TestCheckResourceAttrPair() or ARN check functions (e.g. testAccMatchResourceAttrRegionalARN)
```

```
--- FAIL: TestAccAwsRamResourceShareAccepter_basic (11.40s)
```

The test failure `TestAccAwsRamResourceShareAccepter_basic` is a pre-existing failure.


aws/resource_aws_ram_resource_share_accepter_test.go
```diff
@@ -31,8 +33,8 @@ func TestAccAwsRamResourceShareAccepter_basic(t *testing.T) {
 				Config: testAccAwsRamResourceShareAccepterBasic(shareName),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAwsRamResourceShareAccepterExists(resourceName),
-					resource.TestMatchResourceAttr(resourceName, "share_arn", regexp.MustCompile(`^arn\:aws\:ram\:.*resource-share/.+$`)),
-					resource.TestMatchResourceAttr(resourceName, "invitation_arn", regexp.MustCompile(`^arn\:aws\:ram\:.*resource-share-invitation/.+$`)),
+					resource.TestCheckResourceAttrPair(resourceName, "share_arn", principalAssociationResourceName, "resource_share_arn"),
+					testAccMatchResourceAttrRegionalARN(resourceName, "invitation_arn", "ram", regexp.MustCompile("resource-share-invitation/.+$")),
```
bflad, line 37: Blah, in this case the `invitation_arn` account ID is from the other AWS account:

```
    TestAccAwsRamResourceShareAccepter_basic: testing.go:683: Step 0 error: Check failed: Check 3/9 error: aws_ram_resource_share_accepter.test: Attribute 'invitation_arn' didn't match "arn:aws:ram:us-west-2:111111111111:resource-share-invitation/.+$", got "arn:aws:ram:us-west-2:222222222222:resource-share-invitation/2bce9ffd-c64d-4ec3-9105-92bd450f6545"
```

We do not have a covering helper function for that type of ARN -- maybe we can add one? I don't think the `invitation_arn` is exposed by the other resource.

gdavison: I added `testAccCheckResourceAttrRegionalARNAccountID ` in #13173 that should help with this

aws/provider_test.go
```diff
@@ -73,6 +74,21 @@ func testAccPreCheck(t *testing.T) {
 	}
 }
 
+// testAccAlternateAccountProvider returns the schema.Provider for the alternate account
+// Must be used within a resource.TestCheckFunc
+func testAccAlternateAccountProvider(providers *[]*schema.Provider) (result *schema.Provider) {
```
gdavison, line 79: After thinking about it, this is overly tied to one specific case of using alternate accounts. There’s a better way to handle this.

aws/provider_test.go
```diff
@@ -180,6 +181,27 @@ func testAccMatchResourceAttrRegionalARNNoAccount(resourceName, attributeName, a
 	}
 }
 
+// testAccMatchResourceAttrRegionalARNAccountID ensures the Terraform state regexp matches a formatted ARN with region and specific account ID
+func testAccMatchResourceAttrRegionalARNAccountID(resourceName, attributeName, accountID, arnService string, arnResourceRegexp *regexp.Regexp) resource.TestCheckFunc {
```
bflad, line 185: 😭 Please don't hate me, but I have the smallest nitpicks/questions:

- Since ARNs are in `arn:PARTITION:SERVICE:REGION:ACCOUNT:RESOURCE` format, could we have accountID be the second to last argument?
- Should we enforce a `*regexp.Regexp` type for the accountID argument if the intent is to use strings like `\d{12}`? Otherwise I wonder if we should just default that value in the `arnRegexp`, remove the argument, and name the function something like `testAccMatchResourceAttrRegionalAnyAccountARN()`? I think 12 digit account numbers and the `aws` string are the only expected values, unless there might be a test where we fully know the 12 digit account number, but I cannot think of a case like that anywhere.

gdavison: 1. Yeah, that bugged me a bit when I did it. There was a reason, IIRC, but I'll change it
1. There's also at least `lambda` as an account ID. In this case, I was piggybacking on how `testAccMatch*ARN*()` build the final regex for the match. In the `*_service_account` data sources, we know the expected account ID, but those are `Check` not `Match` tests.

bflad: Cool cool thanks!

PR #13319, opened 2020-05-14
Documentation updates

ksatirli:
Hello maintainers! :wave:

This updates two parts in the documentation.

For the `acm_certficate` resource: the `id` and `arn` attributes are identical, but are exposed as different attributes.

For other [resources](https://www.terraform.io/docs/providers/aws/r/kinesis_video_stream.html), we indicate that they are the same (for now), so I added this identifier here, too.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/website/docs/r/acm_certificate.html.markdown
+++ b/website/docs/r/acm_certificate.html.markdown
@@ -105,8 +105,8 @@ Supported nested arguments for the `options` configuration block:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `id` - The ARN of the certificate
-* `arn` - The ARN of the certificate
+* `id` - The ID of the certificate
+* `arn` - The ARN of the certificate (same as `id`)
 * `domain_name` - The domain name for which the certificate is issued
 * `domain_validation_options` - A list of attributes to feed into other resources to complete certificate validation. Can have more than one element, e.g. if SANs are defined. Only set if `DNS`-validation was used.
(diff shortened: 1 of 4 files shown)


website/docs/r/acm_certificate.html.markdown
```diff
@@ -105,8 +105,8 @@ Supported nested arguments for the `options` configuration block:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `id` - The ARN of the certificate
-* `arn` - The ARN of the certificate
+* `id` - The ID of the certificate
+* `arn` - The ARN of the certificate (same as `id`)
```
bflad, line 109: We typically try to denote that `id` attribute is the ARN when it is an ARN. It can be confusing to say "ID" in this context because that could mean just the part after `certificate/...` in the ARN.

For additional context, we typically add the `arn` attribute by convention, even if the resource ID is the ARN, since folks seem to expect that attribute name.

See also: https://github.com/terraform-providers/terraform-provider-aws/blob/941de448465dd3619fbde6ced64ab4111b1d5fe2/aws/resource_aws_acm_certificate.go#L217

```suggestion
* `id` - The ARN of the certificate
* `arn` - The ARN of the certificate
```

ksatirli:
Hey @bflad thanks for the explanation - that certainly helps!

I've updated the documentation.

PR #13364, opened 2020-05-16
Fix aws_security_group_rule: description overwritten on state refresh

ktham:
### Community Note

Reason for change: Existing state refresh code does not set `description` field when `source_security_group_id` contains account id prefix.

```release-note
* resource/aws_security_group_rule: Fix description field from being overwritten during state refresh when `source_security_group_id` contains account id prefix
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id -timeout 120m
=== RUN   TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id
=== PAUSE TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id
=== CONT  TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id (34.83s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	39.394s
```

### DIFF
--- a/aws/resource_aws_security_group_rule.go
+++ b/aws/resource_aws_security_group_rule.go
@@ -851,15 +851,24 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
 			groupId := components[1]
 
 			for _, gp := range rule.UserIdGroupPairs {
+				if *gp.GroupId != groupId || *gp.UserId != userId {
+					continue
+				}
 				if *gp.GroupId == groupId && *gp.UserId == userId {
-					return aws.StringValue(gp.Description)
+					if desc := aws.StringValue(gp.Description); desc != "" {
+						return desc
+					}
 				}
 			}
 		case 1:
 			groupId := components[0]
 			for _, gp := range rule.UserIdGroupPairs {
-				if *gp.GroupId == groupId {
-					return aws.StringValue(gp.Description)
+				if *gp.GroupId != groupId {
+					continue
+				}
+
+				if desc := aws.StringValue(gp.Description); desc != "" {
+					return desc
 				}
 			}
 		}
--- a/aws/resource_aws_security_group_rule.go
+++ b/aws/resource_aws_security_group_rule.go
@@ -842,20 +842,25 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
 		}
 	}
 
-	// probe UserIdGroupPairs
-	groupIds := make(map[string]bool)
 	if raw, ok := d.GetOk("source_security_group_id"); ok {
-		groupIds[raw.(string)] = true
-	}
+		components := strings.Split(raw.(string), "/")
 
-	if len(groupIds) > 0 {
-		for _, gp := range rule.UserIdGroupPairs {
-			if _, ok := groupIds[*gp.GroupId]; !ok {
-				continue
+		switch len(components) {
+		case 2:
+			userId := components[0]
+			groupId := components[1]
+
+			for _, gp := range rule.UserIdGroupPairs {
+				if *gp.GroupId == groupId && *gp.UserId == userId {
+					return aws.StringValue(gp.Description)
+				}
 			}
-
-			if desc := aws.StringValue(gp.Description); desc != "" {
-				return desc
+		case 1:
+			groupId := components[0]
+			for _, gp := range rule.UserIdGroupPairs {
+				if *gp.GroupId == groupId {
+					return aws.StringValue(gp.Description)
+				}
 			}
 		}
 	}
--- a/aws/resource_aws_security_group_rule.go
+++ b/aws/resource_aws_security_group_rule.go
@@ -842,6 +842,7 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
 		}
 	}
 
+	// probe UserIdGroupPairs
 	if raw, ok := d.GetOk("source_security_group_id"); ok {
 		components := strings.Split(raw.(string), "/")
 
@@ -851,19 +852,18 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
 			groupId := components[1]
 
 			for _, gp := range rule.UserIdGroupPairs {
-				if *gp.GroupId != groupId || *gp.UserId != userId {
+				if aws.StringValue(gp.GroupId) != groupId || aws.StringValue(gp.UserId) != userId {
 					continue
 				}
-				if *gp.GroupId == groupId && *gp.UserId == userId {
-					if desc := aws.StringValue(gp.Description); desc != "" {
-						return desc
-					}
+
+				if desc := aws.StringValue(gp.Description); desc != "" {
+					return desc
 				}
 			}
 		case 1:
 			groupId := components[0]
 			for _, gp := range rule.UserIdGroupPairs {
-				if *gp.GroupId != groupId {
+				if aws.StringValue(gp.GroupId) != groupId {
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


ktham:
Hi @bflad / @anGie44 , I found an issue where the `description` field isn't set properly when `source_security_group_id` contains an account id prefix.  I had discovered this after we merged https://github.com/terraform-providers/terraform-provider-aws/pull/11809.

Would've caught this if the test config had the `description` field set 😢 .

aws/resource_aws_security_group_rule.go
```diff
@@ -842,20 +842,34 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
...
-		for _, gp := range rule.UserIdGroupPairs {
-			if _, ok := groupIds[*gp.GroupId]; !ok {
-				continue
+		switch len(components) {
+		case 2:
+			userId := components[0]
+			groupId := components[1]
+
+			for _, gp := range rule.UserIdGroupPairs {
+				if *gp.GroupId != groupId || *gp.UserId != userId {
```
anGie44, line 854: we should use the `aws.StringValue` for these UserGroupIdPair values

aws/resource_aws_security_group_rule.go
```diff
@@ -842,20 +842,34 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
...
+					continue
+				}
+				if *gp.GroupId == groupId && *gp.UserId == userId {
```
anGie44, line 857: i believe you can omit this logical check as only a `False || False` result in the above check will get us here

aws/resource_aws_security_group_rule.go
```diff
@@ -842,20 +842,34 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
 		}
 	}
 
-	// probe UserIdGroupPairs
```
anGie44, line 845: do you mind adding back the comment here?

aws/resource_aws_security_group_rule.go
```diff
@@ -842,20 +842,34 @@ func descriptionFromIPPerm(d *schema.ResourceData, rule *ec2.IpPermission) strin
...
+				if *gp.GroupId == groupId && *gp.UserId == userId {
+					if desc := aws.StringValue(gp.Description); desc != "" {
+						return desc
+					}
+				}
 			}
+		case 1:
+			groupId := components[0]
+			for _, gp := range rule.UserIdGroupPairs {
+				if *gp.GroupId != groupId {
```
anGie44, line 866: similar comment as above to use the aws SDK's string conversion method

aws/resource_aws_security_group_rule.go
ktham: ah, yes

ktham:
Thanks @anGie44 ! PR updated!

ktham:
> an additional check using testAccCheckAWSSecurityGroupRuleAttributes in the ComposeTestCheckFunc could be a nice addition since the test aims to verify a specific set of attributes for the security_group_rule in question.

Hi @anGie44 , thanks for that suggestion. Unfortunately, I can't use `testAccCheckAWSSecurityGroupRuleAttributes` because that helper function requires me to create an [ec2.IpPermission](https://github.com/terraform-providers/terraform-provider-aws/blob/cd05028c0e8f4066e1a65dc79f3fe77ac43f42db/aws/resource_aws_security_group_rule_test.go#L1219) with the expected values I want to check for. Since the values aren't specific CIDR ranges but are dynamic and dependent on the test environment (e.g. the account ids and resource ids), I cannot easily create a `UserIdGroupPair` with the `source_security_group_id` and `security_group_id` filled with expected values I want to test for.

I do see value in adding additional validation on the resource attributes, so I added some `resource.TestMatchResourceAttr` cases to validate the attribute values. Let me know what you think!

anGie44:
ah true @ktham! I figured there was logic in the helper function where you could pass in a `nil` value of the `ec2.IpPermission` and generate the permission but this defaults to a "tcp"  permission with hardcoded `IpRanges` when in fact we'd need the `UserIdGroupPair` in addition to the other attributes..well in any cases thanks for the response! the test update looks good! 

Output of acceptance tests:
```
--- PASS: TestIpPermissionIDHash (0.00s)
--- PASS: TestAccAWSSecurityGroupRule_ExpectInvalidTypeError (3.99s)
--- PASS: TestAccAWSSecurityGroupRule_ExpectInvalidCIDR (4.82s)
--- PASS: TestAccAWSSecurityGroupRule_Egress (16.68s)
--- PASS: TestAccAWSSecurityGroupRule_EgressDescription (22.59s)
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Classic (24.64s)
--- PASS: TestAccAWSSecurityGroupRule_Issue5310 (25.95s)
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Protocol (26.05s)
--- PASS: TestAccAWSSecurityGroupRule_EgressDescription_updates (26.05s)
--- PASS: TestAccAWSSecurityGroupRule_MultiIngress (26.33s)
--- PASS: TestAccAWSSecurityGroupRule_Ingress_VPC (27.71s)
--- PASS: TestAccAWSSecurityGroupRule_SelfSource (27.97s)
--- PASS: TestAccAWSSecurityGroupRule_IngressDescription_updates (27.94s)
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id (31.44s)
--- PASS: TestAccAWSSecurityGroupRule_Ingress_Ipv6 (34.03s)
--- PASS: TestAccAWSSecurityGroupRule_IngressDescription (34.31s)
--- PASS: TestAccAWSSecurityGroupRule_Description_AllPorts (30.01s)
--- PASS: TestAccAWSSecurityGroupRule_MultipleRuleSearching_AllProtocolCrash (19.02s)
--- PASS: TestAccAWSSecurityGroupRule_PartialMatching_basic (37.28s)
--- PASS: TestAccAWSSecurityGroupRule_PartialMatching_Source (37.99s)
--- PASS: TestAccAWSSecurityGroupRule_PrefixListEgress (39.35s)
--- PASS: TestAccAWSSecurityGroupRule_SelfReference (46.96s)
--- PASS: TestAccAWSSecurityGroupRule_Description_AllPorts_NonZeroPorts (51.85s)
--- PASS: TestAccAWSSecurityGroupRule_MultiDescription (47.15s)
--- PASS: TestAccAWSSecurityGroupRule_Race (100.69s)
```

aws/resource_aws_security_group_rule_test.go
```diff
@@ -168,6 +170,18 @@ func TestAccAWSSecurityGroupRule_Ingress_Source_With_Account_Id(t *testing.T) {
 				Config: testAccAWSSecurityGroupRule_Ingress_Source_with_AccountId(rInt),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckAWSSecurityGroupRuleExists("aws_security_group.web", &group),
+					resource.TestMatchResourceAttr(
```
anGie44, line 173: i don't think a change is needed but as a note, you can also compare to the `security_group` as an alternative to matching the regex by doing something like:
```
resource.TestCheckResourceAttrPair(ruleName, "security_group_id", "aws_security_group.web", "id")
```

ktham: Good point! Updated!

aws/resource_aws_security_group_rule_test.go
```diff
@@ -2089,6 +2103,7 @@ resource "aws_security_group_rule" "allow_self" {
   from_port                = 0
   to_port                  = 0
   protocol                 = "-1"
+  description              = "some description"
```
bflad, line 2106: Nit: It is generally preferred to create a new test and test configuration to ensure we are not introducing a regression. 👍

Issue #13312, opened 2020-05-13
Fix Resources Failing with TF_SCHEMA_PANIC_ON_ERROR


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The acceptance testing framework has supported the `TF_SCHEMA_PANIC_ON_ERROR` environment variable as a methodology for causing a panic instead of (usually unchecked) errors with `d.Set()`. In Terraform Plugin SDK version 2, this behavior will be enabled (in some form, potentially not a panic) by default so provider developers can more easily discover these trickier coding issues. Static analysis and the potential to add typed `d.Get()`/`d.Set()` receiver methods cannot prevent this type of issue everywhere before it reaches testing.

These errors indicate places where Terraform cannot properly set `Computed` attribute values in the state or properly perform drift detection for arguments.

### Affected Resource(s)

Currently failing:

- [x] aws_lb_cookie_stickiness_policy: `cookie_expiration_period: '' expected type 'int', got unconvertible type 'string'`
- [x] aws_lb_ssl_negotiation_policy: `Invalid address to set: []string{"attributes"}`
- [x] aws_lightsail_instance: `ram_size: '' expected type 'int', got unconvertible type 'string'`
- [x] aws_load_balancer_listener_policy: `load_balancer_port: '' expected type 'int', got unconvertible type 'string'`
- [x] aws_load_balancer_backend_server_policy: `instance_port: '' expected type 'int', got unconvertible type 'string'`
- [x] aws_load_balancer_policy: `load_balancer_port: '' expected type 'int', got unconvertible type 'string'`
- [x] aws_opsworks_application: `environment.0.secure: '' expected type 'bool', got unconvertible type 'string'`
- [x] aws_redshift_security_group: `Invalid address to set: []string{"skip_final_snapshot"}` (on import)
- [x] aws_sns_topic: `lambda_success_feedback_sample_rate: '' expected type 'int', got unconvertible type 'string'`
- [x] aws_ssm_activation: `expired: '' expected type 'string', got unconvertible type 'bool'`
[...]

PR #13418, opened 2020-05-20
service/elb: Fix schema errors

appilon:
Fixed string to int type mismatches. Noticed attr name typo `attributes` --> `attribute`.

### Community Note

Relates #13312 

```release-note
* resource/aws_lb_cookie_stickiness_policy: `cookie_expiration_policy` now properly set
* resource/aws_load_balancer_backend_server_policy: `instance_port` now properly set
* resource/aws_load_balancer_listener_policy: `load_balancer_port` now properly set
```

```

...
```

### DIFF
--- a/aws/resource_aws_lb_ssl_negotiation_policy.go
+++ b/aws/resource_aws_lb_ssl_negotiation_policy.go
@@ -146,7 +146,20 @@ func resourceAwsLBSSLNegotiationPolicyRead(d *schema.ResourceData, meta interfac
 	// We can get away with this because there's only one policy returned
 	policyDesc := getResp.PolicyDescriptions[0]
 	attributes := flattenPolicyAttributes(policyDesc.PolicyAttributeDescriptions)
-	return d.Set("attribute", attributes)
+
+	// This was previously erroneously setting "attributes", however this cannot
+	// be changed without introducing problematic side effects. The ELB service
+	// automatically expands the results to include all SSL attributes
+	// (unordered, so we'd need to switch to TypeSet anyways), which we would be
+	// quite impractical to force practitioners to write out and potentially
+	// update each time the API updates since there is nearly 100 attributes.
+
+	// TODO: fix attribute
+	// if err := d.Set("attribute", attributes); err != nil {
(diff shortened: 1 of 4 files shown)


aws/resource_aws_lb_ssl_negotiation_policy.go
```diff
@@ -139,16 +139,14 @@ func resourceAwsLBSSLNegotiationPolicyRead(d *schema.ResourceData, meta interfac
...
-
 	d.Set("name", policyName)
 	d.Set("load_balancer", lbName)
 	d.Set("lb_port", lbPort)
 
-	return nil
+	// We can get away with this because there's only one policy returned
+	policyDesc := getResp.PolicyDescriptions[0]
+	attributes := flattenPolicyAttributes(policyDesc.PolicyAttributeDescriptions)
+	return d.Set("attribute", attributes)
```
bflad, line 149: This attribute is going to be very complicated, if not impractical to fix unfortunately. The ELB service automatically expands the results to include all SSL attributes (unordered, so we'd need to switch to `TypeSet` anyways), which we would be quite impractical to force practitioners to write out and potentially update each time the API updates. 😦 For example from running the acceptance testing:

```
--- FAIL: TestAccAWSLBSSLNegotiationPolicy_basic (95.19s)
testing.go:684: Step 0 error: After applying this step and refreshing, the plan was not empty:
DIFF:
DESTROY/CREATE: aws_lb_ssl_negotiation_policy.test
attribute.#:        "93" => "7" (forces new resource)
attribute.0.name:   "ADH-AES128-GCM-SHA256" => "AES128-GCM-SHA256" (forces new resource)
attribute.0.value:  "false" => "true" (forces new resource)
attribute.1.name:   "ADH-AES128-SHA" => "ECDHE-RSA-AES128-GCM-SHA256" (forces new resource)
attribute.1.value:  "false" => "true" (forces new resource)
attribute.10.name:  "ADH-RC4-MD5" => "" (forces new resource)
attribute.10.value: "false" => "" (forces new resource)
```

My suggestion for now would be just remove or comment out the `d.Set()` with a note mentioning this API expansion issue.

PR #13435, opened 2020-05-20
service/opsworks: Fix schema error

appilon:
### Community Note

Relates #13312

```release-note
* resource/aws_opsworks_application: `environment` `secure` now properly set
```

```

...
```

### DIFF
--- a/aws/resource_aws_opsworks_application.go
+++ b/aws/resource_aws_opsworks_application.go
@@ -387,32 +387,42 @@ func resourceAwsOpsworksApplicationDelete(d *schema.ResourceData, meta interface
 	return err
 }
 
-func resourceAwsOpsworksSetApplicationEnvironmentVariable(d *schema.ResourceData, v []*opsworks.EnvironmentVariable) {
-	log.Printf("[DEBUG] envs: %s %d", v, len(v))
-	if len(v) == 0 {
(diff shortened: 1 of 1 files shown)


bflad:
Ha! I think fixing this one is highlighting another bug (https://github.com/terraform-providers/terraform-provider-aws/issues/5429), as the acceptance testing is failing with:

```
--- FAIL: TestAccAWSOpsworksApplication_basic (39.02s)
testing.go:684: Step 2 error: After applying this step and refreshing, the plan was not empty:
DIFF:
UPDATE: aws_opsworks_application.test
...
environment.#:                   "2" => "2"
environment.0.key:               "key1" => "key1"
environment.0.secure:            "false" => "false"
environment.0.value:             "value1" => "value1"
environment.1.key:               "key2" => "key2"
environment.1.secure:            "true" => "true"
environment.1.value:             "*****FILTERED*****" => "value2"
```

Where I believe the `*****FILTERED*****` value is what the API is returning when it is secure. 🙃 In these cases we typically need to omit setting that special value into state (and hopefully allow the configuration -> state passthrough behavior, but I forget off the top of my head if that works for nested attributes).

appilon:
will do!

appilon:
@bflad I had to significantly change the implementation to allow the state passthrough but the acceptance test passes now.

PR #13439, opened 2020-05-20
Add TODO comment for tricker schema error

appilon:
This is the final schema error in the codebase, it seems there is some history here as there is a warn, currently though this does not work since `result` is a deeply nested map, but the schema is defined as `TypeMap, Elem: TypeString`.

So I'm not sure the best way to fix this, I suspect switching the field to `TypeString` and saving the `json` output directly, with a diffsuppressfunc is likely the best way to save the data? Not sure if that causes a breaking change, maybe that was the purpose of the `result` attribute?

I wanted to add the TODO so we could come back to this searching `TODO: schema error`

### Community Note

Relates #13312

```release-note

```

```

...
```

### DIFF
--- a/aws/data_source_aws_lambda_invocation.go
+++ b/aws/data_source_aws_lambda_invocation.go
@@ -82,6 +82,9 @@ func dataSourceAwsLambdaInvocationRead(d *schema.ResourceData, meta interface{})
 		return err
 	}
 
+	// TODO: schema error: result_map seems to be a deeply nested map which
+	// cannot be set. May need to change type to string, and apply json diff
+	// suppression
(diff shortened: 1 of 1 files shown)


appilon:
Travis is done and passed for this, not sure why still shows pending

bflad:
@appilon some additional context is available here: https://github.com/terraform-providers/terraform-provider-aws/issues/9189

We'd like to remove the `result_map` attribute in the next major version due to the issue you mention here.

bflad:
Since we now have a very solid reason for deprecating and removing the attribute, submitted the deprecation here: #13492

PR #13481, opened 2020-05-25
Increase aws_lambda_function create timeout:

itayadler:
- This is a workaround for a potential issue noticed
in cases where it may take around 2 minute to create
a lambda. It seems a TimeoutError failure which
will occur when a lambda takes 2 minutes to be created
will result in another CreateFunction request, as
the first one isn't canceled with the TimeoutError.
Increase from 1 minute to 5 minutes as a workaround.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```


breathingdust:
Hi @itayadler 👋, thanks for the contribution. It looks like we are using the native waiter now so this is no longer necessary.

PR #13482, opened 2020-05-25
Fix interpolation and update example to latest LTS

mindfulmonk:
Interpolation-only expressions are deprecated.
Switched to Ubuntu 20.04 LTS

### Community Note

### DIFF
--- a/website/docs/r/instance.html.markdown
+++ b/website/docs/r/instance.html.markdown
@@ -25,7 +25,7 @@ data "aws_ami" "ubuntu" {
 
   filter {
     name   = "name"
-    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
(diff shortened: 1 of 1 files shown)


website/docs/r/instance.html.markdown
```diff
@@ -37,7 +37,7 @@ data "aws_ami" "ubuntu" {
 }
 
 resource "aws_instance" "web" {
-  ami           = "${data.aws_ami.ubuntu.id}"
+  ami           = data.aws_ami.ubuntu.id
```
bflad, line 40: Please note that we would like to keep the Terraform 0.11 and 0.12 compatible string interpolation syntax around for a little while longer in the documentation, while the provider natively supports Terraform 0.11 and earlier. We plan on updating all examples with Terraform 0.12 only syntax as part of upcoming major version release work since it will no longer support the older Terraform version, see also: #8950

PR #13556, opened 2020-05-29
[PROPOSAL] Implement TypeSet resource.TestCheckFuncs

appilon:
This proposed PR would add 2 helper `resource.TestCheckFunc` that can substitute current checks that use a pre-calculated hash in the `key`. These functions essentially search all of the instance state for a `schema.TypeSet` element that matches either the simple value or a nested group of attr/value pairs.

In V2 of the SDK, the underlying test driver uses [this library](https://github.com/hashicorp/terraform-json) which naively exports TypeSets as a list so this implementation supports both "element ids" of either number or hash.

TypeSet indexes should be substituted with the sentinel value `*`, this works across deep nesting. `TestCheckTypeSetElemNestedAttrs` should be used with a granular value map to ensure an element is a full match.

#### Examples
Deep TypeSet simple element
```
test.TestCheckTypeSetElemAttr("foo.test", "cors_configuration.0.allow_headers.*", "Authorization"),
```
TypeSet nested object
```
test.TestCheckTypeSetElemNestedAttrs("foo.test", "parameter.*", map[string]string{
    "name":  "character_set_results",
    "value": "utf8",
})
```
TypeSet with downstream TypeList
```
test.TestCheckTypeSetElemNestedAttrs("foo.test", "origin.*", map[string]string{
    "custom_origin_config.0.http_port":  "80",
})
```
TypeSet with downstream TypeList with downstream TypeSet
```
test.TestCheckTypeSetElemNestedAttrs("foo.test", "parameter.*.lifecycle_rule.2.transition.*", map[string]string{
    "storage_class":  "INTELLIGENT_TIERING",
})
```

### DIFF
--- a/aws/internal/test/testing.go
+++ b/aws/internal/test/testing.go
@@ -1,0 +1,114 @@
+package test
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+const (
+	sentinelIndex = "*"
+)
+
+// TestCheckTypeSetElemNestedAttrs is a resource.TestCheckFunc that accepts a resource
+// name, an attribute path, which should use the sentinel value '*' for indexing
+// into a TypeSet. The function verifies that an element matches the whole value
+// map.
+//
+// Use this function over SDK provided TestCheckFunctions when validating a
+// TypeSet where its elements are a nested object with their own attrs/values.
+//
+// Please note, if the provided value map is not granular enough, there exists
+// the possibility you match an element you were not intending to, in the TypeSet.
+// Provide a full mapping of attributes to be sure the unique element exists.
+func TestCheckTypeSetElemNestedAttrs(res, attr string, values map[string]string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		ms := s.RootModule()
+		rs, ok := ms.Resources[res]
+		if !ok {
+			return fmt.Errorf("Not found: %s in %s", res, ms.Path)
+		}
+
+		is := rs.Primary
+		if is == nil {
+			return fmt.Errorf("No primary instance: %s in %s", res, ms.Path)
+		}
+
+		matches := make(map[string]int)
+		attrParts := strings.Split(attr, ".")
+		for stateKey, stateValue := range is.Attributes {
+			stateKeyParts := strings.Split(stateKey, ".")
+			// a Set/List item with nested attrs would have a flatmap address of
+			// at least length 3
+			// foo.0.name = "bar"
+			if len(stateKeyParts) < 3 {
+				continue
+			}
+			var pathMatch bool
+			for i := range attrParts {
+				if attrParts[i] != stateKeyParts[i] && attrParts[i] != sentinelIndex {
+					break
+				}
+				if i == len(attrParts)-1 {
+					pathMatch = true
+				}
+			}
+			if !pathMatch {
+				continue
+			}
+			elementId := stateKeyParts[len(attrParts)-1]
+			nestedAttr := strings.Join(stateKeyParts[len(attrParts):], ".")
+			if v, keyExists := values[nestedAttr]; keyExists && v == stateValue {
+				matches[elementId] = matches[elementId] + 1
+				if matches[elementId] == len(values) {
+					return nil
+				}
+			}
+		}
+
+		return fmt.Errorf("%q no TypeSet element %q, with nested attrs %#v in state: %#v", res, attr, values, is.Attributes)
+	}
+}
+
+// TestCheckTypeSetElemAttr is a resource.TestCheckFunc that accepts a resource
+// name, an attribute path, which should use the sentinel value '*' for indexing
+// into a TypeSet. The function verifies that an element matches the provided
+// value.
+//
+// Use this function over SDK provided TestCheckFunctions when validating a
+// TypeSet where its elements are a simple value
+func TestCheckTypeSetElemAttr(res, attr, value string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		ms := s.RootModule()
+		rs, ok := ms.Resources[res]
+		if !ok {
+			return fmt.Errorf("Not found: %s in %s", res, ms.Path)
+		}
+
+		is := rs.Primary
+		if is == nil {
+			return fmt.Errorf("No primary instance: %s in %s", res, ms.Path)
+		}
+
+		attrParts := strings.Split(attr, ".")
+		for stateKey, stateValue := range is.Attributes {
+			if stateValue == value {
+				stateKeyParts := strings.Split(stateKey, ".")
+				if len(stateKeyParts) == len(attrParts) {
+					for i := range attrParts {
+						if attrParts[i] != stateKeyParts[i] && attrParts[i] != sentinelIndex {
+							break
+						}
+						if i == len(attrParts)-1 {
+							return nil
+						}
+					}
+				}
+			}
+		}
+
+		return fmt.Errorf("%q no TypeSet element %q, with value %q in state: %#v", res, attr, value, is.Attributes)
+	}
+}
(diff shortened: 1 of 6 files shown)


gdavison:
This looks great, @appilon!

How does it handle collections nested inside a `TypeSet`? For example:
* `TestAccAWSCodeBuildProject_Source_Auth` tests for nested values from a `TypeSet` inside another `TypeSet`
* `TestAccAWSMqBroker_allFieldsDefaultVpc` tests for a `TypeSet` inside another `TypeSet`
* `TestAccAWSCloudFrontDistribution_noOptionalItemsConfig` tests for a `TypeList` inside a `TypeSet`

I searched for `"[a-z_]+\.[0-9]{10}\.[a-z_]+\.[0-9]` (the unmatched `"` is expected)

ewbankkit:
Definitely looking forward to this as I have to tweak another set of acceptance test values 😄.
Second the need to ideally extend this to nested collections.

appilon:
Thanks @gdavison 
> TestAccAWSCloudFrontDistribution_noOptionalItemsConfig
This should just work, the "unflattening" part just lifts the path leading up to the Set into a map, so any state past that point should all "just work". I will look for a similar case and verify this, I would try this one but the test mentions deletion takes a while (if you have an example that you know runs relatively quickly please let me know).

> TestAccAWSMqBroker_allFieldsDefaultVpc
> TestAccAWSCodeBuildProject_Source_Auth

😂 I'm not sure how to handle this yet, was deliberately NOT looking for those cases, but I'll focus on this now.

appilon:
@gdavison @ewbankkit The implementation has changed to support nesting sets.

appilon:
🤔 I may still be missing some situations.

gdavison:
Is there a way to check the case of a `TypeSet` inside a `TypeList` where the list has multiple values? For example `TestAccAWSS3Bucket_LifecycleBasic()` and the tests for `"lifecycle_rule.<index>.transition.<item>.<value>"`

I'd like to be able to differentiate the list items, and I'm not sure how I'd do that using the `depth` parameter in `TestCheckTypeSetElemNestedAttrs()`

```go
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.2000431762.date", ""),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.2000431762.days", "30"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.2000431762.storage_class", "STANDARD_IA"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.3601168188.date", ""),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.3601168188.days", "60"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.3601168188.storage_class", "INTELLIGENT_TIERING"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.3854926587.date", ""),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.3854926587.days", "90"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.3854926587.storage_class", "ONEZONE_IA"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.962205413.date", ""),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.962205413.days", "120"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.962205413.storage_class", "GLACIER"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.1571523406.date", ""),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.1571523406.days", "210"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.0.transition.1571523406.storage_class", "DEEP_ARCHIVE"),

resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.2.transition.460947558.days", "0"),

resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.4.transition.460947558.days", "0"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.4.transition.460947558.storage_class", "GLACIER"),

resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.5.transition.460947558.days", "0"),
resource.TestCheckResourceAttr(resourceName, "lifecycle_rule.5.transition.460947558.storage_class", "GLACIER"),
```

appilon:
@gdavison Indeed it appears this implementation wouldn't be able to discern between those list items since the dept is the same 😢 . These are great examples though keep them coming, back to the drawing board!

appilon:
@ewbankkit @gdavison oookay hopefully this covers all cases.

aws/resource_aws_db_parameter_group_test.go
```diff
@@ -132,8 +148,19 @@ func TestAccAWSDBParameterGroup_basic(t *testing.T) {
...
+						"value": "utf8",
+					}),
+					test.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_server",
+						"value": "utf8",
+					}),
+					test.TestCheckTypeSetElemNestedAttrs(resourceName, "parameter.*", map[string]string{
+						"name":  "character_set_client",
+						"value": "utf8",
+					}),
```
gdavison, line 163: This seems to be changing the test

gdavison, line 163: This seems to be changing the test, since there were two items, and no values set for `value`, and now there are three and they have values.. Is that intentional?

appilon: Ah yes this is a bit quirky. There isn't really anyway to represent a specific set element that doesn't exist... My PR proposes helpers to guarantee a single element meets the desired criteria. So I flipped the test to "3 elements with the following property maps, and that the set only contains 3 entries (and not 5)"

appilon: I inferred that this is what was being validated by looking at the test config (goes from 5 parameters to 3)

appilon:
Okay although flaky the last example you mentioned @gdavison has passed 🎉 
```
❯ make testacc TEST=./aws TESTARGS='-run=TestAccAWSS3Bucket_LifecycleBasic'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAWSS3Bucket_LifecycleBasic -timeout 120m
=== RUN   TestAccAWSS3Bucket_LifecycleBasic
=== PAUSE TestAccAWSS3Bucket_LifecycleBasic
=== CONT  TestAccAWSS3Bucket_LifecycleBasic
--- PASS: TestAccAWSS3Bucket_LifecycleBasic (118.38s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	121.065s
```

aws/internal/test/testing.go
```diff
@@ -0,0 +1,114 @@
+package test
```
bflad, line 1: It would be super awesome if these new testing functions had unit testing so they can be updated without needing to verify with various acceptance tests. 👍  I'm also not the biggest fan of the package being named `test` since the function names already relay that (I'd suggest `tfawsresource`), but that's a personal nit since its internal to this module until it is migrated to the SDK.

aws/internal/test/testing.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+const (
+	sentinelIndex = "*"
```
bflad, line 12: Is it worth having the functions return an error if this sentinel value is not present or present more than once?

aws/internal/test/testing.go
appilon: no problem I can change the package name

aws/internal/test/testing.go
appilon: I can return error if the last part in the address isn't `foo.bar.*`, since it always should

PR #13777, opened 2020-06-16
New Data Sources: aws_outposts_outpost(s)

bflad:
### Community Note

```release-note
* **New Data Source:** `aws_outposts_outpost`
* **New Data Source:** `aws_outposts_outposts`
```

```
--- PASS: TestAccAWSOutpostsOutpostDataSource_Id (25.86s)
--- PASS: TestAccAWSOutpostsOutpostDataSource_Name (32.28s)

--- PASS: TestAccAWSOutpostsOutpostsDataSource_basic (21.47s)
```

### DIFF
--- a/aws/data_source_aws_outposts_outpost.go
+++ b/aws/data_source_aws_outposts_outpost.go
@@ -1,0 +1,109 @@
+package aws
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/outposts"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceAwsOutpostsOutpost() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAwsOutpostsOutpostRead,
+
+		Schema: map[string]*schema.Schema{
+			"arn": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"availability_zone": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"availability_zone_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+			},
+			"owner_id": {
+				Type:     schema.TypeString,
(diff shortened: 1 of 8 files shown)


aws/data_source_aws_outposts_outposts.go
```diff
@@ -0,0 +1,103 @@
...
+
+func dataSourceAwsOutpostsOutpostsRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).outpostsconn
+
+	input := &outposts.ListOutpostsInput{}
+
+	var arns, ids []string
+
+	for {
+		output, err := conn.ListOutposts(input)
```
anGie44, line 54: similar comment as above

aws/data_source_aws_outposts_outpost.go
```diff
@@ -0,0 +1,115 @@
...
+
+func dataSourceAwsOutpostsOutpostRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).outpostsconn
+
+	input := &outposts.ListOutpostsInput{}
+
+	var outposts []*outposts.Outpost
+
+	for {
+		output, err := conn.ListOutposts(input)
```
anGie44, line 62: my only comment here would be regarding the `MaxResults` attribute in the `ListOutpostsInput{}` object. if this defaults to 1, I was wondering if this would be an issue at scale, given a service quota/requests limit but from the docs I don't see any reference (https://docs.aws.amazon.com/general/latest/gr/outposts_region.html). so _should_ be OK right?

bflad: Can you point to where this defaults to 1? 😅  I'm not seeing it in the [ListOutposts API Reference](https://docs.aws.amazon.com/outposts/latest/APIReference/API_ListOutposts.html). These do certainly need to be switched to the [paginated SDK function](https://docs.aws.amazon.com/sdk-for-go/api/service/outposts/#Outposts.ListOutpostsPages) though.

bflad:
Fixed merge conflict, switched to paginated list functions, and re-verified:

```
--- PASS: TestAccAWSOutpostsOutpostsDataSource_basic (21.30s)

--- PASS: TestAccAWSOutpostsOutpostDataSource_Name (33.50s)
--- PASS: TestAccAWSOutpostsOutpostDataSource_Id (35.04s)
```

aws/data_source_aws_outposts_outpost.go
anGie44: ahh I was speculating 😅 i figured the min # of results would be used but maybe the # of results is the max available/possible? whoa wouldn't have known by the AWS API ref to use the paginated function 👍 !

Issue #13624, opened 2020-06-04
Add ARN and/or plan time validations to resources/data_sources


anGie44:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

This particular Issue serves as a log of PRs addressing this tech-debt Issue Topic

### ARN Support
- [x] aws_ec2_client_vpn_endpoint 
- [x] aws_glue_connection
- [x] aws_ebs_volume
- [x] aws_api_gateway_rest_api
- [x] aws_wafregional_rate_based_rule
- [x] aws_wafregional_rule_group
- [x] aws_wafregional_rule

### Plan time validations
- [x] aws_ec2_client_vpn_endpoint 
- [x] aws_waf_xss_match_set
- [x] aws_opsworks_application
- [x] aws_spot_fleet_request
- [x] aws_launch_template
- [x] aws_instance

### Relates

* #13601, #13404, #13021, #12777, #12383, #12295, #11486, #10971, #10897, #10896, #10895,

bflad:
@anGie44 is there a definition of done for this issue?

anGie44:
that's a good question @bflad!  I believe this issue tries to address too much 😅 . using a naive grep search 
```
grep -L 'd.Set("arn",' aws/resource*.go | grep -v '_test.go$' | wc -l
```
we still have over 300 resources needed to satisfy just a portion of this issue. 

lmk what your thoughts on this are, but I think we should contain the associated PRs as is (so definition of done will be when those are merged) and then moving forward, we should create enhancement issues on a resource/data-source by resource/data-source basis

anGie44:
Going to close this for the time being as validation changes are captured well enough in changelog files and we aren't monitoring this tech debt item.

Issue #13527, opened 2020-05-28
Replace testAccCheck...Disappears Implementations with testAccCheckResourceDisappears


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Recently, we introduced a shared `TestCheckFunc`, `testAccCheckResourceDisappears()`, that utilizes the existing `helper/schema.Resource.Delete` function to easily perform [disappears testing](https://github.com/terraform-providers/terraform-provider-aws/blob/master/docs/contributing/running-and-writing-acceptance-tests.md#disappears-acceptance-tests) for a resource.

For example, previous disappears tests implemented like:

```go
func TestAccAwsExampleThing_disappears(t *testing.T) {
	// ... other configuration ...

	resource.ParallelTest(t, resource.TestCase{
		// ... other configuration ...
		Steps: []resource.TestStep{
			{
				Config: /* ... */,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckAwsExampleThingExists(resourceName, &thing),
					testAccCheckExampleThingDisappears(&thing),
				),
				ExpectNonEmptyPlan: true,
			},
		},
	})
}

func testAccCheckExampleThingDisappears(/* ... */) resource.TestCheckFunc {
	return func(s *terraform.State) error {
		// ... reimplemented logic to delete resource ...
	}
}
```

Can be replaced with:

```go
func TestAccAwsExampleThing_disappears(t *testing.T) {
	// ... other configuration ...

	resource.ParallelTest(t, resource.TestCase{
		// ... other configuration ...
		Steps: []resource.TestStep{
			{
				Config: /* ... */,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckAwsExampleThingExists(resourceName, &thing),
					testAccCheckResourceDisappears(testAccProvider, resourceAwsExampleThing(), resourceName),
				),
				ExpectNonEmptyPlan: true,
			},
		},
	})
}
```

### Affected Resource(s)

Ideally any submissions for this should be split up per AWS service to ease review and merge process.

- [ ] `aws_accessanalyzer_analyzer`
[...]

PR #13811, opened 2020-06-17
r/ses_receipt_filter - add arn attribute + validations for all arguments + docs

DrFaust92:
### Community Note

Relates #13624, #13527

```release-note
resource_aws_ses_receipt_filter - add arn attribute
resource_aws_ses_receipt_filter - add plan time validations to `name`, `cidr`, and `policy`
resource_aws_ses_receipt_filter - add missing doc for import and attributes
```

```
$ make testacc TESTARGS='-run=TestAccAWSSESReceiptFilter_'
--- PASS: TestAccAWSSESReceiptFilter_basic (46.59s)
--- PASS: TestAccAWSSESReceiptFilter_disappears (30.71s)
```

### DIFF
--- a/aws/resource_aws_ses_receipt_filter.go
+++ b/aws/resource_aws_ses_receipt_filter.go
@@ -30,9 +30,12 @@ func resourceAwsSesReceiptFilter() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
-				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[0-9a-zA-Z][0-9A-Za-z-_]{1,62}[0-9a-zA-Z]$`),
-					"This value can only contain ASCII letters, Start and end with a letter or number,"+
-						" and Contain less than 64 characters"),
+				ValidateFunc: validation.All(
+					validation.StringLenBetween(1, 64),
+					validation.StringMatch(regexp.MustCompile(`^[0-9a-zA-Z_-]+$`), "must contain only alphanumeric, underscore, and hyphen characters"),
+					validation.StringMatch(regexp.MustCompile(`^[0-9a-zA-Z]`), "must begin with a alphanumeric character"),
+					validation.StringMatch(regexp.MustCompile(`[0-9a-zA-Z]$`), "must end with a alphanumeric character"),
+				),
 			},
 
 			"cidr": {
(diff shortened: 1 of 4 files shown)


aws/resource_aws_ses_receipt_filter.go
```diff
@@ -75,21 +93,33 @@ func resourceAwsSesReceiptFilterRead(d *schema.ResourceData, meta interface{}) e
 		return err
 	}
 
-	found := false
-	for _, element := range response.Filters {
```
bflad, line 79: Since the SES API does not provide the ability to fetch a single filter, this looping is unfortunately necessary to handle accounts with multiple Receipt Filters. Otherwise it will always return errors like:

```
    TestAccAWSSESReceiptFilter_basic: testing.go:684: Step 0 error: errors during apply:

        When applying changes to aws_ses_receipt_filter.test, provider "aws" produced
        an unexpected new value for was present, but now absent.

        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
```

You can use a pattern like this to move the filter object up:

```go
var filter 

for _, responseFilter := range response.Filters {
  if aws.StringValue(responseFilter.Name) == d.Id() {
    filter = responseFilter
    break
  }
}

if filter == nil {
  // bye bye
}
```

We should probably add a new covering test for this while we're adjusting this logic, e.g. a test configuration utilizing count and `count.index`, since this resource does especially need to handle multiple resources correctly.

aws/resource_aws_ses_receipt_filter.go
```diff
@@ -19,22 +22,37 @@ func resourceAwsSesReceiptFilter() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[0-9a-zA-Z][0-9A-Za-z-_]{1,62}[0-9a-zA-Z]$`),
+					"This value can only contain ASCII letters, Start and end with a letter or number,"+
+						" and Contain less than 64 characters"),
```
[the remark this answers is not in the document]
DrFaust92: You're right, ill split these, it took me 20 mins to play with a regex checker to build this 😿

aws/resource_aws_ses_receipt_filter.go
```diff
@@ -75,21 +96,34 @@ func resourceAwsSesReceiptFilterRead(d *schema.ResourceData, meta interface{}) e
...
 	}
 
-	found := false
-	for _, element := range response.Filters {
-		if *element.Name == d.Id() {
-			d.Set("cidr", element.IpFilter.Cidr)
-			d.Set("policy", element.IpFilter.Policy)
-			d.Set("name", element.Name)
-			found = true
+	var filter *ses.ReceiptFilter
```
DrFaust92, line 99: @bflad , is that what you had in mind?

aws/resource_aws_ses_receipt_filter.go
DrFaust92: ive made a small change, let me know what you think. i can revert this to the original for now with minor changes

Issue #11801, opened 2020-01-29
Order is lost for data `aws_iam_policy_document` when applied to S3 buckets, iam roles, kms keys, etc


nitrocode:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

See also #21968 

### Terraform Version

```
✗ terraform --version
Terraform v0.12.20
+ provider.aws v2.46.0
```

### Affected Resource(s)

* `aws_ecr_repository_policy`
* `aws_kms_key`
* `aws_s3_bucket`
* `aws_launch_configuration`
* `aws_lb_listener`
* `aws_codeartifact_domain_pernission_policy`

### Terraform Configuration Files

If running locally, please add your own real roles to `local.secrets_roles` otherwise the kms policy will fail.

```hcl
provider "aws" {
  region = "us-east-1"
}

data "aws_caller_identity" "current" {}

locals {
  id = data.aws_caller_identity.current.account_id
  secrets_roles = [
    "asnip",
    "esnip",
    "lsnip",
    "psnip",
    "wsnip",
    "csnip",
#    "default",
  ]

  secrets_roles_arns = sort([
    for role in local.secrets_roles :
    "arn:aws:iam::${local.id}:role/${role}"
  ])

  secrets_admin = [
    "arn:aws:iam::${local.id}:role/sre"
  ]
}

data "aws_iam_policy_document" "secrets" {
  policy_id = "secrets-policy"

  statement {
    sid       = "Enable IAM User Permissions"
    effect    = "Allow"
    resources = ["*"]
    actions = [
      "kms:Create*",
      "kms:Describe*",
      "kms:Enable*",
      "kms:List*",
      "kms:Put*",
      "kms:Update*",
      "kms:Revoke*",
      "kms:Disable*",
      "kms:Get*",
      "kms:Delete*",
      "kms:ScheduleKeyDeletion",
      "kms:CancelKeyDeletion",
    ]

    principals {
      type        = "AWS"
      identifiers = local.secrets_admin
    }
  }

  statement {
    sid       = "Allow use of the key"
    effect    = "Allow"
    resources = ["*"]

[...]

PR #13813, opened 2020-06-17
Prevent spurious diffs in IAM policies

jbergknoff-rival:
IAM policies have several fields that are treated as arrays by the AWS API and Terraform, but are actually sets with no meaningful order (examples: actions, principals, resources). As a result, the order of the results returned by the AWS API can vary from API call to API call. This provider makes liberal use of `diff_suppress_funcs.suppressEquivalentAwsPolicyDiffs` which can suppress such a diff **if the policy isn't changing for any other reason**. However, if the policy **is** changing, then we will (correctly) get a plan, but it will include spurious diffs due to the arbitrary ordering of these sets.

One solution, as in https://github.com/terraform-providers/terraform-provider-aws/pull/11463, is to normalize the policy document as it's read from AWS, and normalize it as it's serialized to Terraform state. This PR does that for several resources (the ones where I've noted the need, though this is probably not exhaustive).

Here's an example:

```
provider "aws" {
  region = "us-east-1"
}

resource "aws_iam_role" "test" {
  name = "test-role"

  assume_role_policy = jsonencode(
    {
      Version = "2012-10-17"
      Statement = [
        {
          Action = ["sts:AssumeRole"],
          Effect = "Allow",
          Principal = {
            AWS = [
              # NOTE: These account ids need to be replaced with valid ids for IAM to accept the policy
              "arn:aws:iam::111111111111:root",
              "arn:aws:iam::222222222222:root",
              "arn:aws:iam::333333333333:root",
              "arn:aws:iam::444444444444:root",
              "arn:aws:iam::555555555555:root",
            ]
          }
        }
      ]
    }
  )
}
```

Apply this, and then remove the middle principal and plan again. Without this patch, the plan has a random element to it, but it can look like:

```
~ Principal = {
    ~ AWS = [
        + "arn:aws:iam::111111111111:root",
        + "arn:aws:iam::222222222222:root",
          "arn:aws:iam::444444444444:root",
          "arn:aws:iam::555555555555:root",
        - "arn:aws:iam::333333333333:root",
        - "arn:aws:iam::222222222222:root",
        - "arn:aws:iam::111111111111:root",
      ]
  }
```

With this patch, the plan is:

```
~ Principal = {
    ~ AWS = [
          "arn:aws:iam::111111111111:root",
          "arn:aws:iam::222222222222:root",
        - "arn:aws:iam::333333333333:root",
          "arn:aws:iam::444444444444:root",
          "arn:aws:iam::555555555555:root",
      ]
  }
```

It would be ideal to centralize this logic, but I don't see a clear way to do that, and I'd appreciate suggestions.

Also, this PR actually doesn't touch actions and resources in the policy because they're not given any structure in `iam_policy_model.go` (i.e. just left as `interface{}`). With some guidance on how this should look, I'd be happy to make that change, as well.

### Community Note

```release-note
BUG FIXES:
* Suppress spurious diffs due to arbitrary ordering of IAM policies
```

Unfortunately, as in https://github.com/terraform-providers/terraform-provider-aws/pull/11463, this doesn't fit the mold of most changes as far as testing goes. I think the most meaningful test would be one that inspects the contents of the plan, and I don't know how to do that with the testing framework.

### DIFF
--- a/aws/iam_policy_model.go
+++ b/aws/iam_policy_model.go
@@ -38,6 +38,39 @@ type IAMPolicyStatementCondition struct {
 type IAMPolicyStatementPrincipalSet []IAMPolicyStatementPrincipal
 type IAMPolicyStatementConditionSet []IAMPolicyStatementCondition
 
+func parseIAMPolicyDoc(b []byte) (*IAMPolicyDoc, error) {
+	doc := &IAMPolicyDoc{}
+	if err := json.Unmarshal(b, &doc); err != nil {
+		return nil, err
+	}
+
+	return doc, nil
+}
+
+func serializeIAMPolicyDoc(doc *IAMPolicyDoc) (string, error) {
+	b, err := json.Marshal(&doc)
+	if err != nil {
+		return "", err
+	}
+
+	return string(b), nil
+}
+
+func normalizeIAMPolicyDoc(policy string) (string, error) {
+	doc, err := parseIAMPolicyDoc([]byte(policy))
+	if err != nil {
+		return "", err
+	}
+
+	for _, s := range doc.Statements {
+		s.Principals.normalize()
+		s.NotPrincipals.normalize()
+		s.Conditions.normalize()
+	}
+
+	return serializeIAMPolicyDoc(doc)
+}
+
 func (s *IAMPolicyDoc) Merge(newDoc *IAMPolicyDoc) {
 	// adopt newDoc's Id
 	if len(newDoc.Id) > 0 {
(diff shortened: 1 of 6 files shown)


shanestarcher-okta:
We are attempting to diff plans between resources and due to the current non-deterministic nature, it makes the plan non-viable to diff in these cases.  Making it deterministic as this does would I believe solve the problem.

ewbankkit:
Related:

* #11801
* https://github.com/jen20/awspolicyequivalence/issues/10

@jbergknoff-rival Thanks for the contribution :tada: :clap:.
I am going to close this PR as we would like a more general solution implemented in `awspolicyequivalence`.

PR #13831, opened 2020-06-18
Rewrite TypeSet checks with new helpers [Part 2]

appilon:
These checks have been rewritten using the new helpers which avoids using set hashes. The helpers have somewhat exhaustive unit tests so the hope is this can be merged in chunks without waiting on acceptance tests.

### Community Note

Relates #13556 

```release-note
NONE
```

```

...
```

### DIFF
--- a/aws/resource_aws_autoscaling_group_test.go
+++ b/aws/resource_aws_autoscaling_group_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-aws/aws/internal/tfawsresource"
 )
 
 func init() {
@@ -86,10 +87,9 @@ func TestAccAWSAutoScalingGroup_basic(t *testing.T) {
 	randName := fmt.Sprintf("terraform-test-%s", acctest.RandString(10))
 
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:            func() { testAccPreCheck(t) },
-		Providers:           testAccProviders,
-		CheckDestroy:        testAccCheckAWSAutoScalingGroupDestroy,
-		DisableBinaryDriver: true,
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSAutoScalingGroupDestroy,
 		Steps: []resource.TestStep{
 			{
 				Config: testAccAWSAutoScalingGroupConfig(randName),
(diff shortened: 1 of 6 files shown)


aws/resource_aws_autoscaling_group_test.go
```diff
@@ -1007,10 +1003,11 @@ func TestAccAWSAutoScalingGroup_initialLifecycleHook(t *testing.T) {
...
-						"aws_autoscaling_group.bar", "initial_lifecycle_hook.391359060.default_result", "CONTINUE"),
-					resource.TestCheckResourceAttr(
-						"aws_autoscaling_group.bar", "initial_lifecycle_hook.391359060.name", "launching"),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs("aws_autoscaling_group.bar", "initial_lifecycle_hook.*", map[string]string{
+						"default_result": "CONTINUE",
+						"name":           "launching",
+					}),
+					// TODO: TypeSet hash
 					testAccCheckAWSAutoScalingGroupInitialLifecycleHookExists(
 						"aws_autoscaling_group.bar", "initial_lifecycle_hook.391359060.name"),
```
anGie44, line 1012: should we still keep `DisableBinaryDriver: true,` to be able to support this check for the time being?

aws/resource_aws_cognito_user_pool_test.go
```diff
@@ -944,32 +944,35 @@ func TestAccAWSCognitoUserPool_withSchemaAttributes(t *testing.T) {
...
-					resource.TestCheckResourceAttr(resourceName, "schema.770828826.name", "mybool"),
-					resource.TestCheckResourceAttr(resourceName, "schema.770828826.number_attribute_constraints.#", "0"),
-					resource.TestCheckResourceAttr(resourceName, "schema.770828826.required", "false"),
-					resource.TestCheckResourceAttr(resourceName, "schema.770828826.string_attribute_constraints.#", "0"),
+					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "schema.*", map[string]string{
+						"attribute_data_type":            "String",
+						"developer_only_attribute":       "false",
+						"mutable":                        "false",
+						"name":                           "email",
+						"number_attribute_constraints.#": "0",
```
anGie44, line 961: 👏 didn't you these count checks were supported !

aws/resource_aws_autoscaling_group_test.go
appilon: Yes we could, I will switch it back

anGie44: oh ok! wasn't sure if it would cause problems while running the tests in the binary driver testing environment in TC

appilon: the test would just error if we switched to V2 of the SDK today is all. `DisableBinaryDriver` is not an option for V2 of the SDK, but keeping it set to true is a good breadcrumb for me (even with the TODO)

PR #13891, opened 2020-06-23
FAQ.md  Link reactions to sorted PRs and issues

abitrolly:
```release-note
NONE
```

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -17,7 +17,7 @@ Unfortunately, due to the volume of issues and new pull requests we receive, we 
 
 ### How do you decide what gets merged for each release?
 
-The number one factor we look at when deciding what issues to look at are your 👍 [reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue/PR description as these can be easily discovered. Comments that further explain desired use cases or poor user experience are also heavily factored. The items with the most support are always on our radar, and we commit to keep the community updated on their status and potential timelines.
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -17,7 +17,7 @@ Unfortunately, due to the volume of issues and new pull requests we receive, we
 
 ### How do you decide what gets merged for each release?
 
-The number one factor we look at when deciding what issues to look at are your 👍 [reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue/PR description as these can be easily discovered. Comments that further explain desired use cases or poor user experience are also heavily factored. The items with the most support are always on our radar, and we commit to keep the community updated on their status and potential timelines.
+The number one factor we look at when deciding what issues to look at are your 👍 [reactions](https://github.com/terraform-providers/terraform-provider-aws/issues?q=is%3Aopen+sort%3Areactions-%2B1-desc) to the original issue/PR description as these can be easily discovered. Comments that further explain desired use cases or poor user experience are also heavily factored. The items with the most support are always on our radar, and we commit to keep the community updated on their status and potential timelines.
```
bflad, line 20: Hi @abitrolly 👋  Thank you for submitting this. I believe we would prefer to keep the original link in this case to show those unfamiliar with how to create reactions, but it would be nice to include the upvote list link as well -- can you keep the original link where it was and maybe convert "easily discoverable" to that link?

```suggestion
The number one factor we look at when deciding what issues to look at are your 👍 [reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue/PR description as these can be [easily discovered](../issues?q=is%3Aopen+sort%3Areactions-%2B1-desc). Comments that further explain desired use cases or poor user experience are also heavily factored. The items with the most support are always on our radar, and we commit to keep the community updated on their status and potential timelines.
```

PR #13907, opened 2020-06-23
Enable UltraWarm in-place

edma2:
Since GA, you can enable UltraWarm in-place on existing domains: https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/ultrawarm.html#ultrawarm-enable. This makes [ForceNew](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#forcenew) unnecessary. Remove `ForceNew` so it does not destroy and recreate the underlying resource.

### Community Note

```release-note
Enable Elasticsearch UltraWarm in-place.
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSElasticSearchDomain_warm'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAWSElasticSearchDomain_warm -timeout 120m
=== RUN   TestAccAWSElasticSearchDomain_warm
=== PAUSE TestAccAWSElasticSearchDomain_warm
=== CONT  TestAccAWSElasticSearchDomain_warm
--- PASS: TestAccAWSElasticSearchDomain_warm (4418.55s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	4420.811s
```

### DIFF
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -242,7 +242,6 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 						"warm_enabled": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							ForceNew: true,
 						},
 						"warm_count": {
 							Type:         schema.TypeInt,
--- a/aws/resource_aws_elasticsearch_domain_test.go
+++ b/aws/resource_aws_elasticsearch_domain_test.go
@@ -190,6 +190,15 @@ func TestAccAWSElasticSearchDomain_warm(t *testing.T) {
 		CheckDestroy: testAccCheckESDomainDestroy,
 		Steps: []resource.TestStep{
 			{
+				Config: testAccESDomainConfigWarm(rName, "ultrawarm1.medium.elasticsearch", false, 6),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckESDomainExists(resourceName, &domain),
+					resource.TestCheckResourceAttr(resourceName, "cluster_config.0.warm_enabled", "false"),
+					resource.TestCheckResourceAttr(resourceName, "cluster_config.0.warm_count", "6"),
+					resource.TestCheckResourceAttr(resourceName, "cluster_config.0.warm_type", "ultrawarm1.medium.elasticsearch"),
+				),
+			},
+			{
 				Config: testAccESDomainConfigWarm(rName, "ultrawarm1.medium.elasticsearch", true, 6),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckESDomainExists(resourceName, &domain),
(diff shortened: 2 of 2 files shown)


aws/resource_aws_elasticsearch_domain.go
```diff
@@ -242,7 +242,6 @@ func resourceAwsElasticSearchDomain() *schema.Resource {
 						"warm_enabled": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							ForceNew: true,
```
amulyas, line 245: this looks good for in-place upgrade

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -1112,7 +1121,37 @@ resource "aws_elasticsearch_domain" "test" {
 `, rName, zoneAwarenessEnabled)
 }
 
-func testAccESDomainConfigWarm(rName, warmType string, enabled bool, warmCnt int) string {
+func testAccESDomainConfigWarmEnabled(rName, warmType string, warmCnt int) string {
```
DrFaust92, line 1124: I think this can changed to be a variable in the original config instead of a separate one.

amulyas:
👍

aws/resource_aws_elasticsearch_domain_test.go
```diff
@@ -1113,7 +1122,36 @@ resource "aws_elasticsearch_domain" "test" {
 }
 
 func testAccESDomainConfigWarm(rName, warmType string, enabled bool, warmCnt int) string {
-	return fmt.Sprintf(`
+	if enabled {
```
DrFaust92, line 1125: can you flatten this to a single config with no if? if warm is disabled it will ignore the other params (you can put 0 and "")

DrFaust92: im taking this comment back. ill let maintainer decide on this. im split on what is the way to go. may be splitting this is the right way to go /:

DrFaust92: @gdavison can you give your input here?

edma2: I tried it with a single config at first but the test failed. After applying the configuration the state file stores `warm_type` as `""` and `warm_count` as `0`, because as you said, those parameters are ignored during application when `warm_enabled` is false. The test then fails because a follow-up plan phase returns a non-empty diff. Yet, we cannot set `warm_type` or `warm_count` to zero values because we constrain them to specific values.

So the solution here is to not set those optional parameters.

gdavison: It looks like the only difference in the two configurations is that when `warm_enabled` is `true`, it also includes `warm_count` and `warm_type`. So you could do something like

```go
warm_config := ""
if enabled {
    warm_config = fmt.Sprintf(`
warm_count = %[1]d
warm_type. = %[2]q
`, warmCount, warmType)
}
```

In the config, then you could put
```go
fmt.Sprintf(`
...
cluster_config {
  ...
  warm_enabled = %t
  %s
  ...
}
...
`, rName, warmEnabled, warmConfig)
```

edma2: that's certainly a lot less repetitive, thanks for the suggestion. I'll try to push an update before the cut off.

PR #14033, opened 2020-07-02
tests/provider: Initial goreleaser configuration and CI workflow

### LINKED ISSUE #8988: Add Cross Compiler Checks to CI
### Community Note

### Description

Update the Travis-CI configuration and project make file `GNUmakefile` with a build step that runs a cross-comipler build for Terraform supported operating systems: Linux, Darwin, Windows, FreeBSD, OpenBSD, Solaris. This check would help catch potential issues with newly updated dependencies before they are merged into master.

An attempt for this check was made on #8458, but it was closed due to the time needed to cross compile for each of the platforms. For anyone picking up this request, the base requirements are as follows:

- [ ] Can run on Travis-CI within a containerized build.
- [ ] Runs as a separate build stage [Travis Matrix Builds](https://docs.travis-ci.com/user/build-matrix/)
- [ ] Runs on PR builds only
- [ ] Total build time for all phases does not exceed 10 minutes 

### New or Affected Files(s)

* .travis.yml

### References

* #0000

bflad:
### Community Note

Closes #8988
Reference: https://www.terraform.io/docs/registry/providers/publishing.html

```release-note
NONE
```

Going forward, the Terraform Provider release process for the Terraform Registry will recommend `goreleaser` for all providers, including the official ones released via a separate process today. Given this decision is now more final, we can begin the process of using this tool for CI. In the future, we can also support nightly builds and later releases via the new GitHub release process. For now though, we leverage it just to verify codebase cross-compilation for all platforms that will be supported after the upcoming 3.0.0 release.

Output from acceptance testing: N/A (CI testing)

### DIFF
--- a/.github/workflows/terraform_provider.yml
+++ b/.github/workflows/terraform_provider.yml
@@ -247,7 +247,7 @@ jobs:
       - name: goreleaser build
         uses: goreleaser/goreleaser-action@v2
         with:
-          args: build --snapshot
+          args: build -p 1 --snapshot --timeout 1h
 
   tfproviderdocs:
     needs: [terraform_providers_schema]
--- a/.github/workflows/terraform_provider.yml
+++ b/.github/workflows/terraform_provider.yml
@@ -241,13 +241,14 @@ jobs:
           path: ~/go/pkg/mod
           key: ${{ runner.os }}-go-pkg-mod-${{ hashFiles('go.sum') }}
       - name: goreleaser check
+        continue-on-error: true
         uses: goreleaser/goreleaser-action@v2
         with:
           args: check
       - name: goreleaser build
         uses: goreleaser/goreleaser-action@v2
         with:
-          args: build -p 1 --snapshot --timeout 1h
+          args: build --snapshot --timeout 1h
 
   tfproviderdocs:
     needs: [terraform_providers_schema]
--- a/.github/workflows/terraform_provider.yml
+++ b/.github/workflows/terraform_provider.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - .github/workflows/terraform_provider.yml
       - .golangci.yml
+      - .goreleaser.yml
       - aws/**
       - awsproviderlint/**
       - docs/index.md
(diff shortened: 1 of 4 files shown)


.github/workflows/terraform_provider.yml
```diff
@@ -225,6 +226,29 @@ jobs:
...
+      - uses: actions/cache@v2
+        continue-on-error: true
+        timeout-minutes: 2
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-pkg-mod-${{ hashFiles('go.sum') }}
+      - name: goreleaser check
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          args: check
```
radeksimko, line 246: FYI: GoReleaser will return non-zero exit code here when it detects deprecated options in the config, which is probably good to catch either way, but you might still want to `continue-on-error` in such case as the config may otherwise be valid and the next step can still check if it builds on all platforms.

.goreleaser.yml
```diff
@@ -0,0 +1,42 @@
+archives:
+  - name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+    format: zip
+before:
+  hooks:
+    - go mod download
+builds:
+  - binary: '{{ .ProjectName }}_{{ .Version }}'
+    env:
+      - CGO_ENABLED=0
```
radeksimko, line 10: Not sure if `go mod download` does anything that this may affect, but you might want to move it to the top-level `env` section, so that any other `go` commands outside of `builds` can pick it up too.

.goreleaser.yml
```diff
@@ -0,0 +1,42 @@
+archives:
+  - name_template: '{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}'
+    format: zip
```
radeksimko, line 3: GoReleaser packages `README` and `LICENSE` files by default here, which you may not want - at least I think we haven't packaged anything other than binaries so far for any providers.

```yaml
files: # Avoid packaging any extra (default) files
      - none*
```

bflad:
Thanks, @radeksimko! Adjusted per your recommendations. Also going to see if re-enabling build concurrency is a good or bad idea, since the timeout is now fixed to match GitHub Actions'.

.goreleaser.yml
```diff
@@ -0,0 +1,44 @@
...
+    goos:
+      - darwin
+      - freebsd
+      - linux
+      - windows
+    ignore:
+      - goarch: '386'
+        goos: darwin
+    ldflags:
+      - -s -w -X github.com/terraform-providers/terraform-provider-aws/aws/version.ProviderVersion={{.Version}}
```
paultyng, line 25: You are probably going to want `-trimpath` for `flags` here as well as overriding the `mod_timestamp`, see these very helpful instructions an awesome contributor put together for Goreleaser: https://goreleaser.com/customization/build/#reproducible-builds

bflad: Is https://github.com/terraform-providers/terraform-provider-aws/pull/14033/commits/aa78000afa74380bb785fae709c8a4559a9d2f49 what you had in mind here? 😄

paultyng: Yep, I can't really speak to the ldflags -X change, but the other parts look good.

PR #14036, opened 2020-07-02
Add static check for hardcoded region

YakDriver:
### Community Note

```release-note
NONE
```

Example output from testing:

```console
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_security_group_rule_test.go:1797:59: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_security_group_test.go:1526:34: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_security_group_test.go:1813:34: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_shield_protection_test.go:263:21: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_shield_protection_test.go:530:21: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_sns_platform_application_test.go:138:22: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_sns_platform_application_test.go:139:22: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead
/Users/yakdriver/go/src/github.com/terraform-providers/terraform-provider-aws/aws/resource_aws_sns_platform_application_test.go:145:22: AWSAT003: regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead

...
```


awsproviderlint/passes/AWSAT003/AWSAT003.go
```diff
@@ -0,0 +1,65 @@
...
+func run(pass *analysis.Pass) (interface{}, error) {
+	ignorer := pass.ResultOf[commentignore.Analyzer].(*commentignore.Ignorer)
+	inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
+
+	nodeFilter := []ast.Node{
+		(*ast.BasicLit)(nil),
+	}
+
+	majorRegions := "ap|eu|sa|us"
+	directions := "central|east|northeast|south|southeast|west"
```
bflad, line 45: Rather than manually maintaining these lists, my suggestion would be to use the [AWS Go SDK `endpoints` package](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/) to enumerate all the potential values. 👍

awsproviderlint/passes/AWSAT003/README.md
```diff
@@ -0,0 +1,44 @@
+# AWSAT003
```
bflad, line 1: If this check is going to be dual purpose (generally the previous checks have not been), it should show all examples  of flagged/passing code. e.g. using the `aws_region` data source.

awsproviderlint/passes/AWSAT003/testdata/src/a/main.go
```diff
@@ -0,0 +1,52 @@
+package a
+
+import (
+	"fmt"
+)
+
+func f() {
+	/* Passing cases */
```
bflad, line 8: It might be nice to show a passing example using one of the [`endpoints` package regional constants](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants), which could be an easy way to bypass the linter failure in normal code. e.g.

```
endpoints.UsEast1RegionID
```

awsproviderlint/passes/AWSAT003/AWSAT003.go
```diff
@@ -0,0 +1,65 @@
...
+
+		if x.Kind != token.STRING {
+			return
+		}
+
+		if !re.MatchString(x.Value) {
+			return
+		}
+
+		pass.Reportf(x.ValuePos, "%s: regions and AZs should not be hardcoded", analyzerName)
```
bflad, line 62: It might be nice to give developers a hint about how to fix these in the message as well, e.g.

```suggestion
		pass.Reportf(x.ValuePos, "%s: regions and AZs should not be hardcoded, use the aws_availability_zones and aws_region data sources instead", analyzerName)
```

awsproviderlint/passes/AWSAT003/README.md
YakDriver: It will find regions, even if they are part of an AZ so perhaps the designation as dual purpose is not necessary.

awsproviderlint/passes/AWSAT003/AWSAT003.go
```diff
@@ -0,0 +1,73 @@
...
+	partitions := resolver.(endpoints.EnumPartitions).Partitions()
+	var regions []string
+
+	for _, p := range partitions {
+		for id, _ := range p.Regions() {
+			regions = append(regions, id)
+		}
+	}
+
+	re := regexp.MustCompile(strings.Join(regions, "|"))
```
bflad, line 54: Ha here's a (relatively minor/fun) false positive:

https://github.com/terraform-providers/terraform-provider-aws/blob/b77516e01695792da30c58b73924e79c753180df/aws/resource_aws_codebuild_project_test.go#L3460

I think the only way to workaround that would be to look for double quotes or something:

```go
regexp.MustCompile(fmt.Sprintf(`"(%s)[a-z]?", strings.Join(regions, "|")))
```

But I'm guessing that could also lead to it missing some things, like when a region/AZ is hardcoded in the string for a TestCheck. 🤔  Not a big deal, just noting.

awsproviderlint/passes/AWSAT003/README.md
bflad: Very true! 👍

awsproviderlint/passes/AWSAT003/AWSAT003.go
YakDriver: I figure there will be a few places that need to use ignore.

PR #14079, opened 2020-07-07
resource/instance: add private_ips field

### LINKED ISSUE #7063: Cannot create eth0 NIC with secondary IP while creating EC2 instance
_This issue was originally opened by @PeterP55P as hashicorp/terraform#19937. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,
I'm trying to create NIC with index 0 while creating EC2 instance with 1 public and 2 privite IPs. All is fine if I'm doing 1 pub and 1 priv.  In my case I need to use 2 private IPs which is possible from AWS GUI but there is no such option in any resource from terraform.

### Current Terraform Version
Terraform v0.11.8
+ provider.aws v1.50.0

### Proposal
Please add an option for adding second IPv4 IP address while creating EC2 instance together with public IP.
- hashicorp/terraform#6017

Thank you,
Peter

anGie44:
### Community Note

Closes #7063

```release-note
resource/instance: add private_ips field
```

```
--- PASS: TestAccAWSInstance_basic (189.80s)
--- PASS: TestAccAWSInstance_primaryNetworkInterface (114.29s)
--- PASS: TestAccAWSInstance_primaryNetworkInterfaceSourceDestCheck (124.33s)
--- PASS: TestAccAWSInstance_NewNetworkInterface_EmptyPrivateIPAndSecondaryPrivateIPs (179.40s)
--- PASS: TestAccAWSInstance_NewNetworkInterface_PrivateIPAndSecondaryPrivateIPs (192.43s)
--- PASS: TestAccAWSInstance_NewNetworkInterface_PrivateIPAndSecondaryPrivateIPsUpdate (264.68s)
--- PASS: TestAccAWSInstance_NewNetworkInterface_PublicIPAndSecondaryPrivateIPs (274.98s)
--- PASS: TestAccAWSInstance_NewNetworkInterface_EmptyPrivateIPAndSecondaryPrivateIPsUpdate (275.76s)
```
```
--- PASS: TestAccAWSInstanceDataSource_secondaryPrivateIPs (81.88s)
```

### DIFF
--- a/aws/resource_aws_instance.go
+++ b/aws/resource_aws_instance.go
@@ -113,19 +113,16 @@ func resourceAwsInstance() *schema.Resource {
 					validation.StringIsEmpty,
 					validation.IsIPv4Address,
 				),
-				ConflictsWith: []string{"private_ips"},
 			},
 
-			"private_ips": {
+			"secondary_private_ips": {
 				Type:     schema.TypeSet,
 				Optional: true,
-				ForceNew: true,
 				Computed: true,
 				Elem: &schema.Schema{
 					Type:         schema.TypeString,
 					ValidateFunc: validation.IsIPv4Address,
 				},
-				ConflictsWith: []string{"private_ip"},
 			},
 
 			"source_dest_check": {
@@ -219,7 +216,7 @@ func resourceAwsInstance() *schema.Resource {
 			},
 
 			"network_interface": {
-				ConflictsWith: []string{"associate_public_ip_address", "subnet_id", "private_ip", "private_ips", "vpc_security_group_ids", "security_groups", "ipv6_addresses", "ipv6_address_count", "source_dest_check"},
+				ConflictsWith: []string{"associate_public_ip_address", "subnet_id", "private_ip", "secondary_private_ips", "vpc_security_group_ids", "security_groups", "ipv6_addresses", "ipv6_address_count", "source_dest_check"},
 				Type:          schema.TypeSet,
 				Optional:      true,
 				Computed:      true,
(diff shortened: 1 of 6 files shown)


aws/resource_aws_instance.go
```diff
@@ -113,6 +113,19 @@ func resourceAwsInstance() *schema.Resource {
 					validation.StringIsEmpty,
 					validation.IsIPv4Address,
 				),
+				ConflictsWith: []string{"private_ips"},
+			},
+
+			"private_ips": {
+				Type:     schema.TypeSet,
```
bflad, line 120: Does this need to be `TypeList` to support the "primary" private IP being the first element? I believe the concept of the primary private IP is [important for DNS hostnames](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#concepts-private-addresses). Not sure about the API response semantics of ordering once it is configured.

anGie44: ah yes thats more sense! i took the idea from the `network_interface` resource which uses a TypeSet as well but didn't think about the response ordering. Is it still ok to have the `private_ip` and `private_ips` conflict as they are set right now? or would be better to introduce the breaking change and use the ordering of the list to do determined primary vs. secondary?

anGie44: Ahh but then we'd have to still support the empty string in the list...hmm id side with having the 2 fields instead

anGie44: Also I'm rethinking the design here given the comment in the original issue https://github.com/terraform-providers/terraform-provider-aws/issues/7063#issuecomment-547513390 . if user is attaching interfaces via the `network_interface` blocks, should the private_ips list be applicable to these network_interfaces too? i.e. `private_ips` no longer conflicts with `network_interface` ?

anGie44: thinking about this more, in 9fc6e9c i've enabled the use of `private_ip` with the renamed `secondary_private_ips` to be more specific and not make things more confusing with IPs specified in the set vs. that provided in `private_ip`

PR #14099, opened 2020-07-08
reverse the order the parameter diff

wburn-sierra-cedar:
do the compare with aws and current  first to remove deleted params
then compare with current and aws to modify  params

### Community Note

Relates to https://github.com/terraform-providers/terraform-provider-aws/pull/5728

```release-note
Reversed the order of the diffs for how db parameters are updated.  The current order leaves a state where if parameters are updated and deleted in the same apply, updated params get the current aws value as opposed to the modified value.  This persists in state and the only fix is to delete the parameter group.  Changing the order of the sequence fixes this bug.

```

```

...
```


bflad:
Hi @wburn-sierra-cedar 👋  Thank you for this contribution. It appears the earlier #12112 was recently merged and released which seems to cover this functionality, so I'm going to close this. Please reach out if that is not the case here. 👍

PR #14068, opened 2020-07-06
New resource: aws_ec2_managed_prefix_list

roberth-k:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request

Closes #13986

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note
FEATURES:
- **New Resource:** `aws_ec2_managed_prefix_list`

ENHANCEMENTS:
- resource/aws_security_group: Update documentation with managed prefix lists in mind
- resource/aws_security_group_rule: Update documentation with managed prefix lists in mind
```

Update to route tables expected with https://github.com/terraform-providers/terraform-provider-aws/pull/14013.

Output from acceptance testing:

```
$ make testacc TEST=./aws TESTARGS=-run=TestAccAwsEc2ManagedPrefixList
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAwsEc2ManagedPrefixList -timeout 120m
--- PASS: TestAccAwsEc2ManagedPrefixList_disappears (20.64s)
--- PASS: TestAccAwsEc2ManagedPrefixList_exceedLimit (28.56s)
--- PASS: TestAccAwsEc2ManagedPrefixList_name (38.95s)
--- PASS: TestAccAwsEc2ManagedPrefixList_basic (41.02s)
--- PASS: TestAccAwsEc2ManagedPrefixList_tags (78.83s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       80.429s
```

Issue #13986, opened 2020-06-29
Support New Resource and Data Source for EC2 Managed Prefix Lists


ewbankkit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

> Virtual Private Cloud (VPC) customers can now create and manage their own Prefix Lists to simplify VPC configurations.

##### New

* `aws_ec2_managed_prefix_list`

##### Affected

* `aws_route_table`
* `aws_route`

### Potential Terraform Configuration

```hcl
resource "aws_ec2_managed_prefix_list" "example" {
  address_family = "IPv4"
  max_entries    = 1
  
  entry {
    cidr        = "0.0.0.0/0"
    description = "Everywhere"
  }
}
```

### References

[Announcement](https://aws.amazon.com/about-aws/whats-new/2020/06/amazon-virtual-private-cloud-customers-use-prefix-lists-simplify-configuration-security-groups-route-tables/).
[User guide](https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html).

Requires [AWS SDK v1.32.12](https://github.com/aws/aws-sdk-go/releases/tag/v1.32.12):
* https://github.com/terraform-providers/terraform-provider-aws/pull/13803

roberth-k:
I'll work on a PR for this.

Additional affected resources:

`aws_security_group`
`aws_security_group_rule`

ewbankkit:
@roberth-k Thanks for offering to tackle this.
I am working on some other fixes/enhancements for routes and route tables and can do the work to change `destination_prefix_list_id` from a `Computed`-only attribute to an `Optional` and `Computed` attribute.
Let us know if you need any assistance.

roberth-k:
Thanks @ewbankkit -- if you could update `destination_prefix_list_id` in `aws_route` it would be helpful. This attribute should be added to the matching data resource as well. One of the vendor prefix lists such as `com.amazonaws.eu-west-1.s3` (via `data_source_aws_prefix_list`) should work for acceptance testing.

yermulnik:
[...]

PR #14109, opened 2020-07-09
d/aws_prefix_list: fixes and enhancements

roberth-k:
### Community Note

Relates #14068, #13986

Contains bugfixes and enhancements to the `aws_prefix_list` data source discovered while implementing support for Managed Prefix Lists.

Users may be affected by the following changes in behavior:

- Where no criteria have been provided to the data source, the data source currently produces an arbitrary prefix list. After this PR, it will raise an error.
- Where a resource depends on the API ordering of CIDR blocks (such as by selecting `cidr_blocks[0]`), the result may now be a different CIDR block.

```release-note
ENHANCEMENTS:
- data-source/aws_prefix_list: The `cidr_blocks` attribute is now lexically sorted
- data-source/aws_prefix_list: Add documentation example about finding a prefix list by name

BUG FIXES:
- data-source/aws_prefix_list: Raise error when criteria match more than one prefix list
- data-source/aws_prefix_list: Specifying the `name` attribute no longer overwrites all filters
```

```
$ make testacc TEST=./aws TESTARGS=-run=TestAccDataSourceAwsPrefixList_
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccDataSourceAwsPrefixList_ -timeout 120m
--- PASS: TestAccDataSourceAwsPrefixList_matchesTooMany (7.09s)
--- PASS: TestAccDataSourceAwsPrefixList_nameDoesNotOverrideFilter (7.10s)
--- PASS: TestAccDataSourceAwsPrefixList_basic (27.39s)
--- PASS: TestAccDataSourceAwsPrefixList_filter (28.37s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       29.752s
...
```

### DIFF
--- a/aws/data_source_aws_prefix_list.go
+++ b/aws/data_source_aws_prefix_list.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"sort"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -68,10 +69,8 @@ func dataSourceAwsPrefixListRead(d *schema.ResourceData, meta interface{}) error
 	d.SetId(*pl.PrefixListId)
 	d.Set("name", pl.PrefixListName)
 
-	cidrs := make([]string, len(pl.Cidrs))
-	for i, v := range pl.Cidrs {
-		cidrs[i] = *v
-	}
+	cidrs := aws.StringValueSlice(pl.Cidrs)
+	sort.Strings(cidrs)
 	d.Set("cidr_blocks", cidrs)
 
 	return nil
--- a/aws/data_source_aws_prefix_list.go
+++ b/aws/data_source_aws_prefix_list.go
@@ -71,7 +71,9 @@ func dataSourceAwsPrefixListRead(d *schema.ResourceData, meta interface{}) error
 
 	cidrs := aws.StringValueSlice(pl.Cidrs)
 	sort.Strings(cidrs)
-	d.Set("cidr_blocks", cidrs)
+	if err := d.Set("cidr_blocks", cidrs); err != nil {
+		return fmt.Errorf("failed to set cidr blocks of prefix list %s: %s", d.Id(), err)
+	}
 
 	return nil
 }
--- a/aws/data_source_aws_prefix_list.go
+++ b/aws/data_source_aws_prefix_list.go
@@ -45,19 +45,22 @@ func dataSourceAwsPrefixListRead(d *schema.ResourceData, meta interface{}) error
 	if prefixListID := d.Get("prefix_list_id"); prefixListID != "" {
 		req.PrefixListIds = aws.StringSlice([]string{prefixListID.(string)})
 	}
-	req.Filters = buildEC2AttributeFilterList(
-		map[string]string{
-			"prefix-list-name": d.Get("name").(string),
-		},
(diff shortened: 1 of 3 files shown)


aws/data_source_aws_prefix_list.go
```diff
@@ -45,31 +46,34 @@ func dataSourceAwsPrefixListRead(d *schema.ResourceData, meta interface{}) error
...
 	pl := resp.PrefixLists[0]
 
 	d.SetId(*pl.PrefixListId)
 	d.Set("name", pl.PrefixListName)
 
-	cidrs := make([]string, len(pl.Cidrs))
-	for i, v := range pl.Cidrs {
-		cidrs[i] = *v
+	cidrs := aws.StringValueSlice(pl.Cidrs)
+	sort.Strings(cidrs)
```
bflad, line 73: Similarly here with respect to introducing a breaking change which should be removed for now.

In this case though, if the ordering of the CIDR Blocks is not significant (which it likely is not), the `cidr_blocks` attribute should be changed to `TypeSet` instead of attempting to add ordering that is inconsistent with the semantics of the upstream API. Terraform configurations can and should use built in language support such as resource `for_each` for handling all elements of an attribute like this to avoid ordering concerns.

At some point in the future we may go back through many of the `TypeList` attributes and convert them to `TypeSet` to ensure that Terraform configurations are not depending on ordering where its not guaranteed or semantically correct.

aws/data_source_aws_prefix_list_test.go
```diff
@@ -98,3 +99,47 @@ data "aws_prefix_list" "s3_by_id" {
...
+				Config:      testAccDataSourceAwsPrefixListConfig_nameDoesNotOverrideFilter,
+				ExpectError: regexp.MustCompile(`no matching prefix list found`),
+			},
+		},
+	})
+}
+
+const testAccDataSourceAwsPrefixListConfig_nameDoesNotOverrideFilter = `
+data "aws_prefix_list" "test" {
+  name = "com.amazonaws.us-west-2.s3"
```
bflad, line 139: Some of our newer codebase linting will likely pick this up on rebase/merge for not being AWS Region agnostic. Similar to  the website example you added (👍 ) we should use the `aws_region` data source here so the configuration can be applied in anywhere. This also means we'll need to remove the hardcoded EC2 Prefix List identifier. Something like this should get you close:

```terraform

data "aws_prefix_list" "dynamodb" {
  name = "com.amazonaws.${data.aws_region.current.name}.dynamodb"
}

data "aws_prefix_list" "test" {
  name = "com.amazonaws.${data.aws_region.current.name}.s3"

  filter {
    name = "prefix-list-id"
    values = [data.aws_prefix_list.dynamodb.id]
  }
}
```

bflad (changes requested):
Hi @roberth-k 👋  Thank you for submitting this! Overall this is good stuff, but I marked this as breaking changes for now since we would want to hold off on purposefully introducing changes that effect practitioner experience until a major version update. We can split this up into fixing the name versus filter bug now and separately introducing the other changes. 

Please reach out if you have any questions. 😄

roberth-k:
@bflad The managed prefix list data source has been spun off as https://github.com/hashicorp/terraform-provider-aws/pull/16738, and the fixes for the original data source at https://github.com/hashicorp/terraform-provider-aws/pull/16739.

PR #14167, opened 2020-07-14
Fix schema set errors

appilon:
A few schema panics seemed to have slipped through the cracks

### Community Note

Relates OR Closes #0000

```release-note
* data-source/lb: `enable_http2` now properly set
* resource/lb_cookie_stickiness_policy: `lb_port` now properly set
* resource/spot_instance_request: `primary_network_interface_id` now properly set
* resource/ssm_association: `parameters` now properly set
* resource/elasticsearch_domain: update method to set advanced_security_options
```

```

...
```

### DIFF
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -709,22 +709,19 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 		return err
 	}
 
-	// Use AdvancedSecurityOptions from resource if possible
-	// because DescribeElasticsearchDomainConfig does not return MasterUserOptions
+	// Populate AdvancedSecurityOptions with values returned from
+	// DescribeElasticsearchDomainConfig, if enabled, else use
+	// values from resource; additionally, append MasterUserOptions
+	// from resource as they are not returned from the API
 	if ds.AdvancedSecurityOptions != nil {
-		if _, ok := d.GetOk("advanced_security_options"); ok {
-			if err := d.Set("advanced_security_options", []interface{}{
-				map[string]interface{}{
-					"enabled":                        ds.AdvancedSecurityOptions.Enabled,
-					"internal_user_database_enabled": ds.AdvancedSecurityOptions.InternalUserDatabaseEnabled,
-				},
-			}); err != nil {
-				return fmt.Errorf("error setting advanced_security_options: %w", err)
-			}
-		} else {
-			if err := d.Set("advanced_security_options", flattenAdvancedSecurityOptions(ds.AdvancedSecurityOptions)); err != nil {
-				return fmt.Errorf("error setting advanced_security_options: %w", err)
-			}
+		advSecOpts := flattenAdvancedSecurityOptions(ds.AdvancedSecurityOptions)
+		if !aws.BoolValue(ds.AdvancedSecurityOptions.Enabled) {
+			advSecOpts[0]["internal_user_database_enabled"] = getUserDBEnabled(d)
+		}
+		advSecOpts[0]["master_user_options"] = getMasterUserOptions(d)
+
+		if err := d.Set("advanced_security_options", advSecOpts); err != nil {
+			return fmt.Errorf("error setting advanced_security_options: %w", err)
 		}
 	}
 
--- a/aws/resource_aws_elasticsearch_domain.go
+++ b/aws/resource_aws_elasticsearch_domain.go
@@ -713,8 +713,14 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 	// because DescribeElasticsearchDomainConfig does not return MasterUserOptions
 	if ds.AdvancedSecurityOptions != nil {
 		if _, ok := d.GetOk("advanced_security_options"); ok {
-			d.Set("advanced_security_options.0.enabled", ds.AdvancedSecurityOptions.Enabled)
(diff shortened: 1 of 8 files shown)


bflad:
The `aws_lambda_invocation` data sources changes also require documentation updates, see also https://github.com/terraform-providers/terraform-provider-aws/pull/14125

aws/resource_aws_elasticsearch_domain.go
```diff
@@ -713,8 +713,14 @@ func resourceAwsElasticSearchDomainRead(d *schema.ResourceData, meta interface{}
 	// because DescribeElasticsearchDomainConfig does not return MasterUserOptions
 	if ds.AdvancedSecurityOptions != nil {
 		if _, ok := d.GetOk("advanced_security_options"); ok {
-			d.Set("advanced_security_options.0.enabled", ds.AdvancedSecurityOptions.Enabled)
-			d.Set("advanced_security_options.0.internal_user_database_enabled", ds.AdvancedSecurityOptions.InternalUserDatabaseEnabled)
+			if err := d.Set("advanced_security_options", []interface{}{
```
bflad, line 716: This seems to be unfortunately cause new test failures because `master_user_options` is missing, @anGie44 may know more:

```
--- FAIL: TestAccAWSElasticSearchDomain_AdvancedSecurityOptions_UserDB (953.01s)
testing.go:684: Step 0 error: After applying this step, the plan was not empty:
DIFF:
UPDATE: aws_elasticsearch_domain.test
advanced_options.rest.action.multi.allow_explicit_index:                "true" => "true"
advanced_security_options.#:                                            "1" => "1"
advanced_security_options.0.enabled:                                    "true" => "true"
advanced_security_options.0.internal_user_database_enabled:             "true" => "true"
advanced_security_options.0.master_user_options.#:                      "0" => "1"
advanced_security_options.0.master_user_options.0.master_user_name:     "" => "testmasteruser"
advanced_security_options.0.master_user_options.0.master_user_password: "" => "Barbarbarbar1!"

--- FAIL: TestAccAWSElasticSearchDomain_AdvancedSecurityOptions_Disabled (965.41s)
testing.go:684: Step 0 error: After applying this step, the plan was not empty:
DIFF:
UPDATE: aws_elasticsearch_domain.test
advanced_options.rest.action.multi.allow_explicit_index:                "true" => "true"
advanced_security_options.#:                                            "1" => "1"
advanced_security_options.0.enabled:                                    "false" => "false"
advanced_security_options.0.internal_user_database_enabled:             "false" => "true"
advanced_security_options.0.master_user_options.#:                      "0" => "1"
advanced_security_options.0.master_user_options.0.master_user_name:     "" => "testmasteruser"
advanced_security_options.0.master_user_options.0.master_user_password: "" => "Barbarbarbar1!"

--- FAIL: TestAccAWSElasticSearchDomain_AdvancedSecurityOptions_IAM (1216.21s)
testing.go:684: Step 0 error: After applying this step, the plan was not empty:
DIFF:
UPDATE: aws_elasticsearch_domain.test
advanced_options.rest.action.multi.allow_explicit_index:           "true" => "true"
advanced_security_options.#:                                       "1" => "1"
advanced_security_options.0.enabled:                               "true" => "true"
advanced_security_options.0.internal_user_database_enabled:        "false" => "false"
advanced_security_options.0.master_user_options.#:                 "0" => "1"
advanced_security_options.0.master_user_options.0.master_user_arn: "" => "arn:aws:iam::*******:user/es-master-user-9000075685154087677"
```

anGie44: ahhh yep the explicit d.Set() were added here to workaround the entire block being overwritten. we wanted to keep the nested `master_user_options` block unchanged while also setting the `enabled` and `internal_user_database_enbaled` attributes which also exist in this outer `advanced_security_options` block

anGie44: it seems a bit unconventional but maybe for this case since we need to rebuild this `advanced_security_options` map when we're setting it back in state, we can explicitly get the `master_user_options` from state (as they're not returned from the API) with a d.Get() and append these attributes to the overall array being constructed in lines 717-720?

bflad: @anGie44 yeah, that's the unfortunate way we've worked around these "missing nested configuration block in API response" situations in the past. Doesn't work on import for obvious reasons but there's only so much we can do. 😄

anGie44: @appilon @bflad https://github.com/terraform-providers/terraform-provider-aws/pull/14196 _should_ take care of these attributes

appilon: no worries @anGie44 thanks for fixing those attrs! I still have more panics to settle before this PR is finally ready

appilon: Let me know when your PR is done and I can review/merge it into this one. I think I got all of them now.

aws/structure.go
```diff
@@ -3262,6 +3262,20 @@ func expandAwsSsmTargets(in []interface{}) []*ssm.Target {
...
+func flattenAwsSsmParameters(parameters map[string][]*string) map[string]string {
+	result := make(map[string]string)
+	for p, values := range parameters {
+		var vs []string
+		for _, vPtr := range values {
+			if v := aws.StringValue(vPtr); v != "" {
+				vs = append(vs, v)
+			}
+		}
+		result[p] = strings.Join(vs, ",")
```
appilon, line 3274: @bflad unsure if this was the right way to do this.. the data structure from the API is `map[string][]string`, so in the scenario where it's multiple values for 1 key I string delimited the list??

PR #14175, opened 2020-07-14
Remove CodePipeline GitHub environment variable

### LINKED ISSUE #2796: Putting GITHUB_TOKEN in terraform config for aws_codepipeline

### LINKED ISSUE #4768: Codepipeline OAuthToken property inside source configuration should be sensitive
### Community Note

### Terraform Version

Terraform v0.11.7
+ provider.aws v1.22.0

### Affected Resource(s)

* aws_codepipeline

```hcl
resource "aws_codepipeline" "pipeline" {
  # ...
  stage {
    name = "Source"

    action {
      name             = "Source"
      category         = "Source"
      owner            = "ThirdParty"
      provider         = "GitHub"
      version          = "1"
      output_artifacts = ["code"]

      configuration {
        Owner                = "owner"
        Repo                 = "repo"
        Branch               = "master"
        PollForSourceChanges = false
        OAuthToken           = "<REDACTED>"
      }
    }
  }
}
```

### Expected Behavior

`stage.0.action.0.configuration.OAuthToken` should appear as sensitive in terraform plan and apply

### Actual Behavior

`stage.0.action.0.configuration.OAuthToken` value is shown in plan and apply.

### Steps to Reproduce

1. `terraform plan -out=plan`
2. `terraform apply plan`

### References

* #4261 : Related to this issue, but for codepipeline

gdavison:
Removes the previously required `GITHUB_TOKEN` environment variable for CodePipeline. The authentication token is now passed as the `OAuthToken` field in the action configuration.

### Community Note

Closes #2796 
Closes #4768

```release-note
resource/aws_codepipeline: Removes `GITHUB_TOKEN` environment variable
```

Output from acceptance testing in commercial partition:

```
$ make testacc TESTARGS='-run=TestAccAWSCodePipeline_'

--- PASS: TestAccAWSCodePipeline_disappears (35.93s)
--- PASS: TestAccAWSCodePipeline_emptyStageArtifacts (40.89s)
--- PASS: TestAccAWSCodePipeline_multiregion_basic (48.58s)
--- PASS: TestAccAWSCodePipeline_deployWithServiceRole (55.08s)
--- PASS: TestAccAWSCodePipeline_WithNamespace (58.69s)
--- PASS: TestAccAWSCodePipeline_basic (63.80s)
--- PASS: TestAccAWSCodePipeline_tags (75.56s)
--- PASS: TestAccAWSCodePipeline_multiregion_Update (79.32s)
--- PASS: TestAccAWSCodePipeline_multiregion_ConvertSingleRegion (94.63s)
```

Output from acceptance testing in GovCloud partition:

```
--- SKIP: TestAccAWSCodePipeline_multiregion_ConvertSingleRegion (2.37s)
--- SKIP: TestAccAWSCodePipeline_multiregion_basic (2.38s)
--- SKIP: TestAccAWSCodePipeline_multiregion_Update (2.39s)
--- PASS: TestAccAWSCodePipeline_disappears (24.64s)
--- PASS: TestAccAWSCodePipeline_emptyStageArtifacts (30.55s)
--- PASS: TestAccAWSCodePipeline_WithNamespace (32.72s)
--- PASS: TestAccAWSCodePipeline_deployWithServiceRole (35.85s)
--- PASS: TestAccAWSCodePipeline_basic (46.98s)
--- PASS: TestAccAWSCodePipeline_tags (57.72s)
```

### DIFF
--- a/aws/resource_aws_codepipeline.go
+++ b/aws/resource_aws_codepipeline.go
@@ -209,7 +209,7 @@ func resourceAwsCodePipelineCreate(d *schema.ResourceData, meta interface{}) err
 		resp, err = conn.CreatePipeline(params)
 	}
 	if err != nil {
-		return fmt.Errorf("Error creating CodePipeline: %s", err)
+		return fmt.Errorf("Error creating CodePipeline: %w", err)
 	}
 	if resp.Pipeline == nil {
 		return fmt.Errorf("Error creating CodePipeline: invalid response from AWS")
@@ -297,12 +297,12 @@ func flattenAwsCodePipelineArtifactStore(artifactStore *codepipeline.ArtifactSto
 	}
 
 	values := map[string]interface{}{}
-	values["type"] = *artifactStore.Type
-	values["location"] = *artifactStore.Location
+	values["type"] = aws.StringValue(artifactStore.Type)
+	values["location"] = aws.StringValue(artifactStore.Location)
 	if artifactStore.EncryptionKey != nil {
 		as := map[string]interface{}{
-			"id":   *artifactStore.EncryptionKey.Id,
-			"type": *artifactStore.EncryptionKey.Type,
+			"id":   aws.StringValue(artifactStore.EncryptionKey.Id),
+			"type": aws.StringValue(artifactStore.EncryptionKey.Type),
 		}
 		values["encryption_key"] = []interface{}{as}
 	}
@@ -339,7 +339,7 @@ func flattenAwsCodePipelineStages(stages []*codepipeline.StageDeclaration) []int
 	stagesList := []interface{}{}
 	for _, stage := range stages {
 		values := map[string]interface{}{}
-		values["name"] = *stage.Name
+		values["name"] = aws.StringValue(stage.Name)
 		values["action"] = flattenAwsCodePipelineStageActions(stage.Actions)
 		stagesList = append(stagesList, values)
 	}
@@ -410,16 +410,16 @@ func flattenAwsCodePipelineStageActions(actions []*codepipeline.ActionDeclaratio
 	actionsList := []interface{}{}
 	for _, action := range actions {
 		values := map[string]interface{}{
-			"category": *action.ActionTypeId.Category,
-			"owner":    *action.ActionTypeId.Owner,
-			"provider": *action.ActionTypeId.Provider,
-			"version":  *action.ActionTypeId.Version,
-			"name":     *action.Name,
+			"category": aws.StringValue(action.ActionTypeId.Category),
+			"owner":    aws.StringValue(action.ActionTypeId.Owner),
+			"provider": aws.StringValue(action.ActionTypeId.Provider),
+			"version":  aws.StringValue(action.ActionTypeId.Version),
+			"name":     aws.StringValue(action.Name),
 		}
 		if action.Configuration != nil {
 			config := flattenAwsCodePipelineStageActionConfiguration(action.Configuration)
 			_, ok := config["OAuthToken"]
-			actionProvider := *action.ActionTypeId.Provider
+			actionProvider := aws.StringValue(action.ActionTypeId.Provider)
 			if ok && actionProvider == "GitHub" {
 				delete(config, "OAuthToken")
 			}
@@ -435,15 +435,15 @@ func flattenAwsCodePipelineStageActions(actions []*codepipeline.ActionDeclaratio
 		}
 
 		if action.RoleArn != nil {
-			values["role_arn"] = *action.RoleArn
+			values["role_arn"] = aws.StringValue(action.RoleArn)
 		}
 
 		if action.RunOrder != nil {
-			values["run_order"] = int(*action.RunOrder)
+			values["run_order"] = int(aws.Int64Value(action.RunOrder))
 		}
 
 		if action.Region != nil {
-			values["region"] = *action.Region
+			values["region"] = aws.StringValue(action.Region)
 		}
 
 		if action.Namespace != nil {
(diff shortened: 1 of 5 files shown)


aws/resource_aws_codepipeline_test.go
```diff
@@ -488,10 +592,29 @@ func testAccPreCheckAWSCodePipeline(t *testing.T) {
 	}
 }
 
+func testAccPreCheckAWSCodePipelineAlternateRegion(t *testing.T) {
+	// There isn't a way to get the alternate region provider at PreCheck time, so hardcode it
+	if testAccGetAlternateRegion() == "us-gov-east-1" {
```
gdavison, line 597: Not sure, but this might be better as a check on `testAccGetAlternateRegionPartition() == "aws-us-gov"` instead…

aws/resource_aws_codepipeline.go
```diff
@@ -331,31 +340,23 @@ func expandAwsCodePipelineStages(d *schema.ResourceData) []*codepipeline.StageDe
...
-func expandAwsCodePipelineActions(s []interface{}) []*codepipeline.ActionDeclaration {
+func expandAwsCodePipelineActions(a []interface{}) []*codepipeline.ActionDeclaration {
 	actions := []*codepipeline.ActionDeclaration{}
-	for _, config := range s {
+	for _, config := range a {
 		data := config.(map[string]interface{})
 
 		conf := expandAwsCodePipelineStageActionConfiguration(data["configuration"].(map[string]interface{}))
-		if data["provider"].(string) == "GitHub" {
-			githubToken := os.Getenv("GITHUB_TOKEN")
```
bflad, line 353: Can you please add information to the Version 3 Upgrade Guide (`website/docs/guides/version-3-upgrade.html.md`) as part of this change?

aws/resource_aws_codepipeline.go
```diff
@@ -605,8 +610,33 @@ func resourceAwsCodePipelineDelete(d *schema.ResourceData, meta interface{}) err
...
+		hash := hashCodePipelineGitHubToken(new)
+		return old == hash
+	}
+
+	return false
+}
+
+const codePipelineGitHubTokenHashPrefix = "hash-"
+
+func hashCodePipelineGitHubToken(token string) string {
```
bflad, line 635: We are actively in the process of [removing the few other instances of hashed state storage](https://github.com/terraform-providers/terraform-provider-aws/issues/13406), however Terraform in general doesn't handle `TypeMap` and `Sensitive: true` in a meaningful way without making the entire map sensitive, if I remember correctly. 😖 

It would be a bummer to introduce this technical debt in, but otherwise we could also be forced down the road of what I think you were suggesting the other day of potentially going against the API design slightly and implementing bespoke Terraform schema so we can control it better (and in this situation, actually implement a single sensitive attribute). We do have some slight precedence of this class of user experience enhancement with resources like `aws_sns_platform_application` and I believe `aws_sns_topic`.

---

I think this leaves us in a tough spot. 🙁   We can try to quickly design something better for this situation to squeeze it in for this major release in the next two weeks, which is fairly risky in my opinion, or punt on this for now and spend later cycles on this issue in general (potentially changing the Core/SDK sensitive map abilities and/or ensuring we get good feedback on a new resource design). What do you think?

aws/resource_aws_codepipeline_test.go
```diff
@@ -488,10 +592,29 @@ func testAccPreCheckAWSCodePipeline(t *testing.T) {
 	}
 }
 
+func testAccPreCheckAWSCodePipelineAlternateRegion(t *testing.T) {
```
bflad, line 595: Ah its a bummer we cannot reuse logic similar to `testAccPreCheckAWSCodePipeline` here (using a real API call to automatically enable/skip the test), but I see the problem is that we don't initialize an "alternate" global instance of the provider in the testing similar to `testAccProvider`. I think this is okay as-is for now until we can figure out things as part of https://github.com/terraform-providers/terraform-provider-aws/issues/8983 / https://github.com/terraform-providers/terraform-provider-aws/issues/8316

aws/resource_aws_codepipeline_test.go
bflad: Wow this is annoying (on the AWS side), I'm fine with either, to be honest.

aws/resource_aws_codepipeline.go
gdavison: I think the environment variable is a huge pain-point for practitioners, and waiting until 4.0 to remove it would be less than ideal. If we don't want to hash the value, would setting the `TypeMap` to `Sensitive` be a better option at this point? And then we could come up with a better solution.

aws/resource_aws_codepipeline_test.go
gdavison: Yeah, I briefly considered the option of rolling my own connection to test the API, but I figured the hard-coded option was quicker. Though it actually might not be that complicated

aws/resource_aws_codepipeline.go
bflad: Our potential path forward will be:

* Removing the state hashing handling
* Marking the `configuration` attribute `Sensitive: true` and `Computed: true`
* Adding a new `github`/`github_configuration` configuration block so we can properly show a difference with only the `oauth` attribute sensitive

These will require apply-time validation to ensure they aren't both defined since `ConflictsWith` does not support relative references. The other configuration blocks will need to be filled in later due to time. The documentation and upgrade guide should recommend using the new configuration block(s) over the map.

bflad:
Once these pending items are commented/addressed, will provide another review:

* Fixing merge conflicts
* Adding a new `github`/`github_configuration` configuration block so we can properly show a difference with only the `oauth` attribute sensitive

PR #14177, opened 2020-07-14
docs/provider: Setup and document release/* branch convention, link 2.x and earlier changelog entries

bflad:
### Community Note

Reference: https://github.com/terraform-providers/terraform-provider-aws/tree/release/2.x

```release-note
NONE
```

Following Vault's branching convention.


CHANGELOG.md
```diff
@@ -29,4725 +29,8 @@ BUG FIXES
...
-* resource/aws_appmesh_virtual_node: Replace `backends` configuration block(s) with `backend` configuration blocks ([#7858](https://github.com/terraform-providers/terraform-provider-aws/issues/7858))
-* resource/aws_appmesh_virtual_node: Replace `service_discovery` configuration block `service_name` argument with `service_discovery` configuration block `dns` configuration block ([#7858](https://github.com/terraform-providers/terraform-provider-aws/issues/7858))
-* resource/aws_appmesh_virtual_router: Remove `spec` configuration block `service_names` argument ([#7858](https://github.com/terraform-providers/terraform-provider-aws/issues/7858))
-
-FEATURES:
-
-* **New
```
breathingdust, line 36: Do you think its worth copying AzureRM and having a CHANGELOGv2.md? It would be easier to search for a particular version a resource was released in that way?

bflad: Personally, I'd prefer to not need to growingly managing additional root files across multiple branches. e.g. if we patch something in 2.x, we'll need to update `master` and `release/2.x` versions of that file whereas linking directly to it means a single update on its tracking branch. If we are worried about exposing when functionality was released whether a new resource/data source or new attribute, we could possibly teach the registry how to do that by exporting the provider schema with each version and ingesting that information for additional display there.

breathingdust: I didn't consider that, makes sense. 👍

PR #14224, opened 2020-07-17
Add wait_until_stable option for ECS services

surminus:
I initially considered raising this as an issue, but then decided I'd be better placed to show an example through a PR.

There are quite a few elements of an ECS service that cause the resource to be recreated. Upon creation, we do not wait for the service to be stable, which means even using lifecycle rules to `create_before_destroy`, we will suffer downtime.

This commit seeks to introduce a basic assessment on whether a service is stable. The attributes for this assessment are:

- if the desired count matches the running count
- if the pending count is zero
- if the service is described as "ACTIVE"

If all 3 of these conditions are met, we can be reasonably sure that the service has at least started and provisioned some tasks. It won't account for tasks that end up crashing, but if you're confident of your service it should be safe to use.

I haven't added any tests yet, because I'm honestly not quite sure the best way to add them. I would appreciate some guidance, if possible! I'm OK to run an acceptance test, so more than happy to poke at finding the right solution.

### Community Note

Relates OR Closes #0000

```release-note

```

```

...
```

### DIFF
--- a/aws/resource_aws_ecs_service.go
+++ b/aws/resource_aws_ecs_service.go
@@ -353,6 +353,11 @@ func resourceAwsEcsService() *schema.Resource {
 				},
 			},
 			"tags": tagsSchema(),
+
+			"wait_until_stable": {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


surminus:
This PR achieves the same so going to close: https://github.com/hashicorp/terraform-provider-aws/pull/3485

Issue #12633, opened 2020-04-02
Api gateway authorizer `authorizer_result_ttl_in_seconds` is not set to 0 on first run


kevinkuszyk:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
Terraform v0.12.24
+ provider.aws v2.55.0
```

### Affected Resource(s)

* `aws_api_gateway_authorizer.api_authorizer`

### Terraform Configuration Files

```
resource "aws_api_gateway_authorizer" "api_authorizer" {
  name                             = "authorizer"
  rest_api_id                      = aws_api_gateway_rest_api.api.id
  authorizer_uri                   = aws_lambda_function.authorizer.invoke_arn
  authorizer_credentials           = aws_iam_role.api_authorizer.arn
  type                             = "REQUEST"
  authorizer_result_ttl_in_seconds = 0
}
```

There is a repo here with the full source:
https://github.com/kevinkuszyk/terraform-aws-apigateway-authorizer-cache

### Expected Behavior

I was expecting the authorizer to be created with the `authorizer_result_ttl_in_seconds` set to zero.

### Actual Behavior

It doesn't seem to be set.  It shows like this in the aws console:

<img width="473" alt="Screenshot 2020-04-02 at 15 56 40" src="https://user-images.githubusercontent.com/2734580/78264056-9d6c7c00-74fa-11ea-8ee7-6a0001df3b85.png">

And a `terraform plan` after applying the config gives this (`300` is the default value):

```
  # aws_api_gateway_authorizer.api_authorizer will be updated in-place
  ~ resource "aws_api_gateway_authorizer" "api_authorizer" {
        authorizer_credentials           = "arn:aws:iam::************:role/api-authorizer"
      ~ authorizer_result_ttl_in_seconds = 300 -> 0
        authorizer_uri                   = "arn:aws:apigateway:eu-west-2:lambda:path/2015-03-31/functions/arn:aws:lambda:eu-west-2:************::function:authorizer/invocations"
        id                               = "7q1ebw"
        identity_source                  = "method.request.header.Authorization"
[...]

PR #14262, opened 2020-07-21
resource/aws_sns_topic_subscription: Use paginated ListSubscriptionsByTopic and return immediately on errors

bflad:
### Community Note

Reference: https://github.com/terraform-providers/terraform-provider-aws/issues/13409

```release-note
NONE
```

Also fixes recurring, unrelated test failure:

```
--- FAIL: TestAccAWSSNSTopicSubscription_autoConfirmingSecuredEndpoint (63.28s)
testing.go:684: Step 0 error: After applying this step, the plan was not empty:
DIFF:
UPDATE: aws_api_gateway_authorizer.test
...
authorizer_result_ttl_in_seconds: "300" => "0"
```

```
--- PASS: TestAccAWSSNSTopicSubscription_basic (13.47s)
--- PASS: TestAccAWSSNSTopicSubscription_rawMessageDelivery (27.13s)
--- PASS: TestAccAWSSNSTopicSubscription_filterPolicy (28.12s)
--- PASS: TestAccAWSSNSTopicSubscription_deliveryPolicy (28.38s)
--- PASS: TestAccAWSSNSTopicSubscription_autoConfirmingEndpoint (48.31s)
--- PASS: TestAccAWSSNSTopicSubscription_autoConfirmingSecuredEndpoint (91.18s)
```

### DIFF
--- a/aws/resource_aws_sns_topic_subscription_test.go
+++ b/aws/resource_aws_sns_topic_subscription_test.go
@@ -749,7 +749,6 @@ resource "aws_api_gateway_authorizer" "test" {
   name                             = "tf-acc-test-api-gw-authorizer-%d"
   rest_api_id                      = "${aws_api_gateway_rest_api.test.id}"
   authorizer_uri                   = "${aws_lambda_function.authorizer.invoke_arn}"
-  authorizer_result_ttl_in_seconds = "0"
   authorizer_credentials           = "${aws_iam_role.invocation_role.arn}"
 }
 
--- a/aws/resource_aws_sns_topic_subscription.go
+++ b/aws/resource_aws_sns_topic_subscription.go
@@ -235,18 +235,16 @@ func subscribeToSNSTopic(d *schema.ResourceData, snsconn *sns.SNS) (output *sns.
 
 			subscription, err := findSubscriptionByNonID(d, snsconn)
 
-			if subscription != nil {
-				output.SubscriptionArn = subscription.SubscriptionArn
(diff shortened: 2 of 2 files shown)


aws/resource_aws_sns_topic_subscription_test.go
```diff
@@ -749,7 +749,6 @@ resource "aws_api_gateway_authorizer" "test" {
   name                             = "tf-acc-test-api-gw-authorizer-%d"
   rest_api_id                      = "${aws_api_gateway_rest_api.test.id}"
   authorizer_uri                   = "${aws_lambda_function.authorizer.invoke_arn}"
-  authorizer_result_ttl_in_seconds = "0"
```
anGie44, line 752: this fix looks good! I also noticed in this test step https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_api_gateway_authorizer_test.go#L190 we're able to update the TTL to 0 without a problem so it looks like the error here is attributed to the `d.GetOk` call instead of `d.Get` when setting this value at creation time https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_api_gateway_authorizer.go#L111 . might be out of scope to address it in this PR but I can make in issue/see if there already is one stating we can't create this authorizer resource w/ a value of 0 (caching disabled)

anGie44: Found: #12633

PR #14334, opened 2020-07-24
feat: Enable S3-compatible providers with no support for bucket tagging

killzoner:
### Community Note

```release-note
* resource/aws_s3_bucket: Add `skip_get_s3_bucket_list_tags` to enable S3-compatible providers with no support for bucket tagging (ie https://wasabi.com)
```

```

...
```

### DIFF
--- a/aws/resource_aws_s3_bucket.go
+++ b/aws/resource_aws_s3_bucket.go
@@ -1319,7 +1319,11 @@ func resourceAwsS3BucketRead(d *schema.ResourceData, meta interface{}) error {
 
 	// Retry due to S3 eventual consistency
 	tags, err := retryOnAwsCode(s3.ErrCodeNoSuchBucket, func() (interface{}, error) {
-		return keyvaluetags.S3BucketListTags(s3conn, d.Id())
+		tags, err := keyvaluetags.S3BucketListTags(s3conn, d.Id())
+		if err != nil && (!isAWSErr(err, "NotImplemented", "") || !isAWSErr(err, "NoSuchTagSetError", "") || isAWSErrRequestFailureStatusCode(err, 404)) {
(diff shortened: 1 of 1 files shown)


killzoner:
Hi there, this is a WIP work, the base idea is here though.

The motivation behind this is to be able to work with some alternative S3-compatible providers which don't support bucket tagging. I checked and `S3BucketListTags` seems to be used only in `aws/resource_aws_s3_bucket.go` so it would not be a big change.

Currently looking into the contribution guides to add tests, even though it's not much change.
Still need to figure out to how to test it locally too.

One thing that maybe somebody can tell me is if it would be preferable to add default empty tag `map` on `d *schema.ResourceData` or if it's ok as is.

bflad:
Hi @killzoner 👋  Thank you for submitting this! Our preference (also why I marked this with the `proposal` label) when implementing best effort support with these alternative implementations is typically to ignore their specific errors with specific API calls, instead of requiring operators to add extra/manual configuration. For example (the `NotImplemented` part):

https://github.com/terraform-providers/terraform-provider-aws/blob/818942acfd59940af2295587cca2b81a6aa4e895/aws/resource_aws_s3_bucket.go#L947

To get the correct error code/message, you will need to have the error handy. Let us know if you have questions. 👍

killzoner:
Hi @bflad thanks for the input! Doing that soon then

killzoner:
Hi @bflad I changed the modifications according to your review, it's working :+1: 
One thing I can't seem to understad is when I do an `plan` after creating a bucket, I get always the same change planned like this

```
- object_lock_configuration {
}

- replication_configuration {
}

- server_side_encryption_configuration {
}
```
even though I didn't define any of this.

Any idea or working this around? I'll look after other possibles fixes if not

PR #14381, opened 2020-07-29
Move semaphore notify to test cleanup

appilon:
In general it is probably not a safe bet to have critical code in a `CheckDestroy` in V1 (as it does not run when state is nil), in SDKv2, we explicitly don't call Destroy if state is empty. As a result the semaphore leaks and deadlocks against SDKv2, because creation fails due to resource limits (I guess in V1, `state != nil` so this problem doesn't happen 🤷‍♂️ ), the semaphore is not released. This change uses go1.14's Cleanup helper to ensure the semaphore is released.

### Community Note

Relates OR Closes #0000

```release-note

```

The tests ran without deadlock against SDKv1 or SDKv2

### DIFF
--- a/aws/resource_aws_ec2_client_vpn_endpoint_test.go
+++ b/aws/resource_aws_ec2_client_vpn_endpoint_test.go
@@ -111,6 +111,9 @@ func TestAccAwsEc2ClientVpn_serial(t *testing.T) {
 		for name, tc := range m {
 			tc := tc
 			t.Run(fmt.Sprintf("%s_%s", group, name), func(t *testing.T) {
+				t.Cleanup(func() {
+					testAccEc2ClientVpnEndpointSemaphore.Notify()
+				})
 				tc(t)
 			})
 		}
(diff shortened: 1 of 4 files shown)


aws/resource_aws_ec2_client_vpn_endpoint_test.go
```diff
@@ -111,6 +111,9 @@ func TestAccAwsEc2ClientVpn_serial(t *testing.T) {
 		for name, tc := range m {
 			tc := tc
 			t.Run(fmt.Sprintf("%s_%s", group, name), func(t *testing.T) {
+				t.Cleanup(func() {
+					testAccEc2ClientVpnEndpointSemaphore.Notify()
+				})
```
bflad, line 116: This is currently deadlocking during normal `go test` (non acceptance testing) since the underlying function is waiting to close a channel that never gets opened. As a first line of defense we can guard this for acceptance testing:

```suggestion
				t.Cleanup(func() {
					if os.Getenv(resource.TestEnvVar) != "" {
						testAccEc2ClientVpnEndpointSemaphore.Notify()
					}
				})
```

But I'm pretty sure this can still fail if `testAccPreCheck(t)` fails in the current test cases. We can try switching the `PreCheck` handlers in the test cases to workaround that particular issue as well (annoying as that is logic-wise):

```go
PreCheck:     func() { testAccPreCheckClientVPNSyncronize(t); testAccPreCheck(t) },
```

appilon:
fixed @bflad

PR #14451, opened 2020-08-03
aws_workspaces_directory: Assign IP Group

### LINKED ISSUE #14669: Add aws_workspaces_ip_group association with directory
### Community Note

### Description

This PR #10904  added the aws_workspaces_ip_group resource but it is useless since it cant' be applied to a directory: neither the directory resources like aws_workspaces_directory or aws_directory_service_directory has ip group argument nor there is such an association resource like aws_workspaces_ip_group_association.

* aws_workspaces_ip_group

### References

*[ IP Access Control Groups for Your WorkSpaces](https://docs.aws.amazon.com/workspaces/latest/adminguide/amazon-workspaces-ip-access-control-groups.html)

Tensho:
### Community Note

Relates  #434
Closes #14669

```release-note
aws_workspaces_directory: Assign IP group
```

### Acceptance Test

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAwsWorkspacesDirectory_ipGroupIds'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAwsWorkspacesDirectory_ipGroupIds -timeout 120m
=== RUN   TestAccAwsWorkspacesDirectory_ipGroupIds
=== PAUSE TestAccAwsWorkspacesDirectory_ipGroupIds
=== CONT  TestAccAwsWorkspacesDirectory_ipGroupIds
--- PASS: TestAccAwsWorkspacesDirectory_ipGroupIds (903.92s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	906.724s
```

### DIFF
--- a/aws/resource_aws_workspaces_directory.go
+++ b/aws/resource_aws_workspaces_directory.go
@@ -55,6 +55,7 @@ func resourceAwsWorkspacesDirectory() *schema.Resource {
 				Type:     schema.TypeSet,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 				Computed: true,
+				Optional: true,
 			},
 			"registration_code": {
 				Type:     schema.TypeString,
@@ -200,6 +201,19 @@ func resourceAwsWorkspacesDirectoryCreate(d *schema.ResourceData, meta interface
 		log.Printf("[DEBUG] WorkSpaces Directory %q creation properties are set", directoryId)
 	}
 
+	if v, ok := d.GetOk("ip_group_ids"); ok && v.(*schema.Set).Len() > 0 {
+		ipGroupIds := v.(*schema.Set)
+		log.Printf("[DEBUG] Associate WorkSpaces Directory %q with IP Groups %s...", directoryId, ipGroupIds.List())
+		_, err := conn.AssociateIpGroups(&workspaces.AssociateIpGroupsInput{
+			DirectoryId: aws.String(directoryId),
+			GroupIds:    expandStringSet(ipGroupIds),
+		})
+		if err != nil {
+			return fmt.Errorf("error asassociating ip groups: %s", err)
+		}
+		log.Printf("[DEBUG] WorkSpaces Directory %q IP Groups are set", directoryId)
+	}
+
 	return resourceAwsWorkspacesDirectoryRead(d, meta)
 }
 
@@ -286,6 +300,34 @@ func resourceAwsWorkspacesDirectoryUpdate(d *schema.ResourceData, meta interface
 			return fmt.Errorf("error updating creation properties: %s", err)
 		}
 		log.Printf("[DEBUG] WorkSpaces Directory %q creation properties are set", d.Id())
+	}
+
+	if d.HasChange("ip_group_ids") {
+		o, n := d.GetChange("ip_group_ids")
+		old := o.(*schema.Set)
+		new := n.(*schema.Set)
+		added := new.Difference(old)
+		removed := old.Difference(new)
+
+		log.Printf("[DEBUG] Associate IP Groups %s from WorkSpaces Directory %q...", added.GoString(), d.Id())
+		_, err := conn.AssociateIpGroups(&workspaces.AssociateIpGroupsInput{
+			DirectoryId: aws.String(d.Id()),
+			GroupIds:    expandStringSet(added),
+		})
+		if err != nil {
+			return fmt.Errorf("error asassociating ip groups: %s", err)
+		}
+
+		log.Printf("[DEBUG] Disassociate IP Groups %s from WorkSpaces Directory %q...", removed.GoString(), d.Id())
+		_, err = conn.DisassociateIpGroups(&workspaces.DisassociateIpGroupsInput{
+			DirectoryId: aws.String(d.Id()),
+			GroupIds:    expandStringSet(removed),
+		})
+		if err != nil {
+			return fmt.Errorf("error disasassociating ip groups: %s", err)
+		}
+
+		log.Printf("[DEBUG] WorkSpaces Directory %q IP Groups are set", d.Id())
 	}
 
 	if d.HasChange("tags") {
--- a/aws/resource_aws_workspaces_directory_test.go
+++ b/aws/resource_aws_workspaces_directory_test.go
@@ -283,6 +283,45 @@ func TestAccAwsWorkspacesDirectory_workspaceCreationProperties(t *testing.T) {
 	})
 }
 
+func TestAccAwsWorkspacesDirectory_ipGroupIds(t *testing.T) {
+	var v workspaces.WorkspaceDirectory
+	rName := acctest.RandString(8)
+
+	resourceName := "aws_workspaces_directory.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t); testAccPreCheckHasIAMRole(t, "workspaces_DefaultRole") },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsWorkspacesDirectoryDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccWorkspacesDirectoryConfig_ipGroupIds_create(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsWorkspacesDirectoryExists(resourceName, &v),
+					resource.TestCheckResourceAttr(resourceName, "ip_group_ids.#", "1"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			{
+				Config: testAccWorkspacesDirectoryConfig_ipGroupIds_update(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsWorkspacesDirectoryExists(resourceName, &v),
+					resource.TestCheckResourceAttr(resourceName, "ip_group_ids.#", "2"),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
 func testAccPreCheckHasIAMRole(t *testing.T, roleName string) {
 	conn := testAccProvider.Meta().(*AWSClient).iamconn
 
@@ -615,6 +654,47 @@ resource "aws_workspaces_directory" "main" {
     enable_maintenance_mode             = false
     user_enabled_as_local_administrator = false
   }
+}
+`, rName))
+}
+
+func testAccWorkspacesDirectoryConfig_ipGroupIds_create(rName string) string {
+	return composeConfig(
+		testAccAwsWorkspacesDirectoryConfig_Prerequisites(rName),
+		fmt.Sprintf(`
+resource "aws_workspaces_ip_group" "test_alpha" {
+  name = "%[1]s-alpha"
+}
+
+resource "aws_workspaces_directory" "test" {
+  directory_id = aws_directory_service_directory.main.id
+
+  ip_group_ids = [
+    aws_workspaces_ip_group.test_alpha.id
+  ]
+}
+`, rName))
+}
+
+func testAccWorkspacesDirectoryConfig_ipGroupIds_update(rName string) string {
+	return composeConfig(
+		testAccAwsWorkspacesDirectoryConfig_Prerequisites(rName),
+		fmt.Sprintf(`
+resource "aws_workspaces_ip_group" "test_beta" {
+  name = "%[1]s-beta"
+}
+
+resource "aws_workspaces_ip_group" "test_gamma" {
+  name = "%[1]s-gamma"
+}
+
+resource "aws_workspaces_directory" "test" {
+  directory_id = aws_directory_service_directory.main.id
+
+  ip_group_ids = [
+    aws_workspaces_ip_group.test_beta.id,
+    aws_workspaces_ip_group.test_gamma.id
+  ]
 }
 `, rName))
 }
--- a/aws/resource_aws_workspaces_ip_group.go
+++ b/aws/resource_aws_workspaces_ip_group.go
@@ -148,13 +148,49 @@ func resourceAwsWorkspacesIpGroupUpdate(d *schema.ResourceData, meta interface{}
 func resourceAwsWorkspacesIpGroupDelete(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*AWSClient).workspacesconn
 
-	log.Printf("[INFO] Deleting Workspaces IP Group")
-	_, err := conn.DeleteIpGroup(&workspaces.DeleteIpGroupInput{
+	var directoryID *string
+
+	log.Printf("[DEBUG] Finding directory associated with Workspaces IP Group %q", d.Id())
+	err := conn.DescribeWorkspaceDirectoriesPages(nil,
+		func(page *workspaces.DescribeWorkspaceDirectoriesOutput, lastPage bool) bool {
+			for _, dir := range page.Directories {
+				for _, ipg := range dir.IpGroupIds {
+					if aws.StringValue(ipg) == d.Id() {
+						directoryID = dir.DirectoryId
+						return true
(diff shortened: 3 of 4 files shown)


aws/resource_aws_workspaces_directory.go
```diff
@@ -225,6 +238,21 @@ func resourceAwsWorkspacesDirectoryUpdate(d *schema.ResourceData, meta interface
 		log.Printf("[DEBUG] WorkSpaces Directory %q self-service permissions are set", d.Id())
 	}
 
+	if d.HasChange("ip_group_ids") {
+		ipGroupIds := d.Get("ip_group_ids").(*schema.Set).List()
+		log.Printf("[DEBUG] Associate WorkSpaces Directory %q with IP Groups %s...", d.Id(), ipGroupIds)
+		// Need to calculate diff
```
Tensho, line 244: https://discuss.hashicorp.com/t/slice-of-strings-diff/12324

aws/resource_aws_workspaces_directory_test.go
```diff
@@ -216,6 +216,47 @@ func TestAccAwsWorkspacesDirectory_tags(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t); testAccPreCheckHasIAMRole(t, "workspaces_DefaultRole") },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAwsWorkspacesDirectoryDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccWorkspacesDirectoryConfig_ipGroupIds_create(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAwsWorkspacesDirectoryExists(resourceName, &v),
+					resource.TestCheckResourceAttr(resourceName, "ip_group_ids.#", "1"),
+					// CONSIDER: For some reason returns error: "Attribute "ip_group_ids.0" not set"
```
Tensho, line 237: Have no idea why test check fails here 🤷‍♂️

aws/resource_aws_workspaces_directory.go
```diff
@@ -144,17 +144,30 @@ func resourceAwsWorkspacesDirectoryCreate(d *schema.ResourceData, meta interface
...
 		}
+		log.Printf("[DEBUG] WorkSpaces Directory %q self-service permissions are set", directoryId)
+	}
+
+	if v, ok := d.GetOk("ip_group_ids"); ok {
+		ipGroupIds := v.(*schema.Set).List()
+		log.Printf("[DEBUG] Associate WorkSpaces Directory %q with IP Groups %s...", directoryId, ipGroupIds)
+		_, err := conn.AssociateIpGroups(&workspaces.AssociateIpGroupsInput{
+			DirectoryId: aws.String(directoryId),
+			GroupIds:    expandStringList(ipGroupIds),
```
DrFaust92, line 164: expandStringSet and there wont be a need to call .List()

aws/resource_aws_workspaces_directory.go
```diff
@@ -144,17 +144,30 @@ func resourceAwsWorkspacesDirectoryCreate(d *schema.ResourceData, meta interface
...
 			ResourceId:             aws.String(directoryId),
 			SelfservicePermissions: expandSelfServicePermissions(v.([]interface{})),
 		})
 		if err != nil {
 			return fmt.Errorf("error setting self service permissions: %s", err)
 		}
+		log.Printf("[DEBUG] WorkSpaces Directory %q self-service permissions are set", directoryId)
+	}
+
+	if v, ok := d.GetOk("ip_group_ids"); ok {
```
DrFaust92, line 159: add `ok && v.(*schema.Set).Len() > 0`

aws/resource_aws_workspaces_directory.go
```diff
@@ -225,6 +238,21 @@ func resourceAwsWorkspacesDirectoryUpdate(d *schema.ResourceData, meta interface
 		log.Printf("[DEBUG] WorkSpaces Directory %q self-service permissions are set", d.Id())
 	}
 
+	if d.HasChange("ip_group_ids") {
+		ipGroupIds := d.Get("ip_group_ids").(*schema.Set).List()
```
DrFaust92, line 242: same as above, schema struct has diff function is remove of old groups is needed by calling d.GetChange

DrFaust92:
```
		o, n := d.GetChange("endpoint_configuration.0.vpc_endpoint_ids")
		prefix := "/endpointConfiguration/vpcEndpointIds"

		old := o.(*schema.Set).List()
		new := n.(*schema.Set).List()
```

aws/resource_aws_workspaces_directory_test.go
DrFaust92: it fails because because `ip_group_ids` is a set and not an ordered list. the "index" will a hash value instead. you can leave as just the count test and the import will test the rest.

aws/resource_aws_workspaces_directory.go
Tensho: Awesome 🙂

aws/resource_aws_workspaces_directory.go
Tensho: I will

aws/resource_aws_workspaces_directory.go
Tensho: I have been looking for something like this, I'll add IP group disassociation with it.

Tensho:
@DrFaust92 Thank you for review, I'll make appropriate changes shortly 🙇

aws/resource_aws_workspaces_directory_test.go
Tensho: Ah, I see. Does it make sense to switch from `Set` to `List` for this attribute? I remember @gdavison suggested to use `TypeList + MaxItem 1` over `TypeSet + MaxItem 1`, but what is the best for many items?

Tensho: How does `import` know what should be tested? Does `ImportStateVerify` compare resource configuration attributes and actual state attributes?

DrFaust92: yep, let say that if would have omitted the read part for the attribute the import test would have showed the diff.

aws/resource_aws_workspaces_directory_test.go
```diff
@@ -516,3 +555,37 @@ resource "aws_workspaces_directory" "main" {
...
+resource "aws_workspaces_directory" "main" {
+  directory_id = "${aws_directory_service_directory.main.id}"
+  ip_group_ids = ["${aws_workspaces_ip_group.test_alpha.id}"]
+}
+`, rName)
+}
+
+func testAccWorkspacesDirectoryConfig_ipGroupIds_update(rName string) string {
+	return testAccAwsWorkspacesDirectoryConfig_Prerequisites(rName) + fmt.Sprintf(`
+resource "aws_workspaces_ip_group" "test_alpha" {
```
Tensho, line 574: @DrFaust92 I caught one issue with workspace directory and IP group mutual dependency. If I remove IP group resource and it's reference in workspaces directory declaration, then I get an error like this:
```
Error: Error Deleting Workspaces IP Group: ResourceAssociatedException: Can not delete IP Group with GroupId - wsipg-5g88k3vby because its attached with these Directories d-9267047639
```
If I understand correctly that happens, because workspace directory depends on IP group implicitly referencing it in `ip_group_ids`, so Terraform deletes dependent resource (IP group) first. But IP group can't be deleted until it is associated with any directory. Sounds like I should declare workspace directory dependency for IP group, but that makes a cycle and violates acyclic graph constraint:
```
resource "aws_workspaces_ip_group" "test_alpha" {
  name        = "%[1]s-alpha"

  depends_on = ["${aws_workspaces_directory.main.id}"]
}

resource "aws_workspaces_directory" "main" {
  directory_id = "${aws_directory_service_directory.main.id}"
  ip_group_ids = ["${aws_workspaces_ip_group.alpha.id}"]
}
```
I have only one idea – disassociate IP group from directory, but it requires to call [DescribeWorkspaceDirectories](https://docs.aws.amazon.com/workspaces/latest/api/API_DescribeWorkspaceDirectories.html) API action and find the directory associated with IP group – [DescribeIpGroups](https://docs.aws.amazon.com/workspaces/latest/api/API_DescribeIpGroups.html) API action doesn't return back refrence to directory. Feels not ideomatic. Could you provide an example for already implenented resources with similar behavior?

aws/resource_aws_workspaces_directory_test.go
Tensho: Thank you 🙇‍♂️

Tensho:
@DrFaust92 Fixed review comments, please take a look now 🙇‍♂️

aws/resource_aws_workspaces_directory_test.go
DrFaust92: There is something similar in API Gateway Usage plan resource, care to take a look for inspiration?
im guessing there isn't much to do to beside listing all directories and dissociating all directories from that ip group.

maybe it's worth splitting the association itself to a separate resource. and leave the ip_groups in this resource computed.

wdyt?

Tensho: TBH, I haven't found anything could help in `aws_api_gateway_usage_plan` resource. `UpdateUsagePlan` action handles all change together with `PatchOperations` attribute, which is designed to handle update operations in ordered manner.

---

I doubt `aws_workspaces_ip_group_attachment` makes any difference and solves the cycle issue.
```
resource "aws_workspaces_directory" "main" {
  directory_id = "${aws_directory_service_directory.main.id}"
}

resource "aws_workspaces_ip_group" "test_alpha" {
  name        = "%[1]s-alpha"
}

resource "aws_workspaces_ip_group_attachment" "main" {
  directory_id = "${aws_directory_service_directory.main.id}"
  ip_group_ids = ["${aws_workspaces_ip_group.test_alpha.id}"]
}
```

WorkspacesIpGroups

As far as attachment refers to directory and ip group resources, they are destroyed first by Terraform. Nor directory, neither ip group can be destroyed until disassociation is completed.

--- 

Seems like we should go with dumb `aws_workspaces_ip_group` resource disassociation on `Delete()`.

@DrFaust92 Thank you for the insight and please don't hesitate to drop any comment regarding my plan 🙇  I'm not a seasoned Terraform provider developer yet, so appreciate any feedback 🙂

DrFaust92: Take a look a the delete function in usage plan. It deletes the associated stages.

I think your plan seems fine to me, @gdavison want if you want to chime in on this.

Tensho: I've just bumped with another idea – reverse (inject) dependency between resources – drop `aws_workspaces_directory.ip_group_ids` attribute in favor of `aws_workspaces_ip_group.directory_id` one.

```
resource "aws_workspaces_directory" "test" {
  directory_id = "${aws_directory_service_directory.test.id}"
}

resource "aws_workspaces_ip_group" "a" {
  name         = "A"
  directory_id = "${aws_workspaces_directory.test.id}" {
}

resource "aws_workspaces_ip_group" "b" {
  name         = "B"
  directory_id = "${aws_workspaces_directory.test.id}" {
}
```

WorkspacesIpGroupsReverse

I know AWS Terraform provider tries to follow AWS API as close as possible, but this option looks more explicit than implicit group disassociation.

@DrFaust92  @bflad @gdavison What do you think?

DrFaust92: Seems like a nicer solution but doesn't it mean there needs to be an IP group per directory? Isn't it supposed to be that an IP group can be attached to several directories? (for less duplication)

Tensho: Good point, indeed IP group is supposed to be associated with many directories. Let's allow multiple values within `directory_ids` attribute:
```
resource "aws_workspaces_directory" "x" {
  directory_id = "${aws_directory_service_directory.x.id}"
}

resource "aws_workspaces_directory" "y" {
  directory_id = "${aws_directory_service_directory.y.id}"
}

resource "aws_workspaces_ip_group" "a" {
  directory_ids = [
    "${aws_workspaces_directory.x.id}",
  ]
}

resource "aws_workspaces_ip_group" "b" {
  directory_ids = [
    "${aws_workspaces_directory.x.id}",
    "${aws_workspaces_directory.y.id}",
  ]
}
```
Doesn't look cool, but solves the problem. @DrFaust92 Is it better from your perspective?

DrFaust92: It looks better, I'm just wondering if it becomes too unintuitive for users. Let's wait for maintainers to comment as well.

Tensho: @gdavison Please could you suggest the best option here?

bensunny1: Thanks for working on this @Tensho. I personally prefer the attachment resource option, this would be more intuitive and most similar to other resources

Tensho: @bensunny1 TBH, I can't grasp how attachment may solve the [problem I mentioned above in this thread](https://github.com/terraform-providers/terraform-provider-aws/pull/14451/files#r466892431). If I remove IP group attachment and IP group from the configuration **all together**, then both resources will be marked as orphaned and deleted in undesired order (first IP group, then IP group attachment) by Terraform. I guess it's OK for other attachments and target resources, e.g. `aws_iam_role_policy_attachment` and `aws_iam_role`/`aws_iam_policy`, but it doesn't work for IP group API. Or maybe I don't understand how Terraform builds the graph nodes for `destroy` operations...

@apparentlymart @bflad Please help us to sort out this chicken-egg question 🙇

bensunny1: @Tensho You're right, https://github.com/terraform-providers/terraform-provider-aws/pull/14451#discussion_r466961078 i think is a pretty good idea and deals with the problem

Tensho:
```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAwsWorkspacesDirectory_ipGroupIds'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAwsWorkspacesDirectory_ipGroupIds -timeout 120m
=== RUN   TestAccAwsWorkspacesDirectory_ipGroupIds
=== PAUSE TestAccAwsWorkspacesDirectory_ipGroupIds
=== CONT  TestAccAwsWorkspacesDirectory_ipGroupIds
--- PASS: TestAccAwsWorkspacesDirectory_ipGroupIds (64.30s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	66.881s
```

aws/resource_aws_workspaces_directory_test.go
gdavison: I think the best option here is to add the `DisassociateIpGroups()` call to the `aws_workspaces_ip_group` delete operation as well. There isn't currently a way to model these mutual relationships in Terraform, so we're stuck with brute-forcing it.

Tensho: @gdavison Got it. Added ip group disassociation from a directory on delete.

gdavison (approved):
LGTM 🚀 

--- PASS: TestAccAwsWorkspacesIpGroup_disappears (65.56s)
--- PASS: TestAccAwsWorkspacesIpGroup_basic (113.13s)
--- PASS: TestAccAwsWorkspacesIpGroup_tags (128.86s)
--- PASS: TestAccAwsWorkspacesDirectory_subnetIds (727.54s)
--- PASS: TestAccAwsWorkspacesDirectory_disappears (747.15s)
--- PASS: TestAccAwsWorkspacesDirectory_workspaceCreationProperties (747.94s)
--- PASS: TestAccAwsWorkspacesDirectory_selfServicePermissions (749.95s)
--- PASS: TestAccAwsWorkspacesIpGroup_MultipleDirectories (761.30s)
--- PASS: TestAccAwsWorkspacesDirectory_basic (771.07s)
--- PASS: TestAccAwsWorkspacesDirectory_ipGroupIds (794.51s)
--- PASS: TestAccAwsWorkspacesDirectory_tags (818.72s)

Issue #14384, opened 2020-07-29
 Technical Debt Theme: Resolve Acceptance Test Failures (RDS)


breathingdust:
### Description

This is a meta-issue which tracks the list of RDS Acceptance Tests which now consistently fail. The August-October 2020 [Roadmap](https://github.com/terraform-providers/terraform-provider-aws/blob/master/ROADMAP.md) for the AWS Provider includes a technical debt theme which decides an area of debt upon which we want to apply focus. 

The focus for this period is improving our Acceptance Test suite. This covers resolving failing and flaky tests. Our initial focus will be to resolve failing tests which exist within our [Core Service](https://github.com/terraform-providers/terraform-provider-aws/blob/master/docs/CORE_SERVICES.md) list.

## RDS Tests Currently Failing

- [x] `TestAccAWSRDSClusterEndpoint_tags` (moved to #14570)
- [x] `TestAccAWSDbInstanceDataSource_basic` #14526
- [x] `TestAccAWSDbClusterSnapshotDataSource_MostRecent` #14527
- [ ] `TestAccAWSDBInstance_S3Import` (will be covered by #13391)
- [ ] `TestAccAWSRDSCluster_SnapshotIdentifier_PreferredBackupWindow`
- [ ] `TestAccAWSRDSCluster_s3Restore` (will be covered by #13391)

bflad:
The S3 import test failures may involve an AWS Support case to resolve if it is not a simple version adjustment since they require a custom backup file and the error message is nondescript. Further tracking for that can be done with: https://github.com/terraform-providers/terraform-provider-aws/issues/13391

The `TestAccAWSRDSCluster_SnapshotIdentifier_PreferredBackupWindow` failure can be tracked with this specific issue: https://github.com/terraform-providers/terraform-provider-aws/issues/14981

PR #14527, opened 2020-08-08
tech-debt/db_cluster_snapshot: retry on create when cluster is not in a valid state

anGie44:
### Community Note

Relates #14384 

```release-note
tech-debt/db_cluster_snapshot: retry on create when cluster is not in a valid state
```
Output from acceptance testing (before):
```
TestAccAWSDbClusterSnapshotDataSource_MostRecent: testing.go:684: Step 0 error: errors during apply:

        Error: error creating RDS DB Cluster Snapshot: InvalidDBClusterStateFault: Cannot create a snapshot because the database instance is not currently in the available state.
```
Output from acceptance testing (after):

```
--- PASS: TestAccAWSDbClusterSnapshotDataSource_MostRecent (222.65s)
```


aws/data_source_aws_db_cluster_snapshot_test.go
```diff
@@ -91,7 +92,30 @@ func TestAccAWSDbClusterSnapshotDataSource_MostRecent(t *testing.T) {
...
+						Timeout:    15 * time.Minute,
+						MinTimeout: 10 * time.Second,
+						Delay:      30 * time.Second,
+					}
+
+					_, err := stateConf.WaitForState()
+					if err != nil {
+						t.Fatal(err)
+					}
+				},
```
bflad, line 116: In general, if we need to implement custom logic in acceptance testing for other resources, it means we aren't doing something properly in the resource which is also a problem that practitioners won't be able to workaround either (or at least easily). We may want to try adding `ContinuousTargetOccurence: 3` to the `aws_rds_cluster` resource creation retries (which hopefully should catch if RDS goes available -> backing-up -> available) and seeing if the single configuration can apply correctly afterwards. 👍

anGie44: yep totally works! retry time i put is arbitrary though, 2min ok?

PR #14655, opened 2020-08-14
Add note on how to use aws_wafv2_web_acl_association with Cloudfront

FnTm:
_Only changes Documentation._

The existing docs mention that this resource should only be used with an ALB or API Gateway, but doesn't mention how to get the association to work in Cloudfronts case. This PR ammends that.

### Community Note

```release-note
NONE
```

```

...
```

### DIFF
--- a/website/docs/r/wafv2_web_acl_association.html.markdown
+++ b/website/docs/r/wafv2_web_acl_association.html.markdown
@@ -10,6 +10,11 @@ description: |-
 
 Creates a WAFv2 Web ACL Association.
 
+~> **NOTE on associating a WAFv2 Web ACL with a Cloudfront distribution:** Do not use this resource to associate a WAFv2 Web ACL with a Cloudfront Distribution. The [AWS API call backing this resource][1] notes that you should use the [`web_acl_id`][2] property on the [`cloudfront_distribution`][2] instead.
+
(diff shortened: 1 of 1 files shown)


FnTm:
I've fixed the link check that was failing, and this is now ready for review

website/docs/r/wafv2_web_acl_association.html.markdown
```diff
@@ -10,6 +10,11 @@ description: |-
 
 Creates a WAFv2 Web ACL Association.
 
+~> **NOTE on associating a WAFv2 Web ACL with a Cloudfront distribution:** Do not use this resource to associate a WAFv2 Web ACL with a Cloudfront Distribution. The [AWS API call backing this resource][1] notes that you should use the [`web_acl_id`][2] property on the [`cloudfront_distribution`][2] instead.
+
+[1]: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/wafv2.html#WAFV2.Client.associate_web_acl
```
anGie44, line 15: this API link is totally correct in its information, I would just switch this with the docs hosted here https://docs.aws.amazon.com/waf/latest/APIReference/API_AssociateWebACL.html since we generally get API specs from the dev guides and defined api actions there as well

FnTm: @anGie44 thanks for the suggestion, I've changed the links and force-pushed an update. This should be good to go.

PR #14709, opened 2020-08-17
resource/aws_rds_cluster: Add allow_major_version_upgrade argument

bflad:
### Community Note

Reference: https://github.com/terraform-providers/terraform-provider-aws/issues/13874

```release-note
* resource/aws_rds_cluster: Add `allow_major_version_upgrade` argument
```

Output from acceptance testing in AWS Commercial:

```
--- PASS: TestAccAWSRDSCluster_AllowMajorVersionUpgrade (1248.28s)
```

Output from acceptance testing in AWS GovCloud (US):

```
--- PASS: TestAccAWSRDSCluster_AllowMajorVersionUpgrade (1457.47s)
```

### DIFF
--- a/aws/resource_aws_rds_cluster_test.go
+++ b/aws/resource_aws_rds_cluster_test.go
@@ -146,6 +146,56 @@ func TestAccAWSRDSCluster_basic(t *testing.T) {
 	})
 }
 
+func TestAccAWSRDSCluster_AllowMajorVersionUpgrade(t *testing.T) {
+	var dbCluster1, dbCluster2 rds.DBCluster
+	rName := acctest.RandomWithPrefix("tf-acc-test")
+	resourceName := "aws_rds_cluster.test"
+	// If these hardcoded versions become a maintenance burden, use DescribeDBEngineVersions
+	// either by having a new data source created or implementing the testing similar
+	// to TestAccAWSDmsReplicationInstance_EngineVersion
+	engine := "aurora-postgresql"
+	engineVersion1 := "10.11"
+	engineVersion2 := "11.7"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckAWSClusterDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccAWSClusterConfig_AllowMajorVersionUpgrade(rName, true, engine, engineVersion1),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSClusterExists(resourceName, &dbCluster1),
+					resource.TestCheckResourceAttr(resourceName, "allow_major_version_upgrade", "true"),
+					resource.TestCheckResourceAttr(resourceName, "engine", engine),
+					resource.TestCheckResourceAttr(resourceName, "engine_version", engineVersion1),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"allow_major_version_upgrade",
+					"apply_immediately",
+					"cluster_identifier_prefix",
+					"master_password",
+					"skip_final_snapshot",
+				},
+			},
+			{
+				Config: testAccAWSClusterConfig_AllowMajorVersionUpgrade(rName, true, engine, engineVersion2),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckAWSClusterExists(resourceName, &dbCluster2),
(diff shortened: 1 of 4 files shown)


Sovietaced:
Was just about to submit this myself as I was running into issues performing a major version upgrade using TF.

aws/resource_aws_rds_cluster_test.go
```diff
@@ -146,6 +146,60 @@ func TestAccAWSRDSCluster_basic(t *testing.T) {
...
+					"apply_immediately",
+					"cluster_identifier_prefix",
+					"master_password",
+					"skip_final_snapshot",
+				},
+			},
+			{
+				Config:      testAccAWSClusterConfig_AllowMajorVersionUpgrade(rName, false, engine, engineVersion2),
+				ExpectError: regexp.MustCompile(`InvalidParameterCombination: The AllowMajorVersionUpgrade flag must be present when upgrading to a new major version`),
+			},
```
anGie44, line 189: looking at this more, would this test still be valid if we omit this step?  because at apply time, the `rds_instance` gets destroyed (as we'd expect) while the `rds_cluster` attempts to update, which ultimately errors out with this `ExpectError` value, but then this leaves us without a recreated `rds_instance`. so in the subsequent step, at plan time, we expect to create an `instance` and modify the `cluster`, but these 2 steps happen sequentially as the `instance` references the `cluster`, leaving the `cluster` to first be modified which then ultimately fails b/c of the missing primary instance :/

anGie44: unless maybe adding some `lifecycle` handling like `create_before_destroy` on the `instance` resource was a workaround to this step's config?

bflad: Your first comment was right here. 👍  We actually do not want to have the `aws_rds_cluster_instance` to respond to `engine_version` changes at all (e.g. `ignore_changes`), preventing the recreation problem you mentioned on both errant and real engine version updates. Updating the documentation and test configuration should help.

website/docs/r/rds_cluster_instance.html.markdown
```diff
@@ -58,7 +58,7 @@ The following arguments are supported:
 For information on the difference between the available Aurora MySQL engines
 see [Comparison between Aurora MySQL 1 and Aurora MySQL 2](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/AuroraMySQL.Updates.20180206.html)
 in the Amazon RDS User Guide.
-* `engine_version` - (Optional) The database engine version.
+* `engine_version` - (Optional) The database engine version. When managing the engine version in the cluster, it is recommended to add the [lifecyle `ignore_changes` configuration](/docs/configuration/resources.html#ignore_changes) for this argument to prevent Terraform from proposing changes the instance engine version directly.
```
anGie44, line 61: Nit:
```suggestion
* `engine_version` - (Optional) The database engine version. When managing the engine version in the cluster, it is recommended to add the [lifecyle `ignore_changes` configuration](/docs/configuration/resources.html#ignore_changes) for this argument to prevent Terraform from proposing changes the instance engine version directly.
```
```suggestion
* `engine_version` - (Optional) The database engine version. When managing the engine version in the cluster, it is recommended to add the [lifecyle `ignore_changes` configuration](/docs/configuration/resources.html#ignore_changes) for this argument to prevent Terraform from proposing changes to the instance engine version directly.
```

anGie44 (approved):
Nice addition of the `lifecycle` block 👍 Just one small doc nit pick otherwise LGTM

website/docs/r/rds_cluster_instance.html.markdown
bflad: Ha looks like I misspelled lifecycle in there too. I'll pull in yours and adjust it again.

website/docs/r/rds_cluster_instance.html.markdown
```diff
@@ -58,7 +58,7 @@ The following arguments are supported:
 For information on the difference between the available Aurora MySQL engines
 see [Comparison between Aurora MySQL 1 and Aurora MySQL 2](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/AuroraMySQL.Updates.20180206.html)
 in the Amazon RDS User Guide.
-* `engine_version` - (Optional) The database engine version.
+* `engine_version` - (Optional) The database engine version. When managing the engine version in the cluster, it is recommended to add the [lifecyle `ignore_changes` configuration](/docs/configuration/resources.html#ignore_changes) for this argument to prevent Terraform from proposing changes to the instance engine version directly.
```
bflad, line 61:
```suggestion
* `engine_version` - (Optional) The database engine version. When managing the engine version in the cluster, it is recommended to add the [lifecycle `ignore_changes` configuration](/docs/configuration/resources.html#ignore_changes) for this argument to prevent Terraform from proposing changes to the instance engine version directly.
```

PR #14820, opened 2020-08-25
fix: remove 's' for consistancy between aws_waf_ipset and aws_wafregi…

julienpierini:
### Community Note

Relates OR Closes #0000

```release-note
resource/aws_waf_ipset: Rename block ip_set_descriptors to ip_set_descriptor for consistancy with ressource/aws_wafregional_ipset
```

```
unable to pay to run acceptance tests
```

### DIFF
--- a/aws/resource_aws_waf_ipset.go
+++ b/aws/resource_aws_waf_ipset.go
@@ -34,7 +34,7 @@ func resourceAwsWafIPSet() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"ip_set_descriptors": {
+			"ip_set_descriptor": {
 				Type:     schema.TypeSet,
 				Optional: true,
 				Elem: &schema.Resource{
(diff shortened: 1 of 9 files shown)


bflad:
Hi @julienpierini 👋  Thank you for submitting this.

For reasons highlighted in https://github.com/terraform-providers/terraform-provider-aws/issues/4137#issuecomment-470108161 and for consistency with our previous decision in https://github.com/terraform-providers/terraform-provider-aws/pull/5319, we would prefer to leave this resource as-is rather than work through the longer maintenance tail of work involved. In this case, Terraform AWS Provider attributes must follow the full [attribute rename best practices](https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename) to ensure that existing Terraform configurations continue to work until a new major version is released, which proved to be more effort than anticipated with other WAF attribute renames.

Apologies that this will not make it in, but we hope that focusing our efforts in other places will be beneficial to you and the community at large. Thank you again.

Issue #14529, opened 2020-08-08
Migrate SES SMTP password to v4 for existing aws_iam_access_key


gileri:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Historically `aws_iam_access_key` only had the `ses_smtp_password` attribute, which was complemented by `ses_smtp_password_v4`, then `ses_smtp_password` was deprecated.

`ses_smtp_password_v4` is [derived from the `SecretAccessKey` during resource creation](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_iam_access_key.go#L108), and currently isn't created during resource update. Users must taint+recreate `aws_iam_access_key` to be able to continue SES SMTP passwords.

Two possible reasons may explain that missing "migration" :
- The `SecretAccessKey` isn't stored by AWS, so during refresh it isn't available from AWS API's response.
- The `secret` (=`SecretAccessKey`) attribute may not be available if user chose to PGP-encrypt the key.

However, if the secret isn't PGP-encrypted, it is stored in state, so `ses_smtp_password_v4` could be calculated during resource update.

Would it be interesting to generate `ses_smtp_password_v4` during resource update ? I know that access key rotation shouldn't be a big deal, but this deprecation makes upgrading less easy.

I made a POC that seems to work, but would like input from maintainers to know if this is a good idea before polishing it up.

```diff
diff --git a/aws/resource_aws_iam_access_key.go b/aws/resource_aws_iam_access_key.go
index f6dfdc3c1..dcc1b65f7 100644
--- a/aws/resource_aws_iam_access_key.go
+++ b/aws/resource_aws_iam_access_key.go
@@ -147,7 +147,7 @@ func resourceAwsIamAccessKeyRead(d *schema.ResourceData, meta interface{}) error
        return nil
 }
 
-func resourceAwsIamAccessKeyReadResult(d *schema.ResourceData, key *iam.AccessKeyMetadata) error {
+func resourceAwsIamAccessKeyReadResult(d *schema.ResourceData, key *iam.AccessKeyMetadata, region *string) error {
[...]

PR #14829, opened 2020-08-25
Add examples of aws_iam_access_key v3 upgrade paths

jdheyburn:
### Community Note

Relates OR Closes #0000

```release-note
NONE
```

```
N/A
```

Adding more examples of upgrade paths for `aws_iam_access_key` from v2 -> v3 to the upgrade documentation.

### DIFF
--- a/website/docs/guides/version-3-upgrade.html.md
+++ b/website/docs/guides/version-3-upgrade.html.md
@@ -1054,6 +1054,10 @@ resource "aws_glue_job" "example" {
 
 In many regions today and in all regions after October 1, 2020, the [SES API will only accept version 4 signatures](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/using-ses-api-authentication.html). If referencing the `ses_smtp_password` attribute, switch your Terraform configuration to the `ses_smtp_password_v4` attribute instead. Please note that this signature is based on the region of the Terraform AWS Provider. If you need the SES v4 password in multiple regions, it may require using [multiple provider instances](/docs/configuration/providers.html#alias-multiple-provider-instances).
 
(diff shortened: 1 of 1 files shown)


Porkepix:
Does that mean there's no way to update `iam_access_key` in-place to just add the missing attribute?

jdheyburn:
@Porkepix I'm not sure - I would have though if it was possible those steps would be included in the documentation. I haven't read enough of the resource to understand exactly what the values are, I just wanted to provide a real-world example of an upgrade path for those who created the resource on an old provider version.

gileri:
Thank you @jdheyburn for this PR ! That would clarify the current situation for end-users.
Could you update your first post to indicate

>Closes #14529 

So that it's closed in case your PR is merged ?

Porkepix:
Well I'm not sure this completely close it.

It sure helps, but I'm currently in a situation where even that solution would be painful: taint the resource would render the currently used keys in production infrastructure completely ineffective for some times, up until  a new one is created and deployed.

I still think it should be possible to upgrade existing `iam_access_key` to have file the new attributes in the state, rather than having to completely replace the resource.

jdheyburn:
I would agree it does not close it out. The issue would still persist even if the PR is there. It is an issue of the `aws_iam_access_key` resource being created prior to that attribute existing.

@Porkepix - if you are looking for a seamless switch then the second paragraph added in the PR should suffice for you:

> Alternatively, you can stage the change by creating a new `aws_iam_access_key` resource and change any downstream dependencies to use the new `ses_smtp_password_v4` attribute. Once dependents have been updated with the new resource you can remove the old one.

One way this can be accomplished is by performing a `terraform state rm aws_iam_access_key.resource_name`.  Then your subsequent `terraform plan` will update the downstream dependencies in place.

IAM Access keys are supposed to be short-lived anyway. I believe this is a good compromise IMO.

breathingdust:
LGTM 🚀 Thanks @jdheyburn for the contribution!

Issue #14383, opened 2020-07-29
Technical Debt Theme: Resolve Acceptance Test Failures (EC2)


breathingdust:
### Description

This is a meta-issue which tracks the list of EC2 Acceptance Tests which now consistently fail. The August-October 2020 [Roadmap](https://github.com/terraform-providers/terraform-provider-aws/blob/master/ROADMAP.md) for the AWS Provider includes a technical debt theme which decides an area of debt upon which we want to apply focus. 

The focus for this period is improving our Acceptance Test suite. This covers resolving failing and flaky tests. Our initial focus will be to resolve failing tests which exist within our [Core Service](https://github.com/terraform-providers/terraform-provider-aws/blob/master/docs/CORE_SERVICES.md) list.

## EC2 Tests Currently Failing

- [x] `TestAccAWSSpotFleetRequest_tags`
- [x] `TestAccAWSSpotFleetRequest_launchSpecToLaunchTemplate`
- [x] `TestAccAWSInstance_EbsRootDevice_ModifyDeleteOnTermination`
- [x] `TestAccAWSInstance_EbsRootDevice_MultipleBlockDevices_ModifyDeleteOnTermination`
- [x] `TestAccAWSInstance_EbsRootDevice_ModifyAll`
- [x] `TestAccAWSInstance_metadataOptions`
- [x] `TestAccAWSSpotFleetRequest_withoutSpotPrice`
- [x] `TestAccAWSFlowLog_LogFormat`
- [ ] `TestAccAWSEIPAssociation_ec2Classic`
- [ ] `TestAccAwsEc2ClientVpn_serial` (related to Max # of certs)
- [ ] `TestAccAWSEIP_Ec2Classic`
- [ ] `TestAccAWSRouteTable_panicEmptyRoute`
- [ ] `TestAccAWSVpcPeeringConnectionOptions_sameRegionDifferentAccount`
- [ ] `TestAccAWSInstance_instanceProfileChange`
- [x] `TestAccAWSInstance_EbsBlockDevice_KmsKeyArn`
- [ ] `TestAccAWSEIPAssociation_spotInstance` (passing since 8/19 in Commercial and Gov)
- [ ] ~~`TestAccAWSInstance_ipv6_supportAddressCountWithIpv4`~~ (passing in Commercial and Gov)
- [ ] ~~`TestAccAWSRoute_changeRouteTable`~~ (passing in Commercial and Gov)
[...]

PR #14841, opened 2020-08-25
tech-debt: update SpotFleetRequest acceptance tests to use data source instance type

anGie44:
### Community Note

Relates #14383

Notes: 
* requests with  `m1.small` instance types consistently are not fulfilled so a switch here to using the `micro`'s seems to work and conveniently can be referenced from the data source in tests configs

```release-note
tech-debt: update SpotFleetRequest acceptance tests to use data source instance type
```

```
--- PASS: TestAccAWSSpotFleetRequest_overriddingSpotPrice (249.48s)
--- PASS: TestAccAWSSpotFleetRequest_multipleInstanceTypesInSameAz (273.53s)
```

### DIFF
--- a/aws/resource_aws_spot_fleet_request_test.go
+++ b/aws/resource_aws_spot_fleet_request_test.go
@@ -641,9 +641,10 @@ func TestAccAWSSpotFleetRequest_multipleInstanceTypesInSameAz(t *testing.T) {
 					testAccCheckAWSSpotFleetRequestExists(resourceName, &sfr),
 					resource.TestCheckResourceAttr(resourceName, "spot_request_state", "active"),
 					resource.TestCheckResourceAttr(resourceName, "launch_specification.#", "2"),
-					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "launch_specification.*", map[string]string{
-						"instance_type":     "m1.small",
-						"availability_zone": "us-west-2a",
+					tfawsresource.TestMatchTypeSetElemNestedAttrs(resourceName, "launch_specification.*", map[string]*regexp.Regexp{
+						// match Regexp as instance_type is one of the t2.micro or t3.micro preferred types
+						"instance_type":     regexp.MustCompile(`t(2|3).micro`),
+						"availability_zone": regexp.MustCompile("us-west-2a"),
 					}),
 					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "launch_specification.*", map[string]string{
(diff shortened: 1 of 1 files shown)


aws/resource_aws_spot_fleet_request_test.go
```diff
@@ -641,9 +641,10 @@ func TestAccAWSSpotFleetRequest_multipleInstanceTypesInSameAz(t *testing.T) {
 					testAccCheckAWSSpotFleetRequestExists(resourceName, &sfr),
 					resource.TestCheckResourceAttr(resourceName, "spot_request_state", "active"),
 					resource.TestCheckResourceAttr(resourceName, "launch_specification.#", "2"),
-					tfawsresource.TestCheckTypeSetElemNestedAttrs(resourceName, "launch_specification.*", map[string]string{
-						"instance_type":     "m1.small",
-						"availability_zone": "us-west-2a",
+					tfawsresource.TestMatchTypeSetElemNestedAttrs(resourceName, "launch_specification.*", map[string]*regexp.Regexp{
+						// match Regexp as instance_type is one of the t2.micro or t3.micro preferred types
+						"instance_type":     regexp.MustCompile(`t(2|3).micro`),
```
bflad, line 646: Drive-by comment: You may find `tfawsresource.TestCheckTypeSetElemAttrPair()` easier to use in these cases, especially as we'll want to remove the hardcoded availability zones too 😄 

```go
					tfawsresource.TestCheckTypeSetElemAttrPair(resourceName, "launch_specification.*", "data.aws_ec2_instance_type_offering.available", "instance_type"),
```

anGie44: i thought so too! but i gave that one a try and from the output it doesn't seem to handle blocks like `launch_specification` ?  
```	
* Check 4/5 error: no TypeSet element "launch_specification.*", with value "t3.micro" 
```
even though you can see that value in the state that gets printed out :/

anGie44: @bflad, https://github.com/terraform-providers/terraform-provider-aws/pull/14843 could be an approach to supporting test checks with nested attributes like we have here.

bflad: Oh yikes, yeah, `TestCheckTypeSetElemAttrPair()` is for attribute TypeSet and not block TypeSet. Sorry about that. I think this is okay for now and we can iterate on it later if necessary. 👍

PR #14968, opened 2020-09-02
[WIP] Allow aws_organizations_organization data source from delegated admin accounts

breser:
AWS has added a feature to AWS Organizations to delegated administration
of some services across the organization to other accounts than the
master account.  To make it possible to do so they've allowed accounts
that have any such delegation to use read APIs against the organization.
For example you can call ListAccounts to retrieve the list of member
accounts in such a delgated admin account.

Prior to this commit, Terraform checked that the account was the master
account and skipped setting a number of attributes for the
aws_organizations_organization data source.  This makes it hard to do
things like automatically write a policy for an S3 bucket that these
services will write to that includes an account number in the path.

### Community Note

```release-note
data-source/aws_organizations_organization: Can be used against delegated admin accounts
```

```
$ make testacc TESTARGS='-run=testAccDataSourceAwsOrganizationsOrganization_basic'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=testAccDataSourceAwsOrganizationsOrganization_basic -timeout 120m
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	1.458s [no tests to run]```

### DIFF
--- a/aws/data_source_aws_organizations_organization.go
+++ b/aws/data_source_aws_organizations_organization.go
@@ -156,7 +156,31 @@ func dataSourceAwsOrganizationsOrganizationRead(d *schema.ResourceData, meta int
 	d.Set("master_account_email", org.Organization.MasterAccountEmail)
 	d.Set("master_account_id", org.Organization.MasterAccountId)
 
+	// Some of these APIs are only allowed if the account you are in has
+	// organization admin privileges.  The org's master account always has
+	// this as well as other accounts that have delegated admin for any
+	// service.
+	var hasOrgAdmin bool
 	if aws.StringValue(org.Organization.MasterAccountId) == meta.(*AWSClient).accountid {
+		hasOrgAdmin = true
+	} else {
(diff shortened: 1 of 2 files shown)


breser:
Feedback I'd like on this.

* Should I be calling ListDelegatedAccounts at all the decide if I want to try the other APIs?  I could have removed the block and made the method try the other functions and ate access denied errors.  That we've shoved many API calls into a single data source and that AWS has added a condition that we can't detect with the API calls we have makes this somewhat ugly with respect to make access denied errors clear to users.
* Should we log a hint to users when LIstDelegatedAccounts fails with access denied with a hint to users that this must not be a delegated admin account or that they don't have permissions to ListDelegatedAdministrators it won't work against a delegated admin account?  If so at what log level?
* Do I need to adjust the acceptance test somehow?  I'm thinking no because who knows what sort of account someone is running it against.
* Should I bother to iterate the results of ListDelegatedAdministrators.  Any results implies I'm in a delegated admin account.  I wrote it this way in part to make it easy to add the delegated admins to the output of the data source potentially at a later date.  But if you think we shouldn't extend that onto this data source then does it still make sense to do that iteration at all?

And of course in general any feedback on style and the code.  I've not made very many Terraform contributions and am not an experienced Go programmer.

PR #15338, opened 2020-09-24
[WIP] Removing Restriction on Zone ID XOR Name for Route53 Hosted Zone Data Source

yhakbar:
### Community Note

Not sure about the order of operations for this PR. I believe that the restriction of requiring the Route 53 Hosted Zone data source to provide `name` or `zone_id` is unnecessary, but I don't know the context in which that restriction was placed. 

It doesn't seem to be a restriction stemming from the AWS SDK call, as the data source is calling `ListHostedZones` and filtering in Go.

Please provide insight into the reason this restriction is currently in place and what I can do to make this change safe and effective!

### DIFF
--- a/aws/data_source_aws_route53_zone.go
+++ b/aws/data_source_aws_route53_zone.go
@@ -134,7 +134,7 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 
 				if matchingTags && matchingVPC {
 					if hostedZoneFound != nil {
-						return fmt.Errorf("multiple Route53Zone found please use vpc_id option to filter")
+						return fmt.Errorf("Multiple Route53 Hosted Zones found please use zone_id or vpc_id option to filter")
 					}
 
 					hostedZoneFound = hostedZone
--- a/aws/data_source_aws_route53_zone.go
+++ b/aws/data_source_aws_route53_zone.go
@@ -73,15 +73,11 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 	name, nameExists := d.GetOk("name")
 	name = name.(string)
 	id, idExists := d.GetOk("zone_id")
-	vpcId, vpcIdExists := d.GetOk("vpc_id")
+	vpcID, vpcIDExists := d.GetOk("vpc_id")
 	tags := keyvaluetags.New(d.Get("tags").(map[string]interface{})).IgnoreAws()
 
 	if nameExists && idExists {
 		return fmt.Errorf("zone_id and name arguments can't be used together")
-	}
-
-	if !nameExists && !idExists {
-		return fmt.Errorf("Either name or zone_id must be set")
 	}
 
 	var nextMarker *string
@@ -100,16 +96,16 @@ func dataSourceAwsRoute53ZoneRead(d *schema.ResourceData, meta interface{}) erro
 			return fmt.Errorf("Error finding Route 53 Hosted Zone: %v", err)
 		}
 		for _, hostedZone := range resp.HostedZones {
-			hostedZoneId := cleanZoneID(aws.StringValue(hostedZone.Id))
-			if idExists && hostedZoneId == id.(string) {
+			hostedZoneID := cleanZoneID(aws.StringValue(hostedZone.Id))
+			if idExists && hostedZoneID == id.(string) {
 				hostedZoneFound = hostedZone
 				break
 				// we check if the name is the same as requested and if private zone field is the same as requested or if there is a vpc_id
-			} else if (trimTrailingPeriod(aws.StringValue(hostedZone.Name)) == trimTrailingPeriod(name)) && (aws.BoolValue(hostedZone.Config.PrivateZone) == d.Get("private_zone").(bool) || (aws.BoolValue(hostedZone.Config.PrivateZone) && vpcIdExists)) {
+			} else if (trimTrailingPeriod(aws.StringValue(hostedZone.Name)) == trimTrailingPeriod(name)) && (aws.BoolValue(hostedZone.Config.PrivateZone) == d.Get("private_zone").(bool) || (aws.BoolValue(hostedZone.Config.PrivateZone) && vpcIDExists)) {
 				matchingVPC := false
-				if vpcIdExists {
+				if vpcIDExists {
 					reqHostedZone := &route53.GetHostedZoneInput{}
-					reqHostedZone.Id = aws.String(hostedZoneId)
+					reqHostedZone.Id = aws.String(hostedZoneID)
 
 					respHostedZone, errHostedZone := conn.GetHostedZone(reqHostedZone)
 					if errHostedZone != nil {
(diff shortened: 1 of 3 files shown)


yhakbar:
@Ninir Hello! I believe #2680 is where the decision to enact this restriction was made. Would you mind sharing some insight into why we have it, and anything I could do to make this PR align more closely with best practices here?

Ninir:
@yhakbar Hi!checking the Git history, it actually came from https://github.com/terraform-providers/terraform-provider-aws/commit/87cee00ba90fd17c172737b73f4cd30133785bd3 which was a long time ago

Does that help?

yhakbar:
> @yhakbar Hi!checking the Git history, it actually came from [87cee00](https://github.com/terraform-providers/terraform-provider-aws/commit/87cee00ba90fd17c172737b73f4cd30133785bd3) which was a long time ago
> 
> Does that help?

I'm not sure. I think I might have mistakenly thought that before [this change](https://github.com/terraform-providers/terraform-provider-aws/pull/2680/files#diff-e01a1d64cf89b976c217be6883c66eebL64-L68), the scenario in which neither the hosted zone id or the hosted zone name is provided would allow for a successful lookup using a unique tag.

Do you happen to know the context of that conditional? I don't know if there's a special reason that we want to ensure that one and only one of zone id/ name is always provided for a hosted zone lookup, from a functional perspective, in Route53.

Ninir:
I don;t have the context, but looking at https://github.com/terraform-providers/terraform-provider-aws/pull/2680/files#diff-e01a1d64cf89b976c217be6883c66eebR18-R27 I imagine it's related to both attributes being optional and the logic making sure either of these 2 is set up to plan correctly.

I'd likely go the CustomizeFunc way now to make sure one of the 2 is required, but it's some refactoring to make :) 

Does that make sense?

yhakbar:
> I don;t have the context, but looking at https://github.com/terraform-providers/terraform-provider-aws/pull/2680/files#diff-e01a1d64cf89b976c217be6883c66eebR18-R27 I imagine it's related to both attributes being optional and the logic making sure either of these 2 is set up to plan correctly.
> 
> I'd likely go the CustomizeFunc way now to make sure one of the 2 is required, but it's some refactoring to make :)
> 
> Does that make sense?

😅 I don't think I'm following. What do you mean by going the CustomizeFunc way? Are you saying that the refactoring would be to make sure that at least one of the optional arguments is passed in? I'm still unclear as to why that's the case. Please forgive my inexperience with this stuff. I appreciate your patience 🙇🏼 .

Ninir:
> 😅 I don't think I'm following. What do you mean by going the CustomizeFunc way? Are you saying that the refactoring would be to make sure that at least one of the optional arguments is passed in? I'm still unclear as to why that's the case. Please forgive my inexperience with this stuff. I appreciate your patience 🙇🏼 .


No worries at all! so in that case, this data source is expected to return one result, based on an ID or a name. This check exists so that we make sure we call the underlying AWS API endpoint and get 0-1 result.

At the time, this check was the only way to make sure this was expecting one or the other. Nowadays, it's recommended to use https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/resource.go#L170 if I remember correctly, to control the attributes themselves.

If you want to get a list of data sources in a given format, I'd recommend creating a new data source object, that would give you a list of zones based on some filters instead.

yhakbar:
> > 😅 I don't think I'm following. What do you mean by going the CustomizeFunc way? Are you saying that the refactoring would be to make sure that at least one of the optional arguments is passed in? I'm still unclear as to why that's the case. Please forgive my inexperience with this stuff. I appreciate your patience 🙇🏼 .
> 
> No worries at all! so in that case, this data source is expected to return one result, based on an ID or a name. This check exists so that we make sure we call the underlying AWS API endpoint and get 0-1 result.
> 
> At the time, this check was the only way to make sure this was expecting one or the other. Nowadays, it's recommended to use https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/resource.go#L170 if I remember correctly, to control the attributes themselves.
> 
> If you want to get a list of data sources in a given format, I'd recommend creating a new data source object, that would give you a list of zones based on some filters instead.

So, I think there's a misunderstanding. I am not looking to create a data source that returns a list of hosted zones, though I'm happy to try to do that in a separate PR, if desired. I am only looking to introduce the ability to select an individual Hosted Zone by tag (if multiple are present with a given tag, I believe [this](https://github.com/terraform-providers/terraform-provider-aws/pull/15338/files#diff-e01a1d64cf89b976c217be6883c66eebR135-R138) will throw an error).

I've added a test case that hopefully illustrates this. I think there's likely room for improvement there, but I'm not 100% sure how.

I don't understand the reference to [this link](https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/resource.go#L170). The comment above it says that it's only allowed on regular resources and not data sources. I tried to find another data source that uses it for an example, but I'm not able to find one. Could you elaborate, please?

Ninir:
> So, I think there's a misunderstanding. I am not looking to create a data source that returns a list of hosted zones, though I'm happy to try to do that in a separate PR, if desired. I am only looking to introduce the ability to select an individual Hosted Zone by tag (if multiple are present with a given tag, I believe [this](https://github.com/terraform-providers/terraform-provider-aws/pull/15338/files#diff-e01a1d64cf89b976c217be6883c66eebR135-R138) will throw an error).
> 
> I've added a test case that hopefully illustrates this. I think there's likely room for improvement there, but I'm not 100% sure how.
> 
> I don't understand the reference to [this link](https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/resource.go#L170). The comment above it says that it's only allowed on regular resources and not data sources. I tried to find another data source that uses it for an example, but I'm not able to find one. Could you elaborate, please?

Oh I see, you want to use _tags_ then - definitely did not get that from your past comments or the PR description, sorry about that!

So yeh a tag is doable, and in this case you actually want to check if one of zone ID, name or a tag is introduced at least (as a precaution at plan time). If we remove the whole block, how do we make sure there is some information when doing the actual API call otherwise?

Let me know if that makes sense! and let's forget about https://github.com/hashicorp/terraform-plugin-sdk/blob/9da7cbac0dfcd4c7096105eeda94ea69f53c1101/helper/schema/resource.go#L170 for now, it was an answer related to my initial understanding.

yhakbar:
@Ninir Sorry for the late response, but does what I'm saying make sense?

I can include a restriction that one of VPC ID, Hosted Zone ID or Tags is populated to ensure that some filtering is going on, and I think it will work for most use cases if that's the right way to go.

I'm still new to working with TF providers, so I'd just like to know if you have any input on that last question, so I can follow best practices.

Ninir:
No worries at all :) 

I'm not sure how the actual maintainers feel about it these days (even though I was a previous maintainer of this repository I have to admit I haven't followed developments and best practices in over 2 years), but I'd suggest keeping the restriction and adding a workaround for your use case

yhakbar:
Closing this PR out. Will create a fresh PR with lessons learned here when ready to tackle this.

Thanks for your guidance @Ninir !

PR #15342, opened 2020-09-24
Use Go modules to manage tflint version

gdavison:
Use Go modules to manage version on `tflint` to use for validation and enable Dependabot.

Enabling `Dependabot` for the module will be a separate PR.

```release-note
NONE
```


tools.go
```diff
@@ -8,4 +8,5 @@ import (
 	_ "github.com/client9/misspell/cmd/misspell"
 	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
 	_ "github.com/katbyte/terrafmt"
+	_ "github.com/terraform-linters/tflint"
```
bflad, line 11: This can now be found in `tools/main.go` 👍

bflad (commented):
Now that #15391 is in, this should just be the important ~7 files. 😄  Thank you so much for doing this, it'll be super nice to have this tracked for dependency updates.

bflad (approved):
Looks good to me. 🚀  Pulling this in to also pull in the golangci-lint update.

PR #12599, opened 2020-03-31
Transfer Server VPC Endpoint Type and User Home Directory Type / Mapping

sshearn:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request

Closes #11281
Closes #11632
Closes #11724
Closes #11569
Closes #11593

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note

Transfer server now supports VPC for endpoint_type.
```

Output from acceptance testing:

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSTransferServer_Vpc'

...
```

```
$ make testacc TEST=./aws TESTARGS='-run=TestAccAWSTransferUser_restricted'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAWSTransferUser_restricted -timeout 120m
=== RUN   TestAccAWSTransferUser_restricted
=== PAUSE TestAccAWSTransferUser_restricted
=== CONT  TestAccAWSTransferUser_restricted
--- PASS: TestAccAWSTransferUser_restricted (40.44s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       41.599s
```

Note:

`TestAccAWSTransferServer_Vpc` test failed because it couldn't release the eip's on destroy due to policies on my account. But the actual test works.

PR #15373, opened 2020-09-27
Doc: `VPC` for endpoint_type to transfer_server

PChambino:
Document `VPC` value for `endpoint_type` attribute in transfer_server resource after being added by #12599.

I put `VPC_ENDPOINT` in parentheses because it is no longer possible to create this type of Transfer servers from AWS dashboard. It is only accessible via the API, so I believe it will be eventually deprecated, but I didn't find any information to support that.

### Community Note

Relates #12599

```release-note
NONE
```


bflad (approved):
Thank you for this update, @PChambino 🚀

andwoods:
@PChambino the doc doesnt show how to set Access type (internal or internet-facing) for endpoint type = VPC. How do we do this?

PR #15426, opened 2020-10-01
provider: Update Makefile fmt target to ignore ./awsproviderlint/vendor

bflad:
### Community Note

Reference: https://github.com/terraform-providers/terraform-provider-aws/pull/15391

```release-note
NONE
```

Previously was not an issue before because we were vendoring in the root directory and explicitly referencing package subdirectories. Now that a vendor directory is included in ./awsproviderlint/ we need to exclude it. Using Makefile's built-in filter-out and wildcard functions to avoid platform specific issues with find, grep, etc.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -37,7 +37,7 @@ testacc: fmtcheck
 
 fmt:
 	@echo "==> Fixing source code with gofmt..."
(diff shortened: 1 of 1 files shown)


YakDriver (approved):
Very nice!

PR #15533, opened 2020-10-06
Adds file path conditional checks for acceptance tests linters

gdavison:
Adds file path conditional checks for acceptance tests linters. Runs terraform validation when `tflint` is updated

Relates #15342

```release-note
NONE
```

N/A


.github/workflows/examples.yml
```diff
@@ -30,14 +31,9 @@ jobs:
...
-        go-version: ${{ matrix.go_version }}
+        go-version: ${{ env.GO_VERSION }}
     - name: go build
-      run: |
-        GOOS=$(go env GOOS)
-        GOARCH=$(go env GOARCH)
-        # Substitute as latest release
-        VERSION=$(git describe --abbrev=0 --match='v*.*.*' --tags || echo -n "v99.99.99")
-        go build -o ${TF_PLUGIN_CACHE_DIR}/${GOOS}_${GOARCH}/terraform-provider-aws_${VERSION}_x4 .
+      run: go build -o terraform-plugin-dir/terraform-provider-aws_v99.99.99_x5 .
```
bflad, line 36: How is `terraform-plugin-dir` being picked up? 😄  This is passing right now because its just downloading the latest released provider version (`- Downloading plugin for provider "aws" (hashicorp/aws) 3.9.0...` in the [logs](https://github.com/terraform-providers/terraform-provider-aws/runs/1222392266?check_suite_focus=true)), which works fine most of the time, but wouldn't work for new examples with unreleased resources.

gdavison: Yeah, I noticed that when I was testing. That was also the behaviour before this change, so I figured I'd leave it and fix later. Part of the problem is that at least one of the examples uses additional providers, causing problems with `terraform init -plugin-dir=terraform-plugin-dir`.

gdavison: Actually, it was easy…

.github/workflows/examples.yml
```diff
@@ -7,8 +7,10 @@ on:
     paths:
       - .github/workflows/examples.yml
       - examples/**
+      - tools/go.mod
 
 env:
+  GO_VERSION: "1.15"
   AWS_DEFAULT_REGION: us-west-2
   TF_PLUGIN_CACHE_DIR: ${{ github.workspace }}/.terraform.d/plugin-cache
```
bflad, line 15: This appears to no longer be used:

```suggestion
```

bflad (approved):
Other than the small suggestion for cleanup, let's do it. 🚀

PR #15619, opened 2020-10-13
docs/resources/wafv2_web_acl: provide a link to the AWS WAF managed rules group names

craiggenner:
Make it easier for people to find an initial list of AWS WAF managed rules groups names - this is what I suspect most people will want to use to get started so let's make it easy to find what to use.

### DIFF
--- a/website/docs/r/wafv2_web_acl.html.markdown
+++ b/website/docs/r/wafv2_web_acl.html.markdown
@@ -458,7 +458,7 @@ The `xss_match_statement` block supports the following arguments:
 
 The `excluded_rule` block supports the following arguments:
 
-* `name` - (Required) The name of the rule to exclude.
+* `name` - (Required) The name of the rule to exclude. See the [documentation](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html) for a list of names in the "AWSManagedRulesCommonRuleSet" managed rule group.
 
 ### Field to Match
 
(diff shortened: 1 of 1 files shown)


website/docs/r/wafv2_web_acl.html.markdown
```diff
@@ -458,7 +458,7 @@ The `xss_match_statement` block supports the following arguments:
 
 The `excluded_rule` block supports the following arguments:
 
-* `name` - (Required) The name of the rule to exclude.
+* `name` - (Required) The name of the rule to exclude. See the [documentation](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html) for a list of names in the "AWSManagedRulesCommonRuleSet" managed rule group.
```
anGie44, line 461: since the `excluded_rule` block is also applicable to rule groups managed by practitioners, (i.e. w/in the `rule_group_reference_statement` configuration block), let's make this statement also applicable only to rule groups managed by AWS e.g.
```suggestion
* `name` - (Required) The name of the rule to exclude. If the rule is managed by AWS, see the [documentation](https://docs.aws.amazon.com/waf/latest/developerguide/aws-managed-rule-groups-list.html) for a list of names in the "AWSManagedRulesCommonRuleSet" managed rule group.
```

anGie44 (changes requested):
Hi @craiggenner, thank you for this PR! The changes look great, just a small addition for specificity 😄

website/docs/r/wafv2_web_acl.html.markdown
craiggenner: Thanks for the suggestion @anGie44, I've taken your words and expanded on this to be clearer for this use case.

Would you be happy with the changes?

anGie44: looks great @craiggenner , thank you!

anGie44 (approved):
Thanks again @craiggenner! 🚀

PR #15683, opened 2020-10-16
Removes `dorny/paths-filter`

gdavison:
Removes `dorny/paths-filter` since it's causing GitHub Workflows problems

```release-note
NONE
```


YakDriver (approved):
Looks good!

PR #15838, opened 2020-10-26
Adds overridden version of `terrafmt` to tools modules

gdavison:
Adds the overridden version of `terrafmt` used in linters to the `tools` modules. This will allow version locking, rather than tracking a branch.

```release-note
NONE
```

N/A


bflad (approved):
I'm okay with continuing this temporarily, but we really should get on a release tag of the tool. 👍

PR #15864, opened 2020-10-27
tests/provider: Increase unit testing timeout from 2 to 5 minutes

bflad:
### Community Note

```release-note
NONE
```

For a variety of reasons, the unit testing timeout is occassionally too low for many systems including the GitHub Actions runners. There are some minor optimizations that can occur, such as ensuring the main provider instance is only configured once per test execution and reducing the number of generated TLS certificates in the acceptance tests (which also occurs during unit testing compilation), but the low timeout is arbitrary and these will not overly help. Longer term the large Go package will be split and more of the acceptance testing configuration will be in HCL instead of Go, which should spread out and overall reduce the execution time against the testing timeout.

Output from acceptance testing: N/A (unit testing in CI)


YakDriver (approved):
🥳

Issue #9330, opened 2019-07-12
Amazon EventBridge


ewbankkit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Amazon EventBridge [announced](https://aws.amazon.com/about-aws/whats-new/2019/07/introducing-amazon-eventbridge/).

At the AWS API level these are additions to the existing CloudWatch Events API.

### New or Affected Resource(s)

Sticking with the existing naming convention of `aws_cloudwatch_event_`

#### New Resource

* aws_cloudwatch_event_event_bus

#### Modified Resources

* [`aws_cloudwatch_event_permission`](https://www.terraform.io/docs/providers/aws/r/cloudwatch_event_permission.html)
* [`aws_cloudwatch_event_rule`](https://www.terraform.io/docs/providers/aws/r/cloudwatch_event_rule.html)
* [`aws_cloudwatch_event_target`](https://www.terraform.io/docs/providers/aws/r/cloudwatch_event_target.html)

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

Requires:
* https://github.com/terraform-providers/terraform-provider-aws/pull/9305

ebekker:
I would just stick the simpler `aws_cloudwatch_event_bus` -- the resource is a _bus_ in the CW Event system.

ewbankkit:
@ebekker Yes, I tend to agree. Of these CloudWatch resources, only [`aws_cloudwatch_metric_alarm`](https://www.terraform.io/docs/providers/aws/r/cloudwatch_metric_alarm.html) matches the [naming guidelines](https://github.com/terraform-providers/terraform-provider-aws/blob/master/.github/CONTRIBUTING.md#new-resource).

tanmng:
@ewbankkit given the change in commit `8295926` in #10256 by @alexanderkalach, I think this issue can no longer be closed by that PR, and we will need separate effort to handle AWS partner event bus. Is that the case? 
[...]

PR #15918, opened 2020-10-29
Create separate documentation category for EventBridge

gdavison:
CloudWatch Events has been rebranded as EventBridge, so create a new documentation category

Relates #9330

```release-note
NONE
```

N/A

### DIFF
--- a/website/allowed-subcategories.txt
+++ b/website/allowed-subcategories.txt
@@ -50,7 +50,7 @@ Elastic Load Balancing v2 (ALB/NLB)
 Elastic Map Reduce (EMR)
 Elastic Transcoder
 ElasticSearch
-EventBridge (fka CloudWatch Events)
+EventBridge (CloudWatch Events)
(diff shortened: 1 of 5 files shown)


website/allowed-subcategories.txt
```diff
@@ -50,6 +50,7 @@ Elastic Load Balancing v2 (ALB/NLB)
 Elastic Map Reduce (EMR)
 Elastic Transcoder
 ElasticSearch
+EventBridge (fka CloudWatch Events)
```
bflad, line 53: I'm not sure if `fka` (formerly known as) is a globally known acronym, especially for non-native English speakers. I would recommend just dropping it.

bflad (changes requested):

bflad (approved):
Onward and upward! 🚀

paulca99:
Just read "on the default EventBus" .... what if you're using a different EventBus ?

PR #15923, opened 2020-10-29
Adds tagging for EventBridge for aws_cloudwatch_event_* resources

gdavison:
Resources matching `aws_cloudwatch_event_*` are currently tagged with `service/cloudwatchevents`. Add tagging with `service/eventbridge` to match new branding.


bflad (approved):
LGTM 🚀 Two questions:

* Should we also stop labeling them with `service/cloudwatchevents`?
* Should we label PRs in the new way?

PR #16093, opened 2020-11-09
Fix workspaces_defaultRole notes

Tensho:
### Community Note

### Note

AWS WorkSpaces service requires `workspaces_defaultRole` IAM role to operate normally in general.

### References

- https://docs.aws.amazon.com/workspaces/latest/adminguide/workspaces-access-control.html#create-default-role

### DIFF
--- a/website/docs/r/workspaces_directory.html.markdown
+++ b/website/docs/r/workspaces_directory.html.markdown
@@ -10,51 +10,11 @@ description: |-
 
 Provides a WorkSpaces directory in AWS WorkSpaces Service.
 
+~> **NOTE:** AWS WorkSpaces service requires [`workspaces_DefaultRole`](https://docs.aws.amazon.com/workspaces/latest/adminguide/workspaces-access-control.html#create-default-role) IAM role to operate normally.
+
 ## Example Usage
 
 ```hcl
-resource "aws_vpc" "example" {
-  cidr_block = "10.0.0.0/16"
-}
-
-resource "aws_subnet" "example_a" {
-  vpc_id            = aws_vpc.example.id
-  availability_zone = "us-east-1a"
-  cidr_block        = "10.0.0.0/24"
-}
-
-resource "aws_subnet" "example_b" {
-  vpc_id            = aws_vpc.example.id
-  availability_zone = "us-east-1b"
-  cidr_block        = "10.0.1.0/24"
-}
-resource "aws_subnet" "example_c" {
-  vpc_id            = aws_vpc.example.id
-  availability_zone = "us-east-1c"
-  cidr_block        = "10.0.2.0/24"
-}
-
-resource "aws_subnet" "example_d" {
-  vpc_id            = aws_vpc.example.id
-  availability_zone = "us-east-1d"
-  cidr_block        = "10.0.3.0/24"
-}
-
-
-resource "aws_directory_service_directory" "example" {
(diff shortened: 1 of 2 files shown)


Tensho:
@gdavison Please check this out 🙇

website/docs/r/workspaces_directory.html.markdown
```diff
@@ -81,6 +109,8 @@ resource "aws_workspaces_directory" "example" {
     enable_maintenance_mode             = true
     user_enabled_as_local_administrator = true
   }
+
+  depends_on = [aws_iam_role.workspaces_default]
```
gdavison, line 113: Typically, services depend on the `aws_iam_role_policy_attachment` resources. In this case does it just check for the existence of the role itself?

website/docs/r/workspaces_directory.html.markdown
```diff
@@ -10,9 +10,37 @@ description: |-
 
 Provides a WorkSpaces directory in AWS WorkSpaces Service.
 
+~> **NOTE:** AWS WorkSpaces service requires [`workspaces_DefaultRole`](https://docs.aws.amazon.com/workspaces/latest/adminguide/workspaces-access-control.html#create-default-role) IAM role to operate normally.
+
 ## Example Usage
 
 ```hcl
+data "aws_iam_policy_document" "workspaces" {
```
gdavison, line 18: I'd prefer to see the resource that's being documented at the top. Otherwise, it can get buried in the details

gdavison (commented):
Looks good. I have a few comments

website/docs/r/workspaces_directory.html.markdown
Tensho: I don't mind. Does the reverse order make sense?

website/docs/r/workspaces_directory.html.markdown
Tensho: Yes. This is quite strange for me too – AWS WorkSpaces service requires the exact `workspaces_DefaultRole` IAM role name along with a particular set of policies. I'm happy with the case when service allows me to provide my custom role, like API Gateway CloudWatch IAM role.

gdavison: Reverse, or even just move the `aws_workspaces_directory` to the top

website/docs/r/workspaces_directory.html.markdown
gdavison: Reverse, or even just move the `aws_workspaces_directory` to the top

website/docs/r/workspaces_directory.html.markdown
gdavison: If it needs the policies as well, we should make the dependency on the `aws_iam_role_policy_attachment` resources. If all it needs is the IAM role, and the policies aren't strictly needed at creation time of the WorkSpaces directory, then the dependency can be on the `aws_iam_role` itself. Otherwise, we can get into race conditons where the configuration works sometimes and not others. Examples should be copy-pastable as much as possible.

Tensho: Good point and it makes sense to me. Pollicies are required, so I'll put a dependency on the attachment to rid of race condtions.

gdavison (approved):
LGTM 🚀

Issue #15952, opened 2020-10-31
[REGRESSION] v3.13.0 causes perpetual diff of resources containing lambda version number


lijok:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* aws_lambda_function

I'm constantly getting diffs for lambdas that have a vpc config. The diffs look like this (minified a bit, only change is always version and qualified_arn), no matter how many times I run apply

```hcl
 # module.a326_us_east_1.module.summary.aws_lambda_function.this will be updated in-place
  ~ resource "aws_lambda_function" "this" {
        arn                            = "arn:aws:lambda:us-east-1:272785785785:function:A326_summary"
        description                    = "Managed by Terraform"
        function_name                  = "A326_summary"
        handler                        = "function.handler"
        id                             = "A326_summary"
        invoke_arn                     = "arn:aws:apigateway:us-east-1:lambda:path/2015-03-31/functions/arn:aws:lambda:us-east-1:272785785785:function:A326_summary/invocations"
        last_modified                  = "2020-10-31T03:33:04.656+0000"
        layers                         = []
        memory_size                    = 512
        publish                        = true
      ~ qualified_arn                  = "arn:aws:lambda:us-east-1:272785785785:function:A326_summary:3" -> (known after apply)
        reserved_concurrent_executions = 10
        role                           = "arn:aws:iam::272785785785:role/A326_summary20201006163630280800000003"
        runtime                        = "python3.8"
        timeout                        = 900
      ~ version                        = "3" -> (known after apply)
    }
```

### Code (you will need to provide s3 bucket and key)

```hcl
terraform {
  backend local {}
}

terraform {
  required_version = "= 0.13.5"

  required_providers {
    aws = {
      version = "= 3.13.0"
      source  = "hashicorp/aws"
    }
  }
}

provider aws { region = "us-east-1" }

resource aws_vpc this {
  cidr_block                       = "10.1.0.0/16"
  instance_tenancy                 = "default"
  enable_dns_support               = true
  enable_dns_hostnames             = true
  enable_classiclink               = false
[...]

PR #16124, opened 2020-11-10
Workaround attempt for perpetual diff in lambda with vpc_config

maciejp-ro:
### Community Note

Relates #15952

```release-note
* resource/aws_lambda_function: workaround for perpetual plan with vpc_config [GH-15952]
```

WIP: Attempt at a workaround for perpetual plan on `aws_lambda_function` with `vpc_config`. #15952 is labeled "good first issue", so I tried.

I tracked the cause to `d.HasChange("vpc_config")` returning `true` even if `vpc_config` did not actually change (two results from `GetChange` are identical as confirmed by `reflect.DeepEqual`).

I'm not very familiar with either this provider's internals or mechanics of Terraform plugins in general. This might be a bug in terraform-plugin-sdk ([search for HasChange in the SDK repo](https://github.com/hashicorp/terraform-plugin-sdk/issues?q=HasChange) returns some open issues, but they seem to refer to v1 of the SDK) or some subtle issue in the provider itself. Falling back on reflection to compare values seems to be code smell, but this fixes the issue for me and might provide some hints for a real fix.

I haven't figured out how to update tests yet, I'll try to add a test case later. Publishing a draft in the meantime to gather feedback (or help maintainers with finding a proper fix).

**TODO:** Output from acceptance testing:

```

...
```

### DIFF
--- a/aws/resource_aws_lambda_function.go
+++ b/aws/resource_aws_lambda_function.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"reflect"
 	"regexp"
 	"time"
 
@@ -283,10 +284,18 @@ func updateComputedAttributesOnPublish(_ context.Context, d *schema.ResourceDiff
 		d.SetNewComputed("version")
 		d.SetNewComputed("qualified_arn")
 	}
+
 	return nil
 }
 
 func hasConfigChanges(d resourceDiffer) bool {
+	hasVpcConfigChange := d.HasChange("vpc_config")
+	if hasVpcConfigChange {
+		// This returns true when vpc_config is set even when the values are the same
+		o, n := d.GetChange("vpc_config")
+		hasVpcConfigChange = reflect.DeepEqual(o, n)
+	}
+
 	return d.HasChange("description") ||
 		d.HasChange("handler") ||
 		d.HasChange("file_system_config") ||
@@ -297,7 +306,7 @@ func hasConfigChanges(d resourceDiffer) bool {
 		d.HasChange("layers") ||
 		d.HasChange("dead_letter_config") ||
 		d.HasChange("tracing_config") ||
-		d.HasChange("vpc_config") ||
+		hasVpcConfigChange ||
 		d.HasChange("runtime") ||
 		d.HasChange("environment")
 }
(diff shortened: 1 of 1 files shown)


gdavison:
Hi @maciejp-ro, thanks for putting this together. It's helped me do some deeper investigation. It's related to https://github.com/hashicorp/terraform-plugin-sdk/issues/617; a nested structure with a `TypeSet` element always returns `true` for `HasChange()`.

aws/resource_aws_lambda_function.go
```diff
@@ -283,10 +284,18 @@ func updateComputedAttributesOnPublish(_ context.Context, d *schema.ResourceDiff
...
+
 	return nil
 }
 
 func hasConfigChanges(d resourceDiffer) bool {
+	hasVpcConfigChange := d.HasChange("vpc_config")
+	if hasVpcConfigChange {
+		// This returns true when vpc_config is set even when the values are the same
+		o, n := d.GetChange("vpc_config")
+		hasVpcConfigChange = reflect.DeepEqual(o, n)
```
gdavison, line 296: It took me a while to see this, but this is setting `hasVpcConfigChange` to `true` when they're `DeepEqual`. However, since `vpc_config` contains `TypeSet` elements, it will always return not equal.

gdavison:
For tests, you could make a copy of `TestAccAWSLambdaFunction_VPC` and its setup function `testAccAWSLambdaConfigWithVPC` and add `publish = true` to the `aws_lambda_function`. For more details on acceptance tests, see https://github.com/hashicorp/terraform-provider-aws/blob/master/docs/contributing/running-and-writing-acceptance-tests.md.

aws/resource_aws_lambda_function.go
maciejp-ro: I have tested this with a real Terraform project and this works: I don't have a plan with this change when a `lambda_function` has a `vpc_config`, and have a plan without that change. Additional debug logs show that `hasChange("vpc_config")` returns `true`, and `reflect.DeepEqual(o, n)` also returns true.

I have also tested it against a real change in `vpc_config` and made sure it shows up in plan.

However, since there can be only one `vpc_config`, if `hasChange("vpc_config.0.subnet_ids") || hasChange("vpc_config.0.security_group_ids")` does the trick, it seems to be a cleaner solution – it doesn't need reflection or conditionals. I'll test it.

ewbankkit:
@maciejp-ro Thanks for the contribution 👏.
I'm going to close this one in favor of https://github.com/hashicorp/terraform-provider-aws/pull/17610.

PR #16149, opened 2020-11-11
Docs Overview AWS [Grammatical changes]

shantanu-parate:


website/docs/index.html.markdown
```diff
@@ -67,7 +67,7 @@ explained below:
 
 ### Static Credentials
 
-!> **Warning:** Hard-coding credentials into any Terraform configuration is not
+!> **Warning:** Hard-coding credentials into any Terraform configuration are not
```
bflad, line 70: This update tripped me up a little bit as the verb here reads to me as attached to the singular subject "configuration" and not the plural "credentials", therefore the original "is" seems more correct to me.

```suggestion
!> **Warning:** Hard-coding credentials into any Terraform configuration is not
```

I think if the sentence was reworded a little around "Hard-coding credentials are not recommended in any Terraform configuration" or "It is not recommended to hard-code credentials into any Terraform configuration" it might remove the ambiguity. 👍

bflad (changes requested):
These are great updates, @shantanu-parate, just not sure about the first one. 👍

website/docs/index.html.markdown
shantanu-parate: @bflad I agree. We can reword a little **Hard-coding credentials are not recommended in any Terraform configuration** or we should proceed with the **are** I think.

website/docs/index.html.markdown
```diff
@@ -67,7 +67,7 @@ explained below:
...
 recommended, and risks secret leakage should this file ever be committed to a
```
bflad, line 71:
```suggestion
!> **Warning:** Hard-coded credentials are not recommended in any Terraform
configuration and risks secret leakage should this file ever be committed to a
```

bflad (approved):
Thanks so much, @shantanu-parate 🚀

PR #16558, opened 2020-12-03
Proposal for nullable value handling

gdavison:
A proposal for handling of nullable int and float values. This should be extended for nullable booleans as well. After evaluation in the AWS Provider, it could be moved to the Plugin SDK.

`nullable.ValidateTypeStringNullableFloat()` replaces the existing `validateTypeStringNullableFloat()` in the AWS Provider. There is currently no Int equivalent in the AWS Provider, but there is one in the Kubernetes Provider (https://github.com/hashicorp/terraform-provider-kubernetes/blob/aae2fefb5bbe8fde1c2083d11ab2eaaaa63de657/kubernetes/validators.go#L204-L220).

Adds nullable value equivalents to `validation.<Type>Between()`, `validation.<Type>AtLeast()`, and `validation.<Type>AtMost()`. Could be extended with equivalents to `validation.IntDivisibleBy()`, `validation.IntInSlice()`, and `validation.IntNotInSlice()`.

Adds `nullable.TypeNullableInt` and `nullable.TypeNullableFloat` as aliases for `schema.TypeString` to make intention more clear and potentially guide linting, e.g. to require one of the appropriate validation functions.

Implements Go types to wrap the string representations of nullable values and adds functions `IsNull()` and `Value()`.

Using the type wrapper, we could replace the following code

```go
if v, ok := m["instance_warmup"]; ok {
	if s, ok := v.(string); ok && s != "" {
		i, _ := strconv.ParseInt(s, 10, 64)
		refreshPreferences.InstanceWarmup = aws.Int64(i)
	}
}
```

with

```go
if v, ok := m["instance_warmup"]; ok {
	if i, null, _ := nullable.Int(v.(string)).Value(); !null {
		refreshPreferences.InstanceWarmup = aws.Int64(i)
	}
}
```

```release-note
NONE
```

N/A

### DIFF
--- a/aws/internal/nullable/float.go
+++ b/aws/internal/nullable/float.go
@@ -66,7 +66,6 @@ func ValidateFloatBetween(min, max float64) schema.SchemaValidateFunc {
 		}
 
 		if value == "" {
-			es = append(es, fmt.Errorf("expected %s to be in the range (%f - %f), got null", k, min, max))
 			return
 		}
 
@@ -95,7 +94,6 @@ func ValidateFloatAtLeast(min float64) schema.SchemaValidateFunc {
 		}
 
 		if value == "" {
-			es = append(es, fmt.Errorf("expected %s to be at least (%f), got null", k, min))
 			return
 		}
 
@@ -124,7 +122,6 @@ func ValidateFloatAtMost(max float64) schema.SchemaValidateFunc {
 		}
 
 		if value == "" {
-			es = append(es, fmt.Errorf("expected %s to be at most (%f), got null", k, max))
 			return
 		}
 
--- a/aws/internal/nullable/float.go
+++ b/aws/internal/nullable/float.go
@@ -1,0 +1,143 @@
+package nullable
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const (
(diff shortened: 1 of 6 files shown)


aws/internal/nullable/float.go
```diff
@@ -0,0 +1,140 @@
...
+	value, err := strconv.ParseFloat(string(f), 64)
+	if err != nil {
+		return 0, false, err
+	}
+	return value, false, nil
+}
+
+// ValidateTypeStringNullableFloat provides custom error messaging for TypeString floats
+// Some arguments require a floating-point value or unspecified, empty field.
+func ValidateTypeStringNullableFloat(v interface{}, k string) (ws []string, es []error) {
```
gdavison, line 40: I'm not sure if this full naming style is better, or the shorter names below. Shorter seems to fit better with Go conventions.

gdavison:
In general, this needs a visit from the refactoring fairy, since there's a lot of duplication, but I'll wait for consensus on the approach before spending more time on this.

aws/internal/nullable/float.go
jrhouston: +1 to shortening the name, it's a bit verbose at the moment.

ewbankkit:
Relates:

* https://github.com/hashicorp/terraform-provider-aws/pull/19066

Issue #12995, opened 2020-04-24
Ignore/Fix and Enable awsproviderlint AWSAT003: Regions should not be hardcoded, use aws_region and aws_availability_zones data sources instead


breathingdust:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

To ensure test configurations are region and partition agnostic, any hardcoded AZ identifiers ([a-z]{2}(-[a-z]+)+-[1-9][0-9]*(-[a-z]+-[1-9][0-9]*)?[a-z]) should be replaced with the aws_availability_zones data source.

To ensure test configurations are region and partition agnostic, any hardcoded Region identifiers ([a-z]{2}(-[a-z]+)+-[1-9][0-9]*) should be replaced with the `aws_region` data source.

#### Flagged Code

In test configuration:

```
us-west-2
```

```
us-west-2a
```

#### Passing Code

In test configuration:

```
data "aws_region" "current" {}
```

```
data "aws_availability_zones" "available" {
  # ... configuration ...
}
```

## Definition of Done

* `AWSAT003` check enabled in `GNUmakefile`
* CI testing passes

YakDriver:
This was added as [AWSAT003](https://github.com/terraform-providers/terraform-provider-aws/tree/master/awsproviderlint/passes/AWSAT003). This is the central management issue.

Fixes needed:
- [ ] cloudfront_distribution_configuration_structure_test
- [ ] config_test
- [ ] data_source_aws_acmpca_certificate_authority_test
- [ ] data_source_aws_arn_test
- [ ] data_source_aws_cloudtrail_service_account
- [ ] data_source_aws_cur_report_definition_test
- [ ] data_source_aws_db_instance_test
- [ ] data_source_aws_ebs_snapshot_ids_test
- [ ] data_source_aws_ec2_transit_gateway_vpc_attachment_test
- [ ] data_source_aws_eip
- [ ] data_source_aws_eip_test
- [ ] data_source_aws_elastic_beanstalk_hosted_zone
- [ ] data_source_aws_elastic_beanstalk_hosted_zone_test
[...]

PR #16602, opened 2020-12-04
tests/resource/workspaces_directory: Fix hardcoded region

YakDriver:
### Community Note

Relates #12995 [Phase 2](https://github.com/hashicorp/terraform-provider-aws/issues/12995#issuecomment-730680494)

```release-note
NONE
```

### Acceptance Tests in GovCloud

```
--- SKIP: TestAccAwsWorkspacesDirectory_basic (2.36s)
```

### Acceptance Tests in `us-west-2`

```
--- PASS: TestAccAwsWorkspacesDirectory_basic (615.53s)
```

### Acceptance Tests in `us-east-1`

```
--- PASS: TestAccAwsWorkspacesDirectory_basic (911.57s)
```

### DIFF
--- a/aws/resource_aws_workspaces_directory_test.go
+++ b/aws/resource_aws_workspaces_directory_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/workspaces"
 	multierror "github.com/hashicorp/go-multierror"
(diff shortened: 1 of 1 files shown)


aws/resource_aws_workspaces_directory_test.go
```diff
@@ -515,7 +516,7 @@ data "aws_availability_zones" "available" {
 
 locals {
   region_workspaces_az_ids = {
```
Tensho, line 518: @YakDriver The basic idea behind this map providing exact AZs (with ids), where the WorkSpaces service is available. If you need to test workspace directory resource in Gov region simply add extra mapping of region name and respective AZ ids. Then `lookup` function will pick up it according to the region, where the test is run.

bflad (dismissed):
In this case, I think lint ignoring is fine and we should prefer to keep the region name in the configuration for clarity, since it looks like the `local` is written to be expanded in the future. 👍  (I have not done any research to see if there is a better solution to the "acceptable AZs for WorkSpaces" problem.)

YakDriver:
@bflad I try to avoid lint ignoring big blocks of config since offending stuff can be inserted anywhere in the config. But, not sure what else to do here. It's "fixed."

bflad (approved):
Looks good to me, (although the test on this branch is failing for unrelated reasons fixed on the main branch).

Output from acceptance testing in AWS Commercial:

```
=== CONT  TestAccAwsWorkspacesDirectory_basic
    resource_aws_workspaces_directory_test.go:63: Step 1/2 error: Check failed: 1 error occurred:
        	* Check 16/24 error: aws_workspaces_directory.main: Attribute 'tags.%' expected "0", got "1"
```

Output from acceptance testing in AWS GovCloud (US):

```
--- SKIP: TestAccAwsWorkspacesDirectory_basic (2.26s)
```

Issue #16554, opened 2020-12-02
provider: Remove extraneous AWS Go SDK value wrapping in d.Set()


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The [`(*schema.ResourceData).Set()` receiver method](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.Set) automatically handles pointer types and `nil` values, so wrapping the values with the AWS Go SDK pointer conversion functions, e.g. `aws.StringValue()` is extraneous.

Note: This should _not_ include slice/map functions calls such as `aws.StringValueSlice()` since those conversions are not automatic.

### Example semgrep Rule

```yaml
rules:
  - id: helper-schema-ResourceData-Set-extraneous-value-pointer-conversion
    languages: [go]
    message: AWS Go SDK pointer conversion function for `d.Set()` value is extraneous
    paths:
      include:
        - aws/
    patterns:
      - pattern-either:
        - pattern: d.Set($ATTRIBUTE, aws.BoolValue($APIOBJECT))
        - pattern: d.Set($ATTRIBUTE, aws.Int64Value($APIOBJECT))
        - pattern: d.Set($ATTRIBUTE, int(aws.Int64Value($APIOBJECT)))
        - pattern: d.Set($ATTRIBUTE, aws.StringValue($APIOBJECT))
    severity: WARNING
```

### Example Reports

```
aws/resource_aws_wafv2_rule_group.go
severity:warning rule:helper-schema-ResourceData-Set-extraneous-value-wrapping: AWS Go SDK value wrapping in `d.Set()` is extraneous
188:	d.Set("name", aws.StringValue(resp.RuleGroup.Name))
189:	d.Set("capacity", aws.Int64Value(resp.RuleGroup.Capacity))
190:	d.Set("description", aws.StringValue(resp.RuleGroup.Description))
191:	d.Set("arn", aws.StringValue(resp.RuleGroup.ARN))
192:	d.Set("lock_token", aws.StringValue(resp.LockToken))

[...]

PR #16655, opened 2020-12-09
tech-debt/data-sources: autofix extraneous value conversion for d.Set() of primitives

anGie44:
### Community Note

Relates #16554 

```release-note
N/A
```

```
N/A semgrep rule (defined in https://github.com/hashicorp/terraform-provider-aws/pull/16650) should not show warning msgs for these files
```

### DIFF
--- a/aws/cloudfront_cache_policy_structure.go
+++ b/aws/cloudfront_cache_policy_structure.go
@@ -224,10 +224,10 @@ func setParametersConfig(parametersConfig *cloudfront.ParametersInCacheKeyAndFor
 }
 
 func setCloudFrontCachePolicy(d *schema.ResourceData, cachePolicy *cloudfront.CachePolicyConfig) {
-	d.Set("comment", aws.StringValue(cachePolicy.Comment))
-	d.Set("default_ttl", aws.Int64Value(cachePolicy.DefaultTTL))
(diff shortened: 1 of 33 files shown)


bflad:
Hey @anGie44 👋  Thinking about this particular change more, we may need to split this up into ones that are "safe" to do before a major release and not. The semantics of the attribute values when wrapped versus not could cause an unexpected change where new expression errors are generated.

Previously, the wrappers would always return a (zero) value. In practice, this means that configuration attribute references would always resolve without needing to worry about `null` or a non-existent attribute reference. After removing the wrapper, if the API does not always send back a value, expressions will need to potentially handle these now `null` attributes.

From a tactical perspective, this probably means we will need to chunk this into two categories after reviewing the AWS API documentation of each attribute being updated and verify that it _should_ be returning with a value:

- If so, those cases are okay to update.
- If not or where it is ambiguous whether it would always return with a value, erring on the side of caution and adding `nosem` ignores. We can add a 4.0 issue to sweep the rest of these.

It is also possible that this change actually may be okay for most use cases, but either which way we would need to find an attribute that is unset from the API before and after to see how configurations handle the change. 👍

bflad (approved):
Looks great! 🚀

PR #16748, opened 2020-12-14
docs/contributing: Add data handling and conversion documentation

bflad:
### Community Note

```release-note
NONE
```

Initial Contributing Guide documentation on how to handle data and its conversion. Provides a walkthrough of the current Terraform Plugin SDK type system, AWS (API and SDK) type systems, and how to convert between the two. Also includes an initial section on how the maintainers treat sensitive values in terms of the Terraform Plugin SDK implementation that exists today.

In the future, this documentation can be augmented with further information on how to handle "nullable" types and other data concerns such as "passthrough" and "virtual" attributes.

Output from acceptance testing: N/A (documentation)


docs/contributing/data-handling-and-conversion.md
```diff
@@ -0,0 +1,809 @@
...
+    - A special type implementation is required within the schema to workaround the limitation.
+
+The maintainers can provide guidance on appropriate solutions for cases not mentioned in the [Recommended Implementation section](#recommended-implementations).
+
+### Root Attributes Versus Block Attributes
+
+All Attributes and Blocks at the top level of `schema.Resource` `Schema` are considered "root" attributes. These will always be handled with receiver methods on `ResourceData`, such as reading with `d.Get()`, `d.GetOk()`, etc. and writing with `d.Set()`. Any nested Attributes and Blocks inside those root Blocks will then be handled with standard Go types according to the table in the [Type Mapping section](#type-mapping).
+
+By convention in the codebase, each level of Block handling beyond root attributes should be separated into "expand" and "flatten" functions, typically named `expand{Service}{Type}` and `flatten{Service}{Type}`. The [Recommended Implementations section](#recommended-implementations) will go into those details.
```
breathingdust, line 129: It would be nice to be explicit here about what an `Expand` function does vs `Flatten`

bflad: Expanded (😂 ) on here: https://github.com/hashicorp/terraform-provider-aws/pull/16748/commits/6521344dc1169ab6f6010f0debde8598917371a8

breathingdust (approved):
This is great, I think this will be a really valuable for new contributors and equally as a quick reference for more established ones. 🚀 👍

docs/contributing/data-handling-and-conversion.md
```diff
@@ -0,0 +1,809 @@
...
+
+To read:
+
+```go
+func expandServiceStructure(tfMap map[string]interface{}) *service.Structure {
+    // ...
+
+    if v, ok := tfMap["nested_attribute_name"].([]interface{}); ok && len(v) > 0 {
+        apiObject.NestedAttributeName = expandServiceStructure(v[0].map[string]interface{})
+    }
```
ewbankkit, line 551: There are cases (e.g. #16816 and I've seen others) where a resource is marked as `Required: true` and all its elements are marked `Optional: true`.
In this case a configuration something like

```hcl
resource "aws_svc_thing" "example" {
  flump {}
}
```

indicates that the default value for the Flump property is to be used and this is to be specified via the API as

```go
input.Flump = &svc.Flump{}
```

But `len(v) == 0` so `input.Flump = nil`.

ewbankkit: I wrote some quick code to log attributes that have this property:

```go
func TestProvider_RequiredContainingAllOptional(t *testing.T) {
	for resourceName, resource := range Provider().ResourcesMap {
		for topLevelAttributeName, v := range resource.Schema {
			if v.Type == schema.TypeList || v.Type == schema.TypeSet {
				switch ty := v.Elem.(type) {
				case *schema.Resource:
					rrrContainer(t, []string{resourceName, topLevelAttributeName}, ty.Schema, false)
				}
			}
		}
	}
}

// The current schema is a container (List or Set)
func rrrContainer(t *testing.T, path []string, m map[string]*schema.Schema, required bool) {
	allOptional := true

	for attributeName, v := range m {
		if v.Type == schema.TypeList || v.Type == schema.TypeSet {
			switch ty := v.Elem.(type) {
			case *schema.Resource:
				rrrContainer(t, append(append([]string{}, path...), attributeName), ty.Schema, v.Required)
			}
		}

		if !(v.Optional && v.Default == nil) {
			allOptional = false
		}
	}

	if required && allOptional {
		t.Logf("%s is Required and has only Optional attributes without Defaults", strings.Join(path, "."))
	}
}
```

```console
$ go test -v ./aws -run TestProvider_RequiredContainingAllOptional
```

and there are only a dozen or so outside the `wafv2` resources.
Some are already covered by having an additional `ExactlyOneOf` attribute.

PR #17067, opened 2021-01-12
.github/workflows: Reduce goreleaser parallelism to match GitHub Actions hosted runners

bflad:
### Community Note

```release-note
NONE
```

Most of the GitHub Actions logging is either missing or cutoff without explanation, but finally found one with a `signal: killed` entry. We may be triggering oom-killer or another abuse protection mechanism, so this attempts to reduce the footprint on the runner. I did not immediately see a way to dynamically fetch the processor count via workflow information.

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -27,7 +27,7 @@ jobs:
       - name: goreleaser release
         uses: goreleaser/goreleaser-action@v2
         with:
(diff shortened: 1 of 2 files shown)


YakDriver (approved):
10 thumbs up!

bflad:
I'm going to trigger failed runs after ensuring they pick up the new setting. 👍

PR #17084, opened 2021-01-13
infrastructure: allow multiple workspaces

breathingdust:
### Description

Change remote to use the prefix attribute allowing multiple workspaces. Added TF 0.15 as a label while I'm here.

### DIFF
--- a/infrastructure/repository/main.tf
+++ b/infrastructure/repository/main.tf
@@ -3,7 +3,7 @@ terraform {
     organization = "hashicorp-v2"
 
     workspaces {
-      name = "terraform-provider-aws-repository"
+      prefix = "terraform-provider-aws-"
     }
   }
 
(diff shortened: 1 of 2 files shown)


infrastructure/repository/main.tf
```diff
@@ -3,7 +3,7 @@ terraform {
     organization = "hashicorp-v2"
 
     workspaces {
-      name = "terraform-provider-aws-repository"
+      prefix = "terraform-provider-aws-"
```
bflad, line 6: I believe this will require some manual `terraform init` work to migrate, but I'm unsure of the details. We'll just want to ensure any instructions for switching workspaces is documented in the `infrastructure/repository` README.

bflad (approved):

breathingdust:
No longer required.

PR #17116, opened 2021-01-14
Update Data Handling and Conversion guide: Fix typo and encourage error checking

ewbankkit:
### Community Note

Encourage checking of any error returned from `d.Set()` for lists and sets.

```release-note
NONE
```

```
n/a
```

### DIFF
--- a/docs/contributing/data-handling-and-conversion.md
+++ b/docs/contributing/data-handling-and-conversion.md
@@ -40,7 +40,7 @@ At the bottom of this documentation is a [Glossary section](#glossary), which ma
 
 ## Data Conversions in Terraform Providers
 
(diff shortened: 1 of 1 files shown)


bflad (approved):
LGTM 🚀 

One thing to consider is that recent updates to the acceptance testing framework in the Terraform Plugin SDK have partially negated the need for this error checking given that it will cause a panic. On the flip side, it is possible for acceptance testing to not cover all data scenarios and hiding this error would prevent operators from knowing about the potential issue. It can be confusing for developers to choose whether or not to include the error check since always including can reduce code readability.

In the future with Terraform Plugin SDK v2 compatible resources, we may want to consider a simple diagnostic wrapper always be used for simplicity and consistency (even better would be different `.SetX` receiver methods that return diagnostics directly). e.g.

```
resource ExampleThingRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
  var diags diag.Diagnostics
// ...
  diags = append(diags, attributeSetDiagnostic(d.Set("attribute_name", /* ... */)))
// ...
  return diags
}
```

PR #17137, opened 2021-01-15
fix imagebuilder-pipeline image_recipe_arn ForceNew

Goryudyuma:
### Community Note

Relates OR Closes #0000

```release-note
fix: "ForceNew" in "aws_imagebuilder_image_pipeline.image_recipe_arn" changed to "True".
```

```

...
```

---

If you keep the previous settings, you will get an error when you change "image_recipe".
```
Error: error deleting Image Builder Image Recipe (arn:aws:imagebuilder:ap-northeast-1:12345:image-recipe/example/1.0.0): ResourceDependencyException: Resource dependency error: The resource ARN 'arn:aws:imagebuilder:ap-northeast-1:12345:image-recipe/example/1.0.0' has other resources depended on it.
```
This is a phenomenon that occurs when "image_recipe" cannot be deleted because "ImagePipeline" depends on "image_recipe".
After making this change, the "ImagePipeline" will be recreated and no error will occur.

What I want is that changing "image_recipe" doesn't cause "ImagePipeline" to be an error.
This PR can do that, but if there's a better way, that's fine.
Please tell me what to do after this.

### DIFF
--- a/aws/resource_aws_imagebuilder_image_pipeline.go
+++ b/aws/resource_aws_imagebuilder_image_pipeline.go
@@ -63,6 +63,7 @@ func resourceAwsImageBuilderImagePipeline() *schema.Resource {
 			"image_recipe_arn": {
 				Type:         schema.TypeString,
 				Required:     true,
+				ForceNew:     true,
 				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^arn:aws[^:]*:imagebuilder:[^:]+:(?:\d{12}|aws):image-recipe/[a-z0-9-_]+/\d+\.\d+\.\d+$`), "valid image recipe ARN must be provided"),
 			},
 			"image_tests_configuration": {


bflad:
Hi @Goryudyuma 👋  Thank you for submitting this.

Unfortunately this change is not valid -- the `aws_imagebuilder_image_pipeline` Terraform resource implements in-place update support for the `image_recipe_arn` attribute by calling the [`UpdateImagePipeline` API](https://docs.aws.amazon.com/imagebuilder/latest/APIReference/API_UpdateImagePipeline.html#imagebuilder-UpdateImagePipeline-request-imageRecipeArn):

https://github.com/hashicorp/terraform-provider-aws/blob/9a94e4fe535ba7986c1e4985aa68f656d76194c0/aws/resource_aws_imagebuilder_image_pipeline.go#L254-L303

Which is verified with the `TestAccAwsImageBuilderImagePipeline_ImageRecipeArn` acceptance test:

https://github.com/hashicorp/terraform-provider-aws/blob/9a94e4fe535ba7986c1e4985aa68f656d76194c0/aws/resource_aws_imagebuilder_image_pipeline_test.go#L238-L270

So while Terraform may not be operating as you expect in this situation, the resource and schema are currently valid. My best suggestion would be to potentially open a GitHub issue following the issue template so the maintainers can attempt to troubleshoot any potential bugs or if you have a question about how Terraform is operating to use the [HashiCorp Community Forums](https://discuss.hashicorp.com/c/terraform-providers) where there are far more people ready to help, whereas the GitHub issues here are generally monitored only by the [small set of code maintainers](https://github.com/terraform-providers/terraform-provider-aws/blob/master/docs/FAQ.md#who-are-the-maintainers).

Very briefly though, you may want to see if your configuration should be using the [`lifecycle` block `create_before_destroy` argument](https://www.terraform.io/docs/configuration/meta-arguments/lifecycle.html#create_before_destroy) in this situation -- that can invert the order of operations in Terraform.

Goryudyuma:
Hi @bflad !
Thank you for your comment!
As you advised, I wrote "create_before_destroy" in "aws_imagebuilder_image_recipe" and it worked as intended.
I am very happy!
Thank you very much for your help!

Issue #17225, opened 2021-01-21
aws_elb_hosted_zone_id are wrong to the China region


shell-autonomic-sh:
Provider version `3.24.1`, it's a bug in here.

https://github.com/hashicorp/terraform-provider-aws/blob/master/aws/data_source_aws_elb_hosted_zone_id.go#L23

I refer the data sheet from here.

https://docs.amazonaws.cn/en_us/general/latest/gr/elb.html

As we can see, the ID of China Ningxia Region (cn-northwest-1) is ZM7IZAIOVVDZF. But in the code it's Z3BX2TMKNYI13Y.

bflad:
The fix for this has been merged and will release with version 3.26.0 of the Terraform AWS Provider, later this week. Thank you to @shell-autonomic-sh for the implementation. 👍

PR #17226, opened 2021-01-21
Update data_source_aws_elb_hosted_zone_id

shell-autonomic-sh:
To fix this bug.
aws_elb_hosted_zone_id are wrong to the China region
https://github.com/hashicorp/terraform-provider-aws/issues/17225


bflad (approved):
Thank you for this fix, @shell-autonomic-sh 🚀

PR #17346, opened 2021-01-28
changelog/add manual trigger

breathingdust:
### Community Note

Adds additional manual trigger action to generate changelog. We may change the original action to `pull_request_target` once further testing has been completed.


.github/workflows/generate_changelog_manual.yml
```diff
@@ -0,0 +1,18 @@
+name: Generate CHANGELOG (Manual)
+on:
+  workflow_dispatch:
```
bflad, line 3: If the only difference is the commit message, I believe these two workflows can be combined. For example, the snapshot  workflow: https://github.com/hashicorp/terraform-provider-aws/blob/f5c1c211773868630fac91a8f47e447ff2c3e398/.github/workflows/snapshot.yml#L3-L6

breathingdust: The triggered one does an additional check to see if the PR trigger was merged rather than closed, and the manual one will not have a context of PR, that's why they are separate:

https://github.com/hashicorp/terraform-provider-aws/blob/ad8293cc2ec8556783e4896eed33ae97e56f5919/.github/workflows/generate_changelog.yml#L6-L8

bflad: Wouldn't this work then? 😄  `if: github.event_name != 'pull_request' || github.event.pull_request.merged`

.github/workflows/generate_changelog.yml
```diff
@@ -4,17 +4,14 @@ on:
...
       - run: cd tools && go install github.com/hashicorp/go-changelog/cmd/changelog-build
+      - run: ./scripts/generate-changelog.sh
       - run: |
-          ./scripts/generate-changelog.sh
-          git config --local user.email "changelogbot@hashicorp.com"
-          git config --local user.name "changelogbot"
-          git add CHANGELOG.md
-          git commit -m "Update CHANGELOG.md for #${{ github.event.pull_request.number }}" 
-          git push   
+          if [ ${{github.event_name == 'workflow_dispatch'}} ]; then MSG="Update CHANGELOG.md (Manual Trigger)"; else MSG="Update CHANGELOG.md for #${{ github.event.pull_request.number }}"; fi
```
breathingdust, line 16: Bit of a different approach, use the same workflow, and allow the changeset handling to skip if no changes are present.

scripts/push-changelog-changes.sh
```diff
@@ -0,0 +1,16 @@
...
+set -o nounset
+
+EMAIL=$1
+UNAME=$2
+MSG=$3
+
+if [[ `git status --porcelain` ]]; then
+    git config --local user.email $1
+    git config --local user.name $2
+    git add CHANGELOG.md
```
breathingdust, line 13: I kept this specific to changelog as if we do have automation pushing other files I want to be intentional about it.

.github/workflows/generate_changelog.yml
```diff
@@ -4,17 +4,14 @@ on:
...
+      - run: ./scripts/generate-changelog.sh
       - run: |
-          ./scripts/generate-changelog.sh
-          git config --local user.email "changelogbot@hashicorp.com"
-          git config --local user.name "changelogbot"
-          git add CHANGELOG.md
-          git commit -m "Update CHANGELOG.md for #${{ github.event.pull_request.number }}" 
-          git push   
+          if [ ${{github.event_name == 'workflow_dispatch'}} ]; then MSG="Update CHANGELOG.md (Manual Trigger)"; else MSG="Update CHANGELOG.md for #${{ github.event.pull_request.number }}"; fi      
+          ./scripts/push-changes.sh "changelogbot@hashicorp.com" "changelogbot" "$MSG"
```
bflad, line 17: Nit: Are we expecting anywhere else to run this script? Seems like more commonly a maintainer would manually doing the changelog update and committing with their Git configuration. (Nota bene: seems preferable to avoid shell script sprawl.)

bflad (approved):
Otherwise, looks good to me 👍

Issue #16717, opened 2020-12-11
Amazon EMR on Amazon EKS


ewbankkit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

> Amazon EMR on EKS provides a deployment option for Amazon EMR that allows you to run open-source big data frameworks on Amazon Elastic Kubernetes Service (Amazon EKS).

### New or Affected Resource(s)

* `aws_emrcontainers_virtual_cluster`

### Potential Terraform Configuration

```hcl
resource "aws_emrcontainers_virtual_cluster" "example" {
  name = "example-cluster"

  container_provider {
    id = aws_eks_cluster.example.id
  }
}
```

### References

[Blog post](https://aws.amazon.com/blogs/aws/new-amazon-emr-on-amazon-elastic-kubernetes-service-eks/).
[Developer Guide](https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/emr-eks.html).
[API Reference](https://docs.aws.amazon.com/emr-on-eks/latest/APIReference/Welcome.html)

Requires [AWS SDK v1.36.4](https://github.com/aws/aws-sdk-go/releases/tag/v1.36.4):

* https://github.com/hashicorp/terraform-provider-aws/pull/16713

jmazumder:
👍

jmazumder:
👍

shuheiktgw:
I'm going to work on this issue!

shuheiktgw:
Added emrcontainers service as a first step so would anybody review the PR? 🙏 
- https://github.com/hashicorp/terraform-provider-aws/pull/17065

shuheiktgw:
While implementing it, I noticed problems...
https://github.com/hashicorp/terraform-provider-aws/pull/17355#issuecomment-772095379

mickmorseMB:
@shuheiktgw any updates on getting this in?

shuheiktgw:
[...]

Issue #16542, opened 2020-12-02
Support Managing EKS Cluster Add-Ons


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Announced at re:Invent 2020:

> Amazon Elastic Kubernetes Service (Amazon EKS) now allows you to install and manage Kubernetes operational software (add-ons) as part of the Amazon EKS console, CLI, and API. Today, EKS supports managing the Amazon VPC CNI networking plugin, with more add-ons coming soon.

This is for supporting the management of an add-on itself.

### New Resource

* aws_eks_addon

### Potential Terraform Configuration

```hcl
resource "aws_eks_addon" "example" {
  addon_name = ""
  addon_version = ""
  cluster_name = aws_eks_cluster.example.name
  resolve_conflicts = ""
  service_account_role_arn = ""
  tags = {}
}
```

### References

* [Announcement](https://aws.amazon.com/about-aws/whats-new/2020/12/amazon-eks-simplifies-installation-and-management-for-kubernetes-culster-add-ons/)
* [User Guide](https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html#update-cluster-add-ons)
* [API Reference](https://docs.aws.amazon.com/eks/latest/APIReference/Welcome.html)
* [AWS Go SDK Service Reference](https://docs.aws.amazon.com/sdk-for-go/api/service/eks/)

shuheiktgw:
Let me see if I can implement it!

juozasget:
Hey @shuheiktgw, I already have some working code, but no acceptance tests yet. Maybe you'd like to collaborate on this?

shuheiktgw:
@juozasget Thank you, but then I'll check other issues so please add ACCs and finish up the PR! If you need any help please let me know! 👍

juozasget:
Thanks! Working on it!

juozasget:
I'll just mention it here - in case the issue has more visibility. ACC test are added and I think the PR is in good shape for a review. Maybe someone could have a look? :)

DonBower:
[...]

PR #17355, opened 2021-01-29
resource/aws_emrcontainers_virtual_cluster: new resource

shuheiktgw:
### Community Note

Relates https://github.com/hashicorp/terraform-provider-aws/issues/16717

```

...
```

### DIFF
--- a/aws/resource_aws_emrcontainers_virtual_cluster.go
+++ b/aws/resource_aws_emrcontainers_virtual_cluster.go
@@ -1,10 +1,8 @@
 package aws
 
 import (
-	"encoding/json"
 	"fmt"
 	"log"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -111,6 +109,8 @@ func resourceAwsEMRContainersVirtualClusterCreate(d *schema.ResourceData, meta i
 	if _, err := waiter.VirtualClusterCreated(conn, aws.StringValue(out.Id)); err != nil {
 		return fmt.Errorf("error waiting for EMR containers virtual cluster (%s) creation: %w", d.Id(), err)
 	}
+
+	d.SetId(aws.StringValue(out.Id))
 
 	return resourceAwsEMRContainersVirtualClusterRead(d, meta)
 }
--- a/aws/resource_aws_emrcontainers_virtual_cluster.go
+++ b/aws/resource_aws_emrcontainers_virtual_cluster.go
@@ -39,17 +39,19 @@ func resourceAwsEMRContainersVirtualCluster() *schema.Resource {
 							Required: true,
 							ForceNew: true,
 						},
+						// According to https://docs.aws.amazon.com/emr-on-eks/latest/APIReference/API_ContainerProvider.html
+						// The info and the eks_info are optional but the API raises ValidationException without the fields
 						"info": {
 							Type:     schema.TypeList,
 							MaxItems: 1,
-							Optional: true,
+							Required: true,
 							ForceNew: true,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"eks_info": {
 										Type:     schema.TypeList,
 										MaxItems: 1,
-										Optional: true,
+										Required: true,
 										ForceNew: true,
 										Elem: &schema.Resource{
 											Schema: map[string]*schema.Schema{
--- a/aws/resource_aws_emrcontainers_virtual_cluster.go
+++ b/aws/resource_aws_emrcontainers_virtual_cluster.go
@@ -81,10 +81,6 @@ func resourceAwsEMRContainersVirtualCluster() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-			"id": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
 			"name": {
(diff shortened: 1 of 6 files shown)


aws/resource_aws_emrcontainers_virtual_cluster.go
```diff
@@ -0,0 +1,385 @@
...
+							ValidateFunc: validation.StringInSlice(emrcontainers.ContainerProviderType_Values(), false),
+						},
+					},
+				},
+			},
+			"created_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"id": {
```
DrFaust92, line 84: Not needed as TF always has a top level id attribute

aws/resource_aws_emrcontainers_virtual_cluster.go
```diff
@@ -0,0 +1,385 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"name": {
```
DrFaust92, line 88: Lets add validation to `name` based on docs: https://docs.aws.amazon.com/emr-on-eks/latest/APIReference/API_CreateVirtualCluster.html#API_CreateVirtualCluster_RequestSyntax

Len is between 1 and 64 
and must conform to the following regex: `[\.\-_/#A-Za-z0-9]+`

aws/resource_aws_emrcontainers_virtual_cluster.go
```diff
@@ -0,0 +1,385 @@
...
+		Name: aws.String(d.Get("name").(string)),
+	}
+
+	log.Printf("[INFO] Creating EMR containers virtual cluster: %s", input)
+	out, err := conn.CreateVirtualCluster(&input)
+	if err != nil {
+		return fmt.Errorf("error creating EMR containers virtual cluster: %w", err)
+	}
+
+	if _, err := waiter.VirtualClusterCreated(conn, aws.StringValue(out.Id)); err != nil {
```
DrFaust92, line 115: Missing SetId to the cluster id

DrFaust92 (changes requested):

aws/resource_aws_emrcontainers_virtual_cluster.go
shuheiktgw: Didn't know that! Thank you!

shuheiktgw:
Hmm, while writing acc tests, I noticed those two steps are not possible through the current terraform-aws-provider:

- https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/setting-up-cluster-access.html
- https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/setting-up-enable-IAM.html

1. We cannot create Kubernetes resources from terraform-aws-provider
2. terraform-aws-provider still hasn't support EKS addons, which is necessary for IAM Roles for service accounts (https://github.com/hashicorp/terraform-provider-aws/issues/16542)

The second problem will be solved eventually but I'm not sure about how I can solve the first one... @DrFaust92 Do you have any ideas? 🙏

DrFaust92:
Hey @shuheiktgw, do acc tests fail without those prerequisites? I don’t how they block creating the resource and using additional resources (and providers) to add access outside the scope of the tests

shuheiktgw:
@DrFaust92 Yes without them, the acc test fails with the following error:

```
make testacc TESTARGS='-run=TestAccAwsEMRContainersVirtualCluster_basic'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAwsEMRContainersVirtualCluster_basic -timeout 120m
=== RUN   TestAccAwsEMRContainersVirtualCluster_basic
=== PAUSE TestAccAwsEMRContainersVirtualCluster_basic
=== CONT  TestAccAwsEMRContainersVirtualCluster_basic
    resource_aws_emrcontainers_virtual_cluster_test.go:65: Step 1/2 error: Error running apply:
        Error: error creating EMR containers virtual cluster: ValidationException: Unauthorized to perform read namespace on default
```

I've looked into eksctl's PR to enable EMR containers, and they seem to create Kubernetes resources...

- https://github.com/weaveworks/eksctl/pull/2913

shuheiktgw:
@bflad @ewbankkit Do you people know how I can write an acc test that requires provisioning Kubernetes resources?

bflad:
@shuheiktgw if it does not work by just including `provider "kubernetes" {}` in the configuration, it may be necessary to also use [`helper/resource.TestCase` type `ExternalProviders` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#TestCase), however we do not currently have any examples of multiple provider testing in this codebase anymore. Attempting to configure the Kubernetes provider using values from other resources may also require two `TestStep`.

The maintainers are very behind on the large backlog of existing pull requests, so unfortunately we cannot provide more guidance ourselves or guarantee when this will be reviewed.

shuheiktgw:
Thank you @bflad! With that information, I think I can start writing the acc tests 👍

anggao:
@shuheiktgw do you have any updates on this PR :) looking forward to use it

ewbankkit:
Superseded by https://github.com/hashicorp/terraform-provider-aws/pull/20003.

Issue #8302, opened 2019-04-12
Feature Request: Allow data sources searching tags/filters to return empty results


lmickh:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

There are use-cases for data sources such as `aws_vpcs` or `aws_security_groups` that do not return results based on the tags/filter.  If there are not results, these data sources will error out during the planning stage.  Having a flag that allowed these data sources to return empty lists could expand their usage to search for resources that might not exist yet, but could in future runs.

### New or Affected Resource(s)

* data.aws_vpcs
* data.aws_security_groups
* probably could be expanded to more data sources

### Potential Terraform Configuration

```hcl
data "aws_security_groups" "fake" {
  allow_empty_results = true
  tags = {
    is_fake = "true"
  }
}

data "aws_vpcs" "fake" {
  allow_empty_results = true
  filter {
    name = "vpc-id"
    values = ["vpc-fake"]
  }
}

resource "aws_security_group" "this" {
  name        = "example-fake-sg-rule"
  description = "Allow inbound traffic for ssh"
  vpc_id      = "${var.vpc_id}"

  ingress {
    from_port       = 22
    to_port         = 22
    protocol        = "tcp"
    security_groups = ["${concat(var.ingress_security_groups, data.aws_security_groups.fake.ids)}"]
  }
}
```

lmickh:
Looks like https://github.com/hashicorp/terraform/issues/16380 is a hold up for implementing this.

stephenl03:
This issue should be addressed. If I use the the AWS SDK, I get an empty list returned, versus 404 or empty response. Something is returned and terraform should respect that response instead of generating an error. Coding standards have progressed along with the AWS API. This should not throw an error.

```
aws ec2 describe-instances --profile test --region ap-northeast-2 --filters Name=instance-state-name,Values=running --query 'Reservations[*].Instances[*][].InstanceId'
[]

aws ec2 describe-instances --profile test --region us-east-1 --filters Name=instance-state-name,Values=running --query 'Reservations[*].Instances[*][].InstanceId'
[
[...]

PR #17382, opened 2021-02-01
Fix data source aws_vpc_peering_connections

borancar:
The data source errors out if there are no connections to return, but
should instead return an empty array and let the user decide.

### Community Note

Relates https://github.com/hashicorp/terraform-provider-aws/issues/17540
Relates https://github.com/hashicorp/terraform-provider-aws/issues/8302

```

...
```


ewbankkit:
Relates:

* https://github.com/hashicorp/terraform-provider-aws/issues/8302

aws/data_source_aws_vpc_peering_connections.go
```diff
@@ -46,9 +44,6 @@ func dataSourceAwsVpcPeeringConnectionsRead(d *schema.ResourceData, meta interfa
 	if err != nil {
 		return err
 	}
-	if resp == nil || len(resp.VpcPeeringConnections) == 0 {
-		return fmt.Errorf("no matching VPC peering connections found")
-	}
```
bflad, line 51: To prevent potential panics, we should still check for `nil`:

```suggestion
	if resp == nil
		return fmt.Errorf("error reading EC2 VPC Peering Connections: empty response")
	}
```

borancar: Thanks! Done.

anGie44 (approved):
Thank you for this fix @borancar, LGTM 🚀 I'll follow up with an acceptance test for this use-case

PR #17468, opened 2021-02-04
Clarify that overriding policy document statements can be added to current document if `sid` does not match

aperiodic:
In the description of `iam_policy_document` data source's
`override_json` argument, state specifically that any statements in the
override document that have `sid`s that do _not_ match a statement in
the current document will be added to the current document, instead of
ignored. Before this change, there was no description of how
non-matching statements in the `override_json` were used, so the
behavior was unclear: the opposite and incorrect conclusion that these
statements were ignored was equally consistent with the documentation.

This behavior is covered by an acceptance test, so from the coverage I'm
inferring that this behavior is both intended and supported.

### Community Note

I am omitting the output from acceptance testing, because this is a documentation-only change.

### DIFF
--- a/website/docs/d/iam_policy_document.html.markdown
+++ b/website/docs/d/iam_policy_document.html.markdown
@@ -78,104 +78,6 @@ resource "aws_iam_policy" "example" {
 ### Example Assume-Role Policy with Multiple Principals
 
 You can specify multiple principal blocks with different types. You can also use this data source to generate an assume-role policy.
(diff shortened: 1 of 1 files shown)


YakDriver:
@aperiodic Thank you for noticing this hole in the documentation and taking the initiative to fill it! 🎉 This builds on a fairly extensive overhaul of the `aws_iam_policy_document` documentation in #12055.

PR #17650, opened 2021-02-17
Sweeper: exclude replication group member clusters from aws_elasticache_cluster sweeper

### LINKED ISSUE #17647: Sweeper: aws_elasticache_cluster does not correctly handle Cluster that is last member of Replication Group
The `aws_elasticache_cluster` sweeper depends on the `aws_elasticache_replication_group` sweeper. If the `aws_elasticache_replication_group` fails for some reason, the member clusters still exists and will be picked up by the `aws_elasticache_cluster` sweeper.

If the Cluster sweeper tries to delete the last member Cluster in a Replication Group, it receives the following error:

```
InvalidCacheClusterState: Cannot delete the cache cluster since it is the only member of a replication group. Delete the replication group instead.
```

Aside from one particular case, the `InvalidCacheClusterState` error is retried for 5 minutes.

### Affected Resource

* aws_elasticache_cluster

gdavison:
Prevents retry when trying to delete the last member cluster from a replication group.

Excludes replication group member clusters from sweeper.

Closes #17647

### DIFF
--- a/aws/resource_aws_elasticache_cluster_test.go
+++ b/aws/resource_aws_elasticache_cluster_test.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/elasticache"
+	"github.com/hashicorp/aws-sdk-go-base/tfawserr"
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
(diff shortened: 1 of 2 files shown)


aws/resource_aws_elasticache_cluster_test.go
```diff
@@ -47,24 +54,31 @@ func testSweepElasticacheClusters(region string) error {
 
 			log.Printf("[INFO] Deleting ElastiCache Cluster: %s", id)
 			err := deleteElasticacheCacheCluster(conn, id, "")
+			if tfawserr.ErrMessageContains(err, elasticache.ErrCodeInvalidCacheClusterStateFault, "only member of a replication group") {
+				log.Printf("[WARN] ElastiCache Cache Cluster (%s) is member of a Replication Group. Ignoring. %s", id, err)
+				continue
+			}
```
anGie44, line 60: qq: if we're filtering these specific clusters out in the `DescribeCacheClustersInput` , is it still possible to hit this type of error? doesn't hurt to catch the error tho 👍

anGie44 (approved):
Looks great! just one small q more for my own curiosity

aws/resource_aws_elasticache_cluster_test.go
gdavison: Good catch, it's redundant since they're filtered out from the results

anGie44: gotchu, sounds good! tho i'm always cautious with AWS APIs 😅

PR #17681, opened 2021-02-18
Fix typo in IP range

cdodd:
Fix a typo in the private class B address range on the EKS `service_ipv4_cidr` configuration.


anGie44 (approved):
Whoa great catch @cdodd! interstingly enough, the typo also appears in the [AWS doc page](https://docs.aws.amazon.com/eks/latest/APIReference/API_KubernetesNetworkConfigRequest.html) as well 😅  Thank you for the fixes 🚀

Issue #17685, opened 2021-02-18
Go 1.16 Upgrade Followup Items


bflad:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Quick checklist from [Go 1.16 release notes](https://golang.org/doc/go1.16) and any preferred changes in the codebase:

- [x] Remove manual `GO111MODULE` environment variable handling from `.github/workflows/*.yml`
- [x] Remove manual `GO111MODULE` environment variable handling from `GNUmakefile`
- [x] Remove `darwin/arm64` build ignore from `.goreleaser.yml`
- [x] Enable `darwin/arm64` releases in TeamCity
- [ ] Refactor off deprecated `io/ioutil` package

### References

- https://github.com/hashicorp/terraform-provider-aws/issues/16948
- https://github.com/hashicorp/terraform-provider-aws/pull/17655
- https://github.com/hashicorp/terraform-provider-aws/pull/17684

PR #17795, opened 2021-02-24
provider: Development and testing updates for Go 1.16 upgrade, proper impi handling in tools Go module, and update golangci-lint@v1.37.1

bflad:
### Community Note

Reference: https://github.com/hashicorp/terraform-provider-aws/pull/17743
Reference: https://github.com/hashicorp/terraform-provider-aws/issues/17685
Reference: https://golang.org/doc/go1.16

This change request bundles a few related items around the Go 1.16 upgrade and changes to the `tools` Go module:

* Removes `GO111MODULE=on` environment variable handling in GitHub Actions and GNUmakefile as that is now the default behavior in Go 1.16
* Adds missing `impi` into `tools/main.go` so it shows as needing to be versioned in the `tools` Go module
* Upgrades `golangci-lint` in `tools` since dependabot has been having trouble with this change for some reason
* Runs `go mod tidy` in `tools` with Go 1.16, which is the majority of the automatic updates in `tools/go.sum`

Now with Go 1.16, we can use `go install PACKAGE@VERSION` if we no longer want to use the `tools` Go module strategy, however that will also remove dependabot pinging us on tooling updates.

Output from acceptance testing: N/A (CI)


YakDriver (approved):
Looks good! 🎉 

Ran some spot tests (acc test, `make tools`, etc.) and all runs as expected.

PR #17808, opened 2021-02-24
deps: Go 1.16 go mod tidy

bflad:
### Community Note

It looks like dependabot was not using Go 1.16 during its last update, so it included additional checksums that Go 1.16 removes. Will pull in this `go mod tidy` and see if the problem persists next time.

Output from acceptance testing: N/A (dependencies)


bflad:
Dependency check is "expected" to fail -- its using `pull_request_target`. Self-merging this fix so pull requests don't all start failing.

PR #17849, opened 2021-02-26
tests/ec2/ami: Rename tests

YakDriver:
### Community Note

Relates #17847 

```
$ make testacc TESTARGS='-run=TestAccEC2AMI'

...
```

### DIFF
--- a/aws/data_source_aws_ami_ids_test.go
+++ b/aws/data_source_aws_ami_ids_test.go
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
-func TestAccDataSourceAwsAmiIds_basic(t *testing.T) {
+func TestAccEC2AMIIDsDataSource_basic(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
@@ -22,7 +22,7 @@ func TestAccDataSourceAwsAmiIds_basic(t *testing.T) {
 	})
 }
 
-func TestAccDataSourceAwsAmiIds_sorted(t *testing.T) {
+func TestAccEC2AMIIDsDataSource_sorted(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
--- a/aws/resource_aws_ami_copy_test.go
+++ b/aws/resource_aws_ami_copy_test.go
@@ -12,7 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-func TestAccAWSAMICopy_basic(t *testing.T) {
+func TestAccEC2AMICopy_basic(t *testing.T) {
 	var image ec2.Image
 	rName := acctest.RandomWithPrefix("tf-acc-test")
 	resourceName := "aws_ami_copy.test"
@@ -39,7 +39,7 @@ func TestAccAWSAMICopy_basic(t *testing.T) {
 	})
 }
 
-func TestAccAWSAMICopy_Description(t *testing.T) {
+func TestAccEC2AMICopy_Description(t *testing.T) {
 	var image ec2.Image
 	rName := acctest.RandomWithPrefix("tf-acc-test")
 	resourceName := "aws_ami_copy.test"
@@ -67,7 +67,7 @@ func TestAccAWSAMICopy_Description(t *testing.T) {
 	})
 }
 
-func TestAccAWSAMICopy_EnaSupport(t *testing.T) {
+func TestAccEC2AMICopy_EnaSupport(t *testing.T) {
 	var image ec2.Image
 	rName := acctest.RandomWithPrefix("tf-acc-test")
 	resourceName := "aws_ami_copy.test"
@@ -88,7 +88,7 @@ func TestAccAWSAMICopy_EnaSupport(t *testing.T) {
 	})
 }
 
-func TestAccAWSAMICopy_tags(t *testing.T) {
+func TestAccEC2AMICopy_tags(t *testing.T) {
 	var ami ec2.Image
 	resourceName := "aws_ami_copy.test"
 	rName := acctest.RandomWithPrefix("tf-acc-test")
--- a/aws/data_source_aws_ami_test.go
+++ b/aws/data_source_aws_ami_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-func TestAccAWSAmiDataSource_natInstance(t *testing.T) {
+func TestAccEC2AMIDataSource_natInstance(t *testing.T) {
 	resourceName := "data.aws_ami.nat_ami"
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
@@ -60,7 +60,7 @@ func TestAccAWSAmiDataSource_natInstance(t *testing.T) {
 	})
 }
 
-func TestAccAWSAmiDataSource_windowsInstance(t *testing.T) {
+func TestAccEC2AMIDataSource_windowsInstance(t *testing.T) {
 	resourceName := "data.aws_ami.windows_ami"
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
@@ -103,7 +103,7 @@ func TestAccAWSAmiDataSource_windowsInstance(t *testing.T) {
 	})
 }
 
-func TestAccAWSAmiDataSource_instanceStore(t *testing.T) {
+func TestAccEC2AMIDataSource_instanceStore(t *testing.T) {
 	resourceName := "data.aws_ami.amzn-ami-minimal-hvm-instance-store"
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
@@ -142,7 +142,7 @@ func TestAccAWSAmiDataSource_instanceStore(t *testing.T) {
 	})
 }
 
-func TestAccAWSAmiDataSource_localNameFilter(t *testing.T) {
+func TestAccEC2AMIDataSource_localNameFilter(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:  func() { testAccPreCheck(t) },
 		Providers: testAccProviders,
@@ -158,7 +158,7 @@ func TestAccAWSAmiDataSource_localNameFilter(t *testing.T) {
 	})
 }
 
-func TestAccAWSAmiDataSource_Gp3BlockDevice(t *testing.T) {
+func TestAccEC2AMIDataSource_Gp3BlockDevice(t *testing.T) {
 	resourceName := "aws_ami.test"
 	datasourceName := "data.aws_ami.test"
 	rName := acctest.RandomWithPrefix("tf-acc-test")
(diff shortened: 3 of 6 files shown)


aws/resource_aws_ami_copy_test.go
```diff
@@ -39,7 +39,7 @@ func TestAccAWSAMICopy_basic(t *testing.T) {
 	})
 }
 
-func TestAccAWSAMICopy_Description(t *testing.T) {
+func TestAccEC2AMICopy_description(t *testing.T) {
```
bflad, line 42: The [current guidance for writing acceptance tests](https://github.com/hashicorp/terraform-provider-aws/blob/main/docs/contributing/running-and-writing-acceptance-tests.md#per-attribute-acceptance-tests) uses lowercase for "meta" tests like `_basic` and `_disappears`, but uppercase for attribute-based and other tests. Not sure if we should be following the documentation or updating the documentation to include more explicit guidance.

aws/data_source_aws_ami_ids_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
-func TestAccDataSourceAwsAmiIds_basic(t *testing.T) {
+func TestAccEC2AMIIDsDataSource_basic(t *testing.T) {
```
bflad, line 10: What determines whether test naming should include a fully capitalized initialism? Should we enforce the standard?

aws/resource_aws_ami_copy_test.go
YakDriver: While Go conventions [do allow for capitalized words](https://golang.org/pkg/testing/#hdr-Examples) after the underscore, they apply to types and methods on types. Drawing the analogy to what we're doing, it would be reasonable to apply the same logic. However, I do not think we should have capitalized words after underscores for one simple reason: a badly followed convention is worse than no convention at all. Currently, the capitalized word after an underscore is as likely to tell you nothing about the test as something. As a result, a clear, followable convention would be better. I vote for updating the documentation.

aws/data_source_aws_ami_ids_test.go
YakDriver: Unlike above, the [Go convention](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) is clear on this point:

> Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url".

Although the AWS SDK for Go breaks this rule a lot, I have seen efforts to change course and some of the newer APIs follow the rules. My vote is that we stick with convention and win the respect of Go coders.

bflad: If we are going to enforce this, we will need to manually enumerate all initialisms and maintain the list going forward. For example, what does `sesv2` convert into?

I would prefer to see a proposal on that before really going down this route since it affects all code and testing contributions, otherwise it will be easy to escape out of the desired convention. I'm guessing it would need to be a text file with a list or a Go file with a slice of strings, but we will need to figure out how to ensure it is usable by tooling that needs it.

For what it is worth, not using initialisms allows you to easily convert snake_case to PascalCase and lowercase to Titlecase without additional logic to also bake in the conversions. Not saying that needs to be an end-all goal, but I'm not sure if folks are really chasing after this project's code for "correctness" in this sense. More likely folks are probably interested in functionality that is readable and works.

aws/resource_aws_ami_copy_test.go
bflad: > Currently, the capitalized word after an underscore is as likely to tell you nothing about the test as something.

Can you please expand on this?

aws/data_source_aws_ami_ids_test.go
YakDriver: I'm specifically limiting the scope of changes to test names. We have non-breaking freedom with test names. However, fixing things in test names but not code introduces a level of inconsistency. For instance, removing "AWS" and "Aws" from test names but not all the resource names. Despite this issue, which I think is small, I love the idea of having one area where we have clean, Go convention adhereing names.

Following Go's conventions, the "V" would either be the next word and capitalized or an initial and capitalized:

package | testing name
-- | --
apigatewayv2 | APIGatewayV2
cloudhsmv2 | CloudHSMV2
elbv2 | ELBV2
greengrassv2 | GreengrassV2
kinesisanalyticsv2 | KinesisAnalyticsV2
lexmodelsv2 | LexModelsV2
lexruntimev2 | LexRuntimeV2
sesv2 | SESV2
wafv2 | WAFV2

aws/resource_aws_ami_copy_test.go
YakDriver: Currently, there are around about 1256 `basic` and `disappears` tests, which use the underscore-lowercase "meta" test convention. There are a further 1904 tests that use underscore-lowercase but many do not seem like "meta" tests:

```go
TestAccAWSAPIGatewayUsagePlan_description()
TestAccAWSAPIGatewayMethod_customrequestvalidator()
TestAccAWSAPIGatewayIntegration_integrationType()
TestAccAWSEBSSnapshot_tags()
TestAccAWSEBSVolume_updateSize()
TestAccAWSEIP_instance()
TestAccAWSGlueTrigger_actions_notify()
TestAccAWSInstance_outpost()
```

aws/data_source_aws_ami_ids_test.go
gdavison: I like the `<service>V2` approach in general. However, it doesn't look right after initialisms, such as `WAFV2`. For straight up consistency, I'd go with `<service>V2`. For ease of readability or aesthetics, I'd go with `<full word service>V2` and `<initialism service>v2`.

gdavison: We can extract the list of initialisms from the SDK packages at `service/<service package>/service.go` files. The constant `ServiceID` could be checked for all-caps. There are some special cases, such as `ACM PCA`, where there's a space.

gdavison: This also somewhat relates to another cleanup issue I'd like to open around standardizing service and resource naming in, e.g. error messages

aws/resource_aws_ami_copy_test.go
gdavison: We should enforce whatever standard we choose. In this case, I think we should pick a standard that improves understandability/readability first, and then update whatever documentation and Semgrep rules to enforce it

anGie44 (commented):
my view on topics covered:

(1) dropping `Aws/AWS` -- YES 🥳 .

(2) initialisms -- also leaning towards 🐫 Case (e.g. `WafV2RuleGroup`) as it seems initialisms are not uniformly defined w/in AWS and in the real-world. also maintaining a list of initialisms can prove tricky/subjective and easy to miss adding new ones to the docs when introduced in PRs (e.g. maintaining our list of Environment variables)

(3) test suffixes (e.g. `_Basic`, `ContentHandling`) -- maintain 🐫 Case; imo, case doesn't map to semantics too intuitively and guidance from our documentation can be easily overlooked; no preference on whether case after underscore is lower (`_contentHandling`) vs. upper (`_ContentHandling`)

gdavison:
One other comment on standardization: Where do we want to move `DataSource` in acceptance test naming? For test name pattern matching, it's sometimes helpful to have it at the beginning if you want to test only the resource (`-run=TestAccAWSMskCluster_` or `-run=TestAccDataSourceAWSMskCluster_`), and easy enough to include for resource and datasource (`-run=TestAcc(DataSource)?AWSMskCluster_`. This was one case where having `AWS` helped with `-run=TestAcc.*AWSMskCluster_`)

YakDriver:
> Where do we want to move DataSource in acceptance test naming?

@bflad suggests, and I heartily agree, that we should move "DataSource" to the end of the first clause, i.e., just before the underscore. Most of the time, the default should be to run both the resource and data source when changing either since there can be dependencies. This fascilitates that. We should be able to still run only one or the other with `TestAccExample*DataSource` and `TestAccExample*[^DataSource]`.

YakDriver:
Re: @anGie44 

> (2) initialisms -- also leaning towards 🐫 Case (e.g. WafV2RuleGroup) as it seems initialisms are not uniformly defined w/in AWS and in the real-world. also maintaining a list of initialisms can prove tricky/subjective and easy to miss adding new ones to the docs when introduced in PRs (e.g. maintaining our list of Environment variables)

I'm don't think that _not_ capitalizing acronyms and initialisms will get rid of all lists. For example, if we decide that `WafV2` is correct, how would automation fix `Wafv2`, `WAFv2`, and `WAFV2` without knowing that `Waf` is a word?

There are about 41 acronyms in common AWS use. We only use some of those in the AWS provider. I don't think it's an unreasonable burden to maintain a list. We will need to maintain a list anyway to fix mixed case as well.

Also, by adhereing to Go's well-established convention, we don't have to explain special, non-conforming Go naming rules. We certainly can and should explain that we follow convention but Go developers will expect it. Many contributors have been trying to adhere for years. Probably some of them forget that we're special.

anGie44:
> I don't think it's an unreasonable burden to maintain a list

I think you're right here 👍  i thought it was a bit rare in golang projects at first but totally overlooked what current linters already do 😅  e.g. https://github.com/golang/lint/blob/master/lint.go#L767-L809. Which then makes me wonder, since we already get those for free from a linter like golint which _could_ be enabled in our `.golangci.yml` file, could we just build off of that list i.e. just keep track of the `AWS` service initialisms. Or enabling that linter creates more warnings than we need to get this proposal started? Additionally, users (via their PRs: https://github.com/golang/lint/pulls) in the go community seem to be wanting to add to that initialisms list. So do we follow a similar pattern, say if `DB` is added, do we also update our tests when the time comes? 

So let's say we go with the idea of having a list of initialisms/acronyms we see fit, i still have a couple Q's:

1. what do they look like in code (e.g. Go enumerations)? 
2. where will those initialisms live? 
3. what do our new naming conventions/mappings look like e.g. mapping of `waf` -> `WAF`? how do we enforce the mapping when changing pre-existing tests vs. new tests?
4. What do these initialisms apply to? only test functions for the moment? what about initialisms used in the resource and datasource files?
5. how are we going to enforce whichever rules we set out? are we set on say using `semgrep` or another linter tool mentioned in the original issue, and how does that look like (can be a brief example).  seems like that could also be a contributing factor as well if one of the end goals is to use the 🤖 🧠 to do the chore of updating tests in bulk where needed. 

p.s. thank you for taking this on @YakDriver and taking this naming convo from the less abstract to more tangible examples!

YakDriver:
@anGie44 Thanks, Angie! 🙏 🙏 

It would be awesome to just use industry-standard linters! Maybe as a first step we could just lint test files and then eventually move to the resource and data source files. I like the idea of just feeding an iinitialism/acronym list to the linter that the community can help maintain!

> 1. what do they [intialisms] look like in code (e.g. Go enumerations)?

The rules are the same. "Capitalisms" (my word for abbreviations, initialisms, acronyms, etc.) should be either all caps or all lowercase. 

```go
vpcCount := 0
namedVPC := "vpc"
```

We have a few complexities here that maybe are not exactly handled in the convention though, e.g., `IoT`.

>2. where will those initialisms live? (could be answered by 1.)

Somewhere in the repo, where we can feed them to the linter(s).

>3. what do our new naming conventions/mappings look like e.g. mapping of waf -> WAF?

Right now, I would expect to handle "WAF" the same as "EBS", "VPC", or "EC2".

>4. What do these initialisms apply to? only test functions for the moment? what about initialisms used in the resource and datasource files?

I think we start with tests. It's beyond the scope of this effort but we can turn on the the enforcement just for tests to get going. I have a script that can fix acceptance test names to avoid this taking much engineering time.

> how does that look

As I'm not sure which direction we're going or if we'll even make any initialism changes, I haven't spent much time on enforcement. I'd be happy to collaborate if you have ideas!

PR #9735, opened 2019-08-12
[WIP] New data sources 'aws_dx_location' and 'aws_dx_locations'

ewbankkit:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Closes https://github.com/terraform-providers/terraform-provider-aws/issues/9708.

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note
* **New Data Source:** `aws_dx_location`
* **New Data Source:** `aws_dx_locations`
```

Output from acceptance testing:

```console
$ make testacc TEST=./aws/ TESTARGS='-run=TestAccDataSourceAwsDxLocation_'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws/ -v -parallel 20 -run=TestAccDataSourceAwsDxLocation_ -timeout 120m
=== RUN   TestAccDataSourceAwsDxLocation_basic
=== PAUSE TestAccDataSourceAwsDxLocation_basic
=== CONT  TestAccDataSourceAwsDxLocation_basic
--- PASS: TestAccDataSourceAwsDxLocation_basic (21.62s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	21.700s
```

PR #9705, opened 2019-08-09
d/aws_dx_connection: Adding new data source for Direct Connect connection

robh007:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Changes proposed in this pull request:
Adds a new data resource for DX  connections: ``` aws_dx_connection ```

Lookup DX connections from a name.

example
```hcl
data "aws_dx_connection" "example" {
  name = "example"
}
```

* Adds data source aws_dx_connection

Output from acceptance testing:

```
$ DX_LOCATION=TCSH make testacc TEST=./aws/ TESTARGS='-run=TestAccDataSourceAwsDxConnection_Basic'                                                                                                   master * ] 6:09 pm
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws/ -v -parallel 20 -run=TestAccDataSourceAwsDxConnection_Basic -timeout 120m
=== RUN   TestAccDataSourceAwsDxConnection_Basic
=== PAUSE TestAccDataSourceAwsDxConnection_Basic
=== CONT  TestAccDataSourceAwsDxConnection_Basic
--- PASS: TestAccDataSourceAwsDxConnection_Basic (24.28s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       24.326s

...
```

PR #17852, opened 2021-02-27
New Data Source: aws_dx_connection

yzguy:
### Community Note

Relates OR Closes #0000

```
-> make testacc TESTARGS='-run=TestAccDataSourceAwsDxConnection'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccDataSourceAwsDxConnection -timeout 120m
=== RUN   TestAccDataSourceAwsDxConnection_basic
=== PAUSE TestAccDataSourceAwsDxConnection_basic
=== CONT  TestAccDataSourceAwsDxConnection_basic
--- PASS: TestAccDataSourceAwsDxConnection_basic (23.06s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	25.404s
```


ewbankkit:
Similar:

* https://github.com/hashicorp/terraform-provider-aws/pull/9705

ewbankkit:
@yzguy Thanks for the contribution 👏.
There is a stale PR #9705 that adds the same data source.
Could you please add the additional attributes from that PR?

yzguy:
@ewbankkit Absolutely! Will update PR soon.

yzguy:
Couple things I believe I did right, but am unsure.

1. You can search by ID or Name. If you used both, it would search for the connection with the specific ID, then only continue if it found the one with the specified name. If it didn't it would error out.

2. It looks like AWS does not allow filtering on DirectConnect connections so I left this out, but I'm unsure if searching by tags would still be a good idea.

3. A couple of the fields (`vlan`, `partner_name`, `lag_id`) are not present in the JSON response from AWS all the time, and I believe depending on if it's a Hosted versus Dedicated connection they will show up, or if you have 2 or more connections in LAG. I don't have access to an environment that uses hosted connections, only dedicated. Therefore I have made them only added when they are present in the JSON.

ewbankkit:
@yzguy The better way here I think would be to add an additional data source `aws_dx_connections` where connections can be filtered by `name` or `tags`. This DS would output just a set of IDs. The `aws_dx_connection` would have the one required attribute `connection_id` and output all the other connection attributes.
You can look at the `aws_apigatewayv2_apis` and `aws_apigatewayv2_api` data sources for an example for how to approach this.
BTW you don't need the `!= nil` check for fields missing from the API response, `d.Set()` handles `nil` input.

Let me know if this is not clear or if you are unable to make progress.
Thanks.

yzguy:
```
-> make testacc TESTARGS='-run=TestAccDataSourceAwsDxConnections'
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccDataSourceAwsDxConnections -timeout 120m
=== RUN   TestAccDataSourceAwsDxConnections_Name
--- PASS: TestAccDataSourceAwsDxConnections_Name (23.60s)
=== RUN   TestAccDataSourceAwsDxConnections_Tags
--- PASS: TestAccDataSourceAwsDxConnections_Tags (19.98s)
PASS
ok  	github.com/terraform-providers/terraform-provider-aws/aws	45.549s
```

yzguy:
@ewbankkit I was able to get back to this after, completing the changes you suggested. Using the API Gateway v2 Datasource as a reference.

Couple things I'm not sure about are on `data.aws_dx_connection` there are a bunch of attributes that are not returned by `aws_dx_connection`, so I'm not checking they match in the test. I'm not sure best way to go about that, perhaps just checking that the attributes are set on the data source.

Other thing is that I pass in the Bandwidth and Locations in the tests, and that works, but I believe because both are `us-east-1` resources, and that's my default region that works fine, but we can't be 100% certain that the tests run in `us-east-1`, or that they would be forever. I believe maybe a better choice would be to have perhaps a map of locations per regions, then just randomly select one for whatever region the tests run it.

aws/data_source_aws_dx_connections.go
```diff
@@ -0,0 +1,65 @@
...
+
+func dataSourceAwsDxConnectionsRead(d *schema.ResourceData, meta interface{}) error {
+	conn := meta.(*AWSClient).dxconn
+	ignoreTagsConfig := meta.(*AWSClient).IgnoreTagsConfig
+
+	tagsToMatch := keyvaluetags.New(d.Get("tags").(map[string]interface{})).IgnoreAws().IgnoreConfig(ignoreTagsConfig)
+
+	connections, err := finder.Connections(conn, &directconnect.DescribeConnectionsInput{})
+	if err != nil {
+		fmt.Errorf("error reading DirectConnect connections: %w", err)
```
ewbankkit, line 41: Fix `golangci` linter error:

```suggestion
		return fmt.Errorf("error reading DirectConnect connections: %w", err)
```

ewbankkit:
@yzguy The location/bandwidth issue will be resolvable once https://github.com/hashicorp/terraform-provider-aws/pull/9735, `aws_dx_location(s)` data sources, has been merged.
Don't worry about checking for specific values of any of those attributes, just checking that they are set is sufficient.
There are also some formatting issues with the acceptance test configurations.
Thanks.

ewbankkit:
Closes #18400.

yzguy:
@ewbankkit Done!

PR #17900, opened 2021-03-03
Fix AWS DB Proxy target example to use correct name from default TG attribute

solarchad:
The example was using a bad output attribute from the default target group and when followed would lead to errors like:

```
Error: error registering RDS DB Proxy ($db_proxy_name/$db_proxy_target_group_name) Target: DBProxyTargetGroupNotFoundFault: Requested resource not found - TargetGroup: $db_proxy_target_group_name
```

We can see in the [docs for the aws_db_proxy_default_target_group
 resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_proxy_default_target_group), that `db_proxy_name` is not a valid attribute.

I wasn't sure if a changelog entry was necessary for small updates to documentation examples but if needed, just let me know.

### Community Note


breathingdust:
Validated that the change to example fixes the issue. In future we should be able to automate validation of examples.

LGTM! 🚀 Thanks @solarchad!

PR #17169, opened 2021-01-19
Add retry when publishing lambda functions that uses images

mauriciocm9:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request

Closes #17153

Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note

```

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccAWSLambdaFunction_imageConfig'


```

PR #18109, opened 2021-03-15
Increase Lambda function publish timeout

skorfmann:
### Community Note

Publishing a Lambda function based on a large Docker image can take longer than 2 minutes. It's a bit more than 2 minutes for an image of ~ 2 GB. Lambda supports images with a size of up to 10 GB. I haven't tried if 5 minutes are really sufficient for the max image size.

Relates #17169


skorfmann:
The actual image size I'm using is about 5 GB - the ~ 2 GB is the packed size in the ECR registry. I think that makes the 5 minutes timeout reasonable

PR #18141, opened 2021-03-17
Fix athena database read

harsimranmaan:
The current athena database implements a query on read. This means that the terraform plan needs
the permission to execute queries against athena as well as permission to write to the
s3 bucket where the execution results are stored. This makes it hard to run plan in an
environment where we don't want it to modify any aws resources. The StartQueryExecution
also does not have a way to restrict the queries. This means that it is possible to use
the permissions set up for a terraform plan to run any queries against athena.

The fix here is to read the database name from the AwsDataCatalog. In the future, this can
be extended to include other catalogs as the support for creating athena catalogs is added to
the provider

### Community Note

Relates OR Closes #0000

```

...
```


harsimranmaan:
Hi @bflad, Can you take a look when you get a chance. Thanks

ewbankkit:
Replaced by #19765.

PR #18326, opened 2021-03-22
d/aws_glue_catalog_database-database-naming-convention

joelpinheiro:
When using Athena with the AWS Glue Data Catalog, we can use AWS Glue to create databases and tables (schema) to be queried in Athena, or we can use Athena to create schema and then use them in AWS Glue and related services. 

Under the hood, Athena uses Presto to process DML statements and Hive to process the DDL statements that create and modify schema. With these technologies, there are a couple of conventions to follow so that Athena and AWS Glue work well together.

One of the conventions is that the the only acceptable characters for database names are lowercase letters, numbers, and the underscore character ([source](https://docs.aws.amazon.com/athena/latest/ug/glue-best-practices.html#schema-names)). Hence the PR.

### Community Note

### DIFF
--- a/.changelog/18320.txt
+++ b/.changelog/18320.txt
@@ -1,3 +1,0 @@
-```release-notes:enhancement
-resource/aws_glue_catalog_database: Add `glue catalog database` naming convention.
-```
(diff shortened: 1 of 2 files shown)


gdavison (approved):
Looks good 🚀

PR #18436, opened 2021-03-26
Fix the `make tools` command

gdavison:
The command `make tools` currently fails due to a publishing error with github.com/hashicorp/go-getter v1.5.1.

```console
$ make tools

cd awsproviderlint && go install .
cd tools && go install github.com/bflad/tfproviderdocs
cd tools && go install github.com/client9/misspell/cmd/misspell
cd tools && go install github.com/golangci/golangci-lint/cmd/golangci-lint
cd tools && go install github.com/katbyte/terrafmt
verifying github.com/hashicorp/go-getter@v1.5.1: checksum mismatch
	downloaded: h1:LZ49OxqBBtdKJymlpX7oTyqGBQRg4xxQDyPW4hzoZqM=
	go.sum:     h1:lM9sM02nvEApQGFgkXxWbhfqtyN+AyhQmi+MaMdBDOI=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
make: *** [tools] Error 1
```

The [v1.5.2 release of github.com/hashicorp/go-getter](https://github.com/hashicorp/go-getter/releases/tag/v1.5.2) states:

> The goreleaser release pipeline that was created and used for the 1.5.1 release was missing the GOPRIVATE env var being set, which caused private modules to be skipped. This release contains the same code as 1.5.1, but with GOPRIVATE set to github.com/hashicorp

Add an `exclude` directive to `tools/go.mod` to skip v.1.5.1


gdavison:
I'm actually a little curious why the GitHub Actions aren't failing. We call `cd tools && go install github.com/katbyte/terrafmt` in four workflows, and they're not failing, even though they fail locally.

PR #18571, opened 2021-04-05
docs: Add AWS_SDK_LOAD_CONFIG to env vars section

justincampbell:
I was unable to get the AWS provider to load the region, key ID, and secret from env vars on a new system. Setting `AWS_SDK_LOAD_CONFIG=1` resolved this, as suggested in [hashicorp/terraform-provider-aws#13057](https://github.com/hashicorp/terraform-provider-aws/issues/13057#issuecomment-620612604).

### Community Note

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -87,12 +87,14 @@ provider "aws" {
 ### Environment Variables
 
 You can provide your credentials via the `AWS_ACCESS_KEY_ID` and
-`AWS_SECRET_ACCESS_KEY`, environment variables, representing your AWS
-Access Key and AWS Secret Key, respectively.  Note that setting your
-AWS credentials using either these (or legacy) environment variables
-will override the use of `AWS_SHARED_CREDENTIALS_FILE` and `AWS_PROFILE`.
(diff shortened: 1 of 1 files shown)


bflad:
Hey @justincampbell 👋  Good to see you and thank you for submitting this!

Since version 3.0.0 of the Terraform AWS Provider, the AWS SDK Go configuration has that option automatically enabled here: https://github.com/hashicorp/aws-sdk-go-base/blob/57529b4c2d2f8f3b5299d66a829b01259fa800d7/session.go#L46 so the environment variable should have no effect with the provider.

I'm personally hesitant to re-introduce the documentation for this environment variable because it should not be necessary any more and could cause confusion. If something is not working as expected without the `AWS_SDK_LOAD_CONFIG` environment variable on 3.x versions and later, we should suspect there is a bug somewhere or a potentially different issue occurring (e.g. the Terraform S3 Backend also needs to be Terraform CLI 0.13.0 or higher (IIRC) to also automatically load the AWS shared configuration). Please feel free to submit a full bug report and maybe we can narrow it down. 👍

justincampbell:
@bflad Hey! Thanks for taking a look.

I'm on AWS provider version 3.35.0, TF 14.9.0. I'll try to reproduce and open a separate issue. Thanks!

Issue #15334, opened 2020-09-24
tests/provider: Add concurrent sweepers


YakDriver:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When running a test suite, sweepers run before to make sure the environment is clean. However, these sweepers can take a very long time to run. Adding concurrency, especially to long-running sweepers, is likely to improve sweeper performance.

### Documentation

This is [the documentation](https://github.com/hashicorp/terraform-provider-aws/blob/main/docs/contributing/running-and-writing-acceptance-tests.md#writing-test-sweepers) for writing and/or updating sweepers in the tests. Also, refer to the PRs below for examples.

### Top 10 Longest Tests

 - [x] resource_aws_waf_web_acl_test.go (#15359)
 - [x] resource_aws_elasticsearch_domain_test.go (#18574)
 - [x] resource_aws_eks_node_group_test.go (#18575)
 - [x] resource_aws_redshift_cluster_test.go (#18576)
 - [x] resource_aws_elasticache_replication_group_test.go (#18587)
 - [x] resource_aws_redshift_snapshot_schedule_test.go (#18592)
 - [x] resource_aws_db_instance_test.go (#15336)
 - [ ] resource_aws_dms_replication_instance_test.go (#18623)
 - [ ] resource_aws_fsx_windows_file_system_test.go (#18626)
 - [ ] resource_aws_waf_rule_group_test.go (#18631)

### 6 Sweepers Running the Longest, April 7, 2021

- [ ] aws_instance (#18652)
- [x] aws_elasticache_replication_group (#18587)
- [ ] aws_subnet (#18653)
- [ ] aws_ecrpublic_repository (#18656)
- [ ] aws_security_group (#18657)
- [ ] aws_vpc (#18658)

### Affected Resources

 - [ ] resource_aws_acm_certificate_test.go
 - [ ] resource_aws_acmpca_certificate_authority_test.go
 - [ ] resource_aws_api_gateway_rest_api_test.go
 - [ ] resource_aws_api_gateway_vpc_link_test.go
[...]

PR #18657, opened 2021-04-08
tests/r/security_group: Add sweeper concurrency

YakDriver:
### Community Note

Relates #15334

Output from acceptance testing: NA

```
2021/04/08 00:16:05 [DEBUG] Security Group destroy: sg-0c83901c9e757f55d
2021/04/08 00:16:05 [DEBUG] Security Group destroy: sg-09aafd52d69555d9e
2021/04/08 00:16:06 [DEBUG] Waiting for state to become: [success]
2021/04/08 00:16:06 [DEBUG] Waiting for state to become: [success]
```

### DIFF
--- a/aws/resource_aws_security_group_test.go
+++ b/aws/resource_aws_security_group_test.go
@@ -73,11 +73,8 @@ func testSweepSecurityGroups(region string) error {
 		errs = multierror.Append(errs, fmt.Errorf("error describing EC2 Security Groups for %s: %w", region, err))
 	}
 
-	if len(sweepResources) > 0 {
-		// Any errors didn't prevent gathering some sweeping work, so do it.
-		if err := testSweepResourceOrchestrator(sweepResources); err != nil {
-			errs = multierror.Append(errs, fmt.Errorf("error sweeping EC2 Security Groups for %s: %w", region, err))
-		}
+	if err := testSweepResourceOrchestrator(sweepResources); err != nil {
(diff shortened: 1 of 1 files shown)


aws/resource_aws_security_group_test.go
```diff
@@ -35,93 +35,54 @@ func init() {
...
-		log.Printf("[WARN] Skipping EC2 Security Group sweep for %s: %s", region, err)
-		return nil
-	}
-
 	if err != nil {
-		return fmt.Errorf("Error retrieving EC2 Security Groups: %s", err)
+		errs = multierror.Append(errs, fmt.Errorf("error describing EC2 Security Groups for %s: %w", region, err))
 	}
 
-	err = conn.DescribeSecurityGroupsPages(input, func(page *ec2.DescribeSecurityGroupsOutput, lastPage bool) bool {
```
bflad, line 88: If I recall correctly, the previous behavior was necessary as you could have security groups that cross-reference each other so it was not possible to delete them with revoking rules first.

YakDriver: The resource itself does all that as far as I can tell when `d.Set("revoke_rules_on_delete", true)`.

bflad: The problem isn't with the resource being deleted, it is a race condition that _other_ Security Groups may have rules that reference the current Security Group, which would cause the failures.

YakDriver:
This is old and will likely take more work to rebase than to simply re-implement. Thus, closing...

Issue #7926, opened 2019-03-13
Default tags for all resources


flowirtz:
### Description
> I'd like to set default tags for all resources created within a Terraform stack.

When creating service stacks on AWS with Terraform, I like to tag all resources of a project. Right now, this involves a lot of manual work.
I would like to set default tags with the provider, that then automatically get applied to all AWS resources that support it.

There has been an [issue over on the main terraform repo (since 2015)](https://github.com/hashicorp/terraform/issues/2283) which had plenty of 👍 and got closed in favor of creating issues with the respective providers. I couldn't find such an issue though, thus creating this one.

### New or Affected Resource(s)
* `aws_*`

### Potential Terraform Configuration
```hcl
provider "aws" {
  # ...

  default_tags = {
    Project = "Project_1"
    Environment = "Production"
  }
}
```

### References
* Original issue: hashicorp/terraform#2283

---

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

gwkunze:
In general (with my limited knowledge on the provider specifics) this should be possible. The one tricky part would be any resources that have additions to the basic tag properties. The one I know that does this is the auto scaling group resources which in addition to a key and value also has a _required_ property which determines whether the tags propagate to instances the ASG creates. So that should probably be taken into account when (if?) implementing this.

FWirtz:
@gwkunze sorry for the late reply, am currently on a long vacation, coming back next week and want to have a proper look at this then.

I've never handled the terraform core nor written any plugins, but want to give this a try anyways as I'd love to have this feature. 

Is there anything else that you can think of that one would need to keep in mind when implementing this? I'm very confident that my TF knowledge is way more limited than yours haha :-)

Thanks already for this suggestion!

FWirtz:
Alright, so I investigated this a little further and am pretty sure that I **won't** have the resources to tackle this. I have no prior experience with go, never contributed to this repo before and have not a whole lot of time.
I'm **happy to help somebody else** with minor things but that's all I can do, I think.

jonscheiding:
[...]

PR #18726, opened 2021-04-09
provider: Support default tags (resources aws_r*)

bflad:
### Community Note

Reference: https://github.com/hashicorp/terraform-provider-aws/issues/7926

Output from acceptance testing: Will be handled via CI, review, and daily acceptance testing

### DIFF
--- a/aws/resource_aws_route53_health_check.go
+++ b/aws/resource_aws_route53_health_check.go
@@ -153,8 +153,11 @@ func resourceAwsRoute53HealthCheck() *schema.Resource {
 				Default:  false,
 			},
 
-			"tags": tagsSchema(),
+			"tags":     tagsSchema(),
+			"tags_all": tagsSchemaComputed(),
 		},
+
+		CustomizeDiff: SetTagsDiff,
 	}
 }
 
@@ -227,8 +230,8 @@ func resourceAwsRoute53HealthCheckUpdate(d *schema.ResourceData, meta interface{
 		return err
 	}
 
-	if d.HasChange("tags") {
-		o, n := d.GetChange("tags")
+	if d.HasChange("tags_all") {
+		o, n := d.GetChange("tags_all")
 
 		if err := keyvaluetags.Route53UpdateTags(conn, d.Id(), route53.TagResourceTypeHealthcheck, o, n); err != nil {
 			return fmt.Errorf("error updating Route53 Health Check (%s) tags: %s", d.Id(), err)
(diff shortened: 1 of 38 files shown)


anGie44 (approved):

aws/resource_aws_route53_health_check.go
```diff
@@ -341,7 +346,7 @@ func resourceAwsRoute53HealthCheckCreate(d *schema.ResourceData, meta interface{
 
 	d.SetId(aws.StringValue(resp.HealthCheck.Id))
 
-	if err := keyvaluetags.Route53UpdateTags(conn, d.Id(), route53.TagResourceTypeHealthcheck, map[string]interface{}{}, d.Get("tags").(map[string]interface{})); err != nil {
+	if err := keyvaluetags.Route53UpdateTags(conn, d.Id(), route53.TagResourceTypeHealthcheck, nil, tags); err != nil {
```
anGie44, line 349: i _think_ here we'll need to pass in `tags.Map()` instead of `tags` since the `Route53UpdateTags` method will call `New` on the tags and result in a new `KeyValueTags` instance but without the values originally in it

aws/resource_aws_route53_zone.go
```diff
@@ -138,7 +143,7 @@ func resourceAwsRoute53ZoneCreate(d *schema.ResourceData, meta interface{}) erro
 		}
 	}
 
-	if err := keyvaluetags.Route53UpdateTags(conn, d.Id(), route53.TagResourceTypeHostedzone, map[string]interface{}{}, d.Get("tags").(map[string]interface{})); err != nil {
+	if err := keyvaluetags.Route53UpdateTags(conn, d.Id(), route53.TagResourceTypeHostedzone, nil, tags); err != nil {
```
anGie44, line 146: similar comment as above to pass in `tags.Map()` instead of `tags`...hmm or maybe we need to adjust the `keyvaluetags.New` method if we find ourselves doing this too often  🤔

aws/resource_aws_route53_health_check.go
bflad: Submitted https://github.com/hashicorp/terraform-provider-aws/pull/19011 to cover this nonintuitive situation 👍

aws/resource_aws_route53_zone.go
bflad: Submitted https://github.com/hashicorp/terraform-provider-aws/pull/19011 to cover this nonintuitive situation 👍

Issue #18329, opened 2021-03-22
ECS service wait_for_steady_state doesn't work


laur1s:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform AWS Provider Version

Terraform v0.14.5
+ provider registry.terraform.io/hashicorp/archive v2.1.0
+ provider registry.terraform.io/hashicorp/aws v3.33.0

### Affected Resource(s)

aws_ecs_service

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

We have a terraform code base that deploys around 10 ECS services. The problem is that sometimes after 3 minutes. Terraform fails with an error code: 
`Error: error waiting for service (arn:aws:ecs:eu-xxx-1:xxx:service/dev-xx/email) to reach a steady state: ResourceNotReady: failed waiting for successful resource state`

The strange part is that terraform instead of updating the services first tries to delete them and only then create new services. This error waiting for service happens at random times from 3-to 3:40minutes. Sometimes it launches services correctly. In AWS console I can see that all services are launched correctly without any errors!

```hcl
resource "aws_ecs_service" "email" {
  cluster                            = local.cluster
  deployment_maximum_percent         = 200
  deployment_minimum_healthy_percent = 50
  desired_count                      = 1
  health_check_grace_period_seconds  = 30
  name                               = local.email_name
  propagate_tags                     = "TASK_DEFINITION"
  scheduling_strategy                = "REPLICA"
  task_definition                    = aws_ecs_task_definition.email.arn
  wait_for_steady_state              = true

  deployment_controller {
    type = "ECS"
  }

  load_balancer {
    container_name   = local.email_name
    container_port   = 443
    target_group_arn = local.email_target_group
  }
}
```

### Expected Behavior
Update should update the ecs services

### Actual Behavior
Some random timeouts
### Steps to Reproduce

1. Create a terraform state with multiple ecs services
2. Try updating multiple serivices at the same time with  `terraform apply`

gdavison:
[...]

Issue #16012, opened 2020-11-04
ECS service parameter wait_for_ready_state can lead to inconsistent new deployments


maximelenair:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform AWS Provider Version

Terraform core version: 0.13.3
Terraform AWS provider version: 3.13.0

### Affected Resource(s)

* aws_ecs_service

### Terraform Configuration Files

```hcl
resource "aws_ecs_service" "my_service" {
  name          = "my-service"
  cluster       = "my-cluster"
  desired_count = 1

  launch_type             = "FARGATE"
  enable_ecs_managed_tags = true
  propagate_tags          = "SERVICE"

  deployment_maximum_percent         = 200
  deployment_minimum_healthy_percent = 50
  wait_for_steady_state              = true

  network_configuration {
    subnets         = ["subnet-*******", "subnet-*******", "subnet-*******"]
    security_groups = ["sg-********"]
  }

  service_registries {
    registry_arn = "arn:aws:servicediscovery:*******:********:service/srv-*********"
  }

  task_definition = "my-service:10"
  tags            = local.common_tags
}
```

### Expected Behavior

When using the `wait_for_steady_state` parameter during the *creation* of a service, we have multiple service / task status possible:

1. Both the service and the task are in a healthy state, `terraform apply` is successful
2. The service has an issue and `terraform apply` fails after a timeout (ie, the specific Docker image does not exist)
3. The service is running but the task has a consistent issue preventing it from passing its initial health check and `terraform apply` fails after a timeout (ie, a container fails to start because of a missing environment variable)

### Actual Behavior

In both case 1. and 2., the actual behaviour follows the expected behavior.
For case 3., the actual behavior is inconsistent given the same Terraform configuration.

Example on 5 different tests (deploy/destroy without any configuration change):
- Test 1: Success - after 2 minutes 30
- Test 2: Success - after 7 minutes 40
- Test 3: Success - after 1 minutes 20
- Test 4: Failure - timeout after 10 minutes
- Test 5: Success - after 5 minutes 20

[...]

Issue #17118, opened 2021-01-14
Terraform continues even if ecs task has not reached steady state


jmbosse19:
### Terraform CLI and Terraform AWS Provider Version

Terraform v0.14.3
+ provider registry.terraform.io/hashicorp/aws v3.23.0
+ provider registry.terraform.io/hashicorp/local v2.0.0
+ provider registry.terraform.io/hashicorp/tls v3.0.0

### Affected Resource(s)

aws_ecs_service

### Terraform Configuration Files

```hcl
resource "aws_ecs_service" "service" {
  name            = "${var.environment}-service"
  cluster         = aws_ecs_cluster.ecs_cluster.id
  task_definition = aws_ecs_task_definition.task_def.arn
  desired_count   = 1
  launch_type     = "EC2"

  load_balancer {
    target_group_arn = var.alb_tg
    container_name   = "container"
    container_port   = var.port
  }

  network_configuration {
    security_groups  = [var.aws_sec_group]
    subnets          = var.application_subnet_ids
    assign_public_ip = false
  }

  wait_for_steady_state = true
}

resource "aws_ecs_service" "service2" {
  name            = "${var.environment}-service2" ...

  depends_on [aws_ecs_service.service]
  wait_for_steady_state = true
}

```

### Debug Output

Output is successful, no errors

### Expected Behavior

Terraform creates the first ecs service. Terraform then waits for the ECS task for this service to reach a steady state. Once this state is reached, service 2 is created because service 1 is its dependency (as specified by a manual `depends_on`. 

### Actual Behavior

Terraform creates the first ecs service. Terraform does not wait for the ECS task to reach a steady state, instead it looks that once the task registers a target group is believes this is satisfied. Service 2 is created and if service 1 isn't actually ready for service 2 to connect, these will loop and fail, and continue to create other services in our application.

### Steps to Reproduce

1. `terraform apply`
2. Monitor the ecs service status.

pv93:
This issue also means that when we apply an update to an ECS service that requires a replacement (e.g changes to `network_configuration`) we will see downtime.

This is because a new replacement service is created and the old service is immediately destroyed without actually waiting for (desiredTasks == runningTasks) on the new service, as described in the docs for `aws ecs wait services-stable` [here](https://docs.aws.amazon.com/cli/latest/reference/ecs/wait/services-stable.html).

Issue #16888, opened 2020-12-23
Failed to update scaling policy: ValidationException: Target group


Zogoo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform AWS Provider Version

Terraform 12.0.6
* provider.aws: version = "~> 3.21"
* provider.local: version = "~> 2.0"
* provider.template: version = "~> 2.2"
* provider.tls: version = "~> 3.0"

### Affected Resource(s)

aws_ecs_service

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```

### Debug Output

```terraform
  # module.ecs_cluster.aws_ecs_service.portal must be replaced
+/- resource "aws_ecs_service" "web" {
        cluster                            = "arn:aws:ecs:ap-northeast-1:****:cluster/name_of_cluster"
        deployment_maximum_percent         = 150
        deployment_minimum_healthy_percent = 50
      ~ desired_count                      = 2 -> 3
        enable_ecs_managed_tags            = false
        health_check_grace_period_seconds  = 60
      ~ iam_role                           = "aws-service-role" -> (known after apply)
      ~ id                                 = "arn:aws:ecs:ap-northeast-1:****:service/name_of_cluster/name_of_service" -> (known after apply)
      + launch_type                        = (known after apply)
        name                               = "name_of_service"
      + platform_version                   = (known after apply)
      - propagate_tags                     = "NONE" -> null
        scheduling_strategy                = "REPLICA"
      - tags                               = {} -> null
      ~ task_definition                    = "name_of_task:28" -> (known after apply)
        wait_for_steady_state              = false
      - capacity_provider_strategy {
          - base              = 0 -> null
          - capacity_provider = "name_of_auto_scaling_group" -> null
          - weight            = 1 -> null
        }
[...]

PR #18868, opened 2021-04-14
Introduce custom timeout when waiting for `aws_ecs_service` to reach a steady state

### LINKED ISSUE #17746: ecs_service with wait_for_steady_state should allow to configure delay and retries
### Community Note

### Description

when applying `ecs_service` with `wait_for_steady_state=true`, terraform will wait for the service to reach a steady state.
currently there is no way to set the delay and max retries to be used while waiting.
the requested feature is to support passing custom parameters for those values

the default value is 10 minutes:
the function call is here: https://github.com/hashicorp/terraform-provider-aws/blob/master/aws/resource_aws_ecs_service.go#L1201
which defines the default values here: https://github.com/aws/aws-sdk-go/blob/master/service/ecs/waiters.go#L31-L32

* aws_ecs_service

jackbatzner:
### Community Note

Fixes #17746, #18329, #16012

```
$ make testacc TESTARGS='-run=TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyStateWithTimeout'
=== RUN   TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyStateWithTimeout
=== PAUSE TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyStateWithTimeout
=== CONT  TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyStateWithTimeout
--- PASS: TestAccAWSEcsService_withLaunchTypeFargateAndUpdateWaitForSteadyStateWithTimeout (179.35s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       185.647s
...
```


aws/resource_aws_ecs_service.go
```diff
@@ -379,6 +379,10 @@ func resourceAwsEcsService() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"wait_for_steady_state_timeout_in_minutes": {
```
brunhil, line 382: I'm not 100% sold on this property name and would love suggestions if anyone can come up with a better idea?

Another option I had was to put this in a element block like so,

```
wait_for_steady_state {
   enabled                     = true
   timeout_in_minutes = 15
}
```

YakDriver:
@brunhil Thank you for this PR! I believe this PR would fix #18329. You mention that it partially addresses it. What doesn't it cover?

Since there is a built-in timeout mechanism, we often avoid a custom timeout argument unless it's tied to an API argument. However, there are currently many timeout arguments and likely a good portion are not directly tied to an API argument (e.g., ASG, batch job, codebuild, cloudfront, codedeploy).

It would be great if an eventual solution could address most/all of the related issues:

Related #16012
Related #16184
Related #16526
Related #16888
Related #17118
Related #17746
Related #18329
Related #18868 (this PR)

brunhil:
Thanks for taking a peek @YakDriver ! I've answered a few of your questions below.

> @brunhil Thank you for this PR! I believe this PR would fix #18329. You mention that it partially addresses it. What doesn't it cover?

This actually does fix all of #18329 it appears. I don't known why I put the "partially addresses it" in the first place 😄 . 

> Since there is a built-in timeout mechanism, we often avoid a custom timeout argument unless it's tied to an API argument. However, there are currently many timeout arguments and likely a good portion are not directly tied to an API argument (e.g., ASG, batch job, codebuild, cloudfront, codedeploy).

I agree with this statement. The issue is the built in function has a [timeout of 10 minutes](https://github.com/aws/aws-sdk-go/blob/main/service/ecs/waiters.go#L31-L32). In my experience, AWS can take longer than 10 minutes to fully switch over from the old service to the new one. This often happens when the service is actively being used. 

> It would be great if an eventual solution could address most/all of the related issues:
> 
> Related #16012
> Related #16184
> Related #16526
> Related #16888
> Related #17118
> Related #17746
> Related #18329
> Related #18868 (this PR)

I've updated based on issues you found. Only one more is actually addressed, the others didn't seem too relevant based on my initial understanding of them.

aws/resource_aws_ecs_service.go
nickfaughey: 1. Block version is definitely more straightforward and easier to read
2. Most other timing-related parameters (health check timeout, deregistration delay, etc) are in seconds. Probably makes sense to follow that convention here also

dr-yd:
I can confirm the issue with deployments taking longer than 10 minutes - in our case, we require a deregistration delay of 20 minutes which causes the deployment to fail if a long-lasting connection is opened shortly before a new deployment. 

As per AWS documentation, the target is allowed to deregister once it has no more open connections. Thus, from my understanding the 20min are the maximum time Terraform would actually run, but it would be very useful to reflect that in the resource timeout.

The hard-coded 10min limit makes this impossible and the "error" is indistinguishable from an actual failure (like e. g. a missing Docker image). This is a large UX problem if the output should be understood by people who aren't intimately familiar with Terraform.

aws/resource_aws_ecs_service.go
tjhiggins: @brunhil Thank you for putting up this pr. Definitely a much needed addition. Could you please update so that it might get merged?

brunhil: I was hoping someone from the providers team would be able to provide some input on this potential config as well? 

Any chance you could provide input @YakDriver / @bflad / @anGie44 ?

luis-miguel-alves (approved):

luis-miguel-alves:
Could you please merge this? Thank you

luis-miguel-alves (approved):

tylerohlsen:
@YakDriver @luis-miguel-alves Looks like this PR has been stalled waiting for review.  Is there someone we can ping to get the review process going again?  This feature is something I would really like to use.

mkielar:
I know I'm late to the party, but why introduce a new option when we could just use:
```
timeout {
  create = "10m"
  update = "20m"
}
```
I mean, these two would still have to be implemented, because `aws_ecs_ervice` only supports `delete` timeout at the moment, but the user experience would at least be consistent with other resources and [general conventions](https://www.terraform.io/language/resources/syntax#operation-timeouts).

breathingdust:
Hi all 👋 Just letting you know that this is issue is featured on this quarters [roadmap](https://github.com/hashicorp/terraform-provider-aws/blob/main/ROADMAP.md). If a PR exists to close the issue a maintainer will review and either make changes directly, or work with the original author to get the contribution merged. If you have written a PR to resolve the issue please ensure the "Allow edits from maintainers" box is checked. Thanks for your patience and we are looking forward to getting this merged soon!

gdavison:
Hi @jackbatzner, thank you for contributing this PR. We think that using the built-in customizable timeouts functionality in Terraform is a better option to address this issue. We've created #25641 instead. Thanks again!

PR #19037, opened 2021-04-20
lb_target_group - Remove unknown attribute for health checks in the documentation

Ninir:
There is no `protocol_version` attribute in the health check configuration, which is confirmed by checking the schema as per https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/resource_aws_lb_target_group.go#L56.

This was added back in https://github.com/hashicorp/terraform-provider-aws/pull/17731.


ewbankkit (approved):
LGTM 🚀.

PR #19237, opened 2021-05-05
Extends semgrep rule to include `resource.RetryContext()`

gdavison:
Extends the existing semgrep check that ensures `resource.Retry()` calls to the API are given one last try if the retry times out to include `resource.RetryContext()`

N/A


ewbankkit (approved):
LGTM 🚀.

```console
% semgrep -f .semgrep.yml
running 22 rules...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████|22/22
ran 22 rules on 2481 files: 0 findings
```

PR #19271, opened 2021-05-07
Clarify key policy vs IAM policy

glerb:
Key policies are not the same as IAM policies. A key policy is technically required for each KMS key. There is no `aws_key_policy_document` data source, but the `aws_iam_policy_document` data source will work. KMS keys that are created programmatically are given a [default policy](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html#key-policy-default) that allows the `:root` of an account full access to the key, which allows IAM policies to be applied to the key.

### Community Note

Relates OR Closes #0000

N/A


breathingdust:
LGTM 🚀  Thanks @glerb!

PR #19276, opened 2021-05-07
Stop re-enabling termination protection during delete of aws_instance

### LINKED ISSUE #19275: Unable to delete aws_instance if disable_api_termination was true
### Community Note

### Terraform CLI and Terraform AWS Provider Version

Terraform: v0.14.8
Provider: v3.38.0

### Affected Resource(s)

* aws_instance

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
  disable_api_termination = true
  # ...
}
```

### Expected Behavior

If I manually enable api termination on instance previously managed by terraform that had `disable_api_termination` set to true, and then apply a change to destroy that instance, the instance should be destroyed, and terraform should not re-enable termination protection as part of the apply.

### Actual Behavior

Apply deletion of an aws_instance always modifies the attribute to match what is in the resource config 

### Steps to Reproduce

1. `terraform apply`

### Cause

I'm pretty sure this was introduced in  commit 0a116d70949241082 with this change:

```
-	// false = enable api termination
-	// true = disable api termination (protected)
-	if !d.Get("disable_api_termination").(bool) {
-		_, err := conn.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
-			InstanceId: aws.String(d.Id()),
-			DisableApiTermination: &ec2.AttributeBooleanValue{
-				Value: aws.Bool(d.Get("disable_api_termination").(bool)),
-			},
-		})
+	err := resourceAwsInstanceDisableAPITermination(conn, d.Id(), d.Get("disable_api_termination").(bool))
 
-		if err != nil {
-			log.Printf("[WARN] attempting to terminate EC2 instance (%s) despite error enabling API termination: %s", d.Id(), err)
-		}
+	if err != nil {
+		log.Printf("[WARN] attempting to terminate EC2 instance (%s) despite error enabling API termination: %s", d.Id(), err)
 	}
```
note that it now unconditionally modifies the disable_api_termination attribute, instead of only changing it if disable_api_termination is false.

## Discussion
Would it be better to just always set the attribute to false before terminating? Rather than depending on the disable_api_termination setting. (or maybe the intention here was that if disable_api_termination was true in the config, then it should be changed to false when the instance is destroyed?)

tmccombs:
Closes #19275

This just restores the previous behavior. Another option would be to
just always modify the attribute to false so that deleting works even if
`disable_api_termination` is true.

### Community Note

### DIFF
--- a/internal/service/ec2/instance.go
+++ b/internal/service/ec2/instance.go
@@ -1726,13 +1726,15 @@ func resourceInstanceUpdate(d *schema.ResourceData, meta interface{}) error {
 func resourceInstanceDelete(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*conns.AWSClient).EC2Conn
 
-	err := resourceInstanceDisableAPITermination(conn, d.Id(), d.Get("disable_api_termination").(bool))
+	if !d.Get("disable_api_termination").(bool) {
+		err := resourceInstanceDisableAPITermination(conn, d.Id(), false)
 
-	if err != nil {
-		log.Printf("[WARN] attempting to terminate EC2 instance (%s) despite error modifying attribute (%s): %s", d.Id(), ec2.InstanceAttributeNameDisableApiTermination, err)
+		if err != nil {
+			log.Printf("[WARN] attempting to terminate EC2 instance (%s) despite error modifying attribute (%s): %s", d.Id(), ec2.InstanceAttributeNameDisableApiTermination, err)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


ewbankkit:
@tmccombs Thanks for the contribution :tada: :clap:.
Let's go with the implementation in #19277 for this.

tmccombs:
Awesome, thank you.

PR #19562, opened 2021-05-27
Fix invalid example: session_number is required

atz:
Otherwise the error is:
```
The argument "session_number" is required, but no definition was found.
```


anGie44 (approved):
Looks great @atz , thank you for this fix! 🚀 🚀

PR #10812, opened 2019-11-08
docs/provider: Remove source content for guides which are no longer hosted by this repository

bflad:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" comments, they generate extra noise for pull request followers and do not help prioritize the request

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):

```release-note
NONE
```

Output from acceptance testing: N/A

Historically we have left the source for these guides in this repository for open source contribution even after they were migrated to learn.hashicorp.com, which has its own source files. This has been okay up until this point with some manual synchronization of any changes, however they will now show up unexpectedly in the Terraform Registry documentation if left here.

PR #20151, opened 2021-07-12
Remove incorrect reference to IAM policy in aws_ecr_lifecycle_policy website docs

emsearcy:
The documentation for the "policy" attribute of aws_ecr_lifecycle_policy incorrectly links out to an IAM policy tutorial which has nothing to do with this resource. (The attribute uses a JSON body, and has an attribute name of "policy", but otherwise has nothing to do with the IAM policies that the second sentence of the attribute description links to.)

The first sentence already links to the correct upstream documentation and this is an obscure enough resource there are no other Hashicorp "learn" links to put in its place, so I've simply removed it.

This only affects documentation; there is no corresponding bug or sample code to reproduce.

### Community Note

### DIFF
--- a/website/docs/r/ecr_lifecycle_policy.html.markdown
+++ b/website/docs/r/ecr_lifecycle_policy.html.markdown
@@ -85,7 +85,7 @@ EOF
 The following arguments are supported:
 
 * `repository` - (Required) Name of the repository to apply the policy.
-* `policy` - (Required) The policy document. This is a JSON formatted string. See more details about [Policy Parameters](http://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lifecycle_policy_parameters) in the official AWS docs. For more information about building IAM policy documents with Terraform, see the [AWS IAM Policy Document Guide](https://learn.hashicorp.com/terraform/aws/iam-policy).
+* `policy` - (Required) The policy document. This is a JSON formatted string. See more details about [Policy Parameters](http://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lifecycle_policy_parameters) in the official AWS docs.
 
 ## Attributes Reference
 


anGie44:
Hi @emsearcy , thank you for raising this PR. It does look though the current state of the documentation is by design and introduced in https://github.com/hashicorp/terraform-provider-aws/pull/10812/files to replace an older link hosted by AWS. I do believe the link helps new practitioners find additional terraform-related resources when new to the ecosystem, though I understand it may be a bit extraneous for those more comfortable with the tools and AWS. I'm going to close this PR in favor of retaining the information for new practitioners that make their way to the provider's documentation in the registry, but thank you for taking the time to review our docs!

emsearcy:
Hi @anGie44 - regardless of the outcome, thank you for your time reviewing and considering this!

I would briefly ask you to reconsider. The former link was introduced in aa49feb by @bflad with the comment "This guide is designed to highlight some examples how Terraform and the AWS provider allow for many configuration styles to implement these policies. We then can cross reference this guide in all documentation *where IAM policy documents are used*." (emphasis mine)  However, this resource (and possibly others introduced in aa49feb) do **not** use an **IAM** policy document, it rather has an entirely-unrelated resource that happens to share the attribute name "policy".  I believe this was a sort of search/replace mistake, not intentional.

I'm not suggesting that linking a tutorial doc isn't needed because the AWS link is sufficient, nor suggesting users ought to be experienced enough to not need an extra tutorial; rather, I'm saying the linked tutorial has nothing to do with the resource that is referencing it (and thus would, in my opinion anyhow, be more confusing to newcomers).

Warm regards, Eric

anGie44:
Sure thing @emsearcy! And I see you definitely bring a valid point, the link looks incorrect even before this `learn` content was introduced as `policy` in this context is strictly JSON, not IAM-related. Let me reopen this and get this merged in! Thanks for your patience and feedback!

anGie44 (approved):

PR #20205, opened 2021-07-15
Remove misleading wording

mootpt:
Peering isn't actually a requirement: https://docs.aws.amazon.com/cli/latest/reference/route53/create-vpc-association-authorization.html


breathingdust:
LGTM 🚀  Thanks @mootpt!

Issue #10715, opened 2019-11-01
Add retry handling when a request's connection is reset by peer 


davegallant:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

Terraform v0.12.10

### Affected Resource(s)

- aws_iam_instance_profile

### Terraform Configuration Files

```hcl
data "aws_iam_role" "my_role" {
  name = "0f9f1e2t-instance"
}
```

### Debug Output

N/A

### Panic Output

N/A

### Expected Behavior

It would be nice if there was a retry mechanism implemented for this resource since it is only doing a read.

### Actual Behavior

```shell
Error: Error reading IAM instance profile 0f9f1e2t-instance: RequestError: send request failed

caused by: Post https://iam.amazonaws.com/: read tcp 172.17.0.2:36404->59.133.22.207:443: read: connection reset by peer
```

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

Does not look like there is any retry logic when reading an IAM instance profile:

https://github.com/terraform-providers/terraform-provider-aws/blob/98b8b848ca94031b20c3e626c9d40484e3af80de/aws/resource_aws_iam_instance_profile.go#L287-L305 

An example of retrying within the same file:
https://github.com/terraform-providers/terraform-provider-aws/blob/98b8b848ca94031b20c3e626c9d40484e3af80de/aws/resource_aws_iam_instance_profile.go#L163-L175 

### References

None

camlow325:
We saw something similar but with an `aws_iam_account_alias` data source instead. In that case, at least, it appeared that Terraform would attempt to perform some number of retries for the failed API call - up to the value configured for [max_retries](https://www.terraform.io/docs/providers/aws/index.html#max_retries) for the AWS provider instance - for cases where the request failed due to an `i/o timeout`. If a `connection reset by peer` failure occurred, though, like the one mentioned in this issue, no further retries were attempted. Would it make sense to make the more generic API retry handling be used for `connection reset by peer` errors?

analogrithems:
This is also happening for us when working with s3 buckets

`
[...]

PR #20300, opened 2021-07-24
WIP: Automatically retry when encounter connection reset by peer error from aws api

jiashuChen:
### Community Note

Relates #10715 

Draft PR, I'm not sure if this is the most correct way to solve this issue, raising as a draft, feedback will be appreciated. The goal of this pull request is to resolve to have a global wrapper function to deal with connection reset by peer error from aws api, which seems to impact multiple aws region and services.

```
│ Error: error deleting IAM Role (IAM-ROLE-NAME): RequestError: send request failed
│ caused by: Post "https://iam.amazonaws.com/": read tcp IP:PORT->DIFFERENT_IP:PORT: read: connection reset by peer
```

Furthermore, this helps resolve the issue with `resource.Retry` being marked deprecated.

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/wait.go#L66-L69

Lastly, this pull request now seems to have more than 100 labels automatically applied to it, which has caused verification to fail. I'm not sure how to resolve it. Some help will be appreciated.

NOTE: although there are many files changes, the core of the change is here in this file:
https://github.com/hashicorp/terraform-provider-aws/blob/03ea02c693ab28c09cbcce39dfe5ba98676eb33a/aws/internal/tfresource/retry.go#L119-L129


aws/internal/tfresource/retry.go
```diff
@@ -114,3 +115,15 @@ func RetryConfigContext(ctx context.Context, delay time.Duration, delayRand time
 	// more likely to be useful
 	return resultErr
 }
+
+func RetryOnConnectionResetByPeer(timeout time.Duration, f resource.RetryFunc) error {
```
jiashuChen, line 119: Main Change

aws/internal/tfresource/retry_test.go
```diff
@@ -97,3 +98,55 @@ func TestRetryConfigContext_error(t *testing.T) {
 		t.Fatal("timeout")
 	}
 }
+
+func TestRetryOnConnectionResetByPeer(t *testing.T) {
```
jiashuChen, line 102: Unit Test

AdamTylerLynch:
Question for the author: the AWS GoSDK has retry configured by default for all client operations. The retry mechanism will retry 3 times by default and use exponential back off with jitter.

I’m curious as to how this PR will be any different? I would suggest modifying the default policy for the SDK rather than implement additional retry logic, specifically because that logic does not include exponential backoff with jitter.

https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/

Additionally, TCP reset could be happening for a number of reasons. I use Terraform all the time, running the Acceptance Tests, 20 threads at a time, and I’ve never gotten a TCP reset error. 

Have you checked your network route to AWS? A proxy, machine-in-the-middle, or AWS API throttling are all potential sources of TCP reset. I would recommend checking your CloudTrail logs for your account to see if you are being throttled.

jiashuChen:
> Question for the author: the AWS GoSDK has retry configured by default for all client operations. The retry mechanism will retry 3 times by default and use exponential back off with jitter.
> 
> I’m curious as to how this PR will be any different? I would suggest modifying the default policy for the SDK rather than implement additional retry logic, specifically because that logic does not include exponential backoff with jitter.
> 
> https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/
> 
> Additionally, TCP reset could be happening for a number of reasons. I use Terraform all the time, running the Acceptance Tests, 20 threads at a time, and I’ve never gotten a TCP reset error.
> 
> Have you checked your network route to AWS? A proxy, machine-in-the-middle, or AWS API throttling are all potential sources of TCP reset. I would recommend checking your CloudTrail logs for your account to see if you are being throttled.

Thanks. I'll close the current pull requests.

PR #20428, opened 2021-08-03
docs: align docs for fields with SchemaConfigModeAttr

radeksimko:
Fields which use `ConfigMode: schema.SchemaConfigModeAttr` are generally treated as attributes by core. See https://github.com/hashicorp/terraform/blob/v1.0.3/internal/lang/blocktoattr/schema.go for more.

Additionally it is proposed to deprecate blocks generally in providers in the long run as per https://github.com/hashicorp/terraform-plugin-framework/issues/85

This documentation update reflects this plan and current reality by encouraging users to use the attribute notation.

I assume that some fields should probably be renamed to plural in the long run, but that seems like a job for another PR and more careful planning.


ewbankkit (approved):
LGTM :rocket:.

PR #20873, opened 2021-09-10
Use better ARN as an example for importing aws_imagebuilder_infrastructure_configuration

szemek:
A tiny fix. An example of import is suggesting ARN for different resource which might be misleading.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@szemek Thanks for the contribution :tada: :clap:.

PR #21088, opened 2021-09-29
Fix syntax in example usage

eihli:
Running `terraform apply` (v1.0.5) with this example resulted in:

│ Error: Incorrect attribute value type
│
│   on main.tf line 94, in resource "aws_default_security_group" "default":
│   94:   ingress = [
│   95:     {
│   96:       protocol  = -1
│   97:       self      = true
│   98:       from_port = 0
│   99:       to_port   = 0
│  100:     }
│  101:   ]
│
│ Inappropriate value for attribute "ingress": element 0: attributes "cidr_blocks",
│ "description", "ipv6_cidr_blocks", "prefix_list_ids", and "security_groups" are
│ required.

Is there a parity mismatch between the two ways of defining a collection of blocks?
```
ingress { ... }
ingress { ... }
```

vs.

```
ingress = [{ ... }, { ... }]
```

From what I gather from this error message and the fact that changing the syntax resolved the error, that means there are certain required-present attributes in the `ingress = []` syntax that aren't required in the `ingress {}` syntax?

### Community Note

Relates OR Closes #0000

```

...
```

### DIFF
--- a/website/docs/r/default_security_group.html.markdown
+++ b/website/docs/r/default_security_group.html.markdown
@@ -32,23 +32,19 @@ resource "aws_vpc" "mainvpc" {
 resource "aws_default_security_group" "default" {
   vpc_id = aws_vpc.mainvpc.id
 
-  ingress = [
-    {
-      protocol  = -1
-      self      = true
-      from_port = 0
-      to_port   = 0
-    }
-  ]
+  ingress {
+    self      = true
+    from_port = 0
+    to_port   = 0
+    protocol  = "-1"
+  }
 
-  egress = [
-    {
-      from_port   = 0
-      to_port     = 0
-      protocol    = "-1"
-      cidr_blocks = ["0.0.0.0/0"]
-    }
-  ]
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
 }
 ```
 


justinretzolk:
Hey @eihli 👋 Thank you for taking the time to open this PR. In looking it over, I think this may actually be a bug that needs to be addressed rather than a documentation fix. If I update the configuration that's described in the documentation now as below, it seems to work (the method you proposed works too, but I suspect the way it's described in the documentation is intended to work).

Would you mind opening an issue for this so that we can look into the root cause of what's causing this? I think that may be a better approach than updating the documentation for now, given that this seems to be unintended behavior.

```hcl
resource "aws_default_security_group" "default" {
  vpc_id = aws_vpc.mainvpc.id

  ingress = [
    {
      protocol  = -1
      self      = true
      from_port = 0
      to_port   = 0
      description = "test"
      cidr_blocks = []
      ipv6_cidr_blocks = []
      prefix_list_ids = []
      security_groups = []
    }
  ]

  egress = [
    {
      from_port   = 0
      to_port     = 0
      protocol    = "-1"
      self        = true
      cidr_blocks = []
      description = "test"
      ipv6_cidr_blocks = []
      prefix_list_ids = []
      security_groups = []
    }
  ]
}
```

eihli:
I'm trying to find the best location to open the issue. I'm not familiar enough with Go or Terraform to identify which of the following repositories the issue stems from. And I want now to make some notes that may help others (or future me) before I run out of time and have to move on to other tasks.

`terraform-provider-aws`

The schema correctly marks these attributes as "optional": https://github.com/hashicorp/terraform-provider-aws/blob/46a8d79d3492654c8c0119d7545385328a9cbea2/aws/resource_aws_default_security_group.go#L22

```
			"ingress": {
				Type:       schema.TypeSet,
				Optional:   true,
				Computed:   true,
				ConfigMode: schema.SchemaConfigModeAttr,
				Elem: &schema.Resource{
                                                // ,,,
						},
						"description": {
							Optional:     true,
							ValidateFunc: validateSecurityGroupRuleDescription,
						},
```

`terraform-plugin-sdk`

And here are some comments related to the `schema.Resource` `Elem` type: https://github.com/hashicorp/terraform-plugin-sdk/blob/4d35f0240509689df3aedf16cd1ba90fb88a00aa/helper/schema/schema.go#L62

```
	// If Elem is *schema.Resource then setting ConfigMode to
	// SchemaConfigModeAttr will force it to be represented in configuration
	// as an attribute, which means that the Computed flag can be used to
	// provide default elements when the argument isn't set at all, while still
	// allowing the user to force zero elements by explicitly assigning an
	// empty list.
```

`hcl2`

The error comes comes from these lines https://github.com/hashicorp/hcl2/blob/fb75b3253c80b3bc7ca99c4bfa2ad6743841b1af/hcldec/spec.go#L204

```
	convVal, err := convert.Convert(val, s.Type)
	if err != nil {
		diags = append(diags, &hcl.Diagnostic{
			Severity: hcl.DiagError,
			Summary:  "Incorrect attribute value type",
			Detail: fmt.Sprintf(
				"Inappropriate value for attribute %q: %s.",
				s.Name, err.Error(),
			),
			Subject: attr.Expr.StartRange().Ptr(),
			Context: hcl.RangeBetween(attr.NameRange, attr.Expr.StartRange()).Ptr(),
		})
		// We'll return an unknown value of the _correct_ type so that the
		// incomplete result can still be used for some analysis use-cases.
		val = cty.UnknownVal(s.Type)
	}
```

`go-cty`

Which comes from these lines https://github.com/zclconf/go-cty/blob/78acef51bab6df1a132424dbf28dd9cdad481cff/cty/convert/public.go#L42. 

```
func Convert(in cty.Value, want cty.Type) (cty.Value, error) {
	if in.Type().Equals(want) {
		return in, nil
	}

	conv := GetConversionUnsafe(in.Type(), want)
	if conv == nil {
		return cty.NilVal, errors.New(MismatchMessage(in.Type(), want))
	}
	return conv(in)
}
```

justinretzolk:
Hey @eihli, thank you for digging in as much as you have already! I think that it's totally fair to open the issue in this repository, and we can work to track down the root cause; particularly given that you mentioned that you're a bit constrained for time.

justinretzolk:
Thank you! With that in mind, I'm going to close out the PR for now. Thank you again for filing this!

sandeepsarma:
> Hey @eihli 👋 Thank you for taking the time to open this PR. In looking it over, I think this may actually be a bug that needs to be addressed rather than a documentation fix. If I update the configuration that's described in the documentation now as below, it seems to work (the method you proposed works too, but I suspect the way it's described in the documentation is intended to work).
> 
> Would you mind opening an issue for this so that we can look into the root cause of what's causing this? I think that may be a better approach than updating the documentation for now, given that this seems to be unintended behavior.
> 
> ```terraform
> resource "aws_default_security_group" "default" {
>   vpc_id = aws_vpc.mainvpc.id
> 
>   ingress = [
>     {
>       protocol  = -1
>       self      = true
>       from_port = 0
>       to_port   = 0
>       description = "test"
>       cidr_blocks = []
>       ipv6_cidr_blocks = []
>       prefix_list_ids = []
>       security_groups = []
>     }
>   ]
> 
>   egress = [
>     {
>       from_port   = 0
>       to_port     = 0
>       protocol    = "-1"
>       self        = true
>       cidr_blocks = []
>       description = "test"
>       ipv6_cidr_blocks = []
>       prefix_list_ids = []
>       security_groups = []
>     }
>   ]
> }
> ```



> Hey @eihli 👋 Thank you for taking the time to open this PR. In looking it over, I think this may actually be a bug that needs to be addressed rather than a documentation fix. If I update the configuration that's described in the documentation now as below, it seems to work (the method you proposed works too, but I suspect the way it's described in the documentation is intended to work).
> 
> Would you mind opening an issue for this so that we can look into the root cause of what's causing this? I think that may be a better approach than updating the documentation for now, given that this seems to be unintended behavior.
> 
> ```terraform
> resource "aws_default_security_group" "default" {
>   vpc_id = aws_vpc.mainvpc.id
> 
>   ingress = [
>     {
>       protocol  = -1
>       self      = true
>       from_port = 0
>       to_port   = 0
>       description = "test"
>       cidr_blocks = []
>       ipv6_cidr_blocks = []
>       prefix_list_ids = []
>       security_groups = []
>     }
>   ]
> 
>   egress = [
>     {
>       from_port   = 0
>       to_port     = 0
>       protocol    = "-1"
>       self        = true
>       cidr_blocks = []
>       description = "test"
>       ipv6_cidr_blocks = []
>       prefix_list_ids = []
>       security_groups = []
>     }
>   ]
> }
> ```

By specifying all of the (Optional) arguments with null arrays I was able to overcome this issue. I am on Terraform v1.0.10

PR #21078, opened 2021-09-28
Removes dependency vendoring for `awsproviderlint`

gdavison:
Removes dependency vendoring for `awsproviderlint`

PR #21114, opened 2021-09-30
Restore dependency vendoring for `awsproviderlint`

gdavison:
Turns out that the `vendor` directory was needed, since the testing framework for the `go/analysis` framework does not work with Go modules.

Relates #21078


ewbankkit (approved):
LGTM :rocket:.

YakDriver:
😞

PR #21120, opened 2021-10-01
Docs/add note for ssm association to target all instances

GlennChia:
add note and example code snippet for `aws_ssm_association` to target all instances in an AWS account

### Community Note

### Description

In the console, one can target all instances for an SSM association as shown below:

image

Doing it from Terraform is slightly confusing as the previous documentation did not specify how one can target all instances in an AWS account for `aws_ssm_association`. This issue was also faced in the CloudFormation documentation. After providing feedback to the CloudFormation team, they have updated their set of docs at [AWS::SSM::Association](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-association.html#aws-resource-ssm-association-properties) to explicitly specify how to use the `targets` attribute to target all instances in an AWS account. Similarly, I felt it would be useful to update the Terraform documentation to help users with the same need and added the note to the Terraform documentation.

### Testing

I've tested it with the following snippet

```terraform
resource "aws_ssm_association" "example" {
  name = "AmazonCloudWatch-ManageAgent"

  targets {
    key    = "InstanceIds"
    values = ["*"]
  }
}
```

which has an output of:

image

Verifying on the console also shows that it now targets all instances

image

### DIFF
--- a/website/docs/r/ssm_association.html.markdown
+++ b/website/docs/r/ssm_association.html.markdown
@@ -27,7 +27,7 @@ resource "aws_ssm_association" "example" {
 
 ### Create an association for all managed instances in an AWS account
 
-To target all instances in an AWS account, set the `key` as `"InstanceIds"` with `values` set as `["*"]`. This example also illustrates how to use an Amazon owned SSM document named `AmazonCloudWatch-ManageAgent`.
+To target all managed instances in an AWS account, set the `key` as `"InstanceIds"` with `values` set as `["*"]`. This example also illustrates how to use an Amazon owned SSM document named `AmazonCloudWatch-ManageAgent`.
 
 ```terraform
 resource "aws_ssm_association" "example" {
--- a/website/docs/r/ssm_association.html.markdown
+++ b/website/docs/r/ssm_association.html.markdown
@@ -53,6 +53,7 @@ Targets specify what instance IDs or tags to apply the document to and has these
 
 -> **Note:** To target all instances in an AWS account, set the `key` as `"InstanceIds"` with `values` set as `["*"]`.
 Example:
+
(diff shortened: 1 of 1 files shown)


website/docs/r/ssm_association.html.markdown
```diff
@@ -51,6 +51,20 @@ Targets specify what instance IDs or tags to apply the document to and has these
...
+resource "aws_ssm_association" "example" {
+  name = "AmazonCloudWatch-ManageAgent"
+
+  targets {
+    key    = "InstanceIds"
+    values = ["*"]
+  }
+}
+```
+
```
justinretzolk, line 67: I think it probably makes sense to move this up under the Examples header and switch from using `**Note**` to an H3 header; perhaps something like `### Targeting All Instances`. Note that that would also require adding an H3 header to the existing example (perhaps `### Targeting A Single Instance`?). 

There's precedence for this in a few places. I grabbed the [aws_s3_bucket](https://github.com/hashicorp/terraform-provider-aws/blob/de0147b7c2d9051bee842b8f6579e9dbbbf831bf/website/docs/r/s3_bucket.html.markdown#L15-L55) documentation as an example.

justinretzolk (changes requested):

website/docs/r/ssm_association.html.markdown
GlennChia: Thanks @justinretzolk for the review! I've made the requested changes. For completeness I've also added an example on how to target **managed** instances based on their tags.

justinretzolk (approved):
Looks good to me! Thank you again for your contribution! 🚀

PR #21153, opened 2021-10-05
wafv2_web_acl doc: Update rate based example

chtitux:
- If we rate-limit clients, we should **Allow by default, block when rule is matching**
- Add a comment so geographical restriction is explicit
- Use `CLOUDFRONT` for scope to have at least one Cloudfront example for the `scope` parameter

### Community Note


anGie44 (approved):
LGTM, thank you @chtitux 🚀

PR #21709, opened 2021-11-10
Docs: Add Transit Gateway filter support options

thatInfrastructureGuy:
Source: https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-transit-gateways.html

### Community Note

### DIFF
--- a/website/docs/d/ec2_transit_gateway.html.markdown
+++ b/website/docs/d/ec2_transit_gateway.html.markdown
@@ -40,24 +40,8 @@ The following arguments are supported:
 
 ### filter Argument Reference
 
-* `name` - (Required) Name of the filter.
+* `name` - (Required) The name of the field to filter by, as defined by the [underlying AWS API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeTransitGateways.html).
 * `values` - (Required) List of one or more values for the filter.
-
-#### Supported values for Filters
-
-One or more filters are supported. The possible values are:
-
-* `options.propagation-default-route-table-id` - The ID of the default propagation route table.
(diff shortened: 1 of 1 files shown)


justinretzolk (changes requested):
Hey @thatInfrastructureGuy 👋 Thank you very much for opening this pull request! In reviewing it, I think we should maybe go a slightly different route in order to match up with precedence. The [`aws_route_tables` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/route_tables) has a similar filter option, but links out to the relevant AWS documentation for the API. This means that if there are any changes to the filter options, the documentation here wouldn't need to be specifically updated, which I think would be a better experience. 

My suggestion would be to update the `name` parameter to something like:

thatInfrastructureGuy:
> Hey @thatInfrastructureGuy 👋 Thank you very much for opening this pull request! In reviewing it, I think we should maybe go a slightly different route in order to match up with precedence. The [`aws_route_tables` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/route_tables) has a similar filter option, but links out to the relevant AWS documentation for the API. This means that if there are any changes to the filter options, the documentation here wouldn't need to be specifically updated, which I think would be a better experience.
> 
> My suggestion would be to update the `name` parameter to something like:
> 
> ```gfm
> - `name` - (Required) The name of the field to filter by, as defined by the [underlying AWS API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeTransitGateways.html).
> ```

@justinretzolk That makes sense to me. I have done the requested changes. Thank you!

justinretzolk (approved):
Looks good to me! Thank you for your contribution!  I'll get this merged as soon as the checks complete 🚀

PR #21868, opened 2021-11-23
Clean up docs on aws_s3_bucket_policy

moskyb:
Change the docs on the `aws_s3_bucket_policy` resource example so that:
- Instead of using the `jsonencode` function, it uses the `aws_iam_policy_document` data source
- Change the policy attached to be a bit simpler and easier to understand

Also, update the docs on the `policy` member of the resource to suggest using the `aws_iam_policy_document` data source.

### Community Note

Relates OR Closes #0000

### DIFF
--- a/website/docs/r/s3_bucket_policy.html.markdown
+++ b/website/docs/r/s3_bucket_policy.html.markdown
@@ -15,24 +15,20 @@ Attaches a policy to an S3 bucket resource.
 ### Basic Usage
 
 ```terraform
-locals {
-  aws_account_b_id = "123456789012"
-}
-
 resource "aws_s3_bucket" "example" {
   bucket = "my-tf-test-bucket"
 }
 
-resource "aws_s3_bucket_policy" "allow_access_from_account_b" {
+resource "aws_s3_bucket_policy" "allow_access_from_another_account" {
   bucket = aws_s3_bucket.example.id
-  policy = data.aws_iam_policy_document.allow_access_from_account_b.json
+  policy = data.aws_iam_policy_document.allow_access_from_another_account.json
 }
 
-data "aws_iam_policy_document" "allow_access_from_account_b" {
+data "aws_iam_policy_document" "allow_access_from_another_account" {
   statement {
     principals {
-      type         = "AWS"
-      identitfiers = [local.aws_account_b_id]
+      type        = "AWS"
+      identifiers = ["123456789012"]
     }
 
     actions = [
--- a/website/docs/r/s3_bucket_policy.html.markdown
+++ b/website/docs/r/s3_bucket_policy.html.markdown
@@ -15,36 +15,33 @@ Attaches a policy to an S3 bucket resource.
 ### Basic Usage
 
 ```terraform
-resource "aws_s3_bucket" "b" {
+locals {
+  aws_account_b_id = "123456789012"
+}
+
+resource "aws_s3_bucket" "example" {
   bucket = "my-tf-test-bucket"
 }
 
-resource "aws_s3_bucket_policy" "b" {
-  bucket = aws_s3_bucket.b.id
+resource "aws_s3_bucket_policy" "allow_access_from_account_b" {
+  bucket = aws_s3_bucket.example.id
+  policy = data.aws_iam_policy_document.allow_access_from_account_b.json
+}
 
-  # Terraform's "jsonencode" function converts a
-  # Terraform expression's result to valid JSON syntax.
-  policy = jsonencode({
-    Version = "2012-10-17"
-    Id      = "MYBUCKETPOLICY"
-    Statement = [
-      {
-        Sid       = "IPAllow"
-        Effect    = "Deny"
-        Principal = "*"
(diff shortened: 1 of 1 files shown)


website/docs/r/s3_bucket_policy.html.markdown
```diff
@@ -53,7 +50,7 @@ resource "aws_s3_bucket_policy" "b" {
 The following arguments are supported:
 
 * `bucket` - (Required) The name of the bucket to which to apply the policy.
-* `policy` - (Required) The text of the policy. For more information about building AWS IAM policy documents with Terraform, see the [AWS IAM Policy Document Guide](https://learn.hashicorp.com/terraform/aws/iam-policy). Note: Bucket policies are limited to 20 KB in size.
+* `policy` - (Required) The text of the policy. Although this isn't technically an IAM policy, you can use the [`aws_iam_policy_document`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) data source for this, so long as it specifies a principal. For more information about building AWS IAM policy documents with Terraform, see the [AWS IAM Policy Document Guide](https://learn.hashicorp.com/terraform/aws/iam-policy). Note: Bucket policies are limited to 20 KB in size.
```
justinretzolk, line 53: Since the note you're adding is covered in the linked AWS IAM policy document guide, I don't believe it's necessary to add it here as well.

website/docs/r/s3_bucket_policy.html.markdown
```diff
@@ -15,36 +15,33 @@ Attaches a policy to an S3 bucket resource.
 ### Basic Usage
 
 ```terraform
-resource "aws_s3_bucket" "b" {
+locals {
+  aws_account_b_id = "123456789012"
+}
```
justinretzolk, line 20: Though I understand why you'd do so, I don't think adding a `locals` block here is necessary, and may be confusing for less experienced Terraform practitioners; particular since it's only being used once in the example. We can omit this and pass it directly below.

```suggestion
```

website/docs/r/s3_bucket_policy.html.markdown
```diff
@@ -15,36 +15,33 @@ Attaches a policy to an S3 bucket resource.
...
+resource "aws_s3_bucket_policy" "allow_access_from_account_b" {
+  bucket = aws_s3_bucket.example.id
+  policy = data.aws_iam_policy_document.allow_access_from_account_b.json
+}
+
+data "aws_iam_policy_document" "allow_access_from_account_b" {
+  statement {
+    principals {
+      type         = "AWS"
+      identitfiers = [local.aws_account_b_id]
```
justinretzolk, line 35: As noted above, I think adding a `locals` block might add more confusion for someone less familiar with Terraform, so it probably makes more sense to just pass this value directly.

```suggestion
      identitfiers = ["123456789012"]
```

justinretzolk (changes requested):
Thank you for taking the time to submit this PR! I added a couple of small notes to try to help with brevity and clarity.

website/docs/r/s3_bucket_policy.html.markdown
moskyb: yep, this makes total sense - the logic behind adding the local was to ensure that the reader knew that the string of numbers was an account ID and not some other piece of AWS deep lore, but i figure that we have to assume some level of familiarity (or ability to look things up) with regards to AWS.

Will update :)

website/docs/r/s3_bucket_policy.html.markdown
moskyb: i'm a little conflicted on this - i modeled this message after the one on [`aws_kms_key`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/kms_key), though with slightly friendlier language. the other thing is that the tutorial, while helpful in learning best practice for IAM policies, isn't necessarily the best way of consuming quick-reference information.

i'm happy to make the changes you're suggesting obviously - you're the maintainer here 😅 but thought it might be a good idea to detail the reasoning behind why i wrote what i did

justinretzolk: Thanks for the link! I see where you're coming from now (and wasn't really ever _against_ the idea, for what it's worth 😃). With that in mind, I think this rephrasing might present a bit clearer.

```suggestion
* `policy` - (Required) The text of the policy. Although this is a bucket policy rather than an IAM policy, the [`aws_iam_policy_document`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) data source may be used, so long as it specifies a principal. For more information about building AWS IAM policy documents with Terraform, see the [AWS IAM Policy Document Guide](https://learn.hashicorp.com/terraform/aws/iam-policy). Note: Bucket policies are limited to 20 KB in size.
```

moskyb: good suggestion 👍 have applied

justinretzolk (approved):
Looks good to me! Thank you for your contribution!

PR #22200, opened 2021-12-14
fix: increase golanci timeout to 25m

anGie44:
### Community Note

### Notes

* `Terraform Provider` Github actions workflows often error out during the `golangci-lint` job after exceeding the 20min run timeout e.g. https://github.com/hashicorp/terraform-provider-aws/runs/4513217734?check_suite_focus=true

```
N/A
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -72,4 +72,4 @@ linters-settings:
       - gomnd
 
 run:
-  timeout: 20m
+  timeout: 25m


.golangci.yml
```diff
@@ -72,4 +72,4 @@ linters-settings:
       - gomnd
 
 run:
-  timeout: 20m
+  timeout: 25m
```
anGie44, line 75: i'm assuming this is where the `20m` is defined though the error messages seem to call out the `--timeout` command line arg so maybe better suited there 🤔 ?

ewbankkit: From [the documentation](https://golangci-lint.run/usage/configuration/):

> The config file has lower priority than command-line options. If the same bool/string/int option is provided on the command-line and in the config file, the option from command-line will be used.

so changing the value in this file seems correct.

ewbankkit (approved):
LGTM :rocket:.

I guess 25 minutes is a suitable new value.
Let's go with it and see how frequently we get timeouts.

PR #22283, opened 2021-12-20
Illustrate SecretsManager secret policy with a safer example

christophetd:
### Community Note

## Context and background

Such policies are generally considered dangerous, as they allow access by any principal in the account, or outside the account that has permissions on the KMS key. Generally, it's a bad idea to illustrate security features with insecure examples, so I'm proposing this change.

### DIFF
--- a/website/docs/r/secretsmanager_secret_policy.html.markdown
+++ b/website/docs/r/secretsmanager_secret_policy.html.markdown
@@ -27,10 +27,10 @@ resource "aws_secretsmanager_secret_policy" "example" {
   "Version": "2012-10-17",
   "Statement": [
 	{
-	  "Sid": "EnableAllPermissions",
+	  "Sid": "EnableAnotherAWSAccountToReadTheSecret",
 	  "Effect": "Allow",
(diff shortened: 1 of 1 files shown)


website/docs/r/secretsmanager_secret_policy.html.markdown
```diff
@@ -27,10 +27,10 @@ resource "aws_secretsmanager_secret_policy" "example" {
   "Version": "2012-10-17",
   "Statement": [
 	{
-	  "Sid": "EnableAllPermissions",
+	  "Sid": "EnableAnotherAWSAccountToReadTheSecret",
 	  "Effect": "Allow",
 	  "Principal": {
-		"AWS": "*"
+		"AWS": "arn:aws:iam::111122223333:root"
```
breathingdust, line 33: Thanks for the contribution @christophetd! I think this makes sense. Can we change the random account id to: `123456789012` as that's what we typically use.

breathingdust (changes requested):

christophetd:
Thanks!

PR #22308, opened 2021-12-21
Fix spelling mistake in s3-api-gateway-integration example.

Hazy87:
Fix small spelling mistake in example.


anGie44 (approved):
Thank you for this fix @Hazy87 , LGTM 🚀

PR #22553, opened 2022-01-12
Improve S3 bucket name validation

rgoomar:
### Community Note

Output from testing:

```
$ make test TEST=./internal/service/s3/
go test ./internal/service/s3/  -timeout=5m
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/s3	3.382s
```

Much of the validation was derived from - https://github.com/hashicorp/terraform-provider-aws/blob/13dfabe4e55e6e4f088350898d3bb5cffdcdfb64/internal/service/s3/validate.go#L15-L44

### DIFF
--- a/internal/service/s3/bucket.go
+++ b/internal/service/s3/bucket.go
@@ -49,7 +49,16 @@ func ResourceBucket() *schema.Resource {
 				Computed:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"bucket_prefix"},
-				ValidateFunc:  validation.StringLenBetween(0, 63),
+				ValidateFunc: validation.All(
+					// Between 3 and 63 characters in length
+					validation.StringLenBetween(3, 63),
+					validation.StringMatch(regexp.MustCompile(`^[0-9a-z-.]+$`), "only lowercase alphanumeric characters and hyphens allowed"),
+					validation.StringDoesNotMatch(regexp.MustCompile(`^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$`), "must not be formatted as an IP address"),
+					validation.StringDoesNotMatch(regexp.MustCompile(`^\.`), "cannot start with a period"),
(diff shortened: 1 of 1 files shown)


rgoomar:
FWIW I had to work around this in a custom s3 bucket module with custom validators. I think since this is something that is required on the AWS provider side, the validation logic should likely live there. Without these validators, a `terraform plan` with a `bucket_name` of `some_bucket_that_isnt_valid` works just fine and then the `terraform apply` fails since the `_` character is invalid. 

```terraform
variable "bucket_name" {
  type        = string
  description = "Name of the bucket"

  validation {
    condition     = length(var.bucket_name) >= 3 && length(var.bucket_name) <= 63
    error_message = "Bucket name must be between 3 and 63 characters."
  }

  validation {
    condition     = can(regex("^[0-9a-z-.]+$", var.bucket_name))
    error_message = "Bucket name must contain only lowercase alphanumeric characters and hyphens allowed."
  }

  validation {
    condition     = !can(regex("^(?:[0-9]{1,3}\\.){3}[0-9]{1,3}$", var.bucket_name))
    error_message = "Bucket name must not be formatted as an IP address."
  }

  validation {
    condition     = !can(regex("^\\.", var.bucket_name))
    error_message = "Bucket name cannot start with a period."
  }

  validation {
    condition     = !can(regex("^.*\\.$", var.bucket_name))
    error_message = "Bucket name cannot end with a period."
  }

  validation {
    condition     = !can(regex("(\\.\\.)", var.bucket_name))
    error_message = "Bucket name must not have two periods in a row."
  }
}
```

fireflycons:
The rule for bucket names is that they must be valid for a DNS entry, therefore you should be able to do it with a single regex validation in addition to the length check. This should do it:
`^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$`
with error message something like "Bucket name must be valid internet host name."

PR #22596, opened 2022-01-14
Switch to Issue Forms

justinretzolk:
### Community Note

### Information

This PR aims to switch over to using [GitHub Issue Forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) rather than Markdown-based templates. With this implementation, there are 6 separate forms to choose from:

- Bugs
- Documentation issues
- Enhancements to existing resources/data sources
- Request for an entirely new resource, data source, or service
- Issues with the repository itself (Actions, workflow things, labels, etc)
- Other; for when the issue isn't one of the above.

Having so many distinct options allows for adding much richer contextual details and helpful links customized to a given scenario, allows specific labels to be added for each form, and potentially opens options for GitHub Actions to be created at a later date.

More information on forms may be found:

- [Creating Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms)
- [Syntax for Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)

Since it's not possible to preview forms in this repository prior to merging into `main`, I've created a [public repository](https://github.com/justinretzolk/issue-forms-test/issues/new/choose) for testing and uploaded the same configuration there. Feel free to go through the flow there to preview what the forms will look like.

### Additional Details

In addition to adding the Forms configuration itself, a few related things will change with this PR:

- Removed the Action that automatically added `needs-triage` to new issues. Forms allow us to specify a set of labels that should be automatically added to new issues, so this is no longer necessary.
- Added an Action to automatically remove `needs-triage` from internal member contributions. Since we aren't excluding these users from having `needs-triage` added in the now-removed action mentioned above, added this to retain functionality.
- Added an Action to automatically comment on newly created issues and PRs with the community note. We're not currently able to retain the community note section at the top of issues submitted by forms without a clunky workaround, so this Action helps us maintain the note while switch to Forms. Since we have to do this for issues, mirrored this change for PRs for consistency.
- "Community Note" is customized between issues and PRs.

### Rendered Issue Forms & Pull Request Comment

I've gone through the flow to create an issue of each type + a pull request to display what each type of form looks like when completed + what the community note comment looks like. Links as follows:

- [Pull Request](https://github.com/justinretzolk/issue-forms-test/pull/28) (to view the updated community note comment)
- [Bug](https://github.com/justinretzolk/issue-forms-test/issues/31)
- [Documentation](https://github.com/justinretzolk/issue-forms-test/issues/32)
- [Enhancement](https://github.com/justinretzolk/issue-forms-test/issues/33)
- [New Resource/Data Source/Service](https://github.com/justinretzolk/issue-forms-test/issues/34)
- [Repo/Meta](https://github.com/justinretzolk/issue-forms-test/issues/35)
- [Other](https://github.com/justinretzolk/issue-forms-test/issues/36)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,96 @@
...
+  - type: textarea
+    attributes:
+      label: Community Note
+      description: Please leave this note as is for the community
+      value: |
+        * Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
+        * Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
+        * If you are interested in working on this issue or have submitted a pull request, please leave a comment
+    validations:
+      required: true
```
kamilturek, line 28: Should it be a `textarea`? I suppose this note should be kept as-is for the community.

kamilturek (commented):
Looks nice 🚀

.github/ISSUE_TEMPLATE/bug_report.yml
breathingdust, line 28: This is the only part which I don't love, but it doesn't seem like the [GitHub Form Schema](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema) allows you to add content to the issue without a editable form element. Just an idea, but I wonder if instead of keeping the note as part of the the issue body, we could add a GitHub action to add the content as a comment on creation?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,96 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        * [Configuration Language](https://www.terraform.io/docs/configuration/index.html) or resource ordering issues
+        * [State](https://www.terraform.io/docs/state/index.html) and [State Backend](https://www.terraform.io/docs/backends/index.html) issues
+        * [Provisioner](https://www.terraform.io/docs/provisioners/index.html) issues
+        * [Registry](https://registry.terraform.io/) issues
+        * Spans resources across multiple providers
+
+        If you are running into one of these scenarios, we recommend opening an issue in the [Terraform core repository](https://github.com/hashicorp/terraform/) instead.
```
breathingdust, line 18:
```suggestion
        If you are running into one of these scenarios, we recommend opening an issue in the [Terraform Core repository](https://github.com/hashicorp/terraform/) instead.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,96 @@
...
+      value: |
+        * Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
+        * Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
+        * If you are interested in working on this issue or have submitted a pull request, please leave a comment
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Terraform CLI and Terraform AWS Provider Version
+      description: |
```
breathingdust, line 32: Could we add placeholder attributes to show an example of our preferred format? I think they would be useful for most of the fields.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,96 @@
...
+        Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.
+
+        For large Terraform configurations, please use a service like Dropbox and share a link to the ZIP file. For your security you can also encrypt the files using our [GPG public key](https://keybase.io/hashicorp).
+
+        [Information about code formatting](https://help.github.com/articles/basic-writing-and-formatting-syntax/#quoting-code).
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Debug Output
```
breathingdust, line 69: Currently we typically don't get `debug.log` for a lot of issues (though we probably should). Do we think this requirement will be a barrier?

breathingdust (changes requested):
Less requesting specific changes, but adding some points that should likely have team discussion. Let's discuss this week.

ewbankkit:
@justinretzolk Do you have some screen shots as to what this will look like for the participants? Thanks.

.github/ISSUE_TEMPLATE/bug_report.yml
kamilturek: Ah, sorry. I didn't realize that `markdown` elements are not submitted.

.github/ISSUE_TEMPLATE/bug_report.yml
justinretzolk: I initially put placeholder values in for most of the `textarea` fields, but there's currently a bug in Forms that means that in Safari, the newlines are removed, so the formatting was off. Because of that, I removed them for now, but I'm happy to go back and re-add them so they're there when the bug is fixed if we'd prefer.

.github/ISSUE_TEMPLATE/bug_report.yml
justinretzolk: In case we decide to go that route, looks like [this action](https://github.com/marketplace/actions/create-or-update-comment) would make it really easy to make that happen.

.github/ISSUE_TEMPLATE/bug_report.yml
justinretzolk: Added placeholders -- I verified they look fine on Firefox and figure it's better to have them and deal with the [GitHub bug](https://github.com/github/feedback/discussions/4290) until it gets fixed in Safari.

.github/ISSUE_TEMPLATE/bug_report.yml
justinretzolk: I went back and marked this as optional so as to not cause friction :+1:

justinretzolk:
@breathingdust this is otherwise complete, but something did occur to me. With this change, since issue forms doesn't have the ability to have a section of markdown that gets rendered into the resulting issue, there's a new Action to comment on issues when they're first opened to leave the community note. Since there's not a Forms equivalent for Pull Requests, that still relies on the submitted to leave the community note from the existing pull request template. Does it make more sense to also switch pull requests to having an action that comments with the community note when it's first opened, rather than having that note put in two different ways?

justinretzolk:
Got this branch dusted off after it sitting for a few months + did quite a bit of overhauling. I've updated the PR description with the relevant information.

justinretzolk:
After speaking with @breathingdust, going to get this merged in.

PR #22746, opened 2022-01-25
vpc/peering_connection Update auto_accept description to suggest the usage of the accepter when creating inter-region peering

luca-pasquali:
### Community Note

This PR does not update any functionality of the provider, however aims to fix some incomplete information by the documentation.
The `auto_accept` option can be used only for VPC peering between VPC in the same region and environment. According to the note at the top of the page it is suggested to use a combination of `aws_vpc_peering_connection` and `aws_vpc_peering_connection_accepter` for every inter-region or cross-account peering. This needs to be reflected also in the description of the `auto_accept` parameter, which would not work in case the two VPCs are not in the same region.

I believe there will be no need to run acceptance testing considering there is no change on the code.

### DIFF
--- a/website/docs/r/vpc_peering_connection.html.markdown
+++ b/website/docs/r/vpc_peering_connection.html.markdown
@@ -150,7 +150,7 @@ In addition to all arguments above, the following attributes are exported:
 
 ## Notes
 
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
This looks good to me! Thank you for your contribution! I'll work on getting this merged once the checks finish 🚀

PR #22876, opened 2022-02-01
fix lambda_config link

jurnho:
fix lambda_config link so the browser will scroll down to the lambda config section


anGie44 (approved):
Thank you for this fix @jurnho , LGTM 🚀 !

Issue #23106, opened 2022-02-10
Discussion around changes made to the `aws_s3_bucket` resource in AWS Provider version 4.0


joe-a-t:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

## Update from HashiCorp, 2022-02-25

Hi everyone,

In light of how much traction this issue has gained, and the amount of information contained within, @joe-a-t has graciously allowed us to make edits to the description of this GitHub Issue to better reflect its current state, and so that we may pin it to make it more visible. We would like to sincerely thank him for the opportunity to do so. For the sake of transparency, and to not lose any context, the original issue description and title will be retained beneath this update.

### Breakdown

On 2022-02-10, the AWS Provider team at HashiCorp released a new major version of the AWS Provider for Terraform (`v4.0.0`), which included breaking changes to the `aws_s3_bucket` resource. The update introduced several independent `aws_s3_bucket_*` resources to manipulate the configuration of S3 buckets, where these configuration options previously were arguments to the `aws_s3_bucket` resource itself. This issue spawned out of that change, has received quite a bit of valuable feedback, and is being used as a centralized meeting point for ongoing discussions around these changes.

Information around these changes may be found:

- [Upcoming changes in Version 4.0 of the AWS Provider](https://github.com/hashicorp/terraform-provider-aws/issues/20433)
- [Terraform AWS Provider 4.0 Refactors S3 Bucket Resource (blog post)](https://www.hashicorp.com/blog/terraform-aws-provider-4-0-refactors-s3-bucket-resource)
- [Terraform AWS Provider Version 4 Upgrade Guide](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade)

### Action Items

Given the amount of feedback that we received after the release, the AWS Provider team has been following this issue (as well as any other avenues of feedback we can find) and taking steps to try to alleviate some of the burden placed on operators with this release. The team continues to investigate additional avenues, and will continue to update this issue with additional information where possible. These efforts include:

**Completed**

[...]

PR #23324, opened 2022-02-22
Update v4 migration guide with more info on imports

justinretzolk:
### Community Note

Relates #23106

N/a, docs

### Information

This PR aims to provide additional clarifying information in the migration guide around the new `aws_s3_bucket_*` resources introduced in AWS Provider v4.0.0. The aim is to provide a better idea of when `terraform import` is required for these new resources with the hopes of easing operator burden when upgrading.


website/docs/guides/version-4-upgrade.html.md
```diff
@@ -193,9 +193,9 @@ Note that the provider can only resolve FIPS endpoints where AWS provides FIPS s
 
 ## S3 Bucket Refactor
 
-To help distribute the management of S3 bucket settings via independent resources, various arguments and attributes in the `aws_s3_bucket` resource
-have become **read-only**. Configurations dependent on these arguments should be updated to use the corresponding `aws_s3_bucket_*` resource.
-Once updated, new `aws_s3_bucket_*` resources should be imported into Terraform state.
+To help distribute the management of S3 bucket settings via independent resources, various arguments and attributes in the `aws_s3_bucket` resource have become **read-only**. Configurations dependent on these arguments should be updated to use the corresponding `aws_s3_bucket_*` resource in order to prevent Terraform from reporting “unconfigurable attribute” errors for read-only arguments. Once updated, it is recommended to import new `aws_s3_bucket_*` resources into Terraform state.
```
anGie44, line 196: Just an open suggestion and not necessarily what I'd encourage a user to do but do you think we should explain the other option one has and the consequences which may not be readily apparent with something like: 

```suggestion
To help distribute the management of S3 bucket settings via independent resources, various arguments and attributes in the `aws_s3_bucket` resource have become **read-only**. 

Configurations dependent on these arguments should be updated to use the corresponding `aws_s3_bucket_*` resource in order to prevent Terraform from reporting “unconfigurable attribute” errors for read-only arguments. Once updated, it is recommended to import new `aws_s3_bucket_*` resources into Terraform state. 

In the event practitioners do not anticipate future modifications to the S3 bucket settings associated with these read-only arguments or drift detection is not needed, these read-only arguments should be removed from `aws_s3_bucket` resource configurations in order to prevent Terraform from reporting “unconfigurable attribute” errors; the states of these arguments will be preserved but are subject to change with modifications made outside Terraform.
```
?

website/docs/guides/version-4-upgrade.html.md
```diff
@@ -193,9 +193,9 @@ Note that the provider can only resolve FIPS endpoints where AWS provides FIPS s
...
+
+~> **NOTE:** Each of the new `aws_s3_bucket_*` resources relies on S3 API calls that utilize a `PUT` action in order to modify the target S3 bucket. These calls follow standard HTTP methods for REST APIs, and therefor **should** handle situations where the target configuration already exists. While it is not strictly necessary to import new `aws_s3_bucket_*` resources where the updated configuration matches the configuration used in previous versions of the AWS provider, skipping this step will lead to a diff in the first plan after a configuration change indicating that any new `aws_s3_bucket_*` resources will be created, making it more difficult to determine whether the appropriate actions will be taken.
```
anGie44, line 198: maybe `therefore` instead of `therefor` in this context?
```suggestion
~> **NOTE:** Each of the new `aws_s3_bucket_*` resources relies on S3 API calls that utilize a `PUT` action in order to modify the target S3 bucket. These calls follow standard HTTP methods for REST APIs, and therefore **should** handle situations where the target configuration already exists. While it is not strictly necessary to import new `aws_s3_bucket_*` resources where the updated configuration matches the configuration used in previous versions of the AWS provider, skipping this step will lead to a diff in the first plan after a configuration change indicating that any new `aws_s3_bucket_*` resources will be created, making it more difficult to determine whether the appropriate actions will be taken.
```

anGie44 (approved):
LGTM, just a couple comments left for review

website/docs/guides/version-4-upgrade.html.md
justinretzolk: TIL! Thank you!

website/docs/guides/version-4-upgrade.html.md
justinretzolk: This seems reasonable to me given the stated goal of making the migration guide as clear as possible. I'll wait for input from additional reviewers before committing.

ewbankkit: Sounds good to me also.

gdavison (approved):
Looks good! 🚀

ewbankkit (approved):
LGTM :rocket:.

website/docs/guides/version-4-upgrade.html.md
maryelizbeth: +1 to suggested changes from @anGie44

PR #23345, opened 2022-02-23
Changes `markdownlint` rules to use names instead of codes

gdavison:
Updates `markdownlint.yml` to use rule names instead of codes.

Also removes unneeded rule exclusion and updates heading spacing in service README files


ewbankkit (approved):
LGTM :rocket:.

PR #23561, opened 2022-03-08
fix "*" key policy in aws_route53_key_signing_key example

keithrozario:
### Community Note

Output from acceptance testing: n/a

### Information

The example policy in the docs uses a Principal of `AWS = "*"` and this is generally accepted as a bad practice. There's a special warning the the [AWS Docs](https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-modifying-external-accounts.html) about this:

>  Do not set the Principal to an asterisk (*) in any key policy statement that allows permissions unless you use conditions to limit the key policy. An asterisk gives every identity in every AWS account permission to use the KMS key, unless another policy statement explicitly denies it. Users in other AWS accounts just need corresponding IAM permissions in their own accounts to use the KMS key.

This request modified the policy to mimic what the Console would create for manual KSK creation. It limits the the use of the KMS Key to the specific account, and adds conditions for Route53 as well.

### DIFF
--- a/website/docs/r/route53_key_signing_key.html.markdown
+++ b/website/docs/r/route53_key_signing_key.html.markdown
@@ -17,6 +17,8 @@ provider "aws" {
   region = "us-east-1"
 }
 
+data "aws_caller_identity" "current" {}
+
 resource "aws_kms_key" "example" {
   customer_master_key_spec = "ECC_NIST_P256"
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks good to me! Thank you very much for your contribution! :rocket:

PR #23591, opened 2022-03-09
Fix incorrect documentation for aws_subnet

m1keil:
The documentation for the `private_dns_hostname_type_on_launch` is incorrect. 
Resource-based naming can be used for ipv4 only, dual stack and ipv6-only subnets, while the legacy IP-based naming cannot be used with ipv6-only subnets.

Docs: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html#instance-naming-types

### DIFF
--- a/website/docs/r/subnet.html.markdown
+++ b/website/docs/r/subnet.html.markdown
@@ -66,7 +66,7 @@ The following arguments are supported:
     that instances launched into the subnet should be assigned
     a public IP address. Default is `false`.
 * `outpost_arn` - (Optional) The Amazon Resource Name (ARN) of the Outpost.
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
The [API documentation](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifySubnetAttribute.html) and [AWS Go SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#ModifySubnetAttributeInput) seem to still indicate that the current documentation is correct, however, based on your linked documentation, I suspect these documents are out of date. With that in mind, this looks good to me! 🚀 

note: I did push a change to switch the word "and" in where there was an ampersand before.

m1keil:
Cheers. Will leave feedback on the API docs to hopefully get it fixed there as well.

PR #23636, opened 2022-03-11
Switch GitHub Actions to use MAINTAINER_LIST secret

justinretzolk:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request

Relates #23633

Output from acceptance testing: n/a, meta PR for GitHub Actions

### Information

This PR comes after the linked PR is merged and the secret has been created by approving the resulting Terraform run. Once that has completed, this PR can be merged to switch the GitHub Actions over to using the `MAINTAINER_LIST` secret rather than a statically defined list.

PR #23640, opened 2022-03-11
Fix broken GitHub Actions after switch to MAINTAINER_LIST

justinretzolk:
### Community Note

Relates  #23636

Output from acceptance testing: n/a, relates to GitHub Actions

### Information

The linked Pull Request updated several GitHub Actions to use the new `MAINTAINER_LIST` secret to pull a list of users for an exclusion list. Unfortunately, this broke many of the workflows, so this PR aims to fix it.

There seems to be some inconsistencies, but the best I could tell, this is a result of using the `secrets` context in `jobs.<job-id>.if` definitions. Per the [context availability docs](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability) this isn't supported (though it seemed to work on some Actions, so I'm still scratching my head a little).

Because of that, this PR introduces a top level `env` definition that evaluates the `contains()` block (the same way we were doing it in `jobs.<job-id>.if` previously) and then moves the conditional to `jobs.<job-id>.steps.if`, where `env` is supported.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -11,15 +11,18 @@ on:
       - CHANGELOG.md
   pull_request_target:
 
+env:
+  IN_MAINTAINER_LIST: ${{ contains(secrets.MAINTAINER_LIST, github.actor) }}
+
 jobs:
   changes:
-    if: github.event_name == 'pull_request_target' && !contains( secrets.MAINTAINER_LIST, github.actor)
     name: Filter Changes
     runs-on: ubuntu-latest
     outputs:
       changed: ${{ steps.filter.outputs.changed }}
     steps:
       - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request_target' && env.IN_MAINTAINER_LIST == 'false'
         id: filter
         with:
           filters: |
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -7,15 +7,18 @@ on:
       - 'release/**'
   pull_request_target:
 
+env:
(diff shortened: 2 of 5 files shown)


.github/workflows/changelog.yml
```diff
@@ -11,15 +11,18 @@ on:
...
 jobs:
   changes:
-    if: github.event_name == 'pull_request_target' && !contains( secrets.MAINTAINER_LIST, github.actor)
     name: Filter Changes
     runs-on: ubuntu-latest
     outputs:
       changed: ${{ steps.filter.outputs.changed }}
     steps:
       - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request_target' && env.IN_MAINTAINER_LIST == 'false'
```
anGie44, line 25: I'm not familiar with the data type that gets returned in `IN_MATAINER_LIST` 😅  but if the equality check works here as a String, LGTM

anGie44 (approved):

.github/workflows/changelog.yml
justinretzolk: Mainly commenting in case anyone ever needs to reference this.

This is what worked in my testing, but just to back myself up, I checked back into the documentation and found [this example](https://docs.github.com/en/actions/learn-github-actions/expressions#example-returning-a-json-data-type). From what I can tell, even if the environment evaluates to a `bool`, when you reference it, it'll be a `string`. In order to use it as a `bool` (so that we could do `!env.IN_MAINTAINER_LIST`), we'd need to use `!fromJSON(env.IN_MAINTAINER_LIST)`, or something like that -- I didn't test this route since this seemed to work just as well.

Issue #20452, opened 2021-08-05
Provider produced inconsistent result after apply ECS Fargate task


GuillermoRosello:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform AWS Provider Version
* Terraform 0.14.5
* aws provider 3.38.0

### Affected Resource(s)

* aws_ecs_service.app

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp

```

### Expected Behavior

Updated tags on ECS Fargate tasks images on 13 services

### Actual Behavior

After updating 10 services the next one we got the error "Provider produced inconsistent result after apply", weirdly enough this same code was run on different clusters at the same time on diferent pipelines multiple times before without any errors.
```hcl
Error: Provider produced inconsistent result after apply

When applying changes to module.linux-app-1.aws_ecs_service.app["app11"],
provider "registry.terraform.io/hashicorp/aws" produced an unexpected new
value: Root resource was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.


##[error]PowerShell exited with code '1'.
##[error]PowerShell wrote one or more lines to the standard error stream.
##[error]
Error: Provider produced inconsistent result after apply

When applying changes to module.linux-app-1.aws_ecs_service.app["app11"],
provider "registry.terraform.io/hashicorp/aws" produced an unexpected new
value: Root resource was present, but now absent.

```
This caused a weirder error when running the pipeline for a second time "creation: InvalidParameterException: Creation of service was not idempotent." So we commented out that service re-ran pipeline it applied, but the service was still in AWS but as INACTIVE. Had to manually destroy it and uncomment it, and the apply an the service went trough.
```hcl
[...]

PR #23747, opened 2022-03-18
Fix for hashicorp/terraform-provider-aws/#20452

rwky:
When using `wait_for_steady_state` retry up to 3 times.

This is due to when a service is replaced there is a possibility the
service will return a status of INACTIVE as the AWS API returns the
status of the old service instead of the new one which hasn't fully
registered yet.

### Community Note

Relates #20452

### Description

* for use with `wait_for_steady_state`, this retries on error
* for new resources, this will ignore the `INACTIVE` state condition during read

### DIFF
--- a/internal/service/ecs/wait.go
+++ b/internal/service/ecs/wait.go
@@ -63,6 +63,7 @@ func waitCapacityProviderUpdated(conn *ecs.ECS, arn string) (*ecs.CapacityProvid
 }
 
 func waitServiceStable(conn *ecs.ECS, id, cluster string) error {
+	var err error
 	input := &ecs.DescribeServicesInput{
 		Services: aws.StringSlice([]string{id}),
 	}
@@ -71,10 +72,12 @@ func waitServiceStable(conn *ecs.ECS, id, cluster string) error {
 		input.Cluster = aws.String(cluster)
 	}
 
-	if err := conn.WaitUntilServicesStable(input); err != nil {
-		return err
+	for i := 0; i < 3; i++ {
+		if err := conn.WaitUntilServicesStable(input); err == nil {
+			return nil
+		}
 	}
-	return nil
+	return err
 }
 
 func waitServiceInactive(conn *ecs.ECS, id, cluster string) error {


anGie44:
Hi @rwky , thank you for your contribution! I'd like to get this fix in, with just a small change to possibly retry on error automatically without having to provide a for-loop with a best-guess count. From some internal logs we have, they show 
```
error: error waiting for ECS service (arn:aws:ecs:us-east-1:xxxxxxxx:service/stage-external-ecs-service/stage-mock-api-11) to become ready: ResourceNotReady: failed waiting for successful resource state
```
so we can catch this err code which is what I believe is getting returning in the for-loop you've added . Additionally, to address the linked issue of the `inconsistent result after apply` error, it looks like there's one particular line we can adjust. If possible, I'll put up a commit to this PR directly else, will open a new one to continue the work here.

rwky:
> Hi @rwky , thank you for your contribution! I'd like to get this fix in, with just a small change to possibly retry on error automatically without having to provide a for-loop with a best-guess count. From some internal logs we have, they show
> 
> ```
> error: error waiting for ECS service (arn:aws:ecs:us-east-1:xxxxxxxx:service/stage-external-ecs-service/stage-mock-api-11) to become ready: ResourceNotReady: failed waiting for successful resource state
> ```
> 
> so we can catch this err code which is what I believe is getting returning in the for-loop you've added . Additionally, to address the linked issue of the `inconsistent result after apply` error, it looks like there's one particular line we can adjust. If possible, I'll put up a commit to this PR directly else, will open a new one to continue the work here.

Would love any improvements you can make! One thing to note that my commit only fixes it if wait_for_steady_state is used, without it this can still trigger the problem, I looked into fixing it without wait_for_steady_state but it seemed pretty complex and enabling wait_for_steady_state seems a reasonable workaround. #20452 should remain open until both pathways are fixed.

anGie44:
Awesome! So i was looking at your comment here: https://github.com/hashicorp/terraform-provider-aws/issues/20452#issuecomment-1005892169 and I tentatively addressed the create behavior in https://github.com/hashicorp/terraform-provider-aws/pull/24223/files#diff-5256ebbe00ecb0cea632f285f1edc023596e6b8b70c1092d46c37188c5c2c41b by allowing the state to be `INACTIVE` for a newly created resource but I'm curious if you by any chance the full set of debug logs from that configuration run? Also are those logs produced with `wait_for_steady_state` enabled?

PR #23927, opened 2022-03-29
r/autoscaling_attachment.html.markdown: Don't use the deprecated parameter in the example

grimm26:
alb_target_group_arn is deprecated as stated in the reference, but was
still used in the example.


anGie44 (approved):
LGTM 🚀 Thank you for this change @grimm26

PR #24250, opened 2022-04-14
stack_set.operation_preferences.failure_tolerance_count can be zero

MarkCBell:
Currently, an aws_cloudformation_stack_set with:
```
  operation_preferences {
    failure_tolerance_count = 0
    max_concurrent_percentage = 100
  }
```
fails with `Validation Error: Exactly one of FailureTolerancePercentage and FailureToleranceCount must be specified since both parameters are present`. This is because currently the `FailureToleranceCount` and `FailureTolerancePercentage` attributes are both set (possibly to a default value of 0) and then one is removed if it is zero and the other is non-zero. Therefore with the `operation_preferences` block above both remain, leading to the `Validation Error` being raised. This PR removes the requirement that the other be non-zero and so allows the above `operation_preferences` block to work.

Note that although this does not apply to max_concurrent_percentage / max_concurrent_count since at least one of these must be at least one, I propose making the same change there just to keep the logic the same.

### DIFF
--- a/.changelog/24250.txt
+++ b/.changelog/24250.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_cloudformation_stack_set: Prevent `Validation` errors when `operation_preferences.failure_tolerance_count` is zero
+```
--- a/internal/service/cloudformation/stack_set_test.go
+++ b/internal/service/cloudformation/stack_set_test.go
@@ -329,6 +329,30 @@ func TestAccCloudFormationStackSet_operationPreferences(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "operation_preferences.0.region_concurrency_type", ""),
 				),
 			},
+			{
+				Config: testAccStackSetOperationPreferencesConfig(rName, 0, 3),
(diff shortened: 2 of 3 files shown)


ewbankkit (approved):
LGTM :rocket:.

```console
% make testacc TESTS=TestAccCloudFormationStackSet_operationPreferences PKG=cloudformation
TF_ACC=1 go test ./internal/service/cloudformation/... -v -count 1 -parallel 20 -run='TestAccCloudFormationStackSet_operationPreferences'  -timeout 180m
=== RUN   TestAccCloudFormationStackSet_operationPreferences
=== PAUSE TestAccCloudFormationStackSet_operationPreferences
=== CONT  TestAccCloudFormationStackSet_operationPreferences
--- PASS: TestAccCloudFormationStackSet_operationPreferences (125.20s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/cloudformation	128.933s
```

ewbankkit:
@MarkCBell Thanks for the contribution :tada: :clap:.

PR #24278, opened 2022-04-16
Fix a subtle bug with beanstalkCnamePrefixRegexp

disconnect3d:
Hey, there seems to be a subtle bug with the beanstalkCnamePrefixRegexp where it may match domains that are likely unexpected to be matched e.g. ebamazonaws.com.cn instead of eb.amazonaws.com.cn.

I've found this by accident, so please confirm if this commit/PR is correct before merging it.

### DIFF
--- a/.changelog/24278.txt
+++ b/.changelog/24278.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_elastic_beanstalk_environment: Correctly set `cname_prefix` attribute
+```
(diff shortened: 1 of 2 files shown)


disconnect3d:
FWIW regex before this PR:
image

and after this PR:
image

(Thx to regexper.com)

ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@disconnect3d Thanks for the contribution :tada: :clap:.

PR #24367, opened 2022-04-22
[WIP] Switch service labels to be dynamically generated

justinretzolk:
### Community Note

Output from acceptance testing: N/a, relates to repository labels Terraform configuration

### Information

Previously, we used a static list of service names that had to be updated manually any time a new service was added. This PR switches the configuration to use the [`github_tree`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/tree) data source to get the `tree` of the repository, then filters down to only the directories under `internal/service/` to get the list of services.

At this time, there's a `trimprefix()` that trims off `internal/service/` from the paths -- this could be switched to leave the `service/`, since that's included in the label name, but given that we already have a Terraform state that has these indexed as _only_ the service names, I've opted to keep that formatting for now. If we feel it would be better, I'm happy to modify the configuration / state file to leave `service/`.

### DIFF
--- a/infrastructure/repository/labels-service.tf
+++ b/infrastructure/repository/labels-service.tf
@@ -1,234 +1,50 @@
-#
-# For future consideration, this list could be automatically generated
-# via the AWS SDK service list.
-#
-
-variable "service_labels" {
-  default = [
-    "accessanalyzer",
-    "account",
-    "acm",
-    "acmpca",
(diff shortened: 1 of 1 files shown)


justinretzolk:
After looking at the plan, it's apparent that this approach isn't going to work. Closing.

PR #24418, opened 2022-04-26
document how to protect sensitive variables in secretsmanager_secret_version docs

jameslamb:
### Community Note

The documentation for creating an `aws_secretsmanager_secret_version` recommends using a `variable` to store a map of key-value pairs.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret_version#key-value-pairs

```hcl
variable "example" {
  default = {
    key1 = "value1"
    key2 = "value2"
  }

  type = map(string)
}

resource "aws_secretsmanager_secret_version" "example" {
  secret_id     = aws_secretsmanager_secret.example.id
  secret_string = jsonencode(var.example)
}
```

Starting in Terraform 0.14, it's possible to add `sensitive = true` to Terraform `variable`s, to prevent their values from being printed in logs and console output.

https://learn.hashicorp.com/tutorials/terraform/sensitive-variables

This PR proposes adding a small note about that to the `secret_version` documentation, to encourage users of this provider to be more protective of their secrets.

Thanks very much for this provider, and for your time and consideration.

### DIFF
--- a/website/docs/r/secretsmanager_secret_version.html.markdown
+++ b/website/docs/r/secretsmanager_secret_version.html.markdown
@@ -45,7 +45,7 @@ resource "aws_secretsmanager_secret_version" "example" {
 }
 ```
 
-In Terraform 0.14 and later, use `sensitive = true` to protect the values of that `variable` from being printed in logs and console output (see ["Protect Sensitive Input Variables"](https://learn.hashicorp.com/tutorials/terraform/sensitive-variables)).
+-> **Note:** In Terraform 0.14 and later, use `sensitive = true` to protect the values of the variable from being printed in logs and console output (see [Protect Sensitive Input Variables](https://learn.hashicorp.com/tutorials/terraform/sensitive-variables)).
 
 Reading key-value pairs from JSON back into a native Terraform map can be accomplished in Terraform 0.12 and later with the [`jsondecode()` function](https://www.terraform.io/docs/configuration/functions/jsondecode.html):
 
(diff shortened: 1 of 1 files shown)


website/docs/r/secretsmanager_secret_version.html.markdown
```diff
@@ -45,6 +45,8 @@ resource "aws_secretsmanager_secret_version" "example" {
 }
 ```
 
+In Terraform 0.14 and later, use `sensitive = true` to protect the values of that `variable` from being printed in logs and console output (see ["Protect Sensitive Input Variables"](https://learn.hashicorp.com/tutorials/terraform/sensitive-variables)).
```
justinretzolk, line 48: A couple of small formatting suggestions here. For linking out to the Learn guide (great idea, by the way!), we don't need to wrap the text in quotes. Additionally, I think that it may be worthwhile to move this up to just below line 28, and turn it into a [callout note](https://www.terraform.io/registry/providers/docs#callouts) (by adding the `->` to the beginning of the line), so that it doesn't get missed by being below the example.

```suggestion
-> **Note:** In Terraform 0.14 and later, use `sensitive = true` to protect the values of the variable from being printed in logs and console output (see [Protect Sensitive Input Variables](https://learn.hashicorp.com/tutorials/terraform/sensitive-variables)).
```

justinretzolk (changes requested):
Thank you very much for your contribution! Overall, I think this is a great addition! I left a couple of small formatting notes and a suggestion on the placement.

website/docs/r/secretsmanager_secret_version.html.markdown
jameslamb: Ah super cool, thanks for that!

justinretzolk (approved):
Looks good to me! Thank you again for your contribution! 🚀

PR #25661, opened 2022-07-02
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>


.github/workflows/issue-comment-created.yml
```diff
@@ -7,8 +7,13 @@ on:
 env:
   IN_MAINTAINER_LIST: ${{ contains(secrets.MAINTAINER_LIST, github.actor) }}
 
+permissions:
+  contents: read
+
 jobs:
   issue_comment_triage:
+    permissions:
+      contents: none
```
justinretzolk, line 16: In order to remove labels, we need write permissions

```suggestion
    permissions:
      issues: write
```

.github/workflows/lock.yml
```diff
@@ -4,8 +4,14 @@ on:
   schedule:
     - cron: '50 1 * * *'
 
+permissions:
+  contents: read
+
```
justinretzolk, line 9: I believe we can omit this entirely given the permissions declaration in the `job.lock.permissions` sections.

```suggestion
```

.github/workflows/issues.yml
```diff
@@ -7,6 +7,9 @@ on:
 env:
   IN_MAINTAINER_LIST: ${{ contains(secrets.MAINTAINER_LIST, github.actor) }}
 
+permissions:
+  contents: read
+
```
justinretzolk, line 12: This workflow uses a custom token, and so we should disable all scopes.

```suggestion
permissions: {}
```

.github/workflows/maintainer-edit.yml
```diff
@@ -3,8 +3,14 @@ on:
   pull_request_target:
     types:
       - opened
+permissions:
+  contents: read
+
```
justinretzolk, line 8: I don't believe we need content read permissions for this workflow, and permissions for the job are defined below, so we should be able to omit this.

```suggestion
```

.github/workflows/pull_requests.yml
```diff
@@ -8,6 +8,9 @@ env:
 
 jobs:
   Labeler:
+    permissions:
+      contents: read  # for actions/labeler to determine modified files
+      pull-requests: write  # for actions/labeler to add labels to PRs
```
justinretzolk, line 13: The `actions/labeler` action is using a custom secret in this case, so we don't need to give it pull request write permissions.

```suggestion
```

.github/workflows/pull_requests.yml
```diff
@@ -17,6 +20,9 @@ jobs:
           configuration-path: .github/labeler-pr-triage.yml
           repo-token: ${{ secrets.GITHUB_TOKEN }}
   NeedsTriageLabeler:
+    permissions:
+      contents: read  # for actions/labeler to determine modified files
+      pull-requests: write  # for actions/labeler to add labels to PRs
```
justinretzolk, line 25: The `actions/labeler` action is using a custom secret in this case, so we don't need to give it pull request write permissions.

```suggestion
```

.github/workflows/pull_requests.yml
```diff
@@ -27,6 +33,8 @@ jobs:
           configuration-path: .github/labeler-pr-needs-triage.yml
           repo-token: ${{ secrets.GITHUB_TOKEN }}
   SizeLabeler:
+    permissions:
+      pull-requests: write  # for bflad/pr-size-labeler to add labels & comment on PRs
```
justinretzolk, line 37: The `bflad/pr-size-labeler` action is using a custom secret in this case, so we don't need to give it pull request write permissions.

```suggestion
```

.github/workflows/stale.yml
```diff
@@ -3,8 +3,14 @@ on:
...
     - cron: "40 17 * * *"
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
```
justinretzolk, line 13: The `actions/stale` action is using a custom secret, so we don't need to grant additional permissions in this case.

```suggestion
```

.github/workflows/stale.yml
```diff
@@ -3,8 +3,14 @@ on:
   schedule:
     - cron: "40 17 * * *"
 
+permissions:
+  contents: read
+
```
justinretzolk, line 8: Given that the only job in this workflow is using a custom token, we should limit all scopes.

```suggestion
permissions: {}

```

justinretzolk (changes requested):
Thank you very much for taking the time to raise this! This is definitely something I'd like to work on getting merged, but I left a few comments I'd like to see addressed before we do so.

naveensrinivasan:
@justinretzolk   I don't have time to work on this PR now. I would request you, please, take it over. Thanks

justinretzolk:
Hey @naveensrinivasan, no problem at all! Thank you for letting me know. We'll gladly take over from here.

PR #25763, opened 2022-07-11
Doc: Update data.aws_ecs_task_definition arn attribute

ndench:
The `aws_ecs_task_definition` data source exposes the `id` attribute (not the `arn` attribute) which is actually the `ARN` of the Task Definition.

### Community Note

Relates OR Closes #0000

```

...
```

### DIFF
--- a/website/docs/d/ecs_task_definition.html.markdown
+++ b/website/docs/d/ecs_task_definition.html.markdown
@@ -65,7 +65,7 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `arn` - The ARN of the task definition
+* `id` - The ARN of the task definition
 * `family` - The family of this task definition
 * `network_mode` - The Docker networking mode to use for the containers in this task.
 * `revision` - The revision of this task definition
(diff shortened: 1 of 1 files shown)


website/docs/d/ecs_task_definition.html.markdown
```diff
@@ -65,7 +65,7 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
-* `arn` - The ARN of the task definition
+* `id` - The ARN of the task definition
```
justinretzolk, line 68:
```suggestion
* `id` - The ARN of the task definition
* `arn` - The ARN of the task definition
```

justinretzolk (changes requested):
It looks like both `id` and `arn` are exported and are both the ARN of the task definition, based on the [data source code](https://github.com/hashicorp/terraform-provider-aws/blob/135dcc40af9835b7ed7dcd3eaf408d84174813ab/internal/service/ecs/task_definition_data_source.go#L70-L71).

In other places (for example, the [`aws_ecs_cluster` resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_cluster#attributes-reference)), both `id` and `arn` are documented, so I think it makes sense to do the same here.

ndench:
Ah, good pickup @justinretzolk. I still have v3.68 so the `arn` property didn't work for me (it was added in v3.71 from the looks of it). Regardless, I've updated with your suggestion.

justinretzolk (approved):
Looks good to me! Thank you for adding the suggested changes! I'll get this merged once the checks finish up 🚀

PR #25848, opened 2022-07-17
Changing Lambda Integration Type to AWS_PROXY

fishcharlie:
According to the [AWS documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-integration-types.html) integration type should be `AWS_PROXY` for Lambda functions. This really tripped me up for a while since it seems like the example is just wrong here.

Previously was getting the following error when trying to use `AWS` like the example shows:

> Error: error creating API Gateway v2 integration: BadRequestException: Currently, an API with a protocol type of HTTP may only be associated with proxy integrations (AWS_PROXY, HTTP_PROXY)

### DIFF
--- a/website/docs/r/apigatewayv2_integration.html.markdown
+++ b/website/docs/r/apigatewayv2_integration.html.markdown
@@ -35,7 +35,7 @@ resource "aws_lambda_function" "example" {
 
 resource "aws_apigatewayv2_integration" "example" {
   api_id           = aws_apigatewayv2_api.example.id
-  integration_type = "AWS"
+  integration_type = "AWS_MOCK"
 
   connection_type           = "INTERNET"
   content_handling_strategy = "CONVERT_TO_TEXT"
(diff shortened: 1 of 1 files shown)


fishcharlie (commented):
Changing the integration type to AWS_MOCK.

website/docs/r/apigatewayv2_integration.html.markdown
```diff
@@ -35,7 +35,7 @@ resource "aws_lambda_function" "example" {
 
 resource "aws_apigatewayv2_integration" "example" {
   api_id           = aws_apigatewayv2_api.example.id
-  integration_type = "AWS"
+  integration_type = "AWS_MOCK"
```
justinretzolk, line 38: Unless I'm misunderstanding, based on the [document you linked in the PR](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-integration-types.html), it seems that this should instead be `AWS_PROXY`. It doesn't appear that `AWS_MOCK` is a valid integration type, and `MOCK` appears to be used only for testing. If I've misunderstood, I'd be happy to take another look! 🙂 

```suggestion
  integration_type = "AWS_PROXY"
```

justinretzolk (changes requested):
Thank you for taking the time to raise this PR! I had one note that I'd like us to take a look at before we get this merged in.

fishcharlie:
@justinretzolk You are correct. My bad. Just committed that change.

justinretzolk (approved):
Looks good to me! Thank you for your contribution!

PR #25894, opened 2022-07-19
tools: Update `skaff` for AWS SDK for Go v2

gdavison:
Makes a number of updates to `skaff` generated code for AWS SDK for Go v2, including

* Uses updated error type checking using `errors.As`
* The AWS SDK for Go v2 always uses context and doesn't include `WithContext` in function names
* API Clients are named `<service>.Client`


ewbankkit (approved):
LGTM :rocket:.

PR #25981, opened 2022-07-26
r/cloudtrail: Add missing dependencies to sample code

amoshimei:
In my environment, it didn't work without "depends_on", so I issue a pull request. I'm sorry for your inconvenience but thank you for your cooperation.

### DIFF
--- a/website/docs/r/cloudtrail.html.markdown
+++ b/website/docs/r/cloudtrail.html.markdown
@@ -38,6 +38,7 @@ resource "aws_s3_bucket" "foo" {
 
 resource "aws_s3_bucket_policy" "foo" {
   bucket = aws_s3_bucket.foo.id
+  depends_on = [aws_s3_bucket.foo]
   policy = <<POLICY
 {
     "Version": "2012-10-17",


justinretzolk:
Hey @amoshimei 👋 Thank you for taking the time to submit this contribution! My immediate thought is that this is unnecessary, as there is already an [implicit dependency](https://www.terraform.io/language/resources/behavior#resource-dependencies) by way of the `bucket` argument:

```hcl
bucket = aws_s3_bucket.foo.id
```

That said, I always like to follow up on things like this to see if perhaps there's something that we've overlooked. Did you encounter unexpected behavior when attempting to use the current example from the documentation?

amoshimei:
I am terribly sorry. I tested it today and it worked fine. Your indication is correct.

 The other day, when "bucket = aws_s3_bucket.foo.id" did not exist, an error occurred in terraform apply, and when "bucket = aws_s3_bucket.foo.id" was added, the build was completed. So I made a pull request. The terraform version at the time of the error was 1.1.9. 

I will Improve to show the error log and do test from the following pull request. Thank you for your confirmation.

PR #26050, opened 2022-07-29
Update outputs.tf

jack17529:
AWS has changed apiVersion to v1alpha1

### Community Note

### DIFF
--- a/examples/eks-getting-started/outputs.tf
+++ b/examples/eks-getting-started/outputs.tf
@@ -41,7 +41,7 @@ users:
 - name: aws
   user:
     exec:
-      apiVersion: client.authentication.k8s.io/v1beta1
+      apiVersion: client.authentication.k8s.io/v1alpha1
       command: aws-iam-authenticator
       args:
         - "token"


bschaatsbergen:
Hi @jack17529, thanks for raising this PR. Correct me if I'm wrong - but I think that `v1beta1` still exists doesn't it (judging by the docs here: https://kubernetes.io/docs/reference/config-api/)

jack17529:
@bschaatsbergen , that is not wrong but AWS EKS uses older versions (latest 1.22) and the aws-iam-authenticator (https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html) which they have told to install is as old as 1.21.2 
So I get this error. 
If I was using aws cli for making kubeconfig, I think v1beta1 would work. Still have to reproduce this in order to test that it works.

jack17529:
@bschaatsbergen , so yeah my hypothesis was correct if you use the most latest aws cli and generate kubeconfig using `aws eks update-kubeconfig --region <region> --name <cluster-name>`, you do get `v1beta1`.

jack17529:
@bschaatsbergen 
So now you have 2 options either change it to `v1alpha1` or completely remove kubeconfig output from the outputs.tf and tell everybody to connect using aws cli.

breathingdust:
Hi @jack17529, thanks for the contribution and sorry about the delay. From what I can tell it looks like `client.authentication.k8s.io/v1alpha1` has been [deprecated](https://github.com/kubernetes/kubernetes/blob/d060d487dcd74bc1a557f9378035696db40c64c2/CHANGELOG/CHANGELOG-1.24.md?plain=1#L1960) and removed from Kubernets 1.24 onwards. Given EKS now supports 1.26 I think its best to keep the `v1beta1` reference.

PR #26149, opened 2022-08-05
linting: Add `yamllint`

gdavison:
Adds [`yamllint`](https://yamllint.readthedocs.io/en/stable/) linting for all YAML files.

Uses the [default ruleset](https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration) with the following exceptions:

* Requires only 1 space before comments (default is 2)
* Does not require the document start marker (`---`)
* Does not limit line length
* Does not check for truthy values in keys. Otherwise this would flag the `on` key in GitHub workflows

Adds GitHub workflow and makefile rule


ewbankkit (approved):
LGTM :rocket:.

PR #26231, opened 2022-08-10
Extends `markdownlint` checks to all Markdown files

gdavison:
Extends `markdownlint` checks to all Markdown files in repo. Previously only ran against `./docs` and `./website/docs` directories.


ewbankkit (approved):
LGTM :rocket:.

`names/caps.md` is generated via `internal/generate/namescapslist`, so it looks like the template there needs to be cleaned up.

PR #26287, opened 2022-08-13
sweeper/aws_ecs_service: No longer fails silently on error

### LINKED ISSUE #25635: sweeper/aws_ecs_service: Fails silently
The sweeper for `aws_ecs_service` fails silently, rather than returning all errors encountered. It also should be updated to use the resource's delete function.

gdavison:
The sweeper for `aws_ecs_service` no longer fails silently. Runs delete operations in parallel.

Closes #25635


ewbankkit (approved):
LGTM :rocket:.

PR #26368, opened 2022-08-18
Add a note that changing resource attributes will stop the task

ankon:
We stumbled over this behavior by accident, and found it surprising, to say the least. Adding a note here in the docs hopefully catches other users that "just need to do this small change to the resource" to think about this, and plan the change accordingly.

### Community Note


breathingdust:
LGTM 🚀 Thanks @ankon. The code stops the task if it is running on before making the update call. AWS API [docs](https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Tasks.Modifying.html) state that modifying the task is only possible when the status is Stopped or Failed (not running).

PR #26441, opened 2022-08-23
lf_tag: remove values max items limit

Tonkonozhenko:


frco9:
Indeed this value is well under estimated, but the doc says that there is a soft limit to 1000 values per tag. Futhermore, when testing the AWS API and client, you'll see that you can only pass at most 50 values per API call. 

I've proposed a PR that fixes this : https://github.com/hashicorp/terraform-provider-aws/pull/26546/files

Tonkonozhenko:
@frco9 I definitely like your MR more

PR #26449, opened 2022-08-23
fix(resource-aws_lambda_invocation): allow lambda invocation without providing any value

rizkimhd:
### Community Note

Relates OR Closes #0000

```

...
```

### DIFF
--- a/internal/service/lambda/invocation.go
+++ b/internal/service/lambda/invocation.go
@@ -34,7 +34,6 @@ func ResourceInvocation() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				Default:  FunctionVersionLatest,
 			},
 			"result": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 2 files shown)


gdavison:
This requires the following changes:

- [ ] `InvokeInput.Qualifier` cannot be set to an empty string, it must be unset if empty
- [ ] Documentation
- [ ] Testing that it works with all forms of `function_name` (See https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_RequestParameters):
  - [ ] Name only
  - [ ] Name with alias
  - [ ] Full ARN
  - [ ] Partial ARN
- [ ] Testing that it works correctly with multiple Lambda function versions, including with an unpublished `$LATEST`

pvbouwel:
Good points raised by gdavison.

Probably also not great to change the default behavior since that is a breaking change? Perhaps it is an option to have the empty string signal that it wants to do an unqualified request since empty string is not a valid qualifier for AWS?

PR #26509, opened 2022-08-28
Lambda Website Example Requires Implied Zipping Step, Updated

KyMidd:
The example requires the lambda to be zipped into a file before terraform is run, which means it doesn't work when the example is copy and pasted (error that the lambda file is not found). 

I updated the example to be automation and CI/CD friendly, and move the zipping into the terraform code using the archive_file data source. This works with copy and pasting and is more useful, with Terraform doing more of the work in an intuitive way (and doesn't require implied steps). 

### Community Note

Output from acceptance testing: n/a - Web site update, didn't run local unit testing.


shaun-guth-allscripts (approved):

website/docs/r/lambda_function.html.markdown
```diff
@@ -52,9 +58,13 @@ resource "aws_lambda_function" "test_lambda" {
   handler       = "index.test"
 
   # The filebase64sha256() function is available in Terraform 0.11.12 and later
```
breathingdust, line 60: Hi @KyMidd! 👋 Thanks for the contribution! I think this is a reasonable change, but I think the extra information within the example comments makes it a bit confusing to read. I'm thinking it might make more sense to drop this qualifier comments completely from the example, and make sure that the attribute description captures the various ways to do this instead. The example can then just capture how to do it with the archive_file data source. WDYT?

breathingdust (changes requested):

website/docs/r/lambda_function.html.markdown
KyMidd: Hey @breathingdust! 👋 . Thanks for the welcome :) That sounds reasonable to me, that example information is starting to be too much. I pushed a commit to make the changes you requested, how does that look? I'll re-request your review also.

KyMidd: Hey @breathingdust, I rebased to fix a merge conflict and all tests passing. Bump for this fix to be included :)

harishbepf (approved):

harishbepf:
👍

breathingdust:
LGTM @KyMidd and really sorry for dropping the ball here 😬 Thanks for the contribution 🚀

KyMidd:
Don't mind at all, thanks for keeping the quality so high! Thanks @breathingdust 🚀

PR #26830, opened 2022-09-15
r/route53_record docs for import: remove "delegated" word as it is unncessary and confusing.

grimm26:
Also change "can" to "should" in the same sentence because you do need
to specify the set identifier in the import command or the provider
won't be able to find the record.


jar-b (approved):
LGTM 🎉

jar-b:
Thanks for your contribution @grimm26!

Issue #22197, opened 2021-12-13
sqs_managed_sse_enabled used to default to false but now defaults to true


tiagomatic:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* aws_sqs_queue

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```

### Expected Behavior

No tf plan due to underlying tf code not changing.

### Actual Behavior

This change causes issues with set-ups that don't support and conflict with `sqs_managed_sse_enabled`, which spontaneously now cause the following ambiguous error:

```
Error: error waiting for SQS Queue (...) attributes to update: SQS Queue policies are not equivalent
```

it was only after doing an attempted import that revealed the issue a bit with the following error:
```
data could not be decoded from
│ the state: unsupported attribute "sqs_managed_sse_enabled".
```

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

jufemaiz:
I'm seeing this too.

```
Terraform version: 1.1.0 on Go runtime version: go1.17.2
--
  | on linux_amd64
  | + provider registry.terraform.io/hashicorp/aws v3.69.0
  | + provider registry.terraform.io/hashicorp/random v2.3.1
  | + provider registry.terraform.io/hashicorp/template v2.2.0
```

LOG trace on apply.

```
Action=GetQueueAttributes&AttributeName.1=All&QueueUrl=https%3A%2F%2Fsqs.REGION.amazonaws.com%2FACCOUNT_ID%2FQUEUE_NAME&Version=2012-11-05
-----------------------------------------------------: timestamp=2021-12-14T10:23:36.111Z
[...]

PR #26843, opened 2022-09-16
Fix: Ensure that SqsManagedSseEnabled (sqs_managed_sse_enabled) is always passed to CreateQueue SQS API

nantiferov:
## Summary

Affected resource: `aws_sqs_queue`

Seems that since 01 of September 2022 [CreateQueue SQS API](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html) changed and if **SqsManagedSseEnabled** parameter is not passed, SQS created as encrypted by default with `aws_sqs_queue` resource.

Current logic in provider (internal/attrmap/attrmap.go, ResourceDataToAPIAttributesCreate) omit SqsManagedSseEnabled (sqs_managed_sse_enabled) if it's equal false. Which causes SQS to be created as encrypted and then it's changed to unencrypted on next apply.

This change fixes it, but probably not in a best way.
Closes: #22197

## Output from acceptance testing:

NOTE: `attrmap` package is used in couple of other services, but this change affects only SQS I believe.

```log
$ make testacc PKG=sqs
...
PASS
ok   github.com/hashicorp/terraform-provider-aws/internal/service/sqs 2314.165s
```


celik0311:
I am seeing this issue as well. When I set the resource variable to `false` I need to apply terraform 2x if the value defined is different than the default. As an example, In the case where [sqs_managed_sse_enabled](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue#sqs_managed_sse_enabled) is set to `false` terraform needs to be applied 2x, since when its created it gets set to aws' default of `true` and then the 2nd apply to set it back to `false`.

ewbankkit:
@nantiferov Thanks for the contribution :tada: :clap:.

It looks like Amazon announced this change a couple of weeks ago: https://aws.amazon.com/about-aws/whats-new/2022/10/amazon-sqs-announces-server-side-encryption-ssq-managed-sse-sqs-default/.

The best way to fix this is to mark the `aws_sqs_queue` `sqs_managed_sse_enabled` attribute as [`Computed`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed). To ensure backwards compatibility this means that Terraform will only perform drift detection of the attribute's value when present in a configuration.

ewbankkit:
@nantiferov Thanks for chasing up on this. As usual, lack of a covering acceptance test case failed to uncover the real problem.
I have made the change to always send a configured `sqs_managed_sse_enabled` value of `false` when creating an SQS Queue in https://github.com/hashicorp/terraform-provider-aws/pull/27335.

Issue #26735, opened 2022-09-09
Confusing : Cookie name being set if not needed.


everops-miked:
I have a case where the cookie name got set somehow but it is not used. 

The plan shows a drift  
`cookie_name     = "lb_cookie" -> null`

But looking at the console this is not even visible. 

This following line sets the cookie name and will cause a change that is not needed. 
https://github.com/hashicorp/terraform-provider-aws/blob/d4aae9242ad71229bc43bebe8a0145f9f0861f6a/internal/service/elbv2/target_group.go#L1040

We should only set the cookie name attribute if the stickiness["type"] == "app_cookie" as well.

Describe of the attributes shows 
`{   "Key": "stickiness.app_cookie.cookie_name",         "Value": "lb_cookie" },`

but looking at the console that attribute is not set and changing it does cause the save to be activated unless the app_cookie type is set.

timothyjrogers:
I picked this up and have a pull request open.

timothyjrogers:
@everops-miked I asked this on the PR but it probably makes more sense to ask it here.

Was this target group previously using app_cookie and updated at some point to use lb_cookie, or has it always used lb_cookie? I'm trying to determine how the cookie name could have been set to begin with on a target group using `lb_cookie`, since that shouldn't be valid.

everops-miked:
I think at some point someone manually set the app_cookie and switched to back setting the name. so start with lb_cookie and switch to app, set the name and switch back in the console.

Chris-Softrams:
Same issue.

timothyjrogers:
I apologize for the long delay, I got pulled away from this after picking it up and then just absolutely forgot about it.

I've been digging back into this for a couple hours here and I think I've convinced myself what you're seeing is actually correct behavior based on the AWS APIs. If I create a new Target Group with no stickiness and then update the settings in this order:

1. Enable stickiness w/ lb_cookie
2. Update stickiness w/ app_cookie, name=TESTCOOKIE
3. Update stickiness back to lb_cookie

And I describe that Target Group's Attributes via the AWS CLI between each step, I get back these cookie details. I've omitted the unrelated attributes for brevity.

```
//After enable stickiness w/ lb_cookie
{
    "Attributes": [
        {
            "Key": "stickiness.enabled",
            "Value": "true"
        },
        {
            "Key": "stickiness.type",
            "Value": "lb_cookie"
        },
        {
            "Key": "stickiness.app_cookie.cookie_name",
[...]

PR #26844, opened 2022-09-16
B fix elbv2 cookie name set on wrong type

timothyjrogers:
Relates #26735

Output from acceptance testing: I did not exercise acceptance testing.

### DIFF
--- a/internal/service/elbv2/target_group.go
+++ b/internal/service/elbv2/target_group.go
@@ -1037,7 +1037,9 @@ func flattenTargetGroupStickiness(attributes []*elbv2.TargetGroupAttribute) ([]i
 				m["cookie_duration"] = duration
 			}
 		case "stickiness.app_cookie.cookie_name":
-			m["cookie_name"] = aws.StringValue(attr.Value)
+			if sType, ok := m["type"].(string); !ok || sType == "app_cookie" {
+				m["cookie_name"] = aws.StringValue(attr.Value)
+			}
 		case "stickiness.app_cookie.duration_seconds":
 			if sType, ok := m["type"].(string); !ok || sType == "app_cookie" {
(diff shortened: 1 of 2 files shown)


internal/service/elbv2/target_group.go
```diff
@@ -1037,7 +1037,9 @@ func flattenTargetGroupStickiness(attributes []*elbv2.TargetGroupAttribute) ([]i
 				m["cookie_duration"] = duration
 			}
 		case "stickiness.app_cookie.cookie_name":
-			m["cookie_name"] = aws.StringValue(attr.Value)
+			if sType, ok := m["type"].(string); !ok || sType == "app_cookie" {
```
everops-miked, line 1040: this looks good but where is sType being set? are we sure that the iteration will happen in the right order?

everops-miked:
Maybe add a test case that will exercise this ?

internal/service/elbv2/target_group.go
timothyjrogers: If I'm understanding this code correctly the stickiness attributes passed into `flattenTargetGroupStickiness` are read from the live Target Group resource via the API, so at the point they are being flattened all necessary entries are guaranteed to be in the array and are just being reformatted into a map.

timothyjrogers:
> Maybe add a test case that will exercise this ?

I'm thinking through the test case now and I'm trying to decide how to set it up as its not clear how the cookie name would become set on a target group using `lb_cookie`. In your scenario from [#26735 ](https://github.com/hashicorp/terraform-provider-aws/issues/26735) was this target group previously using `app_cookie` and updated at some point to use `lb_cookie`, or has it always used `lb_cookie`?

everops-miked:
I think the app cookie was set and unset manually in the console at some point.

PR #26860, opened 2022-09-19
docs(waf): include Amazon Cognito User Pool was an option for WAF ass…

rdsedmundo:
…ociation

https://aws.amazon.com/about-aws/whats-new/2022/08/amazon-cognito-enables-native-support-aws-waf/

Tested with the existing property resource_arn and it worked.


gdavison (approved):
Thanks for the PR, @rdsedmundo 🚀

PR #19482, opened 2021-05-23
resource/aws_eks_node_group: added support for taints

david7482:
### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request

Closes #19310

Release note for [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md):
```release-notes:enhancement
resource/aws_eks_node_group: add `taints` attribute.
```

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccAWSEksNodeGroup_Taints'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./aws -v -count 1 -parallel 20 -run=TestAccAWSEksNodeGroup_Taints -timeout 180m
=== RUN   TestAccAWSEksNodeGroup_Taints
=== PAUSE TestAccAWSEksNodeGroup_Taints
=== CONT  TestAccAWSEksNodeGroup_Taints
--- PASS: TestAccAWSEksNodeGroup_Taints (1132.05s)
PASS
ok      github.com/terraform-providers/terraform-provider-aws/aws       1132.138s
...
```

This PR adds support for the `taints` resource block.
```
taints {
    key = "key1"
    value = "value1"
    effect = "NO_SCHEDULE"
}
taints {
    key = "key2"
    value = "value2"
    effect = "NO_EXECUTE"
}
```

PR #23452, opened 2022-03-02
data-source/aws_eks_node_group: Add `taints` attribute

soleares:
Adds the `taints` attribute to the `aws_eks_node_group` data source. It was added to the resource in https://github.com/hashicorp/terraform-provider-aws/pull/19482 but is missing from the data source.

### Community Note

* Please vote on this pull request by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original pull request comment to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for pull request followers and do not help prioritize the request

Output from acceptance testing:

```
$ make testacc TESTS=TestAccEKSNodeGroupDataSource PKG=eks

=== RUN   TestAccEKSNodeGroupDataSource_basic
        --- PASS: TestAccEKSNodeGroupDataSource_basic (1384.79s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/eks        1386.654s
...
```

PR #26875, opened 2022-09-20
Updating front end api of terraform provider from taint to taints

assankar:
Relates OR Closes #19482 & #23452

```

...
```
The goal is to make taints plural since it can be used up to 50 taints. The documentation and api call shows incorrect here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_node_group#taint

It is also incorrect here:
https://github.com/hashicorp/terraform-provider-aws/blob/586f11b0d227f53a4dec1963faa19dd44b1b4ce3/website/docs/r/eks_node_group.html.markdown

But it is correct here:
https://github.com/hashicorp/terraform-provider-aws/blob/586f11b0d227f53a4dec1963faa19dd44b1b4ce3/website/docs/d/eks_node_group.html.markdown

### DIFF
--- a/.changelog/26875.txt
+++ b/.changelog/26875.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+data-source/aws_eks_node_group: Modify `taint` attribute to `taints`
+```
--- a/internal/service/eks/node_group.go
+++ b/internal/service/eks/node_group.go
@@ -223,7 +223,7 @@ func ResourceNodeGroup() *schema.Resource {
 			},
 			"tags":     tftags.TagsSchema(),
 			"tags_all": tftags.TagsSchemaComputed(),
-			"taint": {
+			"taints": {
 				Type:     schema.TypeSet,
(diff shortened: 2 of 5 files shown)


md5:
Your PR would introduce a breaking change to the provider. The reason the schema element is called `taint` is because it is configured as a block. You can have multiple `taint { }` blocks.

In the case of the `eks_node_group` data source, it is called `taints` because in that case it is an output from the data source.

The different between `taint` and `taints` here is in line with the resources and data sources for other services.

It would be nice if the documentation included an example of using multiple `taint` blocks to make this more clear.

justinretzolk:
Hey @assankar 👋 Thank you for taking the time to submit this! We've discussed this, and agree with the information provided in the above comment. In this case, having the argument be named `taint` rather than `taints` is consistent with other resources, and we don't feel that this warrants making a breaking change. With that in mind, we'll close this PR.

PR #26885, opened 2022-09-20
Update rds_cluster_role_association.html.markdown to fix example typo

ejdigby:
Updated documentation example to use `iam_role.example.arn` attribute instead of `id`, as that is what is expected by the API.

When `.id` is used, the following error is thrown:
```
Error: "role_arn" (<role id>) is an invalid ARN: arn: invalid prefix
```


justinretzolk (approved):
Looks good to me! Thank you for your contribution! I'll get this merged in once the checks finish up 🚀

PR #26916, opened 2022-09-21
Preserve bug label on issues from bug template

jar-b:
This fix attempts to prevent the `bug` label automatically applied by the Bug issue template from being removed by the github `issue-labeler` action. 

By default the [`issue-labeler`](https://github.com/github/issue-labeler) action will remove existing labels if the issue body doesn't contain one of the specified regular expressions. There is an open [issue](https://github.com/github/issue-labeler/issues/12) and [PR](https://github.com/github/issue-labeler/pull/50) to add a `sync-labels` input that would optionally disable this behavior, but it has not yet been merged. As an alternative, this PR extends the `bug` regular expression list to include `[Bug]:`, the heading automatically created by the bug issue template, and sets the `include-title` input parameter to include the issue title in the regular expression check. Assuming the issue author does not replace this default title prefix, this should preserve the bug label for new issues until the more permanent fix is available.


justinretzolk (approved):
LGTM! Thank you for taking care of this! 🚀

PR #26940, opened 2022-09-22
Update cloudfront_distribution.html.markdown

sgoodluck:
corrects typo to reference "locals" instead of "local" in documentation


breathingdust:
LGTM 🚀 Thanks @sgoodluck!

PR #26943, opened 2022-09-22
include warning about services that do not use this endpoint

drewmullen:
Provide a warning for users of `aws_organizations_organization` resource. Not all services support enablement across the organization using `EnableAWSServiceAccess` and it may cause strange errors. For example: adding the ipam service principal to `aws_service_access_principals` will yield errors like: 

> The operation AllocateIpamPoolCidr is not supported. Account xxxx is not monitored by IPAM ipam-xxxxx.

The easiest way to avoid these issues is to use the specific service endpoint: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_ipam_organization_admin_account


johnsonaj (approved):
LGTM 🚀

PR #26955, opened 2022-09-23
resource/aws_iam_role: Add sweeper dependency on `aws_iot_topic_rule_destination`

gdavison:
If the associated IAM role is deleted before a `aws_iot_topic_rule_destination`, it may get stuck in `Deleting` status. Add dependency.


ewbankkit (approved):
LGTM :rocket:.

PR #26956, opened 2022-09-23
resource/aws_network_interface : Add sweeper dependency on `aws_db_proxy`

gdavison:
`aws_db_proxy` creates an ENI, so it must be swept before `aws_network_interface`


ewbankkit (approved):
LGTM :rocket:.

Issue #10835, opened 2019-11-11
Terraform codepipeline action block doesnt support region to run cross-region lambda function


mohitdayalzomato:
The issue is that the action block for a particular stage in AWS Codepipeline as specified here: https://www.terraform.io/docs/providers/aws/r/codepipeline.html#action doesnt supports region key. While integrating cross-region lambda function to be invoked as a codepipeline stage, it is mandatory. The region needs to be specified as a key value in the codepipeline actions. The json reference of the same is given here: https://docs.aws.amazon.com/codepipeline/latest/userguide/action-reference-Lambda.html 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* aws_XXXXX

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

gdavison:
Thanks for taking the time to submit this issue.

This documentation provides some more background on cross-region actions: https://docs.aws.amazon.com/codepipeline/latest/userguide/actions-create-cross-region.html

bflad:
@gdavison is this a duplicate of #6871 (or related)?

gdavison:
Yes,addressing #6871 would address this one, too. I'm going to close this ticket

PR #26959, opened 2022-09-23
docs(iam_access_key): add instructions about the expected `pgp_key` format

Horgix:
Relates to #10835

When providing a base-64 encoded PGP public key as `pgp_key` for the `iam_access_key` resource, only a "raw" export is accepted - not an ASCII "armored" one with `BEGIN` and `END` markers. This commit make this explicit for a better user experience, given that previously users would just get a `invalid data: tag byte does not have MSB set` cryptic error as reported in https://github.com/hashicorp/terraform/issues/10835

This documentation addition was suggested in 2017 but never got contributed until now: https://github.com/hashicorp/terraform/issues/10835#issuecomment-348593664

### DIFF
--- a/website/docs/r/iam_access_key.html.markdown
+++ b/website/docs/r/iam_access_key.html.markdown
@@ -67,7 +67,7 @@ output "aws_iam_smtp_password_v4" {
 
 The following arguments are supported:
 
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@Horgix Thanks for the contribution :tada: :clap:.

Issue #25173, opened 2022-06-04
r/aws_security_group_rule create_before_destroy fails when changing CIDRs


Nuru:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform AWS Provider Version

```
Terraform v1.2.0
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v4.17.1
```

### Affected Resource(s)

* aws_security_group_rule

### Terraform Configuration Files

<details><summary>Click to reveal</summary>

```hcl
variable "ipv6_allowed" {
  type    = bool
  default = false
}

variable "other_vpc_allowed" {
  type    = bool
  default = false
}

provider "aws" {
  region = "us-west-2"
}

resource "aws_vpc" "default" {
  cidr_block = "10.9.8.0/24"

  assign_generated_ipv6_cidr_block = true
}

resource "aws_security_group" "default" {
  vpc_id = aws_vpc.default.id
}

resource "aws_security_group_rule" "default" {
  security_group_id = aws_security_group.default.id
  type              = "ingress"
  from_port         = 0
  to_port           = 0 # [sic] from and to port ignored when protocol is "-1", warning if not zero
  protocol          = "-1"
  description       = "Allow ingress from VPC"
  cidr_blocks       = concat([aws_vpc.default.cidr_block], var.other_vpc_allowed ? ["10.1.2.0/24"] : [])
  ipv6_cidr_blocks  = var.ipv6_allowed ? [aws_vpc.default.ipv6_cidr_block] : []

  lifecycle {
    create_before_destroy = true
  }
}
```

</details>

### Expected Behavior

I want to use `create_before_destroy = true` for Security Group Rules so that I can avoid the service interruption that comes when existing rules are destroyed before the new rules are created. I expect the provider to be smart enough to avoid trying to create an existing rule.

### Actual Behavior

Whenever an CIDR is added or removed from a rule, the apply fails if an existing CIDR is retained.

```
│ Error: [WARN] A duplicate Security Group rule was found on (sg-01b3487e05d48a6a3). This may be
[...]

Issue #20104, opened 2021-07-08
Resource Identifiers and Tags for VPC Security Group Rules


ewbankkit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

> Amazon Virtual Private Cloud (VPC) now supports resource identifiers and tags for security group rules. You can now manage VPC security group rules using the assigned rule IDs and resource tags.

### New or Affected Resource(s)

* [`aws_security_group`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group)
* [`aws_security_group_rule`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule)
* [`aws_default_security_group`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_security_group)

### Potential Terraform Configuration

```hcl
resource "aws_security_group_rule" "example" {
  tags = {
    Env = "prod"
  }
}
```

### References

[Announcement](https://aws.amazon.com/about-aws/whats-new/2021/07/amazon-ec2-adds-resource-identifiers-tags-vpc-security-groups-rules/).
[Blog post](https://aws.amazon.com/blogs/aws/easily-manage-security-group-rules-with-the-new-security-group-rule-id/).

Requires [AWS SDK v1.39.2](https://github.com/aws/aws-sdk-go/releases/tag/v1.39.2):

* #20102

shortcut-opsadmin:
Been over a year. Any way we could get tags on these resources?

virgofx:
Adding tags to the above specified resources would be incredibly useful. Seems like this would be an easy fix. Would love to see this get implemented (Been over a two years now!)

ewbankkit:
@virgofx Thanks for your (and everyone else's) interest in the issue.

The delay in addressing this is best summarized in a section from an internal design document we are working on:

### Background

[...]

PR #27005, opened 2022-09-28
security_group_rule: explictly state that a single resource can manage many security group rules.

aidanmelen:
Explicitly document that a single security group rule resource may result in the EC2 API creating one or many security group rule(s) depending on whether multiple source(s)/destination(s) are provided or whether a source/destination has a list containing multiple elements.

### Description

The security group rule resource explicit states that it "Represents a single ingress or egress group rule" which is false.

For example:

```hcl
resource "aws_security_group_rule" "example" {
  description              = "A single rule resource will create a total of three ingress rules in AWS"
  type                     = "ingress"
  security_group_id        = "sg-123456"
  from_port                = 443
  to_port                  = 443
  protocol                 = "tcp"
  cidr_blocks              = ["10.0.1.0/24", "10.0.2.0/24"] # this will create two ingress rules
  source_security_group_id = "sg-123456789"                 # and this will another
}
```

### Relations

https://github.com/hashicorp/terraform-provider-aws/issues/25173

### DIFF
--- a/website/docs/r/security_group_rule.html.markdown
+++ b/website/docs/r/security_group_rule.html.markdown
@@ -8,19 +8,17 @@ description: |-
 
 # Resource: aws_security_group_rule
 
-Provides a security group rule resource. Represents one or many `ingress` or
-`egress` group rule(s), which can be added to external Security Groups.
+Provides a security group rule resource. Represents a single `ingress` or
+`egress` group rule, which can be added to external Security Groups.
 
 ~> **NOTE on Security Groups and Security Group Rules:** Terraform currently
-provides both a standalone Security Group Rule resource (a one or many `ingress` or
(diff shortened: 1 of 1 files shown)


ewbankkit:
Relates https://github.com/hashicorp/terraform-provider-aws/issues/20104.

aidanmelen:
I think the `(s)` in `source(s)/destination(s)` triggered the spelling check to fail. I also could have spelled something wrong. The action did not specify the specific failure.

aidanmelen:
@ewbankkit I simplified my change. the linter should pass now.

aidanmelen:
@ewbankkit do you have any idea why this GA check is failing? This seems like a really small change and should be failing

breathingdust:
LGTM 🚀 Thanks @aidanmelen!

PR #27008, opened 2022-09-28
update apigatewayv2_deployment.html.markdown

karlpokus:
### Description
`api_id` in the 1st example was pointing to a route instead of an api


ewbankkit (approved):
LGTM :rocket:.

PR #27115, opened 2022-10-05
changes typo in bucket replication parameter

denniswebb:
This should be "name" and not "ARN". The examples reflect this already.

### Description
Documentation states `bucket` should be the ARN of the bucket, when it should be the name. This can be seen in the examples where `bucket.id` is passed as the value.


ewbankkit:
@denniswebb Thanks for the contribution :tada: :clap:.

praneetloke:
I just happened to notice this. But I _think_ the docs may have been correct already. There are two `bucket` properties at in the replication configuration resource. The top-level `bucket` property seems to take the _source_ bucket's name, i.e. `bucket.id` and there is a `bucket` property under the `destination` object which does take the ARN of the _destination_ bucket, at least as per the examples in the doc.

PR #27151, opened 2022-10-07
[Docs]: Change import help-text from 'id' to 'arn'

partlythomas:
### Description
Minor change to documentation of resource aws_ram_resource_share - The import help-text stated that `id` of the resource share should be used when importing, but the example code below clearly uses the complete `arn` instead.
Verified that `arn` is in fact the correct value to use when importing these resources.


breathingdust:
LGTM 🚀 Thanks @partlythomas!

PR #27190, opened 2022-10-11
Remove use of `element()` function with `aws_ses_domain_dkim` documentation example

magnetikonline:
### Description

Dropping use of [`element()` function](https://www.terraform.io/language/functions/element) with `aws_ses_domain_dkim` example of setup of Route53 DKIM records, in place of just the built-in index syntax.

Provides a simpler example to reason with.

`element()` provides "wrap around" - but that won't help here - as you'd end up with conflicting `CNAME` DNS names anyway - which would fail the AWS API on `terraform apply`.


breathingdust:
LGTM 🚀 Thanks @magnetikonline!

magnetikonline:
Thanks @breathingdust 👍

PR #27262, opened 2022-10-15
resource/aws_config_configuration_recorder add retention period

### LINKED ISSUE #13305: AWS Config Set Retention
_This issue was originally opened by @tbo-246 as hashicorp/terraform#24944. It was migrated here as a result of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._
    
<hr>
    
### Current Terraform Version
```
v0.12.24
```

### Use-cases

It is now possible to set the retention policy in days for the AWS Config Service. This is a good way of keeping costs down and there is not currently a way to configure this with Terraform. 

### Attempted Solutions

I have checked all arguments in the resource [config](https://www.terraform.io/docs/providers/aws/r/config_configuration_recorder.html#) resources and none have a retention period setting. I purposefully set this option in the AWS console hoping for Terraform to detect code drift and give me a clue where the setting might be however the setting was ignored. 

### Proposal

Introduce an argument in the aws_config_configuration_recorder resource called "retention_period". It might look something like:

```
resource "aws_config_configuration_recorder" "example" {
  name     = "example"
  role_arn = aws_iam_role.config.arn
  retention_period = 365
  recording_group {
    all_supported = true
    include_global_resource_types = true
  }
}
```

mtt88:
### Description

Adds retention period to aws_config_configuration_recorder as per proposed solution and discussion on https://github.com/hashicorp/terraform-provider-aws/issues/13305

Some observations:

* There is no explicit relationship between a retention period and configuration recorder. However, the retention period is only
used to control the retention of configuration captured under a configuration recorder.
* AWS only lets us create a single retention period and configuration recorder per region, so whilst an explicit relationship does not exist a relationship _does_ exist by this virtue.
    * The AWS Console UI shows/groups the retention period under the other configuration recorder settings.
* After a retention configuration is created its given the name 'default'. It is not possible to change this name and all further
interactions with the retention configuration resource require it to be referenced by this name attribute.

Are we happy adding retention period to the existing `aws_config_configuration_recorder` resource based on this behaviour? Please let me know if we would prefer to create this as a seperate resource and I can refactor this PR to reflect that. 

### Relations

Closes #13305

### References

https://docs.aws.amazon.com/config/latest/APIReference/API_PutRetentionConfiguration.html
https://docs.aws.amazon.com/config/latest/APIReference/API_DeleteRetentionConfiguration.html
https://docs.aws.amazon.com/config/latest/APIReference/API_ConfigurationRecorder.html

https://docs.aws.amazon.com/config/latest/APIReference/API_RetentionConfiguration.html

```
 make testacc TESTS=TestAccConfigServiceConfigurationRecorder_allParams PKG=configservice
TF_ACC=1 go test ./internal/service/configservice/... -v -count 1 -parallel 20 -run='TestAccConfigServiceConfigurationRecorder_allParams'  -timeout 180m
=== RUN   TestAccConfigServiceConfigurationRecorder_allParams
--- PASS: TestAccConfigServiceConfigurationRecorder_allParams (48.68s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/configservice      48.738s
```

### DIFF
--- a/.changelog/27262.txt
+++ b/.changelog/27262.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_config_configuration_recorder: Add retention period
+```
--- a/internal/service/configservice/configservice_test.go
+++ b/internal/service/configservice/configservice_test.go
@@ -24,7 +24,7 @@ func TestAccConfigService_serial(t *testing.T) {
 		},
 		"ConfigurationRecorder": {
 			"basic":       testAccConfigurationRecorder_basic,
-			"allParams":   testAccConfigurationRecorder_allParams,
+			"allParams":   TestAccConfigurationRecorder_allParams,
 			"importBasic": testAccConfigurationRecorder_importBasic,
 		},
 		"ConformancePack": {
--- a/internal/service/configservice/configservice_test.go
+++ b/internal/service/configservice/configservice_test.go
@@ -23,9 +23,8 @@ func TestAccConfigService_serial(t *testing.T) {
 			"importBasic":  testAccConfigurationRecorderStatus_importBasic,
 		},
 		"ConfigurationRecorder": {
-			"basic":       testAccConfigurationRecorder_basic,
-			"allParams":   TestAccConfigurationRecorder_allParams,
-			"importBasic": testAccConfigurationRecorder_importBasic,
+			"basic":     TestAccConfigServiceConfigurationRecorder_basic,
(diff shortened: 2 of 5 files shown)


sherifkayad:
can we maybe get this one merged? looks like an easy win to me and already required by so many people who put a thumbs up on the issue #13305  (not the PR).

PR #27266, opened 2022-10-16
Explicit options on AWS Directory docs for size, edition and type

### LINKED ISSUE #16487: Documentation Update for aws_directory_service_directory
I previously created a bug report for this but closed it before anyone could triage it. The short version is that the documentation for the aws_directory_service_directory module could use some updates. Specifically, the "Size" parameter is ignored in some cases because the "Edition" parameter defaults to "Enterprise" which has no "Small" option.

Details here:
https://github.com/hashicorp/terraform-provider-aws/issues/15839#issue-729919355

abelgvidal:
…_service_directory

### Description
When create an aws_directory_service_directory the fields `edition`, `size` have some exclusions depending on the value of `type`. For example, if `type` is `MicrosoftAD`, field `size` is ignored, and only if `type` is not `MicrosoftAD`, `edition` fields should be used.

### Relations
Closes #16487

### References
[directory cli command](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ds/create-microsoft-ad.html)

### DIFF
--- a/website/docs/r/directory_service_directory.html.markdown
+++ b/website/docs/r/directory_service_directory.html.markdown
@@ -126,7 +126,7 @@ The following arguments are supported:
 
 * `name` - (Required) The fully qualified name for the directory, such as `corp.example.com`
 * `password` - (Required) The password for the directory administrator or connector user.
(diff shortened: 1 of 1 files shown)


breathingdust:
LGTM 🚀 Thanks @abelgvidal!

Issue #27192, opened 2022-10-11
[Docs] aws_vpc_endpoint: proposal to add behavior when service_name is null


uh-zz:
### Description

Hello! This would be first contribution. 

The following document describes `service_name` as follows.

>  For AWS services the service name is usually in the form com.amazonaws.<region>.<service>

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint

As noted, the format of the examples shown in the documentation is correct.

```tf
resource "aws_vpc_endpoint" "s3" {
  vpc_id       = aws_vpc.main.id
  service_name = "com.amazonaws.us-west-2.s3"
}
```

In contrast, you can see that the following module does not specify service_name, but this module successfully creates the service_name as formatted

```tf
module "endpoints" {
  source = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"

  endpoints = {
    s3 = {
      # interface endpoint
      service             = "s3"
      tags                = { Name = "s3-vpc-endpoint" }
    },
   }
}
```

https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/master/modules/vpc-endpoints#input_security_group_ids

I am aware that this works correctly because it handles the following

https://github.com/hashicorp/terraform-provider-aws/blob/19574e39cbbb44a59b5809697030e54daf15c990/internal/service/ec2/vpc_endpoint_service_data_source.go#L107-L111

So, I thought I would suggest that we add the following note to the document.

> [service_name](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint#service_name) - (Required) The service name. For AWS services the service name is usually in the form com.amazonaws.<region>.<service> (the SageMaker Notebook service is an exception to this rule, the service name is in the form aws.sagemaker.<region>.notebook). If service name is null, then service name is set in the above format from region and service.

### References

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint

### Would you like to implement a fix?

Yes

PR #27284, opened 2022-10-18
[Docs] aws_vpc_endpoint: proposal to add behavior when service_name is null

uh-zz:
### Description

Added to the description of the service_name field of the vpc endpoint.

### Relations

Relates #27192

### DIFF
--- a/website/docs/d/vpc_endpoint.html.markdown
+++ b/website/docs/d/vpc_endpoint.html.markdown
@@ -33,7 +33,7 @@ The given filters must match exactly one VPC endpoint whose data will be exporte
 
 * `filter` - (Optional) Custom filter block as described below.
 * `id` - (Optional) ID of the specific VPC Endpoint to retrieve.
-* `service_name` - (Optional) Service name of the specific VPC Endpoint to retrieve. For AWS services the service name is usually in the form `com.amazonaws.<region>.<service>` (the SageMaker Notebook service is an exception to this rule, the service name is in the form `aws.sagemaker.<region>.notebook`).
+* `service_name` - (Optional) Service name of the specific VPC Endpoint to retrieve. For AWS services the service name is usually in the form `com.amazonaws.<region>.<service>` (the SageMaker Notebook service is an exception to this rule, the service name is in the form `aws.sagemaker.<region>.notebook`). If service name is null, then service name is set in the above format from <region> and <service>.
 * `state` - (Optional) State of the specific VPC Endpoint to retrieve.
 * `tags` - (Optional) Map of tags, each pair of which must exactly match
   a pair on the specific VPC Endpoint to retrieve.


justinretzolk:
Hey @uh-zz 👋 Thank you for taking the time to raise this! I read over your description in #27192, and a couple of things stuck out to me.

It looks like you've made an update to the [`aws_vpc_endpoint` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc_endpoint) documentation, while in the issue, it seemed as though you were talking about updating the [`aws_vpc_endpoint` resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint) documentation.

Looking at the module that you referenced, the `service_name` argument of the `aws_vpc_endpoint` resource is being set by first using the [data source to retrieve the service name](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/6a3a9bde634e2147205273337b1c22e4d94ad6ff/modules/vpc-endpoints/main.tf#L5-L19) based on the user input, and the [passing the data from the data source to the `service_name` argument](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/6a3a9bde634e2147205273337b1c22e4d94ad6ff/modules/vpc-endpoints/main.tf#L25).

Based on this information, I don't believe we'll be able to merge this pull request. I'll leave it open for now so that we can discuss further, in case I've missed something.

uh-zz:
Hi @justinretzolk 👋🏼 
Thanks for review!

> It looks like you've made an update to the [aws_vpc_endpoint data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc_endpoint) documentation, while in the issue, it seemed as though you were talking about updating the [aws_vpc_endpoint resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint) documentation.

You're right, it was the `resource` documentation I meant to mention! (I was modifying the `data` docs by mistake.:sweat:)

> Looking at the module that you referenced, the service_name argument of the aws_vpc_endpoint resource is being set by first using the [data source to retrieve the service name](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/6a3a9bde634e2147205273337b1c22e4d94ad6ff/modules/vpc-endpoints/main.tf#L5-L19) based on the user input, and the [passing the data from the data source to the service_name argument](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/6a3a9bde634e2147205273337b1c22e4d94ad6ff/modules/vpc-endpoints/main.tf#L25).

You are right here too, this process is getting the service name from the data source, but when used from the module, the service name is not entered.

The module is configured by the user as follows.

https://github.com/terraform-aws-modules/terraform-aws-vpc/tree/6a3a9bde634e2147205273337b1c22e4d94ad6ff/modules/vpc-endpoints

```tf
module "endpoints" {
  source = "terraform-aws-modules/vpc/aws//modules/vpc-endpoints"

  vpc_id             = "vpc-12345678"
  security_group_ids = ["sg-12345678"]

  endpoints = {
    s3 = {
      # interface endpoint
      service             = "s3"
      tags                = { Name = "s3-vpc-endpoint" }
    },
    dynamodb = {
      # gateway endpoint
      service         = "dynamodb"
      route_table_ids = ["rt-12322456", "rt-43433343", "rt-11223344"]
      tags            = { Name = "dynamodb-vpc-endpoint" }
    },
    sns = {
      service    = "sns"
      subnet_ids = ["subnet-12345678", "subnet-87654321"]
      tags       = { Name = "sns-vpc-endpoint" }
    },
    sqs = {
      service             = "sqs"
      private_dns_enabled = true
      security_group_ids  = ["sg-987654321"]
      subnet_ids          = ["subnet-12345678", "subnet-87654321"]
      tags                = { Name = "sqs-vpc-endpoint" }
    },
  }

  tags = {
    Owner       = "user"
    Environment = "dev"
  }
}
```

If the user of the module omits the service name in this way, the behavior is as mentioned in the previous issue. (If the `service_name` is omitted, the `service_name` is set from the `service` and `region`)

https://github.com/hashicorp/terraform-provider-aws/blob/19574e39cbbb44a59b5809697030e54daf15c990/internal/service/ec2/vpc_endpoint_service_data_source.go#L107-L111

So, as long as the `service` and `region` are properly configured, there is no problem, but I thought that users of the module might be confused by the fact that it works without the required `service_name` when they refer to [the documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint), so I have created an issue and a pull request:+1:

justinretzolk:
Hey @uh-zz 👋 I think we have a bit of a disconnect here. As far as the `aws_vpc_endpoint` resource is concerned, the `service_name` argument [is required](https://github.com/hashicorp/terraform-provider-aws/blob/8750e4008a0fb662c6c54f82e94af59cb6b0e8b5/internal/service/ec2/vpc_endpoint.go#L137-L141), and cannot be inferred by setting the `service` argument (this resource actually doesn't take a `service` argument).

It seems like maybe your intent is to update the documentation for the module that you're referencing to indicate that the `service_name` isn't required when using the module, so long as the `service` is supplied. To do so, you'll want to open a pull request to the [module's repository](https://github.com/terraform-aws-modules/terraform-aws-vpc), rather than for the resource itself.

uh-zz:
Hi @justinretzolk :wave:
Thanks for comment.

I have a feeling that you are right, and that it would be better to put the issue in the module instead of updating the document here:)

So you may close this pull request.
If I find another problem, I will create a pull request. 👋🏼

PR #27296, opened 2022-10-18
skaff: Update datasource `ErrorCheck` generation

jar-b:
This change brings the generation of the `ErrorCheck` field in data source acceptance tests into alignment with how the same field is generated for resources:

https://github.com/hashicorp/terraform-provider-aws/blob/238ded41ee735e9a9f0fd23569c4d80866961624/skaff/resource/resourcetest.tmpl#L70-L72

https://github.com/hashicorp/terraform-provider-aws/blob/238ded41ee735e9a9f0fd23569c4d80866961624/skaff/resource/resourcetest.tmpl#L183-L187


ewbankkit (approved):
LGTM :rocket:.

PR #27354, opened 2022-10-20
Make documentation clearer on the list of endpoint types

pierskarsenbarg:
### Description

The docs say that this is a list, but it's a comma-separated list in a string so I've tried to make that less confusing.


jar-b:
Hi @pierskarsenbarg - the `types` attribute should be a list of strings according to the schema definition:

https://github.com/hashicorp/terraform-provider-aws/blob/00f7164cc6e1c39af6399d5c759556c1ef2378df/internal/service/apigateway/rest_api.go#L82-L91

Do you have an example configuration where you've observed the comma-separated value to work, but not a list?

pierskarsenbarg:
Sorry, that's my bad, it's the wrong resource. I'll fix it.

PR #27376, opened 2022-10-21
Provide multi step example for aws_transfer_workflow resource

hozzer:
### Description
Create an example to demonstrate how multiple steps can be used within the `aws_transfer_workflow` resource.

### References
[Example question](https://discuss.hashicorp.com/t/aws-transfer-workflow-resource-syntax-example-for-multiple-steps/44455) on the forums suggesting clarification would be helpful. 
I, personally, worked backwards using `terraform import` to find out the expected syntax for multiple steps.

### DIFF
--- a/website/docs/r/transfer_workflow.html.markdown
+++ b/website/docs/r/transfer_workflow.html.markdown
@@ -73,8 +73,8 @@ The following arguments are supported:
 
 #### Copy Step Details
 
-* `name` - (Optional) The name of the step, used as an identifier.
 * `destination_file_location` - (Optional) Specifies the location for the file being copied. Use ${Transfer:username} in this field to parametrize the destination prefix by username.
+* `name` - (Optional) The name of the step, used as an identifier.
 * `overwrite_existing` - (Optional) A flag that indicates whether or not to overwrite an existing file of the same name. The default is `FALSE`. Valid values are `TRUE` and `FALSE`.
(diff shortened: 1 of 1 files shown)


website/docs/r/transfer_workflow.html.markdown
```diff
@@ -43,8 +73,8 @@ The following arguments are supported:
 
 #### Copy Step Details
 
-* `destination_file_location` - (Optional) Specifies the location for the file being copied. Use ${Transfer:username} in this field to parametrize the destination prefix by username.
 * `name` - (Optional) The name of the step, used as an identifier.
+* `destination_file_location` - (Optional) Specifies the location for the file being copied. Use ${Transfer:username} in this field to parametrize the destination prefix by username.
```
justinretzolk, line 77: I'd be interested to hear your line of thought for this change. The list of arguments is currently in alphabetical order; is there a reason to consideration moving this line?

justinretzolk (commented):
This looks good overall, but I had one comment I'd like to resolve before we merge.

website/docs/r/transfer_workflow.html.markdown
hozzer: Initially it felt appropriate as all other steps started with `name`. However, it seems I'd overlooked the alphabetic approach - which I am entirely in favour of.

PR #27399, opened 2022-10-23
Update DMS endpoint.go - timestamp_column_name should default to null

asaf400:
The timestamp_column_name is not required as per AWS Documentation, and Terraform Documentation, However, a default value of an empty string, is an invalid input for the aws api <--> terraform aws provider, not including any value results in the following error when trying to create a DMS Replication Task:

```Error: error creating DMS Replication Task (task-name): InvalidParameterValueException: TimestampColumnName cannot be an empty string. status code: 400, request id REDACTED```

the default should be change to null value in order for the provider to exclude that parameter from the aws api call,
so that the only possible options for the 'TimestampColumnName' parameter are either null or any non-empty string value.

preferably this should also be validated with a ValidateFunc something like this:
```ValidateFunc: validation.Any(validation.StringIsNotEmpty, IsNull),```

however at this time, I'm unable to dig around and test a correct ValidateFunc expression,
and changing the default should help others in the future anyways..

provider-version: 4.35.0

### References
https://docs.aws.amazon.com/dms/latest/APIReference/API_S3Settings.html
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dms_endpoint#timestamp_column_name

### DIFF
--- a/internal/service/dms/endpoint.go
+++ b/internal/service/dms/endpoint.go
@@ -616,7 +616,7 @@ func ResourceEndpoint() *schema.Resource {
 						"timestamp_column_name": {
 							Type:     schema.TypeString,
 							Optional: true,
-							Default:  "",
+							Default:  null,
 						},
 						"use_csv_no_sup_value": {
 							Type:     schema.TypeBool,


asaf400:
Closing pull request, null value doesn't solve the issue,
I was under false impressions that it did..

Issue #6448, opened 2018-11-13
Modify aws_db_instance and delete aws_db_parameter_group breaks


jrobison-sb:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

```
$ terraform -v
Terraform v0.11.8
+ provider.aws v1.41.0
```

### Affected Resource(s)

* aws_db_parameter_group
* aws_db_instance

### Terraform Configuration Files

If I have a running RDS instance which is using a DB parameter group, and I want to modify the DB instance to use a different parameter group, and delete the old parameter group, I'll get a TF plan which looks like this:

```
  ~ module.my_module.aws_db_instance.application
      parameter_group_name: "pg10-foo" => "pg10-bar"

  - module.my_module.aws_db_parameter_group.foo
```

### Expected Behavior

Modify the RDS instance, then delete the parameter group which is now unused.

### Actual Behavior

It tries to delete the parameter group first, which fails because the parameter group is still in use. 

```
InvalidDBParameterGroupState: One or more database instances are still members of this parameter group pg10-foo, so the group cannot be deleted
```

If it would have done the `modify` action first on the DB instance, it would then be able to do the `destroy` action on the now unused parameter group.

### Steps to Reproduce

Run a plan which plans to modify a DB instance to change the parameter group to some other parameter group, and which also plans to delete the now unused parameter group.

bflad:
Hi @jrobison-sb 👋 Thanks for reporting this and sorry for the trouble.

Can you confirm a few pieces of information here?
* Are you using a reference such as `parameter_group_name = "${aws_db_parameter_group.foo.name}"` in your `aws_db_instance` configuration?
* Is `apply_immediately` set to `true` in the `aws_db_instance` configuration?
* What is the state of the RDS instance after this apply? Is the parameter group changed? Does it say `pending-reboot` in the console for the parameter group?

Without seeing the debug log here its a not possible to tell definitively, but my initial hunch is that the operations are happening in order, but that the parameter group change is requiring an instance reboot still, something which the `Update` function of `aws_db_instance` does not do currently.

[...]

PR #27497, opened 2022-10-26
r/aws_db_parameter_group: create_before_destroy example

jar-b:
### Description
Adds a new example to the `aws_db_parameter_group` documentation on when to apply the `create_before_destroy` lifecycle configuration.

### Relations
Relates #6448

### DIFF
--- a/website/docs/r/db_parameter_group.html.markdown
+++ b/website/docs/r/db_parameter_group.html.markdown
@@ -50,7 +50,7 @@ resource "aws_db_parameter_group" "default" {
 The [`create_before_destroy`](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#create_before_destroy)
 lifecycle configuration is necessary for modifications that force re-creation of an existing,
 in-use parameter group. This includes common situations like changing the group `name` or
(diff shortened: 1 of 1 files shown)


website/docs/r/db_parameter_group.html.markdown
```diff
@@ -43,6 +45,36 @@ resource "aws_db_parameter_group" "default" {
 }
 ```
 
+### `create_before_destroy` Lifecycle Configuration
+
+The [`create_before_destroy`](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#create_before_destroy)
+lifecycle configuration is necessary for modifications that force re-creation of an existing,
+in-use parameter group. This includes common situations like changing the group `name` or
+bumping `family` during a major version upgrade. This configuration will prevent destruction
```
bschaatsbergen, line 53: To add a bit more clarity/context to the `bumping family` 

```suggestion
bumping the `family` version during a major version upgrade. This configuration will prevent destruction
```

bschaatsbergen (approved):
LGTM :rocket:

PR #27535, opened 2022-10-28
Documentation Improvement: Make examples for S3 bucket references consistent across documentation

JTP3XP:
### Description
It looks like some of the newer documentation prefers to reference S3 bucket names using the `bucket` attribute instead of the `id` attribute. This works, but the rest of the documentation is overwhelmingly using `id`: there are currently 163 places where `id` is used and 35 where `bucket` is used.

This pull request standardizes references to S3 bucket names so that `id` is always used. That seems consistent with most S3 bucket name references and with how other resources are referenced.

This is the regex I used to find all references, which could be modified to replace all `id` references to `bucket` if that is more desirable:
`(bucket\s{0,}=[\sa-zA-Z0-9_.]{1,}\.)(bucket)$`

### DIFF
--- a/website/docs/d/elb_service_account.html.markdown
+++ b/website/docs/d/elb_service_account.html.markdown
@@ -54,7 +54,7 @@ resource "aws_elb" "bar" {
   availability_zones = ["us-west-2a"]
 
   access_logs {
-    bucket   = aws_s3_bucket.elb_logs.bucket
+    bucket   = aws_s3_bucket.elb_logs.id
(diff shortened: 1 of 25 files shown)


justinretzolk:
Hey @JTP3XP 👋 Thank you very much for taking the time to raise this and bring up the inconsistency! We agree that standardizing the references is important, and are going to discuss this internally a bit to decide between usage of `id` vs `bucket` before getting this merged in.

PR #27597, opened 2022-11-02
aws_s3_bucket: Remove optional from arn and hosted_zone_id

cgetzen:
### Description

Currently, the `arn` and `hosted_zone_id` attributes are marked as optional, hinting that they can be set within a resource block (as opposed to being solely `computed` attributes).

However, `arn` and `hosted_zone_id` should never be set with user-generated values. `arn` is derived from other attributes, and `hosted_zone_id` is looked up [here](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/s3/hosted_zones.go#L10).

### Relations
n/a

### References
https://discuss.hashicorp.com/t/providers-schema-how-to-distinguish-attributes-from-arguments/5029/3

```

...
```

### DIFF
--- a/internal/service/s3/bucket.go
+++ b/internal/service/s3/bucket.go
@@ -84,7 +84,6 @@ func ResourceBucket() *schema.Resource {
 
 			"arn": {
 				Type:     schema.TypeString,
-				Optional: true,
 				Computed: true,
 			},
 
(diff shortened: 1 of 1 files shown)


cgetzen:
This PR is from a github organization.
> Who can use this feature
People with push access to the upstream repository of a fork owned by a personal account can commit to the forked branches.

It looks like I can only allow changes from personal account fork.
I'm happy to recreate the PR on a personal fork, or apply feedback from maintainers to this PR.

gdavison (approved):
Great catch, @cgetzen! That's been there for a long time

PR #27159, opened 2022-10-08
r/aws_evidently_segment

GlennChia:
### Description

New Resource for CloudWatch Evidently Segment.

### Relations

Relates #22624

### References

- [API_CreateSegment](https://docs.aws.amazon.com/cloudwatchevidently/latest/APIReference/API_CreateSegment.html)
- [API_DeleteSegment](https://docs.aws.amazon.com/cloudwatchevidently/latest/APIReference/API_DeleteSegment.html)
- [API_GetSegment](https://docs.aws.amazon.com/cloudwatchevidently/latest/APIReference/API_GetSegment.html)

Note: Only Segment tags can be updated at the moment

### Output from Acceptance Testing

```
$ make testacc TESTARGS='-run=TestAccEvidentlySegment' PKG=evidently
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/evidently/... -v -count 1 -parallel 20  -run=TestAccEvidentlySegment -timeout 180m
=== RUN   TestAccEvidentlySegment_basic
=== PAUSE TestAccEvidentlySegment_basic
=== RUN   TestAccEvidentlySegment_description
=== PAUSE TestAccEvidentlySegment_description
=== RUN   TestAccEvidentlySegment_tags
=== PAUSE TestAccEvidentlySegment_tags
=== RUN   TestAccEvidentlySegment_disappears
=== PAUSE TestAccEvidentlySegment_disappears
=== CONT  TestAccEvidentlySegment_basic
=== CONT  TestAccEvidentlySegment_tags
=== CONT  TestAccEvidentlySegment_disappears
=== CONT  TestAccEvidentlySegment_description
--- PASS: TestAccEvidentlySegment_disappears (21.71s)
--- PASS: TestAccEvidentlySegment_basic (28.11s)
--- PASS: TestAccEvidentlySegment_description (28.15s)
--- PASS: TestAccEvidentlySegment_tags (65.95s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/evidently  66.049s

...
```

Issue #22624, opened 2022-01-17
Support CloudWatch Evidently


michimani:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Define and manage the resources associated with CloudWatch Evidently.  
I want to be able to change the ratio of each variation, the start time of the launch, etc. flexibly, so I want to be able to ignore changes made in the management console, etc. using `igunore_changes`.

### New or Affected Resource(s)

* aws_evidently_project
* aws_evidently_feature
* aws_evidently_launch
* aws_evidently_experimental

### Potential Terraform Configuration

```hcl
resource "aws_evidently_project" "evidently_project" {
  name        = "project-name"
  description = "project description"
}

resource "aws_evidently_feature" "evidently_feature" {
  name              = "feature-name"
  description       = "feature description"
  project           = evidently_project.arn
  default_variation = "variation-default"

  variations {
    variation {
      variation_name = "variation-default"
      string_value   = "white"
    }

    variation {
      variation_name = "variation-red"
      string_value   = "red"
    }

    variation {
      variation_name = "variation-green"
      string_value   = "green"
    }

    variation {
      variation_name = "variation-blue"
      string_value   = "blue"
    }

    variation {
      variation_name = "variation-rainbow"
      string_value   = "rainbow"
    }
  }

  entity_overrides {
    entity_override {
      entity_id = "rainbow"
      variation = "variation-rainbow"
    }
  }

  evaluation_strategy = "ALL_RULES"
}

resource "aws_evidently_launch" "evidently_launch" {
  name        = "launch-name"
  description = "launch description"
  project     = evidently_project.arn

  groups {
    launch_group {
      group_name  = "red-group-name"
      description = "red group description"
      feature     = evidently_feature.arn
      variation   = "variation-red"
    }

    launch_group {
      group_name  = "green-group-name"
      description = "green group description"
      feature     = evidently_feature.arn
      variation   = "variation-green"
    }

    launch_group {
      group_name  = "blue-group-name"
      description = "blue group description"
      feature     = evidently_feature.arn
      variation   = "variation-blue"
    }

[...]

PR #27650, opened 2022-11-04
docs(cwevi): id uses arn, import uses arn

GlennChia:
### Description

CloudWatch Evidently Segments use `arn` instead of `name`. Small documentation fix to align with the resource definition. Reference: https://github.com/hashicorp/terraform-provider-aws/blob/1952a1e8e3227b3b919d21e231fcd303f4603ad2/internal/service/evidently/segment.go#L117

### Relations

Relates #27159 and #22624


jar-b (approved):
LGTM 👍

PR #27656, opened 2022-11-04
remove tags field from aws_secretsmanager_secret_rotation resource

### LINKED ISSUE #27655: [Bug]: The aws_secretmanager_secret_rotation plan file includes a `tags` field but tags aren't supported
### Terraform Core Version

1.3.4

### AWS Provider Version

4.38.0

### Affected Resource(s)

The `aws_secretsmanager_secret_rotation` resource produces a plan that contains a `tags` field even though the [Rotate Secret API](https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_RotateSecret.html) does not support tagging.

This cause a problem for me because when writing policies using [open-policy-agent/conftest](https://github.com/open-policy-agent/conftest), the rule assume that if an aws resource has a `tags` field in the plan output, then the resource supports tagging. Since this resource expose the `tags` field but doesn't support tagging my policy fails.

My work around it to add `aws_secretsmanager_secret_rotation` resource to a exception list in my policy code. However, it would be get if I could rely on the presence of a `tags` field to determine when an aws resource supports tagging; Rather than keeping a whitelist/blacklist.

### Expected Behavior

I expect that the plan output for an `aws_secretsmanager_secret_rotation` does not include a `tags` field.

### Actual Behavior

The plan output for the `aws_secretsmanager_secret_rotation` resource does include a `tags` field.

_No response_

```hcl
terraform {
  required_providers {
    aws = {
      source = "hashicorp/aws"
      version = "4.38.0"
    }
  }
}

provider "aws" {
  region = "us-east-1"
}

resource "aws_secretsmanager_secret_rotation" "example" {
  secret_id           = "my-secret-id"
  rotation_lambda_arn = "arn:aws:lambda:::my_lambda"

  rotation_rules {
    automatically_after_days = 30
  }
}
```

### Steps to Reproduce

The the terraform code above in to a file named `main.tf`

# create a plan json
```bash
terraform init
terraform plan -out plan.bin
terraform show -json plan.bin | tee plan.json
```

### Debug Output

The terraform code above produces this snippet of plan output. Notice the `tags` field.

❯ jq '.resource_changes' plan.json
```bash
[
  {
    "address": "aws_secretsmanager_secret_rotation.example",
    "mode": "managed",
    "type": "aws_secretsmanager_secret_rotation",
    "name": "example",
    "provider_name": "registry.terraform.io/hashicorp/aws",
    "change": {
      "actions": [
        "create"
      ],
      "before": null,
      "after": {
        "rotation_lambda_arn": "arn:aws:lambda:::my_lambda",
        "rotation_rules": [
          {
            "automatically_after_days": 30
          }
        ],
        "secret_id": "my-secret-id",
        "tags": null
      },
      "after_unknown": {
        "id": true,
        "rotation_enabled": true,
        "rotation_rules": [
          {}
        ]
      },
      "before_sensitive": false,
      "after_sensitive": {
        "rotation_rules": [
          {}
        ]
      }
    }
  }
]
```

### Panic Output

_No response_

### Important Factoids

_No response_

### References

_No response_

Yes

frek818:
### Description

This change removes the useless `tags` argument from the `aws_secretsmanager_secret_rotation` resource.

### Relations

Closes #27655

```
$ make testacc TESTARGS='-run=TestAccSecretsManagerSecretRotation' PKG=secretsmanager
TF_ACC=1 go test ./internal/service/secretsmanager/... -v -count 1 -parallel 20  -run=TestAccSecretsManagerSecretRotation -timeout 180m
=== RUN   TestAccSecretsManagerSecretRotationDataSource_basic
=== PAUSE TestAccSecretsManagerSecretRotationDataSource_basic
=== RUN   TestAccSecretsManagerSecretRotation_basic
=== PAUSE TestAccSecretsManagerSecretRotation_basic
=== CONT  TestAccSecretsManagerSecretRotationDataSource_basic
=== CONT  TestAccSecretsManagerSecretRotation_basic
--- PASS: TestAccSecretsManagerSecretRotation_basic (53.01s)
--- PASS: TestAccSecretsManagerSecretRotationDataSource_basic (54.90s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/secretsmanager	57.662s

```

### DIFF
--- a/.changelog/27656.txt
+++ b/.changelog/27656.txt
@@ -1,0 +1,4 @@
+```release-note:breaking-change
+resource/aws_secretsmanager_secret_rotation: Remove unused `tags` attribute
+```
+
--- a/internal/service/secretsmanager/secret_rotation.go
+++ b/internal/service/secretsmanager/secret_rotation.go
@@ -11,7 +11,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-aws/internal/conns"
-	tftags "github.com/hashicorp/terraform-provider-aws/internal/tags"
 	"github.com/hashicorp/terraform-provider-aws/internal/tfresource"
 )
 
@@ -52,7 +51,6 @@ func ResourceSecretRotation() *schema.Resource {
 					},
 				},
 			},
-			"tags": tftags.TagsSchema(),
 		},
 	}
 }


jar-b (approved):
LGTM 🎉 

```console
$ make testacc TESTARGS='-run=TestAccSecretsManagerSecretRotation' PKG=secretsmanager
TF_ACC=1 go test ./internal/service/secretsmanager/... -v -count 1 -parallel 20  -run=TestAccSecretsManagerSecretRotation -timeout 180m
=== RUN   TestAccSecretsManagerSecretRotationDataSource_basic
=== PAUSE TestAccSecretsManagerSecretRotationDataSource_basic
=== RUN   TestAccSecretsManagerSecretRotation_basic
=== PAUSE TestAccSecretsManagerSecretRotation_basic
=== CONT  TestAccSecretsManagerSecretRotationDataSource_basic
=== CONT  TestAccSecretsManagerSecretRotation_basic
--- PASS: TestAccSecretsManagerSecretRotation_basic (49.05s)
--- PASS: TestAccSecretsManagerSecretRotationDataSource_basic (52.32s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/secretsmanager     55.051s
```

jar-b:
Nice find and fix @frek818 !

007:
@jar-b Why a breaking change for a point release?

jar-b:
In this case we decided to move forward with removing the attribute (versus deprecating and removing in v5) because the practitioner impact should be minimal while solving the issue documented in #27655. The underlying resource doesn't support tags, and the `tags` attribute was never included in the website documentation (which may have encouraged incorrect configurations).

#27597 is a similar situation where while technically breaking, the attributes modified are always computed and therefore unlikely to break a real world configuration. That changelog entry is categorized as a bug, as this could have been, but I opted for breaking to be transparent in the case this does have an unexpected impact and we need to roll back. 

Hope this helps to explain the line of thinking here, and please open an issue and link to this PR if you have any concerns.

shidel-dev:
@jar-b how do I upgrade in the case where my state already has the `tag` attribute? Seeing errors like `Error decoding "......" from previous state: unsupported attribute "tags` when running a plan

jar-b:
@shidel-dev - Are you still observing the error above? 

If so, do you mind opening a separate issue with your terraform configuration and the plugin versions (previous and version you upgraded to)? We want to track any upgrade issues this may be causing.

shidel-dev:
I was observing it until I rebuilt my remote state with `terraform apply -refresh-only -auto-approve`. I think anyone that has remote state with this resource in it will have this issue if they run plan, or apply with targets. Rebuilding the state seems like a reasonable workaround.

rossbush:
I am not sure if this is the exact line, but I think a line like this causeing a de-serialization issue during tf apply after updating to 4.450.0 ->

`resource "aws_kms_key" "secrets_manager" {
  provider = aws.primary

  deletion_window_in_days = 30
  description             = "Encryption key for the ${local.composite_name}-${local.primary_region_alias}-rds-master-secret"
  enable_key_rotation     = true

  tags = merge(
    var.tags,
    {
      Name = "${local.composite_name}-${local.primary_region_alias}-rds-master-secret-kms-key"
    }
  )`

PR #27788, opened 2022-11-14
DOC: Added missing "domain" attr to ses_domain_identity

ShaharNaveh:
Looking at the source code at: 

https://github.com/hashicorp/terraform-provider-aws/blob/d481be15e66cff7cf9d95f9f1258c148ef63c1cb/internal/service/ses/domain_identity.go#L67

And at the docs of [ses_domain_identity](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ses_domain_identity):

2022-11-14-132753_1895x1032_scrot

---

#### P.S.

093a697 is just adding a dot at the and of the line, I can revert this commit if it's out of scope.

### DIFF
--- a/website/docs/r/ses_domain_identity.html.markdown
+++ b/website/docs/r/ses_domain_identity.html.markdown
@@ -20,6 +20,8 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
+* `domain` - Name of the domain.
+
 * `arn` - The ARN of the domain identity.
 
 * `verification_token` - A code which when added to the domain as a TXT record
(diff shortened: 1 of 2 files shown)


website/docs/r/ses_domain_identity.html.markdown
```diff
@@ -20,6 +20,8 @@ The following arguments are supported:
 
 In addition to all arguments above, the following attributes are exported:
 
+* `domain` - Name of the domain.
```
jar-b: I believe `domain` is intentionally omitted here since it is included in the arguments section above. 

> In addition to all arguments above, the following attributes are exported

jar-b (approved):

jar-b:
Thanks for your contribution @ShaharNaveh! 👏 

I removed the `domain` addition as indicated in the review, but kept the formatting fix. Appreciate it!

Issue #26809, opened 2022-09-14
Add ability to add features to existing feature groups in Sagemaker


aShevc:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently when one adds new feature to the existing Sagemaker feature group, it forces the destruction of resource. However, since [FeatureAdditions](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_UpdateFeatureGroup.html#API_UpdateFeatureGroup_RequestSyntax) functionality has been introduced to the UpdateFeatureGroup API call, this must not be the case.

### New or Affected Resource(s)

* aws_sagemaker_feature_group

### Potential Terraform Example

```hcl
resource "aws_sagemaker_feature_group" "feature_group" {
  feature_group_name                   = "some name"
// ...rest of the configuration and features

// the feature definition below is the new feature in he feature group
    feature_definition {
      feature_name = example_add
      feature_type  = string
    }
}

```

### Expected Behavior

When adding a feature to feature group, the feature group should be updated with the new feature.

### Actual Behavior

Executing Terraform plan gives the following:

```
  # aws_sagemaker_feature_group.feature_group[0] must be replaced
-/+ resource "aws_sagemaker_feature_group" "feature_group" {

      + feature_definition { # forces replacement
          + feature_name = "example_add"
          + feature_type = "String"
        }
...

    }

Plan: 1 to add, 0 to change, 1 to destroy.

```

### References

* [resourceFeatureGroupUpdate](https://github.com/hashicorp/terraform-provider-aws/blob/46cfb1e4e8624c4a865ddecd24fdc6f948ed2275/internal/service/sagemaker/feature_group.go#L318-L330) - it looks like this is the function needs to be updated

DrFaust92:
looking into this

xianwill:
@DrFaust92 just wanted to see if you have made any progress on this?

DrFaust92:
[...]

PR #27933, opened 2022-11-21
[WIP] Add ability to add features to existing feature groups in Sagemaker

aShevc:
### Description

This leverages newly added [UpdateFeatureGroupRequest](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_UpdateFeatureGroup.html#API_UpdateFeatureGroup_RequestSyntax) in order to introduce adding new features to feature groups.

**NOTE:** in order to enable feature additions, I have had to remove `ForceNew` attribute from `feature_definition` field in order to disable recreation of the feature group when the size changes, but replaced it with `CustomizeDiff`. However, I was not able to simultaneously force creating the new feature group in case if the existing feature groups change, it will only force feature groups recreation when the new size of feature definitions is smaller. The reason for not being able to detect change of existing feature definitions seems to be the improper handling of `CustomizeDiff` attribute for `TypeList` fields. I have added an [issue](https://github.com/hashicorp/terraform-plugin-sdk/issues/1103) into Terraform SDK Plugin repo in order to address this. However, while resolving this definitely might take some time, I have fallen back to the solution on returning an error in case if the existing feature groups change upon applying the plan.

At this time, as I have tried quite a bit of different approaches, as seen in the issue above, and I would largely appreciate suggestions on any alternatives, I will gladly try to apply worthwhile suggestions

Some of the potential alternatives on the plate include:
- Set feature definitions field as computed, try to utilize `ComputedIf` and/or `SetNew` in case if undesired feature groups change are introduced.
- Rework Feature group service as the one using Terraform Framework rather than Terraform SDK. This may be usefuls as [PlanModifiers](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#Attribute.PlanModifiers) may potentially be helpful in resolving the issue. 

However, from what I see from my experience with `TypeList` and `CustomizeDiff` function, it may happen that those changes won't bring the required effect. Specifically, redesigning a service to use Terraform Framework rather than SDK may be quite a big change, especially given relatively small impact.

Therefore, the second thing beyond discussing alternative solutions would be to determine the plausibility of a fallback option I introudced. Woud largely appreciate a feedback on this

TODO:
- [ ] Explore additional ways to force changes prior to attempting applying the update
- [ ] Acceptance tests

### Relations

Closes [#26809](https://github.com/hashicorp/terraform-provider-aws/issues/26809)

```

...
```

### DIFF
--- a/internal/service/sagemaker/feature_group.go
+++ b/internal/service/sagemaker/feature_group.go
@@ -29,10 +29,11 @@ func ResourceFeatureGroup() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 		CustomizeDiff: customdiff.All(
+
 			customdiff.ForceNewIf("feature_definition", func(_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
-				old, new := d.GetChange("feature_definition")
-				var featureDefinitionsOld = expandFeatureGroupFeatureDefinition(old.([]interface{}))
-				var featureDefinitionsNew = expandFeatureGroupFeatureDefinition(new.([]interface{}))
+				o, n := d.GetChange("feature_definition")
+				var featureDefinitionsOld = expandFeatureGroupFeatureDefinition(o.([]interface{}))
+				var featureDefinitionsNew = expandFeatureGroupFeatureDefinition(n.([]interface{}))
 
 				return len(featureDefinitionsNew) < len(featureDefinitionsOld)
 			}),
@@ -89,7 +90,6 @@ func ResourceFeatureGroup() *schema.Resource {
 			"feature_definition": {
 				Type:     schema.TypeList,
 				Required: true,
-				ForceNew: false,
 				MinItems: 1,
 				MaxItems: 2500,
 				Elem: &schema.Resource{
@@ -97,8 +97,6 @@ func ResourceFeatureGroup() *schema.Resource {
 						"feature_name": {
 							Type:     schema.TypeString,
 							Optional: true,
-							// TODO: line below will also force change when new features are added, workaround needded
-							// ForceNew: true,
 							ValidateFunc: validation.All(
 								validation.StringLenBetween(1, 64),
 								validation.StringNotInSlice([]string{"is_deleted", "write_time", "api_invocation_time"}, false),
@@ -107,9 +105,8 @@ func ResourceFeatureGroup() *schema.Resource {
 							),
 						},
 						"feature_type": {
-							Type:     schema.TypeString,
-							Optional: true,
-							// ForceNew: true,
+							Type:         schema.TypeString,
+							Optional:     true,
 							ValidateFunc: validation.StringInSlice(sagemaker.FeatureType_Values(), false),
 						},
 					},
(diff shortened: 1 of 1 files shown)


internal/service/sagemaker/feature_group.go
```diff
@@ -318,6 +328,16 @@ func resourceFeatureGroupRead(d *schema.ResourceData, meta interface{}) error {
 func resourceFeatureGroupUpdate(d *schema.ResourceData, meta interface{}) error {
 	conn := meta.(*conns.AWSClient).SageMakerConn
 
+	o, n := d.GetChange("feature_definition")
+	var featureDefinitionsOld = expandFeatureGroupFeatureDefinition(o.([]interface{}))
+	var featureDefinitionsNew = expandFeatureGroupFeatureDefinition(n.([]interface{}))
+
+	if !checkIfDefinitionsUnchanged(featureDefinitionsOld, featureDefinitionsNew) {
+		return fmt.Errorf("existing feature definitions should remain unchanged. Only additions of nes feature"+
```
zbstof, line 336:
```suggestion
		return fmt.Errorf("existing feature definitions should remain unchanged. Only additions of new feature"+
```

aShevc:
Updated the pull request with the following approach:

Replaced initial CustomizeDiff function with the function showing errors in case if feature definitions of existing features change

Hence, currently, it will give an error in case if feature definitions change on plan. Still found no way for now to force changes at that stage.

Will proceed on working on acceptance tests and make it ready for review

aShevc:
@YakDriver @DrFaust92 can I gently ask to have an initial look at this guys?

DrFaust92:
[aShevc](https://github.com/aShevc) looks good but we are still with the issue of needing to force change as it will make operator to target destroy/taint and it think its bad UX

aShevc:
> [aShevc](https://github.com/aShevc) looks good but we are still with the issue of needing to force change as it will make operator to target destroy/taint and it think its bad UX

I understand, but it is really the only option we have at the moment until the [issue](https://github.com/hashicorp/terraform-plugin-sdk/issues/1103) in Terraform plugin SDK is fixed. Do you think we can move forward temporarily with this solution and adjust it later?

Also, if you have other thoughts on how we could better deal with this scenario, this would be largely appreciated. Thank you!

DrFaust92:
If Maintainers are ok with this its up to them, as i personally opted to just not submit a similar PR im against half measures

aShevc:
@ewbankkit would you mind re-approving the actions run? I have had the Lint formatting fixed, so it should build now. Thanks!

aShevc:
@ewbankkit @DrFaust92 since all checks are authorized, would the team be ready to merge it in soon? Do you guys have a plan on releasing this? Please do let know. Thank you!

mpwsh:
sorry to ping but any updates here? <3 would love to get this merged. FeatureGroup updates are a nightmare to handle through terraform as of now

obscure-star:
+1 to mpwsh's comment. Recent changes were made to the resourceFeatureGroupUpdate function, which only initiates an update of the feature group if the online_store_config or the throughput_config has changed. This still does not address the issue where the terraform plan destroys and recreates the feature group when there is a change in a feature definition. It seems this PR addresses that issue. Can someone look at this?

nicosandller:
+1 to REALLY needing this pull request. Please team!

Issue #27852, opened 2022-11-17
[Bug]: compressionType is not a valid parameter in extra connection attributes


shanmugahp:
### Terraform Core Version

1.0.5

### AWS Provider Version

4.8.0

### Affected Resource(s)

aws_dms_endpoint

### Expected Behavior

Terraform should not send the default value for compressionType and Resource is created

### Actual Behavior

Endpoint fails to create as aws has removed the compressionType attribute for dms s3 as source endpoint . Terraform is sending default value of compressionType=None causing the call to fail.

          + cdc_path                         = "changedata"
          + compression_type                 = "NONE"
          + csv_delimiter                    = "|"
          + csv_row_delimiter                = "\\n"
          + data_format                      = "csv"
          + data_page_size                   = 1048576
          + date_partition_delimiter         = "slash"
          + date_partition_enabled           = false
          + date_partition_sequence          = "yyyymmdd"

### Relevant Error/Panic Output Snippet

```shell
Error: Error creating DMS endpoint: InvalidParameterValueException: Unsupported value 'compressionType' for extra connection attributes
	status code: 400, request id: 89916c08-3fa2-454e-affb-2c69c6572df1

  with module.data-lake.module.sources.module.source_activities[0].aws_dms_endpoint.main[0],
  on ../../main/data-sources/s3-data-source/main.tf line 5, in resource "aws_dms_endpoint" "main":
   5: resource "aws_dms_endpoint" "main" {
```

### Terraform Configuration Files

resource "aws_dms_endpoint" "main" {

  endpoint_id   = var.resource_prefix
  endpoint_type = "source"
  ssl_mode      = "none"
  tags          = var.application_tags
  engine_name   = "s3"

  s3_settings {
    bucket_name               = var.source_bucket_id
    bucket_folder             = var.source_bucket_folder
    service_access_role_arn   = var.activities_dms_s3_role_arn
    external_table_definition = var.external_table_mappings
    csv_delimiter             = var.csv_delimiter
    cdc_path                  = var.cdc_path
    csv_null_value            = var.csv_null_value
    csv_row_delimiter         = var.csv_row_delimiter
  }

### Steps to Reproduce

Create a DMS source endpoint with S3 as source

### References

* #23506

rossbush:
Our pipelines are silent. Please make this a high priority.

jeffarredondo:
Ok at least I'm not crazy. Ran into this for MySQL as well when testing a potential change to include the sybase engine

[...]

PR #27957, opened 2022-11-22
Fix s3 compression type no longer supported

rossbush:
### Description

Fix compressionType no longer supported.

### Relations

Relates #27852


rossbush:
This looks like the wrong way.

sbuzan:
Do we know when this will be released?  We have a dozen or so pipelines that are blocked because of this.

Issue #27283, opened 2022-10-18
[Bug]: `aws_dms_replication_task` failed with error InvalidParameterValueException: TimestampColumnName cannot be an empty string.


thaiphv:
### Terraform Core Version

1.3.2

### AWS Provider Version

4.35.0

### Affected Resource(s)

* `aws_dms_replication_task`

### Expected Behavior

The Terraform provider should create a task successfully.

### Actual Behavior

The Terraform provider failed to create the resource and reported the error: `InvalidParameterValueException: TimestampColumnName cannot be an empty string.`

### Terraform Configuration Files

```terraform
resource "aws_dms_endpoint" "source" {
  endpoint_id = "dms-source"

  database_name = var.source_db_name
  endpoint_type = "source"
  engine_name   = "sqlserver"
  username      = var.source_db_username
  password      = var.source_db_password
  server_name   = var.source_db_server_name
  port          = var.source_db_server_port
}

resource "aws_dms_endpoint" "fullload" {
  endpoint_id = "fullload-task"

  endpoint_type = "target"
  engine_name   = "s3"

  s3_settings {
    add_column_name   = true
    bucket_name       = var.bucket
    bucket_folder     = var.s3_folder
    compression_type  = "NONE"
    csv_delimiter     = ","
    csv_row_delimiter = "\\n"
    date_partition_enabled = false
    include_op_for_full_load = true
    rfc_4180                 = false
    service_access_role_arn  = var.dms_s3_iam_role_arn
  }
}

resource "aws_dms_replication_task" "task" {
  replication_task_id = "fullload-and-cdc"

  migration_type            = "full-load-and-cdc"
  replication_instance_arn  = var.dms_replication_instance_arn
  replication_task_settings = var.fullload_cdc_task_settings

  source_endpoint_arn = aws_dms_endpoint.source.endpoint_arn
  target_endpoint_arn = aws_dms_endpoint.fullload.endpoint_arn

  table_mappings = var.mappings
}
```

### Steps to Reproduce

Run `terraform apply`

thaiphv:
Just realised that the Terraform provider applied a whole lot of default settings to the endpoint:
```json
{
    "ServiceAccessRoleArn": "<arn>",
    "ExternalTableDefinition": "",
    "CsvRowDelimiter": "\\n",
    "CsvDelimiter": ",",
    "BucketFolder": "<folder>",
    "BucketName": "<bucket>",
    "CompressionType": "NONE",
[...]

PR #27980, opened 2022-11-23
service-dms-endpoint-s3Settings-compressionType is no longer supported for source endpoints

rossbush:
### Description
Remove default value of "NONE" for dms endpoint S3 compression. Source s3 endpoints no longer support compressionType. When not set, the aws default is no compression. This allows the target to continue to define a compressionType while source endpoints can get around the bug by not setting a value for this property. The only other work around is change source endpoint to target, deploy and the change it back to source and deploy, for each dms task, which is not ideal.

### Relations
Relates #27852

### References
#27283


sbuzan:
Do we have an ETA when this will be released?

YakDriver:
Thank you so much for your work on this and interest! 🎉 

Unfortunately, this fix would not be sufficient.

1. As noted in [this comment](https://github.com/hashicorp/terraform-provider-aws/issues/27852#issuecomment-1319844717), there are other attributes affected as well.
2. Removing the default is generally considered a breaking change (although here no value and `NONE` are equivalent).
3. Removing the default results in diffs. When `COMPRESSION_TYPE` is not included in the configuration, AWS returns `NONE` but the configuration is `""`, which causes a diff.

In order to address these problems and additional continuing challenges with the deprecation of extra connection attributes, we have created a new resource `aws_dms_s3_endpoint` (see #28130). Hopefully this will allow us to move forward in a much cleaner way than continuing to try to rework `aws_dms_endpoint`.

PR #27988, opened 2022-11-23
correct instances.0.console_url description to indicate this references the consule_url for ActiveMQ and RabbitMQ broker types

logan-bobo:
correct `instances.0.console_url` description to indicate this references the `consule_url` for `ActiveMQ` and `RabbitMQ` broker types

I have noticed when reading through the docs for `mq_broker` that the attribute `instances.0.console_url` seems to suggest to users that this is only for the `ActiveMQ Web Console` this could confuse users looking for the `RabbitMQ Management UI` endpoint. The resource provides the URL for both the `ActiveMQ Web Console` and the `RabbitMQ Management UI`. I have updated the docs to reflect this


ewbankkit:
@logan-bobo Thanks for the contribution :tada: :clap:.

PR #28039, opened 2022-11-28
Update the restrictions entry

robin-stucomm:
### Description
The `locations` attribute is required on a `geo_restriction`, while the documentation says it's optional. This attribute can be an empty array.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@robin-stucomm Thanks for the contribution :tada: :clap:.

PR #28116, opened 2022-12-01
docs: aws_lambda_permission add cross-account URL example

baolsen:
### Description

Add an example for `aws_lambda_permission` in a cross-account scenario and allowing a Lambda function URL invoke.

The current examples appear to only support AWS Service principals when AWS IAM principals are also supported.

### References

https://docs.aws.amazon.com/lambda/latest/dg/urls-auth.html

### DIFF
--- a/website/docs/r/lambda_permission.html.markdown
+++ b/website/docs/r/lambda_permission.html.markdown
@@ -179,15 +179,42 @@ EOF
 }
 ```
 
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@baolsen Thanks for the contribution :tada: :clap:.

PR #28232, opened 2022-12-07
Fix CF timeout default values

dtaniwaki:
### Description

The default timeout values of a `aws_cloudfront_distribution` resource are not correct and confusing.

Here's a screenshot of the corresponded CloudFront page.

Screenshot 2022-12-07 at 20 51 12

And, the official document also writes the same.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html

### Relations

N/A

### References

N/A

Only markdown changes.


dtaniwaki:
Ah, I realized the value is not about the default values, but default max values of the configuration...

PR #28239, opened 2022-12-07
(docs): Inform about potential traffic blocking

### LINKED ISSUE #27820: [Docs]: Ingress/Egress Rule (In Security Group) Not Created Without Optional CIDR Block Parameter
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group

### Description

If you create a security group with ingress/egress inside but without the CIDR block parameter (which is marked as optional) Terraform will create a security group in AWS without any ingress/egress rules. This results in blocking all traffic.

I couldn't find this behavior documented on the page and my assumption was that it defaults to allow all traffic or use the specified VPC's CIDR block. Maybe this parameter is optional in another use case but I'd suggest adding a note to avoid other people encountering this issue.

Normally "required" correlates with "resource can't be created without" and if something is missing to create a resource Terraform will complain. In this scenario neither applies and the written ingress/egress gets ignored.

### References

_No response_

_No response_

dev-slatto:
### Description
As mentioned in issue #27820 traffic can be blocked on a resource if you create a security group without any of the available `cidr_block` parameters. It's intentional to have these optional so that you're not forced to have all of them present if you only need one or the other. A note should be present informing the user about that not providing any of them could lead to your traffic being blocked. 

### Relations
Closes #27820


ewbankkit:
LGTM :rocket:.

ewbankkit:
@dev-slatto Thanks for the contribution :tada: :clap:.

PR #28307, opened 2022-12-12
Add d/aws_apigateway_rest_apis

Mongey:
### Relations

Closes #0000

```

...
```


mburisa-glooko:
Any status update on this? Would be useful to get `arns` alongside names, same as in https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/lbs

Use case: module for AWS WAF where we are associating ACL with a list of API Gateways (same as with the list of ALBs)

Issue #428, opened 2017-06-13
aws_s3_bucket force_destroy option has no effect on pre-existing bucket


hashibot:
_This issue was originally opened by @jhoblitt as hashicorp/terraform#10738. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
```
$ ./bin/terraform --version
Terraform v0.8.0

```
### Affected Resource(s)
- `aws_s3_bucket`

### Terraform Configuration Files
```hcl
resource "aws_s3_bucket" "ssh_public_keys" {
  region = "${var.aws_default_region}"
  bucket = "${var.bastion_bucket_name}"
  acl    = "private"
  force_destroy = true
  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "1",
      "Effect": "Allow",
      "Principal": "*",
      "Action": [
        "s3:List*",
        "s3:Get*"
      ],
      "Resource": "arn:aws:s3:::${var.bastion_bucket_name}"
    }
  ]
}
EOF
}
```

### Expected Behavior
bucket is deleted then recreated to modify IAM profile

### Actual Behavior

```
$ ./bin/terraform import aws_s3_bucket.ssh_public_keys jhoblitt-bastion-test
aws_s3_bucket.ssh_public_keys: Importing from ID "jhoblitt-bastion-test"...
aws_s3_bucket.ssh_public_keys: Import complete!
  Imported aws_s3_bucket (ID: jhoblitt-bastion-test)
aws_s3_bucket.ssh_public_keys: Refreshing state... (ID: jhoblitt-bastion-test)
$ ./bin/terraform apply
...
aws_s3_bucket.ssh_public_keys: Modifying...
  force_destroy: "false" => "true"
aws_s3_bucket.ssh_public_keys: Modifications complete
Error applying plan:

1 error(s) occurred:

* aws_s3_bucket.s3_public_keys: Error deleting S3 Bucket: BucketNotEmpty: The bucket you tried to delete is not empty
	status code: 409, request id: 3991CB0128E5871C

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.

```

### Steps to Reproduce
1.  import an existing bucket
2. `terraform apply`

[...]

PR #28351, opened 2022-12-14
docs: clarify behaviour of aws_s3_bucket force_destroy

morremeyer:
### Description

This clarifies the behaviour of the `force_destroy` parameter for an `aws_s3_bucket` resource.

The current documentation is ambiguous as to when the objects in the bucket are actually deleted.

### Relations

* Relates to #428: It could clarify some cases where the behaviour is not correctly understood

### References

The behaviour is not documented well so far, but the code at https://github.com/hashicorp/terraform-provider-aws/blob/e9ea507012986603aec663169da2e8d694990112/internal/service/s3/bucket.go#L1439-L1485 clearly shows that this only happens when the bucket is destroyed.

_None_ since this is a documentation-only change.


breathingdust:
LGTM 🚀  Thanks @morremeyer!

PR #28373, opened 2022-12-15
 Data source aws_s3control_multi_region_access_point

### LINKED ISSUE #28251: [New]: Data source aws_s3control_multi_region_access_point
### Description

There currently a resource `aws_s3control_multi_region_access_point` but no corresponding data source. It seems like a common use case that you might want to get data about a S3 multi region access point in a separate TF configuration so the data source would be required. I thin looking up the access point by name should be possible and the simplest to use.

If you attempt this at the moment you received the following error
```
│ Error: Invalid data source
│ 
│   on exisitng.tf line 21, in data "aws_s3control_multi_region_access_point" "web_server_db_access_point":
│   21: data "aws_s3control_multi_region_access_point" "web_server_db_access_point" {
│ 
│ The provider hashicorp/aws does not support data source "aws_s3control_multi_region_access_point".
│ 
│ Did you intend to use the managed resource type "aws_s3control_multi_region_access_point"? If so, declare this using a "resource" block instead of a "data" block.
```

### Requested Resource(s) and/or Data Source(s)

data "aws_s3control_multi_region_access_point"

```terraform
data "aws_s3control_multi_region_access_point" "web_server_db_access_point" {
  name = "web-server-db-ap"
}
```

### References

_No response_

No

tiborhercz:
### Description
S3 multi region access point data resource.

### Relations
Closes #28251

```
$ make testacc TESTARGS='-run=TestAccS3ControlMultiRegionAccessPointDataSource_basic' PKG=s3control
TF_ACC=1 go test ./internal/service/s3control/... -v -count 1 -parallel 20  -run=TestAccS3ControlMultiRegionAccessPointDataSource_basic -timeout 180m
=== RUN   TestAccS3ControlMultiRegionAccessPointDataSource_basic
=== PAUSE TestAccS3ControlMultiRegionAccessPointDataSource_basic
=== CONT  TestAccS3ControlMultiRegionAccessPointDataSource_basic
--- PASS: TestAccS3ControlMultiRegionAccessPointDataSource_basic (324.84s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/s3control  327.081s


...
```


internal/service/s3control/multi_region_access_point_data_source.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-aws/internal/conns"
+	"github.com/hashicorp/terraform-provider-aws/internal/verify"
+)
+
+func DataSourceMultiRegionAccessPoint() *schema.Resource {
+	return &schema.Resource{
+		ReadWithoutTimeout: dataSourceMultiRegionAccessPointBlockRead,
+
+		Schema: map[string]*schema.Schema{
```
bschaatsbergen, line 20: Drive-by cosmetic comment 💄  - preferably we order the schema alphabetically.

internal/service/s3control/multi_region_access_point_data_source.go
```diff
@@ -0,0 +1,92 @@
...
+			"alias": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"created_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"public_access_block": {
+				Type:     schema.TypeMap,
```
bschaatsbergen, line 39: I believe `public_access_block` contains a few more attributes, such as: `block_public_acls`, `block_public_policy`, etc.. 

You can change the type to a `schema.TypeList` too.

internal/service/s3control/multi_region_access_point_data_source.go
```diff
@@ -0,0 +1,92 @@
...
+				Computed: true,
+			},
+			"regions": {
```
bschaatsbergen, line 42: This differs a bit from the corresponding resource, is that intended? I believe the corresponding resource has a `region` attribute instead of `regions`.

internal/service/s3control/multi_region_access_point_data_source.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err != nil {
+		return diag.Errorf("error reading S3 Multi Region Access Point: %s", err)
+	}
+
+	if output == nil || output.AccessPoint == nil {
+		return diag.Errorf("error reading S3 Multi Region Access Point (%s): missing access point", accountID)
+	}
+
+	d.SetId(accountID)
```
bschaatsbergen, line 84: Perhaps we can make the ID a bit more unique, see function: `MultiRegionAccessPointCreateResourceID` in the corresponding resource.

internal/service/s3control/multi_region_access_point_data_source.go
```diff
@@ -0,0 +1,92 @@
...
+	d.Set("created_at", aws.TimeValue(output.AccessPoint.CreatedAt).Format(time.RFC3339))
```
bschaatsbergen, line 85: I don't think that this is captured in the corresponding resource either, is it intended to add this to the data source?

internal/service/s3control/multi_region_access_point_data_source.go
```diff
@@ -0,0 +1,92 @@
...
+	d.Set("name", output.AccessPoint.Name)
+	d.Set("public_access_block", output.AccessPoint.PublicAccessBlock)
```
bschaatsbergen, line 87: PublicAccessBlock is a type that contains several properties, you should expand this and flatten this type.

bschaatsbergen (commented):
Added a few drive-by comments, nice WIP 👍

internal/service/s3control/multi_region_access_point_data_source.go
tiborhercz: I have ordered the schema order like this because the `account_id` and `name` are both parameters that are needed as argument for the data source. From `alias` down are attributes that are returned when using the data source.

internal/service/s3control/multi_region_access_point_data_source.go
tiborhercz: Good point! Missed that one

internal/service/s3control/multi_region_access_point_data_source.go
tiborhercz: The reason I have added this is because in the API it is outputted
See: https://github.com/aws/aws-sdk-go/blob/main/service/s3control/api.go#L15133

internal/service/s3control/multi_region_access_point_data_source.go
tiborhercz: Good point! 
Implemented it

internal/service/s3control/multi_region_access_point_data_source.go
tiborhercz: Updated the code

ewbankkit:
@tiborhercz Thanks for the contribution :tada: :clap:.

bschaatsbergen:
Nice work @tiborhercz :+1:

PR #28496, opened 2022-12-20
codebuild_webhook example should use BASE_REF

jrobison-sb:
The codebuild_webhook example should use BASE_REF instead of HEAD_REF. BASE_REF would refer to a branch like `master` or `main`, whereas `HEAD_REF` would refer to a feature branch. Since most people are probably using Codebuild to run CI tests on their pull requests, and their pull requests are meant to merge into master, changing this to BASE_REF provides a more easily pasteable example.


breathingdust:
LGTM 🚀 Thanks @jrobison-sb!

Issue #27908, opened 2022-11-20
[Docs]: `website_endpoint` is not listed as deprecated


klovelyPhraseHealth:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#website_endpoint

### Description

Attempting to use this attribute yields

```
│ Warning: Deprecated attribute
│
│   on .terraform/modules/cloudfront_web_app/cloudfront.tf line 16, in resource "aws_cloudfront_distribution" "s3_site_cdn":
│   16:     domain_name = aws_s3_bucket.main_bucket.website_endpoint
│
│ The attribute "website_endpoint" is deprecated. Refer to the provider documentation for details.
```

other documentation gives better info

### References

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade#website-website_domain-and-website_endpoint-arguments

### Would you like to implement a fix?

nahh

tmccombs:
The error also says to refer to the provider documentation, but isn't immediately clear what you should use instead, there are no references to deprecation anywhere near any reference of `website_endpoint`. 

I assume the intention is that you should isntead use an `aws_s3_bucket_website_configuration` and use the `website_endpoint` on that.

PR #28524, opened 2022-12-21
r/aws_s3_bucket: document that website_endpoint and website_domain are deprecated

tmccombs:
Closes: #27908

### DIFF
--- a/website/docs/r/s3_bucket.html.markdown
+++ b/website/docs/r/s3_bucket.html.markdown
@@ -721,12 +721,8 @@ In addition to all arguments above, the following attributes are exported:
 * `hosted_zone_id` - The [Route 53 Hosted Zone ID](https://docs.aws.amazon.com/general/latest/gr/rande.html#s3_website_region_endpoints) for this bucket's region.
 * `region` - The AWS region this bucket resides in.
 * `tags_all` - A map of tags assigned to the resource, including those inherited from the provider [`default_tags` configuration block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-configuration-block).
-* `website_endpoint` - The website endpoint, if the bucket is configured with a website. If not, this will be an empty string.
-  -> **NOTE:** The attribute `website_endpoint` is deprecated.
-  Use the same on attribute on the [`aws_s3_bucket_website_configuration`](s3_bucket_website_configuration.html.markdown) resource instead.
-* `website_domain` - The domain of the website endpoint, if the bucket is configured with a website. If not, this will be an empty string. This is used to create Route 53 alias records.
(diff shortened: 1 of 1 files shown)


tmccombs:
> That said, given that website_endpoint and website_domain are under the attributes section, I can see your point.

Yes, that was my thinking. And in fact I don't think these attributes are mentioned in the section for the website block so doing a search in the (rather long) page for the attribute name doesn't lead to any messaging around depecation.

website/docs/r/s3_bucket.html.markdown
```diff
@@ -722,7 +722,11 @@ In addition to all arguments above, the following attributes are exported:
 * `region` - The AWS region this bucket resides in.
 * `tags_all` - A map of tags assigned to the resource, including those inherited from the provider [`default_tags` configuration block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-configuration-block).
 * `website_endpoint` - The website endpoint, if the bucket is configured with a website. If not, this will be an empty string.
+  -> **NOTE:** The attribute `website_endpoint` is deprecated.
+  Use the same on attribute on the [`aws_s3_bucket_website_configuration`](s3_bucket_website_configuration.html.markdown) resource instead.
```
justinretzolk, line 726:
```suggestion
* `website_endpoint` - (**Deprecated**) The website endpoint, if the bucket is configured with a website. If not, this will be an empty string. Use the resource [`aws_s3_bucket_website_configuration`](s3_bucket_website_configuration.html.markdown) instead.
```

website/docs/r/s3_bucket.html.markdown
```diff
@@ -722,7 +722,11 @@ In addition to all arguments above, the following attributes are exported:
...
 * `website_domain` - The domain of the website endpoint, if the bucket is configured with a website. If not, this will be an empty string. This is used to create Route 53 alias records.
+  -> **NOTE:** The attribute `website_domain` is deprecated.
+  Use the same on attribute on the [`aws_s3_bucket_website_configuration`](s3_bucket_website_configuration.html.markdown) resource instead.
```
justinretzolk, line 729:
```suggestion
* `website_domain` - (**Deprecated**) The domain of the website endpoint, if the bucket is configured with a website. If not, this will be an empty string. This is used to create Route 53 alias records. Use the resource [`aws_s3_bucket_website_configuration`](s3_bucket_website_configuration.html.markdown) instead.
```

justinretzolk (approved):
Hey @tmccombs 👋 After discussing internally, and with your latest comment in mind, I agree and think we should get this merged in. I left a couple of suggestions that I think will make the formatting more similar to how the deprecation is called out in the main `website` argument description. My thought is that this should help to maintain readability, since there won't be yellow callout boxes between the attributes.

PR #28578, opened 2022-12-27
aws_dms_endpoint.cdc_min_file_size is kilobytes, no longer megabytes

bennylu2:
### Description
https://github.com/hashicorp/terraform-provider-aws/issues/27688 captures the details of the bug

`cdc_min_file_size` is 32 which previously was correct when api was in megabytess but has since changed to kilobytes. Updating to 32000 as default which is in aws docs.

Closes https://github.com/hashicorp/terraform-provider-aws/issues/27688

### References
https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Target.S3.html#CHAP_Target.S3.Configuring

### DIFF
--- a/.changelog/28578.txt
+++ b/.changelog/28578.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_dms_endpoint: cdc_min_file_size is in kilobytes, default is 32000
+```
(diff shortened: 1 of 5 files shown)


YakDriver:
Changing a default is generally considered a breaking change. However, I think we have good grounds for changing anyway since the change in multiplier comes from AWS.

YakDriver (approved):
LGTM 🎉 Thanks for your work on this!

```console
% make testacc TESTS=TestAccDMSEndpoint_S3 PKG=dms
TF_ACC=1 go test ./internal/service/dms/... -v -count 1 -parallel 20 -run='TestAccDMSEndpoint_S3'  -timeout 180m
```

PR #28610, opened 2022-12-30
Doc update: Standard way of checking resource disappeared

bennylu2:
### Description
Was going through document and noticed the example for checking if resource disappeared was not what I was seeing when looking at recent code. Updated with what appears to be the standard.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@bennylu2 Thanks for the contribution :tada: :clap:.

PR #28641, opened 2023-01-03
Fix maximum size of zipfile

mgwiings:
### Description

This PR includes change to the synthetic monitor resource, changing the allowed size of `zip_file` accordingly to AWS API.


breathingdust:
LGTM 🚀 Thanks @mgwiings for the correction!

PR #28630, opened 2023-01-03
Doc: glue trigger "EVENT" type

jessehall3:
### Description
This change adds documentation for the glue trigger `EVENT` type, as stated in the [aws glue trigger docs](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-api-jobs-trigger.html#aws-glue-api-jobs-trigger-CreateTrigger) and seen [in the SDK](https://github.com/aws/aws-sdk-go/blob/a9faefa96cb9bee57d55d9702b6e7de4aad69bc0/models/apis/glue/2017-03-31/api-2.json#L11419-L11427). I could not find within the AWS docs anything specifically stating why `enabled` must be set to `false` for the `EVENT` type, but I know this to be true from seeing the error myself.

### Relations

Closes #0000

### Output from Acceptance Testing

```
$ make testacc TESTS=TestAccXXX PKG=ec2

...
```

PR #28650, opened 2023-01-03
Doc: Invalid statement in glue trigger docs

jessehall3:
### Description
@ewbankkit 
This is to remove an invalid statement added with [a previous PR](https://github.com/hashicorp/terraform-provider-aws/pull/28630) just merged about 9 hours ago. My sincere apologies, I have since realized through subsequent attempts, that this statement is not true. Likewise, I am not able to identify why it had been necessary to resolve a previous error message. Regardless, I cannot say definitely that it is true, therefore I think the statement should be removed.

### Relations

Closes #0000

```

...
```


ewbankkit (approved):
LGTM :rocket:.

PR #28661, opened 2023-01-04
Correct documentation to specify ARN is required.

theherk:
When following the documentation, the following error is returned:

```
Error: "rule.0.destination.0.bucket" (given-bucket-name) is an invalid ARN: arn: invalid prefix
```

This is because the API requires the full ARN be given. The resource correctly identifies `verify.ValidARN`, but the documentation is incorrect. This is updated in accordance with AWS's own documentation that can be referenced in [the official API documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Destination.html).

```
terraform-provider-aws
ᐅ make testacc TESTS=testAccBucketReplicationConfigurationConfig_rulesDestination PKG=s3
TF_ACC=1 go test ./internal/service/s3/... -v -count 1 -parallel 20 -run='testAccBucketReplicationConfigurationConfig_rulesDestination'  -timeout 180m
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/s3 22.271s [no tests to run]
```

Not entirely clear on why it ran no tests, since that test does exist, but I only change docs here anyway.


breathingdust:
LGTM 🚀! Thanks @theherk for the correction!

PR #28682, opened 2023-01-05
docs: Fix SI unit used in maximum zip file from Kb (kilobits) to KB (kilobytes)

MatthewJohn:
### Description
The SI units for the maximum zip file for canary lambda zip upload size were incorrect.
This is also incorrect on the AWS documentation (https://docs.aws.amazon.com/AmazonSynthetics/latest/APIReference/API_CanaryCodeInput.html#synthetics-Type-CanaryCodeInput-ZipFile).
I have reached out to AWS support independently, who have confirmed this to also be incorrect and have been told it will be updated once reviewed internally.

Once I hear back from AWS support that it has been corrected, I can post back here so the live AWS documentation can be confirmed.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@MatthewJohn Thanks for the contribution :tada: :clap:.

MatthewJohn:
@ewbankkit  If it's of interest, AWS have responded and have fixed the upstream documentation :D

PR #28811, opened 2023-01-10
`flex` framework flattener/expander updates

jar-b:
### Description
Several updates and additions to the `flex` package's framework specific helper functions, including:

- Add flatten functions for string list/set/map objects that return null values for nil and empty inputs
- Rename flatten functions for string list/set/map objects that return empty values for nil inputs with the `Legacy` suffix adopted by the value conversion helpers
- Add string list expander functions
- Extend various test cases

The non-legacy flatteners will allow omitted list/set attributes to be set as null for net-new resources, while the legacy versions preserve the empty list/set/map return behavior where needed (tags, refactored resources/data sources, etc.). 

```console
$ go test -count=1 ./internal/flex/...
ok      github.com/hashicorp/terraform-provider-aws/internal/flex       0.257s
```

### DIFF
--- a/internal/flex/framework.go
+++ b/internal/flex/framework.go
@@ -10,6 +10,34 @@ import (
 
 // Terraform Plugin Framework variants of standard flatteners and expanders.
 
+func ExpandFrameworkStringList(ctx context.Context, list types.List) []*string {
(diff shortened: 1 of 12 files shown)


ewbankkit (approved):
LGTM :rocket:.

PR #28827, opened 2023-01-11
.github/workflows: Use hashicorp/ghaction-terraform-provider-release setup-go-version-file input

bflad:
### Description

This input is passed through to actions/setup-go's go-version-file input, similar to what is used in other workflows.

### Relations

N/A

### References

Reference: https://github.com/hashicorp/ghaction-terraform-provider-release/releases/tag/v2.2.0
Reference: https://github.com/hashicorp/ghaction-terraform-provider-release/blob/7e53d9adb57f05cb16c227ef0bfd030060fae82e/.github/workflows/hashicorp.yml#L26-L29

```

...
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,14 +9,6 @@ permissions:
   contents: write
 
 jobs:
-  go-version:
-    runs-on: macos-latest
-    outputs:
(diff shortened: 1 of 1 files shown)


ewbankkit:
Should be backported to the `release/3.x` and `release/2.x` branches.

PR #28894, opened 2023-01-13
Remove bare `error` returns from CRUD functions

gdavison:
Many CRUD functions returned raw API errors rather than adding information on the current operation


ewbankkit (approved):
LGTM :rocket: apart from some linter errors.

Issue #28556, opened 2022-12-23
Avoid direct calls to `Resource.Delete` in sweepers and acceptance tests


ewbankkit:
With a future switch to [`WithoutTimeout` CRUD handlers](https://github.com/hashicorp/terraform-provider-aws/issues/15090), avoid direct calls to `Resource.Delete` in sweepers and acceptance tests.

* In acceptance tests use `acctest.DeleteResource`
* In sweepers use the `sweep.SweepOrchestrator` facility

PR #29021, opened 2023-01-20
Fixes sweepers after `Context` migration

gdavison:
A few sweepers made direct calls to `schema.Resource`'s `Read` function. After the `Context` migration, the legacy CRUD functions are `nil`, so those sweepers panicked.

Closes https://github.com/hashicorp/terraform-provider-aws/issues/28556.


jar-b (approved):
LGTM 👍

PR #29025, opened 2023-01-21
r/cloudfront_distribution: Mark custom_origin_config.origin_ssl_proto…

### LINKED ISSUE #28627: [Docs]: Items in [aws_cloudfront_distribution] -> [origin] -> [custom_origin_config] shoud not be required
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#custom-origin-config-arguments

### Terraform Version

```shell
- 1.4.0 (latest)
- aws provider 4.48
```

### What is the docs issue?

The following elements are listed as required in `aws_cloudfront_distribution -> origin -> custom_origin_config` .

- http_port
- https_port
- origin_ssl_protocols 

However, these elements are not written as required in CloudFormation docs(see: [this](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-customoriginconfig.html)).

There should be no reason for Terraform to require this value. Is there a reason?

### References

I already opened the issue in `hashicorp/terraform` , I prompted reopen within here by team member.

see:
https://github.com/hashicorp/terraform/issues/32440

None

mattburgess:
…cols as optional

### Description

`CustomOriginConfig.OriginSslProtocols` is marked as optional in the [API docs](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_CustomOriginConfig.html) and testing with the AWS CLI shows the docs are accurate.

Whilst here, I noticed that reference 3 in the docs wasn't referenced at all so the reference list has been updated with valid links and correct ordering.

### Relations

Closes #28627 

```
$ make testacc TESTS=TestAccCloudFrontDistribution_noOptionalItems PKG=cloudfront
TF_ACC=1 go test ./internal/service/cloudfront/... -v -count 1 -parallel 20 -run='TestAccCloudFrontDistribution_noOptionalItems'  -timeout 180m
=== RUN   TestAccCloudFrontDistribution_noOptionalItems
=== PAUSE TestAccCloudFrontDistribution_noOptionalItems
=== CONT  TestAccCloudFrontDistribution_noOptionalItems
    distribution_test.go:638: Step 1/2 error: After applying this test step, the plan was not empty.
        stdout:
        
        
        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place
        
          # aws_cloudfront_distribution.no_optional_items will be updated in-place
          ~ resource "aws_cloudfront_distribution" "no_optional_items" {
                id                             = "E3UXZOGO4ES5GS"
                # (17 unchanged attributes hidden)
        
              + origin {
                  + connection_attempts = 3
                  + connection_timeout  = 10
                  + domain_name         = "www.example.com"
                  + origin_id           = "myCustomOrigin"
        
                  + custom_origin_config {
                      + http_port                = 80
                      + https_port               = 443
                      + origin_keepalive_timeout = 5
                      + origin_protocol_policy   = "http-only"
                      + origin_read_timeout      = 30
                      + origin_ssl_protocols     = (known after apply)
                    }
                }
              - origin {
                  - connection_attempts = 3 -> null
                  - connection_timeout  = 10 -> null
                  - domain_name         = "www.example.com" -> null
                  - origin_id           = "myCustomOrigin" -> null
        
                  - custom_origin_config {
                      - http_port                = 80 -> null
                      - https_port               = 443 -> null
                      - origin_keepalive_timeout = 5 -> null
                      - origin_protocol_policy   = "http-only" -> null
                      - origin_read_timeout      = 30 -> null
                      - origin_ssl_protocols     = [
                          - "TLSv1",
                          - "TLSv1.1",
                          - "TLSv1.2",
                        ] -> null
                    }
                }
        
                # (3 unchanged blocks hidden)
            }
        
--- FAIL: TestAccCloudFrontDistribution_noOptionalItems (396.47s)
FAIL
FAIL	github.com/hashicorp/terraform-provider-aws/internal/service/cloudfront	396.549s
FAIL
make: *** [GNUmakefile:92: testacc] Error 1
```

### DIFF
--- a/internal/service/cloudfront/distribution.go
+++ b/internal/service/cloudfront/distribution.go
@@ -559,7 +559,8 @@ func ResourceDistribution() *schema.Resource {
 									},
 									"origin_ssl_protocols": {
 										Type:     schema.TypeSet,
-										Required: true,
+										Optional: true,
+										Computed: true,
 										Elem: &schema.Schema{
 											Type:         schema.TypeString,
 											ValidateFunc: validation.StringInSlice(cloudfront.SslProtocol_Values(), false),
--- a/internal/service/cloudfront/distribution_configuration_structure.go
+++ b/internal/service/cloudfront/distribution_configuration_structure.go
@@ -1013,9 +1013,12 @@ func ExpandCustomOriginConfig(m map[string]interface{}) *cloudfront.CustomOrigin
 		OriginProtocolPolicy:   aws.String(m["origin_protocol_policy"].(string)),
 		HTTPPort:               aws.Int64(int64(m["http_port"].(int))),
 		HTTPSPort:              aws.Int64(int64(m["https_port"].(int))),
-		OriginSslProtocols:     ExpandCustomOriginConfigSSL(m["origin_ssl_protocols"].(*schema.Set).List()),
 		OriginReadTimeout:      aws.Int64(int64(m["origin_read_timeout"].(int))),
 		OriginKeepaliveTimeout: aws.Int64(int64(m["origin_keepalive_timeout"].(int))),
+	}
+
+	if v, ok := m["origin_ssl_protocols"].(*schema.Set); ok && len(v.List()) > 0 {
+		customOrigin.OriginSslProtocols = ExpandCustomOriginConfigSSL(v.List())
 	}
 
 	return customOrigin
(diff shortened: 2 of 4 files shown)


mattburgess:
I'm not quite sure what to do to resolve that "diff on refresh" issue. I note that the other 2 parameters that the API will return if not set, `origin_keepalive_timeout` & `origin_read_timeout` have `Default`s set that match what the API returns and therefore don't have `Compute: true`.

Should we do the same here, or do I need a `CustomizeDiff` function to ignore the diff if `origin_ssl_protocols` isn't set by the user?

mattburgess:
> I'm not quite sure what to do to resolve that "diff on refresh" issue. I note that the other 2 parameters that the API will return if not set, `origin_keepalive_timeout` & `origin_read_timeout` have `Default`s set that match what the API returns and therefore don't have `Compute: true`.
> 
> Should we do the same here, or do I need a `CustomizeDiff` function to ignore the diff if `origin_ssl_protocols` isn't set by the user?

Turns out `Default` isn't valid on a Set so I guess we're down to a `CustomizeDiff` function?

PR #29150, opened 2023-01-28
aws_acmpca_certificate_authority doc clarifications

### LINKED ISSUE #28998: [Docs]: aws_acmpca_certificate_authority- Insufficient information for enable parameter
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/3.70.0/docs/resources/acmpca_certificate_authority

### Description

- In the terraform documentation of aws_acmpca_certificate_authority resource description of enable parameter is insufficient.

- enabled parameter is specified based on whether we want the certificate authority enabled or disabled, but this state change is possible only when PCA is in Active state.

- When we create a PCA from terraform it is created with 'pending_certificate' status, some additional steps are included to make status of PCA 'Active', which is mentioned in the same documentation.

- But the usage that you can disable the PCA only when it is in Active status it is not mentioned.

- Even if we specify false value for enable parameter, terraform applies it as a true hence during 2nd terraform apply there is a change in the configuration, and due to this change terraform throws an error because PCA can't be modified when it is in pending_certificate state. 

### References

_No response_

Yes

mattburgess:
### Relations

Closes #28998

```

...
```

### DIFF
--- a/website/docs/r/acmpca_certificate_authority.html.markdown
+++ b/website/docs/r/acmpca_certificate_authority.html.markdown
@@ -10,7 +10,7 @@ description: |-
 
 Provides a resource to manage AWS Certificate Manager Private Certificate Authorities (ACM PCA Certificate Authorities).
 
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@mattburgess Thanks for the contribution :tada: :clap:.

PR #29217, opened 2023-02-02
Moved RequiredWith from wrong attribute on an appmesh route

bjernie:
If you create an aws_appmesh_route and provide a spec.0.grpc_route.0.match.service_name it gives an error that you also need to specify a spec.0.grpc_route.0.match.method_name when it is actually the other way around.

### DIFF
--- a/.changelog/29217.txt
+++ b/.changelog/29217.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_appmesh_route: Fix RequiredWith setting for `spec.0.grpc_route.0.match.0.method_name` attribute
+```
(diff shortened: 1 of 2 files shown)


jar-b (approved):
LGTM 🎉 

`plan` now returns the proper error message:

```
│ Error: Missing required argument
│
│   with aws_appmesh_route.test,
│   on main.tf line 58, in resource "aws_appmesh_route" "test":
│   58:       match {
│
│ "spec.0.grpc_route.0.match.0.method_name": all of `spec.0.grpc_route.0.match.0.method_name,spec.0.grpc_route.0.match.0.service_name` must be specified
```

GRPC Route Match Docs: https://docs.aws.amazon.com/app-mesh/latest/APIReference/API_GrpcRouteMatch.html

jar-b:
Thanks for your contribution, @bjernie! 👏

PR #29291, opened 2023-02-07
Add default_value to the cost category data source.

jimrazmus:
### Description
Adds 'default_value' to the cost category data source.

This is an important addition given that a default_value may be an instrumental part of a rule set. The 'default value' may be capturing material costs not explicitly categorized by other rules.

### Relations
NA

### References
NA

I didn't run acceptance tests as I don't have an AWS management account suitable for testing.


ewbankkit:
@jimrazmus Thanks for the contribution :tada: :clap:.

PR #29097, opened 2023-01-25
r/aws_networkmanager_core_network_policy_attachment

GlennChia:
### Description

Creates a new resource that attaches a policy to an existing Core Network. When doing this, it sets the latest policy as the `LIVE` policy. Refer to the [Core network policies documentation](https://docs.aws.amazon.com/network-manager/latest/cloudwan/cloudwan-policy-change-sets.html) for more information.

#### Clarifications and design choices

1. Since this resource modifies the `policy_document`, which is currently defined as an argument for the `aws_networkmanager_core_network` resource, it will lead to diffs after refresh. This is solved by changing it to `optional` and `computed`.

2. `optional` and `computed` were also chosen because `policy_document` can also be updated since am optional `create_base_policy` was introduced to create a base Core Network policy document at create-time. This base policy will be set to the `LIVE` state. It is important that the Core Network has a `LIVE` policy because this is required if VPC attachments were to be attached to the Core Network at create-time. Hence, in the next plan or apply, the `policy_document` needs to be `computed` to read the policy created from the `create_base_policy` argument. This new argument is not returned by the read APIs and is added to ImportStateVerifyIgnore.

3. The base policy requires a region. Hence, a `base_policy_region` argument was introduced for practitioners to pass in their own region. This is useful in scenarios where an organization may not permit resources to be deployed in certain regions and practitioners want to have the flexibility to deploy resources into an allow-listed region. This argument is optional and defaults to the region defined in the `providers` block. This new argument is not returned by the read APIs and is added to ImportStateVerifyIgnore.

4. The core_network resource also has 2 computed attributes `segments` and `edges` that are computed after the policy has been applied. With the new `aws_networkmanager_core_network_policy_attachment` resource, the computed arguments are not computed immediately after the apply since customized_diff logic will not be triggered because the policy document is not managed by the `aws_networkmanager_core_network` resource. This can potentially be solved by implementing a data source for core network just for reading the latest attributes.
https://github.com/hashicorp/terraform-provider-aws/blob/88e19859acfb992526694c036b1d236f90079176/internal/service/networkmanager/core_network.go#L340-L349

### Relations

Closes #28867

### References

- [ExecuteCoreNetworkChangeSet](https://docs.aws.amazon.com/networkmanager/latest/APIReference/API_ExecuteCoreNetworkChangeSet.html)
- [PutCoreNetworkPolicy](https://docs.aws.amazon.com/networkmanager/latest/APIReference/API_PutCoreNetworkPolicy.html)

[...]

PR #29303, opened 2023-02-08
Docs/aws_networkmanager_core network_policy_attachment

GlennChia:
### Description

Add more information in the documentation

- Describe how the policy attachment works. This first puts the new policy document to the core network before executing the change set, which sets the policy to `LIVE`
- Note that deleting the resource does not delete the policy it manages nor does it revert the current `LIVE` policy to a previous iteration

### Relations

Relates #29097

```

...
```


jar-b (approved):
LGTM

PR #29353, opened 2023-02-11
Replace heredoc with aws_iam_policy_document datasource (d/*)

mattburgess:
### Description

Replaces all instances of heredocs in datasource examples with an `aws_iam_policy_document` datasource so that users can take advantage of the validation and syntax highlighting that the latter provides.

### Relations

Relates #17714

```

...
```


jar-b (approved):
LGTM 👍

Issue #26797, opened 2022-09-13
FR: aws_organizations_account account closure with all SCP's detached causes error.


susiesolis:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As a practice, upon account creation, all policies are attached through Terraform while removing `FULLAWSAccess` using a provisioner. When Terraform tries to destroy the account it first tries to destroy resource `aws_organization_policy_attachment` and hence tries to detach all SCPs one by one, but fails to detach the last SCP from account as AWS does not allow detaching the last SCP from account.

This results in the following error:

```
Error: ConstraintViolationException: You cannot remove the last policy attached to the specified target. You must have at least one attached at all times.
{
  RespMetadata: {
    StatusCode: 400,
    RequestID: "cdb81a90-4a4d-45fc-bdab-d3408028d96b"
  },
  Message_: "You cannot remove the last policy attached to the specified target. You must have at least one attached at all times.",
  Reason: "MIN_POLICY_TYPE_ATTACHMENT_LIMIT_EXCEEDED"
}
```

1. Is there a way or switch outside of manually manipulating the state that will not attempt to detach all SCP's prior to deleting the `aws_organizations_account` ?

2. Can there be a dependency such that resource "aws_organizations_policy_attachment" is destroyed after resource "aws_organizations_account" "account" is destroyed? At which point we do not care about deployment error given account is already closed.

### References

*  https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/organizations_account
*  https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/organizations_policy_attachment

YakDriver:
I don’t have the provider error so can’t see exactly where it’s happening. Assuming it’s coming from `aws_organizations_policy_attachment` when it attempts to detach the policy and it happens to be the last, we could decide that eating that specific error is an acceptable risk or add an argument (something like `ignore_last_policy`) that allows that error to be skipped.

susiesolis:
[...]

PR #29382, opened 2023-02-13
AWS Organizations Policy Attachment: Add `ignore_last_policy_exception` attribute

jar-b:
### Description
Adds an optional `ignore_last_policy_exception` attribute, which when set to `true` will allow the resource to gracefully handle cases where the policy is the last attached to a given target. Instead of throwing an error, the resource is simply removed from state, allowing other related account destruction operations to complete.

### Relations
Relates #26797

```
$ make testacc TESTS=TestAccOrganizations_serial PKG=organizations

...
```


YakDriver (approved):
This LGTM 🎉 . In places we have testing challenges, I think the "best effort" and some reliance on the community is totally reasonable. The community would probably be okay with more reliance if it meant a bit faster turn around I'm guessing.

PR #29491, opened 2023-02-18
aws_cloudfront_public_key: Update docs for issue with string parameters for `encoded_key` when not using `file` terraform function

### LINKED ISSUE #20081: Updating aws_cloudfront_public_key encoded_key with the same value forces replacement
Changing 
```
resource "aws_cloudfront_public_key" "key" {
  name        = "my-cf-pubkey"
  encoded_key = file("${var.cloudfront_public_key_path}")
}
```
to
```
resource "aws_cloudfront_public_key" "key" {
  name        = "my-cf-pubkey"
  encoded_key = var.cloudfront_public_key
}
```
forces recreation despite no diff existing in the `encoded_key` value. Introducing changes to the variable shows up correctly as part of the diff for `encoded_key`.

Creating a new key using `var.cloudfront_public_key` shows no differences between both key values when comparing them in AWS/CloudFront/Public Keys.

Running Terraform 1.0.1 with an S3 and DynamoDB back-end.

teddylear:
### Description
aws_cloudfront_public_key: Update docs for issue with string parameters for `encoded_key` when not using `file` terraform function needing newline at the end. 

I'm also willing to add a follow up PR for a check in the resource for this issue, but figured it was better to add this to docs now with the fix from the linked issue as I ran into this today as well and spent way too much time debugging it until I found thread. 

Partially resolves #20081


breathingdust:
Hi @teddylear! Thanks for the contribution! 🚀 I removed the changelog entry as they are not necessary for [documentation updates](https://hashicorp.github.io/terraform-provider-aws/changelog-process/#changes-that-should-not-have-a-changelog-entry).

PR #29573, opened 2023-02-22
Move CODEOWNERS to root and add commented out default owner line

breathingdust:
### Description

Adds commented out default owner for attribution without review requested noise.


jar-b (approved):
LGTM 🎉

PR #29661, opened 2023-02-25
aws_efs_file_system_policy - updating example

### LINKED ISSUE #29218: [Docs]: Resource: aws_efs_file_system_policy
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_file_system_policy

### Description

I do not think it is appropriate to have such a dangerous example setting.. 

The use of `bypass_policy_lockout_safety_check` is fairly obvious and shouldn't need an example... it's just a boolean. Or at the very least, if you are going to provide this attribute's use in your top example, then do it in a way that won't result in an orphan resource for those who try it.

### References

_No response_

No

bennylu2:
### Description
Example was setting `bypass_policy_lockout_safety_check` to `true` which left orphan resources when testing out. Users can refer to documentation as it's an optional boolean argument with good description

### Relations
Closes #29218


breathingdust:
LGTM 🚀 Thanks @bennylu2 for the contribution!

PR #29677, opened 2023-02-27
Fix route53_delegation_set argument description

bigwheel:
### Description

This PR fixes route53_delegation_set argument description.
Argument `id` is of delegation set, not zone.


breathingdust:
LGTM 🚀 Thanks @bigwheel for the contribution!

PR #29686, opened 2023-02-27
Correct link for using-object-lock-configuration

paullschock:
Correct case for link `U` -> `u`

### Description

Noticed the link didn't work, seems like the upper case was incorrect: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#using-object-lock-configuration

### Relations

N/A

Did not run, feel free to toss this.


breathingdust:
LGTM 🚀 Thanks @paullschock for the contribution!

Issue #29729, opened 2023-03-01
[Docs]: Code Example for aws_sesv2_email_identity (BYODKIM) is misleading


Asteao:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sesv2_email_identity

### Description

In the example code for setting up domain signing with a BYODKIM key a full PEM key is base64 encoded and passed to the `domain_signing_private_key` attribute https://github.com/hashicorp/terraform-provider-aws/blob/main/website/docs/r/sesv2_email_identity.html.markdown?plain=1#L57
However, if you actually try to do this you will be uploading invalid data to the AWS API. The API will still accept it, but it will never successfully verify the domain.

The AWS API documentation states the value needs to be base64 encoded, which is true. https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_DkimSigningAttributes.html 
However, the user guide shows that the value of `DomainSigningPrivateKey` should be the raw PEM value of the cert but without the `-----BEGIN PRIVATE KEY-----` and `-----END PRIVATE KEY-----` lines and without newline characters.  https://docs.aws.amazon.com/ses/latest/dg/send-email-authentication-dkim-bring-your-own.html

It doesn't look like the `tls_private_key` has a suitable attribute which would match the format required by the `aws_sesv2_email_identity` resource. Some formatting would need to be done on the PEM key before it can be passed to the `aws_sesv2_email_identity` resource.

I'm not sure what the best code example would be for this particular case, but the existing documentation is misleading. In the very least the description for the `domain_signing_private_key` should be modified to include a note similar to the AWS documentation. 

### Would you like to implement a fix?

None

PR #29730, opened 2023-03-01
[Docs]: Code Example for aws_sesv2_email_identity (BYODKIM) is misleading

Asteao:
https://github.com/hashicorp/terraform-provider-aws/issues/29729

### DIFF
--- a/website/docs/r/sesv2_email_identity.html.markdown
+++ b/website/docs/r/sesv2_email_identity.html.markdown
@@ -46,15 +46,11 @@ resource "aws_sesv2_email_identity" "example" {
 #### DKIM Signing Attributes (BYODKIM)
 
 ```terraform
-resource "tls_private_key" "example" {
-  algorithm = "RSA"
-}
-
 resource "aws_sesv2_email_identity" "example" {
   email_identity = "example.com"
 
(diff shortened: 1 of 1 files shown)


kamilturek (approved):
LGTM! That's an important note.

Asteao:
@kamilturek The note helps, but the code example is still misleading. Do you have any suggestions on how this could be made more clear?
The only thing that comes to mind is using a local with the expected value and calling that variable for the attribute.

kamilturek:
@Asteao I think what you suggested will be fine and very explanatory. Another idea would be to use a variable without explicitly specifying its value. It's simpler but seems to be less clear.

```terraform
resource "aws_sesv2_email_identity" "example" {
  email_identity = "example.com"

  dkim_signing_attributes {
    domain_signing_private_key = base64encode(var.private_key)
    domain_signing_selector    = "example"
  }
}
```

Asteao:
@kamilturek I have update the example with an example text, since none of the other attributes reference variables it seems strange to use a var for this attribute. If you're okay with this change I can merge it after your approval.

jar-b:
Thanks for your contribution, @Asteao!

PR #29796, opened 2023-03-05
aws_codebuild_project - update secondary_artifacts

### LINKED ISSUE #29614: [Docs]: conflicting information about type under Secondary Artifacts
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/codebuild_project#arn

### Description

Conflicting information under secondary artifacts.   First it states in some of the block variables if Type is this or that,  but  the type description states only valid value is S3.   So is it true that you can have other types,  or is this whole block only for S3?     Also about documentation  in general ,  all these "variables/parameters" must include in the same description the type.(string, number, list, bool etc) 

[location](Optional) Information about the build output artifact location. If type is set to CODEPIPELINE or NO_ARTIFACTS, this value is ignored. If type is set to S3, this is the name of the output bucket. If path is not also specified, then location can also specify the path of the output artifact in the output bucket

[type]- (Required) Build output artifact's type. The only valid value is S3.

(https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/codebuild_project#type) 

### References

_No response_

None

bennylu2:
### Description
Updates to outdated descriptions for `secondary_artifacts` block

### Relations
Closes #29614

### References
https://docs.aws.amazon.com/codebuild/latest/APIReference/API_CreateProject.html
https://docs.aws.amazon.com/codebuild/latest/APIReference/API_ProjectArtifacts.html

### DIFF
--- a/website/docs/r/codebuild_project.html.markdown
+++ b/website/docs/r/codebuild_project.html.markdown
@@ -340,15 +340,15 @@ See [ProjectFileSystemLocation](https://docs.aws.amazon.com/codebuild/latest/API
 ### secondary_artifacts
 
 * `artifact_identifier` - (Required) Artifact identifier. Must be the same specified inside the AWS CodeBuild build specification.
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks good to me! Thank you again for taking the time to raise this! 🚀 

P.s. [If you're interested...](https://hashicorp.github.io/terraform-provider-aws/#join-the-contributors-slack)

PR #29834, opened 2023-03-07
Remove note on Terraform version compatibility changes v4 upgrade guide

breathingdust:
### Description

When v4 of the provider was launched, there was a plan to deprecate older Terraform versions with v5.0. This is no longer the case, so this section should be removed.


jar-b (approved):
LGTM

marcosentino (approved):
Approved.

PR #29835, opened 2023-03-07
Fix file name reference for docs

KyMidd:
### Description

I [previously updated this doc](https://github.com/hashicorp/terraform-provider-aws/pull/26509/files), and realized I missed one reference to a filename. This change makes the output file generated by the `archive_file` match the name of the file ingested by `aws_lambda_function`. 

cc: @breathingdust since we worked together on the last one. 

### Relations

Relates #26509

n/a


breathingdust:
LGTM 🚀 Thanks for the correction @KyMidd!

PR #29905, opened 2023-03-10
Increase aws_dms_replication_instance creation timeout

pedroarthur:
### Description

I've been setting up DMS replication instances for the past two weeks, and most `terraform apply` fail because of the creation timeout. However, the AWS DMS console shows that the instances are still creating. And, almost 2 to 3 minutes later, the console shows the instances as Available. Therefore, this seems the case to increase the instance creation timeout.


johnsonaj:
@pedroarthur thanks for the contribution! 🎉

Issue #29871, opened 2023-03-08
[Enhancement]: Control when `aws_ecs_task_execution` is executed


bryantbiggs:
### Description

Carrying over from here since the PR will close once the release is created https://github.com/hashicorp/terraform-provider-aws/pull/29783

Executing the `RunTask` API call on each read will inevitably be problematic. This issue is to brainstorm and chat about possible further enhancements that would give users more control over when tasks are run, or re-run. 

### Affected Resource(s) and/or Data Source(s)

* aws_ecs_task_execution data source

### Potential Terraform Configuration

```terraform
resource "aws_ecs_task_execution" "everytime" {
  cluster         = aws_ecs_cluster.test.id
  task_definition = aws_ecs_task_definition.test.arn
  desired_count   = 1
  launch_type     = "FARGATE"
  network_configuration {
    subnets          = aws_subnet.test[*].id
    security_groups  = [aws_security_group.test.id]
    assign_public_ip = false
  }

  triggers = {
    everytime = timestamp()
  }
}

resource "aws_ecs_task_execution" "flag" {
  cluster         = aws_ecs_cluster.test.id
  task_definition = aws_ecs_task_definition.test.arn
  desired_count   = 1
  launch_type     = "FARGATE"
  network_configuration {
    subnets          = aws_subnet.test[*].id
    security_groups  = [aws_security_group.test.id]
    assign_public_ip = false
  }

  triggers = {
    flag = true # flip to false to re-run
  }
}
```

### References

https://github.com/hashicorp/terraform-provider-aws/pull/29783

### Would you like to implement a fix?

None

md5:
It feels like there is already a decent amount of control of when you _do_ want an execution to happen using either `reference_id` to provide a reference ID that only changes when your logic dictates (perhaps based on an external variable set by a CI pipeline, or based on `timestamp()` as in your example if you want it to invoke whenever the other arguments are resolvable, whether during `plan` or `apply`).

For me, what I really want is to ensure that this _does not_ run during `plan`, only during `apply`. For that, I feel like the same duality that we currently have between the `aws_lambda_function_invocation` data source and the `aws_lambda_function_invocation` resource would be helpful. In other words, I would like to see `resource "aws_lambda_function_invocation"` introduced alongside `data "aws_lambda_function_invocation"` so that I know that the run task invocation will only happen during an `apply`.

md5:
[...]

PR #29783, opened 2023-03-03
New Data Source: `aws_ecs_task_execution`

jar-b:
### Description
The `aws_ecs_task_execution` data source will enable use cases when a one-time task needs to run on an existing ECS cluster.

As this data source wraps the [`RunTask`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html) API, it will provision tasks on every read operation where the input attributes are known. This may include `terraform plan` commands, and should be used with caution for any tasks that cannot safely run multiple instances in parallel.

### Relations
Adapted from #12096
Closes #1703

### References
This data source was modeled after [`aws_lambda_function_invocation`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/lambda_invocation).

### Output from Acceptance Testing
```console
$ make testacc PKG=ecs TESTS=TestAccECSTaskExecutionDataSource_
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/ecs/... -v -count 1 -parallel 20 -run='TestAccECSTaskExecutionDataSource_'  -timeout 180m
=== RUN   TestAccECSTaskExecutionDataSource_basic
=== PAUSE TestAccECSTaskExecutionDataSource_basic
=== RUN   TestAccECSTaskExecutionDataSource_tags
=== PAUSE TestAccECSTaskExecutionDataSource_tags
=== CONT  TestAccECSTaskExecutionDataSource_basic
=== CONT  TestAccECSTaskExecutionDataSource_tags
--- PASS: TestAccECSTaskExecutionDataSource_basic (310.88s)
--- PASS: TestAccECSTaskExecutionDataSource_tags (310.89s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ecs        313.941s
```

PR #29978, opened 2023-03-14
r/aws_ecs_task_execution: Add new resource

md5:
### Description

This pull request adds an `aws_ecs_task_execution` resource to complement `data.aws_ecs_task_execution` as added in #29783

### Relations

Relates #29871

### References

This is almost an exact copy of `task_execution_data_source.go` and its tests. The main difference between the resource schemas is that `ForceNew` has been added to all non-computed fields in the schema. There are possibly better patterns for making the code between the two more DRY.

I'm unable to run `testacc` successfully because a process in our account automatically tags the default security group of any VPC if it has any `ingress` or `egress` rules. This causes the `refresh` to always come back dirty...

### DIFF
--- a/internal/service/ecs/task_execution_test.go
+++ b/internal/service/ecs/task_execution_test.go
@@ -1,0 +1,181 @@
+package ecs_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/service/ecs"
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-provider-aws/internal/acctest"
+)
+
+func TestAccECSTaskExecutionResource_basic(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping long-running test in short mode")
+	}
+
+	rName := sdkacctest.RandomWithPrefix(acctest.ResourcePrefix)
+	resourceName := "aws_ecs_task_execution.test"
+	clusterName := "aws_ecs_cluster.test"
+	taskDefinitionName := "aws_ecs_task_definition.test"
+
+	resource.ParallelTest(t, resource.TestCase{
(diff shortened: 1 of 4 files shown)


internal/service/ecs/task_execution_test.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	rName := sdkacctest.RandomWithPrefix(acctest.ResourcePrefix)
+	resourceName := "aws_ecs_task_execution.test"
+	clusterName := "aws_ecs_cluster.test"
+	taskDefinitionName := "aws_ecs_task_definition.test"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			acctest.PreCheck(t)
```
md5, line 25: It's unclear to me what this "not enough arguments in call to acctest.PreCheck" error would have me do instead. All the other examples I see in the repo are calling `acctest.PreCheck` and I don't see any clear alternatives in `internal/acctest`

mcfedr:
Please someone merge this, it would be so helpful!

mikicz:
Is there any chance this will get merged? I would like to use it to run an ECS task only when applying, and this should do just that

briantist:
I would love to see support for CRUD as well just like `aws_lambda_invocation` resource

briantist:
_comment spam made necessary by ill-conceived inactivity bot_

> Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request.

This policy doesn't work when you consider lack of comments as inactivity.

> Marking this pull request as stale due to inactivity. This helps our maintainers find and focus on the active pull requests. If this pull request receives no comments in the next 30 days it will automatically be closed. Maintainers can also remove the stale label.

md5:
@justinretzolk I saw there might be some movement on creating an `action` for `ecs:RunTask` (cf #44688), which would actually fit our use case better. I can't actually think of a use case where I would want a `resource` for `ecs:RunTask` instead of an `action`, so I'm wondering whether it makes sense to leave this PR open given that it has been sitting around for three years at this point

PR #30092, opened 2023-03-17
framework validators and sweeper helpers

johnsonaj:
### Description

Adds timestamp and ARN validators for the framework. 
Add `UpgradeState()` method to framework resource wrapper
Add helper for framework supplemental attributes

```

...
```

### DIFF
--- a/internal/framework/validators/arn.go
+++ b/internal/framework/validators/arn.go
@@ -1,0 +1,40 @@
+package validators
+
+import (
+	"context"
+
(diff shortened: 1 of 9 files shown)


internal/framework/validators/arn.go
ewbankkit, line 1: There's also `internal/framework/types/arn.go` - You may be able to share some code?

gdavison: +1 to this. We could have a core types library with SDK and Framework interfaces to it

PR #30097, opened 2023-03-17
Adds region overrides for services in TeamCity

gdavison:
### Description

Adds region overrides for services such as Lightsail or WAF Classic which require `us-east-1`. Support for the GovCloud partition will be added later. Does not allow specific tests to override the required region.

Note: all Route 53 tests are run in `us-east-1`, though only the DNSSEC tests require it

### Relations

Relates #29964


ewbankkit (approved):
LGTM :rocket:.

PR #30219, opened 2023-03-22
Updates version of `golangci-lint` used in GitHub actions

gdavison:
### Description

Updates GitHub Actions to use same version of `golangci-lint` as set in `.ci/tools/go.mod`

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -27,7 +27,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.51.0
+          version: v1.52.0
           args: --config .ci/.golangci.yml
   golangci-lintb:
     name: 2 of 2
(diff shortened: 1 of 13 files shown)


ewbankkit:
Relates https://github.com/hashicorp/terraform-provider-aws/pull/29267.

.github/workflows/golangci-lint.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.51.0
+          version: v1.52.0
```
ewbankkit, line 30: The version was pinned in https://github.com/hashicorp/terraform-provider-aws/pull/29272.
We may want to revert to `latest`?

gdavison: My plan is to make it use the version set in `.ci/tools/go.mod`, so that the same version can be used locally and in GH

ewbankkit: 👍

jar-b (approved):
LGTM 🎉

Issue #30272, opened 2023-03-27
[Docs]: Missing detail about KMS in secretsmanager_secret.html.markdown which prevents cross-account access


v-rosa:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret#kms_key_id

### Description

If a SM instance is intended to be access cross-account the kms-key value to be passed must be the full arn. This little detail is present in https://awscli.amazonaws.com/v2/documentation/api/2.0.33/reference/secretsmanager/create-secret.html but missing in the provider.

This leads to situations like this:

* SM instance created in `acc-a` with `cmk` key-id instead of key arn (e.g. `12345` instead of `arn:aws:kms:region:acc-a:12345`)
* cmk and sm are shared cross account with `acc-b`.
* A given principal in `acc-b` tries to execute `PutSecretValue`  in the SM instance but gets the following error:

```log
Secrets Manager can't encrypt the secret value: Key 'arn:aws:kms:us-east-1:acc-b:key/12345' does not exist (Service: AWSKMS; Status Code: 400; Error Code: NotFoundException; Request ID: ....
```

### References

https://awscli.amazonaws.com/v2/documentation/api/2.0.33/reference/secretsmanager/create-secret.html

### Would you like to implement a fix?

Yes

breathingdust:
Closed in: #30273

PR #30273, opened 2023-03-27
Update secretsmanager_secret.html.markdown

v-rosa:
Add a reference about the strict usage of KMS arn instead of KMS key id, if the SM instance is intended to be used cross account.

### Description

If a SM instance is intended to be access cross-account the kms-key value to be passed must be the full arn. This little detail is present in https://awscli.amazonaws.com/v2/documentation/api/2.0.33/reference/secretsmanager/create-secret.html but missing in the provider.

This leads to situations like this:

* SM instance created in `acc-a` with `cmk` key-id instead of key arn (e.g. `12345` instead of `arn:aws:kms:region:acc-a:12345`)
* cmk and sm are shared cross account with `acc-b`.
* A given principal in `acc-b` tries to execute `PutSecretValue`  in the SM instance but gets the following error:

```log
Secrets Manager can't encrypt the secret value: Key 'arn:aws:kms:us-east-1:acc-b:key/12345' does not exist (Service: AWSKMS; Status Code: 400; Error Code: NotFoundException; Request ID: ....
```

### Relations

Relates #30272


breathingdust:
LGTM 🚀 Thanks for the clarification @v-rosa!

PR #30276, opened 2023-03-27
Switch over to using external source for GHA secrets for community

justinretzolk:
### Description

This PR changes how we create three GitHub Actions Secrets to help with automating via Actions.

### Relations

Before: #30277

### References

- [`github_repository_file` data source](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/repository_file)
- [`github_actions_secret` resource](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_secret)
- [`github_team` data source](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/team)

### Output from Acceptance Testing

N/a, Actions

PR #30277, opened 2023-03-27
Add Community Check reusable workflow

justinretzolk:
### Description

This PR adds a reusable workflow that can be used to determine whether an actor is a Core Contributor, Partner, or Maintainer, for use with other GitHub Actions that depend on these determinations

### Relations

Depends on: #30276
Before: #30279

### References

- [GitHub: Reusing Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)
- [GitHub: Security Hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) (This has info on the reasoning behind the `permissions` block and why we need to `add-mask`)

N/a, Actions

### Usage

Once this is (and the PR that this depends on) is merged, the workflow can be used something like this:

```yaml
name: Label if user is on a list
on: [issues]

jobs:
  community_check:
    uses: ./.github/workflows/community-check.yml # Note: We cannot version pin for local workflows.
    secrets: inherit # The reusable workflow needs access to the repository's GitHub Actions Secrets

  label_groups:
    needs: community_check # Mark this job as dependent on the previous
    runs-on: ubuntu-latest
    steps: # these labels are made up for the sake of showing what the workflow can do
      - name: Label Contributors
        if: needs.community_check.outputs.core_contributor
        uses: actions-ecosystem/action-add-labels@v1
        with:
          labels: contributor
      - name: Label Maintainers
        if: needs.community_check.outputs.maintainer
        uses: actions-ecosystem/action-add-labels@v1
        with:
          labels: maintainer
      - name: Label Partners
        if: needs.community_check.outputs.partner
        uses: actions-ecosystem/action-add-labels@v1
        with:
          labels: partner
```

### DIFF
--- a/.github/workflows/community-check.yml
+++ b/.github/workflows/community-check.yml
@@ -17,7 +17,7 @@ jobs:
     name: Check community lists for username
     runs-on: ubuntu-latest
     outputs:
-      is_contributor: ${{ steps.determination.outputs.is_contributor }}
+      is_core_contributor: ${{ steps.determination.outputs.is_core_contributor }}
       is_maintainer: ${{ steps.determination.outputs.is_maintainer }}
       is_partner: ${{ steps.determination.outputs.is_partner }}
     steps:
--- a/.github/workflows/community-check.yml
+++ b/.github/workflows/community-check.yml
@@ -1,0 +1,54 @@
+name: Community Check
+
+on:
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🎉

PR #30427, opened 2023-04-03
Fix conditionals to compare strings when using community check

breathingdust:
### Description

These conditionals are not working correctly as they assume that the output type is falsy when its actually a string and no implicit conversion occurs. Update these to compare with a string value


ewbankkit (approved):
LGTM :rocket:.

PR #29024, opened 2023-01-21
Fix default values for launch template metadata.

lvthillo:
### Description

Launch Template metadata options does not use default values as [described in the docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httptokens).

Example (`4.50`):

```hcl
resource "aws_launch_template" "lt" {
  name                   = "test-lt"
  image_id               = "ami-0fe0b2cf0e1f25c8a"
  instance_type          = "t3.small"
  vpc_security_group_ids = ["sg-12345678"]

  metadata_options {
    http_tokens = "optional"
  }
}
```

Plan:
```hcl
  + metadata_options {
      + http_endpoint               = (known after apply)
      + http_protocol_ipv6          = "disabled"
      + http_put_response_hop_limit = (known after apply)
      + http_tokens                 = "optional"
      + instance_metadata_tags      = "disabled"
    }
```
Describe using AWS CLI after apply
```json
  "LaunchTemplateData": {
      "ImageId": "ami-0fe0b2cf0e1f25c8a",
      "InstanceType": "t3.small",
      "UserData": "",
      "SecurityGroupIds": [
          "sg-12345678"
      ],
      "MetadataOptions": {
          "HttpProtocolIpv6": "disabled"
      }
}
```

By using this PR + same TF code:
Plan:

```hcl
  + metadata_options {
      + http_endpoint               = "enabled"
      + http_protocol_ipv6          = "disabled"
      + http_put_response_hop_limit = 1
      + http_tokens                 = "optional"
      + instance_metadata_tags      = "disabled"
    }
```

Describe using CLI:
```json
    "LaunchTemplateData": {
        "ImageId": "ami-0fe0b2cf0e1f25c8a",
        "InstanceType": "t3.small",
        "UserData": "",
        "SecurityGroupIds": [
            "sg-12345678"
        ],
        "MetadataOptions": {
            "HttpTokens": "optional",
            "HttpPutResponseHopLimit": 1,
            "HttpEndpoint": "enabled",
            "HttpProtocolIpv6": "disabled",
            "InstanceMetadataTags": "disabled"
        }
    }
```

### Relations

Relates #12564

### Output from Acceptance Testing

```
$ make testacc TESTS=TestAccEC2LaunchTemplate_metadataOptions PKG=ec2
[...]

PR #30520, opened 2023-04-06
Remove incorrect default values from documentation

gdavison:
### Description

Removes default values that were incorrectly documented. The default value for all fields should be blank, to give the Launch Template an unspecified value.

### Relations

Relates #29024


ewbankkit (approved):
LGTM :rocket:.

Issue #23787, opened 2022-03-21
AWS Client VPN Resource Timeouts are too low


jalavoy:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform CLI and Terraform AWS Provider Version
Terraform: 1.1.7
AWS Provider Version: 4.5.0

### Affected Resource(s)

* aws_ec2_client_vpn_authorization_rule
* aws_ec2_client_vpn_route

### Terraform Configuration Files

```hcl
resource "aws_ec2_client_vpn_authorization_rule" "remote" {
  for_each               = { for role in local.roles : "${role.role}-${role.cidr}" => role }
  client_vpn_endpoint_id = aws_ec2_client_vpn_endpoint.vpn.id
  target_network_cidr    = each.value.cidr
  access_group_id        = each.value.role
  description            = "${each.value.role}-${each.value.cidr}"
}

resource "aws_ec2_client_vpn_route" "route_1" {
  for_each               = toset(var.peer_cidrs)
  client_vpn_endpoint_id = aws_ec2_client_vpn_endpoint.vpn.id
  destination_cidr_block = each.value
  target_vpc_subnet_id   = tolist(module.base_infrastructure_state.private_subnet_ids)[0]
}
```

### Expected Behavior

Resources would be created/deleted cleanly. Timeout needs to be raised on these resources to 2 minutes or more. This can be done in hcl, but the default should be increased.

### Actual Behavior

The timeout on these resources is hard coded to 1m. In my tests, creating and destroying these resources can take anywhere from 50-75 seconds. So about half the time the transaction fails. Which results in tainted resources that have to be destroyed and created again. When you have a number of routes/auth rules, this basically can never complete successfully.

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

Raising this value with the `timeouts` metadata block to 2m solved this issue for me.

* #0000

nitrocode:
Previous possibly related issues

* https://github.com/hashicorp/terraform-provider-aws/issues/15680
* https://github.com/hashicorp/terraform-provider-aws/issues/19750

[...]

PR #30552, opened 2023-04-09
Update default timeout for aws_ec2_client_vpn_route

asaf400:
When creating the resource aws_ec2_client_vpn_route, AWS takes between 2 and 3 minutes to create and activate the route, That means the default setting of 1 minute, is not enough, and would most likely fail to either create OR delete.

I propose to raise the default timeout to 4 minutes, which is a lot of time, sure, but  for https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/eks/node_group.go#L38 Node_group timeout is an hour, but node groups are usually active within the first 15 minutes, so I took some headroom as well..

This, _to the best of my knowledge is not an open issue_, but rather something I encountered while creating a resource of 
 `aws_ec2_client_vpn_route`.
 
 Closes: #23787.

### DIFF
--- a/.changelog/30552.txt
+++ b/.changelog/30552.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_ec2_client_vpn_route: Increase Create and Delete timeouts to 4 minutes
+```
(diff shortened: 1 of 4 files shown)


asaf400:
Feel free to modify my PR 👐

ewbankkit (approved):
LGTM :rocket:.

PR #30641, opened 2023-04-11
Typo in ECS Service doc

ktmorgan:
### Description
As title says. This doesn't line up with the actual schema of the object: https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ecs/service.go#L60


breathingdust:
LGTM 🚀 Thanks for the correction @ktmorgan!

PR #30648, opened 2023-04-11
Filter out dependabot from feed pr open action

breathingdust:
### Description

Reduces noise to the channel by filtering out dependabot PR open's


jar-b (approved):
LGTM 🤖

Issue #27457, opened 2022-10-25
[Bug]: Error: unexpected format for ID (arn:aws:events:eu-west-1:xxx:event-bus/aws.partner/auth0.com/xxx/auth0.logs/terraform-20221024154539698300000003), expected EVENTBUSNAME/RULENAME or RULENAME


dascar5:
### Terraform Core Version

1.3.3

### AWS Provider Version

4.36.1

### Affected Resource(s)

aws_cloudwatch_event_rule

### Expected Behavior

Expected behavior is for it to actually deploy properly.

### Actual Behavior

Every other resource deploys fine, except the rules themselves.
They throw the error snippet I posted below, complaining about format. There was this issue (https://github.com/hashicorp/terraform-provider-aws/issues/18431) that implies this was solved with a newer version, but using the latest of everything (TF, auth0 provider, aws provider) - doesn't fix this problem - and I tried various version. The format is correct, as I have tried leaving the name field blank (so it autogenerates), and I also used my own custom name (sign-up-rule). The results are the same.

### Relevant Error/Panic Output Snippet

```shell
│ Error: unexpected format for ID (arn:aws:events:eu-west-1:xxx:event-bus/aws.partner/auth0.com/xxx/auth0.logs/terraform-20221024154539697900000002), expected EVENTBUSNAME/RULENAME or RULENAME
│ 
│   with aws_cloudwatch_event_rule.successful_sign_up_rule,
│   on auth0-log-stream.tf line 95, in resource "aws_cloudwatch_event_rule" "successful_sign_up_rule":
│   95: resource "aws_cloudwatch_event_rule" "successful_sign_up_rule" {
```

### Terraform Configuration Files

```
resource "aws_cloudwatch_event_rule" "successful_sign_up_rule" {
  name           = "Auth0-Successful-Sign-up-Rule"
  description    = "Auth0 User Successfully signed up"
  event_bus_name = aws_cloudwatch_event_bus.event_bridge_event_bus.name
  event_pattern  = <<EOF
  {
    "detail-type": [
      "ss"
    ]
  }
EOF
}
```

### Steps to Reproduce

`terraform apply`

### Important Factoids

There is nothing special about my configuration.

vitorf7:
[...]

Issue #25510, opened 2022-06-21
Event bus name validation error for resource policy


subkanthi:
`Creating EventBridge policy failed: ValidationException: 1 validation error detected: Value 'aws.partner/auth0.com/development-c15bd8ab-17ce-4e74-b94b-9f8697a4c6d6/auth0.logs' at 'eventBusName' failed to satisfy constraint: Member must satisfy regular expression pattern: [\.\-_A-Za-z0-9]+
`
Event bus name for partner source are of the format aws.partner/<partner_name>, it looks the event bus validation is wrong for partner sources.

```
data "aws_iam_policy_document" "auth0_cloudwatch_policy" {
  statement {
    sid = "TrustEventsToStoreLogEvent"
    actions = [
      "logs:CreateLogStream",
      "logs:PutLogEvents"
    ]

    effect = "Allow"

    principals {
      identifiers = ["events.amazonaws.com", "delivery.logs.amazonaws.com"]
      type        = "Service"
    }

    resources = ["${aws_cloudwatch_log_group.auth0_event_logs_group.arn}:/aws/events/*:*"]

    condition {
      test     = "ArnEquals"
      variable = "aws:SourceArn"

      values   = [aws_cloudwatch_event_rule.auth0_cloudwatch_rule.arn]

    }
  }
}

# Create a Resource policy for Event Bus so that it can write logs to cloudwatch
resource "aws_cloudwatch_event_bus_policy" "auth0_eventbus_cloudwatch_policy" {
  depends_on = [aws_cloudwatch_event_bus.auth0]
  policy = data.aws_iam_policy_document.auth0_cloudwatch_policy.json
  event_bus_name = aws_cloudwatch_event_bus.auth0.name
}
```

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* aws_XXXXX

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
[...]

PR #30697, opened 2023-04-13
fix(events): fix arn based partner evernt bus id parse bug

### LINKED ISSUE #30293: [Bug]: Error creating "aws_cloudwatch_event_rule" while using Partner Event Sources - id format doesn't match
### Terraform Core Version

0.13.5, 0.14.0

### AWS Provider Version

4.0.0

### Affected Resource(s)

- aws_cloudwatch_event_rule

### Expected Behavior

Expected behaviour is that the EventBridge rule is created and all validations pass. While the rule does get created, it can't be used since a check fails due to an incorrect format. 

### Actual Behavior

Resource is created however it fails when the state is checked (terraform plan/apply).

```shell
Error: unexpected format for ID (arn:aws:events:eu-west-1:test-id:event-bus/aws.partner/salesforce.com/id1/id2/crm-sync-create-account-rule-terraform-test), expected EVENTBUSNAME/RULENAME or RULENAME
```

resource "aws_cloudwatch_event_rule" "event-bridge-rule" {
  name = "event-bridge-rule"
  description = "Testing creation of rules with Terraform"
  event_bus_name = data.aws_cloudwatch_event_bus.my-bus.name

  event_pattern = jsonencode({
    "source": [{
      "prefix": "aws.partner/salesforce.com"
    }],
    "detail-type": ["AccountChangeEvent"],
    "detail": {
      "payload": {
        "ChangeEventHeader": {
          "changeType": ["CREATE"]
        }
      }
    }
  })
}

### Steps to Reproduce

terraform plan / 
terraform apply

Both show the same error.

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

I already researched the cause of why this happened and I got to the "_aws/internal/service/events/id.go_" file which contains a function called "**RuleParseResourceID**". 
A split of the rule name is done based on "/" and in my case the code will enter the 3rd "if" statement (if len(parts) > 2 ).

The event bus name that I have is under the following format: "arn:aws:events:eu-west-1:test-id:event-bus/aws.partner/salesforce.com/id1/id2"

The "id" that gets passed to the function is under the following format: 
"arn:aws:events:eu-west-1:test-id:event-bus/aws.partner/salesforce.com/id1/id2/test-rule-name"

I think the problem consists in our partner event bus name containing multiple "/". This makes the check of the "eventBusName" against the "eventBusARNPattern" regexp to fail causing the error to appear. I don't have the option to change the event bus name, it's created automatically.

### References

I found a very smilar issue - same resource, same error, same file and function causing the error- however it seems in my case it's still something different caused by a mismatch with the regular expression.
https://github.com/hashicorp/terraform-provider-aws/issues/18431

No

vitorf7:
### Description

This PR aims at fixing a bug related to issue #30293 where it seems that partner event bus rules can also make use be in the format `arn:...` which is causing some people (myself included) to not be able to create or even delete resources once you come across this bug.

### Relations

Closes #30293.
Closes https://github.com/hashicorp/terraform-provider-aws/issues/27457.
Closes https://github.com/hashicorp/terraform-provider-aws/issues/25510.

### References

Please refer to my comment in the initial issue in order to see the configuration I was using when I encountered this bug https://github.com/hashicorp/terraform-provider-aws/issues/30293#issuecomment-1505476917

### Output from Testing

```
$ make test TEST=internal/service/events/id_test.go TESTARGS=-run=TestRuleParseResourceID
go test internal/service/events/id_test.go -run=TestRuleParseResourceID -timeout=5m
ok      command-line-arguments  0.342s
```


TakumiHaruta:
Thanks for this bugfix.

Is anyone able to merge this PR and release?
Actually this bug blocked the development which requires to use EventBridge with a partner provider like Auth0.
It's a bit critical for us...

vitorf7:
@TakumiHaruta Thank you so much. At my work we were also blocked because of this working with partner events as well (Genesys in my case).

Once it gets approved and merged in it should hopefully help you get everything sorted. If you are however in a pinch and wanna solve your problem sooner, then you could clone my branch locally and follow the guidelines on how to do local development from here https://hashicorp.github.io/terraform-provider-aws/development-environment/

I was personally facing some trouble when trying to use it since at work we had the terraform plan and terraform apply run from the docker image provided by hashicorp (and one we have at work) as my own machine is an Mac M1 which means the make build command from the makefile (in the terraform-provider-aws repo) would build an arch64 instead of a linux amd64.

After running env GOOS=linux GOARCH=amd64 go build ., then moving the built binary to the $GOPATH/bin and making sure that both the $GOPATH/bin and my ~/.terraformrc were both mounted to the correct places in the docker image, it all worked fine.

I was able to finally finish the work I was doing and it also meant that I not longer go the error you and I faced.

TakumiHaruta:
@vitorf7 
Thank you for your guidance. That'd be really helpful.
We're actually hosting a private registry for aws provider on Terraform Cloud using goreleaser, but this PR seems not active since it was opened 3 months ago. I think it could be categorized as a hotfix.

@ewbankkit Hi, could you review and merge this PR when you have time? This is a bit critical for developers who use multiple providers.

sean-redmond:
Be good to see this merged in, looks ready and will solve an issue I am running into.

tapadarj:
Please confirm when it will be merged. I am facing the same issue.

jar-b:
The `golangci-lint` action is failing due to network timeouts, not linter findings. Linting checks pass locally.

```console
% make lint PKG=events
==> Checking source code with golangci-lint...
==> Checking source code with providerlint...
==> Checking source code with importlint...
```

Will proceed with merging despite the failing check.

jar-b:
Thanks for your contribution, @vitorf7 ! 👏

vitorf7:
@jar-b no problem at all. Glad I was able to help.

PR #30817, opened 2023-04-19
feat: Add insecure_value to datasource aws_ssm_parameter

### LINKED ISSUE #26915: [Bug]: data.aws_ssm_parameter - Non SecureString SSM parameters marked as sensitive
### Terraform Core Version

0.15.5

### AWS Provider Version

4.31.0

### Affected Resource(s)

data.aws_ssm_parameter

### Expected Behavior

Non SecureString Values should not be flagged sensitive

### Actual Behavior

All values are being flagged sensitive

```
data "aws_ssm_parameter" "logging_name" {
  name = "/logging_name"
}

name = data.aws_ssm_parameter.logging_name.value
```

### Steps to Reproduce

Create `/logging_name` SSM parameter of type `String`.

1. terraform init
2. terraform plan

### Debug Output

```
logging {
          # Warning: this attribute value will be marked as sensitive and will not
          # display in UI output after applying this change. The value is unchanged.
          ~ name = (sensitive)
            # (1 unchanged attribute hidden)
        }
```

### References

Based on issue https://github.com/hashicorp/terraform-provider-aws/issues/9090 the fix was added for insecure_value in provider version 4.22.0 but `data.aws_ssm_parameter` was left out.

_No response_

Jno21:
### Description

Add `insecure_value` to datasource `aws_ssm_parameter`. The idea is to have the possibility to expose the value as insecure like in the resource `aws_ssm_parameter`.

### Relations

Closes #26915

Closes #0000

```
$ make testacc TESTS=TestAccSSMParameterDataSource_basic PKG=ssm
TF_ACC=1 go test ./internal/service/ssm/... -v -count 1 -parallel 20 -run='TestAccSSMParameterDataSource_basic'  -timeout 180m
=== RUN   TestAccSSMParameterDataSource_basic
=== PAUSE TestAccSSMParameterDataSource_basic
=== CONT  TestAccSSMParameterDataSource_basic
--- PASS: TestAccSSMParameterDataSource_basic (107.81s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ssm        122.663s
```

```
$ make testacc TESTS=TestAccSSMParameterDataSource_fullPath PKG=ssm
TF_ACC=1 go test ./internal/service/ssm/... -v -count 1 -parallel 20 -run='TestAccSSMParameterDataSource_fullPath'  -timeout 180m
=== RUN   TestAccSSMParameterDataSource_fullPath
=== PAUSE TestAccSSMParameterDataSource_fullPath
=== CONT  TestAccSSMParameterDataSource_fullPath
--- PASS: TestAccSSMParameterDataSource_fullPath (120.89s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ssm        137.380s
```

### DIFF
--- a/internal/service/ssm/parameter_data_source.go
+++ b/internal/service/ssm/parameter_data_source.go
@@ -21,6 +21,10 @@ func DataSourceParameter() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"insecure_value": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
@@ -69,6 +73,7 @@ func dataParameterRead(ctx context.Context, d *schema.ResourceData, meta interfa
 
 	d.SetId(aws.StringValue(param.Name))
 	d.Set("arn", param.ARN)
+	d.Set("insecure_value", param.Value)
 	d.Set("name", param.Name)
 	d.Set("type", param.Type)
 	d.Set("value", param.Value)
--- a/internal/service/ssm/parameter_data_source_test.go
+++ b/internal/service/ssm/parameter_data_source_test.go
@@ -24,6 +24,7 @@ func TestAccSSMParameterDataSource_basic(t *testing.T) {
 				Config: testAccParameterDataSourceConfig_basic(name, "false"),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttrPair(resourceName, "arn", "aws_ssm_parameter.test", "arn"),
+					resource.TestCheckResourceAttr(resourceName, "insecure_value", "TestValue"),
 					resource.TestCheckResourceAttr(resourceName, "name", name),
 					resource.TestCheckResourceAttr(resourceName, "type", "String"),
 					resource.TestCheckResourceAttr(resourceName, "value", "TestValue"),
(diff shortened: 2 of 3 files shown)


agupta94:
@Jno21 ThankYou for this PR! 
Would it be easier to add `nonsensitive = true` instead on the data object itself? That would prevent us from switching all reference of `value` to `insecure_value`

Jno21:
@agupta94 Thank you for your comment.
> Would it be easier to add `nonsensitive = true` that would flag `value` to be nonsensitive? That would prevent us from switching all references of `value` to `insecure_value`, otherwise this is no different than simply doing `nonsensitive(data.aws_ssm_parameter.xyz.value)` currently.


That would be an idea, my first though was to have the `value` exposed as `sensitive` or not depending on the `SecureString` or `String` in ssm parameter. Unfortunately the way I understand the Terraform SDK it is not possible to change dynamically the value of a sensitive value since it is set in the Schema before any interaction, if I am wrong on this I will be happy to know how to do it.
I also think that the vision of the aws provider is to avoid implicit behaviors, having a behavior that change the way a value work depending on something else is quite risky and can induce error. 

Globally I follow what was done here: https://github.com/hashicorp/terraform-provider-aws/pull/25721 to have the same behavior on the `resource` and `data`.

To answer the last part, yes indeed doing `nonsensitive` would be the same, but that would also mean using locals if you want to use that value multiples time, I think the insecure_value make it easier code wise.

If there is a better solution possible, that I didn't think of and it respect the way this provider wants to go, I would be totally down to work on it :)

internal/service/ssm/parameter_data_source.go
```diff
@@ -73,10 +73,15 @@ func dataParameterRead(ctx context.Context, d *schema.ResourceData, meta interfa
 
 	d.SetId(aws.StringValue(param.Name))
 	d.Set("arn", param.ARN)
-	d.Set("insecure_value", param.Value)
+	if aws.StringValue(param.Type) == ssm.ParameterTypeSecureString {
```
Jno21, line 79: **issue:** I like the idea ! However the condition should be inverted. Since we want to have it in `insecure_value` when it is **not** a  ParameterTypeSecureString.
```
if aws.StringValue(param.Type) != ssm.ParameterTypeSecureString
```

nam054: Hi Yes! That was my mistake - I've changed it now!

PR #30822, opened 2023-04-19
Workflow: Partner project integration

breathingdust:
### Description

- When an PR is raised and contributor is a partner, apply the `partner` label
- When `partner` label is applied to an PR, move it into the partner board.


ewbankkit (approved):
LGTM :rocket:.

PR #30839, opened 2023-04-20
[WIP] ECS Service Primary Task Set

joanayma:
### Description

Adds the Primary Task Set of the ECS Service.

I added option `primary` to `aws_ecs_task_set` that when set to `true` will call API [`UpdateServicePrimaryTaskSet`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateServicePrimaryTaskSet.html) to set it.

There's no API call to unset/remove the Primary Task Set - let me know if there's a way. Because of this, when a Task Set has `primary=true`, the destroy is not allowed by AWS SDK and `force_delete=true` is required. In case `force_delete` is `false` or unset, the Error is clear enough. I think this is what's expected to do from the developer.

If multiple `aws_ecs_task_set` for the same `aws_ecs_service` has `primary=true` it will drift until fixed as there can be only one Primary Task Set. I think it's up to developer to write consistent Terraform code.

I fixed a wrong assumption in `internal/service/ecs/service.go:684` that says the AWS SDK doesn't set `task_definition` when the Deployment Type is `EXTERNAL` - it is setting it and it can't be updated, leading to an error on next refresh/apply. First I was creating the test with a `ignore_changes` (see first commit) but I think this is more clear.

I considered other implementations but I considered this one is better because:
- Adding this on `aws_ecs_service` might make more sense but then it's a dependency cycle.
- Adding it as a separated resource (i.e. `aws_ecs_service_primary_task_set`) makes it unclear which `aws_ecs_task_set` is the primary and why one Task Set can't be destroyed.

Let me know if this can be merged or requires any fix or implemented another way.

### Relations

There's no open issue, thus I opened this Pull Request to add the contribution.

### References

From: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-type-external.html

> The UpdateServicePrimaryTaskSet API action modifies which task set in a service is the primary task set. When you call the DescribeServices API action, it returns all fields specified for a primary task set. If the primary task set for a service is updated, any task set parameter values that exist on the new primary task set that differ from the old primary task set in a service are updated to the new value when a new primary task set is defined. If no primary task set is defined for a service, when describing the service, the task set fields are null.

Added Acceptance Test `TestAccECSTaskSet_withPrimaryTaskSet`:

```
$ TF_ACC_LOG=DEBUG TF_ACC=1 go test ./internal/service/ecs/... -v -count 1 -parallel 20 -run='TestAccECSTaskSet_withPrimaryTaskSet'  -timeout 180m
=== RUN   TestAccECSTaskSet_withPrimaryTaskSet
=== PAUSE TestAccECSTaskSet_withPrimaryTaskSet
=== CONT  TestAccECSTaskSet_withPrimaryTaskSet
--- PASS: TestAccECSTaskSet_withPrimaryTaskSet (68.69s)
PASS
```

Tested `TestAccECSService_basic` and `TestAccECSService_DeploymentControllerType_external` to make sure the above mentioned fix in `internal/service/ecs/service.go:684` is valid:
```
$ TF_ACC_LOG=DEBUG TF_ACC=1 go test ./internal/service/ecs/... -v -count 1 -parallel 20 -run='TestAccECSService_basic'  -timeout 180m
=== RUN   TestAccECSService_basic
=== PAUSE TestAccECSService_basic
=== RUN   TestAccECSService_basicImport
=== PAUSE TestAccECSService_basicImport
=== CONT  TestAccECSService_basic
=== CONT  TestAccECSService_basicImport
--- PASS: TestAccECSService_basic (118.32s)
--- PASS: TestAccECSService_basicImport (212.22s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ecs	212.457s

$ TF_ACC_LOG=DEBUG TF_ACC=1 go test ./internal/service/ecs/... -v -count 1 -parallel 20 -run='TestAccECSService_DeploymentControllerType_external'  -timeout 180m
zsh: correct './internal/service/ecs/...' to './internal/service/ecs/..' [nyae]? n
=== RUN   TestAccECSService_DeploymentControllerType_external
=== PAUSE TestAccECSService_DeploymentControllerType_external
=== CONT  TestAccECSService_DeploymentControllerType_external
--- PASS: TestAccECSService_DeploymentControllerType_external (63.87s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ecs	64.050s
```


joanayma:
Hi @justinretzolk how can I get this merged? Thanks in advance

PR #30866, opened 2023-04-21
Minor `skaff` self-registration annotation updates

ewbankkit:
### Description

Adds `name="<human friendly name>"` to the resource/data source self registration annotations generated by `skaff`.
Also adds an `IsDataSource` flag to the `InContext` structure than can be used for various purposes.


johnsonaj (approved):
LGTM 🚀

PR #31144, opened 2023-05-03
Update permissive ECR policy docs to better demonstrate best practices

james-warren0:
### Description

The documentation for ECR Repository policies is permissive. This is a minor update to documentation that better demonstrates best practices of explicitly setting the type and identifier in the policy.

### References

Policy example in AWS docs can be found [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/repository-policy-examples.html) and [here](https://docs.aws.amazon.com/AmazonECR/latest/public/public-repository-policies.html).


breathingdust:
LGTM 🚀 Thanks for the contribution @james-warren0!

PR #31151, opened 2023-05-04
enhance/aws_ssm_parameter: Validate value max size

### LINKED ISSUE #31133: [Enhancement]: aws_ssm_parameter should warn on max param size
### Description

I was using the `aws_ssm_parameter` resource.

A `terraform plan` succeeded, but a `terraform apply` failed, because the parameter value was > 4096 characters (and I didn't set another tier).

These limits can be checked statically. It would be nice if a plan would already showcase an error, so this doesn't only break during deployment.

* `aws_ssm_parameter`

_No response_

### References

_No response_

None

ldcorentin:
Hello, I tried

### Relations
Closes #31133


internal/service/ssm/parameter.go
```diff
@@ -112,6 +112,7 @@ func ResourceParameter() *schema.Resource {
 				Sensitive:    true,
 				Computed:     true,
 				ExactlyOneOf: []string{"insecure_value", "value"},
+				ValidateFunc: validation.StringLenBetween(0, 4096),
```
erz4, line 115: I think you missing the diffrentiation between tiers:
Standard max is 4k
Advanced/Intelligent-Tiering is 8k

you can see it here: https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html

PR #31248, opened 2023-05-07
data_source: ap-southeast-3 (Jakarta) HostedZoneID

PatMyron:
https://docs.aws.amazon.com/general/latest/gr/elasticbeanstalk.html#elasticbeanstalk_region

https://github.com/hashicorp/terraform-provider-aws/blob/main/docs/add-a-new-region.md#update-region-specific-values-in-static-data-sources


ewbankkit (approved):
LGTM :rocket:.

PR #31276, opened 2023-05-08
Use default retry count from AWS SDK

gdavison:
### Description

The provider was previously setting the default retry count for the AWS SDK to 25. This would  cause the SDK to retry returnable errors for an extremely long time internally without returning control to the provider, which would look like a hang to users.

### DIFF
--- a/.changelog/31276.txt
+++ b/.changelog/31276.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+provider: Now uses the default retry count for AWS SDKs
+```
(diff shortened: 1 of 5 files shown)


bateller (approved):

ewbankkit:
Add a CHANGELOG entry and maybe something to https://github.com/hashicorp/terraform-provider-aws/blob/prerelease/5.x/website/docs/guides/version-5-upgrade.html.md?

jar-b (approved):
LGTM 🎉

PR #31331, opened 2023-05-11
fix: use correct format for ECS parameters tags in scheduler

spiarh:
### Description

The previous tags format was based on RunTask request syntax [1] which consisted in an array of map with key and an optional value but the schedule format is an array of string to string maps [2].

I did tested that this works with the fix.

[1] https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html
[2] https://docs.aws.amazon.com/scheduler/latest/APIReference/API_EcsParameters.html


qchuchu:
@spiarh I guess you didn't have any follow up on this since then ?

spiarh:
No I haven't unfortunately and I'm also not using this resource anymore.

PR #31455, opened 2023-05-17
Doc : ACM certificate validation method optional

### LINKED ISSUE #31425: r/aws_acm_certificate failing with `expected validation_method to be one of [EMAIL DNS], got NONE` for imported certificate
### Terraform Core Version

1.4.4

### AWS Provider Version

4.67.0

### Affected Resource(s)

aws_acm_certificate

### Expected Behavior

The `validation_method = "NONE"` should be a valid value when certificate is imported first into ACM and then this resource is imported to Terraform.

### Actual Behavior

When using `aws_acm_certificate` for imported certificate, `validation_method` can be set to NONE. But this fails with error

```
Error: expected validation_method to be one of [EMAIL DNS], got NONE
```

_No response_

```terraform
resource "aws_acm_certificate" "cert" {
  domain_name       = "testing.custom-example.com"
  validation_method = "NONE"
}
```

### Steps to Reproduce

* Create a certificate and Import it into ACM manually (outsider of Terraform)
* Now import the ACM resource into Terraform using `terraform import` command.
* Run `terraform plan` to check diff with above resource.

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

_No response_

### References

_No response_

Yes

sousmangoosta:
### Description

The documentation state that `validation_method` is required in `aws_acm_certificate` resource.
The only method currently validated by the API are `EMAIL` and `DNS`, so people can have issue following the doc recommendation.
As the `validation_method` is currently optional, remove it from the resource fix the issue.

### Relations

Closes #31425

### References

[My answer and test done here](https://github.com/hashicorp/terraform-provider-aws/issues/31425#issuecomment-1551960350)

### DIFF
--- a/website/docs/r/acm_certificate.html.markdown
+++ b/website/docs/r/acm_certificate.html.markdown
@@ -143,7 +143,7 @@ The following arguments are supported:
 * Creating an Amazon issued certificate
     * `domain_name` - (Required) Domain name for which the certificate should be issued
     * `subject_alternative_names` - (Optional) Set of domains that should be SANs in the issued certificate. To remove all elements of a previously configured list, set this value equal to an empty list (`[]`) or use the [`terraform taint` command](https://www.terraform.io/docs/commands/taint.html) to trigger recreation.
-    * `validation_method` - (Required) Which method to use for validation. `DNS` or `EMAIL` are valid, `NONE` can be used for certificates that were imported into ACM and then into Terraform.
+    * `validation_method` - (Optional) Which method to use for validation. `DNS` or `EMAIL` are valid. This parameter must not be set for certificates that were imported into ACM and then into Terraform.
     * `key_algorithm` - (Optional) Specifies the algorithm of the public and private key pair that your Amazon issued certificate uses to encrypt data. See [ACM Certificate characteristics](https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate.html#algorithms) for more details.
     * `options` - (Optional) Configuration block used to set certificate options. Detailed below.
     * `validation_option` - (Optional) Configuration block used to specify information about the initial validation of each domain name. Detailed below.


website/docs/r/acm_certificate.html.markdown
```diff
@@ -143,7 +143,7 @@ The following arguments are supported:
 * Creating an Amazon issued certificate
     * `domain_name` - (Required) Domain name for which the certificate should be issued
     * `subject_alternative_names` - (Optional) Set of domains that should be SANs in the issued certificate. To remove all elements of a previously configured list, set this value equal to an empty list (`[]`) or use the [`terraform taint` command](https://www.terraform.io/docs/commands/taint.html) to trigger recreation.
-    * `validation_method` - (Required) Which method to use for validation. `DNS` or `EMAIL` are valid, `NONE` can be used for certificates that were imported into ACM and then into Terraform.
+    * `validation_method` - (Required for Amazon-issued certificate) Which method to use for validation. `DNS` or `EMAIL` are valid. This parameter must not be set for certificates that were imported into ACM and then into Terraform.
```
justinretzolk, line 146: I think this change would retain the necessary information while keeping with the standard format of just having `(Required)` or `(Optional)`.

If it feels more clear to mention the Amazon-issued bit, perhaps we could add it into the description itself. 

```suggestion
    * `validation_method` - (Optional) Which method to use for validation. `DNS` or `EMAIL` are valid. This parameter must not be set for certificates that were imported into ACM and then into Terraform.
```

justinretzolk (changes requested):
Thank you for taking the time to raise this! I left one comment that I think we should consider prior to merging.

website/docs/r/acm_certificate.html.markdown
sousmangoosta: > I think this change would retain the necessary information while keeping with the standard format of just having `(Required)` or `(Optional)`.

I got the same feeling, I modified that, thanks 🙏

justinretzolk (approved):
Looks good to me! Thank you again for taking the time to raise this! We'll get this merged in as soon as we're able 🚀 

Note to any future viewers:
- [Schema for `validation_method`](https://github.com/hashicorp/terraform-provider-aws/blob/a13a04f72ec61be611145b079c1017c1d4f0d119/internal/service/acm/certificate.go#L224-L231) indicating this is an optional argument.
- [`acm.ImportCertificateInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/acm#ImportCertificateInput) documentation indicating that `ValidationMethod` is not a field for this type, while it is for [`acm.RequestCertificateInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/acm#RequestCertificateInput) (the other type used, when an ACM certificate is requested rather than imported).

PR #31596, opened 2023-05-26
docs: add CDKTF documentation for `ec2_*` resources/datasource and update script

DanielMSchmidt:
### Description

This PR introduces CDKTF documentation generated from the original HCL-based documentation. The internal tool doing the conversion replaces the code blocks with the CDKTF equivalent of the HCL. If the HCL is invalid we can not translate it, therefore we keep it as is. We also add a comment with a cache key to each generated document to avoid regenerating them if neither the input nor the used cdktf version changed.

For now we only add documentation for Typescript and Python and only for resources or data sources that start with `ec2_`.
Once this is merged more PRs for other resources will follow.

The included GitHub action will run every Wednesday night to create a PR with any documentation updates.


ewbankkit:
The `tfproviderdocs` error requires https://github.com/bflad/tfproviderdocs/pull/50 to fix.

bflad:
tfproviderdocs v0.10.0 introduces some initial CDKTF support. Please reach out with any issues.

website/docs/cdktf/python/d/ec2_client_vpn_endpoint.html.markdown
```diff
@@ -0,0 +1,96 @@
+---
```
breathingdust, line 1: Not a blocker, but previously we had discussed adding a comment header to each generated file stating that the file was generated and not to be edited. It would hopefully minimize contributors submitting PR's with direct edits.

DanielMSchmidt: Yeah good point, I can include this in the next version 👍

.github/workflows/cdktf-documentation.yml
```diff
@@ -0,0 +1,73 @@
...
+          cdktf-registry-docs convert \
+            --files='*/ec2_*.html.markdown' \
+            --languages='typescript,python' \
+            --parallel-conversions-per-document=1 \
+            --provider-from-registry="hashicorp/aws@~> ${{ needs.releasedVersion.outputs.version }}" \
+            .
+        env:
+          TF_PLUGIN_CACHE_DIR: ${{ steps.global-cache-dir-path.outputs.dir }}/terraform-plugins
+
+      - name: Create Pull Request
```
breathingdust, line 68: Would you want the AWS provider team to wait from an approval from yourselves before we merge (Assuming checks pass)?

DanielMSchmidt: No, I don't think we are going to take a look at the PRs on a regular basis. We get informed about problems through sentry by the conversion tool itself :)

.github/workflows/cdktf-documentation.yml
```diff
@@ -0,0 +1,73 @@
...
+    needs:
+      - releasedVersion
+    runs-on: ubuntu-latest
+    container:
+      image: docker.mirror.hashicorp.services/hashicorp/jsii-terraform
+    env:
+      CHECKPOINT_DISABLE: "1"
+    timeout-minutes: 120
+    steps:
+      - uses: actions/checkout@v3.1.0
```
xiehan, line 40: We should be using the SHA hashes instead of versions here as per the new security  policy that's being rolled out -- let's not create more work for other teams.

xiehan (changes requested):

breathingdust (approved):
LGTM @DanielMSchmidt  🚀 Let's see how the job generation goes this evening!

PR #31741, opened 2023-06-02
d/docs: add the missing id attribute to vpc

amazingandyyy:
### Description

the `aws_vpc` did have id attribute

### Reference
update https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc#arn


justinretzolk:
Hey @amazingandyyy 👋 Thank you for taking the time to raise this! It looks like `id` is listed in the arguments section, which is why it was not listed in the attributes section. This is mentioned at the start of the attributes section:

> All of the argument attributes except `filter` blocks are also exported as result attributes. This data source will complete the data by populating any fields that are not included in the configuration with the data for the selected VPC.

I'll wait to hear back from you before closing this one out, to make sure I haven't overlooked anything.

amazingandyyy:
got it! that makes sense!

Issue #27432, opened 2022-10-24
r/aws_chime_voice_connector: Adding tagging support


ewbankkit:
Add tagging support to the `aws_chime_voice_connector` resource.

```hcl
resource "aws_chime_voice_connector" "example" {
  name = "example"

  tags = {
    Environment = "test"
  }
}
```

PR #31742, opened 2023-06-02
[TechDebt]: provider/tags: allow generator to override aws service package

jar-b:
### Description
Adds an optional `-ServicePackage` argument to the tags generator, allowing the AWS Go SDK package to be overridden. Specifically, this will support tagging in the `chime` service package by allowing the tagging functions to use the `chimesdkvoice` AWS SDK package which has proper tagging support unavailable in the legacy `chime` namespace.

This PR will enable supporting `tags` for existing `chime` resources once merged.

### Relations
Relates #27432

### References
- https://docs.aws.amazon.com/chime/latest/APIReference/API_TagResource.html
  - "Applies the specified tags to the specified Amazon Chime SDK meeting resource." (Note only meeting resources are supported).
- https://docs.aws.amazon.com/chime-sdk/latest/APIReference/API_identity-chime_TagResource.html


internal/generate/tags/README.md
```diff
@@ -24,6 +24,7 @@ Some flags control generation a certain section of code, such as whether the gen
 | --- | --- | --- | --- |
 | `GetTag` |  | Whether to generate GetTag | `-GetTag` |
 | `ListTags` |  | Whether to generate ListTags | `-ListTags` |
+| `ServicePackage` |  | AWS Go SDK package to use. Defaults to the provider service package name. | `-ServicePackage=chimesdkvoice` |
```
YakDriver, line 27:
```suggestion
| `TagServicePackage` |  | AWS Go SDK package to use. Defaults to the provider service package name. | `-TagServicePackage=chimesdkvoice` |
```

Maybe? Seems like this will go away eventually and might be easier to extricate with this name.

internal/generate/tags/main.go
```diff
@@ -123,6 +124,7 @@ type TemplateData struct {
 	AWSService             string
 	AWSServiceIfacePackage string
 	ClientType             string
+	InternalServicePackage string
```
YakDriver, line 127:
```suggestion
	TagServicePackage string
```

Similar to above

YakDriver (approved):
Couple notes

internal/generate/tags/README.md
ewbankkit: Or `AWSSDKServicePackage` to match `AWSSDKVersion`?

jar-b: Updated to `AWSSDKServicePackage` and changed the logic around so the `InternalServicePackage` field is no longer necessary. If we opt to back this out later it should only requiring removing the flag and reverting a couple function calls to use the `servicePackage` variable again.

PR #31751, opened 2023-06-03
Updates `skaff` to collect `diags`

gdavison:
### Description

Updates `skaff` templates so that `diags` are collected rather than immediately returning. This ensures that if a Warning Diagnostic is added at some point, it will not get lost accidentally.

Also removes unneeded registration reminder and fixes spacing for data source imports


jar-b (approved):
LGTM 🎸

PR #31753, opened 2023-06-03
fix format for identity pool id

lpmi-13:
### Description
the format currently shown is for a user pool id, and it should be for an identity pool id

### References
it looks like the existing ID for this documentation was copied from the page for user pool.
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cognito_user_pool

The format of the ID for identity pools is very difficult to track down anywhere, but the cloudformation docs seem to show it correctly
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-identitypool.html

### DIFF
--- a/website/docs/r/cognito_identity_pool.markdown
+++ b/website/docs/r/cognito_identity_pool.markdown
@@ -70,7 +70,7 @@ backend and the Cognito service to communicate about the developer provider.
 
 In addition to all arguments above, the following attributes are exported:
 
-* `id` - An identity pool ID, e.g. `us-west-2_abc123`.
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks good to me! Thank you very much for taking the time to raise this! I'll get this merged once the checks have finished up 🚀 

[API documentation reference](https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/API_CreateIdentityPool.html#API_CreateIdentityPool_ResponseSyntax)

Issue #31777, opened 2023-06-05
[Docs]: 


mateusprado:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_target_group

### Description

The doc about stickiness is incorrect, because the actual behavior of the AWS provider is to default stickiness to false, not true. See the reference below.

### References

https://docs.aws.amazon.com/cli/latest/reference/elbv2/modify-target-group-attributes.html

### Would you like to implement a fix?

Yes

PR #31779, opened 2023-06-05
adding a fix to stickiness default value

mateusprado:
### Description

The doc about stickiness is incorrect, because the actual behavior of the AWS provider is to default stickiness to false, not true. See the reference below.

### Relations

[Relates #31777 ](https://github.com/hashicorp/terraform-provider-aws/issues/31777)


justinretzolk:
Hey @mateusprado 👋 Thank you for taking the time to raise this! Based on the [resource schema](https://github.com/hashicorp/terraform-provider-aws/blob/661b2736bbcc601ae856e4d8e7f4869915c8f9d9/internal/service/elbv2/target_group.go#L272) the default appears to be `true`. Are you seeing differing behavior when using the provider?

justinretzolk:
Hey @mateusprado 👋 It does look like the default on the AWS side is `false`, however, the AWS Provider currently [defaults to `true`](https://github.com/hashicorp/terraform-provider-aws/blob/1c429dc9ab71947a63a3def747cceb16b95ae33f/internal/service/elbv2/target_group.go#L275) based on the schema. There's absolutely an argument to be made that the default on the AWS Provider side should be updated to match the default behavior on the AWS side, however, my question was more of whether you're seeing this value default to `true` or `false` when not specified in your Terraform configuration, as that'll give us a better idea of what path should be taken to remediate the issue.

mr-karan:
Although `stickiness` is optional, it doesn't let you create forward block without it. Even then, when `enabled=false`, it still asks to enter duration and 0 isn't a valid duration. Hence have to resort to this hack.

```
      stickiness {
        enabled  = false
        duration = 1
      }
```

PR #31792, opened 2023-06-06
Update cloudwatch_event_target.html.markdown

up-town-down:
### Description

Adding "delivery.logs.amazonaws.com" as a service principal, as this now appears necessary to get a log forwarding target to work (As a result of many hours of experimentation!). Also example here: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-use-resource-based.html#eb-cloudwatchlogs-permissions


breathingdust:
LGTM @up-town-down 🚀 Thanks for the contribution!

PR #31793, opened 2023-06-06
[Compliance]: Removal of `alex-page/github-project-automation-plus`

justinretzolk:
### Description

This PR removes the `alex-page/github-project-automation-plus` action in favor of directly calling the GitHub GraphQL API, similar to the workflows added in #30822. I also took this opportunity to combine the `roadmap_milestone.yml` and `project-roadmap-label.yml` workflow files, since they did very similar things.

### Relations

Relates #30822
Relates #31014
Relates hashicorp/security-tsccr#454

### References

A follow up PR will be needed to change the column IDs to the `next_global_id` rather than the `legacy_global_id` seen here (see #31846). Unfortunately, this value only seems to be retrievable by actually calling the GraphQL API to add an item, so I'll modify the value after we get this merged in and the workflows run (it will still work, just throws a warn message).

N/a, docs


ewbankkit (approved):
LGTM :rocket:.

PR #31856, opened 2023-06-08
Ensure `issue_comment` triggers are correctly evaluated

justinretzolk:
### Description

When creating this reusable workflow, I assumed that checking the `event_type` to see if the event was related to an issue would work well enough for distinguishing between issues and PRs. I failed to account for `issue_comment` triggers, which [are triggered](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment) for both issues _and_ pull requests with an event type of `issue_comment`, bypassing the previous test. This led to failures, as seen in the references section.

### Relations

Relates #31750

### References

- [Failed run due to not accounting for this](https://github.com/hashicorp/terraform-provider-aws/actions/runs/5215614719/jobs/9413376735)
- [GitHub recommendation on a similar approach](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only) (this is only discussing `issue_comment` triggers; the `github.event.pull_request` bit checks other non-comment related pull requests.

N/a, Actions

### DIFF
--- a/.github/workflows/reusable-add-or-remove-labels.yml
+++ b/.github/workflows/reusable-add-or-remove-labels.yml
@@ -21,12 +21,15 @@ jobs:
     name: 'Modify Label(s)'
     runs-on: ubuntu-latest
     env:
-      ITEM_TYPE: 'pr'
+      ITEM_TYPE: 'issue'
     steps:
       - name: 'Determine if the Item is an Issue or Pull Request'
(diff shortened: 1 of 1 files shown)


johnsonaj (approved):
LGTM 🚀

PR #31880, opened 2023-06-09
Add reusable workflow for automating new Projects

justinretzolk:
### Description

This PR introduces a reusable workflow that will enable us to automate [GitHub's new style of Projects](https://docs.github.com/en/issues/planning-and-tracking-with-projects/learning-about-projects/about-projects). This workflow is specific to the team's Project, and so has the project ID and Status field's ID statically defined.

Initially, this workflow was more generic, so that we could reuse the same workflow to update the team's Project as well as any additional Projects we decide to add down the line (for example, the Partner board). I've since modified it to be scoped to this the team's Project so to make usability a bit easier.

If we add additional fields at a later date, an additional step may be added to update that field's value as well.

### Relations

Relates #31881
Before #31883

### References

[Automating Projects using Actions](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions)

N/a, Actions


breathingdust (approved):
LGTM 🚀 Let's try it out!

PR #31905, opened 2023-06-12
Move partner labeler to a distinct job

justinretzolk:
### Description

Reusable workflows can only be used at the job level, not in steps within a job. This PR moves the partner labeler to a distinct job so that it can properly use the reusable flow to add the label.

### References

- [Failed step from a previous job](https://github.com/hashicorp/terraform-provider-aws/actions/runs/5230914202/jobs/9444743245)
- [GitHub Actions documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations):

> **Reusable workflows are called directly within a job, and not from within a job step.** You cannot, therefore, use GITHUB_ENV to pass values to job steps in the caller workflow.

N/a, Actions

### DIFF
--- a/.github/workflows/pull_request_feed.yml
+++ b/.github/workflows/pull_request_feed.yml
@@ -71,8 +71,3 @@ jobs:
                 }
               ]
             }
-      - name: Apply Partner Label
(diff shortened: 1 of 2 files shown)


ewbankkit (approved):
LGTM :rocket:.

PR #31984, opened 2023-06-14
update example with up to date image

lpmi-13:
### Description
the current documentation shows an image that doesn't exist anymore (standard:1.0), and so we should use the most up-to-date image for Amazon Linux2.

### References

https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html


justinretzolk (approved):
Looks good to me! Thank you very much for taking the time to raise this! I'll get this merged once the checks finish up 🚀

PR #32012, opened 2023-06-15
Update cognito_user_pool.markdown

rkaufman13:
Hi Hashicorp!
I spent a lot of yesterday struggling with updating my Cognito userpool email_configuration settings, and I believe that this documentation change will save others the struggle I went through. (In fairness, AWS's documentation on this point is also not great.)

### Description
Adds one line to the `email_provider` documentation block.

### DIFF
--- a/website/docs/r/cognito_user_pool.markdown
+++ b/website/docs/r/cognito_user_pool.markdown
@@ -118,7 +118,7 @@ The following arguments are optional:
 ### email_configuration
 
 * `configuration_set` - (Optional) Email configuration set name from SES.
-* `email_sending_account` - (Optional) Email delivery method to use. `COGNITO_DEFAULT` for the default email functionality built into Cognito or `DEVELOPER` to use your Amazon SES configuration.
+* `email_sending_account` - (Optional) Email delivery method to use. `COGNITO_DEFAULT` for the default email functionality built into Cognito or `DEVELOPER` to use your Amazon SES configuration. Required to be `DEVELOPER` if `from_email_address` is set.
 * `from_email_address` - (Optional) Sender’s email address or sender’s display name with their email address (e.g., `john@example.com`, `John Smith <john@example.com>` or `\"John Smith Ph.D.\" <john@example.com>`). Escaped double quotes are required around display names that contain certain characters as specified in [RFC 5322](https://tools.ietf.org/html/rfc5322).
 * `reply_to_email_address` - (Optional) REPLY-TO email address.
 * `source_arn` - (Optional) ARN of the SES verified email identity to use. Required if `email_sending_account` is set to `DEVELOPER`.


justinretzolk (approved):
Hey @rkaufman13 👋 Thank you for taking the time to raise this! In reviewing this, I took a look around, and found the AWS documentation around this to be a bit confusing, as you'd mentioned. As such, I wanted to provide a bit more context for anyone who may stumble across this in the future.

Looking at the AWS documentation on [Email settings for Amazon Cognito user pools](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-email.html#user-pool-email-default), I noticed that under the **Default email functionality** header, it states:

> When you use the default email configuration, you can use either of the following email addresses as the FROM address:
>
> The default email address, no-reply@verificationemail.com.
>
> A custom email address. Before you can use your own email address, you must verify it with Amazon SES and grant Amazon Cognito permission to use this address.

From there, I navigated to the API documentation, specifically for the [`EmailConfigurationType` data type](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_EmailConfigurationType.html). Under the **EmailSendingAccount** argument (which correlates to the `email_sending_account` argument in the `aws_cognito_user_pool` resource), it states:

> `COGNITO_DEFAULT`
> When Amazon Cognito emails your users, it uses its built-in email functionality. When you use the default option, Amazon Cognito allows only a limited number of emails each day for your user pool. For typical production environments, the default email limit is less than the required delivery volume. To achieve a higher delivery volume, specify `DEVELOPER` to use your Amazon SES email configuration.
> 
> To look up the email delivery limit for the default option, see [Limits](https://docs.aws.amazon.com/cognito/latest/developerguide/limits.html) in the Amazon Cognito Developer Guide.
> 
> The default `FROM` address is `no-reply@verificationemail.com`. To customize the `FROM` address, provide the Amazon Resource Name (ARN) of an Amazon SES verified email address for the `SourceArn` parameter.

Initially, this information led me to believe that specifying the `email_sending_account` as `COGNITO_DEFAULT` was possible when setting the `from_email_address`, so long as the `from_email_address` was an SES verified email address (and the `source_arn` argument was set to the ARN of said SES verified email address). I then happened across one final section of the documentation that made the requested change in this PR more clear to me.

In the previously linked "Email settings for Amazon Cognito user pools" document, under the header **Step 4: Configure your user pool**, there is the following note:

> 5. On the **Edit email configuration** page, select **Send email from Amazon SES** or **Send email with Amazon Cognito**. You can customize the **SES Region**, **Configuration Set**, and **FROM sender name** only when you choose **Send email from Amazon SES**.

With that information in mind, it made it more clear that configuring the `FROM` address is possible in two ways:

1. When using `COGNITO_DEFAULT`, by passing the ARN of a verified SES email address in the `source_arn` argument of the resource.

Issue #30041, opened 2023-03-15
Private DNS options for VPC endpoints


ewbankkit:
> Virtual Private Cloud (VPC) interface endpoints for Amazon S3 now offer private DNS options that can help you more easily route S3 requests to the lowest-cost endpoint in your VPC.

* [Announcement](https://aws.amazon.com/about-aws/whats-new/2023/03/amazon-s3-private-connectivity-on-premises-networks/)
* [Blog post](https://aws.amazon.com/blogs/aws/celebrate-amazon-s3s-17th-birthday-at-aws-pi-day-2023/)

Requires [AWS SDK for Go v1.44.221](https://github.com/aws/aws-sdk-go/releases/tag/v1.44.221): https://github.com/hashicorp/terraform-provider-aws/pull/30015.

[`aws_vpc_endpoint`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpc_endpoint):

```hcl
resource "aws_vpc_endpoint" "example" {
  dns_options {
    private_dns_only_for_inbound_resolver_endpoint = true
  }
}
```

ewbankkit:
Relates https://github.com/hashicorp/terraform-provider-aws/pull/31873.
Relates https://github.com/hashicorp/terraform-provider-aws/issues/31117.

PR #31873, opened 2023-06-09
ec2: fix aws_vpc_endpoint missing exposed dns_option

FabianPonce:
### Description
This PR exposes the field `PrivateDnsOnlyForInboundResolverEndpoint` from [DnsOptionsSpecification](https://github.com/aws/aws-sdk-go-v2/blob/main/service/ec2/types/types.go#L2880). 

In order to create S3 Interface endpoints, this field must be set to false. The absence of this field results in an error message:
```
Error: creating EC2 VPC Endpoint (com.amazonaws.us-east-1.s3): InvalidParameter: To set PrivateDnsOnlyForInboundResolverEndpoint to true, the VPC vpc-abc123 must have a Gateway endpoint for the service.
```

### Relations
Closes #31117

### Output from Acceptance Testing

```
$ make testacc TESTS=TestAccXXX PKG=ec2
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/ec2/... -v -count 1 -parallel 20 -run='TestAccVPCEndpoint_'  -timeout 180m
=== RUN   TestAccVPCEndpoint_gatewayBasic
=== PAUSE TestAccVPCEndpoint_gatewayBasic
=== RUN   TestAccVPCEndpoint_interfaceBasic
=== PAUSE TestAccVPCEndpoint_interfaceBasic
=== RUN   TestAccVPCEndpoint_interfacePrivateDNS
=== PAUSE TestAccVPCEndpoint_interfacePrivateDNS
=== RUN   TestAccVPCEndpoint_disappears
=== PAUSE TestAccVPCEndpoint_disappears
=== RUN   TestAccVPCEndpoint_tags
=== PAUSE TestAccVPCEndpoint_tags
=== RUN   TestAccVPCEndpoint_gatewayWithRouteTableAndPolicy
=== PAUSE TestAccVPCEndpoint_gatewayWithRouteTableAndPolicy
=== RUN   TestAccVPCEndpoint_gatewayPolicy
=== PAUSE TestAccVPCEndpoint_gatewayPolicy
=== RUN   TestAccVPCEndpoint_ignoreEquivalent
=== PAUSE TestAccVPCEndpoint_ignoreEquivalent
=== RUN   TestAccVPCEndpoint_ipAddressType
=== PAUSE TestAccVPCEndpoint_ipAddressType
=== RUN   TestAccVPCEndpoint_interfaceWithSubnetAndSecurityGroup
[...]

PR #32021, opened 2023-06-15
r/vpc_endpoint: fix private_dns_only_for_inbound_resolver_endpoint

FabianPonce:
### Description
This PR is a follow-on from #31873 , which seems to have a defect. In order to create a VPC Endpoint for S3 in Interface mode, the field `private_dns_only_for_inbound_resolver_endpoint` MUST be set to `false`. 

In the previous PR, the value was only sent when `true`, otherwise as `nil`. AWS interprets the `nil` option the same as if it were set to `true`, which is odd.

In any case, it was impossible to set this value to `false`. 

In the case of this value being set to `false`, the test should have no dependency on a VPC Endpoint of GATEWAY type. The test passed in the previous PR because the gateway was created and the value was interpreted as `true`. 

In the case of this value being set to `true`, this test continues to have a dependency on the VPC Endpoint of GATEWAY type.

### Relations

Closes https://github.com/hashicorp/terraform-provider-aws/issues/30041.
Relates #31873.

```
make testacc TESTS=TestAccVPCEndpoint_interfacePrivateDNS PKG=ec2  2 ↵  6578  16:12:03 
TF_ACC=1 go test ./internal/service/ec2/... -v -count 1 -parallel 20 -run='TestAccVPCEndpoint_interfacePrivateDNS'  -timeout 180m
=== RUN   TestAccVPCEndpoint_interfacePrivateDNS
=== PAUSE TestAccVPCEndpoint_interfacePrivateDNS
=== CONT  TestAccVPCEndpoint_interfacePrivateDNS
--- PASS: TestAccVPCEndpoint_interfacePrivateDNS (449.81s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ec2	450.003s

...
```

### DIFF
--- a/.changelog/32021.txt
+++ b/.changelog/32021.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_vpc_endpoint: Fix error creating S3 _Interface_ VPC endpoint
+```
--- a/internal/service/ec2/vpc_endpoint.go
+++ b/internal/service/ec2/vpc_endpoint.go
@@ -492,7 +492,7 @@ func expandDNSOptionsSpecification(tfMap map[string]interface{}) *ec2.DnsOptions
 		apiObject.DnsRecordIpType = aws.String(v)
 	}
 
-	if v, ok := tfMap["private_dns_only_for_inbound_resolver_endpoint"].(bool); ok && v {
+	if v, ok := tfMap["private_dns_only_for_inbound_resolver_endpoint"].(bool); ok {
 		apiObject.PrivateDnsOnlyForInboundResolverEndpoint = aws.Bool(v)
 	}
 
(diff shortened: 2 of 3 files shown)


ewbankkit:
```
=== CONT  TestAccVPCEndpoint_ipAddressType
    vpc_endpoint_test.go:360: Step 1/3 error: Error running apply: exit status 1
        
        Error: creating EC2 VPC Endpoint (com.amazonaws.vpce.us-west-2.vpce-svc-00809ad67b7780d3f): InvalidParameter: PrivateDnsOnlyForInboundResolverEndpoint not supported for this service.
        	status code: 400, request id: 7751fc22-b3ff-4ac6-b44d-3109e2f7d05c
        
          with aws_vpc_endpoint.test,
          on terraform_plugin_test.tf line 60, in resource "aws_vpc_endpoint" "test":
          60: resource "aws_vpc_endpoint" "test" {
        
--- FAIL: TestAccVPCEndpoint_ipAddressType (237.20s)
```

FabianPonce:
Sorry for letting this go stale, will have an update in the next day or so.

ewbankkit:
@FabianPonce Thanks for the contribution :tada: :clap:. The functionality has been implemented in #32355.

Issue #28353, opened 2022-12-14
Notice: S3 - Changes to Defaults (Block Public Access and ACL's)


ewbankkit:
> Starting in April 2023, Amazon S3 will introduce two new default bucket security settings by automatically enabling S3 Block Public Access and disabling S3 access control lists (ACLs) for all new S3 buckets.

[Announcement](https://aws.amazon.com/about-aws/whats-new/2022/12/amazon-s3-automatically-enable-block-public-access-disable-access-control-lists-buckets-april-2023/)
[Blog post](https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/)
[FAQ](https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-faq.html)

Relevant resources:

* `aws_s3_bucket`
* `aws_s3_bucket_acl`
* `aws_s3_bucket_ownership_controls`
* `aws_s3_bucket_public_access_block`

Relates #22069
Relates #6607

---

## Terraform AWS Provider Impact

This section will detail the impact of the [upcoming AWS S3 security changes](https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/) on the Terraform AWS Provider.

### Summary

The impending S3 security changes will not require breaking changes to v3 or v4 of the Terraform AWS provider. However, some previously valid S3 bucket ACL configurations will begin returning errors for net-new buckets. Existing buckets (and their corresponding terraform configuration) are not impacted.

### ACL Changes

ACLs can be defined either directly on the `aws_s3_bucket` resource, or via the standalone `aws_s3_bucket_acl` resource. This is true in both v3 and v4 or the provider, though the `acl` attribute on the S3 bucket resource was officially deprecated in v4.

**Inline:**

```hcl
resource "aws_s3_bucket" "example" {
  bucket = "my-tf-test-bucket"
  acl    = "private"
}
```

**Standalone Resource:** 

```hcl
resource "aws_s3_bucket" "example" {
  bucket = "my-tf-test-bucket"
}

resource "aws_s3_bucket_acl" "example" {
  bucket = aws_s3_bucket.example.id
  acl    = "private"
}
```

While existing buckets are not impacted by the impending security changes, the following configuration patterns are impacted for net-new buckets:

1. Inline `acl` where the value is not `private`. 
2. Standalone `aws_s3_bucket_acl` resource with default security settings.

[...]

PR #32063, opened 2023-06-16
Update codepipeline.markdown to make the example work again

savar:
Since AWS changed the default to disable ACLs on S3 (setting "Bucket owner enforced" as the default), the code example does not work anymore but fail because it tries to set ACLs which are disabled.

The change is using the `aws_s3_bucket_public_access_block` resource to disable public access which should be the equivalent to set `acl = private`.

### Relations

Closes #0000

```

...
```


justinretzolk (changes requested):
Hey @savar 👋 Thank you for taking the time to raise this! I actually believe we can just remove the `aws_s3_bucket_acl` resource entirely (rather than replacing it with `aws_s3_bucket_public_access_block`) according to https://github.com/hashicorp/terraform-provider-aws/issues/28353.

savar:
Hey @justinretzolk,

yeah fair point. I wanted to make it explicit but we probably can also remove it. Feel free to change my code or if you want me to do it, tell me. I will then do it over the next day.

Cheers
Simon

Am 16. Juni 2023 22:52:42 UTC schrieb Justin Retzolk ***@***.***>:
>@justinretzolk requested changes on this pull request.
>
>Hey @savar 👋 Thank you for taking the time to raise this! I actually believe we can just remove the `aws_s3_bucket_acl` resource entirely (rather than replacing it with `aws_s3_bucket_public_access_block`) according to https://github.com/hashicorp/terraform-provider-aws/issues/28353. 
>
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/hashicorp/terraform-provider-aws/pull/32063#pullrequestreview-1484529539
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

ewbankkit:
@savar Thanks for the contribution :tada: :clap:.

PR #32087, opened 2023-06-19
Bugfix for aws_lambda_invocation lifecycle scope

theipster:
### Relations

Closes #0000

```

...
```


theipster:
Fixed via #32118.

Issue #30876, opened 2023-04-21
[Enhancement]: Update to Golang 1.20


stephen-opal:
### Description

Hey folks, when possible can we update to Golang 1.20? Reference [here](https://github.com/hashicorp/terraform-provider-aws/blob/main/.go-version). We're having some DNS over HTTPS issues with using our VPN provider that should be fixed in 1.20.

### Would you like to implement a fix?

None

ewbankkit:
Relevant: https://tip.golang.org/doc/go1.20#ports.

tristanmorgan:
Related issue [terraform#32778](https://github.com/hashicorp/terraform/issues/32778)

PR #32108, opened 2023-06-20
bump go.mod to 1.20

jar-b:
### Description
Updates `go.mod` to Go `1.20`. Also replaces package-scoped calls to `rand.Seed()` with randomized values, as Go 1.20 now seeds the global number generator in `math/rand` with a randomized value by default. From the [1.20 release notes](https://tip.golang.org/doc/go1.20):

> The [math/rand](https://tip.golang.org/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level Seed function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).

### Relations

Relates #32082 (see [failed checks](https://github.com/hashicorp/terraform-provider-aws/actions/runs/5308865624/jobs/9608821669?pr=32082)).
Closes https://github.com/hashicorp/terraform-provider-aws/issues/30876.

### DIFF
--- a/.changelog/32108.txt
+++ b/.changelog/32108.txt
@@ -1,0 +1,7 @@
+```release-note:note
+provider: Updates to Go 1.20, the last release that will run on macOS 10.13 High Sierra or 10.14 Mojave. A future release will update to Go 1.21, and these platforms will no longer be supported.
+```
+
+```release-note:note
+provider: Updates to Go 1.20, the last release that will run on any release of Windows 7, 8, Server 2008 and Server 2012. A future release will update to Go 1.21, and these platforms will no longer be supported.
+```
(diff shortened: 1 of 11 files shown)


ewbankkit:
We may need to call out the potentially now unsupported platforms (see https://tip.golang.org/doc/go1.20#ports) in the release notes. Or maybe not. I can't remember how where our internal discussions landed on this.

ewbankkit:
In theory we can replace uses of `github.com/hashicorp/go-multierror` with `errors.Join`: https://tip.golang.org/doc/go1.20#errors. Maybe open an Issue for this?

ewbankkit:
Do we want to call out the DNS changes -- https://tip.golang.org/doc/go1.20#minor_library_changes -- as they are explicitly mentioned in https://github.com/hashicorp/terraform-provider-aws/issues/30876.
See e.g. _UPGRADE NOTES_ in https://github.com/hashicorp/terraform/blob/v1.5/CHANGELOG.md#150-june-12-2023.

ewbankkit (approved):
LGTM :rocket:.

PR #31933, opened 2023-06-13
r/aws_lambda_provisioned_concurrency_config: Use comma-delimited `id`

jar-b:
### Description
Updates the resource `id` to a comma-separated string, joining `function_name` and `qualifier`. This allows the `function_name` argument to reliably accept ARN values which previously caused Read operations to fail, and migrates this resource to the preferred ID separator.

### Relations
Fixes #11152
Fixes #12923
Fixes #22392
Closes #19868
Closes #11679
Closes #22808
Relates #27843

### Output from Acceptance Testing
```console
$ make testacc PKG=lambda TESTS=TestAccLambdaProvisionedConcurrencyConfig_
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/lambda/... -v -count 1 -parallel 20 -run='TestAccLambdaProvisionedConcurrencyConfig_'  -timeout 180m

--- PASS: TestAccLambdaProvisionedConcurrencyConfig_Qualifier_aliasName (190.12s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_Disappears_lambdaProvisionedConcurrency (232.09s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_idMigration530 (243.59s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_Disappears_lambdaFunction (248.04s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_basic (307.14s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_provisionedConcurrentExecutions (310.16s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_skipDestroy (412.08s)
--- PASS: TestAccLambdaProvisionedConcurrencyConfig_FunctionName_arn (495.91s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/lambda     499.303s
```

PR #32121, opened 2023-06-20
[Docs] Update lambda_provisioned_concurrency_config.html.markdown Import Syntax

skyfox675:
Updated the import statement to match the import's error code and formatting.


jar-b:
HI @skyfox675 👋 - The separator for this resource's ID was changed to a `,` in #31933 to fix some long standing issues when passing function ARNs to the `function_name` argument. Documentation for versions >=5.4.0 will show the new separator. 

Switching to the provider version in the Terraform Registry will show the correct import command for earlier versions, for example:
https://registry.terraform.io/providers/hashicorp/aws/5.3.0/docs/resources/lambda_provisioned_concurrency_config.html#import

PR #32136, opened 2023-06-21
Fix string values in reusable team project automation

justinretzolk:
### Description

Since the newly added reusable workflow for automating the team project did not wrap the IDs of the project, field in quotes, it caused failures. This PR switches to interpolating them so that they may be passed as quote-wrapped strings.

### References

- [Failed run](https://github.com/hashicorp/terraform-provider-aws/actions/runs/5335208853/jobs/9668107745)

N/a, workflow


jar-b (approved):
LGTM

PR #32139, opened 2023-06-21
Fix community check for label events

justinretzolk:
### Description

Previously, we were comparing the list of users in our Secrets to the `github.actor`. For events such as labeling an issue or pull request, however, the `github.actor` is the person who labeled the item rather than the actual issue or pull request author. This means that there's false positives when maintainers (or partners, core contributors) label an item. This PR aims to fix that by using `github.event.*.user.login` instead (the `*` means that this will work for `github.event.issue.user.login` as well as `github.event.pull_request.user.login`).

### References

- [False positive result](https://github.com/hashicorp/terraform-provider-aws/actions/runs/5335208853/jobs/9668104459)

N/a, workflows

### DIFF
--- a/.github/workflows/community-check.yml
+++ b/.github/workflows/community-check.yml
@@ -1,5 +1,4 @@
 name: Community Check
-
 on:
   workflow_call:
     outputs:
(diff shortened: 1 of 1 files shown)


breathingdust (approved):
LGTM 🚀

PR #32181, opened 2023-06-23
tooling: updating the TeamCity configuration

tombuildsstuff:
### Description

This PR updates the TeamCity Configuration to use the newer syntax (which is now versionless).

I'm having a little trouble validating the `jenv` setup, but this is the latest version available from the Corretto site for v17, so I think should work fine?


jar-b (approved):
LGTM 🚀

PR #32241, opened 2023-06-27
tooling: updating the TeamCity Configuration

tombuildsstuff:
### Description

This PR makes 2 changes to the TeamCity Configuration:

1. Rather than pulling/updating `goenv`, we can assume that it's up to date (since these are updated regularly) - so we can rely on the version of Go we need being present rather than mutating the Build Agent.
2. Adds a feature-toggle to the Build Configuration for determining if we're running on-premise or not, to enable the migration by conditionally enabling/disabling all of the Build Triggers outside of the change here (so this repo can be configured on both instances whilst triggering the tests only on one).


jar-b (approved):
LGTM 🎉

PR #29424, opened 2023-02-15
Replace heredoc with jsonencode (r/a*, r/b*, r/c*)

mattburgess:
### Relations

Relates #17714

### Output from Acceptance Testing

```
$ make testacc TESTS=TestAccXXX PKG=ec2

...
```

PR #32282, opened 2023-06-29
docs: update aws_ecr_repository_policy to be written in jsonencode.

yuukiyo:
### Description
The resource for creating ECR resource policies, aws_ecr_repository_policy, currently only includes a description of using data source in the examples.
How about putting this one here as it is simpler to write in jsonencode?
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecr_repository_policy

### DIFF
--- a/website/docs/r/ecr_repository_policy.html.markdown
+++ b/website/docs/r/ecr_repository_policy.html.markdown
@@ -19,38 +19,38 @@ resource "aws_ecr_repository" "foo" {
   name = "bar"
 }
 
-data "aws_iam_policy_document" "foopolicy" {
-  statement {
-    sid    = "new policy"
-    effect = "Allow"
-
-    principals {
-      type        = "AWS"
-      identifiers = ["123456789012"]
-    }
-
-    actions = [
-      "ecr:GetDownloadUrlForLayer",
-      "ecr:BatchGetImage",
-      "ecr:BatchCheckLayerAvailability",
-      "ecr:PutImage",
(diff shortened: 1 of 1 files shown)


justinretzolk:
Hey @yuukiyo 👋 Thank you for taking the time to raise this! I gave this a bit of thought, but have ultimately decided that we'll not be able to accept this particular pull request. One of the things we mention in our contribution documentation on [documentation changes](https://hashicorp.github.io/terraform-provider-aws/documentation-changes/) is:

> **Avoid Terraform Configuration Language Features:** Individual resource documentation pages and examples should refrain from highlighting particular Terraform configuration language syntax workarounds or features such as variable, local, count, and built-in functions.

This includes `jsonencode()`, as it is a [Terraform function](https://developer.hashicorp.com/terraform/language/functions/jsonencode). I'd be remiss if I didn't note that there _are_ places throughout the AWS Provider documentation where `jsonencode()` is used, however, these cases fit into two main categories:

1. Documentation that has not been updated in quite some time, and _should_ be updated to use the `aws_iam_policy_document` data source.
1. Situations where the alternative to `jsonencode()` is using `HEREDOC` or escaped strings to produce JSON for a particular argument (for example, the changes introduced in #29424).

Given that in this particular case, the documentation is already using `data.aws_iam_policy_document` rather than `jsonencode()`, we'd prefer to stick with the former. Regardless, we really appreciate you taking the time to open this pull request so that we can take time to reconsider our approach, and hope that you continue to submit contributions in the future!

PR #32365, opened 2023-07-04
itsprout

blinovmaksym:
Hello. We are the IT Sprout team, and we would like to provide an example of using popular tools (such as Terraform, Wordpress, MySQL, Tekton) in the most popular cloud provider, AWS, to automate and optimize the deployment of a WordPress website.

Please take a look at our README.


blinovmaksym:
Good day , @ewbankkit !
Can you review our PR?
We want to demonstrate an example of using various technologies in the AWS cloud

PR #32386, opened 2023-07-05
Fix TeamCity configuration

gdavison:
### Description

The step "Configure GOENV" was not being added to the build steps, causing builds to fail


jar-b (approved):
LGTM 👍

PR #32387, opened 2023-07-05
workflow: GHA and text update in end user documentation highlight the resources/data-source count.

breathingdust:
### Description

Adds a resource/data-source count for the most recent version to the registry index page for the provider. This is a question often asked and will now be easier to find the answer.


jar-b (approved):
LGTM 🔢

PR #32409, opened 2023-07-07
Update kinesis_firehose_delivery_stream.html.markdown

benben:
fix `buffer_size` parameter in example

### DIFF
--- a/website/docs/r/kinesis_firehose_delivery_stream.html.markdown
+++ b/website/docs/r/kinesis_firehose_delivery_stream.html.markdown
@@ -161,9 +161,9 @@ resource "aws_kinesis_firehose_delivery_stream" "extended_s3_stream" {
   name        = "terraform-kinesis-firehose-extended-s3-test-stream"
   destination = "extended_s3"
   extended_s3_configuration {
-    role_arn    = aws_iam_role.firehose_role.arn
-    bucket_arn  = aws_s3_bucket.bucket.arn
-    buffer_size = 64
+    role_arn       = aws_iam_role.firehose_role.arn
+    bucket_arn     = aws_s3_bucket.bucket.arn
+    buffering_size = 64
     # https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
     dynamic_partitioning_configuration {
       enabled = "true"


website/docs/r/kinesis_firehose_delivery_stream.html.markdown
```diff
@@ -107,7 +107,7 @@ resource "aws_kinesis_firehose_delivery_stream" "extended_s3_stream" {
     role_arn   = aws_iam_role.firehose_role.arn
     bucket_arn = aws_s3_bucket.bucket.arn
 
-    buffering_size = 64
+    buffer_size = 64
```
justinretzolk, line 110: Based on the [documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/kinesis_firehose_delivery_stream#s3-configuration) and [schema](https://github.com/hashicorp/terraform-provider-aws/blob/dfd52e40ea10872efff5ec5b797ebcdcff85585e/internal/service/firehose/delivery_stream.go#L156), `buffering_size` is the correct argument. While this particular line should not be changed, line 166 incorrectly uses `buffer_size`. Was that the line you intended to edit?

```suggestion
    buffering_size = 64
```

justinretzolk (changes requested):

website/docs/r/kinesis_firehose_delivery_stream.html.markdown
benben: Hi @justinretzolk, thanks for taking a look but this is not correct. The block `extended_s3_configuration` only accepts `buffer_size` but not `buffering_size`. Line 166 shows `buffer_size` to me so this is correct. Here is a screenshot. 

image

It's funny that you are linking to the docs I am trying to fix. Also the schema for [this particular block is here](https://github.com/hashicorp/terraform-provider-aws/blob/dfd52e40ea10872efff5ec5b797ebcdcff85585e/internal/service/firehose/delivery_stream.go#L1025) I believe. It still says `buffering_size` so there might be another problem somewhere or I am not getting something.

```
Terraform v1.4.0
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v4.58.0
```

benben: just tried it with tf `1.5.2` and still the same behavior

justinretzolk: Hey @benben 👋 It looks like this is a result of your use of AWS Provider version `4.58.0`. In version `5.0.0`, the `buffer_size` argument [was renamed to `buffering_size`](https://github.com/hashicorp/terraform-provider-aws/blob/5ce64c172eff331b0de12e9e86b8773b75eb30a1/CHANGELOG.md?plain=1#L345). Looks like line 166 was missed when this change occurred, which is why it still reads as `buffer_size`.

justinretzolk (approved):
Looks good to me! Thank you again for taking the time to raise this, and for taking the time to work through discussing it! I'll get this merged once the checks finish up 🚀

PR #32427, opened 2023-07-07
Don't manually set `Status` field to `To Do` for new Projects automation

justinretzolk:
### Description

Previously, this workflow called the [reusable workflow](https://github.com/hashicorp/terraform-provider-aws/blob/main/.github/workflows/reusable-team-project.yml) that adds items to the new style of GitHub Project and also passed the requisite input to set the `Status` field to `To Do`. This Project, however, uses the built-in Projects workflow (reference below) to automatically set the `Status` field to `To Do` for newly added items. Removing this means that we save an API call (and associated time, rate-limiting) for every item that gets added.

YAML formatter also removed the extraneous newlines; figured I'd leave that be to try to maintain consistency. 🤷 

### References

- [GitHub Projects Automated Workflows](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-built-in-automations)

N/a, Actions

### DIFF
--- a/.github/workflows/new-project-automation.yml
+++ b/.github/workflows/new-project-automation.yml
@@ -1,16 +1,13 @@
 name: 'Team GitHub Projects (new) Automation'
-
 on:
   issues:
     types: ["labeled", "milestoned", "opened"]
   pull_request_target:
     types: ["labeled", "opened", "ready_for_review"]
-
 jobs:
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🎉

PR #32434, opened 2023-07-10
cdkft: use reusable workflow and translate more

DanielMSchmidt:
### Description

This refactors the action into the reusable github action workflow as a better abstraction and it bumps the cdktf version we use which should handle partial snippets better
I also enabled the translation of lambda, eks, and vpc prefixed docs.
Lastly I removed the org scoped token, it might be possible to create the automatic PR with just the GITHUB_TOKEN (since we don't act on the push or PR trigger where this is problematic)

### Relations

Closes #0000

```

...
```


DanielMSchmidt:
Oh no, the docs seem to be off, I'll be right back with a new PR

PR #32457, opened 2023-07-11
cdkft: use reusable workflow and translate more

DanielMSchmidt:
### Description

This refactors the action into the reusable github action workflow as a better abstraction and it bumps the cdktf version we use which should handle partial snippets better
I also enabled the translation of lambda, eks, and vpc prefixed docs.
Lastly I removed the org scoped token, it might be possible to create the automatic PR with just the GITHUB_TOKEN (since we don't act on the push or PR trigger where this is problematic)

### Relations

Closes #0000


DanielMSchmidt:
Got a PR to change tfproviderdocs to ignore cdktf files in the content check: https://github.com/bflad/tfproviderdocs/pull/80

.github/workflows/cdktf-documentation.yml
```diff
@@ -10,48 +10,13 @@ permissions:
...
-          commit-message: "docs: update cdktf documentation"
-          title: "docs: update cdktf documentation"
-          body: "This PR updates the cdktf related documentation based on the current HCL-based documentation. It is automatically created by the cdktf-documentation GitHub action."
-          token: ${{ secrets.ORGSCOPED_GITHUB_TOKEN }}
+    uses: hashicorp/terraform-cdk/.github/workflows/registry-docs-pr-based.yml@304e2507209c9657135dc7b4b7ee68030327468f
+    secrets:
+      GH_PR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    with:
+      providerFqn: "hashicorp/aws"
+      files: "*/ec2*.html.markdown,*/lambda*.html.markdown,*/eks*.html.markdown,*/vpc*.html.markdown"
```
ansgarm, line 18: Can we include `aws_instance` as that's the most known EC2 resource 😄

```suggestion
      files: "*/instance.html.markdown,*/ec2*.html.markdown,*/lambda*.html.markdown,*/eks*.html.markdown,*/vpc*.html.markdown"
```

DanielMSchmidt: I’ll do more PRs step by step, we can do it in the next one 👍 I’d rather keep this one as is. Over time expanding the documentation should be a non event, but right now I’m still with every PR adjusting the provider docs linter.

breathingdust (approved):
LGTM though I think that the job will probably fail due to PR creation being blocked at org level. If so we can update it to use our GitHub App token. However it may make more sense to create your own GitHub app that you could use and manage across all the repositories that you are creating cdktf docs for.

DanielMSchmidt:
Sounds good 👍 I'll do a follow-up PR anyways to include a few more missing resources, so just let me know which token to use there :)

DanielMSchmidt:
I also added `aws_instance` docs as per @ansgarm request. Can we get this in before the release?

PR #32474, opened 2023-07-12
workflow: Resource Counts - Use org token to all PR to be raised

breathingdust:
### Description

Update workflow to specify token for create pr step.


jar-b (approved):
LGTM

PR #32486, opened 2023-07-12
Document max items for `aws_wafv2_regex_pattern_set.regular_expression`

### LINKED ISSUE #32311: [Docs]: aws_wafv2_regex_pattern_set can only handle 10 regular_expression
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_regex_pattern_set#regular-expression

### Description

I have a module which takes a list of regexp strings. That list is then used by the `aws_wafv2_regex_pattern_set` resource to create one `regular_expression { regex_string = <string> }` per entry:

```
variable "string_match" {
  description                           = "Strings to match and block"
  type                                  = list(string)
  default                               = null
}

resource "aws_wafv2_regex_pattern_set" "string_match" {
  count                         = var.string_match != null ? 1 : 0

  name                          = var.prefix
  description                   = "STRING set - ${var.prefix}"
  scope                         = var.scope

  dynamic "regular_expression" {
    for_each = var.string_match
    content {
      regex_string = regular_expression.value
    }
  }
}
```

When I added the eleventh entry in that list, I got an error from TF/AWS that said it can only handle ten entries.

The solution is to split this up with `chunklist()` in a local and then use that instead:
```
locals {
  # The `aws_wafv2_regex_pattern_set` resource can only handle ten regex entries.
  # So split it up in entries of ten and create a map from that.
  string_match                          = {
    for key, val in chunklist(var.string_match, 10) : key => val
  }
}

resource "aws_wafv2_regex_pattern_set" "string_match" {
  for_each                      = local.string_match

  name                          = "${var.prefix}-${each.key}"
  description                   = "STRING set - ${var.prefix}-${each.key}"
  scope                         = var.scope

  dynamic "regular_expression" {
    for_each = each.value
    content {
      regex_string = regular_expression.value
    }
  }
}
```

This will create one `aws_wafv2_regex_pattern_set` resource for every ten regex entry. Another important update is the `name` attribute. Each resource must be unique, so suffixing it with the number will suffice.

A note about this limitation would suffice, but feel free to use my code as additional example.

### References

_No response_

None

justinretzolk:
### Description

The `aws_wafv2_regex_pattern_set.regular_expression` argument can take a maximum of 10 regular expressions. This limitation was not documented, so this PR adds a note indicating the limitation.

### Relations

Closes #32311

### References

- [Resource Schema](https://github.com/hashicorp/terraform-provider-aws/blob/d54824dccd7179c11347981505be66d90db750a5/internal/service/wafv2/regex_pattern_set.go#L80)
- [AWS Quotas documentation](https://docs.aws.amazon.com/waf/latest/developerguide/limits.html) (this limit is not mentioned in the API or SDK documentation, but is indicated here)

N/a, docs

### DIFF
--- a/website/docs/r/wafv2_regex_pattern_set.html.markdown
+++ b/website/docs/r/wafv2_regex_pattern_set.html.markdown
@@ -40,7 +40,7 @@ The following arguments are supported:
 * `name` - (Required) A friendly name of the regular expression pattern set.
 * `description` - (Optional) A friendly description of the regular expression pattern set.
 * `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the region `us-east-1` (N. Virginia) on the AWS provider.
-* `regular_expression` - (Optional) One or more blocks of regular expression patterns that you want AWS WAF to search for, such as `B[a@]dB[o0]t`. See [Regular Expression](#regular-expression) below for details.
+* `regular_expression` - (Optional) One or more blocks of regular expression patterns that you want AWS WAF to search for, such as `B[a@]dB[o0]t`. See [Regular Expression](#regular-expression) below for details. A maximum of 10 `regular_expression` blocks may be specified.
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🎉

PR #32500, opened 2023-07-13
Add `username` input to community check reusable workflow

justinretzolk:
### Description

Previously, the community check reusable workflow based its determination on `github.event.issue.user.login || github.event.pull_request.user.login`, which maps to the username of the user who created the issue or pull request. This, however, isn't always accurate when it comes to certain things we'd like to automate.

For example, for `issue_comment` events (such as is used in the `issue-comment-created` workflow, used to remove the `waiting-response` label when a non-maintainer comments on an issue or PR), this meant that when a comment was made, the determination was based on the original issue/PR author's username, rather than the commenter's username.

This PR adds a `username` input to the community check workflow to allow us to check a _specific_ username, and updates all workflows that use the workflow to include this now-required input. This also enables additional possibilities, such as adding an item to the team's Project board when it gets assigned to a maintainer, and possibly additional things down the line.

### Relations

N/a

### References

N/a

N/a, Actions

### DIFF
--- a/.github/workflows/autoremove_labels.yml
+++ b/.github/workflows/autoremove_labels.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   community_check:
-    if: github.event_name == 'issues'
+    if: github.event.action == 'opened'
     uses: ./.github/workflows/community-check.yml
     secrets: inherit
     with:
@@ -17,7 +17,7 @@ jobs:
   RemoveNeedsTriageFromMaintainers:
     name: Remove needs-triage for Maintainers
     needs: community_check
-    if: github.event.action == 'opened' && needs.community_check.outputs.maintainer == 'true'
+    if: needs.community_check.outputs.maintainer == 'true'
     uses: ./.github/workflows/reusable-add-or-remove-labels.yml
     with:
       remove: needs-triage
--- a/.github/workflows/autoremove_labels.yml
+++ b/.github/workflows/autoremove_labels.yml
@@ -8,11 +8,14 @@ on:
 
 jobs:
   community_check:
+    if: github.event_name == 'issues'
     uses: ./.github/workflows/community-check.yml
     secrets: inherit
+    with:
+      username: ${{ github.event.issue.user.login }}
 
   RemoveNeedsTriageFromMaintainers:
-    name: 'Remove needs-triage for Maintainers'
+    name: Remove needs-triage for Maintainers
     needs: community_check
     if: github.event.action == 'opened' && needs.community_check.outputs.maintainer == 'true'
     uses: ./.github/workflows/reusable-add-or-remove-labels.yml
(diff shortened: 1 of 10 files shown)


.github/workflows/autoremove_labels.yml
```diff
@@ -8,19 +8,22 @@ on:
 
 jobs:
   community_check:
+    if: github.event.action == 'opened'
```
justinretzolk, line 11: This is only needed for the `RemoveNeedsTriageFromMaintainers` job, which previously had this check in its conditional. Moving it here allows us to skip this job entirely if the event type is `pull_request_target`. This also means that if it's skipped, `RemoveNeedsTriageFromMaintainers` is skipped, since it's in the `needs` of that job. As a third benefit, it makes the conditional in that job a bit cleaner looking.

.github/workflows/changelog.yml
```diff
@@ -8,8 +8,11 @@ on:
 
 jobs:
   community_check:
+    if: github.event_name == 'pull_request_target'
```
justinretzolk, line 11: This one has a similar idea as is mentioned in my comment on `autoremove_labels.yml`, but has a bit of an interesting caveat that I'd like to look into more.

Previously, this conditional was in the `changes` job. This meant that if the event type was `push`, the `changes` job would be skipped, which would caused the `comment` job to be skipped as well, since `changes` is in its `needs`. This means that `push` events never trigger any of the jobs in this workflow.

.github/workflows/dependencies.yml
```diff
@@ -9,8 +9,11 @@ on:
 
 jobs:
   community_check:
+    if: github.event_name == 'pull_request_target'
```
justinretzolk, line 12: See comment on `changelog.yml`.

.github/workflows/pull_requests.yml
```diff
@@ -1,12 +1,14 @@
+name: Pull Request Target (All types)
```
justinretzolk, line 1: There's really no point in moving this up, but pretty much every other workflow has `name` at the top so 🤷

.github/workflows/changelog.yml
justinretzolk: In looking at this further, the `comment` job also uses `github.event.pull_request.number`, which isn't included in `push` events, so it seems safe to remove the `on.push` triggers entirely. Will wait for review before doing so, in case I'm overlooking anything.

.github/workflows/dependencies.yml
justinretzolk: In looking at this further, the `comment` action uses `github.event.pull_request.number`, which is not included in `push` events. I'm hesitant to remove the `on.push` triggers from this one, since the `go_mod` job would still run for those events, though I'm unsure of the utility of that job at the time of writing.

breathingdust (approved):
LGTM 🚀

PR #32523, opened 2023-07-14
Workflow: Update resource count action to use app token

breathingdust:
### Description

Updates action to use GitHub app in order to source a token with the correct scopes to be able to create a PR.


jar-b (approved):
LGTM

PR #32569, opened 2023-07-18
Label Updates

breathingdust:
### Description

Some label updates relating to internal process improvements

### DIFF
--- a/infrastructure/repository/labels-workflow.tf
+++ b/infrastructure/repository/labels-workflow.tf
@@ -3,7 +3,7 @@
 
 variable "workflow_labels" {
   default = {
-    # Stop the search. Keep these alphabetibelized.
+    # Stop the search. Keep these alphabetized.
 
     "authentication" = {
       color       = "844fba", # color:terraform (main)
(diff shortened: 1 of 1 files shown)


infrastructure/repository/labels-workflow.tf
```diff
@@ -117,9 +125,9 @@ variable "workflow_labels" {
       color       = "828a90", # color:stale grey
       description = "Repository modifications; GitHub Actions, developer docs, issue templates, codeowners, changelog."
     },
-    "roadmap" = {
+    "prioritized" = {
```
justinretzolk, line 128: My only concern with removing `roadmap` before we get `prioritized` in place is that it'll remove that label from everything that currently has it. I almost wonder if we should do it in two passes so that we can make sure everything that has `roadmap` currently gets the `prioritized` label before we remove it. Other than that, this looks good to me

breathingdust: Theres only 2 open ones and I've made a note of them. That said, it doesn't make sense to have `prioritized` on closed issues so we should add removing that label on merge/release to the automation list

PR #32625, opened 2023-07-20
TeamCity: Add failure condition to sweeper run to match region failures

gdavison:
### Description

Adds failure condition to sweeper run to match region failures. This will highlight each regional failure in the Build Problems list

image


ewbankkit (approved):
LGTM :rocket:.

Issue #31992, opened 2023-06-14
[Docs]: Changing abort_incomplete_multipart_upload to abort_incomplete_multipart_upload_days


dunefro:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration#abort_incomplete_multipart_upload

### Description

When trying to use `abort_incomplete_multipart_upload` with a non-zero positive integer value for `days_after_initiation` the following error comes up
```
Error: updating S3 Bucket Lifecycle Configuration (MYBUCKETNAME): InvalidRequest: At least one action needs to be specified in a rule
```
After changing this `abort_incomplete_multipart_upload_days = <positiveint>` it is working fine

### Would you like to implement a fix?

Yes

justinretzolk:
Hey @dunefro 👋 Thank you for taking the time to raise this! I'm not quite following what you've described here, as the schema for this resource includes a `abort_incomplete_multipart_upload` argument, but not a `abort_incomplete_multipart_upload_days` argument. Can you supply a sample Terraform configuration that illustrates your findings?

dunefro:
I was trying to run this
```
module "some_bucket" {
  source  = "terraform-aws-modules/s3-bucket/aws"
  version = "3.13.0"

  bucket        = "BUCKET_NAME_HERE"

  force_destroy = true


  # S3 bucket-level Public Access Block configuration
  block_public_acls       = true
  block_public_policy     = true
  ignore_public_acls      = true
  restrict_public_buckets = true

  lifecycle_rule = [
    {
      id      = "abort-multipart-uploads"
      status  = "Enabled"
      prefix  = ""
      enabled = true

      abort_incomplete_multipart_upload = {
        days_after_initiation = 7
      }
    }
  ]
}
```
and was getting the following error
```
│ Error: updating S3 Bucket Lifecycle Configuration (BUCKET_NAME_HERE): InvalidRequest: At least one action needs to be specified in a rule
│       status code: 400, request id: QRD22QRX035YNWJ5, host id: nqmc4mvzbcMrUYu1Fg2qnNbkWRI6wJH2Q5vgNCiJM5SoexD/HLxjwT/A1P2rX8F7srVjMYO+oMwTgvX76yjXeg==
│ 
│   with module.some_bucket.aws_s3_bucket_lifecycle_configuration.this[0],
[...]

PR #32640, opened 2023-07-21
issue-31992-abort_incomplete_multipart_upload-doc-updated

divyafun:
Issue https://github.com/hashicorp/terraform-provider-aws/issues/31992 fixed updated document

### DIFF
--- a/website/docs/r/s3_bucket_lifecycle_configuration.html.markdown
+++ b/website/docs/r/s3_bucket_lifecycle_configuration.html.markdown
@@ -373,7 +373,7 @@ Since `prefix` is deprecated by Amazon S3 and will be removed in the next major 
 
 The `rule` configuration block supports the following arguments:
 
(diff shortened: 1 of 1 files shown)


justinretzolk:
Hey @divyafun 👋 Thank you for taking the time to raise this! Based on the [schema](https://github.com/hashicorp/terraform-provider-aws/blob/a932c7b38e1e337bca8d7ffed480a3a1adbb3b33/internal/service/s3/bucket_lifecycle_configuration.go#L59-L71), the current documentation appears to be correct. It looks like the original issue was caused by the configuration for the module that the original poster was using. For more information, [see my comment](https://github.com/hashicorp/terraform-provider-aws/issues/31992#issuecomment-1648636385) on the original issue.

With that in mind, I'll close this pull request out, but we really appreciate your contribution, and hope to see more in the future!

Issue #28582, opened 2022-12-27
[Docs]: aws_cloudtrail advanced_event_selector - not_equals should be not_starts_with


andrewmichaelsmith:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudtrail

### Description

It's possible I'm missing something here but this is what I've observed in my AWS account. Following the cloud trail docs, the following example is intended to limit data events for 2 specific buckets:

```
data "aws_s3_bucket" "not-important-bucket-1" {
  bucket = "not-important-bucket-1"
}

data "aws_s3_bucket" "not-important-bucket-2" {
  bucket = "not-important-bucket-2"
}

resource "aws_cloudtrail" "example" {
  # ... other configuration ...

  advanced_event_selector {
    name = "Log all S3 objects events except for two S3 buckets"

    field_selector {
      field  = "eventCategory"
      equals = ["Data"]
    }

    field_selector {
      field = "resources.ARN"

      not_equals = [
        "${data.aws_s3_bucket.not-important-bucket-1.arn}/",
        "${data.aws_s3_bucket.not-important-bucket-2.arn}/"
      ]
    }

    field_selector {
      field  = "resources.type"
      equals = ["AWS::S3::Object"]
    }
  }
```

When I tried this myself, I still saw data events for those two buckets (which is counter to the intended behaviour). However, when I changed from `not_equals` not `not_starts_with` I no longer observed data events from those two buckets.

The AWS docs are not explicit about this, but it does say that to *include* data events to use `StartsWith`, so I believe it's correct to use `NotStartsWith` ([ref](https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_AdvancedFieldSelector.html)): 

```
 resources.ARN - You can use any operator with resources.ARN, but if you use Equals or NotEquals, the value must exactly match the ARN of a valid resource of the type you've speciﬁed in the template as the value of resources.type. For example, if resources.type equals AWS::S3::Object, the ARN must be in one of the following formats. To log all data events for all objects in a specific S3 bucket, use the StartsWith operator, and include only the bucket ARN as the matching value.
```

### References

Terraform docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudtrail
AWS docs: https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_AdvancedFieldSelector.html)

### Would you like to implement a fix?

None

PR #32670, opened 2023-07-24
Fix `aws_cloudtrail` docs examples

### LINKED ISSUE #32619: [Docs]: aws_cloudtrail advanced_event_selector - equals should be starts_with
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudtrail

### Description

This is similar to https://github.com/hashicorp/terraform-provider-aws/issues/28582. 
It seems there is still one example left where `equals` needs to be replaced with `starts_with`.

```
field_selector {
      field = "resources.ARN"

      #The trailing slash is intentional; do not exclude it.
      equals = [
        "${data.aws_s3_bucket.important-bucket-1.arn}/",
        "${data.aws_s3_bucket.important-bucket-2.arn}/"
      ]
    }

    field_selector {
      field  = "readOnly"
      equals = ["false"]
    }

    field_selector {
      field  = "resources.type"
      equals = ["AWS::S3::Object"]
    }
  }
 ```

### References

https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_AdvancedFieldSelector.html

None

justinretzolk:
### Description

This PR fixes an example in the `aws_cloudtrail` documentation. This `advanced_event_selector`'s name is `"Log PutObject and DeleteObject events for two S3 buckets"`, so with a `field_selector.field` of `"resources.ARN"`, `starts_with` should be used instead of `equals` in order to catch the objects in the bucket, rather than the bucket itself.

Also removed a comment of `#The trailing slash is intentional; do not exclude it.`, as there was no trailing slash where it was referenced.

### Relations

Closes #32619
Relates #28582

N/a, docs

### DIFF
--- a/website/docs/r/cloudtrail.html.markdown
+++ b/website/docs/r/cloudtrail.html.markdown
@@ -244,7 +244,7 @@ resource "aws_cloudtrail" "example" {
       field = "resources.ARN"
 
       #The trailing slash is intentional; do not exclude it.
-      equals = [
+      starts_with = [
         "${data.aws_s3_bucket.important-bucket-1.arn}/",
         "${data.aws_s3_bucket.important-bucket-2.arn}/"
       ]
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

PR #32683, opened 2023-07-25
Add conditionals to community comment jobs

justinretzolk:
### Description

This PR corrects a few jobs related to `pull_request_target` and `issues` events so that community notes are only added when issues/PRs are opened, rather than on every event type.

### Relations

Relates #32605

N/a, actions


breathingdust (approved):
LGTM 🚀

PR #32752, opened 2023-07-29
Update DLM Docs to show that count and interval cannot be used together 

ceciliaames695:
Added some documentation to note that you can't use both count and interval in a single DLM retention schedule retain_rule block (as experienced with terraform and confirmed in AWS documentation here: https://docs.aws.amazon.com/dlm/latest/APIReference/API_RetainRule.html)

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/website/docs/r/dlm_lifecycle_policy.markdown
+++ b/website/docs/r/dlm_lifecycle_policy.markdown
@@ -307,9 +307,9 @@ This resource supports the following arguments:
 
 #### Retain Rule arguments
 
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Thank you for taking the time to raise this! I noticed that there were a couple of other `*_rule` blocks that had similar restrictions, so I pushed a change for those as well. I'll get this merged in once the checks finish up. Thanks for your contribution! 🚀

PR #32771, opened 2023-07-31
Improve conditionals for adding PRs to Project

justinretzolk:
### Description

Previously, we were only adding PRs that were ready for review to the Project board, however, this caused some PRs to be missed due to a race condition with labeling. This PR simplifies the conditional such the drafts are now added as well, and the items are added as soon as they're labeled (in the case of maintainers, as soon as it's opened).

N/a, Actions

### DIFF
--- a/.github/workflows/pull_request_target.yml
+++ b/.github/workflows/pull_request_target.yml
@@ -119,13 +119,13 @@ jobs:
           gh project item-edit --id "$PROJECT_ITEM_ID" --project-id ${{ env.PROJECT_ID }} --field-id ${{ env.VIEW_FIELD_ID }} --single-select-option-id ${{ vars.team_project_view_working_board }}
 
       - name: 'Ready Partner Pull Requests'
-        if: contains(github.event.pull_request.labels.*.name, 'partner') && !github.event.pull_request.draft
+        if: contains(fromJSON('["opened", "ready_for_review"]'), github.event.action) && needs.community_check.outputs.partner == 'true' && !github.event.pull_request.draft
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_target.yml
```diff
@@ -119,13 +119,13 @@ jobs:
           gh project item-edit --id "$PROJECT_ITEM_ID" --project-id ${{ env.PROJECT_ID }} --field-id ${{ env.VIEW_FIELD_ID }} --single-select-option-id ${{ vars.team_project_view_working_board }}
 
       - name: 'Ready Partner Pull Requests'
-        if: contains(github.event.pull_request.labels.*.name, 'partner') && !github.event.pull_request.draft
+        if: contains(fromJSON('["opened", "ready_for_review"]'), github.event.action) && needs.community_check.outputs.partner == 'true' && !github.event.pull_request.draft
```
breathingdust, line 122: Would it not be simpler to have those checks just work on labelled event?

justinretzolk: It could be, but the label could be applied while the PR was still a draft, so we wouldn't be able to differentiate on whether it was a draft PR or not before adding it to the Project. If we're okay with adding draft PRs to the Project, I can make that change.

breathingdust: I'd definitely prefer to include the drafts. It seems simpler to disconnect the events, and just have one trigger apply labels and another triggering view assignment when labelled. Don't feel strongly about it if you would prefer this method.

justinretzolk: That makes sense to me -- I'd based the conditionals on excluding drafts, but including them makes this quite a bit easier. I've pushed a change with that in mind.

breathingdust (approved):
LGTM 🚀

PR #31806, opened 2023-06-06
[New Resource]: `aws_finspace_kx_cluster`

kavan-aws:
### Description

This pull request introduces the `finspace_kx_cluster` resource, along with its associated tests and documentation. This is one component of four new resources being introduced to enable terraform support for FinSpace Managed kdb.

### Relations

Relates #22658

### References

- https://github.com/aws/aws-sdk-go-v2/tree/main/service/finspace
- https://docs.aws.amazon.com/finspace/latest/userguide/finspace-managed-kdb.html

### Output from Acceptance Testing

```
$ make testacc TESTS=TestAccKxCluster PKG=finspace
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/finspace/... -v -count 1 -parallel 20 -run='TestAccFinSpaceKxCluster_basic'  -timeout 180m
=== RUN   TestAccFinSpaceKxCluster_basic
=== PAUSE TestAccFinSpaceKxCluster_basic
=== CONT  TestAccFinSpaceKxCluster_basic
--- PASS: TestAccFinSpaceKxCluster_basic (3641.54s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/finspace	3665.680s

==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/finspace/... -v -count 1 -parallel 20 -run='TestAccFinSpaceKxCluster'  -timeout 180m
=== RUN   TestAccFinSpaceKxCluster_disappears
=== PAUSE TestAccFinSpaceKxCluster_disappears
=== RUN   TestAccFinSpaceKxCluster_description
=== PAUSE TestAccFinSpaceKxCluster_description
=== RUN   TestAccFinSpaceKxCluster_database
=== PAUSE TestAccFinSpaceKxCluster_database
=== CONT  TestAccFinSpaceKxCluster_description
=== CONT  TestAccFinSpaceKxCluster_database
=== CONT  TestAccFinSpaceKxCluster_disappears
--- PASS: TestAccFinSpaceKxCluster_description (4034.09s)
--- PASS: TestAccFinSpaceKxCluster_disappears (4056.67s)
--- PASS: TestAccFinSpaceKxCluster_database (4122.80s)
PASS
ok	github.com/hashicorp/terraform-provider-aws/internal/service/finspace	4147.613s

==> Checking that code complies with gofmt requirements...
[...]

PR #32800, opened 2023-08-01
B finspace kx cluster savedown size

kavan-aws:
### Description

Increase minimum of savedown storage size attribute to be 10 GiB rather than 4, in line with recent model changes to address bug with sizes smaller than this. Does not require any test changes, as test used value of 500 GiB.

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/31806

### References

Will soon also be updated on https://docs.aws.amazon.com/finspace/latest/management-api/API_KxSavedownStorageConfiguration.html


jar-b (approved):
LGTM 🎉

PR #32849, opened 2023-08-03
Move `issues` and `pull_request_target` workflows to GitHub App

justinretzolk:
### Description

This PR moves the `issues.yml` and `pull_request_target.yml` workflows over to use a GitHub App for authentication when labeling or adding items to the GitHub Project. This is necessary, as using the automatically generated GitHub token does not allow the workflow to trigger itself, so when the workflows label something (e.g. with `partner`), it won't re-trigger the workflow, causing the `label` event, which is what we rely on to add the item to the project.

Note: this will require ensuring that our GitHub App has write permissions for issues and pull requests, as well as read/write for organization projects.

### References

- [An example](https://github.com/hashicorp/terraform-provider-aws/actions/runs/5754481875/job/15599855690) of a partner PR being opened, labeled appropriately, and not being added to the project
- [GitHub Documentation](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) indicating the limitation that we're hitting:

> When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN`, with the exception of `workflow_dispatch` and `repository_dispatch`, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's `GITHUB_TOKEN`, a new workflow will not run even when the repository contains a workflow configured to run when push events occur. For more information, see "[Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)."
> 
> If you do want to trigger a workflow from within a workflow run, you can use a GitHub App installation access token or a personal access token instead of GITHUB_TOKEN to trigger events that require a token.

N/a, Actions

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -20,12 +20,26 @@ jobs:
       # https://docs.github.com/en/actions/learn-github-actions/expressions#example
       username: ${{ github.event.action == 'assigned' && github.event.assignee.login || github.event.issue.user.login }}
 
+  generate_token:
+    name: 'Generate Token'
+    runs-on: ubuntu-latest
+    outputs:
+      token: ${{ steps.generate.outputs.token }}
+    steps:
+      - name: 'Generate'
+        id: generate
+        uses: tibdex/github-app-token@b62528385c34dbc9f38e5f4225ac829252d1ea92 # v1.8.0
+        with:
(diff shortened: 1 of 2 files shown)


ewbankkit (approved):
LGTM :rocket:.

PR #32866, opened 2023-08-04
Ignore additional error when deleting fleet

### LINKED ISSUE #31375: [Bug]: EMR aws_emr_instance_fleet tries to modify a deleted resource and fails when emr was manually terminated.
### Terraform Core Version

1.2.8

### AWS Provider Version

4.67.0

### Affected Resource(s)

aws_emr_instance_fleet

### Expected Behavior

When an EMR with an attached Instance Fleet terminates for a reason or another (failure, end of steps, manual termination from the console or CLI), it also terminates the Instance Fleet. This is the AWS imposed behavior. There is no API to delete an Instance Fleet, when created, its lifecycle is attached to the one of the EMR cluster. This is also written in the AWS Provider documentation for `aws_emr_instance_fleet`.

So when we re-create the same EMR with a `terraform apply`, we expect the AWS provider to know this rule and not trying to modify the no-more existing Instance Fleet.

### Actual Behavior

Until May, 10 2023, this was working perfectly as expected since months everyday (we have a CI/CD that do that).

Then, on May, 11, the same Terraform code, with the same Terraform version and the same AWS Provider fails because it tries to modify the Instance Fleet on `terraform apply`, and this Instance Fleet does not exists anymore because it was destroyed by the preceding cluster "terminate". We suspect that something changed on AWS side because nothing changed on Terraform side. Perhaps it reveals an always-existing bug that was not visible before.

The behavior is systematic and easy to reproduce : provision an EMR cluster and an Instance Fleet (for TASKS) on this EMR with Terraform. Then terminate the EMR from the console/cli/api/whatever. Then re-do a `terraform apply`. The provider tries to modify the Instance Fleet, but it does not exists anymore. The whole `apply` then fails. We think the provider should not fail if the resource does not exists anymore.

Also, the error occurs during a "Destroying" step of the Instance Fleet, which is an impossible action (there is no API to destroy an Instance Fleet). The only actions that exists on an Instance Fleet in the API are: add, modify, list. When created, the lifecycle depends on the EMR cluster on which it is attached. So "destroy" on this resource should be a no-op.

```shell
aws_emr_instance_fleet.emr-task-fleet: Destroying... [id=if-18V3MU5UCXJGI]
╷
│ Error: deleting EMR Instance Fleet (if-18V3MU5UCXJGI): InvalidRequestException: A job flow that is shutting down, terminated, or finished may not be modified.
│ {
│   RespMetadata: {
│     StatusCode: 400,
│     RequestID: "495707fb-97ae-425a-afde-74284d8a91d2"
│   },
│   ErrorCode: "INVALID_JOB_SHUT_DOWN",
│   Message_: "A job flow that is shutting down, terminated, or finished may not be modified."
│ }
│ 
│ 
╵
Cleaning up project directory and file based variables 00:01
ERROR: Job failed: exit code 1
```

Provisioning an EMR is very long, this requires a lot of resources... And our code is full of specific (unsharable) variables and stack dependencies. We currently have no bandwidth to produce a minimal context independent stack.

### Steps to Reproduce

* Provision a `aws_emr_cluster` and a `aws_emr_instance_fleet` attached to it with Terraform.
* Terminate the EMR cluster with the console, CLI or API.
* Re-apply the same Terraform : it fails because the no more existing resources try to be modified.

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

This was working perfectly until May, 10. Then since May, 11, it fails without any change on Terraform, provider or code. Something may have changed on AWS side, however, AWS Provider should be aware of the special Instance Fleet lifecycle and should behave accordingly (if the emr does not exists anymore, the Instance Fleet does not exists also).

To bypass the problem, we currently cleanup the state by removing the Instance Fleet with a `terraform state rm aws_emr_instance_fleet.emr-task-fleet`. Then we can `apply` successfully.

### References

_No response_

No

erpel:
### Description
This adds another error message from the AWS api to be ignored when trying to delete an AWS EMR Instance Fleet which does not support "delete". Its lifecycle is instead linked to the EMR cluster.

### Relations

Closes #31375

The acceptance testing failed but I'm fairly certain this is due to the locked down nature of the AWS Accounts that I have access to in my org. I would be very grateful if someone with a known working acceptance testing setup could reproduce this and maybe add their output.
I verified the fix with a cluster setup that was failing for my org using the published provider versions.

```console
% make testacc TESTARGS='-run=TestAccEMRCluster_ebs\|TestAccEMRCluster_basic' PKG=emr ACCTEST_PARALLELISM=2
TF_ACC=1 go test ./internal/service/emr/... -v -count 1 -parallel 2  -run=TestAccEMRCluster_ebs\|TestAccEMRCluster_basic -timeout 180m
=== RUN   TestAccEMRCluster_basic
=== PAUSE TestAccEMRCluster_basic
=== RUN   TestAccEMRCluster_ebs
=== PAUSE TestAccEMRCluster_ebs
=== CONT  TestAccEMRCluster_basic
=== CONT  TestAccEMRCluster_ebs
    cluster_test.go:1537: Step 1/2 error: Error running apply: exit status 1
        
        Error: running EMR Job Flow: ValidationException: Invalid InstanceProfile: EMR_EC2_DefaultRole.
        	status code: 400, request id: e7786da7-64bd-4280-adf3-6b41ae0028b5
        
          with aws_emr_cluster.test,
          on terraform_plugin_test.tf line 88, in resource "aws_emr_cluster" "test":
          88: resource "aws_emr_cluster" "test" {
        
--- FAIL: TestAccEMRCluster_ebs (189.08s)
--- PASS: TestAccEMRCluster_basic (457.25s)
FAIL
FAIL	github.com/hashicorp/terraform-provider-aws/internal/service/emr	459.257s
FAIL
make: *** [testacc] Error 1
```

### DIFF
--- a/.changelog/32866.txt
+++ b/.changelog/32866.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_emr_instance_fleet: Fix fleet deletion failing for terminated clusters
+```
--- a/internal/service/emr/instance_fleet.go
+++ b/internal/service/emr/instance_fleet.go
@@ -332,7 +332,11 @@ func resourceInstanceFleetDelete(ctx context.Context, d *schema.ResourceData, me
 		},
 	})
 
+	// Ignore certain errors that indicate the fleet is already (being) deleted
 	if tfawserr.ErrMessageContains(err, emr.ErrCodeInvalidRequestException, "instance fleet may only be modified when the cluster is running or waiting") {
+		return diags
+	}
+	if tfawserr.ErrMessageContains(err, emr.ErrCodeInvalidRequestException, "A job flow that is shutting down, terminated, or finished may not be modified") {
 		return diags
 	}
 


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@erpel Thanks for the contribution :tada: :clap:.

PR #32869, opened 2023-08-04
Increase mwaa_environment create timeout to 4 hours

toadjaune:
### Description

While the current value (2 hours) may already sound huge, it may not always be enough.
For example, I had a run with a configuration that turned out to be invalid, that took around 3h (I believe my network setup was lacking NAT gateways during this specific attempt).

I'm unsure, reading the guidelines, if this change is big enough to mandate a changelog entry. Please let me know if I should add one.

### Relations
N/A

### References
N/A

I don't believe timeout values are covered by acceptance tests, are they ?


toadjaune:
Hi folks,

Any chance this could be prioritized ? :)

PR #32926, opened 2023-08-09
Change the type of aws_codeartifact_domain asset_size_bytes to float …

misak113:
…(64) instead of int (32)

Because this value can get really big and the Integer (signed 32 size) can overflow. The maximum int32 value is 2147483647. See https://pkg.go.dev/builtin#int That is roughly 2 GBi of data that is easy to reach.

Example of error:
```
panic: Error reading level state: strconv.ParseInt: parsing "15679188538": value out of range
```


johnsonaj:
Closes #32926

johnsonaj:
@misak113 thank you for the contribution! 🎉  I incorporated your changes into a a new PR so that we could add tests

PR #32988, opened 2023-08-13
update lightsail_database docs to use the correct name argument

k4rtikp4til:
### Description

The correct argument for the name of the database in `aws_lightsail_database` is `relational_database_name` instead of `name`

### References

See `relational_database_name` in https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/lightsail/database.go

### DIFF
--- a/website/docs/r/lightsail_database.html.markdown
+++ b/website/docs/r/lightsail_database.html.markdown
@@ -20,13 +20,13 @@ for more information.
 
 ```terraform
 resource "aws_lightsail_database" "test" {
-  name                 = "test"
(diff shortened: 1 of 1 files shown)


k4rtikp4til:
Unless the correct argument name should indeed be `name` instead of `relational_database_name`, unsure if the argument was set as such to accomodate future introduction of non-relational databases by aws in lightsail.

justinretzolk (approved):
Looks good to me! I'll get this merged as soon as the checks finish up. Thank you for taking the time to contribute, and for helping us to keep improving our documentation! 🚀

PR #33010, opened 2023-08-14
Fix `issue_comment` workflow

justinretzolk:
### Description

The `issue_comment` workflow was intended to only remove `waiting-response` and `stale` when a non-maintainer commented on issues/pull requests. Since the conditional was missing, it was running when maintainers commented too.

N/a, workflow


ewbankkit (approved):
LGTM :rocket:.

Issue #32976, opened 2023-08-11
[Provider]: Migration to aws-sdk-go-v2


breathingdust:
### Description

The majority of the Terraform AWS Providers resources and data sources are based on [aws-sdk-go](https://github.com/aws/aws-sdk-go). Currently we support the use of both [aws-sdk-go](https://github.com/aws/aws-sdk-go) and [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2) in the provider, and have a significant number of resources based on [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2). 

AWS have now [announced](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/) the End-Of-Support plan for the AWS SDK for Go v1 which presents these key dates for the provider to work towards:

**Start of Maintenance Mode: July 31, 2024**

Effectively this means that feature development on existing resources will require migration to v2 of the provider, as new features will not be added to that version of the SDK.

**End of Support: July 31, 2025**

By this time the provider should not have any functionality dependent on V1 of the SDK.

The maintainer team has [stopped](https://github.com/hashicorp/terraform-provider-aws/issues/32917) accepting net-new resources/data sources based on [aws-sdk-go](https://github.com/aws/aws-sdk-go).

### References

- #32917
- [Announcing end-of-support for AWS SDK for Go (v1) effective July 31, 2025](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/)
- https://aws.github.io/aws-sdk-go-v2/docs/migrating/

### Final Items to Complete to Mark As Complete

- [x] #41775
- [x] Remove SDK v1 Session and Endpoints (`internal/conns/awsclient.go`)
- [ ] `.ci/scripts/update-aws-sdk-go-dependencies.sh` #42098
- [ ] Skaff data source template (`skaff/datasource/datasourcefw.gtpl`) #42098
- [ ] Docs: `docs/dependency-updates.md` #42098
- [x] ~Test fixtures for CloudFormation~ (`internal/service/cloudformation/test-fixtures/examplecompany-exampleservice-exampleresource`) (not possible, indirect, whole test-fixture needs updating)

### Services Requiring Migration

- [x] #36103 
- [x] #36110
- [x] #36111 
- [x] #36112
- [x] #36113
- [x] #36114
- [x] #36115
[...]

Issue #32917, opened 2023-08-08
Policy Change: Net-new resources/data sources must be based on aws-sdk-go-v2 / terraform-plugin-framework


breathingdust:
### Description

In order to take advantage of automation opportunities presented by the [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2), and to reduce the long term upgrade burden we will no longer accept net-new resources or data sources based on [aws-sdk-go](https://github.com/aws/aws-sdk-go).

In a similar vein to take advantage of the improved developer experience of the [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework) we will no longer accept net-new resources or data sources based on [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk).

The date of this change will be dictated by the following being completed:

- [ ] All examples and guidance in contributor focused documentation should be focused on [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2) / [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework). We should keep older examples in order to facilitate best practice for existing resources. 
- [ ] Our templating tool ([skaff](https://github.com/hashicorp/terraform-provider-aws/tree/main/skaff)) should default to using [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2) / [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework)
- [ ] A GitHub action or check to comment/perform check on any PR containing a new datasource/resource explaining that the use of [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2) / [terraform-plugin-framework](https://github.com/hashicorp/terraform-plugin-framework) is required.
- [ ] Messaging to indicate that the priority of existing PRs based on [aws-sdk-go](https://github.com/aws/aws-sdk-go)/[terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk) is not affected by their use. While the maintainers appreciate the work to update them to [aws-sdk-go-v2](https://github.com/aws/aws-sdk-go-v2), we are happy to update them if they are not.

PR #33022, opened 2023-08-15
[Docs]: Config and example content for content tabs

breathingdust:
### Description

To help with the transition to requiring terraform-plugin-framework / aws-sdk-go-v2 all examples in the contributor documentation should lead with the preferred format. We should however preserve legacy examples to help contributors with existing resources. In order to present these various formats in a way that doesn't make the documentation hard to navigate we could use content tabs. This PR includes the mkdocs configuration to allow the use of content tabs, as well as some sample tabs.

### References

- #32917
- #32976

### DIFF
--- a/docs/add-a-new-service.md
+++ b/docs/add-a-new-service.md
@@ -51,31 +51,116 @@ If an AWS service must be created in a non-standard way, for example the service
 
 1. Add a file `internal/<service>/service_package.go` that contains an API client factory function, for example:
 
-```go
-package globalaccelerator
+=== "framework / aws-go-sdk-v2"
 
-import (
-	"context"
+    ```go
+    package globalaccelerator
+    
+    import (
+        "context"
+    
+        aws_sdkv1 "github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


docs/add-a-new-service.md
```diff
@@ -51,62 +51,98 @@ If an AWS service must be created in a non-standard way, for example the service
...
-
-	// Force "global" services to correct Regions.
-	if p.config["partition"].(string) == endpoints_sdkv1.AwsPartitionID {
-		config.Region = aws_sdkv1.String(endpoints_sdkv1.UsWest2RegionID)
-	}
-
-	return globalaccelerator_sdkv1.New(sess.Copy(config)), nil
-}
-```
+<!-- markdownlint-disable code-block-style -->
```
jar-b, line 54: I opted to disable this linter check inline rather than globally for now as the code block checking is valuable for the website docs. If this becomes too burdensome for the contributor documentation we can extend the root markdownlint configuration with a second file stored in the `docs/` subfolder and disable it globally there instead.

jar-b (approved):

PR #33056, opened 2023-08-16
teamcity: failure condition using regex

johnsonaj:
### Description

Check for string in build logs without case sensitivity

```console

...
```


jar-b (approved):
LGTM 🎸

PR #33241, opened 2023-08-30
Align doc file endings to be consistent with .html.markdown standard

breathingdust:
### Description

Correcting a few doc files which do not conform to the `.html.markdown` file extension standard that the majority of the doc files align too. This is conflicting with some in progress tooling development. The CDK generated docs are `.html.md`, but at present this does not interfere with that tooling.

Ideally we have a check to prevent new files being added with non-standard extensions.


justinretzolk (approved):
LGTM 🚀

PR #33509, opened 2023-09-18
Add MediaConvert endpoints data source

luqasz:
### Description
New MediaConvert endpoints datasource.

I need some help in this manner. I don't know how to mock api responses.


dimaman2001:
MediaConvert endpoints are designed to be obtained once and then hardcoded. The `DescribeEndpoints` action can only be invoked [once per minute](https://docs.aws.amazon.com/general/latest/gr/mediaconvert.html). This Data Source may not be a good idea.

PR #33567, opened 2023-09-21
Don't return a retry.NotFoundError for servicequotas.ErrCodeDependencyAccessDeniedException

rdark:
### Description

When the applying role is missing a permission (such as `rds:DescribeAccountAttributes`), the API returns a `DEPENDENCY_ACCESS_DENIED_ERROR` (with status code 200/non-actual-error). The resulting error from `findServiceQuotaByID` is currently a `retry.NotFoundError` which results in the error being silently swallowed by [this block](https://github.com/hashicorp/terraform-provider-aws/blob/8963cdd345854ade46f6bae056b69f441c879a2e/internal/service/servicequotas/service_quota.go#L215-L217)

Looking at the docs:

https://docs.aws.amazon.com/servicequotas/2019-06-24/apireference/API_ErrorReason.html

`DEPENDENCY_ACCESS_DENIED_ERROR` seems to be the only one that is not retry-able - but I can't see that any retry logic is actually being done for this.

Current behaviour seems to be on any of the above errors, the error is silently swallowed, and the default service quota values are used, which can result in inconsistent state.


internal/service/servicequotas/find.go
```diff
@@ -85,10 +85,7 @@ func findServiceQuotaByID(ctx context.Context, conn *servicequotas.ServiceQuotas
 	}
 
 	if output.Quota.ErrorReason != nil {
-		return nil, &retry.NotFoundError{
-			Message:     fmt.Sprintf("%s: %s", aws.StringValue(output.Quota.ErrorReason.ErrorCode), aws.StringValue(output.Quota.ErrorReason.ErrorMessage)),
-			LastRequest: input,
-		}
+		return nil, fmt.Errorf("%s: %s", aws.StringValue(output.Quota.ErrorReason.ErrorCode), aws.StringValue(output.Quota.ErrorReason.ErrorMessage))
```
acaban-nuharbor, line 88: In case a retry is ever implemented in the future, consider an if-statement to return `Errorf` only for `DEPENDENCY_ACCESS_DENIED_ERROR`.

rdark: Thanks for the feedback!

I'd argue with the current lack of retry implementation, it's better just to fail on a temporary server error. Otherwise that scenario can result in erroneous resource creation.

PR #33581, opened 2023-09-22
Update port description in vpclattice_target_group_attachment docs

posquit0:
### Description

- `port` doesn't use default port by `protocol`. The default port is inherited from the parent target group.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@posquit0 Thanks for the contribution :tada: :clap:.

PR #33590, opened 2023-09-22
docs(route53-dnssec): Improve documentation for route53 dnssec resources

essjayhch:
# Description
The documentation for the route53_hosted_zone_dnssec resource doesn't explicitly explain why the aws provider is defined with the us-east-1 region. This is not arbitrary, it is done as a side effect of where the signing key's CMK needs to be located.

# Improvements
Adds an info block to slightly more explicitly explain the requirement. When generating these through the console, you never get exposed to this requirement until such time as you provision the resources through terraform.

This is purely a documentation adjustment, hence there is no acceptance testing of any underlying provider objects.


ewbankkit (approved):
LGTM :rocket:.

Issue #33588, opened 2023-09-22
[Bug]: new provider version not published on terraform registry


dhohengassner:
### Terraform Core Version

1.5.7

### AWS Provider Version

5.17.0

### Affected Resource(s)

It seems like the provider version 5.18.0 containing some important S3 fixes is not published yet on terraform registry.

https://registry.terraform.io/providers/hashicorp/aws/latest

Did there something happen during the sync/upload?

Thanks a lot!

### Expected Behavior

latest available version of provider can be used

### Actual Behavior

version 5.18.0 cannot be used since it is not completely released yet

### Relevant Error/Panic Output Snippet

```shell
│ Could not retrieve the list of available versions for provider
│ hashicorp/aws: no available releases match the given constraints >= 5.0.0,
│ 5.18.0
```

### Terraform Configuration Files

```shell
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
     version = "5.18.0"
    }
  }
}
```

### Steps to Reproduce

terraform init

### Would you like to implement a fix?

None

ewbankkit:
Hmm. The release GHA was successful: https://github.com/hashicorp/terraform-provider-aws/actions/runs/6267061933, and the binaries are in https://releases.hashicorp.com/terraform-provider-aws/5.18.0/.
We'll investigate.
@dhohengassner Thanks for bringing this to our attention 👏.

breathingdust:
Just an update that the registry team is actively investigating. Hopefully we'll have a resolution for you soon.

vampirechicken:
Need the S3 fixes in 5.18. Any updates?

el-chazmo:
> Need the S3 fixes in 5.18. Any updates?

Until this is resolved, you could always just block version v5.17 for now
```terraform
required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 4.0, != 5.17.0"
    }
  }
```

vampirechicken:
> > Need the S3 fixes in 5.18. Any updates?
> 
> Until this is resolved, you could always just block version v5.17 for now
> 
> ```terraform
> required_providers {
>     aws = {
>       source  = "hashicorp/aws"
>       version = ">= 4.0, != 5.17.0"
>     }
>   }
> ```

[...]

PR #33629, opened 2023-09-26
doc(cdktf): remove duplicate .html.md extensions

jar-b:
### Description

#33501 added CDKTF guides with an `html.markdown` extension, leaving files of the same name with an `.html.md` extension in the same directory. This has caused downstream issues when publishing releases.

### Relations

Relates #33501 
Relates #33588 

N/a


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
Need to re-release `v5.18.0` with this commit (but no others 😄).

PR #31232, opened 2023-05-05
rds/instance: NEW ID configuration

YakDriver:
### Relations

Closes #16782
Closes #507

### Output from Acceptance Testing

(Failures unrelated to these changes.)

```console
% make t T=TestAccRDSInstance_ K=rds P=5 ACCTEST_TIMEOUT=780m
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/rds/... -v -count 1 -parallel 5 -run='TestAccRDSInstance_'  -timeout 780m
=== RUN   TestAccRDSInstance_basic
=== PAUSE TestAccRDSInstance_basic
=== RUN   TestAccRDSInstance_manage_password
=== PAUSE TestAccRDSInstance_manage_password
=== RUN   TestAccRDSInstance_identifierPrefix
=== PAUSE TestAccRDSInstance_identifierPrefix
=== RUN   TestAccRDSInstance_identifierGenerated
=== PAUSE TestAccRDSInstance_identifierGenerated
=== RUN   TestAccRDSInstance_disappears
=== PAUSE TestAccRDSInstance_disappears
=== RUN   TestAccRDSInstance_tags
=== PAUSE TestAccRDSInstance_tags
=== RUN   TestAccRDSInstance_onlyMajorVersion
=== PAUSE TestAccRDSInstance_onlyMajorVersion
=== RUN   TestAccRDSInstance_kmsKey
=== PAUSE TestAccRDSInstance_kmsKey
=== RUN   TestAccRDSInstance_customIAMInstanceProfile
=== PAUSE TestAccRDSInstance_customIAMInstanceProfile
=== RUN   TestAccRDSInstance_subnetGroup
=== PAUSE TestAccRDSInstance_subnetGroup
=== RUN   TestAccRDSInstance_networkType
=== PAUSE TestAccRDSInstance_networkType
=== RUN   TestAccRDSInstance_optionGroup
=== PAUSE TestAccRDSInstance_optionGroup
=== RUN   TestAccRDSInstance_iamAuth
=== PAUSE TestAccRDSInstance_iamAuth
=== RUN   TestAccRDSInstance_allowMajorVersionUpgrade
=== PAUSE TestAccRDSInstance_allowMajorVersionUpgrade
=== RUN   TestAccRDSInstance_dbSubnetGroupName
=== PAUSE TestAccRDSInstance_dbSubnetGroupName
=== RUN   TestAccRDSInstance_DBSubnetGroupName_ramShared
[...]

PR #33640, opened 2023-09-27
Update db_instance.html.markdown

jrsherry:
Add 'identifier' attribute to docs.

Was added in code here: 
[NEW ID configuration](https://github.com/hashicorp/terraform-provider-aws/pull/31232/files#diff-e48aedc61af35e9ec060ba3ba12c9601ceb9d93d5dbf5ced12f882d7b65edb4dR386)

### DIFF
--- a/website/docs/r/db_instance.html.markdown
+++ b/website/docs/r/db_instance.html.markdown
@@ -439,6 +439,7 @@ DB instance.
 * `hosted_zone_id` - The canonical hosted zone ID of the DB instance (to be used
 in a Route 53 Alias record).
 * `id` - RDS DBI resource ID.
+* `identifier` - The RDS instance ID.
 * `instance_class`- The RDS instance class.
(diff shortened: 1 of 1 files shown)


gdavison:
Thanks for submitting this, @jrsherry.

The "Attribute Reference" section is used only for parameters that are exported by the resource, and cannot be set in the configuration. The parameter `identifier` can be set, and is documented at https://github.com/hashicorp/terraform-provider-aws/blob/2828878f7edf98dfa60eb6672bc69f3a702f36f4/website/docs/r/db_instance.html.markdown?plain=1#L281

We appreciate you contributing to updating the provider's documentation.

jrsherry:
@gdavison, please re-open this PR.  'identifier' is both an attribute and an argument.

[resource](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/rds/instance.go#L1673)
[example](https://github.com/hashicorp/terraform-provider-aws/blob/main/examples/rds/outputs.tf#L9)

PR #33663, opened 2023-09-27
sweeper/aws_cloudfront_distribution: Fixes infinite loop

gdavison:
### Description

The `aws_cloudfront_distribution` sweeper has an infinite loop when there are no CloudFront Distributions


johnsonaj (approved):
LGTM 🚀

PR #33664, opened 2023-09-27
D-fix documentation typos and grammars mistakes

akhateeb22:


gdavison (approved):
HI @akhateeb22, thanks for your contribution. I've made some adjustments that better preserve the meaning of the original documentation

Issue #33639, opened 2023-09-26
[Enhancement]: Support region for target_table argument in aws_glue_catalog_table


raman-aro:
### Description

Per the following issue, can region be added to the target_table argument for aws_glue_catalog_table as well?
https://github.com/hashicorp/terraform-provider-aws/issues/32099#issue-1764919771

The fix added it to aws_glue_catalog_database but this doesn't address the case when creating resource links to specific tables rather than whole databases. 

### Affected Resource(s) and/or Data Source(s)

aws_glue_catalog_table

### Potential Terraform Configuration

```terraform
resource "aws_lakeformation_permissions" "lf_grant_on_target" {

  principal = local.arn
  permissions = ["SELECT"]

  table {
    region = "us-east-2"
    catalog_id = local.catalog
    database_name = local.database
    name = local.table
  }
}
```

### References

https://github.com/hashicorp/terraform-provider-aws/issues/32099

### Would you like to implement a fix?

No

autotune:
I am working on this.

autotune:
While this is available in the AWS docs: https://docs.aws.amazon.com/glue/latest/webapi/API_TableIdentifier.html#Glue-Type-TableIdentifier-Region

This is not a supported argument in the go sdk: https://github.com/aws/aws-sdk-go/blob/737322b8f8b54fc78d076334d1e433e2e5f2993d/service/lakeformation/api.go#L13775

When attempting to add it in I get an idempotent error where it constantly tries to force a new replacement. I'll submit what I have so far as a WIP but I'm not sure that this is, in fact, supported.

Issue #32099, opened 2023-06-20
[Enhancement]: support for AWS Lake Formation and Glue Data Catalog cross-Region table access


rtoma-btg:
### Description

AWS has released a feature to support cross-region Glue table access. Related blog / docs:
- https://aws.amazon.com/about-aws/whats-new/2023/06/aws-lake-formation-glue-data-catalog-cross-region-access/
- https://docs.aws.amazon.com/lake-formation/latest/dg/data-access-across-region.html
- https://docs.aws.amazon.com/lake-formation/latest/dg/setup-cross-region-access.html

Please add support for this feature in the Terraform AWS provider.

### Affected Resource(s) and/or Data Source(s)

* aws_glue_catalog_table

### Potential Terraform Configuration

As of June 20th the API docs for `glue:CreateTable` https://docs.aws.amazon.com/glue/latest/webapi/API_CreateTable.html has not yet been updated, but docs at https://docs.aws.amazon.com/lake-formation/latest/dg/create-resource-link-table.html mention how to do it using the AWS CLI, so we can assume the API is there.

```
$ aws glue create-table —region eu-west-1 —cli-input-json '{
    "CatalogId": "111122223333",
    "DatabaseName": "ireland_db",
    "TableInput": {
        "Name": "rl_useast1salestb_ireland",
        "TargetTable": {
            "CatalogId": "444455556666",
            "DatabaseName": "useast1_salesdb",
            "Region": "us-east-1", <------------------
            "Name":"useast1_salestb"
        }
    }
}‘
```

When I map this to the current Terraform resource I believe the `target_table` parameter (see https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/glue_catalog_table#target_table), needs to be extended with a `region`.

### Would you like to implement a fix?

No

DrFaust92:
looking into this

PR #33687, opened 2023-09-29
Add region to lakeformation table permission grant (WIP)

autotune:
> ### Description
> 
> Per the following issue, can region be added to the target_table argument for aws_glue_catalog_table as well? [#32099 (comment)](https://github.com/hashicorp/terraform-provider-aws/issues/32099#issue-1764919771)
> 
> The fix added it to aws_glue_catalog_database but this doesn't address the case when creating resource links to specific tables rather than whole databases.
> ### Affected Resource(s) and/or Data Source(s)
> 
> aws_glue_catalog_table
> ### Potential Terraform Configuration
> 
> ```terraform
> resource "aws_lakeformation_permissions" "lf_grant_on_target" {
> 
>   principal = local.arn
>   permissions = ["SELECT"]
> 
>   table {
>     region = "us-east-2"
>     catalog_id = local.catalog
>     database_name = local.database
>     name = local.table
>   }
> }
> ```
> 
> ### References
> 
> #33639
> ### Would you like to implement a fix?
> 
> Yes

### Relations
<!---
If your pull request fully resolves and should automatically close the linked issue, use Closes. Otherwise, use Relates.

For Example:

Relates #33639

### References

This does not appear to be a supported identifier in the go sdk struct: https://github.com/aws/aws-sdk-go/blob/737322b8f8b54fc78d076334d1e433e2e5f2993d/service/lakeformation/api.go#L13775

When attempting to add it in I get an idempotent error where it constantly tries to force a new replacement. I'll submit what I have so far as a WIP but I'm not sure that this is, in fact, supported.

```
    permissions_test.go:504: Step 1/1 error: After applying this test step, the plan was not empty.
        stdout:


        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
        -/+ destroy and then create replacement

          # aws_lakeformation_permissions.test must be replaced
        -/+ resource "aws_lakeformation_permissions" "test" {
              ~ id                            = "2476167902" -> (known after apply)

              ~ table {
                  ~ catalog_id    = "504376484015" -> (known after apply)
                    name          = "tf-acc-test-537905478652321360"
                  + region        = "us-east-2" # forces replacement
                    # (2 unchanged attributes hidden)
                }
            }

        Plan: 1 to add, 0 to change, 1 to destroy.
```

### DIFF
--- a/internal/service/lakeformation/permissions.go
+++ b/internal/service/lakeformation/permissions.go
@@ -293,6 +293,11 @@ func ResourcePermissions() *schema.Resource {
 								"table.0.wildcard",
 							},
 						},
+						"region": {
+							Type:     schema.TypeString,
+							ForceNew: true,
+							Optional: true,
+						},
 					},
 				},
 			},
@@ -361,6 +366,12 @@ func ResourcePermissions() *schema.Resource {
 								"table_with_columns.0.column_names",
 								"table_with_columns.0.wildcard",
 							},
+						},
+						"region": {
+							Type:     schema.TypeString,
+							ForceNew: true,
+							Optional: true,
+							Computed: true,
 						},
 					},
 				},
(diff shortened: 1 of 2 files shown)


bryantbiggs:
this is not in the Lakeformation grantPermissions API and therefore wouldn't be valid to support

bryantbiggs:
this would need to be added to the glue catalog table, not the permissions grant https://github.com/hashicorp/terraform-provider-aws/blob/07b42a52a5864e79c2b8376bdecd109d2b541256/internal/service/glue/catalog_table.go#L326

PR #33770, opened 2023-10-05
skaff resource test template is missing errors stdlib import.

gbw:
### Description
The resource test template makes use of errors in multiple places, so `errors` should be added by default in the import list.

### Relations
N/A

### References
N/A

N/A


jar-b (approved):
LGTM 👍

jar-b:
Nice catch - thanks for your contribution, @gbw!

PR #33927, opened 2023-10-13
Replace the values' data type from string to number in the document

shusann01116:
### Description

Replaced the `build_timeout` and `queued_timeout` values' data type from a string to a number in the document because it is expected to be a Int in the [code](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/codebuild/project.go#L627-L632).
While it is not illegal to use a string value, it should be of the number data type.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@shusann01116 Thanks for the contribution :tada: :clap:.

PR #34041, opened 2023-10-20
ci: Don't allow multiple sweeper, performance, or sanity builds in queue

gdavison:
### Description

Because full acceptance test runs take a significant amount of time, multiple daily sweeper, performance, and sanity builds can be queued. Allow older runs to be replaced if they are queued


johnsonaj (approved):
LGTM 🚀

PR #34089, opened 2023-10-24
ci: fix diff grep regex

jar-b:
### Description

Adjusts the grep regular expression so that only added import lines will match. Previously unmodified lines which are displayed with the diff for context could also match the provided expression. Now the line must start with a literal +, contain only whitespace prior to a double quote, and match the root path of the AWS SDK for Go v1 library.

Also fixes the `body-includes` argument in the check for existing comments so that the comment is not repeated on subsequent commits.

### Relations

Relates #34083


ewbankkit (approved):
LGTM :rocket:.

Issue #34072, opened 2023-10-23
[Docs]: aws_db_snapshot documentation formatting error, not sure if parameter is supported


lisakoivu:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/db_snapshot

### Description

The documentation states that a "tags" parameter is supported, optional, but it's formatted incorrectly, without a bullet. 
<img width="549" alt="image" src="https://github.com/hashicorp/terraform-provider-aws/assets/43785335/6353531d-775f-4d1f-ab72-1151b977a9b5">

That being said, I can't use tags as a filter. Is it supported, or not? 

Thank you

```
data "aws_db_snapshot" "clone" {
  db_instance_identifier = "ylvoprd"
  snapshot_type = "manual"
  tags = {
    key   = "tag:data-systems:golden-snapshot-date"
    value = "20231022"
  }
}

...

│ Error: Unsupported argument
│ 
│   on main.tf line 48, in data "aws_db_snapshot" "clone":
│   48:   tags = {
│ 
│ An argument named "tags" is not expected here.
```

### Would you like to implement a fix?

None

lisakoivu:
I realized I am using version 4.x of the provider, which does not support this functionality.

stefanfreitag:
Hi @justinretzolk , I was looking  for some work in the documentation space - this issue seems to be resolved. Maybe the issue can be set from open to closed.

PR #34115, opened 2023-10-25
Update db_snapshot.html.markdown

network-charles:
### Original Sentence
- include_public - (Optional) Set this value to true to include manual DB snapshots that are public and can be copied or restored by any AWS account, otherwise set this value to false. The default is false. tags - (Optional) Mapping of tags, each pair of which must exactly match a pair on the desired DB snapshot.

### Issues
Incorrect formatting of the `tags` argument reference.

### New Sentence
- include_public - (Optional) Set this value to true to include manual DB snapshots that are public and can be copied or restored by any AWS account, otherwise set this value to false. The default is false. 
- tags - (Optional) Mapping of tags, each pair of which must exactly match a pair on the desired DB snapshot.

This PR properly formats the argument reference, hence, improving readability.

### Relations
Closes #0000

### References
https://github.com/hashicorp/terraform-provider-aws/issues/34072

### DIFF
--- a/website/docs/d/db_snapshot.html.markdown
+++ b/website/docs/d/db_snapshot.html.markdown
@@ -63,7 +63,7 @@ AWS accounts that this AWS account has been given permission to copy or restore,
 The default is `false`.
 * `include_public` - (Optional) Set this value to true to include manual DB snapshots that are public and can be
 copied or restored by any AWS account, otherwise set this value to false. The default is `false`.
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@network-charles Thanks for the contribution :tada: :clap:.

PR #34184, opened 2023-10-31
ci: switch library_versions action to pull_request_target event

jar-b:
### Description

This change should allow the Github comment job to have the necessary write permissions, even when pull requests are opened from forks.

### Relations

Relates #34083

### References

- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

PR #34207, opened 2023-11-01
ci: pass repository argument during head checkout

jar-b:
### Description

This change should allow the checkout of the head ref to function correctly for pull requests coming from forks.

### Relations

Relates #34184


johnsonaj (approved):
LGTM 🚀

PR #34272, opened 2023-11-06
docs: Mark service_role as optional for aws_batch_compute_environment

### LINKED ISSUE #34143: [Docs]: service_role should not be required for aws_batch_compute_environment
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/batch_compute_environment

### Description

The provider documentation states that the `service_role` argument to `aws_batch_compute_environment is required`: https://github.com/hashicorp/terraform-provider-aws/blob/main/website/docs/r/batch_compute_environment.html.markdown?plain=1#L164

The provider does not actually require the argument to be given and works perfectly well without it; the AWS API method `CreateComputeEnvironment` documentation also states that the service role argument is optional, in the case of which AWS creates a service-linked role automatically.

### References

https://docs.aws.amazon.com/batch/latest/APIReference/API_CreateComputeEnvironment.html#Batch-CreateComputeEnvironment-request-serviceRole

None

acwwat:
### Description

This PR is to fix the `service_role` argument in the `aws_batch_compute_environment` resource documentation, which is optional instead of required.

### Relations

Closes #34143

### References

n/a

n/a


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@acwwat  Thanks for the contribution :tada: :clap:.

PR #34313, opened 2023-11-08
deps: ignore aws-sdk-go-v2/config

jar-b:
### Description

See https://github.com/hashicorp/terraform-provider-aws/pull/34300#issuecomment-1801819150

### Relations

Relates #34300
Relates #34307

PR #34300, opened 2023-11-07
Rolls back AWS SDK modules with shared config parsing bug

gdavison:
### Description

Rolls back `github.com/aws/aws-sdk-go-v2/config` and `github.com/aws/aws-sdk-go-v2/internal/ini` versions which contain shared config file parsing bug https://github.com/aws/aws-sdk-go-v2/issues/2349

Adds exclusions to `go.mod` file and adds test for this regression

Relates #34234.
Closes #34257.

### Output from Acceptance Testing

```console
% go test -v ./internal/provider/... -run=TestSharedConfigFileParsing

--- PASS: TestSharedConfigFileParsing (0.08s)
    --- PASS: TestSharedConfigFileParsing/leading_whitespace (0.08s)
```

PR #34327, opened 2023-11-09
deps: add feature/s3/manager to temporary ignore list

jar-b:
### Description

This will prevent the need for manual intervention on automated dependabot updates until the upstream `config` package regression is fixed. Also added some additional context and links to the comments of this ignore section.

### Relations

Relates #34300
Relates #34313


ewbankkit (approved):
LGTM :rocket:.

It looks like a fix for the breaking INI file change is coming, https://github.com/aws/aws-sdk-go-v2/pull/2359, but we shouldn't wait for it.

PR #33745, opened 2023-10-04
F finspace kx cluster cache type and size changes

kavan-aws:
### Description

This PR adds support for new cache type enums CACHE_12 and CACHE_250. New acceptance tests that create clusters with the 2 new enums are included. This PR also adds support for larger cache sizes. Further validation on both cache size and type is still done after the API call is made.

### Relations

Relates #31806

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccFinSpaceKxCluster_cache12Configurations PKG=finspace
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/finspace/... -v -count 1 -parallel 20 -run='TestAccFinSpaceKxCluster_cache12Configurations'  -timeout 360m
=== RUN   TestAccFinSpaceKxCluster_cache12Configurations
=== PAUSE TestAccFinSpaceKxCluster_cache12Configurations
=== CONT  TestAccFinSpaceKxCluster_cache12Configurations
--- PASS: TestAccFinSpaceKxCluster_cache12Configurations (2873.13s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/finspace	2901.285s

% make testacc TESTS=TestAccFinSpaceKxCluster_cache250Configurations PKG=finspace
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/finspace/... -v -count 1 -parallel 20 -run='TestAccFinSpaceKxCluster_cache250Configurations'  -timeout 360m
=== RUN   TestAccFinSpaceKxCluster_cache250Configurations
=== PAUSE TestAccFinSpaceKxCluster_cache250Configurations
=== CONT  TestAccFinSpaceKxCluster_cache250Configurations
--- PASS: TestAccFinSpaceKxCluster_cache250Configurations (2641.30s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/finspace	3471.082s

# Tested with larger cache size (36000)
% make testacc TESTS=TestAccFinSpaceKxCluster_cacheConfigurations PKG=finspace
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/finspace/... -v -count 1 -parallel 20 -run='TestAccFinSpaceKxCluster_cacheConfigurations'  -timeout 360m
=== RUN   TestAccFinSpaceKxCluster_cacheConfigurations
=== PAUSE TestAccFinSpaceKxCluster_cacheConfigurations
[...]

PR #34398, opened 2023-11-14
finspace: Increase kx_cluster create/update and kx_environment delete timeouts

kavan-aws:
### Description

Increases create and update default timeouts of kx_cluster resource to 18 hours to prepare for upcoming changes involving cache warming.

Also increase delete default timeout of kx_environment resource to 75 minutes to prepare for bug fix coming to deletion workflow when environment has deleting clusters (as in acceptance tests).

### Relations

Relates #33745 and #31806

Fix on cluster/environment deletion still in progress for kx_cluster tests, but ran TestAccFinSpaceKxCluster_code after making timeout changes and verified that create and update both worked as expected, along with their checks. Will rerun all cluster tests once deletion bug is fixed.

### DIFF
--- a/.changelog/34398.txt
+++ b/.changelog/34398.txt
@@ -1,0 +1,6 @@
+```release-note:enhancement
+resource/aws_finspace_kx_cluster: Increase default create and update timeouts to 4 hours to allow for increased startup times with large volumes of cached data
+```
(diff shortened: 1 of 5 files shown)


jar-b (approved):
LGTM 🎉

PR #34454, opened 2023-11-17
add note about what to do with existing resources

lpmi-13:
### Description

I was very confused when attempting to import an existing resource, since there was no Import section [in the docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sns_sms_preferences).

So I just tried an apply, and it seemed to work. This PR adds a note to the documentation if/when somebody else runs into a similar situation and is confused that the resource docs don't say anything about importing.

### References
[docs page that led to this PR](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sns_sms_preferences)

### DIFF
--- a/website/docs/r/sns_sms_preferences.html.markdown
+++ b/website/docs/r/sns_sms_preferences.html.markdown
@@ -30,3 +30,7 @@ This resource supports the following arguments:
 ## Attribute Reference
 
 This resource exports no additional attributes.
+
+## Import
+
+You cannot import the SMS preferences.


lpmi-13:
I see CI is upset with the format of the section

```
Error checking Terraform Provider documentation: 1 error occurred:
	* website/docs/r/sns_sms_preferences.html.markdown: error checking file contents: import section should conclude with ". For example:" (or state "You cannot import ...")
```

I'll update that.

lpmi-13:
The error messsage for this workflow is a bit confusing, since I assume by `import section should conclude with` it really means that the section can't contain anything else (ie, a helpful comment about what to do if you have existing state).

Nevertheless, a section saying the resource doesn't support importing is better than nothing, so I'll update the commit to just be that.

justinretzolk (approved):
Looks good to me! Thank you very much for your contribution! 🚀

PR #34620, opened 2023-11-29
tech debt: Collect diags instead of returning directly - `b` and `c` services

gdavison:
### Description

Covers services starting with `b` and `c`


ewbankkit (approved):
LGTM :rocket:.

PR #34460, opened 2023-11-17
r/aws_elasticache_replication_group: add auth_token_update_strategy argument

jar-b:
### Description

This optional argument will have a default value of `ROTATE`, as this was the previously hardcoded value sent during modify requests in which the `auth_token` was changed. Practitioners will now have the ability to explicitly use the `SET` and `DELETE` strategies as well.

### Relations

Closes #11524
Closes #33439

### References

- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ModifyReplicationGroup.html
- https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth.html

### Output from Acceptance Testing

```console
% make testacc PKG=elasticache TESTS="TestAccElastiCacheReplicationGroup_"
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/elasticache/... -v -count 1 -parallel 20 -run='TestAccElastiCacheReplicationGroup_'  -timeout 360m

--- PASS: TestAccElastiCacheReplicationGroup_clusteringAndCacheNodesCausesError (37.89s)
=== CONT  TestAccElastiCacheReplicationGroup_Engine_Redis_LogDeliveryConfigurations_ClusterMode_Disabled
--- PASS: TestAccElastiCacheReplicationGroup_Validation_noNodeType (60.51s)
=== CONT  TestAccElastiCacheReplicationGroup_dataTiering
--- PASS: TestAccElastiCacheReplicationGroup_basic (947.34s)
=== CONT  TestAccElastiCacheReplicationGroup_GlobalReplicationGroupIDClusterModeValidation_numNodeGroupsOnSecondary
--- PASS: TestAccElastiCacheReplicationGroup_ClusterMode_singleNode (1004.69s)
=== CONT  TestAccElastiCacheReplicationGroup_GlobalReplicationGroupIDClusterMode_basic
--- PASS: TestAccElastiCacheReplicationGroup_enableSnapshotting (1032.43s)
=== CONT  TestAccElastiCacheReplicationGroup_GlobalReplicationGroupID_disappears
--- PASS: TestAccElastiCacheReplicationGroup_tags (1284.05s)
=== CONT  TestAccElastiCacheReplicationGroup_finalSnapshot
--- PASS: TestAccElastiCacheReplicationGroup_Validation_globalReplicationGroupIdAndNodeType (1292.54s)
=== CONT  TestAccElastiCacheReplicationGroup_autoMinorVersionUpgrade
[...]

PR #34633, opened 2023-11-29
r/aws_elasticache_replication_group(doc): clarify auth_token_update_strategy argument dependencies

jar-b:
### Description

While this value is only sent to the AWS API if auth_token is set (and by association, transit_encryption_enabled is set to true), the default value means it is always set in state. As such, indiciating in the documentation that it can only specified when transit encrpytion is enabled is no longer true.

### Relations

Relates #34460 
Originally suggested in this comment: https://github.com/hashicorp/terraform-provider-aws/pull/34460#issuecomment-1829317390


ewbankkit (approved):
LGTM :rocket:.

PR #34727, opened 2023-12-04
tech debt: Collect diags instead of returning directly - `e`, `f`, `g`, `i`, and `k` services

gdavison:
### Description

Covers services starting with `e` (other than `ec2`), `f`, `g`, `i`, and `k`


ewbankkit (approved):
LGTM :rocket:.

PR #34748, opened 2023-12-05
skaff: remove unintended newline in imports

jar-b:
This newline was causing the import linter (impi) to fail when using skaff for resources with AWS SDK for Go V2.


ewbankkit (approved):
LGTM :rocket:.

PR #34797, opened 2023-12-07
lookoutmetrics: add generated files

jar-b:
### Description

Adds generated files missed in #34793.


ewbankkit:
I think you'll need to run `make gen` at the top level.

PR #34837, opened 2023-12-09
tech debt: Collect diags instead of returning directly - `l`, `m`, and `n` services

gdavison:
### Description

Covers services starting with `l`, `m`, and `n`


ewbankkit (approved):
LGTM :rocket:.

PR #34903, opened 2023-12-13
Mention "cloudfront-js-2.0" now is an available runtime for CloudFront funcs

Tenzer:
### Description

This seems to have been a recent addition that hasn't been announced anywhere. It was mentioned in https://aws.amazon.com/blogs/aws/introducing-amazon-cloudfront-keyvaluestore-a-low-latency-datastore-for-cloudfront-functions/ but otherwise it just seems to have appeared in the documentation

I have updated the runtime version shown in the example as I assume most people would prefer to use the more recent runtime version.

### References

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/functions-javascript-runtime-20.html
https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_FunctionConfig.html

I have not run this since it's a documentation-only change.

### DIFF
--- a/website/docs/r/cloudfront_function.html.markdown
+++ b/website/docs/r/cloudfront_function.html.markdown
@@ -21,7 +21,7 @@ See [CloudFront Functions](https://docs.aws.amazon.com/AmazonCloudFront/latest/D
 ```terraform
 resource "aws_cloudfront_function" "test" {
   name    = "test"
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@Tenzer Thanks for the contribution :tada: :clap:.

PR #34914, opened 2023-12-13
Correct Slack notifier for regressions

justinretzolk:
### Description

This PR corrects the Slack notifier for regressions so that it works with titles, even if there's quotes in them.

N/a, Actions


jar-b (approved):
LGTM 🥇

PR #35001, opened 2023-12-20
Updates service data to have function calls instead of raw column numbers

gdavison:
Centralizes access to service data CSV.

Adds accessor functions to service records instead of exposing column numbers.


ewbankkit (approved):
LGTM :rocket:.

PR #35012, opened 2023-12-20
Glue Ray jobs fail to update

bobthemighty:
### Relations

Closes #0000

```console

...
```

### DIFF
--- a/internal/service/glue/job_test.go
+++ b/internal/service/glue/job_test.go
@@ -739,7 +739,7 @@ func TestAccGlueJob_rayJob(t *testing.T) {
 		CheckDestroy:             testAccCheckJobDestroy(ctx),
 		Steps: []resource.TestStep{
 			{
-				Config: testAccJobConfig_rayJob(rName),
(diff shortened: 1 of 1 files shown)


bobthemighty:
Any chance of a bump on this?

bobthemighty:
Friends! This bug makes the Ray job resource *useless* because it can't be updated through Terraform. The fix is here and tested, and has sat for 2 months. Is there any way we can get this prioritised? I could go ask a few dozen friends to add emojis if that helps?

nunezpaul:
bump on this

Geoiv:
Mirroring the above, we've found that destroying and recreating the relevant glue job allows the changes to be applied, but this is cumbersome and results in the connection between the job and previous run logs being lost. Would greatly like to see this implemented.

silveiraluiza:
Is anyone still looking into this? This is a great fix for an annoying issue with the ray jobs. I've been destroying and recreating them a lot lately...

ewbankkit:
@bobthemighty Thanks for the contribution :tada: :clap:.

PR #35088, opened 2023-12-28
Remove unnecessary depends_on+hard coded ID of dependency

### LINKED ISSUE #35084: [Docs]: Example usage for aws_route uses a hard-coded RTB ID with an explicit depends_on on a(n unmentioned) route table resource
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route

### Description

To the best of my understanding an explicit `depends_on` should be a last resort used to express dependencies that aren't implied, e.g. not being able to allocate an EIP without first having an IGW. In this case, when the resource being referenced is available, the id should be accessed as a reference rather than being given as a string literal. Alternatively, if the route table object is out of scope of the example, it’s not necessarily bad to use the string literal (`rtb-4fbb3ac4`), but in that case the explicit dependency should be eliminated.

### References

_No response_

Yes

Michagogo:
### Description

The current example usage specifies a hard-coded route table ID, and then proceeds to use an explicit `depends_on` on the route table. Besides the fact that logically it doesn’t make sense since if it’s being deployed together you wouldn’t have the ID yet, `depends_on` shouldn’t be promoted in general as it can cause issues when used unnecessarily.

### Relations

Closes #35084

### DIFF
--- a/website/docs/r/route.html.markdown
+++ b/website/docs/r/route.html.markdown
@@ -18,10 +18,9 @@ Provides a resource to create a routing table entry (a route) in a VPC routing t
 
 ```terraform
 resource "aws_route" "r" {
-  route_table_id            = "rtb-4fbb3ac4"
+  route_table_id            = aws_route_table.testing.id
   destination_cidr_block    = "10.0.1.0/22"
   vpc_peering_connection_id = "pcx-45ff3dc1"
-  depends_on                = [aws_route_table.testing]
 }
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@Michagogo Thanks for the contribution :tada: :clap:.

PR #35206, opened 2024-01-09
Increase Scaling Group Timeouts to match the FinSpace API.

hiranimd:
### Description
Increasing the Timeout for Create and Update Scaling Group Operations to match the FinSpace API in order to support customer use-cases.


jar-b (approved):
LGTM 🎉

PR #35245, opened 2024-01-11
[New Resource] launchwizard_deployment

brakf:
### Description

Introduces the first resource for the new API of AWS LaunchWizard.
LaunchWizard provides a guided experience to deploy enterprise applications on EC2 (e.g. SAP, MS ISS, etc.).
The new API currently only supports SAP workloads.

The API requires specifications to be send as a JSON file. As specifications are different for each deployment pattern, I decided to use a key-value map for this. Long term, I would add the possibilty to set specifications in pattern-specific blocks.

There is no Update API.

As SAP is a licensed software, LaunchWizard requires that the installation media is provided in an S3 bucket. For full acceptance testing, the bucket needs to be provided via the environment variable LAUNCHWIZARD_SAP_INST_MEDIA_S3_URI. Basic acceptance testing works without.

### References
https://docs.aws.amazon.com/launchwizard/latest/APIReference/API_CreateDeployment.html
https://docs.aws.amazon.com/launchwizard/latest/APIReference/launch-wizard-specifications.html
https://docs.aws.amazon.com/launchwizard/

```console
% make testacc TESTS=TestAccLaunchWizardDeployment PKG=launchwizard
TF_ACC=1 go test ./internal/service/launchwizard/... -v -count 1 -parallel 20 -run='TestAccLaunchWizardDeployment'  -timeout 360m
=== RUN   TestAccLaunchWizardDeployment_basic
=== PAUSE TestAccLaunchWizardDeployment_basic
=== RUN   TestAccLaunchWizardDeployment_basicFullInstallation
=== PAUSE TestAccLaunchWizardDeployment_basicFullInstallation
=== RUN   TestAccLaunchWizardDeployment_disappears
=== PAUSE TestAccLaunchWizardDeployment_disappears
=== RUN   TestAccLaunchWizardDeployment_SkipDestroyAfterFailure
=== PAUSE TestAccLaunchWizardDeployment_SkipDestroyAfterFailure
=== CONT  TestAccLaunchWizardDeployment_basic
=== CONT  TestAccLaunchWizardDeployment_disappears
=== CONT  TestAccLaunchWizardDeployment_SkipDestroyAfterFailure
=== CONT  TestAccLaunchWizardDeployment_basicFullInstallation
--- PASS: TestAccLaunchWizardDeployment_basic (1204.23s)
--- PASS: TestAccLaunchWizardDeployment_disappears (1289.58s)
--- PASS: TestAccLaunchWizardDeployment_basicFullInstallation (2670.04s)
--- PASS: TestAccLaunchWizardDeployment_SkipDestroyAfterFailure (3113.62s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/launchwizard       3119.984s

```

I defined 4 cases. 1 is requires SAP installation files that need to be placed in an S3 bucket. However, it is optional as the other cases should cover the same basic functionality and behaviours. To run the fourth test case, place the installation media in an S3 bucket and provide the URI as environment parameter LAUNCHWIZARD_SAP_INST_MEDIA_S3_URI

One of the test cases fails intentionally in order to test a virtual attribute that defines whether the stack should be deleted after a failed deployment or not. Leaving it avaiblable often makes sense for troubleshooting.


brakf:
Dear Team (@ewbankkit),
I am struggling with how to handle the fact that AWS masks the response for passwords in the GetDeployment Reponse.  For normal usage this is not a big deal as I have a workaround to ignore the value coming from the API and overwrite it with the password that was stored in state.

However, I am running into problems duing Import of the resource. The workaround doesn't work. I need a solution to write the password value from the configuration into the state before the plan is computed.

Root cause is the fact that I use a key/value map for the specifications and the complete map requires replace. I need a solution to check a single value in that map (if it is set) and ensure a change to it doesn't cause a replace. I found  mapplanmodifier.RequiresReplaceIf but no active implementation and couldn't get it to run on my own...

(changing specification to a block with explicit values is an alternative but would introduce a lot of complexity as there are a lot of possible values and combinations. I'd rather offload those validations to the API. https://docs.aws.amazon.com/launchwizard/latest/APIReference/launch-wizard-specifications-sap.html)

brakf:
How to properly handle an ExpiredTokenException while waiting for the process to finish? I do have a workaround that ensures that the resource isn't tainted, but it feels "hacky". Depending on the deployment pattern, it might take a few hours for the deployment to finish.

I have thought about a solution similar to the validation of acm certificates (creating the deployment in one resource and waiting for it to finish with another)

brakf:
I did some final changes after finally realizing my timeout problems came from my local configuration. Adjusted the error handling and fixed the test cases. Updated the acceptance test output after the last commit.

brakf:
@justinretzolk ; any update on this or info about priorisation?

brakf:
Dear Team,
@justinretzolk , @ewbankkit

It has been almost half a year and I am growing a bit frustrated. Can you please look at this PR?

Thank you!

PR #35298, opened 2024-01-14
docs: Improve response models and params descs for aws_api_gateway_method_response

### LINKED ISSUE #31614: [Enhancement]: Display the argument type in the documentation
### Description

The documentation includes if the argument is required or optional, and a description, but not the actual type of the argument.
 It is sometimes unclear what exactly the type should be, and we need to resort to guessing / or searching through examples to understand how to pass the parameter. 
It also sometimes results in having to look into AWS cli documentation to understand what terraform is expecting.

---

Secondly, even if the argument type would already be very useful, it would also be very helpful to include in the argument description some details of how the value should be provided, when it is a complex value, mostly for `string` and `map(string)` (or at least an example with that argument used).

It is particularly confusing for API Gateway resources, as they often contain mappings that are implemented as maps of strings, but the actual structure of the map is not specified. 
For instance, `api_gateway_method_response.response_models` has no explainations on what can be the key/values (should `value` be the model JSON ? Or the model name ? Or the model object ? Or something else ?).
However, `api_gateway_method_response.response_parameters` has more details (it explains that the key must be in the form `method.response.header.X-Some-Header`, even if there is no explanation on why the key must look like that, nor if we can specify `method.response.body.something`).

`api_gateway_method_response` exemples section does not contain any exemple of `response_models`. In order to use it, I had to look into AWS Cli documentation, to find the command to get an existing response_model and look at the response's payload to understand how to use it in terraform.  

All ressources 

```terraform

- rest_api_id - (Required) [string] ID of the associated REST API
- resource_id - (Required) [string] API resource ID
- http_method - (Required) [string] HTTP Method (GET, POST, PUT, DELETE, HEAD, OPTIONS, ANY)
- status_code - (Required) [integer] HTTP status code
- response_models - (Optional) [map(string)] Map of the API models used for the response's content type (TODO: add some details here on the format of the keys/values of the map)
- response_parameters - (Optional) [map(string)] Map of response parameters that can be sent to the caller. For example: `response_parameters = { "method.response.header.X-Some-Header" = true }` would define that the header X-Some-Header can be provided on the response.
```

### References

[api_gateway_method_response](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_method_response#argument-reference)

None

acwwat:
### Description

This PR aims to improve the description of the `response_models` and `response_parameters` and provide a sample usage on how to use them.

### Relations

Closes #31614

### References

Used the [MethodResponse API data type documentation](https://docs.aws.amazon.com/apigateway/latest/api/API_MethodResponse.html#responseParameters) for reference on wordings.

n/a

### DIFF
--- a/website/docs/r/api_gateway_method_response.html.markdown
+++ b/website/docs/r/api_gateway_method_response.html.markdown
@@ -8,9 +8,11 @@ description: |-
 
 # Resource: aws_api_gateway_method_response
 
-Provides an HTTP Method Response for an API Gateway Resource.
+Provides an HTTP Method Response for an API Gateway Resource. More information about API Gateway method responses can be found in the [Amazon API Gateway Developer Guide](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-method-settings-method-response.html).
 
 ## Example Usage
+
+### Basic Response
 
 ```terraform
 resource "aws_api_gateway_rest_api" "MyDemoAPI" {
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks great to me! Thank you very much for the contribution! 🚀

PR #35433, opened 2024-01-22
Update ecs_service.html.markdown

### LINKED ISSUE #35410: [Docs]: aws_ecs_service: force_new_deployment example doesn't work
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service

### Description

The example to "Redeploy Service On Every Apply" doesn't work because it uses the `timestamp()` function.
```
  # ... other configurations ...

  force_new_deployment = true

  triggers = {
    redeployment = timestamp()
  }
}
```

According to the [timestamp function documentation](https://developer.hashicorp.com/terraform/language/functions/timestamp), the return value can't be evaluated during the plan:
> Due to the constantly changing return value, the result of this function cannot be predicted during Terraform's planning phase, and so the timestamp will be taken only once the plan is being applied.

If we replace `timestamp()` with `plantimestamp()` (available in Terraform v1.5 and later), the forced deployment is triggered:
```
  # ... other configurations ...

  force_new_deployment = true

  triggers = {
    redeployment = plantimestamp()
  }
}
```

### References

[dev-east1-bai2.tfstate](https://s3.console.aws.amazon.com/s3/object/wextag-subprod-terraform-state-store-us-east-1?region=us-east-1&bucketType=general&prefix=dev-east1-bai2.tfstate) 
https://developer.hashicorp.com/terraform/language/functions/plantimestamp
https://developer.hashicorp.com/terraform/language/functions/timestamp

No

kirk-wex:
### Description
The example for using `timestamp()` as a trigger to force a new deployment doesn't work because `timestamp()` is evaluated once the plan is applied.

This update replaces `timestamp()` with `plantimestamp(),` which will change for every plan operation.

### Relations
Closes #35410 

### References
[timestamp()](https://developer.hashicorp.com/terraform/language/functions/timestamp)
[plantimestamp()](https://developer.hashicorp.com/terraform/language/functions/plantimestamp)
[Original documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service#redeploy-service-on-every-apply)

### DIFF
--- a/website/docs/r/ecs_service.html.markdown
+++ b/website/docs/r/ecs_service.html.markdown
@@ -113,7 +113,7 @@ resource "aws_ecs_service" "example" {
   force_new_deployment = true
 
   triggers = {
-    redeployment = timestamp()
+    redeployment = plantimestamp()
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks good to me! Thank you for taking the time to not only report this, but follow it up with a fix! I was just looking at your issue during triage, and happened to see your PR pop up 🙂 . I'll get this merged once the checks finish up 🚀

kirk-wex:
> Looks good to me! Thank you for taking the time to not only report this, but follow it up with a fix! I was just looking at your issue during triage, and happened to see your PR pop up 🙂 . I'll get this merged once the checks finish up 🚀

My pleasure. I originally didn't think I'd have time to submit a fix, but I did!

Issue #35401, opened 2024-01-19
[Bug]: Cognito user pools recreated (ie: nuked) after using aws provider v5.33.0


joshlang:
### Terraform Core Version

1.6.6

### AWS Provider Version

5.33.0

### Affected Resource(s)

aws_cognito_user_pool

### Expected Behavior

Behavior matched previous terraform versions... User pool is NOT deleted & recreated

### Actual Behavior

We lost our cognito user pool when 5.33.0 came out and the next pipeline executed.

Every pipeline execution, the user pool is recreated (even with zero changes)

### Relevant Error/Panic Output Snippet

```shell
# aws_cognito_user_pool.pool must be replaced
+/- resource "aws_cognito_user_pool" "pool" {
      ~ arn                        = "arn:aws:cognito-idp:ca-central-1:178543299566:userpool/ca-central-1_xxxxxxxx" -> (known after apply)
      - auto_verified_attributes   = [] -> null
      ~ creation_date              = "2024-01-19T15:41:41Z" -> (known after apply)
      + custom_domain              = (known after apply)
      + domain                     = (known after apply)
      ~ email_verification_message = <<-EOT
            <html xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
            ...
            </html>
        EOT -> (known after apply)
      ~ email_verification_subject = "Program Admin Portal - Your Password Reset Code" -> (known after apply)
      ~ endpoint                   = "cognito-idp.ca-central-1.amazonaws.com/ca-central-1_xxxxxxxx" -> (known after apply)
      ~ estimated_number_of_users  = 1 -> (known after apply)
      ~ id                         = "ca-central-1_xxxxxxxx" -> (known after apply)
      ~ last_modified_date         = "2024-01-19T15:41:41Z" -> (known after apply)
        name                       = "xxxxxxxx"
      ~ sms_verification_message   = "Your password reset code is {####}" -> (known after apply)
      - tags                       = {} -> null
      ~ tags_all                   = {} -> (known after apply)
      - username_attributes        = [] -> null
        # (2 unchanged attributes hidden)

      - account_recovery_setting {
          - recovery_mechanism {
              - name     = "verified_email" -> null
              - priority = 1 -> null
            }
          - recovery_mechanism {
              - name     = "verified_phone_number" -> null
              - priority = 2 -> null
            }
        }

[...]

PR #35439, opened 2024-01-23
r/aws_cognito_user_pool: Mark `username_configuration` as `Optional` and `Computed`

maxfielb:
2025-03-25 Maintainer edit: The implementation of this change was modified from the original submission. Please see the comments below for additional details. The original PR body is preserved below for completeness.

---

… uses schema version upgrade.

### Description
This change is to begin fix for bug #35401. The issue described is that the username_configuration field of user_pool Resource isn't being returned in response by our Create or Read APIs when it was not set in the Create request (it is a create-only field). Even so, there is inherently a default value for it (`case_sensitive: true`) when the field isn't set. This situation is a bug for us at Cognito and we want to begin returning this field in responses to Read and Create even when it was not set by the customer. This will more accurately communicate to the customer their resource's settings.

When we tried to do this through API changes we received an issue that customers were getting Terraform failures the field wasn't expected during Read, and the customer would have to recreate the user pool to continue with the plan. This wasn't acceptable. We rolled back that API behavior. In order to continue we realized we would need an integration change for Terraform. This is that change.

We aim to update both the Default attribute for `username_configuration` to `{ case_sensitive: true }`, which should be applied to future plans that don't set the `username_configuration`. For new user pools being created, having no state, this should fix the issue. Then we also apply a Schema Version Upgrade on top of this to fix customers with existing user pool states that DON'T have `username_configuration` set. This Upgrader will apply the `username_configuration` default setting in this case, therefore matching the state with the plan configuration (resulting in no-change).

Once this change has been published to the AWS Provider, Cognito will re-release their API behavior change that ALWAYS returns `usernameConfiguration` in UserPool Create and Read responses to complete the effort.

### Concerns

* This is my first time using Go.
* This is my first time developing or even USING Terraform.
* I have a feeling I did the `userPoolStateUpgradeV0()` `rawState["username_configuration"]` part wrong. I may be setting it for all Upgrades of the schema version, when I only want to set the `username_configuration` when it is null. Please correct me if that is in-fact the case.

### Relations
Relates #35401

### References
* Applies a state upgrade as described here: https://developer.hashicorp.com/terraform/plugin/framework/resources/state-upgrade

* Used elasticache's replication_group code as reference: https://github.com/hashicorp/terraform-provider-aws/blob/697c29e95d8d0795d337a3898e6f965f99df6186/internal/service/elasticache/replication_group.go

```console
% make testacc TESTS=TestAccXXX PKG=cognitoidp
TF_ACC=1 go test ./internal/service/cognitoidp/... -v -count 1 -parallel 20 -run='TestAccXXX'  -timeout 360m
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/cognitoidp	38.770s [no tests to run]
```


maxfielb:
Please read my "Concerns" section. Any help appreciated :)

YakDriver:
I have not reviewed this thoroughly but from a glance, this is typically considered a breaking change. If, in fact, it is, we'd have to wait until v6. Having a migration path in place, is very helpful though!

One other item I noticed, for changelog entries, you only need a file at `.changelog/35439.txt` and the changelog will be generated.

jar-b:
If in the future omitting the `username_configuration` block will result in the API returning a default value for `username_configuration.case_sensitive`, I believe we can resolve this by marking both the block and it's nested argument as Optional and Computed.

This is not a breaking change and should allow existing configurations to continue functioning as-is while handling the new Read API behavior whenever it is released. For now I'd propose removing the `breaking-change` label and dropping from the V6 milestone.

jar-b:
@maxfielb - Thanks for your contribution! 👍 

Given the implementation has changed from the original intent, I'm going to adjust the title and add a section above the original PR body just to add extra clarity for anyone clicking through to this PR from the changelog.

johnsonaj (approved):
LGTM 🚀

PR #35485, opened 2024-01-25
Update vpc_peering_connection.html.markdown

Deku-shrub:
Disambiguate the peer / target documentation

### Description
Many hours wasted with the 'owner' attribute which owns nothing. Instead using the more intuitive jargon 'target'.

Failure to set the peer_owner_id  attribute when needed will try and fail to find the cross account VPC ID locally, erroring with:
```
Error: creating EC2 VPC Peering Connection: InvalidVpcID.NotFound: The vpc ID 'vpc-xxxxx' does not exist status code: 400, request id: xxx
```

```console

...
```

### DIFF
--- a/website/docs/r/vpc_peering_connection.html.markdown
+++ b/website/docs/r/vpc_peering_connection.html.markdown
@@ -105,9 +105,9 @@ more information.
 
 This argument supports the following arguments:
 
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks good to me! That seems like a totally reasonable clarification. Thank you for your contribution! I'll get it merged shortly 🚀

PR #35518, opened 2024-01-27
docs: Improve access_endpoint attr desc for aws_appstream_image_builder

### LINKED ISSUE #33029: [Docs]: "vpce_id" parameter in "aws_appstream_image_builder" is required when you specify "endpoint_type"
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appstream_image_builder#endpoint_type

### Description

Resource documentation state following for "vpce_id":
`[access_endpoint](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appstream_image_builder#access_endpoint)
The access_endpoint block supports the following arguments:

[endpoint_type](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appstream_image_builder#endpoint_type) - (Required) Type of interface endpoint.
[vpce_id](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appstream_image_builder#vpce_id) - (Optional) Identifier (ID) of the VPC in which the interface endpoint is used.

As you could noticed it mentioned that only "endpoint_type" is required, but actually when you provide it you must also provide an endpoint id:
`│ Error: creating AppStream ImageBuilder (test-image-builder): InvalidParameter: 1 validation error(s) found.
│ - minimum field size of 1, CreateImageBuilderInput.AccessEndpoints[0].VpceId.
│ 
│ 
│   with module.appstream.aws_appstream_image_builder.image_builder,
│   on ../main.tf line 4, in resource "aws_appstream_image_builder" "image_builder":
│    4: resource "aws_appstream_image_builder" "image_builder" {
`
Also the I was confused of the wording in the descriptions "Identifier (ID) of the VPC in which the interface endpoint is used". It need the id of the VPC ENDPOINT and not the actual id of the vpc.

Another thing which I noticed is related to the other parameter in this block "endpoint_type". It description states "Type of interface endpoint." but this could not be anything different from "STREAMING". Could this be added in the documentation also? 
Here is output from terraform apply with "endpoint_type" different from "STREAMING":
`│ Error: expected access_endpoint.0.endpoint_type to be one of [STREAMING], got ruk
│ 
│   with module.appstream.aws_appstream_image_builder.image_builder,
│   on ../main.tf line 4, in resource "aws_appstream_image_builder" "image_builder":
│    4: resource "aws_appstream_image_builder" "image_builder" {
`

### References

_No response_

No

acwwat:
### Description

This PR is to improve the description of the `access_endpoint` config block attributes for the `aws_appstream_image_builder` resource. What I noticed is that the description for `vpce_id` is consistently incorrect in AWS documentations, so I just used my own wordings for more accuracy. Also, the schema seems to set a lot of attributes to optional but in fact they would be required. For consistency, I still left `vpce_id` as optional.

### Relations

Closes #33029

### References

Referred to [CreateImageBuilder](https://docs.aws.amazon.com/appstream2/latest/APIReference/API_CreateImageBuilder.html#AppStream2-CreateImageBuilder-request-DomainJoinInfo) in the API reference for schema and descriptions.

n/a

### DIFF
--- a/website/docs/r/appstream_image_builder.html.markdown
+++ b/website/docs/r/appstream_image_builder.html.markdown
@@ -56,8 +56,8 @@ The following arguments are optional:
 
 The `access_endpoint` block supports the following arguments:
 
-* `endpoint_type` - (Required) Type of interface endpoint.
-* `vpce_id` - (Optional) Identifier (ID) of the VPC in which the interface endpoint is used.
+* `endpoint_type` - (Required) Type of interface endpoint. Valid values: `STREAMING`
+* `vpce_id` - (Optional) Identifier (ID) of the interface VPC endpoint.
 
 ### `domain_join_info`
 
@@ -80,7 +80,7 @@ This resource exports the following attributes in addition to the arguments abov
 * `arn` - ARN of the appstream image builder.
 * `created_time` -  Date and time, in UTC and extended RFC 3339 format, when the image builder was created.
 * `id` - Name of the image builder.
-* `state` - State of the image builder. Can be: `PENDING`, `UPDATING_AGENT`, `RUNNING`, `STOPPING`, `STOPPED`, `REBOOTING`, `SNAPSHOTTING`, `DELETING`, `FAILED`, `UPDATING`, `PENDING_QUALIFICATION`
+* `state` - State of the image builder. Valid values: `PENDING`, `UPDATING_AGENT`, `RUNNING`, `STOPPING`, `STOPPED`, `REBOOTING`, `SNAPSHOTTING`, `DELETING`, `FAILED`, `UPDATING`, `PENDING_QUALIFICATION`
 * `tags_all` - Map of tags assigned to the resource, including those inherited from the provider [`default_tags` configuration block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags-configuration-block).
 
 ## Import
--- a/website/docs/r/appstream_image_builder.html.markdown
+++ b/website/docs/r/appstream_image_builder.html.markdown
@@ -56,7 +56,7 @@ The following arguments are optional:
 
 The `access_endpoint` block supports the following arguments:
 
-* `endpoint_type` - (Required) Type of interface endpoint. Valid values: `STREAMING`
+* `endpoint_type` - (Required) Type of interface endpoint. For valid values, refer to the [AWS documentation](https://docs.aws.amazon.com/appstream2/latest/APIReference/API_AccessEndpoint.html).
 * `vpce_id` - (Optional) Identifier (ID) of the interface VPC endpoint.
 
 ### `domain_join_info`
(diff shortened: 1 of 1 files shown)


website/docs/r/appstream_image_builder.html.markdown
```diff
@@ -56,8 +56,8 @@ The following arguments are optional:
 
 The `access_endpoint` block supports the following arguments:
 
-* `endpoint_type` - (Required) Type of interface endpoint.
-* `vpce_id` - (Optional) Identifier (ID) of the VPC in which the interface endpoint is used.
+* `endpoint_type` - (Required) Type of interface endpoint. Valid values: `STREAMING`
```
justinretzolk, line 59: We've had a lot of times where listing valid values winds up with the options getting out of sync, so I tend towards linking directly to the AWS API docs for this type of thing.

```suggestion
* `endpoint_type` - (Required) Type of interface endpoint. For valid values, refer to the [AWS documentation](https://docs.aws.amazon.com/appstream2/latest/APIReference/API_AccessEndpoint.html)
```

website/docs/r/appstream_image_builder.html.markdown
```diff
@@ -56,8 +56,8 @@ The following arguments are optional:
...
+* `vpce_id` - (Optional) Identifier (ID) of the interface VPC endpoint.
```
justinretzolk, line 60: No notes here, just _wow_ I had to read the AWS docs, this, and the argument name several times to wrap my head around it because of the weird description issues 😅. Thanks for wording this in a way that made it make sense!

website/docs/r/appstream_image_builder.html.markdown
```diff
@@ -80,7 +80,7 @@ This resource exports the following attributes in addition to the arguments abov
 * `arn` - ARN of the appstream image builder.
 * `created_time` -  Date and time, in UTC and extended RFC 3339 format, when the image builder was created.
 * `id` - Name of the image builder.
-* `state` - State of the image builder. Can be: `PENDING`, `UPDATING_AGENT`, `RUNNING`, `STOPPING`, `STOPPED`, `REBOOTING`, `SNAPSHOTTING`, `DELETING`, `FAILED`, `UPDATING`, `PENDING_QUALIFICATION`
+* `state` - State of the image builder. Valid values: `PENDING`, `UPDATING_AGENT`, `RUNNING`, `STOPPING`, `STOPPED`, `REBOOTING`, `SNAPSHOTTING`, `DELETING`, `FAILED`, `UPDATING`, `PENDING_QUALIFICATION`
```
justinretzolk, line 83: Same suggestion as above, but this one does challenge my perception of this a bit. I added the `#state` anchor link tag, but it doesn't seem to really work, meaning this link goes to a fairly length page that you need to scroll down to the `state` header on.

In your opinion, does it make more sense to list here vs. linking? 

```suggestion
* `state` - State of the image builder. For valid values, see the [AWS documentation](https://docs.aws.amazon.com/appstream2/latest/APIReference/API_ImageBuilder.html#state)
```

website/docs/r/appstream_image_builder.html.markdown
acwwat: Thanks, will follow this convention going forward.

website/docs/r/appstream_image_builder.html.markdown
acwwat: I was looking at the HTML of the documentation for the anchor and it seems that `#AppStream2-Type-ImageBuilder-State` works:

https://docs.aws.amazon.com/appstream2/latest/APIReference/API_ImageBuilder.html#AppStream2-Type-ImageBuilder-State

I'll update the PR with all the proposed changes with a new commit.

acwwat:
@justinretzolk FYI the PR is ready for re-review. I couldn't remove the waiting-response tag but hopefully adding this comment will trigger it. Thanks.

justinretzolk (approved):
Looks good to me! Thanks for another great contribution, and apologies for the delay! 🚀

PR #35581, opened 2024-01-31
Quick fix: Loosen restriction on command line argument length

kavan-aws:
### Description

This minor change increases the upper limit on command line argument key/value length from 50 to 1024, in accordance with a model change on our side. We perform validation on both of these on our side as well.


breathingdust:
LGTM 🚀 Thanks @kavan-aws!

Issue #23567, opened 2022-03-08
Resource aws_iam_user_login_profile "forces replacement" after upgrade to aws provider v4x


Arisfx:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

- aws_iam_user_login_profile

Hi team, as the title says, the resource aws_iam_user_login_profile breaks after the upgrade to AWS provider v4x.

While the terraform code dictates `password_reset_required = true` and in the state file i can see that `"password_reset_required": true,` (in the aws console of course its enabled). 
When i am running terraform apply i get the `~ password_reset_required = false -> true # forces replacement` 

I understand that the upgrade to aws provider v4x is introducing many breaking changes but i wasn’t able to find anything useful in the upgrade documentation [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade)

### Terraform CLI and Terraform AWS Provider Version

P.S We are not using multiple providers, there is only 1 piece of code that this provider is declared:
```
  required_providers {
aws = {
  source  = "hashicorp/aws"
  version = "4.4.0"
 }
}
```
Am i missing something?
Thanks!

rustige:
Had the same issue. Worked around it by explicitly ignoring changes on password_reset_required by lifcycle policy on the resource. Which should be the case automatically according to the manual.

Workaround:
`  lifecycle {
    ignore_changes = [password_reset_required]
  }`

rowleyaj:
Is this possibly related to https://github.com/hashicorp/terraform-provider-aws/pull/12384/? The default value for `password_reset_required` was removed and it was changed to a computed field in that PR.

I've tested `4.0.0` and this issue doesn't happen, but in `4.1.0` which included that PR the issue does happen.

cc @DrFaust92 as the PR author as I'm unsure on the context for that change

amontalban:
I'm having the same problem, unfortunately, I am using [terraform-aws-iam](https://github.com/terraform-aws-modules/terraform-aws-iam) module and I cannot do the workaround without creating a fork.

[...]

PR #35601, opened 2024-02-01
DOCS: Change aws_iam_user_login_profile's encrypted_password text.

sierra-alpha:
### Description

There is a descrepency between the docs and observed behaviour for `aws_iam_user_login_profile`'s `encrypted_password`. Unfortunatly I don't know any golang so can't have a crack at changing the implementation but I can change the docs to save others time with this in the future.

This is to make the docs align with observed behaviour of not "_only applies on resource creation_" but on all `terraform apply`s. 

This also provides documentation of an interim work around. I read this as maybe a not good practise in the Contributing Guide so I'm willing to modify this, but we should at least provide a hint for how to use this without it trying to be re-applied every time a user has changed their password from the initial one. 

I'm open to any wording or formatting changes and happy for a maintainer to make changes as they see fit to this PR, the only thing that is needed is for the docs to align with the resources behaviour.

### Relations

Relates #23567

### References

Steps to reproduce the reasoning for this documentation change. 

1. Use terraform config:
```
variable "developers" {
  description = "A list of objects of developers usernames and their public PGP key to setup."
  type        = set(object({ user-name : string, their-public-pgp-key : string }))
}

resource "aws_iam_user" "iam-users-developers" {
  for_each = { for obj in var.developers : obj.user-name => obj }

  name = each.value.user-name
}

resource "aws_iam_user_login_profile" "iam-login-developers" {
  depends_on = [aws_iam_user.iam-users-developers]
  for_each   = { for obj in var.developers : obj.user-name => obj }

  user                    = each.value.user-name
  pgp_key                 = each.value.their-public-pgp-key
  password_reset_required = true
}
```
2. `terraform apply` it.
3. Log into the AWS console as this new user.
4. Change the password as required.
5. Perform a `terraform plan` and see that it's showing destroy and create on the user.
6. (Optionally) `terraform apply` and see that you need to use the new pgp encrypted password to log into the user.

Documentation change so doesn't require tests?

### DIFF
--- a/website/docs/r/iam_user_login_profile.html.markdown
+++ b/website/docs/r/iam_user_login_profile.html.markdown
@@ -38,7 +38,17 @@ This resource supports the following arguments:
 * `user` - (Required) The IAM user's name.
 * `pgp_key` - (Optional) Either a base-64 encoded PGP public key, or a keybase username in the form `keybase:username`. Only applies on resource creation. Drift detection is not possible with this argument.
 * `password_length` - (Optional) The length of the generated password on resource creation. Only applies on resource creation. Drift detection is not possible with this argument. Default value is `20`.
-* `password_reset_required` - (Optional) Whether the user should be forced to reset the generated password on resource creation. Only applies on resource creation.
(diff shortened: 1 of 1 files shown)


sierra-alpha:
So I read the contributing guide etc before submitting this PR, is there anything else I need to do from my end to get this looked at?

justinretzolk:
Hey @sierra-alpha 👋 Thank you for taking the time to raise this! I'm going to bring this up to the team, so no further action needed from you at this point.

PR #35659, opened 2024-02-06
[docs] CodeBuild Project attribute clarifications

Michagogo:
The resource access role is only applicable to public log access (and isn’t annotated as optional), and the log stream name is actually a prefix.

### Description

Clarify the two points mentioned above

### Relations

N/A

### References

https://docs.aws.amazon.com/codebuild/latest/userguide/change-project-console.html#change-project-console.public-builds
https://docs.aws.amazon.com/codebuild/latest/APIReference/API_UpdateProjectVisibility.html

https://docs.aws.amazon.com/codebuild/latest/APIReference/API_CloudWatchLogsConfig.html

N/A


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@Michagogo Thanks for the contribution :tada: :clap:.

Issue #34416, opened 2023-11-15
[Docs]: Misleading attribute documentation on cloudfront_origin_access_identity datasource


benja-M-1:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/cloudfront_origin_access_identity

### Description

Hello, on the https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/cloudfront_origin_access_identity, the description of the `id` attribute specifies `The identifier for the distribution` with an example.
This misled me: I put the id of the cloudfront distribution but it didn't exist. I replaced it with the id of the distribution origin and then it worked.

I suggest something like `The identifier for the origin of the distribution` instead.

### Would you like to implement a fix?

None

jtyrus:
I can update this doc

rama280290:
Can i work on this?

PR #34483, opened 2023-11-20
Fixed docs for data.aws_cloudfront_origin_access_identity

jtyrus:
### Description
Updated description for the `id` attribute of `data.aws_cloudfront_origin_access_identity`. Id should be that of the access identity, not the CLoudfront distribution.

### Relations
Closes #34416

### References
[Cloudfront Access Control](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html) (Legacy OAI at the bottom)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/cloudfront_origin_access_identity

PR #35692, opened 2024-02-07
Modified the cloudfront origin access identity doc

rama280290:
Description

Updated description for the id attribute of aws_cloudfront_origin_access_identity since it is misleading users with Cloudfront distribution.

Relations
Closes https://github.com/hashicorp/terraform-provider-aws/issues/34416

References

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/cloudfront_origin_access_identity


justinretzolk:
Hey @rama280290 👋 Thank you for taking the time to raise this! It looks like this is a duplicate of #34483, which was opened a bit earlier, so we're going to close this one in favor of #34483.

PR #35716, opened 2024-02-08
Update go.mod for better module version

maxmeng-hw:
Update go.mod for valid version path

### Description

For now, go require version matches SCM version in module name  while version > 1.0

### References
https://go.dev/ref/mod#version


gdavison:
Thanks for the contribution, @maxmeng-hw. Because the `module` directive is the root for the module, it should not include the version. I'm going to close the PR

Issue #36332, opened 2024-03-12
[New Resource]: `aws_devopsguru_resource_collection`


jar-b:
### Description

Manages a collection of resources for AWS DevOps Guru to analyze.

### Requested Resource(s) and/or Data Source(s)

- `aws_devopsguru_resource_collection`

### Potential Terraform Configuration

```terraform
resource "aws_devopsguru_resource_collection" "example" {
  resource_collection {
    cloudformation {
      stack_names = ["foo", "bar"]
    }
  }
}
```

### References
- Relates #17919
- https://docs.aws.amazon.com/devops-guru/latest/APIReference/API_UpdateResourceCollection.html
- https://docs.aws.amazon.com/devops-guru/latest/APIReference/API_GetResourceCollection.html

### Would you like to implement a fix?

None

PR #36437, opened 2024-03-18
autoflex: wire up option functions, add `IgnoreFieldNames` option

jar-b:
### Description

This change implements AutoFlex options, allowing custom options to be passed to `flex.Expand` or `flex.Flatten`. The changes include:

- Adding a new `AutoFlexOptions` struct, which stores all configurable options
- Embedding the `AutoFlexOptions` struct into the `autoExpand` and `autoFlatten` flexer structs
- Adjusting the `AutoFlexOptionsFunc` signature to accept a pointer to `AutoFlexOptions`

As a first use case, the `IgnoreFieldNames` option was added. This option is a slice of strings, with a default value of `[]string{"Tags"}`. This replaces the previously hardcoded logic which skipped processing of any fields named `Tags`, and instead makes this configurable. Test cases have been added to exercise the behavior with default options and overrides to remove defaults and add custom values.

### Relations

Relates #36332
Depends on https://github.com/hashicorp/terraform-plugin-framework/pull/961
Depends on #36992

### References

- AWS SDK for Go V2 retry options, the pattern on which these options are based: https://github.com/aws/aws-sdk-go-v2/blob/main/aws/retry/standard.go

```console
% go test ./internal/framework/flex/...
ok      github.com/hashicorp/terraform-provider-aws/internal/framework/flex     0.488s
```

### DIFF
--- a/internal/framework/flex/auto_expand.go
+++ b/internal/framework/flex/auto_expand.go
@@ -26,11 +26,7 @@ import (
 // target data type) are copied.
 func Expand(ctx context.Context, tfObject, apiObject any, optFns ...AutoFlexOptionsFunc) diag.Diagnostics {
 	var diags diag.Diagnostics
-	expander := &autoExpander{}
-
-	for _, optFn := range optFns {
-		optFn(expander)
-	}
+	expander := newAutoExpander(optFns)
 
 	diags.Append(autoFlexConvert(ctx, tfObject, apiObject, expander)...)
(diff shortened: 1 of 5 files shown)


ewbankkit (approved):
LGTM :rocket:.

```console
% go test ./internal/framework/flex 
ok  	github.com/hashicorp/terraform-provider-aws/internal/framework/flex	0.817s
```

PR #36581, opened 2024-03-26
AutoFlEx: Add function to `null` out "model" structure fields

ewbankkit:
### Description

Adds `NullOutObjectPtrFields`, a function to `null` out all the applicable fields in a newly allocated "model" structure.


jar-b (approved):
LGTM 🥳

PR #36615, opened 2024-03-27
r/aws_applicationinsights_application: ACTIVE is a valid create target status

### LINKED ISSUE #27277: [Bug]: aws_applicationinsights_application unexpected state with auto_config_enabled=true
### Terraform Core Version

0.13.6

### AWS Provider Version

4.30.0,4.34.0

### Affected Resource(s)

* aws_applicationinsights_application

### Expected Behavior

It should be able to create and destroy the resource without issue.

### Actual Behavior

When `auto_config_enabled` is true, it gets an "unexpected state" error on both create and destroy. This results in the resource being tainted on create, and then another "unexpected state" error when it destroys the resource before it can recreate it.

```
Error: waiting for ApplicationInsights Application (test) create: unexpected state 'CONFIGURING', wanted target 'NOT_CONFIGURED'. last error: %!s(<nil>)
```

```
Error: waiting for ApplicationInsights Application (test) delete: unexpected state 'ACTIVE', wanted target ''. last error: %!s(<nil>)
```

```terraform
resource "aws_applicationinsights_application" "test" {
  resource_group_name = aws_resourcegroups_group.test.name

  auto_config_enabled = true
}

resource "aws_resourcegroups_group" "test" {
  name = "test"

  resource_query {
    query = jsonencode({
      ResourceTypeFilters = ["AWS::AllSupported"]
      TagFilters          = [{ Key = "test" }]
    })
  }
}
```

### Steps to Reproduce

Apply the above config, or any applicationinsights_application with auto_config_enabled=true.

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

With auto_config_enabled=false, there is no error, since the expected state of `NOT_CONFIGURED` is reached.

It seems like the solution would simply be to add additional possible states to the wait functions.

### References

https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/applicationinsights/wait.go is the location of the wait functions where the expected states are listed.

The AWS API docs don't include any information of the actual values of the LifeCycle field, so I just watched the value in a loop as I ran apply to create and destroy an application. It's possible that it could also be in `CREATING` when deleting, but I wasn't able to replicate that since it goes from `CREATING` to the next state pretty quickly.

When creating an application, the LifeCycle goes `CREATING` -> [`CONFIGURING`] -> (`ACTIVE`|`NOT_CONFIGURED`)

When deleting an application, it goes (`ACTIVE`|`CONFIGURING`|`NOT_CONFIGURED`) -> `DELETING` -> *gone* 

The `waitApplicationCreated` function would need `Pending` to be `[]string{"CREATING", "CONFIGURING"}`, and `Target` would need to be `[]string{"ACTIVE", "NOT_CONFIGURED"}`. Maybe `"CONFIGURING"` would belong in the `Target` slice instead? I guess it depends on whether it should wait for the application to be configured or not.

The `waitApplicationTerminated` function would need `Pending` to be `[]string{"ACTIVE", "CONFIGURING", "NOT_CONFIGURED", "DELETING"}` (`"CONFIGURING` is included for the case where something else failed during apply, caused the application to be tainted, and a new apply is started before it is done configuring). `Target` would stay `[]string{}`.

I would have opened a PR with these changes, but I am pretty sure the tests will need to be updated and I don't have the free time to update and add those currently.

No

jar-b:
### Description

In some configurations with `auto_config_enabled = true`, intermittent errors are observed waiting for application creation.

```
│ Error: waiting for ApplicationInsights Application (example) create: unexpected state 'ACTIVE', wanted target 'NOT_CONFIGURED'. last error: %!s(<nil>)
```

The current assumption is that all applications enter a `NOT_CONFIGURED` lifecycle status upon initial creation (which is why this succeeds most of the time in minimal configurations such as our acceptance test), but in some instances can proceed into an `ACTIVE` status before the create waiter has polled for and observed a `NOT_CONFIGURED` status. The AWS documentation on the [LifeCycle argument](https://docs.aws.amazon.com/cloudwatch/latest/APIReference/API_ApplicationInfo.html#appinsights-Type-ApplicationInfo-LifeCycle) is limited, so while we cannot be certain about the expected values in the lifecycle flow, there is enough evidence from issue reports and manual testing to indicate `ACTIVE` should be considered a valid target state.

### Relations

Closes #27277 

### References

- https://docs.aws.amazon.com/cloudwatch/latest/APIReference/API_ApplicationInfo.html#appinsights-Type-ApplicationInfo-LifeCycle

```console
% make testacc PKG=applicationinsights TESTS=TestAccApplicationInsightsApplication_
TF_ACC=1 go1.21.8 test ./internal/service/applicationinsights/... -v -count 1 -parallel 20 -run='TestAccApplicationInsightsApplication_'  -timeout 360m

--- PASS: TestAccApplicationInsightsApplication_disappears (22.83s)
--- PASS: TestAccApplicationInsightsApplication_autoConfig (25.47s)
--- PASS: TestAccApplicationInsightsApplication_basic (34.21s)
--- PASS: TestAccApplicationInsightsApplication_tags (43.17s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/applicationinsights        48.867s
```

### DIFF
--- a/.changelog/36615.txt
+++ b/.changelog/36615.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_applicationinsights_application: Make `ACTIVE` a valid create target status
+```
--- a/internal/service/applicationinsights/wait.go
+++ b/internal/service/applicationinsights/wait.go
@@ -20,7 +20,7 @@ const (
 func waitApplicationCreated(ctx context.Context, conn *applicationinsights.ApplicationInsights, name string) (*applicationinsights.ApplicationInfo, error) {
 	stateConf := &retry.StateChangeConf{
 		Pending: []string{"CREATING"},
-		Target:  []string{"NOT_CONFIGURED"},
+		Target:  []string{"NOT_CONFIGURED", "ACTIVE"},
 		Refresh: statusApplication(ctx, conn, name),
 		Timeout: ApplicationCreatedTimeout,
 	}


ewbankkit (approved):
LGTM :rocket:.

PR #36662, opened 2024-03-31
Auto-generate patches to help with AWS SDK v2 migration

mattburgess:
### Description

Adds a generator to produce [gopatch](https://github.com/uber-go/gopatch) patches that do some, but not all, of the translations from v1 to v2 of the AWS SDK. This is definitely POC quality and may not be the automation approach you want to adopt. However, running the patch against the `backup` service shows it reduces the number of errors post-import and post-client migration from 245 to 130 so it will save a reasonable amount of time.

### Relations

Relates #32976

```console

...
```


.ci/tools/go.mod
```diff
@@ -270,6 +272,7 @@ require (
 	github.com/timonwong/loggercheck v0.9.4 // indirect
 	github.com/tomarrell/wrapcheck/v2 v2.8.3 // indirect
 	github.com/tommy-muehle/go-mnd/v2 v2.5.1 // indirect
+	github.com/uber-go/gopatch v0.3.0 // indirect
```
jar-b, line 275: Pinned to `0.3.0` for now as the latest version (`0.4.0`) requires Go 1.22, which the provider has not yet upgraded to. Will open an issue reminder to upgrade this as part of the provider upgrade.

jar-b:
@mattburgess - First off, thank you for the effort on this and submitting the proof of concept! Automating these base level changes will result in a huge time savings given the scale of the SDK migration across the provider.

I made some high level changes to the initial design, primarily around swapping the entry point from a generator to a CLI. This will allow us to chain together generation of the patch file and application of the changes behind a make target.

```console
PKG=ec2 make awssdkpatch-apply
```

I've also added `uber-go/gopatch` to the list of provider tools so that `make tools` will now install it automatically. I'm really excited about the automation this is going to enable going forward - thanks again!

jar-b (approved):
LGTM 🎉 🤖

jar-b:
Thanks for your contribution, @mattburgess! 👍

PR #36735, opened 2024-04-04
[Tech Debt] Split EC2 tests in CI

johnsonaj:
### Description

Currently EC2 tests time out consistently after running for 12 hours in the same runner. Splitting these tests into their sub groups could be a first step with getting them to complete. Each sub group would have its own allocation of resources to run tests.

```console

...
```


internal/generate/teamcity/acctest_services.hcl
YakDriver, line 1: Eventually this all needs to be centralized with names so we're not having to change several places (and forgetting any).

gdavison (approved):
Looks good! It will be nice to finally get these broken out 🚀

PR #36983, opened 2024-04-18
r/ecs_sevice doc: removes duplicate references

thatderek:
### Description
`name`, `Iam_role`, `desired_count` are all attribute arguments to `ecs_service` but the docs listed them both as attribute-arguments and as additional attributes. My brain tripped over this when it came to `iam_role` because I thought to myself "wait, I thought the IAM role was an argument on this resource". 

This isn't a big change and maybe it too much of a nit pick vs perhaps it's useful to keep this thing in multiple places, but I know I tripped over it and figured I cast a PR out there to see if this is thing kind of thing that's worth a quick clean up for the next me/ person like me. 

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/.changelog/36983.txt
+++ b/.changelog/36983.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_ecs_service: Removes duplicate argument references to `name`, `cluster`, and `iam_role` in docs.
+```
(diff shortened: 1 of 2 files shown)


thatderek:
btw, if this is way too nit picky, I'll take zero offense as it being closed without comment or explanation.

justinretzolk (approved):
This seems totally reasonable to me, and fits the well-established pattern. I removed the changelog entry (we don't add changelog entries for docs fixes), but other than that, looks good to me! I'll get this merged shortly. Thanks for that contribution! 🚀

PR #35095, opened 2023-12-29
Add support for optional KMS key ID in repository resource

obounaim:
### Description
Add support for optional KMS key ID in repository resource

### Relations

Closes #35040

### References
[Announcing Customer Managed Key (CMK) support in AWS CodeCommit](https://aws.amazon.com/about-aws/whats-new/2023/12/customer-managed-key-support-aws-codecommit/)
[Go SDK CreateRepository](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.49.12/service/codecommit#CodeCommit.CreateRepository)
[Go SDK UpdateRepositoryEncryptionKey](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.49.12/service/codecommit#CodeCommit.UpdateRepositoryEncryptionKey)
[AWS Key Management Service and encryption for AWS CodeCommit repositories](https://docs.aws.amazon.com/codecommit/latest/userguide/encryption.html)

PR #37047, opened 2024-04-22
[docs] Typo fix in website docs

mbbush:
### Description

Fix a typo in the documentation for `r/aws_codecommit_repository`

### References

When the `aws_codecommit_repository` resource was enhanced to add the `kms_key_id` parameter in https://github.com/hashicorp/terraform-provider-aws/pull/35095, the documentation added for the new parameter had a typo in it, and closed the name of the fallback AWS-managed key with a double backtick ` `` ` instead of a single `. 

This results in [the name not rendering properly on the terraform provider registry website](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/codecommit_repository#kms_key_id), and for some strange reason I haven't pinned down yet, only sometimes breaks a tool I use that parses the terraform provider documentation. I'll take care of fixing the tool, but I thought I'd open the PR to clean up the website. 

The same typo is present in the cdktf docs, but it looks like the normal practice is for those to be updated by codegen scripts, not by humans. If I'm wrong about that I prepared 3a3dfd1c1975cb02c8f2ba7b5ea0dbb33b2789be as well, but didn't open a PR.

I didn't run acceptance testing because it's only a documentation change.

### DIFF
--- a/website/docs/r/codecommit_repository.html.markdown
+++ b/website/docs/r/codecommit_repository.html.markdown
@@ -41,7 +41,7 @@ This resource supports the following arguments:
 * `repository_name` - (Required) The name for the repository. This needs to be less than 100 characters.
 * `description` - (Optional) The description of the repository. This needs to be less than 1000 characters
 * `default_branch` - (Optional) The default branch of the repository. The branch specified here needs to exist.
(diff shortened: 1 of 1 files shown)


justinretzolk (approved):
Looks good to me! Thank you for taking the time to contribute the fix! I'll get this merged as soon as the checks finish up 🚀

PR #37089, opened 2024-04-25
aws_wafv2_web_acl_association Data Source Get by Resource ARN

### LINKED ISSUE #36998: [Enhancement]: allow to discover wafv2 acl by resource
### Description

Right now [aws_wafv2_web_acl](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/wafv2_web_acl) data source allows discovery by `name` and `scope` only. 
This data source can be enhanced by allowing `resource` field that can be used to discover attached ACLs to that specific resource by the use of [GetWebACLForResource](https://docs.aws.amazon.com/waf/latest/APIReference/API_wafRegional_GetWebACLForResource.html) API where `scope` is "REGIONAL" or [GetDistributionConfig](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_GetDistributionConfig.html) for the "CLOUDFRONT" `scope`

This enhancement will help specifically in cases where dynamic ACL attachments is used, for example with Firewall manager (FMS).

* [aws_wafv2_web_acl](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/wafv2_web_acl)

```terraform
data "aws_wafv2_web_acl" "alb_example" {
  resource  = "arn:aws:elasticloadbalancing:us-east-1:0123456789:loadbalancer/app/my-alb/xxxxx"
  scope = "REGIONAL"
}

data "aws_wafv2_web_acl" "cfront_example" {
  resource  = "arn:aws:cloudfront::0123456789:distribution/XXX"
  scope = "CLOUDFRONT"
}
```

### References

CLI documentation for the [GetWebACLForResource](https://docs.aws.amazon.com/cli/latest/reference/wafv2/get-web-acl-for-resource.html) 

Related issue: https://github.com/hashicorp/terraform-provider-aws/issues/36941

No

jtyrus:
### Description
Added `resource_arn` to the `aws_wafv2_web_acl_association` data source.

### Relations
Closes #36998

```console
$ make testacc TESTS=TestAccWAFV2WebACLDataSource PKG=wafv2
TF_ACC=1 go1.22.2 test ./internal/service/wafv2/... -v -count 1 -parallel 20 -run='TestAccWAFV2WebACLDataSource'  -timeout 360m
=== RUN   TestAccWAFV2WebACLDataSource_basic
=== PAUSE TestAccWAFV2WebACLDataSource_basic
=== RUN   TestAccWAFV2WebACLDataSource_arn
=== PAUSE TestAccWAFV2WebACLDataSource_arn
=== CONT  TestAccWAFV2WebACLDataSource_basic
=== CONT  TestAccWAFV2WebACLDataSource_arn
--- PASS: TestAccWAFV2WebACLDataSource_basic (20.21s)
--- PASS: TestAccWAFV2WebACLDataSource_arn (54.76s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/wafv2      54.961s

```

### DIFF
--- a/internal/service/wafv2/web_acl_data_source.go
+++ b/internal/service/wafv2/web_acl_data_source.go
@@ -32,12 +32,18 @@ func DataSourceWebACL() *schema.Resource {
 				},
 				"name": {
 					Type:     schema.TypeString,
-					Required: true,
+					Computed: true,
+					Optional: true,
+				},
+				"resource_arn": {
+					Type:     schema.TypeString,
+					Computed: true,
+					Optional: true,
 				},
 				"scope": {
-					Type:         schema.TypeString,
-					Required:     true,
-					ValidateFunc: validation.StringInSlice(wafv2.Scope_Values(), false),
+					Type:     schema.TypeString,
+					Computed: true,
+					Optional: true,
 				},
 			}
 		},
@@ -47,6 +53,15 @@ func DataSourceWebACL() *schema.Resource {
 func dataSourceWebACLRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	var diags diag.Diagnostics
 	conn := meta.(*conns.AWSClient).WAFV2Conn(ctx)
+	diags = validateDataSourceInput(d)
+	if diags.HasError() {
+		return diags
+	}
+
+	if v, ok := d.GetOk("resource_arn"); ok {
+		return getWebAclDataSourceByArn(ctx, conn, v.(string), d, diags)
+	}
+
 	name := d.Get("name").(string)
 
 	var foundWebACL *wafv2.WebACLSummary
@@ -87,4 +102,45 @@ func dataSourceWebACLRead(ctx context.Context, d *schema.ResourceData, meta inte
 	d.Set("description", foundWebACL.Description)
 
 	return diags
+}
+
+func getWebAclDataSourceByArn(ctx context.Context, conn *wafv2.WAFV2, arn string, d *schema.ResourceData, diags diag.Diagnostics) diag.Diagnostics {
+	input := &wafv2.GetWebACLForResourceInput{
+		ResourceArn: aws.String(arn),
+	}
+
+	output, err := conn.GetWebACLForResourceWithContext(ctx, input)
+	if err != nil {
+		return sdkdiag.AppendErrorf(diags, "reading WAFv2 WebACLs: %s", err)
+	}
+
+	if output == nil || output.WebACL == nil {
+		return sdkdiag.AppendErrorf(diags, "WAFv2 WebACL not found for source: %s", arn)
+	}
+
+	d.SetId(*output.WebACL.Id)
+	d.Set("arn", *output.WebACL.ARN)
+	d.Set("name", *output.WebACL.Name)
+	d.Set("description", *output.WebACL.Description)
+	return diags
+}
+
+func validateDataSourceInput(d *schema.ResourceData) diag.Diagnostics {
+	if _, ok := d.GetOk("resource_arn"); ok {
+		return nil
+	}
+
+	if _, ok := d.GetOk("name"); !ok {
+		return diag.Errorf("name is required")
+	}
+
+	if v, ok := d.GetOk("scope"); !ok {
+		return diag.Errorf("scope is required")
+	} else {
+		_, err := validation.StringInSlice(wafv2.Scope_Values(), false)(v, "scope")
+		if len(err) == 0 {
+			return nil
+		}
+		return diag.FromErr(err[0])
+	}
 }
--- a/website/docs/d/wafv2_web_acl.html.markdown
+++ b/website/docs/d/wafv2_web_acl.html.markdown
@@ -23,8 +23,9 @@ data "aws_wafv2_web_acl" "example" {
 
 This data source supports the following arguments:
 
-* `name` - (Required) Name of the WAFv2 Web ACL.
-* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the region `us-east-1` (N. Virginia) on the AWS provider.
+* `name` - (Required) Name of the WAFv2 Web ACL. Optional if `resource_arn` is provided.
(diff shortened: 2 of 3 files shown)


website/docs/d/wafv2_web_acl.html.markdown
```diff
@@ -23,8 +23,9 @@ data "aws_wafv2_web_acl" "example" {
 
 This data source supports the following arguments:
 
-* `name` - (Required) Name of the WAFv2 Web ACL.
-* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the region `us-east-1` (N. Virginia) on the AWS provider.
+* `name` - (Required) Name of the WAFv2 Web ACL. Optional if `resource_arn` is provided.
+* `resource_arn` - (Optional) Name of a resource associated with a WAF. 
+* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the region `us-east-1` (N. Virginia) on the AWS provider. Optional if `resource_arn` is provided.
```
stumyp, line 28: There's trailing space that breaks the linter

```suggestion
* `name` - (Required) Name of the WAFv2 Web ACL. Optional if `resource_arn` is provided.
* `resource_arn` - (Optional) Name of a resource associated with a WAF.
* `scope` - (Required) Specifies whether this is for an AWS CloudFront distribution or for a regional application. Valid values are `CLOUDFRONT` or `REGIONAL`. To work with CloudFront, you must also specify the region `us-east-1` (N. Virginia) on the AWS provider. Optional if `resource_arn` is provided.
```

internal/service/wafv2/web_acl_data_source.go
```diff
@@ -32,12 +32,18 @@ func DataSourceWebACL() *schema.Resource {
...
+					Computed: true,
+					Optional: true,
 				},
 				"scope": {
-					Type:         schema.TypeString,
-					Required:     true,
-					ValidateFunc: validation.StringInSlice(wafv2.Scope_Values(), false),
+					Type:     schema.TypeString,
+					Computed: true,
+					Optional: true,
```
stumyp, line 46: I think this should stay as required

internal/service/wafv2/web_acl_data_source.go
```diff
@@ -88,3 +103,44 @@ func dataSourceWebACLRead(ctx context.Context, d *schema.ResourceData, meta inte
...
+		ResourceArn: aws.String(arn),
+	}
+
+	output, err := conn.GetWebACLForResourceWithContext(ctx, input)
+	if err != nil {
+		return sdkdiag.AppendErrorf(diags, "reading WAFv2 WebACLs: %s", err)
+	}
+
+	if output == nil || output.WebACL == nil {
+		return sdkdiag.AppendErrorf(diags, "WAFv2 WebACL not found for source: %s", arn)
```
stumyp, line 118: Should we instead just return nothing, if there is no attachment instead of  throwing an error?

stumyp (commented):
Thank you for doing this.
Does this PR add REGIONAL scope handling only? I do not see [GetDistributionConfig](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_GetDistributionConfig.html) call for CLOUDFRONT.

internal/service/wafv2/web_acl_data_source.go
jtyrus: It's checked later. If resource_arn isn't there this becomes required

internal/service/wafv2/web_acl_data_source.go
jtyrus: Went with the existing functionality, if the scope and name aren't there it will also throw an error

internal/service/wafv2/web_acl_data_source.go
stumyp: The scope should be required as this is defining what route the check will go, REGIONAL check is using one API call, and CLOUDFRONT another.

internal/service/wafv2/web_acl_data_source.go
stumyp: Got it, that totally makes sense when the `name` is provided, because then we’re specifically searching for a certain ACL. For `resource`, it might be nicer from a user’s perspective to see an empty value rather than an error if nothing is attached. Also, when we use FMS, attachments don't happen immediately. So, if the same code is tagging the resource and then immediately trying to use data from that source, it could run into issues due to the delay with FMS.

internal/service/wafv2/web_acl_data_source.go
jtyrus: Scope isn't used in the api call though.

internal/service/wafv2/web_acl_data_source.go
jtyrus: The name call is actually less direct than using resource arn. It lists everything by scope then iterates through to filter out the name, this call gets the WAF explicitly by resource_arn input. I think in general if a data source isn't returned you want an error to be thrown at the source, otherwise it will just get thrown later when you try to use an empty datasource attribute.

internal/service/wafv2/web_acl_data_source.go
jtyrus: Ah nevermind, I though the resource arn would take Cloudfront as an input, but that's a different call.

jtyrus:
Looks like the checks don't like calling another service. Would it be better for resource_arn in global scopes to accept the arn of the wafv2 so we don't have to fetch it?

```terraform
data "aws_wafv2_web_acl" "test" {
  resource_arn = aws_cloudfront_distribution.test.web_acl_id
  scope        = "CLOUDFRONT"
}
```

internal/service/wafv2/web_acl_data_source.go
stumyp: Using empty results data sources is expected and can be worked around on terraform level, but if the data source itself fails it blocks the terraform flow completely and will require targeted deployment.

stumyp:
This has been implemented here: https://github.com/hashicorp/terraform-provider-aws/pull/43597
Feel free to close. Thank you!

PR #36847, opened 2024-04-11
New Resource Export

nam054:
### Description

Add new resource Export to service BCM Data Exports.

### Relations

Closes #35343

### Output from Acceptance Testing

```console
> make testacc TESTARGS="-run=TestAccBCMDataExportsExport_" PKG=bcmdataexports
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.21.8 test ./internal/service/bcmdataexports/... -v -count 1 -parallel 20  -run=TestAccBCMDataExportsExport_ -timeout 360m
=== RUN   TestAccBCMDataExportsExport_basic
=== PAUSE TestAccBCMDataExportsExport_basic
=== RUN   TestAccBCMDataExportsExport_disappears
=== PAUSE TestAccBCMDataExportsExport_disappears
=== RUN   TestAccBCMDataExportsExport_tags
=== PAUSE TestAccBCMDataExportsExport_tags
=== RUN   TestAccBCMDataExportsExport_updateTable
=== PAUSE TestAccBCMDataExportsExport_updateTable
=== CONT  TestAccBCMDataExportsExport_basic
=== CONT  TestAccBCMDataExportsExport_disappears
=== CONT  TestAccBCMDataExportsExport_tags
=== CONT  TestAccBCMDataExportsExport_updateTable
--- PASS: TestAccBCMDataExportsExport_disappears (18.74s)
--- PASS: TestAccBCMDataExportsExport_basic (20.92s)
--- PASS: TestAccBCMDataExportsExport_updateTable (32.98s)
--- PASS: TestAccBCMDataExportsExport_tags (43.09s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/bcmdataexports     48.967s
...
```

PR #37144, opened 2024-04-28
Improve import docs for aws_bcmdataexports_export resource

yob:
This resource was added in v5.47.0 and I took it for a test run. We have an existing v2.0 export that was manually created, and I wanted to import it into terraform management.

It wasn't clear from the docs what format the ID should be in, but by reading the provider source I could see it was the export ARN. I thought I'd mention that in the docs to help others.

Interestingly, finding the export ARN from the AWS web console wasn't easy. The URL was the only option I could find.

### Relations

Relates: #36847


johnsonaj (approved):
LGTM 🚀

PR #37239, opened 2024-05-02
awssdkpatch: Fix aliased import substitutions

mattburgess:
`dms` imports its service package with an alias. The current patch template didn't cope with templated imports. gopatch [supports](https://github.com/uber-go/gopatch/blob/main/docs/PatchesInDepth.md#changing-any-import) changing both named and unnamed imports so this converts over to using that generic solution.

On top of that, fix up the other patches so that they match aliased names rather than fully-qualified names.

### Relations

Relates #32976


tools/awssdkpatch/patch.tmpl
```diff
@@ -59,18 +59,22 @@ var x identifier
...
 # to get most things right
 @@
 var x identifier
 @@
+{{- if eq .ImportAlias "" }}
 -{{ .GoV1Package }}.x
+{{- else }}
+-{{ .ImportAlias }}.x
+{{- end }}
 +awstypes.x
```
jar-b, line 78: Is this condition possible, given `ImportAlias` is set to `GoV2Package` when empty during `getProviderData`?

```suggestion
-{{ .ImportAlias }}.x
+awstypes.x
```

tools/awssdkpatch/patch.tmpl
```diff
@@ -59,18 +59,22 @@ var x identifier
...
 # Step 2: Convert AWS SDK v1 (service) to v2
 @@
-var x identifier
+var i, x identifier
 @@
--import "github.com/aws/aws-sdk-go/service/{{ .GoV1Package }}"
-+import "github.com/aws/aws-sdk-go-v2/service/{{ .GoV2Package }}"
-{{ .GoV1Package }}.x
+-import i "github.com/aws/aws-sdk-go/service/{{ .GoV1Package }}"
++import i "github.com/aws/aws-sdk-go-v2/service/{{ .GoV2Package }}"
```
jar-b, line 65:
```suggestion
-import {{ .ImportAlias }} "github.com/aws/aws-sdk-go/service/{{ .GoV1Package }}"
+import {{ .ImportAlias }} "github.com/aws/aws-sdk-go-v2/service/{{ .GoV2Package }}"
```

tools/awssdkpatch/patch.tmpl
```diff
@@ -59,18 +59,22 @@ var x identifier
 
 # Step 2: Convert AWS SDK v1 (service) to v2
 @@
-var x identifier
+var i, x identifier
```
jar-b, line 62: From the best practices section, it sounds like this metavariable should match the import name exactly:
https://github.com/uber-go/gopatch/blob/main/docs/PatchesInDepth.md#best-practices-for-imports

```suggestion
var {{ .ImportAlias }}, x identifier
```

tools/awssdkpatch/main.go
```diff
@@ -102,16 +105,21 @@ func getServiceData(service string) (data.ServiceRecord, error) {
 }
 
 // getPackageData parses AWS SDK V1 package data, collecting inputs for the patch template
-func getPackageData(sd data.ServiceRecord) (TemplateData, error) {
+func getPackageData(sd data.ServiceRecord, importAlias string) (TemplateData, error) {
```
jar-b, line 108: Nit: Rather than accepting this as an argument we can just reference the global `importalias` variable within the function body.

jar-b (commented):
Overall LGTM - just a couple suggestions on naming of metavariables and extra condition checks in the patch template.

tools/awssdkpatch/patch.tmpl
mattburgess: Good catch; that was a remnant of an earlier approach I'd taken.

tools/awssdkpatch/patch.tmpl
mattburgess: So, I originally had the `{{ .ImportAlias }}` variant and thought I'd be clever by making it more generic. I hadn't seen that far down in the docs though; will change it back.

jar-b (approved):
LGTM 🎉

jar-b:
Thanks for your contribution, @mattburgess ! 👍

PR #35991, opened 2024-02-27
[New Resource] : `aws_globalaccelerator_cross_account_attachment`

kathmbeck:
### Description

PR for new resource `aws_globalaccelerator_cross_account_attachment`

### Relations

Closes https://github.com/hashicorp/terraform-provider-aws/issues/34254.

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccGlobalAcceleratorCrossAccountAttachment PKG=globalaccelerator

...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/globalaccelerator/... -v -count 1 -parallel 20 -run='TestAccGlobalAcceleratorCrossAccountAttachment'  -timeout 360m
=== RUN   TestAccGlobalAcceleratorCrossAccountAttachment_basic
=== PAUSE TestAccGlobalAcceleratorCrossAccountAttachment_basic
=== RUN   TestAccGlobalAcceleratorCrossAccountAttachment_principals
=== PAUSE TestAccGlobalAcceleratorCrossAccountAttachment_principals
=== RUN   TestAccGlobalAcceleratorCrossAccountAttachment_resources
=== PAUSE TestAccGlobalAcceleratorCrossAccountAttachment_resources
=== RUN   TestAccGlobalAcceleratorCrossAccountAttachment_disappears
=== PAUSE TestAccGlobalAcceleratorCrossAccountAttachment_disappears
=== CONT  TestAccGlobalAcceleratorCrossAccountAttachment_basic
=== CONT  TestAccGlobalAcceleratorCrossAccountAttachment_resources
=== CONT  TestAccGlobalAcceleratorCrossAccountAttachment_principals
=== CONT  TestAccGlobalAcceleratorCrossAccountAttachment_disappears
--- PASS: TestAccGlobalAcceleratorCrossAccountAttachment_resources (12.14s)
--- PASS: TestAccGlobalAcceleratorCrossAccountAttachment_disappears (13.06s)
--- PASS: TestAccGlobalAcceleratorCrossAccountAttachment_basic (15.49s)
--- PASS: TestAccGlobalAcceleratorCrossAccountAttachment_principals (15.50s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/globalaccelerator  18.282s
```

PR #37269, opened 2024-05-03
feat: add field for cross account support in endpoint groups for AGA

rybit:
### Description

In https://github.com/hashicorp/terraform-provider-aws/pull/35991 we added support for making cross account attachments ([docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/globalaccelerator_cross_account_attachment)). There was one part missing, using them in the endpoint group for the Global Accelerator (AGA). 

This add support for setting this field: https://docs.aws.amazon.com/global-accelerator/latest/api/API_EndpointConfiguration.html

...in progress...

### DIFF
--- a/internal/service/globalaccelerator/endpoint_group.go
+++ b/internal/service/globalaccelerator/endpoint_group.go
@@ -67,6 +67,11 @@ func resourceEndpointGroup() *schema.Resource {
 							Optional:     true,
 							ValidateFunc: validation.IntBetween(0, 255),
 						},
+						"cross_account_attachment_arn": {
+							Type:         schema.TypeString,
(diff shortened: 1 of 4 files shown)


rybit:
What is currently a blocker is that to test this, we need to create two accounts and then link them. 

What we'd need to configure is:
- NLB/ALB in account_one
- cross account attachment in account_two
- the AGA in account_two with attachments to account_one

I'm looking through the tests around AGA and it all seems geared towards operating in one account with one provider. I was wondering if you can point me to examples (if they exist) for leveraging two accounts. If so I can happily get the integration tests to pass.

PR #37287, opened 2024-05-06
Adjust documentation of enable_http_endpoint option of RDS cluster

### LINKED ISSUE #35370: [Docs]: aws_rds_cluster: Incorrect description for enable_http_endpoint Argument
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster#enable_http_endpoint

### Description

The description of that argument says that it is only valid, when engine_mode is set to "serverless", but this is not true anymore. Serverless v2 now also supports Data API for postgresql DBs and neither this module nor the AWS API stops one from doing so, which means code with enable_http_endpoint on engine_mode "provisioned" (= Serverless v2) does work. It is just the documentation which is wrong.

It should say: (Optional) Enable HTTP endpoint (data API). Only valid when engine_mode is set to serverless or (if engine is set to aurora-postgresql) to provisioned.

### References

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.Aurora_Fea_Regions_DB-eng.Feature.Data_API.html

No

katafrakt:
### Description

The documentation for `enable_http_endpoint` option is outdated and no longer true. It says it only works for serverless clusters, but it [also works for provisioned ones](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/data-api.html#data-api.regions).

This also adds a note about other options making this option not being applied. I'm not sure if they are bugs or features, but that's wha I can infer from the code: if one of `snapshot_identifier`, `replication_source_identifier`, `s3_import` is used, `enable_http_endpoint` is not applied. It's a good thing to have stated in the docs, I think after spending more than a week figuring it out.

Closes #35370


justinretzolk (approved):
Looks good to me! Thank you for taking the time to raise this! I linked an issue that this'll close as well, and will get this merged once the checks finish up 🚀

PR #37405, opened 2024-05-09
r/aws_bedrockagent_agent: use configured `prepare_agent` value 

### LINKED ISSUE #37162: [Enhancement]: Need a mechanism to prepare an agent with `aws_bedrockagent*` resources
### Description

It seems that the Terraform resources for Agents for Amazon Bedrock do not [prepare an agent](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_PrepareAgent.html) so that the latest changes are available for testing. How I noticed the issue was that I was creating an agent in Terraform, but my tests failed as the agent wasn't recognizing the tool associated with my action group.

Since preparation is an an action and not a resource, I am not sure what the best way is to model that in Terraform. Here are some options that I can think of:

- An argument such as `prepare` added to the resources that would cause a change to the agent. That would probably include `aws_bedrockagent_agent` and `aws_bedrockagent_agent_action_group`.
- A new resource such as `aws_bedrockagent_preparation` that would prepare the agent on create or update. To trigger it to run, perhaps you need to add dependencies to the above resources. Feels kind of hacky TBH.

Either way, the solution isn't complete without a way to prepare the agent to allow creation of alias/version based on the latest changes.

* `aws_bedrockagent_agent`
* `aws_bedrockagent_agent_action_group`

_No response_

### References

- https://docs.aws.amazon.com/bedrock/latest/userguide/agents-test.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_PrepareAgent.html

None

jar-b:
### Description

This changes removes logic in the read operation which adjusted the `prepare_agent` argument based on whether or not the agent status was `Prepared`. Now the configured `prepare_agent` value is always used, regardless of the status of the agent. If omitted, the default value of `true` is inherited, meaning that by default agents will always be prepared as part of the create or update operation. The `prepare_agent` value is now also set to `true` when an agent is imported.

### Relations

Closes #37162

### References

- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_PrepareAgent.html

```console
% make testacc PKG=bedrockagent TESTS=TestAccBedrockAgentAgent_
TF_ACC=1 go1.22.2 test ./internal/service/bedrockagent/... -v -count 1 -parallel 20 -run='TestAccBedrockAgentAgent_'  -timeout 360m

--- PASS: TestAccBedrockAgentAgent_basic (19.66s)
--- PASS: TestAccBedrockAgentAgent_full (19.67s)
--- PASS: TestAccBedrockAgentAgent_tags (35.51s)
--- PASS: TestAccBedrockAgentAgent_update (37.46s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/bedrockagent       42.493s
```

### DIFF
--- a/internal/service/bedrockagent/agent.go
+++ b/internal/service/bedrockagent/agent.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
@@ -55,7 +56,6 @@ func newAgentResource(context.Context) (resource.ResourceWithConfigure, error) {
 
 type agentResource struct {
 	framework.ResourceWithConfigure
-	framework.WithImportByID
 	framework.WithTimeouts
 }
 
@@ -249,8 +249,6 @@ func (r *agentResource) Read(ctx context.Context, request resource.ReadRequest, 
 		return
 	}
 
-	data.PrepareAgent = types.BoolValue(agent.AgentStatus == awstypes.AgentStatusPrepared)
-
 	response.Diagnostics.Append(response.State.Set(ctx, &data)...)
 }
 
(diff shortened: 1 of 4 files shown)


internal/service/bedrockagent/agent.go
```diff
@@ -366,6 +365,12 @@ func (r *agentResource) Delete(ctx context.Context, request resource.DeleteReque
 	}
 }
 
+func (r *agentResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
```
ewbankkit, line 368: Can now remove embedded `framework.WithImportByID`.

ewbankkit (approved):
LGTM :rocket:.

internal/service/bedrockagent/agent.go
jar-b: Nice catch! I removed this initially and think I left that change on a different branch.

jar-b: Fixed.

PR #37421, opened 2024-05-10
role requirements for token

Deku-shrub:
Add info on the role type needed for the token. Base on gitlab, not sure if this tightly maps for github.

```console

...
```


justinretzolk:
The personal access token permissions can (and one could argue should) be controlled using GitHub's [fine-grained personal access tokens](https://docs.github.com/en/rest/authentication/permissions-required-for-fine-grained-personal-access-tokens), which wouldn't have the named role mentioned here.

With that in mind, the current documentation covers the necessary information, so I'm going to close this pull request for now. Regardless, thank you very much for the contribution, and giving us the opportunity to review this again!

PR #37504, opened 2024-05-14
Don't use attribute literals for serialized test names

gdavison:
### Description

Updates the `semgrep` rules for attribute literals to exclude serialized test names.


ewbankkit (approved):
LGTM :rocket:.

PR #37555, opened 2024-05-15
docs!: remove al1 references from image_id for `aws_cloud9_environment_ec2`

### LINKED ISSUE #37554: [Docs]: Remove Amazon Linux 1 references from `aws_cloud9_environment_ec2`
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloud9_environment_ec2

### Description

Currently the documentation says Amazon Linux 1 is available as an option. But AWS have updated the API documentation for `imageId` where this is no longer an option.

What is yet to confirm is even if its possible to create a Cloud9 still with Amazon Linux 1, or have AWS updated the API where this is no longer a valid option? If this is the case, then I think it would be best to create another issue tracking that, relating that issue to this one. It might still be an option for those running it as Terraform could error out?

In the meantime, since Amazon Linux 1 has reached end of life, I think it would be good to remove the Amazon Linux 1 options from the documentation so it aligns with the AWS API one.

### References

- https://docs.aws.amazon.com/cloud9/latest/APIReference/API_CreateEnvironmentEC2.html#cloud9-CreateEnvironmentEC2-request-imageId

Yes

kieran-lowe:
### Description

As mentioned in #37554, the AWS API docs for [CreateEnvironmentEC2](https://docs.aws.amazon.com/cloud9/latest/APIReference/API_CreateEnvironmentEC2.html#cloud9-CreateEnvironmentEC2-request-imageId) have been updated where `imageId` no longer specifies Amazon Linux 1. Probably because it has reached end of life.

Again as mentioned in the linked issue, it would be worth investigating if create a Cloud9 instance with Amazon Linux 1 via Terraform is even possible or would the API return an error but that should be tracked in another issue/pr.

### Relations

Closes #37554

### References

- https://docs.aws.amazon.com/cloud9/latest/APIReference/API_CreateEnvironmentEC2.html#cloud9-CreateEnvironmentEC2-request-imageId

N/A

### DIFF
--- a/website/docs/r/cloud9_environment_ec2.html.markdown
+++ b/website/docs/r/cloud9_environment_ec2.html.markdown
@@ -74,12 +74,10 @@ This resource supports the following arguments:
 * `name` - (Required) The name of the environment.
 * `instance_type` - (Required) The type of instance to connect to the environment, e.g., `t2.micro`.
 * `image_id` - (Required) The identifier for the Amazon Machine Image (AMI) that's used to create the EC2 instance. Valid values are
-    * `amazonlinux-1-x86_64`
     * `amazonlinux-2-x86_64`
     * `amazonlinux-2023-x86_64`
     * `ubuntu-18.04-x86_64`
     * `ubuntu-22.04-x86_64`
-    * `resolve:ssm:/aws/service/cloud9/amis/amazonlinux-1-x86_64`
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@kieran-lowe Thanks for the contribution :tada: :clap:.

PR #37644, opened 2024-05-22
Generates tagging tests for core EC2 resource types

gdavison:
### Description

Generates tagging tests for core EC2 resource types


internal/generate/tagstests/main.go
```diff
@@ -198,7 +270,7 @@ var testTfTmpl string
 
 // Annotation processing.
 var (
-	annotation = regexache.MustCompile(`^//\s*@([0-9A-Za-z]+)(\((.*)\))?\s*$`)
+	annotation = regexp.MustCompile(`^//\s*@([0-9A-Za-z]+)(\((.*)\))?\s*$`) // nosemgrep:ci.calling-regexp.MustCompile-directly
```
johnsonaj, line 273: is there a difference with using `regexp` vs `regexp2` for this call? Both are imported so I'm curious

gdavison: `regexp2` is needed for negative lookahead, but the API takes a few more parameters for most calls than the standard `regexp` so I used `regexp` for most cases. Maybe overly complicated

johnsonaj (approved):
LGTM 🚀

PR #37713, opened 2024-05-26
r/aws_key_pair: Include public key in read

### LINKED ISSUE #27023: [Enhancement]: aws_key_pair import should fetch public key material
### Description

The documentation for importing an aws_key_pair states:

> The AWS API does not include the public key in the response, so terraform apply will attempt to replace the key pair. There is currently no supported workaround for this limitation.

This might have been true when that was written, but it is no longer the case. At least, it doesn't have to be. 

The request for DescribeKeyPairs includes an "IncludePublicKey" property. If this is set to true, then the the response includes the Public Key. 

This could also make it so that terraform can track if the public key matches the configuration. 

* aws_key_pair

```terraform
N/A, this would just make it so that importing an aws_key_pair would work as expected
```

### References

* AWS API documentation: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeKeyPairs.html
* Go SDK documentation: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#DescribeKeyPairsInput

It does look like this would require using the aws-sdk-go v2 branch. I'm not sure if there is already an issue or pull request tracking migrating to that (it's kind of a difficult thing to search for in github issues), or if it would be feasible to incrementally switch to the new provider. 

I might be willing to implement this, if I had some guidance on how to approach the aws-sdk-go-v2 situation.

No

tmccombs:
This allows import to fetch the existing public key material.

Fixes #27023


riccardomassullo:
any news on this issue?

tmccombs:
I'm still interested in this

PR #37742, opened 2024-05-28
Generate tagging tests for SQS resources

gdavison:
### Description

Generates tagging tests for SQS resources


internal/acctest/const/consts_or_quote_gen.go
```diff
@@ -0,0 +1,60 @@
...
+package acctest
+
+import (
+	"fmt"
+)
+
+// ConstOrQuote returns the constant name for the given attribute if it exists.
+// Otherwise, it returns the attribute quoted. This is intended for use in
+// generated code and templates.
+func ConstOrQuote(constant string) string {
```
YakDriver, line 14: Is this just to save time when compiling and linking? It only takes a long time for the first one and then the rest are cached so they run at normal speed. Diff between generation with/without is only a minute or so. I'm not a fan of proliferating packages unless there's a compelling reason.

internal/generate/tagstests/main.go
```diff
@@ -22,7 +22,7 @@ import (
 	"text/template"
 
 	"github.com/YakDriver/regexache"
-	"github.com/hashicorp/terraform-provider-aws/internal/acctest"
+	acctestconsts "github.com/hashicorp/terraform-provider-aws/internal/acctest/const"
```
YakDriver, line 25: I've gone back and forth on the location of this stuff (in my mind). My thoughts are maybe names is the right place since it's lightweight. Most of the time (~99%), where the constants are used, `acctest` is already imported. The one exception being generators. Or maybe we have a whole new package with attribute and testing consts that just dedicated to constants. Depends on if you're splitting functionality horizontally (constant stuff) or vertically (test/non-test), as it were.

YakDriver (commented):
I left some thoughts. Other than that, it LGTM.

internal/acctest/const/consts_or_quote_gen.go
gdavison: During a `make gen` the first one takes a lot of time, but each additional service package takes an extra 10-15 seconds. So as we add generated tests to more services, it will add a lot of time

internal/generate/tagstests/main.go
gdavison: I like the testing constants being in the `acctest` package, since it scopes them to tests. In this case, the only thing in `acctest/const` is the function `ConstOrQuote`, so maybe the name `const` is wrong for the package.

The generator never uses the contstants in `acctest`, it only needs to map from string values to the constant names. Essentially, `ConstOrQuote` is a "generator helper function", so it isn't directly part of `acctest` but is strongly-associated with it

internal/acctest/const/consts_or_quote_gen.go
gdavison: I've been thinking about this, and because it's generator-related (but not a generator itself) we could name this `internal/acctest/generate` or something similar:

* being a sub-package of `acctest` makes it clear that it is directly associated with `acctest`
* calling it `generate` (or similar) indicates that it's intended for use from generators and shouldn't be used elsewhere
* it's a separate package: `acctest` is used **in** acctests, `acctest/generate` is used by generators to **create** acctests

ewbankkit: Aah I see, it's just for generated test code.

ewbankkit (approved):
LGTM :rocket:.

PR #37759, opened 2024-05-29
fix: change route53_traffic_policy update to create new version

### LINKED ISSUE #27767: [Bug]: Impossible to update traffic policy
### Terraform Core Version

1.3.4, 1.2.9

### AWS Provider Version

4.39.0

### Affected Resource(s)

* aws_route53_traffic_policy
* aws_route53_traffic_policy_instance

### Expected Behavior

We should be able to modify a traffic policy and that it creates a new version, enabling to choose the version to use in the traffic policy instance.

### Actual Behavior

When a change is done to a traffic policy, the AWS API rejects it. It is the consequence that modifying a traffic policy never creates a new version which could be used by the aws_route53_traffic_policy_instance.
Thus it is not possible to "change" the policy because it is still used by the traffic policy instance

```shell
`TrafficPolicyInUse: Policy is in use for at least one instance`
```

data "aws_route53_traffic_policy_document" "example" {
  record_type = "A"
  start_endpoint  = "endpoint-start-YYYY"

  endpoint {
    id     = "endpoint-start-YYYY"
    type   = "cloudfront"
    region = data.aws_region.current.name
    value  = "ZZZZZZZZZ.cloudfront.net"
  }
}

resource "aws_route53_traffic_policy" "test-policy" {
    name     = "playground-policy"

    document = data.aws_route53_traffic_policy_document.example.json
}

resource "aws_route53_traffic_policy_instance" "test-instance" {
  name                   = "my.playground.com"
  traffic_policy_id      = aws_route53_traffic_policy.test-policy.id
  traffic_policy_version = aws_route53_traffic_policy.test-policy.version
  hosted_zone_id         = "XXXXXXXXXXXXXX"
  ttl                    = 360
}

### Steps to Reproduce

terraform apply

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

_No response_

### References

_No response_

_No response_

brainsam:
### Description

This PR changes route53_traffic_policy update method from updating current version to creating new version of policy.

### Relations

Closes #27767

```console
$ make testacc TESTS=TestAccRoute53TrafficPolicy_update PKG=route53
cat: .go-version: No such file or directory
cat: .go-version: No such file or directory
cat: .go-version: No such file or directory
cat: .go-version: No such file or directory
cat: .go-version: No such file or directory
cat: .go-version: No such file or directory
cat: .go-version: No such file or directory
TF_ACC=1 go test ./internal/service/route53/... -v -count 1 -parallel 20 -run='TestAccRoute53TrafficPolicy_update'  -timeout 360m
=== RUN   TestAccRoute53TrafficPolicy_update
=== PAUSE TestAccRoute53TrafficPolicy_update
=== CONT  TestAccRoute53TrafficPolicy_update
--- PASS: TestAccRoute53TrafficPolicy_update (36.29s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/route53	41.859s
...
```

### DIFF
--- a/internal/service/route53/traffic_policy.go
+++ b/internal/service/route53/traffic_policy.go
@@ -63,7 +63,7 @@ func resourceTrafficPolicy() *schema.Resource {
 			"document": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ForceNew:     true,
+				ForceNew:     false,
 				ValidateFunc: validation.StringLenBetween(0, 102400),
 			},
 			names.AttrName: {
--- a/internal/service/route53/traffic_policy.go
+++ b/internal/service/route53/traffic_policy.go
@@ -30,7 +30,7 @@ func resourceTrafficPolicy() *schema.Resource {
 	return &schema.Resource{
 		CreateWithoutTimeout: resourceTrafficPolicyCreate,
 		ReadWithoutTimeout:   resourceTrafficPolicyRead,
-		UpdateWithoutTimeout: resourceTrafficPolicyUpdate,
+		UpdateWithoutTimeout: resourceTrafficPolicyVersionCreate,
 		DeleteWithoutTimeout: resourceTrafficPolicyDelete,
 
 		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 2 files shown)


brainsam:
Root cause of the problem #27767 is that update method in fact updates only `comment` argument and trying to forcibly recreate whole object if `route53_traffic_policy_document` has changed. This operation will fail when a `aws_route53_traffic_policy` used in `aws_route53_traffic_policy_instance`.

Appropriate way - to create new version of traffic policy and then set dependent `route53_traffic_policy_instance` `traffic_policy_version` to the new value

internal/service/route53/traffic_policy.go
```diff
@@ -136,23 +136,20 @@ func resourceTrafficPolicyRead(ctx context.Context, d *schema.ResourceData, meta
...
-	if d.HasChange(names.AttrComment) {
-		input.Comment = aws.String(d.Get(names.AttrComment).(string))
+	input := &route53.CreateTrafficPolicyVersionInput{
+		Document: aws.String(d.Get("document").(string)),
+		Id:       aws.String(d.Id()),
+		Comment:  aws.String(d.Get(names.AttrComment).(string)),
 	}
 
-	_, err := conn.UpdateTrafficPolicyComment(ctx, input)
+	_, err := conn.CreateTrafficPolicyVersion(ctx, input)
```
pwilczynskiclearcode, line 149: Unfortunately this is not the full solution.
It should read the output - esp. new version ID and save it on the resource data so it's marked as modified. 

In other words we need to inform `aws_route53_traffic_policy_instance` the a new version got created.



I tried sth like but it doesn't fully work:
```diff
-       _, err := conn.CreateTrafficPolicyVersion(ctx, input)
+       output, err := conn.CreateTrafficPolicyVersion(ctx, input)
+       d.Set(names.AttrVersion, output.TrafficPolicy.Version)
```

pwilczynskiclearcode: A similar case for aws secret version -> https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/secretsmanager/secret_version.go#L116 
It concats ID with version ID. Writing the code that changes the resource identifier that is backward-compatible might be hard.

PR #37800, opened 2024-06-01
Add new optional attribute validation_timeout to aws_acm_certificate

heldersepu:
### Description
The `validation_timeout` is just exposing what was already done by `certificateDNSValidationAssignmentTimeout` default remains the same, nothing should change from a user persective
Now advanced users with edge cases can increase the timeout or completely bypass the wait by setting it to 0

### Relations
None

Since this is an optional attribute and default remains the same testing remains the same


heldersepu:
@justinretzolk any update on this? 
This should be an easy PR to triage, nothing radically new here, just changing a constant to an optional attribute

PR #37830, opened 2024-06-04
Improve community check

justinretzolk:
### Description

This PR improves the community check workflow by:
- Reducing complexity
- Stops usage of deprecated `::add-mask::` feature
- Following best practice and using an intermediate environment variable for the input

Most importantly, it doesn't functionally change the workflow, so there's no impact to dependent workflows.

N/a, workflows


justinretzolk:
On further thought, I think we may wind up ditching this workflow in the future, so with that in mind, closing this PR since it doesn't really help us in any meaningful way.

PR #37833, opened 2024-06-04
Improved projects workflow

justinretzolk:
### Description

This PR improves our workflow for projects by:

- Combining the relevant parts of `issues.yml` and `pull_request_target.yml`, reducing duplicated config
- Specifying a top-level permissions key that completely limits the permissions of the automatically generated GitHub auth token
- Using intermediary environment variables for any values that come from the issue/pull request to improve security and readability

### Relations

Relates https://github.com/hashicorp/terraform-provider-awscc/pull/1777

N/a, workflow

### DIFF
--- a/.github/workflows/projects.yml
+++ b/.github/workflows/projects.yml
@@ -1,4 +1,5 @@
 name: GitHub Projects Automations
+permissions: {}
 
 on:
   issues:
@@ -11,13 +12,6 @@ on:
       - assigned
       - labeled
       - opened
-
-env:
-  ISSUE_URL: ${{ github.event.issue.html_url || github.event.pull_request.html_url }}
-
-# Any automation with Projects will need a token from the GitHub App,
-# so restrict the permissions of the automated token.
-permissions: {}
 
 jobs:
   community_check:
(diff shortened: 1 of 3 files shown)


.github/workflows/projects.yml
```diff
@@ -0,0 +1,133 @@
...
+      - labeled
+
+  pull_request_target:
+    types:
+      - assigned
+      - labeled
+      - opened
+
+env:
+  ISSUE_URL: ${{ github.event.issue.html_url || github.event.pull_request.html_url }}
```
justinretzolk, line 17: The idea behind separating the `ISSUE_URL` environment variable is that if we were to add additional jobs (for automating future projects), this URL would be common across them, while the environment variables defined at the job level are job-specific.

.github/workflows/projects.yml
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+      - name: Check if Author is Maintainer
+        id: author
+        if: github.event.action == 'opened'
+        env:
+          MAINTAINERS: ${{ secrets.MAINTAINERS }}
+          AUTHOR_LOGIN: ${{ github.event.pull_request.user.login }}
+        run: |
+          echo "is_maintainer=$(echo $MAINTAINERS | base64 --decode | jq --arg u $AUTHOR_LOGIN '. | contains([$u])')" >> "$GITHUB_OUTPUT"
```
justinretzolk, line 56: Doing "community check" with this `bash` one-liner is the same amount of lines as using the reusable workflow (if you don't count the `env` block that could _technically_ be eliminated, to the detriment of readability, in my opinion).

justinretzolk:
Closing this after further research and consideration as to how I think we'll organize workflows going forward.

Issue #35261, opened 2024-01-12
[New Resource]: Add Pinpoint FCM channel resource


mike-ntk:
### Description

Existing resource `aws_pinpoint_gcm_channel` uses Google Cloud Messaging API which is going to be deprecated by **6/20/2024**

"If you are an existing user of the legacy HTTP or XMPP APIs (deprecated on 6/20/2023), you must migrate to the latest Firebase Cloud Messaging API (HTTP v1) by 6/20/2024"

Official migration manual https://firebase.google.com/docs/cloud-messaging/migrate-v1

### Requested Resource(s) and/or Data Source(s)
```
aws_pinpoint_fcm_channel
```

### References

- https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-channels-gcm.html

### Would you like to implement a fix?

No

breathingdust:
Hi @mike-ntk! The provider doesn't directly integrate with the Google Cloud Messaging API, but via the AWS Pinpoint API. It looks like the same Pinpoint GCM API endpoint is used to allow FCM configuration with the addition of the two missing attributes `ServiceJson` and `DefaultAuthenticationMethod` so I think the best way forward here is to add those attributes rather than create a new resource.

mike-ntk:
Hi @breathingdust! Thanks for your input. 
Main idea behind the new resource was to match current AWS terminology, but the proposed solution with the existing resource would work just fine. 
![image](https://github.com/hashicorp/terraform-provider-aws/assets/119694137/0537d627-f623-4afb-84d8-947917d65048)

mfournier:
There is the same issue with `aws_sns_platform_application`: #35721

BenPhegan:
Just wondering if it is expected that the same fix provided as part of https://github.com/hashicorp/terraform-provider-aws/issues/35721 (ie just sending the json token file as the credentials) is expected to work for this resource as well?  We are about a week away from Google deprecating FCM and the previous approach failing.  

When we apply the same fix associated with the `aws_sns_platform_application` by sending the json token file contents to the `aws_pinpoint_gcm_channel` via the `api_key` we get a failure:

```
 Error: putting Pinpoint GCM Channel for application asdfasdfasdfasdfasdfasdfasdfasdfsasdf: BadRequestException: FCM returned 401 Unauthorized
 {
   RespMetadata: {
     StatusCode: 400,
[...]

PR #37976, opened 2024-06-14
update gcm_channel upsert schema

LiamAttClarke:
### Relations

Closes #0000

```console

...
```

https://github.com/hashicorp/terraform-provider-aws/issues/35261


BenPhegan:
I believe that there is an updated deadline for migration of resources to use a `service_json` until late this month (20th July).  What blockers are there preventing this change going in?  It would be great to be able to use Terraform to affect this change prior to the deadline.

PR #37019, opened 2024-04-20
f-aws_ecs_service: support for EBS

nikhil-goenka:
### Relations

Closes #35279

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccECSTaskDefinition_configuredAtLaunch PKG=ecs
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.22.2 test ./internal/service/ecs/... -v -count 1 -parallel 20 -run='TestAccECSTaskDefinition_configuredAtLaunch'  -timeout 360m
=== RUN   TestAccECSTaskDefinition_configuredAtLaunch
=== PAUSE TestAccECSTaskDefinition_configuredAtLaunch
=== CONT  TestAccECSTaskDefinition_configuredAtLaunch
--- PASS: TestAccECSTaskDefinition_configuredAtLaunch (54.53s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ecs	126.627s
% make testacc TESTS=TestAccECSTaskDefinition_ PKG=ecs                    
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.22.2 test ./internal/service/ecs/... -v -count 1 -parallel 20 -run='TestAccECSTaskDefinition_'  -timeout 360m
=== RUN   TestAccECSTaskDefinition_basic
=== PAUSE TestAccECSTaskDefinition_basic
=== RUN   TestAccECSTaskDefinition_scratchVolume
=== PAUSE TestAccECSTaskDefinition_scratchVolume
=== RUN   TestAccECSTaskDefinition_configuredAtLaunch
=== PAUSE TestAccECSTaskDefinition_configuredAtLaunch
=== RUN   TestAccECSTaskDefinition_DockerVolume_basic
=== PAUSE TestAccECSTaskDefinition_DockerVolume_basic
=== RUN   TestAccECSTaskDefinition_DockerVolume_minimal
=== PAUSE TestAccECSTaskDefinition_DockerVolume_minimal
=== RUN   TestAccECSTaskDefinition_runtimePlatform
=== PAUSE TestAccECSTaskDefinition_runtimePlatform
=== RUN   TestAccECSTaskDefinition_Fargate_runtimePlatform
=== PAUSE TestAccECSTaskDefinition_Fargate_runtimePlatform
=== RUN   TestAccECSTaskDefinition_Fargate_runtimePlatformWithoutArch
=== PAUSE TestAccECSTaskDefinition_Fargate_runtimePlatformWithoutArch
[...]

PR #38109, opened 2024-06-25
Bug fix: fixes throughput type for managed ebs volume

jnsaff:
### Description
When specifying throughput as an attribute for the block terraform plan fails with the following error message:

```
│ Error: expected type of volume_configuration.0.managed_ebs_volume.0.throughput to be integer
```

This is an attempt to fix this issue.

### Relations
Relates: #37019

### References

Type definition in aws-sdk-go-v2: 
https://github.com/aws/aws-sdk-go-v2/blob/main/service/ecs/types/types.go#L4174


ewbankkit (approved):
LGTM :rocket:.

PR #38159, opened 2024-06-27
ci: Sets the parallelism for AMP acceptance tests to 10

gdavison:
### Description

AMP has a maximum of 10 scrapers per region, and this quota cannot be adjusted.


YakDriver (approved):
Assuming tests pass, LGTM 🎉

PR #38192, opened 2024-06-28
autoflex: Adds support for Expanding union types

gdavison:
### Description

Adds the `fwflex.Expander` interface to allow model types to specify how they expand. Used for "tagged union types" implemented in the AWS SDK as an interface type.

Implemented in

* `aws_m2_application`
* `aws_m2_environment`
* `aws_bedrockagent_agent_action_group`

PR #38258, opened 2024-07-04
Adds validation of Diagnostics and logging to AutoFlEx tests

gdavison:
### Description

Currently the tests for AutoFlEx check simply for the presence of an error, not the contents. Tests do not capture logs.

This PR replaces the boolean `WantErr` field with the `diag.Diagnostics` `expectedDiags` field to validate the error messages. It also adds the field `expectedLogLines` to capture logs.

In addition to explicitly expecting error messages and log entries, this sets the baseline for what is currently output so that we can update them as we define the expected behaviour.

### Relations

Relates #38192

### DIFF
--- a/internal/framework/flex/auto_expand_test.go
+++ b/internal/framework/flex/auto_expand_test.go
@@ -4,7 +4,9 @@
 package flex
 
 import (
+	"bytes"
 	"context"
+	"reflect"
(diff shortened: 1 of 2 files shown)


ewbankkit (approved):
LGTM :rocket:.

Issue #36938, opened 2024-04-16
[Docs]: S3 Object Lock no longer needs a token field


tanelso2:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object_lock_configuration#token

### Description

Hi there,

The `aws_s3_bucket_object_lock_configuration` object type has a field called `token` that it claims is needed to enable Object Lock on an existing S3 bucket.

This is no longer the case. As of November 2023, AWS allows you to enable Object Lock without specifying a token, and the current versions of the Object Lock documentation makes no reference to the token field.

I just tested this out myself with a quick demo. I was able to add Object Lock to an existing bucket without a token.

We should update the documentation to make it clear that the token is deprecated and no longer needed. The current documentation implies the opposite.

### References

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-configure.html
https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-s3-enabling-object-lock-buckets/

Relates https://github.com/hashicorp/terraform-provider-aws/pull/36530.

### Would you like to implement a fix?

Yes

Jyots6914:
I will work on it and I will raise PR

tanelso2:
Fantastic, thanks everyone who worked on this!

PR #38298, opened 2024-07-09
Update s3_bucket_object_lock_configuration.html.markdown

Jyots6914:
Referred Doc:  https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-configure.html https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-s3-enabling-object-lock-buckets/

and Tested it for existing S3 bucket , no token is getting generated we just need to enable version and then lock

### Description
As per above document no token is needed to enable S3 Object Lock for existing buckets. We can enable versioning and then enable the lock.

### Relations
This will be closing the below issue 
Closes :  https://github.com/hashicorp/terraform-provider-aws/issues/36938

Go to the AWS Console --> Go to the S3 Bucket --> Go to Properties --> Enable versioning first --> Enable to Lock 
Step1
Step2
Step3
Step4


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@Jyots6914 Thanks for the contribution :tada: :clap:.

PR #38349, opened 2024-07-12
add explicit reference to the type of the expected value

lpmi-13:
### Description
I was confused as to whether this should be a string (eg, `enabled`) or a boolean. It was easy enough to find out, but it would have been nicer if the docs had been more explicit.

This commit makes the boolean type of the value explicit.

### References
current docs
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/docdb_cluster#deletion_protection


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@lpmi-13 Thanks for the contribution :tada: :clap:.

PR #38386, opened 2024-07-16
Updates to `tfsdk2fw`

gdavison:
### Description

Makes the following updates to `tfsdk2fw`:

* Uses the `stringdefault` package for string defaults
* Adds support for defaults for other primitive types
* Sets `Computed` for attributes with default values
* No longer adds the non-existent `SetMigratedFromPluginSDK` to resource initialization


ewbankkit (approved):
LGTM :rocket:.

PR #38410, opened 2024-07-18
Added `AMPLIFY_MANAGED` as the default value for `aws_amplify_domain_association.certificate_settings.type`

### LINKED ISSUE #38387: [Enhancement]: aws_amplify_domain_association certificate_settings AMPLIFY_MANAGED should be default
### Description

the newly added (>=5.57.0) certificate_settings attribute should default to AMPLIFY_MANAGED rather than null. Existing resources appear to result in plans where the certificate settings are being removed, despite amplify managed domains being the default up to this point. The new certificate setting block should respect this an only be required to override the default settings.

aws_amplify_domain_association

_No response_

### References

_No response_

None

aayushsss1:
### Description

Added `AMPLIFY_MANAGED` as the default value for the type attribute. Without the default value, the resource will be replaced on the next plan/apply. 

The issue occurs when users update the provider version to 5.57 or above, without making any changes to the resource.

### Relations

Closes #38387

### References

- https://docs.aws.amazon.com/amplify/latest/APIReference/API_CertificateSettings.html 

```console

...
```

### DIFF
--- a/internal/service/amplify/domain_association.go
+++ b/internal/service/amplify/domain_association.go
@@ -62,6 +62,7 @@ func resourceDomainAssociation() *schema.Resource {
 						names.AttrType: {
 							Type:             schema.TypeString,
 							Required:         true,
+							Default:          "AMPLIFY_MANAGED",
 							ValidateDiagFunc: enum.Validate[types.CertificateType](),
 						},
 						"custom_certificate_arn": {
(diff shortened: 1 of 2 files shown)


internal/service/amplify/domain_association.go
```diff
@@ -62,6 +62,7 @@ func resourceDomainAssociation() *schema.Resource {
 						names.AttrType: {
 							Type:             schema.TypeString,
 							Required:         true,
```
taraspos, line 64: This won't work unless type changed to `Optional: true`.

I fixed the same issue while working on another bug in my PR:
- https://github.com/hashicorp/terraform-provider-aws/pull/40351

gdavison:
Thanks for taking the time to submit this PR, @aayushsss1. Assigning a default value to `certificate_settings.type` is unfortunately not sufficient to address the issue. I've created #40589 instead.

PR #38424, opened 2024-07-18
Update `issue_comment` workflow to use new community check

justinretzolk:
### Description

Similar to other recent PRs, moves the workflow to use the new composite action style community check. Also introduces:

- `permissions` block to limit token permissions
- Only runs the label removing job if the labels exist on the issue

N/a, Actions


breathingdust (approved):
LGTM! 🚀

PR #38527, opened 2024-07-24
fix(ec2_instance): adding placement_group_id to aws_instance resource for shared placement groups from different account

jracollins:
Recently when sharing a placement group to another AWS account, and trying to launch EC2s via `aws_instance` or an ASG with a launch template with the placement group name provided, I ran into the following error:

```
InvalidPlacementGroup.Unkown: The placement group 'xxxx' is unknown
```

I could create manually in the UI using the same shared placement group causing the error, and after opening a support request it was confirmed from AWS that referencing the placement group by name is only possible when used in the account that created the placement group.  

A simple cut down example of how I encountered it is below: 

```hcl
resource "aws_placement_group" "local" {
  name     = "test"
  strategy = "cluster"
}

  ami           = var.ami
  instance_type = var.instance_type

  placement_group = "shared_placement_group_name"
  # placement_group = aws_placement_group.local.name # works  

  launch_template {
    id      = var.launch_template_id
    version = var.launch_template_id.version
  }
}
```

I have looked through the provider code and can see it only references by name, and have made this PR that should resolve it for the `aws_instance` resource use case. 

However the AWS Go SDK does not actually support the placement group ID field in all places where it would be needed (ie launch templates, spot requests) so it is not as simple or easy as the small changes I've made here. 

- SpotPlacement: https://github.com/aws/aws-sdk-go-v2/blob/service/ec2/v1.171.0/service/ec2/types/types.go#L16618
vs
- Placement: https://github.com/aws/aws-sdk-go-v2/blob/service/ec2/v1.171.0/service/ec2/types/types.go#L12904 

I have left a commented section in this PR where the error happens in a spot request within this single resource. 

It would be great if it could be merged with just this small change and adding  `instance_market_options` to the `ConflictsWith` to protect against any errors until the AWS Go SDK version gets bumped/supports it for all the other resources.


ewbankkit:
@jracollins Thanks for the contribution :tada: :clap:.

PR #38562, opened 2024-07-26
Use generated token for PR labeler

justinretzolk:
### Description

The labeler job hasn't been working quite right (for example [here](https://github.com/hashicorp/terraform-provider-aws/actions/runs/10111695655/job/27964156924), where `needs-triage`, and `external-maintainer` should have been added, but those steps were skipped.

This seems to be caused by the use of the GitHub App token, which will [trigger a `labeled` event](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow), which doesn't happen with the generated token. 

The related workflow in the [AWSCC repository](https://github.com/hashicorp/terraform-provider-awscc/actions/runs/10103293975/job/27940403639) is nearly identical, but uses the generated token and is working. Thus far, that's the only difference I've noticed. When testing locally with `act`, the GitHub App token portion is omitted, and the workflow works fine, which seems to back up the theory that this is the issue.

Thank you to @DanielRieske for bringing this to my attention!

N/a, workflows.

### DIFF
--- a/.github/workflows/pull_request_target.yml
+++ b/.github/workflows/pull_request_target.yml
@@ -16,14 +16,11 @@ jobs:
   labels:
     name: Labelers
     runs-on: ubuntu-latest
-    steps:
-      - name: Generate GitHub App Token
-        id: token
-        uses: actions/create-github-app-token@31c86eb3b33c9b601a1f60f98dcbfd1d70f379b4 # v1.10.3
-        with:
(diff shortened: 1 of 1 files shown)


justinretzolk:
I've done a bit more investigating and testing and realized that this is indeed _not_ the issue. Closing this PR to raise a proper fix.

PR #38608, opened 2024-07-31
Add warning to s3_bucket_notification doc

zisom-hc:
### Description

Adding more insight to the warning note of this documentation. It'd be a benefit to inform users of the results they would see if they were to attempt to use this resource against an S3 bucket with existing notification configurations.

### Relations

Closes #0000

```console

...
```


jar-b:
Thanks for the addition, @zisom-hc! 👍

PR #38616, opened 2024-07-31
[Docs] Improve route_53_zone and route_53_zone_association examples

cfernhout:
### Description
I was using the example code and I believe it could be a bit more clearer. 
1. Having the notes under the description and not in-line with the examples, like often the case in other docs in this repo. 
2. Making it more explicit what makes an hosted zone a _private_ hosted zone.
3. Adding an example for cross-account zone association.

I am aware this is opinionated, so feel free to disagree. Let's work together to make the best of it.

### Relations
NA

### References
NA

NA


cfernhout:
Sorry, there are mistakes in this PR.

PR #38643, opened 2024-08-01
internal/framework/flex: add field name prefix option 

jar-b:
### Description

This AutoFlex option will allow a field name prefix to be passed into the `Flatten` and `Expand` functions to improve fuzzy matching between Terraform and AWS data structures. Replaces the previous method of providing the prefix via a context key.

Also adjusts the entry point of the ignored field names option from methods on the option struct to exported constructor functions. This should simplify the syntax when used in service packages.

### Relations

Closes #37266

### Output from Acceptance Testing

```console
% make testacc PKG=drs TESTS=TestAccDRSReplicationConfigurationTemplate_
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.22.5 test ./internal/service/drs/... -v -count 1 -parallel 20 -run='TestAccDRSReplicationConfigurationTemplate_'  -timeout 360m
=== RUN   TestAccDRSReplicationConfigurationTemplate_serial
=== PAUSE TestAccDRSReplicationConfigurationTemplate_serial
=== CONT  TestAccDRSReplicationConfigurationTemplate_serial
=== RUN   TestAccDRSReplicationConfigurationTemplate_serial/disappears
=== RUN   TestAccDRSReplicationConfigurationTemplate_serial/basic
--- PASS: TestAccDRSReplicationConfigurationTemplate_serial (180.58s)
    --- PASS: TestAccDRSReplicationConfigurationTemplate_serial/disappears (89.44s)
    --- PASS: TestAccDRSReplicationConfigurationTemplate_serial/basic (91.14s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/drs        186.422s
```

```console
% make testacc PKG=lexv2models TESTS=TestAccLexV2ModelsIntent_
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.22.5 test ./internal/service/lexv2models/... -v -count 1 -parallel 20 -run='TestAccLexV2ModelsIntent_'  -timeout 360m

--- PASS: TestAccLexV2ModelsIntent_disappears (45.17s)
--- PASS: TestAccLexV2ModelsIntent_basic (47.25s)
--- PASS: TestAccLexV2ModelsIntent_updateSampleUtterance (61.25s)
--- PASS: TestAccLexV2ModelsIntent_updateDialogCodeHook (61.58s)
--- PASS: TestAccLexV2ModelsIntent_updateInputContext (61.59s)
[...]

PR #38649, opened 2024-08-01
Td autoflex options resource name prefix

gdavison:
### Description

Moves resource name prefix to `AutoFlexOptions`


jar-b:
@gdavison - #38643 implemented this option along with introducing some constructors. The documentation updates here aren't covered though. Should we consolidate?

ewbankkit (approved):
LGTM :rocket:.

PR #38691, opened 2024-08-05
Various sweeper fixes

ewbankkit:
### Description

* Make AWS SDK for Go v1 and v2 skip checks equivalent
* Actually sweep FSx resources
* Prevent crash in DataZone sweeper


jar-b (approved):
LGTM 🚀

PR #37780, opened 2024-05-30
dms: Migrate to AWS SDK v2

mattburgess:
### Relations

Closes #36141

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccXXX PKG=ec2

...
```

PR #38694, opened 2024-08-05
internal/provider(test): remove DMS from unusual endpoint tests

jar-b:
### Description

DMS has been migrated to AWS SDK for Go V2, and thus the corresponding `DMSConn` method to create a V1 client no longer exists. The generic `TestAccProvider_endpoints` test effectively ignores services which use V2 AWS SDK clients with the following conditional in the test check function (`serviceConn` will always return a method name ending in `Conn`):

```go
			methodName := serviceConn(serviceKey)
			method := reflect.ValueOf(providerClient).MethodByName(methodName)
			if !method.IsValid() {
				continue
			}
```

Because this generic test check ignores V2 clients, the DMS case for the unusual endpoints tests has been removed entirely rather than including new conditional logic which would effectively just skip the case before executing any real checks.

### Relations

Relates #37780

```console
% TF_ACC=1 go test -count=1 ./internal/provider/... -run="TestAccProvider_unusualEndpoints"
?       github.com/hashicorp/terraform-provider-aws/internal/provider/fwprovider        [no test files]
ok      github.com/hashicorp/terraform-provider-aws/internal/provider   23.263s
```

### DIFF
--- a/internal/provider/provider_acc_test.go
+++ b/internal/provider/provider_acc_test.go
@@ -217,8 +217,7 @@ func TestAccProvider_unusualEndpoints(t *testing.T) {
 	ctx := acctest.Context(t)
 	var provider *schema.Provider
 	unusual1 := unusualEndpoint{"es", "elasticsearch", "http://notarealendpoint"}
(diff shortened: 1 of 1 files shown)


YakDriver (approved):
LGTM 🎉

PR #38720, opened 2024-08-06
sweeper/aws_directory_service_directory:Addresses API rate limiting when sweeping many directories

gdavison:
### Description

Sweeping a large number of directories, e.g. 20 would hit rate limits that prevented any directories from being swept. Adds some delay, which addresses the rate limit to some extent.


ewbankkit (approved):
LGTM :rocket:.

PR #38803, opened 2024-08-09
Map Block Key fields should have meaningful names

gdavison:
### Description

While AutoFlex currently requires the struct field name of a Map Block Key to be `MapBlockKey`, the attribute name in the schema should be something meaningful rather than `map_block_key`. Adds a Semgrep rule to catch this and allows existing `map_block_key` attributes to ignore the rule.

### DIFF
--- a/.ci/semgrep/framework/map_block_key.go
+++ b/.ci/semgrep/framework/map_block_key.go
@@ -1,0 +1,23 @@
+package main
+
+import (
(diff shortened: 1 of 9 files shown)


jar-b (approved):
LGTM 🚀

ewbankkit (approved):
LGTM :rocket:.

breathingdust (approved):
LGTM! 🚀

PR #38852, opened 2024-08-13
resource/aws_appfabric_app_authorization: Convert `credentialModel` to `fwflex.Expander`

gdavison:
### Description

Converts `credentialModel` to `fwflex.Expander`. Prevents error log when expanding


jar-b (approved):
LGTM 🚀

PR #38884, opened 2024-08-15
ci: remove library_versions workflow

jar-b:
### Description

This workflow was intended to check for the presence of net-new usage of the AWS SDK for Go V1. We have now migrated almost 80% of services off of V1 of the SDK and this check, while well intended, often misfires due to the way in which merges are applied on pull requests before producing a diff. For these reasons this check has likely outlived it's benefit and can be removed to reduce unnecessary pull request comment noise.


ewbankkit (approved):
LGTM :rocket:.

breathingdust (approved):
LGTM 🚀

PR #38903, opened 2024-08-16
Correct case name in new resource dev guide

mungojam:
### Description

The guide was making a distinction between different cases but used the same name for both. It looks like it intended to use the actual case name for the `snake_case` example


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@mungojam Thanks for the contribution :tada: :clap:.

PR #38904, opened 2024-08-16
Use `attr.Value` `Value<type>Pointer` functions instead of `aws.<type>(v.Value<type>())`

gdavison:
### Description

Take `attr.Value` pointers directly instead of address of value. i.e. instead of

```go
aws.String(data.Alias.ValueString())
```

use

```go
data.Alias.ValueStringPointer 
```

In addition to looking cleaner, this will reduce the number of small memory allocations.

### DIFF
--- a/.ci/semgrep/framework/flex.yml
+++ b/.ci/semgrep/framework/flex.yml
@@ -1,15 +1,13 @@
 rules:
   - id: flex-type-from-framework
     languages: [go]
-    message: Prefer `flex.<type>FromFramework` to `aws.<type>(x.Value<type>())`
-    paths:
-      include:
-        - internal/service/cognitoidp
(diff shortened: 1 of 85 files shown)


ewbankkit:
In those cases where the assignment isn't inside `if ...IsNull() {}`, these aren't strictly equivalent -- using `aws.ToString(s.ValueString())` returns a pointer to `""` when `s` is Null whereas `s.ValueStringPointer()` returns `nil`. However an eyeballing of the changes suggests there is no practical difference.
There's also `internal/framework/flex.StringFromFramework()` which does the same thing -- I think that is from before the `ValueStringPointer()` method existed.

gdavison:
This is for the take-a-pointer functions, not the dereference-a-pointer functions, e.g.

```go
func String(v string) *string {
	return &v
}
```

ewbankkit (approved):
LGTM :rocket:.

PR #38921, opened 2024-08-19
r/aws_ses_configuration_set: fix crash when `reputation_metrics_enabled` is set

jar-b:
### Description

Migration of this service to AWS SDK for Go V2 unintentionally introduced an assertion to a pointer to bool, rather than a bool.

### Relations

Closes #38918
Relates #38568

### Output from Acceptance Testing

```console
% make testacc PKG=ses TESTS=TestAccSESConfigurationSet_
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.22.6 test ./internal/service/ses/... -v -count 1 -parallel 20 -run='TestAccSESConfigurationSet_'  -timeout 360m

--- PASS: TestAccSESConfigurationSet_disappears (13.24s)
--- PASS: TestAccSESConfigurationSet_deliveryOptions (15.27s)
--- PASS: TestAccSESConfigurationSet_basic (15.29s)
--- PASS: TestAccSESConfigurationSet_emptyDeliveryOptions (20.05s)
--- PASS: TestAccSESConfigurationSet_sendingEnabled (29.13s)
--- PASS: TestAccSESConfigurationSet_Update_emptyDeliveryOptions (35.44s)
--- PASS: TestAccSESConfigurationSet_reputationMetricsEnabled (37.59s)
--- PASS: TestAccSESConfigurationSet_Update_deliveryOptions (37.92s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ses        43.956s
```

PR #38945, opened 2024-08-20
chore: prepare changelog for v5.63.1

jar-b:
### Description

Preparing a patch release to address an SES configuration set regression which triggers a crash (#38921).


ewbankkit (approved):
LGTM :rocket:.

PR #38993, opened 2024-08-22
Update ECS cluster name validation

gregops312:
Previous regex was not complete and allowed for apply time failures

### Description

Updating the ECS `validateClusterName` which contained and incomplete regex allowing for many cases with plans that wouldn't fail as expected

### Relations

n/a

### References

n/a

n/a

### DIFF
--- a/.changelog/38993.txt
+++ b/.changelog/38993.txt
@@ -1,0 +1,7 @@
+```release-note:bug
+resource/aws_ecs_cluster: Fix validation error with `name` attribute.
+```
+
+```release-note:bug
+resource/aws_ecs_cluster_capacity_providers: Fix validation error with `name` attribute.
+```
--- a/internal/service/ecs/validate.go
+++ b/internal/service/ecs/validate.go
@@ -14,7 +14,7 @@ func validateClusterName(v interface{}, k string) (ws []string, errors []error) 
 	return validation.All(
 		validation.StringLenBetween(1, 255),
 		validation.StringMatch(
-			regexache.MustCompile("^[a-zA-Z0-9-_]+$"),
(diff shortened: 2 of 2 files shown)


gregops312:
So I know there's a lot that goes on in this provider, and while this is pretty small, it like many other bugs can have consequences to deployments and availability of resources. I would think fixing bugs would be fairly high priority. 

That said I've read the contributing docs and am still curious what is a normal timeframe for getting PRs reviewed?

gdavison (approved):
Thanks for submitting this, @gregops312.

I've made a couple adjustments:

* I've removed the `.` that was in the regex, since that's not valid for ECS cluster names
* We have an [order that the character classes should appear in](https://github.com/hashicorp/terraform-provider-aws/blob/main/docs/regular-expressions.md), since we cache compiled regexes. I've re-ordered the regexp.

gregops312:
> Thanks for submitting this, @gregops312.
> 
> I've made a couple adjustments:
> 
> * I've removed the `.` that was in the regex, since that's not valid for ECS cluster names
> * We have an [order that the character classes should appear in](https://github.com/hashicorp/terraform-provider-aws/blob/main/docs/regular-expressions.md), since we cache compiled regexes. I've re-ordered the regexp.

@gdavison Thank you very much. I'll keep those things in mind should I come across anything else in the future. I also see you added a `changelog` file. I didn't see that in the contributing docs, is that something I should do when making a change or is that something that is reserved for maintainers?

gdavison:
Yes, there is some documentation about adding a changelog entry (https://hashicorp.github.io/terraform-provider-aws/changelog-process/), but we have recently reorganized the contributor guide, so it may have been more buried when you created the PR originally.

Issue #34189, opened 2023-10-31
[Bug]: ASG Instance Refresh Rollback Impaired behavior


adeeb-ma:
### Terraform Core Version

1.0.5

### AWS Provider Version

5.32.1

### Affected Resource(s)

aws_autoscaling_group

### Expected Behavior

Applying the Autoscaling group with Instance Refresh enabled (i.e. changing launch template version, for example) would trigger the instance refresh to deploy the new configurations, **but shouldn't update the configuration beforehand**.

In the AWS console, you would see under the "details" tab the old Launch Template Version, while in the Instance Refresh tag it would run with the target configuration being the newer launch template version.

This would imply that in case the new launch template version (2) fails on healthcheck for a while, it would rollback to the previous version (1).

### Actual Behavior

Running apply would update the launch template version for the ASG, **AND THEN** triggers instance refresh with target configuration being the latest launch template version.

This imposes that should the newer launch template version (2) fail on healthcheck for a while, it would "roll back" to the same newer launch template version (2), making the ASG stuck in a perpetual rollback replacing the unhealthy launch template over and over.

### Terraform Configuration Files

```hcl
data "aws_ami" "ami" {
  most_recent = true
  filter {
    name   = "name"
    values = ["ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*"]
  }
  filter {
    name   = "virtualization-type"
    values = ["hvm"]
  }
  owners = ["099720109477"] # Canonical
}

resource "aws_launch_template" "sample" {
  name = "sample-lt"

  image_id                             = data.aws_ami.ami.id
  instance_initiated_shutdown_behavior = "terminate"
  instance_type                        = "t3.small"
# user_data = base64encode("#!/bin/bash\nsudo ifconfig eth0 down") # This is to purposefully fail the EC2 health check, to test out the Instance Refresh Rollback
}

resource "aws_autoscaling_group" "sample" {
  name  = "sample-asg"

  launch_template {
    id      = aws_launch_template.sample.id
    version = aws_launch_template.sample.latest_version
  }
  min_size                  = 1
  max_size                  = 1
  desired_capacity          = 1
  health_check_type         = "EC2" # Can be ELB as well
  health_check_grace_period = 20
  wait_for_capacity_timeout = "5m"

  instance_refresh {
    strategy = "Rolling"
    # Preferences can be anything, as long as auto_rollback = true
    preferences {
      auto_rollback = true
[...]

PR #39007, opened 2024-08-23
autoscaling: Fix auto-rollback for instance refresh

arianvp:
### Description

The auto-rollback behaviour for instance refresh relies on the fact that the desired configuration is *not* modified until the instance refresh finishes. If the instance refresh fails, it uses the values from the Auto Scaling Group for the rollback.

So when we do an Instance Refresh we should not use UpdateAutoScalingGroup for modify the LaunchTemplate or MixedInstancesPolicy.

Instead we should always rely on the DesiredConfiguration parameter of StartInstanceRefresh

StartInstanceRefresh will modify the Auto Scaling Group for us on success or else rollback (if enabled) to the previous version.

If Auto Rollback is disabled; and the instance will simply fail, causing the apply to fail.

### Relations

Closres #34189 

```console

...
```

### DIFF
--- a/internal/service/autoscaling/group.go
+++ b/internal/service/autoscaling/group.go
@@ -1360,6 +1360,18 @@ func resourceGroupUpdate(ctx context.Context, d *schema.ResourceData, meta inter
 	var shouldWaitForCapacity bool
 	var shouldRefreshInstances bool
 
+	// When instance refresh is enabled, the way we update the launch_template
+	// or mixed_instances_policy fields is different.  In this case,
+	// StartInstanceRefresh is passed the new launch_template or
+	// mixed_instances_policy, and the API will automatically update the Auto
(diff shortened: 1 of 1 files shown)


arianvp:
I didn't test this yet.  But I think the change like this makes sense.   The function is huge though.   Suggestions on how to restructure the code to make this change more obvious would be appreciated

arianvp:
>  This resource does not wait for the instance refresh to complete.

This might be problematic. As we can't observe the change to the autoscaling group after the instance refresh completed. Breaking my entire idea.

I think how instance refresh is currently implemented is kind of broken. I'm not sure if we can make it work :/

arianvp:
I guess what we can do is   read any in progress instance refreshes in the `read` operation and use that as the "current" state for `LaunchTemplate` and `MixedInstancesPolicy`.

PR #39015, opened 2024-08-23
chore(ci): notify and prioritize `crash` labeled issues

jar-b:
### Description

This change extends the existing workflow which prioritized issues labeled `regression`, including sending a slack notification. The workflow has been named more generically and completes the same steps for issues labeled `crash`.

### DIFF
--- a/.github/workflows/label-notifications.yml
+++ b/.github/workflows/label-notifications.yml
@@ -24,7 +24,6 @@ jobs:
         uses: slackapi/slack-github-action@70cd7be8e40a46e8b0eced40b0de447bdb42f68e # v1.26.0
         env:
           ISSUE_TITLE: ${{ toJSON(github.event.issue.title || github.event.pull_request.title) }}
-          LABEL_NAME: ${{ github.event.label.name }}
           SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/label-notifications.yml
```diff
@@ -34,7 +35,7 @@ jobs:
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                    "text": ":warning: The following has been labeled as a regression:"
+                    "text": ":warning: The following has been labeled as a $LABEL_NAME:"
```
justinretzolk, line 38: I'm fairly confident that `$LABEL_NAME` would work here too, but using `${{ env.LABEL_NAME }}` keeps consistency with line 45 below.

```suggestion
                    "text": ":warning: The following has been labeled as a ${{ env.LABEL_NAME }}:"
```

justinretzolk (approved):
LGTM! One small suggestion, but it's not critical

.github/workflows/label-notifications.yml
jar-b: Since it's only referenced once, I like using the initial value directly. I'll get this updated.

jar-b: Fixed with [6ec8dea](https://github.com/hashicorp/terraform-provider-aws/pull/39015/commits/6ec8dea4f2126894a44a785bbc58dd8de9b6ec5d)

PR #39017, opened 2024-08-23
Add Team Working Board Composite Action

justinretzolk:
### Description

Adds a new Composite Action to automate the team's working board. This lets us add logic to prevent no-op calls to `gh project item-edit` that have been failing lately. Bonus points, it cleans up the calling workflow files quite a bit, and will allow us to remove some of the infrastructure around the working board variables.

I used `act` to test locally using my test project, so it wasn't perfectly 1:1, but it worked. Since I'm not able to test this particularly well locally, I only moved `pull_request_target.yml` over to the new composite action, until we validate it.

### DIFF
--- a/.github/actions/team_working_board/README.md
+++ b/.github/actions/team_working_board/README.md
@@ -1,0 +1,33 @@
+# Team Working Board
+
+Used to automate the AWS Provider Team's working board.
+
+## Usage
+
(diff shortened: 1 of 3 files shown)


jar-b (approved):
LGTM 🚀

PR #39057, opened 2024-08-27
Use ServiceRecord instead of `names` for `skaff`

gdavison:
### Description

Uses ServiceRecord data instead of `names` for `skaff`


jar-b (approved):
LGTM 🚀

Issue #38153, opened 2024-06-27
[Bug]: ecs_task_definition → volume → configure_at_launch attribute change not detected


izeau:
### Terraform Core Version

1.9.0

### AWS Provider Version

5.55.0

### Affected Resource(s)

* aws_ecs_task_definition

### Expected Behavior

A diff is detected and a new task definition revision is created

### Actual Behavior

No diff detected

### Terraform Configuration Files

```terraform
resource "aws_ecs_task_definition" "service" {
  family                   = "test"
  network_mode             = "awsvpc"
  requires_compatibilities = ["FARGATE"]
  cpu                      = 1024
  memory                   = 2048

  container_definitions = jsonencode([{
    name = "demo"
    image = "alpine"
  }])

  volume {
    name = "test"
  }
}
```

```terraform
resource "aws_ecs_task_definition" "service" {
  ...

  volume {
    name                = "test"
    configure_at_launch = true
  }
}
```

### Steps to Reproduce

- Apply the first configuration as usual
- Update the volume block to add the `configure_at_launch` attribute
- Plan

### Would you like to implement a fix?

None

juicetin:
There are conditions which this isn't triggered, i.e. I don't _lose_ the fact that I have set `configuredAtLaunch: true`

Extending on the original bug:
However, more often than not, even when I have already set it, it gets rolled back on the subsequent `terraform apply`, even though I have it set to true in tf code

PR #39062, opened 2024-08-28
fix: configure_at_launch -> configured_at_launch

ryanleecode:
### Description
Fixes typo in `configure_at_launch`. it should be `configured_at_launch`. All downstream sdks that generate based off of terraform's are broken b/c of this.

### Relations
https://github.com/hashicorp/terraform-provider-aws/issues/38153
https://github.com/pulumi/pulumi-aws/issues/4413

### DIFF
--- a/internal/service/ecs/service_test.go
+++ b/internal/service/ecs/service_test.go
@@ -2229,7 +2229,7 @@ TASK_DEFINITION
 
   volume {
     name                = "vol1"
-    configure_at_launch = true
+    configured_at_launch = true
   }
 }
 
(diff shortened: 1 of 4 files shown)


jar-b (commented):
Hello @ryanleecode 👋 - thanks for your submission. Changing a schema argument is a breaking change, which we typically will not do in minor releases.

Can you elaborate on how downstream SDKs are broken by this? I would expect reliance on the Terraform AWS provider schema definition would result in the same naming convention (which does vary from the AWS API).

ryanleecode:
Might be a false alarm. see https://github.com/pulumi/pulumi-aws/issues/4413

closing for now :smiley_cat:

PR #39089, opened 2024-08-29
Fixes `aws_fms_admin_account` sweeper

gdavison:
### Description

When the current account has an admin account, sweepers running outside of `us-east-1` will return an error


nam054 (approved):
LGTM 🚀

PR #39127, opened 2024-09-03
(doc): Update default version for creating grafana workspace

bonclay7:
Simple doc update about the default behaviour of Grafana workspace creation

### DIFF
--- a/website/docs/r/grafana_workspace.html.markdown
+++ b/website/docs/r/grafana_workspace.html.markdown
@@ -79,7 +79,7 @@ The following arguments are optional:
 * `configuration` - (Optional) The configuration string for the workspace that you create. For more information about the format and configuration options available, see [Working in your Grafana workspace](https://docs.aws.amazon.com/grafana/latest/userguide/AMG-configure-workspace.html).
 * `data_sources` - (Optional) The data sources for the workspace. Valid values are `AMAZON_OPENSEARCH_SERVICE`, `ATHENA`, `CLOUDWATCH`, `PROMETHEUS`, `REDSHIFT`, `SITEWISE`, `TIMESTREAM`, `XRAY`
 * `description` - (Optional) The workspace description.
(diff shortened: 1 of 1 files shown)


jar-b:
Thanks for your contribution, @bonclay7! 👍

website/docs/r/grafana_workspace.html.markdown
```diff
@@ -79,7 +79,7 @@ The following arguments are optional:
 * `configuration` - (Optional) The configuration string for the workspace that you create. For more information about the format and configuration options available, see [Working in your Grafana workspace](https://docs.aws.amazon.com/grafana/latest/userguide/AMG-configure-workspace.html).
 * `data_sources` - (Optional) The data sources for the workspace. Valid values are `AMAZON_OPENSEARCH_SERVICE`, `ATHENA`, `CLOUDWATCH`, `PROMETHEUS`, `REDSHIFT`, `SITEWISE`, `TIMESTREAM`, `XRAY`
 * `description` - (Optional) The workspace description.
-* `grafana_version` - (Optional) Specifies the version of Grafana to support in the new workspace. Supported values are `8.4`, `9.4` and `10.4`. If not specified, defaults to `10.4`.
+* `grafana_version` - (Optional) Specifies the version of Grafana to support in the new workspace. Supported values are `8.4`, `9.4` and `10.4`. If not specified, defaults to the latest version.
```
bonclay7, line 82: makes sense, but this might not be true for a while after the next version is released

bonclay7: at least for a while

jar-b: That's fair. Let's go with matching the [AWS documentation](https://docs.aws.amazon.com/grafana/latest/APIReference/API_CreateWorkspace.html#ManagedGrafana-CreateWorkspace-request-grafanaVersion) for now, and if the default becomes something other than the latest version we can revert at the same time the list of supported values are updated.

> Specifies the version of Grafana to support in the new workspace. If not specified, defaults to the latest version (for example, 10.4).

PR #39234, opened 2024-09-09
r/aws_imagebuilder_image_recipe(doc): remove uninstall_after_build default

jar-b:
### Description

The `systems_manager_agent.uninstall_after_build` argument does not have a default value. This change removes mention of a default from the registry documentation to avoid confusion about whether the argument must be provided.

Also includes some general cleanup for this resource documentation.


ewbankkit (approved):
LGTM :rocket:.

PR #39244, opened 2024-09-10
Enable quota name lookup for `aws_servicequotas_service_quota`

alexwilcox9:
### Description

The data source `aws_servicequotas_service_quota` allows for either `quota_name` or `quota_code` to be used when looking up a quota. This PR aims to bring the same flexibility to the matching resource so users can avoid hard coding quota IDs

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/internal/service/servicequotas/service_quota.go
+++ b/internal/service/servicequotas/service_quota.go
@@ -50,17 +50,22 @@ func ResourceServiceQuota() *schema.Resource {
 			},
 			"quota_code": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
+				Computed: true,
 				ForceNew: true,
 				ValidateFunc: validation.All(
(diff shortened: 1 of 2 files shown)


ewbankkit:
It seems the more natural thing to do here would be to chain the `aws_servicequotas_service_quota` data source and resource together in configuration:

```hcl
data "aws_servicequotas_service_quota" "example" {
  quota_name   = "VPCs per Region"
  service_code = "vpc"
}

resource "aws_servicequotas_service_quota" "example" {
  quota_code   = data.aws_servicequotas_service_quota.example.quota_code
  service_code = data.aws_servicequotas_service_quota.example.service_code
  value        = "42"
}
```

@alexwilcox9 Does this work for you?

alexwilcox9:
That does work but I thought it might be cleaner to make the data and resource work in the same way rather than having to use both.

PR #39329, opened 2024-09-13
provider: generates `resolveEndpoint`

gdavison:
### Description

The `AWSClient` function `resolveEndpoint` function has overrides for AWS SDK v1 services to correctly handle endpoints defined in environment variables or shared config files. Moves determination of which services need the override behaviour from run-time to generate-time


ewbankkit (approved):
LGTM :rocket:.

PR #39350, opened 2024-09-16
provider: Generates `endpoints` schema

gdavison:
### Description

Generates `endpoints` schema instead of dynamically creating at runtime

Updates custom endpoints documentation to include environment variables and shared config file parameters


ewbankkit (approved):
LGTM :rocket:.

PR #39374, opened 2024-09-18
[chore] Remove unnecessary closed id field

breathingdust:
### Description

Update to this action removes the closed id action parameter so it can be removed. Also corrected casing of token parameter. This can also actually be removed but needs to happen in the action first.


ewbankkit (approved):
LGTM :rocket:.

Issue #39287, opened 2024-09-12
Top-level `dependabot` `go.mod` checks timing out


ewbankkit:
The top-level `dependabot` `go.mod` checks have been timing out recently and we have had no AWS SDK for Go v2 updates.
See e.g.
* https://github.com/hashicorp/terraform-provider-aws/network/updates/883521579
* https://github.com/hashicorp/terraform-provider-aws/network/updates/882627808

ewbankkit:
No timeout last night (9/12): https://github.com/hashicorp/terraform-provider-aws/pull/39299.

PR #39422, opened 2024-09-20
dependabot: Split top-level Go dependency checks into 4

ewbankkit:
### Description

In an attempt to address regular _dependabot_ timeouts (https://github.com/hashicorp/terraform-provider-aws/issues/39287), split the checking of the top-level Go dependencies into 4 using [`allow`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow) and [`ignore`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) directives:

* Everything but the AWS SDK for Go v2 service modules
* The AWS SDK for Go v2 service modules, roughly divided into 3 groups based on the first letter of the service name and how many services are in each group

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/39409.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,28 +5,27 @@ updates:
     schedule:
       interval: "daily"
 
+  # Split the top-level Go dependency check into 4 to avoid timeouts.
+  #  1) Everything but the AWS SDK for Go v2 service modules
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🚀

PR #39492, opened 2024-09-26
doc: Update object_size_greater_than of s3 lifecycle defaults to 128000

mrnonz:
### Description
As announcement from AWS https://aws.amazon.com/about-aws/whats-new/2024/09/amazon-s3-default-minimum-object-size-lifecycle-transition-rules/. I think it's better to note for user to notice about the default value changed to 128KB.


ewbankkit:
@mrnonz Thanks for the contribution :tada: :clap:.
This work has been included in https://github.com/hashicorp/terraform-provider-aws/pull/39578.

PR #39628, opened 2024-10-08
doc: update route53 import command

nimboya:
### Relations

Closes #0000

```console

...
```

### DIFF
--- a/website/docs/r/route53_record.html.markdown
+++ b/website/docs/r/route53_record.html.markdown
@@ -253,11 +253,11 @@ import {
 Using the ID of the record, which is the zone identifier, record name, and record type, separated by underscores (`_`):
 
 ```console
-% terraform import aws_route53_record.myrecord Z4KAPRWWNC7JR_dev.example.com_NS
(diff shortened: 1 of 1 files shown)


justinretzolk:
Hey @nimboya 👋 Thank you for taking the time to raise this! I'm not certain I follow this change. Can you provide a bit more information as to why you feel this is a necessary change?

nimboya:
> Hey @nimboya 👋 Thank you for taking the time to raise this! I'm not certain I follow this change. Can you provide a bit more information as to why you feel this is a necessary change?

The command to import Route53 DNS record was wrong. It does not need the Domain and Record type. When I used it that way it did not import the resource correctly to the Terraform state. The correct command was without the Domain/Sub-domain and record type.

It should be this

```terraform import aws_route53_record.myrecord Z4KAPRWWNC7JR_dev_NS_dev

and not this

```terraform import aws_route53_record.myrecord Z4KAPRWWNC7JR_dev.example.com_NS

ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@nimboya Thanks for the contribution :tada: :clap:.

PR #39634, opened 2024-10-08
Add support to restore a serverless cache snapshot as a provisioned replication group

### LINKED ISSUE #39632: [Enhancement]: add serverless_snapshot_name attribute to aws_elasticache_replication_group resource
### Description

Add `serverless_snapshot_name` attribute to the `aws_elasticache_replication_group` resource, to support restoring a Serverless Cache snapshot as a provisioned replication group.

* aws_elasticache_replication_group

_No response_

### References

* https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html (search for `ServerlessCacheSnapshotName`)

Yes

vtstanescu:
### Description

r/aws_elasticache_replication_group: add `serverless_cache_snapshot_name` attribute
r/aws_elasticache_replication_group: fix restoring a Multi-AZ enabled replication group snapshot into a Multi-AZ disabled replication group
r/aws_elasticache_serverless_cache: add `final_snapshot_name` attribute

Test steps:
1. Create a Redis serverless cache, a Redis replication group with Multi-AZ enabled, and a Redis replication group with Multi-AZ disabled (1 shard, 0 replicas)
2. Take a snapshot of each cache, let's call them: _serverless_, _maz_, _saz_
3. Restore each in a replication group with Multi-AZ disabled. It won't work with the current AWS provider for the _serverless_ and _maz_ snapshots, only for the _saz_ one.
That's because `multi_az_enabled` resolves to a zero value (boolean false) so it is not added to the CreateReplicationGroupInput that is passed to the AWS API.
For restoring the _serverless_ snapshot, use the new `serverless_cache_snapshot_name` attribute.
4. Following this patch, step 3 works for all snapshots, with `multi_az_enabled = false`, and does not work for the _serverless_ and _maz_ snapshots if setting `multi_az_enabled = null`, or not specifying it at all.

### Relations

Closes #39632

None


internal/service/elasticache/replication_group.go
```diff
@@ -490,7 +499,12 @@ func resourceReplicationGroupCreate(ctx context.Context, d *schema.ResourceData,
 		input.PreferredMaintenanceWindow = aws.String(v.(string))
 	}
 
-	if v, ok := d.GetOk("multi_az_enabled"); ok {
+	// We want to explictly set MultiAZEnabled even if it is `false` in the configuration.
+	// This addresses the error InvalidParameterCombination
+	// "Each Node Group needs to have at least one replica for Multi-AZ enabled Replication Group"
+	// when creating a replication group with 1 shard/node-group and 0 replicas with Multi AZ disabled,
+	// using the snapshot of either a multi-AZ enabled replication group or of a serverless cache.
+	if v, ok := d.GetOkExists("multi_az_enabled"); ok {
```
vtstanescu, line 507: This seems to have an unintended effect of always setting `input.MultiAZEnabled`, even if the `multi_az_enabled` attribute is not specified in the resource or is set as `null`.
What's the best way to check if an attribute exists in the configuration, regardless of whether it's value is zero or not.
`GetOk` function has the downside that will return the second value as `false` if the first one is a "type's zero"

Issue #39651, opened 2024-10-09
[Bug]: allow aws_iam_account_alias to return nil or empty on account without alias


jose-pablo-camacho:
### Terraform Core Version

1.5.7

### AWS Provider Version

5.70.0

### Affected Resource(s)

Datasource: aws_iam_account_alias.

### Expected Behavior

Datasource `aws_iam_account_alias` should NOT return an error when there's no Aliases associated with an account, this makes the whole TF flow fail for newly created AWS accounts or accounts that simply don't have an alias set. This should return `nil` or `empty array of strings` since that's actually the data we got from the `ListAccountAliases` API call and not an `error` based on the [implementation](https://github.com/hashicorp/terraform-provider-aws/blob/v5.70.0/internal/service/iam/account_alias_data_source.go#L43-L46).

### Actual Behavior

Datasource `aws_iam_account_alias` returns an error when trying to get an alias for an account without one.

### Relevant Error/Panic Output Snippet

```shell
Error: reading IAM Account Alias: empty result
 
  with module.onboarding.data.aws_iam_account_alias.current,
  on <module-path>/main.tf line 57, in data "aws_iam_account_alias" "current":
  57: data "aws_iam_account_alias" "current" {}
```

### Terraform Configuration Files

data "aws_iam_account_alias" "current" {}

### Steps to Reproduce

Run TF plan/apply on an account without an `alias` and try to get the data from `data "aws_iam_account_alias" "current" {}`.

### Debug Output

Error: reading IAM Account Alias: empty result
 
  with module.onboarding.data.aws_iam_account_alias.current,
  on <module-path>/main.tf line 57, in data "aws_iam_account_alias" "current":
  57: data "aws_iam_account_alias" "current" {}

### Panic Output

N/A

### Important Factoids

N/A

### References

N/A

### Would you like to implement a fix?

None

jose-pablo-camacho:
I opened a PR with some fix, however I have some questions regarding testing. Please see: https://github.com/hashicorp/terraform-provider-aws/pull/39653

justinretzolk:
Duplicate of #29324

justinretzolk:
[...]

PR #39653, opened 2024-10-09
b: allow aws_iam_account_alias-datasource to return nil

jose-pablo-camacho:
### Description

This PR is a small change that allows the `aws_iam_account_alias` datasource to return `nil` instead of erroring out when there's no aliases associated to an account. For me it makes sense to return the `accountID` as the ID for the datasource since it's the identifier at the time but I'm open to suggestions.

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/issues/39651

I have a couple of questions related to Acceptance Testing:
1. Can you help me running these tests?
2. Related to the tests setup now, looks like we're executing an account alias creation and then getting it. In order to test this we will need an account without an alias and don't call the `resource "aws_iam_account_alias"` before calling `data "aws_iam_account_alias"` - is this a possibility? I'm basing off this code https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/iam/account_alias_data_source_test.go#L16-L52.

I'd like to get some advice here related to tests.

```console

...
```

### DIFF
--- a/internal/service/iam/account_alias_data_source.go
+++ b/internal/service/iam/account_alias_data_source.go
@@ -42,7 +42,10 @@ func dataSourceAccountAliasRead(ctx context.Context, d *schema.ResourceData, met
 
 	// 'AccountAliases': [] if there is no alias.
 	if resp == nil || len(resp.AccountAliases) == 0 {
-		return sdkdiag.AppendErrorf(diags, "reading IAM Account Alias: empty result")
+		log.Printf("[DEBUG] No Aliases Associated to Account %s.", meta.(*conns.AWSClient).AccountID)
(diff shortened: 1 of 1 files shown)


justinretzolk:
Hey @jose-pablo-camacho 👋 Thank you for taking the time to raise this! As far as your question about the tests -- whoever picks this up for review will be able to help with those (I didn't want to leave your question completely unaddressed whilst triaging).

jose-pablo-camacho:
Hi @justinretzolk, is there any estimated time for someone to pickup this PR for review?

jose-pablo-camacho:
Hi @justinretzolk, is this something that could be picked up soon?

PR #39656, opened 2024-10-09
Corrects regex pattern `/+.` in various acceptance tests

gdavison:
### Description

A number of acceptance tests check for the regex `/+.`, which checks for one or more `/` followed by any character (and since it doesn't include `$`, it can be followed by any additional characters.

Replaces the regex with `/.+$`, which checks for a single `/` followed by at least one character until the end of the string.


ewbankkit (approved):
LGTM :rocket:.

PR #39751, opened 2024-10-16
d/aws_efs_file_system: Document 'file_system_id' attribute

acmcelwee:
### Description

[This attribute already exists on the data source](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/efs/file_system_data_source.go#L56-L60), but the documentation was simply missing that attribute.

### References
* [Relevant attribute in data source schema](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/efs/file_system_data_source.go#L56-L60)

### DIFF
--- a/website/docs/d/efs_file_system.html.markdown
+++ b/website/docs/d/efs_file_system.html.markdown
@@ -46,6 +46,7 @@ This data source exports the following attributes in addition to the arguments a
 * `availability_zone_id` - The identifier of the Availability Zone in which the file system's One Zone storage classes exist.
 * `dns_name` - DNS name for the filesystem per [documented convention](http://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-dns-name.html).
 * `encrypted` - Whether EFS is encrypted.
+* `file_system_id` - The ID that identifies the file system (e.g., fs-ccfc0d65).
 * `kms_key_id` - ARN for the KMS encryption key.
 * `lifecycle_policy` - File system [lifecycle policy](https://docs.aws.amazon.com/efs/latest/ug/API_LifecyclePolicy.html) object.
 * `name` - The value of the file system's `Name` tag.


website/docs/d/efs_file_system.html.markdown
```diff
@@ -46,6 +46,7 @@ This data source exports the following attributes in addition to the arguments a
 * `availability_zone_id` - The identifier of the Availability Zone in which the file system's One Zone storage classes exist.
 * `dns_name` - DNS name for the filesystem per [documented convention](http://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-dns-name.html).
 * `encrypted` - Whether EFS is encrypted.
+* `file_system_id` - The ID that identifies the file system (e.g., fs-ccfc0d65).
```
acmcelwee, line 49: You know, as I re-read these docs, I see above that it says `This data source exports the following attributes in addition to the arguments above:`.

Some AWS resources/data-sources fully lay out all attributes in the Attribute Reference section - [example](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/autoscaling_group#attribute-reference). Maybe what would be more useful is a removal of the statement `This data source exports the following attributes in addition to the arguments above:` and ensure that the Attribute Reference section covers the exhaustive list.

Thoughts?

justinretzolk: Hey @acmcelwee 👋 Thank you for taking the time to put this PR together! As you've mentioned, it looks like `file_system_id` is indeed documented in the argument section. The pattern our documentation follows is to split arguments (things that you can supply a value to in your configuration) and attributes (read-only values). This holds true in the `aws_autoscaling_group` data source that you linked to; you may only supply a value to the `name` argument, while the rest of are read-only attributes.

acmcelwee: @justinretzolk, I'm not sure if I fully follow your guidance here. Are you saying that the preference is for the docs to clearly lay out all arguments and a separate call out for all attributes? Or, are you saying that this current doc follows the current preference of the hashcorp team?

The docs don't currently seem to consistently follow one way or the other. For example, that ASG data source I referenced takes the hard split of arguments and then an exhaustive list of attributes (that duplicates the `name` attribute argument that was listed in the arguments section). On the other side of that, it looks like [~1.4k files](https://github.com/search?q=repo%3Ahashicorp%2Fterraform-provider-aws+%22This+data+source+exports+the+following+attributes+in+addition+to+the+arguments+above%22&type=code) use the approach of not restating the arguments in the attributes section and just points the reader to the arguments section as additional attributes. I'm not sure if I'm a lone data point, but I will at least say that multiple people on my team did completely overlook the reference to the arguments as attributes that are also available. I'm happy to reshape this PR in whatever way you want to guide it, but I just need to fully understand your thoughts on how you'd like to proceed.

PR #39768, opened 2024-10-17
Update Verified Access Group SSE Config Example

u-blavins:
The example for verified access group set with SSE encryption was mentioned incorrectly.

Referred Doc : https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/verifiedaccess_group

Open Issue: #39769


jar-b (approved):
LGTM 🚀

jar-b:
Thanks for the correction, @u-blavins! 👍

PR #39781, opened 2024-10-17
Update the name of the resource to be more accurate.

thoward:
The resource name `subnet` is misleading in this example, because it's returning a security group, not the subnet. It makes more sense to call this `subnet_security_group` to indicate that this is the security group attached to the subnet.


ewbankkit:
@thoward Thanks for the contribution :tada: :clap:.

PR #39802, opened 2024-10-18
docs: add argument description to resource aws_ami_copy

### LINKED ISSUE #39794: [Docs]: aws_ami_copy references undocumented "description"
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ami_copy

### Description

The example in this documentation has:

```text
description       = "A copy of ami-xxxxxxxx"
```

But this `description` argument is not listed in the argument reference. It should be included in both places or neither, depending on whether this argument is actually supported.

### References

_No response_

No

stefanfreitag:
### Description

The example in resource `aws_ami_copy` contains a `description` field that is not documented in the _Arguments_ section.

```hcl
resource "aws_ami_copy" "example" {
  name              = "terraform-example"
  description       = "A copy of ami-xxxxxxxx"
  source_ami_id     = "ami-xxxxxxxx"
  source_ami_region = "us-west-1"

  tags = {
    Name = "HelloWorld"
  }
}
```

This PR added the missing field.

Please note: The resource documentation contains the statement below

> This resource also exposes the full set of arguments from the [aws_ami](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ami) resource.

So, it may be the case that adding the  `description` field is not required, but it helps readers as they to not need to open another documentation page for getting details on `description` . 

### Relations

Closes #39794 

### References

- [Link to resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ami_copy)
- [Usage of description in AWS Go SDK v2](https://github.com/aws/aws-sdk-go-v2/blob/6a30b3f71b8aaf38fef1816652f5b8847471681c/service/ec2/api_op_CopyImage.go#L84)

Not required. Only documentation is updated.

### DIFF
--- a/website/docs/r/ami_copy.html.markdown
+++ b/website/docs/r/ami_copy.html.markdown
@@ -40,6 +40,7 @@ resource "aws_ami_copy" "example" {
 This resource supports the following arguments:
 
 * `name` - (Required) Region-unique name for the AMI.
+* `description` - (Optional) Longer, human-readable description for the AMI in the destination region.
 * `source_ami_id` - (Required) Id of the AMI to copy. This id must be valid in the region
   given by `source_ami_region`.
 * `source_ami_region` - (Required) Region from which the AMI will be copied. This may be the
--- a/website/docs/r/ami_copy.html.markdown
+++ b/website/docs/r/ami_copy.html.markdown
@@ -25,7 +25,6 @@ block until the new AMI is available for use on new instances.
 ```terraform
 resource "aws_ami_copy" "example" {
   name              = "terraform-example"
-  description       = "A copy of ami-xxxxxxxx"
   source_ami_id     = "ami-xxxxxxxx"
   source_ami_region = "us-west-1"
 
@@ -40,7 +39,6 @@ resource "aws_ami_copy" "example" {
 This resource supports the following arguments:
 
 * `name` - (Required) Region-unique name for the AMI.
-* `description` - (Optional) Longer, human-readable description for the AMI in the destination region.
 * `source_ami_id` - (Required) Id of the AMI to copy. This id must be valid in the region
   given by `source_ami_region`.
 * `source_ami_region` - (Required) Region from which the AMI will be copied. This may be the


justinretzolk:
Hey @stefanfreitag 👋 Thank you for taking the time to raise this! Since the documentation mentions the whole "exposes the full set of arguments from the [aws_ami](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ami) resource" thing, I'm not sure it's appropriate to add the `description` argument here. I see where you're coming from on not needing to navigate to another page, but my take would be that we either need to leave the documentation as is or add all of the argument documentation from `aws_ami` to `aws_ami_copy` rather than linking. 

I generally try to get docs changes merged in as quick as possible, but in this case, I'd like to leave this for now for further input from the team / community.

stefanfreitag:
Hi  @justinretzolk
thanks for the feedback! 
I think it is also the only resource in the complete code base that contains

> This resource also exposes the full set of arguments from the ...

As you may have noticed I was also a bit sceptical on adding the description argument.
An alternative approach would be to remove of `description` from the example.  Then the example uses only attributes documented on the [aws_ami_copy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ami_copy) page. 

```hcl
resource "aws_ami_copy" "example" {
  name              = "terraform-example"
  source_ami_id     = "ami-xxxxxxxx"
  source_ami_region = "us-west-1"
  tags = {
    Name = "HelloWorld"
  }
```

Would that make more sense?

jar-b:
It's fairly common practice to link the `Attributes` section of data sources to the corresponding resource documentation to ensure consistency. While this is a slightly different case (resource arguments), I think the purpose is similar in avoiding diverging argument definitions.

My vote would be removing `description` from the example rather than copying all of the arguments already documented on the `aws_ami` resource page. We could also optionally move this note from the bottom of the `Arguments` section to the top:

> This resource also exposes the full set of arguments from the [`aws_ami`](ami.html) resource.

jar-b (approved):
LGTM 🚀

jar-b:
Thanks for the investigation and clarification, @stefanfreitag!

PR #39848, opened 2024-10-23
Updated the description for the id attributes in the aws_kms_alias da…

satyajeetjsr:
…ta source documentation to reflect their distinct purposes

### Relations

Closes #0000

```console

...
```


jar-b (commented):
In this case the `id` attribute is explicitly set to the alias ARN:

https://github.com/hashicorp/terraform-provider-aws/blob/7d0ae4383729508853d09524581a84fd0524b0a4/internal/service/kms/alias_data_source.go#L55

Are you observing something different in the output of this data source?

PR #39745, opened 2024-10-16
Add Valkey engine support for ElastiCache resources

relu:
### Description

Adding support for the newly introduced Valkey engine in AWS ElastiCache.

### Relations

Closes #39641.
Closes https://github.com/hashicorp/terraform-provider-aws/issues/39711.

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccElastiCacheCluster PKG=elasticache
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.23.2 test ./internal/service/elasticache/... -v -count 1 -parallel 20 -run='TestAccElastiCacheCluster'  -timeout 360m
2024/10/16 21:29:03 Initializing Terraform AWS Provider...
=== RUN   TestAccElastiCacheClusterDataSource_basic
=== PAUSE TestAccElastiCacheClusterDataSource_basic
=== RUN   TestAccElastiCacheClusterDataSource_Engine_Redis_LogDeliveryConfigurations
=== PAUSE TestAccElastiCacheClusterDataSource_Engine_Redis_LogDeliveryConfigurations
=== RUN   TestAccElastiCacheCluster_Engine_memcached
=== PAUSE TestAccElastiCacheCluster_Engine_memcached
=== RUN   TestAccElastiCacheCluster_Engine_redis
=== PAUSE TestAccElastiCacheCluster_Engine_redis
=== RUN   TestAccElastiCacheCluster_disappears
=== PAUSE TestAccElastiCacheCluster_disappears
=== RUN   TestAccElastiCacheCluster_Engine_redis_v5
=== PAUSE TestAccElastiCacheCluster_Engine_redis_v5
=== RUN   TestAccElastiCacheCluster_Engine_None
=== PAUSE TestAccElastiCacheCluster_Engine_None
=== RUN   TestAccElastiCacheCluster_PortRedis_default
=== PAUSE TestAccElastiCacheCluster_PortRedis_default
=== RUN   TestAccElastiCacheCluster_ParameterGroupName_default
=== PAUSE TestAccElastiCacheCluster_ParameterGroupName_default
=== RUN   TestAccElastiCacheCluster_ipDiscovery
=== PAUSE TestAccElastiCacheCluster_ipDiscovery
=== RUN   TestAccElastiCacheCluster_port
=== PAUSE TestAccElastiCacheCluster_port
=== RUN   TestAccElastiCacheCluster_snapshotsWithUpdates
[...]

PR #39950, opened 2024-10-30
Fix invalid terraform drifting to redis in global replication groups

jramosf:
### Description

This change follows-up on https://github.com/hashicorp/terraform-provider-aws/pull/39745 and fixes a bug which causes drift for ElastiCache Global Datastore in Valkey.

This is due to the current hard-coded default value for the `engine` argument, which inevitably causes the drift even in a `terraform plan` in an empty state importing an existing `aws_elasticache_replication_group` resource with the `global_replication_group_id` argument set (if this argument is set, the `redis` value in `engine` argument should **not** be passed).

```terraform
      ~ engine                         = "valkey" -> "redis" # forces replacement
```

The API parameter in AWS for `Engine` is [not required](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html) and has a default value, which brings the Terraform provider in line with AWS API behaviour.

### Relations

Relates #39745 

As per the [contributor guide](https://hashicorp.github.io/terraform-provider-aws/#4-write-tests) I would prefer the tests run on HashiCorp side.

### DIFF
--- a/.changelog/39950.txt
+++ b/.changelog/39950.txt
@@ -1,0 +1,4 @@
+
+```release-note:bug
+resource/aws_elasticache_replication_group: Fix invalid drifting to redis engine in global replication groups
+```
--- a/internal/service/elasticache/replication_group.go
+++ b/internal/service/elasticache/replication_group.go
@@ -124,7 +124,7 @@ func resourceReplicationGroup() *schema.Resource {
 			names.AttrEngine: {
 				Type:         schema.TypeString,
 				Optional:     true,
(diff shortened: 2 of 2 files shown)


ptallapureddy (approved):

jramosf:
@justinretzolk this Pull Request is over 3 months old open with no activity. I am going to wait for another week but afterwards I'll close it if there is no traction from HashiCorp to merge.

Thanks,

PR #39964, opened 2024-10-31
Add Script to Check Registry Version to Github and Initialize 

nam054:
### Description

The Terraform Provider release process can cause a “partial release“ if an interruption occurs, leading to a SHASUM mismatch which will prevent the provider version from being installed. This occurred  recently leading to widespread reported issues from users. This can appear as a successful release via the GHA release process, but will only result in issues once the Registry ingests and makes that release available (which can occur an hour after an apparently successful release)

This new action will install the latest version of the provider from the registry and install in the 3 architectures - Windows, Linux, Mac. If the installation fails, or if the version on the registry does not match Github in an hrs time, it will fail.

### Relations

Closes #0000

```console

...
```


.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,52 @@ jobs:
...
+
+
+  registry-check:
+    # strategy:
+    #   matrix:
+    #     os: [ubuntu-latest, windows-latest, macOS-latest]
+    # runs-on: ${{ matrix.os }}
+    # runs-on: ubuntu-latest
+    runs:
+      using: "composite"
```
justinretzolk, line 63: These are only needed for composite actions, so you'll need to remove them since this is a "normal" workflow

```suggestion
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,52 @@ jobs:
...
+      steps:
+        - using: composite
```
justinretzolk, line 65: Similar to the other suggestion

```suggestion
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,52 @@ jobs:
...
+        - name: Registry Version Check
+          id: registry_latest_ver
+          shell: bash
+          run: |
+            curl -s "https://registry.terraform.io/v2/providers/323/provider-versions/latest" | jq '.data.attributes.version' >> "$GITHUB_OUTPUT"
```
justinretzolk, line 70: For the output to get registered properly, you'll need whatever's being sent to `$GITHUB_OUTPUT` to be in the format of `key=value`

```suggestion
            echo "current=$(curl -s https://registry.terraform.io/v2/providers/323/provider-versions/latest | jq '.data.attributes.version')" >> "$GITHUB_OUTPUT"
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,52 @@ jobs:
...
+
+        - name: Check Version Values
+          if: ${{ needs.on-success-or-workflow-dispatch.outputs.release-tag }} == steps.registry_latest_ver.outputs.current
```
justinretzolk, line 73: Thinking about this a bit more, I'm not sure this `if` is going to do what we want it to after all. Currently, if `needs.on-success-or-workflow-dispatch.outputs.release-tag != steps.registry_latest_ver.outputs.current`, this step will just get skipped and the workflow will move on to the next step.

I'm kinda thinking we need to have this step run every time (no `if`), and then do a loop until the current version matches the one from

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,84 @@ jobs:
...
+      latest-version: ${{ steps.registry_latest_ver.outputs.current }}
+    steps:
+      - name: Registry Version Check
+        id: registry_latest_ver
+        shell: bash
+        run: |
+          for i in 1 2
+          do
+            LATEST_VERSION=$(curl -s "https://registry.terraform.io/v2/providers/323/provider-versions/latest" | jq -r '.data.attributes.version')
+            if [[ " ${{ needs.on-success-or-workflow-dispatch.outputs.release-tag }}" != ${LATEST_VERSION} ]]; then
```
justinretzolk, line 68: Off of the top of my head, I'm not sure if this extra space would cause the comparison to fail, but it wouldn't surprise me.

```suggestion
            if [[ "${{ needs.on-success-or-workflow-dispatch.outputs.release-tag }}" != ${LATEST_VERSION} ]]; then
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,84 @@ jobs:
...
+              sleep 1h
```
justinretzolk, line 69: Does it generally take over an hour to get into the registry? If we can make this shorter, I think it's probably worthwhile.

```suggestion
              sleep 1h
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,84 @@ jobs:
...
+            else
+              echo "Registry and Github Version matches"
+                echo "current=$LATEST_VERSION >> "$GITHUB_OUTPUT"
+            fi
+          done
+
+  os-version-init:
+    strategy:
```
justinretzolk, line 77: Adding a name to help eliminate `echo` calls (more info in later comment)

```suggestion
  os-version-init:
    name: Run terraform init On Supported Platforms
    strategy:
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,84 @@ jobs:
...
+
+          provider "aws" {
+            region = "us-east-1"
+          }
+          EOF
+
+          echo "Terraform configuration file (main.tf) created with the latest version."
+
+          echo "Initializing Terraform with the latest AWS provider..."
+          terraform init -upgrade
```
justinretzolk, line 104: If we break the different parts of this inline script into several steps, we get a couple of benefits. Whatever we put for the `name` shows up in the Actions UI for the step, so there's no need to do the `echo` calls to indicate what the script is doing. With that, it's easier to identify what part of the script failed in the UI, since that (named) step will show as failed. Nicely, if the command in a step fails, the step will fail, any further steps are skipped (unless instructed otherwise), and the job fails (more on that in a separate comment).

```suggestion
    steps:
      - name: Specify Provider Version in Terraform Configuration
        run: |
          cat <<EOF > main.tf
          terraform {
            required_providers {
              aws = {
                source  = "hashicorp/aws"
                version = ${{ needs.registry-check.outputs.latest-version }}
              }
            }
          }

          provider "aws" {
            region = "us-east-1"
          }
          EOF
          
      - name: Initialize the Provider
        run: terraform init -upgrade
```

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,84 @@ jobs:
...
+
+          if [ $? -ne 0 ]; then
+            echo "Terraform initialization failed."
+            exit 1
+          fi
```
justinretzolk, line 109: I believe we can eliminate this part of the script with the above comment in mind, since the step would fail if `terraform init -upgrade` fails.

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,84 @@ jobs:
...
+              "blocks": [
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "ERROR: Registry Provder Initiation Failure"
+                  }
+                }
+              ]
+            }
```
justinretzolk, line 134: With what I mentioned up on line 82, I think we could combine these jobs. The `Send Slack Notification On Failure` step could be added as a step to the end of the `os-version-init` job with an `if` condition that it only runs [on failure](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#failure). If we go that route, we should be able to make the Slack message more rich by adding the `${{ matrix.os }}` to the message, since there would be a separate message for each OS type.

justinretzolk (commented):
I had a few suggestions, but overall this is looking awesome 🙂

.github/workflows/post_publish.yml
nam054: Yeah I think it's expected it'll take about an hr to show up on the registry.

.github/workflows/post_publish.yml
```diff
@@ -55,28 +55,30 @@ jobs:
...
+          for i in 1 2
+          do
+            LATEST_VERSION=$(curl -s "https://registry.terraform.io/v2/providers/323/provider-versions/latest" | jq -r '.data.attributes.version')
+            if [[ " ${{ needs.on-success-or-workflow-dispatch.outputs.release-tag }}" != ${LATEST_VERSION} ]]; then
+              sleep 1h
+            else
+              echo "Registry and Github Version matches"
+                echo "current=$LATEST_VERSION >> "$GITHUB_OUTPUT"
+            fi
+          done
```
justinretzolk, line 74: We probably want this job to fail if `$LATEST_VERSION` and `needs.on-success-or-workflow-dispatch.outputs.release-tag` never end up matching, that way `os-version-init` won't run.

If we're introducing a failure here, do you think we should do a Slack notification, like you've added for the `os-version-init` job? I'm unsure of how chatty we want these to be.

```suggestion
            else
                echo "Registry and Github Version matches"
                echo "current=$LATEST_VERSION" >> "$GITHUB_OUTPUT"
                exit
            fi
        done

        echo "Registry does not contain ${{ needs.on-success-or-workflow-dispatch.outputs.release-tag }}"
        exit 1
```

justinretzolk (approved):
One minor suggestion, but other than that, this LGTM!

.github/workflows/post_publish.yml
```diff
@@ -51,3 +51,72 @@ jobs:
...
+      latest-version: ${{ steps.registry_latest_ver.outputs.current }}
+    steps:
+      - name: Registry Version Check
+        id: registry_latest_ver
+        shell: bash
+        run: |
+          for i in 1 2
+          do
+            LATEST_VERSION=$(curl -s "https://registry.terraform.io/v2/providers/323/provider-versions/latest" | jq -r '.data.attributes.version')
+            if [[ "${{ needs.on-success-or-workflow-dispatch.outputs.release-tag }}" != ${LATEST_VERSION} ]]; then
```
justinretzolk, line 68: Missed this in my review, but I think we to either quote `${LATEST_VERSION}` or remove the `-r` from the `jq` call here since the output from the `needs` will be a quoted string

PR #39975, opened 2024-10-31
linter: Add `perfsprint`

gdavison:
### Description

Adds linter `perfsprint` to catch uses of `fmt.Sprint` or `fmt.Sprintf` used for conversion to string when a direct call to, e.g. a `strconv` call can be used


jar-b (approved):
LGTM 🚀

Issue #39818, opened 2024-10-21
[Bug]: aws_iam_role_policy_attachments_exclusive doesnt detach policies on destroy?


lapkritinis:
### Terraform Core Version

1.9.7

### AWS Provider Version

5.72.1

### Affected Resource(s)

* aws_iam_policy
* aws_iam_role_policy_attachments_exclusive

### Expected Behavior

Created resources should have deleted.

### Actual Behavior

`│ Error: deleting IAM Policy (arn:aws:iam::123456789012:policy/terraform-20241021154342453400000001): operation error IAM: DeletePolicy, https response error StatusCode: 409, RequestID: 51ddf8fa-2fa5-44e3-a47c-81c67a6b3043, DeleteConflict: Cannot delete a policy attached to entities.`

### Terraform Configuration Files

```
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "5.72.1"
    }
  }
}

resource "aws_iam_role" "this" {
  assume_role_policy = data.aws_iam_policy_document.trust-ec2-service.json
}

resource "aws_iam_policy" "this" {
  policy = data.aws_iam_policy_document.ebs-encrypt-decrypt.json
}

resource "aws_iam_role_policy_attachments_exclusive" "this" {
  role_name = aws_iam_role.this.name
  policy_arns = [
    aws_iam_policy.this.arn,
  ]
}

data "aws_iam_policy_document" "trust-ec2-service" {
  statement {
    sid     = "EC2AssumeRole"
    actions = ["sts:AssumeRole"]

    effect = "Allow"

    principals {
      type        = "Service"
      identifiers = ["ec2.amazonaws.com"]
    }
  }
}

data "aws_iam_policy_document" "ebs-encrypt-decrypt" {
  statement {
    effect = "Allow"
    actions = [
      "kms:DescribeKey",
    ]
    resources = ["*"]
  }
}


```

### Steps to Reproduce

terraform apply <- resources are created as expected.
terraform apply -destroy

```
aws_iam_role_policy_attachments_exclusive.this: Destroying...
aws_iam_role_policy_attachments_exclusive.this: Destruction complete after 0s
aws_iam_policy.this: Destroying... [id=arn:aws:iam::654654141215:policy/terraform-20241021155037463400000001]
aws_iam_role.this: Destroying... [id=terraform-20241021155037463400000002]
aws_iam_role.this: Destruction complete after 1s
╷
[...]

PR #40005, opened 2024-11-04
[Docs]: Describe the destroy behavior of `_exclusive` IAM policy attachment resources

jar-b:
### Description

This change adds a note section to each of the role, user, and policy exclusive IAM policy attachment resources to clarify the behavior of resource destruction. For these resources, destroy is a no-op as the intent is simply for Terraform to no longer own reconciliation of the attached policies going forward.

### Relations

Relates #39818


YakDriver (approved):
LGTM 🎉

PR #40008, opened 2024-11-04
internal/framework/types: surface ARN validation errors

### LINKED ISSUE #39227: [Bug]: arnType custom type implementation prevents validation to work properly
### Terraform Core Version

1.9.5

### AWS Provider Version

5.66.0

### Affected Resource(s)

Any resource which currently uses ARNType custom type implementation and there are many

### Expected Behavior

When invalid ARN is passed to attribute of type ARNType, some kind of validation error should appear during `terraform plan`

### Actual Behavior

An error suggesting a bug in the provider is displayed:

```shell
Error: Provider produced invalid plan
│
│ Provider "registry.terraform.io/hashicorp/aws" planned an invalid value for <RESOURCE_WITH_ARN_ATTRIBUTE>: planned value cty.UnknownVal(cty.String) does not match
│ config value cty.StringVal("<SOME_INVALID_ARN>").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

_No response_

Problem can be reproduced using any resource which has ARNType attributes. For example:

```hcl
resource "aws_bedrockagent_agent" "test" {
  agent_name              = "test"
  agent_resource_role_arn = "INVALID-ARN"
  foundation_model        = "test"
}
```

will cause the problem.

Also, the minimal resource implementation reproduces the problem:

```hcl
// @FrameworkResource("aws_test_arn", name="TestArn")
func newResourceTestArn(_ context.Context) (resource.ResourceWithConfigure, error) {
	r := &resourceTestArn{}
	return r, nil
}

type resourceTestArn struct {
	framework.ResourceWithConfigure
}

func (r *resourceTestArn) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
	resp.TypeName = "aws_test_arn"
}

func (r *resourceTestArn) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
		Attributes: map[string]schema.Attribute{
			"arn": schema.StringAttribute{
				CustomType: fwtypes.ARNType,
				Required:   true,
			},
		},
	}
}

func (r *resourceTestArn) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
}
func (r *resourceTestArn) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
}
func (r *resourceTestArn) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
}
func (r *resourceTestArn) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
}
```

this resource will demonstrate the error with the following config:

```hcl
resource "aws_test_arn" "test" {
  arn = "INVALID-ARN"
}
```

### Steps to Reproduce

Use any of the snippets provided above and just run `terraform plan`

### Debug Output

[plan.log](https://github.com/user-attachments/files/16933091/plan.log)
This is the plan for `aws_bedrockagent_agent` snippet suggested above. 

### Panic Output

_No response_

### Important Factoids

The cause of the issue seems to be implementation of the [ARNType](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/framework/types/arn.go)

This method:

```go
func (t arnType) ValueFromString(_ context.Context, in types.String) (basetypes.StringValuable, diag.Diagnostics) {
	var diags diag.Diagnostics

	if in.IsNull() {
		return ARNNull(), diags
	}
	if in.IsUnknown() {
		return ARNUnknown(), diags
	}

	valueString := in.ValueString()
	if _, err := arn.Parse(valueString); err != nil {
		return ARNUnknown(), diags // Must not return validation errors.
	}

	return ARNValue(valueString), diags
}
```

returns ARNUnknown if an invalid ARN is passed. The unknown result will cause the validation method:

```go
func (v ARN) ValidateAttribute(ctx context.Context, req xattr.ValidateAttributeRequest, resp *xattr.ValidateAttributeResponse) {
	if v.IsNull() || v.IsUnknown() {
		return
	}

	if !arn.IsARN(v.ValueString()) {
		resp.Diagnostics.AddAttributeError(
			req.Path,
			"Invalid ARN Value",
			"The provided value cannot be parsed as an ARN.\n\n"+
				"Path: "+req.Path.String()+"\n"+
				"Value: "+v.ValueString(),
		)
	}
}
```

to never pass the first condition thus validation will not take place.

If the lines:

```go
	if _, err := arn.Parse(valueString); err != nil {
		return ARNUnknown(), diags // Must not return validation errors.
	}
```

are removed from `ValueFromString` method, then `terraform plan` for the `aws_bedrockagent_agent` snippet provided above, shows this error instead:

```shell
Error: Invalid ARN Value
│
│   with aws_bedrockagent_agent.test,
│   on main.tf line 85, in resource "aws_bedrockagent_agent" "test":
│   85:   agent_resource_role_arn = "INVALID-ARN"
│
│ The provided value cannot be parsed as an ARN.
│
│ Path: agent_resource_role_arn
│ Value: INVALID-ARN
```

I don't feel competent to suggest that indeed those two lines should be gone as they might be critical to some other behavior, but the fact that the method returns unknown value for invalid ARNs and unknown values are never validated, causes the issue here.

### References

_No response_

None

jar-b:
### Description

Previously the `ARNType`s `ValueFromString` method parsed the string value and set the value to Unknown in cases where validation failed. This caused a plan time failure indicating a difference between the planned value (unknown) and config value (known string). This obfuscated the underlying error which is actually an invalid ARN value.

Invalid ARNs provided in a configuration will now produce errors similar to the following:

```
│ Error: Invalid ARN Value
│
│   with aws_bedrockagent_agent.test,
│   on main.tf line 16, in resource "aws_bedrockagent_agent" "test":
│   16:   agent_resource_role_arn = "INVALID-ARN"
│
│ The provided value cannot be parsed as an ARN.
│
│ Path: agent_resource_role_arn
│ Value: INVALID-ARN
```

```console
% go test ./internal/framework/types/...
ok      github.com/hashicorp/terraform-provider-aws/internal/framework/types    0.631s
```

### Relations

Closes #39227

### DIFF
--- a/.changelog/40008.txt
+++ b/.changelog/40008.txt
@@ -1,0 +1,3 @@
+```release-note:note
+provider: validation of arguments implementing the custom `ARNType` will properly surface validation errors
+```
--- a/internal/framework/types/arn.go
+++ b/internal/framework/types/arn.go
@@ -52,12 +52,9 @@ func (t arnType) ValueFromString(_ context.Context, in types.String) (basetypes.
 		return ARNUnknown(), diags
 	}
 
-	valueString := in.ValueString()
-	if _, err := arn.Parse(valueString); err != nil {
-		return ARNUnknown(), diags // Must not return validation errors.
-	}
-
-	return ARNValue(valueString), diags
+	// The ValidateAttribute method will surface errors if the value is an invalid
+	// ARN. This method simply passes the value through.
+	return ARNValue(in.ValueString()), diags
 }
 
 func (t arnType) ValueFromTerraform(ctx context.Context, in tftypes.Value) (attr.Value, error) {
--- a/internal/framework/types/arn_test.go
+++ b/internal/framework/types/arn_test.go
@@ -36,7 +36,7 @@ func TestARNTypeValueFromTerraform(t *testing.T) {
 		},
 		"invalid ARN": {
 			val:      tftypes.NewValue(tftypes.String, "not ok"),
-			expected: fwtypes.ARNUnknown(),
+			expected: fwtypes.ARNValue("not ok"),
 		},
 	}
 


johnsonaj (approved):
LGTM 🚀

PR #40017, opened 2024-11-05
website/docs/functions: remove required version note

jar-b:
### Description

The registry now includes a banner above all functions pages indicating that provider defined functions are only supported in Terraform 1.8 and later. This now makes the notes embedded in the pages themselves redundant.

image


johnsonaj (approved):
LGTM 🚀

PR #39988, opened 2024-11-01
lambda: Ephemeral resource `aws_lambda_invocation`

YakDriver:
### Description

Thoughts on an ephemeral resource for Lambda function invocations?

### Relations

Depends #39835
Depends #40203 
Relates #40009

### Output from Acceptance Testing

```console
% make t T=TestAccLambdaInvocationEphemeral_basic K=lambda
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.23.2 test ./internal/service/lambda/... -v -count 1 -parallel 20 -run='TestAccLambdaInvocationEphemeral_basic'  -timeout 360m
2024/11/20 16:50:43 Initializing Terraform AWS Provider...
=== RUN   TestAccLambdaInvocationEphemeral_basic
=== PAUSE TestAccLambdaInvocationEphemeral_basic
=== CONT  TestAccLambdaInvocationEphemeral_basic
--- PASS: TestAccLambdaInvocationEphemeral_basic (27.79s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/lambda	31.716s
```

PR #40018, opened 2024-11-05
autoflex: Allow []byte-string conversion

YakDriver:
### Relations

Relates #39988 

### Output from Testing

```console
% cd internal/framework/flex
% go test ./...
ok  	github.com/hashicorp/terraform-provider-aws/internal/framework/flex	0.272s
```


jar-b (approved):
LGTM 🚀

PR #40087, opened 2024-11-11
Increase operation count of stale workflow

justinretzolk:
### Description

I noticed some items with the  `stale` label that had been open for longer than the prescribed 30 additional days. Upon review, the stale workflow [is finishing](https://github.com/hashicorp/terraform-provider-aws/actions/runs/11746909216) with a warning indicating that we should increase [`operations-per-run`](https://github.com/actions/stale?tab=readme-ov-file#operations-per-run).

This PR increases the operations count by 50 to see if that's enough. Will monitor and increase further as needed.

N/a, workflow


jar-b (approved):
LGTM 🚀

PR #40096, opened 2024-11-12
fix typo errors in document for semantic_chunking_configuration block

### LINKED ISSUE #39819: [Docs]: `max_tokens` argument in `aws_bedrockagent_data_source`
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/5.72.1/docs/resources/bedrockagent_data_source

### Description

The documentation refers to the `max_tokens` argument several times, but the correct argument is `max_token`.

```
Error: Missing required argument

  on bedrock/main.tf line 128, in resource "aws_bedrockagent_data_source" "unstructured_data_source":
 128:       semantic_chunking_configuration {

The argument "max_token" is required, but no definition was found.

Error: Unsupported argument

  on bedrock/main.tf line 131, in resource "aws_bedrockagent_data_source" "unstructured_data_source":
 131:         max_tokens = 300

An argument named "max_tokens" is not expected here. Did you mean
"max_token"?
```

### References

_No response_

None

thiennguyen56:
### Description
I have updated the max_tokens -> max_token of sematic_chunking_configuration block in documents

### Relations

Closes #39819 

```console

...
```

### DIFF
--- a/website/docs/r/bedrockagent_data_source.html.markdown
+++ b/website/docs/r/bedrockagent_data_source.html.markdown
@@ -106,7 +106,7 @@ The `semantic_chunking_configuration` block supports the following arguments:
 
 * `breakpoint_percentile_threshold` - (Required, Forces new resource) The dissimilarity threshold for splitting chunks.
 * `buffer_size` - (Required, Forces new resource) The buffer size.
-* `max_tokens` - (Required, Forces new resource) The maximum number of tokens a chunk can contain.
+* `max_token` - (Required, Forces new resource) The maximum number of tokens a chunk can contain.
 
 ### `custom_transformation_configuration` block
 


jar-b:
Thanks for the catch, @thiennguyen56. I believe the Terraform schema is actually incorrect in this case as the corresponding AWS data structure has a `maxTokens` field:
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_SemanticChunkingConfiguration.html#bedrock-Type-agent_SemanticChunkingConfiguration-maxTokens

That said, to fix the schema we'd need to introduce a new `max_tokens` argument, deprecate `max_token`, and handle both variants within AutoFlex (something I'm not sure is currently possible). At this point I think correcting the documentation is sufficient absent some other issue preventing the misnamed schema attribute from operating normally.

jar-b (approved):
LGTM 🚀

jar-b:
Thanks for your contribution, @thiennguyen56! 👍

thiennguyen56:
@jar-b yeah I had seen struct of sematic_chunking_configuration block in code, it look like this 
`
type semanticChunkingConfigurationModel struct {
	BreakpointPercentileThreshold types.Int32 `tfsdk:"breakpoint_percentile_threshold"`
	BufferSize                    types.Int32 `tfsdk:"buffer_size"`
	MaxTokens                     types.Int32 `tfsdk:"max_token"`
}
`
I reckon that the tfsdk tag is mismatched, so we need to create a new field to deprecate it?

jar-b:
That's correct. I'm going to add this to the agenda in our team meeting later this week to discuss how this looks with AutoFlex. With legacy flatten/expand functions we can simply add handling for both cases, but because AutoFlex handles fuzzily matching field names we'll need to investigate whether it can correctly handle multiple schema fields mapping to the same underlying field in the struct.

PR #40120, opened 2024-11-13
Correct typo in elasticache_replication_group regarding number of allowed Secondary replica groups

xmclallen:
Prior typo was confusing and had no semantic meaning. Updated to be accurate per AWS documentation

https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/Redis-Global-Datastores-Getting-Started.html

The above documentation from AWS states:
"You can set up replication for a primary cluster from one AWS Region to a secondary cluster in up to two other AWS Regions. Note: The exception to this are China (Beijing) Region and China (Ningxia) regions, where replication can only occur between the two regions."

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/website/docs/r/elasticache_replication_group.html.markdown
+++ b/website/docs/r/elasticache_replication_group.html.markdown
@@ -130,7 +130,7 @@ for full details on using Replication Groups.
 
 ### Creating a secondary replication group for a global replication group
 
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🚀

jar-b:
Thanks for the correction, @xmclallen! 👍

Issue #29817, opened 2023-03-06
[Enhancement]: resource/aws_batch_job_description use block for `container_properties` instead of JSON string


gdavison:
### Description

Presumably for historical reasons, the `container_properties` field of `aws_batch_job_description` is represented as a JSON string, instead of using a block. This defers validation of the actual fields to apply-time. It also leads to confusing error messages, as in #14024.

### Affected Resource(s) and/or Data Source(s)

aws_batch_job_description

### Would you like to implement a fix?

None

PR #40149, opened 2024-11-15
[Help Wanted] Migrate Batch Job Definition to Framework + AutoFlex

danquack:
### Description

Looking to migrate this resource to framework. A lot of the code can be removed with auto flex, and will make migration on #29817 a little easier down the line. I'm currently running into a couple bugs that I am struggling to fix, and would love some assistance on.

- [ ] SDK uses `DiffSuppressFunc` to do comparison of the `container_properties` `node_properties` and `ecs_properties`. Currently when writing back to state, the sdk will add the empty arrays, so the strings are different. 
```
 new value: .container_properties: was
        cty.StringVal("{\"command\":[\"echo\",\"test\"],\"image\":\"busybox\",\"memory\":128,\"vcpus\":1}"),
        but now
        cty.StringVal("{\"command\":[\"echo\",\"test\"],\"environment\":[],\"image\":\"busybox\",\"memory\":128,\"mountPoints\":[],\"resourceRequirements\":[],\"secrets\":[],\"ulimits\":[],\"vcpus\":1,\"volumes\":[]}").
```

Maybe [relevant code](https://github.com/hashicorp/terraform-provider-aws/blob/37786fefbe3d4855608345568a00f7f629e466da/internal/service/batch/ecs_properties.go#L28-L69)

- [ ] Null to empty. I think this is something to do with a missing upgrade function, but adding the `legacy` tag on auto flex doesn't seem to resolve it.
```
.parameters: was null, but now cty.MapValEmpty(cty.String).
```

```console

...
```

### DIFF
--- a/internal/service/batch/exports_test.go
+++ b/internal/service/batch/exports_test.go
@@ -6,7 +6,7 @@ package batch
 // Exports for use in tests only.
 var (
 	ResourceComputeEnvironment = resourceComputeEnvironment
-	ResourceJobDefinition      = resourceJobDefinition
+	ResourceJobDefinition      = newResourceJobDefinition
 	ResourceJobQueue           = newJobQueueResource
 	ResourceSchedulingPolicy   = resourceSchedulingPolicy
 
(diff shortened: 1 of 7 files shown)


danquack:
Going to do the full migration with the resource attribute change. The string format gets manipulated not only with suppression, but also some reduction. Since the end goal is to move to a schema defined attribute, a new PR will be opened with the full schema + framework migration.

PR #40203, opened 2024-11-19
testing: add echo provider for testing ephemeral resources

johnsonaj:
### Description

- adds echo provider for easier testing on ephemeral resources
- add logging helper to mask sensitive values by key

```console

...
```

### DIFF
--- a/internal/logging/keys.go
+++ b/internal/logging/keys.go
@@ -3,6 +3,30 @@
 
 package logging
 
-const (
-	KeyResourceId = "id"
+import (
+	"context"
+
+	baselogging "github.com/hashicorp/aws-sdk-go-base/v2/logging"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
+
+const (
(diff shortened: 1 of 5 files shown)


internal/logging/keys.go
```diff
@@ -3,6 +3,30 @@
...
+
 const (
-	KeyResourceId = "id"
+	HTTPKeyRequestBody  = "http.request.body"
+	HTTPKeyResponseBody = "http.response.body"
+	KeyResourceId       = "id"
 )
+
+// MaskSensitiveValuesByKey masks sensitive values using tflog
+func MaskSensitiveValuesByKey(ctx context.Context, keys ...string) context.Context {
```
YakDriver, line 20: Still feel uneasy about this but I guess we wait on guidance from upstream?

YakDriver (approved):
LGTM 🎉

internal/logging/keys.go
johnsonaj: yeah, upstream needs to figure out logging/tracing as well.

PR #40213, opened 2024-11-20
.github/workflows: address breaking changes to `slack-github-action`

jar-b:
### Description

The `webhook-type` argument is now required. The `webhook` argument is optional and can also be configured by setting the `SLACK_WEBHOOK_URL` environment variable, but I've migrated to the explicit argument to align with the documentation for the release. See the [release notes](https://github.com/slackapi/slack-github-action/releases/tag/v2.0.0) for complete details on the breaking changes.

### Relations

Relates #40160 

### References

- https://github.com/slackapi/slack-github-action/releases/tag/v2.0.0


justinretzolk (approved):
This looks good to me. I'm a little 🤔 on why a couple of them didn't have the same environment variables set, but were still working. Something I'll look into more after we get the immediate problem solved.

PR #40253, opened 2024-11-22
r/chatbot: Fix SNS topic ARNs ordering in Slack channel configuration

max-ts0gt:
### Description
This PR fixes an inconsistency in the aws_chatbot_slack_channel_configuration resource where it unnecessarily triggers plan changes when SNS topics are returned in a different order from the AWS API.

The fix implements sorting of SNS topic ARNs before comparison in the slackChannelConfigurationHasChanges function, ensuring that the order of topics doesn't trigger unnecessary updates.

Before this fix:
Plan would show changes when AWS returns topics in different order:
```
sns_topic_arns = [
    - "arn:aws:sns:us-west-2:123456789012:topic2"
    - "arn:aws:sns:us-west-2:123456789012:topic1"
    - "arn:aws:sns:us-west-2:123456789012:topic1"
    - "arn:aws:sns:us-west-2:123456789012:topic2"
]
```
After this fix:
No changes shown when only order differs:
```
sns_topic_arns = [
    - "arn:aws:sns:us-west-2:123456789012:topic1"
    - "arn:aws:sns:us-west-2:123456789012:topic2"
]
```

Closes #39025
Closes #39870

PR #40291, opened 2024-11-25
r/chatbot_teams_channel_configuration: Fix SNS topic ARNs ordering

johnsonaj:
### Description

Changes to autflex so that list ordering is not considered a diff. 

Best effort changes since we do not have a teams application configured

### Relations

Relates #40253

```console

...
```


ewbankkit (approved):
LGTM :rocket:.

PR #40341, opened 2024-11-28
docs(r-lb-listener-rule): Notes the relationship between a tag key `Name` to the AWS Console

### LINKED ISSUE #36386: [Enhancement]: Include disclaimer about name tag field for aws_lb_listener_rule resource
### Description

**What I have found:** Today when I created the AWS ELB listener rule it works but I can't find the instruction to fulfill the the Name Tag. After searching for this option in a AWS console, I notice the name tag had the same value for tags. Of course this is a small improvement, but it can save time during the implementation process.

**What my suggestion:** Include a disclaimer in a lb_listener_rule page to explain to users if they want to use the name tag field they should use the tags.

Thank you.

lb_listener_rule

```terraform
Not Applicable
```

### References

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener_rule

None

Bexanderthebex:
### Description

It has been pointed out by a community member that when a key `Name` is defined in the lb_listener_rule resource, it gets picked up by the AWS Console and displays the value within the `Name Tag` column. In order to save time for other users so that they don't perform the same investigation as the reporter did, this PR documents that quirk.

Screenshot 2024-11-28 at 4 13 43 PM

### Closes

Closes #36386


ewbankkit:
@Bexanderthebex Thanks for the contribution :tada: :clap:.

PR #40355, opened 2024-11-28
sweeper/aws_docdb_cluster: Fix error when sweeping members of global cluster

gdavison:
### Description

When sweeping an `aws_docdb_cluster` which is a member of a global cluster, the cluster `arn` value is needed. Otherwise it returns an error

> removing DocumentDB Cluster () from DocumentDB Global Cluster (tf-acc-test-7157141747640475231-0): operation error DocDB: RemoveFromGlobalCluster, https response error StatusCode: 400, api error InvalidParameterValue: Db cluster identifier  is not a valid arn

Note that the "()" should contain the ARN of the cluster.

### DIFF
--- a/internal/service/docdb/sweep.go
+++ b/internal/service/docdb/sweep.go
@@ -4,6 +4,7 @@
 package docdb
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strings"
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

PR #40403, opened 2024-12-03
d/aws_standards_control_association: Fix resource name in docs

henkka:
### Description

`aws_standards_control_association` resource does not exist in the provider. Instead, the resource in question is `aws_securityhub_standards_control_association`


jar-b (approved):
LGTM 🚀

jar-b:
Thanks for the fix, @henkka! 👍

PR #40245, opened 2024-11-21
.github/workflows: fix formatting in `post_publish` shell script

jar-b:
### Description

A missed quote caused the script to fail during the `v5.77.0` release.

```
/home/runner/work/_temp/b9a14e53-10ad-497f-a91b-d57575e541f4.sh: line 12: unexpected EOF while looking for matching `"'
Error: Process completed with exit code 2.
```

Ref: https://github.com/hashicorp/terraform-provider-aws/actions/runs/11959899920/job/33342735174

### Relations

Relates #39964

PR #40406, opened 2024-12-03
.github/workflows: fix registry version check comparison

jar-b:
### Description

The version returned from the `curl` request to the registry will contain only the version number, while the release tag is prefixed with `v`.

For example, here is what is stored in the `LATEST_VERSION` environment variable:

```console
% curl -s "https://registry.terraform.io/v2/providers/323/provider-versions/latest" | jq -r '.data.attributes.version'
5.79.0
```

This changes the conditional to apply a prefix to the version number fetched from the registry to allow an accurate comparison of the two values.

### Relations

Relates #40245 
Relates #39964 

https://github.com/hashicorp/terraform-provider-aws/actions/runs/12038203742/job/33563358216


ewbankkit (approved):
LGTM :rocket:.

Issue #39955, opened 2024-10-30
[Bug]: Can not disable `transit encryption` and `at_rest_encryption`


EugenKon:
### Terraform Core Version

v1.9.8

### AWS Provider Version

v5.67.0

### Affected Resource(s)

* aws_elasticache_replication_group

### Expected Behavior

When `at_rest_encryption_enabled = true` and `transit_encryption_enabled = true` options are removed terraform should detect that.

Default values should be described here:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group#transit_encryption_enabled

### Actual Behavior

Nothing is detected.
No changes. Your infrastructure matches the configuration.

But:
![image](https://github.com/user-attachments/assets/457eabe0-bd9a-4e62-8bd6-3bf3bc83ba52)

### Terraform Configuration Files

```
resource "aws_elasticache_replication_group" "redis" {
  replication_group_id  = "redis"
  description           = "Redis cluster"
  subnet_group_name     = aws_elasticache_subnet_group.redis.name

  node_type                  = "cache.t2.micro"
  port                       = 6379
  parameter_group_name       = "default.redis7.cluster.on"
  automatic_failover_enabled = true
  num_cache_clusters         = 2

  at_rest_encryption_enabled = true
  transit_encryption_enabled = true
}
```

### Steps to Reproduce

Create Redis cluster with `at_rest_encryption_enabled = true` and `transit_encryption_enabled = true` options
Remove them
terraform plan/apply

### Would you like to implement a fix?

None

rsafonseca:
On top of this, the default for at_rest_encryption_enabled has changed in the AWS API, but is not reflected in the API Docs, just in the Elasticache docs. I've already raised this with AWS to fix their docs.

To make it worse, when you set `at_rest_encryption_enabled = false` (or when you don't set it al all), it shows up in the plan that will be set to false, but in fact get set to true, which makes it appear on every run and require the ReplicationGroup to be recreated, rendering it unmanageable unless you explicitly set it to true.

rsafonseca:
Hi @gdavison , thanks for taking this up.
Your MR only fixes one of the problems, which is the perpetual loop. This property should still be configurable beyond the default though and currently because it's set to Computed it cannot be configured.

PR #40460, opened 2024-12-05
Fix: at_rest_encryption_enabled incorrectly processed

rsafonseca:
### Description
at_rest_encryption_enabled incorrectly processed leading to unexpected replication group state and endless replacement when set to false

Current situation:
- AWS docs specify that the default is false, but this has changed on the API for valkey, new ReplicationGroups are all being created with at_rest_encryption_enabled set to true
- By default, this provider assumes at_rest_encryption_enabled=false, and doesn't include the parameter in the CreateReplicationGroup API call
- When creating a Valkey ReplicationGroup with the provider, unless explicitly specifying  `at_rest_encryption_enabled=true`, every terraform plan/apply will try to recreate the ReplicationGroup, which is a destructive operation

This PR changes `at_rest_encryption_enabled` from Computed to a normal parameter, so it is included with the CreateReplicationGroup API call
### Relations

Relates #39955

<!--
Replace TestAccXXX with a pattern that matches the tests affected by this PR.

Replace ec2 with the service package corresponding to your tests.

For more information on the `-run` flag, see the `go test` documentation at https://tip.golang.org/cmd/go/#hdr-Testing_flags.

```console

...
-->


```

### DIFF
--- a/internal/service/elasticache/replication_group.go
+++ b/internal/service/elasticache/replication_group.go
@@ -70,7 +70,7 @@ func resourceReplicationGroup() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				ForceNew: true,
-				Default: true,
+				Default:  true,
 			},
 			"auth_token": {
 				Type:          schema.TypeString,
--- a/internal/service/elasticache/replication_group.go
+++ b/internal/service/elasticache/replication_group.go
@@ -70,7 +70,7 @@ func resourceReplicationGroup() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				ForceNew: true,
-				Computed: true,
+				Default: true,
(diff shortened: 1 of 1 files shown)


gdavison:
Hi @rsafonseca. Thanks for contributing a fix for an issue. Unfortunately, we won't be able to use it. Changing this default value will be a breaking change and will also cause problems for existing deployments. In addition, due to a known issue with Boolean values and the Terraform Plugin SDK, `at_rest_encryption_enabled` would no longer be able to be disabled.

I've created #40514 to fix this issue

rsafonseca:
Hi @gdavison , thanks for taking this up.
Your MR only fixes one of the problems, which is the perpetual loop. This property should still be configurable beyond the default though and currently because it's set to Computed it cannot be configured.

PR #40464, opened 2024-12-05
.github/workflows: exit successfully once registry and github versions match

jar-b:
### Description

Previously a version match would print the correct success message, but did not exit after doing so. This caused the loop to continue and eventually fall through to a failure message with a non-zero exit code.

Ref: https://github.com/hashicorp/terraform-provider-aws/actions/runs/12164517617/job/33926378837

### Relations

Relates #40406 
Relates #40245 
Relates #39964


ewbankkit (approved):
LGTM :rocket:.

PR #40472, opened 2024-12-05
ci: Remove on-premise related parameters

gdavison:
### Description

Removes parameter overrides that were only needed on the old on-premise setup


ewbankkit (approved):
LGTM :rocket:.

PR #40513, opened 2024-12-10
Removes unused `Check`s when test step has `ExpectError`

gdavison:
### Description

Acceptance test steps with `ExpectError` ignore any `Check`s. Adds Semgrep rule to prevent `Check`s when `ExpectError` is set


ewbankkit (approved):
LGTM :rocket:.

PR #40515, opened 2024-12-11
Fix incorrect word in eks_access_policy_association import documention.

jpterry:
Substitute incorrect "a colon" for more correct "an octothorp" in eks_access_policy_association import instructions, so that english sentence matches the referenced symbol.

### Description
Updates the eks_access_policy_association import documentation to use the correct word for the separator in the import id.
The examples, and other text refer to `#` as the correct separator. I've also proven that `#` is the expected separator for this import.

### References
`#` is called an "octothorp". It is also known by other names as well. I checked and I don't see any consistency issues with using "octothorp", but if "pound sign" or "hashtag symbol" or something else is more consistent, I'm happy to update.

...

I am unsure how to properly build the docs, but I will look into it.

### DIFF
--- a/website/docs/r/eks_access_policy_association.html.markdown
+++ b/website/docs/r/eks_access_policy_association.html.markdown
@@ -64,7 +64,7 @@ The `associated_access_policy` block has the following attributes.
 
 ## Import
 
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@jpterry Thanks for the contribution :tada: :clap:.

PR #40557, opened 2024-12-13
kms: add support for external key store

### LINKED ISSUE #28092: AWS KMS External Key Store
> Today, [AWS Key Management Service (AWS KMS)](https://aws.amazon.com/kms/) introduces the External Key Store (XKS), a new feature for customers who want to protect their data with encryption keys stored in an external key management system under their control.

[Announcement](https://aws.amazon.com/about-aws/whats-new/2022/11/aws-kms-external-key-store/).
[Blog post](https://aws.amazon.com/blogs/aws/announcing-aws-kms-external-key-store-xks/).

Requires [AWS SDK for Go v1.44.148](https://github.com/aws/aws-sdk-go/releases/tag/v1.44.148): https://github.com/hashicorp/terraform-provider-aws/pull/28085.

### Affected Resource(s)

* [`aws_kms_custom_key_store`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/kms_custom_key_store)

bschaatsbergen:
Closes #28092

* Adds XKS (External Key Store) arguments to the `aws_kms_custom_key_store` resource
* Changes CloudHSM related arguments from Required to Optional.
* Updates documentation with CloudHSM and XKS (public and private) examples.


internal/service/kms/custom_key_store.go
```diff
@@ -198,3 +303,39 @@ func findCustomKeyStores(ctx context.Context, conn *kms.Client, input *kms.Descr
...
+	if apiObject == nil {
+		return []interface{}{}
+	}
+
+	tfMap := map[string]interface{}{
+		"access_key_id": aws.ToString(apiObject.AccessKeyId),
+	}
+
+	return []interface{}{tfMap}
+}
```
jar-b, line 341: I'm wondering if setting only `access_key_id` here will cause perpetual differences if `raw_secret_access_key` is set in the configuration. We could either omit the flattener entirely or perhaps set only `xks_proxy_authentication_credential.access_key_id` rather than the whole block.

Do you have a preference either way @bschaatsbergen?

bschaatsbergen: Thanks for spotting this, @jar-b. With the latter approach, are you suggesting that we only set `xks_proxy_authentication_credential.access_key_id` during creation to avoid any drift from the flattening function, since it already excludes raw_secret_access_key? Given that both `access_key_id` and `raw_secret_access_key` are required, I think technically we could skip the flattening, as these values are always persisted to state during creation?

jar-b: > With the latter approach, where we only set `xks_proxy_authentication_credential.access_key_id`, how would you go about implementing that? Would you use something like `d.Set("xks_proxy_authentication_credential.0.access_key_id", output.XksProxyConfiguration.AccessKeyId)`? I’m just trying to get a better understanding.

Yes, exactly. That would avoid overwriting all of `xks_proxy_authentication_credential` on the `Set` call. 

That said, I'm also leaning toward omitting the flattener for this initial implementation as it removes a path for unexpected behavior given we can't acceptance test this. If we find users miss/want drift detection on the `xks_proxy_authentication_credential.access_key_id` argument (the only argument we can detect drift on since it's all AWS returns), we can revisit and add a flattener for that argument in a future release.

jar-b (approved):
LGTM 🚀

**Community Note**:  Due to the dependency on an external key store, we are unable to acceptance test the changes made in this pull request. Any community feedback in testing these new arguments would be greatly appreciated!

jar-b:
Thanks for your contribution, @bschaatsbergen! 👍

PR #40619, opened 2024-12-18
docs: refine terraform plugin package verbiage

jar-b:
### Description

This page is now more explicit about the requirement for all new resources and data sources to be written with Terraform Plugin Framework. It also switches the ordering of the plugin libraries such that Terraform Plugin Framework is always mentioned first.


ewbankkit (approved):
LGTM :rocket:.

PR #40695, opened 2024-12-25
Cloudfront distribution logging config documentation update

MykKuc:
### Description
This is the documentation change for `logging_config` block existing in the `aws_cloudfront_distribution` page. This documentation change specifies that the version used for CloudWatch Standard (Access) logs is Legacy instead of v2. I think this change makes it more clear for the newer AWS provider and CLoudFront user.

### References
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html
image


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@MykKuc Thanks for the contribution :tada: :clap:.

Issue #35559, opened 2024-01-30
[Bug]: Service Discovery health check custom config failure threshold deprecated


fszymanski-blvd:
### Terraform Core Version

1.6.6

### AWS Provider Version

4.67.0

### Affected Resource(s)

aws_service_discovery_service

### Expected Behavior

Setting the failure_threshold to any value other than 1 and running multiple times should not produce plan changes

### Actual Behavior

When running a subsequent plan, the service discovery service is marked as needing to be replaced because the failure_threshold is set to 1 rather than the other value that is defined.

### Terraform Configuration Files

```terraform
resource "aws_service_discovery_public_dns_namespace" "example" {
  name        = "example.terraform.com"
  description = "example"
}

resource "aws_service_discovery_service" "example" {
  name = "example"

  dns_config {
    namespace_id = aws_service_discovery_public_dns_namespace.example.id

    dns_records {
      ttl  = 10
      type = "A"
    }
  }

  health_check_custom_config {
    failure_threshold = 5
  }
}
```

### Steps to Reproduce

1. Create a aws_service_discovery_service with a health_check_custom_config containing a failure_threshold set to a value other than 1
2. Apply
3. Run another plan with the same config

### Important Factoids

The Failure Threshold parameter is marked as deprecated and is always set to 1

### References

https://docs.aws.amazon.com/cloud-map/latest/api/API_HealthCheckCustomConfig.html

### Would you like to implement a fix?

None

tvaleriy:
Expected behavior might be better to change to something like "Setting failure_threshold to any value other than 1 should result in an error." IMO.

fszymanski-blvd:
> Expected behavior might be better to change to something like "Setting `health_check_custom_config {failure_threshold = 1}` to any value other than 1 should result in an error." IMO.

This would make sense, but any existing service discovery services that had different values will still have them. If Terraform enforces the value of 1, it would require major changes to fix the existing resources since changes to this value require recreating the service discovery service and recreation requires the service is empty.

yermulnik:
@fszymanski-blvd On the other hand silently discarding user-provided value and hence making user think their changes were successfully accepted by the AWS API (including perpetual changes in TF plan and hence drift) is even more evil as opposite to throwing an error to let user know there's something wrong with their configuration.
I wouldn't even suggest a deprecation notice to be thrown by TF since many of us rely upon automation that might be not accounting deprecation notices as something that is worth of human engineer attention.
[...]

PR #40777, opened 2025-01-05
r/aws_service_discovery_service: Deprecates `health_check_custom_config.failure_threshold`

### LINKED ISSUE #37955: [Attribute Deprecation] r/aws_service_discovery_service: Deprecate `health_check_custom_config.failure_threshold
The `aws_service_discovery_service`'s `health_check_custom_config.failure_threshold` argument is deprecated in the [AWS API](https://docs.aws.amazon.com/cloud-map/latest/api/API_HealthCheckCustomConfig.html#API_HealthCheckCustomConfig_Contents):

Screenshot 2024-06-13 at 7 11 01 AM

Deprecate the argument in the provider.

Relates https://github.com/hashicorp/terraform-provider-aws/issues/35559.

Bexanderthebex:
### Description

This PR will try to resolve the issue with the `aws_service_discovery_service` resource where setting the `health_check_custom_config.failure_threshold` to a value other than `1` results into the following behaviors:

- During resource creation, the value is getting ignored as the AWS SDK overrides it to `1`
- After the initial resource creation, the resource always gets recreated as terraform detects that the value written in the terraform file is not on par with the setting in the AWS Cloudformation Stack. This causes uninteded `terraform apply` failures since terraform will try to destroy and then recreate the resource again even with tight depedencies that terraform won't simply allow.

### Relations

Closes #37955 

### References

#35559 

```console

...
```

### DIFF
--- a/internal/service/servicediscovery/service.go
+++ b/internal/service/servicediscovery/service.go
@@ -126,7 +126,7 @@ func resourceService() *schema.Resource {
 							Type:       schema.TypeInt,
 							Optional:   true,
 							ForceNew:   true,
-							Deprecated: `The attribute "failure_threshold" is now unsupported in the AWS API and is always set to 1. The attribute will be removed in a future major version.`,
+							Deprecated: `"failure_threshold" is deprecated and no longer supported by AWS. The attribute will be removed in a future major version.`,
 						},
 					},
 				},
--- a/internal/service/servicediscovery/service.go
+++ b/internal/service/servicediscovery/service.go
@@ -126,7 +126,7 @@ func resourceService() *schema.Resource {
 							Type:       schema.TypeInt,
 							Optional:   true,
 							ForceNew:   true,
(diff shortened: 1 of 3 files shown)


internal/service/servicediscovery/service.go
```diff
@@ -123,9 +123,10 @@ func resourceService() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"failure_threshold": {
-							Type:     schema.TypeInt,
-							Optional: true,
-							ForceNew: true,
+							Type:       schema.TypeInt,
+							Optional:   true,
+							ForceNew:   true,
+							Deprecated: "failure_threshold is deprecated and no longer supported by AWS. The attribute will be removed in a future major version.",
```
jar-b, line 129: We should maybe mention here that AWS always sets this value to `1` since it'll display any time the argument is set. 

That way if the user misses the registry documentation or CHANGELOG they'll still have appropriate expectations for the behavior.

jar-b (approved):
LGTM 🚀

PR #40778, opened 2025-01-05
Add a comment to default_ttl, min and max ttl arguments in `aws_cloudfront_distribution` resource telling that TTL configuration coming from cache policy overrides other TTL arguments

MykKuc:
### Description
In the resource `aws_cloudfront_distribution` in configuration blocks `default_cache_behavior` and `ordered_cache_behavior` there exists arguments `default_ttl`,`min_ttl` and `max_ttl`. These same arguments also exist in resource `aws_cloudfront_cache_policy` and when setting up behaviors and providing argument `cache_policy_id`  it is possible to leave to sets of TTL configurations. For that reason I thought it would be good to clarify that TTL arguments from `aws_cloudfront_cache_policy` take priority.

Correct me if I'm wrong or there is some better way to explain this in the docs.


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@MykKuc Thanks for the contribution :tada: :clap:.

PR #40822, opened 2025-01-07
testing: Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes: `a` services

gdavison:
### Description

Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes in favour of explicit checks


ewbankkit (approved):
LGTM :rocket:.

PR #40854, opened 2025-01-09
testing: Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes: `b` and `c` services

gdavison:
### Description

Addresses `b` and most of the `c` services


ewbankkit (approved):
LGTM :rocket:.

Issue #40874, opened 2025-01-10
[Enhancement]: Make organization api more flexible


vladimir-romashov:
### Description

Currently, `aws_organizations_organization` makes calls in such order:

_output, err := conn.DescribeOrganization(ctx, input) [Link1](https://github.com/hashicorp/terraform-provider-aws/blob/62e3d25b6538d91a2545226f6f63fc40545b4c43/internal/service/organizations/organization.go#L463)_
if not failed then
_pages := organizations.NewListAccountsPaginator(conn, input) [link2](https://github.com/hashicorp/terraform-provider-aws/blob/62e3d25b6538d91a2545226f6f63fc40545b4c43/internal/service/organizations/organization.go#L486)_
if not failed then
_pages := organizations.NewListRootsPaginator(conn, input)  [link3](https://github.com/hashicorp/terraform-provider-aws/blob/62e3d25b6538d91a2545226f6f63fc40545b4c43/internal/service/organizations/organization.go#L534)_

If I only need to retrieve organizations.id, I don't need information from ListAccounts and/or ListRoots. 

Why is that a problem? - AWS imposes [limits](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_reference_limits.html) for their APIs:
![image](https://github.com/user-attachments/assets/03b73825-51de-404f-882f-9df69490aa60)

### Affected Resource(s) and/or Data Source(s)

aws_organizations_organization

### Would you like to implement a fix?

Yes

ewbankkit:
`ListRoots` also has a very low API limit:
<img width="680" height="625" alt="Image" src="https://github.com/user-attachments/assets/ac97cb65-0e4a-492d-a325-4d4883460361" />

PR #40884, opened 2025-01-12
feat: organization isolated describe organization

vladimir-romashov:
### Description
This is a rough draft of a naive attempt to solve #40874. The first question I have for the community is:
- currently, `organizations` pull data not only from `DescribeOrganization`, but also from `ListAccounts` and `ListRoots`. This is an attempt to limit this to just `DescribeOrganization` by introducing `aws_organizations_describe_organization` next to the generic `aws_organizations_organizations`. Is this the right approach or do we want to refactor the existing `aws_organizations_organization`?

### Relations
Related #40874

### DIFF
--- a/internal/service/organizations/organization_describe_organization.go
+++ b/internal/service/organizations/organization_describe_organization.go
@@ -1,0 +1,215 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package organizations
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


DrFaust92:
[vladimir-romashov](https://github.com/vladimir-romashov). I Believe it would less confusing and less difficult to maintain if this is logic in the existing data source

ewbankkit:
@vladimir-romashov Thanks for the contribution :tada: :clap:.

PR #40907, opened 2025-01-13
Fix redishift issue 20944

ivankatliarchuk:
### Relations

Closes #0000

```console

...
```


cjtim:
+1 bumping this MR. This would fix my issue applying TF when cluster is in paused state.

ivankatliarchuk:
Yeah, terraform at the moment not an option if one cluster is paused.

PR #40913, opened 2025-01-14
Tech debt: Require `name=` value for `@SDKResource` annotation

ewbankkit:
### Description

Enforce setting a friendly name for every resource declared with the `@SDKResource` annotation.
Backfill those resource that were missing such a value.


jar-b (approved):
LGTM 🚀

PR #40917, opened 2025-01-14
Tech debt: Require `name=` value for `@SDKDataSource` annotation

ewbankkit:
### Description

Enforce setting a friendly name for every data source declared with the `@SDKDataSource` annotation.
Backfill those data sources that were missing such a value.

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/40913.


jar-b (approved):
LGTM 🚀

PR #40930, opened 2025-01-15
testing: Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes: `f` to `i` services

gdavison:
### Description

Addresses `f` through `i` services

Adds `arn` attribute to `globalaccelerator` resources


ewbankkit (approved):
LGTM :rocket:.

PR #40993, opened 2025-01-17
testing: Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes: `k` to `m` services

gdavison:
### Description

Addresses `k` through `m` services


ewbankkit (approved):
LGTM :rocket:.

PR #41107, opened 2025-01-27
testing: Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes: `oam`

gdavison:
### Description

Reduce uses of `resource.TestCheckResourceAttrSet` on ARN attributes in favour of explicit checks.

Also:
* replaces use of `id` in acceptance tests and documentation with `arn`
* adds `depends_on` relationship with `aws_oam_sink_policy` to prevent intermittent failures


jar-b (approved):
LGTM 🚀

PR #41109, opened 2025-01-27
Tech debt: Generate AWS API clients with override Regions

ewbankkit:
### Description

Enhances the AWS API client generation to support override Regions.

```console

...
```


internal/generate/servicepackage/file.gtpl
```diff
@@ -132,11 +135,40 @@ func (p *servicePackage) ServicePackageName() string {
...
+}
+
+func servicePackageExtraOptFns(ctx context.Context, sp conns.ServicePackage, config map[string]any) []func(*{{ .GoV2Package }}.Options) {
+	if v, ok := sp.(interface {
+		extraOptFns(context.Context, map[string]any) []func(*{{ .GoV2Package }}.Options)
+	}); ok {
+		return v.extraOptFns(ctx, config)
+	}
+
+	return nil
```
jar-b, line 171: Will these "extra" options be exported and referenced elsewhere in the future? I find the client setup a bit harder to read with this function called via a separate `append` operation, but understand if the tradeoff is necessary to future proof for multi-region support.

names/data/names_data.hcl
```diff
@@ -2477,13 +2339,11 @@ service "costoptimizationhub" {
...
-    skip_client_generate = true
-  }
-
   endpoint_info {
-    endpoint_api_call        = "GetPreferences"
-    endpoint_region_override = "us-east-1"
+    endpoint_api_call = "GetPreferences"
+    endpoint_region_overrides = {
+      "aws" = "us-east-1"
+    }
```
jar-b, line 2346: Is scoping these override to the commercial partition potentially a breaking change for other partitions? Perhaps we include a changelog entry noting the change?

internal/generate/servicepackage/file.gtpl
ewbankkit: The immediate reason is so that we can generate most of the API client factory method for those dozen or so API services that are doing it manually -- they are doing that because of setting up custom retry handling which is done via an `Options` function. This will allow that functionality to be moved into a per-service `extraOptFns` method.
Agreed that the `append` is not nice, would help if you could do something like this in Go:

```go
optFns := []func(*Options) {
    blah,
   extraOptFns(ctx, config)...,
}
```

names/data/names_data.hcl
ewbankkit: The manually written API client factories always did a partition check for `aws` (and others for Route 53) -- this map captures that.

internal/generate/servicepackage/file.gtpl
ewbankkit: Realistically we could just have at most one extra `OptFn` per service and insert it in the slice initializer and instead of returning `nil` by default, return a no-op `OptFn`.

ewbankkit: Added `withExtraOptions` which can now be called directly in the slice initializer.

jar-b (approved):
LGTM 🚀

PR #41240, opened 2025-02-05
aws_lambda_invocation import support

Giaco9NN:
### Description

This pull requests allows to import an aws_lambda_invocation resource.
The imported resource will be updated in the next apply because it is not possible to understand properties like input, terraform_key, lifecycle_scope, result and triggers starting from the resource id, which is made by the function name, the qualifier and md5 of the return value.

```console
% make testacc TESTS=TestAccLambdaInvocation_ PKG=lambda

...
```


ewbankkit (approved):
LGTM :rocket:.

PR #41242, opened 2025-02-05
.ci/tools: remove gopatch

jar-b:
### Description

This was previously installed to assist in automating migrations from AWS SDK V1. Now that that effort has completed, we no longer need to bundle this during tool installation.

### Relations

Relates #32976


ewbankkit (approved):
LGTM :rocket:.

PR #41267, opened 2025-02-06
docs: add storage configuration block for eks_cluster resource

### LINKED ISSUE #41226: [Docs]: aws_eks_cluster is missing details for storage_config
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#storage_config-1

### Description

> [storage_config](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#storage_config-1) - (Optional) Configuration block with storage configuration for EKS Auto Mode. Detailed below.

Unless I'm missing it, It is not detailed below unlike all the other options.

### References

_No response_

No

stefanfreitag:
### Description

* The resource `aws_eks_cluster`  contains a `storage_config` block, see e.g. example code available [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#eks-cluster-with-eks-auto-mode).
* The provider documentation is missing details on the `storage_config`. 

The pull requests 
- adds the missing documentation for `storage_config`
- adds links from "config block" argument to the corresponding description.

### Relations

Closes #41226 

### References

- [AWS Provider Source code](https://github.com/hashicorp/terraform-provider-aws/blob/16b3f1f59e22a57aedfe0fbe37926031608910b2/internal/service/eks/cluster.go#L394)

### DIFF
--- a/website/docs/r/eks_cluster.html.markdown
+++ b/website/docs/r/eks_cluster.html.markdown
@@ -407,11 +407,13 @@ The `remote_pod_networks` configuration block supports the following arguments:
 
 ### vpc_config Arguments
 
+* `cluster_security_group_id` - (Computed) Cluster security group that is created by Amazon EKS for the cluster. Managed node groups use this security group for control-plane-to-data-plane communication.
 * `endpoint_private_access` - (Optional) Whether the Amazon EKS private API server endpoint is enabled. Default is `false`.
 * `endpoint_public_access` - (Optional) Whether the Amazon EKS public API server endpoint is enabled. Default is `true`.
 * `public_access_cidrs` - (Optional) List of CIDR blocks. Indicates which CIDR blocks can access the Amazon EKS public API server endpoint when enabled. EKS defaults this to a list with `0.0.0.0/0`. Terraform will only perform drift detection of its value when present in a configuration.
 * `security_group_ids` – (Optional) List of security group IDs for the cross-account elastic network interfaces that Amazon EKS creates to use to allow communication between your worker nodes and the Kubernetes control plane.
 * `subnet_ids` – (Required) List of subnet IDs. Must be in at least two different availability zones. Amazon EKS creates cross-account elastic network interfaces in these subnets to allow communication between your worker nodes and the Kubernetes control plane.
+* `vpc_id` - (Computed) ID of the VPC associated with your cluster.
 
 ### kubernetes_network_config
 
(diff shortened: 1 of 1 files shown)


website/docs/r/eks_cluster.html.markdown
```diff
@@ -477,7 +489,7 @@ This resource exports the following attributes in addition to the arguments abov
 * `endpoint` - Endpoint for your Kubernetes API server.
 * `id` - Name of the cluster.
 * `identity` - Attribute block containing identity provider information for your cluster. Only available on Kubernetes version 1.13 and 1.14 clusters created or upgraded on or after September 3, 2019. Detailed below.
-* `kubernetes_network_config` - Attribute block containing Kubernetes network configuration for the cluster. Detailed below.
+* `kubernetes_network_config` - Attribute block containing Kubernetes network configuration for the cluster. [Detailed](#kubernetes_network_config-1) below.
```
justinretzolk, line 492: Since this is an argument, I think we should remove it from the attributes section entirely to be more consistent with other areas of the docs.

justinretzolk (approved):
Had a minor suggestion, but otherwise looks good to me! Thank you for taking the time to raise this @stefanfreitag! :tada:

website/docs/r/eks_cluster.html.markdown
stefanfreitag: Hi @justinretzolk,

we can do so. One question I have before going that route:

As part of  `kubernetes_network_config` there is  `service_ipv6_cidr` which is `Computed: true`. 
Is this maybe the reason why `kubernetes_network_config` block is listed under the Attributes section? 

(rephrased: How would we need to document `service_ipv6_cidr` then without mentioning the block under Attributes? - sorry for this possibly stupid question)

justinretzolk: This is an interesting one, and I think a great question @stefanfreitag!

From my perspective, the right approach would be to combine the two `kubernetes_network_config` block definitions (the ones starting at lines 416 and 506 respectively) and then note `service_ipv6_cidr` as `(Read-Only)` or `(Computed)` similar to the way we do for `(Optional)`. I'm not aware of another place in the documentation that we do this, so I'll spend some time looking around to see if we have a precedent here.

justinretzolk: In looking, it seems the `(Computed)` note is an established pattern:

- [`aws_apigatewayv2_domain_name.domain_name_configuration.hosted_zone_id`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apigatewayv2_domain_name#hosted_zone_id-1)
- [`aws_lex_bot_alias.log_settings.resource_prefix`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lex_bot_alias#resource_prefix-1)

Another time where I wish we were generating all of the docs, but alas 🙂

stefanfreitag: I introduced `(Computed)` and applied our approach on `kubernetes_network_config` also on `vpc_config` to be consistent in the document.

justinretzolk (approved):
Looks good to me! Thanks again for taking care of this! I'll get it merged once the checks go green 🚀

PR #41271, opened 2025-02-06
testing: Reduce uses of `resource.TestCheckResourceAttrSet` with ARN attributes: `n` to `z` services

gdavison:
### Description

Addresses `n` through `z` services. Excludes `organizations` and `redshift`

### DIFF
--- a/internal/service/rekognition/stream_processor.go
+++ b/internal/service/rekognition/stream_processor.go
@@ -73,7 +73,7 @@ var (
 )
 
 // @FrameworkResource("aws_rekognition_stream_processor", name="Stream Processor")
-// @Tags(identifierAttribute="stream_processor_arn")
+// @Tags(identifierAttribute="arn")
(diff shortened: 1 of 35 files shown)


internal/service/rekognition/stream_processor.go
```diff
@@ -134,6 +135,7 @@ func (r *resourceStreamProcessor) Schema(ctx context.Context, req resource.Schem
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
 				},
+				DeprecationMessage: "Use 'arn' instead. This attribute will be removed in a future verion of the provider.",
```
ewbankkit, line 138:
```suggestion
				DeprecationMessage: "Use 'arn' instead. This attribute will be removed in a future version of the provider.",
```

ewbankkit: Should this have a CHANGELOG entry?

jar-b (approved):
LGTM 🚀

PR #41276, opened 2025-02-07
feat: Workspaces Directory Dedicated Tenancy Support

brandonldixon:
### Description
Workspaces tenancy was hard coded to always be shared. Leveraging the existing tenancy enum type, I updated the schema to accept tenancy options of "SHARED" or "DEDICATED".

I left the default behavior to default to SHARED. Updated so the state file will reflect this change. 

The tests I wrote for this came back with "no applicable tests to run"

We did live test this against deployed / newly deployed workspace directories and the flags worked as expected. 

This is my first ever PR so any feedback is appreciated!

### DIFF
--- a/.changelog/40972.txt
+++ b/.changelog/40972.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_medialive_multiplex_program: Add configurable `create` timeout
+```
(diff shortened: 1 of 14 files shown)


and-ku:
Yes please!

Issue #41191, opened 2025-02-02
[Bug]: AWS Backup  - aws_backup_selection is not implementing correctly Equal in Tag filtering


sw360cab:
### Terraform Core Version

Terraform v1.10.1

### AWS Provider Version

terraform-provider-aws_v5.84.0

### Affected Resource(s)

* aws_backup_selection

### Expected Behavior

Creating a filter on tags for backup selection, is not producing the same result of doing it manually on the console

Given the following snippet

```yaml
resource "aws_backup_selection" "ebs_tagged_resources" {
  name         = "net_backup_selection"
  iam_role_arn = data.aws_iam_role.eks_backup.arn
  plan_id      = aws_backup_plan.testnet-backup.id
  resources    = ["arn:aws:ec2:*:*:volume/*"]

  selection_tag {
    type  = "StringEquals"
    key   = "aws:ResourceTag/Name"
    value = "reserved-storage"
  }

}
```

the following output is expected to be seen on AWS console

```
Summary
Resource assignment name
  net_backup_selection
IAM role
  _REDACTED_
Assign by ARN prefix
  arn:aws:ec2:*:*:volume/*
Tags - equal
  aws:ResourceTag/Name - validator-storage
```

however the result is the following

```
Summary
Resource assignment name
  net_backup_selection
IAM role
  _REDACTED_
Assign by ARN prefix
  arn:aws:ec2:*:*:volume/*
Tags
  aws:ResourceTag/Name - validator-storage
```

Note that missing ***equal*** in the Tags header make a great difference

### Actual Behavior

See above

### Relevant Error/Panic Output Snippet

```shell

```

### Terraform Configuration Files

A simple AWS backup plan

```
resource "aws_backup_plan" "net-backup" {
  name = "nets_backup_plan"

  rule {
    rule_name         = "net-backup-rule"
    target_vault_name = ...
    schedule          = "cron(0 4 * * ? *)"

    start_window = 60
    completion_window = 120
    lifecycle {
      cold_storage_after = 10
      delete_after = 100
    }
  }
}


resource "aws_backup_selection" "ebs_tagged_resources" {
  name         = "nets_backup_selection"
  iam_role_arn = ...
  plan_id      = aws_backup_plan.net-backup.id
  resources    = ["arn:aws:ec2:*:*:volume/*"]

  selection_tag {
    type  = "StringEquals"
    key   = "aws:ResourceTag/Name"
    value = "validator-storage"
  }

[...]

PR #41293, opened 2025-02-07
docs: fix invalid value for type in resource aws_backup_selection

stefanfreitag:
### Description
The [documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/backup_selection) for the resource `aws_backup_selection` mentions

> [type](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/backup_selection#type-1) - (Required) An operation, such as `StringEquals`, that is applied to a key-value pair used to filter resources in a selection.

Actually `StringEquals` is not correct. It should read `STRINGEQUALS`.

### Relations

- Relates #41191 
  The wrong value was spotted while working on this issue.

### References

- [AWS Provider source code](https://github.com/hashicorp/terraform-provider-aws/blob/5772b4568fdadb632f8769c921864a041f72f166/internal/service/backup/selection.go#L166)  
 Here one can see for the argument `type`:
  ```go
  ValidateDiagFunc: enum.Validate[awstypes.ConditionType](),
  ```
     In the referenced AWS SDK Go v2 there is
  
  ```go
  type ConditionType string

  // Enum values for ConditionType
  const (
  	ConditionTypeStringequals ConditionType = "STRINGEQUALS"
  )
  ```

### DIFF
--- a/website/docs/r/backup_selection.html.markdown
+++ b/website/docs/r/backup_selection.html.markdown
@@ -140,7 +140,7 @@ This resource supports the following arguments:
 
 Tag conditions (`selection_tag`) support the following:
 
-* `type` - (Required) An operation, such as `StringEquals`, that is applied to a key-value pair used to filter resources in a selection.
+* `type` - (Required) An operation, such as `STRINGEQUALS`, that is applied to a key-value pair used to filter resources in a selection.
 * `key` - (Required) The key in a key-value pair.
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🚀

jar-b:
Thanks for the fix, @stefanfreitag! 🚀

PR #41336, opened 2025-02-11
chore: remove `gopatch` from `make tools` target

jar-b:
### Description

This dependency has already been removed from the `tools` subdirectory, but the installation step was missed in the previous PR.

### Relations

Relates #41242


johnsonaj (approved):
LGTM 🚀

PR #41444, opened 2025-02-18
[Docs]: `target_ip.ip` is optional in Route 53 Resolver rule

Michagogo:
### Description
The documentation currently claims that the IPv4 address parameter is mandatory. This doesn’t make sense, as the API Reference specifies that you can specify IPv4 or IPv6 addresses but not both in the same rule, and checking the actual code, sure enough, it’s optional.

### Relations
N/A

### References
https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/route53resolver/rule.go#L95

N/A — minor documentation fix only


breathingdust (approved):
LGTM 🚀 Thanks for the contribution @Michagogo!

PR #41535, opened 2025-02-24
Docs - Update the example for aws_eks_access_entry resource

### LINKED ISSUE #40951: [Docs]:
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_access_policy_association

### Description

When creating an EKS access policy association, a corresponding EKS access entry with the same `principal_arn` must exist already. Since none of the `aws_eks_access_policy_association` arguments access derivative attributes of an `aws_eks_access_entry` it appears that terraform is unaware of this dependency. 

Even worse if you are only creating a couple EKS access entries and corresponding policy associations the terraform will not fail as it appears the `aws_eks_access_policy_association` will keep retrying until it can be created. As soon as you have a large number of paired entries and associations you will see failed terraform runs as terraform seems to apply batches of `aws_eks_access_policy_association` resources before even trying to create the corresponding `aws_eks_access_entry` resources. The only reliable fix is to add a `depends_on` block to the `aws_eks_access_policy_association` to make the dependency explicit.

If this is how these resources are intended to operate then the docs for `aws_eks_access_policy_association` should definitely give a complete example using a `depends_on` block to reference the associated `aws_eks_access_entry` resource as this is the only reliable way to create both resources at scale.

### References

_No response_

None

obounaim:
### Description
Update `aws_eks_access_policy_association` documentation to illustrate the dependency with `aws_eks_access_entry` resource.

### Relations
Closes #40951

### DIFF
--- a/website/docs/r/eks_access_policy_association.html.markdown
+++ b/website/docs/r/eks_access_policy_association.html.markdown
@@ -14,9 +14,9 @@ Access Entry Policy Association for an EKS Cluster.
 
 ```terraform
 resource "aws_eks_access_entry" "example" {
-  cluster_name      = aws_eks_cluster.example.name
-  principal_arn     = aws_iam_role.example.arn
-  type              = "STANDARD"
+  cluster_name  = aws_eks_cluster.example.name
+  principal_arn = aws_iam_role.example.arn
+  type          = "STANDARD"
 }
 
 resource "aws_eks_access_policy_association" "example" {
@@ -29,7 +29,7 @@ resource "aws_eks_access_policy_association" "example" {
     namespaces = ["example-namespace"]
   }
 
-  depends_on = [ aws_eks_access_entry.example ]
+  depends_on = [aws_eks_access_entry.example]
 }
 ```
 


justinretzolk (commented):
If I understand the linked issue, the `principal_arn` for the `aws_eks_access_entry` and `aws_eks_access_policy_association` need to be the same, but it appears that your update references an `aws_iam_role` vs. the `aws_iam_user` that the `aws_eks_access_policy_association` uses.

Additionally, by interpolating the value from `aws_eks_access_entry` within the `aws_eks_access_policy_association`, the need for an explicit `depends_on` can be eliminated since there would be an implicit dependency. That said, I'm not sure if that makes it less clear, so I'm open to not going that route. For example:

```hcl
resource "aws_eks_access_entry" "example" {
   cluster_name  = aws_eks_cluster.example.name
   principal_arn = aws_iam_role.example.arn
   type          = "STANDARD"
 }

 resource "aws_eks_access_policy_association" "example" {
   cluster_name  = aws_eks_cluster.example.name
   policy_arn    = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy"
   principal_arn = aws_eks_access_entry.example.principal_arn
   access_scope {
     type       = "namespace"
     namespaces = ["example-namespace"]
   }
 }
```

obounaim:
@justinretzolk, yes that would be a complete solution for the issue.
The idea behind my PR is to quickly update the documentation to make it clear for people.

obounaim:
@justinretzolk , sure.

First, I completely agree that the solution you suggest is the correct way to implement the fix. My changes are just to update the documentation to show a workaround while someone takes care of implementing the correct fix.

But I fine, if you think we should directly implement the proper solution. However, I don't I will have time for it.

website/docs/r/eks_access_policy_association.html.markdown
```diff
@@ -13,6 +13,12 @@ Access Entry Policy Association for an EKS Cluster.
 ## Example Usage
 
 ```terraform
+resource "aws_eks_access_entry" "example" {
+  cluster_name  = aws_eks_cluster.example.name
+  principal_arn = aws_iam_role.example.arn
```
justinretzolk, line 18: The `aws_eks_access_entry.principal_arn` and `aws_eks_access_policy_association.principal_arn` need to be the same, but in your example here, one references an `aws_iam_role` while the other references a `aws_iam_user`, leading to a mismatch.

website/docs/r/eks_access_policy_association.html.markdown
```diff
@@ -22,6 +28,8 @@ resource "aws_eks_access_policy_association" "example" {
     type       = "namespace"
     namespaces = ["example-namespace"]
   }
+
+  depends_on = [aws_eks_access_entry.example]
```
justinretzolk, line 32: This is the part where I'm unsure of the right way to represent in our documentation. By adding this `depends_on`, you're creating an explicit dependency between the two resources. That will work, and if that conveys the message most clearly, I'm okay with us leaving it as is.

On the other hand, an implicit dependency can be created by interpolating a value from `aws_eks_access_entry` within the `aws_eks_access_policy_association` resource. That would have the same effect as adding the `depends_on`, but could be viewed as cleaner or less clear depending on the experience level of whoever is reading the documentation. The updated configuration would then look like what I mentioned in my previous comment:

```hcl
resource "aws_eks_access_entry" "example" {
   cluster_name  = aws_eks_cluster.example.name
   principal_arn = aws_iam_role.example.arn
   type          = "STANDARD"
 }

 resource "aws_eks_access_policy_association" "example" {
   cluster_name  = aws_eks_cluster.example.name
   policy_arn    = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy"
   principal_arn = aws_eks_access_entry.example.principal_arn
   access_scope {
     type       = "namespace"
     namespaces = ["example-namespace"]
   }
 }
```

With the above configuration, the resources are created in the correct order.

justinretzolk (commented):
I think I see our disconnect now, @obounaim. I added a couple of comments within this review that I hope will help clear things up. My proposal is also only a documentation change.

website/docs/r/eks_access_policy_association.html.markdown
thefire: Actually @justinretzolk I don't believe your example of an implicit dependency will solve the issue that I found. I tried this approach and in my testing the `aws_eks_access_policy_association` only appeared to depend on the creation of the "principal" (in my case a role) and not the creation of the `aws_eks_access_entry`. 

Not knowing exactly how the internals of the dependency graph works I assumed this is because the `aws_eks_access_entry.principal_arn` is actually **known** before the creation of the access entry (since it's an argument and not an attribute or result) and therefore does not cause the `aws_eks_access_policy_association` to truly depend on the `aws_eks_access_entry`. 

In my testing with a lot of `aws_eks_access_entry` and associated `aws_eks_access_policy_association` resources the only fix was to make an explicit link via `depends_on`.

PR #41539, opened 2025-02-24
autoflex: Rename Calculate to Diff

### LINKED ISSUE #41512: [TechDebt]: Use `flex.Compare` in the `skaff` `Update` method
### Description

The `internal/framework/flex` package contains a `Compare` function for calculating whether state and plan have differences. The `skaff` template for Terraform Plugin Framework based resources should utilize this method for comparing state and plan in the update method rather than the current logic which explicitly compares each argument individually.

https://github.com/hashicorp/terraform-provider-aws/blob/23f4b59fe367ddf64b4f145702e518af62a681f3/skaff/resource/resourcefw.gtpl#L431-L434

https://github.com/hashicorp/terraform-provider-aws/blob/23f4b59fe367ddf64b4f145702e518af62a681f3/internal/framework/flex/diff.go#L40-L93

### References

_No response_

None

YakDriver:
### Description

In Go, function names are typically short, descriptive, and clear in their purpose without unnecessary redundancy. Given the function's role in computing differences between read AWS data and Terraform state, an idiomatic name could be:

#### Best options:

1. **`Diff`** – Simple, clear, and follows Go conventions (like `cmp.Diff`).  
   ```go
   diff, d := flex.Diff(ctx, plan, state)
   ```
   This is concise and immediately communicates the function's purpose.

2. **`ComputeDiff`** – Explicitly states that it calculates a difference.  
   ```go
   diff, d := flex.ComputeDiff(ctx, plan, state)
   ```
   This follows Go’s `ComputeX` pattern, which is seen in some standard library and popular Go packages.

#### Other reasonable options:

3. **`DetermineDiff`** – If the emphasis is on figuring out *what* changed rather than computing values.  
4. **`PlanDiff`** – More domain-specific.  
5. **`StateDiff`** – Emphasizes that the diff is between AWS and Terraform state.  
6. **`Delta`** – A more abstract name, though `Diff` is likely clearer.  

`flex.Diff` is unambiguous, the cleanest, and most Go-idiomatic choice. If we need more clarity, `ComputeDiff` is a strong alternative.

### Relations

Closes #41512

```console
% make build
make: Building provider...
```

### DIFF
--- a/internal/framework/flex/diff.go
+++ b/internal/framework/flex/diff.go
@@ -37,8 +37,8 @@ func (r *Results) IgnoredFieldNames() []string {
 	return r.ignoredFieldNames
 }
 
-// Calculate compares the plan and state values and returns whether there are changes
-func Calculate(ctx context.Context, plan, state any, options ...ChangeOption) (*Results, diag.Diagnostics) {
+// CalculateDiff compares the plan and state values and returns whether there are changes
+func CalculateDiff(ctx context.Context, plan, state any, options ...ChangeOption) (*Results, diag.Diagnostics) {
 	var diags diag.Diagnostics
 	opts := NewChangeOptions(options...)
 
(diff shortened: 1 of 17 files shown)


johnsonaj (approved):
LGTM 🚀

PR #41562, opened 2025-02-25
validations: remove `PreferWriteOnlyAttribute` validations

johnsonaj:
### Description

`PreferWriteOnlyAttribute` validation will be displayed as a `warning` when the following conditions are met:
- Using Terraform 1.11.0
- Not using a `write-only` attribute on a resource when one is available

This warning can be polarizing since there is nothing wrong with the configuration and the validation is simply a suggestion.

This PR also adds additional `RequiredWith` validation to `_wo` and `_wo_version` attributes

image

```console

...
```

### DIFF
--- a/website/docs/r/db_instance.html.markdown
+++ b/website/docs/r/db_instance.html.markdown
@@ -27,7 +27,7 @@ See the AWS Docs on [RDS Instance Maintenance][instance-maintenance] for more in
 ~> **Note:** All arguments including the username and password will be stored in the raw state as plain-text.
 [Read more about sensitive data instate](https://www.terraform.io/docs/state/sensitive-data.html).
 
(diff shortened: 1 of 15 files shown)


website/docs/r/db_instance.html.markdown
```diff
@@ -27,6 +27,8 @@ See the AWS Docs on [RDS Instance Maintenance][instance-maintenance] for more in
 ~> **Note:** All arguments including the username and password will be stored in the raw state as plain-text.
 [Read more about sensitive data instate](https://www.terraform.io/docs/state/sensitive-data.html).
 
+-> **Note:** Write-Only attribute `password_wo` is available to use in place of `password`. Write-Only attributes are supported in HashiCorp Terraform 1.11.0 and later.
```
jar-b, line 30: Should we link to write only argument docs here? Maybe
https://developer.hashicorp.com/terraform/language/v1.11.x/resources/ephemeral#write-only-arguments

PR #41564, opened 2025-02-26
Tech Debt: Improve memory behaviour of type-specific Flattener and Expander functions

gdavison:
### Description

Currnently, type-specific Flattener and Expander functions call through into `flex.Flatten` or `flex.Expand`, which need to determine source and target types and perform debug logging. These are not needed when the source and target types are known. Logging and type reflection performs on approximately 100 allocations for a total of 6500 to 9000 bytes.

By directly calling Framework `types` functions, allocations are reduced to 0 to 2 allocations of minor size. Execution time is also reduced by approximately 99%.

See commit messages for specific improvements.


ewbankkit (approved):
LGTM :rocket:.

PR #41574, opened 2025-02-26
skaff: follow preferred `input` conventions in finder

jar-b:
### Description

Modifies the finder function in the standard Plugin Framework resource template to define the `input` as a value, rather than a pointer. This aligns with a broader refactoring occurring across the codebase, and will keep net-new resources following the appropriate conventions.


ewbankkit (approved):
LGTM :rocket:.

Issue #41234, opened 2025-02-05
[Bug]: aws_sqs_queue, timeout while waiting for state to become


YakDriver:
### Terraform Core Version

1.10.5

### AWS Provider Version

5.85.0

### Affected Resource(s)

* aws_sqs_queue

### Expected Behavior

Resource should wait the time it takes to create

### Actual Behavior

It times out during creation.

### Relevant Error/Panic Output Snippet

```shell
Error: waiting for SQS Queue (https://sqs.us-east-1.amazonaws.com/012345678901/Test-name.fifo) attributes create: timeout while waiting for state to become 'equal' (last state: 'notequal', timeout: 2m0s)
│ 
│   with aws_sqs_queue.queue,
│   on main.tf line 16, in resource "aws_sqs_queue" "queue":
│   16: resource "aws_sqs_queue" "queue" {
```

### Terraform Configuration Files

```terraform
resource "aws_sqs_queue" "queue" {
  # kms_master_key_id is not set so no KMS data key but reuse period is set
  kms_data_key_reuse_period_seconds = "60" # not default (300)
  sqs_managed_sse_enabled           = false # not enabled but reuse period is set
  max_message_size                  = "261244"
  message_retention_seconds         = "60"
  fifo_queue                        = true
  name                              = "Test-name.fifo"
  visibility_timeout_seconds        = "60"
  receive_wait_time_seconds         = "10"
}
```

### Steps to Reproduce

1. Use `us-east-1` where this is most often reported
2. Use an account that performs slowly (perhaps having many SQS queues or being an older account?)
3. `terraform apply`

### References

* #29497

PR #41590, opened 2025-02-27
docs/aws_sqs_queue: Timeout with incompatible arguments

YakDriver:
### Description

SQS has an unusual behavior where certain misconfigurations cause it to hang indefinitely rather than return an error. Specifically, creating or updating an `aws_sqs_queue` results in a `timeout while waiting` error if:  

- `kms_data_key_reuse_period_seconds` is set to a non-default value,  
- `sqs_managed_sse_enabled` is `false` (explicitly or by default), and  
- `kms_master_key_id` is not set.  

While `resourceQueueCustomizeDiff` already handles `kms_data_key_reuse_period_seconds` in some cases, we don't modify configurations to prevent this hanging behavior. Instead of enforcing a change, this PR updates the documentation to make users aware of the issue. We could also connect `kms_data_key_reuse_period_seconds`, `sqs_managed_sse_enabled`, and `kms_master_key_id` to provide invalid configuration errors.

### Relations

Relates #41234

N/A


jar-b (approved):
LGTM 🚀

Issue #35813, opened 2024-02-14
Disallow use of `SingleNestedBlock` in resource schemas


ewbankkit:
The use of `SingleNestedBlock` resource schemas for resources implemented using Terraform Plugin Framework is problematic.
NestedBlocks have no Required/Optional/Computed flags – they are effectively always there, so an “Optional” `SingleNestedBlock` doesn’t work as expected if the block isn’t present in configuration (e.g. Required attributes).
The pattern with Terraform Plugin SDK v2 is to use ListNestedBlock with SizeAtMost(1) and then a missing block is equivalent to an empty list. In fact `SingleNestedBlock`s are not available in Terraform Plugin SDK v2.

* https://developer.hashicorp.com/terraform/language/syntax/configuration#arguments-and-blocks
* https://developer.hashicorp.com/terraform/language/attr-as-blocks
* https://stackoverflow.com/a/71443376
* https://developer.hashicorp.com/terraform/plugin/framework/handling-data/blocks

Relates https://github.com/hashicorp/terraform-provider-aws/issues/34664.

There are only 3 uses in the current codebase:

```console
% grep SingleNestedBlock internal/service/*/*.go
internal/service/bedrock/model_invocation_logging_configuration.go:			"logging_config": schema.SingleNestedBlock{
internal/service/bedrock/model_invocation_logging_configuration.go:					"cloudwatch_config": schema.SingleNestedBlock{
internal/service/bedrock/model_invocation_logging_configuration.go:							"large_data_delivery_s3_config": schema.SingleNestedBlock{
internal/service/bedrock/model_invocation_logging_configuration.go:					"s3_config": schema.SingleNestedBlock{
internal/service/opensearchserverless/security_config.go:			"saml_options": schema.SingleNestedBlock{
internal/service/opensearchserverless/security_config_data_source.go:			"saml_options": schema.SingleNestedBlock{
internal/service/verifiedpermissions/schema.go:			"definition": schema.SingleNestedBlock{
```

Add a semgrep rule to prevent additional occurrences and migrate the existing occurrences to `ListNestedBlock`s.

acwwat:
Updated list of references as of 2025-02-12:

```text
$ grep SingleNestedBlock internal/service/*/*.go
internal/service/elbv2/listener_rule_data_source.go:                                            "authenticate_cognito": schema.SingleNestedBlock{
[...]

PR #41610, opened 2025-02-28
.ci/semgrep: add `avoid-SingleNestedBlock` rule

jar-b:
### Description

This rule will check for any usage of `schema.SingleNestedBlock` within the `internal/service` subfolder. Pre-existing instances of `SingleNestedBlock`s have been annotated with `nosemgrep` trailing comments. These will be migrated in the future.

### Relations

Relates #35813


ewbankkit (approved):
LGTM :rocket:.

PR #41612, opened 2025-02-28
docs: group generated tag test options with headings

jar-b:
### Description

As tag test generation has expanded across the provider, the number of parameters which can be passed to the tags test annotation has grown significantly. Given the number of options, the contributor documentation on this section has become difficult to read in its entirety. This change bundles the various parameters into related groups and adds headings to break up the available options into smaller standalone sections.

### References
- https://hashicorp.github.io/terraform-provider-aws/resource-tagging/#controlling-test-generation


ewbankkit (approved):
LGTM :rocket:.

PR #41637, opened 2025-03-03
[TechDebt]: Standardize deprecation messages

### LINKED ISSUE #40050: [TechDebt]: Include attribute names in deprecation messages
### Description

The default warning diagnostic summaries for both Plugin SDKV2 and Plugin Framework based resources only include a generic title (`Argument is deprecated` and `Attribute Deprecated`, respectively). This means that omitting the attribute name in the `Message` field can result in a diagnostic lacking context about _what_ is actually deprecated. 

To identify the attribute in question the user needs to follow the source file and line number printed in the result. Here is a recent example (fixed via #39430).

```
│ Warning: Argument is deprecated
│
│   with module.main.aws_iam_role.backup_default_service_role,
│   on ../main_module/main.tf line 8290, in resource "aws_iam_role" "backup_default_service_role":
│ 8290: resource "aws_iam_role" "backup_default_service_role" {
│
│ Use the aws_iam_role_policy resource instead. If Terraform should exclusively manage all inline policy associations (the current
│ behavior of this argument), use the aws_iam_role_policies_exclusive resource as well.
```

The `inline_policies` attribute is the source of the issue, but nothing in the diagnostic states this. 

### Proposal

Many of our existing deprecation messages follow a patterns like:

- `Use the 'aws_service_resource' instead`
- `Use 'other_argument' instead`

To provide additional clarity to users, we should always include the argument name in the message body.

### References

- https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema.Deprecated
- https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource/schema#StringAttribute.DeprecationMessage
- Example of a deprecation message lacking context: https://github.com/hashicorp/terraform-provider-aws/issues/22336#issuecomment-2363944697

None

jar-b:
### Description

Standardizes the format of all deprecation messages such that they include the affected argument name. This will prevent confusion for practitioners as the default message does not always include the attribute path (Terraform diagnostic messages [do not support set paths](https://github.com/hashicorp/terraform-plugin-sdk/issues/1094#issuecomment-1307536150)).

Also adds a `semgrep` rule to enforce inclusion of the argument name in new deprecation messages.

### Relations

Closes #40050

### DIFF
--- a/.ci/.semgrep.yml
+++ b/.ci/.semgrep.yml
@@ -680,3 +680,21 @@ rules:
     patterns:
       - pattern: schema.SingleNestedBlock{ ... }
     severity: ERROR
+
+  - id: include-argument-name-in-deprecation-message
+    languages: [go]
+    message: Deprecation messages should begin with `argument_name is deprecated`.
+    paths:
+      include:
+        - internal/service
+    patterns:
+      - pattern-inside: "Schema: map[string]*schema.Schema{ ... }"
+      - pattern: |
+          {
+            ..., Deprecated: $MESSAGE, ...
+          }
+      - metavariable-pattern:
+          metavariable: $MESSAGE
+          patterns:
+            - pattern-not-regex: ([a-zA-Z0-9_]+ is deprecated\.)
+    severity: WARNING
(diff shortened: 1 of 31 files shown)


ewbankkit (approved):
LGTM :rocket:.

Issue #41583, opened 2025-02-27
[TechDebt]: Update skaff to use accepted/linter conventions


nam054:
### Description

There has been design/technical changes required by new/existing resources and datasources.

It is now the expected standard to utilize `ValueStringPointer`over the traditional `aws.String(x.ValueString())` approach. This shift ensures consistency and alignment with current best practices. As a result, the generated resources and data sources through skaff should be updated to reflect this change, as adherence to this convention is also enforced by the linters. 

Similarly, it is now required that commonly reused attribute names use the `names.` package. A common example is the `arn` attribute is now required to be `names.AttrARN`. 

### Would you like to implement a fix?

None

PR #41652, opened 2025-03-04
skaff: use `names` package attribute constants

jar-b:
### Description

Currently the hardcoded attribute names will be flagged by the attribute constants semgrep rule. This change brings the preferred pattern into `skaff`.

### Relations

Relates #41583


ewbankkit (approved):
LGTM :rocket:.

Issue #40992, opened 2025-01-17
[Enhancement]: Remove properties related to Elastic Inference as it has been deprecated


gmarciani:
### Description

Hi,

Elastic Inference has been deprecated on April, 2024.

The following items related to Elastic Inference must be removed together with the respective documentation:
1. properties in EC2 and ECS: `inference_accelerator_overrides` `elastic_inference_accelerator` `elastic_inference_accelerators` `LaunchTemplateElasticInferenceAccelerator`
2. The Elastic Inference service

### Potential Terraform Configuration

```terraform

```

### References

See deprecation notice from [official AWS documentation](https://docs.aws.amazon.com/cli/latest/reference/elastic-inference/#cli-aws-elastic-inference):

<img width="731" alt="Image" src="https://github.com/user-attachments/assets/ade319ba-ede6-49e1-b653-e071903f6727" />

### Would you like to implement a fix?

None

YakDriver:
This equates to the following changes:

- [ ] From `aws_ecs_task_execution` data source, remove `inference_accelerator_overrides`
- [ ] From `aws_launch_template` data source, remove `elastic_inference_accelerator`
- [ ] From `aws_launch_template` resource, remove `elastic_inference_accelerator`

bryantbiggs:
this also includes `aws_ecs_task_definition` - https://github.com/hashicorp/terraform-provider-aws/blob/63aed68fd48035b85f04fba9cd13efe7af0151e2/internal/service/ecs/task_definition.go#L153C5-L153C26

YakDriver:
Closed by #42137

Issue #40991, opened 2025-01-17
[Service EOL]: Amazon Elastic Graphics reached end of life on January 8, 2024


gmarciani:
### Description

Hi,

Elastic Graphics has been deprecated on January, 2024.

The following items related to Elastic Graphics must be removed together with the respective documentation:
1. property `elastic_gpu_specifications` in EC2
2. references to the EC2 operation `DescribeElasticGpus`

### Affected Resource(s) and/or Data Source(s)

EC2 resources and data sources.

### References

See deprecation notice from [official AWS documentation](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-elastic-gpus.html):

<img width="732" alt="Image" src="https://github.com/user-attachments/assets/116fdc64-b7c6-4781-8fcc-15b88faf81e1" />

### Would you like to implement a fix?

None

YakDriver:
This equates to the following changes:

- [ ] From `aws_launch_template` data source, remove `elastic_gpu_specifications`
- [ ] From `aws_launch_template` resource, remove `elastic_gpu_specifications`

jar-b:
Closed via #42312. These changes will be available in the `v6.0.0` beta release later this month.

PR #41677, opened 2025-03-05
Deprecate Elastic Graphics/Inference in `aws_launch_template`

YakDriver:
### Description

Deprecate arguments related to Elastic Graphics and Elastic Inference, which AWS no longer supports, so that they can be removed in the next major version.

### Relations

Relates:
* #40991 
* #40992

```console

...
```


johnsonaj (approved):
LGTM 🚀

PR #41678, opened 2025-03-05
B account region enable

### LINKED ISSUE #41430: [Bug]: `aws_account_region` errors when attempting to enable a region that it already enabled
### Terraform Core Version

1.10.0

### AWS Provider Version

5.84.0

### Affected Resource(s)

* aws_account_region

### Expected Behavior

When the Terraform configuration wishes to set a region to be enabled, if that region was already enabled, then silently import the enablement and continue.

### Actual Behavior

Got an error: `Unable to switch region status because of its current opt-in status.`

```shell
Error: enabling Account Region (123456789012,eu-south-1): operation error Account: EnableRegion, https response error StatusCode: 400, RequestID: abcdef01-2345-6789-0abc-def012345678, ValidationException: Unable to switch region status because of its current opt-in status.
  with module.bootstrap.aws_account_region.enabled["eu-south-1"],
  on resources-regions.tf line 1, in resource "aws_account_region" "enabled":
   1: resource "aws_account_region" "enabled" {
```

N/A

### Steps to Reproduce

1. (optional) override the resource create timeout to `10s` to force a timeout error and show the problem
2. Terraform provider calls `EnableRegion`: https://github.com/hashicorp/terraform-provider-aws/blob/ddc7bceb2a7ac23982507fe4a6315d7f1d933127/internal/service/account/region.go#L102
3. AWS accepts the API call, and puts the region into the "Enabling..." status
4. Terraform waits for enablement to finish before proceeding: https://github.com/hashicorp/terraform-provider-aws/blob/ddc7bceb2a7ac23982507fe4a6315d7f1d933127/internal/service/account/region.go#L108
5. Something happens and the Terraform process stops before the region enters "Enabled" status
6. Attempt to re-run Terraform
7. `terraform plan` will attempt to create `aws_account_region` again, since it didn't think that the region was successfully "created" in the last run
8. Terraform provider calls `EnableRegion` on a region that has since entered into `Enabled` status
9. get the error. Error persists on subsequent retries without manual intervention.

### Debug Output

N/A

### Panic Output

N/A

### Important Factoids

_No response_

### References

https://docs.aws.amazon.com/accounts/latest/reference/API_EnableRegion.html#API_EnableRegion_RequestBody

The documentation warns:

> This process takes a few minutes for most accounts, but it can take several hours.

Particularly since the default timeout is 60 minutes, if for some reason the region is not enabled/created successfully on the first attempt, then the error will trigger.

None

nam054:
### Description

Currently, when attempting to enable a region that is already in the enabling  or  enabled status, an error occurs. This change will prevent this error by allowing the process to skip the EnableRegion API call. Instead, it would go directly to the waitRegionEnabled status if the region is already enabling or enabled. Essentially, this change modifies the behavior by letting the process continue without error/failing.

### Relations

Closes #41430

```console
> make testacc TESTARGS="-run=TestAccAccount_serial/Region" PKG=account
TF_ACC=1 go1.23.5 test ./internal/service/account/... -v -count 1 -parallel 20  -run=TestAccAccount_serial/Region -timeout 360m -vet=off
2025/03/06 10:23:09 Initializing Terraform AWS Provider...
=== RUN   TestAccAccount_serial
=== PAUSE TestAccAccount_serial
=== CONT  TestAccAccount_serial
=== RUN   TestAccAccount_serial/Region
=== RUN   TestAccAccount_serial/Region/basic
=== RUN   TestAccAccount_serial/Region/AccountID
    region_test.go:70: skipping test because at least one environment variable of [AWS_ALTERNATE_PROFILE AWS_ALTERNATE_ACCESS_KEY_ID] must be set. Usage: credentials for running acceptance testing in alternate AWS account.
--- PASS: TestAccAccount_serial (413.81s)
    --- PASS: TestAccAccount_serial/Region (413.81s)
        --- PASS: TestAccAccount_serial/Region/basic (413.81s)
        --- SKIP: TestAccAccount_serial/Region/AccountID (0.00s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/account    420.625s

...
```

### DIFF
--- a/.changelog/41678.txt
+++ b/.changelog/41678.txt
@@ -1,7 +1,7 @@
 ```release-note:enhancement
-resource/aws_account_region: Prevent errors when the region is an ENABLING or DISABLING state during creation.
+resource/aws_account_region: Prevent errors when the region is an ENABLING or DISABLING state during creation
 ```
 
 ```release-note:enhancement
-resource/aws_account_region: Allow adoption of regions in an ENABLED or DISABLED state without an explicit import operation.
+resource/aws_account_region: Allow adoption of regions in an ENABLED or DISABLED state without an explicit import operation
 ```
--- a/.changelog/41678.txt
+++ b/.changelog/41678.txt
@@ -1,3 +1,7 @@
-```release-note:note
-resource/aws_account_region: Change the way to handle region enable/disable when already enabled/disabled in state
+```release-note:enhancement
+resource/aws_account_region: Prevent errors when the region is an ENABLING or DISABLING state during creation.
+```
+
+```release-note:enhancement
+resource/aws_account_region: Allow adoption of regions in an ENABLED or DISABLED state without an explicit import operation.
 ```
--- a/.changelog/41678.txt
+++ b/.changelog/41678.txt
@@ -1,0 +1,3 @@
+```release-note:note
+resource/aws_account_region: Change the way to handle region enable/disable when already enabled/disabled in state
+```
--- a/internal/service/account/region.go
+++ b/internal/service/account/region.go
@@ -247,7 +247,7 @@ func waitRegionDisabled(ctx context.Context, conn *account.Client, accountID, re
 
 func requiresStatusChange(status types.RegionOptStatus, enable bool) bool {
 	if enable {
-		return status != types.RegionOptStatusEnabled
+		return status != types.RegionOptStatusEnabled && status != types.RegionOptStatusEnabling
 	}
 	return status != types.RegionOptStatusDisabled && status != types.RegionOptStatusDisabling
 }
--- a/internal/service/account/region.go
+++ b/internal/service/account/region.go
@@ -104,7 +104,7 @@ func resourceRegionUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 			input.AccountId = aws.String(accountID)
 		}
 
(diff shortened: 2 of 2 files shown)


internal/service/account/region.go
```diff
@@ -239,3 +244,10 @@ func waitRegionDisabled(ctx context.Context, conn *account.Client, accountID, re
 
 	return nil, err
 }
+
+func requiresStatusChange(status types.RegionOptStatus, enable bool) bool {
+	if enable {
+		return status != types.RegionOptStatusEnabled
```
jar-b, line 250: To ensure we also skip when the region is actively being enabled.

```suggestion
		return status != types.RegionOptStatusEnabled && status != types.RegionOptStatusEnabling
```

.changelog/41678.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:note
+resource/aws_account_region: Change the way to handle region enable/disable when already enabled/disabled in state
```
jar-b, line 2: We might want to split this into two distinct enhancement items.

1. Prevent errors when the region is an `ENABLING` or `DISABLING` state during creation.
2. Allow adoption of regions in an `ENABLED` or `DISABLED` state without an explicit import operation.

jar-b (approved):
LGTM 🚀

PR #41683, opened 2025-03-06
Fix interpolation of token output in triage workflow

justinretzolk:
### Description

Fixes interpolation because `string` != `${{ string }}`, and I am once again wishing for a better way to test workflows locally 😅


ewbankkit (approved):
LGTM :rocket:.

PR #41722, opened 2025-03-07
Block Terraform 0.12 from v6.0.0 beta

ewbankkit:
### Description

Terraform 0.12 [does not support pre-releases](https://github.com/hashicorp/terraform/issues/36586) in provider version constraints. Explicitly block TF 0.12 (and below) during provider configuration.

This change must **NOT** make its way into the v6.0.0 GA release.


jar-b (dismissed):
LGTM 🚀

internal/provider/provider.go
```diff
@@ -445,6 +445,14 @@ func configure(ctx context.Context, provider *schema.Provider, d *schema.Resourc
 		terraformVersion = "0.11+compatible"
 	}
 
+	// Terraform 0.12.31 will download beta version of the provider as the latest available.
+	// See https://github.com/hashicorp/terraform/issues/36586.
+	// This check must be removed before the GA release.
+	switch {
+	case terraformVersion == "0.11+compatible", strings.HasPrefix(terraformVersion, "0.12."):
+		return nil, sdkdiag.AppendErrorf(diags, "unsupported Terraform version: %s", terraformVersion)
```
jar-b, line 453: Overlooked this on the first pass, but since this is displayed in a diagnostic we could potentially get away with a longer message than what is typical for go errors. 

Perhaps we can include some additional context indicating how to pin versions and/or that `v0.12` will still be supported, just not until after GA?

internal/provider/provider.go
```diff
@@ -445,6 +445,16 @@ func configure(ctx context.Context, provider *schema.Provider, d *schema.Resourc
...
 	}
 
+	// Terraform 0.12.31 will download beta version of the provider as the latest available.
+	// See https://github.com/hashicorp/terraform/issues/36586.
+	// This check must be removed before the GA release.
+	switch {
+	case terraformVersion == "0.11+compatible", strings.HasPrefix(terraformVersion, "0.12."):
+		return nil, sdkdiag.AppendErrorf(diags, "unsupported Terraform version: %s. "+
+			"This version of Terraform is not supported with pre-release version of the Terraform AWS Provider but will be supported at GA. "+
+			"See https://developer.hashicorp.com/terraform/language/providers/requirements#v0-12-compatible-provider-requirements for details of how to specify an exact provider version to use", terraformVersion)
```
jar-b, line 455: This is great 👍

jar-b (approved):
LGTM 🚀

PR #41762, opened 2025-03-10
autoflex: Handle all `StringValuable` values consistently

gdavison:
### Description

The function `mapBlockKey` currently uses `reflect.Value.MethodByName` to detect values that can be converted to `StringValue`. Instead, check for any type that implements `StringValuable`

Note: Uses of `reflect.Value.MethodByName` weaken the dead code elimination in the Go linker


ewbankkit (approved):
LGTM :rocket:.

PR #41768, opened 2025-03-11
Backup restore testing plan requires name but example didn't have one

AresiusXP:


ewbankkit (dismissed):
LGTM :rocket:.

AresiusXP:
@ewbankkit terrafmt was complaining about the block so I fixed that. Could you review again? 🙏

PR #41791, opened 2025-03-12
skaff: add not found warning diagnostic to read

jar-b:
### Description

This is a more recent pattern resources have been adopting which emits a warning diagnostic when the read operation detects a `NotFound` error from the corresponding AWS API. Adding this diagnostic provides practitioners with additional context about why a resource may be re-created (it was deleted out of band), and in general is a good practice to embed into `skaff`.

### Relations

Relates #41583


ewbankkit (approved):
LGTM :rocket:.

PR #41824, opened 2025-03-13
init

AliAllomani:
### Relations

Closes #0000

```console

...
```

### DIFF
--- a/internal/service/invoicing/invoice_unit.go
+++ b/internal/service/invoicing/invoice_unit.go
@@ -1,0 +1,133 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package invoicing
+
+import (
+	"context"
+	"log"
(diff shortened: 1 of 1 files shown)


jar-b:
Hello - Thanks for your submission!

As of 2023-10-20 all net-new resources are required to use [Terraform Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework). More context is available in this pinned issue and the provider contributor guide.

- https://github.com/hashicorp/terraform-provider-aws/issues/32917
- https://hashicorp.github.io/terraform-provider-aws/terraform-plugin-development-packages/#which-plugin-version-should-i-use

Once prioritized, maintainers will convert Plugin-SDK based resources as part of the review process, but if you'd like to do the conversion in the meantime we would greatly appreciate it!

niccoloborrelli:
Hello, do you know if and when the resource will be integrated into the aws official provider?

PR #41826, opened 2025-03-13
skaff: add sweeper function to resource template

jar-b:
### Description

This will encourage new resources to implement the corresponding sweeper function. Comments are included with direction on how to register the generated function in the services `sweep.go` file.

Also includes documentation updates to the example sweeper functions:

- Provide both Plugin SDK and Plugin Framework based implementations 
- Remove much of the boilerplate setup and error checking which is now handled centrally by the `awsv2.Register` helper function


ewbankkit (approved):
LGTM :rocket:.

PR #41900, opened 2025-03-18
docs: link to schema union type issue

jar-b:
### Description

Adds a proper link to the Terraform core issue to replace the plain text URL which is currently rendered.

image


ewbankkit (approved):
LGTM :rocket:.

PR #41947, opened 2025-03-21
docs: fix table and namespace name of s3tables-related examples

k5ki:
### Description

Fix S3Tables-related examples.
On S3 Tables, table name and namespace name cannot include hyphens (`-`). So I replaced it with underscore (`_`).

### References

table naming rules can be found in:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-buckets-naming.html#naming-rules-table


ewbankkit (approved):
LGTM :rocket:.

PR #41967, opened 2025-03-24
Fix regressions notifier

justinretzolk:
### Description

This PR updates the regressions notifier to use a `rich_text` block rather than a `section` with a `mrkdwn` block for linking to the affected item. This helps fix weird quoting issues that let to some items being posted without working links. While I was at it, I made some slight adjustments to formatting.

Whilst writing this PR description, I've started to wonder if the text should be updated to say "The following **(item|Pull Request|Issue)** has been labeled...", (change bolded) so if that reads better to the team, I'm happy to make those changes prior to merge.

### References

- [A recent example of a failure](https://github.com/hashicorp/terraform-provider-aws/actions/runs/14000394430)
- [`rich_text` block reference](https://api.slack.com/reference/block-kit/blocks#rich_text)
- [`header` block reference](https://api.slack.com/reference/block-kit/blocks#header)
- [`divider` block reference](https://api.slack.com/reference/block-kit/blocks#divider)

The following screenshot was the result of previewing these changes in Slack's [block kit builder](https://app.slack.com/block-kit-builder)

Screenshot 2025-03-24 at 11 04 51

### DIFF
--- a/.github/workflows/maintainer_helpers.yml
+++ b/.github/workflows/maintainer_helpers.yml
@@ -173,7 +173,7 @@ jobs:
                   "type": "header",
                   "text": {
                     "type": "plain_text",
-                    "text": ":warning: The following has been labeled as a ${{ github.event.label.name }}:",
+                    "text": ":warning: The following ${{ github.event_name == 'issues' && 'issue' || 'pull request' }} has been labeled as a ${{ github.event.label.name }}:",
                     "emoji": true
                   }
                 },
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🎉 

I personally like the idea of including Issue or PR in the heading text to distinguish new regression reports from fixes, but I'm also fine to merge this fix as-is and follow up with that later.

justinretzolk:
@jar-b thanks for the feedback! I'd forgotten that there was already prior art in the triage workflow for having the distinguisher, so I went ahead and cribbed that so a follow up PR wouldn't be needed.

jar-b (approved):
LGTM 🚀

Issue #41999, opened 2025-03-27
[Enhancement]: Allow enabling and disabling of lambda source aws_lambda_event_source_mapping


alvarosaavedrau:
### Description

Is there any parameter to disable the source of a lambda? The `Enabled` parameter only affects when this resource is created, not whether it acts as a switch afterwards.

The equivalent I am looking for is the cli `aws command lambda update-event-source-mapping --uuid <mapping-uuid> --enabled false`

There is a possibility to turn off the lambda with the `reserved_concurrent_executions` parameter of the `aws_lambda_function` resource but that is not what I want for the lambda operation, I need to disable the source that triggers the lambda.

In serverless framework there is this possibility to disable the lambda triggerer:

![Image](https://github.com/user-attachments/assets/28a3111c-a9df-4281-b9d7-f8480b94c456)

### Affected Resource(s) and/or Data Source(s)

* aws_lambda_event_source_mapping

### Potential Terraform Configuration

```terraform

```

### Would you like to implement a fix?

None

alvarosaavedrau:
The “enabled” parameter is described as:
```
enabled - (Optional) Determines if the mapping will be enabled on creation. Defaults to true.
```

I have tested it and it is not simply done on resource creation, it also works for already created resources, this serves to enable and disable it, the description is not the most adequate.

PR #42001, opened 2025-03-27
d/aws_lambda_event_source_mapping improve 'enabled' parameter description

alvarosaavedrau:
### Description

Improve the `enabled` parameter on aws_lambda_event_source_mapping resource.

The `enabled` parameter is described as:
```
enabled - (Optional) Determines if the mapping will be enabled on creation. Defaults to true.
```

This parameter not only applies on the resource creation. It also works for already created resources.
I have updated the documentation to:
```
enabled - (Optional) Determines if the mapping is enabled. This parameter can be used to enable or disable the mapping, both during resource creation and for already created resources. Defaults to true.
```

### Relations

Closes [#41999](https://github.com/hashicorp/terraform-provider-aws/issues/41999)


ewbankkit:
@alvarosaavedrau Thanks for the contribution :tada: :clap:.

Issue #41160, opened 2025-01-30
[Proposal]: Consider alternatives for surfacing resource registration errors


jar-b:
### Description

Currently errors encountered during resource registration are logged at the `WARNING` level. This can result in a poor developer experience under certain conditions. 

https://github.com/hashicorp/terraform-provider-aws/blob/a2291d945ed7f13f3adb7f49ad5d46d292612f4b/internal/provider/fwprovider/provider.go#L466-L470

For example, applying the `@Tags` annotation when no `tags` attribute is defined in the schema will surface a warning level error similar to the following. 

```
2025-01-30T13:30:46.842-0500 [WARN]  aws: registering resources: tf_req_id=bb0096ca-7c0a-1be0-e702-4374921fce06 tf_provider_addr=registry.terraform.io/hashicorp/aws tf_rpc=GetProviderS
chema tf_mux_provider="*proto5server.Server" error="no `tags` attribute defined in schema: aws_billing_view"
```

The provider will still build, but the impacted resource will not be registered. If the developer does not have warning level logging enabled, this can prove difficult to debug as referencing the resource (e.g in acceptance tests) will simply result in an error message like `Error: Invalid resource type`. 

### Proposal

Unfortunately, the [`Resources` method](https://github.com/hashicorp/terraform-plugin-framework/blob/v1.13.0/provider/provider.go#L51-L56) does not include a mechanism to surface errors or diagnostics while registering resources. This means an alternative approach must be used to surface these errors in place of logged warnings which require explicit configuration of the `TF_LOG` environment variable.

#### Evaluate for errors in the `Configure` method

Unlike the `Resource` method, the provider's [`Configure` method](https://github.com/hashicorp/terraform-plugin-framework/blob/v1.13.0/provider/provider.go#L35-L42) does expose a response object with embedded diagnostics. If we repeated the walk of the service package data done during resource registration in this method, we could theoretically expose the errors here. This would involve traversing the service packages data  a second time (in addition to the `Resources`, `DataSources`, and `EphemeralResources` methods), but this trade off may be worthwhile if we can extract the logic into a shared function and better surface registration issues.

#### Panic

A alternative option with the existing registration workflow is to `panic` rather than log warnings.

```go
	if err := errors.Join(errs...); err != nil {
		panic(fmt.Errorf("registering resources: %w", err))
	}

```

[...]

Issue #41101, opened 2025-01-27
[Major Version]: Terraform AWS Provider v6.0.0


breathingdust:
## Description

The next major version of the Terraform AWS Provider is planned to release in beta in April 2025. There will be a 6 week beta period in which we will ask the community for feedback, to better understand the impact of the upgrade experience, to assess the implementation of multi-region and other enhancements and to resolve any issues found. Please submit any feedback by creating an issue in the provider using the [beta-feedback](https://github.com/hashicorp/terraform-provider-aws/issues/new?template=05_beta_feedback.yml) form.

After the release of the beta, we will endeavor to maintain and keep both v5 and v6 branches up to date with new service/resource coverage and bugfixes during the beta period. After v6.0.0 exits beta with a stable release, v5 will only receive critical security updates.

## Opting into the v6.0.0 beta

The builds will be versioned as "beta" pre-releases and uploaded to the Terraform Registry as usual. Users with permissive version constraints, such as `> 5.8,` will not automatically use the new beta version. Instead, they will need to opt-in by specifying the pre-release version in their provider requirements:

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "= 6.0.0-beta"
    }
  }
}
```

### Terraform 0.12

Terraform 0.12 does not [support](https://github.com/hashicorp/terraform/issues/36586) pre-releases in provider version constraints. For this reason, 0.12 users will receive a runtime error when attempting to use the v6.0.0 beta either by specifying the version or via a non-specific version constraint. This will be removed after the stable release so that 0.12 users can use v6.0.0. We advise that users of v0.12 should upgrade to a more recent version of Terraform at the earliest opportunity.

- https://github.com/hashicorp/terraform-provider-aws/issues/27758#issuecomment-2587102363

### Upgrade Instructions

As always comprehensive upgrade instructions will be found in the Registry. We would welcome in particular contributions to documentation if these instructions are unclear in any way to ensure that the upgrade process is as easy as possible for practitioners.

_Note: The upgrade guide is under development and will be added to as we work though the scope of the major version._

- [Terraform AWS Provider v6.0.0 Upgrade Guide (WIP)](https://github.com/hashicorp/terraform-provider-aws/blob/release/6.0.0-beta/website/docs/guides/version-6-upgrade.html.markdown)

### Enhanced Multi-Region Support

- #25308
- #27758
- #31517

[...]

PR #42014, opened 2025-03-27
Per-resource Region override phase I - SDK resources

ewbankkit:
### Description

Phase I of the per-resource Region override work for **v6.0.0**.

### Relations

Closes https://github.com/hashicorp/terraform-provider-aws/issues/41160.
Relates https://github.com/hashicorp/terraform-provider-aws/issues/27758.

```console

...
```


internal/types/service_package.go
```diff
@@ -24,6 +31,7 @@ type ServicePackageEphemeralResource struct {
 	Factory  func(context.Context) (ephemeral.EphemeralResourceWithConfigure, error)
 	TypeName string
 	Name     string
+	Region   *ServicePackageResourceRegion
```
gdavison, line 34: In this and other specifications, `Region` is always configured by the generator. Consider using `ServicePackageResourceRegion` instead of `* ServicePackageResourceRegion`

carlcauchi:
what is the eta for this PR to be merged to have the beta version available please? @ewbankkit

ewbankkit:
@carlcauchi See https://github.com/hashicorp/terraform-provider-aws/issues/41101 -- the first beta should be available towards the end of this month.

carlcauchi:
> @carlcauchi See #41101 -- the first beta should be available towards the end of this month.

With the new region parameter introduced in AWS provider 6.0.0 beta, if I pass it into a module like the community aws-vpc module:
```
module "vpc" {
  source = "terraform-aws-modules/vpc/aws"
  version = "5.0.0"
  region  = var.region
}
```
Will the provider inside the module automatically pick up the region from the new provider-level support, or do we need to update directly all the community modules to support passing the region parameter like a normal variable?

internal/types/service_package.go
ewbankkit: Relates https://github.com/hashicorp/terraform-provider-aws/pull/42186.

ewbankkit:
@carlcauchi Not for the beta -- We have considered the module case and intend to implement transparently in a future phase (so that each resource in the module doesn't have to add a `region = var.region` argument).
Please add a comment to the root issue https://github.com/hashicorp/terraform-provider-aws/issues/27758 if you wish.

PR #42133, opened 2025-04-04
Update docs to make addresses optional to match the code

rjw1:
The [code](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/wafv2/ip_set.go#L63) suggests that actually addresses is optional. In practice this also seems to the case


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@rjw1 Thanks for the contribution :tada: :clap:.

PR #42186, opened 2025-04-09
Updates Transparent Tagging structs to use `unique.Handle` for deduplication

gdavison:
### Description

Uses [`unique.Handle`](https://pkg.go.dev/unique#Handle) for `ServicePackageResourceTags` deduplication.

Before change

BenchmarkSDKProviderInitialization-10    	       1	  10538708 ns/op	27357848 B/op	  139373 allocs/op

After change

BenchmarkSDKProviderInitialization-10    	       1	  11033041 ns/op	27333384 B/op	  138609 allocs/op

Eliminates 764 allocations and approximately 25 KB when initializing SDKv2 schemas. (Framework was not measured)

### DIFF
--- a/.teamcity/scripts/provider_tests/acceptance_tests.sh
+++ b/.teamcity/scripts/provider_tests/acceptance_tests.sh
@@ -61,6 +61,7 @@ TF_ACC=1 go test \
     ./internal/tags/... \
     ./internal/tfresource/... \
     ./internal/types/... \
+    ./internal/unique/... \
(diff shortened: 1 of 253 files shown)


ewbankkit:
Nice 🎉.
We can do something similar for the new `ServicePackageResourceRegion` structs being generated for per-resource Region override: https://github.com/hashicorp/terraform-provider-aws/pull/42014.

jar-b (approved):

Issue #42164, opened 2025-04-07
Support new security policies for AWS Transfer Family servers


jmiller-ca:
### Description

Planning failed. Terraform encountered an error while generating this plan.

```
╷
│ Error: expected security_policy_name to be one of ["TransferSecurityPolicy-2018-11" "TransferSecurityPolicy-2020-06" "TransferSecurityPolicy-2022-03" "TransferSecurityPolicy-2023-05" "TransferSecurityPolicy-2024-01" "TransferSecurityPolicy-FIPS-2020-06" "TransferSecurityPolicy-FIPS-2023-05" "TransferSecurityPolicy-FIPS-2024-01" "TransferSecurityPolicy-FIPS-2024-05" "TransferSecurityPolicy-PQ-SSH-Experimental-2023-04" "TransferSecurityPolicy-PQ-SSH-FIPS-Experimental-2023-04" "TransferSecurityPolicy-Restricted-2018-11" "TransferSecurityPolicy-Restricted-2020-06" "TransferSecurityPolicy-Restricted-2024-06"], got TransferSecurityPolicy-SshAuditCompliant-2025-02
│ 
│   with aws_transfer_server.sftp_server,
│   on sftp.tf line 4, in resource "aws_transfer_server" "sftp_server":
│    4:   security_policy_name   = var.sftp_server_security_policy_name
│ 
```

### Affected Resource(s) or Data Source(s)

* `aws_transfer_server`

### Potential Terraform Configuration

```hcl
resource "aws_transfer_server" "sftp_server" {
  identity_provider_type = SERVICE_MANAGED"
  logging_role           = aws_iam_role.sftp_server.arn
  security_policy_name   = TransferSecurityPolicy-SshAuditCompliant-2025-02
}
```

### References

https://docs.aws.amazon.com/transfer/latest/userguide/security-policies.html#security-policy-transferSecurityPolicy-SshAuditCompliant-2025-02

### Would you like to implement the enhancement?

No

fitsnips:
https://github.com/hashicorp/terraform-provider-aws/pull/42167

PR #42191, opened 2025-04-10
Fix "Ssh" in variable name to "SSH" to comply with naming caps rule to prevent CI failure (Naming Scan Caps/AWS/EC2)

tabito-hara:
### Description
* Since #42164 was merged, the CI check *Naming Scan Caps/AWS/EC2* continues to fail.  
  * Example log: https://github.com/hashicorp/terraform-provider-aws/actions/runs/14372058248/job/40296841581?pr=42188  
* This is because a variable name containing *SSH* does not comply with the naming caps rule.  
* This PR fixes the variable name to prevent the CI check from failing.
  * It has been confirmed that the renamed variable is not referenced in any other files.


ewbankkit:

PR #42212, opened 2025-04-12
r/eks_addon add incremental_version_update argument

### LINKED ISSUE #39182: [Enhancement]: add argument to aws_eks_addon for changing version in steps
### Description

The [documentation](https://docs.aws.amazon.com/eks/latest/userguide/add-ons-vpc-cni.html) for the Amazon VPC CNI plugin for Kubernetes says:

> You can only update one minor version at a time.

(That's actually advice, not currently enforced by AWS.)

Likewise, the [documentation](https://docs.aws.amazon.com/eks/latest/userguide/add-ons-pod-id.html) of the EKS Pod Identity Agent says:

> You can only update one minor version at a time.

Therefore, it would be very useful if the `aws_eks_addon` resource had an argument to tell the provider to follow that advice when changing the version of an add-on.

For example, suppose that the provider is changing the version of the `vpc-cni` add-on from v1.13.4-eksbuild.1 to v1.16.4-eksbuild.2. Rather than making the change in one step, it could:

* change from v1.13.4-eksbuild.1 to v1.14.1-eksbuild.1;
* then change to v1.15.5-eksbuild.1;
* then change to v1.16.4-eksbuild.2.

* `aws_eks_addon`

```terraform
resource "aws_eks_addon" "example" {
  cluster_name            = aws_eks_cluster.example.name
  addon_name              = "vpc-cni"
  change_version_in_steps = true
}
# I actually don't much like the name "change_version_in_steps";
# you may be able to think of something better.
```

### References

_No response_

No

naruse666:
### Description

When `incremental_version_update = true` is set and the following version upgrade is attempted:

```hcl
addon_version = "v1.30.9-eksbuild.3" → "v1.32.0-eksbuild.2"
```

The apply will fail with the following error:
```console
incremental_version_update is enabled: minor version upgrade must be done one version at a time (detected jump: 30 → 32)
```

### Relations
Closes #39182

```console
% make testacc TESTS=TestAccEKSAddon_addonVersion PKG=eks
--- PASS: TestAccEKSAddon_addonVersion (342.44s)
```


internal/service/eks/addon.go
```diff
@@ -84,6 +97,11 @@ func resourceAddon() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"incremental_version_update": {
```
naruse666, line 100: The current name `incremental_version_update` may not be perfect.
Suggestions for a clearer or more conventional name are very welcome.

internal/service/eks/addon_test.go
```diff
@@ -114,9 +115,10 @@ func TestAccEKSAddon_addonVersion(t *testing.T) {
 	var addon1, addon2 types.Addon
 	rName := sdkacctest.RandomWithPrefix(acctest.ResourcePrefix)
 	resourceName := "aws_eks_addon.test"
-	addonName := "vpc-cni"
-	addonVersion1 := "v1.17.1-eksbuild.1"
-	addonVersion2 := "v1.18.5-eksbuild.1"
+	addonName := "kube-proxy"
```
naruse666, line 118: When I ran the tests locally, the Kubernetes version was 1.32.
I switched to using kube-proxy for testing, based on the available versions returned by `aws eks describe-addon-versions`.

PR #42208, opened 2025-04-11
r/aws_elasticache_serverless_cache: allow in-place `redis` to `valkey` engine updates

jar-b:
### Description

Previously changing the `engine` from `redis` to `valkey` would have forced a replacement of the resource. Migrations from `redis` to `valkey` are now supported in-place as intended in the original implementation of `valkey` support.

### Relations

Relates #39745
Closes #42204 

### References

- https://discuss.hashicorp.com/t/in-place-update-for-aws-elasticache-serverless-cache-from-redis-to-valkey/73319
- https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ModifyServerlessCache.html

### Output from Acceptance Testing

```console
% make testacc PKG=elasticache TESTS=TestAccElastiCacheServerlessCache_
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.23.8 test ./internal/service/elasticache/... -v -count 1 -parallel 20 -run='TestAccElastiCacheServerlessCache_'  -timeout 360m -vet=off
2025/04/11 16:04:25 Initializing Terraform AWS Provider...

--- PASS: TestAccElastiCacheServerlessCache_full (399.87s)
--- PASS: TestAccElastiCacheServerlessCache_basicValkey (556.51s)
--- PASS: TestAccElastiCacheServerlessCache_basicRedis (567.35s)
--- PASS: TestAccElastiCacheServerlessCache_disappears (584.55s)
--- PASS: TestAccElastiCacheServerlessCache_description (584.98s)
--- PASS: TestAccElastiCacheServerlessCache_fullValkey (586.27s)
--- PASS: TestAccElastiCacheServerlessCache_tags (632.07s)
--- PASS: TestAccElastiCacheServerlessCache_fullRedis (638.54s)
--- PASS: TestAccElastiCacheServerlessCache_redisUpdateWithUserGroup (797.42s)
--- PASS: TestAccElastiCacheServerlessCache_cacheUsageLimits (874.83s)
--- PASS: TestAccElastiCacheServerlessCache_engine (1456.48s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/elasticache        1463.285s
```

PR #42222, opened 2025-04-14
internal/framework/flex: skip planned unknown values in `Diff`

jar-b:
### Description

Previously a computed or optional/computed attribute in state without a `UseStateForUnknown` plan modifier would cause `flex.Diff` to indicate the plan has changes due to the previously known value now being known after apply. In this scenario we now skip this field as it can't be ensured whether or not the value has changed.

### Relations

Relates #42208 (behavior was uncovered while implementing this fix)

```console
% make testacc PKG=elasticache TESTS=TestAccElastiCacheServerlessCache_tags
TF_ACC=1 go1.23.8 test ./internal/service/elasticache/... -v -count 1 -parallel 20 -run='TestAccElastiCacheServerlessCache_tags'  -timeout 360m -vet=off
2025/04/14 11:13:08 Initializing Terraform AWS Provider...

--- PASS: TestAccElastiCacheServerlessCache_tags (381.21s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/elasticache        388.060s
```

### DIFF
--- a/internal/framework/flex/diff.go
+++ b/internal/framework/flex/diff.go
@@ -76,6 +76,10 @@ func Diff(ctx context.Context, plan, state any, options ...ChangeOption) (*Resul
 			continue
 		}
 
+		if planFieldValue.IsUnknown() {
+			continue
+		}
+
 		if !planFieldValue.Equal(stateFieldValue) {
 			hasChanges = true
 		} else {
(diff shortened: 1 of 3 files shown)


johnsonaj (approved):
LGTM 🚀

PR #42294, opened 2025-04-18
Add check to notify if a changelog entry is needed

justinretzolk:
### Description

Adds a section to the warning comment workflow to indicate that a changelog entry is needed on non-test changes to files within `internal/service/` subdirectories

Also adds the requisite `no-changelog-needed` label to help with exclusions where the general check provides a false-positive.

N/a, workflow

### DIFF
--- a/infrastructure/repository/labels-workflow.tf
+++ b/infrastructure/repository/labels-workflow.tf
@@ -125,7 +125,7 @@ variable "workflow_labels" {
       color       = "ac72f0", # color:terraform (link on black)
       description = "Introduces a new service."
     },
-    "no-changelog" = {
+    "no-changelog-needed" = {
       color       = "828a90", # color:stale grey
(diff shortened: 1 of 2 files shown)


infrastructure/repository/labels-workflow.tf
```diff
@@ -125,6 +125,10 @@ variable "workflow_labels" {
       color       = "ac72f0", # color:terraform (link on black)
       description = "Introduces a new service."
     },
+    "no-changelog" = {
```
jar-b, line 128: In case we want to be consistent with Terraform core, that project uses a `no-changelog-needed` label.

https://github.com/hashicorp/terraform/pulls?q=is%3Apr+is%3Aopen+label%3Ano-changelog-needed

jar-b (approved):
LGTM 🚀

PR #42384, opened 2025-04-25
Restore `global` setting to `@Region` annotation

gdavison:
### Description

Resource Identity leverages the `@Region` annotation from Per-Resource Region Override. However, Identity needs to identify global resource types.

Restores the `global` flag to the `@Region` annotations, with a default of `false`. 

`ServicePackageResourceRegion.IsOverrideEnabled` is now set to `overrideEnabled && !global`


ewbankkit:
It is better to base this PR on `f-per-resource-override-Region-sdkv2-phaseII` as that is the branch that extends `f-per-resource-override-Region-sdkv2-phaseI` to add support for Framework resources and is closer to what will be the `v6.0.0-beta1`.

ewbankkit:
I have been capturing the list of "global resources in region services" here: https://github.com/hashicorp/terraform-provider-aws/issues/27758#issuecomment-2587102363

ewbankkit:
Added all the resources and data source for _Global Resources In Regional Services_ (https://github.com/hashicorp/terraform-provider-aws/issues/27758#issuecomment-2587102363). `make gen` shows no diff (a good sign).

johnsonaj (approved):
LGTM 🚀

Issue #40292, opened 2024-11-25
AWS User Notifications


ewbankkit:

abstrask:
Duplicate of #34969 ?

ewbankkit:
Indeed it is -- I didn't realize the service had been around (with an API) for so long.

yshollander-selerity:
My AWS Health emails now have the following header:


[Starting September 15, 2025, emails about AWS Health events will be delivered via the AWS User Notifications service. You can opt-in to this enhanced experience today. See footer for details.]

There is in fact an opt-in button in User Notifications.

Now there is a timeframe.  There is a proposal in [#34969](https://github.com/hashicorp/terraform-provider-aws/issues/34969).  

Particularly for large organizations, with many accounts each with unique notification requirements, it would be really helpful to be able to automate this configuration...

Issue #34969, opened 2023-12-18
[New Service]: AWS User Notifications


F-Xu:
### Description

AWS announced User Notifications general availability on 3rd May 2023.

### Requested Resource(s) and/or Data Source(s)

aws_usernotification_configuration
aws_usernotification_deliverychannel
aws_usernotification_hub

### References

https://aws.amazon.com/about-aws/whats-new/2023/05/aws-user-notifications-available/
https://docs.aws.amazon.com/notifications/latest/userguide/what-is-service.html

### Would you like to implement a fix?

None

colonmelvin:
Per [AWS documentation](https://aws.amazon.com/notifications/faqs/) there is no API for User Notifications. 

_Does AWS User Notifications provide users with APIs to access notifications?_
No, AWS User Notifications does not provide programmatic access to notifications.

iwamot:
AWS User Notifications API is now available.
https://docs.aws.amazon.com/notifications/latest/APIReference/Welcome.html

jatin-mehrotra-colorkrew:
https://aws.amazon.com/about-aws/whats-new/2024/11/aws-user-notifications-sdk/

SDK is available which also means API is also available

DanHyland:
Bump now that the API and SDK are now available. Is there an ETA when we can have TF resources to manage AWS User Notifications?

marcinbelczewski:
@srgoni are willing to contribute the design proposed above or possibly you'd welcome some assistance? I could contribute as well.

marcinbelczewski:
Thanks for comment @srgoni. I will try to take a stab and if I'm successful I will create a draft PR. Cheers!

marcinbelczewski:
Here are my comments @srgoni:

Ad 1: I believe it absolutely makes sense to reference AWS docs here. Good idea.

Ad 2: I decided to skip those after struggling with `status_summary` for a moment. "Flexed" response would need to land into nested attributes which are currently supported in protocol ver 6 while this provider is still on 5. So I'd need to make this a little less pretty and since those aren't really used for anything meaningful (I reckon) I decided to skip those.

Ad 3: Current implementation already waits for the deregistration of the hub. I believe what takes place here, is parallel invocation of destruction and creation starting almost instanteanously resulting in conflict. The way I'd migrate the code from one to another would be:

```terraform
resource "aws_notifications_notification_hub" "test" {
  for_each = toset(["us-west-1"])
  notification_hub_region = each.value
}

[...]

PR #42399, opened 2025-04-28
New Service - Add support for AWS User Notifications

srgoni:
### Description

This PR adds the the scaffolding for supporting AWS User Notifications in the Terraform AWS provider.

AWS has split functionality over two separate service tags, so there will be two separate service packages: `notifications` and `notificationscontacts`. Notifications Contacts only contains a few APIs for managing email addresses for User Notifications.

I attempted to put both into the same Terraform provider package, but failed. It doesn't look like there is a single arrangement of parameters in `names_data.hcl` that wouldn't fail one of the verifications. Therefore, and because AWS has released both service APIs and Go packages under separate namespaces, I opted for a separate package in the Terraform provider, but set it up as a sub-service and used the same documentation prefix. Maybe this needs to be changed as well.

This is the first time I'm contributing here, so any guidance would be appreciated.

### Relations

Contributes to #34969 and #40292 

### References

* https://docs.aws.amazon.com/notifications/latest/APIReference/Welcome.html
* https://docs.aws.amazon.com/notificationscontacts/latest/APIReference/Welcome.html

N/A (this PR is just the scaffolding, no related unit tests are defined yet)


justinretzolk:
Hey @srgoni 👋 Thank you for taking the time to raise this! That warning comment is a fairly new workflow that we've introduced to try to help ensure PR reviews are as streamlined as possible. I'll make a note to update it to help exclude the warning when a new service is added. Thanks for calling that out!

ewbankkit (approved):
LGTM :rocket:.

johnsonaj (approved):

ewbankkit:
@srgoni Thanks for the contribution :tada: :clap:.

PR #42378, opened 2025-04-25
Move warning comment to it's own workflow

justinretzolk:
### Description

While modifying the "warning comment" job to cope with not warning of missing changelog entries on draft PRs, it became apparent that the scope of it had changed enough to warrant breaking it out into its own workflow. 

Aside from separating the job into its own workflow, the changes here include:

- Introduces the `ready_for_review` trigger type to re-run the job when a draft PR is marked as ready for review. This allows us to skip the "Needs Changelog Entry" step when a draft PR is opened, allowing contributors to open the PR as a draft first, add any necessary changelog files, and then mark the PR as ready for review without receiving the warning comment.
- Removes `edited` from the trigger types, since those types of changes don't have an impact on any of the steps within the job
- Runs the community check when a PR is determined to need a changelog entry, allowing the "Needs Changelog Entry" step to be skipped for maintainers (mostly a convenience for maintainers)
- In the Auto-Triage workflow, removed a "Debug Output" step that was leftover from local testing

### Output from Acceptance Testing

N/a, workflows

PR #42415, opened 2025-04-29
Enhance Auto Triage Workflows

justinretzolk:
### Description

As mentioned in the linked, but now closed PR, while modifying the "warning comment" job to cope with not warning of missing changelog entries on draft PRs, it became apparent that the scope of it had changed enough to warrant breaking it out into its own workflow. 

Aside from separating that job into its own workflow, the changes here include:

Community note:
- Separates the job out into its own workflows since it only ever runs when a PR or issue is opened, and is otherwise skipped every time, creating unnecessary costs.

Closed items:
- Separates the job out into its own workflows since it only ever runs when a PR or issue is closed, and is otherwise skipped every time, creating unnecessary costs.

Auto Triage:
- Removed a "Debug Output" step that was leftover from local testing
- Removed extraneous triggers leftover from moving other jobs out to their own workflows

Warning comment:
- Introduces the `ready_for_review` trigger type to re-run the job when a draft PR is marked as ready for review. This allows us to skip the "Needs Changelog Entry" step when a draft PR is opened, allowing contributors to open the PR as a draft first, add any necessary changelog files, and then mark the PR as ready for review without receiving the warning comment.
- Adds an additional filter to the check for dependency updates so that the "dependencies updated by non-maintainer" comment is skipped when a new service is being introduced
- Removes `edited` from the trigger types, since those types of changes don't have an impact on any of the steps within the job
- Runs the community check when a PR is determined to need a changelog entry, allowing the "Needs Changelog Entry" step to be skipped for maintainers (mostly a convenience for maintainers)
- Updates the formatting of the warning messages so they're all within disclosure triangles

### Relations

Relates #42378

n/a, workflow


.github/workflows/readiness_comment.yml
```diff
@@ -0,0 +1,177 @@
...
+      - name: Add Start Message
+        id: start
+        if: |
+          steps.maintainer_editability.outcome != 'skipped'
+          || steps.dependencies.outcome != 'skipped'
+          || steps.changelog.outcome != 'skipped'
+          || steps.needs-changelog-entry.outcome != 'skipped'
+        shell: bash
+        run: |
+          { echo $START_TEXT; echo "---"; cat note.md; } > tmpnote && mv tmpnote note.md
```
jar-b, line 148: Could this write directly to `note.md` instead of the `mv` step?

```suggestion
          { echo $START_TEXT; echo "---"; cat note.md; } > note.md
```

jar-b (approved):
LGTM, just one optional suggestion

.github/workflows/readiness_comment.yml
jar-b: Ah - I missed that `note.md` was referenced when constructing the text. This approach seems fine if we want to merge and defer on an alternative solution for later?

PR #32090, opened 2023-06-19
Deprecate "legacy" behaviour when parsing nullable Bool values

gdavison:
### Description

The parsing rules for `TypeNullableBool` currently allow `0` and `1` as valid values, matching behaviour of very old versions of Terraform. This PR deprecates this behaviour and return a warning in the validation function.

Note: The ideal solution is to update the resources to use `terraform-plugin-framework` instead of `terraform-plugin-sdk/v2`. `terraform-plugin-framework` allows typed nullable values.

The parsing behaviour will be removed by #29379.

### Relations

Closes #29378

### Output from Testing

```
$ go test ./internal/experimental/nullable/... -v

--- PASS: TestNullableBool (0.00s)
--- PASS: TestNullableFloat (0.00s)
--- PASS: TestDiffSuppressBoolFalseAsNull (0.00s)
--- PASS: TestDiffSuppressBool (0.00s)
--- PASS: TestValidationBool (0.00s)
--- PASS: TestValidationFloat (0.00s)
--- PASS: TestNullableInt (0.00s)
--- PASS: TestValidationInt (0.00s)
--- PASS: TestValidationIntAtLeast (0.00s)
```

PR #42434, opened 2025-04-30
internal/sdkv2/types/nullable: refine `TypeNullableBool` validation

### LINKED ISSUE #29379: Restrict parsing of nullable Bool values to `""`, `false`, and `true`
### Description

In order to simulate values that support Boolean values that can also be unset, the provider currently uses `TypeNullableBool` which is implemented as a `TypeString` with special validation and parsing rules to restrict to boolean values.

The parsing rules for `TypeNullableBool` currently allow `0` and `1` as valid values, matching behaviour of very old versions of Terraform. Only allow the values `""`, `false`, and `true`.

Note: The ideal solution is to update the resources to use `terraform-plugin-framework` instead of `terraform-plugin-sdk/v2`. `terraform-plugin-framework` allows typed nullable values.

The previous parsing behaviour was deprecated in #29378.

_No response_

_No response_

### References

_No response_

None

jar-b:
### Description

Arguments of the type `TypeNullableBool` will now only accept one of `""` (empty string), `true`, or `false`. An upgrade guide entry is included enumerating the resources and arguments which are affected by this validation change.

### Relations

Closes #29379
Relates #32090
Relates #41101

### Output from Testing

```console
% go test -count=1 ./internal/sdkv2/types/nullable/...
ok      github.com/hashicorp/terraform-provider-aws/internal/sdkv2/types/nullable       0.208s
```

### DIFF
--- a/.changelog/42434.txt
+++ b/.changelog/42434.txt
@@ -1,0 +1,36 @@
+```release-note:breaking-change
+resource/aws_accessanalyzer_archive_rule: `filter.exists` now only accepts one of `""` (empty string), `true`, or `false`
+```
+```release-note:breaking-change
+resource/aws_cloudtrail_event_data_store: `suspend` now only accepts one of `""` (empty string), `true`, or `false`
+```
+```release-note:breaking-change
+resource/aws_ec2_spot_instance_fleet: `terminate_instances_on_delete` now only accepts one of `""` (empty string), `true`, or `false`
+```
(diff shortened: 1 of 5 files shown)


ewbankkit (approved):
LGTM :rocket:.

As the `ValueBool()` function has not change (still using `strconv.ParseBool()`), any `"1"`s or `"0"`s in state will still work as `true` or `false` respectively.

PR #42444, opened 2025-05-01
security/tests: Add `enable_key_rotation` to `aws_kms_key`

YakDriver:
### Description

Best practices for security are that key rotation should be enabled. This updates tests, which, strictly speaking, do not need adhere to these standards. However, to avoid additional findings and provide a good example, we are adding them.

### Relations

Closes #0000

### References

* [Example finding](https://github.com/hashicorp/terraform-provider-aws/security/code-scanning/364)

```console

...
```

### DIFF
--- a/internal/service/mwaa/environment_test.go
+++ b/internal/service/mwaa/environment_test.go
@@ -914,8 +914,10 @@ resource "aws_mwaa_environment" "test" {
 data "aws_region" "current" {}
 
 resource "aws_kms_key" "test" {
-  description = "Key for a Terraform ACC test"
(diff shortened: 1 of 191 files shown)


internal/service/mwaa/environment_test.go
```diff
@@ -964,6 +966,8 @@ resource "aws_s3_object" "startup_script" {
 }
 
 
+
+
```
jar-b, line 970: Remove this extra whitespace?

YakDriver: Good catch. Very strange where that came from...

jar-b (approved):
LGTM 🚀

PR #42486, opened 2025-05-05
Documentation: Standardize `Argument Reference` byline

ewbankkit:
### Description

Should be one of

* "This resource supports the following arguments:"
* "The following arguments are required:"
* "The following arguments are optional:"
* "This resource does not support any arguments."


jar-b (approved):
LGTM 🚀

PR #42537, opened 2025-05-08
Deep update dependencies

YakDriver:
### Description

This uses a new version of tfproviderdocs (v0.19.0) that allows data sources to not be checked, and does not check the renegade data sources `aws_kms_secrets` and `aws_kms_secret`.

### Relations

Closes #42528

### References

* https://github.com/YakDriver/tfproviderdocs/pull/180

### Output from Testing

```console
% make tfproviderdocs
make: Provider Checks / go-build (darwin_arm64)...
make: Provider Checks / tfproviderdocs...
Initializing the backend...
Initializing provider plugins...
- Finding latest version of hashicorp/aws...
- Installing hashicorp/aws v99.99.99...
- Installed hashicorp/aws v99.99.99 (unauthenticated)
Terraform has created a lock file .terraform.lock.hcl to record the provider
selections it made above. Include this file in your version control repository
so that Terraform can guarantee to make the same selections by default when
you run "terraform init" in the future.

╷
│ Warning: Incomplete lock file information for providers
│ 
│ Due to your customized provider installation methods, Terraform was forced to calculate lock file checksums locally for the
│ following providers:
│   - hashicorp/aws
│ 
│ The current .terraform.lock.hcl file only includes checksums for darwin_arm64, so Terraform running on another platform will fail
│ to install these providers.
│ 
│ To calculate additional checksums for another platform, run:
│   terraform providers lock -platform=linux_amd64
│ (where linux_amd64 is the platform to generate)
╵
Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.
```

PR #42543, opened 2025-05-08
Go mod tidy

YakDriver:
### TL;DR

This PR:

1. Fixes the CI workflow in `.github/workflows/dependencies.yml` so that it checks the actual changes in pull requests, not the base branch. The previous use of `pull_request_target` caused broken dependencies in PRs to go undetected. Switching to `pull_request` ensures accurate validation moving forward.
2. Rolls back upgrades to tflint that cause problems.
3. Adds a CI check for use of `pull_request_target` except in approved places.
4. Is ❌ because the base branch `release/6.0.0-beta` has a badness that is being checked. 👇 See for additional details.

### Description

This PR updates `.github/workflows/dependencies.yml` to use the correct GitHub Actions trigger for validating pull request changes.

Previously, the workflow was using the `pull_request_target` event, which causes the workflow to run in the context of the **base branch**, not the pull request (PR) itself. This behavior is useful for workflows that need elevated permissions (e.g., to apply labels or comment on PRs), but it's not appropriate for CI checks that need to validate the actual code being proposed.

As a result, if a PR introduced broken dependencies (e.g., missing or outdated modules), the CI checks still passed — because they were validating the base branch, where those issues didn't exist. This allowed regressions to be merged undetected.

### What this PR changes

This PR replaces the `pull_request_target` trigger with the `pull_request` trigger, which runs the workflow against the **PR’s changes**, as expected for dependency validation and similar CI checks.

### Why the job is currently running twice

Until this PR is merged, the `main` branch still contains the old `pull_request_target` trigger. When a PR like this one is opened, both triggers are active:
- `pull_request_target` runs from the base branch (main)
- `pull_request` runs from the PR branch (this change)

That’s why you'll see the job run twice in this PR.

### What happens after merging

Once this PR is merged, the `pull_request_target` trigger will be removed from the base branch, so future PRs will only run the workflow once — correctly, against the PR content.

This change ensures that dependency checks (and any other validations in the `dependencies.yml` workflow) reflect the actual code being submitted in the PR.

### Relations

Relates #42537

### References

- Events that trigger workflows: `GITHUB_REF` for [`pull_request`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) vs [`pull_request_target`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target)
- [Security hardening for `pull_request_target`](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-pull_request_target)
- [Workflow permissions and the `permissions` key](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)

```console
% make deps-check
make: Tidying Go mods...
make: tidying with go1.24.2
make: Go mods tidied
make: Dependency Checks / go_mod...
% make pr-target-check 
make: Checking for pull request target...
make: pull_request_target check passed.
```


johnsonaj (approved):
LGTM 🚀

PR #42549, opened 2025-05-08
Enhanced Regions import: allow setting region directly from ARN or region name used as Import ID

### LINKED ISSUE #42527: When using Enhanced Region, allow setting region directly from ARN or region name used as Import ID
### Description

Some resource types use an ARN or a region name as their Import ID. When using Enhanced Region, this makes specifying the region by appending `@<region>` redundant.

If no additional region is specified, use the region from the Import ID. If an additional region is specified, it should match.

_No response_

```hcl

```

### References

_No response_

No

gdavison:
### Description

Some resource types use an ARN or a region name as their Import ID. When using Enhanced Region, this makes specifying the region by appending `@<region>` redundant.

If no additional region is specified, use the region from the Import ID. If an additional region is specified, it should match.

### Relations

Closes #42527

```console

...
```

### DIFF
--- a/internal/acctest/state_id.go
+++ b/internal/acctest/state_id.go
@@ -21,3 +21,21 @@ func AttrImportStateIdFunc(resourceName, attrName string) resource.ImportStateId
 		return rs.Primary.Attributes[attrName], nil
 	}
 }
+
+// CrossRegionImportStateIdFunc is a resource.ImportStateIdFunc that appends the region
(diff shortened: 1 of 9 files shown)


ewbankkit (dismissed):
LGTM :rocket:.

For future consideration, can we add _semgrep_ rules to check that new resources have the correct annotations?

PR #42615, opened 2025-05-14
Fix DataExchange sweepers

ewbankkit:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

Fixes

```
2025/05/14 08:25:29 [DEBUG] Running Sweeper (aws_dataexchange_revision) in region (us-gov-west-1)
2025/05/14 08:25:29 [INFO]  sweeper: listing resources: sweeper_region=us-gov-west-1 tf_resource_type=aws_dataexchange_revision
2025/05/14 08:25:29 [DEBUG] Completed Sweeper (aws_dataexchange_revision) in region (us-gov-west-1) in 209µs
2025/05/14 08:25:29 [ERROR] Error running Sweeper (aws_dataexchange_revision) in region (us-gov-west-1): listing "aws_dataexchange_revision" (us-gov-west-1): operation error DataExchange: ListDataSetRevisions, 1 validation error(s) found.
- missing required field, ListDataSetRevisionsInput.DataSetId.
```

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/42272.

### Output from Acceptance Testing

```console
% SWEEP=us-gov-west-1 SWEEPERS=aws_dataexchange_data_set make sweep
# make sweep SWEEPARGS=-sweep-run=aws_example_thing
# set SWEEPARGS=-sweep-allow-failures to continue after first failure
WARNING: This will destroy infrastructure. Use only in development accounts.
go1.23.8 test ./internal/sweep -v -sweep=us-gov-west-1 -sweep-run='aws_dataexchange_data_set' -timeout 360m -vet=off
2025/05/14 09:12:35 [DEBUG] Running Sweepers for region (us-gov-west-1):
2025/05/14 09:12:35 [DEBUG] Sweeper (aws_dataexchange_data_set) has dependency (aws_dataexchange_event_action), running..
2025/05/14 09:12:35 [DEBUG] Running Sweeper (aws_dataexchange_event_action) in region (us-gov-west-1)
2025/05/14 09:12:35 [DEBUG] sweeper: Configuring Terraform AWS Provider: sweeper_region=us-gov-west-1 tf_resource_type=aws_dataexchange_event_action
[...]

PR #42620, opened 2025-05-14
sweeper/aws_dataexchange_revision: listing revisions requires data set ID

gdavison:
Fixes `aws_dataexchange_revision` sweeper to pass data set ID when listing revisions


ewbankkit:
The fix is in https://github.com/hashicorp/terraform-provider-aws/pull/42615 which I will now merge...

Issue #42598, opened 2025-05-13
resource/aws_redshift_cluster: The AWS default for `encrypted` is `true` but provider defaults to `false`


gdavison:
### Terraform and AWS Provider Version

```console
v5
```

### Affected Resource(s) or Data Source(s)

* aws_redshift_cluster

### Expected Behavior

The default for `encrypted` should be `true`

### Actual Behavior

The default for `encrypted` is `false`, this causes a permadiff unless `encrypted` is explicitly set to `true`. This also causes multiple CI failures

### Relevant Error/Panic Output

```console

```

### Sample Terraform Configuration

```terraform
resource "aws_redshift_cluster" "test" {
  cluster_identifier                  = %[1]q
  availability_zone                   = data.aws_availability_zones.available.names[0]
  database_name                       = "mydb"
  master_username                     = "foo_test"
  master_password                     = "Mustbe8characters"
  multi_az                            = false
  node_type                           = "dc2.large"
  automated_snapshot_retention_period = 0
  allow_version_upgrade               = false
  skip_final_snapshot                 = true
}
```

### Steps to Reproduce

1. Apply the configuration

### Debug Logging

<details>
<summary>Click to expand log output</summary>

```console

```
</details>

### GenAI / LLM Assisted Development

n/a

### Important Facts and References

The `_basic` test should also not set `encrypted`, as it should only set `Required` attributes

### Would you like to implement a fix?

No

gdavison:
The `_basic` test should also not set `encrypted`

ewbankkit:
Closed via https://github.com/hashicorp/terraform-provider-aws/pull/42631.

PR #42635, opened 2025-05-15
docs: add default values recommendation

jar-b:
## Rollback Plan

### Description

Documents the recommended practice for the use of provider side default values. In general, arguments with server side default should store the value returned from AWS rather than setting a matching provider side default.

### Relations

Relates #42598


ewbankkit (approved):
LGTM :rocket:.

Issue #33693, opened 2023-09-29
[Bug]: AWS Provider: ec2_client_vpn_endpoint authentication_options type is missing options


justinmdudley:
### Terraform Core Version

1.5.7

### AWS Provider Version

5.18.1

### Affected Resource(s)

ec2_client_vpn_endpoint 
authentication_options 
type

### Expected Behavior

The Terraform AWS documentation list three possible options for authentication_options type (certificate-authentication, directory-service-authentication, and federation-authentication). AWS Client VPN also supports a combination of mutual authentication (certificate-authentication) and user-based methods, as documented below. 

-- Terraform Documentation --
[type](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ec2_client_vpn_endpoint#type) - (Required) The type of client authentication to be used. Specify certificate-authentication to use certificate-based authentication, directory-service-authentication to use Active Directory authentication, or federated-authentication to use Federated Authentication via SAML 2.0.

-- AWS Documentation --
Client VPN offers the following types of client authentication:

[Active Directory authentication](https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/ad.html) (user-based)

[Mutual authentication](https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/mutual.html) (certificate-based)

[Single sign-on (SAML-based federated authentication)](https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/federated-authentication.html) (user-based)

You can use one of methods listed above alone, or a combination of mutual authentication with a user-based method such as the following:

Mutual authentication and federated authentication

Mutual authentication and Active Directory authentication

### Actual Behavior

The Terraform AWS provider does not accept the combined authentication methods listed in the AWS documentation. I need to create an AWS Client VPN Endpoint using mutual authentication and federated authentication. 

### Terraform Configuration Files

.

### Steps to Reproduce

.

### Would you like to implement a fix?

No

acwwat:
Since there's no Terraform configuration provided, I cannot tell exactly what the problem might be. But should be able to use two `authentication_options` configuration blocks to implement the combined mutual + user-based authentication scenario. Please give it a try and report back on the results.

sixty4k:
Not the original bug reporter, but can confirm that 2 `authentication_options` blocks works. 

However, this is non-obvious in the docs, nor the source code (that did start me in the right direction); perhaps this can be shifted from bug to documentation?

PR #42815, opened 2025-05-30
Documentation Improvement: ec2_client_vpn_endpoint, authentication options

sixty4k:
### Description

Clarification to the `ec2_client_vpn_endpoint` documentation, making more explicit how to use combinations of authentication options (called out in the AWS docs [here](https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/client-authentication.html))

I'm not entirely sold on the way I've worded this.  Will dig around to see if other resources with similar patterns of required _if_ another argument is set.

### Relations

Relates #33693

### References

* https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/client-authentication.html

n/a


sixty4k:
Any feedback is appreciated ...

PR #42825, opened 2025-05-30
test generation: Improves test generation time

gdavison:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the provider.

None

### Description

Some generated test code had missing Go imports, which causes `goimports` to take longer time.


ewbankkit (approved):
LGTM :rocket:.

`make gen` running much faster.

PR #42913, opened 2025-06-09
docs: enhanced aws_ssm_association for multiple instances/tag values

### LINKED ISSUE #42431: Improve ssm association documentation
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssm_association#argument-reference

### Description

Hi there, the documentation about `targets` is a bit confusing. Specifically, I had trouble figuring out how to target multiple instance ids. The parts that confused me were (emphasis added by me):
>Targets specify what instance IDs or tags to apply the document to and has these keys:
> - key - (Required) Either `InstanceIds` or `tag:Tag Name` to specify an EC2 tag.
> - values - (Required) A list of instance IDs or tag values. **AWS currently limits this list size to one value.**

After trying out this resource and the aws console, I determined that should not be creating multiple target blocks for multiple instance ids, and should instead have a single target block with a list of multiple instance ids.

I think the sentence in bold only applies to when the key is set to `tag:Tag Name` and should be updated to indicate that. It would also be helpful to update at least one example to use multiple instance ids.

### References

https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-state-manager-targets-and-rate-controls.html

No

quixoticmonk:
## Rollback Plan

None.

### Description

Updated the example documentation and attribute description to resolve #42431

### Relations

Closes https://github.com/hashicorp/terraform-provider-aws/issues/42431

```console

...
```

### DIFF
--- a/website/docs/r/ssm_association.html.markdown
+++ b/website/docs/r/ssm_association.html.markdown
@@ -81,7 +81,6 @@ resource "aws_ssm_association" "example" {
 
 ### Create an association with multiple instances with their instance ids
 
-
 ```
 # Removed EC2 provisioning dependencies for brevity
 
@@ -242,7 +241,6 @@ resource "aws_instance" "database_server" {
   }
 }
 ```
-
 
 ## Argument Reference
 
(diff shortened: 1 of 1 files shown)


ewbankkit:
@quixoticmonk Thanks for the contribution :tada: :clap:.

PR #42934, opened 2025-06-09
lambda: Revise documentation (part 1)

YakDriver:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Relations

Relates #42935

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccXXX PKG=ec2

...
```

PR #42935, opened 2025-06-09
lambda: Revise documentation (part 2)

YakDriver:
## Rollback Plan

### Relations

Relates #42934

```console

...
```


website/docs/d/lambda_invocation.html.markdown
```diff
@@ -3,50 +3,144 @@ subcategory: "Lambda"
...
+  })
 }
-JSON
+
+output "result" {
+  value = jsondecode(data.aws_lambda_invocation.example.result)
+}
+```
+
+### Configuration Validation
```
jar-b: These example feel a bit contrived. For example, Terraform variables already offer [validation](https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules), which would accomplish the same thing with less overhead.

I'm not against adding additional examples showing how to use the data source output, but perhaps only one is necessary?

website/docs/r/lambda_function_recursion_config.html.markdown
```diff
@@ -3,49 +3,98 @@ subcategory: "Lambda"
...
+
+```terraform
+# Explicitly set recursion to terminate (this is the default behavior)
+resource "aws_lambda_function_recursion_config" "example" {
+  function_name  = aws_lambda_function.example.function_name
+  recursive_loop = "Terminate"
+}
+```
+
+### Production Safety Configuration
```
jar-b, line 36: This is the same thing as the example above.

website/docs/r/lambda_invocation.html.markdown
```diff
@@ -3,117 +3,160 @@ subcategory: "Lambda"
...
+    }
   })
 
   lifecycle_scope = "CRUD"
+  terraform_key   = "terraform_lifecycle"
 }
 ```
 
-~> **NOTE:** `lifecycle_scope = "CRUD"` will inject a key `tf` in the input event to pass lifecycle information! This allows the lambda function to handle different lifecycle transitions uniquely.  If you need to use a key `tf` in your own input JSON, the default key name can be overridden with the `terraform_key` argument.
+~> **Note:** `lifecycle_scope = "CRUD"` will inject a key `terraform_lifecycle` (or the value of `terraform_key`) in the input event to pass lifecycle information! This allows the Lambda function to handle different lifecycle transitions uniquely.
```
jar-b: This change isn't quite correct - the `terraform_key` argument defaults to `tf` when not configured, but in the example above `terraform_key` is explicitly overridden.

website/docs/r/lambda_invocation.html.markdown
```diff
@@ -3,117 +3,160 @@ subcategory: "Lambda"
...
 
   input = jsonencode({
-    key1 = "value1"
-    key2 = "value2"
+    resource_name = "database_setup"
+    database_url  = aws_db_instance.example.endpoint
+    credentials = {
+      username = var.db_username
+      password = var.db_password
+    }
```
jar-b, line 85: For this example I think the trivial `key1` and `key2` arguments are easier to follow. Perhaps we also remove the `terraform_key` argument to show how the default key is injected when an alternate key name is not explicitly configured?

website/docs/r/lambda_invocation.html.markdown
YakDriver: Agreed and fixed!

website/docs/d/lambda_invocation.html.markdown
YakDriver: Agreed and removed!

website/docs/r/lambda_function_recursion_config.html.markdown
YakDriver: Excellent catch! I concur. Removed the simpler one.

website/docs/r/lambda_invocation.html.markdown
YakDriver: This one I disagree. Although it's not a big deal, the string literals aren't as realistic.

I did remove the alternate key name to show the `tf` default functionality.

jar-b (approved):
LGTM 🚀

PR #42952, opened 2025-06-10
generators: Shares common templates between Identity and Tagging tests

gdavison:
## Rollback Plan

### Description

Shares common templates between Identity and Tagging tests


internal/generate/testgen/acctest.tf.gtpl
ewbankkit, line 1: Maybe `tests` is a better package/directory name than `testsgen`?

jar-b (approved):
LGTM 🚀

PR #43012, opened 2025-06-13
Don't validate `region` if partition is empty

ewbankkit:
## Rollback Plan

### Description

When connecting to non-Amazon, [AWS-compatible APIs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/custom-service-endpoints) the `AWSClient.partition` field may be empty. In these cases don't validate the top-level `region` argument.


jar-b (approved):
LGTM 🚀

PR #43039, opened 2025-06-16
Go-VCR: Use internal state change retry functions

jar-b:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

A proof-of-concept for replacing state change refresh helpers from the Terraform Plugin-SDK V2 `retry` package with internal variants that enable replayed VCR acceptance testing. Specifically, when VCR is enabled all delay between state change checks is removed to allow replayed interactions to quickly pass through the CRUD handler's waiting logic. This can significantly reduce test execution times for resources which are slow to provision. For example, ElastiCache clusters:

Standard:

```
--- PASS: TestAccElastiCacheCluster_Engine_redis (703.67s)
--- PASS: TestAccElastiCacheCluster_Engine_redis_v5 (703.69s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/elasticache        709.501s
```

Replayed Interaction:

```
--- PASS: TestAccElastiCacheCluster_Engine_redis (10.20s)
--- PASS: TestAccElastiCacheCluster_Engine_redis_v5 (10.23s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/elasticache        15.940s
```

### Relations

Depends on #43070
Relates #25602
Relates #42930 (an alternative implementation)

### Output from Acceptance Testing

```console
% VCR_MODE=REPLAY_ONLY VCR_PATH=/Users/jaredbaker/development/_worktrees/f-vcr-status-check/testdata/ make testacc PKG=elasticache TESTS=TestAccElastiCacheCluster_Engine_redis
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.24.4 test ./internal/service/elasticache/... -v -count 1 -parallel 20 -run='TestAccElastiCacheCluster_Engine_redis'  -timeout 360m -vet=off
2025/06/17 11:12:14 Creating Terraform AWS Provider (SDKv2-style)...
2025/06/17 11:12:14 Initializing Terraform AWS Provider (SDKv2-style)...

=== NAME  TestAccElastiCacheCluster_Engine_redis
    cluster_test.go:84: stopping VCR recorder
    cluster_test.go:84: persisting randomness seed
--- PASS: TestAccElastiCacheCluster_Engine_redis (10.20s)
[...]

Issue #42554, opened 2025-05-09
Problematic interaction of explicit and implicit retries


ewbankkit:

rafael-debug:
When I try to update the EC2 AMI, TF returns the same error.
For it to work, I need to delete EC2 in the AWS console and run Terraform again.
I've already updated to v5.48.0, without success.

Issue #25602, opened 2022-06-28
Proposal: Enable HTTP Response Recording for long running Acceptance Tests 


breathingdust:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Some AWS services can take considerable time to provision, leading to Acceptance Test run times that can take in some cases nearly an hour for a single fixture. This creates a very slow development cycle time and slow development cycles and CI suite runs.

Prototype the integration of HTTP Response Recording ie ([go-vcr](https://github.com/dnaeon/go-vcr)) into the provider Acceptance Test suite to see if this would help us improve development cycle times. A balance will need to be struck in when to use the recorded responses, and when to require a "real" response from the AWS API's. We need to improve the development cycle, but not at the expense of detecting regressions.

This library is already in use within the [google](https://registry.terraform.io/providers/hashicorp/google/latest) provider.

### Tasks

- [x] Prototype Integration of [go-vcr](https://github.com/dnaeon/go-vcr)
- [ ] Propose usage pattern for development
- [ ] Propose usage pattern in CI

Relates #18419

ewbankkit:
[Phase 1](https://github.com/hashicorp/terraform-provider-aws/pull/25587) implemented the infrastructure required to VCR-enable a resource's acceptance tests and did so for 2 resources, `aws_cloudwatch_log_group` (which uses AWS SDK for Go v1) and `aws_scheduler_schedule` (which uses AWS SDK for Go v1).

As most acceptance testing configurations contain more than 1 resource type (e.g. many configurations contains `aws_iam_role` and `aws_vpc` resources), to get any significant value from `go-vcr` we need to enable VCR everywhere.
This will be the focus of Phase 2:

* Complete Graham's work converting all CRUD handlers to `WithoutTimeout(context.Context, ...)` variants: https://github.com/hashicorp/terraform-provider-aws/issues/15090
* Complete migration of all resources to self-registration mechanism to allow CRUD handler wrapping and custom context: https://github.com/hashicorp/terraform-provider-aws/issues/26626
* Add VCR flags to the custom context
[...]

PR #43070, opened 2025-06-18
internal/retry: replace SDK V2 `WaitForState` helper

jar-b:
## Rollback Plan

### Description

This change begins to port functionality from the Terraform Plugin SDK V2 `helper/retry` package to the provider's `internal/retry` package. Specifically, various error types are copied and the state change logic is replicated and modified to support generic response types and status codes should we choose to refactor existing implementations which currently use type assertions and string conversion of status enums to shim into the Plugin SDK V2 implementation.

The ultimate goal is to drop the dependency on Plugin SDK V2's `helper/retry` package which no longer receives new features and achieve more granular control over retry behavior which can be tuned to the specific needs of the AWS provider. Long term these changes will support refactoring retry behavior (#42554) and enabling `go-vcr` support (#25602).

This port was authored by @ewbankkit, with minor adjustments by myself while re-using the work for `go-vcr` proof-of-concepts.

### Relations

Relates #42554
Relates #25602
Relates #43039

### DIFF
--- a/internal/backoff/backoff.go
+++ b/internal/backoff/backoff.go
@@ -145,7 +145,6 @@ type Loop struct {
 	config      LoopConfig
 	deadline    inttypes.Deadline
 	gracePeriod time.Duration
-	timedOut    bool
 }
 
 // NewLoopWithOptions returns a new loop configured with the provided options.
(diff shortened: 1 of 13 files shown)


ewbankkit (approved):
LGTM :rocket:.

PR #43089, opened 2025-06-19
Resource Identity: Don't set Identity when removing resource

gdavison:
## Rollback Plan

### Description

When removing a resource during Read because it is not found, the Identity must not be set for Framework-based resource types. Skips setting Identity for both SDKv2- and Framework-based types.


ewbankkit (approved):
LGTM :rocket:.

PR #43156, opened 2025-06-24
add note about what to do with an existing resource

lpmi-13:
### Description

I was just dealing with a plan that showed creation of an existing resource here, and I thought I would have to import it. However, there is no mention of the syntax for import on the [existing docs page](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate_validation). This makes sense, since you don't need to import it. A `terraform apply` is enough to update the existing resource.

This commit adds that explicit note for future travellers.

### References

[docs page that led to this PR](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate_validation)

### DIFF
--- a/website/docs/r/acm_certificate_validation.html.markdown
+++ b/website/docs/r/acm_certificate_validation.html.markdown
@@ -144,3 +144,8 @@ This resource exports the following attributes in addition to the arguments abov
 [Configuration options](https://developer.hashicorp.com/terraform/language/resources/syntax#operation-timeouts):
 
 - `create` - (Default `75m`)
+
+
+## Import
+
+You cannot import the ACM Certificate Validation resource. Running an apply will update the existing one, even if the plan shows a resource to add.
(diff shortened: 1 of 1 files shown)


justinretzolk (commented):
I'm not sure that I agree with this change. Line 18 has the following call out, indicating that this resource doesn't correlate to a "real" resource within AWS:

> ~> **WARNING:** This resource implements a part of the validation workflow. It does not represent a real-world entity in AWS, therefore changing or deleting this resource on its own has no immediate effect.

The wording that you've added implies that an update is being made, which I think could add additional confusion. We generally omit the "Import" section of the documentation entirely when a resource isn't able to be imported, so this wouldn't fit that convention.

Overall, I can definitely see where you were coming from with this suggestion, but I don't think it's something we should add given the existing warning context and the convention of omission of the "Import" section, so I'm going to close this pull request. If you see anything else in the documentation that seems off, please do continue to raise pull requests! We always appreciate the chance to review the documentation and make sure that it's accurate and helpful.

PR #43230, opened 2025-06-30
Update index page to show new provider version 6

jensvog:
Updates the provider version to 6 on the main website readme file.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -29,7 +29,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5.0"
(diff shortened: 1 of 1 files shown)


ewbankkit:
@jensvog Thanks for the contribution :tada: :clap:.

HasseJohansen:
@ewbankkit I think there is a problem with this change. I think we will now try to pull the 6.1.0 which is removed because of a bug. So when you bumped from 5.0 to 6.0 it created problems for people because the latest 6.0.x is still 6.0.10 which does not exist. Some where you need to change 6.1.0. to 6.0.0 (or revert back to 5.0 I think without knowing too much about how your stuff works)

PR #43284, opened 2025-07-06
Add possible `ARN` value for `function_name` in `aws_lambda_permission` resource

l-teles:
### Description

This PR improves the documentation for the `aws_lambda_permission` resource by specifying that the function ARN is also a possible value for the `function_name` argument, as seen [here](https://github.com/hashicorp/terraform-provider-aws/blob/229b588b9676522d110ebdf0accffa821aef6b51/website/docs/r/config_organization_custom_rule.html.markdown?plain=1#L22) for example.


justinretzolk (approved):
Looks good to me! Thank you for taking the time to raise this! I'll get this merged once the checks finish up.

[Resource definition reference](https://github.com/hashicorp/terraform-provider-aws/blob/9afe5b341c36f1324fe2b20411ea91d9a9ba30ba/internal/service/lambda/validate.go#L25)

PR #43298, opened 2025-07-07
Simplify Framework registration

gdavison:
## Rollback Plan

### Description

Uses `sync.Once` instead of mutex and boolean to ensure schema validation is called once. 

Simplifies Framework Data Source registration:

* Defers creating `inner` wrapped resource until wrapper is called
* Makes `boostrapContext` a method on wrappers instead of function pointer
* Encapsulates configuration of interceptors

### DIFF
--- a/internal/provider/framework/provider.go
+++ b/internal/provider/framework/provider.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"iter"
 	"log"
+	"reflect"
 	"slices"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


internal/provider/framework/provider.go
```diff
@@ -35,7 +36,7 @@ import (
 )
 
 var (
-	resourceSchemasValidated bool
+	resourceSchemasValidated sync.Once
```
ewbankkit, line 40: As this returns an `error` should we prefer [`sync.OnceValue`](https://pkg.go.dev/sync#OnceValue)?

internal/provider/framework/wrap.go
```diff
@@ -26,34 +31,83 @@ type getAttributeFunc func(context.Context, path.Path, any) diag.Diagnostics
...
-		inner: inner,
-		opts:  opts,
+		inner:              inner,
+		servicePackageName: servicePackageName,
+		spec:               spec,
+		interceptors:       interceptors,
 	}
 }
 
+// bootstrapContext is run on all wrapped methods before any interceptors.
```
ewbankkit: Maybe just `context` (or `Context`) is now a better name for this method?

internal/provider/framework/provider.go
gdavison: Yeah, I thought about that. I see `sync.OnceValue` as more "return a useful value", but since the `error` manages the control flow, that counts as a "useful value"

PR #43307, opened 2025-07-08
d-apigatewayv2-stage-throttling-defaults: Clarify throttling default values

ysfiqbl:
## Description
Clarifies that `throttling_burst_limit` and `throttling_rate_limit` default to `0` in the `default_route_settings` block, and adds a note warning that enabling `detailed_metrics_enabled` without setting explicit throttling values will cause 429 errors.

## Reasoning for Change
The current documentation for `aws_apigatewayv2_stage` does not specify the default values for `throttling_burst_limit` and `throttling_rate_limit` parameters. This omission causes confusion when users enable `detailed_metrics_enabled` without explicitly setting throttling values, resulting in unexpected 429 throttling errors due to the default values being `0`.

This change aligns the documentation with AWS API Gateway behavior where unset throttling limits default to `0` (no throttling allowed).

## AWS Documentation Reference
[AWS API Gateway v2 Throttling Documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-throttling.html)

## Changes
- Added default values (`0`) for `throttling_burst_limit` and `throttling_rate_limit` parameters
- Added warning note explaining the 429 error scenario when `detailed_metrics_enabled` is true without explicit throttling configuration
- Applied changes to `default_route_settings` only.


website/docs/r/apigatewayv2_stage.html.markdown
```diff
@@ -56,8 +56,10 @@ Defaults to `false`. Supported only for WebSocket APIs.
 * `detailed_metrics_enabled` - (Optional) Whether detailed metrics are enabled for the default route. Defaults to `false`.
 * `logging_level` - (Optional) Logging level for the default route. Affects the log entries pushed to Amazon CloudWatch Logs.
 Valid values: `ERROR`, `INFO`, `OFF`. Defaults to `OFF`. Supported only for WebSocket APIs. Terraform will only perform drift detection of its value when present in a configuration.
-* `throttling_burst_limit` - (Optional) Throttling burst limit for the default route.
-* `throttling_rate_limit` - (Optional) Throttling rate limit for the default route.
+* `throttling_burst_limit` - (Optional) Throttling burst limit for the default route. Defaults to `0`.
```
justinretzolk, line 59: I'm not sure that I agree with this change. Typically in our resource documentation, we only mention a default value if the default value is coming from the resource schema. In this case, [a default value is not specified](https://github.com/hashicorp/terraform-provider-aws/blob/c6cef71fa1c4cd785a9aae33c448bbcf4273a8e6/internal/service/apigatewayv2/stage.go#L108-L111) in the resource definition.

website/docs/r/apigatewayv2_stage.html.markdown
```diff
@@ -56,8 +56,10 @@ Defaults to `false`. Supported only for WebSocket APIs.
...
+* `throttling_rate_limit` - (Optional) Throttling rate limit for the default route. Defaults to `0`.
```
justinretzolk, line 60: Same thought as above -- [the schema doesn't define a default here](https://github.com/hashicorp/terraform-provider-aws/blob/c6cef71fa1c4cd785a9aae33c448bbcf4273a8e6/internal/service/apigatewayv2/stage.go#L112-L116). 

Edit: this was supposed to be a comment on line 60

website/docs/r/apigatewayv2_stage.html.markdown
```diff
@@ -56,8 +56,10 @@ Defaults to `false`. Supported only for WebSocket APIs.
...
+
+-> **Note:** When `detailed_metrics_enabled` is set to `true` without explicitly setting throttling values, both limits default to `0`, which will cause all requests to be throttled with 429 errors.
```
justinretzolk, line 62: I generally like this note, but I'm curious as to whether we should consider that this presumably could also happen for `route_settings.throttling_[burst|rate]_limit`. It feels a bit excessive to have the same warning callout in both places, but maybe it's worthwhile. I'd be interested in your thoughts, as someone who seems to have needed the information for yourself.

Edit: This was supposed to be a comment on line 62 🤦

schilling-infrm:
@justinretzolk 
We also just encountered this issue, but the other way around. A member manually enabled route metrics and a `terraform apply` from the pipeline removed the metrics. 
But it also set the rates to `0/0`. Additionally the terraform plan output did also not display the changes made to the setting:
```text
resource "aws_apigatewayv2_stage" "api_stage" {
    id   = "redacted"
    [...]
    # (11 unchanged attributes hidden)
  ~ default_route_settings {
      ~ detailed_metrics_enabled = true -> false
    }
}
```
I think displaying the change to the rates should be possible already on the `terraform plan`

PR #43345, opened 2025-07-10
Add support for `aws_glue_crawler` to retry updates when the crawler is running

RicePatrick:
### Description

AWS Doesn't allow updates made to running crawlers, which results in terraform immediately throwing an error when attempting to make an update to a running crawler. Users of the resource simply need to wait a bit for their run to finish, then retry their update with the exact same configuration, which is a poor user experience and can cause frustration with the resource.

This PR adds the `Cannot update Crawler while running` error to the list of retryable errors that are currently checked when performing an update to the crawler, which will allow the resource to retry the operation after receiving that error.

A new test is included that creates a crawler with an s3 target, then _runs_ that crawler before updating the name to validate the error handling retries properly.

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library without the included error handling

N/A

TODO

```console

...
```

### DIFF
--- a/internal/service/glue/crawler.go
+++ b/internal/service/glue/crawler.go
@@ -612,6 +612,11 @@ func resourceCrawlerUpdate(ctx context.Context, d *schema.ResourceData, meta any
 					return retry.RetryableError(err)
 				}
 
+				// Glue will allow updating crawlers when a crawler is running. Retry occassionally to persist the updates once a crawler has finished the run.
+				if errs.IsAErrorMessageContains[*awstypes.CrawlerRunningException](err, "Cannot update Crawler while running") {
+					return retry.RetryableError(err)
+				}
+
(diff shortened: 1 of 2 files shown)


RicePatrick:
Note - this PR also adds support for a configurable timeout on the Update operation, and replaces the `propagationTimeout` with that value. This is for multiple reasons:

- Test stability, since a crawler run takes ~1 minute to finish minimum, and then between 45-75 seconds to  "stop", with the default 2 minute timeout, the test suffered from race conditions
- User experience, since different Glue jobs may take a different amount of time to complete, this allows users to configure how long they would want to wait.

RicePatrick:
Proof of test passing:

image

RicePatrick:
@justinretzolk - apologies for the ping, but would you be able to see if someone has a chance to take a look at this PR for potential merger?

PR #43374, opened 2025-07-14
Remove duplicated words from the aws_bedrock_inference_profiles documentation

Collinux:
Removed two words from the documentation which says the string  "AWS Bedrock" twice back-to-back.


jar-b:
Thanks for the correction, @Collinux! 👍

PR #43379, opened 2025-07-15
[Documentation] Amplify: Proper link to build instance type compute

oycyc:
The current link that is associated with job_config is wrong. It doesn't contain any information about the job config for setting the build compute. The updated link is the actual link that contains the info. 
image


ewbankkit:
@oycyc Thanks for the contribution :tada: :clap:.

Issue #43406, opened 2025-07-16
aws_bedrockagent_agent_alias produces inconsistent result: routing_configuration.agent_version    changes from null to string after apply


cmmo:
### Terraform and AWS Provider Version

```console
Terraform Version
  1.11.4

AWS Provider Version
  ~> 6.3
```

### Affected Resource(s) or Data Source(s)

  - aws_bedrockagent_agent_alias

### Expected Behavior

  The provider should not change the agent_version value from null to a string after apply without being
  explicitly configured.

### Actual Behavior

  The provider changes routing_configuration[0].agent_version from null to a string value (e.g., "37", "24")
  after apply, causing inconsistent state.

### Relevant Error/Panic Output

```console
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.aifp_financial_bedrock_agent.aws_bedrockagent_agent_alias.this[0], provider
│ "provider[\"registry.terraform.io/hashicorp/aws\"]" produced an unexpected new value: .routing_configuration[0].agent_version: was null, but now
│ cty.StringVal("37").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.aifp_general_bedrock_agent.aws_bedrockagent_agent_alias.this[0], provider "provider[\"registry.terraform.io/hashicorp/aws\"]"
│ produced an unexpected new value: .routing_configuration[0].agent_version: was null, but now cty.StringVal("24").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

```

### Sample Terraform Configuration

<details>
<summary>Click to expand configuration</summary>

```hcl
  resource "aws_bedrockagent_agent_alias" "this" {
    agent_alias_name = "example-alias"
    agent_id         = "AGENT123"
    description      = "Example agent alias"
    
    routing_configuration {
      agent_version = null  # or not specified
    }
  }
```
</details>

### Steps to Reproduce

  1. Create aws_bedrockagent_agent_alias with routing_configuration.agent_version as null
  2. Run terraform apply
  3. Observer the inconsistent result error

### Debug Logging

<details>
<summary>Click to expand log output</summary>

```console
  Error: Provider produced inconsistent result after apply

  When applying changes to module.aifp_financial_bedrock_agent.aws_bedrockagent_agent_alias.this[0], provider
[...]

PR #43411, opened 2025-07-16
Fix inconsistent result error in aws_bedrockagent_agent_alias routing_configuration

cmmo:
### Description
Fixes issue where aws_bedrockagent_agent_alias produces inconsistent result when routing_configuration.agent_version changes from null to string after apply.

### Problem
When routing_configuration is not specified or agent_version is explicitly set to null, AWS automatically assigns a default agent_version (like "37" or "24"). The provider was reading this value back into state, causing inconsistency with the user's configuration.

### Solution
- Preserve the original plan's null state for routing_configuration
- Maintain consistency between plan and state when values are not explicitly configured
- Enhanced handling in Create, Read, and Update methods

### Testing
- Added TestAccBedrockAgentAgentAlias_routingConfigurationNull
- Added TestAccBedrockAgentAgentAlias_routingConfigurationExplicitNull
- Verified existing tests continue to pass

### Relations
Closes https://github.com/hashicorp/terraform-provider-aws/issues/43406.

```console
% make testacc TESTS=TestAccBedrockAgentAgentAlias_routingConfiguration PKG=bedrockagent
# Tests would be run here by maintainers
```

### DIFF
--- a/internal/service/bedrockagent/agent_alias.go
+++ b/internal/service/bedrockagent/agent_alias.go
@@ -30,7 +30,6 @@ import (
 	"github.com/hashicorp/terraform-provider-aws/internal/flex"
 	"github.com/hashicorp/terraform-provider-aws/internal/framework"
 	fwflex "github.com/hashicorp/terraform-provider-aws/internal/framework/flex"
-	fwtypes "github.com/hashicorp/terraform-provider-aws/internal/framework/types"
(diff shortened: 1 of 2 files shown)


drewtul:
Unfortunately we can't validate the `agent_version` inside `routing_configuration`, `null` is not a valid value, technically it should be optional and computed.

Returning `null`  to the user I don't think is correct either, we should reject the `null` and return the computed `agent_version` as it is never `null`.

PR #43121, opened 2025-06-19
smarterr CloudWatch Integration: Declarative, Config-Driven Error Handling

YakDriver:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

This PR begins integration of [smarterr](https://github.com/YakDriver/smarterr) into the Terraform AWS Provider, starting with CloudWatch, establishing a new foundation for standardized, maintainable, and user-focused error diagnostics.

### Why smarterr?

Our current error handling is spread across \~15,000 call sites, many of which manually format error messages. This leads to:

* Repetitive and inconsistent messaging
* Walls of duplicated diagnostic text
* Difficulties evolving how we talk to users

With `smarterr`, we move toward a **config-driven model** where error formatting—summary, detail, and logs—is handled centrally and declaratively.

### PR To Do

- [x] semgrep (or other) check to ensure a migrated service has all callsites using smarterr
- [x] add CI to validate configs from every service directory (this will just be testing the global config over and over again early on)

### What this enables

* 💬 **Improved user experience** with clean, friendly, and informative error messages—plus optional, per-error suggestions—all powered by config
* 🔄 **Centralized updates** to formatting and messaging via HCL config (not cross-codebase rewrites)
* 🛠️ **Better developer experience** with simple, low-friction APIs
* 🧪 **Validation and previewing of error output** via CLI during development

### User experience

This is a template-driven error formatted by smarterr. And we can change it as needs change by just updating declarative config.

```
        Error: creating CloudWatch Composite Alarm
        
          with aws_cloudwatch_composite_alarm.test,
          on terraform_plugin_test.tf line 29, in resource "aws_cloudwatch_composite_alarm" "test":
          29: resource "aws_cloudwatch_composite_alarm" "test" {
        
        ID: r-1234567890
        Underlying issue: operation error CloudWatch: ModifyServerlessCache
        If you are trying to modify a serverless cache, please use the
        `aws_cloudwatch_serverless_cache` resource instead of
        `aws_cloudwatch_log_group`.
```

### Integration Pattern

The integration pattern uses the `smerr` package (a thin wrapper around smarterr) for idiomatic use in this codebase. Example SDK-style usage:

```go
return smerr.Append(ctx, diags, err, smerr.ID, d.Id())
```

And wrapped errors in helpers:

```go
return nil, smarterr.NewError(err)
```

### What’s in this PR

[...]

PR #43524, opened 2025-07-23
Smarterr Migration: Datazone

ThomasZalewski:
## Rollback Plan

### Description

Added a tool for simplifying smarterr conversion workflows along with a smarterr migration for datazone.

### Relations

Relates #43121 

```console
terraform-provider-aws % make build
make: Building provider...
```


tools/smarterr-parser/smarterr-parser
jar-b, line 1: Can we remove the binary from version control?

tools/smarterr-parser/main.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Fprintf(os.Stderr, "Usage: %s <folder_path> [--fix]\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	folderPath := os.Args[1]
+	autoFix := len(os.Args) > 2 && os.Args[2] == "--fix"
```
jar-b: Could this be a `-fix` flag instead of parsing arguments for a string literal?

tools/smarterr-parser/main.go
```diff
@@ -0,0 +1,299 @@
...
+		"create.AddError":                           regexp.MustCompile(`create\.AddError\s*\(`),
+		"bare_return_nil_err":                       regexp.MustCompile(`^\s*return\s+nil\s*,\s*err\s*$`),
+		"return_retry_NotFoundError":                regexp.MustCompile(`return\s+nil\s*,\s*&retry\.NotFoundError\s*{`),
+		"return_tfresource_NewEmptyResultError":     regexp.MustCompile(`return\s+nil\s*,\s*tfresource\.NewEmptyResultError\s*\(`),
+		"return_tfresource_AssertSingleValueResult": regexp.MustCompile(`return\s+tfresource\.AssertSingleValueResult\s*\(`),
+		"resp.Diagnostics.Append":                   regexp.MustCompile(`resp\.Diagnostics\.Append\s*\(`),
+		"response.Diagnostics.Append":               regexp.MustCompile(`response\.Diagnostics\.Append\s*\(`),
+		"diags.AddError":                            regexp.MustCompile(`diags\.AddError\s*\(`),
+		"diagnostics.AddError":                      regexp.MustCompile(`diagnostics\.AddError\s*\(`),
+	}
```
jar-b: This could be pulled into a global var so it only compiles once.

tools/smarterr-parser/main.go
```diff
@@ -0,0 +1,299 @@
...
+
+		if isResourceOrDataSource(line) {
+			result.IsResource = true
+			break
+		}
+	}
+
+	if !result.IsResource {
+		return result, nil
+	}
```
jar-b, line 230: Are there cases where we would want to use `smarterr` in a non-resource/data source file? One example might be logic shared between multiple resources in a service. 

Is the assumption that the upstream call sites in each resource will wrap the error instead?

tools/smarterr-parser/main.go
```diff
@@ -0,0 +1,299 @@
...
+			lines = append(lines[:i+1], append([]string{
+				"\t" + existingImport,
+				"\t\"github.com/YakDriver/smarterr\"",
+				")",
+			}, lines[i+1:]...)...)
+			return lines
+		}
+	}
+	return lines
+}
```
jar-b, line 388: Could the `make fix-imports` target be paired with this tool to handle cleaning up imports?

tools/smarterr-parser/main.go
```diff
@@ -0,0 +1,299 @@
+// Copyright (c) HashiCorp, Inc.
```
jar-b, line 1: Perhaps include a make target to build and install this (similar to `make skaff`)?

internal/service/datazone/asset_type.go
```diff
@@ -149,18 +150,12 @@ func (r *assetTypeResource) Create(ctx context.Context, req resource.CreateReque
 
 	out, err := conn.CreateAssetType(ctx, &in)
 	if err != nil {
-		resp.Diagnostics.AddError(
-			create.ProblemStandardMessage(names.DataZone, create.ErrActionCreating, ResNameAssetType, plan.Name.String(), err),
-			err.Error(),
-		)
+		smerr.AddError(ctx, &resp.Diagnostics, err, smerr.ID, plan.Name.String())
```
jar-b: Should the key of the key/value pair be `name` here? This might apply to all cases when the identifier has not been returned from the API yet.

```suggestion
		smerr.AddError(ctx, &resp.Diagnostics, err, "name" plan.Name.String())
```

tools/smarterr-parser/main.go
ThomasZalewski: Added it so you can manually choose files for scanning but I will double check this tommorow

PR #43555, opened 2025-07-25
r/aws_quicksight_data_set: add `output_columns` output description

posquit0:
## Rollback Plan

### Description

Updated `r/aws_quicksight_data_set` docs to include missing `output_columns` output

### DIFF
--- a/website/docs/r/quicksight_data_set.html.markdown
+++ b/website/docs/r/quicksight_data_set.html.markdown
@@ -395,6 +395,7 @@ This resource exports the following attributes in addition to the arguments abov
 
 * `arn` - Amazon Resource Name (ARN) of the data set.
 * `id` - A comma-delimited string joining AWS account ID and data set ID.
+* `output_columns` - The final set of columns available for use in analyses and dashboards after all data preparation and transformation steps have been applied within the data set.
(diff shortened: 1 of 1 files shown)


website/docs/r/quicksight_data_set.html.markdown
```diff
@@ -395,6 +395,7 @@ This resource exports the following attributes in addition to the arguments abov
 
 * `arn` - Amazon Resource Name (ARN) of the data set.
 * `id` - A comma-delimited string joining AWS account ID and data set ID.
+* `output_columns` - The final set of columns available for use in analyses and dashboards after all data preparation and transformation steps have been applied within the data set.
```
justinretzolk: [It looks like](https://github.com/hashicorp/terraform-provider-aws/blob/a9e20d8d832355eb3faeb8bf2b5f132612a8de8c/internal/service/quicksight/schema/data_set.go#L358-L379) this is a block that has additional attributes within in. Since we're adding the `output_columns`, I think it's probably worthwhile to add `output_columns.description`, `output_columns.name`, and `output_columns.type` as well (similar to how we do on, say, lines 375 and 379 above).

posquit0: I agree! 👍🏼 
I've taken your feedback and updated.

posquit0:
@justinretzolk can you check the latest changes?

PR #43561, opened 2025-07-25
New resource: aws_wafv2_web_acl_rule_group_association

YakDriver:
- **Add docs for new resource**
- **Add new resource**
- **Add tests for new resource**
- **Updates for new resource**

## Rollback Plan

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/internal/service/wafv2/web_acl_rule_group_association.go
+++ b/internal/service/wafv2/web_acl_rule_group_association.go
@@ -1,0 +1,529 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package wafv2
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/wafv2"
+	awstypes "github.com/aws/aws-sdk-go-v2/service/wafv2/types"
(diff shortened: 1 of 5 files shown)


GNUmakefile
```diff
@@ -383,6 +383,9 @@ provider-lint: ## [CI] ProviderLint Checks / providerlint
 		-XS002=false \
 		$(SVC_DIR)/... ./internal/provider/...
 
+quick-fix: fmt testacc-lint-fix fix-imports modern-fix semgrep-fix ## Some quick fixes
```
ewbankkit, line 386: 🙌

internal/service/wafv2/web_acl_rule_group_association.go
```diff
@@ -0,0 +1,1317 @@
...
+		ruleGroupIdentifier,
+	}, webACLRuleGroupAssociationResourceIDPartCount, false)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error creating resource ID",
+			fmt.Sprintf("Could not create resource ID: %s", err),
+		)
+		return
+	}
+	plan.ID = types.StringValue(id)
```
jar-b: Is a composite `id` attribute necessary here? Our standard recommendation for net-new Framework resources is to omit one when the identifier is a combination of arguments, but I can't quite tell from context if its necessary here to store the rule group type. If that's all it is for, perhaps we can just make that it's own computed attribute and drop `id`?

https://hashicorp.github.io/terraform-provider-aws/id-attributes/#standard-usage

PR #43632, opened 2025-07-31
(feat) Create aws_route53_resolver_firewall_domain_lists datasource

twelsh-aw:
## Rollback Plan

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/internal/service/route53resolver/firewall_domain_lists_data_source.go
+++ b/internal/service/route53resolver/firewall_domain_lists_data_source.go
@@ -1,0 +1,119 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package route53resolver
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/route53resolver"
+	awstypes "github.com/aws/aws-sdk-go-v2/service/route53resolver/types"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-aws/internal/conns"
+	"github.com/hashicorp/terraform-provider-aws/internal/errs/sdkdiag"
+)
+
+// @SDKDataSource("aws_route53_resolver_firewall_domain_lists", name="Firewall Domain Lists")
+func dataSourceFirewallDomainLists() *schema.Resource {
+	return &schema.Resource{
+		ReadWithoutTimeout: dataSourceFirewallDomainListsRead,
+
+		Schema: map[string]*schema.Schema{
+			"firewall_domain_lists": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"arn": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"creator_request_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"managed_owner_name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceFirewallDomainListsRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	var diags diag.Diagnostics
+	conn := meta.(*conns.AWSClient).Route53ResolverClient(ctx)
+
+	input := &route53resolver.ListFirewallDomainListsInput{}
+	var domainLists []awstypes.FirewallDomainListMetadata
+
+	pages := route53resolver.NewListFirewallDomainListsPaginator(conn, input)
+	for pages.HasMorePages() {
+		page, err := pages.NextPage(ctx)
+
+		if err != nil {
+			return sdkdiag.AppendErrorf(diags, "listing Route53 Resolver Firewall Domain Lists: %s", err)
+		}
+
+		for _, domainList := range page.FirewallDomainLists {
+			domainLists = append(domainLists, domainList)
+		}
+	}
+
+	d.SetId(meta.(*conns.AWSClient).Region(ctx))
+
+	if err := d.Set("firewall_domain_lists", flattenFirewallDomainListsMetadata(domainLists)); err != nil {
+		return sdkdiag.AppendErrorf(diags, "setting firewall_domain_lists: %s", err)
+	}
+
+	return diags
+}
+
+func flattenFirewallDomainListsMetadata(apiObjects []awstypes.FirewallDomainListMetadata) []any {
+	if len(apiObjects) == 0 {
+		return nil
+	}
+
+	var tfList []any
+
+	for _, apiObject := range apiObjects {
+		tfList = append(tfList, flattenFirewallDomainListMetadata(apiObject))
+	}
+
+	return tfList
+}
+
+func flattenFirewallDomainListMetadata(apiObject awstypes.FirewallDomainListMetadata) map[string]any {
+	tfMap := map[string]any{}
+
+	if apiObject.Arn != nil {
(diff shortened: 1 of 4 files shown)


internal/service/route53resolver/firewall_domain_lists_data_source.go
```diff
@@ -0,0 +1,118 @@
+// Copyright IBM Corp. 2014, 2025
```
cchristous, line 1: The copyright header here reads `Copyright IBM Corp. 2014, 2025` — this should be `Copyright (c) HashiCorp, Inc.` to match every other file in this repository.

```suggestion
// Copyright (c) HashiCorp, Inc.
```

internal/service/route53resolver/firewall_domain_lists_data_source_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright IBM Corp. 2014, 2025
```
cchristous, line 1: Same copyright header issue here — should be `Copyright (c) HashiCorp, Inc.`

```suggestion
// Copyright (c) HashiCorp, Inc.
```

.changelog/43632.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:new-data-source
+aws_route53_resolver_firewall_domain_lists
+```
```
cchristous, line 3: Nit: missing trailing newline at end of file.

internal/service/route53resolver/firewall_domain_lists_data_source_test.go
```diff
@@ -0,0 +1,56 @@
...
+					acctest.CheckResourceAttrGreaterThanValue(dsResourceName, "firewall_domain_lists.#", 0),
+					// Check that our created resource appears in the list
+					resource.TestCheckTypeSetElemNestedAttrs(dsResourceName, "firewall_domain_lists.*", map[string]string{
+						names.AttrName: rName,
+					}),
+					// Check that AWS managed malware domain list appears in the list
+					resource.TestCheckTypeSetElemNestedAttrs(dsResourceName, "firewall_domain_lists.*", map[string]string{
+						names.AttrName:       "AWSManagedDomainsMalwareDomainList",
+						"managed_owner_name": "Route 53 Resolver DNS Firewall",
+					}),
```
cchristous, line 39: This assertion depends on `AWSManagedDomainsMalwareDomainList` existing in every test account/region. If AWS ever renames, removes, or doesn't provide this managed list in a particular region or partition, this test will fail for reasons unrelated to the code. Other list-type data source tests in this package (e.g., `rules_data_source_test.go`) don't assert on specific AWS-managed resources.

Consider removing this assertion from the basic test, or moving it to a separate test function (e.g., `TestAccRoute53ResolverFirewallDomainListsDataSource_managed`) so a failure here doesn't block the core functionality test.

internal/service/route53resolver/firewall_domain_lists_data_source_test.go
```diff
@@ -0,0 +1,56 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccFirewallDomainListsDataSourceConfig_basic(rName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(dsResourceName, "firewall_domain_lists.#"),
+					acctest.CheckResourceAttrGreaterThanValue(dsResourceName, "firewall_domain_lists.#", 0),
+					// Check that our created resource appears in the list
+					resource.TestCheckTypeSetElemNestedAttrs(dsResourceName, "firewall_domain_lists.*", map[string]string{
+						names.AttrName: rName,
+					}),
```
cchristous, line 34: Two things here:

1. **`TestCheckTypeSetElemNestedAttrs` on a `TypeList`**: The schema defines `firewall_domain_lists` as `schema.TypeList`, but `TestCheckTypeSetElemNestedAttrs` is designed for `TypeSet`. It works in practice (iterates all indices), but is semantically misleading. Since API ordering isn't guaranteed, this is pragmatic — but worth a comment explaining the choice, or consider changing the schema to `TypeSet`.

2. **Missing attribute assertions**: This only checks `name`. The flatten function sets `arn`, `id`, `creator_request_id`, and `managed_owner_name` — a bug in any of those branches would go undetected. Consider also verifying `arn` and `id` for the created resource, e.g. via `TestCheckResourceAttrPair` against `aws_route53_resolver_firewall_domain_list.test`.

website/docs/d/route53_resolver_firewall_domain_lists.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+    list if list.name == "AWSManagedDomainsMalwareDomainList"
+  ])
+}
+
+resource "aws_route53_resolver_firewall_rule" "example" {
+  name                    = "block-malware-domains"
+  action                  = "BLOCK"
+  block_response          = "NXDOMAIN"
+  firewall_domain_list_id = local.malware_list.id
+  firewall_rule_group_id  = aws_route53_resolver_firewall_rule_group.example.id
```
cchristous, line 33: The example references `aws_route53_resolver_firewall_rule_group.example.id` but never defines that resource, so the example isn't self-contained. Consider either adding the missing `aws_route53_resolver_firewall_rule_group` resource block, or simplifying the example to only show the data source + locals filtering (removing the `aws_route53_resolver_firewall_rule` block).

website/docs/d/route53_resolver_firewall_domain_lists.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+* `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
+
+## Attribute Reference
+
+This data source exports the following attributes in addition to the arguments above:
+
+* `id` - AWS Region.
+* `firewall_domain_lists` - List of firewall domain lists with detailed information. Each element contains:
+    * `arn` - The Amazon Resource Name (ARN) of the firewall domain list.
+    * `creator_request_id` - A unique string defined by you to identify the request.
```
cchristous, line 56: Nit: "A unique string defined by you to identify the request" is slightly misleading for a read-only data source that also returns AWS-managed lists (where the `creator_request_id` was not defined by the user). Consider rephrasing to something like: "A unique string that identifies the request that created the firewall domain list."

website/docs/d/route53_resolver_firewall_domain_lists.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+    * `id` - The ID of the firewall domain list.
+    * `managed_owner_name` - The owner of the firewall domain list. For AWS managed domain lists, this is `Route 53 Resolver DNS Firewall`.
```
cchristous, line 58: Nit: It would be helpful to note that this field is empty/null for user-created domain lists. The singular data source docs use: "The owner of the list, used only for lists that are not managed by you." Consider something like: "The owner of the firewall domain list, if managed. For AWS managed domain lists, this is `Route 53 Resolver DNS Firewall`. Empty for user-created lists."

website/docs/d/route53_resolver_firewall_domain_lists.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+output "malware_list_arn" {
+  value = local.malware_list.arn
+}
+```
+
+## Argument Reference
+
+This data source supports the following arguments:
+
+* `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
```
cchristous, line 47: Nit: Since this data source has no required or optional arguments, listing only `region` (a provider-level attribute) as an argument is a bit unusual. The singular data source (`route53_resolver_firewall_domain_list`) doesn't list `region`. Consider either stating "This data source does not require any arguments" or checking the project's convention for zero-argument data sources.

cchristous (commented):
Thanks for adding this data source! Left some review comments below.

cchristous:
Thanks for implementing this — I'd really like to see it merged\! Left some inline comments, mostly minor. The main thing to fix is the copyright header; the rest are suggestions to strengthen the tests and docs.

PR #43807, opened 2025-08-11
Fix SageMaker user profile regex

gregops312:
## Rollback Plan

n/a

### Description

This is a minor change to the regex matching pattern for SageMaker user profiles, as it had not caught many failure cases until an apply

### Relations
 
n/a

### References

I'm not entirely sure how to make sure this gets properly tested. The testing cases are a little over my head.

I have not yet run any acceptance testing, but will update when I have, just wanted to get the PR open

```console

...
```


jar-b:
Thanks for your contribution, @gregops312 ! 👍

PR #43842, opened 2025-08-12
Fixes returned diags from SDKv2 interceptors

gdavison:
## Rollback Plan

### Description

Previously:

* Any `Warning` diags returned by `Before` interceptors were suppressed if all `Before` interceptors succeded
* Some diags returned by interceptors where returned out of order

### Relations

Closes #0000

```console

...
```


internal/provider/sdkv2/intercept.go
```diff
@@ -57,28 +58,18 @@ type interceptor1[D, E any] interface {
 	run(context.Context, interceptorOptions[D]) E
```
ewbankkit, line 58: Maybe rename this to just `interceptor` as there are no longer `1` and `2` variants?

Issue #42988, opened 2025-06-12
Resource Identity - Parameterized Phase One (Core Resources)


breathingdust:
### Description

If the Primary API Identifier is not the remote object’s ARN, and it consists of one or more identifying parameters, the resource type can use the Parameterized Identity pattern. The Identity object will contain multiple attributes which are all used to identify the remote object. This is the class that will feature the majority of resources. 

In order to make sure we prioritize the implementation of this effectively, we will focus on the most popular resources in the provider first.

Issue #42983, opened 2025-06-12
[Meta] Resource Identity Adoption


breathingdust:
### Description

This issue captures the overall effort in bringing Resource Identity to all possible resources in the provider. 

### Tasklist

- [ ] Resource Identity - Provider Support
  - [x] #42487
  - [x] #42549
  - [ ] #42527
- [ ] https://github.com/hashicorp/terraform-provider-awscc/issues/2339
- [ ] #42984
- [ ] #42986
- [ ] #42987
- [ ] #42988
- [ ] #42989
- [x] #42990
- [ ] https://github.com/YakDriver/tfproviderdocs/issues/187

### References

- https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/identity
- https://developer.hashicorp.com/terraform/plugin/framework/resources/identity

PR #43843, opened 2025-08-12
Add parameterized resource identity to `aws_route_table_association`

jar-b:
### Description

Adds paramerterized resource identity to the `aws_route_table_association` resource. Because the `id` can be changed on update, this resource is marked as a mutable identity. This behavior may change in a future major version.

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/issues/42983
Relates https://github.com/hashicorp/terraform-provider-aws/issues/42988
Relates #43844

```console
% make testacc PKG=ec2 TESTS=TestAccVPCRouteTableAssociation_

```


jar-b:
This resource identity implementation is currently blocked. The mismatch between the resource identifier attribute (association ID) and expected import attributes (`route_table_id` and one of `subnet_id` or `gateway_id`) prevents the existing import logic from being ported to a custom importer. 

Additional support must be added from "at most one of" optional arguments within the import schema before proceeding. Alternatively, we could make a breaking change to the `id` argument and import attributes such that they both become a combination of `association_id` and `route_table_id`, the two required arguments for the read operation.

PR #43865, opened 2025-08-13
Update api_gateway_gateway_response response_type argument reference with a link to the documentation of response types

digitalfiz:
This change simply adds a link to the documentation for the response_types API Gateway exposes.

### DIFF
--- a/website/docs/r/api_gateway_gateway_response.html.markdown
+++ b/website/docs/r/api_gateway_gateway_response.html.markdown
@@ -38,7 +38,7 @@ This resource supports the following arguments:
 
 * `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
 * `rest_api_id` - (Required) String identifier of the associated REST API.
-* `response_type` - (Required) Response type of the associated GatewayResponse.
+* `response_type` - (Required) [Response type](https://docs.aws.amazon.com/apigateway/latest/developerguide/supported-gateway-response-types.html) of the associated GatewayResponse.
 * `status_code` - (Optional) HTTP status code of the Gateway Response.
 * `response_templates` - (Optional) Map of templates used to transform the response body.
 * `response_parameters` - (Optional) Map of parameters (paths, query strings and headers) of the Gateway Response.


website/docs/r/api_gateway_gateway_response.html.markdown
```diff
@@ -38,7 +38,7 @@ This resource supports the following arguments:
 
 * `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
 * `rest_api_id` - (Required) String identifier of the associated REST API.
-* `response_type` - (Required) Response type of the associated GatewayResponse.
+* `response_type` - (Required) [Response type](https://docs.aws.amazon.com/apigateway/latest/developerguide/supported-gateway-response-types.html) of the associated GatewayResponse.
```
justinretzolk: In other places, we're a bit more explicit with linking to the AWS documentation, so I think we should follow that pattern.

```suggestion
* `response_type` - (Required) Response type of the associated GatewayResponse. See the [AWS Documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/supported-gateway-response-types.html) for supported values.
```

justinretzolk (changes requested):
Thank you very much for taking the time to raise this! I always think linking out to the AWS documentation like this is the best way to ensure the docs stay up to date should they add any additional supported values. I did have one thought regarding formatting that I've left inline.

website/docs/r/api_gateway_gateway_response.html.markdown
dialupdisaster: Ah, I see! Makes sense.

justinretzolk (approved):
Thanks for the quick turnaround, and thanks again for the contribution! I'll get this merged once the checks finish up 🚀

dialupdisaster:
> Thank you very much for taking the time to raise this! I always think linking out to the AWS documentation like this is the best way to ensure the docs stay up to date should they add any additional supported values. I did have one thought regarding formatting that I've left inline.

Should we change the text for the [region links](https://github.com/hashicorp/terraform-provider-aws/blob/513a871fa8bc600d2288ffd5cdb0f6fa2fbd758a/website/docs/r/api_gateway_gateway_response.html.markdown?plain=1#L39) as well?

justinretzolk:
If you're up for that, that would be great @digitalfiz! Ideally, we'll get that completely standardized at some point, but chipping away at it always helps 😅

website/docs/r/api_gateway_gateway_response.html.markdown
```diff
@@ -36,12 +36,12 @@ resource "aws_api_gateway_gateway_response" "test" {
 
 This resource supports the following arguments:
 
-* `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
-* `rest_api_id` - (Required) String identifier of the associated REST API.
-* `response_type` - (Required) Response type of the associated GatewayResponse.
-* `status_code` - (Optional) HTTP status code of the Gateway Response.
-* `response_templates` - (Optional) Map of templates used to transform the response body.
-* `response_parameters` - (Optional) Map of parameters (paths, query strings and headers) of the Gateway Response.
+- `region` - (Optional) Region where this resource will be manage. See the [AWS Documentation](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints) for supported values. Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
```
justinretzolk:
```suggestion
- `region` - (Optional) Region where this resource will be managed. See the [AWS Documentation](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints) for supported values. Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
```

justinretzolk (changes requested):
Suggested a change for a small typo. I noticed the bulleted list changed from `*` to `-` -- I'm not _that_ worried about it (I'm fairly confident it'll render the same regardless), but figured if you're pushing a change for the typo fix, it may be worthwhile to include reverting that too.

dialupdisaster:
Ah, yeah, that would be vscode auto-formatting.

On a side note, I feel like all the extra unrelated tests it runs when you're just editing the website docs really slow down the process.

dialupdisaster:
Fixed the list format since the rest of the documentation is using `*`, not `-`.

justinretzolk (approved):

justinretzolk:
Thanks again @digitalfiz! I'll get this merged once the checks finish up.

PR #43888, opened 2025-08-14
Release: Update changelog and version/version

YakDriver:
## Rollback Plan

### Relations

Closes #0000

```console

...
```


CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-## 6.9.0 (Unreleased)
+## 6.10.0 (Unreleased)
```
johnsonaj, line 1: we should remove the header for the next version. This update should only add the date for the current release that is about to happen

PR #43938, opened 2025-08-18
[Docs] aws_s3_bucket_logging: Add an example using bucket policy to grant permission and clarify S3 key format

tabito-hara:
## Rollback Plan

### Description

* Added an example configuration using a bucket policy to grant permission for S3 server access logging.

  * While bucket ACLs can also be used to grant permission, this approach is no longer recommended.
* Additionally, added a description of the S3 key format.

### References
https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html


ewbankkit:

PR #43940, opened 2025-08-18
[doc]: add example values to reserved instance offering

tinazhouhui:
## Rollback Plan

### Description

Adds most common product types for the reserved instance offering data source. This is to explicitly show that there are differences between aurora postgresql and rds postgresql offerings.


ewbankkit:
@tinazhouhui Thanks for the contribution :tada: :clap:.

Issue #43978, opened 2025-08-21
Resource "aws_bcmdataexports_export" produced an unexpected new value


AlejandroJL:
### Terraform and AWS Provider Version

```console
Terraform: v1.11
# Provider
- hashicorp:aws: 6.9.0
```

### Affected Resource(s) or Data Source(s)

aws_bemdataexports_export

### Expected Behavior

Applying correct

### Actual Behavior

When applying changes to module.core.module.data_export.aws_bcmdataexports_export.cur, provider "provider[\"registry.terraform.io/hashicorp/aws)"]" produced an unexpected new value:
export[0].data_query|0].table_configurations "COST_AND_USAGE_REPORT"]: new element "BILLING_VIEW_ARN" has appeared.
This is a bug in the provider, which should be reported in the provider's own issue tracker.

### Relevant Error/Panic Output

```console
When applying changes to module.core.module.data_export.aws_bcmdataexports_export.cur, provider "provider[\"registry.terraform.io/hashicorp/aws)"]" produced an unexpected new value:
export[0].data_query|0].table_configurations "COST_AND_USAGE_REPORT"]: new element "BILLING_VIEW_ARN" has appeared.
This is a bug in the provider, which should be reported in the provider's own issue tracker.

```

### Sample Terraform Configuration

<details open>
<summary>Click to expand configuration</summary>

```hcl
resource "aws_bemdataexports_export" "cur" {
    export {
        name = "cur-20"
        data_query {
            query_statement = "SELECT identity_line_item_id, identity_time_interval, line_item_product_code,line_item unblended_cost FROM COST_AND_USAGE_REPORT"
            table_configurations = {
                COST_AND_USAGE_REPORT = {
                    TIME_GRANULARITY = "HOURLY",
                    INCLUDE_RESOURCES = "TRUE",
                    INCLUDE_MANUAL_DISCOUNT_COMPATIBILITY = "FALSE",
                    INCLUDE_SPLIT_COST_ALLOCATION_DATA = "FALSE"
                }
            }
        }
        destination_configurations {
            s3_destination {
                s3_bucket = local.landing_bucket
                s3_prefix = "product_monitorization/inbound/"
                s3_region = var.aws.region
                s3_output_configurations {
                    overwrite = "OVERWRITE_REPORT"
                    format = "PARQUET"
                    compression = "PARQUET"
                    output_type = "CUSTOM"
                }
            }
        }
        refresh_cadence {
[...]

PR #43983, opened 2025-08-21
[Docs] aws_bcmdataexports_export: Add `BILLING_VIEW_ARN` to documentation example and clarify `table_configurations` and `query`

tabito-hara:
## Rollback Plan

### Description

* As reported in #43978, if `COST_AND_USAGE_REPORT.BILLING_VIEW_ARN` is not included in `table_configuration`, it is automatically added in the refreshed state. This mismatch between the configuration and the refreshed state causes the following error:

  > produced an unexpected new value:
  > export\[0].data\_query\[0].table\_configurations\["COST\_AND\_USAGE\_REPORT"]: new element "BILLING\_VIEW\_ARN" has appeared.

  * The type of `table_configuration` is `fwtypes.MapOfMapOfStringType`, and part of it is marked as `Computed`. It would not be straightforward to treat only a subset of the map as `Computed`.

* As a workaround, adding `BILLING_VIEW_ARN` to the configuration allows `terraform apply` to succeed.

* Although this does not address the root cause, the documentation example has been updated to include `BILLING_VIEW_ARN`.

* References for `query` and `table_configuration` have also been added.

### Relations
Relates #43978

### References
https://docs.aws.amazon.com/cur/latest/userguide/table-dictionary-cur2.html

### DIFF
--- a/website/docs/r/bcmdataexports_export.html.markdown
+++ b/website/docs/r/bcmdataexports_export.html.markdown
@@ -69,8 +69,8 @@ The following arguments are required:
 
 ### `data_query` Argument Reference
 
-* `query_statement` - (Required) Query statement.
-* `table_configurations` - (Optional) Table configuration.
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:

PR #43144, opened 2025-06-23
Fix/update tests data with spn data souce

myerscf:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

N/A

### Description

Updates all IAM tests to use the SPN Data Provider, so they do not fail when running in Partitions where the polocy suffix is not the regions domain.

### Relations

Relates #38693

### References

Previously done for S3 [here](https://github.com/hashicorp/terraform-provider-aws/pull/38693)

### Output from Acceptance Testing
```console
% make testacc TESTS=TestAccIAMRole PKG=iam
        
--- PASS: TestAccIAMRolePolicy_invalidJSON (5.83s)
--- PASS: TestAccIAMRolePoliciesExclusive_empty (54.63s)
--- PASS: TestAccIAMRoleDataSource_tags (58.37s)
--- PASS: TestAccIAMRolePolicyAttachmentsExclusive_disappears_Role (59.09s)
--- PASS: TestAccIAMRolePolicyAttachmentsExclusive_basic (59.15s)
--- PASS: TestAccIAMRolePolicyAttachmentsExclusive_empty (60.38s)
--- PASS: TestAccIAMRolePolicyAttachmentsExclusive_disappears_Policy (67.47s)
--- PASS: TestAccIAMRolePolicyAttachment_disappears (68.68s)
--- PASS: TestAccIAMRolePolicyAttachment_Disappears_role (70.14s)
--- PASS: TestAccIAMRolePolicy_Policy_invalidResource (32.54s)
--- PASS: TestAccIAMRolePoliciesExclusive_outOfBandAddition (102.12s)
--- PASS: TestAccIAMRolePolicyAttachment_basic (113.78s)
--- PASS: TestAccIAMRolePoliciesExclusive_outOfBandRemoval (108.94s)
--- PASS: TestAccIAMRole_tags_DefaultTags_updateToProviderOnly (115.69s)
--- PASS: TestAccIAMRolePolicy_unknownsInPolicy (55.43s)
--- PASS: TestAccIAMRole_badJSON (7.76s)
--- PASS: TestAccIAMRole_tags_EmptyTag_OnUpdate_Replace (124.93s)
--- PASS: TestAccIAMRole_maxSessionDuration (125.87s)
--- PASS: TestAccIAMRolePolicyAttachmentsExclusive_multiple (126.86s)
--- PASS: TestAccIAMRole_basic (70.10s)
[...]

PR #38693, opened 2024-08-05
tests(s3): updated s3 test data config to use aws_service_principal data source

myerscf:
### Description

With the release of the AWS Service Principal Name Data Source in [v5.60.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.60.0) this PR updates S3 test resources to use it, in order to provide more effective testing across partitions

### Relations

Relates #38307

### References

N/A

### Output from Acceptance Testing

```console
%  TF_ACC=1 go test ./internal/service/s3/ -v -run="TestAccS3BucketPolicyDataSource"

=== RUN   TestAccS3BucketPolicyDataSource_basic
=== PAUSE TestAccS3BucketPolicyDataSource_basic
=== CONT  TestAccS3BucketPolicyDataSource_basic
--- PASS: TestAccS3BucketPolicyDataSource_basic (18.81s)
PASS

% TF_ACC=1 go test ./internal/service/s3/ -v -run="TestAccS3BucketPolicy"

=== RUN   TestAccS3BucketPolicyDataSource_basic
=== PAUSE TestAccS3BucketPolicyDataSource_basic
=== RUN   TestAccS3BucketPolicy_basic
=== PAUSE TestAccS3BucketPolicy_basic
=== RUN   TestAccS3BucketPolicy_disappears
=== PAUSE TestAccS3BucketPolicy_disappears
=== RUN   TestAccS3BucketPolicy_disappears_bucket
=== PAUSE TestAccS3BucketPolicy_disappears_bucket
=== RUN   TestAccS3BucketPolicy_policyUpdate
=== PAUSE TestAccS3BucketPolicy_policyUpdate
=== RUN   TestAccS3BucketPolicy_IAMRoleOrder_policyDoc
=== PAUSE TestAccS3BucketPolicy_IAMRoleOrder_policyDoc
=== RUN   TestAccS3BucketPolicy_IAMRoleOrder_policyDocNotPrincipal
=== PAUSE TestAccS3BucketPolicy_IAMRoleOrder_policyDocNotPrincipal
=== RUN   TestAccS3BucketPolicy_IAMRoleOrder_jsonEncode
=== PAUSE TestAccS3BucketPolicy_IAMRoleOrder_jsonEncode
=== RUN   TestAccS3BucketPolicy_migrate_noChange
=== PAUSE TestAccS3BucketPolicy_migrate_noChange
=== RUN   TestAccS3BucketPolicy_migrate_withChange
[...]

PR #43987, opened 2025-08-21
Fix/update eks tests with spn data souce

myerscf:
## Rollback Plan

N/A

### Description

Updates all EKS tests to use the SPN Data Provider, so they do not fail when running in Partitions where the policy suffix is not the regions domain.

Additionally updated some tests there where stale a failing, so had to make them work again to do this PR so my tweaked tests passed 😁

### References

Previously done for S3 https://github.com/hashicorp/terraform-provider-aws/pull/38693
Previously done for IAM https://github.com/hashicorp/terraform-provider-aws/pull/43144

Have attached output as file, as very long

[eks_test_ouput.log](https://github.com/user-attachments/files/21920986/eks_test_ouput.log)

```console
% make t T="TestAccEKS(Cluster|NodeGroup|FargateProfile|Addon|IdentityProviderConfig|AccessEntry|AccessPolicyAssociation)" K=eks P=10

....

PASS
ok    github.com/hashicorp/terraform-provider-aws/internal/service/eks        12009.975s
```


myerscf:
From previous conversation with @ewbankkit 
> There are some under the EKS service which currently fail as they are a bit stale.
The default cluster version for EKS creates wtith 1.33 which no longer supports the al2 image which the tests pulling.
I looked at updating them use the al-2023  but i think i've started down a rabbit hole that I need to stop....since al-2023 requires a different user_data format
and i am in no way a person who understands wtf EKS is doing
Would I be ok, pin tests to using 1.32 and and leave it with current al2 image, allowing them to work as expected?
Future tests would need to be written/upgrade the suite to account for al-2023 and the different user_data format

> Hi Fred,
As long as we keep at least one aws_eks_cluster test (TestAccEKSCluster_basic) with version unconfigured, the maintainers are happy with you pinning.
We'll have to deal with that once 1.32 is retired...

ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@myerscf Thanks for the contribution :tada: :clap:.

Issue #42409, opened 2025-04-29
SaaS Manager for Amazon CloudFront


lbertorello:
### Description

Hi,
AWS announces CloudFront SaaS Manager, a new Amazon CloudFront feature designed to efficiently manage content delivery across multiple websites for Software-as-a-Service (SaaS) providers, web development platforms, and companies with multiple brands/websites.

More information:

* [Announcement](https://aws.amazon.com/about-aws/whats-new/2025/04/saas-manager-amazon-cloudfront/)
* [Blog post](https://aws.amazon.com/blogs/aws/reduce-your-operational-overhead-today-with-amazon-cloudfront-saas-manager/)
* [Developer Guide](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-config-options.html)

### Affected Resource(s) or Data Source(s)

- [x] `aws_cloudfront_multitenant_distribution` (#45535)
- [x] New resource `aws_cloudfront_distribution_tenant` (#45088)
- [ ] New resource `aws_cloudfront_connection_group` (#44885)

### Potential Terraform Configuration

```hcl

```

### References

* https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudfront#CreateDistributionInput
* https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudfront#CreateDistributionTenantInput

Requires https://github.com/hashicorp/terraform-provider-aws/pull/42425.

### Would you like to implement the enhancement?

No

rshayman:
There also needs to be another resource: `aws_cloudfront_connection_group`

References:
* https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudfront#CreateConnectionGroupInput

heindrickdumdum0217:
Do we have plan with this feature?
We have to use this feature with Terraform.

stylianosom:
Any update on this?

ran-azrad-sp:
Like others here, I'm also looking forward to native support for the `aws_cloudfront_distribution_tenant` resource.

In the meantime, I've created a workaround using a null_resource with local-exec provisioners. This solution handles both the create and destroy lifecycles, including the necessary logic to disable the tenant and wait for the operation to complete before deletion.

Prerequisites: This workaround requires the AWS CLI and jq to be installed on the machine running Terraform.

Here is the code for anyone who needs a temporary solution. I've created it as a self-contained module for easier use.


[...]

PR #44034, opened 2025-08-26
[Enhancement] [WIP] r/aws_cloudfront_distribution: Added SaaS Manager configuration support for distributions

codyzhao2770:
## Rollback Plan

N/A

### Relations
Updates `aws_cloudfront_distribution` as requested by https://github.com/hashicorp/terraform-provider-aws/issues/42409

```console
% make testacc TESTS=TestAccCloudFrontDistribution PKG=cloudfront 
TF_ACC=1 go1.24.6 test ./internal/service/cloudfront/... -v -count 1 -parallel 20 -run='TestAccCloudFrontDistribution'  -timeout 360m -vet=off
2025/08/26 10:22:17 Creating Terraform AWS Provider (SDKv2-style)...
2025/08/26 10:22:17 Initializing Terraform AWS Provider (SDKv2-style)...        
--- PASS: TestAccCloudFrontDistribution_s3Origin (515.41s)
--- PASS: TestAccCloudFrontDistribution_Origin_connectionAttempts (507.09s)
--- PASS: TestAccCloudFrontDistribution_ViewerCertificateACMCertificateARN_conflictsWithCloudFrontDefaultCertificate (327.65s)
--- PASS: TestAccCloudFrontDistribution_noCustomErrorResponse (461.97s)
--- PASS: TestAccCloudFrontDistribution_orderedCacheBehaviorResponseHeadersPolicy (494.69s)
--- PASS: TestAccCloudFrontDistribution_http11 (485.81s)
--- PASS: TestAccCloudFrontDistribution_disappears (269.51s)
--- PASS: TestAccCloudFrontDistribution_basic (270.86s)
--- PASS: TestAccCloudFrontDistribution_isIPV6Enabled (572.06s)
--- PASS: TestAccCloudFrontDistribution_waitForDeployment (580.28s)
--- PASS: TestAccCloudFrontDistribution_tags (303.26s)
--- PASS: TestAccCloudFrontDistribution_orderedCacheBehavior (520.11s)
--- PASS: TestAccCloudFrontDistribution_noOptionalItems (540.33s)
--- PASS: TestAccCloudFrontDistribution_orderedCacheBehaviorCachePolicy (531.73s)
--- PASS: TestAccCloudFrontDistribution_enabled (821.89s)
--- PASS: TestAccCloudFrontDistribution_forwardedValuesToCachePolicy (609.34s)
--- PASS: TestAccCloudFrontDistribution_vpcOriginConfig (1225.58s)
PASS
ok    github.com/hashicorp/terraform-provider-aws/internal/service/cloudfront 1226.874s
...
```

### DIFF
--- a/.changelog/43950.txt
+++ b/.changelog/43950.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_inspector2_filter: Support `code_repository_project_name`, `code_repository_provider_type`, `ecr_image_in_use_count`, and `ecr_image_last_in_use_at` in `filter_criteria`
+```
--- a/.changelog/43967.txt
+++ b/.changelog/43967.txt
@@ -1,0 +1,7 @@
+```release-note:enhancement
+resource/aws_vpc_ipam: Add `metered_account` argument
+```
+
+```release-note:enhancement
+data-source/aws_vpc_ipam: Add `metered_account` attribute
+```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 6.11.0 (Unreleased)
 
+ENHANCEMENTS:
+
+* resource/aws_distribution: Add SaaS manager configuration support ([#42409](https://github.com/hashicorp/terraform-provider-aws/issues/42409))
+
 ## 6.10.0 (August 21, 2025)
 
 NOTES:
--- a/internal/service/cloudfront/distribution.go
+++ b/internal/service/cloudfront/distribution.go
@@ -73,6 +73,13 @@ func resourceDistribution() *schema.Resource {
 				Optional:     true,
 				ValidateFunc: validation.StringLenBetween(0, 128),
 			},
+			"connection_mode": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          awstypes.ConnectionModeDirect,
+				ValidateDiagFunc: enum.Validate[awstypes.ConnectionMode](),
+				ConflictsWith:    []string{"is_ipv6_enabled", "price_class", "aliases", "staging", "continuous_deployment_policy_id"},
+			},
 			"continuous_deployment_policy_id": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -790,6 +797,58 @@ func resourceDistribution() *schema.Resource {
 			},
 			names.AttrTags:    tftags.TagsSchema(),
 			names.AttrTagsAll: tftags.TagsSchemaComputed(),
+			"tenant_config": {
+				Type:          schema.TypeList,
+				MaxItems:      1,
+				Optional:      true,
+				ConflictsWith: []string{"is_ipv6_enabled", "price_class", "aliases", "staging", "continuous_deployment_policy_id"},
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"parameter_definitions": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"definition": {
+										Type:     schema.TypeList,
+										MaxItems: 1,
+										Required: true,
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"string_schema": {
+													Type:     schema.TypeList,
+													MaxItems: 1,
+													Optional: true,
+													Elem: &schema.Resource{
+														Schema: map[string]*schema.Schema{
+															"comment": {
+																Type:     schema.TypeString,
+																Optional: true,
+															},
+															"default_value": {
+																Type:     schema.TypeString,
+																Optional: true,
+															},
+															"required": {
+																Type:     schema.TypeBool,
+																Required: true,
+															},
+														},
+													},
+												},
+											},
+										},
+									},
+									"name": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
 			"trusted_key_groups": {
 				Type:     schema.TypeList,
 				Computed: true,
@@ -964,10 +1023,17 @@ func resourceDistributionRead(ctx context.Context, d *schema.ResourceData, meta 
 	if aws.ToString(distributionConfig.Comment) != "" {
 		d.Set(names.AttrComment, distributionConfig.Comment)
 	}
+	if distributionConfig.ConnectionMode != "" {
+		d.Set("connection_mode", distributionConfig.ConnectionMode)
+	} else {
+		d.Set("connection_mode", awstypes.ConnectionModeDirect)
+	}
 	// Not having this set for staging distributions causes IllegalUpdate errors when making updates of any kind.
 	// If this absolutely must not be optional/computed, the policy ID will need to be retrieved and set for each
 	// API call for staging distributions.
-	d.Set("continuous_deployment_policy_id", distributionConfig.ContinuousDeploymentPolicyId)
+	if distributionConfig.ContinuousDeploymentPolicyId != nil {
+		d.Set("continuous_deployment_policy_id", distributionConfig.ContinuousDeploymentPolicyId)
+	}
 	if distributionConfig.CustomErrorResponses != nil {
 		if err := d.Set("custom_error_response", flattenCustomErrorResponses(distributionConfig.CustomErrorResponses)); err != nil {
 			return sdkdiag.AppendErrorf(diags, "setting custom_error_response: %s", err)
@@ -983,7 +1049,11 @@ func resourceDistributionRead(ctx context.Context, d *schema.ResourceData, meta 
 	d.Set("http_version", distributionConfig.HttpVersion)
 	d.Set(names.AttrHostedZoneID, meta.(*conns.AWSClient).CloudFrontDistributionHostedZoneID(ctx))
 	d.Set("in_progress_validation_batches", output.Distribution.InProgressInvalidationBatches)
-	d.Set("is_ipv6_enabled", distributionConfig.IsIPV6Enabled)
+	if distributionConfig.IsIPV6Enabled != nil {
+		d.Set("is_ipv6_enabled", distributionConfig.IsIPV6Enabled)
+	} else {
+		d.Set("is_ipv6_enabled", false)
+	}
 	d.Set("last_modified_time", aws.String(output.Distribution.LastModifiedTime.String()))
 	if distributionConfig.Logging != nil && aws.ToBool(distributionConfig.Logging.Enabled) {
 		if err := d.Set("logging_config", flattenLoggingConfig(distributionConfig.Logging)); err != nil {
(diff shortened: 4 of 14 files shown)


codyzhao2770:
@ewbankkit @justinretzolk Hi I'm current working on another new resource that is part of Cloudfront's SaaS manager, but acceptance testing it requires the changes in this PR. My team suggested I write tests for the new resource using pre-deployed resources on my own AWS account so I can meet the deadline to publish the PR for that, but I understand its bad practice. Will this get merged soon? Or I was thinking I could write the tests normally and make it clear in my PR that this PR should be merged first.

YakDriver:
@codyzhao2770 Thank you for your work on this!

I'm going to do a little cleanup on this PR. In the meantime, here are some considerations:
1. Would it conceptually make more sense to have a separate resource for multi-tenant distributions? The console splits distributions and multi-tenant distributions. There is a lot of overlap between the two but also many things are different/not used. We can extract schema and other pieces to reuse.
2. [New resources should be implemented in plugin framework](#32917). The modification to `aws_cloudfront_distribution` is fine in SDK since the legacy resource is SDK. But, the following should be implemented in framework:
    * #44181
    * #44183
    * #44184
    * #44185

gahykiamzn:
Hi @YakDriver,
I'll be taking over from Cody to drive this till completion. I took some time to onboard and ramp up to the context. Regarding this
> Would it conceptually make more sense to have a separate resource for multi-tenant distributions? The console splits distributions and multi-tenant distributions. There is a lot of overlap between the two but also many things are different/not used. We can extract schema and other pieces to reuse.

We plan to keep multi-tenant distributions as the same distributions as limitations such as throttles are shared. 

and as for the framework comment, I'll look into it and get them on the framework

YakDriver:
After reviewing this more, there's so much that's not carried over between distrbution and multi-tenant distribution, it makes more sense to split them.

gahykiamzn:
> there's so much that's no carried over between distrbution and multi-tenant distribution

What do you mean by that? Can you explain what is this so much that is not carried over?

The main difference between the two is that multi-tenant distributions are not routable and cannot have aliases added to them. Cache behaviors and origins, which constitute the majority of what makes a distribution, are still present which is why we are not going to split them.

gahykiamzn:
A small comment about requirements, this code must be merged **AFTER** connection groups new resource (PR #44885 )

YakDriver:
@gahykiamzn Thank you so much for your input on how the Terraform AWS Provider works. As maintainers, we value input from our partners when we make provider decisions!

When deciding on multiple vs. a single resource, we consider many factors. There is always a balance for what makes sense. Here are some of the reasons we've decided these should be separate resources:

## 1) The APIs encode different **products**, not just flags

Multi-tenant distributions introduce a **template + inheritance** model (via `tenant_config`, connection groups, certificate/domain inheritance, and stricter policy usage). That model **removes** whole classes of knobs (TTL fields, price class, aliasing, legacy signers, etc.) and **adds** tenant-template schema you don’t see in standard. Practically:

* **Standard** = a single distribution you fully tune (behaviors, TTLs, price class, aliases, cert choice, etc.).
* **Multi-tenant** = a distribution that **templates** and **routes** for many tenants, pushing domains/IP family/certs/routing into **connection groups** and **parameterized config** (and enforcing modern policies).

Trying to bolt these together yields a huge “if type == X then disallow Y” matrix, confusing users and increasing error-prone conditionals.

## 2) Clean schema lines = clearer UX and safer configs

With two resources you get:

* **Focused schemas** that mirror the product: no walls of mutually-exclusive fields.
* **Type-specific validation** that reads like the docs (e.g., “WAF v2 only” enforced at plan time).
* **Docs/examples** that speak to the audience (CDN vs. SaaS multi-tenant).

**This avoids the “giant resource with half the arguments conditionally ignored” anti-pattern.**

## 3) Maintenance & testability

A single mega-resource would balloon acceptance tests (branching for every unsupported field). Two resources let you:

* Write **narrow, scenario-based tests** per product.
* Keep provider code split by **API surface and invariants**.
* Reduce regression risk: changes for one product don’t destabilize the other.

## 4) Strategic Framework migration

* Keep `aws_cloudfront_distribution` on **Plugin SDK v2** for stability.
* Build `aws_cloudfront_multitenant_distribution` on **Plugin Framework** to prove patterns (nested object types, validators, param schemas), collect feedback, and then port lessons back when we migrate the standard resource.

This de-risks the Framework move without blocking users.

## 5) Standard-only (not in Multi-tenant)

**Distribution-level**

* `aliases` → Multi-tenant ties hostnames to **connection groups**; you don’t attach them directly per distribution.
* `anycast_ip_list_id` → Managed via connection group mechanics; not set on the distribution.
* `continuous_deployment_policy_id` → Not supported in multi-tenant.
* `is_ipv6_enabled` → IP family/edge networking handled by the connection group; not user-set per multi-tenant distribution.
* `price_class` → Not supported (multi-tenant abstracts edge footprint selection).
* `staging` → Not supported.

**Cache behaviors**

* `default_ttl`, `max_ttl`, `min_ttl` → **Enforced** to use cache policies; the raw TTL knobs are absent.
* `smooth_streaming` → Not supported.
* `trusted_signers` → **Deprecated legacy signers**; use `trusted_key_groups`. Multi-tenant **requires** key groups if you do signed URLs/cookies.
* `forwarded_values {}` → Deprecated block; multi-tenant uses **Origin Request Policy** + **Cache Policy** exclusively.

**Viewer certificate**

* `iam_certificate_id` → **Standard** still tolerates IAM-uploaded certs (discouraged); **multi-tenant** requires ACM.

**Other features commonly in Standard but not in Multi-tenant (today)**

* Lambda@Edge / Function associations, field-level encryption, fine-grained real-time logs knobs, some geo-restriction patterns—multi-tenant purposely narrows/standardizes here. (If/when parity lands, you can add it to the multi-tenant resource without muddying the standard one.)

## 6) Multi-tenant-only

* `tenant_config` (required) - A nested template block defining **parameter definitions** (schema for per-tenant overrides). This is the heart of multi-tenant and has no analogue in standard.
* **Automatic/forced behaviors**
    * `connection_mode` = `"tenant-only"` (fixed, not user-set).
    * `web_acl_id` = **WAF v2 only** (CloudFront “global” WAF Classic may still be attachable in legacy standard flows, but multi-tenant forbids v1).
* **Certificate & domain handling** via **connection groups** with **inheritance/association**, not in-resource per-distribution toggles.

## 7) Shared fields with different rules

* `comment` → **Required** in multi-tenant (for clarity in tenant fleet views); optional in standard.
* `enabled` → Required in both, but multi-tenant validation/context is stricter due to template/connection dependencies.
* Certificate handling → Multi-tenant is **streamlined** (ACM + group-level inheritance). Standard allows more per-distribution variance.

## 8) Product & console alignment

* AWS surfaces multi-tenant distributions in **distinct console flows and docs**, and the API rejects many standard-only args with clear messages like *“This field only supports standard distributions. You can’t specify this field for multi-tenant distributions.”*
* Mirroring that separation in Terraform gives users an immediate mental model: *“pick the resource that matches your CloudFront product.”*

## 9) User experience

Two resources remove ambiguity at author-time and review-time:

```hcl
# SaaS, tenant-aware template
resource "aws_cloudfront_multitenant_distribution" "saas" {
  comment = "SaaS edge template"
  tenant_config {
    parameter_definition {
      name     = "tenant_theme"
      type     = "string"
      required = false
    }
    # ...
  }
  # No aliases/TTLs/price_class/etc. — those belong to policies/connection groups.
}

# Classic CDN for a site/app
resource "aws_cloudfront_distribution" "cdn" {
  comment     = "Marketing site CDN"
  aliases     = ["www.example.com"]
  price_class = "PriceClass_200"
  viewer_certificate {
    acm_certificate_arn = aws_acm_certificate.cert.arn
    ssl_support_method  = "sni-only"
  }
  # Full behavior toolbox, including legacy knobs for existing estates.
}
```

Reviewers instantly understand intent without decoding conditionals.

## 10) Testing & docs

* **Separate ATs** (acceptance tests) per resource: faster, clearer, and more realistic scenarios (tenant routing vs. single-tenant behavior tunables).
* **Docs/examples** can be concise and opinionated: multi-tenant walkthroughs focus on parameter schemas, connection groups, and policy-first config; standard docs continue to cover the broad feature map.

## 11) Bottom line

Separating into `aws_cloudfront_distribution` and `aws_cloudfront_multitenant_distribution`:

* Reflects **real product boundaries** (template/inheritance vs. bespoke).
* Delivers a **cleaner UX** (no giant matrix of “forbidden when type=X”).
* Improves **safety** (stronger, type-specific validation).
* Reduces **maintenance risk** and **test complexity**.
* Enables a **measured Plugin Framework rollout** on new surface area without churning the mature resource.

All of that maps directly to fewer user mistakes, clearer docs, faster tests, and a smoother path to Framework across the provider.

gahykiamzn:
There's a couple nitpickable information that is wrong around connection groups and in `Other features commonly in Standard but not in Multi-tenant (today)` section, but I won't get into that as I understand your point of view on the maintainability for this.

I have a question about your perspective on this. Give the fact that we use `connection-mode` (not to confuse with Connection Groups) to determine if a distribution is a standard distribution or a multi-tenant distribution, won't it be confusing to the user if we are to abstract that all away by making it a separate resource?

Furthermore, although there are differences the bulk of the payload is the same. The resource is also created through the same API and so does CFN. Wouldn't this divergence confuse users?

YakDriver:
Great questions—thanks for pushing on the user-experience angle. Also, you're under no obligation to create the new framework multi-tenant resource. I've begun work on it.

Totally fair on the nitpicks. I’m sure I’ve got a couple edges wrong re: connection groups and that “other features” list; happy to tighten those as we go. The core argument here is UX + safety, not that the internals are 100% orthogonal.

### “But connection-mode is the switch—won’t a separate resource hide that and confuse users?”

I see it the other way around: exposing two resources _reduces_ confusion at author time and review time.

* **Author intent is explicit.** `aws_cloudfront_multitenant_distribution` says “you are building a tenant template.” That’s easier to reason about than “set connection_mode=TENANT_ONLY and now half the arguments silently don’t apply.”
* **Plan-time guardrails > runtime surprises.** With two resources, we can make the multi-tenant schema _only_ accept what’s valid, rather than asking users to learn which 30% of a giant schema becomes forbidden after they flip a single enum.
* **We don’t hide connection-mode.** We can still surface it as a computed attribute in both resources so users see exactly what the API thinks. This preserves observability without forcing users to toggle it directly.
* **Friendly mismatches.** If someone tries to `import` a tenant-only distribution into the standard resource (or vice-versa), we can emit a precise diagnostic: “This distribution’s connection_mode is TENANT_ONLY. Import it as aws_cloudfront_multitenant_distribution.”

### “Bulk of the payload is the same; same API & CFN—won’t Terraform divergence confuse people?”

Terraform’s resource boundaries are about (or should be) clear contracts and safety, not 1:1 parity with an API type. We’ve taken both paths in the provider before:

* Sometimes we _merge_ (e.g., consolidated load balancer types).
* Sometimes we _split_ when modes imply large sets of mutually exclusive fields, different invariants, and different mental models (good examples across RDS, EKS, and S3 controls).

Multi-tenant has a template/inheritance model, connection-group mediation for domains/IP/certs, policy-only enforcement (TTLs, request policies), WAF v2 only. The payloads are similar but these aren’t minor toggles; they’re different operating assumptions. Keeping a single resource requires conditional validation webs, “this field is ignored when…” cases, longer docs and examples in docs, and more foot-guns.

### Mitigations to keep the experience smooth

* State visibility: expose `connection_mode` read-only so the Terraform state matches the console/API.
* Import guardrails: detect wrong resource/import and give a one-line fix suggestion.

gahykiamzn:
@YakDriver what is the estimated completion time for this work? 

The other two resources are ready and we would like to get this all merged as soon as possible to unblock some of our customers

Connection Group: https://github.com/hashicorp/terraform-provider-aws/pull/44885
Distribution Tenant: https://github.com/hashicorp/terraform-provider-aws/pull/45088

PR #44162, opened 2025-09-05
Add AppConfig multi-variate example

KyMidd:
## Rollback Plan

No. 

### Description

Adding an example of multi-variate AppConfig feature flags. These are feature flags that are part of a Configuration, and which can have Rules which provide different values for the boolean flags. In the example provided, a variable populated with a list of userIds is read and iterated, and for these users, we set loggingEnabled to True. All others (and those lacking Context) will receive loggingEnabled to False. 

### Relations

Closes #0000

### References

[Creating a multi-variant feature flag AWS Documentation](https://docs.aws.amazon.com/appconfig/latest/userguide/appconfig-creating-multi-variant-feature-flags-procedures.html)

n/a documentation only

### DIFF
--- a/website/docs/r/appconfig_hosted_configuration_version.html.markdown
+++ b/website/docs/r/appconfig_hosted_configuration_version.html.markdown
@@ -79,6 +79,42 @@ resource "aws_appconfig_hosted_configuration_version" "example" {
 }
 ```
 
+### Multi-variate Feature Flags
+
+```terraform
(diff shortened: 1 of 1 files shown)


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@KyMidd Thanks for the contribution :tada: :clap:.

PR #44166, opened 2025-09-06
docs: add missing `AWS_IAM` to `auth_provider` in `appsync_api`

auvred:
## Rollback Plan

### Description

`AWS_IAM` is a valid auth provider type.

I copied the order of possible values from here for better consistency:

https://github.com/hashicorp/terraform-provider-aws/blob/d74e0fd412f19f4fbb931d9ae66eec0a4a94b9fb/website/docs/r/appsync_api.html.markdown?plain=1#L164-L168

### References

https://docs.aws.amazon.com/appsync/latest/APIReference/API_AuthProvider.html#appsync-Type-AuthProvider-authType

### DIFF
--- a/website/docs/r/appsync_api.html.markdown
+++ b/website/docs/r/appsync_api.html.markdown
@@ -131,7 +131,7 @@ The `event_config` block supports the following:
 
 The `auth_provider` block supports the following:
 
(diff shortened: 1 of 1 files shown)


jar-b:
Thank you for the correction, @auvred! 🔍

PR #44174, opened 2025-09-08
attribute names and description for managedRuleGroupConfigATPResponse…

megashby:
…InspectionSchema json object updated

## Rollback Plan

N/A

### Description

Update managedRuleGroupConfigATPResponseInspectionSchema json object to have attributes `failure_values` & `success_values` as used in schema and AWS docs. Also update the description of these attributes.

The go schema correctly uses `failure_values` and `success_values`: https://github.com/hashicorp/terraform-provider-aws/blob/d74e0fd412f19f4fbb931d9ae66eec0a4a94b9fb/internal/service/wafv2/schemas.go#L1671

it appears this section may have been copy pasted incorrectly from here: https://github.com/hashicorp/terraform-provider-aws/blob/d74e0fd412f19f4fbb931d9ae66eec0a4a94b9fb/website/docs/r/wafv2_web_acl.html.markdown?plain=1#L900

### References
AWS docs reference here: https://docs.aws.amazon.com/waf/latest/APIReference/API_ResponseInspectionJson.html

### DIFF
--- a/website/docs/cdktf/python/r/wafv2_web_acl.html.markdown
+++ b/website/docs/cdktf/python/r/wafv2_web_acl.html.markdown
@@ -915,8 +915,8 @@ The `managed_rule_group_configs` block support the following arguments:
 ### `json` Block
 
 * `identifier` (Required) The identifier for the value to match against in the JSON.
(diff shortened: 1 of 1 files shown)


jar-b:
Thank you for the correction, @megashby! 👍

PR #44187, opened 2025-09-08
Update lambda_function.html.markdown - `log_format` option under `logging_config` is optional and not required

0xabdi:
According to the official [AWS API reference doc. for lambda](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-lambda-function-loggingconfig.html), the `log_format` option under `logging_config` argument is optional and not required.

## Rollback Plan

### Relations

Closes #0000

```console

...
```


0xabdi:
I am not very familiar with the GO code but it looks like [this](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/lambda/function.go#L265) needs to be updated to `Optional:     true,` too

PR #44228, opened 2025-09-10
r/aws_secretsmanager_secret: warning when `policy` is invalid

### LINKED ISSUE #29496: [Bug]: v4.55.0 report invalid principals in SecretsManager resource policy
### Terraform Core Version

1.3.5

### AWS Provider Version

4.55.0

### Story behind this issue

We have some poorly written Terraform code in our infrastructure codebase. One of the Terraform root module that we have created some SecretsManager secrets and populated their resource policy with hardcoded ARN of IAM User defined outside of this Terraform root module. We moved the secret from the poorly written Terraform root module to a new one, but sadly we haven't migrated the resource policy. One day, the IAM User was destroyed because it was not used anymore, but we did not updated the resource policy which eventually became broken (it contained an invalid principal). After updating to the provider 4.55.0 we are not able to plan the Terraform root module that contains the broken secret. The plan and apply was still possible with the provider in the v4.54.0 version.

### Affected resources

* aws_secretsmanager_secret

### Related commit

From my understanding this behavior is caused by this [code](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/secretsmanager/secret.go#L304). When retrying to access the secret, the code also check if the policy is valid or not. This was implemented with [this commit](https://github.com/hashicorp/terraform-provider-aws/commit/d1d848b157dbdd8803142160d6caf3b16383f58d). This change was not showcased in the Changelog, but I found [this PR](https://github.com/hashicorp/terraform-provider-aws/pull/29427).

### Expected Behavior

We are able to update attribute of a secrets even when it has a resource policy that contains invalid principals. Error may arise when updating the resource policy tho.

### Actual Behavior

Terraform cannot refresh the state as the policy contains invalid principals

```shell
aws_secretsmanager_secret.this: Refreshing state... [id=arn:aws:secretsmanager:eu-west-3:012345678912:secret:terraform-aws-provider-bug-QNrCc-if-q748a4]
╷
│ Error: reading Secrets Manager Secret (arn:aws:secretsmanager:eu-west-3:012345678912:secret:terraform-aws-provider-bug-QNrCc-if-q748a4) policy: contains invalid principals
│ 
│   with aws_secretsmanager_secret.this,
│   on main.tf line 19, in resource "aws_secretsmanager_secret" "this":
│   19: resource "aws_secretsmanager_secret" "this" {
│ 
╵
```

terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "4.55.0"
    }
  }
}

provider "aws" {}

resource "random_id" "this" {
  byte_length = 6
}

resource "aws_secretsmanager_secret" "this" {
  name = "terraform-provider-aws-bug-${random_id.this.id}"
}

resource "aws_iam_user" "this" {
  name = "terraform-provider-aws-bug"
}

data "aws_iam_policy_document" "secret" {
  statement {
    sid    = "BrokenUser"
    effect = "Allow"
    principals {
      type        = "AWS"
      identifiers = [aws_iam_user.this.arn]
    }
    actions   = ["secretsmanager:GetSecretValue"]
    resources = [aws_secretsmanager_secret.this.arn]
  }
}

output "policy" {
  value = data.aws_iam_policy_document.secret.json
}

### Steps to Reproduce

1. Create a root module with a secret, an IAM User (code provided above)
2. Manually configure a resource policy on the secret allowing the IAM user to retrieve its value. You can use the policy in output of the Terraform configuration provided earlier
4. Remove the IAM User by targeting it with the terraform command: `terraform apply -destroy -target aws_iam_user.this`
5. Wait some time (around 5 min). AWS will update the resource policy with the UID of the IAM User and remove its ARN from the policy.
6. Try to plan the configuration again. It should now expose the error, if not just retry one more time (I had the issue after some retries)

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

I think that having an invalid policy should end up with a warning and not an error. The error will comes from AWS if the user wants to modify the policy. Maybe we can move the [`PolicyHasValidAWSPrincipals` call](https://github.com/hashicorp/terraform-provider-aws/blob/ae683662502d49e73caf3a8d01fb4689ab400592/internal/service/secretsmanager/secret.go#L314) to be performed only when Terraform sees update in the resource policy ? We could still have the warning during refresh time tho.

[The PR](https://github.com/hashicorp/terraform-provider-aws/pull/29427) talks about a race condition with Secrets Manager Secrets and I'm not sure how we could find a solution to both of these problems. Maybe accepting the principal UID as last resort and ouput a warning to the user could make the trick

### References

_No response_

Yes

johnsonaj:
## Rollback Plan

### Description

Inline policies that are not managed by Terraform can become invalid by external influences. Validation in the `aws_secretsmanager_secret` resource produces an error, which then makes the secret unmanageable. 

### Relations

Closes #29496

```console

...
```

### DIFF
--- a/internal/service/secretsmanager/secret.go
+++ b/internal/service/secretsmanager/secret.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -269,6 +270,11 @@ func resourceSecretRead(ctx context.Context, d *schema.ResourceData, meta any) d
 	})
 
 	if err != nil {
+		if strings.Contains(err.Error(), "contains invalid principals") {
+			d.Set(names.AttrPolicy, "")
+			return sdkdiag.AppendWarningf(diags, "reading Secrets Manager Secret (%s) policy: %s", d.Id(), err)
+		}
+
 		return sdkdiag.AppendErrorf(diags, "reading Secrets Manager Secret (%s) policy: %s", d.Id(), err)
 	} else if v := policy.ResourcePolicy; v != nil {
 		policyToSet, err := verify.PolicyToSet(d.Get(names.AttrPolicy).(string), aws.ToString(v))


ewbankkit (dismissed):
LGTM :rocket:.

internal/service/secretsmanager/secret.go
```diff
@@ -269,6 +270,11 @@ func resourceSecretRead(ctx context.Context, d *schema.ResourceData, meta any) d
 	})
 
 	if err != nil {
+		if strings.Contains(err.Error(), "contains invalid principals") {
+			d.Set(names.AttrPolicy, "")
+			return sdkdiag.AppendWarningf(diags, "reading Secrets Manager Secret (%s) policy: %s", d.Id(), err)
```
jar-b: Should this append the warning without returning so that `setTagsOut` can still be run?

johnsonaj: @jar-b updated to return only if there is a diagnostic error

internal/service/secretsmanager/secret.go
```diff
@@ -269,9 +270,19 @@ func resourceSecretRead(ctx context.Context, d *schema.ResourceData, meta any) d
...
+		if strings.Contains(err.Error(), "contains invalid principals") {
+			diags = sdkdiag.AppendWarningf(diags, "reading Secrets Manager Secret (%s) policy: %s", d.Id(), err)
+		} else {
+			diags = sdkdiag.AppendErrorf(diags, "reading Secrets Manager Secret (%s) policy: %s", d.Id(), err)
+		}
+	}
+
+	if diags.HasError() {
+		return diags
+	}
```
jar-b: Slight simplification.

```suggestion
		} else {
			return sdkdiag.AppendErrorf(diags, "reading Secrets Manager Secret (%s) policy: %s", d.Id(), err)
		}
	}
```

PR #44233, opened 2025-09-11
Added `deletion_protection` to `aws_msk_cluster`

### LINKED ISSUE #44209: Add deletion_protection to `aws_msk_cluster`
### Description

Since `prevent_destroy` does not allow for variables, we can't allow for deletion in development environments. `aws_rds_cluster` has this argument, and it would be nice to have this option for many other resources, but `aws_msk_cluster` is the most important to us currently.

`aws_msk_cluster`

```hcl
resource "aws_msk_cluster" "provisioned" {
  cluster_name                     = local.name
  kafka_version                     = var.kafka_version
  number_of_broker_nodes = var.kafka_broker_count
  deletion_protection            = var.deletion_protection
}
```

### References

_No response_

Yes

dialupdisaster:
Adds deletion protection support to the `aws_msk_cluster` resource, implementing Terraform-level protection similar to the existing `aws_rds_cluster` resource.

###   Reasoning

Since you cannot use variables in the `lifecycle` block, preventing destruction is either always on or always off. This would allow for it to be variable, like with `aws_rds_cluster`, so users can turn it off for development if they so choose, and on for production. While AWS MSK does not currently support deletion protection at the API level (unlike RDS), this implementation provides Terraform-level protection, preventing cluster deletion when enabled.

###   Changes Made

  - Added deletion_protection optional boolean argument to `aws_msk_cluster` resource schema
  - Implemented deletion protection logic in the delete function to block deletion when enabled
  - Added comprehensive test coverage following existing patterns from `aws_rds_cluster` tests
  - Updated resource documentation to match `aws_rds_cluster` documentation format

###   Expected User Experience

  - Users can now set `deletion_protection = true` to protect MSK clusters from accidental deletion
  - Attempting to delete a protected cluster will result in a clear error message
  - To delete a protected cluster, users must first set `deletion_protection = false` and apply the change
  - The parameter is optional and defaults to false to maintain backward compatibility

  ### Testing

  - Added `TestAccMSKCluster_deletionProtection` test that validates both enabling and disabling deletion protection
  - Test includes proper import state verification
  - Follows the same test pattern established by `TestAccRDSCluster_deletionProtection`

  Closes #44209

### DIFF
--- a/.changelog/44233.txt
+++ b/.changelog/44233.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_msk_cluster: Add `deletion_protection` argument
+```
--- a/internal/service/kafka/cluster.go
+++ b/internal/service/kafka/cluster.go
@@ -528,6 +528,10 @@ func resourceCluster() *schema.Resource {
 			},
 			names.AttrTags:    tftags.TagsSchema(),
 			names.AttrTagsAll: tftags.TagsSchemaComputed(),
+			names.AttrDeletionProtection: {
+				Type:     schema.TypeBool,
+				Optional: true,
+			},
 			"zookeeper_connect_string": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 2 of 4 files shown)


allebacco:
Hi, could you clarify what is the difference between your proposal and setting the following, please?
```hcl
lifecycle {
    prevent_destroy = true
  }
```

dialupdisaster:
> Hi, could you clarify what is the difference between your proposal and setting the following, please?
> 
> ```hcl
> lifecycle {
>     prevent_destroy = true
>   }
> ```

You can not use a variable with the lifecycle block, so it's either always on or always off. I will add that to the description

dialupdisaster:
@justinretzolk Not to be impatient, but what's the typical turnaround for simple contributions like this?

dialupdisaster:
marco?

PR #44276, opened 2025-09-14
[Docs] aws_cloudfront_distribution: Modernize example and add related resources

tabito-hara:
## Rollback Plan

### Description
Update the `aws_cloudfront_distribution` resource documentation example to modern standards.

* Use a bucket policy instead of a bucket ACL.
* Add an Origin Access Control (OAC) resource (`aws_cloudfront_origin_access_control`).
* Remove legacy logging configuration.
  * Standard logging configuration (v2) is explained in other examples.
* Add arguments related to ACM.
  * If `aliases` is specified, a TLS certificate is required. Arguments and a data source related to ACM are added.
* Add Route 53 resources to register aliases.
  * If `aliases` is specified, they must be registered as DNS records. An example of adding these records to Route 53 is included.

### References
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html


jar-b:
Thanks for your contribution, @tabito-hara! 👍

PR #44278, opened 2025-09-15
make new reference for requestInspection for ACFP to differentiate th…

megashby:
…e 2 data types

Both `awsManagedRulesAcfpRuleSet` and `awsManagedRulesAtpRuleSet` have an attribute `requestInspection`, however the data types are not the same.

according to the AWS docs: https://docs.aws.amazon.com/waf/latest/APIReference/API_RequestInspection.html & https://docs.aws.amazon.com/waf/latest/APIReference/API_RequestInspectionACFP.html

and the providers schema: https://github.com/hashicorp/terraform-provider-aws/blob/a789cbec225b5a7c5d54154c55b588fb7c54102f/internal/service/wafv2/schemas.go#L1562 & https://github.com/hashicorp/terraform-provider-aws/blob/a789cbec225b5a7c5d54154c55b588fb7c54102f/internal/service/wafv2/schemas.go#L1464

However, the docs had both blocks pointing to the same `requestInspection` block, which was actually reflective only of ACFP.

## Rollback Plan

### Description
To align with AWS docs, I have updated the `requestInspection` block to match the ATP pattern, and created a new block `requestInspectionAcfp` to match the ACFP pattern with extended fields

### DIFF
--- a/website/docs/r/wafv2_web_acl.html.markdown
+++ b/website/docs/r/wafv2_web_acl.html.markdown
@@ -838,8 +838,17 @@ The `managed_rule_group_configs` block support the following arguments:
 * `creation_path` - (Required) The path of the account creation endpoint for your application. This is the page on your website that accepts the completed registration form for a new user. This page must accept POST requests.
 * `enable_regex_in_path` - (Optional) Whether or not to allow the use of regular expressions in the login page path.
 * `registration_page_path` - (Required) The path of the account registration endpoint for your application. This is the page on your website that presents the registration form to new users. This page must accept GET text/html requests.
-* `request_inspection` - (Optional) The criteria for inspecting login requests, used by the ATP rule group to validate credentials usage. See [`request_inspection`](#request_inspection-block) for more details.
+* `request_inspection` - (Optional) The criteria for inspecting login requests, used by the ATP rule group to validate credentials usage. See [`request_inspection_acfp`](#request_inspection_acfp-block) for more details.
 * `response_inspection` - (Optional) The criteria for inspecting responses to login requests, used by the ATP rule group to track login failure rates. Note that Response Inspection is available only on web ACLs that protect CloudFront distributions. See [`response_inspection`](#response_inspection-block) for more details.
+
+### `request_inspection_acfp` Block
+
+* `addressFields` (Optional) The names of the fields in the request payload that contain your customer's primary physical address. See [`addressFields`](#address_fields-block) for more details.
+* `emailField` (Optional) The name of the field in the request payload that contains your customer's email. See [`emailField`](#email_field-block) for more details.
+* `passwordField` (Optional) Details about your login page password field. See [`passwordField`](#password_field-block) for more details.
+* `payloadType` (Required) The payload type for your login endpoint, either JSON or form encoded.
+* `phoneNumberFields` (Optional) The names of the fields in the request payload that contain your customer's primary phone number. See [`phoneNumberFields`](#phone_number_fields-block) for more details.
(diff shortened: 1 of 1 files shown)


justinretzolk (dismissed):
Hey @megashby 👋 Thank you for taking the time to raise this! The CDKTF documentation is generated based on the resource documentation, and shouldn't be edited directly. You'll instead want to edit the related file in the `website/docs/r` directory. We discuss this a bit more in the [End User Documentation Changes](https://hashicorp.github.io/terraform-provider-aws/documentation-changes/) section of the contribution guide.

Can you update the pull request with that in mind?

megashby:
> Hey @megashby 👋 Thank you for taking the time to raise this! The CDKTF documentation is generated based on the resource documentation, and shouldn't be edited directly. You'll instead want to edit the related file in the `website/docs/r` directory. We discuss this a bit more in the [End User Documentation Changes](https://hashicorp.github.io/terraform-provider-aws/documentation-changes/) section of the contribution guide.
> 
> Can you update the pull request with that in mind?

updated!

website/docs/r/wafv2_web_acl.html.markdown
```diff
@@ -838,9 +838,18 @@ The `managed_rule_group_configs` block support the following arguments:
...
 * `enable_regex_in_path` - (Optional) Whether or not to allow the use of regular expressions in the login page path.
 * `registration_page_path` - (Required) The path of the account registration endpoint for your application. This is the page on your website that presents the registration form to new users. This page must accept GET text/html requests.
-* `request_inspection` - (Optional) The criteria for inspecting login requests, used by the ATP rule group to validate credentials usage. See [`request_inspection`](#request_inspection-block) for more details.
+* `request_inspection` - (Optional) The criteria for inspecting login requests, used by the ATP rule group to validate credentials usage. See [`request_inspection_acfp`](#request_inspection_acfp-block) for more details.
 * `response_inspection` - (Optional) The criteria for inspecting responses to login requests, used by the ATP rule group to track login failure rates. Note that Response Inspection is available only on web ACLs that protect CloudFront distributions. See [`response_inspection`](#response_inspection-block) for more details.
 
+### `request_inspection_acfp` Block
```
jar-b: The argument name is still `request_inspection` in both cases, so perhaps the new heading can change to something like the following which does not imply an _acfp suffix on the argument name itself.

```suggestion
### `request_inspection` Block (ACFP)
```

megashby: Hey Jared, 

A question on this if you don't mind - if I do this, how would this interpretation:

```
* `request_inspection` - (Optional) The criteria for inspecting login requests, used by the ATP rule group to validate credentials usage. See [`request_inspection_acfp`](#request_inspection_acfp-block) for more details.
```
know to reference this new block? I was under the assumption that if I change the block title, the `#request_inspection_acfp-block` won't work anymore

jar-b: Yes, the anchor link will also need to change (I should have used a text block instead of a suggestion since it has cascading effects). 

I _think_ it will become `#request_inspection-block-acfp` as the parenthesis get stripped out, but I'm not an expert on how Github markdown works here. If you have tools installed (`make tools`) you can run the website linters locally to verify, otherwise CI will pick up any issues.

megashby: ah ok, updated, you were right btw: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links

```
Spaces are replaced by hyphens (-). Any other whitespace or punctuation characters are removed.
```

jar-b:
Thanks for the correction, @megashby! 👍

PR #44312, opened 2025-09-16
Update `skip_matching` documentation for `aws_autoscaling_group`

### LINKED ISSUE #44292: Incorrect documentation for aws_autoscaling_group and skip_matching.
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#skip_matching-1

### Description

The documentations says:

> Replace instances that already have your desired configuration. Defaults to false.

This seems to be the exact opposite of the value, according to AWS docs

> When you specify a new launch template or a new version of the current launch template for your desired configuration, consider enabling the SkipMatching property in preferences. If it’s enabled, Amazon EC2 Auto Scaling skips replacing instances that already use the specified launch template and instance types. This can help you reduce the number of replacements that are required to apply updates.

### References

https://docs.aws.amazon.com/cli/latest/reference/autoscaling/start-instance-refresh.html

No

justinretzolk:
## Rollback Plan

N/a

### Description

Fixes the details for the `skip_matching` argument of the `aws_autoscaling_group` resource.

### Relations

Closes #44292

### References

[AWS Documentation](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/start-instance-refresh.html):

> Indicates whether skip matching is enabled. If enabled (`true`), then Amazon EC2 Auto Scaling skips replacing instances that match the desired configuration. If no desired configuration is specified, then it skips replacing instances that have the same launch template and instance types that the Auto Scaling group was using before the start of the instance refresh. The default is `false`.

N/a, docs


jar-b (approved):
LGTM 🚀

PR #44335, opened 2025-09-18
sweeper: aws_athena_database

gdavison:
## Rollback Plan

### Description

Prevents infinite loop on database listing error


jar-b (approved):
LGTM 🚀

PR #44362, opened 2025-09-19
Fix `aws_vpc` reading incorrect `ipv6_cidr_block` if it has multiple attached.

jwhitaker-gridcog:
## Rollback Plan

None

### Description

Currently, when importing an `aws_vpc`, with multiple IPv6 CIDRs, the wrong cidr gets picked for `ipv6_cidr_block` and `ipv 6_cidr_block_association`.

The code to pick the default ipv6 cidr when reading a VPC has a comment that says "pick the first", but then actually goes and picks the last.

First commit is to export the relevant function so I can hit it with a unit test.
Second commit changes the behaviour to actually pick the first.

I tried to find doco from AWS confirming that the first is indeed the default and came up empty, but this seems to be true for our VPCs now.

```console

...
```


ewbankkit:
@jwhitaker-gridcog Thanks for the contribution :tada: :clap:.

jwhitaker-gridcog:
thanks guys!

Issue #43250, opened 2025-07-02
List correct values for


EugenKon:
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution#origin_ssl_protocols-1

### Description

```
│ Error: expected minimum_protocol_version to be one of ["SSLv3" "TLSv1" "TLSv1_2016" "TLSv1.1_2016" "TLSv1.2_2018" "TLSv1.2_2019" "TLSv1.2_2021"], got TLSv1.2
│
│   with aws_cloudfront_distribution.survey-simulator-cloudfront-net,
│   on cloudfront.tf line 36, in resource "aws_cloudfront_distribution" "survey-simulator-cloudfront-net":
│   36:     minimum_protocol_version       = "TLSv1.2"

```

### Would you like to implement a fix?

No

bradleyGamiMarques:
Hello 👋 
I want to make sure I understand the ask. 
Instead of:
> See all possible values in [this](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html) table under "Security policy." Some examples include: TLSv1.2_2019 and TLSv1.2_2021.

Something like this:
> Valid values: `SSLv3`, `TLSv1`, `TLSv1_2016`… etc.

@EugenKon

PR #44382, opened 2025-09-19
fix: s/TLSv1.2/TLSv1.2_2021/g

chenrui333:
per [this table](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html), `TLSv1.2` is not a valid TLS protocol, thus updating it to `TLSv1.2_2021`

image

relates to https://github.com/hashicorp/terraform-provider-aws/issues/43250


chenrui333:
@justinretzolk do I need to add a changelog entry?

chenrui333:
this PR is wrong, closing it.

PR #44385, opened 2025-09-20
Fix grammar, formatting, and a command on EKS add-on page

davidxia:
Signed-off-by: David Xia <david@davidxia.com>


jar-b:
Thank you for the cleanup, @davidxia! 👍

PR #44451, opened 2025-09-25
34787 Make sqs wait times configurable 34787

### LINKED ISSUE #34787: [Enhancement]: Reduce SQS verification
### Description

Hi!
Recently we started to investigate at Localstack why our SQS implementation hangs for unreasonably long with Terraform and we came to the conclusion that the provider hangs for some time before allowing the resource to be added to the state file. Same happens during delete but with less wait time.
We understand this is not an issue with AWS as the queue creation takes time, but in case of local testing/development slows the process significantly.

Related code:
https://github.com/hashicorp/terraform-provider-aws/blob/017bbcd20d61f2b002c4bdede386d0d5ea2d90ba/internal/service/sqs/queue.go#L577
and
https://github.com/hashicorp/terraform-provider-aws/blob/017bbcd20d61f2b002c4bdede386d0d5ea2d90ba/internal/service/sqs/queue.go#L593

We're wondering if it would be possible to reduce this value or customise it somehow via cli?

* aws_sqs_queue

```terraform
terraform {
  required_providers {
    aws = {
      source = "hashicorp/aws"
      version = "5.29.0"
    }
  }
}

provider "aws" {
  # Configuration options
}

resource "aws_sqs_queue" "terraform_queue" {
}
```

### References

https://docs.localstack.cloud/user-guide/aws/sqs/

No

JackMcBride98:
## Rollback Plan

N/A

N/A

### Description
Make sqs wait times configurable for the benefit of increasing local development speed. See #34787 

### Relations

Closes #34787


srsholmes:
Great work, thanks!

leonaves:
Amazing, would love to see this merged!

leonaves (approved):
LGTM!

srsholmes (approved):

bert-huang-halter:
Hey guys; thanks for taking the effort for making this work!
Possible to get this in soon? cheers

bert-huang-halter:
Hey @ewbankkit @YakDriver sorry for the tagging you guys, I'm unsure who to tag to gain traction for a PR buried away.. let me know if there is a better team to tag.

Could either one of you help with reviewing this one please?

niloyc (approved):
Great stuff!

bert-huang-halter (approved):

fabito (approved):

YakDriver:
Thanks for putting this together, and for tying it back to #34787. The local-dev pain is real and it's a popular ask, so it's useful to have a concrete proposal to talk against.

Some background on what those numbers are doing, since it shapes the design. The `ContinuousTargetOccurence` values (6 on create, 15 on delete) aren't really there to catch a value that flip-flops on the SQS endpoint itself. In practice they act as a wall-clock budget for propagation that happens out of band: you can poll the SQS endpoint and get several clean, stable reads while another service that depends on the queue still can't see it yet. The path a dependent resource takes to the queue isn't the same one we're polling. That's the tricky part: the thing the wait is compensating for isn't observable from where the waiter sits, which is why the counts were tuned conservatively and carry the "avoid lowering" notes. Dropping them globally tends to resurface as occasional spurious diffs and cross-service races that are hard to reproduce.

That's the core concern with exposing `create/delete_continuous_target_occurrence` as provider config: it turns an internal retry knob into permanent public API, and it asks users to pick a number for a mechanism that's fairly opaque even to us. Provider schema is effectively forever, and a per-service `*_wait_times` block sets a precedent every service would then reach for.

A few smaller things in the implementation:

- Passing `meta any` into `waitQueueAttributesPropagated` / `waitQueueDeleted` and type-asserting back to `*conns.AWSClient` inside the waiter couples the low-level wait to the client. Those functions already take `conn`; the resolved value should arrive as a plain typed argument instead.
- The defaults live in two places now (the schema `Default`, and the hardcoded `6`/`15` fallbacks in the waiters), and the "accommodate GovCloud/commercial/China, avoid lowering" comment got dropped from the occurrence line.
- The changelog file is named after the issue (34787) rather than the PR.

On direction: the part that's genuinely safe to change is narrow. About the only environment where collapsing these waits is correct is one with no cross-service propagation at all, like an in-process emulator, and that's something the operator knows but the provider can't detect. So a single opt-in signal that says "assume no propagation delay in this environment," kept out of the provider schema and off by default, is a much smaller and more defensible surface than per-operation tunables, and it leaves the real-AWS behavior untouched.

Appreciate the contribution and the patience here. Wanted to lay out the reasoning and the constraints so it's clear where the friction is.

PR #44478, opened 2025-09-29
fix: Corrects grammar in EMR and KMS validation messages

BenjaminNolan:
This PR fixes an admittedly minor grammar error in the validation messages for the EMR and KMS services.

`be comprised of only` -> `only contain`

## Rollback Plan

I don't think this needs one, as it's just a message fix. The actual text also isn't checked for in the tests.

None.

### DIFF
--- a/.changelog/44478.txt
+++ b/.changelog/44478.txt
@@ -1,4 +1,0 @@
-```release-note:note
-resource/aws_emr_service: Updates validation messages from `be comprised of only` to `only contain`
-resource/aws_kms_service: Updates validation messages from `be comprised of only` to `only contain`
-```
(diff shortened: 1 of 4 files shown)


BenjaminNolan:
Hi @justinretzolk. That's a fair point, I've updated the wording to `only contain`, which is consistent with the messages already in the `kms` module. :)

justinretzolk:
Glad you're on board @BenjaminNolan! I've also added the `no-changelog-needed` label here to stop the checks from yelling at you. In this case, since we're really just changing some wording, we won't need to add CHANGELOG entries.

BenjaminNolan:
Changelog is ~~beyoten~~ deleted. :)

PR #44542, opened 2025-10-03
[bug] Fix inconsistent final plan for tgw association/propagation

YakDriver:
## Rollback Plan

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/internal/service/ec2/transitgateway_route_table_propagation.go
+++ b/internal/service/ec2/transitgateway_route_table_propagation.go
@@ -13,7 +13,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/hashicorp/aws-sdk-go-base/v2/tfawserr"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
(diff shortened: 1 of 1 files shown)


YakDriver:
### TL;DR

We discussed long-standing inconsistencies in how Direct Connect and Transit Gateway attachment dependencies are handled. The consensus was to simplify behavior, remove legacy workarounds, and rely on the attachment ID being available directly from the resource rather than through a data source chain.

### Discussion Summary

* The original `CustomizeDiff` logic was a workaround to prevent resource re-creation when attachment IDs were not yet exposed.
* All current attachment resources now expose their `attachment_id` after creation, making that workaround obsolete.
* Removing `CustomizeDiff` aligns behavior with other attachment-related resources and avoids hidden dependencies like:
  ```
  resource → data source → resource
  ```
* With this change, `attachment_id` attributes on both association and propagation resources will properly trigger replacement when needed, removing the need for customized diffs.
* The team verified there are no remaining attachment types lacking `attachment_id` outputs.
* Documentation warnings were added to help users migrate away from patterns relying on the old data source approach, which could otherwise lead to temporary loss of routing during `apply`.
* The group agreed to move forward, noting that the new behavior reduces inconsistent plan states and that the number of users negatively affected should be minimal.

**Lesson learned:** when resource chaining is needed, prefer exposing computed attributes (e.g., IDs) directly on the source resource instead of forcing users through data sources.

PR #44543, opened 2025-10-04
Avoid using d.Get("task") for each task in appflow_flow

MatthewJohn:
## Rollback Plan

A new attribute has been added, but believe this can _just_ be reverted.

No

### Description

The original implementation called d.Get("task") for each task's source_fields attribute. As a result, for each task, d.Get("task") is called - which turns out to be expensive. For example for ~350 tasks, this seemed to take about 15 minutes. Profiling for this original behavoir can be seen in the PR.

This change introduces a new single_task_flag attribute, which is computed once during the read operation and this is used to determine if a single task is present (which is what the original result of d.Get("task") was used for. From local testing of the use-case, this completely fixes the slow-down

### Relations

Closes #0000

```console

...
```


MatthewJohn:
Flame graph from original:
go_profiling_flamegraph

PR #44564, opened 2025-10-07
Docs: Clarify deprecations in `aws_service_discovery_service`

RafaelWO:
## Rollback Plan

n/a

### Description

This change adapts the example usage to use the non-deprecated block `health_check_config` instead of `health_check_custom_config`. 
Additionally, a note regarding the deprecation of `health_check_custom_config` is added.

### Relations

Closes #0000

n/a

### DIFF
--- a/website/docs/r/service_discovery_service.html.markdown
+++ b/website/docs/r/service_discovery_service.html.markdown
@@ -39,7 +39,7 @@ resource "aws_service_discovery_service" "example" {
     routing_policy = "MULTIVALUE"
   }
 
-  health_check_custom_config {
+  health_check_config {
     failure_threshold = 1
   }
 }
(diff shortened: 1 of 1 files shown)


website/docs/r/service_discovery_service.html.markdown
```diff
@@ -111,6 +111,8 @@ The `health_check_config` configuration block supports the following arguments:
 
 ### `health_check_custom_config` Block
 
+~> The `health_check_custom_config` argument is deprecated. Use [`health_check_custom_config`](#health_check_config-block) instead, which supports additional attributes.
```
ewbankkit:
```suggestion
~> The `health_check_custom_config` argument is deprecated. Use [`health_check_config`](#health_check_config-block) instead, which supports additional attributes.
```

RafaelWO: Good catch!

ewbankkit:
@RafaelWO Thanks for the contribution :tada: :clap:.

PR #44575, opened 2025-10-07
docs(rds): add example + note for blue_green_update block on aws_db_instance (#44488)

### LINKED ISSUE #44488: Error: Unsupported argument: blue_green_update
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#blue_green_update

### Description

I added the next block into my TF config:
```
  blue_green_update = {
    enabled = true
  }
```
But I got this error:
```
│ Error: Unsupported argument
│
│   on modules/private-cloud/replica-rds.tf line 50, in resource "aws_db_instance" "rpostgres_v18":
│   50:   blue_green_update = {
│
│ An argument named "blue_green_update" is not expected here. Did you mean to define a block of type
│ "blue_green_update"?
╵
```
It seems I did something wrong, but documentation has no examples.

### References

_No response_

No

GabrielOvie:
## What
Add a small HCL example for `blue_green_update` on `aws_db_instance` and a short note that it is a **block** (not an argument).

## Why
Users often hit an “Unsupported argument” error when using `blue_green_update = { ... }`. This shows the correct block syntax and calls it out.

## Linked issue
Fixes #44488

## Scope
Docs-only change.

### DIFF
--- a/website/docs/r/db_instance.html.markdown
+++ b/website/docs/r/db_instance.html.markdown
@@ -474,6 +474,29 @@ This will not recreate the resource if the S3 object changes in some way.  It's 
 
 * `enabled` - (Optional) Enables [low-downtime updates](#low-downtime-updates) when `true`.
   Default is `false`.
+#### Example
+
+```hcl
+resource "aws_db_instance" "example" {
+  identifier          = "example"
+  engine              = "postgres"
+  engine_version      = "15"
+  instance_class      = "db.t3.micro"
+  allocated_storage   = 20
+  username            = "admin"
+  password            = "example-password"
+  skip_final_snapshot = true
+
+  # Blue/Green updates are set with a BLOCK (not an argument).
+  blue_green_update {
+    enabled = true
+  }
+}
+``` 
+> **Note:** `blue_green_update` is a **block**, not an argument.  
+> Using `blue_green_update = { ... }` will cause an error like:  
+> *"An argument named `blue_green_update` is not expected here. Did you mean to define a block of type `blue_green_update`?"*
+
 
 [instance-replication]:
 https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Replication.html


website/docs/r/db_instance.html.markdown
```diff
@@ -474,6 +474,29 @@ This will not recreate the resource if the S3 object changes in some way.  It's
...
+  username            = "admin"
+  password            = "example-password"
+  skip_final_snapshot = true
+
+  # Blue/Green updates are set with a BLOCK (not an argument).
+  blue_green_update {
+    enabled = true
+  }
+}
+```
```
justinretzolk, line 495: We'll need to move this example up to the Examples section of the document. I'd say around line 276 should do the trick.

Note that when using `blue_green_update`, the `backup_retention_period` argument must be set to a value greater than `0`, so that will need to be added to the sample configuration in order to ensure its validity.

website/docs/r/db_instance.html.markdown
```diff
@@ -474,6 +474,29 @@ This will not recreate the resource if the S3 object changes in some way.  It's
...
+  identifier          = "example"
+  engine              = "postgres"
+  engine_version      = "15"
+  instance_class      = "db.t3.micro"
+  allocated_storage   = 20
+  username            = "admin"
+  password            = "example-password"
+  skip_final_snapshot = true
+
+  # Blue/Green updates are set with a BLOCK (not an argument).
```
justinretzolk, line 490: We can omit this comment. The block vs argument syntax is a Terraform Core level concept, and not something we call out in the documentation, particularly because Terraform will throw an error and let you know if you've supplied it as an argument rather than a block.

website/docs/r/db_instance.html.markdown
```diff
@@ -474,6 +474,29 @@ This will not recreate the resource if the S3 object changes in some way.  It's
...
+  blue_green_update {
+    enabled = true
+  }
+}
+``` 
+> **Note:** `blue_green_update` is a **block**, not an argument.  
+> Using `blue_green_update = { ... }` will cause an error like:  
+> *"An argument named `blue_green_update` is not expected here. Did you mean to define a block of type `blue_green_update`?"*
+
```
justinretzolk, line 499: See comment above re: adding details of block vs. argument to the documentation

justinretzolk (changes requested):
Hey @GabrielOvie 👋🏻 Thank you for taking the time to raise this! I added a few comments to try to polish this up before we merge. Once those changes have been made, I'd be happy to review again.

alexlux58:
Hi, I noticed this PR is currently waiting on requested changes.

I am looking for a small first contribution and would be happy to help address the maintainer feedback if the original author is no longer actively working on it.

I do not want to duplicate work, so I will hold off unless the maintainers or author are comfortable with me picking it up.

PR #44641, opened 2025-10-13
chore: update release notification channel

jar-b:
### Description

The channel ID has updated following migration to IBM slack.


johnsonaj:
we should also update the release action: https://github.com/hashicorp/terraform-provider-aws/blob/4b32fdf68b3bedce07c703265402c8e2e71622e1/.github/workflows/provider-release.yml#L37

PR #44059, opened 2025-08-27
[Enhancement] aws_ec2_client_vpn_endpoint: Support IPv6 connectivity

tabito-hara:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

* This PR adds the `endpoint_ip_address_type` and `traffic_ip_address_type` arguments to the `aws_ec2_client_vpn_endpoint` resource to support IPv6 connectivity, and makes the currently required `client_cidr_block` optional.

  * Because the update API does not accept `endpoint_ip_address_type` and `traffic_ip_address_type`, they are marked as `ForceNew`.
  * Because the read API returns `endpoint_ip_address_type` and `traffic_ip_address_type`, they are marked as `Computed`.

* The implementation corresponds to the related [SDK update](https://github.com/aws/aws-sdk-go-v2/commit/3b73a3be8423cd3e099e3754830ebeefb5518afe#diff-14b62e8dd1e4586aa283c66c14d53d00ed2b697098959be96aea1644d92d5ae3)

  * The following fields have been added to the `CreateClientVpnEndpointInput` type:

    * `EndpointIpAddressType`
    * `TrafficIpAddressType`
  * `ClientCidrBlock` has been changed from `Required` to `Optional`.

    * `ClientCidrBlock` must not be specified when `TrafficIpAddressType` is set to `ipv6`.

* The `endpoint_ip_address_type` and `traffic_ip_address_type` attributes have also been added to the `aws_ec2_client_vpn_endpoint` data source.

### Relations

Closes #44052

### References
Related SDK updates:
https://github.com/aws/aws-sdk-go-v2/commit/3b73a3be8423cd3e099e3754830ebeefb5518afe#diff-14b62e8dd1e4586aa283c66c14d53d00ed2b697098959be96aea1644d92d5ae3

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateClientVpnEndpoint.html

### Output from Acceptance Testing

```console
$ make testacc TESTS=TestAccClientVPNEndpoint_serial/Endpoint_ PKG=ec2     
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go1.24.6 test ./internal/service/ec2/... -v -count 1 -parallel 20 -run='TestAccClientVPNEndpoint_serial/Endpoint_'  -timeout 360m -vet=off
[...]

PR #44926, opened 2025-11-04
resource/aws_ec2_client_vpn_route: Allow IPv6 address ranges

A1bi:
### Description

AWS has recently introduced full [IPv6 support](https://docs.aws.amazon.com/vpn/latest/clientvpn-admin/ipv6-considerations.html) for their Client VPN. #44059 already added the necessary changes for endpoints. What's missing is routes and authorization rules (I only had to update the documentation for the latter). Apparently the [AWS documentation](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#CreateClientVpnRouteInput) hasn't been updated yet accordingly (still mentions IPv4 only), but you can definitely now create both with IPv6 prefixes.

### DIFF
--- a/.changelog/44926.txt
+++ b/.changelog/44926.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_ec2_client_vpn_route: Allow IPv6 address ranges for `destination_cidr_block`
+```


ewbankkit:
@A1bi Thanks for the contribution :tada: :clap:.

PR #44971, opened 2025-11-06
tag test generation: Sets import ID attribute from ARN Identity

gdavison:
## Rollback Plan

### Description

Sets import ID attribute from ARN Identity in tag test generation


ewbankkit (approved):
LGTM :rocket:.

PR #44999, opened 2025-11-08
new aws_dynamodb_global_secondary_index resource

### LINKED ISSUE #671: No method to ignore changes in DynamoDB GSI
_This issue was originally opened by @cypai as hashicorp/terraform#13393. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

### Terraform Version
$ terraform -v
Terraform v0.8.8

### Affected Resource(s)
- aws_dynamodb_table

```hcl
resource "aws_dynamodb_table" "FooBar" {
    name = "FooBar"
    read_capacity = 10
    write_capacity = 10
    hash_key = "foo-id"
    attribute {
        name = "foo-id"
        type = "S"
    }
    attribute {
        name = "bar-id"
        type = "N"
    }
    global_secondary_index {
        name = "bar-index"
        hash_key = "bar-id"
        read_capacity = 10
        write_capacity = 10
        projection_type = "ALL"
    }
    lifecycle {
        ignore_changes = ["read_capacity", "write_capacity"]
    }
}
```

### References
- hashicorp/terraform#5617 

We would like to ignore read/write capacity for the GSI as well, but this configuration only ignores the read/write capacity of the top-level table. We weren't able to find any usable workaround, since we can't interpolate variables in the lifecycle, and wildcards are not supported.

Technically we could hardcode the randomly generated number like ```ignore_changes = ["global_secondary_index.3291674533.read_capacity"]```, but that's obviously a terrible solution since it would change every time we make a change.

raduendv:
### Description

Introduces a new optional resources `aws_dynamodb_global_resource_index` which allows users to use `lifecycle.ignore_changes` for properties like `read_capacity` and `write_capacity`.

Adoption is optional, old `aws_dynamodb_table` resource works the same. The new resource does not work together with the old `global_secondary_index` block, one must be chosen over the other.

Users migrating to this new resource must import existing indexes into new resource blocks.  Users that change their mind can remove the new resources from state and revert back to old config.

### Relations

Closes #671 

```console
% make testacc TESTS=TestAccDynamoDBGlobalSecondaryIndex_ PKG=dynamodb

make: Running acceptance tests on branch: 🌿 f-aws_dynamodb_global_resource_index-new-resource 🌿...
TF_ACC=1 /Users/radugribincea/.gvm/gos/go1.24.6/bin/go test ./internal/service/dynamodb/... -v -count 1 -parallel 20 -run='TestAccDynamoDBGlobalSecondaryIndex_'  -timeout 360m -vet=off
2025/11/08 01:59:32 Creating Terraform AWS Provider (SDKv2-style)...
2025/11/08 01:59:32 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_basic
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_basic
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange_ignoreChanges
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange_ignoreChanges
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange_ignoreChanges
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange_ignoreChanges
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_attributeValidation
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_attributeValidation
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_payPerRequest_to_provisioned
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_payPerRequest_to_provisioned
=== RUN   TestAccDynamoDBGlobalSecondaryIndex_provisioned_to_payPerRequest
=== PAUSE TestAccDynamoDBGlobalSecondaryIndex_provisioned_to_payPerRequest
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_basic
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_payPerRequest_to_provisioned
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange_ignoreChanges
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_attributeValidation
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange_ignoreChanges
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_provisioned_to_payPerRequest
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput
=== CONT  TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_attributeValidation (60.12s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput (416.30s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest (416.55s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_basic (428.00s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange (455.81s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_capacityChange_ignoreChanges (461.51s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange (461.52s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_billingPayPerRequest_onDemandThroughput_capacityChange_ignoreChanges (462.98s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_payPerRequest_to_provisioned (558.47s)
--- PASS: TestAccDynamoDBGlobalSecondaryIndex_provisioned_to_payPerRequest (634.68s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/dynamodb	643.506s
```


internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+	"github.com/hashicorp/terraform-provider-aws/internal/flex"
+	"github.com/hashicorp/terraform-provider-aws/internal/slices"
+	"github.com/hashicorp/terraform-provider-aws/names"
+)
+
+const (
+	resNameGlobalSecondaryIndex = "global_secondary_index"
+)
+
+// @SDKResource("aws_dynamodb_global_secondary_index", name="Global Secondary Index")
```
gdavison: All new resource types must be implemented using [Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework) instead of [Plugin SDK v2](https://developer.hashicorp.com/terraform/plugin/sdkv2).

The resource implementation can be partially converted using the tool `tfsdk2fw` found at https://github.com/hashicorp/terraform-provider-aws/tree/main/tools/tfsdk2fw

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	awstypes "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"github.com/dlclark/regexp2"
```
gdavison: We prefer the use of `github.com/YakDriver/regexache`, which caches compiled regexes

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+	"github.com/hashicorp/terraform-provider-aws/internal/create"
+	"github.com/hashicorp/terraform-provider-aws/internal/enum"
+	"github.com/hashicorp/terraform-provider-aws/internal/errs"
+	"github.com/hashicorp/terraform-provider-aws/internal/flex"
+	"github.com/hashicorp/terraform-provider-aws/internal/slices"
+	"github.com/hashicorp/terraform-provider-aws/names"
+)
+
+const (
+	resNameGlobalSecondaryIndex = "global_secondary_index"
```
gdavison: This should be a display type name, not the resource type name, e.g.

```suggestion
	resNameGlobalSecondaryIndex = "Global Secondary Index"
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,2248 @@
...
+					resource.TestCheckResourceAttr(resourceNameTable, "server_side_encryption.#", "0"),
+					resource.TestCheckResourceAttr(resourceNameTable, names.AttrStreamARN, ""),
+					resource.TestCheckResourceAttr(resourceNameTable, "stream_enabled", acctest.CtFalse),
+					resource.TestCheckResourceAttr(resourceNameTable, "stream_label", ""),
+					resource.TestCheckResourceAttr(resourceNameTable, "stream_view_type", ""),
+					resource.TestCheckResourceAttr(resourceNameTable, "table_class", "STANDARD"),
+					resource.TestCheckResourceAttr(resourceNameTable, acctest.CtTagsPercent, "0"),
+					resource.TestCheckResourceAttr(resourceNameTable, acctest.CtTagsAllPercent, "0"),
+					resource.TestCheckResourceAttr(resourceNameTable, "read_capacity", "1"),
+					resource.TestCheckResourceAttr(resourceNameTable, "write_capacity", "1"),
```
gdavison: Since this is an acceptance test for `aws_dynamodb_global_secondary_index`, the test checks should validate the values returned by the `aws_dynamodb_global_secondary_index` resource, not the `aws_dynamodb_table` resource

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,2248 @@
...
+					acctest.ImportCheckResourceAttr(names.AttrName, rNameGSI),
+					acctest.ImportCheckResourceAttr("table", rNameTable),
+					acctest.ImportCheckResourceAttr("hash_key", rNameTable),
+					acctest.ImportCheckResourceAttr("hash_key_type", "S"),
+					acctest.ImportCheckResourceAttr("range_key", rNameGSI),
+					acctest.ImportCheckResourceAttr("range_key_type", "S"),
+					acctest.ImportCheckResourceAttr("projection_type", "ALL"),
+					acctest.ImportCheckResourceAttr("read_capacity", "1"),
+					acctest.ImportCheckResourceAttr("write_capacity", "1"),
+				),
```
gdavison: `ImportStateCheck` is usually not needed, as `ImportStateVerify: true` will validate against the values from Step 1. If the results are expected to be different, add a comment explaining the expected change

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,2248 @@
...
+					acctest.ImportCheckResourceAttr("stream_enabled", acctest.CtFalse),
+					acctest.ImportCheckResourceAttr("stream_label", ""),
+					acctest.ImportCheckResourceAttr("stream_view_type", ""),
+					acctest.ImportCheckResourceAttr("table_class", "STANDARD"),
+					acctest.ImportCheckResourceAttr(acctest.CtTagsPercent, "0"),
+					acctest.ImportCheckResourceAttr(acctest.CtTagsAllPercent, "0"),
+					acctest.ImportCheckResourceAttr("read_capacity", "1"),
+					acctest.ImportCheckResourceAttr("write_capacity", "1"),
+				),
+			},
```
gdavison, line 96: It isn't necessary to check that the `aws_dynamodb_table` can be imported

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,2248 @@
...
+resource "aws_dynamodb_global_secondary_index" "test" {
+  table           = aws_dynamodb_table.test.name
+  name            = %[2]q
+  hash_key        = %[1]q
+  range_key       = %[2]q
+  read_capacity   = 1
+  write_capacity  = 1
+  range_key_type  = "S"
+  projection_type = "ALL"
+}
```
gdavison: Since the `aws_dynamodb_global_secondary_index` is the resource type being tested, move it to the top of the configuration

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+		},
+
+		SchemaVersion: 1,
+
+		Schema: map[string]*schema.Schema{
+			names.AttrIndexARN: {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"table": {
```
gdavison: This attribute should be `table_name` to match the underlying API

```suggestion
			"table_name": {
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(createTableTimeout),
+			Delete: schema.DefaultTimeout(deleteTableTimeout),
+			Update: schema.DefaultTimeout(updateTableTimeoutTotal),
+		},
+
+		SchemaVersion: 1,
+
+		Schema: map[string]*schema.Schema{
```
gdavison: The schema should be sorted alphabetically

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+			"write_capacity": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+		},
```
gdavison, line 213: The GSI schema is missing `warm_throughput`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+			"table": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"hash_key": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
```
gdavison: Rather than separate `hash_key` and `range_key` attributes, this should match the underlying API with `attribute_name` and `key_type`, where `key_type` can be one of `HASH` or `RANGE`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+				ForceNew: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"on_demand_throughput": onDemandThroughputSchema(),
+			"projection_type": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: enum.Validate[awstypes.ProjectionType](),
+			},
```
gdavison: To match the underlying API, `projection_type` and `non_key_attributes` should be moved into a nested block `projection`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+				AttributeType: awstypes.ScalarAttributeType(hashKeyType.(string)),
+			})
+			knownAttributes[hashKey] = awstypes.ScalarAttributeType(hashKeyType.(string))
+		} else {
+			diags = append(diags, errs.NewErrorDiagnostic(
+				`"hash_key_type" must be set in this configuration`,
+				`When using "hash_key" that is not defined in the attributes list of the table, you must specify the "hash_key_type"`,
+			))
+		}
+	}
```
gdavison: Table attribute management should be handled by the `aws_dynamodb_table`, rather than split between `aws_dynamodb_table` and `aws_dynamodb_global_secondary_index`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+	if utRes, err := conn.UpdateTable(ctx, ut); err != nil {
+		return create.AppendDiagError(diags, names.DynamoDB, create.ErrActionCreating, resNameGlobalSecondaryIndex, res.Id(), fmt.Errorf("creating GSI (%s): %w", res.Get(names.AttrName).(string), err))
+	} else {
+		for _, gsi := range utRes.TableDescription.GlobalSecondaryIndexes {
+			if aws.ToString(gsi.IndexName) == res.Get(names.AttrName).(string) {
+				if err := res.Set(names.AttrIndexARN, aws.ToString(gsi.IndexArn)); err != nil {
+					diags = append(diags, errs.NewErrorDiagnostic(fmt.Sprintf(`Unable to set field "%s"`, names.AttrIndexARN), fmt.Sprintf("Error: %v", err)))
+				}
+
+				res.SetId(aws.ToString(gsi.IndexArn))
```
gdavison: Once we've found the GSI, we can `break` out of the loop

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+	}
+
+	if len(diags) > 0 {
+		return diags
+	}
+
+	return diags
+}
+
+func waitAllGSIActive(ctx context.Context, conn *dynamodb.Client, tableName string, timeout time.Duration) error {
```
gdavison: To match the signature of the standard waiter functions in the provider, `waitAllGSIActive` should also return the underlying table. Especially because in all cases where `waitAllGSIActive` is called, it is immediately followed by `findTableByName`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,710 @@
...
+	if output, ok := outputRaw.(*awstypes.TableDescription); ok {
+		for _, g := range output.GlobalSecondaryIndexes {
+			if g.IndexStatus != awstypes.IndexStatusActive {
+				if _, err := waitGSIActive(ctx, conn, tableName, aws.ToString(g.IndexName), timeout); err != nil {
+					return err
+				}
+			}
+		}
+	}
+
```
gdavison: Rather than calling `statusTable` (and recreating `waitTableActive`), the `Refresh` function could handle checking the status of all of the GSIs on the table with a single call to `DescribeTable`. Currently, `DescribeTable` is called at least once for `stateConf.WaitForStateContext` and at least once for each GSI.

names/attr_constants.csv
```diff
@@ -10,6 +10,7 @@ application_id,ApplicationID
 apply_immediately,ApplyImmediately
 arn,ARN
 arns,ARNs
+index_arn,IndexARN
```
gdavison: This isn't strictly required, as the attribute name is rarely used

gdavison:
Also, the CHANGELOG entry file should be named after the PR instead of the Issue

internal/service/dynamodb/global_secondary_index.go
raduendv: unfortunately this is a technical limitation of the api

if we send attributes with the table update or create that are not part of any index or the table itself (range or hash key for the table) they are ignored/dropped by dynamodb

having them handled both in the table and gsi resource is a must if we want to comply with the api validations

raduendv:
@gdavison fixed the issues reported by semgrep, should be all good now

please re-run the pipeline

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+				},
+				Validators: []validator.List{
+					listvalidator.SizeAtMost(1),
+				},
+				PlanModifiers: []planmodifier.List{
+					listplanmodifier.UseStateForUnknown(),
+				},
+			},
+		},
+		Version: 1,
```
gdavison: `Version` is zero-indexed. Since this is a new resource type, the `Version` can be left unset so that it defaults to `0`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+	}
+
+	if s.Blocks == nil {
+		s.Blocks = make(map[string]schema.Block)
+	}
+	s.Blocks[names.AttrTimeouts] = timeouts.Block(ctx, timeouts.Opts{
+		Create: true,
+		Update: true,
+		Delete: true,
+	})
```
gdavison: The `timeouts` block can be defined inline in `Blocks` above

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/YakDriver/regexache"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	ddbtypes "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	gocmp "github.com/google/go-cmp/cmp"
```
gdavison: The package `github.com/google/go-cmp` should not be used here. According to the documentation, " It is intended to only be used in tests".

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+import (
+	"context"
+	"fmt"
+	"slices"
+	"time"
+
+	"github.com/YakDriver/regexache"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	ddbtypes "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
```
gdavison: By convention, we alias the package `github.com/aws/aws-sdk-go-v2/service/<whatever>/types` for the current service as `awstypes`.

```suggestion
	awstypes "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1139 @@
...
+					resource.TestCheckResourceAttr(resourceNameGSI, "write_capacity", "1"),
+				),
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue(resourceNameTable, tfjsonpath.New("ttl"), knownvalue.ListExact([]knownvalue.Check{
+						knownvalue.ObjectExact(map[string]knownvalue.Check{
+							"attribute_name":  knownvalue.StringExact(""),
+							names.AttrEnabled: knownvalue.Bool(false),
+						}),
+					})),
+				},
```
gdavison, line 88: Since this test is for `aws_dynamodb_global_secondary_index`, checking the `ttl` on the `aws_dynamodb_table` is not necessary

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+			"hash_key_type": schema.StringAttribute{
+				CustomType: fwtypes.StringEnumType[ddbtypes.ScalarAttributeType](),
+				Optional:   true,
+				Computed:   true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			names.AttrID: framework.IDAttribute(),
```
gdavison: Since the ARN value is used to look up the GSI, the attribute `id` should be replaced with the attribute `arn`. See https://hashicorp.github.io/terraform-provider-aws/id-attributes/#standard-usage for information on updating tests when a resource type does not define an `id` attribute.

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+						if old != nil {
+							slices.Sort(old)
+						}
+
+						if new != nil {
+							slices.Sort(new)
+						}
+
+						response.RequiresReplace = !gocmp.Equal(old, new)
+					}, "", ""),
```
gdavison: Would `setplanmodifier.RequiresReplace` or `setplanmodifier.RequiresReplaceIfConfigured` work here?

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+		tableName = grps[1]
+	} else {
+		tableName = data.Table.ValueString()
+	}
+
+	conn := r.Meta().DynamoDBClient(ctx)
+
+	table, err := findTableByName(ctx, conn, tableName)
+	if err != nil {
+		if tfresource.NotFound(err) {
```
gdavison: `tfresource.NotFound` is deprecated. It should be replaced with `github.com/hashicorp/terraform-provider-aws/internal/retry.NotFound`


```suggestion
		if retry.NotFound(err) {
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+
+	table, err := findTableByName(ctx, conn, data.Table.ValueString())
+	if err != nil {
+		response.Diagnostics.AddError(
+			fmt.Sprintf(`Unable to read table "%s"`, data.Table.ValueString()),
+			err.Error(),
+		)
+
+		return
+	}
```
gdavison: Since `waitAllGSIActive` returns the table, calling `findTableByName` isn't needed here


```suggestion
	if table, err := waitAllGSIActive(ctx, conn, data.Table.ValueString(), r.ReadTimeout(ctx, data.Timeouts)); err != nil {
		response.Diagnostics.AddError(
			fmt.Sprintf(`Error while waiting for all GSIs on table "%s" to be active`, data.Table.ValueString()),
			err.Error(),
		)

		return
	}

	createTimeout := r.CreateTimeout(ctx, data.Timeouts)
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+	var data resourceGlobalSecondaryIndexModel
+
+	response.Diagnostics.Append(request.Plan.Get(ctx, &data)...)
+
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	conn := r.Meta().DynamoDBClient(ctx)
+	if _, err := waitAllGSIActive(ctx, conn, data.Table.ValueString(), r.ReadTimeout(ctx, data.Timeouts)); err != nil {
```
gdavison: Is `waitAllGSIActive` sufficient to prevent concurrent changes to the table, or do we need some other synchronization mechanism?

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+	}
+
+	if !found {
+		response.Diagnostics.Append(
+			fwdiag.NewResourceNotFoundWarningDiagnostic(
+				fmt.Errorf(`unable to find global secondary index with id "%s", treating it as new`, data.ID.ValueString()),
+			),
+		)
+		response.State.RemoveResource(ctx)
+	}
```
gdavison: The function `findGSIByTwoPartKey` can be used to retrieve the GSI using the table name and index name

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1139 @@
...
+			return errors.New("table config is empty")
+		}
+
+		for _, g := range tbl.GlobalSecondaryIndexes {
+			if rs.Primary.Attributes[names.AttrName] == aws.ToString(g.IndexName) {
+				*gsi = g
+
+				break
+			}
+		}
```
gdavison: Look up the GSI using `findGSIByTwoPartKey`. You will likely have to export it by adding it to `exports_test.go`.

This function will then no longer need the parameter `tbl`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+		response.Diagnostics.Append(
+			data.NonKeyAttributes.ElementsAs(ctx, &projection.NonKeyAttributes, false)...,
+		)
+	}
+
+	action := &ddbtypes.CreateGlobalSecondaryIndexAction{
+		IndexName:  data.Name.ValueStringPointer(),
+		KeySchema:  keySchema,
+		Projection: projection,
+	}
```
gdavison: If the schema more closely matches the API structs, AutoFlex can be used to populate the API structs. See https://hashicorp.github.io/terraform-provider-aws/data-handling-and-conversion/#autoflex-for-terraform-plugin-framework-preferred

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+		)
+		response.State.RemoveResource(ctx)
+	}
+
+	if !data.WarmThroughputs.IsNull() && !data.WarmThroughputs.IsUnknown() {
+		var wt []warmThroughputModel
+
+		response.Diagnostics.Append(data.WarmThroughputs.ElementsAs(ctx, &wt, false)...)
+
+		if !response.Diagnostics.HasError() && len(wt) > 0 {
```
gdavison:
```suggestion
		response.Diagnostics.Append(data.WarmThroughputs.ElementsAs(ctx, &wt, false)...)
		if response.Diagnostics.HasError() {
            return
        }

        if len(wt) > 0 {
```

internal/service/dynamodb/global_secondary_index.go
gdavison: Even though `waitAllSGIActive` is technically a read call, it's part of the resource's Create operation, so we should use the `CreateTimeout` here


```suggestion
	if _, err := waitAllGSIActive(ctx, conn, data.Table.ValueString(), r.CreateTimeout(ctx, data.Timeouts)); err != nil {
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+			rups := wt[0].ReadUnitsPerSecond.ValueInt64()
+			wups := wt[0].WriteUnitsPerSecond.ValueInt64()
+
+			if (rups < warmThoughtPutMinReadUnitsPerSecond && wups < warmThoughtPutMinWriteUnitsPerSecond) || (rups == warmThoughtPutMinReadUnitsPerSecond && wups == warmThoughtPutMinWriteUnitsPerSecond) {
+				data.WarmThroughputs = fwtypes.NewListNestedObjectValueOfValueSliceMust[warmThroughputModel](ctx, []warmThroughputModel{})
+			}
+		}
+	}
```
gdavison: What does this section do? Please add a comment that explains it and add test that exercises the code. I commented it out and ran all tests, and it succeeded.

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+		var wt []warmThroughputModel
+
+		response.Diagnostics.Append(data.WarmThroughputs.ElementsAs(ctx, &wt, false)...)
+
+		if !response.Diagnostics.HasError() && len(wt) > 0 {
+			rups := wt[0].ReadUnitsPerSecond.ValueInt64()
+			wups := wt[0].WriteUnitsPerSecond.ValueInt64()
+
+			if (rups < warmThoughtPutMinReadUnitsPerSecond && wups < warmThoughtPutMinWriteUnitsPerSecond) || (rups == warmThoughtPutMinReadUnitsPerSecond && wups == warmThoughtPutMinWriteUnitsPerSecond) {
+				data.WarmThroughputs = fwtypes.NewListNestedObjectValueOfValueSliceMust[warmThroughputModel](ctx, []warmThroughputModel{})
```
gdavison: Specifying `warmThroughputModel` is redundant here


```suggestion
				data.WarmThroughputs = fwtypes.NewListNestedObjectValueOfValueSliceMust(ctx, []warmThroughputModel{})
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+			response.Diagnostics.AddError(
+				fmt.Sprintf(`Unable to read table "%s"`, data.Table.ValueString()),
+				err.Error(),
+			)
+		}
+
+		return
+	}
+
+	knownAttributes := map[string]int{}
```
gdavison: Do we need to get the counts for `table.KeySchema` as well?

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+				CustomType: fwtypes.StringEnumType[ddbtypes.ScalarAttributeType](),
+				Optional:   true,
+				Computed:   true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			names.AttrID: framework.IDAttribute(),
+			names.AttrName: schema.StringAttribute{
```
gdavison: To match the API schema and disambiguate the attribute, rename to `index_name`

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+		})
+	}
+
+	ut.GlobalSecondaryIndexUpdates = []ddbtypes.GlobalSecondaryIndexUpdate{
+		{
+			Create: action,
+		},
+	}
+
+	response.Diagnostics.Append(validateNewGSIAttributes(data, *table)...)
```
gdavison: Let's move validation to after we've retrieved the table but before we start building the input

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+	}
+
+	if _, err := waitGSIActive(ctx, conn, data.Table.ValueString(), data.Name.ValueString(), r.UpdateTimeout(ctx, data.Timeouts)); err != nil {
+		response.Diagnostics.AddError(
+			fmt.Sprintf(`Error while waiting for GSI "%s" on table "%s" to be active`, data.Name.ValueString(), data.Table.ValueString()),
+			err.Error(),
+		)
+	}
+
+	response.Diagnostics.Append(response.State.Set(ctx, &data)...)
```
gdavison, line 356: Do we need to wait for `waitGSIWarmThroughputActive` as well?

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,934 @@
...
+	}
+
+	if _, err = waitTableActive(ctx, conn, data.Table.ValueString(), createTimeout); err != nil {
+		response.Diagnostics.AddError(
+			fmt.Sprintf(`Error while waiting for table "%s" to be active`, data.Table.ValueString()),
+			err.Error(),
+		)
+
+		return
+	}
```
gdavison: In Create, Update, and Delete, we call `UpdateTable`, then `waitTableActive`. This could be factored out.

For both Create and Update, we also call `waitGSIActive`. This could also be factored out.

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1139 @@
...
+
+					resource.TestCheckResourceAttr(resourceNameGSI, names.AttrName, rNameGSI),
+					resource.TestCheckResourceAttr(resourceNameGSI, "table", rNameTable),
+					resource.TestCheckResourceAttr(resourceNameGSI, "hash_key", rNameTable),
+					resource.TestCheckResourceAttr(resourceNameGSI, "hash_key_type", "S"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "range_key", rNameGSI),
+					resource.TestCheckResourceAttr(resourceNameGSI, "range_key_type", "S"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "projection_type", "ALL"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "read_capacity", "1"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "write_capacity", "1"),
```
gdavison: The basic test should verify the values of all attributes

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1139 @@
...
+
+resource "aws_dynamodb_global_secondary_index" "test" {
+  table           = aws_dynamodb_table.test.name
+  name            = %[2]q
+  hash_key        = %[1]q
+  range_key       = %[2]q
+  read_capacity   = 1
+  write_capacity  = 1
+  range_key_type  = "S"
+  projection_type = "ALL"
```
gdavison: The configuration for the basic test should include only required attributes

gdavison (changes requested):
There are still some items from the previous review that need to be addressed regarding the resource schema and testing.

In addition to the new review comments, there also should be tests for

- [ ] GSI keys that are different from table keys
- [ ] `non_key_attributes`

internal/service/dynamodb/global_secondary_index.go
raduendv: unfortunately there's quite a bit of logic necessary here, which was ported from the global_secondary_index blocks of the table resource, this isn't feasible here

gdavison:
Hi @raduendv, a couple more notes:

- [ ] Please rebase onto the current `main` branch
- [ ] When updating the PR, please don't squash to a single commit. If each update after a review is in at least one separate commit, it's easier to see what's changed since the last review

internal/service/dynamodb/global_secondary_index.go
raduendv: i opted for this because i kept seeing changes from [] -> null

internal/service/dynamodb/global_secondary_index_test.go
raduendv: the table information is needed for automatically determining the `hash_key_type` and `range_key_type` when using already existing attributes

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"read_capacity": schema.Int64Attribute{
+				Optional: true,
+				Computed: true,
+			},
+			"table": schema.StringAttribute{
```
gdavison: This attribute should be table_name to match the underlying API


```suggestion
			"table_name": schema.StringAttribute{
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+			if rs.Primary.Attributes[names.AttrName] == aws.ToString(g.IndexName) {
+				*gsi = g
+
+				break
+			}
+		}
+
+		return nil
+	}
+}
```
gdavison: Look up the GSI using `findGSIByTwoPartKey`. You will likely have to export it by adding it to `exports_test.go`. The function can also be renamed.

 Something like

```suggestion
func testAccCheckGSIExists(ctx context.Context, t *testing.T, n string, gsi *awstypes.GlobalSecondaryIndexDescription) resource.TestCheckFunc {
	return func(s *terraform.State) error {
		rs, ok := s.RootModule().Resources[n]
		if !ok {
			return fmt.Errorf("Not found: %s", n)
		}

		conn := acctest.ProviderMeta(ctx, t).DynamoDBClient(ctx)

		tableName := rs.Primary.Attributes["table"]
		indexName := rs.Primary.Attributes["name"]

		output, err := tfdynamodb.FindGSIByTwoPartKey(ctx, conn, tableName, indexName)
		if err != nil {
			return err
		}

		*gsi = *output

		return nil
	}
}
```

raduendv:
@gdavison all points have been addressed, please re-run the pipelines

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+				Config: testAccGlobalSecondaryIndexConfig_basic(rNameTable, rNameGSI),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					testAccCheckInitialTableExists(ctx, resourceNameTable, &conf),
+					testAccCheckInitialGSIExists(resourceNameGSI, &conf, &gsi),
+
+					resource.TestCheckResourceAttr(resourceNameGSI, "hash_key", rNameTable),
+					resource.TestCheckResourceAttr(resourceNameGSI, "hash_key_type", "S"),
+					func(s *terraform.State) error {
+						return resource.TestCheckResourceAttr(resourceNameGSI, names.AttrARN, aws.ToString(gsi.IndexArn))(s)
+					},
```
gdavison: Since the ARN is composed from known values (see https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazondynamodb.html#amazondynamodb-resources-for-iam-policies), this can be replaced by

```suggestion
					acctest.CheckResourceAttrRegionalARNFormat(ctx, resourceName, names.AttrARN, "dynamodb", "table/{table_name}/index/{index_name}"),
```

(Note that this code suggestion uses the updated attribute names)

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+	"github.com/hashicorp/terraform-provider-aws/names"
+)
+
+func TestAccDynamoDBGlobalSecondaryIndex_basic(t *testing.T) {
+	ctx := acctest.Context(t)
+	var conf awstypes.TableDescription
+	var gsi awstypes.GlobalSecondaryIndexDescription
+
+	resourceNameTable := "aws_dynamodb_table.test"
+	resourceNameGSI := "aws_dynamodb_global_secondary_index.test"
```
gdavison: By convention, the name for the resource being tested is `resourceName`, and others can have clarifying names

```suggestion
	resourceName := "aws_dynamodb_global_secondary_index.test"
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+
+	rNameTable := sdkacctest.RandomWithPrefix(acctest.ResourcePrefix)
+	rNameGSI := sdkacctest.RandomWithPrefix(acctest.ResourcePrefix)
```
gdavison: By convention, the random name for the resource being tested is `rName`, and others can have clarifying names

```suggestion
	rName := acctest.RandomWithPrefix(t, acctest.ResourcePrefix)
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+
+func TestAccDynamoDBGlobalSecondaryIndex_basic(t *testing.T) {
+	ctx := acctest.Context(t)
+	var conf awstypes.TableDescription
+	var gsi awstypes.GlobalSecondaryIndexDescription
+
+	resourceNameTable := "aws_dynamodb_table.test"
+	resourceNameGSI := "aws_dynamodb_global_secondary_index.test"
+
+	rNameTable := sdkacctest.RandomWithPrefix(acctest.ResourcePrefix)
```
gdavison: Prefer `acctest.RandomWithPrefix()`

```suggestion
	rNameTable := acctest.RandomWithPrefix(t, acctest.ResourcePrefix)
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+					resource.TestCheckResourceAttr(resourceNameGSI, "table", rNameTable),
+					resource.TestCheckResourceAttr(resourceNameGSI, "write_capacity", "1"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "on_demand_throughput.#", "0"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "warm_throughput.#", "0"),
+				),
+			},
+			{
+				ResourceName:                         resourceNameGSI,
+				ImportState:                          true,
+				ImportStateIdFunc:                    testAccGlobalSecondaryIndexImportStateFunc(resourceNameGSI),
```
gdavison: This can be replaced with

```suggestion
				ImportStateIdFunc:                    acctest.AttrImportStateIdFunc(resourceNameGSI, names.AttrARN),
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+					resource.TestCheckNoResourceAttr(resourceNameGSI, "non_key_attributes"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "projection_type", "ALL"),
+					resource.TestCheckNoResourceAttr(resourceNameGSI, "range_key"),
+					resource.TestCheckNoResourceAttr(resourceNameGSI, "range_key_type"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "read_capacity", "1"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "table", rNameTable),
+					resource.TestCheckResourceAttr(resourceNameGSI, "write_capacity", "1"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "on_demand_throughput.#", "0"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "warm_throughput.#", "0"),
+				),
```
gdavison: For collection types, prefer `ConfigStateChecks`, which can differentiate between empty and `null` values. E.g.

```suggestion
				),
				ConfigStateChecks: []statecheck.StateCheck{
					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("on_demand_throughput"), knownvalue.ListExact([]knownvalue.Check{})), // e.g. for empty list
					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("warm_throughput"), knownvalue.Null()), // e.g. for null list
				},
```

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+    type = "S"
+  }
+}
+
+resource "aws_dynamodb_global_secondary_index" "test" {
+  table           = aws_dynamodb_table.test.name
+  index_name      = %[2]q
+  hash_key        = %[1]q
+  projection_type = "ALL"
+}
```
gdavison: Move the resource under test to the top of the configuration

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+  table              = aws_dynamodb_table.basic-dynamodb-table.name
+  name               = "GameTitleIndex"
+  hash_key           = "GameTitle"
+  range_key          = "TopScore"
+  range_key_type     = "N"
+  write_capacity     = 10
+  read_capacity      = 10
+  projection_type    = "INCLUDE"
+  non_key_attributes = ["UserId"]
+}
```
gdavison: Move the resource being documented to the top of the configuration

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+* `non_key_attributes` - (Optional) Only required with `INCLUDE` as a projection type; a list of attributes to project into the index. These do not need to be defined as attributes on the table.
+* `on_demand_throughput` - (Optional) Sets the maximum number of read and write units for the specified on-demand index. See below.
+* `range_key` - (Optional) Name of the range key; must be defined
+* `range_key_type` - (Optional) Type of the range key; One of `S`, `N` or `B`; If the attribute exists in the table, the value will be inferred
+* `read_capacity` - (Optional) Number of read units for this index. Must be set if billing_mode is set to PROVISIONED.
+* `warm_throughput` - (Optional) Sets the number of warm read and write units for this index. See below.
+* `write_capacity` - (Optional) Number of write units for this index. Must be set if billing_mode is set to PROVISIONED.
+* `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
```
gdavison, line 201: The argument lists should be sorted alphabetically

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,114 @@
...
+
+## Attribute Reference
+
+This resource exports the following attributes in addition to the arguments above:
+
+* `arn` - name of the GSI
```
gdavison:
```suggestion
* `arn` - ARN of the GSI
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+func (r *resourceGlobalSecondaryIndex) Schema(ctx context.Context, request resource.SchemaRequest, response *resource.SchemaResponse) {
+	s := schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			names.AttrARN: framework.ARNAttributeComputedOnly(),
+			"hash_key": schema.StringAttribute{
+				Required: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
```
gdavison: `hash_key` and `range_key` should be removed in favour of a set `key_schema`, to

* match the AWS API
* support multi-key GSIs (https://aws.amazon.com/blogs/database/multi-key-support-for-global-secondary-index-in-amazon-dynamodb/)

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+	}
+
+	re := regexache.MustCompile(":table/([^\\/]+)/index/(.+)")
+	grps := re.FindStringSubmatch(data.ARN.ValueString())
+	var tableName string
+	if len(grps) == 3 {
+		tableName = grps[1]
+	} else {
+		tableName = data.Table.ValueString()
+	}
```
gdavison: Rather than parsing the `arn`, we already have the `table_name` and `index_name` attributes that we can use

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+
+	if !found {
+		response.Diagnostics.Append(
+			fwdiag.NewResourceNotFoundWarningDiagnostic(
+				fmt.Errorf(`unable to find global secondary index with id "%s", treating it as new`, data.ARN.ValueString()),
+			),
+		)
+		response.State.RemoveResource(ctx)
+	}
```
gdavison: There are multiple ways used to find a GSI from a table, including here, in `findGSIByTwoPartKey`, in `Create`, in `Read`, `Update`, `validateNewGSIAttributes`, etc.

Add a function, e.g. `findGSIFromTable` that takes the `awstypes.TableDescription` and `index_name` and returns the GSI. It can then be used everywhere the GSI must be retrieved

E.g.

```go
func findGSIByTwoPartKey(ctx context.Context, conn *dynamodb.Client, tableName, indexName string) (*awstypes.GlobalSecondaryIndexDescription, error) {
	table, err := findTableByName(ctx, conn, tableName)

	if err != nil {
		return nil, err
	}

	return findGSIFromTable(table, indexName)
}

func findGSIFromTable(table *awstypes.TableDescription, indexName string) (*awstypes.GlobalSecondaryIndexDescription, error) {
	for _, v := range table.GlobalSecondaryIndexes {
		if aws.ToString(v.IndexName) == indexName {
			return &v, nil
		}
	}

	return nil, &retry.NotFoundError{}
}
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+	ReadCapacity        types.Int64                                              `tfsdk:"read_capacity"`
+	Table               types.String                                             `tfsdk:"table"`
+	WriteCapacity       types.Int64                                              `tfsdk:"write_capacity"`
+	OnDemandThroughputs fwtypes.ListNestedObjectValueOf[onDemandThroughputModel] `tfsdk:"on_demand_throughput"`
+	WarmThroughputs     fwtypes.ListNestedObjectValueOf[warmThroughputModel]     `tfsdk:"warm_throughput"`
+
+	Timeouts timeouts.Value `tfsdk:"timeouts"`
+}
+
+func (data *resourceGlobalSecondaryIndexModel) Flatten(ctx context.Context, v any) diag.Diagnostics {
```
gdavison: Instead of adding a `Flatten` function to `resourceGlobalSecondaryIndexModel`, custom flattening should be in a standalone function, e.g. `flattenGlobalSecondardIndex` (with a directive `// nosemgrep:ci.semgrep.framework.manual-flattener-functions`). This way it can take the table as a direct parameter instead of using the `Context` to pass it.

Note that the closer the resource schema is to the API structure, the more `fwflex.Flatten` can populate automatically. See https://hashicorp.github.io/terraform-provider-aws/data-handling-and-conversion/#autoflex-for-terraform-plugin-framework-preferred for more information how conversion can be customized.

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"global_secondary_index",
+				},
+				ImportStateCheck: acctest.ComposeAggregateImportStateCheckFunc(
+					acctest.ImportCheckResourceAttr(names.AttrName, rNameTable),
+					acctest.ImportCheckResourceAttr("attribute.#", "1"),
+				),
+			},
```
gdavison, line 2069: Since two back-to-back `ImportStateVerify` steps is an uncommon pattern, add a comment explaining what the step is for

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+			{
+				ResourceName:      resourceNameTable,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
+					"global_secondary_index",
+				},
+				ImportStateCheck: acctest.ComposeAggregateImportStateCheckFunc(
+					acctest.ImportCheckResourceAttr(names.AttrName, rNameTable),
+					acctest.ImportCheckResourceAttr("attribute.#", "1"),
```
gdavison: In addition to the count of `attribute`, add checks for the expected values

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+				ImportState:                          true,
+				ImportStateIdFunc:                    testAccGlobalSecondaryIndexImportStateFunc(resourceNameGSI),
+				ImportStateVerify:                    true,
+				ImportStateVerifyIdentifierAttribute: "index_name",
+			},
+		},
+	})
+}
+
+func TestAccDynamoDBGlobalSecondaryIndex_nonKeyAttributes(t *testing.T) {
```
gdavison: There should also be a test for creating the GSI with `non_key_attributes`

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+				ImportState:                          true,
+				ImportStateIdFunc:                    testAccGlobalSecondaryIndexImportStateFunc(resourceNameGSI),
+				ImportStateVerify:                    true,
+				ImportStateVerifyIdentifierAttribute: "index_name",
+			},
+		},
+	})
+}
+
+func TestAccDynamoDBGlobalSecondaryIndex_differentKeys(t *testing.T) {
```
gdavison: There should also be a test for creating the GSI with keys not in the base table attributes

internal/service/dynamodb/global_secondary_index_test.go
```diff
@@ -0,0 +1,1293 @@
...
+					resource.TestCheckResourceAttr(resourceNameGSI, "on_demand_throughput.#", "0"),
+					resource.TestCheckResourceAttr(resourceNameGSI, "warm_throughput.#", "0"),
+				),
+			},
+			{
+				ResourceName:                         resourceNameGSI,
+				ImportState:                          true,
+				ImportStateIdFunc:                    testAccGlobalSecondaryIndexImportStateFunc(resourceNameGSI),
+				ImportStateVerify:                    true,
+				ImportStateVerifyIdentifierAttribute: "index_name",
```
gdavison: This can be

```suggestion
				ImportStateVerifyIdentifierAttribute: names.AttrARN,
```

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+		)
+
+		return
+	}
+
+	if _, err := waitGSIActive(ctx, conn, data.Table.ValueString(), data.IndexName.ValueString(), createTimeout); err != nil {
+		response.Diagnostics.AddError(
+			fmt.Sprintf(`Error while waiting for GSI "%s" on table "%s" to be active`, data.IndexName.ValueString(), data.Table.ValueString()),
+			err.Error(),
+		)
```
gdavison, line 337: This should `return` on error

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+	}
+
+	if err = waitGSIWarmThroughputActive(ctx, conn, data.Table.ValueString(), data.IndexName.ValueString(), createTimeout); err != nil {
+		response.Diagnostics.AddError(
+			fmt.Sprintf(`Error while waiting for warm throughput on GSI "%s" on table "%s" to be active`, data.IndexName.ValueString(), data.Table.ValueString()),
+			err.Error(),
+		)
```
gdavison, line 351: This should `return` on error

internal/service/dynamodb/global_secondary_index.go
```diff
@@ -0,0 +1,912 @@
...
+	if !ok || table == nil {
+		diags.AddError(
+			"Bad argument",
+			"Context is missing the table description",
+		)
+
+		return diags
+	}
+
+	data.ARN = types.StringValue(aws.ToString(g.IndexArn))
```
gdavison: Replace `types.StringValue(aws.ToString(val))` with `fwflex.StringToFramework(val)` to get proper handling including `nil` values

```suggestion
	data.ARN = fwflex.StringToFramework(g.IndexArn)
```

internal/service/dynamodb/global_secondary_index.go
gdavison: Where are you seeing that? I replaced the `setplanmodifier.RequiresReplaceIf` with `setplanmodifier.RequiresReplace`, and didn't see any errors with the acceptance tests.

gdavison (changes requested):
It's easier to see and track the issues raised if they are still open, and I can resolve them once I've reviewed them, so please don't resolve them.

Some changes from earlier reviews, including

- [ ] The resource schema should closely match the API struct
- [ ]  I replaced the setplanmodifier.RequiresReplaceIf with setplanmodifier.RequiresReplace, and didn't see any errors with the acceptance tests (I un-resolved the issue from earlier, but I can't link it)

This will need some additional tests:

- [ ] Creating, Updating, and/or Deleting multiple GSIs concurrently
- [ ] Migrating a single GSI from `aws_dynamodb_table` to `aws_dynamodb_global_secondary_index`
- [ ] Migrating multiple GSIs from `aws_dynamodb_table` to `aws_dynamodb_global_secondary_index`
- [ ] Partially migrating GSIs from `aws_dynamodb_table` to `aws_dynamodb_global_secondary_index`, e.g. start with 2 GSIs on `aws_dynamodb_table` and move 1 to `aws_dynamodb_global_secondary_index`
- [ ] Two GSIs both adding a key with the same name but with a different type.

internal/service/dynamodb/global_secondary_index.go
raduendv: with all the recents changes, i can no longer reproduce that

updated to use the suggested function

internal/service/dynamodb/global_secondary_index.go
raduendv: during the import, the state doesn't seem to be set and only the ARN is available due to importing with it:

```
terraform import aws_dynamodb_global_secondary_index.test arn:aws:dynamodb:eu-west-1:123456789012:table/tf-test/index/tf-test-gsi

aws_dynamodb_global_secondary_index.test: Import prepared!
  Prepared aws_dynamodb_global_secondary_index for import
aws_dynamodb_global_secondary_index.test: Refreshing state...
╷
│ Error: Unable to read table ""
│ 
│ operation error DynamoDB: DescribeTable, https response error StatusCode: 400, RequestID:
│ 2PJHUSR9JE60JJGP24L3EP2JF3VV4KQNSO5AEMVJF66Q9ASUAAJG, api error ValidationException: 2 validation
│ errors detected: Value '' at 'tableName' failed to satisfy constraint: Member must have length
│ greater than or equal to 3; Value '' at 'tableName' failed to satisfy constraint: Member must satisfy
│ regular expression pattern: [a-zA-Z0-9_.-]+
```

if you have another idea to get over this, please let me know

internal/service/dynamodb/table.go
```diff
@@ -188,6 +201,7 @@ func resourceTable() *schema.Resource {
 				"global_secondary_index": {
 					Type:     schema.TypeSet,
 					Optional: true,
+					Computed: true,
```
ewbankkit, line 204: Does this need a CHANGELOG entry?

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,256 @@
...
+subcategory: "DynamoDB"
+layout: "aws"
+page_title: "AWS: aws_dynamodb_global_secondary_index"
+description: |-
+  Provides a DynamoDB Global Secondary Index resource
+---
+
+# Resource: aws_dynamodb_global_secondary_index
+
+!> The resource type `aws_dynamodb_global_secondary_index` is an experimental feature. The schema or behavior may change without notice, and it is not subject to the backwards compatibility guarantee of the provider.
```
ewbankkit, line 11: Does the (current) experimental nature of this resource warrant something in the CHANGELOG (a _note_?)

internal/service/dynamodb/global_secondary_index_key_schema_list_validator.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (v globalSecondaryIndexKeySchemaListValidator) Description(_ context.Context) string {
+	return ""
+}
+
+func (v globalSecondaryIndexKeySchemaListValidator) MarkdownDescription(_ context.Context) string {
+	return ""
+}
+
+func (v globalSecondaryIndexKeySchemaListValidator) ValidateList(ctx context.Context, request validator.ListRequest, response *validator.ListResponse) {
```
jar-b, line 32: Non-blocker: A unit test against this might be a faster/easier to locate option to supplement the `_validate` acceptance test(s).

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,256 @@
...
+### `warm_throughput`
+
+* `read_units_per_second` - (Required) Number of read operations this index can instantaneously support.
+* `write_units_per_second` - (Required) Number of write operations this index can instantaneously support.
+
+## Attribute Reference
+
+This resource exports the following attributes in addition to the arguments above:
+
+* `arn` - ARN of the GSI
```
jar-b:
```suggestion
* `arn` - ARN of the GSI.
```

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,256 @@
...
+* `provisioned_throughput` - (Optional) Provisioned throughput for the index.
+  See [`provisioned_throughput` below](#provisioned_throughput).
+  Required if the table's `billing_mode` is `PROVISIONED`.
+* `region` - (Optional) Region where this resource will be [managed](https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints). Defaults to the Region set in the [provider configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#aws-configuration-reference).
+* `warm_throughput` - (Optional) Sets the number of warm read and write units for this index.
+  See [`warm_throughput` below](#warm_throughput).
+
+### `key_schema`
+
+* `attribute_name` - (Required) Name of the attribute
```
jar-b:
```suggestion
* `attribute_name` - (Required) Name of the attribute.
```

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,256 @@
...
+
+* `index_name` - (Required) Name of the index.
+* `key_schema` - (Required) Set of nested attribute definitions.
+  At least 1 element defining a `HASH` is required.
+  All elements with the `key_type` of `HASH` must precede elements with `key_type` of `RANGE`.
+  Changing any values in `key_schema` will re-create the resource.
+  See [`key_schema` below](#key_schema).
+* `projection` - (Required) Describes which attributes from the table are represented in the index.
+  See [`projection` below](#projection).
+* `table_name` - (Required) Name of the table this index belongs to
```
jar-b:
```suggestion
* `table_name` - (Required) Name of the table this index belongs to.
```

website/docs/r/dynamodb_global_secondary_index.html.markdown
```diff
@@ -0,0 +1,256 @@
...
+* `read_units_per_second` - (Required) Number of read operations this index can instantaneously support.
+* `write_units_per_second` - (Required) Number of write operations this index can instantaneously support.
+
+## Attribute Reference
+
+This resource exports the following attributes in addition to the arguments above:
+
+* `arn` - ARN of the GSI
+
+## Import
```
jar-b, line 241: Also needs instructions for importing by identity.

internal/service/dynamodb/global_secondary_index_key_schema_list_validator.go
gdavison: I'll add this in a new PR

PR #45106, opened 2025-11-17
Allows use of explicitly typed `retry.StateRefreshFuncOf`

gdavison:
## Rollback Plan

### Description

When using a `retry.StateRefreshFuncOf` with a typed return value rather than `any`, the not-found case is not recognized, by the check at

https://github.com/hashicorp/terraform-provider-aws/blob/09d8ae5baf03bbe2c887cebb1597fc7811543776/internal/retry/state.go#L109

Instead, use the `types.IsZero` check to compare the value with the zero value.

### DIFF
--- a/internal/retry/state.go
+++ b/internal/retry/state.go
@@ -12,7 +12,7 @@ import (
 	"github.com/hashicorp/terraform-provider-aws/internal/backoff"
 	"github.com/hashicorp/terraform-provider-aws/internal/conns"
 	tfslices "github.com/hashicorp/terraform-provider-aws/internal/slices"
-	itypes "github.com/hashicorp/terraform-provider-aws/internal/types"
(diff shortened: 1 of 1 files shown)


internal/retry/state.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-provider-aws/internal/backoff"
 	"github.com/hashicorp/terraform-provider-aws/internal/conns"
 	tfslices "github.com/hashicorp/terraform-provider-aws/internal/slices"
+	itypes "github.com/hashicorp/terraform-provider-aws/internal/types"
```
ewbankkit: Nit: I think at one point we agreed to name this alias **inttypes** (but didn't enforce it).

ewbankkit (approved):
LGTM :rocket:.

Issue #45103, opened 2025-11-17
Documentation Error: Incorrect attribute name for aws_odb_db_servers data source


kampilis:
### Documentation Link(s)

The documentation for the aws_odb_db_servers data source contains an incorrect attribute name.

Documentation URL: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/odb_db_servers#db_server_id-2

### Description

Problem

The documentation suggests using db_server_id as an attribute of the data source, but this attribute does not exist and causes a Terraform error.

Error Message

Error: Unsupported attribute

on ../modules/odbnetwork/odb_cloud_vm_cluster.tf line 15, in resource "aws_odb_cloud_vm_cluster" "odb_cloud_vm_cluster":
15: db_servers = data.aws_odb_db_servers.odb_db_servers.*.db_server_id

This object does not have an attribute named "db_server_id".
Incorrect Syntax (as documented)

data.aws_odb_db_servers.odb_db_servers.*.db_server_id

Correct Syntax (actually works)

data.aws_odb_db_servers.odb_db_servers.db_servers[*].id

Expected Fix

### References

Update the documentation to reflect the correct attribute path: db_servers[*].id instead of db_server_id.

Additional Context

Note: The aws_odb_db_servers data source requires the Exadata infrastructure to already exist. It would be beneficial if the db_servers list was also available as an attribute directly on the aws_odb_exadata_infrastructure resource to avoid the dependency issue.

### Would you like to implement a fix?

No

lukzhang:
@kampilis Hello, can I work on this fix?

kampilis:
@lukzhang Please do it. Thanks alot

lukzhang:
@kampilis PR made

PR #45107, opened 2025-11-18
fix(docs): correct attribute name for aws_odb_db_servers data source (#45103)

lukzhang:
This PR fixes the documentation for the aws_odb_db_servers data source by replacing the incorrect attribute name db_server_id with the correct attribute path db_servers[*].id.

### DIFF
--- a/website/docs/d/odb_db_servers.html.markdown
+++ b/website/docs/d/odb_db_servers.html.markdown
@@ -46,7 +46,7 @@ This data source exports the following attributes in addition to the arguments a
 * `cpu_core_count` - The number of CPU cores enabled on the database server.
 * `created_at` - The date and time when the database server was created.
 * `db_node_storage_size_in_gbs` - The amount of local node storage, in gigabytes (GB), that's allocated on the database server.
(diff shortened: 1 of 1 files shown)


website/docs/d/odb_db_servers.html.markdown
```diff
@@ -46,7 +46,7 @@ This data source exports the following attributes in addition to the arguments a
 * `cpu_core_count` - The number of CPU cores enabled on the database server.
 * `created_at` - The date and time when the database server was created.
 * `db_node_storage_size_in_gbs` - The amount of local node storage, in gigabytes (GB), that's allocated on the database server.
-* `db_server_id` - The unique identifier of the database server.
+* `db_servers[*].id` - The unique identifiers of the database servers.
```
jar-b: Since this list of attributes documents what is contained in the `db_servers` attribute, I think this should just be `id`.


```suggestion
* `id` - The unique identifier of the database server.
```

jar-b:
Thank you for the correction, @lukzhang! 👍

PR #44273, opened 2025-09-13
F/athena workgroup support managed query results configuration

manaty226:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to security controls. This PR adds support for configuring managed query results of Athena Workgroup.

### Description

Adds support for `managed_query_results_configuration` to the `aws_athena_workgroup` resource. 
This enables store query results into Athena owned storage.

Also support exclusive settings with `results_configuration`.

### Relations

Closes #43246

### References

- https://docs.aws.amazon.com/athena/latest/ug/managed-results.html

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccAthenaWorkGroup_managedQueryResultsConfiguration PKG=athena

TF_ACC=1 go1.24.6 test ./internal/service/athena/... -v -count 1 -parallel 20 -run='TestAccAthenaWorkGroup_managedQueryResultsConfiguration'  -timeout 360m -vet=off
2025/09/13 21:27:11 Creating Terraform AWS Provider (SDKv2-style)...
2025/09/13 21:27:11 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccAthenaWorkGroup_managedQueryResultsConfigurationEnabled
=== PAUSE TestAccAthenaWorkGroup_managedQueryResultsConfigurationEnabled
=== RUN   TestAccAthenaWorkGroup_managedQueryResultsConfiguration_kms
=== PAUSE TestAccAthenaWorkGroup_managedQueryResultsConfiguration_kms
=== CONT  TestAccAthenaWorkGroup_managedQueryResultsConfigurationEnabled
=== CONT  TestAccAthenaWorkGroup_managedQueryResultsConfiguration_kms
--- PASS: TestAccAthenaWorkGroup_managedQueryResultsConfiguration_kms (32.76s)
--- PASS: TestAccAthenaWorkGroup_managedQueryResultsConfigurationEnabled (35.25s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/athena     41.848s
```

Issue #37917, opened 2024-06-11
[Bug]: Error when removing Athena workgroup encryption


mlnrt:
### Terraform Core Version

1.7.4

### AWS Provider Version

5.35.0

### Affected Resource(s)

aws_athena_workgroup

### Expected Behavior

Removing the configuration.result_configuration.encryption_configuration block from the resource should call the UpdateWorkgroup API setting ResultConfigurationUpdates={RemoveEncryptionConfiguration=true} 

### Actual Behavior

Terraform seems to try to set the encryption configuration to empty values which is not allowes

### Relevant Error/Panic Output Snippet

```shell
When removing the configuration.result_configuration.encryption_configuration block from the resource, terraform fails on the apply stage with the following error:

Error: updating Athena Workgroup (test): operation error Athena: UpdateWorkGroup, 1 validation error(s) found. - missing required field, UpdateWorkGroupInput.Configuration.ResultConfigurationUpdates.EncryptionConfiguration.EncryptionOption
```

### Terraform Configuration Files

```
resource "aws_athena_workgroup" "test" {
  name = "test"
  force_destroy = true

  configuration {
  bytes_scanned_cutoff_per_query     = 107374182400
  enforce_workgroup_configuration    = true
  publish_cloudwatch_metrics_enabled = true
    result_configuration {
    output_location = "s3://..."
      encryption_configuration {
        encryption_option = "SSE_KMS"
        kms_key_arn          = <KMS key resource reference ARN>
      }
    }
  }
}
```

### Steps to Reproduce

1. create an Athena workgroup with encryption
```
resource "aws_athena_workgroup" "test" {
  name = "test"
  force_destroy = true

  configuration {
  bytes_scanned_cutoff_per_query     = 107374182400
  enforce_workgroup_configuration    = true
  publish_cloudwatch_metrics_enabled = true
    result_configuration {
    output_location = "s3://..."
      encryption_configuration {
        encryption_option = "SSE_KMS"
        kms_key_arn          = <KMS key resource reference ARN>
      }
    }
  }
}
```

2. remove the configuration.result_configuration.encryption_configuration block, then apply the terraform plan

### Would you like to implement a fix?

No

olanre-atlassian:
This also occurs when a new encryption configuration is added to a workgroup

gdavison:
When `output_location` is set along with `encryption_configuration`, it returns the error

[...]

PR #45223, opened 2025-11-24
r/aws_athena_workgroup - Add support for managed_query_results_configuration

### LINKED ISSUE #43246: Add support for aws_athena_workgroup managed storage
### Description

AWS has introduced managed storage for Athena query results. 

https://aws.amazon.com/about-aws/whats-new/2025/06/amazon-athena-managed-query-results-streamline-analysis-workflows/

This enhancement request is to add the new managed storage API features to the aws_athena_workgroup resource.

* `aws_athena_workgroup`

```hcl
resource "aws_athena_workgroup" "primary" {
  ...
  managed_query_results_configuration {
    enabled = true
    encryption_configuration = {
      kms_key = "arn:aws:my:kms:arn"
    }
  }
  ...
}
```

### References

* https://docs.aws.amazon.com/athena/latest/APIReference/API_CreateWorkGroup.html
* https://aws.amazon.com/about-aws/whats-new/2025/06/amazon-athena-managed-query-results-streamline-analysis-workflows/
* https://aws.amazon.com/blogs/big-data/introducing-managed-query-results-for-amazon-athena/
* https://docs.aws.amazon.com/athena/latest/ug/managed-results.html

No

subham-ibmhc:
## Rollback Plan

### Description

Adds support to aws_athena_workgroup resource for managed storage for Athena query results.

### Relations

Closes #43246

### References

https://docs.aws.amazon.com/athena/latest/APIReference/API_ManagedQueryResultsConfiguration.html
https://docs.aws.amazon.com/athena/latest/APIReference/API_WorkGroupConfiguration.html

```console
% make testacc TESTS=TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration PKG=athena
make: Running acceptance tests on branch: 🌿 r/aws_athena_workgroup-add_managed_storage 🌿...
TF_ACC=1 go1.24.10 test ./internal/service/athena/... -v -count 1 -parallel 20 -run='TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration'  -timeout 360m -vet=off
2025/11/20 13:35:33 Creating Terraform AWS Provider (SDKv2-style)...
2025/11/20 13:35:33 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration
=== PAUSE TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration
=== CONT  TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration
--- PASS: TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration (34.18s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/athena     38.854s
...
```

### DIFF
--- a/internal/service/athena/workgroup.go
+++ b/internal/service/athena/workgroup.go
@@ -14,7 +14,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/athena"
 	"github.com/aws/aws-sdk-go-v2/service/athena/types"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
@@ -41,9 +40,7 @@ func resourceWorkGroup() *schema.Resource {
 			StateContext: schema.ImportStatePassthroughContext,
 		},
 
-		CustomizeDiff: customdiff.All(
-			managedQueryResultsValidation,
-		),
+		CustomizeDiff: managedQueryResultsValidation,
 
 		Schema: map[string]*schema.Schema{
 			names.AttrARN: {


subham-ibmhc:
For the attribute managed_query_results_configuration.enabled, when set to true, the attribute result_configuration.output_location cannot be set.
I am not sure how to implement a validation for this, as I am not sure ConflictsWith works here?
Should a validation be added for this or is the documentation enough?

internal/service/athena/workgroup.go
```diff
@@ -106,6 +106,34 @@ func resourceWorkGroup() *schema.Resource {
...
+						"managed_query_results_configuration": {
+							Type:     schema.TypeList,
+							Optional: true,
+							MaxItems: 1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									names.AttrEnabled: {
+										Type:     schema.TypeBool,
+										Optional: true,
+										Default:  false,
```
jar-b: Unless necessary for the API to function, we typically avoid provider-side defaults.
https://hashicorp.github.io/terraform-provider-aws/data-handling-and-conversion/#default-values

internal/service/athena/workgroup_test.go
```diff
@@ -686,6 +686,38 @@ func TestAccAthenaWorkGroup_tags(t *testing.T) {
 	})
 }
 
+func TestAccAthenaWorkGroup_ManagedQueryResultsConfiguration(t *testing.T) {
```
jar-b, line 690: Can we add additional steps to this test case to exercise disabling, then re-enabling the feature to ensure update works as expected?

website/docs/r/athena_workgroup.html.markdown
```diff
@@ -64,6 +65,11 @@ This resource supports the following arguments:
 * `enable_identity_center` - (Optional) Specifies whether the workgroup is IAM Identity Center supported.
 * `identity_center_instance_arn` - (Optional) The IAM Identity Center instance ARN that the workgroup associates to.
 
+#### Managed Query Results Configuration
+
+* `enabled` - (Optional) If set to true, allows you to store query results in Athena owned storage. If set to false, workgroup member stores query results in location specified under result_configuration.output_location. The default is false. A workgroup cannot have the result_configuration.output_location parameter when you set this field to true.
```
jar-b: nit - making values and attribute names format as code can make the rendered result a bit more readable.

```suggestion
* `enabled` - (Optional) If set to `true`, allows you to store query results in Athena owned storage. If set to `false`, workgroup member stores query results in the location specified by `result_configuration.output_location`. The default is `false`. A workgroup cannot have the `result_configuration.output_location` set when this is `true`.
```

jar-b:
> For the attribute managed_query_results_configuration.enabled, when set to true, the attribute result_configuration.output_location cannot be set.
> I am not sure how to implement a validation for this, as I am not sure ConflictsWith works here?

Because the attributes are split across arguments, you'll likely need a `CustomizeDiff` function which is able to retrieve configured values from the entire schema and then perform actions such as forcing replacements, marking attributes as computed, or throwing errors. The Plugin SDK V2 documentation on the topic is here, and grepping for `CustomizeDiff` will yield some prior art for handwritten customize diff logic in other resources.

https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences

manaty226:
@subham-ibmhc 

Thanks for your work on this!
I noticed that #45223 addresses the same problem as my existing PR, #44273.
To avoid duplicated review effort and accelerate the merge, could we perhaps merge your changes into #44273?

Thanks!

subham-ibmhc:
> @subham-ibmhc
> 
> Thanks for your work on this! I noticed that #45223 addresses the same problem as my existing PR, #44273. To avoid duplicated review effort and accelerate the merge, could we perhaps merge your changes into #44273?
> 
> Thanks!

@manaty226 

Thank you for your contribution. I would be glad to merge my changes into #44273.

However, while trying to test the update functionality, I have noticed a few issues with our previous approach. I suspect the same issues may be causing #37917, as the logic surrounding `result_configuration` is similar. I will try to summarise the issues below -

1. When the resource is initially created with `enabled` and `encryption_configuration` set, and later as part of update, `encryption_configuration` block is removed, the expand function (`expandWorkGroupManagedQueryResultsConfigurationUpdates` in my case) still returns an empty struct for `encryptionConfiguration`. This is because the config passed is a list with one element (empty kms key) , i.e. - [map[kms_key:]] instead of an empty list. Hence the `removeEncryptionConfiguration` flag is never set.
2. When the resource is initially created with `enabled` and `encryption_configuration` set, and later as part of update, `enabled` is set to false, the expand function maps enabled as false, and either ends up sending the encryption_configuration struct, which leads to the upstream API complaining to remove all other attributes in the `managed_query_results_configuration` block, or ends up sending encryption_configuration struct with an empty kms key (same as point 1), which leads to InvalidInputError from the upstream API as the kms key is required field with `encryption_configuration`.
3. As highlighted above, to implement the constraint that exists between `enabled` and `output_location`, a `customDiff` function will be needed since `conflictsWith` will not work for attributes split across arguments.

To fix these issues, I have changed the approach of the `expandWorkGroupManagedQueryResultsConfigurationUpdates` function, now it sets the remove encryption attribute if `enabled` is false and returns early, but also checks if the returned encryption_config from the `expandManagedQueryResultsEncryptionConfiguration` is meaningful and only then expands it, otherwise sets the remove_encryption attribute.
To test the above scenarios, I have also added some more test cases.

manaty226:
Hi @subham-ibmhc

Thank you for the detailed investigation and for your willingness to merge your changes into #44273.

I fully understand the issues you identified regarding the update logic for `encryption_configuration`.
Your proposed fix for `expandWorkGroupManagedQueryResultsConfigurationUpdates` looks good, and I appreciate you adding the extra test cases to cover these scenarios.

Please proceed with the merge. Thanks again for your hard work on this!

subham-ibmhc:
Changes for this have been merged as part of #44273, hence closing this PR.

PR #45289, opened 2025-11-26
aws_bedrockagentcore_agent_runtime:  Fix `lifecycle_configuration` attributes to be properly computed when not specified

### LINKED ISSUE #45290: [Bug]: aws_bedrockagentcore_agent_runtime lifecycle_configuration produces inconsistent result after apply

```console
Terraform v1.13.5
provider registry.terraform.io/hashicorp/aws v6.22.1
```

* `aws_bedrockagentcore_agent_runtime`

### Expected Behavior

When specifying only `max_lifetime` in `lifecycle_configuration` without `idle_runtime_session_timeout`, Terraform should accept the API-computed default value for `idle_runtime_session_timeout` without error.

### Actual Behavior

When only `max_lifetime` is specified AWS automatically sets `idle_runtime_session_timeout` to default - or to matching values if `max_lifetime` is lower than detauls idle timeout. This causes Terraform to report an inconsistent result after apply error because it expected `null` but received a computed value.

```console

When applying changes to aws_bedrockagentcore_agent_runtime.test, provider
"provider[\"registry.terraform.io/hashicorp/aws\"]" produced an unexpected
new value: .lifecycle_configuration[0].idle_runtime_session_timeout: was
null, but now cty.NumberIntVal(600).

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

```hcl
resource "aws_bedrockagentcore_agent_runtime" "example" {
  agent_runtime_name = "example_agent_runtime"
  role_arn           = aws_iam_role.example.arn

  agent_runtime_artifact {
    container_configuration {
      container_uri = "123456789012.dkr.ecr.us-east-1.amazonaws.com/my-image:latest"
    }
  }

  lifecycle_configuration {
    max_lifetime = 600  # Lower than default idle_runtime_session_timeout of 900
  }

  network_configuration {
    network_mode = "PUBLIC"
  }
}
```

### Steps to Reproduce

1. Create an `aws_bedrockagentcore_agent_runtime` resource with only `max_lifetime` set in `lifecycle_configuration`
2. Run `terraform apply`
3. Observe the "Provider produced inconsistent result after apply" error

Yes

The root cause is that the `lifecycle_configuration` was using `schema.ListAttribute` which only marks the list itself as Optional+Computed, but does not mark the individual nested attributes (`idle_runtime_session_timeout`, `max_lifetime`) as Computed. When the AWS API returns a computed default value for an unspecified attribute, Terraform sees this as an inconsistent result.

Yes

marcinbelczewski:
## Rollback Plan

### Description

Fixes the `aws_bedrockagentcore_agent_runtime` resource to properly handle computed default values for `lifecycle_configuration` attributes.

When using `lifecycle_configuration` with only `max_lifetime` specified (without `idle_runtime_session_timeout`), AWS automatically computes a default value for `idle_runtime_session_timeout`. The previous implementation used `schema.ListAttribute` which only marked the list as Optional+Computed, but the nested attributes were not marked as Computed. This caused Terraform to report "Provider produced inconsistent result after apply" when the API returned computed values.

The fix converts `lifecycle_configuration` from a `schema.ListAttribute` to a `schema.ListNestedBlock` where both `idle_runtime_session_timeout` and `max_lifetime` are explicitly marked as `Optional: true, Computed: true`. This allows the AWS API to return computed default values without triggering the inconsistent result error.

### Relations

Closes #45290

### References

- [Terraform Plugin Framework - Computed Attributes](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#computed)

```console
% make testacc TESTS=TestAccBedrockAgentCoreAgentRuntime_lifecycleConfiguration PKG=bedrockagentcore

=== RUN   TestAccBedrockAgentCoreAgentRuntime_lifecycleConfiguration
=== PAUSE TestAccBedrockAgentCoreAgentRuntime_lifecycleConfiguration
=== RUN   TestAccBedrockAgentCoreAgentRuntime_lifecycleConfigurationMaxLifetimeOnly
=== PAUSE TestAccBedrockAgentCoreAgentRuntime_lifecycleConfigurationMaxLifetimeOnly
=== CONT  TestAccBedrockAgentCoreAgentRuntime_lifecycleConfiguration
=== CONT  TestAccBedrockAgentCoreAgentRuntime_lifecycleConfigurationMaxLifetimeOnly
--- PASS: TestAccBedrockAgentCoreAgentRuntime_lifecycleConfigurationMaxLifetimeOnly (32.96s)
--- PASS: TestAccBedrockAgentCoreAgentRuntime_lifecycleConfiguration (48.21s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/bedrockagentcore   55.494s
```

### DIFF
--- a/.changelog/45289.txt
+++ b/.changelog/45289.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_bedrockagentcore_agent_runtime: Fix `lifecycle_configuration` attributes to be properly computed when not specified
+```
--- a/internal/service/bedrockagentcore/agent_runtime.go
+++ b/internal/service/bedrockagentcore/agent_runtime.go
@@ -87,7 +87,6 @@ func (r *agentRuntimeResource) Schema(ctx context.Context, request resource.Sche
 				CustomType: fwtypes.MapOfStringType,
 				Optional:   true,
 			},
-			"lifecycle_configuration": framework.ResourceOptionalComputedListOfObjectsAttribute[lifecycleConfigurationModel](ctx, 1, nil, listplanmodifier.UseStateForUnknown()),
 			names.AttrRoleARN: schema.StringAttribute{
 				CustomType: fwtypes.ARNType,
 				Required:   true,
@@ -98,6 +97,24 @@ func (r *agentRuntimeResource) Schema(ctx context.Context, request resource.Sche
 		},
 
 		Blocks: map[string]schema.Block{
+			"lifecycle_configuration": schema.ListNestedBlock{
+				CustomType: fwtypes.NewListNestedObjectTypeOf[lifecycleConfigurationModel](ctx),
+				Validators: []validator.List{
+					listvalidator.SizeAtMost(1),
+				},
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"idle_runtime_session_timeout": schema.Int32Attribute{
+							Optional: true,
+							Computed: true,
+						},
+						"max_lifetime": schema.Int32Attribute{
+							Optional: true,
+							Computed: true,
+						},
+					},
+				},
+			},
 			"agent_runtime_artifact": schema.ListNestedBlock{
 				CustomType: fwtypes.NewListNestedObjectTypeOf[agentRuntimeArtifactModel](ctx),
 				Validators: []validator.List{
--- a/internal/service/bedrockagentcore/agent_runtime_test.go
+++ b/internal/service/bedrockagentcore/agent_runtime_test.go
@@ -642,6 +642,119 @@ func TestAccBedrockAgentCoreAgentRuntime_artifactCode(t *testing.T) {
 	})
 }
 
+func TestAccBedrockAgentCoreAgentRuntime_lifecycleConfiguration(t *testing.T) {
+	ctx := acctest.Context(t)
+	var agentRuntime bedrockagentcorecontrol.GetAgentRuntimeOutput
+	rName := strings.ReplaceAll(sdkacctest.RandomWithPrefix(acctest.ResourcePrefix), "-", "_")
+	resourceName := "aws_bedrockagentcore_agent_runtime.test"
+	rImageUri := acctest.SkipIfEnvVarNotSet(t, "AWS_BEDROCK_AGENTCORE_RUNTIME_IMAGE_V1_URI")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck: func() {
+			acctest.PreCheck(ctx, t)
+			acctest.PreCheckPartitionHasService(t, names.BedrockEndpointID)
+			testAccPreCheckAgentRuntimes(ctx, t)
+		},
+		ErrorCheck:               acctest.ErrorCheck(t, names.BedrockAgentCoreServiceID),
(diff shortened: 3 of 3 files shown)


aidandaly24:
I ran this branch (head `e07dfc3`) against real AWS and hit two issues on the block-omitted path, both coming from `lifecycle_configuration` being a `schema.ListNestedBlock`. The underlying thing is that a plugin-framework Block can't be Computed at the list level, only its nested attributes can be (which is why `idle_runtime_session_timeout` / `max_lifetime` being Optional+Computed correctly handles "specify one, compute the other", and that test passes). But AWS always returns a populated `LifecycleConfiguration` from GetAgentRuntime even when the block isn't specified, and the block can't represent "fill this in from the API". That gives two failures.

1. Creating with the block omitted errors with inconsistent-result:
```
.lifecycle_configuration: block count changed from 0 to 1
```
The runtime is created in AWS first, then the apply fails. With the block omitted the plan value is a known empty list (count 0) since the block can't be list-level Computed, and Create does findAgentRuntimeByID then flattens the full GetAgentRuntimeOutput into state (agent_runtime.go:392-399), and `GetAgentRuntimeOutput.LifecycleConfiguration` comes back populated with the API defaults, so the count goes to 1. Repro:
```hcl
resource "aws_bedrockagentcore_agent_runtime" "test" {
  agent_runtime_name = "example"
  role_arn           = aws_iam_role.test.arn
  agent_runtime_artifact { container_configuration { container_uri = "<uri>" } }
  network_configuration { network_mode = "PUBLIC" }
  # no lifecycle_configuration block
}
```

2. After importing an existing runtime, or removing the block from a managed one, plan perpetually shows the block being removed and never converges:
```
- lifecycle_configuration {
    - idle_runtime_session_timeout = <default> -> null
    - max_lifetime                 = <default> -> null
  }
~ agent_runtime_version = "..." -> (known after apply)
```
Read (agent_runtime.go:428) flattens the whole GetAgentRuntimeOutput into state on every refresh so the block gets repopulated with the API defaults each time, and the set->unset Update does apply but it flattens UpdateAgentRuntimeOutput (agent_runtime.go:469) which carries `AgentRuntimeVersion` but not `LifecycleConfiguration`, so it bumps the runtime version without reconciling the block and the next refresh re-adds the defaults. So every apply is effectively a no-op that still cuts a new runtime version. (`agent_runtime_version` at :77-79 is Computed with no `UseStateForUnknown`, which is why it flips to known-after-apply on these plans.)

One way I think this could be fixed is to model `lifecycle_configuration` as a `schema.ListNestedAttribute` that's list-level Optional+Computed with `listplanmodifier.UseStateForUnknown()` (keeping the two nested fields Optional+Computed) instead of a Block. Then omitted plans as known-after-apply and Create's flatten fills it with no 0 to 1 error, `UseStateForUnknown` stops the perpetual diff and the spurious version bumps, and the per-field computed behavior your `..._maxLifetimeOnly` test exercises still works. The tradeoff is the HCL moves from block form to attribute form (`lifecycle_configuration = [{ ... }]`), which matches how other always-defaulted nested objects are handled in the provider.

Let me know what you think, happy to re-run the omit / import / set->unset cases against real AWS once you pick an approach. Thanks!

PR #45349, opened 2025-12-02
Replaces uses of `tfresource.TimedOut` with `retry.TimedOut`

gdavison:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

Replaces uses of `tfresource.TimedOut` with `retry.TimedOut`.

Relates #25602

PR #45350, opened 2025-12-02
Replaces uses of `tfresource.SetLastError` with `retry.SetLastError`

gdavison:
## Rollback Plan

### Description

Replaces uses of `tfresource.SetLastError` with `retry.SetLastError`.

Relates #25602

### DIFF
--- a/internal/service/cloudfront/trust_store.go
+++ b/internal/service/cloudfront/trust_store.go
@@ -285,7 +285,7 @@ func waitTrustStoreActive(ctx context.Context, conn *cloudfront.Client, id strin
 
 	outputRaw, err := stateConf.WaitForStateContext(ctx)
 	if out, ok := outputRaw.(*cloudfront.GetTrustStoreOutput); ok {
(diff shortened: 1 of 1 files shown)


internal/tfresource/errors_test.go
```diff
@@ -1,88 +0,0 @@
...
-import (
-	"errors"
-	"strings"
-	"testing"
-
-	sdkretry "github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
-	"github.com/hashicorp/terraform-provider-aws/internal/tfresource"
-)
-
-func TestSetLastError(t *testing.T) {
```
jar-b, line 15: Can we also remove `tfresource.SetLastError` itself, or are there certain places we need to continue referencing it?

gdavison: I thought about that, and in the three PRs (this one, ##45320, and #45349), I decided to leave them for now so that we're not breaking PRs, but the Semgrep validations will catch them. It's not a huge deal, but means a lot fewer failing checks

jar-b: Makes sense, thanks!

PR #45383, opened 2025-12-03
Update PAT username after migration

breathingdust:
## Rollback Plan

### Description

Currently this action is silently reporting no results, in concert with changes to that action, this modification allow it to work correctly (old identity is no longer valid). Also switching to a more stable custom field name rather than ID (which can change)


ewbankkit (approved):
LGTM :rocket:.

PR #45245, opened 2025-11-25
opensearch: Remove 10-minute delays from domain creation and deletion

berimbolo13:
Closes #45244

Removes hardcoded delays before polling domain state during creation, update, and deletion operations. The provider will now begin polling immediately with a 10-second interval for all operations.

This significantly reduces wait times for:
- Fast AWS provisioning scenarios
- Local testing with LocalStack (lightweight AWS emulator)

**Changes:**
- Removed `tfresource.WithDelay(10*time.Minute)` from `waitForDomainCreation()`
- Removed `tfresource.WithDelay(1*time.Minute)` from `waitForDomainUpdate()` (for consistency)
- Removed `tfresource.WithDelay(10*time.Minute)` from `waitForDomainDelete()`
- Kept `tfresource.WithPollInterval(10*time.Second)` to prevent API rate limiting
- Updated changelog entry in `.changelog/45245.txt`

**Domain Creation Performance Test (LocalStack):**
- Before: 600+ seconds (10+ minutes minimum)
- After: ~19 seconds
- **Improvement: ~30x faster**

**Note on Consistency:**
PR #32209 reduced the update delay from 10 minutes to 1 minute. This PR goes further by removing delays entirely from all operations (create/update/delete) for consistency and maximum performance improvement.

**Related:**
- #32189 - Original update delay issue
- #32209 - Update delay fix (reduced to 1 min)
- #28523 - LocalStack timeout discussion
- #44925 - Similar delay issues with RDS

PR #32209, opened 2023-06-25
opensearch update delay: 10 min -> 1 min

dsturnbull:
### Description
https://github.com/hashicorp/terraform-provider-aws/issues/32189

Pragmatic option: reduce minimum OpenSearch update to 1 minute (from 10 minutes).

### Relations
Closes #32189

### References
https://github.com/hashicorp/terraform-provider-aws/issues/32189

### Output from Acceptance Testing
Not sure about tests for this one. No prior delay-specific tests exist for this.

Issue #32189, opened 2023-06-23
[Bug]: opensearch module waits 10 minutes for every update


dsturnbull:
### Terraform Core Version

1.5.0

### AWS Provider Version

5.2.0

### Affected Resource(s)

aws_opensearch_domain

### Expected Behavior

It should take a few seconds maximum to change an access_policy on an existing domain.

### Actual Behavior

```
module.opensearch.aws_opensearch_domain.opensearch: Modifying... [id=arn:aws:es:ap-southeast-2:123456789012:domain/xxx]
```

This is correlated with trace logs showing the provider posts the policy change at `2023-06-23T20:52:04.992+1000`.

We see a response at `2023-06-23T20:52:05.811+1000` (~800ms) later, with `"State":"Processing",` in the response. `waitForDomainUpdate` keeps retrying until state becomes something else.

The problem seems to be that the minimum wait time is 10 minutes. Every access_policy change will be affected. If I ctrl-c terraform and rerun immediately, or verify by hand in the console, it's clear that the policy changed and the domain is not in the `Processing` state.

```
	err := tfresource.Retry(ctx, timeout, func() *retry.RetryError {
		var err error
		out, err = FindDomainByName(ctx, conn, domainName)
		if err != nil {
			return retry.NonRetryableError(err)
		}

		if !aws.BoolValue(out.Processing) {
			return nil
		}

		return retry.RetryableError(
			fmt.Errorf("%q: Timeout while waiting for changes to be processed", domainName))
	}, tfresource.WithDelay(10*time.Minute), tfresource.WithPollInterval(10*time.Second))
```
[introduced here](https://github.com/hashicorp/terraform-provider-aws/blob/d68aa1dddc9d412109323f1d554a744098c4ec3c/internal/service/opensearch/wait.go#L86)

### Terraform Configuration Files

```
resource "aws_opensearch_domain" "example" {
  domain_name     = "example"
  access_policies = var.access_policy

  cluster_config {
    instance_type = "t2.small.search"
  }

  ebs_options {
    ebs_enabled = true
    volume_size = 10
  }
}
```

where `access_policy` is 

```
access_policy = jsonencode({
  "Version" : "2012-10-17",
  "Statement" : [
    {
      "Action" : ["es:ESHttpGet", "es:ESHttpPost"],
      "Principal" : { "AWS" : "*" },
      "Effect" : "Allow",
      "Resource" : [
[...]

Issue #28523, opened 2022-12-21
[Enhancement]: Adjust timing of resource deployments for local testing (e.g., with LocalStack)


whummer:
### Description

The Terraform AWS provider has several timings (delays/timeouts/intervals etc) built into the code base - to perform retries, checks and assertions to ensure that the deployment converges to the desired target state. These timings are (obviously) optimized to work well against the real AWS cloud environment, making the testing experience against local cloud emulators like [LocalStack](https://github.com/localstack/localstack) suboptimal. 

It would be fantastic if there was a way to parameterize/configure the resource timeouts, e.g., by introducing a _local test mode_ optimized for quick development cycles. This would make testing much more time-efficient, and could in some cases reduce the local deployment time of resources from several minutes down to a few seconds. Given that more and more users are testing their AWS Terraform scripts against emulators like LocalStack, this would be huge UX improvement and could further foster the adoption of TF as the de-facto standard for defining IaC for AWS resources.
 
One simple example is creation of S3 buckets - let's assume we deploy the following TF script locally against LocalStack:
```
resource "aws_s3_bucket" "bucket1" {
  bucket = "bucket1"
}
```

Below is a sequence of the API calls made by Terraform against LocalStack.It can be seen that there is a ~5-6 seconds delay between requests, e.g., GetBucketPolicy -> GetBucketAcl -> GetBucketCors ... , leading to overall deployment time of ~70+ seconds.

```
12:25:52.385 - AWS s3.CreateBucket => 200
12:25:57.550 - AWS s3.HeadBucket => 200
12:26:02.644 - AWS s3.GetBucketPolicy => 404 (NoSuchBucketPolicy)
12:26:07.951 - AWS s3.HeadBucket => 200
12:26:07.953 - AWS s3.GetBucketAcl => 200
12:26:13.063 - AWS s3.HeadBucket => 200
12:26:13.065 - AWS s3.GetBucketCors => 404 (NoSuchCORSConfiguration)
12:26:18.289 - AWS s3.GetBucketWebsite => 404 (NoSuchWebsiteConfiguration)
12:26:23.607 - AWS s3.GetBucketVersioning => 200
12:26:28.931 - AWS s3.GetBucketAccelerateConfiguration => 200
12:26:34.004 - AWS s3.HeadBucket => 200
12:26:34.007 - AWS s3.GetBucketRequestPayment => 200
12:26:39.125 - AWS s3.GetBucketLogging => 200
[...]

PR #45480, opened 2025-12-08
rds: Remove delays and ContinuousTargetOccurence from DB instance operations

### LINKED ISSUE #44925: Slow RDS instance creation
### Description

Hello 👋 

When I create an RDS instance via AWS Console, it takes roughly 30 seconds to become available. However, when created via terraform, it takes at least 1 minute 30 seconds.

I traced it down to the following commits:
- https://github.com/hashicorp/terraform-provider-aws/commit/9a914869d05938b735d1bca359fe481e3992f2ac introduced the intentional slowdown - originally committed to terraform repo without a PR or explanation in 2014 https://github.com/hashicorp/terraform/commit/dde83f7cddd7d56ab675719bf31a3b58bbd154fc
- when Blue/Green deployments were introduced, it was slowed down even further by increasing the `Delay` https://github.com/hashicorp/terraform-provider-aws/pull/28046/commits/183be6e8f4c86625f1b33800246f1812de371d5e, again with no explanation

I also noticed that specifying `Delay`, `PollInterval`, and `ContinuousTargetOccurence` seems to be quite unique to RDS resources - e.g. I can't see deployments of MSK being intentionally slowed down. Where this becomes really visible if when used in conjunction with Localstack: https://github.com/hashicorp/terraform-provider-aws/issues/28523

My question is: can anyone provide me with a reasoning behind setting those values? What if we removed them?

_No response_

### Would you like to implement a relevant change?

No

berimbolo13:
Closes #44925

Removes hardcoded delays and continuous target occurrence checks before polling instance state during creation, stop, and deletion operations. The provider will now begin polling immediately with a 10-second interval.

This significantly reduces wait times for:
- Fast AWS provisioning scenarios
- Local testing with LocalStack (lightweight AWS emulator)

**Background:**
These delays were introduced in 2014 without explanation (originally committed to terraform repo without a PR). When Blue/Green deployments were introduced, delays were increased further with no documented rationale.

The `ContinuousTargetOccurence` parameter was added in commit 4e1733ed9e9 (Nov 20, 2024) with only "Add ContinuousTargetOccurence" as the commit message, providing no context for why multiple consecutive checks are needed.

This change removes both the initial delays and the continuous target checks, allowing the provider to begin polling immediately whilst keeping the 10-second poll interval to prevent RDS API rate limiting.

**Related:**
- #28523 - LocalStack timeout discussion
- #32189 - OpenSearch update delay issue
- #32209 - OpenSearch update delay fix
- #45245 - OpenSearch creation/deletion delay fix

### DIFF
--- a/.changelog/44925.txt
+++ b/.changelog/44925.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_db_instance: Remove hardcoded delays and continuous target occurrence checks before polling instance state during creation, stop, and deletion operations
+```
--- a/internal/service/rds/instance.go
+++ b/internal/service/rds/instance.go
@@ -2890,9 +2890,7 @@ func statusDBInstance(conn *rds.Client, id string, optFns ...func(*rds.Options))
 
 func waitDBInstanceAvailable(ctx context.Context, conn *rds.Client, id string, timeout time.Duration, optFns ...tfresource.OptionsFunc) (*types.DBInstance, error) {
 	options := tfresource.Options{
-		PollInterval:              10 * time.Second,
-		Delay:                     1 * time.Minute,
-		ContinuousTargetOccurence: 3,
+		PollInterval: 10 * time.Second,
 	}
 	for _, fn := range optFns {
 		fn(&options)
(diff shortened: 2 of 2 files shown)


kujon:
hi @justinretzolk: is there any appetite to move forward on this? as well as [this](https://github.com/hashicorp/terraform-provider-aws/pull/45245)?

kujon:
hi @justinretzolkm polite bump

alipievg:
@hc-github-team-terraform-aws

PR #36045, opened 2024-03-01
feat: Add evaluation_window_sec arg to rate_based_statement for aws_wafv2_rule_group and aws_wafv2_web_acl

acwwat:
### Description

This PR adds the `evaluation_window_sec` argument to the `rate_based_statement` block for both the `aws_wafv2_rule_group` and `aws_wafv2_web_acl` resources.

### Relations

Closes #36028

### References

Referred to [RateBasedStatement](https://docs.aws.amazon.com/waf/latest/APIReference/API_RateBasedStatement.html#WAF-Type-RateBasedStatement-EvaluationWindowSec) in the API reference for attribute specs and wordings.

### Output from Acceptance Testing

For `aws_wafv2_rule_group`:
```console
$ make testacc TESTS=TestAccWAFV2RuleGroup_rateBasedStatement PKG=wafv2
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/wafv2/... -v -count 1 -parallel 20 -run='TestAccWAFV2RuleGroup_rateBasedStatement'  -timeout 360m
=== RUN   TestAccWAFV2RuleGroup_rateBasedStatement
=== PAUSE TestAccWAFV2RuleGroup_rateBasedStatement
=== CONT  TestAccWAFV2RuleGroup_rateBasedStatement
--- PASS: TestAccWAFV2RuleGroup_rateBasedStatement (320.28s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/wafv2      320.470s

$
```

For `aws_wafv2_web_acl`:
```console
$ make testacc TESTS=TestAccWAFV2WebACL_RateBased_ PKG=wafv2
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/wafv2/... -v -count 1 -parallel 20 -run='TestAccWAFV2WebACL_RateBased_'  -timeout 360m
=== RUN   TestAccWAFV2WebACL_RateBased_basic
=== PAUSE TestAccWAFV2WebACL_RateBased_basic
=== RUN   TestAccWAFV2WebACL_RateBased_customKeys
=== PAUSE TestAccWAFV2WebACL_RateBased_customKeys
=== RUN   TestAccWAFV2WebACL_RateBased_forwardedIP
=== PAUSE TestAccWAFV2WebACL_RateBased_forwardedIP
=== RUN   TestAccWAFV2WebACL_RateBased_maxNested
[...]

PR #45481, opened 2025-12-08
docs: wafv2 rate_based_statement: Fix limit description

fungusakafungus:
### Description

`wafv2_rule_group` and `wafv2_web_acl` docs of `rate_based_statement`: Fix limit description for non-default `evaluation_window_sec` or aggregation key.

The description of limit became incorrect with https://github.com/hashicorp/terraform-provider-aws/pull/36045, when `evaluation_window_sec` became configurable.

Here's AWS description: 

> The limit on requests during the specified evaluation window for a single aggregation instance for the rate-based rule. If the rate-based statement includes a ScopeDownStatement, this limit is applied only to the requests that match the statement.
>
> Examples:
>
> If you aggregate on just the IP address, this is the limit on requests from any single IP address.
>
> If you aggregate on the HTTP method and the query argument name "city", then this is the limit on requests for any single method, city pair.

### Relations

Relates #36045

### References

https://docs.aws.amazon.com/waf/latest/APIReference/API_RateBasedStatement.html


ewbankkit:
@fungusakafungus Thanks for the contribution :tada: :clap:.

PR #45552, opened 2025-12-11
Fix for ram shared vmc delete

pooja-choudhury:
## Rollback Plan

No

### Description

AWS ODB exa vm cluster deletion was failing for RAM shared account  as AWS GO SDK was not returning exa infra arn and odb network arn. Post latest release of aws go sdk, new fields are introduced and same has been incorpoareted in this PR. 

### References
https://docs.aws.amazon.com/odb/latest/UserGuide/sharing-resources-task.html
https://github.com/aws/aws-sdk-go-v2/releases/tag/release-2025-12-10

```console
% make testacc TESTS=TestAccODBCloudVmCluster_disappears PKG=odb 
...
```
Screenshot 2025-12-15 at 17 01 37

% make testacc TESTS=TestAccODBCloudVmClusterDataSource_basic PKG=odb 
Screenshot 2025-12-15 at 18 12 33

### DIFF
--- a/internal/service/odb/cloud_vm_cluster_data_source.go
+++ b/internal/service/odb/cloud_vm_cluster_data_source.go
@@ -50,6 +50,10 @@ func (d *dataSourceCloudVmCluster) Schema(ctx context.Context, req datasource.Sc
 				Computed:    true,
 				Description: "The ID of the Cloud Exadata Infrastructure.",
 			},
+			"cloud_exadata_infrastructure_arn": schema.StringAttribute{
+				Computed:    true,
+				Description: "The ARN of the Cloud Exadata Infrastructure.",
+			},
 			names.AttrClusterName: schema.StringAttribute{
 				Computed:    true,
 				Description: "The name of the Grid Infrastructure (GI) cluster.",
@@ -137,6 +141,10 @@ func (d *dataSourceCloudVmCluster) Schema(ctx context.Context, req datasource.Sc
 			"odb_network_id": schema.StringAttribute{
 				Computed:    true,
 				Description: "The ID of the ODB network.",
+			},
+			"odb_network_arn": schema.StringAttribute{
+				Computed:    true,
+				Description: "The ARN of the ODB network.",
 			},
 			"percent_progress": schema.Float64Attribute{
 				Computed:    true,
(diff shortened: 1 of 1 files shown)


pooja-choudhury:
@johnsonaj Can you please take a look. Its an urgent PR to fix customer reported issue.

internal/service/odb/cloud_vm_cluster.go
```diff
@@ -576,40 +576,10 @@ func (r *resourceCloudVmCluster) Create(ctx context.Context, req resource.Create
...
-			err.Error(),
-		)
-		return
-	}
-	plan.CloudExadataInfrastructureId = flex.StringToFramework(ctx, cloudExadataInfraOutput.CloudExadataInfrastructure.CloudExadataInfrastructureId)
-	plan.CloudExadataInfrastructureArn = flex.StringToFramework(ctx, cloudExadataInfraOutput.CloudExadataInfrastructure.CloudExadataInfrastructureArn)
+	plan.OdbNetworkId = flex.StringToFramework(ctx, createdVmCluster.OdbNetworkId)
+	plan.OdbNetworkArn = flex.StringToFramework(ctx, createdVmCluster.OdbNetworkArn)
+	plan.CloudExadataInfrastructureId = flex.StringToFramework(ctx, createdVmCluster.CloudExadataInfrastructureId)
+	plan.CloudExadataInfrastructureArn = flex.StringToFramework(ctx, createdVmCluster.CloudExadataInfrastructureArn)
```
johnsonaj, line 582: @pooja-choudhury we should be able to remove these lines of code and let the `flex.Flatten(ctx, createdVmCluster, &plan)` handle setting these values. Since the structure names are identical then it should just work

internal/service/odb/cloud_vm_cluster.go
```diff
@@ -652,32 +622,10 @@ func (r *resourceCloudVmCluster) Read(ctx context.Context, req resource.ReadRequ
...
-			err.Error(),
-		)
-		return
-	}
-	state.CloudExadataInfrastructureId = flex.StringToFramework(ctx, cloudExadataInfraOutput.CloudExadataInfrastructure.CloudExadataInfrastructureId)
-	state.CloudExadataInfrastructureArn = flex.StringToFramework(ctx, cloudExadataInfraOutput.CloudExadataInfrastructure.CloudExadataInfrastructureArn)
+	state.OdbNetworkId = flex.StringToFramework(ctx, out.OdbNetworkId)
+	state.OdbNetworkArn = flex.StringToFramework(ctx, out.OdbNetworkArn)
+	state.CloudExadataInfrastructureId = flex.StringToFramework(ctx, out.CloudExadataInfrastructureId)
+	state.CloudExadataInfrastructureArn = flex.StringToFramework(ctx, out.CloudExadataInfrastructureArn)
```
johnsonaj, line 628: same as above. `flex` should take care of setting these

johnsonaj:
@pooja-choudhury just a few small comments.

internal/service/odb/cloud_vm_cluster.go
pooja-choudhury: @johnsonaj I have tried the same but it gives error while testing, thats why I left it there. As its not going to break anything, can we go ahead with this block here.

Screenshot 2025-12-16 at 10 44 53

internal/service/odb/cloud_vm_cluster.go
pooja-choudhury: @johnsonaj I have tried the same but it gives error while testing, thats why I left it there. As its not going to break anything, can we go ahead with this block here.

Screenshot 2025-12-16 at 10 44 53

internal/service/odb/cloud_vm_cluster.go
johnsonaj: Thanks for confirming that the explicit flex function is setting a value that auto flex is not

johnsonaj:
@pooja-choudhury thank you for the contribution! 🎉

PR #45470, opened 2025-12-06
[Enhancement] aws_s3vectors_index: Add `metadata_configuration` and `encryption_configuration` blocks

tabito-hara:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description
This PR adds `metadata_configuration` and `encryption_configuration` blocks to `aws_s3vectors_index` resource.

### Relations

Closes #45447 

### References
https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_CreateIndex.html

### Output from Acceptance Testing
```console
$ make testacc TESTS='TestAccS3VectorsIndex_' PKG=s3vectors 
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
make: Running acceptance tests on branch: 🌿 f-aws_s3vectors_idnex-add_metatdata_configuration 🌿...
TF_ACC=1 go1.24.11 test ./internal/service/s3vectors/... -v -count 1 -parallel 20 -run='TestAccS3VectorsIndex_'  -timeout 360m -vet=off
2025/12/06 23:31:18 Creating Terraform AWS Provider (SDKv2-style)...
2025/12/06 23:31:18 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccS3VectorsIndex_Identity_Basic
=== PAUSE TestAccS3VectorsIndex_Identity_Basic
=== RUN   TestAccS3VectorsIndex_Identity_RegionOverride
=== PAUSE TestAccS3VectorsIndex_Identity_RegionOverride
=== RUN   TestAccS3VectorsIndex_tags
=== PAUSE TestAccS3VectorsIndex_tags
=== RUN   TestAccS3VectorsIndex_tags_null
=== PAUSE TestAccS3VectorsIndex_tags_null
=== RUN   TestAccS3VectorsIndex_tags_EmptyMap
=== PAUSE TestAccS3VectorsIndex_tags_EmptyMap
=== RUN   TestAccS3VectorsIndex_tags_AddOnUpdate
=== PAUSE TestAccS3VectorsIndex_tags_AddOnUpdate
=== RUN   TestAccS3VectorsIndex_tags_EmptyTag_OnCreate
=== PAUSE TestAccS3VectorsIndex_tags_EmptyTag_OnCreate
[...]

PR #45558, opened 2025-12-11
docs(s3vectors_index): use correct attribute name kms_key_arn

evan-cohen:
## Rollback Plan

### Description

Resolves an inconsistency between the documentation and implementation for the `encryption_configuration` block of the `aws_s3vectors_index` resource as introduced in v6.26.0.

The enhancement to the `aws_s3vectors_index` resource in https://github.com/hashicorp/terraform-provider-aws/pull/45470 includes an `encryption_configuration` block with a `kms_key_arn` attribute:

https://github.com/hashicorp/terraform-provider-aws/blob/b6c16ac16873c26d130e27e838155525d8c98ddf/internal/service/s3vectors/index.go#L297-L300

https://github.com/hashicorp/terraform-provider-aws/blob/b6c16ac16873c26d130e27e838155525d8c98ddf/internal/service/s3vectors/index_test.go#L418-L430

...but the corresponding documentation uses the attribute name "kms_key_id" instead:
https://github.com/hashicorp/terraform-provider-aws/blob/b6c16ac16873c26d130e27e838155525d8c98ddf/website/docs/r/s3vectors_index.html.markdown#L47-L50

### Relations
Relates #45470

### DIFF
--- a/website/docs/r/s3vectors_index.html.markdown
+++ b/website/docs/r/s3vectors_index.html.markdown
@@ -46,7 +46,7 @@ The following arguments are optional:
 
 The `encryption_configuration` block supports the following attributes:
 
-* `kms_key_arn` - (Optional, Forces new resource) AWS Key Management Service (KMS) customer managed key ID to use for the encryption configuration. This parameter is allowed if and only if `sse_type` is set to `aws:kms`.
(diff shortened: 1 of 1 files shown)


ewbankkit:
@evan-cohen Thanks for the contribution :tada: :clap:.

PR #45577, opened 2025-12-15
docs: clarify user_group_id supports Redis and Valkey

### LINKED ISSUE #44796: Incorrectly states user groups are not supported with serverless Valkey
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_serverless_cache#user_group_id-5

"The identifier of the UserGroup to be associated with the serverless cache. Available for Redis only. Default is NULL."

### Description

AWS docs state that user groups are available for both Redis and Valkey serverless.

### References

AWS docs state that user groups are available for both Redis and Valkey serverless.
https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ServerlessCache.html

AWS console also provides a way to setup user groups.

It's worth noting the tests only validate user group setup with Redis.
https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/elasticache/serverless_cache_test.go#L807-L842

No

ElsaDevOps:
Closes #44796 

The documentation stated `user_group_id` was available for Redis only, 
but AWS docs confirm it works for both Redis and Valkey.

https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ServerlessCache.html


ewbankkit:
@ElsaDevOps Thanks for the contribution :tada: :clap:.

Issue #43848, opened 2025-08-12
`aws_mpa_approval_team`


m0un10:
### What new functionality are you requesting?

AWS have recently launched multi-party approvals within IAM Identity Center.

The CLI commands are available at `aws mpa help`.

At present these resources are not exposed as Terraform resources. At a glance, it appears that the most useful resource to expose would be the approval team.

### Description

The new functionality would provide resources to support:

- [Create team](https://docs.aws.amazon.com/mpa/latest/userguide/create-team.html)
- [View team](https://docs.aws.amazon.com/mpa/latest/userguide/admin-view-team.html)
- [Update team](https://docs.aws.amazon.com/mpa/latest/userguide/update-team.html)
- [Delete team](https://docs.aws.amazon.com/mpa/latest/userguide/delete-team.html)

### Potential Terraform Configuration

```hcl

```

### Would you like to implement the enhancement?

Yes

jackroberts-gh:
PR is open here to add the service*: https://github.com/hashicorp/terraform-provider-aws/pull/45594

PR #45594, opened 2025-12-15
s/mpa: Add `mpa` (Multi-party Approval) service

jackroberts-gh:
## Rollback Plan

### Changes to Security Controls

N/A

### Description

Adds a new Service `mpa` as per instructions in [docs](https://hashicorp.github.io/terraform-provider-aws/add-a-new-service/#adding-a-new-aws-service).

### Relations

Relates #43848 as a pre-requisite.

### References

https://github.com/hashicorp/terraform-provider-aws/issues/43848

N/A

### Other Notes

`make gen` made some updates seemingly unrelated to `mpa` - see [commit](https://github.com/hashicorp/terraform-provider-aws/pull/45594/commits/06504d47215264a7c256fe39618f14be55235a58) (not all changes in that commit, but some)

### DIFF
--- a/internal/generate/serviceendpointtests/main.go
+++ b/internal/generate/serviceendpointtests/main.go
@@ -83,7 +83,7 @@ func main() {
 
 		switch packageName {
 		// TODO: This case should be handled in service data
-		case "costoptimizationhub", "cur", "globalaccelerator", "notifications", "notificationscontacts", "route53domains", "route53recoverycontrolconfig", "route53recoveryreadiness":
+		case "costoptimizationhub", "cur", "globalaccelerator", "mpa", "notifications", "notificationscontacts", "route53domains", "route53recoverycontrolconfig", "route53recoveryreadiness":
 			td.OverrideRegionRegionalEndpoint = true
 
 		case "chatbot":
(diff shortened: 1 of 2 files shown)


jackroberts-gh:
❌ 🆗  **Dependency Changes** - I believe these are legitimate when adding a new service, following the "Adding a new service docs".

:x: 🆘  **Changelog Entry Required** - Not clear if a CHANGELOG file is needed for this type of change? Or whether it would only be needed for the resources that come later.

jackroberts-gh:
`go test` seems to be failing because the test wants to use expected provider with region set to `us-west-2`, but the service is `us-east-1` only, which I specified in `names`.

 Unsure what to do in this case - I could do with some guidance! I can see that `arcregionswitch` is a similar service (us-east-1, not global) and it doesn't seem to have a `_test.go` file for it's service. However, I can't see how that is controlled or if that's the correct approach? Thanks

jar-b:
Hey @jackroberts-gh - Our service endpoint test generation logic has some lingering technical debt which requires some services which only resolve to a single regional endpoint to make a modification to the generator itself (see https://github.com/hashicorp/terraform-provider-aws/pull/45594/commits/d468f86fc1b2a42f5ef313c5933a7a222b0271a3). As service additions are infrequent, refactoring this process hasn't been prioritized yet, but this change should unblock this PR. Thanks for your effort on this.

jar-b:
Thanks for your contribution, @jackroberts-gh! 👍

Issue #19138, opened 2021-04-28
Add OpenBSD amd64 support


fallertsen:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Looks like there was OpenBSD support for this provider, but the support was dropped on version 3.0.0, but I couldn't find any reason why. To re-enable the support it should be as simple as adding openbsd as goos in the `.goreleaser.yml` file.

### References

https://registry.terraform.io/v1/providers/hashicorp/aws/versions

This shows that all the releases up until 3.0.0 openbsd/amd64 was supported.

georgeleege:
I would like to work on this. Do I just need to add openbsd to the `.goreleaser.yml` file and run the local and acceptance tests from an OpenBSD machine?

fallertsen:
I think that should be the place to start. Thanks a lot.

knightjoel:
>
```

**Comment by knightjoel** [comment]
`make test` output is below. I'm uncertain about the cost implications of running the acceptance tests so haven't done that yet.

`OpenBSD build.packetmischief.ca 6.9 GENERIC.MP#1 amd64`

```
==> Checking that code complies with gofmt requirements...
go test ./...  -timeout=5m -parallel=4
?   	github.com/terraform-providers/terraform-provider-aws	[no test files]
ok  	github.com/terraform-providers/terraform-provider-aws/aws	(cached)
?   	github.com/terraform-providers/terraform-provider-aws/aws/internal/attrmap	[no test files]
?   	github.com/terraform-providers/terraform-provider-aws/aws/internal/encryption	[no test files]
ok  	github.com/terraform-providers/terraform-provider-aws/aws/internal/envvar	(cached)
ok  	github.com/terraform-providers/terraform-provider-aws/aws/internal/experimental/nullable	(cached)
?   	github.com/terraform-providers/terraform-provider-aws/aws/internal/experimental/sync	[no test files]
ok  	github.com/terraform-providers/terraform-provider-aws/aws/internal/hashcode	(cached)
ok  	github.com/terraform-providers/terraform-provider-aws/aws/internal/json	(cached)
[...]

PR #45609, opened 2025-12-16
Fix build.yml?

YakDriver:
## Rollback Plan

### Relations

Closes #0000

```console

...
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,7 +123,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        goos: [freebsd, windows, linux, darwin]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -123,7 +123,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        goos: [freebsd, windows, linux, darwin, openbsd]
+        goos: [freebsd, windows, linux, darwin]
```
johnsonaj: do we not need `openbsd` anymore? I remember us adding this a while back: https://github.com/hashicorp/terraform-provider-aws/issues/19138

PR #45246, opened 2025-11-25
Feat: Cloud WAN routing policy

pooriaghaedi:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to security controls.

### Description

Adds support for the `routing_policy_label` argument to three AWS Network Manager attachment resources and adds documentation for the new core network policy version `2025.11` features.

**Attachment Resources Enhanced:**
- `aws_networkmanager_site_to_site_vpn_attachment`
- `aws_networkmanager_dx_gateway_attachment` 
- `aws_networkmanager_transit_gateway_route_table_attachment`
- `aws_networkmanager_connect_attachment`
- `aws_networkmanager_vpc_attachment`

The `routing_policy_label` argument allows users to specify a routing policy label for traffic routing decisions when creating attachments. This is an optional, write-only field that forces resource recreation when changed.

**Documentation Updates:**
- Added comprehensive documentation for `routing_policies` block in core network policy document data source
- Documented new `edge_location_association` for segment actions
- Updated attachment policy condition types (`account-id` → `account`, added `tag-name`)
- Added new segment action type `associate-routing-policy`
- Documented policy version `2025.11` support

### Relations

Closes #45194.
Closes https://github.com/hashicorp/terraform-provider-aws/issues/40564.
Closes https://github.com/hashicorp/terraform-provider-aws/pull/40603.

### References

- [AWS Cloud WAN Routing Policies Documentation](https://docs.aws.amazon.com/vpc/latest/cloudwan/cloudwan-routing-policies.html)
- [AWS Network Manager API Reference](https://docs.aws.amazon.com/networkmanager/latest/APIReference/Welcome.html)

### Output from Acceptance Testing
=== RUN   TestAccNetworkManagerCoreNetworkPolicyAttachment_routingPolicies
=== PAUSE TestAccNetworkManagerCoreNetworkPolicyAttachment_routingPolicies
=== CONT  TestAccNetworkManagerCoreNetworkPolicyAttachment_routingPolicies
--- PASS: TestAccNetworkManagerCoreNetworkPolicyAttachment_routingPolicies (668.16s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/networkmanager	674.457s
</stdout>

<timestamp>2025-11-23T09:51:31.411Z</timestamp>

% make testacc TESTS=TestAccNetworkManagerTransitGatewayRouteTableAttachment_routingPolicyLabel PKG=networkmanager
[...]

PR #45617, opened 2025-12-16
Handle 64bit on 32bit, broken builds

YakDriver:
## Rollback Plan

### Description

The build error occurred because `validation.IntBetween(1, 4294967295)` was trying to pass the literal `4294967295` (which equals 2^32- 1, the maximum 32-bit unsigned integer value) as an `int` parameter to the validation function. On 32-bit architectures, Go's `int` type is only 32 bits and has a maximum value of 2^31 - 1 (2147483647), so the literal `4294967295` overflows when used as an `int` argument. The fix creates a new `IntBetween64` validation function that accepts `int64` parameters instead of `int` parameters, allowing it to handle the full 32-bit ASN range (1-4294967295) without overflow issues. The function still validates `int` values from the schema, but internally converts them to `int64` for range checking, maintaining backward compatibility while supporting the complete ASN value range on all architectures.

### Relations

Problem introduced by #45246

```console

...
```

### DIFF
--- a/internal/service/networkmanager/connect_peer.go
+++ b/internal/service/networkmanager/connect_peer.go
@@ -6,6 +6,7 @@ package networkmanager
 import (
 	"context"
 	"log"
+	"math"
 	"time"
 
 	"github.com/YakDriver/regexache"
@@ -64,7 +65,7 @@ func resourceConnectPeer() *schema.Resource {
 						"peer_asn": {
 							Type:         schema.TypeInt,
 							Optional:     true,
-							ValidateFunc: verify.IntBetween64(1, 1<<32-1), // 4294967295
+							ValidateFunc: verify.IntBetween64(1, math.MaxUint32),
 						},
 					},
 				},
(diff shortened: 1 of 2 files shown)


YakDriver:
Couldn't find SDK int64 validator. If there is one, we could use that instead. But, one thing we know, `validation.IntBetween(1, 4294967295)` is DOA due to `int` arg.

internal/service/networkmanager/connect_peer.go
```diff
@@ -63,7 +64,7 @@ func resourceConnectPeer() *schema.Resource {
 						"peer_asn": {
 							Type:         schema.TypeInt,
 							Optional:     true,
-							ValidateFunc: validation.IntBetween(1, 4294967295),
+							ValidateFunc: verify.IntBetween64(1, 1<<32-1), // 4294967295
```
johnsonaj: There is a const in the in `math` package for this: `math.MaxUint32`

johnsonaj (dismissed):
LGTM 🚀

johnsonaj (approved):

PR #45674, opened 2025-12-20
docs(elasticache_replication_group): clarify auth_token_update_strategy

### LINKED ISSUE #45310: Elasticache auth_token_update_strategy setting incorrectly stated as required when auth_token is set?
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticache_replication_group#auth_token_update_strategy-1

### Description

With the last major upgrade the `auth_token_update_strategy` setting default was changed to null and the release guide / documentation updated to state that this setting must be set to one of it's options whenever the `auth_token` setting was set.

In practice though this doesn't look to be the case. I can continue to deploy an elasticache instance with only `auth_token` set, my existing ones also upgraded fine.

The docs should be reviewed to give clear guidance on what is actually best to set here.

### References

_No response_

No

AnuragEkkati:
## Rollback Plan

Revert this commit (or follow up with a PR) to restore the prior documentation wording.

No. Documentation-only change; no provider behavior changes.

### Description

Clarifies the `auth_token_update_strategy` argument documentation for `aws_elasticache_replication_group`.

The previous text implied `auth_token_update_strategy` is required whenever `auth_token` is set. AWS only applies an auth token update strategy when modifying an existing replication group’s auth token (ModifyReplicationGroup). Initial create uses CreateReplicationGroup, which supports `AuthToken` but does not accept an `AuthTokenUpdateStrategy` parameter.

### Relations

Closes #45310

### References

- AWS API Reference: ModifyReplicationGroup (AuthTokenUpdateStrategy semantics/defaults)  
  https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ModifyReplicationGroup.html
- AWS API Reference: CreateReplicationGroup (AuthToken; no AuthTokenUpdateStrategy)  
  https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html
- (Optional) ElastiCache AUTH user guide  
  https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/auth.html

Docs-only change.

```console
% make website-lint
# (pass)

% make website-link-check
# (pass)   # or: not run locally (docker not available); CI will validate

### DIFF
--- a/website/docs/r/elasticache_replication_group.html.markdown
+++ b/website/docs/r/elasticache_replication_group.html.markdown
@@ -232,7 +232,7 @@ The following arguments are optional:
   When `engine` is `redis`, default is `false`.
   When `engine` is `valkey`, default is `true`.
 * `auth_token` - (Optional) Password used to access a password protected server. Can be specified only if `transit_encryption_enabled = true`.
-* `auth_token_update_strategy` - (Optional) Strategy to use when updating the `auth_token`. Valid values are `SET`, `ROTATE`, and `DELETE`. Required if `auth_token` is set.
+* `auth_token_update_strategy` - (Optional) Strategy used when modifying `auth_token` on an existing replication group. Not used during initial create. Valid values are `SET`, `ROTATE`, and `DELETE`. If omitted during an auth token change, AWS defaults to `ROTATE`.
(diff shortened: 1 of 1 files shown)


ewbankkit:
@AnuragEkkati Thanks for the contribution :tada: :clap:.

PR #45747, opened 2025-12-28
docs: Change import ID from account ID to region for r/aws_glue_resource_policy

### LINKED ISSUE #45568: aws_glue_registry_provider import requires account ID when region is required
### Documentation Link(s)

https://registry.terraform.io/providers/hashicorp/aws/6.26.0/docs/resources/glue_resource_policy#import

### Description

The documentation for aws_glue_resource_policy indicates that the import process should use the AWS account ID in the `id` field of an import. This results in errors, because the `id` field value is being used by the provider to build out the AWS endpoint that it uses to perform a [GetResourcePolicy](https://docs.aws.amazon.com/glue/latest/webapi/API_GetResourcePolicy.html) call:

```
import {
  to = aws_glue_resource_policy.mypolicy
  id = "1234567890"
}
```

> Error: reading Glue Resource Policy (1234567890): operation error Glue: GetResourcePolicy, https response error StatusCode: 0, RequestID: , request send failed, Post "https://glue.1234567890.amazonaws.com/": dial tcp: lookup glue.1234567890.amazonaws.com on 10.10.10.10:53: no such host

Where 1234567890 is the AWS account ID and 10.10.10.10:53 is the dns server for the TFE executer.

The referenced endpoint seems to be an attempt to build the [AWS endpoint](https://docs.aws.amazon.com/general/latest/gr/glue.html) for Glue, but is using the account ID instead of the region.

When the region is used in the import block, the import succeeds.
```
import {
  to = aws_glue_resource_policy.mypolicy
  id = "us-east-1"
}
```

Technically, the provider knows (or can know) both the region it operates in and the account ID, so the requirement for the GetResourcePolicy call, to build a resource ARN for the parameter, and to know what API endpoint to request this from, could be done without an ID at all, but either way, using the account ID doesn't work, and using the region does work, making it likely that the intended value for the provider is actually the region.

### References

https://docs.aws.amazon.com/glue/latest/webapi/API_GetResourcePolicy.html
https://docs.aws.amazon.com/general/latest/gr/glue.html

No

acwwat:
## Rollback Plan

n/a

### Description

This PR is to fix the import ID in the `aws_glue_resource_policy` resource documentation to use the region instead of the account ID as the import ID. This resource is a regional singleton as indicated by the `@SingletonIdentity` annotation, so the import ID must behave in a certain way.

### Relations

Closes #45568

### References

Referred to the source code (such as `singleton.go`) to understand the import behavior.

```console
n/a
```

### DIFF
--- a/website/docs/r/glue_resource_policy.html.markdown
+++ b/website/docs/r/glue_resource_policy.html.markdown
@@ -51,17 +51,17 @@ This resource exports no additional attributes.
 
 ## Import
 
-In Terraform v1.5.0 and later, use an [`import` block](https://developer.hashicorp.com/terraform/language/import) to import Glue Resource Policy using the account ID. For example:
+In Terraform v1.5.0 and later, use an [`import` block](https://developer.hashicorp.com/terraform/language/import) to import Glue Resource Policy using the region where the resource resides. For example:
 
 ```terraform
 import {
   to = aws_glue_resource_policy.Test
-  id = "12356789012"
+  id = "us-east-1"
 }
 ```
 
-Using `terraform import`, import Glue Resource Policy using the account ID. For example:
+Using `terraform import`, import Glue Resource Policy using the region where the resource resides. For example:
 
 ```console
-% terraform import aws_glue_resource_policy.Test 12356789012
+% terraform import aws_glue_resource_policy.Test us-east-1
 ```


ewbankkit:
@acwwat Thanks for the contribution :tada: :clap:.

PR #45971, opened 2026-01-14
ec2: fix updates for aws_vpc_security_group_{ingress,egress}_rule with AccountId/sg-id format

### LINKED ISSUE #30664: [Bug]: aws_vpc_security_group_ingress_rule should not send UserId in ModifySecurityGroupRules request
### Terraform Core Version

1.1.7

### AWS Provider Version

4.61.0

### Affected Resource(s)

* aws_vpc_security_group_ingress_rule

### Expected Behavior

Cross-Account security group ingress rules should be able to update and not contain a diff once updated.

### Actual Behavior

If the AccountId is specified in `referenced_security_group_id`, the `ModifySecurityGroupRules` request will fail with `InvalidGroupId.Malformed`.

If the AccountId is _not_ specified in `referenced_security_group_id`, the `ModifySecurityGroupIngressRules` request works as expected (AWS TAM confirmed the API is documented incorrectly and should not be passed the AccountId), but terraform views this as a diff and will always have a non-empty plan even though the update of `<accountId>/<sgId>` -> `<sgId>` is a noop.

```shell
Error: updating VPC Security Group Rule (sgr-XYZ)

  with aws_vpc_security_group_ingress_rule.XYZ,
  on XYZ.tf line 377, in resource "aws_vpc_security_group_ingress_rule" "XYZ":
 377: resource "aws_vpc_security_group_ingress_rule" "XYZ" {

InvalidGroupId.Malformed: Invalid id: "ACCOUNTID/sg-XYZ"
```

This is not the actual file, but it will trigger the error.

```
resource "aws_security_group" "some-sg" {
  name        = "some-sg"
  description = "testing testing"
  vpc_id      = data.aws_vpc.some-vpc.id

  tags = {
    Name = "some-sg"
  }
}

resource "aws_vpc_security_group_ingress_rule" "some-sg-rule" {
  description                  = "testing testing"
  security_group_id            = aws_security_group.some-sg.id
  ip_protocol                  = "tcp"
  from_port                    = 80
  to_port                      = 80
  referenced_security_group_id = "<otherAccountId>/sg-XYZ"
}
```

### Steps to Reproduce

1. create a security group in another account
2. reference that security group in a rule and create the rule (this works because the create API accepts UserId)
3. try to change the rule's description (this fails because `ModifySecurityGroupRules` should not have UserId/GroupId, it only needs GroupId)

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

I have talked to our AWS support and they mentioned that the documentation needs to be updated. Both the create and update here do _not_ need UserId in the request. To confirm this, I completely omitted the account IDs in `referenced_security_group_id` and I was able to create new rules and modify them just fine. The problem is that terraform views this as a diff and has a non-empty plan once the update goes through. I suggest modifying:

https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_security_group_ingress_rule.go#L412

and possibly

https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_security_group_ingress_rule.go#L427

to remove `UserId/GroupId` and just switch to `GroupId` only. If modifying both is seen to potentially break too many users, modifying the first place to only send `GroupId` will solve the issue without changing user-facing terraform behavior.

### References

_No response_

None

gustavo-gustavo-gustavo:
## Rollback Plan

No.

### Description

When using `aws_vpc_security_group_ingress_rule` and `aws_vpc_security_group_egress_rule` to reference cross-account security groups, there are two possible scenarios:

1. **Specifying `AccountId/sg-id`**

  If `referenced_security_group_id` is set using the `AccountId/sg-id` format, updates will fail.
  The provider calls [`ModifySecurityGroupRules`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifySecurityGroupRules.html) ([call site](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_security_group_ingress_rule.go#L320-L356)), which expects only `sg-id`, so passing `AccountId/sg-id` causes `InvalidGroupId.Malformed`.

2. **Specifying only `sg-id` (no account ID)**

  If users avoid the prefix and configure only `sg-id`, updates can succeed, but Terraform can fall into a plan loop.
  During refresh/read, AWS returns the referenced group as two fields (`UserId` and `GroupId`), and the provider can represent the value as `AccountId/sg-id`.
  That makes state drift away from config (`sg-id` != `AccountId/sg-id`), so Terraform plans an update every run.

Example AWS response for a cross-account referenced SG rule:

```json
"ReferencedGroupInfo": {
  **"GroupId": "sg-42400000",** # sg-id
  "PeeringStatus": "active",
  **"UserId": "170099999999",** # account id
  "VpcId": "vpc-e5******", 
  "VpcPeeringConnectionId": "pcx-0dc*********"
}
```

This PR fixes both resources by:

- **Normalizing update requests**: if `referenced_security_group_id` is in `AccountId/sg-id` format, send only the `sg-id` portion to `ModifySecurityGroupRules`.
- **Stabilizing state**: preserve the user-specified format during refresh to avoid diffs/plan loops.

Acceptance tests were added to cover the cross-account peered-VPC scenario for both ingress and egress.

### Relations

Closes #30664

### References

- https://github.com/hashicorp/terraform-provider-aws/issues/30664
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifySecurityGroupRules.html

```console
# don't forget to also set AWS_ALTERNATE_PROFILE for vpc peering
% make testacc TESTS=TestAccVPCSecurityGroupIngressRule_ReferencedSecurityGroupID_peerVPCUpdate PKG=ec2
make: Running acceptance tests on branch: 🌿 b-vpc-sg-rule-cross-account-fix 🌿...
TF_ACC=1 go1.25.5 test ./internal/service/ec2/... -v -count 1 -parallel 20 -run='TestAccVPCSecurityGroupIngressRule_ReferencedSecurityGroupID_peerVPCUpdate'  -timeout 360m -vet=off
2026/01/14 13:41:28 Creating Terraform AWS Provider (SDKv2-style)...
2026/01/14 13:41:28 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccVPCSecurityGroupIngressRule_ReferencedSecurityGroupID_peerVPCUpdate
=== PAUSE TestAccVPCSecurityGroupIngressRule_ReferencedSecurityGroupID_peerVPCUpdate
=== CONT  TestAccVPCSecurityGroupIngressRule_ReferencedSecurityGroupID_peerVPCUpdate
--- PASS: TestAccVPCSecurityGroupIngressRule_ReferencedSecurityGroupID_peerVPCUpdate (96.85s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ec2        97.029s
...
```

### DIFF
--- a/internal/service/ec2/vpc_security_group_egress_rule_test.go
+++ b/internal/service/ec2/vpc_security_group_egress_rule_test.go
@@ -187,63 +187,63 @@ resource "aws_vpc_security_group_egress_rule" "test" {
 func testAccVPCSecurityGroupEgressRuleConfig_referencedSecurityGroupIDPeerVPCWithDescription(rName, description string) string {
 	return acctest.ConfigCompose(acctest.ConfigAlternateAccountProvider(), testAccVPCSecurityGroupRuleConfig_base(rName), fmt.Sprintf(`
 resource "aws_vpc" "peer" {
-	provider = "awsalternate"
+  provider = "awsalternate"
 
-	cidr_block = "10.1.0.0/16"
+  cidr_block = "10.1.0.0/16"
 
-	tags = {
-		Name = %[1]q
-	}
+  tags = {
+    Name = %[1]q
+  }
 }
 
 resource "aws_security_group" "peer" {
-	provider = "awsalternate"
+  provider = "awsalternate"
 
-	vpc_id = aws_vpc.peer.id
-	name   = %[1]q
+  vpc_id = aws_vpc.peer.id
+  name   = %[1]q
 
-	tags = {
-		Name = %[1]q
-	}
+  tags = {
+    Name = %[1]q
+  }
 }
 
 data "aws_caller_identity" "peer" {
-	provider = "awsalternate"
+  provider = "awsalternate"
 }
 
 resource "aws_vpc_peering_connection" "test" {
-	vpc_id        = aws_vpc.test.id
-	peer_vpc_id   = aws_vpc.peer.id
-	peer_owner_id = data.aws_caller_identity.peer.account_id
-	peer_region   = %[2]q
-	auto_accept   = false
+  vpc_id        = aws_vpc.test.id
+  peer_vpc_id   = aws_vpc.peer.id
+  peer_owner_id = data.aws_caller_identity.peer.account_id
+  peer_region   = %[2]q
+  auto_accept   = false
 
-	tags = {
-		Name = %[1]q
-	}
+  tags = {
+    Name = %[1]q
+  }
 }
 
 resource "aws_vpc_peering_connection_accepter" "peer" {
-	provider = "awsalternate"
+  provider = "awsalternate"
 
-	vpc_peering_connection_id = aws_vpc_peering_connection.test.id
-	auto_accept               = true
+  vpc_peering_connection_id = aws_vpc_peering_connection.test.id
+  auto_accept               = true
 
-	tags = {
-		Name = %[1]q
-	}
+  tags = {
+    Name = %[1]q
+  }
 }
 
 resource "aws_vpc_security_group_egress_rule" "test" {
-	security_group_id = aws_security_group.test.id
+  security_group_id = aws_security_group.test.id
 
-	referenced_security_group_id = "${data.aws_caller_identity.peer.account_id}/${aws_security_group.peer.id}"
-	description                  = %[3]q
-	from_port                    = 80
-	ip_protocol                  = "tcp"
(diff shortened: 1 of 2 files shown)


gustavo-gustavo-gustavo:
@justinretzolk should this be tagged as services/ec2 instead of services/vpc?

PR #45976, opened 2026-01-14
Fixes calling pattern for `yield` in iterators

gdavison:
## Rollback Plan

### Description

Fixes calling pattern for `yield` in iterators


jar-b (approved):
LGTM 🚀

PR #46055, opened 2026-01-20
Test generation: Import `github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema` when using alternate accounts

ewbankkit:
## Rollback Plan

### Description

Explicitly import `github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema` when using alternate accounts, else `goimports` has to search for a `schema` package and may end up mistakenly importing `github.com/hashicorp/terraform-plugin-sdk/helper/schema`.
Required for

```go
providers := make(map[string]*schema.Provider)
```

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/46013.


internal/generate/tests/annotations.go
```diff
@@ -331,6 +331,9 @@ func ParseTestingAnnotations(args common.Args, stuff *CommonArgs) error {
 				Code: "acctest.PreCheckAlternateAccount(t)",
 			})
 			stuff.GoImports = append(stuff.GoImports,
+				common.GoImport{
+					Path: "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema",
+				},
```
ewbankkit, line 337:
```suggestion
				// Required to initialize `schema.Provider` map.
				common.GoImport{
					Path: "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema",
				},
```

PR #46244, opened 2026-02-01
docs: Fix casing of target_storage_class for ecr_lifecycle_policy_document

critchtionary:
## Rollback Plan

None

### Description

Correct casing of `target_storage_class` in docs for `ecr_lifecycle_policy_document` as specifying the current casing of  `targetStorageClass` does not work and results in an error: `An argument named "targetStorageClass" is not expected here.`


jar-b:
Thank you for the correction, @critchtionary! 👍

PR #46262, opened 2026-02-02
`s3`: Deprecates `expected_bucket_owner` attribute on Bucket configuration sub-resource types

gdavison:
## Rollback Plan

### Description

The `ExpectedBucketOwner` parameter for S3 Bucket configuration sub-resources will always be the owning account ID, and therefore the current account ID. Setting it is redundant.

The attributes will be removed in v7.0.0. #46292


jar-b:
Just needs a corresponding issue in the `v7.0.0` milestone to track removal.

PR #46333, opened 2026-02-05
Fix primary_key_arn reference in KMS replica key

VirtualTim:
## Rollback Plan

### Description
Fix a reference in the documentation. I'm pretty sure this should be `aws_kms_external_key`, not `aws_kms_external`.


jar-b:
Thank you for the correction, @VirtualTim! 👍

PR #46351, opened 2026-02-05
docs(lakeformation): Improve aws_lakeformation_data_cells_filter docu…

### LINKED ISSUE #41376: [Docs]: aws_lakeformation_data_cells_filter disambiguation for argument configuration blocks
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lakeformation_data_cells_filter

### Description

I'm struggling with implementing this filter based on iteration through a local variable and conditionals on the arguments. In some cases we have excluded columns and row filters, sometimes just excluded columns and sometimes data filter only. When a data filter has excluded columns set and data filter set to null, I then get the `Exactly one of these attributes must be configured: [table_data[0].row_filter[0].filter_expression,table_data[0].row_filter[0].all_rows_wildcard]` but it's not shown how to use the all_rows_wildcard argument.

1. There are no examples on when and how to use all_row_wildcards in the data_filter configuration block, it's not clear on when this needs to be provided or what it should look like.
2. Attribute etries are referencing additional examples below which don't appear to exist.

Overall, this documentation just requires a little more clarity on how this data filter resource works.

### References

Related Issue: https://github.com/hashicorp/terraform-provider-aws/issues/36574

No

sjlewis25:
Fixes #41376

## Description
Improves the `aws_lakeformation_data_cells_filter` documentation by adding clarity and examples for common usage scenarios.

## Changes
- Added three additional usage examples covering:
  - Excluded columns only (requires `all_rows_wildcard {}`)
  - Row filter with excluded columns
  - Row filter only (all columns included with `excluded_column_names = []`)
- Clarified that exactly one of `filter_expression` or `all_rows_wildcard` must be specified
- Explained when `all_rows_wildcard` is required and its syntax
- Improved description of `filter_expression` parameter

## Context
Users were confused about when to use `all_rows_wildcard` and how to properly configure different filtering scenarios, leading to validation errors.

### DIFF
--- a/website/docs/r/lakeformation_data_cells_filter.html.markdown
+++ b/website/docs/r/lakeformation_data_cells_filter.html.markdown
@@ -16,10 +16,10 @@ Terraform resource for managing an AWS Lake Formation Data Cells Filter.
 ```terraform
 resource "aws_lakeformation_data_cells_filter" "example" {
   table_data {
-    database_name    = aws_glue_catalog_database.test.name
+    database_name    = aws_glue_catalog_database.example.name
     name             = "example"
     table_catalog_id = data.aws_caller_identity.current.account_id
-    table_name       = aws_glue_catalog_table.test.name
+    table_name       = aws_glue_catalog_table.example.name
 
     column_names = ["my_column"]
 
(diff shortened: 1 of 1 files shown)


jar-b (approved):

johnsonaj:
@sjlewis24 thank you for the contribution! 🎉

PR #45357, opened 2025-12-02
Feature: DynamoDB Multi-Attribute Keys for Global Secondary Index

shawnaws:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description
This change adds support for Multi-Attribute keys in Dynamo DB which was recently added. [Documentation page](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html)

### Relations
Closes #45216

### References
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html

### Output from Acceptance Testing
```console
% make build && make testacc PKG=dynamodb TESTS=TestAccDynamoDBTable_extended_gsiMultiHashKey
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
make: Building provider...
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
make: Running acceptance tests on branch: 🌿 f/ddb-gsi-mutlihashrangekey 🌿...
TF_ACC=1 go1.24.10 test ./internal/service/dynamodb/... -v -count 1 -parallel 20 -run='TestAccDynamoDBTable_extended_gsiMultiHashKey'  -timeout 360m -vet=off
2025/12/01 13:45:31 Creating Terraform AWS Provider (SDKv2-style)...
2025/12/01 13:45:31 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccDynamoDBTable_extended_gsiMultiHashKey
=== PAUSE TestAccDynamoDBTable_extended_gsiMultiHashKey
=== RUN   TestAccDynamoDBTable_extended_gsiMultiHashKey_transition
    table_test.go:900: Skipping test for validation
--- SKIP: TestAccDynamoDBTable_extended_gsiMultiHashKey_transition (0.00s)
=== RUN   TestAccDynamoDBTable_extended_gsiMultiHashKeyMutliRangeKey_maxSet
=== PAUSE TestAccDynamoDBTable_extended_gsiMultiHashKeyMutliRangeKey_maxSet
=== RUN   TestAccDynamoDBTable_extended_gsiMultiHashKey_singleAndMultiSet
=== PAUSE TestAccDynamoDBTable_extended_gsiMultiHashKey_singleAndMultiSet
[...]

PR #46359, opened 2026-02-06
Fix documentation for GSI multi attr

marceloavan:
### Description

The implementation made on #45357 seems to be changed before merge, and the documentation was not updated, so the references are to the first approach that wasn't merged.

### Relations

Relates #44999, #45357


jar-b (approved):
LGTM 🚀

jar-b:
Thank you for the update, @marceloavan! 👍

PR #46522, opened 2026-02-17
docs: correct mutability options for ecr_repository_creation_template

mrivnak1:
### Description

Updates `image_tag_mutability` argument description to include all supported values, aligning it with the argument of the same name for the `aws_ecr_repository` resource.

The `_WITH_EXCLUSION` values were missing, although they were referenced in the description for the `image_tag_mutability_exclusion_filter` argument, which was a bit confusing because of the inconsistency.


ewbankkit:
@mrivnak1 Thanks for the contribution :tada: :clap:.

PR #46589, opened 2026-02-20
Vmc gi version tag

ab-cs13:
## Rollback Plan

NA

### Description
gi_version is a mandatory field when creating a cloud_vm_cluster. The underlying API accepts only the major gi_version values 19.0.0.0 or 23.0.0.0, or a value sourced from the list-gi-versions data source.

After creation, the cloud-vm-cluster read operation returns the full GI version (for example, 23.1.4.0.5). In Terraform, this is stored in the computed_gi_version attribute. We then derive the major GI version by parsing the major portion (e.g., 23.1.4.0.5 → 23.0.0.0).

Recently, the underlying API introduced a change that is not compatible with the current parsing logic. Specifically, for the major version 26.0.0.0, the read operation returns 23.1.2.5, which causes Terraform to fail.

This PR introduces a tag-based approach to preserve the user-provided major gi_version input.

Screenshot 2026-02-20 at 13 43 01


internal/service/odb/cloud_vm_cluster.go
```diff
@@ -748,15 +761,29 @@ func findCloudVmClusterForResourceByID(ctx context.Context, conn *odb.Client, id
...
+	if err != nil {
 		return nil, err
 	}
-	return &giVersionMajor, nil
+	var inputGiVersion *string
+	if tagsRead.KeyExists(GiVersionSystemTag) {
+		inputGiVersion = tagsRead.KeyValue(GiVersionSystemTag)
+		return inputGiVersion, nil
+	} else {
+		//This regx based approach is for backward compatibility
```
jar-b:
```suggestion
		// This regexp based approach is for backward compatibility
```

internal/service/odb/cloud_vm_cluster.go
```diff
@@ -748,15 +761,29 @@ func findCloudVmClusterForResourceByID(ctx context.Context, conn *odb.Client, id
...
+	tagsRead, err := listTags(ctx, conn, *arn)
+	if err != nil {
 		return nil, err
 	}
-	return &giVersionMajor, nil
+	var inputGiVersion *string
+	if tagsRead.KeyExists(GiVersionSystemTag) {
+		inputGiVersion = tagsRead.KeyValue(GiVersionSystemTag)
+		return inputGiVersion, nil
+	} else {
```
jar-b, line 776: To ensure fallback works as expected, it might be worth adding an additional acceptance tests which creates the cluster with a previous version that does not have the GI version tag, then upgrade to the latest version. 

You can use an `ExternalProviders` block to pin to a previous version of the provider for the first test step. For example,

https://github.com/hashicorp/terraform-provider-aws/blob/c7c7d55801cbac408e162194d233d88f1b0dda69/internal/service/resourcegroups/resource_test.go#L65-L86

.changelog/46589.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+resource/aws_odb_cloud_vm_cluster: gi_version bug.
```
jar-b:
```suggestion
resource/aws_odb_cloud_vm_cluster: Attempt to read GI Version from resource tags to avoid failures due to new API response values
```

internal/service/odb/cloud_vm_cluster.go
ab-cs13: - Added TestAccODBCloudVmCluster_previousProviderVersion

Screenshot 2026-05-01 at 15 07 13

website/docs/r/odb_cloud_vm_cluster.html.markdown
```diff
@@ -67,6 +67,35 @@ resource "aws_odb_cloud_vm_cluster" "with_all_parameters" {
     is_incident_logs_enabled      = true
   }
 }
+
+
+resource "aws_odb_cloud_vm_cluster" "gi_version_tag_example" {
```
jar-b, line 78: Typically these three different examples would be split into separate subheadings (e.g. `Basic Usage`, `With All Parameters`, `With GI Version Tag`).

internal/service/odb/cloud_vm_cluster_test.go
```diff
@@ -772,3 +815,155 @@ resource "aws_odb_cloud_vm_cluster" "test" {
...
+	vmcDisplayName := sdkacctest.RandomWithPrefix(vmClusterTestEntity.vmClusterDisplayNamePrefix)
+	resourceName := "aws_odb_cloud_vm_cluster.test"
+
+	publicKey, _, err := sdkacctest.RandSSHKeyPair(acctest.DefaultEmailAddress)
+	if err != nil {
+		t.Fatal(err)
+		return
+	}
+
+	vmcWithGiVersionTag := vmClusterTestEntity.cloudVmClusterConfigWithOlderTfProviderAndUpgradeToLatest(vmcDisplayName, publicKey)
```
jar-b: This variable name is a bit misleading in that the underlying configuration does not have the GI version tag. This is what I would expect for versions < 6.43.0 given the tag guidance did not yet exist, so perhaps simply changing the variable name to something more generic would make sense here.

internal/service/odb/cloud_vm_cluster_test.go
```diff
@@ -772,3 +815,155 @@ resource "aws_odb_cloud_vm_cluster" "test" {
...
+				),
+			},
+			{
+				ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories,
+				Config:                   vmcWithGiVersionTag,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					vmClusterTestEntity.testAccCheckCloudVmClusterExists(ctx, resourceName, &cloudvmcluster),
+					resource.TestCheckResourceAttrSet(resourceName, names.AttrID),
+				),
+			},
```
jar-b, line 942: It might also make sense to include a third step where the configuration adds the GI version tag.

website/docs/r/odb_cloud_vm_cluster.html.markdown
```diff
@@ -76,7 +105,7 @@ The following arguments are required:
 * `cpu_core_count` - (Required) The number of CPU cores to enable on the VM cluster. Changing this will create a new resource.
 * `db_servers` - (Required) The list of database servers for the VM cluster. Changing this will create a new resource.
 * `display_name` - (Required) A user-friendly name for the VM cluster. Changing this will create a new resource.
-* `gi_version` - (Required) A valid software version of Oracle Grid Infrastructure (GI). To get the list of valid values, use the ListGiVersions operation and specify the shape of the Exadata infrastructure. Example: 19.0.0.0 Changing this will create a new resource.
+* `gi_version` - (Required) A valid Oracle Grid Infrastructure (GI) software version. To get valid values, use the ListGiVersions operation for the Exadata infrastructure shape. Example: `19.0.0.0`. Changing this creates a new resource. Prefer to provide `odb:input_gi_version` tag. If `odb:input_gi_version` tag is provided, its value must exactly match `gi_version`, otherwise Terraform returns an error. Refer gi_version_tag_example given above.
```
jar-b: If the examples are split into distinct subheadings the last sentence can link directly to the GI version tag example instead of referring to the resource instance name.

ab-cs13: Done

internal/service/odb/cloud_vm_cluster_test.go
ab-cs13: Done 

Screenshot 2026-05-04 at 11 32 02

website/docs/r/odb_cloud_vm_cluster.html.markdown
ab-cs13: Done

internal/service/odb/cloud_vm_cluster_test.go
ab-cs13: Done

website/docs/r/odb_cloud_vm_cluster.html.markdown
```diff
@@ -16,6 +16,8 @@ You can find out more about Oracle Database@AWS from [User Guide](https://docs.a
 
 ### Basic Usage
 
+#### Resource: `aws_odb_cloud_vm_cluster.with_minimum_parameter`
+
```
jar-b: The `Basic Usage` subheading already applies to this example.

```suggestion
```

website/docs/r/odb_cloud_vm_cluster.html.markdown
```diff
@@ -38,8 +40,11 @@ resource "aws_odb_cloud_vm_cluster" "with_minimum_parameter" {
     is_incident_logs_enabled      = false
   }
 }
+```
 
+#### Resource: `aws_odb_cloud_vm_cluster.with_all_parameters`
```
jar-b:
```suggestion
### With Optional Arguments
```

website/docs/r/odb_cloud_vm_cluster.html.markdown
```diff
@@ -69,14 +74,46 @@ resource "aws_odb_cloud_vm_cluster" "with_all_parameters" {
 }
 ```
 
+#### Resource: `aws_odb_cloud_vm_cluster.gi_version_tag_example`
```
jar-b:
```suggestion
### With GI Version Tag
```

website/docs/r/odb_cloud_vm_cluster.html.markdown
```diff
@@ -69,14 +74,46 @@ resource "aws_odb_cloud_vm_cluster" "with_all_parameters" {
...
 
 * `cpu_core_count` - (Required) The number of CPU cores to enable on the VM cluster. Changing this will create a new resource.
 * `db_servers` - (Required) The list of database servers for the VM cluster. Changing this will create a new resource.
 * `display_name` - (Required) A user-friendly name for the VM cluster. Changing this will create a new resource.
-* `gi_version` - (Required) A valid software version of Oracle Grid Infrastructure (GI). To get the list of valid values, use the ListGiVersions operation and specify the shape of the Exadata infrastructure. Example: 19.0.0.0 Changing this will create a new resource.
+* `gi_version` - (Required) A valid Oracle Grid Infrastructure (GI) software version. To get valid values, use the ListGiVersions operation for the Exadata infrastructure shape. Example: `19.0.0.0`. Changing this creates a new resource. Prefer to provide `odb:input_gi_version` tag. If `odb:input_gi_version` tag is provided, its value must exactly match `gi_version`, otherwise Terraform returns an error. Refer [Resource: `aws_odb_cloud_vm_cluster.gi_version_tag_example`](#resource-aws_odb_cloud_vm_clustergi_version_tag_example).
```
jar-b:
```suggestion
* `gi_version` - (Required) A valid Oracle Grid Infrastructure (GI) software version. To get valid values, use the ListGiVersions operation for the Exadata infrastructure shape. Example: `19.0.0.0`. Changing this creates a new resource. Prefer to provide `odb:input_gi_version` tag. If `odb:input_gi_version` tag is provided, its value must exactly match `gi_version`, otherwise Terraform returns an error. See the [`With GI Version Tag`](#with-gi-version-tag) example above.
```

jar-b (approved):
LGTM 🚀

jar-b:
Thanks for your contribution, @ab-cs13! 👍

Issue #43717, opened 2025-08-05
Go-VCR: Add support to all services


jar-b:
This issue will track progress towards enabling `go-vcr` in all services within the Terraform AWS provider.

- [x] `accessanalyzer`
- [x] `account`
- [x] `acm`
- [x] `acmpca`
- [x] `amp`
- [x] `amplify`
- [x] `apigateway`
- [x] `apigatewayv2`
- [x] `appautoscaling`
- [x] `appconfig`
- [x] `appfabric`
- [x] `appflow`
- [x] `appintegrations`
- [x] `applicationinsights`
- [x] `applicationsignals`
- [x] `appmesh`
- [x] `apprunner`
- [x] `appstream`
- [x] `appsync`
- [x] `athena`
- [x] `auditmanager`
- [x] `autoscaling`
- [x] `autoscalingplans`
- [x] `backup`
- [x] `batch`
- [x] `bcmdataexports`
- [x] `bedrock`
- [x] `bedrockagent`
- [x] `bedrockagentcore`
- [x] `billing`
- [x] `budgets`
- [x] `ce`
- [x] `chatbot`
- [x] `chime`
- [x] `chimesdkmediapipelines`
- [x] `chimesdkvoice`
- [x] `cleanrooms`
- [x] `cloud9`
- [x] `cloudcontrol`
- [x] `cloudformation`
- [x] `cloudfront`
- [x] `cloudfrontkeyvaluestore`
- [x] `cloudhsmv2`
- [x] `cloudsearch`
- [x] `cloudtrail`
- [x] `cloudwatch`
- [x] `codeartifact`
- [x] `codebuild`
- [x] `codecatalyst`
- [x] `codecommit`
- [x] `codeconnections`
- [x] `codeguruprofiler`
- [x] `codegurureviewer`
- [x] `codepipeline`
- [x] `codestarconnections`
- [x] `codestarnotifications`
- [x] `cognitoidentity`
- [x] `cognitoidp`
- [x] `comprehend`
- [x] `computeoptimizer`
- [x] `configservice`
- [x] `connect`
- [x] `connectcases`
- [x] `controltower`
- [x] `costoptimizationhub`
- [x] `cur`
- [x] `customerprofiles`
- [x] `databrew`
- [x] `dataexchange`
- [x] `datapipeline`
- [x] `datasync`
- [x] `datazone`
- [x] `dax`
- [x] `deploy`
- [x] `detective`
- [x] `devicefarm`
[...]

PR #46638, opened 2026-02-24
`vcr`: Add Semgrep rules to prevent new instances of non-`vcr`-friendly functions

gdavison:
## Rollback Plan

### Description

Some long-standing PRs or copied code have introduced non-`vcr`-friendly functions to service packages that were upgraded for `vcr` support.

Adds Semgrep rules to detect these cases. Currently excludes the following packages:

- "/internal/provider/sdkv2"
- "/internal/service/apigatewayv2"
- "/internal/service/arcregionswitch"
- "/internal/service/billing"
- "/internal/service/cleanrooms"
- "/internal/service/docdb"
- "/internal/service/ec2"
- "/internal/service/networkflowmonitor"
- "/internal/service/observabilityadmin"
- "/internal/service/odb"
- "/internal/service/savingsplans"

### Relations

Relates #43717

### DIFF
--- a/internal/framework/flex/autoflex_primitives_test.go
+++ b/internal/framework/flex/autoflex_primitives_test.go
@@ -113,7 +113,7 @@ func testStringRoundtrip(t *testing.T) {
 		// Random value for property-based testing feel
 		{
 			name:        "random_value",
-			stringValue: sdkacctest.RandomWithPrefix("tf-test"),
+			stringValue: sdkacctest.RandomWithPrefix("tf-test"), // nosemgrep:ci.semgrep.acctest.vcr.use-acctest-randomwithprefix
 			variants:    []string{"standard", "legacy", "tf_to_aws_pointer", "legacy_tf_to_aws_pointer"},
 		},
 		// Omitempty tests - flatten-only (expand direction not defined in original tests)


.ci/semgrep/acctest/vcr/random.yml
```diff
@@ -0,0 +1,75 @@
...
+        - "/internal/service/apigatewayv2"
+        - "/internal/service/arcregionswitch"
+        - "/internal/service/billing"
+        - "/internal/service/cleanrooms"
+        - "/internal/service/docdb"
+        - "/internal/service/ec2"
+        - "/internal/service/networkflowmonitor"
+        - "/internal/service/observabilityadmin"
+        - "/internal/service/odb"
+        - "/internal/service/savingsplans"
```
jar-b, line 27: `ec2` and `odb` _should_ be the last remaining services missing VCR support (tracked in https://github.com/hashicorp/terraform-provider-aws/issues/43717). If there are other valid instances for using `sdkacctest` in these packages it might be worth adding `nosemgrep` comments inline for those edge cases rather than ignoring the entire package.

gdavison: The checks are for specific `sdkacctest` functions that would have been updated by `make vcr-enable`, so I don't think there should be any "legit" cases for those.

The additional service packages are either new or have new resource types since the initial `make vcr-enable` run

gdavison: Found one semi-legit use in `internal/framework/flex/autoflex_primitives_test.go`, but it just needs a random string of any kind

PR #46745, opened 2026-03-04
ELBv2 List Resources: Batch reading tags

gdavison:
## Rollback Plan

### Description

The ELBv2 API for reading tags, `DescribeTags`, can read tags for multiple resources at once. While listing ELBv2 resources, batch tag reading instead of one API call per resource.


ewbankkit:
Future consideration: Can `batchListTags` be generated?

gdavison:
> Future consideration: Can `batchListTags` be generated?

Yes. But it looks like there are only four services with batchable tags, based on `-ListTagsInIDNeedValueSlice` in the tags generator

PR #46763, opened 2026-03-04
List Resource testing: Move `ProtoV5ProviderFactories` out of `TestStep` level and into `TestCase` level

gdavison:
## Rollback Plan

### Description

Moves `ProtoV5ProviderFactories` out of `TestStep` level and into `TestCase` level

### DIFF
--- a/internal/service/secretsmanager/secret_version_list_test.go
+++ b/internal/service/secretsmanager/secret_version_list_test.go
@@ -38,7 +38,6 @@ func TestAccSecretsManagerSecretVersion_List_basic(t *testing.T) {
 		ErrorCheck:               acctest.ErrorCheck(t, names.SecretsManagerServiceID),
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories,
 		CheckDestroy:             testAccCheckSecretVersionDestroy(ctx, t),
-		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories,
 		Steps: []resource.TestStep{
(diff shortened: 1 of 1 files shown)


internal/service/ec2/ec2_secondary_subnet_list_test.go
```diff
@@ -28,6 +28,7 @@ func testAccEC2SecondarySubnet_listSerial(t *testing.T) {
 		acctest.CtBasic:   testAccEC2SecondarySubnet_List_basic,
 		"includeResource": testAccEC2SecondarySubnet_List_includeResource,
 		"regionOverride":  testAccEC2SecondarySubnet_List_regionOverride,
+		"filtered":        testAccEC2SecondarySubnet_List_filtered,
```
jar-b, line 31: Nice catch 👍

internal/service/iam/role_policy_attachment_list.go
```diff
@@ -82,7 +82,7 @@ func (l *rolePolicyAttachmentListResource) List(ctx context.Context, request lis
 					tflog.Warn(ctx, "Resource disappeared during listing, skipping", map[string]any{
 						logging.ResourceAttributeKey(names.AttrRole): aws.ToString(role.RoleName),
 					})
-					continue
+					break
```
jar-b, line 85: That's a sneaky one - good find!

gdavison: Yeah, I was getting infinite loops intermittently

PR #46799, opened 2026-03-06
Fix aws_lightsail_static_ip_attachment example to use .name instead of .id

fredrikekre:


jar-b:
Thank you for the correction, @fredrikekre! 👍

PR #46807, opened 2026-03-07
Increase default timeout of inspector2 enable from 5 to 30 minutes

JJamali:
## Rollback Plan

None

### Description
In many production-level AWS accounts (accounts with many scannable resources), it takes inspector2 much longer than 5 minutes to activate the enabled scans. This simple change prevents unnecessary taint of the enabler resource.

### Relations
n/a
Closes #0000

### References
n/a

n/a

```console

...
```


edobrowo:
👍

Issue #46842, opened 2026-03-10
aws_guardduty_filter for severity field is created with date value


ajax-mykhailo-oleksiuk:
### Terraform and AWS Provider Version

```console
Terraform v1.13.4
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v6.35.1
```

### Affected Resource(s) or Data Source(s)

* `aws_guardduty_filter`

### Expected Behavior

When I create a filter for `severity` field with one of operation `greater_than`, `greater_than_or_equal`, `less_than`, `less_than_or_equal` - it has to be created with one of severity values (LOW, MEDIUM, HIGH, CRITICAL).

For example:
```hcl
resource "aws_guardduty_filter" "myfilter" {
  detector_id = "<id>"
  name        = "MyFilter"
  action      = "ARCHIVE"
  rank        = 1

  finding_criteria {
    criterion {
      field     = "severity"
      less_than = 4
    }
  }
}
```

I expect a filter with such attributes:
<img width="1125" height="316" alt="Image" src="https://github.com/user-attachments/assets/61de63c8-4f90-48c0-b73c-f13c32ce79b7" />

### Actual Behavior

When I create a filter with one of operation `greater_than`, `greater_than_or_equal`, `less_than`, `less_than_or_equal` and specify "integer" according to the [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/guardduty_filter#greater_than-1) I'm getting a filter with date value:

<img width="1024" height="313" alt="Image" src="https://github.com/user-attachments/assets/de8ed77c-1945-44ad-a89c-ede085c92ea0" />

### Relevant Error/Panic Output

```console

```

### Sample Terraform Configuration

<details open>
<summary>Click to expand configuration</summary>

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 6.35.0"
    }
  }
}

provider "aws" {
  region = "eu-west-1"
}

resource "aws_guardduty_detector" "this" {
  enable = true
}

resource "aws_guardduty_filter" "myfilter" {
  detector_id = aws_guardduty_detector.this.id
  name        = "MyFilter"
  action      = "ARCHIVE"
  rank        = 1

  finding_criteria {
    criterion {
      field     = "severity"
      less_than = 4
    }
  }
}
```
</details>

### Steps to Reproduce

1. Get simple code above and run `terraform apply`
[...]

PR #46914, opened 2026-03-13
[Docs] aws_guardduty_filter: Clarify that an integer value given as a string can be accepted as a value of `less_than` etc.

tabito-hara:
## Rollback Plan

### Description

In the current implementation, the `greater_than`, `greater_than_or_equal`, `less_than`, and `less_than_or_equal` arguments are defined as strings and validated using `ValidStringDateOrPositiveInt`.

[https://github.com/hashicorp/terraform-provider-aws/blob/a9e74da42c8fe92d04de746a5ae160ecb84b90a9/internal/service/guardduty/filter.go#L92-L112](https://github.com/hashicorp/terraform-provider-aws/blob/a9e74da42c8fe92d04de746a5ae160ecb84b90a9/internal/service/guardduty/filter.go#L92-L112)

However, the documentation states:

> [less_than](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/guardduty_filter#less_than-1)- (Optional) A value to be evaluated. Accepts either an integer or a date in [RFC 3339 format](https://tools.ietf.org/html/rfc3339#section-5.8)

This description may suggest that an integer value can be provided directly. However, the value must be provided as a string (i.e., enclosed in quotation marks).

The example in the documentation also shows that an integer value for these arguments must be enclosed in quotation marks:

https://github.com/hashicorp/terraform-provider-aws/blob/a9e74da42c8fe92d04de746a5ae160ecb84b90a9/website/docs/r/guardduty_filter.html.markdown?plain=1#L39-L42

This PR updates the documentation to clarify that integer values must be provided as strings (i.e., enclosed in quotation marks) when used with these arguments.

### Relations

Relates #46842

### DIFF
--- a/website/docs/r/guardduty_filter.html.markdown
+++ b/website/docs/r/guardduty_filter.html.markdown
@@ -64,10 +64,10 @@ The `criterion` block supports the following:
 * `field` - (Required) The name of the field to be evaluated. The full list of field names can be found in [AWS documentation](https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_filter-findings.html#filter_criteria).
 * `equals` - (Optional) List of string values to be evaluated.
 * `not_equals` - (Optional) List of string values to be evaluated.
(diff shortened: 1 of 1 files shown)


ewbankkit:

PR #46929, opened 2026-03-13
docs: remove text from  argument "arn" in data source "aws_batch_job_definition"

stefanfreitag:
## Rollback Plan

### Description

The description for the argument [`arn`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/batch_job_definition#arn-1) of the data source [`aws_batch_job_definition`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/batch_job_definition) contains 

> Do not begin the description with "An", "The", "Defines", "Indicates", or "Specifies," as these are verbose. In other words, "Indicates the amount of storage," can be rewritten as "Amount of storage," without losing any information.

This is the "default"/ lorem ipsum text and this PR removes it.

### DIFF
--- a/website/docs/d/batch_job_definition.html.markdown
+++ b/website/docs/d/batch_job_definition.html.markdown
@@ -12,7 +12,7 @@ Terraform data source for managing an AWS Batch Job Definition.
 
 ## Example Usage
 
-### Lookup via Arn
+### Lookup via ARN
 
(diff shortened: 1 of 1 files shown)


jar-b:
Thank you for the correction, @stefanfreitag! 👍

PR #46937, opened 2026-03-14
fix lambda_permission list resource documentation example

tmanti:
### Description
Updates the example usage of the `aws_lambda_permission` list resource to include the config block and pass the `function_name` within there instead. This is because provider specific attributes belong within the config block.

### References
https://developer.hashicorp.com/terraform/language/block/tfquery/list#config


ewbankkit:
@tmanti Thanks for the contribution :tada: :clap:.

PR #46953, opened 2026-03-17
r/aws_vpc, r/aws_subnet: Add GuardDuty managed resource cleanup on destroy

### LINKED ISSUE #46747: Automatic GuardDuty Runtime Monitoring resource cleanup during destroy
### Description

When GuardDuty Runtime Monitoring is enabled, AWS automatically creates VPC endpoints and security groups in customer VPCs. These resources are not tracked in Terraform state, causing `terraform destroy` to fail with DependencyViolation errors. Users must manually delete GuardDuty resources or disable GuardDuty (which may be enabled at organization level and unable to disable in IaC), and then run `terraform destroy` again.

This enhancement would adda automatic detection and deletion of GuardDuty managed resources when DependencyViolation occurs during subnet and VPC deletion, allowing `terraform destroy` to succeed in a single run.

- `aws_subnet (internal/service/ec2/vpc_subnet.go)`
- `aws_vpc (internal/service/ec2/vpc.go)`

No configuration changes required. This improves the behavior of existing terraform destroy operations.

**Before:** terraform destroy fails with DependencyViolation, requires manual cleanup

**After:** terraform destroy automatically cleans up GuardDuty resources and succeeds

## Proposed Solution

**For vpc_subnet.go:**
- Catch DependencyViolation during subnet deletion
- Detect GuardDuty VPC endpoints (service name pattern: *guardduty-data*)
- Delete detected VPC endpoints
- Surface clear errors if IAM permissions are missing
- Retry subnet deletion

**For vpc.go:**
- Catch DependencyViolation during VPC deletion
- Detect GuardDuty security groups (name pattern: GuardDutyManagedSecurityGroup-*)
- Delete detected security groups
- Surface clear errors if IAM permissions are missing
- Retry VPC deletion

```hcl
resource "aws_vpc" "example" {
  cidr_block = "10.0.0.0/16"
}

resource "aws_subnet" "example" {
  vpc_id     = aws_vpc.example.id
  cidr_block = "10.0.1.0/24"
}

# With this enhancement, terraform destroy will:
# 1. Attempt subnet deletion
# 2. Detect DependencyViolation from GuardDuty VPC endpoints
# 3. Automatically delete GuardDuty VPC endpoints
# 4. Retry and succeed without manual intervention
```

### References

- AWS GuardDuty VPC endpoint docs: https://docs.aws.amazon.com/guardduty/latest/ug/security-vpc-endpoints.html
- AWS SDK Go v2 DescribeVpcEndpoints: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpoints.html
- AWS SDK Go v2 DeleteVpcEndpoints: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteVpcEndpoints.html
- AWS SDK Go v2 DescribeSecurityGroups: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSecurityGroups.html
- AWS SDK Go v2 DeleteSecurityGroup: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteSecurityGroup.html

Yes

JMRacke:
### Description

When AWS GuardDuty is enabled, it automatically creates VPC endpoints (service name `com.amazonaws.{region}.guardduty-data`) and security groups (`GuardDutyManagedSecurityGroup-{vpc-id}`) in customer VPCs. These resources are not tracked in Terraform state, causing `terraform destroy` to fail with `DependencyViolation` errors on `aws_subnet` and `aws_vpc` resources.

This PR adds automatic detection and cleanup of GuardDuty-managed resources during destroy operations:

- **Subnet deletion** (`aws_subnet`): When `DependencyViolation` occurs, the provider detects GuardDuty VPC endpoints (tagged `GuardDutyManaged=true`) associated with the subnet and dissociates them via `ModifyVpcEndpoint`, then retries the deletion.
- **VPC deletion** (`aws_vpc`): When `DependencyViolation` occurs, the provider detects and deletes GuardDuty VPC endpoints and security groups, then retries the deletion.
- Cleanup is a best-effort operation — if the IAM role lacks the optional permissions, the provider logs warnings and continues with the deletion attempt.
- Website documentation for both `aws_vpc` and `aws_subnet` updated with the required IAM permissions in copy-pasteable JSON policy format.

### Relations

Closes #46747

### References

- [AWS GuardDuty VPC endpoints documentation](https://docs.aws.amazon.com/guardduty/latest/ug/security-vpc-endpoints.html)

```console
% TF_ACC=1 go test ./internal/service/ec2/ -run "TestAccVPC.*guardDuty|TestAccVPCSubnet.*guardDuty" -v -timeout 45m

=== RUN   TestAccVPCSubnet_guardDutyLastSubnetDeletion
--- PASS: TestAccVPCSubnet_guardDutyLastSubnetDeletion (262.07s)
=== RUN   TestAccVPCSubnet_guardDutyCleanup
--- PASS: TestAccVPCSubnet_guardDutyCleanup (176.93s)
=== RUN   TestAccVPCSubnet_guardDutyCleanup_timeout
--- PASS: TestAccVPCSubnet_guardDutyCleanup_timeout (0.00s)
=== RUN   TestAccVPCSubnet_guardDutyNotAssociated
--- PASS: TestAccVPCSubnet_guardDutyNotAssociated (0.00s)
=== RUN   TestAccVPC_guardDutySecurityGroupCleanup
--- PASS: TestAccVPC_guardDutySecurityGroupCleanup (290.00s)
=== RUN   TestAccVPC_guardDutyEndpointAlreadyCleaned
--- PASS: TestAccVPC_guardDutyEndpointAlreadyCleaned (291.00s)
=== RUN   TestAccVPC_guardDutyNoResources
--- PASS: TestAccVPC_guardDutyNoResources (31.00s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ec2	439.000s
```


JMRacke:
Hey team, Seeing the above workflows run, I ran them locally, addressed the issues, and amended the PR. I also re-based to ensure I had the latest main changes in the branch. I am not sure how to re-run the workflows that were triggered above the other day.

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
 }
 `, acctest.ProviderName))
 }
+
+func TestGuardDutySecurityGroupNamePattern(t *testing.T) {
```
gdavison: What is this test testing? It looks like it's checking the result of string addition.

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+// code is a safe no-op when there are no GuardDuty resources. The DescribeVpcEndpoints
+// and DescribeSecurityGroups calls should return empty results and the code should return
+// immediately without interfering with normal VPC deletion.
+//
+// Note: This test only exercises the GuardDuty cleanup path if DeleteVpc returns
+// DependencyViolation. Without GuardDuty resources, there may be no DependencyViolation
+// at all, in which case the cleanup code is never called.
+//
+// EXPECTED: Test PASSES. Normal VPC deletion unaffected by GuardDuty cleanup code.
+func TestAccVPC_guardDutyNoResources(t *testing.T) {
```
gdavison: The case of deleting a VPC with no GuardDuty resources should be covered by all of the existing tests. This test is redundant

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+
+			if matches != tc.shouldMatch {
+				t.Errorf("Security group name pattern matching failed for '%s' with VPC '%s'\nExpected match: %v, Got: %v",
+					tc.groupName, tc.vpcID, tc.shouldMatch, matches)
+			}
+		})
+	}
+}
+
+func TestIsDependencyViolationError(t *testing.T) {
```
gdavison: Unit tests should be placed at the top of the file

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+			t.Parallel()
+
+			result := tfec2.IsDependencyViolationError(tc.err)
+			if result != tc.shouldBeDependencyViolation {
+				t.Errorf("Expected %v, got %v for error: %v", tc.shouldBeDependencyViolation, result, tc.err)
+			}
+		})
+	}
+}
+
```
gdavison: Acceptance tests should be placed before the configuration functions and any helpers such as `testAccCheckVPCDestroy`

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+// TestAccVPC_guardDutySecurityGroupCleanup validates UC-V1: VPC destroy with out-of-band
```
gdavison: What does `UC-V1` refer to?

internal/service/ec2/vpc_.go
```diff
@@ -756,3 +787,205 @@ func resourceVPCFlatten(ctx context.Context, client *conns.AWSClient, vpc *awsty
...
+	input := &ec2.DescribeSecurityGroupsInput{
+		Filters: []awstypes.Filter{
+			{
+				Name:   aws.String("vpc-id"),
+				Values: []string{vpcID},
+			},
+			{
+				Name:   aws.String("group-name"),
+				Values: []string{guardDutyGroupName},
+			},
```
gdavison: Should this check the tag value as well?

internal/service/ec2/vpc_.go
```diff
@@ -390,11 +390,37 @@ func resourceVPCDelete(ctx context.Context, d *schema.ResourceData, meta any) di
...
+			}
+
+			_, sgErr, sgWarning := detectAndDeleteGuardDutySecurityGroups(ctx, conn, d.Id())
+			if sgWarning != "" {
+				guardDutyWarnings = append(guardDutyWarnings, sgWarning)
+			}
+			if sgErr != nil {
+				log.Printf("[WARN] Error cleaning up GuardDuty security groups in VPC %s: %s", d.Id(), sgErr)
+			}
+		}
```
gdavison: Why not check for VPC Endpoints and Security Groups before seeing a failure?

internal/service/ec2/vpc_.go
```diff
@@ -756,3 +787,205 @@ func resourceVPCFlatten(ctx context.Context, client *conns.AWSClient, vpc *awsty
...
+				Values: []string{guardDutyGroupName},
+			},
+		},
+	}
+
+	output, err := conn.DescribeSecurityGroups(ctx, input)
+	if err != nil {
+		if IsUnauthorizedError(err) {
+			log.Printf("[WARN] Insufficient IAM permissions to describe GuardDuty security groups in VPC %s: %s", vpcID, err)
+			warningMsg := fmt.Sprintf(
```
gdavison: Why treat this error as a warning?

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+			}
+
+			if err != nil {
+				return err
+			}
+
+			return fmt.Errorf("EC2 VPC %s still exists", rs.Primary.ID)
+		}
+
+		if vpcID != nil && *vpcID != "" {
```
gdavison: To simplify this, you could do something like

```suggestion
		if id := aws.ToString(vpcID); id != "" {
```

and then use `id` where `*vpcID` is used

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+						Name:   aws.String("vpc-id"),
+						Values: []string{*vpcID},
+					},
+					{
+						Name:   aws.String("service-name"),
+						Values: []string{"*guardduty-data*"},
+					},
+					{
+						Name:   aws.String("tag:GuardDutyManaged"),
+						Values: []string{acctest.CtTrue},
```
gdavison: Consider using the constants added to `consts.go`. You will have to export them via `exports_test.go`

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+					},
+				},
+			}
+			endpointsOutput, err := conn.DescribeVpcEndpoints(ctx, endpointsInput)
+			if err != nil {
+				return nil
```
gdavison: Errors should not fail silently

internal/service/ec2/vpc_subnet_test.go
```diff
@@ -1703,3 +1704,1541 @@ resource "aws_subnet" "test" {
...
+					Name:   aws.String("tag:GuardDutyManaged"),
+					Values: []string{acctest.CtTrue},
+				},
+			},
+		}
+		endpointsOutput, err := conn.DescribeVpcEndpoints(ctx, &describeEndpointsInput)
+		if err != nil {
+			// VPC may already be deleted, treat as success
+			return nil
+		}
```
gdavison: This should not fail silently

internal/service/ec2/vpc_subnet_test.go
```diff
@@ -1703,3 +1704,1541 @@ resource "aws_subnet" "test" {
...
+
+		return nil
+	}
+}
+
+// testAccDeleteGuardDutyResources describes and deletes any GuardDuty-tagged endpoints and security
+// groups in the VPC. Used in CheckDestroy to clean up out-of-band resources that Terraform doesn't
+// know about. Handles NotFound errors gracefully since resources may already be cleaned up.
+func testAccDeleteGuardDutyResources(ctx context.Context, t *testing.T, vpcIDPtr *string) resource.TestCheckFunc {
```
gdavison: Typically, we don't actually delete remote resources in the `CheckDestroy` functions, they are simply to validate that remote resources have been deleted. See https://hashicorp.github.io/terraform-provider-aws/running-and-writing-acceptance-tests/#anatomy-of-an-acceptance-test for more information.

Stray remote resources can be cleaned up using sweepers. See https://hashicorp.github.io/terraform-provider-aws/running-and-writing-acceptance-tests/#acceptance-test-sweepers

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+				Check: resource.ComposeTestCheckFunc(
+					acctest.CheckVPCExists(ctx, t, vpcResourceName, &vpc),
+					testAccCheckSubnetExists(ctx, t, subnetResourceName, &subnet),
+					testAccCaptureVPCIDFromVPC(&vpc, &vpcID),
+					testAccCreateGuardDutyResourcesFromSubnet(ctx, t, &subnet),
+					testAccCheckVPCGuardDutySecurityGroupExists(ctx, t, &vpc),
+					testAccCheckVPCGuardDutyEndpointExists(ctx, t, &vpc),
+				),
+			},
+		},
```
gdavison: This should be broken up into multiple steps:

* The first step should simply create the VPC and subnet
* The second step should use a `PreConfig` function to configure the Security Group and VPC Endpoints. The second step should then validate that they exist

internal/service/ec2/vpc_test.go
```diff
@@ -1657,3 +1658,699 @@ import {
...
+	})
+}
+
+func testAccCheckVPCGuardDutySecurityGroupExists(ctx context.Context, t *testing.T, vpc *awstypes.Vpc) resource.TestCheckFunc {
```
gdavison: Since we're capturing the `vpcID`, we can just pass the id instead of the whole VPC

internal/service/ec2/vpc_.go
```diff
@@ -756,3 +787,205 @@ func resourceVPCFlatten(ctx context.Context, client *conns.AWSClient, vpc *awsty
...
+			return deletedCount, formatGuardDutyError("deleting", "security group", groupID, wrapThrottlingError(err)), ""
+		}
+
+		log.Printf("[INFO] Successfully deleted GuardDuty security group: %s", groupID)
+		deletedCount++
+	}
+
+	return deletedCount, nil, ""
+}
+func detectAndDeleteGuardDutyVPCEndpoints(ctx context.Context, conn *ec2.Client, vpcID string) (error, string) {
```
gdavison:
```suggestion
}

func detectAndDeleteGuardDutyVPCEndpoints(ctx context.Context, conn *ec2.Client, vpcID string) (error, string) {
```

internal/service/ec2/vpc_.go
```diff
@@ -756,3 +787,205 @@ func resourceVPCFlatten(ctx context.Context, client *conns.AWSClient, vpc *awsty
...
+				Values: []string{guardDutyServiceNamePattern},
+			},
+			{
+				Name:   aws.String("tag:" + guardDutyManagedTagKey),
+				Values: []string{guardDutyManagedTagValue},
+			},
+		},
+	}
+
+	output, err := conn.DescribeVpcEndpoints(ctx, input)
```
gdavison: This same code to find GuardDuty VPC Endpoints is used in multiple places. It can be factored out

gdavison (changes requested):
Thanks for the PR, @JMRacke. I've taken a partial first pass and have a number of comments.

There are some formatting issues that are preventing the remaining linters from running. Please run `make fmt` to resolve them. You can run the linters locally by running `make lint`

internal/service/ec2/vpc_.go
JMRacke: This is intentional for the best-effort design described in the PR description. Users who don't have GuardDuty enabled shouldn't need to add extra IAM permissions for VPC deletion to continue working as it does today. If the IAM role lacks the optional permissions, we warn and let the original delete retry proceed — if GuardDuty wasn't actually the cause of the `DependencyViolation`, the delete will fail with the original error anyway, which is the same behavior as before this change. Surfacing it as a warning gives operators visibility into what was attempted without blocking their workflow.

internal/service/ec2/vpc_.go
JMRacke: The reactive approach was intentional. By only checking for GuardDuty resources when we get a `DependencyViolation`, this change is invisible to the vast majority of customers who have never enabled GuardDuty — no extra API calls on their VPC deletes. If we checked proactively, every `terraform destroy` on every VPC would make additional `DescribeVpcEndpoints` and `DescribeSecurityGroups` calls, even when there's nothing to clean up.

This was also the approach agreed upon in earlier discussions between our team and HashiCorp when scoping this work.

internal/service/ec2/vpc_.go
JMRacke: Yes will submit fix

internal/service/ec2/vpc_test.go
JMRacke: Removed.

internal/service/ec2/vpc_.go
gdavison: OK. Since logs are only written when requested, we should also emit warning `diag.Diagnostic` messages. These should only be added to the diagnostics returned by `Delete` if there is a `DependencyViolation` error.

internal/service/ec2/vpc_.go
gdavison: Good point. Can you add a comment stating that, please?

internal/service/ec2/vpc_.go
```diff
@@ -390,11 +390,37 @@ func resourceVPCDelete(ctx context.Context, d *schema.ResourceData, meta any) di
 	conn := meta.(*conns.AWSClient).EC2Client(ctx)
 
 	log.Printf("[INFO] Deleting EC2 VPC: %s", d.Id())
+
```
gdavison: Since we're adding a lot of log statements in `resourceVPCDelete`, let's update to using `tflog`.

We can set the `id` once by using `tflog.SetField`.

Replace this line with

```go
ctx := tflog.SetField(ctx, logging.ResourceAttributeKey(names.AttrID), d.Id())
tflog.Info(ctx, "Deleting EC2 VPC")
```

internal/service/ec2/vpc_.go
```diff
@@ -390,11 +390,37 @@ func resourceVPCDelete(ctx context.Context, d *schema.ResourceData, meta any) di
...
 	}
+
+	var guardDutyWarnings []string
+
 	_, err := tfresource.RetryWhenAWSErrCodeEquals(ctx, d.Timeout(schema.TimeoutDelete), func(ctx context.Context) (any, error) {
-		return conn.DeleteVpc(ctx, &input)
+		result, err := conn.DeleteVpc(ctx, &input)
+
+		if tfawserr.ErrCodeEquals(err, errCodeDependencyViolation) {
+			log.Printf("[DEBUG] VPC deletion failed with DependencyViolation, checking for GuardDuty resources")
```
gdavison: The `log.Printf` statements in this conditional can be replaced with `tflog.<Level>(ctx, "...")`

e.g.

```go
tflog.Debug(ctx, "VPC deletion failed with DependencyViolation, checking for GuardDuty resources")
```

internal/service/ec2/vpc_.go
```diff
@@ -390,11 +390,37 @@ func resourceVPCDelete(ctx context.Context, d *schema.ResourceData, meta any) di
...
 
 	log.Printf("[INFO] Deleting EC2 VPC: %s", d.Id())
+
 	input := ec2.DeleteVpcInput{
 		VpcId: aws.String(d.Id()),
 	}
+
+	var guardDutyWarnings []string
+
 	_, err := tfresource.RetryWhenAWSErrCodeEquals(ctx, d.Timeout(schema.TimeoutDelete), func(ctx context.Context) (any, error) {
```
gdavison: Trying to remove the GuardDuty dependencies should be moved out of the Retry. Previously, this would loop for the Delete timeout (default 20 min) allowing for eventual consistency for the dependencies to be deregistered after deletion.

By putting the GuardDuty checking inside the loop, if we're waiting for, e.g. Subnet deletion to be seen by the EC2 service, we're going to be checking for GuardDuty each time. A better approach would be to do a first attempt at deletion and check for GuardDuty resources if it fails with a dependency violation and then start the Retry loop.

internal/service/ec2/vpc_.go
```diff
@@ -756,3 +787,190 @@ func resourceVPCFlatten(ctx context.Context, client *conns.AWSClient, vpc *awsty
...
+		}
+		return "", fmt.Errorf("describing GuardDuty VPC endpoints in VPC %s: %w", vpcID, err)
+	}
+
+	if len(endpoints) == 0 {
+		log.Printf("[DEBUG] No GuardDuty VPC endpoints found in VPC %s", vpcID)
+		return "", nil
+	}
+
+	log.Printf("[INFO] Found %d GuardDuty VPC endpoint(s) in VPC %s, deleting", len(endpoints), vpcID)
```
gdavison: This can be Debug level

internal/service/ec2/consts.go
```diff
@@ -70,6 +70,13 @@ const (
 	vpnStateModifying = "modifying"
 )
 
+const (
+	guardDutyServiceNamePattern  = "*guardduty-data*"
+	GuardDutySecurityGroupPrefix = "GuardDutyManagedSecurityGroup-"
```
gdavison: Since this is only used in tests, it can be made private and exported in `exports_test.go`

internal/service/ec2/exports.go
```diff
@@ -34,6 +34,8 @@ var (
 	WaitTransitGatewayAttachmentAccepted                                    = waitTransitGatewayAttachmentAccepted
 	WaitTransitGatewayAttachmentDeleted                                     = waitTransitGatewayAttachmentDeleted
 	WaitVPCEndpointAvailable                                                = waitVPCEndpointAvailable
+	WaitVPCEndpointDeleted                                                  = waitVPCEndpointDeleted
+	VPCEndpointDeletionTimeout                                              = vpcEndpointDeletionTimeout
```
gdavison: Are these used anywhere?

JMRacke:
Thanks @gdavison! I added that comment in `vpc_.go` and `vpc_subnet.go` above the error handing for GuardDuty resources.

internal/service/ec2/vpc_.go
```diff
@@ -756,3 +787,190 @@ func resourceVPCFlatten(ctx context.Context, client *conns.AWSClient, vpc *awsty
...
+				Values: []string{guardDutyGroupName},
+			},
+			{
+				Name:   aws.String("tag:" + guardDutyManagedTagKey),
+				Values: []string{guardDutyManagedTagValue},
+			},
+		},
+	}
+
+	output, err := conn.DescribeSecurityGroups(ctx, input)
```
gdavison: Finding GuardDuty security groups can be factored out, since it's used in multiple places.

The implementation can also make use of the existing `findSecurityGroups` function and use `newAttributeFilterList` to build the filters

internal/service/ec2/vpc_subnet.go
```diff
@@ -767,3 +804,176 @@ func resourceSubnetFlatten(ctx context.Context, subnet *awstypes.Subnet, rd *sch
...
+	for _, tag := range tags {
+		if aws.ToString(tag.Key) == guardDutyManagedTagKey &&
+			aws.ToString(tag.Value) == guardDutyManagedTagValue {
+			return true
+		}
+	}
+	return false
+}
+
+func findGuardDutyVPCEndpoints(ctx context.Context, conn *ec2.Client, vpcID string) ([]awstypes.VpcEndpoint, error) {
```
gdavison, line 851: This implementation can make use of the existing `findVPCEndpoints` and use `newAttributeFilterList` to build the filters.

This function should be called in several places in testing as well

gdavison (changes requested):
Thanks for the updates. Please do not flatten changes into a single commit. It makes it challenging to view the changes made.

I have some more comments to address.

internal/service/ec2/vpc_.go
```diff
@@ -795,31 +800,14 @@ func vpcARN(ctx context.Context, c *conns.AWSClient, accountID, vpcID string) st
 }
 
 func detectAndDeleteGuardDutySecurityGroups(ctx context.Context, conn *ec2.Client, vpcID string) (int, string, error) {
-	log.Printf("[INFO] Detecting GuardDuty security groups in VPC: %s", vpcID)
+	tflog.Debug(ctx, "Detecting GuardDuty security groups in VPC", map[string]any{names.AttrVPCID: vpcID})
```
gdavison: Because we've called `tflog.SetField(ctx, logging.KeyResourceId, d.Id())` in `resourceVPCDelete`, the `id` value is already included in the log data, so we can skip setting it in the log statements

internal/service/ec2/vpc_.go
gdavison: Since logs are only written when requested, we should also emit warning diag.Diagnostic messages. These should only be added to the diagnostics returned by Delete if there is a DependencyViolation error.

internal/service/ec2/exports_test.go
```diff
@@ -318,6 +318,13 @@ var (
 	VPCMigrateState                                             = vpcMigrateState
 	VPNGatewayRoutePropagationParseID                           = vpnGatewayRoutePropagationParseID
 	WaitVolumeAttachmentCreated                                 = waitVolumeAttachmentCreated
+	HasGuardDutyManagedTag                                      = hasGuardDutyManagedTag
+	FindGuardDutyVPCEndpoints                                   = findGuardDutyVPCEndpoints
+	FindGuardDutySecurityGroups                                 = findGuardDutySecurityGroups
```
gdavison, line 323: This doesn't appear to be used

internal/service/ec2/vpc_subnet_test.go
```diff
@@ -1703,3 +1704,1422 @@ resource "aws_subnet" "test" {
...
+					Name:   aws.String("service-name"),
+					Values: []string{tfec2.GuardDutyServiceNamePattern},
+				},
+				{
+					Name:   aws.String("tag:" + tfec2.GuardDutyManagedTagKey),
+					Values: []string{acctest.CtTrue},
+				},
+			},
+		}
+		endpointsOutput, err := conn.DescribeVpcEndpoints(ctx, endpointsInput)
```
gdavison: We can also use `findGuardDutyVPCEndpoints` here and other places in tests

internal/service/ec2/vpc_subnet_test.go
```diff
@@ -1703,3 +1704,1422 @@ resource "aws_subnet" "test" {
...
+					Name:   aws.String("group-name"),
+					Values: []string{fmt.Sprintf("%s%s", tfec2.GuardDutySecurityGroupPrefix, vpcID)},
+				},
+				{
+					Name:   aws.String("tag:" + tfec2.GuardDutyManagedTagKey),
+					Values: []string{acctest.CtTrue},
+				},
+			},
+		}
+		sgOutput, err := conn.DescribeSecurityGroups(ctx, sgInput)
```
gdavison: We can also use `findGuardDutySecurityGroups` here and other places in tests

internal/service/ec2/vpc_subnet.go
```diff
@@ -413,21 +415,61 @@ func resourceSubnetDelete(ctx context.Context, d *schema.ResourceData, meta any)
...
+				if cleanupErr != nil {
+					tflog.Warn(ctx, "Error dissociating GuardDuty VPC endpoints", map[string]any{
+						names.AttrVPCID: vpcID, "error": cleanupErr.Error(),
+					})
+				}
+				return true, err
+			}
+			return false, err
+		},
+	)
```
gdavison: This should use the same flow as deleting the VPC rather than `tfresource.RetryWhen`

internal/service/ec2/vpc_subnet.go
```diff
@@ -767,3 +809,182 @@ func resourceSubnetFlatten(ctx context.Context, subnet *awstypes.Subnet, rd *sch
...
+	return findSecurityGroups(ctx, conn, &ec2.DescribeSecurityGroupsInput{
+		Filters: newAttributeFilterList(map[string]string{
+			"vpc-id":                        vpcID,
+			"group-name":                    groupName,
+			"tag:" + guardDutyManagedTagKey: guardDutyManagedTagValue,
+		}),
+	})
+}
+
+func isSubnetOwnedBySameAccount(ctx context.Context, conn *ec2.Client, subnetID, vpcID string) (bool, error) {
```
gdavison: We already should know the owning account ID for the subnet: the provider will only be managing subnets in the configured account, so it can be retrieved by calling the `AccountID(ctx)` function on the `*conns.AWSClient` passed in the `meta` parameter to the `Delete` function.

Should the logic here be that the VPC Endpoint should only be disassociated from the Subnet if the Subnet is owned by the same account as the VPC? So that, for example, if the Subnet is owned by the current account, but it's in a VPC shared by another account, the Subnet will **not** be disassociated?

internal/service/ec2/vpc_subnet.go
```diff
@@ -767,3 +809,182 @@ func resourceSubnetFlatten(ctx context.Context, subnet *awstypes.Subnet, rd *sch
...
+		}
+		tflog.Warn(ctx, "Error checking subnet ownership for GuardDuty cleanup", map[string]any{
+			"subnet_id": subnetID, names.AttrVPCID: vpcID, "error": err.Error(),
+		})
+		return "", nil
+	}
+	if !sameAccount {
+		tflog.Debug(ctx, "Subnet is a shared subnet, skipping GuardDuty cleanup", map[string]any{"subnet_id": subnetID})
+		return "", nil
+	}
```
gdavison: Because of how Terraform works, only working with resources in the account configured on the provider, the Subnet will not be shared. If `sameAccount` is false, it means that the VPC will be shared.

internal/service/ec2/vpc_subnet_test.go
```diff
@@ -1703,3 +1704,1422 @@ resource "aws_subnet" "test" {
...
+				// Step 2: Remove the subnet from the config (VPC only remains).
+				// Terraform tries to delete the subnet, gets DependencyViolation
+				// because the out-of-band GuardDuty endpoint is still associated,
+				// and dissociateGuardDutyVPCEndpoints runs to clean it up.
+				Config: testAccVPCSubnetConfig_guardDutyLastSubnet_withoutSubnet(rName),
+			},
+		},
+	})
+}
+
```
gdavison, line 1809: We should also add tests for working with shared VPCs

gdavison (dismissed):
Thanks for the updates. I have some more recommendations.

Please don't rebase the PR, as the moving references make the PR harder to work with.

gdavison:
Hi @JMRacke, can you include instructions for how to run the test `TestAccVPCSubnet_guardDutySharedVPC`, please? I keep getting the error 

> UnauthorizedOperation: You are not authorized to perform CreateSubnet operation. A subnet in this vpc is shared but the provided object is not owned by you

I did a quick reading of documentation, and it suggests that Subnets can only be created by the same account that owns the VPC. Is that correct?

JMRacke:
Hi @gdavison, thanks for the question about `TestAccVPCSubnet_guardDutySharedVPC`.

After investigating this with two AWS accounts in the same Organization (management account as VPC owner, member account as participant), I've determined that a shared VPC acceptance test for GuardDuty cleanup is not viable. Previously it was skipping and I missed that in my run of the acceptance tests. Here's why:

1. **Participant accounts cannot create or delete subnets in a shared VPC** — they receive `UnauthorizedOperation` (403), never `DependencyViolation`. This means the GuardDuty cleanup code path in `dissociateGuardDutyVPCEndpoints` is unreachable from a participant account.

2. **When GuardDuty is enabled with shared VPC support, the VPC endpoint is always created in the VPC owner's account**, regardless of which account triggered it ([AWS docs](https://docs.aws.amazon.com/guardduty/latest/ug/runtime-monitoring-shared-vpc.html)). So the only account that can hit `DependencyViolation` on subnet deletion is the VPC owner — and for them, `isVPCOwnedByAccount` always returns `true`.

3. The `isVPCOwnedByAccount` check serves as a **defensive guard** in case these assumptions change in the future. The ownership comparison logic is covered by the unit test `TestIsVPCOwnedByAccount_logic`.

In this commit I've:
- Removed `TestAccVPCSubnet_guardDutySharedVPC` and its config functions
- Added a doc comment on `dissociateGuardDutyVPCEndpoints` explaining the shared VPC behavior with a link to the AWS documentation
- Added a note on the unit test explaining why the acceptance test is not viable
- Kept `TestIsVPCOwnedByAccount_logic` for the ownership comparison coverage

internal/service/ec2/vpc_.go
```diff
@@ -407,6 +411,52 @@ func resourceVPCDelete(ctx context.Context, d *schema.ResourceData, meta any) di
 		return diags
 	}
 
+	// Defers checking for GuardDuty-managed resources until we get a DependencyViolation error so tha tno new permissions,
```
jar-b:
```suggestion
	// Defers checking for GuardDuty-managed resources until we get a DependencyViolation error so that no new permissions,
```

internal/service/ec2/vpc_subnet_test.go
```diff
@@ -1703,3 +1786,238 @@ resource "aws_subnet" "test" {
...
+	return `
+resource "aws_vpc" "test" {
+  cidr_block           = "10.1.0.0/16"
+  enable_dns_support   = true
+  enable_dns_hostnames = true
+}
+`
+}
+
+func TestIsUnauthorizedError(t *testing.T) {
```
jar-b: nit - move this up above acceptance tests.

jar-b (approved):
LGTM 🚀

PR #47082, opened 2026-03-24
r/aws_instance: Fix perpetual `source_dest_check` drift with `primary_network_interface`

### LINKED ISSUE #44768: Perpetual drift when source_dest_check is false on a primary_network_interface

```console
Terraform v1.13.3
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v6.17.0
```

* `aws_instance`
* `aws_network_interface`

### Expected Behavior

When `source_dest_check` is set to `false` on an `aws_network_interface` that is then referenced in a `primary_network_interface` block on an `aws_instance` terraform should not perpetually plan to change the value of `source_dest_check` on either resource.

### Actual Behavior

After the first terraform apply subsequent apply's plan to toggle `source_dest_check` on the `aws_instance` and `aws_network_interface` in turn. i.e. after first application on subsequent applies
 - `source_dest_check` will be changed from `false -> true` on the `aws_instance`.
 - If applied, the next run will change `source_dest_check` from `true -> false` on the `aws_network_interface`
 - If applied, we go back to `source_dest_check` `false -> true` on the `aws_instance`.
 - The cycle repeats infinitely

_No response_

<details open>

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "6.17.0"
    }
  }
}

provider "aws" {
  region = "eu-west-1"
}

resource "aws_instance" "this" {
  ami           = "ami-033a3fad07a25c231"
  instance_type = "t3.micro"

  root_block_device {
    volume_size = 20
    volume_type = "gp3"
  }

  primary_network_interface {
    network_interface_id = aws_network_interface.this.id
  }
}

resource "aws_network_interface" "this" {
  subnet_id = aws_subnet.this.id

  source_dest_check = false
}

resource "aws_vpc" "this" {
  cidr_block = "10.0.0.0/24"
}

resource "aws_subnet" "this" {
  vpc_id     = aws_vpc.this.id
  cidr_block = "10.0.0.0/24"
}

```
</details>

### Steps to Reproduce

Repeatedly apply the terraform and observe the cycling behaviour.

### Debug Logging

_No response_

n/a

## Context

There are three relevant facts here about how source dest check works in AWS:

1. The 'source dest check' flag in AWS exists on two levels: at the instance level, and at the network interface level.
2. The EC2 `ModifyInstanceAttribute` API call to change the SourceDestCheck value on the instance in actuality changes *both* flags.
3. The EC2 `ModifyNetworkInterfaceAttribute` API call to change the SourceDestCheck value on the network interface also changes *both* flags.

In general, it does not make sense to manage the instance flag if the network interface flag is set. Hence, the provider only attempts to modify the instance-level flag if the instance does not explicitly specify any network interfaces. However, 'explicitly specifies any network interfaces' is [based on whether 'network_interface' is set on the resource](https://github.com/hashicorp/terraform-provider-aws/blob/03d14bbce7ebf52181345fe6d1a41cc6502fc2be/internal/service/ec2/ec2_instance.go#L1356). This attribute has been **deprecated** since this logic was written: now 'primary_network_interface' is used instead to explicitly designate a network interface, but the logic does not account for this.

Thus, when `primary_network_interface` is used, the provider invokes `ModifyInstanceAttribute` to change the source-dest flag at the instance level. **This is a bug, as in this case, the flag should be managed at the network interface level only.**

In addition, it is not possible to set a non-default value of the source_dest_check attribute on the instance in this case, as it (correctly) conflicts with `primary_network_interface`.

## When the bug manifests

In the case where we set source_dest_check to false on a network interface, and then assign this as the primary network interface of an EC2 instance:

1. Terraform creates the instance and interface; AWS sets the instance flag to false when the instance is created, as it knows the primary network interface has the flag set to false.
2. Terraform detects drift, as the desired state is the default `true` value for `source_dest_check` on the instance, but AWS returns this value (correctly) as `false`
3. Terraform tries to correct the drift by invoking `ModifyInstanceAttribute` to set `source_dest_check` to `true` on the instance
4. Per fact (2), this causes the **network interface** to also have its `source_dest_check` set to `true`
5. On the next Terraform apply, Terraform detects drift, as the desired state is source_dest_check as `false` for the **network interface** but AWS reports this value (correctly) as `true`
6. Terraform corrects the drift by invoking the `ModifyNetworkInterfaceAttribute` to `false` on the network interface
7. Per fact (3) this causes the **instance** to also have its `source_dest_check` set to `false`
8. We loop back to 2 and continue the cycle indefinitely

## Fixing

The core issue here is the terraform provider attempting to manage the `source_dest_check` flag on the instance when a `primary_network_interface` is set. As mentioned, in provider versions before `6.x` where `network_interface` was not deprecated the provider does not attempt to manage the `source_dest_check` flag on the `aws_instance` due to the presence of the `network_interface` block.

Interestingly this means this behaviour does not manifests on `aws_instance`'s that were _created_ with AWS provider version prior to 6.x - as even if they have been upgraded to use the `primary_network_interface` block the `network_interface` block still appears in the instances terraform state. e.g. an instance created with version 5.x of the provider and then upgraded to 6.x and changed to `primary_network_interface` will have a state that looks like (output from `terraform state show`):
```
    network_interface {
        delete_on_termination = false
        device_index          = 0
        network_card_index    = 0
        network_interface_id  = "eni-xxxxxxxxxxxxxxxx"
    }

    primary_network_interface {
        delete_on_termination = false
        network_interface_id  = "eni-xxxxxxxxxxxxxxxx"
    }
```

vs an instance _created_ with 6.x:
```
    primary_network_interface {
        delete_on_termination = false
        network_interface_id  = "eni-xxxxxxxxxxxxxxxx"
    }
```

The fact that the `network_interface` hangs around in the state causes us to never enter the if statement [here](https://github.com/hashicorp/terraform-provider-aws/blob/03d14bbce7ebf52181345fe6d1a41cc6502fc2be/internal/service/ec2/ec2_instance.go#L1356).

I suspect the fix is as simple as making the condition here check if either `network_interface` or `primary_network_interface` are set, rather than just `network_interface`.

## Workaround

For now this issue can be worked around by setting `ignore_changes` on `source_dest_check` in the `aws_instance` if the instance has a `primary_network_interface` block.

Yes

TusharSariya:
## Rollback Plan

### Description

The `DiffSuppressFunc` and update guard for `source_dest_check` only checked for the deprecated `network_interface` block, not the newer `primary_network_interface` block. This caused an infinite plan/apply loop when an ENI with `source_dest_check = false` was attached via `primary_network_interface`, because the provider would repeatedly call `ModifyInstanceAttribute` to reset it to the default of `true`.

### Relations

Closes #44768

```console
% make testacc TESTS="TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift|TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck" PKG=ec2 GO_VER=go                                                  
make: Running acceptance tests on branch: 🌿 44768 🌿...
TF_ACC=1 go test ./internal/service/ec2/... -v -count 1 -parallel 20 -run='TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift|TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck'  -timeout 360m -vet=off
2026/03/24 14:12:08 Creating Terraform AWS Provider (SDKv2-style)...
2026/03/24 14:12:08 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck
=== PAUSE TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck
=== RUN   TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift
=== PAUSE TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift
=== CONT  TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck
=== CONT  TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift
--- PASS: TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck (318.95s)
--- PASS: TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift (342.57s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ec2        347.514s

...
```

### DIFF
--- a/.changelog/44768.txt
+++ b/.changelog/44768.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_instance: Fix perpetual `source_dest_check` drift when using `primary_network_interface` with an ENI that has `source_dest_check` set to `false`
+```
--- a/internal/service/ec2/ec2_instance.go
+++ b/internal/service/ec2/ec2_instance.go
@@ -890,9 +890,10 @@ func resourceInstance() *schema.Resource {
 				Optional: true,
 				Default:  true,
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					// Suppress diff if network_interface is set
-					_, ok := d.GetOk("network_interface")
-					return ok
+					// Suppress diff if network_interface or primary_network_interface is set
+					_, niOk := d.GetOk("network_interface")
+					_, pniOk := d.GetOk("primary_network_interface")
+					return niOk || pniOk
 				},
 			},
 			"spot_instance_request_id": {
@@ -1414,9 +1415,11 @@ func resourceInstanceUpdate(ctx context.Context, d *schema.ResourceData, meta an
 		}
 	}
 
-	// SourceDestCheck can only be modified on an instance without manually specified network interfaces.
-	// SourceDestCheck, in that case, is configured at the network interface level
-	if _, ok := d.GetOk("network_interface"); !ok {
+	// SourceDestCheck should only be modified at the instance level when no explicit network interface
+	// is attached via network_interface or primary_network_interface. Otherwise it is managed on the ENI.
+	_, niOk := d.GetOk("network_interface")
+	_, pniOk := d.GetOk("primary_network_interface")
+	if !niOk && !pniOk {
 		// If we have a new resource and source_dest_check is still true, don't modify
 		sourceDestCheck := d.Get("source_dest_check").(bool)
 
--- a/internal/service/ec2/ec2_instance_test.go
+++ b/internal/service/ec2/ec2_instance_test.go
@@ -3436,6 +3436,39 @@ func TestAccEC2Instance_NetworkInterface_primaryNetworkInterfaceSourceDestCheck(
 	})
 }
 
+func TestAccEC2Instance_PrimaryNetworkInterface_sourceDestCheckNoDrift(t *testing.T) {
+	ctx := acctest.Context(t)
+	var instance awstypes.Instance
+	var eni awstypes.NetworkInterface
+	resourceName := "aws_instance.test"
+	eniResourceName := "aws_network_interface.test"
+	rName := acctest.RandomWithPrefix(t, acctest.ResourcePrefix)
+
+	acctest.ParallelTest(ctx, t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(ctx, t) },
+		ErrorCheck:               acctest.ErrorCheck(t, names.EC2ServiceID),
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories,
+		CheckDestroy:             testAccCheckInstanceDestroy(ctx, t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccInstanceConfig_primaryNetworkInterface_sourceDestCheck(rName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckInstanceExists(ctx, t, resourceName, &instance),
+					testAccCheckENIExists(ctx, t, eniResourceName, &eni),
+					resource.TestCheckResourceAttr(eniResourceName, "source_dest_check", acctest.CtFalse),
+				),
+			},
+			// Re-plan same config — must produce no diff.
+			// Before the fix for #44768, this step fails because
+			// source_dest_check drifts from false to true on the instance.
+			{
+				Config:   testAccInstanceConfig_primaryNetworkInterface_sourceDestCheck(rName),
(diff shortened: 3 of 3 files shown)


dyegoe:
Any reviewer for this? :)
The drift is still present 6.43

PR #47096, opened 2026-03-25
resource/aws_ecs_service: clear lifecycle hooks on removal

boxa99:
## Rollback Plan

If this change needs to be reverted, we can ship a follow-up provider release that restores the previous `aws_ecs_service` update behavior.

### Description

This fixes `aws_ecs_service` so removing the final `deployment_configuration.lifecycle_hook` from configuration also removes it from the ECS service in AWS.

Previously, when `deployment_configuration` changed and no `lifecycle_hook` values remained configured, the provider omitted `DeploymentConfiguration.LifecycleHooks` from the ECS `UpdateService` request. ECS requires an explicit empty lifecycle hook list to clear existing hooks, so the remote service could retain the old hook after Terraform removed it from configuration.

This change initializes `DeploymentConfiguration.LifecycleHooks` to an explicit empty slice during update and overwrites it only if hooks are still configured.

This PR also adds an acceptance regression test covering a transition from one lifecycle hook to zero lifecycle hooks and verifies that the remote ECS service no longer has lifecycle hooks.

### Relations

N/A

### References

N/A

Acceptance tests not run locally; requires AWS acceptance test environment.

Local validation run:

```console
% go test ./internal/service/ecs -run '^$' -timeout 90s
ok      github.com/hashicorp/terraform-provider-aws/internal/service/ecs        [no tests to run]

### DIFF
--- a/.changelog/47096.txt
+++ b/.changelog/47096.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_ecs_service: Fix `deployment_configuration.lifecycle_hook` not being removed from the ECS service when removed from configuration
+```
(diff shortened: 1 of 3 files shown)


boxa99:
Bump. Any chance of a review of this? I think it's quite a straightforward fix.

Issue #38984, opened 2024-08-22
[Bug]: RDS DB upgrade from 14.9 to 16.3 and aws_db_parameter_group breaks


Selva-Shan:
### Terraform Core Version

1.5.7

### AWS Provider Version

5.47.0

### Affected Resource(s)

- aws_rds_cluster
- aws_rds_cluster_instance
- aws_db_parameter_group

### Expected Behavior

RDS version upgrade  from 14.9 to 16.3. RDS version should be upgraded successfully. 

### Actual Behavior

For some reason db parameter group name Cluster instance and Cluster are different, Which should not be the case as we are not modifying the db parameter group here. 

Upon checking the status of the cluster instance parameter group it says "pending-reboot". 

Cluster  parameter group name shows default group name "default:aurora-postgresql-16"
Cluster instance parameter group name shows custom one "postgres"

Below is the error message while applying the plan,

```console
│ Error: Provider produced inconsistent final plan
│ 
│ When expanding the plan for
│ module.use_cases.module.iota_utilization_and_benchmarking_db[0].aws_rds_cluster_instance.this[0]
│ to include new values learned so far during apply, provider
│ "registry.terraform.io/hashicorp/aws" produced an invalid new value for
│ .engine_version: was cty.StringVal("16.3"), but now cty.StringVal("14.9").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
╵
╷
│ Error: Provider produced inconsistent final plan
│ 
│ When expanding the plan for
│ module.use_cases.module.iota_utilization_and_benchmarking_db[0].aws_rds_cluster_instance.this[1]
│ to include new values learned so far during apply, provider
│ "registry.terraform.io/hashicorp/aws" produced an invalid new value for
│ .engine_version: was cty.StringVal("16.3"), but now cty.StringVal("14.9").
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```

### Terraform Configuration Files

Code Snippet

```hcl
#########################
# Aurora DB AWS Cluster #
#########################
resource "aws_rds_cluster" "this" {
  cluster_identifier          ="dev-cluster-db"
  engine                      ="aurora-postgresql"
  engine_mode                 = "provisioned"
  engine_version              = 16,3
  database_name               = "postgres"
  master_username             = "**************"
  manage_master_user_password = true
  port                        = 5432
  apply_immediately           = true
  allow_major_version_upgrade = true
[...]

Issue #35385, opened 2024-01-19
[Bug]: Issue is RDS Aurora engine version upgrade from 5.7 to 8.0 because of parameter group issue


varunkumarmnnit:
### Terraform Core Version

0.13

### AWS Provider Version

v4.32.0

### Affected Resource(s)

aws_rds_cluster
aws_rds_cluster_instance
aws_rds_cluster_parameter_group

### Expected Behavior

RDS cluster engine version should be upgraded with 8.0 with new updated parameter grp

### Actual Behavior

Getting error where rds cluster is referring to older parameter group even though I haven't referred older one in the updated engine version of rds cluster .

InvalidParameterCombination: The current DB instance parameter group {service-name}-{service-name}-db-parameter-group is custom. You must explicitly specify a new DB instance parameter group, either default or custom, for the engine version upgrade

### Relevant Error/Panic Output Snippet

```shell
Error: updating RDS Cluster (brands-{service-name}-{service-name}-db): InvalidParameterCombination: The current DB instance parameter group {service-name}-{service-name}-db-parameter-group is custom. You must explicitly specify a new DB instance parameter group, either default or custom, for the engine version upgrade.
status code: 400, request id: 6cde5aab-59ae-4b98-85ce-cb38915d29ce

on rds.tf line 2, in resource "aws_rds_cluster" "{service-name}-db":
2: resource "aws_rds_cluster" "{service-name}-db" {
```

### Terraform Configuration Files

```
  apply_immediately                   = false
  engine                              = local.rds_engine
  engine_version                      = local.rds_engine_aurora8
  cluster_identifier                  = "${var.rds_cluster_name}-db"
  database_name                       = var.rds_name
  master_username                     = var.rds_username
  master_password                     = module.vault_db_password.secret_value
  allow_major_version_upgrade          = true
  backup_retention_period             = 7
  preferred_backup_window             = var.backup_window
  preferred_maintenance_window        = var.maintenance_window
  vpc_security_group_ids              = [aws_security_group.{service-name}-db-sg.id]
  db_subnet_group_name                = aws_db_subnet_group.{service-name}-db.name
  iam_database_authentication_enabled = true
  deletion_protection                 = true
  tags                                = module.{service-name}_tags.tags
  db_cluster_parameter_group_name     = aws_rds_cluster_parameter_group.{service-name}_cluster_parameter_group_updated.id
[...]

Issue #28339, opened 2022-12-14
[Bug]: aws_rds_cluster does not wait properly for modifications


jcarlson:
### Terraform Core Version

1.3.6

### AWS Provider Version

4.42.0

### Affected Resource(s)

* aws_rds_cluster

### Expected Behavior

Terraform should wait for pending changes to be applied asynchronously

### Actual Behavior

Terraform is only monitoring the `<Status>` attribute of the `DescribeDBClusters` operation. 

Some changes, for example changing the database engine major version, AWS applies those changes asynchronously. When Terraform applies the change and then begins polling for the cluster status to be "available", it finds that the status is _immediately_ available because AWS does not apply the changes right away.

While the cluster is still in the `available` state, the pending changes can be found in another attribute, `PendingModifiedValues`. Terraform should wait until _both_ `Status=available` _and_ `PendingModifiedValues=[]`.

### Terraform Configuration Files

```hcl
terraform {
  required_version = ">= 1.3"

  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "4.42.0"
    }
  }
}

variable "name" {
  type    = string
  default = "issue-28339"
}

variable "postgres_version" {
  type    = string
}

variable "subnet_ids" {
  type = list(string)
}

variable "tags" {
  type = map(string)
  default = {
    repository = "https://github.com/hashicorp/terraform-provider-aws"
    issue      = "https://github.com/hashicorp/terraform-provider-aws/issues/28339"
  }
}

variable "vpc_id" {
  type = string
}

locals {
  engine         = "aurora-postgresql"
  engine_version = var.postgres_version
  family         = "aurora-postgresql${local.major_version}"
  major_version  = split(".", local.engine_version)[0]
}

resource "random_pet" "username" {
  separator = "_"
}

resource "random_password" "password" {
  length  = 24
  special = false
}

resource "random_pet" "database" {
  separator = "_"
}

resource "aws_db_subnet_group" "main" {
  name       = var.name
  subnet_ids = var.subnet_ids
  tags       = var.tags

  lifecycle {
    create_before_destroy = true
    ignore_changes        = [name]
  }
}

resource "aws_security_group" "main" {
  description = "Database security group"
  name_prefix = "${var.name}-"
  tags        = var.tags
  vpc_id      = var.vpc_id

  lifecycle {
    create_before_destroy = true
    ignore_changes        = [description, name_prefix]
  }
}

[...]

Issue #28219, opened 2022-12-06
[Bug]: aws_rds_cluster major version upgrade throws InvalidParameterCombination in second apply attempt


riji-ramachandran:
### Terraform Core Version

1.0.8, 0.14.11

### AWS Provider Version

4.30.0

### Affected Resource(s)

aws_rds_cluster
db_instance_parameter_group_name

### Expected Behavior

Trying to upgrade RDS Aurora cluster with one instance from Postgres engine version 13.7 to 14.3
apply_immediately set to true at cluster level. Also db_instance_parameter_group_name is passed to the cluster.
Expect cluster and instance to be upgraded.

### Actual Behavior

On first plan and apply there are no parameter group related error. However if the first run fails for any given reason on second attempt upgrade fails with error:
Error: updating RDS Cluster: InvalidParameterCombination: The current DB instance parameter group is custom. You must explicitly specify a new DB instance parameter group, either default or custom, for the engine version upgrade
If the db_instance_parameter_group_name is changed then apply works again.

### Relevant Error/Panic Output Snippet

```shell
Error: updating RDS Cluster: InvalidParameterCombination: The current DB instance parameter group is custom. You must explicitly specify a new DB instance parameter group, either default or custom, for the engine version upgrade
```

### Terraform Configuration Files

resource "aws_rds_cluster" "aurora_postgres" {
  depends_on = [
    aws_rds_cluster_parameter_group.aurora_postgres_db,
    aws_db_parameter_group.aurora_instance_parameter_group
  ]
  cluster_identifier              = "test-aurora-upgrade-cluster"
  engine                          = "aurora-postgresql"
  engine_version                  = 14.3
  port                            = 6160
  database_name                   = "test"
  allow_major_version_upgrade     = true
  db_cluster_parameter_group_name = aws_rds_cluster_parameter_group.aurora_postgres_db.id
  db_instance_parameter_group_name =aws_db_parameter_group.aurora_instance_parameter_group.id
  apply_immediately          = true
}

### Steps to Reproduce

- Create an rds aurora postgres cluster with one writer instance - engine version 13.7 and custom cluster and instance parameter groups. Set apply_immediately to true
- Once cluster and instance is created change the engine version to 14.3 also change the parameter groups for cluster and instance to family 14
- Make a deliberate mistake so plan completes but apply fails (i had set apply_immediately to false). Plan succeeds but apply will fail
- Revert apply_immediately to true and run apply again. Will get the below error:
[...]

PR #47121, opened 2026-03-26
resource/aws_db_instance: fix perpetual drift on parameter_group_name during pending major version upgrade

etiennechabert:
### Description

When performing a major PostgreSQL version upgrade (e.g., 16 → 17) with `apply_immediately = false`, the engine version change is correctly deferred to the maintenance window. However, the parameter group — which must also change family (e.g., `postgres16` → `postgres17`) — causes perpetual drift and failed applies.

**Root cause:** During refresh, the provider reads `parameter_group_name` directly from `DBParameterGroups[0].DBParameterGroupName`, which still reflects the current (pre-upgrade) parameter group. Unlike `engine_version`, which has the `engine_version_actual` companion attribute and checks `PendingModifiedValues.EngineVersion` to avoid drift, there is no equivalent mechanism for `parameter_group_name`.

The cycle is:
1. State has the new parameter group (postgres17 family) after initial apply
2. Refresh reads the current parameter group from AWS (still postgres16 family, since the upgrade hasn't happened yet)
3. Terraform plans an update: old pg16 group → new pg17 group
4. Apply fails: `InvalidParameterCombination: The parameter group ... with DBParameterGroupFamily postgres17 can't be used for this instance`
5. Repeat on every plan/apply until the maintenance window

The AWS Console handles this correctly — it accepts both engine version and parameter group changes together, deferred to the maintenance window.

**This PR:**
- Skips overwriting `parameter_group_name` during refresh when `PendingModifiedValues.EngineVersion` is set, preventing the drift loop
- Adds a new `parameter_group_name_actual` computed attribute that always reflects the currently active parameter group, mirroring the existing `engine_version` / `engine_version_actual` pattern

### Relations

Relates #28219
Relates #28339
Relates #35385
Relates #38984

### DIFF
--- a/internal/service/rds/instance.go
+++ b/internal/service/rds/instance.go
@@ -492,6 +492,10 @@ func resourceInstance() *schema.Resource {
 				Optional: true,
 				Computed: true,
 			},
+			"parameter_group_name_actual": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			names.AttrPassword: {
 				Type:          schema.TypeString,
 				Optional:      true,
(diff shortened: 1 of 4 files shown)


internal/service/rds/instance.go
```diff
@@ -2041,7 +2045,8 @@ func resourceInstanceRead(ctx context.Context, d *schema.ResourceData, meta any)
 		d.Set("option_group_name", v.OptionGroupMemberships[0].OptionGroupName)
 	}
 	if len(v.DBParameterGroups) > 0 {
-		d.Set(names.AttrParameterGroupName, v.DBParameterGroups[0].DBParameterGroupName)
+		hasPendingEngineVersionUpgrade := v.PendingModifiedValues != nil && v.PendingModifiedValues.EngineVersion != nil
+		setParameterGroupName(d, aws.ToString(v.DBParameterGroups[0].DBParameterGroupName), hasPendingEngineVersionUpgrade)
```
camilacremoneze, line 2049: 🚀

PR #47122, opened 2026-03-26
docs: clarify aws_msk_bootstrap_brokers attribute descriptions

### LINKED ISSUE #35748: [Docs]: data.aws_msk_bootstrap_brokers has wrong name in example and misleading attribute documentation
### Documentation Link

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/msk_bootstrap_brokers

### Description

The example should read
```
data "aws_msk_bootstrap_brokers" {
```
*NOT*: `data "aws_msk_cluster" "example" {` 

All of the attribute descriptions that say "One or more DNS Names..." is false.  I don't know what they return, but those attributes must allow for the fact that *any* of those lists can be empty; perhaps they return NULL in those cases, but that should be documented.    Those lists should only be populated if the cluster actually supports that particular authentication method.

For serverless clusters, the only attributes that could have *one* or more entries would be [bootstrap_brokers_sasl_iam](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/msk_bootstrap_brokers#bootstrap_brokers_sasl_iam) , since that's the only type supported.

For provisioned clusters, any of them could be empty.

### References

_No response_

None

kovan:
## Description

Clarify the `aws_msk_bootstrap_brokers` data source attribute descriptions:

- Each attribute returns a **comma-separated string**, not a list. The previous descriptions ("One or more DNS names") were ambiguous about the type.
- Each attribute returns an **empty string** when its corresponding authentication method is not enabled on the cluster. The previous descriptions implied values are always populated.

## References

Fixes #35748

## Would you like to implement a fix?

- [x] Implemented

### DIFF
--- a/website/docs/d/msk_bootstrap_brokers.html.markdown
+++ b/website/docs/d/msk_bootstrap_brokers.html.markdown
@@ -29,13 +29,13 @@ This data source supports the following arguments:
 
 This data source exports the following attributes in addition to the arguments above:
 
-* `bootstrap_brokers` - Comma separated list of one or more hostname:port pairs of kafka brokers suitable to bootstrap connectivity to the kafka cluster.
-* `bootstrap_brokers_public_sasl_iam` - One or more DNS names (or IP addresses) and SASL IAM port pairs.
-* `bootstrap_brokers_public_sasl_scram` - One or more DNS names (or IP addresses) and SASL SCRAM port pairs.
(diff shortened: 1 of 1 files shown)


jar-b:
This change appears to have removed the `.archive/changelog` directory. Can you fix up the branch so only the documentation change is present?

kovan:
Fixed - force-pushed a clean branch with only the documentation change. The `.archive/changelog` deletion was an artifact of a partial clone. Sorry about that.

PR #34929, opened 2023-12-14
feat(r/aws_autoscaling_group): add max_healthy_percentage arg

rafaelleonardocruz:
### Description

The aws_autoscaling_group resource doesn't supports the attribute max_healthy_percentage at instance_refresh.preferences, which restricts the sensibility of instance-refresh operations.

### References

-  ([StartInstanceRefresh API spec](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_StartInstanceRefresh.html)) describes the expected behaviour for this field.

### Output from Acceptance Testing

```console
make testacc TESTS=TestAccAutoScalingGroup_InstanceRefresh PKG=autoscaling
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./internal/service/autoscaling/... -v -count 1 -parallel 20 -run='TestAccAutoScalingGroup_InstanceRefresh'  -timeout 360m
=== RUN   TestAccAutoScalingGroup_InstanceRefresh_basic
=== PAUSE TestAccAutoScalingGroup_InstanceRefresh_basic
=== RUN   TestAccAutoScalingGroup_InstanceRefresh_start
=== PAUSE TestAccAutoScalingGroup_InstanceRefresh_start
=== RUN   TestAccAutoScalingGroup_InstanceRefresh_triggers
=== PAUSE TestAccAutoScalingGroup_InstanceRefresh_triggers
=== RUN   TestAccAutoScalingGroup_InstanceRefresh_autoRollback
=== PAUSE TestAccAutoScalingGroup_InstanceRefresh_autoRollback
=== CONT  TestAccAutoScalingGroup_InstanceRefresh_basic
=== CONT  TestAccAutoScalingGroup_InstanceRefresh_triggers
=== CONT  TestAccAutoScalingGroup_InstanceRefresh_start
=== CONT  TestAccAutoScalingGroup_InstanceRefresh_autoRollback
--- PASS: TestAccAutoScalingGroup_InstanceRefresh_start (139.94s)
--- PASS: TestAccAutoScalingGroup_InstanceRefresh_autoRollback (148.68s)
--- PASS: TestAccAutoScalingGroup_InstanceRefresh_triggers (157.52s)
--- PASS: TestAccAutoScalingGroup_InstanceRefresh_basic (218.46s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/autoscaling        220.709s
...
```

### Comments:
- My organization policy doesn't allow me to grant write access to external users in this PR

PR #47188, opened 2026-04-01
fix: remove erroneous default for max_healthy_percentage in instance_refresh preferences

### LINKED ISSUE #39872: [Enhancement]:  Revise https://github.com/hashicorp/terraform-provider-aws/pull/34929 to allow omitting maxHealthyPercentage
### Description

`MaxHealthyPercentage` is an **optional** preference as per AWS documentation, so we should be able to omit it in terraform.  

https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_RefreshPreferences.html
 
> **MaxHealthyPercentage**
> Specifies the maximum percentage of the group that can be in service and healthy, or pending, to support your workload when replacing instances. The value is expressed as a percentage of the desired capacity of the Auto Scaling group. Value range is 100 to 200.
> 
> If you specify MaxHealthyPercentage, you must also specify MinHealthyPercentage, and the difference between them cannot be greater than 100. A larger range increases the number of instances that can be replaced at the same time.
> 
> If you do not specify this property, the default is 100 percent, or the percentage set in the instance maintenance policy for the Auto Scaling group, if defined.
> 
> Type: Integer
> 
> Valid Range: Minimum value of 100. Maximum value of 200.
> 
> **Required: No**

aws_autoscaling_group  

 null is ignored in max_healthy_percentage and is set to 100 all the time.

```
 instance_refresh {
    preferences {
      max_healthy_percentage = null 
   }
```

https://github.com/hashicorp/terraform-provider-aws/pull/34929 set maxHealthyPercentage to 100 by default, but that breaks our instant refresh automation.

We used to omit this preference and terminate our instances one at a time, and detach its volume. The new instance would then be launched, the volume would attach to it. It worked as we wanted, but from `hashicorp/aws v5.32.0` onwards where this PR was added. 

Now we get a new instance launched first which cannot pass the health checks without a volume attachment. As a result we get infinitely looping instance refreshes. 
 
We need to be able to set  `max_healthy_percentage` to **_null_** and have the API not generate max HealthyPercentage 
in StartInstanceRefresh API. It works via AWS CLI and AWS console, so Terraform should support it too. 

This is an example API generated by AWS Console when we choose cost based terminate and launch option and enable `Violate min healthy percentage`. We can see in CloudTrail that StartInstanceRefresh did not include maxHealthyPercentage. 

We cannot recreate this API via Terraform anymore, unless we **pin** Terraform/AWS provider to` hashicorp/aws v5.31.0` or earlier versions. 

     "preferences": {
            "minHealthyPercentage": 100,
            "skipMatching": true,
            "autoRollback": false,
            "scaleInProtectedInstances": "Ignore",
            "standbyInstances": "Ignore"
        } 

  

```terraform
aws_autoscaling_group  


null should we allowed for max_healthy_percentage


 instance_refresh {
    preferences {
      max_healthy_percentage = null 
   }

  
This should then omit maxHealthyPercentage in the StartInstanceRefresh API.
```

### References

This is what broke our automation
PR https://github.com/hashicorp/terraform-provider-aws/pull/34929
https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#5320-january-11-2024 

None

heyimdima:
Fixes #39872

Since PR #34929 (v5.32.0), the `max_healthy_percentage` attribute in `instance_refresh.preferences` had `Default: 100` in the schema. This caused the provider to always send `MaxHealthyPercentage=100` to the AWS `StartInstanceRefresh` API even when the user did not configure it.

The AWS API treats these two cases differently:
- **Omitted**: legacy "terminate and launch" behavior (terminates instance first, then launches replacement)
- **Explicitly set to 100**: "launch before terminating" behavior (launches new instance first, waits for warmup, then terminates old one)

This silently changed the instance replacement strategy for all users who didn't set `max_healthy_percentage` since v5.32.0.

## Fix

1. Remove `Default: 100` from the `max_healthy_percentage` schema field — the field is now `0` (unset) when not configured by the user.
2. Guard the API call with `&& v != 0` in `expandRefreshPreferences` — `0` is outside the valid 100–200 range so it reliably indicates the attribute was not configured, and `MaxHealthyPercentage` is omitted from the API request.
3. Update 4 test assertions that expected `"100"` from the old default to now expect `"0"` (MinHealthyPercentage, SkipMatching, ScaleInProtectedInstances, StandbyInstances test steps where `max_healthy_percentage` is not explicitly set).

## References
- Regression introduced by #34929


heyimdima:
## Implementation Note

The `&& v != 0` sentinel in `expandRefreshPreferences` works correctly here because `0` is outside the valid range (100–200) and can never be a legitimate user-supplied value. However, it relies on implicit domain knowledge to understand why `0` means "not set".

The more idiomatic pattern in this codebase would be `nullable.TypeNullableInt`, which is already used by `instance_warmup` in the same preferences block to explicitly distinguish "not configured" from "explicitly set":

```go
"max_healthy_percentage": {
    Type:         nullable.TypeNullableInt,
    Optional:     true,
    ValidateFunc: nullable.ValidateTypeStringNullableIntBetween(100, 200),
},
```

```go
if v, null, _ := nullable.Int(tfMap["max_healthy_percentage"].(string)).ValueInt32(); !null {
    apiObject.MaxHealthyPercentage = aws.Int32(v)
}
```

I kept this PR minimal (bug fix only) since `v != 0` is functionally correct and the nullable approach would touch the state schema. Happy to switch to the nullable pattern if the maintainers prefer it — just let me know.

PR #47294, opened 2026-04-05
Fix aws_organizations_account delete when account already closed

### LINKED ISSUE #46627: aws_organizations_account delete fails with AccountAlreadyClosedException when account is closed outside of Terraform

This error comes up using pulumi-aws v6.83.0 which bridges to terraform-provider-aws v5.100.0.

However, this bug is still in the latest version of terraform-provider-aws since there's nothing handling `AccountAlreadyClosedException` in https://github.com/hashicorp/terraform-provider-aws/blob/4bc11f1fa03c45a3d685e3d153e081c1621067e6/internal/service/organizations/account.go#L331C24-L331C48.

- `aws_organizations_account`

### Expected Behavior

When destroying an `aws_organizations_account` resource with `close_on_deletion = true`, if the account is already closed, the delete should succeed since the desired end state (account closed) is already achieved.

### Actual Behavior

The delete fails with the error:
```
AccountAlreadyClosedException: You attempted to close an account that is already closed.
```

This can happen when a previous destroy attempt partially succeeds or times out. Or when the account was closed manually outside of Terraform.

```console
AccountAlreadyClosedException: You attempted to close an account that is already closed.
```

```hcl
resource "aws_organizations_account" "example" {
  name              = "example-account"
  email             = "example@example.com"
  close_on_deletion = true
}
```
The configuration above should be all that's needed to reproduce, but I think the code itself shows that `AccountAlreadyClosedException` is not taken into account.

I think that there needs to be a way to handle `AccountAlreadyClosedException` similar to `AccountNotFoundException` as it is written here: https://github.com/hashicorp/terraform-provider-aws/blob/4bc11f1fa03c45a3d685e3d153e081c1621067e6/internal/service/organizations/account.go#L331C24-L331C48

### Steps to Reproduce

1. Apply the configuration (set up an AWS organization account using Terraform with `close_on_deletion = true`
2. Close the account outside of Terraform (e.g. through AWS console)
3. Run `terraform destroy` and get the `AccountAlreadyClosedException` error

### Debug Logging

_No response_

n/a

This was reported in #24694 but was closed as stale.

No

TomWilliams-12:
### Description

Fixes #46627.

When destroying an `aws_organizations_account` resource with `close_on_deletion = true`, the delete fails with `AccountAlreadyClosedException` if the account was already closed outside of Terraform.

This PR handles `AccountAlreadyClosedException` as a no-op during deletion, matching the existing handling of `AccountNotFoundException` on the same code path. The desired state (account closed) is already achieved, so the error can be safely ignored.

### Changes

- Handle `AccountAlreadyClosedException` alongside `AccountNotFoundException` in `resourceAccountDelete`
- Added changelog entry

### Output from acceptance testing

Difficult to feasibly test, it would require a real AWS Organizations accounts with close cooldowns. The change follows the identical established pattern used for AccountNotFoundException on the same line.

### DIFF
--- a/.changelog/46627.txt
+++ b/.changelog/46627.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_organizations_account: Fix `AccountAlreadyClosedException` error when deleting an account that has already been closed with `close_on_deletion` set to `true`
+```
--- a/internal/service/organizations/account.go
+++ b/internal/service/organizations/account.go
@@ -329,7 +329,7 @@ func resourceAccountDelete(ctx context.Context, d *schema.ResourceData, meta any
 		})
 	}
 
-	if errs.IsA[*awstypes.AccountNotFoundException](err) {
+	if errs.IsA[*awstypes.AccountNotFoundException](err) || errs.IsA[*awstypes.AccountAlreadyClosedException](err) {
 		return diags
 	}
 


ewbankkit (approved):
LGTM :rocket:.

ewbankkit:
@TomWilliams-12 Thanks for the contribution :tada: :clap:.

PR #47302, opened 2026-04-06
Add `AGENTS.md`

ewbankkit:
## Rollback Plan

### Description

Adds an [`AGENTS.md`](https://agents.md/) file.
The initial commit was created via [opencode](https://opencode.ai/docs/rules#initialize).
The second commit was create by Kiro.


AGENTS.md
jar-b, line 1: The more technical sections (e.g. below `Common Commands`) may eventually make sense for task-specific documentation to avoid bloating the basics, but as we don't have that broken out today I'm good with starting here and seeing how it works.

jar-b (dismissed):

AGENTS.md
ewbankkit: I know, DRY. All the examples out there have this kind of instructions in `AGENTS.md` but that could be because they have no other task-specific documentation.
We should ensure that those task-specific documents are more agent friendly:  sequence of steps the agent follows, ending in a defined exit criterion rather an essay.

ewbankkit: We must iterate. Nobody knows...

ewbankkit: I've left them in for now but I feel we should extract.

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+Always use `skaff` — do not copy an older resource:
+
+```bash
+make skaff
+skaff resource --name ExampleThing --service example
+```
+
+After scaffolding:
+1. Fill in the schema and CRUD handlers.
+2. Add `@FrameworkResource("aws_example_thing", name="Example Thing")` (Framework) or `@SDKResource(...)` (SDKv2) annotation.
```
jar-b: To avoid implying that _new_ SDK V2 resources are acceptable.

```suggestion
2. Add `@FrameworkResource("aws_example_thing", name="Example Thing")` annotation.
```

jar-b (approved):
LGTM 🚀

PR #47328, opened 2026-04-07
generate/listpages: Extract AST manipulation code

ewbankkit:
## Rollback Plan

### Description

Make the `listpages` generator more like the other generators (e.g. `tags`).
Extract the AST manipulation code to `common` for use elsewhere (https://github.com/hashicorp/terraform-provider-aws/pull/47035).


internal/generate/listpages/main.go
```diff
@@ -61,212 +61,117 @@ func main() {
...
-func (g *Generator) addPackage(pkg *packages.Package) {
-	g.pkg = &Package{
-		name:  pkg.Name,
-		files: make([]*PackageFile, len(pkg.Syntax)),
-	}
-
-	for i, file := range pkg.Syntax {
-		g.pkg.files[i] = &PackageFile{
-			file: file,
+	functions := strings.Split(*listOps, ",")
```
YakDriver, line 100: Bug: `function` is never checked for `nil` — if `pkg.FindFunction(functionName)` returns `nil`, the next line `function.Name()` will panic.

The `if err != nil` check here is also wrong — `err` at this point refers to the earlier `common.LoadPackage` error (which was already handled), not the result of `FindFunction`. The nil check should be on `function`:

```go
function := pkg.FindFunction(functionName)
if function == nil {
    g.Fatalf("function (%q) not found", functionName)
}
```

internal/generate/common/ast.go
```diff
@@ -0,0 +1,78 @@
...
+	"golang.org/x/tools/go/packages"
+)
+
+type PackageFile struct {
+	file *ast.File
+}
+
+type Package struct {
+	name  string
+	files []*PackageFile
```
YakDriver: The `Package.name` field is stored but never used — `FindFunction` and the accessor methods don't reference it. If it's intended for future use by callers, consider either exporting it (`Name() string`) or removing it until needed.

internal/generate/common/ast.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"fmt"
+	"go/ast"
+
+	tfslices "github.com/hashicorp/terraform-provider-aws/internal/slices"
+	"golang.org/x/tools/go/packages"
+)
+
+type PackageFile struct {
```
YakDriver: The `PackageFile` wrapper struct adds no value — it's a single-field struct whose only field is `*ast.File`. The `FindFunction` method accesses `file.file` directly. Consider just using `[]*ast.File` in `Package.files` to simplify.

YakDriver (changes requested):
Overall this is a clean refactor — good use of `common.Generator`, the template moves make sense, and the `TemplateData` struct is a nice improvement over the old `FuncSpec`/`HeaderInfo` split. One bug needs fixing before merge.

YakDriver (approved):
LGTM 🎉

Issue #47003, opened 2026-03-19
Error: timeout while waiting for plugin to start


sreekanth3107:
### Terraform and AWS Provider Version

```console
terrafrom 1.15
hashicorp/aws v6.37.0 (signed by HashiCorp)\
```

### Affected Resource(s) or Data Source(s)

```terraform
provider "aws" {
  region = "us-east-1"
}

resource "aws_s3_bucket" "example" {
  bucket = "my-tf-test-bucket"

  tags = {
    Name        = "My bucket"
    Environment = "Dev"
  }
}
```

### Expected Behavior

The apply run without error 

### Actual Behavior

After the new version is released we stared seeing an 
`Error: timeout while waiting for plugin to start` while doing terrafrom apply

### Relevant Error/Panic Output

```console

```

### Sample Terraform Configuration

<details open>
<summary>Click to expand configuration</summary>

```hcl

```
</details>

### Steps to Reproduce

Terrafrom apply

### Debug Logging

<details open>
<summary>Click to expand log output</summary>

```console

```
</details>

### GenAI / LLM Assisted Development

n/a

### Would you like to implement a fix?

No

justinretzolk:
Hey @sreekanth3107 👋🏻 Thank you for taking the time to raise this! Are you able to supply debug logging (redacted as necessary) so that whoever picks this up has the information necessary to investigate?

sreekanth3107:
its working fine when i do local build but its failing on ECS container where i can't get more logs

jjasiek191:
We facing similar issue, and I think it's not problem with resources itself, it's problem with memory. 
I rememebr few years ago (months) there was similar issue, and it was problem with memory consumtion.
Our setup: we have multiple provider blocks, in single terrafrom code. 
We have multiple repository, and see that problem, only when we have multiple provider blocks (currentlly 2 of them are affected)

```╷
│ Error: timeout while waiting for plugin to start
│ 
│ 
╵
╷
│ Error: timeout while waiting for plugin to start
│ 
│ 
╵
╷
│ Error: timeout while waiting for plugin to start
│ 
│ 
╵
╷
│ Error: timeout while waiting for plugin to start```

Adding more memory to runner solved problem, But probablly some analyze is needed why sudentlly consuming more mem

**Comment by jjasiek191** [comment]
My analyze, single resource:

```
resource "aws_s3_bucket" "example" {
  bucket = "my-tf-test-bucket"

  tags = {
[...]

PR #46682, opened 2026-02-26
New resource: `aws_wafv2_web_acl_rule`

YakDriver:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

New resource: `aws_wafv2_web_acl_rule`

## How this helps with #17601 and #28331: _Dependencies the wrong way round_

These issues describe a frustrating problem: when you remove an IP set (or rule group) reference from a `aws_wafv2_web_acl`, Terraform tries to delete the IP set *before* updating the Web ACL — and AWS rejects this because the IP set is still in use.

**Why this happens:**

This isn't purely a provider bug. It's a fundamental mismatch between how Terraform Core builds its dependency graph and how AWS WAFv2 works. Core doesn't handle detach-before-delete scenarios.

- Terraform sees that the Web ACL *references* the IP set, so it creates a dependency edge: Web ACL → IP set
- When destroying, Terraform processes dependents before dependencies — normally correct (delete instances before the cluster)
- But WAFv2 requires the *opposite*: detach the IP set from the Web ACL first, *then* delete it

When you remove the IP set reference from config, Terraform no longer sees that dependency edge in the final graph. It happily tries to delete the IP set before the Web ACL update. AWS says no.

**Why `depends_on` doesn't help:**

You might think "just add `depends_on`" — but that creates a cycle. The Web ACL depends on the IP set (for creation), and the IP set would depend on the Web ACL (for deletion). Terraform correctly rejects cycles.

This is a known class of Terraform Core limitation, tracked across several issues:
- hashicorp/terraform#35763,
- hashicorp/terraform#30439
- hashicorp/terraform#37968
- hashicorp/terraform#35649
- hashicorp/terraform#20196

**What `aws_wafv2_web_acl_rule` enables:**

This new resource lets you manage individual rules separately from the Web ACL itself. Instead of:

```hcl
resource "aws_wafv2_web_acl" "example" {
  rule {
    statement {
      ip_set_reference_statement {
        arn = aws_wafv2_ip_set.blocked.arn
      }
    }
  }
}
```

You can now write:

```hcl
resource "aws_wafv2_web_acl" "example" {
  # No inline rules
}

resource "aws_wafv2_web_acl_rule" "block_ips" {
  web_acl_id = aws_wafv2_web_acl.example.id
  # ...
  statement {
    ip_set_reference_statement {
[...]

PR #47335, opened 2026-04-07
resource/aws_wafv2_web_acl_rule: Adds schema element caching

gdavison:
## Rollback Plan

### Description

Caches elements of the `aws_wafv2_web_acl_rule` schema.

Before the change, memory allocation for all Framework resource type schema is

```
BenchmarkFrameworkProviderResourceSchemaInitialization-10      	       3	 386733083 ns/op	420244816 B/op	 6100751 allocs/op
```

After changes in this PR

```
BenchmarkFrameworkProviderResourceSchemaInitialization-10      	       5	 243161533 ns/op	261602288 B/op	 3601241 allocs/op
```

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/issues/47003.
Relates https://github.com/hashicorp/terraform-provider-aws/issues/46682.

```console

...
```

### DIFF
--- a/internal/provider/framework/provider_test.go
+++ b/internal/provider/framework/provider_test.go
@@ -1,0 +1,76 @@
+// Copyright IBM Corp. 2014, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package framework
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-provider-aws/internal/provider/sdkv2"
+)
+
+// go test -bench=. -benchmem -run=Bench -v ./internal/provider/framework
+
+// This logs Initialization an annoying number of times
+// func BenchmarkFrameworkProviderInitialization(b *testing.B) {
+// 	ctx := b.Context()
+// 	primary, err := sdkv2.NewProvider(ctx)
+// 	if err != nil {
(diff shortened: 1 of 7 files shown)


ewbankkit:
There's also a whole lot of memory getting allocated during model validation
https://github.com/hashicorp/terraform-provider-aws/blob/a89890dd2a68f1ecb9957483c1ce508b4656d1e6/internal/framework/data_source_with_model.go#L43-L54
This memory shouldn't be getting retained but it is allocated during initialization.
I added that model validation code whilst working on Enhanced Region Support to quickly verify that a `Region` field was added to all applicable model structs; we may want to revisit whether this is needed any more. Any per-resource methods will fail if the model doesn't match the schema.

internal/provider/framework/provider_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-provider-aws/internal/provider/sdkv2"
+)
+
+// go test -bench=. -benchmem -run=Bench -v ./internal/provider/framework
```
YakDriver, line 14: Would all the commented out stuff here be better suited to documentation somewhere rather than comments? I can see the justification but commented out stuff is messy but may be the lesser of two weevils.

YakDriver (approved):
LGTM

PR #47422, opened 2026-04-13
resource/aws_dynamodb_table: Updates error message in check

gdavison:
## Rollback Plan

### Description

The error message returned by the AWS API has changed. Typically, the error message should not be checked, but in this case, the error code is not specific enough to determine the correct action.

```console
% make testacc PKG=dynamodb TESTS=TestAccDynamoDBTable_Replica_MRSC_TooManyReplicas

--- PASS: TestAccDynamoDBTable_Replica_MRSC_TooManyReplicas (97.81s)
```


jar-b (approved):
LGTM 🚀

PR #47426, opened 2026-04-13
Returns `unique.Handle` from region spec functions

gdavison:
## Rollback Plan

### Description

Returns `unique.Handle` from region spec functions like `ResourceRegionDefault` rather than wrapping the results for every use.


jar-b (dismissed):
LGTM 🚀

jar-b (approved):
LGTM 🚀

PR #47491, opened 2026-04-16
Update Lake formation permissions documentation to remove 'region' th…

MSilva98:
…at is not part of the schema

## Rollback Plan

### Description

Update documentation of the lake formation permissions resource that mentions region as an optional parameter for tables, but that is not allowed by the schema.

### Relations

Closes #0000

### References

https://docs.aws.amazon.com/lake-formation/latest/APIReference/API_TableResource.html
https://docs.aws.amazon.com/lake-formation/latest/APIReference/API_TableWithColumnsResource.html

```console

...
```

### DIFF
--- a/website/docs/r/lakeformation_permissions.html.markdown
+++ b/website/docs/r/lakeformation_permissions.html.markdown
@@ -254,7 +254,6 @@ The following argument is required:
 
 The following arguments are optional:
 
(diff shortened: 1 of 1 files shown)


jar-b (approved):
LGTM 🚀

jar-b:
Looks like our automation may have inadvertently injected `region` into these sub-headings in addition to the root arguments. Thanks for the correction, @MSilva98 ! 👍

PR #47497, opened 2026-04-16
docs/route53_records_exclusive: ChangeResourceRecordSets AWS API limits

oycyc:
### Description

The `aws_route53_records_exclusive` resource batches every diffed change (`CREATE`/`DELETE`/`UPSERT`) from an `apply` into a single `ChangeResourceRecordSets` request. The provider does not split changes across multiple requests, so applies that exceed the AWS per-request limit fail at `apply` time with no automatic retry or chunking.

The AWS limit is 1,000 `ResourceRecord` elements per request, and `UPSERT` actions count each element twice — meaning modifications hit the cap at ~500 records. This was previously undocumented, so users managing larger hosted zones could hit unexpected failures on initial creates or bulk updates.

This PR adds a callout to the resource documentation covering:

- The single-request behavior (no chunking).
- The 1,000 element cap and `UPSERT` double-count rule.
- A link to the AWS Route 53 API request limits documentation.
- Worked examples distinguishing steady-state operation from failure cases (initial bulk create, large `UPSERT` batches).

No code changes.

### References

- [Route 53 API request limits](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests)

N/A — documentation only.


website/docs/r/route53_records_exclusive.html.markdown
justinretzolk: The Registry [doesn't support](https://developer.hashicorp.com/terraform/registry/providers/docs#callouts) multi-paragraph callouts, so we'll need to combine this into one. Below is my suggestion (with some slight wording changes) to accommodate that. If you're so inclined, you can preview using the [Registry doc preview tool ](https://registry.terraform.io/tools/doc-preview)

```suggestion
~> **API request limits:** All record set changes produced by a single `apply` are sent in one `ChangeResourceRecordSets` request. AWS limits a single request to 1,000 `ResourceRecord` elements, however, each element is counted twice when the change action is `UPSERT`. See the [Route 53 API request limits](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests) for details. This resource does not split changes across multiple requests, so an apply that exceeds these limits will fail. The limit applies to the diff against the live hosted zone, **not** the total number of records managed. For example, a resource managing 2,000 records where a single `apply` modifies 10 of them will send 10 `UPSERT` changes (20 elements counted) and succeed. Attempting to create all 2,000 records in the initial `apply` sends 2,000 `CREATE` elements in one request and will be rejected. Similarly, modifying 600 existing records in a single `apply` produces 600 `UPSERT` changes (counted as 1,200 elements) and will be rejected.
```

justinretzolk (changes requested):

website/docs/r/route53_records_exclusive.html.markdown
oycyc: this sounds & lgtm!

oycyc:
```suggestion
```

oycyc:
@justinretzolk  updated PR accordingly, thanks for feedback

ewbankkit:
@oycyc Thanks for the contribution 👏.
I'm going to merge this one and also create a new issue to capture that this is really a bug in the resource implementation; it is the provider's job to work with those API limits, not error out 😄.

ewbankkit (approved):
LGTM :rocket:.

lucasoares:
Anything else missing here for this to be merged?

PR #47547, opened 2026-04-21
service/sns: Add write-only support for platform_credential and platform_principal in aws_sns_platform_application

### LINKED ISSUE #47293: Add `aws_sns_platform_application` platform_credential write only support
### Description

private keys are being written to the platform application. Adding WO support would enhance the security of the resource.

* `aws_sns_platform_application`

```hcl
resource "aws_sns_platform_application" "apns_application" {
  name                = "apns_application"
  platform            = "APNS"
  platform_credential_wo = "<APNS PRIVATE KEY>"
  platform_principal_wo  = "<APNS CERTIFICATE>"
}
```

### References

_No response_

No

baltasarblanco:
## Rollback Plan

**Yes.** This PR introduces `platform_credential_wo` and `platform_principal_wo` as write-only alternatives. Both attributes are marked `Sensitive` and are excluded from standard Terraform outputs, reducing the risk of accidental private key exposure in state files and logs.

### Description

This PR implements write-only credential support for the `aws_sns_platform_application` resource.

**Key changes:**
- **Schema:** Added `platform_credential_wo` and `platform_principal_wo` attributes.
- **Validation:** Implemented `ExactlyOneOf` and `ConflictsWith` to prevent simultaneous use of standard (`platform_credential`) and write-only (`platform_credential_wo`) attributes.
- **Update Logic:** Modified `resourcePlatformApplicationUpdate` to correctly detect and propagate changes when only the new write-only fields are modified.
- **Testing:** Added `TestAccSNSPlatformApplication_GCM_writeOnly` to verify write-only behavior.

### Relations

Closes #47293

### References

- [AWS SNS API - CreatePlatformApplication](https://docs.aws.amazon.com/sns/latest/api/API_CreatePlatformApplication.html)
- [AWS SNS API - SetPlatformApplicationAttributes](https://docs.aws.amazon.com/sns/latest/api/API_SetPlatformApplicationAttributes.html)

```console
% make testacc TESTS=TestAccSNSPlatformApplication_GCM_writeOnly PKG=sns

--- PASS: TestAccSNSPlatformApplication_GCM_writeOnly (12.42s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/sns  12.85s

### DIFF
--- a/.changelog/47547
+++ b/.changelog/47547
@@ -1,0 +1,1 @@
+notes: service/sns: Add write-only credential support for platform applications
--- a/.changelog/47547
+++ b/.changelog/47547
@@ -1,1 +1,7 @@
-notes: service/sns: Add write-only credential support for platform applications
+```release-note:enhancement
+service/sns: Add write-only credential support for platform applications
+```
+
+```release-note:upgrade-note
+service/sns: When migrating an existing `aws_sns_platform_application` resource from `platform_credential` to `platform_credential_wo`, an explicit state move or manual reconfiguration is required.
+```
--- a/internal/service/sns/platform_application.go
+++ b/internal/service/sns/platform_application.go
@@ -73,9 +73,10 @@ var (
 			ForceNew: true,
 		},
 		"platform_credential": {
-			Type:         schema.TypeString,
-			Optional:     true,
-			Sensitive:    true,
+			Type:      schema.TypeString,
+			Optional:  true,
+			Sensitive: true,
+			// ExactlyOneOf is used here because a credential is required by the AWS API.
 			ExactlyOneOf: []string{"platform_credential", "platform_credential_wo"},
 		},
 		"platform_credential_wo": {
@@ -85,9 +86,10 @@ var (
 			ExactlyOneOf: []string{"platform_credential", "platform_credential_wo"},
 		},
 		"platform_principal": {
-			Type:          schema.TypeString,
-			Optional:      true,
-			Sensitive:     true,
+			Type:      schema.TypeString,
+			Optional:  true,
+			Sensitive: true,
+			// ConflictsWith is used here (instead of ExactlyOneOf) because the principal attribute is optional in the AWS API.
 			ConflictsWith: []string{"platform_principal_wo"},
 		},
 		"platform_principal_wo": {
@@ -241,6 +243,10 @@ func resourcePlatformApplicationUpdate(ctx context.Context, d *schema.ResourceDa
 		} else if v, ok := d.GetOk("platform_principal_wo"); ok {
 			attributes[platformApplicationAttributeNamePlatformPrincipal] = v.(string)
 		}
+	}
+
+	if len(attributes) == 0 {
+		return append(diags, resourcePlatformApplicationRead(ctx, d, meta)...)
 	}
 
 	input := &sns.SetPlatformApplicationAttributesInput{
--- a/internal/service/sns/platform_application.go
+++ b/internal/service/sns/platform_application.go
@@ -73,14 +73,28 @@ var (
 			ForceNew: true,
 		},
 		"platform_credential": {
-			Type:      schema.TypeString,
-			Required:  true,
-			Sensitive: true,
+			Type:         schema.TypeString,
+			Optional:     true,
+			Sensitive:    true,
+			ExactlyOneOf: []string{"platform_credential", "platform_credential_wo"},
+		},
+		"platform_credential_wo": {
+			Type:         schema.TypeString,
+			Optional:     true,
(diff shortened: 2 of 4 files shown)


AndrewCharlesHay:
Positive security change — adding write-only variants for `platform_credential` and `platform_principal` means secrets can live in an ephemeral source (`ephemeral.x.value`) and never land in state. This is the right shape for the write-only pattern rolling out across the provider.

A few review notes:

1. **Test refactor typo**: `testAccPlatformApplicationConfig_basicApnsWithTokenCredentials` now takes a `*TestAccPlatformApplication_gcmBasiPlatform` — that type name looks like a typo (`Basi` missing the `c`). Likely a rename artifact; worth a quick scan to confirm the exported type spelling matches what other tests use.

2. **`ExactlyOneOf` vs `ConflictsWith` asymmetry**: `platform_credential` / `platform_credential_wo` uses `ExactlyOneOf` (one must be set, matching the previous `Required: true` shape). But `platform_principal` / `platform_principal_wo` uses `ConflictsWith`, allowing neither to be set — matching the previous `Optional: true` behavior. Correct, but worth a code comment explaining why the two attribute pairs have different constraints; otherwise someone in a future cleanup PR will "normalize" them and break the APNS-token-cert flow.

3. **State migration**: what happens on first apply with an existing resource that has a SHA256 hash in `platform_credential` state from pre-3.0 provider versions? The `isChangeSha256Removal` handling is extended to the `_wo` variants, but a user migrating from `platform_credential` to `platform_credential_wo` will need to do an explicit move or reconfigure. A note in the upgrade guide / CHANGELOG would help.

4. **`WithSkipUpdate` added for both `_wo` variants** — this means the provider won't detect state drift on these fields. That's intentional for write-only (there's no stored value to diff against), but it also means a user who rotates the actual credential *in AWS* won't see Terraform detect the drift and push the new value. For write-only this is unavoidable, but worth documenting: "drift detection does not apply to `_wo` fields; use a `terraform apply` after a credential rotation to push the new value."

5. **Diff on the conditional write path**: the new logic at the end of `resourcePlatformApplicationUpdate`:
   ```go
   if v, ok := d.GetOk("platform_credential"); ok { ... }
   else if v, ok := d.GetOk("platform_credential_wo"); ok { ... }
   ```
   If *both* are unset (which `ExactlyOneOf` should prevent at plan time but is theoretically reachable after a destroy-and-recreate error path), `attributes[platformApplicationAttributeNamePlatformCredential]` never gets set, and the subsequent `SetPlatformApplicationAttributesInput` may send a partial update that clears the credential at AWS. Worth either asserting one-of or adding a guard before the API call.

6. **No new secret in logs check**: both fields are `Sensitive: true`. Good. Consider a CI/lint check that any new attribute matching `/_wo$/` is also `Sensitive: true` — it's the kind of invariant that's easy to forget on future write-only additions.

baltasarblanco:
Hi @AndrewCharlesHay, thanks for the thorough review and the positive feedback on the security approach! I really appreciate the catch on these edge cases.

Here is my plan to address your notes:
1. **Test typo:** Good catch. I'll fix the `*TestAccPlatformApplication_gcmBasicPlatform` typo right away.
2. **Schema comments:** I will add comments explicitly explaining the `ExactlyOneOf` (required) vs `ConflictsWith` (optional) asymmetry so no one breaks the APNS flow in future cleanups.
3. **Update Guard:** I will add a guard before the `SetPlatformApplicationAttributes` API call to ensure we don't accidentally send an empty credential update.
4. **Docs & State Migration:** I will update the website documentation to clearly state that `_wo` fields do not detect drift and require a `terraform apply` after rotation, plus a note for users migrating from the standard credential.
5. **CI/Lint:** Fully agree on the CI check for `/_wo$/` attributes being `Sensitive: true`. It would be a great addition for a future provider-wide linting PR!

I'll push these code changes and the documentation updates shortly. Thanks again!

baltasarblanco:
@AndrewCharlesHay all requested changes have been pushed! 

* Fixed the typo in the acceptance test struct.
* Added schema comments explaining the `ExactlyOneOf` vs `ConflictsWith` rationale.
* Added a safety guard in the update logic to prevent empty credential API calls.
* Updated the docs to clarify drift detection, and added the requested state migration note to the CHANGELOG.

**Minor doc note:** I updated the `platform_credential` description from `(Required)` to `(Optional)` to accurately reflect the new `ExactlyOneOf` behavior.

Let me know if you need anything else to get this over the finish line! 🚀

AndrewCharlesHay (approved):
Changes look great! Thanks for quickly addressing the review notes, @baltasarblanco. LGTM! 🚀

PR #47586, opened 2026-04-23
docs(ssm_service_setting): Fix setting_id in example usage

handlerbot:
The example usage showed a full ARN for setting_id, but the resource expects a short-form setting ID path like /ssm/parameter-store/high-throughput-enabled.


jar-b (approved):
LGTM 🚀

jar-b:
Thank you for the correction, @handlerbot! 👍

PR #47700, opened 2026-04-29
elasticache: retry tagging on UserNotFoundFault / UserGroupNotFoundFault

### LINKED ISSUE #47638: [Bug]: aws_elasticache_user — UserNotFound on AddTagsToResource (Write path race, sibling to #34002)

```console
Terraform v1.14.7
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v6.18.0
```

* `aws_elasticache_user`

### Expected Behavior

When `default_tags` adds new tag keys to several `aws_elasticache_user` resources in the same plan, the apply should succeed. The provider should wait for each user to leave the transitioning state before issuing the next tagging call, the same way the Read path does after #34396.

### Actual Behavior

`AddTagsToResource` returns `404 UserNotFound: <user> is not available for tagging` against users that exist and are reported as `Status: active` by `describe-users`. After a successful tag op, the user enters a brief internal transitioning window during which any subsequent tag op on the same (or a sibling) user gets the 404. The provider doesn't retry or wait for that window to close.

#34002 / #34396 fixed this exact race, but only on the Read path: the fix description is explicit ("Fix `UserNotFound: ... is not available for tagging` errors on resource Read when there is a concurrent update to the user **by waiting for the resource to be in active state**"). The Write path has no equivalent waiter, so it still trips on plain tag updates.

```console
Error: updating tags for ElastiCache User (arn:aws:elasticache:eu-west-1:<account>:user:tagrepro-a):
  operation error ElastiCache: AddTagsToResource,
  https response error StatusCode: 404,
  RequestID: 9413ef79-31b1-43bc-xxxx-xxxxxxxxxxxx,
  UserNotFound: tagrepro-a is not available for tagging
```

In the apply where I hit this, 15 users across 3 namespaces failed the same way, all `active` in the AWS console.

<details open>

```hcl
provider "aws" {
  region = "eu-west-1"

  default_tags {
    tags = {
      Compliance      = "SOC2-Type2 ISO27001"
      ComplianceScope = "iso27001"
      OwnerTeam       = "platform"
    }
  }
}

resource "aws_elasticache_user_group" "this" {
  user_group_id = "tagrepro-grp"
  engine        = "valkey"
  user_ids      = ["default"]
}

resource "aws_elasticache_user" "this" {
  for_each = toset(["a", "b", "c", "d", "e"])

  user_id       = "tagrepro-${each.key}"
  user_name     = "tagrepro-${each.key}"
  engine        = "valkey"
  access_string = "on ~tagrepro-${each.key}:* +@all"

  authentication_mode {
    type = "iam"
  }

  tags = {
    repro = "true"
  }
}
```

</details>

### Steps to Reproduce

1. Apply the config above without the `default_tags` block. The 5 users get created cleanly.
2. Add the `default_tags` block (or change any key/value in it).
3. Apply again.

The provider fires `AddTagsToResource` against each user back to back. The first one or two land, the rest fail with 404 `UserNotFound`.

It also reproduces deterministically against a single user with two consecutive AWS CLI calls:

```console
$ aws elasticache add-tags-to-resource --resource-name <arn> --tags Key=Probe,Value=Y
{ ...success... }

$ aws elasticache remove-tags-from-resource --resource-name <arn> --tag-keys Probe
An error occurred (UserNotFound) when calling the RemoveTagsFromResource operation:
  <user> is not available for tagging
```

So the underlying race is in the AWS API, not Terraform. The provider just needs the same waiter on the Write path that #34396 added on the Read path.

* Same root cause as #34002 / #34396, but the v5.26.0 fix only handles the Read path.
* Hits reliably as soon as several users get tagged in one apply cycle (e.g. a new `default_tags` rollout). Single-user updates flake more rarely.
* User `Status` reads `active` from `describe-users` while the API still rejects further tag ops, so a state-based wait against the public field isn't enough on its own. The provider may need to retry on `UserNotFound` from `Add/RemoveTagsToResource` for a short window, or serialize tag ops per cluster.
* My current workaround is `lifecycle { ignore_changes = [tags_all] }` on the affected resources, which means I can't push my compliance tags through Terraform until this is fixed.

Used an LLM to help search related issues and put the repro together.

No

Keralin:
Fixes #47638.

For a tag-only update the framework fires `RemoveTagsFromResource` and `AddTagsToResource` back-to-back from `tagsResourceCRUDInterceptor.Before(Update)`, before `resourceUserUpdate` ever runs. The first call leaves the user in `modifying` for ~5s, the second hits during that window and returns `UserNotFoundFault` with "not available for tagging". Same race for `aws_elasticache_user_group`. Full ordering trace in #47638.

The retry inside `tags_gen.go:updateTags` already handles the equivalent `InvalidReplicationGroupStateFault` + "not in available state" race for replication groups. This PR extends the generator so the same mechanism covers users and user_groups.

## Changes

- **`internal/tfresource/retry.go`** — adds `RetryWhenIsOneOf{2,3}ErrorMessageContains`, mirroring the existing type-only `RetryWhenIsOneOf{2,3,4}` helpers.
- **`internal/generate/tags/main.go` + 3 templates** — `-RetryErrorCode` / `-RetryErrorMessage` are now repeatable, paired by index. Templates branch on `len(.RetryConditions)` and emit the matching helper. Single-pair invocations regenerate byte-identically — verified by running `go generate ./internal/service/...` and checking that only `internal/service/elasticache/tags_gen.go` changes.
- **`internal/service/elasticache/generate.go`** — directive now declares retries for `UserNotFoundFault` and `UserGroupNotFoundFault` alongside the existing `InvalidReplicationGroupStateFault` one. The 3 retry sites in `tags_gen.go` regenerate to use `RetryWhenIsOneOf3ErrorMessageContains`.

## Verification

**Unit tests** (`internal/tfresource/retry_test.go`): added 11 sub-tests covering both new helpers — happy path, non-retryable error, retries on each typed error, type-matches-but-needle-doesn't, and needle-paired-with-its-own-type-only. All pass deterministically without AWS.

**End-to-end against the original failing scenario**: built this branch as a local provider via `dev_overrides` and ran it against a development environment with 5 `aws_elasticache_user` resources.

- Apply 1 (additive baseline): added a probe tag on all 5 users — `Add` only, no race expected. Succeeded normally.
- Apply 2 (race trigger): replaced the probe tag's KEY (forces `RemoveTagsFromResource` + `AddTagsToResource` consecutively per user, in parallel across 5 users — same operational shape that originally hit us with 15 users in prod). All 5 users completed in `~7m25s` each — that timing is the retry helper waiting through the modifying state and re-attempting. **Zero `UserNotFoundFault` errors escaped to the caller.**
- Apply 3 (cleanup): removed the probe tag. Succeeded normally.

Compliance tags preserved across all three applies.

CLI repro (no Terraform needed):

```
$ aws elasticache add-tags-to-resource --resource-name <user-arn> --tags Key=Probe,Value=A
200 OK
$ aws elasticache remove-tags-from-resource --resource-name <user-arn> --tag-keys Probe
UserNotFound: <user-id> is not available for tagging
```

Originally hit during a `default_tags` rollout firing 7-8 tags simultaneously across 15 ElastiCache users in production. Local workaround was `ignore_changes = [tags_all]`, which conflicts with compliance/default_tags policy and is what motivated coming upstream.

## AI usage

Drafted with Claude (LLM assistant) under human supervision. The retry helper, generator extension, and template branching were AI-generated then reviewed by the author. Unit tests were drafted by the LLM and run manually before submission.


Keralin:
Any new on this? @justinretzolk

PR #47728, opened 2026-05-01
Simplify route_table_id assignment in markdown

templier2:
### Description

As far as data.aws_route_tables.this.ids returns an object with type "list", there is no need to use tolist() for it.


justinretzolk (approved):
Looks good to me! Thank you for taking the time to improve the docs! I'll get this merged once the checks finish up

PR #47750, opened 2026-05-04
Provider: Removes run-time validation for Framework-based resource types

gdavison:
## Rollback Plan

### Description

Removes run-time validation for Framework-based resource types and adds validation step to Makefile directives.

Will reduce memory usage when initializing the provider. Only the schemas for resource types that are used in the configuration should be allocated.

Relates: #47003

| Before |   58 | 17366403 ns/op | 15072518 B/op | 167863 allocs/op |
| After  | 6097 |   185166 ns/op |   244381 B/op |   2420 allocs/op |

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -925,7 +925,7 @@ testacc: prereq-go fmt-check testacc-schema ## Run acceptance tests
 		echo "See the contributing guide for more information: https://hashicorp.github.io/terraform-provider-aws/running-and-writing-acceptance-tests"; \
 		exit 1; \
 	fi
-	TF_ACC=1 $(GO_VER) test ./$(PKG_NAME)/... -v -count $(TEST_COUNT) -parallel $(ACCTEST_PARALLELISM) $(RUNARGS) $(TESTARGS) -timeout $(ACCTEST_TIMEOUT) -vet=off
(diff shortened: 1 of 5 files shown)


GNUmakefile
ewbankkit, line 1: Plus add new target to `docs/makefile-cheat-sheet.md`.

GNUmakefile
```diff
@@ -912,7 +912,7 @@ test-shard: prereq-go ## Run unit tests for a specific shard (CI only: SHARD=0 T
 test-naming: ## Check test naming conventions
 	@.ci/scripts/check-test-naming.sh
 
-testacc: prereq-go fmt-check ## Run acceptance tests
+testacc: prereq-go fmt-check testacc-schema ## Run acceptance tests
```
jar-b: Should we add this to the `test` target so that it runs in CI?

gdavison: This should be picked up by the `test` target, since it runs all of the packages

jar-b: Ah, yes. The `testacc` prefix took my brain in the wrong direction, thanks for the clarification.

jar-b (approved):
LGTM 🚀

PR #47758, opened 2026-05-05
Adds Semgrep rule to convert from `Schema` to `SchemaFunc`

gdavison:
## Rollback Plan

### Description

Adds Semgrep rule to convert from `Schema` to `SchemaFunc`


.ci/semgrep/pluginsdk/schema.fixed.go
jar-b, line 1: TIL `.fixed`. 👍

jar-b (approved):
LGTM 🚀

ewbankkit:
Is there any way we can prove (`terraform providers -json`?) that the before and after resulting schemas are identical?

gdavison:
We can do something like:

Get the initial schema:

```console
echo 'data "aws_partition" "example" {}' > example.tf
mkdir -p terraform-providers-schema
make go-build
rm -f .terraform.lock.hcl
terraform init -plugin-dir terraform-plugin-dir
terraform providers schema -json | jq . > terraform-providers-schema/schema.original.json
```

Get the updated schema:

```console
make go-build
rm -f .terraform.lock.hcl
terraform init -plugin-dir terraform-plugin-dir
terraform providers schema -json | jq . > terraform-providers-schema/schema.new.json
```

And then:

```console
diff terraform-providers-schema/schema.original.json terraform-providers-schema/schema.new.json
```

PR #47052, opened 2026-03-23
[bugfix] allow for variables and conditional expressions in db_parameter_group_identifier field of aws_timestreaminfluxdb_db_cluster

akostadinov-nasuni:
## Rollback Plan

N/A

## Changes to Security Controls

N/A

### Description
This is a proposed fix for issue [46436](https://github.com/hashicorp/terraform-provider-aws/issues/46436): V3 parameter group validation fails when db_parameter_group_identifier is set via variable or conditional. The root cause of the issue was that parameter group validation was happening in the `ValidateConfig` phase, which runs before variables are fully resolved and the AWS client is properly initialized for API calls. 

The fix was to move parameter group V3 validation from ValidateConfig to ModifyPlan (where the variables and conditional statements are fully resolved). 

### Relations

Closes #46436.
Closes https://github.com/hashicorp/terraform-provider-aws/issues/45330.
Closes https://github.com/hashicorp/terraform-provider-aws/pull/45782.

PR #47768, opened 2026-05-05
generate/identitytests: Correct PlanCheck for single identity attribute when `plannableImportAction="Replace"`

ewbankkit:
## Rollback Plan

### Description

Noticed while adding RI to Timestream InfluxDB (#47052). If an import step results in resource replacement then the identity attribute will plan to `Unknown`.
Restricted to the single identity attribute case as for many multi-attribute parameterized identity resource, not all identity attributes will be unknown.


jar-b (approved):
LGTM 🚀

PR #47744, opened 2026-05-04
Decision Doc: No Import Support For Exclusive Resources

taruntej-a:
### Description

Adds design doc - No Import Support For Exclusive Resources after considering the risks associated with importing exclusive resources.

### Relations
Closes https://github.com/hashicorp/terraform-provider-aws/pull/47620

PR #47772, opened 2026-05-05
Adds `@ExclusiveResource` annotation

taruntej-a:
### Description

This PR adds the `@ExclusiveResource` annotation to help prevent contributors from adding import or list support to exclusive resources, which can be risky.

Once this PR is merged, we can add the `@ExclusiveResource` annotation to all "exclusive‑style" resources.

This annotation permits identity annotations but blocks import and list support at generation time.

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/47744

#### Import

```console
go generate internal/service/iam/generate.go
Generating internal/service/iam/list_pages_gen.go
  iam.listGroupsForUser
  iam.listServiceSpecificCredentials
Generating internal/service/iam/tags_gen.go
Generating internal/service/iam/service_package_gen.go
iam.resourcePolicyAttachment: @ExclusiveResource must not have import annotations (@WrappedImport, @CustomImport, @ImportIDHandler). Exclusive resources do not support import. See docs/design-decisions/no-import-support-for-exclusive-resource-types.md
exit status 1
internal/service/iam/generate.go:6: running "go": exit status 1
```

#### List

```console
go generate generate.go
Generating internal/service/iam/list_pages_gen.go
  iam.listGroupsForUser
  iam.listServiceSpecificCredentials
Generating internal/service/iam/tags_gen.go
Generating internal/service/iam/service_package_gen.go
SDK List Resource "aws_iam_policy_attachment" cannot be created for an @ExclusiveResource. Exclusive resources must not support list. See docs/design-decisions/no-import-support-for-exclusive-resource-types.md
exit status 1
generate.go:6: running "go": exit status 1
```

### DIFF
--- a/internal/generate/identitytests/main.go
+++ b/internal/generate/identitytests/main.go
@@ -605,7 +605,7 @@ func (v *visitor) processFuncDecl(funcDecl *ast.FuncDecl) {
 					}
 				}
 
-			case "NoImport":
+			case "NoImport", "ExclusiveResource":
 				d.NoImport = true
 
 			case "Testing":
--- a/internal/generate/servicepackage/main.go
+++ b/internal/generate/servicepackage/main.go
@@ -103,6 +103,9 @@ func main() {
 
 		for key, value := range v.frameworkListResources {
 			if val, exists := v.frameworkResources[key]; exists {
+				if val.exclusiveResource {
+					g.Fatalf("Framework List Resource %q cannot be created for an @ExclusiveResource. Exclusive resources must not support list. See docs/design-decisions/no-import-support-for-exclusive-resource-types.md", key)
+				}
 				value.Name = val.Name
 				value.ResourceIdentity = val.ResourceIdentity
 				value.TransparentTagging = val.TransparentTagging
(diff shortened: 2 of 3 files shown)


internal/generate/identitytests/main.go
```diff
@@ -605,7 +605,7 @@ func (v *visitor) processFuncDecl(funcDecl *ast.FuncDecl) {
 					}
 				}
 
-			case "NoImport":
+			case "NoImport", "ExclusiveResource":
```
taruntej-a, line 608: We have not discussed the annotation naming yet, `@ExclusiveResource` is just a reasonable starting point.  If it seems unclear, we can explore alternative naming.

internal/generate/servicepackage/main.go
```diff
@@ -103,6 +103,9 @@ func main() {
 
 		for key, value := range v.frameworkListResources {
 			if val, exists := v.frameworkResources[key]; exists {
```
taruntej-a, line 105: While implementing this, I did not fully consider whether identity should also be restricted. Since RI is a prerequisite for list, blocking identity would have implicitly blocked list as well but my understanding was to explicitly block only import/list support, while still allowing RI. That is the reason for having this explicit handling for "list annotations" and  "import annotations".

taruntej-a:
Another approach preferred.

Approach : We will add a source file with the standard List Resource naming, <resource>_list.go, which will contain only a comment stating why the List Resource will not be implemented. This will make the exclusion highly-visible. 

Closing this.

Issue #29014, opened 2023-01-20
[New Resource]: aws_inspector2_update_configuration Add support for updating inspector configuration


Luke1298:
### Description

Simliar to 'aws_inspector2_organization_configuration' which allows you to change the 'autoenable' status of various resources this would allow you to set the rescan duration in inspector2. 

### Requested Resource(s) and/or Data Source(s)

* aws_inspector2_update_configuration for allowing one to update the rescan duration. (See [boto3 inspector 2 client](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/inspector2.html#Inspector2.Client.update_configuration)

### References

Relates https://github.com/hashicorp/terraform-provider-aws/pull/27000
https://github.com/hashicorp/terraform-provider-aws/issues/22330

### Would you like to implement a fix?

None

calvinbui:
not stale

b-dean:
Should it maybe be `aws_inspector2_configuration` not `update_configuration`? After all the [boto3 api for inspector organization configuration](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/inspector2/client/update_organization_configuration.html) is also `update_organization_configuration` but the terraform resource is `aws_inspector2_organization_configuration`

nomeelnoj:
It has been almost 3 years since this request was opened.  Is it possible to get an update? We do not like having to run Terraform and then follow up in the UI to make such a simple change.  AWS also just released support for new methods of scanning (based on image last in-use date) and it would be helpful to be able to make changes if/when they update again in the future. Thanks!

PR #47826, opened 2026-05-08
r/aws_inspector2_configuration: new resource

### LINKED ISSUE #47824: [New Resource]: aws_inspector2_configuration (EC2 scan mode and ECR rescan duration)
### Description

Amazon Inspector v2 supports two EC2 scan modes per region: `EC2_SSM_AGENT_BASED` (default for new accounts) and `EC2_HYBRID` (agent-based plus EBS-snapshot-based agentless scanning for instances without SSM agent).

This is configured at the delegated administrator level via the [`UpdateConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_UpdateConfiguration.html) API, which sets `ec2Configuration.scanMode`.

The current `aws_inspector2_organization_configuration` resource only exposes `auto_enable` (ec2 / ecr / lambda / code_repository / lambda_code) and the read-only `max_account_limit_reached`. There is no Terraform resource or argument that maps to `UpdateConfiguration` / `ec2Configuration.scanMode`, so this setting cannot be managed via Terraform today.

The same API also exposes `ecrConfiguration` (`rescanDuration`, `pullDateRescanDuration`, `pullDateRescanMode`) which is similarly missing.

- `aws_inspector2_organization_configuration` (extend) — most natural home, since `UpdateConfiguration` is also a delegated-admin-level call.

  *or*

- `aws_inspector2_configuration` (new) — a dedicated resource if the team prefers to keep `UpdateConfiguration` separate from `UpdateOrganizationConfiguration`.

Option 1 — extend `aws_inspector2_organization_configuration`:

```terraform
resource "aws_inspector2_organization_configuration" "this" {
  auto_enable {
    ec2    = true
    ecr    = true
    lambda = true
  }

  ec2_configuration {
    scan_mode = "EC2_HYBRID" # or EC2_SSM_AGENT_BASED
  }

  ecr_configuration {
    rescan_duration             = "DAYS_14"
    pull_date_rescan_duration   = "DAYS_14"
    pull_date_rescan_mode       = "LAST_IN_USE_AT"
  }
}
```

Option 2 — new dedicated resource `aws_inspector2_configuration`:

```terraform
resource "aws_inspector2_configuration" "this" {
  ec2_configuration {
    scan_mode = "EC2_HYBRID"
  }

  ecr_configuration {
    rescan_duration           = "DAYS_14"
    pull_date_rescan_duration = "DAYS_14"
    pull_date_rescan_mode     = "LAST_IN_USE_AT"
  }
}
```

### References

- AWS API: [`UpdateConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_UpdateConfiguration.html)
- AWS API: [`GetConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_GetConfiguration.html)
- AWS docs: [Hybrid scanning for Amazon EC2](https://docs.aws.amazon.com/inspector/latest/user/scanning-ec2.html#ec2-scan-modes)
- Existing resource: [`aws_inspector2_organization_configuration`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/inspector2_organization_configuration)

### Why this matters

Without provider support, organizations end up with one of:
1. ClickOps for the scan mode toggle (silent drift, no audit trail).
2. `null_resource` / `terraform_data` + `local-exec` calling the AWS CLI (fragile, no drift detection, requires `aws` + `jq` on the runner, requires extra IAM dance for cross-account assume-role).
3. Custom Lambda + `aws_lambda_invocation` (significant boilerplate for a one-line config).

A first-class resource would let teams manage Inspector configuration alongside the rest of their security baseline (Security Hub, GuardDuty, Macie — all of which have full provider support).

Open to discussing — happy to provide additional context or test against a draft PR.

Cesarsk:
## Rollback Plan

This PR adds a new managed resource for an Amazon Inspector v2 configuration API. It does not change any existing security controls — it exposes the existing `UpdateConfiguration`/`GetConfiguration` APIs, which control EC2 scan mode and ECR re-scan duration. Both are settings that previously could only be managed through the AWS console or CLI; this PR makes them manageable as code.

### Description

Adds a new resource `aws_inspector2_configuration` that wraps the Inspector v2 [`UpdateConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_UpdateConfiguration.html) and [`GetConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_GetConfiguration.html) APIs.

Two configuration domains are exposed:

- **EC2 scan mode** (`ec2_configuration.scan_mode`): `EC2_HYBRID` or `EC2_SSM_AGENT_BASED`. When called by the Inspector delegated administrator, applies org-wide for the region.
- **ECR re-scan settings** (`ecr_configuration`): `rescan_duration`, `pull_date_rescan_duration`, `pull_date_rescan_mode`.

Neither setting was previously expressible in Terraform — `aws_inspector2_organization_configuration` only manages `auto_enable` (a different API: `UpdateOrganizationConfiguration`).

The resource is regional, follows the existing SDKv2 patterns in `internal/service/inspector2/` (mutex-serialized writes, `RetryUntilEqual` poll on the read API), and has no `Delete` API on the AWS side — `destroy` resets to AWS defaults (`EC2_HYBRID`, ECR `LIFETIME` with `LAST_IN_USE_AT`).

### Relations

Closes #47824

### References

- AWS API — [`UpdateConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_UpdateConfiguration.html)
- AWS API — [`GetConfiguration`](https://docs.aws.amazon.com/inspector/v2/APIReference/API_GetConfiguration.html)
- AWS Go SDK v2 — [`inspector2.UpdateConfigurationInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/inspector2#UpdateConfigurationInput)
- AWS docs — [Managed scanning of EC2 instances](https://docs.aws.amazon.com/inspector/latest/user/scanning-ec2.html#choosing-scan-mode)

Run from a power-user role in a standalone AWS account (no Inspector delegated-admin association). All five Configuration sub-tests pass.

```console
% TF_ACC=1 go test -v -timeout=60m \
    -run='TestAccInspector2_serial/Configuration$|TestAccInspector2_serial/Configuration/' \
    ./internal/service/inspector2/

=== RUN   TestAccInspector2_serial/Configuration
=== RUN   TestAccInspector2_serial/Configuration/basic
=== RUN   TestAccInspector2_serial/Configuration/disappears
=== RUN   TestAccInspector2_serial/Configuration/ec2ScanMode
=== RUN   TestAccInspector2_serial/Configuration/ecrRescan
=== RUN   TestAccInspector2_serial/Configuration/combined
    --- PASS: TestAccInspector2_serial/Configuration (133.20s)
        --- PASS: TestAccInspector2_serial/Configuration/basic (20.68s)
        --- PASS: TestAccInspector2_serial/Configuration/disappears (31.53s)
        --- PASS: TestAccInspector2_serial/Configuration/ec2ScanMode (24.25s)
        --- PASS: TestAccInspector2_serial/Configuration/ecrRescan (39.09s)
        --- PASS: TestAccInspector2_serial/Configuration/combined (17.65s)
```

The pre-existing `OrganizationConfiguration` sub-tests in the same serial wrapper fail in this environment because the test account already has a different account configured as Inspector delegated administrator (`ConflictException: Delegated administrator ... is already enabled`). Those failures are unrelated to this change.

### DIFF
--- a/.changelog/47824.txt
+++ b/.changelog/47824.txt
@@ -1,0 +1,3 @@
+```release-note:new-resource
+aws_inspector2_configuration
+```
--- a/internal/service/inspector2/configuration.go
+++ b/internal/service/inspector2/configuration.go
@@ -1,0 +1,380 @@
+// Copyright IBM Corp. 2014, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+// DONOTCOPY: Copying old resources spreads bad habits. Use skaff instead.
+
+package inspector2
+
+import (
+	"context"
+	"errors"
+	"log"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/inspector2"
+	awstypes "github.com/aws/aws-sdk-go-v2/service/inspector2/types"
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-aws/internal/conns"
+	"github.com/hashicorp/terraform-provider-aws/internal/enum"
+	"github.com/hashicorp/terraform-provider-aws/internal/errs/sdkdiag"
+	"github.com/hashicorp/terraform-provider-aws/internal/retry"
+	"github.com/hashicorp/terraform-provider-aws/internal/tfresource"
+)
+
+// @SDKResource("aws_inspector2_configuration", name="Configuration")
+func resourceConfiguration() *schema.Resource {
+	return &schema.Resource{
+		CreateWithoutTimeout: resourceConfigurationCreate,
+		ReadWithoutTimeout:   resourceConfigurationRead,
+		UpdateWithoutTimeout: resourceConfigurationUpdate,
+		DeleteWithoutTimeout: resourceConfigurationDelete,
+
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(5 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"ec2_configuration": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"scan_mode": {
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: enum.Validate[awstypes.Ec2ScanMode](),
+						},
+						"vm_scanner": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Computed: true,
+						},
+					},
+				},
+			},
+			"ecr_configuration": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"rescan_duration": {
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: enum.Validate[awstypes.EcrRescanDuration](),
+						},
+						"pull_date_rescan_duration": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							Computed:         true,
+							ValidateDiagFunc: enum.Validate[awstypes.EcrPullDateRescanDuration](),
+						},
+						"pull_date_rescan_mode": {
+							Type:             schema.TypeString,
+							Optional:         true,
+							Computed:         true,
+							ValidateDiagFunc: enum.Validate[awstypes.EcrPullDateRescanMode](),
+						},
+					},
+				},
+			},
+		},
+
+		// At least one configuration block must be provided. UpdateConfiguration
+		// fails when both ec2_configuration and ecr_configuration are nil.
+		CustomizeDiff: func(_ context.Context, diff *schema.ResourceDiff, _ any) error {
+			ec2 := diff.Get("ec2_configuration").([]any)
+			ecr := diff.Get("ecr_configuration").([]any)
+			if len(ec2) == 0 && len(ecr) == 0 {
+				return errConfigurationRequired
+			}
+			return nil
+		},
+	}
+}
+
+// errConfigurationRequired surfaces the requirement that at least one of
+// ec2_configuration or ecr_configuration is set.
+var errConfigurationRequired = errors.New("at least one of ec2_configuration or ecr_configuration must be set")
+
+func resourceConfigurationCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	d.SetId(meta.(*conns.AWSClient).AccountID(ctx))
+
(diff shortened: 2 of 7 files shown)


Cesarsk:
Hello, any news about this? Thanks!

akuzminsky:
Thanks for putting this together — this is the only route I've found to managing `ec2Configuration` from Terraform, and the `UpdateConfiguration`-only gap has been open a long time (#29014 since 2023).

Two questions, both about `activateVMScanner`.

### 1. Would you consider adding `vm_scanner` to `ec2_configuration`?

`Ec2Configuration` gained `activateVMScanner` when the Inspector VM Scanner launched (2026-05-31), and `GetConfiguration` reads it back as `ec2Configuration.vmScannerState.{activated,status}`. The `service/inspector2` version already vendored in this repo exposes `ActivateVMScanner *bool`, so it looks like a small addition to the schema you have:

```hcl
ec2_configuration {
  scan_mode  = "EC2_HYBRID"
  vm_scanner = true
}
```

As it stands the resource manages one of the two fields on that object, which leaves the other one un-codified for anyone who has activated the scanner.

### 2. What does an apply do to an already-activated VM scanner?

Create/update build `&awstypes.Ec2Configuration{ScanMode: ...}` with `ActivateVMScanner` left nil. I couldn't find anything in the API reference stating whether `UpdateConfiguration` treats an omitted `activateVMScanner` as "leave as-is" or "set to false", and I don't see a test covering it.

If it's the latter, applying this resource against an account that has the scanner on would deactivate it — a silent reduction in scan coverage with nothing in the plan to indicate it. Did you get a chance to test that case? If it does need a fix, exposing the field in (1) would also give people a way to assert it explicitly rather than relying on omission semantics.

Happy to help with either if that's useful.

Cesarsk:
@akuzminsky 
Thanks for the thorough review. Both points were well taken, and (2) was a real gap.

**1. `vm_scanner` added.**

```hcl
ec2_configuration {
  scan_mode  = "EC2_HYBRID"
  vm_scanner = true
}
```

It is Optional + Computed, so existing configurations keep working and the current activation state is reflected in state. This required rebasing onto latest `main`: the field only exists since the `service/inspector2` bump to v1.54.0.

**2. Omission semantics.** I could not find any documentation on whether an omitted `activateVMScanner` means "leave as-is" or "deactivate" either, so the resource no longer relies on it in either direction: when `vm_scanner` is unset, it reads the current configuration and echoes `vmScannerState.activated` back in the `UpdateConfiguration` call. An apply without `vm_scanner` is therefore guaranteed to preserve the activation state regardless of how the API treats a nil field. The `ec2VMScanner` acceptance test covers exactly the case you raised (activate, apply without `vm_scanner`, assert still activated) plus the full activate/deactivate lifecycle.

One consequence worth noting: `destroy` now also deactivates the scanner, consistent with the resource's documented "reset to AWS defaults" behavior (newly onboarded accounts have it off).

Cesarsk:
Any news here? @akuzminsky

akuzminsky:
@Cesarsk  sorry for going quiet. Thanks for addressing both points.
I went through the updated diff and both concerns went away:

1. `vm_scanner`. Optional + Computed is the right shape; existing configs are unaffected and the activation state shows up in state.
2. Omission semantics. reading `vmScannerState.activated` and echoing it back when `vm_scanner` is unset (checked via raw config, so the Computed value in prior state doesn't mask it) sidesteps the undocumented nil behavior entirely. The `ec2VMScanner` test's middle step - activate, re-apply without the argument, assert still activated - is exactly the case I was worried about.

The `destroy` -> scanner-deactivated behavior makes sense given there's no Delete API, and it's already called out in the resource docs, so no concerns there.
LGTM from a user's perspective (I'm not a maintainer, so this is just a +1 on the design). We manage Inspector as the org delegated admin and currently have to set `scanMode/activateVMScanner` out of band, so we'd adopt this as soon as it lands.

PR #47831, opened 2026-05-08
EC2 schema caching: Updates schema sharing for `aws_instance` and `aws_spot_instance_request`

gdavison:
## Rollback Plan

### Description

Updates schema sharing for aws_instance` and `aws_spot_instance_request`. Schemas are no longer retained if the resource types are not used.

Benchmark results

| Before | 124 | 9553866 ns/op | 18,732,854 B/op | 105,952 allocs/op |
| After  | 124 | 9462079 ns/op | 18,502,236 B/op | 104,871 allocs/op |


jar-b (approved):
LGTM 🚀

PR #47845, opened 2026-05-10
Technical debt: Add and fix semgrep rule `list-iterator-calls-stopped-yield`

### LINKED ISSUE #47844: Technical debt: List iterators may call `yield` after stop
Take a look at `listAPIs`:https://github.com/hashicorp/terraform-provider-aws/blob/d568e6abbacfd12be320634621dea6c68c2099c5/internal/service/apigatewayv2/api_list.go#L96-L113
If the call to `yield` made while iterating through `page.Items` returns `false` then the outer `yield` called when `err != nil` will still be called (when `err != nil`). This could be problematic.

This is only occurring when the list iterator function is used a paged finder function with a callback.

ewbankkit:
## Rollback Plan

### Description

Ensure that a stopped `yield` function isn't called when a list iterator uses a callback-style paged finder.

### Relations

Closes https://github.com/hashicorp/terraform-provider-aws/issues/47844.

```console

...
```

### DIFF
--- a/.ci/semgrep/list/callback-pagination.yaml
+++ b/.ci/semgrep/list/callback-pagination.yaml
@@ -1,0 +1,19 @@
+rules:
+  - id: list-iterator-calls-stopped-yield
+    message: >-
+      List iterator function uses callback-based paginated finder (lastPage pattern)
+      without ensuring that a stopped yield isn't called on error.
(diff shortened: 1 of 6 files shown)


jar-b (approved):
LGTM 🚀

PR #47891, opened 2026-05-13
Schema caching: Caches `enum` validators

gdavison:
## Rollback Plan

### Description

Caches SDKv2 `enum` validators and results of `enum.Values` conversion to `string`

| Before | 126 | 9421101 ns/op | 18,502,251 B/op | 104,871 allocs/op |
| After  | 134 | 8927013 ns/op | 17,803,273 B/op |  95,370 allocs/op |

### Relations

Relates #47003

### DIFF
--- a/internal/enum/validate.go
+++ b/internal/enum/validate.go
@@ -4,10 +4,13 @@
 package enum
 
 import (
+	"reflect"
+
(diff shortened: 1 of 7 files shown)


jar-b (approved):
LGTM 🚀

PR #47904, opened 2026-05-14
r/aws_db_instance: Fix `password_wo` value not sent when migrating from `password`

### LINKED ISSUE #42582: [aws_db_instance] Can't migrate from plain `password` to `password_wo`

```console
Terraform v1.11.4
on darwin_arm64
```

* `aws_db_instance`

### Expected Behavior

I'm expecting that if I migrate from the `password` attribute of the **existing** `aws_db_instance`, the `password_wo` password will be used.

### Actual Behavior

I found that the actual password after applying is something like an empty string. Also, I can see this in the plan:
```
  ~ resource "aws_db_instance" "rds_metabase" {
        id                                    = "some_id"
      - password                              = (sensitive value) -> null
      + password_wo_version                   = 1
        tags                                  = {
            "Name" = "some_name"
        }
        # (71 unchanged attributes hidden)
    }
```
So, looks like the `password_wo` is missing in the plan, and the empty old password is applied.

```console
  ~ resource "aws_db_instance" "rds_metabase" {
        id                                    = "some_id"
      - password                              = (sensitive value) -> null
      + password_wo_version                   = 1
        tags                                  = {
            "Name" = "some_name"
        }
        # (71 unchanged attributes hidden)
    }
```

<details>

```hcl
ephemeral "random_password" "metabase_db_password" {
  length  = 32
  special = false
}

resource "aws_db_instance" "rds" {
  allocated_storage       = 5
  max_allocated_storage   = 20
  engine                  = "postgres"
  engine_version          = "14.13"
  instance_class          = "db.t3.micro"
  db_name                 = local.db_name
  identifier              = local.db_identifier
  username                = local.db_username
  password_wo             = ephemeral.random_password.metabase_db_password.result
  password_wo_version     = 1
  db_subnet_group_name    = aws_db_subnet_group.rds.name
  vpc_security_group_ids  = [aws_security_group.rds.id]
  skip_final_snapshot     = true
  backup_retention_period = 30
  maintenance_window      = "Tue:02:00-Tue:04:00"
  backup_window           = "04:00-06:00"
  apply_immediately       = true
  tags = {
    Name = "${local.db_identifier}"
  }
}
```
</details>

### Steps to Reproduce

1. Apply the `aws_db_instance` resource with `password`
2. Migrate the configuration to the `password_wo` and `password_wo_version`
3. Run `plan`
4. Run `apply`

### Debug Logging

<details>

```console

```
</details>

n/a

_No response_

No

Slevy35:
## Description

Fixes the migration path from the plain `password` attribute to the write-only `password_wo` + `password_wo_version` pair on `aws_db_instance`. Today, when a user removes `password` from their config and adds `password_wo` / `password_wo_version`, the `ModifyDBInstance` call is issued without `MasterUserPassword` and the master user password ends up effectively cleared on the running database.

Root cause is in [`internal/service/rds/instance.go`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/rds/instance.go), in `dbInstancePopulateModify`:

```go
if d.HasChange(names.AttrPassword) {
    needsModify = true
    if v, ok := d.GetOk(names.AttrPassword); ok {
        input.MasterUserPassword = aws.String(v.(string))
    }
}

if d.HasChange("password_wo_version") {
    passwordWO, di := flex.GetWriteOnlyStringValue(d, cty.GetAttrPath("password_wo"))
    diags = append(diags, di...)
    if diags.HasError() {
        return false, diags
    }
    if passwordWO != "" {
        input.MasterUserPassword = aws.String(passwordWO)
    }
}
```

In the migration scenario the first block fires (password removed -> nil), but it intentionally does not set `MasterUserPassword` because the new value is empty. The second block was expected to take over, but the two handlers are uncoordinated and the result is that `MasterUserPassword` may end up unset. Compounding this, the caller at line 2322 silently discarded the `diag.Diagnostics` returned by `dbInstancePopulateModify`, so any error reading the write-only attribute from the raw config was hidden from the user.

## Changes

- **`internal/service/rds/instance.go`** — Unify the `password` and `password_wo_version` handlers in `dbInstancePopulateModify`. Whenever either attribute changes, the new master user password is taken from `password` if set, otherwise from `password_wo` resolved via `flex.GetWriteOnlyStringValue`. Propagate diagnostics returned by `dbInstancePopulateModify` at the caller (`resourceInstanceUpdate`) so any failure to resolve `password_wo` surfaces as a real error.
- **`internal/service/rds/instance_test.go`** — Add `TestAccRDSInstance_passwordMigrationToWriteOnly`, a regression acceptance test that creates an instance with `password = "valid-password-1"`, then in a second step switches to `password_wo = "valid-password-2"` + `password_wo_version = 1`, and asserts the instance is not recreated, `password_wo_version` is recorded, and `password` is no longer present in state. Modeled on the existing `TestAccRDSInstance_password` and `TestAccRDSInstance_passwordWriteOnly` tests.

A changelog entry will be added in a follow-up commit using the PR number once this is opened, per `docs/changelog-process.md`.

## Closes

Closes #42582

## Acceptance test

`make testacc TESTS=TestAccRDSInstance_passwordMigrationToWriteOnly PKG=rds` — not yet run by submitter (no AWS credentials available); please verify in your acceptance test environment.

## Output from acceptance testing

_To be filled in by reviewers._

## Notes

- I did not change the `manage_master_user_password` path; that branch remains driven by its own `HasChange` block elsewhere.
- The reporter also noted similar-looking behavior on `aws_secretsmanager_secret_version` `secret_string_wo`. That is a different resource and out of scope here — happy to open a separate issue/PR if you confirm it is also affected.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changelog/47904.txt
+++ b/.changelog/47904.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_db_instance: Fix `password_wo` value not being sent to RDS when migrating an existing instance from `password` to `password_wo`
+```
--- a/internal/service/rds/instance.go
+++ b/internal/service/rds/instance.go
@@ -2319,7 +2319,11 @@ func resourceInstanceUpdate(ctx context.Context, d *schema.ResourceData, meta an
 				log.Println("[INFO] Only settings updating, instance changes will be applied in next maintenance window")
 			}
 
-			dbInstancePopulateModify(input, d)
+			_, di := dbInstancePopulateModify(input, d)
+			diags = append(diags, di...)
+			if diags.HasError() {
+				return diags
+			}
 
 			if d.HasChange(names.AttrEngineVersion) {
 				input.EngineVersion = aws.String(d.Get(names.AttrEngineVersion).(string))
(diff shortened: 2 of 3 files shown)


Slevy35:
@justinretzolk what do you think?

Slevy35:
@justinretzolk any updates?

PR #47911, opened 2026-05-14
.ci/semgrep: add `testcase-use-paralleltest` rule 

jar-b:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

Adds a new `semgrep` rule to enforce the use of `acctest.ParallelTest` on all exported acceptance tests. This explicitly enforces tests requiring serialized execution to be unexported and passed into a `_serial` test runner. 

There were 70+ existing violations of this rule which have been resolved by either:

1. Switching to `acctest.ParallelTest` when serialization is not necessary
2. Unexporting the test and wrapping in a `_serial` test runner
3. Adding a `nosemgrep` comment when a non-standard (but correct) testing pattern is being used 

**AI Disclosure:** An AI agent was used to write this `semgrep` rule and refactor impacted tests. All changes were manually reviewed and edited for correctness as needed. I fully understand all changes made.

### Relations

Relates #47888

PR #47888, opened 2026-05-13
.ci/semgrep: additional acceptance test rules

jar-b:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

Adds `semgrep` rules to ensure acceptance tests:

- Include a `PreCheck`
- Include an `ErrorCheck`
- Use `acctest.Context` to construct test case context

This enforces best practices already documented in the contributor guide: https://hashicorp.github.io/terraform-provider-aws/running-and-writing-acceptance-tests/

Issue #42261, opened 2025-04-16
[TechDebt]: Use `plancheck.ExpectResourceAction` with `_disappears` acceptance tests


jar-b:
### Description

Following the [design decision](https://hashicorp.github.io/terraform-provider-aws/design-decisions/expect-resource-action-with-disappears-tests/) published in #42206, this tech debt issue will track the effort to add the appropriate plan check to existing `_disappears` tests.

```go
        ConfigPlanChecks: resource.ConfigPlanChecks{
          PostApplyPostRefresh: []plancheck.PlanCheck{
            plancheck.ExpectResourceAction(resourceName, plancheck.ResourceActionCreate),
          },
        },
```

Relates #42206 
Relates #42207

PR #47914, opened 2026-05-14
.ci/semgrep: add `disappears-expect-resource-action` rule

jar-b:
## Rollback Plan

### Description

This rule requires that `_disappears` acceptance tests include a `PostApplyPostRefresh` config plan check that asserts the resource under test is planned for creation. This ensures that the expected non-empty plan is caused by the resource being tested and not a dependency or other side-effect of the apply.

**AI Disclosure:** An AI agent was used to write this `semgrep` rule. All changes were manually reviewed and edited for correctness as needed. I fully understand all additions made.

### Relations

Relates #42261
Relates #47911
Relates #47888


YakDriver (approved):
LGTM 🎉

PR #47939, opened 2026-05-15
.ci: lint for `TIP:` comments

jar-b:
## Rollback Plan

### Description

These comments are typically left over when `skaff` is used to outline an initial resource, data source, or list resource. Uses the [`godox`](https://golangci-lint.run/docs/linters/configuration/#godox) linter rather than rolling our own custom analyzer or using semgrep for a simple regexp search.


YakDriver (approved):
Nice, LGTM 🎉

PR #47940, opened 2026-05-16
r/aws_vpc_endpoint: Fix policy removal not resetting endpoint to default policy

### LINKED ISSUE #40973: [Bug]: aws_vpc_endpoint does not remove custom policy from resource if removed from Terraform config
### Terraform Core Version

1.5.7

### AWS Provider Version

5.84.0, 5.67.0

### Affected Resource(s)

- aws_vpc_endpoint

### Expected Behavior

If the policy parameter is removed from the `aws_vpc_endpoint` resource, it should remove the custom policy from the endpoint and revert to the default policy. This is the behaviour when modifying the endpoint on the AWS web console.

### Actual Behavior

```shell
No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed
```

```shell

```

```hcl
resource "aws_vpc_endpoint" "this" {
  vpc_id = var.vpc_id
  service_name = "com.amazonaws.us-east-1.elasticmapreduce"
  vpc_endpoint_type = "Interface"
  security_group_ids = [aws_security_group.vpc-endpoints.id]
  subnet_ids = var.private_subnets
  private_dns_enabled = true
  tags = {
    Name = "${module.vpc.name}-emr-endpoint"
  }
  policy = data.aws_iam_policy_document.vpc-endpoint-policy.json
}

resource "aws_security_group" "vpc-endpoints" {
 name        = "vpc-endpoints"
  description = "Security group for VPC Endpoints"
  vpc_id      = var.vpc_id
  tags = {
    "Name" = "${module.vpc.name}-endpoints"
  }
  lifecycle {
    create_before_destroy = true
  }
}

resource "aws_vpc_security_group_ingress_rule" "vpc-endpoints-ingress" {
  description       = "Allow traffic to VPC endpoints"
  security_group_id = aws_security_group.vpc-endpoints.id
  cidr_ipv4         = "10.0.0.0/8"
  from_port         = 443
  to_port           = 443
  ip_protocol       = "tcp"
}

data "aws_iam_policy_document" "vpc-endpoint-policy" {
  statement {
    sid       = "DenyIfNotFromVpc"
    effect    = "Deny"
    actions   = ["*"]
    resources = ["*"]
    principals {
      type        = "*"
      identifiers = ["*"]
    }
    condition {
      test     = "StringNotEquals"
      values   = [var.vpc_id]
      variable = "aws:sourceVpc"
    }
  }
}
```

### Steps to Reproduce

1. Apply terraform to create the VPC endpoint with the custom policy
2. Comment out or remove the policy parameter from the `aws_vpc_endpoint` resource and replan
3. Subsequent terraform plan will not show any diff, even though the policy parameter has been removed from the endpoint. This also has the same result if the `aws_iam_policy_document` data source is entirely removed from config

### Debug Output

_No response_

### Panic Output

_No response_

### Important Factoids

- Policy can still be modified, Terraform detects the changes to the content of the policy data source and updates the endpoint policy accordingly.

### References

_No response_

No

Sagargupta16:
### Description

Fix a long-standing bug where removing the \`policy\` attribute from an \`aws_vpc_endpoint\` configuration leaves the previously-applied custom policy in place forever. After this change, removing \`policy\` from configuration triggers an update that resets the endpoint to the AWS-default permissive policy, matching the behaviour of the AWS web console and the AWS CLI.

### Relations

Closes #40973.

### Root cause

The \`policy\` attribute is declared with \`sdkv2.IAMPolicyDocumentSchemaOptionalComputed()\`, which sets both \`Optional: true\` and \`Computed: true\`:

- [\`internal/service/ec2/vpc_endpoint.go\` line 143](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_endpoint.go#L143)
- [\`internal/sdkv2/schema.go\` lines 112-124](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/sdkv2/schema.go#L112-L124)

The \`Computed\` flag is required because the AWS API auto-assigns a default permissive policy when none is supplied at \`CreateVpcEndpoint\` time (see [\`ec2:ModifyVpcEndpoint\` API reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyVpcEndpoint.html), which exposes a \`ResetPolicy\` boolean for exactly this case). The side effect is that when an Optional+Computed attribute is removed from configuration, the SDKv2 plan layer fills the missing value from state, so \`d.HasChange(\"policy\")\` returns \`false\`. The existing update path (\`resourceVPCEndpointUpdate\`) is gated behind \`d.HasChanges(...names.AttrPolicy...)\`, so without a detected change, no \`ModifyVpcEndpoint\` call is made, and \`ResetPolicy\` is never sent.

The user-visible symptom: \`terraform plan\` reports \"No changes\" after removing the \`policy\` block, but the endpoint still enforces the old custom policy. Re-applying a different policy works correctly because that does generate a diff. Only **removal** is silently dropped.

### Fix

Add a third function to the existing \`CustomizeDiff: customdiff.All(...)\` block. It inspects the **raw config** via \`diff.GetRawConfig()\` -- the unmodified HCL representation, before SDKv2 fills computed values from state. When raw config has \`policy\` null and state holds a value, the function calls \`diff.SetNew(\"policy\", \"\")\` to surface the diff. The existing update path at [\`vpc_endpoint.go\` lines 458-473](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_endpoint.go#L458-L473) already converts an empty policy string into \`ResetPolicy: aws.Bool(true)\`, so no change is needed there.

Why \`GetRawConfig\` rather than another mechanism: this is the canonical SDKv2 idiom for \`Optional+Computed\` removal detection. The same pattern is used elsewhere in this provider, including [\`vpc_nat_gateway.go\`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_nat_gateway.go#L478) and [\`transitgateway_vpc_attachment.go\`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/transitgateway_vpc_attachment.go#L152). Setting \`Computed: false\` would force-replace the resource on every existing config that omits \`policy\` at create time -- a breaking change. \`SetNew\` to empty string is non-breaking and routes through the existing reset code path.

### Test

\`TestAccVPCEndpoint_gatewayPolicyRemoval\` exercises the regression directly:

1. Create a gateway endpoint with a custom \`DenyAll\` policy. Assert \`policy\` is non-empty.
2. Remove the \`policy\` attribute from configuration. Assert via \`plancheck.ExpectResourceAction(... ResourceActionUpdate)\` that the plan is an update (not a no-op). After apply, assert that \`policy\` is non-empty (because the API repopulates with the default) AND that it is **not equivalent** to the original \`DenyAll\` policy.

The new test follows the same shape as the existing \`TestAccVPCEndpoint_gatewayPolicy\`, reuses \`testAccVPCEndpointConfig_gatewayPolicy\`, and adds one new helper, \`testAccVPCEndpointConfig_gatewayNoPolicy\`, that omits the \`policy\` attribute.

### Output from acceptance testing

Acceptance tests have not been run locally -- this PR is from a community contributor without access to the upstream test AWS account. Local verification:

\`\`\`
$ go vet ./internal/service/ec2/...
(no output)
\`\`\`

Please run on a maintainer-controlled environment:

\`\`\`
$ make testacc TESTS=TestAccVPCEndpoint_gatewayPolicyRemoval PKG=ec2
\`\`\`

### References

- Issue: #40973
- AWS \`ec2:ModifyVpcEndpoint\` API (note the \`ResetPolicy\` boolean): <https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyVpcEndpoint.html>
- AWS VPC Endpoint policy semantics: <https://docs.aws.amazon.com/vpc/latest/privatelink/vpc-endpoints-access.html>
- Terraform SDKv2 \`CustomizeDiff\` and \`GetRawConfig\` reference: <https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences>
- HashiCorp guidance on \`Optional+Computed\` semantics: <https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#optional>
- Prior art in this provider for raw-config inspection on Optional+Computed fields:
  - \`vpc_nat_gateway.go\` (\`secondary_allocation_ids\`): <https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_nat_gateway.go#L478>
  - \`transitgateway_vpc_attachment.go\` (\`transit_gateway_default_route_table_*\`): <https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/transitgateway_vpc_attachment.go#L152>
- Working precedent for \`ResetPolicy\` in the dedicated \`aws_vpc_endpoint_policy\` resource: <https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/vpc_endpoint_policy.go#L68-L72>

### Changelog entry

Will be added in a follow-up commit once this PR has a number assigned (\`.changelog/<PR-NUMBER>.txt\`):

\`\`\`
\`\`\`release-note:bug
resource/aws_vpc_endpoint: Fix policy removal not resetting endpoint to default policy
\`\`\`
\`\`\`

### DIFF
--- a/.changelog/47940.txt
+++ b/.changelog/47940.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_vpc_endpoint: Fix policy removal not resetting endpoint to default policy
+```
--- a/internal/service/ec2/vpc_endpoint.go
+++ b/internal/service/ec2/vpc_endpoint.go
@@ -256,6 +256,42 @@ func resourceVPCEndpoint() *schema.Resource {
 			customdiff.ComputedIf("network_interface_ids", func(_ context.Context, diff *schema.ResourceDiff, meta any) bool {
 				return diff.HasChange("subnet_configuration") || diff.HasChange(names.AttrSubnetIDs)
 			}),
+			// Detect removal of the `policy` block from configuration.
+			//
+			// `policy` is `Optional + Computed` because the AWS API auto-assigns a
+			// default permissive policy at creation time. The Computed flag causes
+			// SDKv2 to silently keep the prior state value when the attribute is
+			// absent from configuration, which suppresses the diff and prevents the
+			// update path from emitting `ResetPolicy`. The result is that removing
+			// the `policy` block from a user's HCL is a no-op until the user also
+			// reapplies a default policy by hand.
+			//
+			// Inspecting the raw config (which reflects the user's HCL exactly,
+			// before any computed fill-in) lets us recognize the unset case and
+			// mark the attribute as changed, so the update path can call
+			// ModifyVpcEndpoint with `ResetPolicy: true`.
+			//
+			// Resolves https://github.com/hashicorp/terraform-provider-aws/issues/40973.
+			func(_ context.Context, diff *schema.ResourceDiff, meta any) error {
+				if diff.Id() == "" {
+					return nil
+				}
+				rawConfig := diff.GetRawConfig()
+				if !rawConfig.IsKnown() || rawConfig.IsNull() {
+					return nil
+				}
+				policyAttr := rawConfig.GetAttr(names.AttrPolicy)
+				if !policyAttr.IsKnown() {
+					return nil
+				}
+				stateHasPolicy := diff.Get(names.AttrPolicy).(string) != ""
+				if policyAttr.IsNull() && stateHasPolicy {
+					if err := diff.SetNew(names.AttrPolicy, ""); err != nil {
+						return err
+					}
+				}
+				return nil
+			},
 		),
 
 		Timeouts: &schema.ResourceTimeout{
--- a/internal/service/ec2/vpc_endpoint.go
+++ b/internal/service/ec2/vpc_endpoint.go
@@ -254,6 +254,42 @@ func resourceVPCEndpoint() *schema.Resource {
 			customdiff.ComputedIf("network_interface_ids", func(_ context.Context, diff *schema.ResourceDiff, meta any) bool {
 				return diff.HasChange("subnet_configuration") || diff.HasChange(names.AttrSubnetIDs)
 			}),
+			// Detect removal of the `policy` block from configuration.
+			//
+			// `policy` is `Optional + Computed` because the AWS API auto-assigns a
+			// default permissive policy at creation time. The Computed flag causes
+			// SDKv2 to silently keep the prior state value when the attribute is
+			// absent from configuration, which suppresses the diff and prevents the
+			// update path from emitting `ResetPolicy`. The result is that removing
+			// the `policy` block from a user's HCL is a no-op until the user also
+			// reapplies a default policy by hand.
+			//
+			// Inspecting the raw config (which reflects the user's HCL exactly,
+			// before any computed fill-in) lets us recognize the unset case and
+			// mark the attribute as changed, so the update path can call
+			// ModifyVpcEndpoint with `ResetPolicy: true`.
+			//
+			// Resolves https://github.com/hashicorp/terraform-provider-aws/issues/40973.
+			func(_ context.Context, diff *schema.ResourceDiff, meta any) error {
+				if diff.Id() == "" {
+					return nil
+				}
+				rawConfig := diff.GetRawConfig()
+				if !rawConfig.IsKnown() || rawConfig.IsNull() {
+					return nil
+				}
+				policyAttr := rawConfig.GetAttr(names.AttrPolicy)
+				if !policyAttr.IsKnown() {
+					return nil
+				}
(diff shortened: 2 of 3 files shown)


piyush-rathi-glean:
Any progress on this?

Sagargupta16:
Yes - the PR is current with main (no conflicts) and all checks are passing. It's waiting on maintainer review. Happy to rebase or adjust if the maintainers want any changes.

PR #47955, opened 2026-05-18
swissshepherd: Skill + Fixes 2

YakDriver:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

1. Add skill for fixing with swissshepherd
2. Fix several ignored targets
3. Update swissshepherd version

### Relations

Relates #47921

### Output from Acceptance Testing

```console
% make testacc TESTS=TestAccXXX PKG=ec2

...
```

PR #47959, opened 2026-05-18
swissshepherd: Fixes using skill

YakDriver:
## Rollback Plan

### Description

Experiment, using zero provider context and only the `.agents/skills/fixdocs` skill, agent was able to fix many issues. Needed a little guidance for some and swissshepherd needed some fixes as well.

### Relations

Relates #47955

```console

...
```


website/docs/d/auditmanager_control.html.markdown
```diff
@@ -64,4 +64,11 @@ This data source supports the following arguments:
 
 This data source exports the following attributes in addition to the arguments above:
 
-See the [`aws_auditmanager_control` resource](/docs/providers/aws/r/auditmanager_control.html) for details on the returned attributes - they are identical.
```
jar-b, line 67: May have missed this previously, but are we moving away from this pattern?

jar-b (approved):
LGTM 🚀

website/docs/d/auditmanager_control.html.markdown
YakDriver: I don't think it's very widely used

PR #47979, opened 2026-05-19
skaff: add `PreApply` config plan check for `_disappears` test

jar-b:
### Description

For consistency with the standard pattern being applied as existing `_disappears` tests are refactored.

### Relations

Relates #42261


ewbankkit (approved):
LGTM :rocket:.

PR #48036, opened 2026-05-23
`meta` service: Deprecate `id` attribute

gdavison:
## Rollback Plan

### Description

Deprecates the `id` attribute for:

* `data.aws_arn`
* `data.aws_default_tags`
* `data.aws_ip_ranges`
* `data.aws_partition`
* `data.aws_region`
* `data.aws_regions`
* `data.aws_service`

### DIFF
--- a/internal/service/meta/arn_data_source.go
+++ b/internal/service/meta/arn_data_source.go
@@ -13,7 +13,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-provider-aws/internal/framework"
-	"github.com/hashicorp/terraform-provider-aws/internal/framework/datasourceattribute"
 	fwflex "github.com/hashicorp/terraform-provider-aws/internal/framework/flex"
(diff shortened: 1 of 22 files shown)


internal/service/meta/arn_data_source.go
```diff
@@ -39,10 +41,7 @@ func (d *arnDataSource) Schema(ctx context.Context, request datasource.SchemaReq
 				CustomType: fwtypes.ARNType,
 				Required:   true,
 			},
-			names.AttrID: schema.StringAttribute{
-				Optional: true,
-				Computed: true,
-			},
+			names.AttrID: datasourceattribute.IDAttributeDeprecatedWithAlternate(path.Root(names.AttrARN)),
```
jar-b: This would change from optional/computed to computed only.

jar-b (commented):
These are technically breaking changes because `id` _could_ have been set as an optional argument, even though doing so has no effect. 

Assuming these should always have been computed-only, we just need to decide whether this breakage is acceptable or we preserve the existing behavior through until removal in v7. I lean a bit toward the latter personally, but am open to either.

internal/service/meta/arn_data_source.go
gdavison: Good point

jar-b (approved):
LGTM 🚀

PR #47873, opened 2026-05-12
feat: changelog automation

taruntej-a:
### Description

Automates changelog addition

PR #48047, opened 2026-05-25
Github action to invoke cloud agent for changelog automation

subham-ibmhc:
## Rollback Plan

### Description

Adds a github action which fires for PRs raised by maintainers to invoke the github copilot cloud agent. The agent adds a changelog by pickup up the changelog skill present in `.copilot/skills` directory.

## Relations

Relates https://github.com/hashicorp/terraform-provider-aws/pull/47873


.copilot/skills/changelog/SKILL.md
```diff
@@ -0,0 +1,47 @@
+---
```
YakDriver, line 1: Argh. We can't use .agents/skills? I don't like the idea of having two parallel skills areas. It becomes confusing fast.

.github/workflows/changelog-agent.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Trigger Copilot to add changelog
+        if: steps.community_check.outputs.maintainer == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.COPILOT_REQUESTS_TOKEN }}
+        run: |
+          PR_NUMBER=${{ github.event.pull_request.number }}
+          PR_URL="https://github.com/${{ github.repository }}/pull/${PR_NUMBER}"
+          gh pr comment "$PR_NUMBER" \
+            --repo "${{ github.repository }}" \
+            --body "@copilot Add changelog for ${PR_URL}. Skip the push confirmation."
```
jar-b: I wonder if we should reword the `.copilot` skill to be scoped to this CI task. I realize it is currently a copy of the existing changelog skill in `.agents`, but it seems odd to put so much emphasis on asking for confirmation in the skill, only to have the comment tell the bot to ignore it.

This does complicate reconciling the two skills back together in the future, but since we are still proving this out in CI removing ambiguity from the task seems like a good thing.

.github/workflows/changelog-agent.yml
```diff
@@ -0,0 +1,40 @@
+# Copyright IBM Corp. 2014, 2026
+# "SPDX-License-Identifier: MPL-2.0"
+
+name: Changelog automation
+
+on:
+  pull_request:
```
justinretzolk: Since this workflow uses secrets, we'll need to use `pull_request_target`, since workflows triggered by `pull_request` events from forks [don't get access to secrets](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflows-in-forked-repositories).

.github/workflows/changelog-agent.yml
```diff
@@ -0,0 +1,40 @@
...
+    types:
+      - ready_for_review
```
justinretzolk, line 9: By only having `ready_for_review` as a type, this workflow will only be triggered when the PR is move from draft to ready for review or when a PR is opened as ready for review rather than as a draft, and won't re-trigger if there are any additional commits that might change what the output is. It might be worth exploring whether we want to add additional trigger types, and what that would mean for the Skill configuration.

.github/workflows/changelog-agent.yml
```diff
@@ -0,0 +1,40 @@
...
+          user_login: ${{ github.event.pull_request.user.login }}
+          maintainers: ${{ secrets.MAINTAINERS }}
+
+      - name: Trigger Copilot to add changelog
+        if: steps.community_check.outputs.maintainer == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.COPILOT_REQUESTS_TOKEN }}
+        run: |
+          PR_NUMBER=${{ github.event.pull_request.number }}
+          PR_URL="https://github.com/${{ github.repository }}/pull/${PR_NUMBER}"
```
justinretzolk: We should be able to just use `${{ github.event.pull_request.html_url }}` here, and since these variables are only used once, we could either:

- Move these variable definitions into the `env` key
- Use the interpolations directly in the bash command below (e.g. `gh pr comment ${{ github.event.pull_request.number}}`) rather than defining them explicitly

Unless I'm forgetting something, that suggestion is more stylistic than anything, but figured it was worth mentioning.

justinretzolk (changes requested):
Made some initial suggestions for the workflow itself, some of which would prevent the workflow from working as intended.

Not to throw a wrench in things and slow down progress on getting this merged, but I do wonder if this is a use-case where we should explore [GitHub Agentic Workflows](https://github.github.com/gh-aw/) rather than using "standard" GitHub Actions. This seems like exactly the type of thing that Agentic Workflows are aimed at solving, though it's worth mentioning that the Agentic Workflow documentation indicates that it's still in early development and may change significantly while they continue to improve the offering.

.github/workflows/changelog-agent.yml
subham-ibmhc: My initial thought process was that once the PR is marked for ready-for-review, the overall context of what the PR means to fix/implement is already set. So even though a review may change/refactor some underlying implementation detail, the overall context of the PR remains the same. 
We had also thought of a labels approach for this, where the contributor could add a specific label and only then this action would be triggered, but since we wanted a completely hands-off approach with low friction, went with this approach instead.
Additional trigger types could be investigated.

.github/workflows/changelog-agent.yml
subham-ibmhc: Agreed with removing ambiguity, will reword the skill and prompt on the action.

.copilot/skills/changelog/SKILL.md
subham-ibmhc: Unfortunately, the cloud agent does not seem to be able to load in skills from `.agents/skills` and requires it to be present in the `.copilot` directory instead.

ewbankkit: https://docs.github.com/en/copilot/concepts/agents/about-agent-skills

.github/workflows/changelog-agent.yml
```diff
@@ -0,0 +1,40 @@
...
+        id: community_check
+        uses: ./.github/actions/community_check
+        with:
+          user_login: ${{ github.event.pull_request.user.login }}
+          maintainers: ${{ secrets.MAINTAINERS }}
+
+      - name: Trigger Copilot to add changelog
+        if: steps.community_check.outputs.maintainer == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.COPILOT_REQUESTS_TOKEN }}
```
ewbankkit:
```suggestion
          GH_TOKEN: ${{ secrets.COPILOT_GITHUB_TOKEN }}
```

For consistency (repo secret has been added).

.copilot/skills/changelog/SKILL.md
taruntej-a: We initially tried this but did not seem to work and found this comment/workaround - https://github.com/github/copilot-cli/issues/1080#issuecomment-3807486889 in one of the issue reported - https://github.com/github/copilot-cli/issues/1080

But it’s still just a workaround.

ewbankkit: Pragmatically I'm happy with the skill being in the `copilot`-specific directory as this is its only use.
Could you remove the copy from `.agents/skills` so we don't start out with duplicates all over the place.

subham-ibmhc: Removed the copy from `.agents/skills` directory.

ewbankkit:
@justinretzolk RE GitHub Agentic Workflows I have created https://github.com/hashicorp/terraform-provider-aws/pull/48447 (using AWs) for some work I have been doing.
Initially it is probably worth getting some experience with both (and additional) approaches.

justinretzolk (approved):

PR #48091, opened 2026-05-27
AutoFlex: Simplification

gdavison:
## Rollback Plan

### Description

Simplifies AutoFlex by making function calls standalone and removing need for `autoFlexer` interface


ewbankkit (approved):
LGTM :rocket:.

I think the reason I had that `interface` originally was to support e.g. "legacy autoflex", but YAGNI.

PR #48108, opened 2026-05-28
docs: replace `sdkaccest.RandomWithPrefix` with internal `acctest` variant

jar-b:
## Rollback Plan

### Description

We now have linters in place to enforce use of the internal `acctest` package for this helper function, so the contributor documentation should reflect current best practices.


YakDriver (approved):
LGTM 🎉

PR #48141, opened 2026-05-31
resource/aws_emrserverless_application: Add `session_enabled` to `interactive_configuration` (Spark Connect) 🤖🤖🤖

### LINKED ISSUE #49281: [Bug]: aws_emrserverless_application resets interactive session support on every update

Terraform v1.15.8, AWS provider v6.47.0. Also present on provider `main`.

* `aws_emrserverless_application`

### Expected Behavior

`interactiveConfiguration.sessionEnabled` survives an apply that changes some other attribute of the application.

### Actual Behavior

Any apply that updates the application resets it to `false`.

`interactive_configuration` is `Optional + Computed`, so a refresh writes the block into state from the API even when it was never configured. `resourceApplicationUpdate` then re-sends `InteractiveConfiguration` on any attribute change, built by `expandInteractiveConfiguration` from `livy_endpoint_enabled` and `studio_enabled` only — there is no `session_enabled` argument. `SessionEnabled` goes out unset and the API resets it.

Every Spark Connect `StartSession` call against the application then fails, so any Spark job using it breaks until the flag is set again outside Terraform — where it survives only until the next apply.

There is no AWS CLI fallback: `aws-cli 2.34.32` rejects `--interactive-configuration sessionEnabled=true` with `Unknown parameter in interactiveConfiguration`, and `aws emr-serverless get-application` omits the field from its output, so the CLI can neither set nor read it. Cloud Control API or a direct SDK call are the only options.

```console
ValidationException: Sessions are not enabled on this application
```

```terraform
resource "aws_emrserverless_application" "example" {
  name          = "example"
  release_label = "emr-7.13.0"
  type          = "spark"

  runtime_configuration {
    classification = "spark-defaults"
    properties = {
      "spark.jars.ivy" = "/tmp/.ivy2"
    }
  }
}
```

### Steps to Reproduce

1. Apply the configuration above.
2. Set `sessionEnabled=true` out of band — `aws cloudcontrol update-resource --type-name AWS::EMRServerless::Application --identifier <id> --patch-document '[{"op":"replace","path":"/InteractiveConfiguration/SessionEnabled","value":true}]'`.
3. Change any attribute of the application (adding a `runtime_configuration` property is enough) and apply.
4. `sessionEnabled` is `false` again.

### Workaround

Two parts, in case they are useful to anyone else hitting this.

To set the flag, a small Go program using the SDK field the provider does not expose. Read the application, change only `InteractiveConfiguration.SessionEnabled`, and pass every other setting back unchanged — whether `UpdateApplication` clears members it is not sent is undocumented, and losing `RuntimeConfiguration` here would break every job:

```go
out, err := client.GetApplication(ctx, &emrserverless.GetApplicationInput{ApplicationId: aws.String(id)})
// ...
_, err = client.UpdateApplication(ctx, &emrserverless.UpdateApplicationInput{
    ApplicationId: aws.String(id),
    ClientToken:   aws.String(uuid.NewString()),
    InteractiveConfiguration: &types.InteractiveConfiguration{
        SessionEnabled:      aws.Bool(true),
        LivyEndpointEnabled: out.Application.InteractiveConfiguration.LivyEndpointEnabled,
        StudioEnabled:       out.Application.InteractiveConfiguration.StudioEnabled,
    },
    // plus Architecture, AutoStart/AutoStopConfiguration, MaximumCapacity,
    // NetworkConfiguration, ReleaseLabel, RuntimeConfiguration, ... echoed back
})
```

The application must be `CREATED` or `STOPPED`; `UpdateApplication` rejects a `STARTED` one, and stopping it terminates live sessions. `aws cloudcontrol update-resource` with a patch document on `/InteractiveConfiguration/SessionEnabled` does the same thing in one call and preserves the other properties for you.

To notice when Terraform has reset it, a `check` block, since `emr-serverless get-application` cannot show the field:

```terraform
check "interactive_sessions_enabled" {
  data "aws_cloudcontrolapi_resource" "app" {
    identifier = aws_emrserverless_application.example.id
    type_name  = "AWS::EMRServerless::Application"
  }

  assert {
    condition = try(
      jsondecode(data.aws_cloudcontrolapi_resource.app.properties).InteractiveConfiguration.SessionEnabled,
      false
    )
    error_message = "Interactive sessions are disabled; every StartSession call will fail."
  }
}
```

A failed `check` assert is a warning rather than an error, so this does not block an apply — it just reports the state on every plan. Auto-remediating it from Terraform needs a `local-exec` shelling out to the AWS CLI, which is not viable on a CI runner without the CLI installed.

Claude Code, used to trace the update path in the provider and to verify the behaviour against a live account. Conclusions checked by hand.

* #48141 adds the missing `session_enabled` argument and would resolve this.
* `SessionEnabled` exists in `aws-sdk-go-v2` `service/emrserverless/types.InteractiveConfiguration` (v1.44.2), and the `AWS::EMRServerless::Application` Cloud Control schema exposes `InteractiveConfiguration/SessionEnabled`.
* The argument requires `emr-7.13.0` or later; earlier release labels reject it with `Sessions are not supported on the selected release label`.

No — #48141 already implements it.

Idan-Or:
### Description

Adds a `session_enabled` argument to `aws_emrserverless_application.interactive_configuration`, wiring up the existing `SessionEnabled` field on `types.InteractiveConfiguration` from the AWS SDK Go v2.

When set to `true`, AWS EMR Serverless allows callers to start interactive sessions on the application using the `StartSession` API — required for [Apache Spark Connect](https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/spark-connect.html) with EMR release `emr-7.13.0` and later. This is a separate flag from `livy_endpoint_enabled` (which enables the Livy REST endpoint) and `studio_enabled` (which enables EMR Studio integration).

Without this argument, the only way to enable Spark Connect sessions on an EMR Serverless application is to bypass Terraform and use the AWS CLI or SDK directly, after which Terraform sees state drift.

```hcl
resource "aws_emrserverless_application" "example" {
  name          = "spark-connect-app"
  release_label = "emr-7.13.0"
  type          = "spark"

  interactive_configuration {
    session_enabled = true
  }
}
```

### Relations

Closes #49281

### References

- AWS docs: [Run interactive sessions with Amazon EMR Serverless through Spark Connect](https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/spark-connect.html)
- AWS SDK Go v2 [`types.InteractiveConfiguration`](https://github.com/aws/aws-sdk-go-v2/blob/main/service/emrserverless/types/types.go) — already exposes `SessionEnabled *bool`

The existing `TestAccEMRServerlessApplication_interactiveConfiguration` was updated to also exercise `session_enabled`, with the release label bumped to `emr-7.13.0` (minimum supported by AWS for Spark Connect sessions). I do not have an AWS account configured for the provider's acceptance test harness; happy to follow up if a maintainer can run the acceptance test, or to adjust the test if there's a preferred approach.

```
$ go vet ./internal/service/emrserverless/...
ok
$ go build ./internal/service/emrserverless/...
ok
$ go test -c -o /tmp/emrs ./internal/service/emrserverless/...
ok
```

### Changelog

`.changelog/00000.txt` is a placeholder — should be renamed to the PR number once assigned.


Idan-Or:
~Closing as redundant — `session_enabled` was added to `main` independently in 72db697a544 by @gdavison on May 21. The feature is already available in the provider.~

My bad, this was not the case. PR reopened and rebased.
@justinretzolk / team - would appreciate it if you can prioritize this one.

Idan-Or:
PR reopened - my assumption this change was already done in another commit was wrong.

Idan-Or:
Hi @justinretzolk 👋 
Any chance this one will be merged soon?

Idan-Or:
@justinretzolk - would appreciate it if you can push this one forward as well as https://github.com/hashicorp/terraform-provider-aws/pull/48673 .

pequalsnp:
We hit the underlying bug in production (filed separately as #49281) and had implemented the same change before finding this PR, so rather than open a competing one: I ran your acceptance test against a real account. It fails in two places, both in the test rather than in the resource code, and both fixable with a test-only diff. With the diff below all 6 steps pass.

### 1. Step 1 fails on `scheduler_configuration` drift

```
application_test.go:273: Step 1/6 error: After applying this test step, the non-refresh plan was not empty.

  # aws_emrserverless_application.test will be updated in-place
  ~ resource "aws_emrserverless_application" "test" {
      - scheduler_configuration {
          - max_concurrent_runs   = 15 -> null
          - queue_timeout_minutes = 360 -> null
        }
    }

```

From `emr-7.13.0`, `GetApplication` returns `schedulerConfiguration {maxConcurrentRuns: 15, queueTimeoutMinutes: 360}` whether or not one was requested. `scheduler_configuration` is `Optional` **without** `Computed` at the block level (the two inner ints are `Computed`, the block is not), so the API-supplied value lands in state and never matches an empty config.

This is invisible on the release labels the other tests use — on `emr-7.1.0` the API returns no `schedulerConfiguration` at all. Since `session_enabled` requires `emr-7.13.0` (`CreateApplication` on older labels returns `ValidationException: Sessions are not supported on the selected release label`), any test for this argument runs into it.

### 2. Step 6 asserts attributes that cannot exist

```
application_test.go:273: Step 6/6 error: Check failed: Check 3/5 error:
  aws_emrserverless_application.test: Attribute 'interactive_configuration.0.livy_endpoint_enabled' not found
```

With all three flags `false`, the API returns `interactiveConfiguration: null`. `flattenInteractiveConfiguration` returns `nil` for that, but Read wraps it unconditionally:

```go
d.Set("interactive_configuration", []any{flattenInteractiveConfiguration(application.InteractiveConfiguration)})
```

so state holds a one-element list containing null. `interactive_configuration.#` is `1` (your check 2 passes) while `interactive_configuration.0.*` do not exist (check 3 fails). Asserting absence matches what the provider actually stores.

Arguably Read should set an empty list when the API returns nil, which would let the step assert `interactive_configuration.# = 0` instead — but that is a behaviour change beyond this PR, so the diff below just asserts the current shape.

### Diff that makes it pass

```diff
@@ testAccApplicationConfig_interactiveConfiguration
   interactive_configuration {
     livy_endpoint_enabled = %[2]t
     session_enabled       = %[3]t
     studio_enabled        = %[4]t
   }
+  # From emr-7.13.0 the API always returns a schedulerConfiguration, and
+  # scheduler_configuration is Optional without Computed at the block level, so
+  # declaring the API defaults is what keeps the post-apply plan empty.
+  scheduler_configuration {
+    max_concurrent_runs   = 15
+    queue_timeout_minutes = 360
+  }
 }

@@ step 6 (all three flags off)
+					// With all three flags false the API returns interactiveConfiguration:
+					// null, and Read stores []any{flattenInteractiveConfiguration(nil)} —
+					// a one-element list holding null — so the count is 1 but none of the
+					// attributes exist in state.
 					resource.TestCheckResourceAttr(resourceName, "interactive_configuration.#", "1"),
-					resource.TestCheckResourceAttr(resourceName, "interactive_configuration.0.livy_endpoint_enabled", acctest.CtFalse),
-					resource.TestCheckResourceAttr(resourceName, "interactive_configuration.0.session_enabled", acctest.CtFalse),
-					resource.TestCheckResourceAttr(resourceName, "interactive_configuration.0.studio_enabled", acctest.CtFalse),
+					resource.TestCheckNoResourceAttr(resourceName, "interactive_configuration.0.livy_endpoint_enabled"),
+					resource.TestCheckNoResourceAttr(resourceName, "interactive_configuration.0.session_enabled"),
+					resource.TestCheckNoResourceAttr(resourceName, "interactive_configuration.0.studio_enabled"),
```

Result, `us-east-1`, your branch at 7495c7cb4c2 plus the diff above:

```
--- PASS: TestAccEMRServerlessApplication_interactiveConfiguration (128.98s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/emrserverless	129.071s
```

No resources leaked — only pre-existing applications remained afterwards.

### One note on the changelog

The resource-code change here also fixes a bug, not only a missing argument: because `interactive_configuration` is `Optional + Computed`, a refresh populates the block from the API even when it was never configured, and `resourceApplicationUpdate` then re-sends `InteractiveConfiguration` on any attribute change built from `livy_endpoint_enabled` / `studio_enabled` alone — so `SessionEnabled` goes out unset and the API resets it to `false`. Today that silently disables interactive sessions on any apply that touches the application for an unrelated reason, and there is no AWS CLI path to restore it (`aws-cli 2.34.32` rejects `sessionEnabled` as an unknown parameter, and `get-application` omits the field). Details in #49281. A `release-note:bug` entry alongside the enhancement note may be worth adding.

Happy to push these test changes as a PR against your branch if that is easier than applying them yourself.

Idan-Or:
@pequalsnp Thanks for tracking down both failures and running the acceptance test against a real account — really appreciate it. Pushed your diff verbatim:

- `12f6c32` — the `scheduler_configuration` block in the test config (drift fix for `emr-7.13.0`) and the `TestCheckResourceAttr` → `TestCheckNoResourceAttr` swap for the all-flags-off step.
- `4730d84` — added a `release-note:bug` entry alongside the enhancement note per your comment, since this also fixes the `session_enabled` reset you hit in #49281.

Let me know if anything still looks off.

PR #48207, opened 2026-06-04
fix(fms): resource_tag_logical_operator silently changes from OR to AND

AruneshDwivedi:
The `resource_tag_logical_operator` attribute in `aws_fms_policy` was being set inside the resource tags loop using `d.Get()`, which returns an empty string when the attribute is not explicitly set. This silently changed the API value from OR to AND without showing a diff in the plan.

Switched to `d.GetOk()` so the attribute is only set when explicitly provided, preserving the existing API behavior when it's not configured.

### DIFF
--- a/.changelog/48207.txt
+++ b/.changelog/48207.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_fms_policy: Fix `resource_tag_logical_operator` silently changing from OR to AND when not set in config
+```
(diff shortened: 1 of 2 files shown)


justinretzolk:
Similar #47933

AruneshDwivedi:
I see #47933 takes a different approach (using Default in the schema). My fix uses GetOk to only set the field when it's explicitly in the config, which avoids changing the schema definition. Both address the same root cause but with different trade-offs. Happy to close if the team prefers the #47933 approach.

AruneshDwivedi:
Thanks for the pointer. #47933 defaults the operator to AND when it is omitted, whereas this PR fixes the opposite direction: d.Get() was returning an empty string for resource_tag_logical_operator when it was not set, silently flipping the API value from OR to AND with no plan diff. The two changes are complementary, not duplicates. I can add a changelog entry if you want this bundled with the #47933 work.

PR #48257, opened 2026-06-08
resource/aws_glue_catalog: Ignore InvalidInputException when listing tags on the root catalog

### LINKED ISSUE #47923: [Bug]: aws_glue_catalog Read fails with InvalidInputException — catalog is not a taggable Glue resource
## Terraform Version

```
Terraform v1.14.9
provider registry.terraform.io/hashicorp/aws v6.44.0
```

## Affected Resource(s)

- `aws_glue_catalog`

## Description

Every plan/apply/import against the root (account-level) catalog fails because `Read` unconditionally calls `listTags`, but the [AWS Glue tagging documentation](https://docs.aws.amazon.com/glue/latest/dg/monitor-tags.html) explicitly does not list the catalog as a taggable resource type. The API confirms this at runtime:

```
Error: reading Glue Catalog (<account_id>)
While listing tags, operation error Glue: GetTags,
InvalidInputException: Tag operations not supported on root catalog
```

This is an AWS API restriction (`InvalidInputException`, not `AccessDeniedException`) — it is not an IAM or Lake Formation permissions issue.

haidargit:
## Rollback Plan

No.

### Description

The `Read` method for `aws_glue_catalog` always calls `listTags`, but the  glue API does not support tag operations on the root (account-level) catalog. It returns `InvalidInputException`, which caused every plan, apply, and import against the root catalog to fail. This skips the tag-listing error only when it is an `InvalidInputException`, since that signals the API-level restriction.

Thanks to @jganucheau-exiger for the helpful input on the data-source fix and its root-catalog test. 🙂

### Relations

Closes #47923

### References

- taggable resource: <br> https://docs.aws.amazon.com/glue/latest/dg/monitor-tags.html

image <br>
`go test ./internal/service/glue/ -list 'TestAccGlue_serial'`
Screen Shot 2026-08-05 at 11 40 02 
<br>`TF_ACC=1 go test ./internal/service/glue/ -v -timeout 20m \ -run 'TestAccGlue_serial/CatalogDataSource/rootCatalog'`
Screen Shot 2026-08-05 at 11 44 34


jganucheau-exiger:
This addresses our case exactly — thanks for picking it up.

We're blocked by #47923 on provider 6.56.0 in HCP Terraform: importing the
account's root catalog (name = account ID) fails at refresh with
`InvalidInputException: Tag operations not supported on root catalog` from
the unconditional `listTags` in Read. Swallowing it for the root catalog is
the right shape as far as we can tell — `terraform validate` already passes,
so the tag read is the only thing standing between us and managing
`CatalogProperties.IcebergOptimizationProperties` declaratively.

Two notes that may be worth folding in:

- **`data.aws_glue_catalog` has the identical unconditional `listTags`** and
  would still fail against the root catalog after this merges. Same
  treatment there would also unblock `check`-block drift detection for
  anyone using the out-of-band workaround in the meantime.
- Root-catalog management is how AWS documents enabling
  [catalog-level Iceberg table optimizers](https://docs.aws.amazon.com/glue/latest/dg/enable-auto-table-optimizers.html),
  so this unblocks a supported AWS feature that currently has no declarative
  path at all (Cloud Control's `AWS::Glue::Catalog` schema doesn't model
  `IcebergOptimizationProperties`).

Happy to validate against a real root catalog if a pre-release build or
branch artifact is available.

haidargit:
Hi @jganucheau-exiger, thanks for the input. 👍🏻 👍🏻  i am looking into it

haidargit:
hi @jganucheau-exiger, I've updated the aws_glue_catalog data, its Read had the identical  list tags call + an acceptance test against the root catalog. you can reproduce it. thanks

`go test ./internal/service/glue/ -list 'TestAccGlue_serial'`

`TF_ACC=1 go test ./internal/service/glue/ -v -timeout 20m \
  -run 'TestAccGlue_serial/CatalogDataSource/rootCatalog'`
  
  
Screen Shot 2026-08-05 at 11 40 02  <br> 
Screen Shot 2026-08-05 at 11 44 34

PR #48272, opened 2026-06-08
Build the AWS Provider on S390x

JenGoldstrich:
### Description

This PR adds a build for s390x linux for the AWS Provider.  My team uses the AWS Provider to provision resources in CI, and we'd like to use it on our s390x runners as well.  For now I just compile it locally, but would prefer to switch to the official artifact

### DIFF
--- a/.changelog/48272.txt
+++ b/.changelog/48272.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+Add Linux s390x support
+```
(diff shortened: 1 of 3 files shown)


ewbankkit:
Based on e.g. https://github.com/hashicorp/terraform-provider-local/pull/504 it seems you'll also have to update `.release/terraform-provider-aws-artifacts.hcl`.

Also a CHANGELOG entry , see https://hashicorp.github.io/terraform-provider-aws/changelog-process/#region-validation-support, an enhancement for the provider as a whole saying something like `Add Linux s390x support`.

ewbankkit (approved):
LGTM :rocket:.

jar-b (approved):
LGTM 🚀

Issue #48159, opened 2026-06-02
[Bug]: aws_bedrockagentcore_harness produces inconsistent result after apply (environment.agentcore_runtime_environment computed fields)


roman-laguta-rs:
### Terraform and AWS Provider Version

```
Terraform v1.13.x
hashicorp/aws v6.47.0
```

### Affected Resource(s)

- `aws_bedrockagentcore_harness`

### Description

When creating an `aws_bedrockagentcore_harness` with an `environment` block (to
set `network_configuration` for VPC mode), `terraform apply` fails with
**"Provider produced inconsistent result after apply"**.

The API populates several fields under
`environment[0].agentcore_runtime_environment[0]` server-side after creation,
but those fields are **not marked `Computed`** in the provider schema. They are
planned as `null`, but the apply response returns values, which fails Terraform
core's post-apply consistency check. The resource **is** created in AWS, but the
run errors and the resource is left **tainted**, so the next apply tries to
destroy/recreate it — which fails the same way, producing a permanent
create/taint loop.

The four fields:

- `lifecycle_configuration` — API defaults it to
  `{ idle_runtime_session_timeout = 900, max_lifetime = 28800 }`
- `agent_runtime_arn` — server-assigned (random suffix)
- `agent_runtime_id` — server-assigned (random suffix)
- `agent_runtime_name` — server-derived (`harness_<harness_name>`)

`agent_runtime_arn`/`agent_runtime_id` contain a random suffix, so they cannot
be pre-set in configuration to dodge the inconsistency.

### Expected Behavior

The harness is created and the apply succeeds, with the server-populated
`agentcore_runtime_environment` fields treated as computed/read-only values.

### Actual Behavior

```
Error: Provider produced inconsistent result after apply

When applying changes to aws_bedrockagentcore_harness.agent, provider
"registry.terraform.io/hashicorp/aws" produced an unexpected new value:
.environment[0].agentcore_runtime_environment[0].lifecycle_configuration: was
null, but now cty.ListVal([...{idle_runtime_session_timeout=900,
max_lifetime=28800}...]).

(and the same for .agent_runtime_arn, .agent_runtime_id, .agent_runtime_name)

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

### Sample Terraform Configuration

```hcl
resource "aws_bedrockagentcore_harness" "agent" {
  harness_name       = "example"
  execution_role_arn = aws_iam_role.example.arn

  model {
[...]

PR #48317, opened 2026-06-10
Add `TestAccBedrockAgentCoreHarness_environment`

ewbankkit:
## Rollback Plan

### Description

Adds an `aws_bedrockagentcore_harness` acceptance test for the `environment {}` block.

### Relations

Relates https://github.com/hashicorp/terraform-provider-aws/issues/48159.

```console

...
```

### DIFF
--- a/internal/service/bedrockagentcore/harness.go
+++ b/internal/service/bedrockagentcore/harness.go
@@ -569,7 +569,7 @@ func (r *harnessResource) Create(ctx context.Context, request resource.CreateReq
 	}
 
 	// Set values for unknowns.
-	smerr.AddEnrich(ctx, &response.Diagnostics, fwflex.Flatten(ctx, harness, &data, fwflex.WithFieldNamePrefix("AgentRuntime")))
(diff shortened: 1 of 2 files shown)


harishvadali:
@ewbankkit  Thank you for fixing this issue. I see the MR is in Draft status. Any idea on when this would be prioritized for release?

aidandaly24:
Hi @ewbankkit, heads-up on some overlap we have #48815 open which fixes the same #48159 `environment` "inconsistent result after apply" by making the server-assigned `environment` leaves (`agent_runtime_arn`/`id`/`name`, `lifecycle_configuration`) `Computed`, and it includes an `environment` regression acctest. It also makes the same one-line Create-flatten cleanup you have here (`fwflex.Flatten(..., WithFieldNamePrefix("AgentRuntime"))` → `r.flatten`), so there's a small collision on that line.

We're moving forward with #48815 since it carries the actual schema fix for #48159, just flagging so we're not duplicating effort (once it lands this should be covered). Thanks for digging into #48159 in parallel!

PR #48345, opened 2026-06-11
aws_cloudfront_distribution_tenant: Fix inconsistent result after apply error when creating an aws_cloudfront_distribution_tenant with a managed_certificate_request block and no explicit customizations block

### LINKED ISSUE #45987: aws_cloudfront_distribution_tenant: "Provider produced inconsistent result after apply" when using managed_certificate_request

```console
Terraform v1.14.3
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v6.28.0
```

- aws_cloudfront_distribution_tenant

### Expected Behavior

When creating a distribution tenant with `managed_certificate_request`, the resource should be created successfully and the state should reflect the certificate that CloudFront automatically attaches to `customizations.certificate`.

### Actual Behavior

After apply, the provider throws an error because CloudFront automatically populates `customizations[0].certificate` with the issued certificate, but the provider expects block count 0 (matching HCL config without explicit `customizations.certificate` block).

```console

When applying changes to
aws_cloudfront_distribution_tenant.domains["example.com"],
provider "provider["registry.terraform.io/hashicorp/aws"]" produced an
unexpected new value: .customizations[0].certificate: block count changed
from 0 to 1.

This is a bug in the provider, which should be reported in the provider's
own issue tracker.
```

<details open>

```hcl
resource "aws_cloudfront_distribution_tenant" "example" {
  name            = "example-tenant"
  distribution_id = aws_cloudfront_multitenant_distribution.main.id
  enabled         = true

  domain {
    domain = "example.com"
  }

  # CloudFront issues and manages the certificate automatically
  managed_certificate_request {
    primary_domain_name   = "example.com"
    validation_token_host = "cloudfront"
  }

  parameter {
    name  = "domain"
    value = "example.com"
  }
}
```
</details>

### Steps to Reproduce

1. Create a multi-tenant distribution and connection group
2. Create a distribution tenant with `managed_certificate_request` block
3. Run `terraform apply`
4. Observe the "Provider produced inconsistent result" error

### Debug Logging

<details open>

```console

```
</details>

n/a

_No response_

No

tony-yu1:
## Description

  Fixes the Provider produced inconsistent result after apply error when creating an aws_cloudfront_distribution_tenant with a managed_certificate_request block and no explicit customizations block.

 ## Root cause

  When managed_certificate_request is used, CloudFront issues a managed certificate and auto-populates customizations.certificate in the API response — even though the practitioner never configured a customizations block. The provider flattened that response straight
  into state, so:

  - plan had customizations block count 0 (nothing in config), but
  - state had block count 1 (the service-injected certificate).

  Terraform core rejects this mismatch (.customizations[0].certificate: block count changed from 0 to 1). The resource is tainted, and the next plan schedules a destroy + recreate of a working tenant — at scale (e.g. CloudFront-for-SaaS migrations) this is
  operationally untenable and can cause customer-facing TLS gaps.

 ##  Fix

  In Create, Read, and Update, reconcile the API-injected certificate against the practitioner's configuration via a helper (suppressManagedCertCustomizations) that runs only when a managed certificate is in play and the user did not explicitly configure a
  certificate sub-block. It operates at the certificate sub-block level (not the whole customizations block), so it covers both:

  1. No customizations block at all — the reported case; planned value (null) is preserved.
  2. customizations with web_acl/geo_restriction but no certificate — only the injected certificate sub-block is suppressed; the practitioner's other sub-blocks and any server-computed values are preserved.

  An explicitly configured customizations.certificate is always passed through unchanged. In Read, the reconciliation is gated on the prior managed_certificate_request so that terraform import (which passes through only the id) round-trips an explicitly configured
  certificate correctly.

  This makes plan == apply, eliminating both the create-time error and the spurious destroy + recreate. No HCL changes are required and the change is non-breaking.

 ## Testing

  - Unit test — TestSuppressManagedCertCustomizations exercises every branch of the reconciliation helper (planned null, planned unknown, planned empty, explicit certificate pass-through, geo_restriction without certificate, and the combined geo_restriction +
  injected-certificate case). ✅ PASS
  - Acceptance test — TestAccCloudFrontDistributionTenant_managedCertificateRequest (create with managed_certificate_request, assert empty follow-up plan). ✅ PASS
  - go vet passes; gofmt clean.
  - Verified end-to-end against a live AWS account across three scenarios: CloudFront-validated managed cert (own cert issued), self-hosted-validated managed cert (own cert issued), and a wildcard-covered domain (tenant reuses the distribution's shared certificate,
  no per-tenant cert). All three: clean apply + empty follow-up plan.

  Plan: 1 to add, 0 to change, 0 to destroy.
  aws_cloudfront_distribution_tenant.repro: Creating...
  aws_cloudfront_distribution_tenant.repro: Still creating... [00m10s elapsed]
  ...
  aws_cloudfront_distribution_tenant.repro: Creation complete after 1m17s [id=xxxxxxxxxxxxxxxxxxxx]

  Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

  aws_cloudfront_distribution_tenant.repro: Refreshing state... [id=xxxxxxxxxxxxxxxxxxxx]

  No changes. Your infrastructure matches the configuration.

## Rollback Plan

##  AI Usage Disclosure

  AI tools (Claude) were used to assist with code analysis, implementation, and test authoring. All code was reviewed, understood, and verified end-to-end by the human contributor.

  ## Relations

  Closes #45987

##  References

  - Issue #45987

  ## Output from Acceptance Testing

  yulanyi@842f578f15d1 terraform-provider-aws % GOPROXY=direct GOSUMDB=off TF_ACC=1 AWS_PROFILE=default \
    ACM_CERTIFICATE_ROOT_DOMAIN="xxxxxxxxxxxxxxxxxxxx.aws.dev" \
    /opt/homebrew/bin/go test ./internal/service/cloudfront/ -v \
    -run 'TestAccCloudFrontDistributionTenant_managedCertificateRequest' -timeout 120m
2026/06/15 16:04:57 Creating Terraform AWS Provider (SDKv2-style)...
2026/06/15 16:04:57 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccCloudFrontDistributionTenant_managedCertificateRequest
=== PAUSE TestAccCloudFrontDistributionTenant_managedCertificateRequest
=== CONT  TestAccCloudFrontDistributionTenant_managedCertificateRequest
--- PASS: TestAccCloudFrontDistributionTenant_managedCertificateRequest (605.53s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/cloudfront 611.997s


internal/service/cloudfront/distribution_tenant.go
```diff
@@ -354,6 +360,13 @@ func (r *distributionTenantResource) Create(ctx context.Context, req resource.Cr
 		}
 	}
 
+	// Restore planned customizations to maintain plan/state consistency.
+	// The API auto-populates customizations.certificate when managed_certificate_request
+	// is used. Restore the planned value when customizations was entirely null
+	if managedCertRequested && plannedCustomizations.IsNull() {
+		data.Customizations = plannedCustomizations
+	}
+
```
tony-yu1, line 368: The restore patch here feels... hacky. But that was the solution i could thought of with the least blast radius, I felt like.

The solution I was hoping for is to mark the field injected by Cloudflare API as "Computed", but the plugin only allows attributes to be Computed, not blocks (which is what the certificate that's being injected is)

I'd love to hear any suggestions!

tony-yu1: I think this is the right direction of what to do, `glue/catalog.go:634-647` could be a prior similar example

internal/service/cloudfront/distribution_tenant_customizations_test.go
```diff
@@ -0,0 +1,170 @@
+// Copyright IBM Corp. 2014, 2026
```
gahykiamzn, line 1: I don't think we want a separate file for this testing, let's merge the two test files

internal/service/cloudfront/distribution_tenant.go
gahykiamzn: I wonder if there's opportunity here to maybe provide a generalized solution for this in the Compute method. Might be worth exploring

internal/service/cloudfront/distribution_tenant_customizations_test.go
tony-yu1: I think it should be okay to keep the unit tests & acceptance tests in separate files? The unit tests use `package cloudfront`, and the acceptance tests use `package cloudfront_test`. There's this distinction because the acceptance tests run end to end with actual AWS resources, while the unit tests are just simple white box tests. There's also similar unit test file examples in other services such as route53's unit tests: `route53/record_flex_test.go`, `route53/clean_test.go`, and `route53/equal_test.go`

Please let me know what you think 🙏

skalugin:
my customer hit this in production

internal/service/cloudfront/distribution_tenant.go
gahykiamzn: TLDR; this works for now. Terraform should look into creating a generalized solution for this case.

mayrop:
It will be very nice if this one gets merged/fixed.

PR #48352, opened 2026-06-11
r/aws_rds_cluster: Add auto_minor_version_upgrade argument

### LINKED ISSUE #38539: [Enhancement]: AutoMinorVersionUpgrade for non-Aurora rds_cluster
### Description

The AutoMinorVersionUpgrade attribute can used to control whether a Multi-AZ DB Cluster has minor version upgrades automatically applied.

* aws_rds_cluster

```terraform
resource "aws_rds_cluster" "cluster" {
    ...
    auto_minor_version_upgrade = true
    ...
}
```

### References

https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html#:~:text=Required%3A%20No-,AutoMinorVersionUpgrade,-Specifies%20whether%20minor

None

mattrobinsonsre:
### Description

Adds an `auto_minor_version_upgrade` argument to the `aws_rds_cluster` resource. The underlying `CreateDBCluster`, `ModifyDBCluster`, and `DescribeDBClusters` APIs already carry `AutoMinorVersionUpgrade` for both Aurora DB clusters and Multi-AZ DB clusters; the resource simply never wired it up, so the setting could only be managed out-of-band (CLI/console).

Design notes:

- The attribute is **`Optional + Computed` with no `Default`**. AWS defaults `AutoMinorVersionUpgrade` to `true`, so an unset configuration must not force a value. On create it is sent only when explicitly set (`GetOkExists`), it is read back as Computed, and update is guarded by `HasChange`. Net effect: **an unset attribute makes no change and produces no diff**, so existing clusters (which read back `true`) do not drift after upgrading the provider.
- No client-side validation of cluster-type validity — the value is passed through and the API remains the authority, consistent with the existing engine/mode-specific arguments on this resource (`allocated_storage`, `db_cluster_instance_class`, etc.).
- The restore-from-snapshot / S3 / point-in-time create inputs do **not** carry `AutoMinorVersionUpgrade` in the SDK, so it is not wired into those branches; on restored clusters the value is read back as Computed and can be set on a subsequent apply via the update path. Happy to add post-restore deferral (the existing `modifyDbClusterInput` pattern) if preferred.
- Docs note the interaction with a pinned full `engine_version` (pinning a full minor while auto-minor is enabled produces a persistent plan diff; use a major-only `engine_version`, or set `auto_minor_version_upgrade = false`).

### Relations

Closes #38539

### References

* [`CreateDBCluster` — `AutoMinorVersionUpgrade`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html)

New tests: `TestAccRDSCluster_autoMinorVersionUpgrade` (Aurora MySQL: create `false` → import → update `true`, asserting in-place modify), `TestAccRDSCluster_AutoMinorVersionUpgrade_multiAZ` (non-Aurora Multi-AZ DB cluster create). The existing `TestAccRDSCluster_basic` re-plan covers the unset-no-diff invariant.

```console
# pending — to be run by the author against a real account:
# make testacc PKG=rds TESTS='TestAccRDSCluster_autoMinorVersionUpgrade|TestAccRDSCluster_AutoMinorVersionUpgrade_multiAZ|TestAccRDSCluster_basic'
```

> [!NOTE]
> Draft until acceptance test output is attached.


mattrobinsonsre:
Closing as a duplicate of #42472, which already implements this with an equivalent approach (Optional+Computed, `GetOkExists` on create, `HasChange` on update). Apologies for the noise — I missed the existing PR before opening this.

Issue #3382, opened 2018-02-14
using aws_vpc_endpoint_subnet_association with count for multiple subnet hangs


GElkayam:
Hello,

I've created a vpc endpoint for using ec2 api, and tried to add aws_vpc_endpoint_subnet_association so it will be in multiple AZs.

I've configured the aws_vpc_endpoint as a resource without any subnets, and added a aws_vpc_endpoint_subnet_association resource with a list of subnets with count, after subnets are in place.

### Terraform Configuration Files
```hcl
resource "aws_vpc_endpoint_subnet_association" "ec2_private-elb-main" {
  count           = "${length(module.subnet__private-elb.subnet_ids)}"
  vpc_endpoint_id = "${module.vpc__main.ep_ec2_id}"
  subnet_id       = "${module.subnet__private-elb.subnet_ids[count.index]}"
}
```

in two different accounts that I've tried that it created only one aws_vpc_endpoint_subnet_association, and hanged for the rest.

aws_vpc_endpoint_subnet_association.ec2_private-elb-main[2]: Creating...
  subnet_id:       "" => "subnet-12341ec2"
  vpc_endpoint_id: "" => "vpce-12343dccc075156d3"
aws_vpc_endpoint_subnet_association.ec2_private-elb-main[0]: Creating...
  subnet_id:       "" => "subnet-1234b585"
  vpc_endpoint_id: "" => "vpce-12343dccc075156d3"
aws_vpc_endpoint_subnet_association.ec2_private-elb-main[1]: Creating...
  subnet_id:       "" => "subnet-1234d612"
  vpc_endpoint_id: "" => "vpce-12343dccc075156d3"
aws_vpc_endpoint_subnet_association.ec2_private-elb-main[2]: Creation complete after 2s (ID: a-vpce-12343dccc075156d31244978879)
aws_vpc_endpoint_subnet_association.ec2_private-elb-main.0: Still creating... (10s elapsed)
aws_vpc_endpoint_subnet_association.ec2_private-elb-main.1: Still creating... (10s elapsed)
aws_vpc_endpoint_subnet_association.ec2_private-elb-main.0: Still creating... (20s elapsed)
aws_vpc_endpoint_subnet_association.ec2_private-elb-main.1: Still creating... (20s elapsed)
aws_vpc_endpoint_subnet_association.ec2_private-elb-main.0: Still creating... (30s elapsed)
.
.
.
[...]

PR #48364, opened 2026-06-11
🤖🤖🤖 r/aws_vpc_endpoint_subnet_association: retry OperationInProgress on ModifyVpcEndpoint

JuanCanham:
## Rollback Plan

None.

### Description

When multiple `aws_vpc_endpoint_subnet_association` resources targeting the same VPC endpoint are created or destroyed in parallel, AWS serializes `ModifyVpcEndpoint` calls per endpoint and returns `OperationInProgress` to any call that arrives while a previous one is still in flight. The provider currently surfaces this as a hard error, forcing users to chain resources with `depends_on` to serialize operations.

The existing `modifyVPCEndpointExclusive` helper already uses a `GlobalMutexKV` to serialize calls within a single Terraform process (added in #3382). However that mutex provides no protection across separate processes or concurrent Terraform runs. This change handles `OperationInProgress` inside the `StateChangeConf.Refresh` closure by returning `(nil, "", nil)`, which causes the existing wait loop to treat it as "not ready yet" and keep polling — consuming no extra timeout budget and requiring no new imports or helpers.

**AI disclosure:** This change was researched and drafted with Claude Code (Claude Opus 4.7). The author reviewed and owns the clanker generated code.

### Relations

Relates #3382

### References

- Original mutex addition: #3382
- Existing in-package precedent for `OperationInProgress` retry on `ModifyVpcEndpoint`: `internal/service/ec2/vpc_subnet.go` (GuardDuty VPC endpoint dissociation path)
- `StateChangeConf` NotFound-counter behavior verified against `internal/retry/state.go`

```console
% make testacc TESTS=TestAccVPCEndpointSubnetAssociation_ PKG=ec2
```

_(Acceptance tests require live AWS credentials — pending CI run)_

### DIFF
--- a/.changelog/48364.txt
+++ b/.changelog/48364.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_vpc_endpoint_subnet_association: Fix `OperationInProgress` errors when multiple subnet associations targeting the same VPC endpoint are created or destroyed in parallel
+```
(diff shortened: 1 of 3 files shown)


justinretzolk:
Similar #42884

PR #48396, opened 2026-06-13
r/aws_api_gateway_rest_api: apply `binary_media_types` removal when argument is deleted

### LINKED ISSUE #48385: `aws_api_gateway_rest_api` does not remove `binary_media_types` when attribute is deleted from config

```console
Terraform v1.11.4
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v6.47.0
```

* `aws_api_gateway_rest_api`

### Expected Behavior

Removing `binary_media_types` from an `aws_api_gateway_rest_api` resource should produce a plan that clears the list on the deployed API (i.e. set it to `[]`). Subsequent `terraform plan` should show no diff once applied.

### Actual Behavior

After removing `binary_media_types` from the config and running `terraform apply`, the attribute persists on the deployed REST API in AWS. Subsequent `terraform plan` reports no changes — Terraform does not detect the drift between config (no `binary_media_types`) and actual state (list still set).

```console

```

<details open>

```hcl
resource "aws_api_gateway_rest_api" "example" {
  name                = "example"
  binary_media_types  = ["application/gzip"]

  endpoint_configuration {
    types = ["REGIONAL"]
  }
}
```
</details>

### Steps to Reproduce

1. Run `terraform apply` — `binary_media_types` is set on the API.
2. Remove `binary_media_types` from the resource block entirely.
3. Run terraform plan — no changes are planned.
4. Check the deployed API via AWS CLI:
```
aws apigateway get-rest-api --rest-api-id <id> --query 'binaryMediaTypes'
# Returns: ["application/gzip"]  ← still set
```

### Debug Logging

<details open>

```console

```
</details>

n/a

_No response_

No

Sagargupta16:
## Rollback Plan

### Description

`binary_media_types` is `Optional + Computed` so that values derived from an imported OpenAPI `body` are not treated as drift. That combination has a side effect: removing the argument from configuration is a no-op. The prior value is retained from state, no diff is planned, and the media types remain on the deployed REST API indefinitely.

This adds a plan-time `CustomizeDiff` (`binaryMediaTypesPlantimeClear`) that, **only when no `body` is configured**, clears the planned `binary_media_types` value if the argument is absent from configuration but a non-empty value exists in state. The existing update logic already emits the `OpRemove` patch operations once a diff is present, so no change to the update path is needed. The `body`/OpenAPI-import path is explicitly skipped, preserving the Computed behavior there.

The new diff function is composed with the existing `endpointConfigurationPlantimeValidate` via `customdiff.All`, and guards against the null raw-config seen during destroy (same pattern as `resourceRestAPIPolicyConfigured`).

### Relations

Closes #48385

Added a removal step to `TestAccAPIGatewayRestAPI_binaryMediaTypes`: after setting a media type, the config drops the argument entirely; the step asserts `binary_media_types.# = 0` with a plan-check expecting an `Update` action (this step shows no diff / no update on the current code, demonstrating the bug). The existing `_overrideBody` test continues to exercise the unchanged `body` path.

```
%  make testacc TESTS=TestAccAPIGatewayRestAPI_binaryMediaTypes PKG=apigateway
```

`go build`, `go vet`, and `gofmt` pass; the test package compiles and runs green locally. I don't have an AWS account suitable for the live acceptance run -- happy to iterate on the maintainer-triggered run.

### DIFF
--- a/.changelog/48385.txt
+++ b/.changelog/48385.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_api_gateway_rest_api: Fix `binary_media_types` removal being silently ignored when the argument is deleted from configuration without `body`
+```
--- a/internal/service/apigateway/rest_api.go
+++ b/internal/service/apigateway/rest_api.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/apigateway/types"
 	awspolicy "github.com/hashicorp/awspolicyequivalence"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/structure"
 	"github.com/hashicorp/terraform-provider-aws/internal/conns"
@@ -183,7 +184,10 @@ func resourceRestAPI() *schema.Resource {
 			}
 		},
 
-		CustomizeDiff: endpointConfigurationPlantimeValidate,
+		CustomizeDiff: customdiff.All(
+			endpointConfigurationPlantimeValidate,
+			binaryMediaTypesPlantimeClear,
+		),
 	}
 }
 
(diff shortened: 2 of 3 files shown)


AdamTylerLynch (changes requested):
This is a really great pull requests, thank you for your contribution! 

I would like to see AccTests for all tests that use an OpenAPI body for completeness, not just the one test the author modified.

If you are concerned with the cost of running the tests, just comment here and I’ll run them for you over the weekend.

Sagargupta16:
Added the requested coverage. Rather than touch every body test, I extended the two that exercise `binary_media_types` through an OpenAPI body, since that is where the new plan-time clear has to skip:

- `TestAccAPIGatewayRestAPI_BinaryMediaTypes_overrideBody`: added a final step that drops the `binary_media_types` argument while the body still sets it, asserting the value is retained (not force-cleared). This is the regression the `body`-skip guard protects against.
- `TestAccAPIGatewayRestAPI_BinaryMediaTypes_setByBody`: added a step that updates the body-derived media type, asserting the Computed value still tracks the body.

Both reuse the existing `*SetByBody` config helper, so no new fixtures. `gofmt`, `go vet`, and a compile of the apigateway test package all pass.

I still don't have an AWS account set up for the live run, so I'd gladly take you up on the offer to run `TestAccAPIGatewayRestAPI_binaryMediaTypes`, `_BinaryMediaTypes_overrideBody`, and `_BinaryMediaTypes_setByBody` over the weekend.

PR #48398, opened 2026-06-13
Add SSO Admin region resource

### LINKED ISSUE #48397: Add 'aws_ssoadmin_region' resource
### What new functionality are you requesting?

* `aws_ssoadmin_region`

### Description

Add a new `aws_ssoadmin_region` resource to manage additional AWS Regions for an IAM Identity Center instance.

This resource would allow Terraform to add, read, import, and remove IAM Identity Center Regions using the SSO Admin Region management APIs. It would manage the upstream IAM Identity Center Region association identified by an `instance_arn` and `region_name`.

```hcl
data "aws_ssoadmin_instances" "example" {}

resource "aws_ssoadmin_region" "example" {
  instance_arn = one(data.aws_ssoadmin_instances.example.arns)
  region_name  = "us-west-2"
}
```

### References

AWS SDK for Go v2 SSO Admin supports the Region management APIs needed for this resource:

* `AddRegion`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.AddRegion
* `DescribeRegion`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.DescribeRegion
* `RemoveRegion`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.RemoveRegion
* `ListRegions`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.ListRegions

The SDK package index lists these operations under the SSO Admin client. See the SSO Admin package documentation: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin

Yes

pooriaghaedi:
## Rollback Plan

This pull request adds support for managing IAM Identity Center Region configuration. It does not change provider-side access controls, encryption, or logging behavior. Users will need IAM permissions for the relevant SSO Admin Region management APIs, such as `sso:AddRegion`, `sso:DescribeRegion`, and `sso:RemoveRegion`.

### Description

Adds the `aws_ssoadmin_region` resource for managing additional AWS Regions for an IAM Identity Center instance.

The resource supports adding an additional Region, reading Region status, removing the Region, and importing existing Region associations using `instance_arn,region_name`.

### Relations

Closes #48397

### References

* AWS SDK for Go v2 SSO Admin `AddRegion`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.AddRegion
* AWS SDK for Go v2 SSO Admin `DescribeRegion`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.DescribeRegion
* AWS SDK for Go v2 SSO Admin `RemoveRegion`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.RemoveRegion
* AWS SDK for Go v2 SSO Admin `ListRegions`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssoadmin#Client.ListRegions

```console
% go1.26.3 test -vet=off -p=1 -run '^$' ./internal/service/ssoadmin

ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ssoadmin	7.011s [no tests to run]

### DIFF
--- a/.changelog/48397.txt
+++ b/.changelog/48397.txt
@@ -1,0 +1,3 @@
+```release-note:new-resource
+aws_ssoadmin_region
+```
--- a/internal/service/ssoadmin/exports_test.go
+++ b/internal/service/ssoadmin/exports_test.go
@@ -18,6 +18,7 @@ var (
 	ResourcePermissionsBoundaryAttachment             = resourcePermissionsBoundaryAttachment
 	ResourcePermissionSet                             = resourcePermissionSet
 	ResourcePermissionSetInlinePolicy                 = resourcePermissionSetInlinePolicy
+	ResourceRegion                                    = newRegionResource
 	ResourceTrustedTokenIssuer                        = newTrustedTokenIssuerResource
 
 	FindAccountAssignmentByFivePartKey               = findAccountAssignmentByFivePartKey
@@ -33,5 +34,6 @@ var (
 	FindPermissionsBoundaryByTwoPartKey              = findPermissionsBoundaryByTwoPartKey
 	FindPermissionSetByTwoPartKey                    = findPermissionSetByTwoPartKey
 	FindPermissionSetInlinePolicyByTwoPartKey        = findPermissionSetInlinePolicyByTwoPartKey
+	FindRegionByID                                   = findRegionByID
 	FindTrustedTokenIssuerByARN                      = findTrustedTokenIssuerByARN
 )
--- a/internal/service/ssoadmin/region.go
+++ b/internal/service/ssoadmin/region.go
@@ -1,0 +1,324 @@
+// Copyright IBM Corp. 2014, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package ssoadmin
+
+import (
+	"context"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ssoadmin"
+	awstypes "github.com/aws/aws-sdk-go-v2/service/ssoadmin/types"
+	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
+	"github.com/hashicorp/terraform-plugin-framework-timetypes/timetypes"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-provider-aws/internal/create"
+	"github.com/hashicorp/terraform-provider-aws/internal/enum"
+	"github.com/hashicorp/terraform-provider-aws/internal/errs"
+	intflex "github.com/hashicorp/terraform-provider-aws/internal/flex"
+	"github.com/hashicorp/terraform-provider-aws/internal/framework"
+	"github.com/hashicorp/terraform-provider-aws/internal/framework/flex"
+	fwtypes "github.com/hashicorp/terraform-provider-aws/internal/framework/types"
+	"github.com/hashicorp/terraform-provider-aws/internal/retry"
+	"github.com/hashicorp/terraform-provider-aws/internal/tfresource"
+	"github.com/hashicorp/terraform-provider-aws/names"
+)
+
+// @FrameworkResource("aws_ssoadmin_region", name="Region")
+// @Testing(preCheck="github.com/hashicorp/terraform-provider-aws/internal/acctest;acctest.PreCheckSSOAdminInstances")
+func newRegionResource(_ context.Context) (resource.ResourceWithConfigure, error) {
+	r := &regionResource{}
+
+	r.SetDefaultCreateTimeout(120 * time.Minute)
+	r.SetDefaultDeleteTimeout(120 * time.Minute)
+
+	return r, nil
+}
+
+const (
+	ResNameRegion = "Region"
+
+	regionIDPartCount = 2
+)
+
+type regionResource struct {
+	framework.ResourceWithModel[regionResourceModel]
+	framework.WithImportByID
+	framework.WithTimeouts
+}
+
+func (r *regionResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by danbaugher** [comment]
Is this a duplicate change? https://github.com/hashicorp/terraform-provider-aws/pull/48126

**Review by tsushanth — CHANGES_REQUESTED** [review]
Thanks for taking on `aws_ssoadmin_region` — the resource fills a real gap. Before diving into code feedback, the most important piece of context: there's an open competing PR for the same resource, and several substantive concerns in this implementation worth resolving before merge.

### Context: #48126 covers the same resource and is further along

[#48126](https://github.com/hashicorp/terraform-provider-aws/pull/48126) by @danbaugher adds the same `aws_ssoadmin_region` resource. He flagged this on 2026-06-13 ("Is this a duplicate change?") — no response in the thread yet. Some material differences worth surfacing:

- **Resource identity testing**: #48126 includes `region_identity_gen_test.go` (224 LOC) and `testdata/Region/basic/main_gen.tf` + `testdata/Region/region_override/main_gen.tf` + `testdata/tmpl/region_basic.gtpl`. This is the newer TF-AWS resource-identity test generator output; PRs that don't include it usually get asked to add it during maintainer review (recent precedent in the lake-formation and IAM space). This PR has no equivalent.
- **Code volume**: #48126 is 838 LOC vs this one's 567 — most of the gap is the identity test scaffolding above.

Coordinating with @danbaugher before iterating further would save duplicated work — either rebase the schema/Create/Read/Delete improvements you've made on top of his identity-test scaffolding, or move the conversation to a single PR. Reviewing both in parallel risks two near-identical resources landing at different times with subtle schema drift.

### Must-fix

**1. `Update` clobbers computed attributes when called.**

```go
func (r *regionResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
    var plan regionResourceModel
    resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
    if resp.Diagnostics.HasError() {
        return
    }

    resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
}
```

Both `instance_arn` and `region_name` are `RequiresReplace`, so Update *seems* unreachable — but it's not. Update fires when only non-replacing attributes change (e.g. `timeouts.create`). When that happens, `plan` carries Unknown for the `Computed: true` attributes (`added_date`, `is_primary_region`, `status`) because there's no `UseStateForUnknown` plan modifier on them. The current implementation writes those Unknowns straight into state, blanking out values that were correct on the prior refresh.

Two fixes, either works:

- **(a)** Add `stringplanmodifier.UseStateForUnknown()` (and the `boolplanmodifier` / `timetypes` equivalents) to `added_date`, `is_primary_region`, and `status`. With that, the plan carries the prior state's values forward when nothing's actually changed at the backend, and the no-op Update becomes safe.
- **(b)** Have Update do a `findRegionByID` + `flex.Flatten` to refresh the computed attrs from the backend before writing state. More API traffic, but bulletproof.

I'd recommend (a) — it's the standard TF-AWS pattern for "this attribute is computed at create time and doesn't drift."

**2. Delete has no handling for the primary region.**

```go
func (r *regionResource) Delete(...) {
    ...
    _, err := conn.RemoveRegion(ctx, input)
    if errs.IsA[*awstypes.ResourceNotFoundException](err) {
        return
    }
    if err != nil {
        resp.Diagnostics.AddError(...)
        return
    }
    ...
}
```

`is_primary_region` is exposed as a computed bool on the resource, which implies the resource *can* be a primary region. AWS rejects `RemoveRegion` on the primary region of an IAM Identity Center instance. If a user imports the primary region into Terraform and then runs `terraform destroy`, this code path returns the raw AWS error, which is fine — but the docs/example should explicitly call out that the primary region can be managed but not removed, and the resource should probably refuse to even attempt deletion when `state.IsPrimaryRegion.ValueBool()` is true (with a clear diagnostic explaining why). Otherwise users will hit a confusing API-level rejection during destroy.

If the resource is meant to *only* model non-primary regions, then `is_primary_region` doesn't need to be a computed attribute at all — it would always be `false`, and the schema could omit it (or add a `Read`-time check that errors out if AWS reports `is_primary_region == true`, preventing accidental import).

**3. No acceptance tests were actually run.**

The "Output from Acceptance Testing" block in the PR body:

```console
% go1.26.3 test -vet=off -p=1 -run '^$' ./internal/service/ssoadmin
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/ssoadmin	7.011s [no tests to run]
```

`-run '^$'` matches no test names; "no tests to run" is the literal output. The tests defined in `region_test.go` (`TestAccSSOAdminRegion_serial/basic` and `/disappears`) were never executed against AWS. The maintainer team will need to run them before merge — worth saying so explicitly in the PR body rather than implying tests passed.

### Worth discussing

**A.** `regionResourceModel.AddedDate` typed as `timetypes.RFC3339`, but AWS SDK's `DescribeRegionOutput.CreatedDate` (the underlying field) is typically `*time.Time`. The `flex.Flatten` machinery handles this — but worth confirming via a unit test (or at least manual verification) that the resulting Terraform state value is what users expect (RFC3339 string, not Unix epoch). Easy to miss if the underlying field name differs slightly from the schema field name.

**B.** The 120-minute create/delete timeouts feel generous for what's typically a sub-minute operation on AWS's side. If there's a documented case where `AddRegion` or `RemoveRegion` actually takes that long (e.g. cross-region replication setup), worth a comment at `r.SetDefaultCreateTimeout(120 * time.Minute)` explaining what scenario justifies the 2h cap. Otherwise, dropping to 30m would be more in line with similar SSO Admin resources in the codebase.

**C.** `WithRegionModel` is embedded in `regionResourceModel`, which exposes a `region` attribute on the resource — separately from `region_name`, which is the *target* region being added. Users will hit a confusion point here: "which region is which?" A note in the docs (and possibly a different schema name to avoid the `region` / `region_name` collision visually) would help.

### Decision

Requesting changes. The Update state-clobber on computed attrs is a real bug that will surface as soon as someone has a non-trivial plan diff; the primary-region Delete path is a UX hazard waiting to happen; and the dueling PR situation with #48126 needs to be resolved one way or the other before either lands. The identity-test scaffolding gap is something maintainers will ask for anyway — easier to add it now than in the third round of review.

Happy to re-review once 1–3 are addressed and the coordination with @danbaugher's PR is settled.

justinretzolk:
Hey @pooriaghaedi  👋 Thanks for taking the time to raise this! It looks like this is a duplicate of #48126. Since that pull request was submitted earlier and is a bit more complete, I'm going to close this pull request in favor of that one.

PR #48400, opened 2026-06-15
New Resource: aws_lexv2models_bot_alias

### LINKED ISSUE #35780: [New]: LexV2 Bot Alias
### Description

Add resource `bot_alias` 
API docs https://docs.aws.amazon.com/lexv2/latest/APIReference/API_CreateBotAlias.html

### Requested Resource(s) and/or Data Source(s)

aws_lexv2models_bot_alias

_No response_

### References

_No response_

None

CaseyLeask:
## Rollback Plan

No changes to security controls. The resource manages AWS Lex V2 bot aliases; conversation log destinations (S3/CloudWatch) and Lambda code hook ARNs are user-supplied and validated as ARNs.

### Description

Adds a new resource `aws_lexv2models_bot_alias` for managing [Amazon Lex V2 bot aliases](https://docs.aws.amazon.com/lexv2/latest/dg/aliases.html). An alias points to a specific version of a bot so client applications can use a stable name to invoke that version.

The resource is built on the Terraform Plugin Framework and supports:

- `bot_alias_name`, `bot_id`, `bot_version`, and `description`
- `bot_alias_locale_settings` — per-locale enablement and Lambda `code_hook_specification`
- `conversation_log_settings` — `audio_log_settings` (S3 destination) and `text_log_settings` (CloudWatch destination)
- `sentiment_analysis_settings`
- Tagging (`identifierAttribute = arn`)
- Import by `bot_id,bot_alias_id`
- A sweeper, registered as a dependency of the `aws_lexv2models_bot` sweeper

### Relations

Closes #35780

### References

* [Amazon Lex V2 — Versions and aliases](https://docs.aws.amazon.com/lexv2/latest/dg/aliases.html)
* [`CreateBotAlias` API reference](https://docs.aws.amazon.com/lexv2/latest/APIReference/API_CreateBotAlias.html)

```console
% make testacc TESTS=TestAccLexV2ModelsBotAlias_ PKG=lexv2models

--- PASS: TestAccLexV2ModelsBotAlias_conversationLogSettings (51.12s)
--- PASS: TestAccLexV2ModelsBotAlias_disappears (52.64s)
--- PASS: TestAccLexV2ModelsBotAlias_basic (57.44s)
--- PASS: TestAccLexV2ModelsBotAlias_description (74.20s)
--- PASS: TestAccLexV2ModelsBotAlias_sentimentAnalysisSettings (74.88s)
--- PASS: TestAccLexV2ModelsBotAlias_tags (98.56s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/lexv2models
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changelog/48400.txt
+++ b/.changelog/48400.txt
@@ -1,0 +1,3 @@
+```release-note:new-resource
+aws_lexv2models_bot_alias
+```
--- a/internal/service/lexv2models/bot_alias.go
+++ b/internal/service/lexv2models/bot_alias.go
@@ -1,0 +1,750 @@
+// Copyright IBM Corp. 2014, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+// DONOTCOPY: Copying old resources spreads bad habits. Use skaff instead.
+
+package lexv2models
+
+// **PLEASE DELETE THIS AND ALL TIP COMMENTS BEFORE SUBMITTING A PR FOR REVIEW!**
+//
+// TIP: ==== INTRODUCTION ====
+// Thank you for trying the skaff tool!
+//
+// You have opted to include these helpful comments. They all include "TIP:"
+// to help you find and remove them when you're done with them.
+//
+// While some aspects of this file are customized to your input, the
+// scaffold tool does *not* look at the AWS API and ensure it has correct
+// function, structure, and variable names. It makes guesses based on
+// commonalities. You will need to make significant adjustments.
+//
+// In other words, as generated, this is a rough outline of the work you will
+// need to do. If something doesn't make sense for your situation, get rid of
+// it.
+
+import (
+	// TIP: ==== IMPORTS ====
+	// This is a common set of imports but not customized to your code since
+	// your code hasn't been written yet. Make sure you, your IDE, or
+	// goimports -w <file> fixes these imports.
+	//
+	// The provider linter wants your imports to be in two groups: first,
+	// standard library (i.e., "fmt" or "strings"), second, everything else.
+	//
+	// Also, AWS Go SDK v2 may handle nested structures differently than v1,
+	// using the services/lexmodelsv2/types package. If so, you'll
+	// need to import types and reference the nested types, e.g., as
+	// awstypes.<Type Name>.
+	"context"
+	"errors"
+	"time"
+
+	"github.com/YakDriver/smarterr"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/lexmodelsv2"
+	awstypes "github.com/aws/aws-sdk-go-v2/service/lexmodelsv2/types"
+	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-provider-aws/internal/conns"
+	"github.com/hashicorp/terraform-provider-aws/internal/errs"
+	"github.com/hashicorp/terraform-provider-aws/internal/errs/fwdiag"
+	"github.com/hashicorp/terraform-provider-aws/internal/framework"
+	"github.com/hashicorp/terraform-provider-aws/internal/framework/flex"
+	fwtypes "github.com/hashicorp/terraform-provider-aws/internal/framework/types"
+	"github.com/hashicorp/terraform-provider-aws/internal/retry"
+	"github.com/hashicorp/terraform-provider-aws/internal/smerr"
+	"github.com/hashicorp/terraform-provider-aws/internal/sweep"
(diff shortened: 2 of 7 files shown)


tsushanth:
Solid new-resource PR. The structure mirrors the surrounding `lexv2models` resources well (composite ID via `intflex.FlattenResourceId`, two-part `findByTwoPartKey` pattern, locally-constructed ARN since `DescribeBotAlias` doesn't return one), and a few details stand out as actively good engineering: `ContinuousTargetOccurence: 2` on the waiter to filter transient-Available reads, the inline comment at `waitBotAliasUpdated` documenting AWS's "no distinct Updating status" API quirk, and the Update path's explicit per-attribute change-detection (avoiding `UpdateBotAlias` calls when nothing the API cares about has actually changed). A few items worth surfacing before merge.

### Worth verifying

**1. The `$LATEST` alias will break the sweeper.**

AWS auto-creates a `$LATEST` (sometimes `TestBotAlias`) alias for every Lex V2 bot, and that alias cannot be deleted via `DeleteBotAlias` — the API rejects with `PreconditionFailedException`. The new `sweepBotAliases` in `internal/service/lexv2models/sweep.go` enumerates *all* aliases returned by `ListBotAliases` and adds each one to `sweepResources`:

```go
for _, a := range aliasPage.BotAliasSummaries {
    botAliasID := aws.ToString(a.BotAliasId)
    id, _ := intflex.FlattenResourceId([]string{botID, botAliasID}, botAliasResourceIDPartCount, false)
    // ...adds to sweepResources unconditionally
}
```

The resource's `Delete` does already handle the precondition-failed-"does not exist" case (line ~470) — but the `$LATEST` error message is different (typically *"cannot delete the $LATEST alias"* or similar). When the sweeper hits it in CI, it'll fail the whole region's sweep run. Two clean options:

- **(a)** Skip `$LATEST` (and any other reserved-name aliases) by name in the sweeper loop:
  ```go
  if name := aws.ToString(a.BotAliasName); name == "$LATEST" || name == "TestBotAlias" {
      continue
  }
  ```
- **(b)** Widen the precondition-error message match in `Delete` to also accept the `$LATEST`-specific message, so the sweeper's call paths through `newBotAliasResource` succeed silently.

(a) is more conservative — won't accidentally swallow real errors. I'd lean (a).

**2. The primary alias workflow isn't covered by acceptance tests.**

The whole point of bot aliases is **repointing them at a different bot version** as part of a deployment workflow. The current tests cover:

- `_basic` (create + import)
- `_tags` (tag CRUD)
- `_disappears`
- `_description` (string field update)
- `_conversationLogSettings` (nested block create-only)
- `_sentimentAnalysisSettings` (toggle `detect_sentiment` bool)

Missing:

- **`_botVersionUpdate`** — create alias pointing at version N, then update config to point at version N+1, verify the alias's `bot_version` attribute reflects the new value and no replace happens. This is the canonical use case of the resource and is currently untested.
- **`_botAliasLocaleSettings`** — the schema has the most surface area here (`code_hook_specification` → `lambda_code_hook` → `code_hook_interface_version` + `lambda_arn`), and the validators (`IsRequired() + SizeAtLeast(1) + SizeAtMost(1)` on `lambda_code_hook`) are doing real work that should be regression-guarded. Currently zero coverage.

Neither is a merge blocker on its own, but the `_botVersionUpdate` gap is conspicuous given that "repoint at a new version" is what an alias *is*.

### Worth a comment-pass

**A.** Identity-test scaffolding missing.

No `bot_alias_identity_gen_test.go` and no `testdata/BotAlias/*/main_gen.tf`. This is the newer TF-AWS resource-identity test infrastructure (visible on recently-landed resources in lake-formation, sso-admin, etc., and the gap I flagged on #48398). Maintainers have been asking for it on new resources during review — easier to add it preemptively than in the third round.

**B.** Delete-path error-message string matching is brittle.

```go
if errs.IsA[*awstypes.ResourceNotFoundException](err) ||
    errs.IsAErrorMessageContains[*awstypes.PreconditionFailedException](err, "does not exist") {
    return
}
```

The `"does not exist"` substring is doing real work — distinguishing "the alias is already gone, that's fine" from other `PreconditionFailedException` cases (which include `$LATEST` and concurrent-modification errors). AWS could change the message string in a future API revision and this would silently start treating unrelated errors as success. Common TF-AWS pattern, not unique to this PR, but if there's an SDK error code or a stable sentinel (looking at `awstypes`, there isn't one — fine, just flagging).

**C.** `botAliasResourceModel` uses `MapBlockKey` tag on `BotAliasLocaleSettings.MapBlockKey` (`tfsdk:"locale_id"`).

That's the `SetNestedBlock` convention for treating one of the attributes as the natural-key for set semantics. Worth a short comment at the field — to the next reader, `MapBlockKey` looks like a misnomer (it's a *set* block, not a map) and the connection to `locale_id` isn't obvious from the type name alone.

### Worth acknowledging

A few things this PR does *better* than the average new-resource PR in tf-aws and worth not regressing in review iteration:

- The Update path's `if !new.X.Equal(old.X) || ...` chain is explicit and avoids unnecessary `UpdateBotAlias` calls when only Terraform-managed metadata (e.g. tags) changed. Some recently-landed resources just call Update unconditionally on every plan.
- The `waitBotAliasUpdated` comment documenting AWS's "Creating-during-Update" quirk is the right level of detail — exactly what saves the next maintainer from a wild-goose chase.
- Sweeper dependency declaration (`Dependencies: []string{"aws_lexv2models_bot_alias"}` on the bot sweeper) is correct ordering — aliases must be swept before their parent bot.

### Decision

Requesting changes — but only because of item 1 (the `$LATEST` sweeper bug). The resource itself is well-constructed and the structural patterns are sound; if the sweeper guard lands, this is a clean approve. Items 2 (test coverage) and the comment-pass items are quality-of-merge polish and shouldn't gate the merge by themselves.

Happy to re-review once the `$LATEST` filter is in.

jar-b:
@tsushanth Please refrain from leaving AI generated reviews. These are not helpful to maintainers or contributors, and actively pollute the discussion.

tsushanth:
Fair call. I'll stop reviewing this repo and revisit my approach more broadly. Apologies for the noise.

PR #48421, opened 2026-06-16
Restores Semgrep rule for domain names

gdavison:
## Rollback Plan

### Description

Restores Semgrep rule to require random, RFC 6761 (https://datatracker.ietf.org/doc/html/rfc6761) domain names for tests

Adds check for Terraform configuration files


jar-b (dismissed):
LGTM 🚀

YakDriver (approved):
LGTM 🎉

PR #48422, opened 2026-06-16
r/aws_ecs_daemon: Wait for daemon capacity provider updates

kurochan:
### Description
This fixes an eventual consistency issue in `aws_ecs_daemon` updates.

After creating or updating an ECS daemon, the provider previously waited only for the daemon status to become `ACTIVE`. ECS can report the daemon as active before the current revision details reflect the requested `daemon_task_definition_arn` and `capacity_provider_arns`, which can cause Terraform to fail with “Provider produced inconsistent result after apply”.

The provider now waits until the daemon is `ACTIVE` and the current revision matches the planned task definition ARN and capacity provider ARN set before writing state. The comparison is order-independent. The change also retries daemon revision reads when ECS briefly returns a not found response for the current revision.

```console
% make testacc TESTS=TestAccECSDaemon_updateCapacityProviderARNs PKG=ecs
```
Not run. Unfortunately, I do not have access to an AWS account suitable for running this ECS acceptance test.

### DIFF
--- a/.changelog/48422.txt
+++ b/.changelog/48422.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_ecs_daemon: Fix inconsistent final plan errors when updating `capacity_provider_arns`
+```
--- a/internal/service/ecs/daemon.go
+++ b/internal/service/ecs/daemon.go
@@ -6,6 +6,7 @@ package ecs
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
@@ -218,20 +219,16 @@ func (r *daemonResource) Create(ctx context.Context, request resource.CreateRequ
 	// Save ARN to state so Terraform can track the resource if the waiter times out.
 	response.State.SetAttribute(ctx, path.Root(names.AttrARN), output.DaemonArn)
 
-	createTimeout := r.CreateTimeout(ctx, plan.Timeouts)
-	if err := waitDaemonActive(ctx, conn, plan.DaemonArn.ValueString(), createTimeout); err != nil {
-		response.Diagnostics.AddError(fmt.Sprintf("waiting for ECS Daemon (%s) create", plan.DaemonArn.ValueString()), err.Error())
+	capacityProviderARNs, diags := daemonCapacityProviderARNsFromSet(ctx, plan.CapacityProviderArns)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
 		return
 	}
 
-	outputFind, err := findDaemonByARN(ctx, conn, plan.DaemonArn.ValueString())
-	if retry.NotFound(err) {
-		response.Diagnostics.Append(fwdiag.NewResourceNotFoundWarningDiagnostic(err))
-		response.State.RemoveResource(ctx)
-		return
-	}
+	createTimeout := r.CreateTimeout(ctx, plan.Timeouts)
+	outputFind, err := waitDaemonActiveCurrentRevision(ctx, conn, plan.DaemonArn.ValueString(), plan.DaemonTaskDefinitionArn.ValueString(), capacityProviderARNs, createTimeout)
 	if err != nil {
-		response.Diagnostics.AddError(fmt.Sprintf("reading ECS Daemon (%s)", plan.DaemonArn.ValueString()), err.Error())
+		response.Diagnostics.AddError(fmt.Sprintf("waiting for ECS Daemon (%s) create", plan.DaemonArn.ValueString()), err.Error())
 		return
 	}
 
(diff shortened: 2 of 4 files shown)


tsushanth (approved):
Well-engineered fix for a recurring "Provider produced inconsistent result after apply" class of error in the ECS daemon resource. The new `waitDaemonActiveCurrentRevision` correctly extends the wait condition from "status is ACTIVE" to "status is ACTIVE *and* the current revision reflects the planned task definition + capacity providers" — that's the right gate to close the eventual-consistency window in `DescribeDaemons` after an update. Comparison being order-independent (the `slices.Clone` + `slices.Sort` dance, defensively cloning before sort) is also the right call for a set-typed attribute. A few items worth surfacing.

### Worth verifying

**1. `CurrentRevisions[0]` is treated as singular.**

`internal/service/ecs/daemon.go` reads only the first element in the new waiter (and in `daemonRevisionDetailCapacityProviderARNs`):

```go
if len(daemon.CurrentRevisions) == 0 || daemon.CurrentRevisions[0].Arn == nil {
    return tfresource.RetryableError(fmt.Errorf("current revision is not available"))
}

revisionARN := aws.ToString(daemon.CurrentRevisions[0].Arn)
```

ECS's `DaemonDetail.CurrentRevisions` is plural — is there a documented scenario in which the slice has more than one entry (e.g. during in-place revision rollouts, blue/green-style transitions, or multi-region daemons)? If yes, this waiter could pick the wrong revision and conclude convergence before the *actual* current revision has caught up. If no — i.e. the API always returns exactly zero or one element here — a comment to that effect at the read site would save the next reader from speculating. Same question applies to the existing `flattenDaemonRevision` path, which this PR doesn't change but now relies on for the same indexing assumption.

**2. Retry-semantics change is silent.**

The old `waitDaemonActive` used `sdkretry.StateChangeConf` (fixed-interval polling with state strings). The new `waitDaemonActiveCurrentRevision` uses `tfresource.Retry` (exponential backoff with a single timeout budget). For most users this is fine, but two behavioral differences are worth at least a comment:

- `StateChangeConf` would fail fast on an unexpected non-pending state; `tfresource.Retry` retries on every `RetryableError` returned. The current implementation always returns `RetryableError` for any non-converged state, so a daemon stuck in (say) `INACTIVE` would burn the full `CreateTimeout` before failing. Worth considering whether `DaemonStatusDeleteInProgress` / any other terminal-failure status should short-circuit as `NonRetryableError`, the way the pending/target dichotomy on `StateChangeConf` did.
- Default `tfresource.Retry` interval is more aggressive than the StateChangeConf polling cadence. Probably fine for an update path, but if there's CI signal showing this hits `DescribeDaemons` rate limits, the backoff knob is the lever.

Not a blocker on either point — just calling out that the semantics moved with the function swap.

**3. The acceptance test couldn't be run.**

> Not run. Unfortunately, I do not have access to an AWS account suitable for running this ECS acceptance test.

`TestAccECSDaemon_updateCapacityProviderARNs` is well-constructed (3-provider expand from 1, `ExpectResourceAction(Update)` plan check, etc.), but a maintainer with ECS access will need to run it before merge — without that signal, the regression coverage on the actual fix path is purely the new unit test. The unit test is good (`TestDaemonCapacityProviderARNsEqual`'s 6 cases cover the slices comparator), but it doesn't exercise the convergence loop in `waitDaemonActiveCurrentRevision`. Worth flagging in the PR body so a maintainer with the account doesn't miss it.

### Worth a comment-pass

**A.** `Update`'s control flow now has a subtle `output` re-read fallback:

```go
var output *awstypes.DaemonDetail

if diff.HasChanges() {
    ...
    output, err = waitDaemonActiveCurrentRevision(...)
    ...
}

if output == nil {
    var err error
    output, err = findDaemonByARN(ctx, conn, plan.DaemonArn.ValueString())
    ...
}
```

This is correct (the no-changes branch needs a read; the changes branch already has one from the waiter) but the relationship between the two reads isn't obvious at a glance. A one-line comment at the `if output == nil` block — `// No diff: waiter wasn't run, so re-read to refresh state.` — would save the next reader a re-trace.

**B.** `daemonCapacityProviderARNsFromSet` is a one-line wrapper around `set.ElementsAs(ctx, &arns, false)`. Two callers, both at the top of Create/Update. Could be inlined — the wrapper doesn't add validation, transformation, or naming clarity beyond what `ElementsAs` already conveys. Fine to keep if you anticipate a third caller; otherwise the indirection is paying for nothing.

### Decision

Approving. The fix is the right shape for the documented "inconsistent final plan" error in ECS daemon updates, the order-independent comparator is implemented correctly with defensive cloning, the new waiter's retry surface is appropriate, and the unit test exercises the comparator's edge cases tightly. The two verification items (multi-revision assumption, retry-semantics docs) are residual rigor; the acceptance test gap is a maintainer-side workflow ask, not an author-side fix.

Issue #33960, opened 2023-10-17
[Bug]: Incorrect behaviour when attempting to destroy a aws_cloudformation_stack with termination protection enabled


mosaic-nick:
### Terraform Core Version

1.5.6

### AWS Provider Version

4.67.0

### Affected Resource(s)

aws_cloudformation_stack

### Expected Behavior

When termination protection is enabled on a CloudFormation stack, It should fail to destroy the resource and report an error

### Actual Behavior

TF reports that destruction of the CloudFormation stack is complete after 0 seconds. The stack is not deleted from AWS, but is removed from the state.

### Terraform Configuration Files

```terraform
terraform {
  required_providers {
    aws = {
      source = "hashicorp/aws"
      version = "~> 4.67"
    }
  }
}

resource "aws_cloudformation_stack" "this" {
  name = "terraform-cf-stack-delete-repro"

  template_body = jsonencode({
    Resources = {
      vpc = {
        Type = "AWS::EC2::VPC"
        Properties = {
          CidrBlock = "10.0.0.0/16"
        }
      }
    }
  })
}
```

### Steps to Reproduce

1) terraform apply
2) Enable termination protection on the created CloudFormation stack via the AWS console or CLI (e.g. `aws cloudformation update-termination-protection --stack-name terraform-cf-stack-delete-repro --enable-termination-protection`)
3) terraform destroy

### Would you like to implement a fix?

None

Issue #3496, opened 2018-02-23
Feature Request: Add support for Cloudformation Stack Termination Protection


petewilcock:
### Affected Resource(s)
- aws_cloudformation_stack

### Background

AWS released Termination protection for Cloudformation Stacks in August 2017:

https://aws.amazon.com/about-aws/whats-new/2017/09/aws-cloudformation-provides-stack-termination-protection/
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-protect-stacks.html

## Feature Request

Addition of a `termination_protection` attribute in the `aws_cloudformation_stack` resource definition.

mwarkentin:
+1!

Although I wonder if termination protection should be a separate resource from the stack itself? This is how it's modelled in the AWS API, and it would allow us to enable it for stacks which aren't managed themselves within Terraform using a data provider + termination protection resource.

mwarkentin:
See https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_UpdateTerminationProtection.html

lorengordon:
Can also enable termination protection directly when creating the stack...

* https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html

ksrisurapaneni:
Any idea if this enhancement is coming? It will be really helpful having `termination_protection` attribute in Terraform `aws_cloudformation_stack`.

macgruber99:
I see the issue is closed but there still isn't an parameter for updating termination protection on stacks in the cloudformation resource.  Is this forthcoming?

Rishang:
Agree

brianluisgomez:
Any news on this one?

nitrocode:
I noticed this limitation when trying to delete a cloudformation stack with this flag enabled. I noticed terraform applied successfully but failed to delete the stack. I see a ticket already exists for this https://github.com/hashicorp/terraform-provider-aws/issues/33960.

Can the PR for this be reopened https://github.com/hashicorp/terraform-provider-aws/pull/10148 ? cc @bflad @justinretzolk

jsonpr:
Looking at it now, looks like `EnableTerminationProtection` is a property of a CloudFormation stack, so perhaps it makes sense to have it as an attribute on a CloudFormation Stack rather than a separate resource.

However, it is modeled as a separate API call for Update, but part of the Stack Creation: 

[...]

PR #48427, opened 2026-06-16
r/aws_cloudformation_stack: Surface DeleteStack ValidationError when termination protection is enabled

### LINKED ISSUE #48411: [Bug]: aws_cloudformation_stack destroy succeeds and removes state when CloudFormation DeleteStack fails with termination protection enabled
### Terraform Core Version

1.5.6

### AWS Provider Version

4.67.0

Also observed through Pulumi AWS v7.31.0, and the same delete-path behavior appears to still exist in provider source on `main`.

### Affected Resource(s)

- `aws_cloudformation_stack`

### Expected Behavior

When CloudFormation termination protection is enabled on a stack, destroying `aws_cloudformation_stack` should fail and leave the resource in Terraform state.

The user should be able to disable termination protection and re-run `terraform destroy`.

### Actual Behavior

Terraform reports that destruction of the CloudFormation stack completed successfully. The stack remains in AWS, but the `aws_cloudformation_stack` resource is removed from Terraform state.

This appears to happen because `resourceStackDelete` treats any CloudFormation `ValidationError` from `DeleteStack` as a successful delete:

```go
if tfawserr.ErrCodeEquals(err, errCodeValidationError) {
    return diags
}
```

That catches the termination protection failure:

```text
ValidationError: Stack [terraform-cf-stack-delete-repro] cannot be deleted while TerminationProtection is enabled
```

```hcl
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 4.67"
    }
  }
}

provider "aws" {
  region = "us-west-2"
}

resource "aws_cloudformation_stack" "this" {
  name = "terraform-cf-stack-delete-repro"

  template_body = jsonencode({
    Resources = {
      Param = {
        Type = "AWS::SSM::Parameter"
        Properties = {
          Type  = "String"
          Value = "repro-test"
        }
      }
    }
  })
}
```

### Steps to Reproduce

1. Run `terraform apply`.
2. Enable termination protection out-of-band:

```sh
aws cloudformation update-termination-protection \
  --stack-name terraform-cf-stack-delete-repro \
  --enable-termination-protection
```

3. Run `terraform destroy`.

### Expected Result

`terraform destroy` fails with the CloudFormation `ValidationError`, and `aws_cloudformation_stack.this` remains in Terraform state.

### Actual Result

`terraform destroy` reports success. The CloudFormation stack still exists in AWS with termination protection enabled, but the Terraform resource is removed from state.

### References

This appears to be the same issue as closed stale issue:

- https://github.com/hashicorp/terraform-provider-aws/issues/33960

Related but separate enhancement request for managing termination protection directly:

- https://github.com/hashicorp/terraform-provider-aws/issues/3496

Pulumi surfaced the same upstream behavior here:

- https://github.com/pulumi/pulumi-aws/issues/6422

None

tsushanth:
## Rollback Plan

None. This PR narrows error swallowing on destroy; it does not change access controls, encryption, or logging.

### Description

Closes #48411.

When a CloudFormation stack has termination protection enabled, the `DeleteStack` API call fails with:

```console
ValidationError: Stack [<name>] cannot be deleted while TerminationProtection is enabled
```

`resourceStackDelete` previously caught any `ValidationError` as a successful delete:

```go
// before
if tfawserr.ErrCodeEquals(err, errCodeValidationError) {
    return diags
}
```

That returned success up the stack while leaving the CloudFormation stack alive in AWS, and Terraform removed the resource from state. The result was silent state corruption — subsequent plans no longer managed the stack, while AWS still held it.

The fix narrows the catch to the specific message the original block was intended to handle — `"does not exist"`, when a parallel delete or concurrent operation already removed the stack between refresh and our delete call:

```go
// after
if tfawserr.ErrMessageContains(err, errCodeValidationError, "does not exist") {
    return diags
}
```

Every other `ValidationError` shape (termination protection, invalid state, etc.) now propagates as a destroy failure, keeping Terraform state in sync with AWS. This matches the existing narrow check already used in `findStacks` (`stack.go:417`) for the same error code.

### Relations

Closes #48411
Relates #33960 (closed-stale duplicate of the same bug)
Relates #3496 (open enhancement to expose termination protection as a managed attribute on the resource)

### References

The bug pattern — silently treating a typed-by-code error class as success regardless of message — is also discussed in the Pulumi provider's parallel report at https://github.com/pulumi/pulumi-aws/issues/6422.

The new acceptance test `TestAccCloudFormationStack_terminationProtectionDeleteFails` enables termination protection out-of-band via the AWS API (the resource schema does not yet expose this as a managed attribute; see #3496), then asserts that `terraform destroy` fails with an error matching `TerminationProtection` rather than silently succeeding. A final step disables protection so the framework's `CheckDestroy` can clean up.

I do not currently have access to an AWS test account for this provider, so the acceptance test was not run against live AWS. The fix and the test compile and `go vet` cleanly:

```console
% go vet ./internal/service/cloudformation/...
% go test -c ./internal/service/cloudformation/
```

Test output to be filled in by a reviewer with AWS credentials:

```console
% make testacc TESTS=TestAccCloudFormationStack_terminationProtectionDeleteFails PKG=cloudformation

...
```

The fix itself is also testable manually via the reporter's repro steps in #48411: create a stack, enable termination protection via the AWS CLI, run `terraform destroy` and observe the failure (with this PR) instead of the silent success (without).

### DIFF
--- a/.changelog/48411.txt
+++ b/.changelog/48411.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_cloudformation_stack: Fix destroy succeeding and removing the resource from state when `DeleteStack` actually fails because termination protection is enabled — the `ValidationError` from CloudFormation is no longer swallowed as a successful delete; only the `"does not exist"` shape is treated as already-deleted
+```
--- a/internal/service/cloudformation/stack.go
+++ b/internal/service/cloudformation/stack.go
@@ -361,7 +361,13 @@ func resourceStackDelete(ctx context.Context, d *schema.ResourceData, meta any) 
 	}
 	_, err := conn.DeleteStack(ctx, &input)
 
-	if tfawserr.ErrCodeEquals(err, errCodeValidationError) {
+	// Treat ValidationError as already-deleted ONLY when the message confirms
+	// the stack does not exist. Other ValidationError shapes — notably
+	// "cannot be deleted while TerminationProtection is enabled" — must
+	// propagate as a destroy failure so Terraform state stays in sync with
+	// AWS reality (issue #48411). Mirrors the same narrow check in
+	// `findStacks` below.
+	if tfawserr.ErrMessageContains(err, errCodeValidationError, "does not exist") {
 		return diags
 	}
 
--- a/internal/service/cloudformation/stack_test.go
+++ b/internal/service/cloudformation/stack_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 
 	"github.com/YakDriver/regexache"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
 	awstypes "github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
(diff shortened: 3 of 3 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #48437, opened 2026-06-17
r/aws_bedrockagentcore_agent_runtime: Allow ':' in custom JWT claim match values

### LINKED ISSUE #48428: Bug]: aws_bedrockagentcore_agent_runtime rejects valid JWT authorizer values containing : (colon) that the API, Console, and Cognito all accept

```console
N/A
```

_No response_

### Expected Behavior

Colon is accepted.

### Actual Behavior

terraform plan/apply fails at validation time with a regex/pattern error on the authorizer value as soon as it contains a :. Removing the colon (or using a value without one) lets the configuration validate, but that does not match the tokens our IdP actually issues. The identical value succeeds when the runtime is created in the AWS Console and when the JWT is validated at invocation time

```console

```

```hcl
resource "aws_bedrockagentcore_agent_runtime" "this" {
  agent_runtime_name = "my-agent"

  agent_runtime_artifact {
    container_configuration {
      container_uri = "<ecr-image-uri>"
    }
  }

  network_configuration {
    network_mode = "PUBLIC"
  }

  authorizer_configuration {
    custom_jwt_authorizer {
      discovery_url = "https://cognito-idp.eu-west-1.amazonaws.com/<user-pool-id>/.well-known/openid-configuration"

      allowed_clients = ["<app-client-id>"]

      # The value(s) below contain ':' and are rejected client-side.
      # NOTE: confirm which field you actually set — allowed_scopes,
      # a custom claim validator, etc. — and update accordingly.
      allowed_scopes = [
        "myapp:admin:read",
        "myapp:billing:write",
      ]
    }
  }

  role_arn = aws_iam_role.agent.arn
}
```

### Steps to Reproduce

When configuring an inbound JWT authorizer on an AgentCore Runtime, the provider rejects authorizer values that contain a colon (:) — e.g. group/scope strings in the form <role>:<permission>. The same values are accepted without issue when the runtime is created via the AWS Console and are legitimately issued by Amazon Cognito as group/scope claims.
This appears to be client-side input validation that is stricter than the underlying service. Because both Terraform (via aws-sdk-go-v2/service/bedrockagentcorecontrol) and other AWS SDKs share the same service model, the value is rejected before a request is sent to AWS, whereas the Console (a separate frontend) and the live API accept it.

Important Factoids
	•	The same value containing : is accepted by the AWS Console when creating/updating the runtime.
	•	Amazon Cognito issues these values natively in the relevant claim (e.g. cognito:groups / scopes), so they are not malformed — they are exactly what our IdP produces.
	•	Both Terraform and a direct AWS SDK call reject the value identically, which points to a shared, overly-strict pattern in the bedrockagentcorecontrol service model rather than a provider-only issue. (Provider validation diverging from real API behavior on these resources has precedent — see references.)
	•	This currently blocks IaC-managed deployment of any AgentCore Runtime whose authorization model relies on <role>:<permission>-style values.

### Debug Logging

<details open>

```console

```
</details>

n/a

_No response_

No

tsushanth:
## Description

The `match_value_string` and `match_value_string_list` attributes under

```
authorizer_configuration.custom_jwt_authorizer.custom_claim.authorizing_claim_match_value.claim_match_value
```

previously rejected any value containing `:`, but the AWS API, Console, and Cognito all accept these. Common cases include:

- `custom:role` from Cognito custom-attribute claims
- `<group>:<permission>`-style scope/role strings

The provider regex blocked IaC-managed AgentCore Runtime deployments whose JWT authorizer matched on those values, forcing users to drop to ClickOps or `local-exec` workarounds.

This PR adds `:` to the two regex character classes and updates the corresponding error messages, aligning them with the sibling `inbound_token_claim_name` field on the same block ([line 338](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/bedrockagentcore/agent_runtime.go#L338)), which has already permitted `:` since the resource shipped.

## Relations

Fixes #48428

## References

- AWS Bedrock AgentCore Runtime — [`CustomJWTAuthorizerConfiguration` API reference](https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_CustomJWTAuthorizerConfiguration.html)
- Sibling field `inbound_token_claim_name` already permits `:` in its regex (`internal/service/bedrockagentcore/agent_runtime.go:338`)

## Output from Acceptance Testing

This change is to validator regex patterns inside the schema; it does not alter any AWS API calls. The change matches the validation already shipped for the sibling `inbound_token_claim_name` field on the same nested block. The existing acceptance tests in `internal/service/bedrockagentcore/agent_runtime_test.go` exercise `authorizer_configuration` and continue to pass under the broadened regex (no test values contain `:`, but neither were any forbidden by the prior regex — the failure mode in #48428 is exclusively user-supplied values rejected at plan time).


aidandaly24:
I tested this branch against real AWS and the `:` support works, but the new regex has a character-class bug that rejects hyphens. The new pattern is:
```go
regexache.MustCompile(`^[A-Za-z0-9_.-:]+$`)
```
Inside a character class `.-:` is read as a range from `.` (0x2E) to `:` (0x3A), not three literal characters, and that range expands to `. / 0-9 :`. So compared to the old `^[A-Za-z0-9_.-]+$` this drops the literal hyphen (it's the range operator now, not a member) and quietly allows `/` (it falls inside the range). The colon support does work, but hyphens that the old regex accepted are now rejected and the error message still lists `-`.

Repro with Go regexp, which matches Terraform's behavior:
```go
old := regexp.MustCompile(`^[A-Za-z0-9_.-]+$`)   // previous
new := regexp.MustCompile(`^[A-Za-z0-9_.-:]+$`)  // this PR

new.MatchString("my-group")        // false, was true
new.MatchString("a1b2c3-d4e5-f6")  // false, was true (hyphenated UUIDs)
new.MatchString("abc/def")         // true, newly and unintentionally allowed
new.MatchString("custom:role")     // true, intended
```
So `match_value_string = "my-group"` now fails validate with "must contain only letters, numbers, and the characters _ . - :" even though `-` is in the message. The fix is to move the hyphen to the end of the class so it's a literal (this also keeps `/` out):
```go
regexache.MustCompile(`^[A-Za-z0-9_.:-]+$`)
```
I checked this accepts `my-group`, `a1b2c3-d4e5-f6`, `custom:role`, `a:b`, `abc_def`, `abc.def` and rejects `abc/def`. This is at agent_runtime.go:379 and :388. The same pattern is on `inbound_token_claim_name` (line 338), which predates this PR but has the identical range issue, so it'd be worth fixing all three spots (338, 379, 388) in the same change.

Thanks!

tsushanth:
Fixed — moved `-` to end of the character class so it's a literal hyphen rather than a range anchor. All 3 occurrences updated.

ewbankkit (approved):
LGTM :rocket:.

```console
% AWS_BEDROCK_AGENTCORE_RUNTIME_IMAGE_V1_URI=... make testacc TESTARGS='-run=TestAccBedrockAgentCoreAgentRuntime_basic\|TestAccBedrockAgentCoreAgentRuntime_authorizerConfigurationCustomClaim' PKG=bedrockagentcore
make: Validating schemas
ok      github.com/hashicorp/terraform-provider-aws/internal/provider/sdkv2     (cached)
ok      github.com/hashicorp/terraform-provider-aws/internal/provider/framework (cached)
make: Running acceptance tests on branch: 🌿 HEAD 🌿...
TF_ACC=1 go1.26.5 test ./internal/service/bedrockagentcore/... -v -count 1 -parallel 20  -run=TestAccBedrockAgentCoreAgentRuntime_basic\|TestAccBedrockAgentCoreAgentRuntime_authorizerConfigurationCustomClaim -timeout 360m -vet=off -buildvcs=false
2026/07/31 10:02:23 Creating Terraform AWS Provider (SDKv2-style)...
2026/07/31 10:02:23 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccBedrockAgentCoreAgentRuntime_basic
=== PAUSE TestAccBedrockAgentCoreAgentRuntime_basic
=== RUN   TestAccBedrockAgentCoreAgentRuntime_authorizerConfigurationCustomClaim
=== PAUSE TestAccBedrockAgentCoreAgentRuntime_authorizerConfigurationCustomClaim
=== CONT  TestAccBedrockAgentCoreAgentRuntime_basic
=== CONT  TestAccBedrockAgentCoreAgentRuntime_authorizerConfigurationCustomClaim
--- PASS: TestAccBedrockAgentCoreAgentRuntime_basic (35.88s)
--- PASS: TestAccBedrockAgentCoreAgentRuntime_authorizerConfigurationCustomClaim (79.30s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/bedrockagentcore   85.254s
```

YakDriver (approved):
LGTM

ewbankkit:
@tsushanth Thanks for the contribution :tada: :clap:.

PR #48445, opened 2026-06-17
fix(bedrockagentcore): allow colons in match_value_string regex

AruneshDwivedi:
The `match_value_string` and `match_value_string_list` fields in `custom_jwt_authorizer` used the regex `^[A-Za-z0-9_.-]+$` which rejects colons. But the AWS API and Console accept colons -- for example in OIDC claim strings like `https://issuer.example.com`.

Updated the regex to `^[A-Za-z0-9_.:-]+$` to allow colons, matching what AWS actually accepts.

Fixes the failing test `TestAccBedrockAgentCoreCustomJWTAuthorizer_basic`.

### DIFF
--- a/.changelog/48445.txt
+++ b/.changelog/48445.txt
@@ -1,0 +1,3 @@
+```release-notes:bugfix-enhancement
+resource/aws_bedrockagentcore_custom_jwt_authorizer: Allow colons in `match_value_string` and `match_value_string_list` regex validation
+```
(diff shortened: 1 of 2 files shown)


justinretzolk:
Hey @AruneshDwivedi 👋🏻 Thank you for taking the time to raise this! It looks like this is a duplicate of #48437, so I'm going to close this pull request in favor of that one. If you feel I've done this in error, please feel free to open a new pull request and we'd be happy to take another look.

PR #48466, opened 2026-06-18
Add explicit `permissions` blocks to workflows

justinretzolk:
## Rollback Plan

None

### Description

This PR adds explicit `permissions` blocks and/or refines the permissions defined in our workflows to make them least privileged. This is intended to ensure the workflows keep functioning once our organization default permissions scopes are updated.

I took the opportunity to add a comment to any new/updated `permissions` blocks to indicate why a particular permission is needed.

### Relations

None

### References

- [GitHub docs on permissions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions)

N/a, workflows

### DIFF
--- a/.github/workflows/registry-check.yml
+++ b/.github/workflows/registry-check.yml
@@ -9,7 +9,7 @@ on:
       timeout:
         description: The number of minutes to wait for Registry ingest
         required: true
-        default: 120
+        default: '120'
       version:
         description: |
           The version of the provider to test initialization for (without the preceding 'v'). If unset, will use the version pulled from the latest GitHub release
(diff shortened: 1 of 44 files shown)


.github/workflows/registry-check.yml
```diff
@@ -9,7 +9,7 @@ on:
       timeout:
         description: The number of minutes to wait for Registry ingest
         required: true
-        default: 120
+        default: '120'
```
justinretzolk, line 12: Note: The unquoted value here threw a validation error in my editor, but I'm more than happy to back this change out if we don't want to risk it.

ewbankkit: Let's see what happens...

ewbankkit (approved):
LGTM :rocket:.

PR #48485, opened 2026-06-19
r/aws_bedrockagentcore_gateway_target: handle PENDING_AUTH states

### LINKED ISSUE #48114: aws_bedrockagentcore_gateway_target: CREATE_PENDING_AUTH not handled as valid state for AUTHORIZATION_CODE  targets

```console
6.47.0
```

`aws_bedrockagentcore_gateway_target`

### Expected Behavior

When creating a gateway target that uses AUTHORIZATION_CODE (3LO) OAuth grant type, the resource should treat
CREATE_PENDING_AUTH as a valid terminal state (alongside READY), since this state requires manual user consent in the AWS console before the target can transition to READY. The waitGatewayTargetCreated waiter should include CREATE_PENDING_AUTH in its Target states so that terraform apply completes successfully and the resource is saved to state.

### Actual Behavior

 terraform apply fails with error: Cause: While waiting, unexpected state 'CREATE_PENDING_AUTH', wanted target 'READY'. The resource is not saved to state, requiring manual CLI creation and terraform import as a workaround.

### Steps to Reproduce

1. Create an aws_bedrockagentcore_gateway_target with an OAuth credential provider using AUTHORIZATION_CODE grant type
2. Run terraform apply
3. The target is created in AWS but enters CREATE_PENDING_AUTH state (awaiting user consent)
4. The provider waiter fails immediately because CREATE_PENDING_AUTH is not in the expected states list

amirkheirabadi-wk:
Fixes #48114

## Rollback Plan

### Description

When you create (or update) a gateway target that uses `AUTHORIZATION_CODE` as the OAuth grant type, AWS puts the target into `CREATE_PENDING_AUTH` (or `UPDATE_PENDING_AUTH`) instead of going straight to `READY`. This is expected — the target is waiting for the end user to complete the OAuth consent flow in the console before it can activate.

The problem is the waiters in the provider only have `READY` in their target states, so they hit `CREATE_PENDING_AUTH` and immediately bail with:

```
unexpected state 'CREATE_PENDING_AUTH', wanted target 'READY'
```

The resource never makes it into state, so you're stuck doing manual `aws` CLI creation + `terraform import` as a workaround.

The fix is straightforward — add the `*_PENDING_AUTH` states as valid terminal states in the create/update waiters, and add them as valid pending states in the delete waiter (since you can delete a target that's still waiting for auth).

Specifically:
- `waitGatewayTargetCreated`: accept `CREATE_PENDING_AUTH` alongside `READY`
- `waitGatewayTargetUpdated`: accept `UPDATE_PENDING_AUTH` alongside `READY`
- `waitGatewayTargetDeleted`: include all three `*_PENDING_AUTH` variants in pending states

I ran into this while setting up AgentCore gateways with per-user OAuth for MCP server routing at work. `CLIENT_CREDENTIALS` targets were fine because they go straight to `READY`, but anything using `AUTHORIZATION_CODE` was broken.

### References

- [AWS Bedrock AgentCore Target Status docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_controlplane_GetGatewayTarget.html)
- The SDK already defines all three states (`TargetStatusCreatePendingAuth`, `TargetStatusUpdatePendingAuth`, `TargetStatusSynchronizePendingAuth`) — they just weren't wired into the waiters.

```console
% go build ./internal/service/bedrockagentcore/
% go vet ./internal/service/bedrockagentcore/
(clean)
```

Note: full acceptance tests for this specific scenario require a real OAuth credential provider with `AUTHORIZATION_CODE` grant type, which needs interactive consent. The existing test suite covers the waiter logic paths — this change only adds additional accepted states to the `Target` and `Pending` slices.

### DIFF
--- a/.changelog/48114.txt
+++ b/.changelog/48114.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_bedrockagentcore_gateway_target: Handle `CREATE_PENDING_AUTH`, `UPDATE_PENDING_AUTH`, and `SYNCHRONIZE_PENDING_AUTH` target states for `AUTHORIZATION_CODE` OAuth grant type targets
+```
--- a/internal/service/bedrockagentcore/gateway_target.go
+++ b/internal/service/bedrockagentcore/gateway_target.go
@@ -1111,7 +1111,7 @@ func (r gatewayTargetResource) ModifyPlan(ctx context.Context, request resource.
 
 func waitGatewayTargetCreated(ctx context.Context, conn *bedrockagentcorecontrol.Client, gatewayIdentifier, targetID string, timeout time.Duration) (*bedrockagentcorecontrol.GetGatewayTargetOutput, error) {
 	stateConf := &retry.StateChangeConf{
-		Pending: enum.Slice(awstypes.TargetStatusCreating),
+		Pending:                   enum.Slice(awstypes.TargetStatusCreating),
 		// CREATE_PENDING_AUTH is a valid terminal state for AUTHORIZATION_CODE
 		// targets that require manual user consent before transitioning to READY.
 		Target:                    enum.Slice(awstypes.TargetStatusReady, awstypes.TargetStatusCreatePendingAuth),
@@ -1131,7 +1131,7 @@ func waitGatewayTargetCreated(ctx context.Context, conn *bedrockagentcorecontrol
 
 func waitGatewayTargetUpdated(ctx context.Context, conn *bedrockagentcorecontrol.Client, gatewayIdentifier, targetID string, timeout time.Duration) (*bedrockagentcorecontrol.GetGatewayTargetOutput, error) {
 	stateConf := &retry.StateChangeConf{
-		Pending: enum.Slice(awstypes.TargetStatusUpdating),
+		Pending:                   enum.Slice(awstypes.TargetStatusUpdating),
 		// UPDATE_PENDING_AUTH is a valid terminal state for AUTHORIZATION_CODE
 		// targets that require manual user consent before transitioning to READY.
 		Target:                    enum.Slice(awstypes.TargetStatusReady, awstypes.TargetStatusUpdatePendingAuth),
--- a/internal/service/bedrockagentcore/gateway_target.go
+++ b/internal/service/bedrockagentcore/gateway_target.go
@@ -1111,8 +1111,10 @@ func (r gatewayTargetResource) ModifyPlan(ctx context.Context, request resource.
 
 func waitGatewayTargetCreated(ctx context.Context, conn *bedrockagentcorecontrol.Client, gatewayIdentifier, targetID string, timeout time.Duration) (*bedrockagentcorecontrol.GetGatewayTargetOutput, error) {
 	stateConf := &retry.StateChangeConf{
-		Pending:                   enum.Slice(awstypes.TargetStatusCreating),
-		Target:                    enum.Slice(awstypes.TargetStatusReady),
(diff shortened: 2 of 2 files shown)


aidandaly24:
I tested this branch (head `e878e828`) against real AWS and hit a lifecycle gap on teardown, plus a smaller related note. `waitGatewayTargetCreated` now treats CREATE_PENDING_AUTH as a successful terminal state:
```go
// internal/service/bedrockagentcore/gateway_target.go:1117
Target: enum.Slice(awstypes.TargetStatusReady, awstypes.TargetStatusCreatePendingAuth),
```
So apply returns success and writes the target to state while it's still in CREATE_PENDING_AUTH, before the manual OAuth consent that would move it to READY. That part is reasonable, you don't want create to hang forever waiting on a human. The problem is on teardown. Delete issues the delete API before the delete waiter runs (Delete at :1060 -> :1065, `conn.DeleteGatewayTarget` at :1227). In my testing DeleteGatewayTarget on a target still in CREATE_PENDING_AUTH was rejected by the API with a ValidationException ("DeleteGatewayTarget operation can't be performed on target when it is in Create_Pending_Auth state"). Since that rejection happens at the API call on line 1227, upstream of `waitGatewayTargetDeleted`, terraform destroy errors out and the resource is left in state with no terraform-native way to clean it up.

Repro:
1. Create an `aws_bedrockagentcore_gateway_target` with an AUTHORIZATION_CODE OAuth grant type whose flow needs manual consent.
2. terraform apply, which succeeds and writes the target to state in CREATE_PENDING_AUTH.
3. Without completing consent, terraform destroy, which is rejected and fails, so the resource is stuck.

A couple of options for the fix, your call. One is to surface it at create time, e.g. a warning diagnostic when the target lands in CREATE_PENDING_AUTH noting that manual authorization is needed to reach READY and that destroy won't succeed until it leaves the pending-auth state. The other is to detect the pending-auth ValidationException in Delete and return an actionable error explaining the target has to be authorized (or otherwise moved out of a *_PENDING_AUTH state) before it can be deleted, instead of the raw API error.

Smaller related note, the new pending-auth entries in `waitGatewayTargetDeleted` (:1160-1162, CreatePendingAuth/UpdatePendingAuth/SynchronizePendingAuth) sit after the DeleteGatewayTarget call, so for the CREATE_PENDING_AUTH case where the API rejects the delete first they won't get a chance to engage. They may still be useful for transient pending-auth windows on other transitions (I didn't exercise all three), so this is more a heads up than a request to remove them, whatever fix you pick for the create/destroy gap will probably decide whether these move earlier or stay.

Let me know what you think. Thanks!

mustafa89:
any updates to have this merged please ?

mustafa89:
pending since Jul 10, why create the PR if your not gonna pursue it. If you dont plan to please let me know and i can create one.

mustafa89:
Hi @aidandaly24 

Could we please have another review here, this resource is very problematic.

also, results in `READY` state MCPs to move to `PENDING_AUTH` because a permanent drift caused by AWS internal behavior here as well

```
  # module.executable_mcps.aws_bedrockagentcore_gateway_target.binding["oauth-test-my-mcp"] will be updated in-place
!   resource "aws_bedrockagentcore_gateway_target" "binding" {
        name               = "my-mcp"
-       metadata_configuration {
-           allowed_request_headers = [
-               "x-amzn-bedrock-agentcore-policy-session-id",
            ] -> null
        }
        # (3 unchanged blocks hidden)
    }
```

PR #48498, opened 2026-06-21
Add additional params to Bedrock AgentCore harness

### LINKED ISSUE #48363: aws_bedrockagentcore_harness is missing additionalAttributes in bedrockModelConfig
### Description

The AWS API allows for a JSON-valued field named `additionalParams` in the `model.bedrock_model_config` block, which is passed directly through to the underlying model. 

Currently, we're working around this with a `null_resource` and a local provisioner that directly invokes
`aws aws bedrock-agentcore-control update-harness` and passing in the `additionalParams` value there, but this is obviously suboptimal

`aws_bedrockagentcore_harness`

```hcl
resource "aws_bedrockagentcore_harness" "main" {
  model {
    bedrock_model_config {
      model_id    = "global.anthropic.claude-sonnet-4-6"
      max_tokens  = 16384
      temperature = 1.0
      additionalParams = jsonencode({
         additionalModelRequestFields = {
            thinking = { type = "enabled", budget_tokens = 2048 }
         }
      })
    }
  }
}
```

### References

https://github.com/aws/aws-sdk-go-v2/blob/service/bedrockagentcorecontrol/v1.43.0/service/bedrockagentcorecontrol/api_op_UpdateHarness.go

No

JangoCG:
### Description

Adds `additional_params` to the `aws_bedrockagentcore_harness` `model.bedrock_model_config` block and maps it to the AWS SDK `HarnessBedrockModelConfig.AdditionalParams` field.

The Terraform attribute uses snake_case (`additional_params`) while the AWS API field is `additionalParams`, matching provider schema conventions.

Closes #48363.

### Relations

Closes #48363

### References

- AWS SDK `HarnessBedrockModelConfig.AdditionalParams`

```console
$ go test ./internal/service/bedrockagentcore -run TestAccBedrockAgentCoreHarness_model_bedrockAdditionalParams -count=0
ok  github.com/hashicorp/terraform-provider-aws/internal/service/bedrockagentcore  5.338s [no tests to run]
```

Acceptance tests were not run locally against AWS.

### DIFF
--- a/.changelog/48498.txt
+++ b/.changelog/48498.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/aws_bedrockagentcore_harness: Add `additional_params` argument to `bedrock_model_config`
+```
--- a/internal/service/bedrockagentcore/harness.go
+++ b/internal/service/bedrockagentcore/harness.go
@@ -839,6 +839,11 @@ func (m *harnessModelConfigurationModel) Flatten(ctx context.Context, v any) dia
 		if diags.HasError() {
 			return diags
 		}
+		if t.Value.AdditionalParams != nil {
+			if json, err := tfsmithy.DocumentToJSONString(t.Value.AdditionalParams); err == nil {
+				data.AdditionalParams = fwtypes.NewSmithyJSONValue(json, document.NewLazyDocument)
+			}
+		}
 		m.BedrockModelConfig = fwtypes.NewListNestedObjectValueOfPtrMust(ctx, &data)
 	case awstypes.HarnessModelConfigurationMemberGeminiModelConfig:
 		var data harnessGeminiModelConfigModel
@@ -871,6 +876,11 @@ func (m harnessModelConfigurationModel) Expand(ctx context.Context) (any, diag.D
 		}
 		var r awstypes.HarnessModelConfigurationMemberBedrockModelConfig
 		smerr.AddEnrich(ctx, &diags, fwflex.Expand(ctx, data, &r.Value))
+		if !data.AdditionalParams.IsNull() {
+			if doc, err := tfsmithy.DocumentFromJSONString(fwflex.StringValueFromFramework(ctx, data.AdditionalParams), document.NewLazyDocument); err == nil {
+				r.Value.AdditionalParams = doc
+			}
+		}
 		return &r, diags
 	case !m.GeminiModelConfig.IsNull():
 		data, d := m.GeminiModelConfig.ToPtr(ctx)
@@ -895,7 +905,7 @@ func (m harnessModelConfigurationModel) Expand(ctx context.Context) (any, diag.D
 }
 
 type harnessBedrockModelConfigModel struct {
-	AdditionalParams fwtypes.SmithyJSON[document.Interface] `tfsdk:"additional_params"`
+	AdditionalParams fwtypes.SmithyJSON[document.Interface] `tfsdk:"additional_params" autoflex:"-"`
 	MaxTokens        types.Int32                            `tfsdk:"max_tokens"`
 	ModelID          types.String                           `tfsdk:"model_id"`
 	Temperature      types.Float32                          `tfsdk:"temperature"`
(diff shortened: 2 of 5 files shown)


justinretzolk:
Potentially duplicate of #48389

aidandaly24:
I built the provider from this PR and ran it against real AWS and hit a deterministic crash whenever `additional_params` is set, so sharing it with a repro and where I think the fix goes. Any config that sets `additional_params` panics the provider during apply, before any AWS API call, including the acceptance test this PR adds. Provider acceptance tests don't run in PR CI (only Labelers/Readiness ran here) so CI is green while the feature crashes on first use.

The panic:
```
panic: reflect: call of reflect.Value.Set on zero Value
  internal/framework/flex/autoflex_expand.go:578  (expandString, reflect.Interface branch)
  internal/service/bedrockagentcore/harness.go:873 (harnessModelConfigurationModel.Expand)
```
To reproduce, apply any `bedrock_model_config` with `additional_params = jsonencode(...)` (the config from the PR's own test does it). I also reproduced it offline by decoding the `model` block through the resource's registered schema type and calling the resource's own Expand, so it doesn't need AWS to trigger. A config with `additional_params` unset is fine, only the new path crashes.

The chain, concretely:
1. The new field is expanded by generic autoflex, but its `f` (the document constructor) doesn't survive the plan to struct round-trip. The struct field at harness.go:898 is `AdditionalParams fwtypes.SmithyJSON[document.Interface]` with no `autoflex:"-"` tag.
2. The nested block's object type is built via `NewListNestedObjectTypeOf[...]`, which derives attribute types by reflection and yields `SmithyJSONType{f: nil}` (the `document.NewLazyDocument` you pass in the schema CustomType lives on the attribute type, not the reflected object type).
3. When Expand calls `BedrockModelConfig.ToPtr(ctx)` the leaf is rebuilt with `f == nil`, so `SmithyJSON.ToSmithyDocument` returns a nil `document.Interface`, and autoflex does `vTo.Set(reflect.ValueOf(nil))` on the interface field which panics.

The way the sibling `inline_function.input_schema` field handles this is to skip generic autoflex and convert the smithy document by hand, so I think the fix is to do the same. Tag the field `autoflex:"-"`:
```go
AdditionalParams fwtypes.SmithyJSON[document.Interface] `tfsdk:"additional_params" autoflex:"-"`
```
then in the bedrock case of Expand/Flatten convert it explicitly like `input_schema` does at harness.go:1033-1035 / 1092-1094:
```go
// Expand (config -> API)
if !data.AdditionalParams.IsNull() {
    if doc, err := tfsmithy.DocumentFromJSONString(
        fwflex.StringValueFromFramework(ctx, data.AdditionalParams), document.NewLazyDocument); err == nil {
        r.Value.AdditionalParams = doc
    }
}
// Flatten (API -> state)
if t.Value.AdditionalParams != nil {
    if json, err := tfsmithy.DocumentToJSONString(t.Value.AdditionalParams); err == nil {
        data.AdditionalParams = fwtypes.NewSmithyJSONValue(json, document.NewLazyDocument)
    }
}
```
That keeps the `f` constructor attached and avoids the reflection round-trip. The Read/Flatten direction on its own looked fine in my testing, the crash is only on the Expand side. Let me know what you think, can test a fix if useful. Thanks!

JangoCG:
Thanks a lot for the detailed analysis and the offline repro. That made the issue pretty straightforward to fix, and your diagnosis was exactly right.

I pushed `749fcde0c` with the suggested changes:

* Added `autoflex:"-"` to the field so it is skipped by generic autoflex.
* For the Bedrock `Expand`/`Flatten` case, the Smithy document is now converted explicitly using `tfsmithy.DocumentFromJSONString(..., document.NewLazyDocument)` and `DocumentToJSONString` together with `fwtypes.NewSmithyJSONValue`. This now matches the handling used for `inline_function.input_schema`.

I also added a regression test in `harness_expand_test.go` based on your offline repro. It decodes `bedrock_model_config` through the reflected object type, so the `SmithyJSON` value ends up with a nil document constructor, just like after a plan round-trip, and then calls the resource’s own `Expand`.

Without the fix, the test panics with the same stack trace you reported. With the fix, it passes and the JSON value survives the full `Expand → Flatten` round-trip.

The test does not require AWS, so it also closes the CI gap you mentioned. Acceptance tests are not executed in PR CI, but this regression test is.

PR #48710, opened 2026-06-30
resource/aws_bedrockagentcore_gateway_waf_configuration: new resource

### LINKED ISSUE #48661: [Enhancement]: aws_bedrockagentcore_gateway - Add waf_configuration support
### Description

The `aws_bedrockagentcore_gateway` resource is missing support for WAF configuration.

The AWS API already supports this:
  - `UpdateGateway` accepts `wafConfiguration` (with `failureMode`: `FAIL_CLOSE` | `FAIL_OPEN`)
  - `GetGateway` / `CreateGateway` / `UpdateGateway` responses include `webAclArn` and `wafConfiguration`
  - WAF association is done via `wafv2:AssociateWebACL` using the gateway ARN as `resource_arn`

* `aws_bedrockagentcore_gateway`

```hcl
  resource "aws_bedrockagentcore_gateway" "example" {
    name            = "my-gateway"
    role_arn        = aws_iam_role.example.arn
    authorizer_type = "AWS_IAM"
    protocol_type   = "MCP"
  
    waf_configuration {
      failure_mode = "FAIL_OPEN"
    }
  }
 
```
The resource should:
  1. Pass wafConfiguration.failureMode to the UpdateGateway / CreateGateway API
  2. Export web_acl_arn as a computed attribute (from API response)

### References

-  AWS CLI update-gateway docs (https://docs.aws.amazon.com/cli/latest/reference/bedrock-agentcore-control/update-gateway.html)

- AWS CLI create-gateway docs (https://docs.aws.amazon.com/cli/latest/reference/bedrock-agentcore-control/create-gateway.html)

- WafConfiguration API reference (https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_WafConfiguration.html)

No

arditti:
### Description

Adds a new resource `aws_bedrockagentcore_gateway_waf_configuration` to manage the AWS WAF configuration (`failure_mode`) for an Amazon Bedrock AgentCore Gateway.

This is modeled as a separate resource rather than an attribute on `aws_bedrockagentcore_gateway` because of how the AgentCore API works:

- `CreateGateway` does not accept `wafConfiguration` — only `UpdateGateway` does.
- `UpdateGateway` with `wafConfiguration` fails (`ValidationException: wafConfiguration can only be set when a WebACL is associated with the gateway`) unless a WAF web ACL is already associated with the gateway.
- WAF association is performed with `aws_wafv2_web_acl_association` using the gateway ARN as `resource_arn`.

Putting `waf_configuration` on the gateway resource itself would create a circular dependency (the gateway can't accept the config until the association — a separate resource pointing at the gateway — exists). A separate resource that `depends_on` the association breaks the cycle: `gateway → web_acl_association → gateway_waf_configuration`. This mirrors the satellite-resource pattern used by `aws_s3_bucket_versioning` and similar.

The resource performs a read-modify-write against `UpdateGateway` (which is a full replacement): it reads the current gateway, preserves all existing fields, and sets only `wafConfiguration`, so it does not clobber attributes managed by `aws_bedrockagentcore_gateway`. Delete clears `wafConfiguration`, tolerating the case where the web ACL association has already been torn down.

### Changes

- New resource `aws_bedrockagentcore_gateway_waf_configuration` with `gateway_identifier` (required, forces replacement), `failure_mode` (required, `FAIL_OPEN` / `FAIL_CLOSE`), and computed `web_acl_arn`.
- Acceptance test covering create (`FAIL_OPEN`), update (`FAIL_CLOSE`), import, and destroy.
- Website documentation.

### Relations

Closes #48661

### References

- https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_UpdateGateway.html
- https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_WafConfiguration.html
- https://docs.aws.amazon.com/waf/latest/APIReference/API_AssociateWebACL.html

```
$ TF_ACC=1 go test ./internal/service/bedrockagentcore/ -run '^TestAccBedrockAgentCoreGatewayWAFConfiguration_basic$' -v -timeout 60m

--- PASS: TestAccBedrockAgentCoreGatewayWAFConfiguration_basic (108.83s)
PASS
ok  	github.com/hashicorp/terraform-provider-aws/internal/service/bedrockagentcore	117.497s
```

### DIFF
--- a/.changelog/48710.txt
+++ b/.changelog/48710.txt
@@ -1,0 +1,3 @@
+```release-note:new-resource
+aws_bedrockagentcore_gateway_waf_configuration
+```
--- a/internal/service/bedrockagentcore/exports_test.go
+++ b/internal/service/bedrockagentcore/exports_test.go
@@ -13,6 +13,7 @@ var (
 	ResourceEvaluator                = newEvaluatorResource
 	ResourceGateway                  = newGatewayResource
 	ResourceGatewayTarget            = newGatewayTargetResource
+	ResourceGatewayWAFConfiguration  = newGatewayWAFConfigurationResource
 	ResourceMemory                   = newMemoryResource
 	ResourceResourcePolicy           = newResourcePolicyResource
 	ResourceMemoryStrategy           = newResourceMemoryStrategy
--- a/internal/service/bedrockagentcore/exports_test.go
+++ b/internal/service/bedrockagentcore/exports_test.go
@@ -15,6 +15,7 @@ var (
 	ResourceGateway                  = newGatewayResource
 	ResourceGatewayRule              = newGatewayRuleResource
 	ResourceGatewayTarget            = newGatewayTargetResource
+	ResourceGatewayWAFConfiguration  = newGatewayWAFConfigurationResource
 	ResourceMemory                   = newMemoryResource
 	ResourceResourcePolicy           = newResourcePolicyResource
 	ResourceMemoryStrategy           = newResourceMemoryStrategy
--- a/internal/service/bedrockagentcore/gateway.go
+++ b/internal/service/bedrockagentcore/gateway.go
@@ -411,7 +411,7 @@ func (r *gatewayResource) Update(ctx context.Context, request resource.UpdateReq
 			return
 		}
 
-		if _, err := waitGatewayUpdated(ctx, conn, gatewayID, r.UpdateTimeout(ctx, new.Timeouts)); err != nil {
+		if err := waitGatewayUpdated(ctx, conn, gatewayID, r.UpdateTimeout(ctx, new.Timeouts)); err != nil {
 			smerr.AddError(ctx, &response.Diagnostics, err, smerr.ID, gatewayID)
 			return
 		}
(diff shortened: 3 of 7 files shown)


aidandaly24:
Small one I noticed while exercising the resource against real AWS, not a correctness problem. The read-modify-write approach that preserves the other gateway fields is the right way to do it, but Update always calls `putWAFConfiguration`, which does a GetGateway -> full-replacement UpdateGateway -> waitGatewayUpdated round trip. The only mutable attributes here are `failure_mode` and the `timeouts` block (`gateway_identifier` is RequiresReplace, `web_acl_arn` is Computed), and `timeouts` is client-side only, so a plan that changes only the timeouts block still lands in Update and fires a real UpdateGateway plus waiter even though nothing on the server needs to change.

Repro:
1. Apply with `timeouts { update = "30m" }`.
2. Change it to `timeouts { update = "20m" }`, no change to `failure_mode`.
3. apply runs a full GetGateway/UpdateGateway/wait cycle on the gateway for a purely client-side edit, and briefly drives the whole gateway through UPDATING.

One way to avoid it is to short-circuit the WAF write when the server-relevant value is unchanged, e.g. compare planned vs state `failure_mode` before calling `putWAFConfiguration`:
```go
var plan, state gatewayWAFConfigurationResourceModel
// load both
if !plan.FailureMode.Equal(state.FailureMode) {
    if err := r.putWAFConfiguration(ctx, conn, gatewayID, wafConfiguration, r.UpdateTimeout(ctx, plan.Timeouts)); err != nil {
        // ...
    }
}
```
This is at gateway_waf_configuration.go:140 (Update) -> :154/:207 (putWAFConfiguration). Either way I think this is low and not blocking. Thanks!

arditti:
Good catch. Pushed 1a9b791: `Update` now loads both plan and state and only calls `putWAFConfiguration` when `failure_mode` actually changes (`!plan.FailureMode.Equal(state.FailureMode)`). Since `failure_mode` is the only server-relevant mutable attribute — `gateway_identifier` is RequiresReplace, `web_acl_arn` is Computed, `timeouts` is client-side only — a plan that edits only the timeouts block now short-circuits with no `GetGateway`/`UpdateGateway`/waiter cycle and no transition through UPDATING. On the skip path `web_acl_arn` is carried over from state. `go build` / `gofmt` / `go vet` pass.

arditti:
Rebased onto current `main` (d987ba0fd9); no content changes. Re-ran `go generate` for the package against the current generators and `service_package_gen.go` is byte-identical to what is committed. Build, vet, and test compile pass locally.

PR #48759, opened 2026-07-02
feat: add deletion_protection attribute to data source aws_rds_cluster

stefanfreitag:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to security controls are implemeted.

### Description

Add missing `deletion_protection` attribute to the data source [`aws_rds_cluster`](https://registry.terraform.io/providers/hashicorp/awS/latest/docs/data-sources/rds_cluster).

Updating the data source documentation is intentionally skipped as we have there

> See the [RDS Cluster Resource](https://registry.terraform.io/providers/hashicorp/awS/latest/docs/resources/rds_cluster) for details on the returned attributes - they are identical for all attributes, except the tags_all. If you need to get the tags for this resource, use the attribute tags as described below.

### Relations

Closes #48708

### Output from Acceptance Testing

```console
make testacc T=TestAccRDSClusterDataSource_ K=rds                  
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
make: Validating schemas
ok      github.com/hashicorp/terraform-provider-aws/internal/provider/sdkv2     0.308s
ok      github.com/hashicorp/terraform-provider-aws/internal/provider/framework 0.191s
make: Running acceptance tests on branch: 🌿 f-rds-cluster-datasource-deletion-protection 🌿...
TF_ACC=1 go1.26.4 test ./internal/service/rds/... -v -count 1 -parallel 20 -run='TestAccRDSClusterDataSource_'  -timeout 360m -vet=off -buildvcs=false
2026/07/02 19:51:38 Creating Terraform AWS Provider (SDKv2-style)...
2026/07/02 19:51:38 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccRDSClusterDataSource_basic
=== PAUSE TestAccRDSClusterDataSource_basic
=== RUN   TestAccRDSClusterDataSource_ManagedMasterPassword_managed
=== PAUSE TestAccRDSClusterDataSource_ManagedMasterPassword_managed
=== RUN   TestAccRDSClusterDataSource_deletionProtection
=== PAUSE TestAccRDSClusterDataSource_deletionProtection
=== CONT  TestAccRDSClusterDataSource_basic
=== CONT  TestAccRDSClusterDataSource_deletionProtection
=== CONT  TestAccRDSClusterDataSource_ManagedMasterPassword_managed
[...]

PR #48768, opened 2026-07-02
data-source/aws_rds_cluster: Add `deletion_protection` attribute

### LINKED ISSUE #48708: [Bug]: Data Source: aws_rds_cluster does not export deletion_protection

Observed with the current `hashicorp/aws` provider (the data source schema on `main` still omits the attribute — see below).

- `data "aws_rds_cluster"`

### Description

The documentation for the `aws_rds_cluster` **data source** states:

> This data source exports the following attributes in addition to the arguments above:
> See the RDS Cluster Resource for details on the returned attributes - they are identical for all attributes, except `tags_all`.

This is inaccurate. The data source does **not** expose `deletion_protection`, even though it is documented on the `aws_rds_cluster` resource. Referencing it fails at plan time:

```
Error: Unsupported attribute

  on main.tf line N, in locals:
   N:   x = data.aws_rds_cluster.example.deletion_protection

This object has no argument, nested block, or exported attribute named
"deletion_protection".
```

The data source's schema confirms the attribute is simply not declared (no `deletion_protection` key, and no corresponding `d.Set("deletion_protection", ...)` in the read function):

https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/rds/cluster_data_source.go

`deletion_protection` is an input-only argument on the resource, so it is never read back by the data source. This also appears to be the case for other resource arguments that are absent from the data source schema.

### Expected Behavior

Either:
1. The data source exports `deletion_protection` (its value is available from the RDS `DescribeDBClusters` API response field `DeletionProtection`), matching the documentation's "identical attributes" statement; or
2. The documentation is corrected to explicitly list the attributes the data source actually exports (and to note that input-only arguments such as `deletion_protection` are not returned).

### Actual Behavior

`data.aws_rds_cluster.<name>.deletion_protection` errors with `Unsupported attribute` because the attribute is not in the data source schema.

```hcl
data "aws_rds_cluster" "example" {
  cluster_identifier = "my-existing-cluster"
}

output "deletion_protection" {
  value = data.aws_rds_cluster.example.deletion_protection
}
```

### Steps to Reproduce

1. Point the config at any existing Aurora/RDS cluster.
2. `terraform plan`.

### References

- Data source docs (claim of identical attributes): https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/rds_cluster
- Data source schema (attribute absent): https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/rds/cluster_data_source.go
- The AWS API does return it: `DescribeDBClusters` -> `DBClusters[].DeletionProtection`.

ksudana:
### Description

Adds the `deletion_protection` attribute to the `aws_rds_cluster` data source. It was missing from the data source schema even though it is documented on the resource and the data source docs state the returned attributes are "identical" to the resource. The value comes from the RDS `DescribeDBClusters` API response (`DBClusters[].DeletionProtection`).

### Relations

Closes #48708

### References

- AWS API field: `DescribeDBClusters` -> `DBClusters[].DeletionProtection`

Read-only schema addition; covered by the existing `aws_rds_cluster` data source acceptance tests.

### DIFF
--- a/.changelog/48768.txt
+++ b/.changelog/48768.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+data-source/aws_rds_cluster: Add `deletion_protection` attribute
+```
--- a/internal/service/rds/cluster_data_source.go
+++ b/internal/service/rds/cluster_data_source.go
@@ -82,6 +82,10 @@ func dataSourceCluster() *schema.Resource {
 					Type:     schema.TypeString,
 					Computed: true,
 				},
+				"deletion_protection": {
+					Type:     schema.TypeBool,
+					Computed: true,
+				},
 				"enabled_cloudwatch_logs_exports": {
 					Type:     schema.TypeList,
 					Computed: true,
@@ -233,6 +237,7 @@ func dataSourceClusterRead(ctx context.Context, d *schema.ResourceData, meta any
 	d.Set("db_cluster_parameter_group_name", dbc.DBClusterParameterGroup)
 	d.Set("db_subnet_group_name", dbc.DBSubnetGroup)
 	d.Set("db_system_id", dbc.DBSystemId)
+	d.Set("deletion_protection", dbc.DeletionProtection)
 	d.Set("enabled_cloudwatch_logs_exports", dbc.EnabledCloudwatchLogsExports)
 	d.Set(names.AttrEndpoint, dbc.Endpoint)
 	d.Set(names.AttrEngine, dbc.Engine)


ksudana:
Closing as a duplicate of #48759, which adds the same `deletion_protection` attribute to the `aws_rds_cluster` data source and was opened earlier today. Thanks @<author of 48759>!

PR #46690, opened 2026-02-27
[Enhancement] aws_opensearch_package_association: Add import support

tabito-hara:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### Description

This PR implements an importer for the `aws_opensearch_package_association` resource.

* In the `Read` function, `domain_name` and `package_id` are used to identify the resource. Therefore, both must be specified when importing.
* A colon (`:`) is used as the separator between the two attributes, since `:` is not used in either argument.

  * Although `-` is used to construct the resource ID from `domain_name` and `package_id`, it cannot be used for importing because `-` may appear in the `domain_name` attribute. For this reason, a different separator is used for the import identifier.

### Relations

Closes #46666

### Output from Acceptance Testing

```console
$ make testacc TESTS='TestAccOpenSearchPackageAssociation_basic' PKG=opensearch
make: Verifying source code with gofmt...
==> Checking that code complies with gofmt requirements...
make: Running acceptance tests on branch: 🌿 f-aws_opensearch_package_association-support_import 🌿...
TF_ACC=1 go1.25.7 test ./internal/service/opensearch/... -v -count 1 -parallel 20 -run='TestAccOpenSearchPackageAssociation_basic'  -timeout 360m -vet=off
2026/02/28 02:12:56 Creating Terraform AWS Provider (SDKv2-style)...
2026/02/28 02:12:56 Initializing Terraform AWS Provider (SDKv2-style)...
=== RUN   TestAccOpenSearchPackageAssociation_basic
=== PAUSE TestAccOpenSearchPackageAssociation_basic
=== CONT  TestAccOpenSearchPackageAssociation_basic
--- PASS: TestAccOpenSearchPackageAssociation_basic (2159.43s)
PASS
ok      github.com/hashicorp/terraform-provider-aws/internal/service/opensearch 2163.815s

```

PR #48803, opened 2026-07-06
resource/aws_opensearch_package_association: Fix import

minzeng-clickup:
### Description

The `aws_opensearch_package_association` importer used `schema.ImportStatePassthroughContext`, which only sets the resource ID and never populates the required `domain_name` and `package_id` attributes. Because `Read` derives the lookup from `d.Get("domain_name")`/`d.Get("package_id")` (not from the ID), import always looked up an empty domain and failed. No import format worked, and there was no import documentation.

This PR replaces the passthrough importer with one that parses a `domain_name,package_id` import ID (comma is the standard `flex.ResourceIdSeparator`, unambiguous unlike the `-` used to build the resource ID), sets both attributes, and normalizes the stored ID to the same `domain_name-package_id` form `Create` produces — so imported and created resources are identical in state. Adds an import test step and an Import section to the docs.

### Relations

Closes #0000

### References

n/a

```
$ make testacc TESTS=TestAccOpenSearchPackageAssociation_basic PKG=opensearch
... (pending maintainer run)
```

### DIFF
--- a/.changelog/48803.txt
+++ b/.changelog/48803.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+resource/aws_opensearch_package_association: Fix `terraform import`
+```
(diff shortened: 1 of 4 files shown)


justinretzolk:
Hey @minzeng-clickup 👋🏻 Thank you for taking the time to raise this! It looks like this is a duplicate of #46690, which was opened a couple of months ago and is pending review/merge. With that in mind, I'm going to close this pull request in favor of that one.

